PR #6557, opened 2024-12-18
Add SmmCpuPlatformHookLib IsCpuSyncAlwaysNeeded interface

sweeaun:
# Description
Determines whether all CPUs sync is needed.

This function determine if all CPUs needed to be synced unconditional during the 1st APs sync.

By default, it returns FALSE which indicates not all CPUs needed to be synced. In case of all CPUs
needed to be synced, platform specific implementation can return TRUE.

## How This Was Tested

No test except compile the code successfully as new interface not being used by other code now. Test can be done once it called by other code later.

### DIFF
--- a/OvmfPkg/Library/SmmCpuPlatformHookLibQemu/SmmCpuPlatformHookLibQemu.c
+++ b/OvmfPkg/Library/SmmCpuPlatformHookLibQemu/SmmCpuPlatformHookLibQemu.c
@@ -131,3 +131,26 @@ SmmCpuPlatformHookBeforeMmiHandler (
 {
   return EFI_UNSUPPORTED;
 }
+
+/**
+  This function determines whether the first CPU Synchronization should be executed unconditionally
+  when a SMI occurs.
+
+  If the function returns true, it indicates that there is no need to check the system configuration
+  and status, and the first CPU Synchronization should be executed unconditionally.
+
+  If the function returns false, it indicates that the first CPU Synchronization is not executed
+  unconditionally, and the decision to synchronize should be based on the system configuration and status.
+
+  @retval TRUE   The first CPU Synchronization is executed unconditionally.
+  @retval FALSE  The first CPU Synchronization is not executed unconditionally.
+
+**/
+BOOLEAN
+EFIAPI
+IsCpuSyncAlwaysNeeded (
+  VOID
+  )
+{
+  return FALSE;
+}
--- a/UefiCpuPkg/Include/Library/SmmCpuPlatformHookLib.h
+++ b/UefiCpuPkg/Include/Library/SmmCpuPlatformHookLib.h
@@ -115,4 +115,24 @@ SmmCpuPlatformHookBeforeMmiHandler (
   VOID
   );
 
+/**
+  This function determines whether the first CPU Synchronization should be executed unconditionally
+  when a SMI occurs.
+
+  If the function returns true, it indicates that there is no need to check the system configuration
+  and status, and the first CPU Synchronization should be executed unconditionally.
+
+  If the function returns false, it indicates that the first CPU Synchronization is not executed
+  unconditionally, and the decision to synchronize should be based on the system configuration and status.
+
+  @retval TRUE   The first CPU Synchronization is executed unconditionally.
+  @retval FALSE  The first CPU Synchronization is not executed unconditionally.
+
+**/
+BOOLEAN
+EFIAPI
+IsCpuSyncAlwaysNeeded (
+  VOID
+  );
+
 #endif
--- a/UefiCpuPkg/Library/SmmCpuPlatformHookLibNull/SmmCpuPlatformHookLibNull.c
+++ b/UefiCpuPkg/Library/SmmCpuPlatformHookLibNull/SmmCpuPlatformHookLibNull.c
@@ -118,3 +118,26 @@ SmmCpuPlatformHookBeforeMmiHandler (
 {
   return EFI_UNSUPPORTED;
 }
+
+/**
+  This function determines whether the first CPU Synchronization should be executed unconditionally
+  when a SMI occurs.
+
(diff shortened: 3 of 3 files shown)


OvmfPkg/Library/SmmCpuPlatformHookLibQemu/SmmCpuPlatformHookLibQemu.c
```diff
@@ -131,3 +131,24 @@ SmmCpuPlatformHookBeforeMmiHandler (
...
+  if a platform would like to skip LMCE status check due to any reason, platform can
+  skip LMCE status check by return FALSE.
+
+  @retval TRUE     Platform LMCE status check is needed.
+  @retval FALSE    Platform LMCE status check is skipped.
+
+**/
+BOOLEAN
+EFIAPI
+IsPlatformLmceStatusCheck (
```
jiaxinwu, line 149: /**
  Determines if the LMCE status check is needed.

  This function checks whether the LMCE status check is required.
  By default, it returns TRUE, indicating that the LMCE status check
  should be performed. However, if the check should be skipped for
  any reason, it can return FALSE.

  @retval TRUE     LMCE status check is needed.
  @retval FALSE    LMCE status check is skipped.
**/
BOOLEAN
EFIAPI
IsLmceStatusCheckNeeded (
  VOID
  )

jiaxinwu:
please also refine the commit message according the new name: IsLmceStatusCheckNeeded

jiaxinwu:
use 2 commits in 1 PR for those 2 different packages changes

UefiCpuPkg/Include/Library/SmmCpuPlatformHookLib.h
```diff
@@ -115,4 +115,21 @@ SmmCpuPlatformHookBeforeMmiHandler (
...
+  if a platform would like to skip LMCE status check due to any reason, platform can
+  skip LMCE status check by return FALSE.
+
+  @retval TRUE     Platform LMCE status check is needed.
+  @retval FALSE    Platform LMCE status check is skipped.
+
+**/
+BOOLEAN
+EFIAPI
+IsPlatformLmceStatusCheck  (
```
jiaxinwu, line 132: /**
Determines if the LMCE status check is needed.

This function checks whether the LMCE status check is required.
By default, it returns TRUE, indicating that the LMCE status check
should be performed. However, if the check should be skipped for
any reason, it can return FALSE.

@RetVal TRUE LMCE status check is needed.
@RetVal FALSE LMCE status check is skipped.
**/
BOOLEAN
EFIAPI
IsLmceStatusCheckNeeded (
VOID
)

sweeaun: Updated the function name and comments

jiaxinwu (approved):

kraxel:
LMCE expands to "Local Machine Check Exception" I assume?

The commit messages describes what is changed, but not why this change is needed.  Can you please add that information?

sweeaun:
> LMCE expands to "Local Machine Check Exception" I assume?
> 
> The commit messages describes what is changed, but not why this change is needed. Can you please add that information?

Hi @kraxel,
Yup. LMCE here stands for Local Machine Check Exception.

This new interface will be consumed by UefiCpuPkg/PiSmmCpuDxeSmm/MpService.c SmmWaitForApArrival() later. However, the consumer part PR will be submitted later to avoid immediate downstream build error. This PR allow downstream to implement platform specific implementation, thus the transition will be smoother when downstream received the consumer part PR later.
I shall add this info into the commit message. Let me know if you have any more questions. Thanks.

kraxel (approved):

jiaxinwu (approved):

jiaxinwu (approved):

jiaxinwu (approved):

sweeaun:
Based on input from @niruiyu, renamed the interface to IsCpuSyncAlwaysNeeded which is more relevant and generic to the code consumer in MP service later.

UefiCpuPkg/Include/Library/SmmCpuPlatformHookLib.h
```diff
@@ -115,4 +115,22 @@ SmmCpuPlatformHookBeforeMmiHandler (
...
+  By default, it returns FALSE which indicates not all CPUs needed to be synced. In case of all CPUs
+  needed to be synced, platform specific implementation can return TRUE.
+
+  @retval TRUE   All CPUs sync is needed.
+  @retval FALSE  Not all CPUs sync is needed.
+
+**/
+BOOLEAN
+EFIAPI
+IsCpuSyncAlwaysNeeded (
```
jiaxinwu, line 132: This function determines whether CPU Synchronization should be executed unconditionally when a SMI occurs. 

If the function returns true, it indicates that there is no need to check the system configuration and status, and CPU Synchronization should be executed unconditionally.

If the function returns false, it indicates that CPU Synchronization is not executed unconditionally, and the decision to synchronize should be based on the system configuration and status.

@retval TRUE   CPU Synchronization is executed unconditionally.
@retval FALSE  CPU Synchronization is not executed unconditionally.

jiaxinwu: The commit message log:

This patch adds the IsCpuSyncAlwaysNeeded interface to the SmmCpuPlatformHookLib. This interface will determine whether the first CPU Synchronization should be executed unconditionally when an SMI occurs.
If the function returns true, it indicates that there is no need to check the system configuration and status, and the first CPU Synchronization should be executed unconditionally.
If the function returns false, it indicates that the first CPU Synchronization is not executed unconditionally, and the decision to synchronize should be based on the system configuration and status.

sweeaun: Updated

niruiyu (approved):

PR #6559, opened 2024-12-19
MdePkg: UefiDevicePathLib: Fix missing the Display Only format of Media Device Path SubType Hard Drive.

huangyq13:
```
HD(Partition,Type,Signature,Start, Size)
HD(Partition,Type,Signature) (Display Only)
```

Reference: [UEFI Specification Version 2.1 (Errata D) (released October  2008)](https://uefi.org/sites/default/files/resources/UEFI_Spec_2_1.pdf)

### DIFF
--- a/MdePkg/Library/UefiDevicePathLib/DevicePathToText.c
+++ b/MdePkg/Library/UefiDevicePathLib/DevicePathToText.c
@@ -1882,7 +1882,7 @@ DevPathToTextHardDrive (
     case SIGNATURE_TYPE_MBR:
       UefiDevicePathLibCatPrint (
         Str,
-        L"HD(%d,%s,0x%08x,",
+        L"HD(%d,%s,0x%08x",
         Hd->PartitionNumber,
         L"MBR",
         *((UINT32 *)(&(Hd->Signature[0])))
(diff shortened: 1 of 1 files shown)


lgao4:
Please aslo update DevPathFromTextHD ()

huangyq13:
Hi lgao4,

There is no need for updating the `DevPathFromTextHD()` since the string produced by `DevPathToTextHardDrive(XXX, TRUE, XXX)` is used for **Display Only**.

Also, the logic of the `DevPathFromTextHD()` is similar to the `DevPathToTextCDROM()` in the current code.

It will definitely return an instance with wrong data if you call the `ConvertTextToDevicePath()` using a Display Only instance as parameter, because the **Display Only** instance lacks information.

Check the current code below.
```
DevPathFromTextCDROM (
   CHAR16 *TextDeviceNode
  )
{
  // ...
  StartStr              = GetNextParamStr (&TextDeviceNode); // This returns a pointer to the "\0", if the TextDeviceNode is Display Only.
  SizeStr               = GetNextParamStr (&TextDeviceNode); // This returns a pointer to the "\0", if the TextDeviceNode is Display Only.
  // ...
  Strtoi64 (StartStr, &CDROMDevPath->PartitionStart); // Get a random PartitionStart value, because Strtoi64 returns an uninitialized local variable.
  Strtoi64 (SizeStr, &CDROMDevPath->PartitionSize); // Get a random PartitionSize, because Strtoi64 returns an uninitialized local variable.

  return (EFI_DEVICE_PATH_PROTOCOL *) CDROMDevPath;
}



```

lgao4 (approved):

huangyq13:
Hi there,

Can we merge this PR? @lgao4 @LiuZhiguang001 @mdkinney

PR #6561, opened 2024-12-19
BaseTools: Clean up os.path.normcase and os.path.normpath usage

YangGangUEFI:
Refer to the docs of python, `os.path.normcase(path)` function: "Normalize the case of a pathname. On Windows, convert all characters in the pathname to lowercase, and also convert forward slashes to backward slashes. On other operating systems, return the path unchanged."

`os.path.normpath(path)` also convert forward slashes to backward slashes.

So call `os.path.normcase` after `os.path.normpath` just convert path to lowercase on Windows(only).

And Windows is case-insensitive but case-preserving.

So the usage of `os.path.normcase(os.path.normpath(path))` can be simplified to `os.path.normpath(path)`. Then we can use case-preserving paths rather than lowercase paths in compile_commands.json file or build log.

But this patch continue to use `os.path.normcase`
when comparing/searching paths.

# Description

- [ ] Breaking change?
  - Not Select this but file path in generated .fd binary has been changed(In Windows, lowercase to uppercase). Do not known if it's "breaking".
    image

- [ ] Impacts security?
  - N/A
- [ ] Includes tests?
  - N/A

## How This Was Tested

Compare build log.

Or use `-y report.txt -Y COMPILE_INFO` and compare workspace path string in report.txt or CompileInfo/compile_commands.json file.

e.g.

In build log output:

before: `WORKSPACE        = d:\opensource\edk2`

after: `WORKSPACE        = D:\OpenSource\edk2`

D:\OpenSource is matched with real charactor case in FileExplorer/CMD...

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Source/Python/Common/MultipleWorkspace.py
+++ b/BaseTools/Source/Python/Common/MultipleWorkspace.py
@@ -34,7 +34,7 @@ class MultipleWorkspace(object):
     #
     @classmethod
     def convertPackagePath(cls, Ws, Path):
-        if str(os.path.normcase (Path)).startswith(Ws):
+        if str(os.path.normcase (os.path.normpath(Path))).startswith(os.path.normcase(os.path.normpath(Ws))):
             return os.path.join(Ws, os.path.relpath(Path, Ws))
         return Path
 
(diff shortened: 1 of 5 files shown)


YangGangUEFI:
Not sure if this is acceptable. But please give some comments or discussion.🤝

lgao4:
I agree to keep original path instead of always lower string. So, I am OK for this change.

lgao4 (approved):

lgao4:
@YangGangUEFI , please rebase it by hand.

YangGangUEFI:
> @YangGangUEFI , please rebase it by hand.

Done. @lgao4

PR #6396, opened 2024-11-01
ArmPkg/GenericWatchdogDxe: Removed fixed PCDs

sureshkumarpMSFT:
Removed the "fixed" attribute for the PCD PcdGenericWatchdogEl2IntrNum.
 This enables platforms to modify it dynamically.

# Description

This change is made to enable the platform code to modify the PcdGenericWatchdogEl2IntrNum dynamically.

<_For each item, place an "x" in between `[` and `]` if true. Example: `[x]` (you can also check items in GitHub UI)_>

<_Create the PR as a Draft PR if it is only created to run CI checks._>

<_Delete lines in \<\> tags before creating the PR._>

- [ ] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [ ] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

## How This Was Tested

Verified the build with this change

## Integration Instructions

N/A

PR #6562, opened 2024-12-19
ArmPkg: Update Generic Watchdog PCD Type

ashishsingha:
# Description
Do not use the PCDs as only fixed when there is an option to use them as dynamic that can be set per SKU.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

SOCs with different addresses and INTR numbers governed via SKU were booted with this change.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/GenericWatchdogDxe/GenericWatchdog.h
+++ b/ArmPkg/Drivers/GenericWatchdogDxe/GenericWatchdog.h
@@ -14,15 +14,15 @@
 #define GENERIC_WATCHDOG_H_
 
 // Refresh Frame:
-#define GENERIC_WDOG_REFRESH_REG  ((UINTN)FixedPcdGet64 (PcdGenericWatchdogRefreshBase) + 0x000)
+#define GENERIC_WDOG_REFRESH_REG  ((UINTN)PcdGet64 (PcdGenericWatchdogRefreshBase) + 0x000)
 
 // Control Frame:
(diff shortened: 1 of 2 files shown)


ashishsingha:
@leiflindholm @ardbiesheuvel @samimujawar This PR can also act as a superset on top of the pending PR https://github.com/tianocore/edk2/pull/6396. Please let me know if you have any concerns.

samimujawar (approved):
@ashishsingha this patch looks good to me. 

I would be curious to know how you intend to use the PCDs as dynamic PCDs. Is there a library/module that will initialise the PCDs?

ashishsingha:
@samimujawar, the DSC file can have SKUs defined, and then you can have Dynamic and HII PCDs per SKU. Very early during boot, the SKU can be set, and PCDs for that SKU will be returned upon query. I see you have approved the change. Can you please see what is needed next to get it merged?

samimujawar:
@ashishsingha thanks for the explanation. 
I will wait for other maintainers to respond as well.

leiflindholm:
> @ashishsingha thanks for the explanation. I will wait for other maintainers to respond as well.

I'm all for it.
Ultimately, I would prefer a higher-level standardised discovery interface across all vendors rather than the platform-specific SKU swizzling, but both require this to move to non-fixed.

leiflindholm (approved):

ashishsingha:
Thanks @leiflindholm. Who should I be requesting to merge the change now that we’ve received 2 approvals?

PR #8637, opened 2024-12-20
Update pipeline filenames

makubacki:
# Description

Details about a particular pipeline change over time - such as the specific Linux distro or toolchain version. At the same time, the pipeline files are directly referenced in Azure Pipelines to run the pipeline and few users have access to change those details.

To make the files less likely to refer to outdated information over time and reduce the need for an Azure Pipeline administrator to get involved, the files are updated to drop specific details apart from the host OS environment used in the pipeline.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Azure pipelines

## Integration Instructions

- N/A


Javagedes (approved):

makubacki:
@mdkinney, I can do the work needed in Azure Pipelines, but do you have a preference?

1. Update the existing pipelines to the new filename?
2. Create new pipelines for the new files and deprecate the old pipelines?

(2) is more work but can make the transition cleaner by keeping the previously named status checks associated with the pipelines as they were when those checks were run. In either case, the pipelines moving forward will use new (more accurate) names that will be reflected in the GitHub status checks.

mdkinney:
If we want to add CLANG support to Windows and Linux builds, would that be added in these file.  The tool chain name was removed.  Should we keep the tool chain family in the file names?

mdkinney:
> @mdkinney, I can do the work needed in Azure Pipelines, but do you have a preference?
> 
> 1. Update the existing pipelines to the new filename?
> 2. Create new pipelines for the new files and deprecate the old pipelines?
> 
> (2) is more work but can make the transition cleaner by keeping the previously named status checks associated with the pipelines as they were when those checks were run. In either case, the pipelines moving forward will use new (more accurate) names that will be reflected in the GitHub status checks.

Changing the names of the status checks presented by Azure Pipelines will require a lot of coordination because those check names have to be configured in the GitHub branch protection rules.

makubacki:
> If we want to add CLANG support to Windows and Linux builds, would that be added in these file. The tool chain name was removed. Should we keep the tool chain family in the file names?

Technically, the Clang and non-Clang toolchain could run in parallel and have their individual jobs reported (like they are today from those generated in the matrix).

However, I was already considering including the toolchain name (i.e. `Windows-VS.yml`). In this case, I think that would clean up the reporting and artifacts in the pipeline.

makubacki:
> Changing the names of the status checks presented by Azure Pipelines will require a lot of coordination because those check names have to be configured in the GitHub branch protection rules.

Right now, the required status checks are tied to the toolchain:

image

As part of this transition, at a minimum, the pipeline names will change (either in the existing pipelines or in newly added pipelines). Once they are agnostic tool specific toolchains, they, like the filenames, won't be as brittle to change in the future.

PR #10558, opened 2024-12-20
MdeModulePkg|UsbBusDxe: support USB Interface Association

olegilyasov:
# Description
USB Interface Association is a group of UsbIo that implement a USB function. UEFI device driver manages multiple UsbIo instances. Examples of such devices are: USB camera, USB serial, USB network, etc.

Current approach for supporting these devices is to respond on UsbIo installation and analyze if the current UsbIo belongs to the USB association. This algorithm is based on assumptions that may not be correct for different device configurations. Having USB association protocol that reports its associates (UsbIo) simplifies the USB device driver.

For the USB configurations that implement USB association the UsbDxeBus driver:
- creates USB association device
- installs device path
- installs USB association IO protocol

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions

### DIFF
--- a/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbBus.h
+++ b/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbBus.h
@@ -3,6 +3,7 @@
     Usb Bus Driver Binding and Bus IO Protocol.
 
 Copyright (c) 2004 - 2018, Intel Corporation. All rights reserved.<BR>
+Copyright (c) 2025, American Megatrends International, LLC. All rights reserved.<BR>
 SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
@@ -15,6 +16,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Protocol/Usb2HostController.h>
 #include <Protocol/UsbHostController.h>
 #include <Protocol/UsbIo.h>
+#include <Protocol/UsbAssociationIo.h>
 #include <Protocol/DevicePath.h>
 
 #include <Library/BaseLib.h>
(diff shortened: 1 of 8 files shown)


MdeModulePkg/Bus/Usb/UsbBusDxe/UsbBus.h
```diff
@@ -752,6 +779,7 @@ UsbBusControllerDriverStop (
   );
 
 extern EFI_USB_IO_PROTOCOL           mUsbIoProtocol;
+extern EDKII_USB_IA_PROTOCOL         mUsbIaProtocol;
```
mdkinney, line 782: Expand "IA" to full words.

MdeModulePkg/Include/Protocol/UsbAssociationIo.h
```diff
@@ -0,0 +1,81 @@
...
+  Get the USB Interface Association Descriptor from the current USB configuration.
+
+  @param[in]  This              A pointer to the EFI_USB_IA_PROTOCOL instance.
+  @param[out]  Descriptor       A pointer to the caller allocated USB Interface Association Descriptor.
+
+  @retval EFI_SUCCESS           The descriptor was retrieved successfully.
+  @retval EFI_INVALID_PARAMETER Descriptor is NULL.
+
+**/
+typedef
```
mdkinney, line 41: Can these be library services instead of a new protocol?  If it is a new protocol, then USB Device Drivers will depends on EDK II specific protocols instead of just UEFI Protocols.

olegilyasov: In this PR we create USB Association handle and install DP and Association protocols on it. This handle is essential If USB Device Driver follows UEFI driver model, it can not be implemented using the library.
If this EDKII protocol is successful, it can be adopted by UEFI specification.
USB Association is a part of USB specification and this PR is an attempt to implement its functionality.
If there is any other way of advertising the new protocol, please advise.

mdkinney: If the intent is to add to UEFI Spec, then there is a process for that and the doc and code updates would be done in edk2-staging repo.

https://github.com/tianocore/tianocore.github.io/wiki/EDK-II-Code-First-Process

olegilyasov: Thank you, will move it to edk2-staging. The document above talks about creating a bugzilla issue, I understand bugzilla function has been replaced with github issue. Should the issue be created in edk2 repo or edk2-staging repo?

mdkinney: good question.  Please create in edk2 repo because that is the location the feature will be added once it is available in published spec.

olegilyasov: https://edk2.groups.io/g/devel/topic/staging_usb_iad/110414302 has a patch to create a edk2-staging/usb_iad branch.

PR #10560, opened 2024-12-23
IntelFsp2WrapperPkg: Save FspHobListPtr Right After FspMemoryInit Exits

dulin0:
# Description

Save FspHobList pointer to HOB right after FspMemoryInit exits so that FspHobList pointer is available when performing platform related reset in CallFspWrapperResetSystem(). Some platforms may consume FSP HOBs prior to performing platform related reset.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This PR was tested with FSP-M, FSP-S, FSP multi-phase memory initialization and FSP multi-phase silicon initialization enabled. And it's confirmed in the test that the bootloader can successfully get and save the FspHobList pointer in different phases.

## Integration Instructions
N/A

### DIFF
--- a/IntelFsp2WrapperPkg/Library/FspWrapperMultiPhaseProcessLib/PeiFspWrapperMultiPhaseProcessLib.c
+++ b/IntelFsp2WrapperPkg/Library/FspWrapperMultiPhaseProcessLib/PeiFspWrapperMultiPhaseProcessLib.c
@@ -17,6 +17,8 @@
 #include <Ppi/Variable.h>
 #include <Library/PeiServicesLib.h>
 #include <Library/FspWrapperPlatformMultiPhaseLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/HobLib.h>
 
 /**
   Execute 32-bit FSP API entry code.
@@ -156,6 +158,8 @@ FspWrapperVariableRequestHandler (
   EDKII_PEI_VARIABLE_PPI                            *VariablePpi;
   BOOLEAN                                           WriteVariableSupport;
   FSP_MULTI_PHASE_COMPLETE_VARIABLE_REQUEST_PARAMS  CompleteVariableRequestParams;
+  VOID                                              *GuidHob;
+  VOID                                              *HobData;
 
   WriteVariableSupport = TRUE;
   Status               = PeiServicesLocatePpi (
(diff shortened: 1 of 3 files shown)


IntelFsp2WrapperPkg/Library/FspWrapperMultiPhaseProcessLib/PeiFspWrapperMultiPhaseProcessLib.c
```diff
@@ -352,6 +356,15 @@ FspWrapperMultiPhaseHandler (
     FspMultiPhaseParams.MultiPhaseParamPtr = NULL;
     Status                                 = CallFspMultiPhaseEntry (&FspMultiPhaseParams, FspHobListPtr, ComponentIndex);
 
+    //
+    // Refresh FspHobList pointer stored in HOB.
```
ChaselChiu, line 360: When the platforms only support VariableRequest, we will need similar code to refresh FspHobList in FspWrapperVariableRequestHandler() too. (to refresh the pointer after FspHobList point migrated to permanent memory)

dulin0: Done. Thanks for pointing this out.

IntelFsp2WrapperPkg/Library/FspWrapperMultiPhaseProcessLib/PeiFspWrapperMultiPhaseProcessLib.c
```diff
@@ -288,6 +292,15 @@ FspWrapperVariableRequestHandler (
     }
   }
 
+  //
+  // Refresh FspHobList pointer stored in HOB.
+  //
+  GuidHob = GetFirstGuidHob (&gFspHobGuid);
+  if (GuidHob != NULL) {
```
niruiyu, line 299: can we add an assertion ASSERT (GuidHob != NULL)?
If-check is ok to be kept.

dulin0: Done. Thanks for the comment.

niruiyu (approved):

AshrafAliS (approved):

PR #10570, opened 2024-12-26
Upl

sjg20:
# Description

Changes to bring UPL in line with the spec and make it work with U-Boot

- [X] Breaking change?
- [ ] Includes tests?

## How This Was Tested

Using U-Boot branch 'upl' at sjg.u-boot.org

./scripts/build-qemu.sh -a x86 -w -u ~dev/edk2/Build/UefiPayloadPkgIA32/UniversalPayload.fit -rs

## Integration Instructions

N/A


UefiPayloadPkg/Tools/MkFitImage.py
```diff
@@ -54,6 +54,7 @@ def BuildConfNode(Fdt, ParentNode, MultiImage):
 
     Fdt.setprop(ConfNode1, 'require-fit', b'', 0)
     Fdt.setprop_str(ConfNode1, 'firmware', 'tianocore')
+    Fdt.setprop_str(ConfNode1, 'loadables', 'uefi-fv\0bds-fv')
```
benjamindoron, line 57: This is a good idea, but we should do it properly by iterating `MultiImage` in `BuildFitImage()` below

UefiPayloadPkg/UniversalPayloadBuild.py
```diff
@@ -130,10 +130,10 @@ def BuildUniversalPayload(Args):
     BuildDir     = os.path.join(os.environ['WORKSPACE'], os.path.normpath("Build/UefiPayloadPkg{}").format (Args.Arch))
     if Args.Arch == 'X64':
         BuildArch      = "X64"
-        FitArch        = "x86_64"
+        FitArch        = "i386"
```
benjamindoron, line 136: No, this is a coreboot-specific point. *coreboot* starts its payloads in 32-bit on x86, without condition on what coreboot is compiled as. So, coreboot will build UPL passing in `-a IA32`. EDK2-based Platform Init might call UPL already in long mode, or it might do a jump from their 32-bit PEI, to our 32-bit SEC, and let us do the mode switch.

U-Boot should build with `-a IA32` or `-a X64` depending on the bitness of the instructions right before you jump.

UefiPayloadPkg/Tools/MkFitImage.py
```diff
@@ -63,6 +63,7 @@ def BuildFvImageNode(Fdt, InfoHeader, ParentNode, DataOffset, DataSize, Descript
     Fdt.setprop_str(ParentNode, 'project', 'tianocore')
     Fdt.setprop_str(ParentNode, 'arch',Arch)
     Fdt.setprop_str(ParentNode, 'type', 'firmware')
+    Fdt.setprop_u64(ParentNode, 'load', InfoHeader.LoadAddr + DataOffset - InfoHeader.DataOffset)
```
benjamindoron, line 69: FVs aren't place-specific. Do we need a `load` property if we don't care?

UefiPayloadPkg/Tools/MkFitImage.py
benjamindoron, line 1: This is going to break PayloadLoaderPeim and coreboot. If the changes have to be made, I'll likely be making coreboot's, but please sync this with FitPayloadLoaderPeim. This is also getting hard to read, (well, harder, after what I did to `load`) and I'm not sure if we can avoid that

UefiPayloadPkg/Tools/MkFitImage.py
```diff
@@ -61,8 +61,11 @@ def BuildFvImageNode(Fdt, InfoHeader, ParentNode, DataOffset, DataSize, Descript
     Fdt.setprop_u32(ParentNode, 'data-offset', DataOffset)
     Fdt.setprop_str(ParentNode, 'compression', 'none')
     Fdt.setprop_str(ParentNode, 'project', 'tianocore')
-    Fdt.setprop_str(ParentNode, 'arch',Arch)
+    if InfoHeader.Arch is not None:
```
benjamindoron, line 64: Can this or the below ever be None? Also, arch's value is the same, as this is what the caller passed in

PR #10571, opened 2024-12-27
UefiPayloadPkg: Minor fixes

sjg20:
# Description

Here are a few fixes for Universal Payload (UPL) which are applicable to master, to help it work with U-Boot

These are brought in from my main PR[1], as that is still dependent on Benjamin Doron's work and cannot be reviewed yet.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Using U-Boot branch 'upl' at sjg.u-boot.org

./scripts/build-qemu.sh -a x86 -w -u ~dev/edk2/Build/UefiPayloadPkgIA32/UniversalPayload.fit -rs

## Integration Instructions

N/A

[1] https://github.com/tianocore/edk2/pull/10570

### DIFF
--- a/UefiPayloadPkg/UniversalPayloadBuild.py
+++ b/UefiPayloadPkg/UniversalPayloadBuild.py
@@ -130,10 +130,10 @@ def BuildUniversalPayload(Args):
     BuildDir     = os.path.join(os.environ['WORKSPACE'], os.path.normpath("Build/UefiPayloadPkg{}").format (Args.Arch))
     if Args.Arch == 'X64':
         BuildArch      = "X64"
-        FitArch        = "x86_64"
+        FitArch        = "i386"
     elif Args.Arch == 'IA32':
         BuildArch      = "IA32 -a X64"
-        FitArch        = "x86"
+        FitArch        = "i386"
     elif Args.Arch == 'RISCV64':
         BuildArch      = "RISCV64"
         FitArch        = "RISCV64"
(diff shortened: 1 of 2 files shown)


UefiPayloadPkg/UniversalPayloadBuild.py
```diff
@@ -130,10 +130,10 @@ def BuildUniversalPayload(Args):
     BuildDir     = os.path.join(os.environ['WORKSPACE'], os.path.normpath("Build/UefiPayloadPkg{}").format (Args.Arch))
     if Args.Arch == 'X64':
         BuildArch      = "X64"
-        FitArch        = "x86_64"
+        FitArch        = "i386"
```
benjamindoron, line 133: As I said on the other change, this shouldn't be here. coreboot will call payloads in 32-bit mode no matter what, so coreboot will always pass `-a IA32` to the script

sjg20: UPL needs to work with other projects, though. The 'arch' property in UPL images indicates which architecture the image was built for. So if we have x86_64 here, U-Boot will happily boot it in 64-bit mode, thus causing a crash.

lgao4:
The first commit is the change in MdePkg. The commit message should be MdePkg: xxxx.

UefiPayloadPkg/UniversalPayloadBuild.py
benjamindoron: I don't understand how you're building UPL then. If you pass `-a X64` to the script, then the entrypoint image is a PE32+, which you should be calling in 64-bit mode. Therefore, it should say x86_64 here. If you pass `-a IA32`, the entrypoint is a PE32, which should be called in 32-bit mode. Of course, the FVs should always be x86_64 images (assuming x86). This still needs to be fixed, I see "IA32" and "X64" when I build UPL.

PR #10577, opened 2025-01-01
DynamicTablesPkg: Adds X64 FACS generator library

abdattar:
Adds a library to generate ACPI FACS table for X64.

Cc: Sami Mujawar <Sami.Mujawar@arm.com>
Cc: Pierre Gondois <pierre.gondois@arm.com>

# Description

Adds X64 generator library for creating ACPI FACS table.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Tested on AMD platform.

## Integration Instructions
N/A

### DIFF
--- a/DynamicTablesPkg/Include/X64NameSpaceObjects.h
+++ b/DynamicTablesPkg/Include/X64NameSpaceObjects.h
@@ -2,7 +2,7 @@
 
   Defines the X64 Namespace Object.
 
-  Copyright (C) 2024 Advanced Micro Devices, Inc. All rights reserved.
+  Copyright (C) 2024 - 2025 Advanced Micro Devices, Inc. All rights reserved.
 
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
@@ -50,7 +50,8 @@ typedef enum X64ObjectID {
   EX64ObjIoApicInfo,             ///< 14 - IO APIC info
   EX64ObjIntrSourceOverrideInfo, ///< 15 - Interrupt Source Override info
   EX64ObjLocalApicX2ApicNmiInfo, ///< 16 - Local APIC and X2APIC NMI info
-  EX64ObjMax                     ///< 17 - Maximum Object ID
+  EX64ObjFacsInfo,               ///< 17 - FACS info
+  EX64ObjMax                     ///< 18 - Maximum Object ID
 } EX64_OBJECT_ID;
 
 /** A structure that describes the
@@ -264,6 +265,18 @@ typedef struct CmX64LocalApicX2ApicNmiInfo {
   UINT32    AcpiProcessorUid;
   UINT8     LocalApicLint;
 } CM_X64_LOCAL_APIC_X2APIC_NMI_INFO;
+
+/**
+  A structure that describes the FACS information.
+
+  ID: EX64ObjFacsInfo
+*/
+typedef struct CmX64FacsInfo {
+  UINT32    FirmwareWakingVector;
+  UINT32    Flags;
+  UINT64    XFirmwareWakingVector;
+  UINT32    OspmFlags;
+} CM_X64_FACS_INFO;
 
 #pragma pack()
 #endif // X64_NAMESPACE_OBJECTS_H_
--- a/DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
+++ b/DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
@@ -1,0 +1,465 @@
+/** @file
+
+  Generate ACPI FACS table for AMD platforms.
+
+  Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+
+  SPDX-License-Identifier BSD-2-Clause-Patent
+**/
+
+#include <AcpiTableGenerator.h>
+#include <ConfigurationManagerHelper.h>
+#include <ConfigurationManagerObject.h>
+#include <Library/DebugLib.h>
+#include <Library/TableHelperLib.h>
+#include <Protocol/AcpiTable.h>
+#include <Protocol/ConfigurationManagerProtocol.h>
+#include <X64NameSpaceObjects.h>
+#include <Library/UefiBootServicesTableLib.h>
+#include <Library/UefiLib.h>
+#include <Library/BaseMemoryLib.h>
+
+/** This macro expands to a function that retrieves the
+    FACS information from the Configuration Manager.
+*/
+GET_OBJECT_LIST (
+  EObjNameSpaceX64,
+  EX64ObjFacsInfo,
+  CM_X64_FACS_INFO
+  );
+
+/** The ACPI FACS Table.
+*/
+STATIC
+EFI_ACPI_6_5_FIRMWARE_ACPI_CONTROL_STRUCTURE  AcpiFacs = {
+  EFI_ACPI_6_5_FIRMWARE_ACPI_CONTROL_STRUCTURE_SIGNATURE,
+  // Length
+  sizeof (EFI_ACPI_6_5_FIRMWARE_ACPI_CONTROL_STRUCTURE),
+  // Hardware Signature
+  0,
+  // Firmware Waking Vector
+  0,
+  // Global Lock
+  0,
+  // Flags
+  0,
+  // XFirmware Waking Vector
+  0,
+  // Version
+  0,
+  // Reserved0
+  { 0,                                                   0,  0 },
+  // OSPM Flags
+  0,
+  // Reserved1
+  {
+    0,                                                   0,  0, 0, 0, 0, 0, 0,
+    0,                                                   0,  0, 0, 0, 0, 0, 0,
+    0,                                                   0,  0, 0, 0, 0, 0, 0
+  }
+};
+
+/**
+  Update the hardware signature in the FACS table.
+**/
+VOID
+AcpiFacsUpdateHardwareSignature (
+  VOID
+  )
+{
+  EFI_ACPI_6_5_FIRMWARE_ACPI_CONTROL_STRUCTURE  *Facs;
+  EFI_ACPI_6_5_FIXED_ACPI_DESCRIPTION_TABLE     *Fadt;
+  EFI_ACPI_6_5_ROOT_SYSTEM_DESCRIPTION_POINTER  *Rsdp;
+  EFI_ACPI_DESCRIPTION_HEADER                   *Table;
+  EFI_ACPI_DESCRIPTION_HEADER                   *Xsdt;
+  EFI_ACPI_DESCRIPTION_HEADER                   *Dsdt;
+  EFI_STATUS                                    Status;
+  UINT32                                        CollectedCrc[2];
+  UINT32                                        ComputedCrc;
+  UINT32                                        TableCount;
+  UINT64                                        XsdtTablePtr;
+  UINTN                                         Index;
+  UINTN                                         XsdtPtr;
+
+  DEBUG ((DEBUG_INFO, "Updating hardware signature in FACS Table.\n"));
+
+  Facs   = NULL;
+  Fadt   = NULL;
+  Dsdt   = NULL;
+  Status = EfiGetSystemConfigurationTable (&gEfiAcpiTableGuid, (VOID **)&Rsdp);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "Failed to get RSDP. Status(%r)\n", Status));
+    ASSERT_EFI_ERROR (Status);
+    return;
+  }
+
+  if (  (Rsdp->Revision >= EFI_ACPI_2_0_ROOT_SYSTEM_DESCRIPTION_POINTER_REVISION)
+     && (Rsdp->XsdtAddress != 0))
+  {
+    CollectedCrc[0] = Rsdp->ExtendedChecksum;
+    Xsdt            = (EFI_ACPI_DESCRIPTION_HEADER *)(UINTN)Rsdp->XsdtAddress;
+    CollectedCrc[1] = Xsdt->Checksum;
+    gBS->CalculateCrc32 (
+           (UINT8 *)CollectedCrc,
+           ARRAY_SIZE (CollectedCrc),
+           &ComputedCrc
+           );
+    CollectedCrc[0] = ComputedCrc;
+
+    TableCount = (Xsdt->Length - sizeof (EFI_ACPI_DESCRIPTION_HEADER)) / sizeof (UINT64);
+    XsdtPtr    = (UINTN)(Xsdt + 1);
+    for (Index = 0; Index < TableCount; Index++) {
+      CopyMem (
+        &XsdtTablePtr,
+        (VOID *)(XsdtPtr + Index * sizeof (UINT64)),
+        sizeof (UINT64)
+        );
+      Table = (EFI_ACPI_DESCRIPTION_HEADER *)((UINTN)(XsdtTablePtr));
+      DEBUG ((
+        DEBUG_INFO,
+        "FACS: Collecting CRC of %c%c%c%c\n",
+        Table->Signature & 0xFF,
+        (Table->Signature >> 8) & 0xFF,
+        (Table->Signature >> 16) & 0xFF,
+        (Table->Signature >> 24) & 0xFF
+        ));
+      CollectedCrc[1] = Table->Checksum;
+      gBS->CalculateCrc32 (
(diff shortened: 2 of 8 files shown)


DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
```diff
@@ -0,0 +1,430 @@
...
+             );
+      CollectedCrc[0] = ComputedCrc;
+    }
+  }
+
+  Facs = (EFI_ACPI_6_5_FIRMWARE_ACPI_CONTROL_STRUCTURE *)EfiLocateFirstAcpiTable (EFI_ACPI_6_5_FIRMWARE_ACPI_CONTROL_STRUCTURE_SIGNATURE);
+  if (Facs != NULL) {
+    /// Update FACS signature
+    Facs->HardwareSignature = ComputedCrc;
+  }
```
pierregondois, line 156: Maybe we should print an error message if there is no FACS table installed.

DynamicTablesPkg/Include/X64NameSpaceObjects.h
```diff
@@ -265,5 +266,19 @@ typedef struct CmX64LocalApicX2ApicNmiInfo {
...
 
+/**
+  A structure that describes the FACS information.
+
+  ID: EX64ObjFacsInfo
+*/
+typedef struct CmX64FacsInfo {
+  UINT32    HardwareSignature;
+  UINT32    FirmwareWakingVector;
+  UINT32    GlobalLock;
```
pierregondois, line 277: The global lock doesn't seem to be data or a pointer to a memory location. It seems that it is instead an area reserved to implement the global lock itself. Thus I don't think this field is necessary.

DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
```diff
@@ -0,0 +1,430 @@
...
+      "ERROR: FACS: Failed to update table info. Status = %r\n",
+      Status
+      ));
+    return Status;
+  }
+
+  //
+  // Register notify function
+  //
+  Status = gBS->CreateEventEx (
```
pierregondois, line 327: @samimujawar 
What do you think about this ? The FACS table needs to have a `Hardware Signature` field allowing to identify if the ACPI tables have been modified while the firmware operated some modifications. If the ACPI tables were modified and the `Hardware Signature` has been modified, the OS should reboot.
The `Hardware Signature` used here is a CRC of all the ACPI tables.

An alternative would be to build/install the FACS last in:
DynamicTablesPkg/Drivers/DynamicTableManagerDxe/DynamicTableManagerDxe.c::ProcessAcpiTables()

DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
```diff
@@ -0,0 +1,430 @@
...
+
+  CollectedCrc[0] = Rsdp->Checksum;
+  CollectedCrc[1] = Rsdp->ExtendedChecksum;
+  gBS->CalculateCrc32 (
+         (UINT8 *)CollectedCrc,
+         ARRAY_SIZE (CollectedCrc),
+         &ComputedCrc
+         );
+  CollectedCrc[0] = ComputedCrc;
+
```
pierregondois, line 101: I think there is also the FADT table also contains pointers to other ACPI tables: DSDT, FACS. Their CRC should also be taken into account and the `Hardware Signature` field of the FACS table should be ignored when computing the CRC.

DynamicTablesPkg/Include/X64NameSpaceObjects.h
```diff
@@ -265,5 +266,19 @@ typedef struct CmX64LocalApicX2ApicNmiInfo {
   UINT8     LocalApicLint;
 } CM_X64_LOCAL_APIC_X2APIC_NMI_INFO;
 
+/**
+  A structure that describes the FACS information.
+
+  ID: EX64ObjFacsInfo
+*/
+typedef struct CmX64FacsInfo {
+  UINT32    HardwareSignature;
```
pierregondois, line 275: If the `HardwareSignature` field is dynamically populated, it should not be possible to provided a static value I think.

DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
```diff
@@ -0,0 +1,430 @@
...
+    0,                                                   0,  0, 0, 0, 0, 0, 0,
+    0,                                                   0,  0, 0, 0, 0, 0, 0
+  }
+};
+
+/**
+  Update the hardware signature in the FACS table.
+**/
+VOID
+AcpiFacsUpdateHardwareSignature (
```
pierregondois, line 66: FYIU, the `HardwareSignature` field allows to OS to check if the firmware modified the ACPI tables while the firmware was holding the global lock. However, I didn't find any example where the global lock is actually taken/released in edk2/edk2-platforms [1]. Thus I am wondering:
- Is it actually useful to compute the `HardwareSignature` as edk2 is not able to modify the signature anyway.
- If a module is modifying the `HardwareSignature` of the FACS table at runtime, then it is necessary to expose this function. Otherwise the code modifying the ACPI tables at runtime has no mean to indicate that the ACPI tables have changed.

[1] Cf. the suggested implementation to acquire/release the global lock in the ACPI spec at `5.2.10.1 Global Lock`.

DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
```diff
@@ -0,0 +1,430 @@
...
+  DEBUG ((DEBUG_INFO, "Updating hardware signature in FACS Table.\n"));
+
+  Status = EfiGetSystemConfigurationTable (&gEfiAcpiTableGuid, (VOID **)&Rsdp);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "Failed to get RSDP. Status(%r)\n", Status));
+    ASSERT_EFI_ERROR (Status);
+    return;
+  }
+
+  CollectedCrc[0] = Rsdp->Checksum;
```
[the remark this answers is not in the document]
abdattar: Ack. will remove unnecessary code.

DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
```diff
@@ -0,0 +1,430 @@
...
+  } else if (Rsdp->RsdtAddress != 0) {
+    Rsdt = (EFI_ACPI_DESCRIPTION_HEADER *)(UINTN)Rsdp->RsdtAddress;
+    /// compute the CRC
+    CollectedCrc[1] = Rsdt->Checksum;
+    gBS->CalculateCrc32 ((UINT8 *)Rsdt, Rsdt->Length, &ComputedCrc);
+    CollectedCrc[0] = ComputedCrc;
+
+    TableCount = (Rsdt->Length - sizeof (EFI_ACPI_DESCRIPTION_HEADER)) / sizeof (UINT32);
+    RsdtPtr32  = (UINT32 *)(Rsdt + 1);
+    for (Index = 0; Index < TableCount; Index++, RsdtPtr32++) {
```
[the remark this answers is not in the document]
abdattar: Index is used to break the for loop.
You are right, most of the platform dont use RSDT, this part of code is to be in align with the spec.
This block of code or RSDT block of code will be executed only if RSDP version is less than 2, which in unlikely.

DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
abdattar: Ack.

DynamicTablesPkg/Include/X64NameSpaceObjects.h
abdattar: Ack. will remove it.

DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
abdattar: 1) Below loop will cover all the tables that are installed on the system.
    for (Index = 0; Index < TableCount; Index++) {
2) FACS table do not have CRC hence its ignore.
   Spec says the hardware signature should contains the CRC of the tables, Including FACS tables fields to CRC check will be extra. suppose(I am not sure about OS code) if OS also computes the checksum w.r.t to spec then there will mismatch.
Hence I think restricting to each table's CRC would be suffice.

DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
abdattar: ProcessAcpiTables() is within the DynamicTables framework.
Its possible that platform can use other method to install/modify/remove acpi table.

DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
abdattar: I dont know much about linux kernel, I can see usage of HardwareSignature

https://github.com/torvalds/linux/blob/master/drivers/acpi/sleep.c#L960

DynamicTablesPkg/Include/X64NameSpaceObjects.h
abdattar: The global lock is an ACPI memory region, and the platform configuration manager should supply the correct information. Therefore, I believe this field is necessary.

DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
pierregondois: yes right indeed

DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
pierregondois: 1- I m not sure the RSDT contains the pointers to all the tables. For instance I don't thing it contains a pointer to itself. Also, cf. `5.1 Overview of the System Description Table Architecture` and:
https://github.com/tianocore/edk2/blob/d51baa02a6f37b3aeb0b2e0772ca446831aca2d2/MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiTableProtocol.c#L654 

2- Is it possible to point me to the section stating the `Hardware signature` should be a CRC ? I couldn't find it.

DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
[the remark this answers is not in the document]
abdattar: it takes account of first table.
    **CollectedCrc[0] = Rsdp->ExtendedChecksum;   /// RSDP checksum**
    Xsdt            = (EFI_ACPI_DESCRIPTION_HEADER *)(UINTN)Rsdp->XsdtAddress;
    **CollectedCrc[1] = Xsdt->Checksum;  // first table checksum**
    gBS->CalculateCrc32 (
           (UINT8 *)CollectedCrc,
           ARRAY_SIZE (CollectedCrc),
           **&ComputedCrc /// final computation**
           );
    CollectedCrc[0] = ComputedCrc;
   
  Sure i will remove the support for RSDT.

DynamicTablesPkg/Include/X64NameSpaceObjects.h
abdattar: The Global Lock is a field in the ACPI FACS table. Without this field, the generation of the FACS table is not possible and its wrong.
 If it is not provided by the platform configuration manager, then FacsGenerator should supply it.
What value should the FacsGenerator provide for the Global Lock? I didnt fine any API or function which can give me global lock.
 Therefore, it is required to be provided by the configuration.

DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
pierregondois: I think all the tables are available from the XSDT, except the DSDT and the FACS (and maybe SSDT tables). The DSDST/FACS tables are accessed through the FADT. The figure Fig. `5.2: Description Table Structures` illustrates how ACPI tables are organized.

5.2.11.1 Differentiated System Description Table (DSDT)

```
OSPM finds the pointer to the DSDT in the Fixed ACPI Description Table (using the FADT’s DSDT or X_DSDT fields) and then loads the DSDT to create the ACPI Namespace
```

5.2.10 Firmware ACPI Control Structure (FACS)

```
The FACS is passed to an ACPI-compatible OS using the FADT.
```

- Is it possible to check whether the loop in `AcpiFacsUpdateHardwareSignature()` enumerates the DSDT/FACS tables ?
- Ok understood for the CRC
- What I meant with the first comment is that when computing the CRC of the FACS table, the HardwareSignature field should be ignored, otherwise the HardwareSignature will change at each computation even though the tables didn't change.

E.g. if the HardwareSignature=0x11:
- We compute the CRC of all the ACPI tables, including the FACS.HardwareSignature field
- We obtain HardwareSignature2=0x55
- We re-compute the CRC of all the ACPI tables (the tables didn't change)
- We obtain HardwareSignature3=0x88. HardwareSignature3 != HardwareSignature2.
- This is because HardwareSignature3 was computed including HardwareSignature2 and HardwareSignature2 was computed including HardwareSignature1

DynamicTablesPkg/Include/X64NameSpaceObjects.h
abdattar: Done. Removed the GlobalLock from configuration manager.

DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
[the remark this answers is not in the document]
abdattar: 1) Done.
Thank you for noticing the DSDT table; I have made the necessary changes to print the table signature and ensure the DSDT table is accounted for.
2) Removed the RSDT table support.

DynamicTablesPkg/Library/Acpi/X64/AcpiFacsLib/FacsGenerator.c
abdattar: @pierregondois, I did not include the FACS table in the CRC computation due to the fact that
   - it holds the CRC (checksum) values in its HardwareSignature field of all remaining tables and will always be updated.
   - Additionally, the FACS table does not have a CRC(checksum) field for its own table,
      Assuming the spec left it intentionally.
 Should I include the checksum of FACS table?

abdattar:
Hi @pierregondois ,
Thanks for the quick follow-up review.
I have made the necessary changes based on your feedback.
Please take a look at it again.
Thank you
AbduL

pierregondois:
Hello Abdul,
The following commit effectively takes into account the Dsdt and removes the global lock:
- https://github.com/tianocore/edk2/commit/30fec7c881ab9d3052298ea7897c21026790438a

But it doesn't seem to be the branch you pushed for the PR. Is it possible to check the branch you submitted ? Or maybe I did a mismanipulation

abdattar:
Sorry @pierregondois,
Looks like I have overwritten old "Facs" branch with new one with force push.
Not sure how I can revert to old branch, but i do have a copy of old branch on my local system.
Do you want me to push the old branch?
Thanks
AbduL

pierregondois:
> Sorry @pierregondois, Looks like I have overwritten old "Facs" branch with new one with force push. Not sure how I can revert to old branch, but i do have a copy of old branch on my local system. Do you want me to push the old branch? Thanks AbduL

Please submit the branch that you would like to be accepted. I cannot make comments on the old versions of the PR

abdattar:
Sorry for the inconvenience, I have pushed the correct branch with correct patch set.

pierregondois (approved):
Hi Adbul,
Thanks for the modifications !

abdattar:
Hi @pierregondois ,
Rebase and submitted the patch.
Thanks
AbduL

PR #4874, opened 2023-09-28
EmulatorPkg: Fix Source Level Debug on Windows

nate-desimone:
The Visual Studio Windows debugger will only load symbols for PE/COFF images that Windows is aware of. Therefore, to enable source level debugging, all PEI/DXE modules must be loaded via LoadLibrary() or LoadLibraryEx() and the the instance in memory created by LoadLibrary() must be the one that is actually executed.

The current source level debug implementation in EmulatorPkg for Windows is inherited from the old Nt32Pkg. This implementation makes the assumption that all PEI/DXE modules have a DLL export tables with a symbol named InitializeDriver. Therefore, this source level debug implementation requires all modules to be linked in a non-PI spec defined manner. Support for adding the InitializeDriver symbol was removed in EmulatorPkg, which broke source level debugging.

To fix this, the source level debugging implementation has been modified to use the PE/COFF entry point directly. This brings the implementation into compliance with the PI spec and should work with any PEIM/DXE driver. Implementing this requires parsing the in-memory instance of the PE/COFF image created by Windows to find the entrypoint and since PEIMs/DXE drivers are not garunteed to have 4KB aligned sections, it also requires explicit configuration of the page table using VirtualProtect().

With this fix, the debugging experience is now so good it is unprecedented! In Visual Studio Code, add the following to launch.json:

{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "EmulatorPkg Launch",
      "type": "cppvsdbg",
      "request": "launch",
      "program": "${workspaceFolder}/<path_to_build>/Build/EmulatorX64/DEBUG_<tool_chain>/X64/WinHost",
      "args": [],
      "stopAtEntry": false,
      "cwd": "${workspaceFolder}/<path_to_build>/Build/EmulatorX64/DEBUG_<tool_chain>/X64/",
      "environment": [],
      "console": false,
    }
  ]
}

Make modifications to the above template as nessesary and build EmulatorPkg. Now, just add breakpoints directly in Visual Studio Code the way you would with any other software project. When you start the debugger, it will halt at the breakpoint automatically without any extra configuration required.

Reviewed-by: Michael D Kinney <michael.d.kinney@intel.com>
Cc: Andrew Fish <afish@apple.com>
Cc: Ray Ni <ray.ni@intel.com>
Cc: Chasel Chiu <chasel.chiu@intel.com>

PR #10585, opened 2025-01-03
BaseTools: Break Build on Linker Warnings

os-d:
# Description

Today VS2022 and GCC are set to treat all compiler warnings as errors and break the build. However, linker warnings for both do not break the build. There are critical errors that can be treated as warnings as the linker, such as not finding the module entry point and use a default address as the entry point. This will cause a runtime crash for something that should be caught at build time.

This PR has three commits:

PrmPkg: Clarify Architecture Support
--
Remove IA32 from supported architectures, as the MSVC build actually fails due to the linker expecting a __stdcall calling convention. IA32 PRMs are not supported anyway, only 64 bit OSes support PRM, so simply drop the building of IA32 PrmPkg and clarify the README. While there, clean up some markdown errors.

EmulatorPkg: Fix IA32 MSVC Linker Warnings
--
IA32 EmulatorPkg had many linker warnings because with the current set of linker flags, the MSVC linker was expecting the __stdcall calling convention on all entry points. This was an effect of having /SUBSYSTEM:CONSOLE on all binaries built in EmulatorPkg; this is only needed on WinHost, as that is what Windows launches. The linker options are adjusted to only set /SUBSYSTEM:CONSOLE on WinHost.

BaseTools: Break Build on Linker Warnings
--
Today VS2022 and GCC are set to treat all compiler warnings as errors and break the build. However, linker warnings for both do not break the build. There are critical errors that can be treated as warnings as the linker, such as not finding the module entry point and use a default address as the entry point. This will cause a runtime crash for something that should be caught at build time.

This commit adds /WX to VS2022's DLINK_FLAGS and --fatal-warnings to GCC's DLINK_FLAGS for IA32, X64, ARM, and AARCH64 in order to break the build on linker warnings.

VS2022 linker warning 4210 is ignored for all builds because it checks for static initializers and the linking of the VCRuntime. edk2 never links the VCRuntime (except for HOST_APPLICATIONs) and so the presence of static initializers will always cause this warning, even when the edk2 code calls these initializers that would otherwise be called in the _CRT_INIT function of the
VCRuntime. At the time of this commit, it only fails in CryptoPkg for building OpenSSL, but could fail anywhere a static initializer is used.

- [x] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on a platform with a driver that didn't specify an entry point and seeing a runtime crash as result. After this change, building with GCC, the build breaks.

## Integration Instructions

This is marked as a breaking change as any linker warnings a platform may have had now will break the build. They should be fixed. This forces GCC12 because `--no-warn-rwx-segments` was not added until then. To continue to use earlier versions of GCC, simply remove this flag.


makubacki (approved):
Approval for PrmPkg changes

os-d:
@ajfish @niruiyu can you please review from the EmulatorPkg side?

@bexcran @lgao4 can you please review from the BaseTools side?

bexcran (approved):

os-d:
@ajfish @niruiyu can you please review from the EmulatorPkg side?

@mdkinney can you help review from the EmulatorPkg side?

EmulatorPkg/EmulatorPkg.dsc
```diff
@@ -537,15 +537,15 @@
...
 
 !if $(WIN_HOST_BUILD) == TRUE
   #
   # CLANGPDB tool chain depends on WIN_HOST_BUILD flag to generate the windows application.
   #
-  GCC:*_CLANGPDB_*_DLINK_FLAGS     = /ALIGN:4096 /FILEALIGN:4096 /SUBSYSTEM:CONSOLE
-  GCC:DEBUG_CLANGPDB_*_DLINK_FLAGS = /EXPORT:InitializeDriver=$(IMAGE_ENTRY_POINT) /BASE:0x10000
-  GCC:NOOPT_CLANGPDB_*_DLINK_FLAGS = /EXPORT:InitializeDriver=$(IMAGE_ENTRY_POINT) /BASE:0x10000
+  GCC:*_CLANGPDB_*_DLINK_FLAGS     = /ALIGN:4096 /FILEALIGN:4096
+  GCC:DEBUG_CLANGPDB_*_DLINK_FLAGS = /BASE:0x10000
```
mdkinney, line 549: Why was the EXPORT removed?

os-d: @mdkinney , the `EXPORT` was only for WinHost.inf, but specifying it here made it true for all binaries, which produced a linker warning that this was not the `__stdcall` interface that the linker expected for an export, which for the FW modules itself it is not, it is `__cdecl`. So, this is removed from applying to all modules built in EmulatorPkg to only WinHost.inf, which is the only valid module to apply this. This removes the linker warning (and continues to boot EmulatorPkg).

mdkinney: This is incorrect.  At one time, source level debug from VS required the export.  Has that been resolved in a different way?

os-d: Let me follow up on this. I did not try source level debug from VS after this, just booting EmulatorPkg.

os-d: @mdkinney , paged back the information on this. This InitializeDriver based source level debugging was removed a while ago and VS source level debugging has been broken on EmulatorPkg (regardless of my change above) unless you cherry-pick this change: https://github.com/tianocore/edk2/pull/4874. I'm not sure why this change never went in, I did follow up with @nate-desimone , but I did not hear a response.

With Nate's change and my change, I can do source level debugging on VS EmulatorPkg.

mdkinney: I can do source level debug of EmulatorPkg with VS Code without any changes from current HEAD

mdkinney:
This is the code that is using the DLL export "InitializeDriver" 

```c
    //
    // Load the .DLL file into the user process's address space for source
    // level debug
    //
    Library = LoadLibraryEx (DllFileName, NULL, DONT_RESOLVE_DLL_REFERENCES);
    if (Library != NULL) {
      //
      // InitializeDriver is the entry point we put in all our EFI DLL's. The
      // DONT_RESOLVE_DLL_REFERENCES argument to LoadLIbraryEx() suppresses the
      // normal DLL entry point of DllMain, and prevents other modules that are
      // referenced in side the DllFileName from being loaded. There is no error
      // checking as the we can point to the PE32 image loaded by Tiano. This
      // step is only needed for source level debugging
      //
      DllEntryPoint = (VOID *)(UINTN)GetProcAddress (Library, "InitializeDriver");
    }
```

EmulatorPkg/EmulatorPkg.dsc
os-d: Interesting, I took my statement from Nate's PR, but I also can do source level debugging from top of tree.

os-d:
> This is the code that is using the DLL export "InitializeDriver"
> 
> ```c
>     //
>     // Load the .DLL file into the user process's address space for source
>     // level debug
>     //
>     Library = LoadLibraryEx (DllFileName, NULL, DONT_RESOLVE_DLL_REFERENCES);
>     if (Library != NULL) {
>       //
>       // InitializeDriver is the entry point we put in all our EFI DLL's. The
>       // DONT_RESOLVE_DLL_REFERENCES argument to LoadLIbraryEx() suppresses the
>       // normal DLL entry point of DllMain, and prevents other modules that are
>       // referenced in side the DllFileName from being loaded. There is no error
>       // checking as the we can point to the PE32 image loaded by Tiano. This
>       // step is only needed for source level debugging
>       //
>       DllEntryPoint = (VOID *)(UINTN)GetProcAddress (Library, "InitializeDriver");
>     }
> ```

@mdkinney , right and this is what Nate's PR changes. I don't have the full history here on why he made that change, but I do think that the existing InitializeDriver implementation is brittle.

Nate's PR had a Reviewed-By from you and the push label, but never went in. Would you be okay if I included his commit here? I have used it many times for source level debug and had good success with it and he seemingly did originally create it a couple years ago to address issues he was seeing.

Again, with that commit and removing InitializeDriver, EmulatorPkg both boots and has source level debug.

os-d:
The other option being I ignore linker warnings in EmulatorPkg.dsc and drop my change to remove the export. But, again, the export generates a lot of linker warnings, so I think Nate's change + the removal of the export is the safer option.

mdkinney:
I have looked at #4874 from Nate again.  I see what it is doing.  The main advantage is getting rid of GetProcAddress() call and instead parsing the PE/COFF image for the entry point.  Also sets permissions on pages based on sections and handles cases where PE/COFF sections are not 4KB aligned.  I think PR #4874 should be merged first and then I agree that the DLL exports can be removed from all link operations.

os-d:
> I have looked at #4874 from Nate again. I see what it is doing. The main advantage is getting rid of GetProcAddress() call and instead parsing the PE/COFF image for the entry point. Also sets permissions on pages based on sections and handles cases where PE/COFF sections are not 4KB aligned. I think PR #4874 should be merged first and then I agree that the DLL exports can be removed from all link operations.

Sounds good to me. Thanks for catching this.

mdkinney (approved):

mdkinney:
#4874 has been merged now.  Please rebase this one and verify all check pass.

os-d:
@mdkinney all checks passed after rebasing. Adding the push label.

sudeep-holla:
This change seem to be breaking CLANGDWARF build on AARCH64 with the below error:
ld.lld: error: unknown argument '--no-warn-rwx-segments'
clang: error: linker command failed with exit code 1 (use -v to see invocation)

mdkinney:
Please see description at top that describes this as a breaking change and the way to resolve.  Either move to GCC12 or newer or make local change to remove this option to use lower than GCC12.

sudeep-holla:
Oh OK, I was not checking the host GCC version. I just checked the cross compiler version, sorry for that.

ardbiesheuvel:
> Please see description at top that describes this as a breaking change and the way to resolve. Either move to GCC12 or newer or make local change to remove this option to use lower than GCC12.

@sudeep-holla is reporting that the Clang build is broken too. Upgrading the GCC version is not going to help when using Clang.

Using no-warn-rwx-segments is not the right fix here. Instead, the ELF linker scripts should be updated to declare separate R-X and RW- segments using ELF program headers.

PR #10586, opened 2025-01-03
StandaloneMmPkg: Call PeCoffLoaderUnloadImage When Unloading Image

os-d:
# Description

Today, StandaloneMmCore calls PeCoffLoaderRelocateImage() when loading images, which calls PeCoffLoaderRelocateImageExtraAction(). On AARCH64, this sets the image memory protections accordingly, RO + E on code sections, RW + NX on data sections.

However, if an image fails to start (i.e. its entry point returns a failure) StandaloneMmCore does not call the corresponding PeCoffLoaderUnloadImage, which calls PeCoffLoaderUnloadImageExtraAction, which on AARCH64 undoes the memory protections on the image, setting the whole memory region back to RW + NX. The core then frees this memory and the next allocation attempts to use it, which results in a data abort if a read only memory region is attempted to be written to. Theoretically, other instances of the PeCoffExtraActionLib could take other actions and so regardless of architecture, the contract with the PeCoffLoader should be maintained.

This patch calls PeCoffLoaderUnloadImage when an image's entry point returns a failure, before freeing the image memory. This meets the contract and follows the DXE core behavior: https://github.com/tianocore/edk2/blob/dca265a8ca9bc01656ff6bc1fc881bdee132de9a/MdeModulePkg/Core/Dxe/Image/Image.c#L975-L978

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on an AARCH64 platform that was crashing in Standalone MM when a driver was unloaded and confirmed that the system could boot after this patch.

## Integration Instructions

N/A.


StandaloneMmPkg/Core/Dispatcher.c
```diff
@@ -108,33 +108,37 @@ BOOLEAN  gRequestDispatch = FALSE;
 EFI_STATUS
 EFIAPI
 MmLoadImage (
-  IN OUT EFI_MM_DRIVER_ENTRY  *DriverEntry
+  IN OUT EFI_MM_DRIVER_ENTRY           *DriverEntry,
+  IN OUT PE_COFF_LOADER_IMAGE_CONTEXT  *ImageContext
```
samimujawar, line 112: Nit: Can we update the documentation for this function to reflect the addition of this new param, please?

samimujawar (approved):
Hi @os-d,
Thank you for this patch.
I have requested a minor documentation update, With that fixed, this patch looks good to me.

LeviYeoReum (approved):

samimujawar:
@niruiyu, @td36, @jiaxinwu,  can someone confirm if this patch is good for X86/x64, please?

ardbiesheuvel (approved):

Issue #10510, opened 2024-04-25
Null Pointer dereference in Private Data Macros (Bugzilla Bug 4766)


tianocore-issues:
This issue was created automatically with [bugzilla2github](https://github.com/makubacki/bugzilla2github/tree/bz2gh_tianocore)
# Bugzilla Bug 4766
Date: 2024-04-25T14:00:00+00:00
From: cglosner
To:   unassigned &lt;<unassigned>&gt;
CC:   @lgao4

Last updated: 2024-05-29T21:18:51+00:00

tianocore-issues:
## Comment 22982
Date: 2024-04-25 14:00:00 +0000
From: cglosner

- Industry Specification: ---
- Release Observed: EDK II Master
- Releases to Fix: EDK II Master
- Target OS: ---
- Bugzilla Assignee(s): unassigned &lt;<unassigned>&gt;

I&apos;ve run across a NULL pointer dereference in MdeModulePkg/Universal/Disk/DiskIoDxe/DiskIo.c inside the DiskIo2Cancel functions protocol private data macro.

EFI_STATUS
EFIAPI
DiskIo2Cancel (
  IN EFI_DISK_IO2_PROTOCOL  *This
  )
{
  DISK_IO_PRIVATE_DATA  *Instance;
  DISK_IO2_TASK         *Task;
  LIST_ENTRY            *Link;

  Instance = DISK_IO_PRIVATE_DATA_FROM_DISK_IO2 (This);

If NULL is passed in then it results in a crash from a NULL pointer dereference since there is no check that This is not null inside the macro.

That being said I have checked this in many other protocols and it is the same thing where there is no check on the instance of the protocol being passed in. I haven&apos;t checked them all, but 5 or 6 that require a valid instance of the protocol to be passed in.

tianocore-issues:
## Comment 23077
Date: 2024-05-29 21:18:51 +0000
From: @lgao4

The proposal is to update CR definition in DebugLib.h to add the checker for NULL pointer.

PR #10591, opened 2025-01-06
MdePkg/DebugLib: Add null pointer check to CR macro

r1chard-lyu:
- Added a null pointer check in the CR macro.
- If Record is NULL, an assertion is triggered.
- This change improves the CR by preventing dereferencing of null pointers.

REF: https://github.com/tianocore/edk2/issues/10510

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions

### DIFF
--- a/MdePkg/Include/Library/DebugLib.h
+++ b/MdePkg/Include/Library/DebugLib.h
@@ -657,11 +657,12 @@ UnitTestDebugAssert (
 **/
 #if !defined (MDEPKG_NDEBUG)
 #define CR(Record, TYPE, Field, TestSignature)                                              \
+    ((Record == NULL) ? (TYPE *) (_ASSERT (CR has Null Pointer), Record) :                    \
     (DebugAssertEnabled () && (BASE_CR (Record, TYPE, Field)->Signature != TestSignature)) ?  \
     (TYPE *) (_ASSERT (CR has Bad Signature), Record) :                                       \
     (BASE_CR (Record, TYPE, Field)->Signature != TestSignature) ?                             \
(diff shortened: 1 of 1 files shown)


MdePkg/Include/Library/DebugLib.h
```diff
@@ -657,11 +657,12 @@ UnitTestDebugAssert (
 **/
 #if !defined (MDEPKG_NDEBUG)
 #define CR(Record, TYPE, Field, TestSignature)                                              \
+    ((Record == NULL) ? (_ASSERT (CR has Null Pointer), NULL) :                               \
```
lgao4, line 660: Record may point to zero address.

r1chard-lyu: Thank you for your comment. The line was added to check if `Record` is pointing to zero. If Record is pointing to zero, the assertion `_ASSERT (CR has Null Pointer)` is triggered, and `Record` (which is NULL) is returned. I have updated the code accordingly. Please review the updated code and provide any suggestions.

r1chard-lyu: @lgao4 I have updated the code. Could you please review the latest changes?

lgao4: I mean NULL as Record may be the valid case, because zero address may be accessed. So, this checker should not be added.

r1chard-lyu: I agree with it, and I will close this PR. Thank you for your comments.

PR #10593, opened 2025-01-08
MdeModulePkg/UsbBusPei: Improve PEI USB enumeration

philnoh2:
Unlike DXE USB enumeration that enumerates all changed ports on timer interrupt, PEI USB Enumeration runs once at the driver entry point. USB3.x devices initially appear in USB2.0 ports. When the USB2.0 port is reset, the USB3.x device disappears from the USB2.0 port and appears on the USB3.0 port. The USB3.x device won't be enumerated if the USB2.0 port number is greater than the USB3.0 port number. Re-enumerate USB to make sure USB3.x devices in this case.

REF: [eXtensible Host Controller Interface for Universal Serial Bus (xHCI)
](https://www.intel.com/content/www/us/en/content-details/625472/extensible-host-controller-interface-for-universal-serial-bus-xhci-requirements-specification.html)4.19.3 Connect Status Change Reporting
For a USB2 Protocol port, “device presence” is indicated by the PORTSC PLS field transitioning from the RxDetect to the Polling state. Software shall reset the port to transition it to the U0 state. For a USB3 Protocol port, “device presence” is indicated by the PORTSC PLS field transitioning from the Polling to the U0 state.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

The update was verified on some AMD platforms with the case. There was an issue that system is failed for Crisis recovery when the USB device is on a specific Type-C port. Crisis recovery is based on the PEI driver. It resolved the issue.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Bus/Usb/UsbBusPei/UsbPeim.c
+++ b/MdeModulePkg/Bus/Usb/UsbBusPei/UsbPeim.c
@@ -1,6 +1,7 @@
 /** @file
 The module to produce Usb Bus PPI.
 
+Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.<BR>
 Copyright (c) 2006 - 2018, Intel Corporation. All rights reserved.<BR>
 
 SPDX-License-Identifier: BSD-2-Clause-Patent
@@ -428,6 +429,7 @@ PeiUsbEnumeration (
   UINT8                 CurrentAddress;
   UINTN                 InterfaceIndex;
   UINTN                 EndpointIndex;
+  UINT8                 UsbEnumLoop;
 
   CurrentAddress = 0;
   if (Usb2HcPpi != NULL) {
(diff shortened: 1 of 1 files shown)


lgao4:
If this change is request, can it be made in PeiUsbEnumeration () implementation? If so, the caller is not required to be updated.

philnoh2:
> If this change is request, can it be made in PeiUsbEnumeration () implementation? If so, the caller is not required to be updated.

I understand. Yes, we have reviewed 2 options for the required update. The PR update is our second option. We thought that it would be better to explain the issue and update. The first option (patch attached) is to update PeiUsbEnumeration () internally. In the case, the caller's update is not required. If maintainers think that the first option is better or there is the other option, I could follow up. Please review it and let me know. Thank you.

[UsbPeim.1stoption.patch](https://github.com/user-attachments/files/18462478/UsbPeim.1stoption.patch)

philnoh2:
Hi Liming (@lgao4),

It is the updated PR for another option that we have reviewed. It could satisfy the point below. Please help to review it.

> can it be made in PeiUsbEnumeration () implementation? If so, the caller is not required to be updated.

Thank you.

lgao4:
> Hi Liming (@lgao4),
> 
> It is the updated PR for another option that we have reviewed. It could satisfy the point below. Please help to review it.
> 
> > can it be made in PeiUsbEnumeration () implementation? If so, the caller is not required to be updated.
> 
> Thank you.

Can you attach the boot log for the changed logic? I would like to verify the boot functionality.

philnoh2:
> > Hi Liming (@lgao4),
> > It is the updated PR for another option that we have reviewed. It could satisfy the point below. Please help to review it.
> > > can it be made in PeiUsbEnumeration () implementation? If so, the caller is not required to be updated.
> > 
> > 
> > Thank you.
> 
> Can you attach the boot log for the changed logic? I would like to verify the boot functionality.

Based on EDk2 Master (Jan 23), I am attaching 2 bios log files for UsbBusPei driver. One is based on the original version. Another is based on the PR version. On our platform, the driver is loaded in the recovery mode. When a USB3.0 flash key that includes a recovery image is plugged in a USB Type C port (Port 0 - USB3, Port 5 - USB2) on the error case, the PR version makes the device initialized and detected. In the result, the recovery is completed successfully. Please review the log files. Thank you.

[EDK2_Master_0123_UsbBusPei_Original_BiosLog (Crisis Recovery - Fail).txt](https://github.com/user-attachments/files/18528402/EDK2_Master_0123_UsbBusPei_Original_BiosLog.Crisis.Recovery.-.Fail.txt)
[EDK2_Master_0123_UsbBusPei_PR10593_BiosLog (Crisis Recovery - Success).txt](https://github.com/user-attachments/files/18528403/EDK2_Master_0123_UsbBusPei_PR10593_BiosLog.Crisis.Recovery.-.Success.txt)

mdkinney:
@lgao4 This was opened before soft freeze and looks like a good improvement.  Should this be added to edk2-stable202502 milestone?

mdkinney:
@lgao4 Please see question above

mdkinney:
@ajfish @leiflindholm Please review for edk-stable202502 milestone.

MdeModulePkg/Bus/Usb/UsbBusPei/UsbPeim.c
```diff
@@ -449,27 +451,37 @@ PeiUsbEnumeration (
 
   DEBUG ((DEBUG_INFO, "PeiUsbEnumeration: NumOfRootPort: %x\n", NumOfRootPort));
 
-  for (Index = 0; Index < NumOfRootPort; Index++) {
+  //
+  // USB3.x devices initially appear in USB2.0 ports. When the USB2.0 port is reset, the USB3.x device disappears
+  // from the USB2.0 port and appears on the USB3.0 port. The USB3.x device won't be enumerated if the USB2.0 port
+  // number is greater than the USB3.0 port number. Re-enumerate USB to make sure USB3.x devices in this case.
+  //
+  for (Index = 0; Index < (UINT8)(NumOfRootPort * 2); Index++) {
```
leiflindholm, line 459: This line holds way too much magic for my liking. While it may not break anything in our current implementation, it makes no sense semantically to iterate over twice as many elements as our discovery calls above have told us exist.

ardbiesheuvel: Agreed. If the entire loop needs to run twice, we should stick another `for ()` loop around it.

But I'd like to understand better what is happening here. The comment claims that resetting the port is what causes the USB 2.0 device to disappear (I assume this refers to the reset that is part of the LS/FS/HS enumeration sequence), and only after this reset has been issued, the device reappears on a different USB 3.x port, which may appear before or after the USB 2.0 sibling port in the enumeration order. If it appears before, the device appearing there will go unnoticed.

Does this have any basis in the USB spec? Or is this based on your observation? Is there real cause and effect here, or are we just racing with the hardware, and running the loop twice happens to work around it?

I understand that USB 3.x capable devices must only appear once in the USB hierarchy. But I cannot find any references for the requirement to reset the USB 2.0 device in order for the USB 3.x device to appear.

philnoh2: > If the entire loop needs to run twice, we should stick another for () loop around it.

After reviewing the feedback, firstly I have updated the PR to reflect it. Please refer to the updated code.

> Does this have any basis in the USB spec?

Like the PR's description, the reference in the XHCI specification has been added. Please refer to more references below in the specification. 
 
> [XHCI spec 1.2b](https://www.intel.com/content/www/us/en/content-details/625472/extensible-host-controller-interface-for-universal-serial-bus-xhci-requirements-specification.html)
> 4.19.7 Port Routing and Control and Figure 4-35: Port Routing Example
> A USB3 hub is the logical combination of USB2 and USB3 hubs. A USB Peripheral Device is required to only utilize one connection at a time.
>  
> 4.19.8 Cold Attach Status
> For USB2 protocol ports the Current Connect Status (CCS) flag is capable of reporting a device attach in any xHC power state. However, for USB3 protocol ports CCS is asserted only after the link has successfully trained and advanced to the U0 state.

Any USB3.x key supports both USB2 and USB3 signals. Based on the above spec description, a USB3.x key connected to USB3.x port (in polling state) will not have CCS=1 in USB3 port PORTSC register but have CCS=1 in USB2 port (in disabled state) PORTSC register initially. The spec doesn’t explicitly state issuing reset on USB2 port will deactivate USB2 hub and activate USB3 hub, but logically it is true because only one protocol is active at a time and USB2 is one to start with.

The log file also shows the status.

USB Enumeration (Port 0 - USB3.x, Port 5 - USB2) : 

> [EDK2_Master_0123_UsbBusPei_PR10593_BiosLog (Crisis Recovery - Success).txt](https://github.com/user-attachments/files/18528403/EDK2_Master_0123_UsbBusPei_PR10593_BiosLog.Crisis.Recovery.-.Success.txt)
> ...
> XhcPeiGetRootHubPortStatus: Port: 0 State: 2A0 (First Enumeration, CCS : 0, PLS : 5 - RxDetect)
> ...
> XhcPeiGetRootHubPortStatus: Port: 5 State: 202E1 (First Enumeration, CCS : 1, PLS : 7 - Polling)
> ...
> XhcPeiGetRootHubPortStatus: Port: 0 State: 21203 (Second Enumeration, CCS : 1, PLS : 0 - U0)
> ...
> XhcPeiGetRootHubPortStatus: Port: 5 State: 2A0 (Second Enumeration, CCS : 0, PLS : 5 - RxDetect)
> ...

ardbiesheuvel: Thanks for the explanation. I'm not fully convinced though.

What happens if a USB 3.x peripheral is using a USB 2.0 only cable? The second enumeration will not observe the peripheral on the USB 3.x port, but does that interfere with its operation in USB 2.0 mode?

Is it possible to enumerate all USB 2.0 ports first, before the USB 3.x ports? Or is it difficult for software to make this distinction upfront?

philnoh2: Thank you for reviewing the update and information. Let me answer the questions additionally based on our review and understanding.

> What happens if a USB 3.x peripheral is using a USB 2.0 only cable? The second enumeration will not observe the peripheral on the USB 3.x port, but does that interfere with its operation in USB 2.0 mode?

For the case of USB3.x device on USB2.0 cable only, there is no USB3.x signal routed to the port. USB3 hub is not connected to anything. USB3.x device is USB2 device only. The device is still on USB2.0 hub after issuing port reset and moves to next state on the first enumeration round. We also tried to put insulation tape on USB3.x device’s D+/- signal to make the device USB3.x only. In this case, USB3.x is never enumerated in PEI, DXE, windows except hot-plug under shell or windows.

> Is it possible to enumerate all USB 2.0 ports first, before the USB 3.x ports? Or is it difficult for software to make this distinction upfront?

As “4.19.8 Cold Attach Status” states, USB3 port CCS =0 initially-> (Port: 0 State: 2A0 (First Enumeration, CCS : 0, PLS : 5 - RxDetect). There is no way to tell if there is USB3.x device on the port or not initially. For example, empty port1 log: Port: 1 State: 2A0 is the same as port0 with USB3.x device on it. USB2 port is enumerated first one by one in first round. Issuing reset on USB2 port with CCS =1 will wake up USB3.x device on its usb3.x port (CCS=1) for a USB3.x device. For S/W side, it only enumerates the port with CCS=1.

Generally, USB3 port number is greater than USB2 port number. One round will be good to enumerate all USB2 and USB3 devices. But USB3 port number is allowed to be smaller than its USB2 port number. This PR is for addressing this configuration. USB DXE driver is designed in the same way: timer driven event on CCS=1 on any order. Therefore, there is no such problem with USB DXE driver.

lgao4:
@philnoh2 , I think the updated version is better than before. I still keep my approval. I aslo agree to add it to milestone 202502. Let's wait for other people feedback.

philnoh2:
> @philnoh2 , I think the updated version is better than before. I still keep my approval. I aslo agree to add it to milestone 202502. Let's wait for other people feedback.

@lgao4, Thank you for your review. I can see the approval status. Yes, let me expect other people's feedback.

lgao4:
Seemly, there is no more comment. I will merge it today.

PR #10594, opened 2025-01-08
NetworkPkg/IScsiDxe:Fix for out of bound memory access for bz4207 (CVE-2024-38805)

MadhavanJey:
# Description

A malicious iSCSI target could reply to the iSCSI initiator with a malformed packet, causing out-of-bounds memory reads and writes. This most likely leads to a denial of service, as the write primitive should not be exploitable.

To Fix this,
In IScsiBuildKeyValueList, check if we have any data left (Len > 0) before advancing the Data pointer and reducing Len. Avoids wrapping Len. Also replace the AsciiStrLen() call with an open-coded loop which likewise checks Len to make sure we don't overrun the buffer.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions

### DIFF
--- a/NetworkPkg/IScsiDxe/IScsiProto.c
+++ b/NetworkPkg/IScsiDxe/IScsiProto.c
@@ -1903,9 +1903,8 @@ IScsiBuildKeyValueList (
       Data++;
     }
 
-    if (*Data == '=') {
+    if ((Len > 0) && (*Data == '=')) {
       *Data = '\0';
-
       Data++;
       Len--;
     } else {
@@ -1917,8 +1916,17 @@ IScsiBuildKeyValueList (
 
     InsertTailList (ListHead, &KeyValuePair->List);
 
-    Data += AsciiStrLen (KeyValuePair->Value) + 1;
-    Len  -= (UINT32)AsciiStrLen (KeyValuePair->Value) + 1;
+    while ((Len > 0) && (*Data != '\0')) {
+      Len--;
+      Data++;
+    }
+
+    if ((Len > 0) && (*Data == '\0')) {
+      Data++;
+      Len--;
+    } else {
+      goto ON_ERROR;
+    }
   }
 
   return ListHead;


leiflindholm:
This change does not have a commit message.
Please write one, following https://github.com/tianocore/tianocore.github.io/wiki/Commit-Message-Format.

leiflindholm:
> This change does not have a commit message. Please write one, following https://github.com/tianocore/tianocore.github.io/wiki/Commit-Message-Format.

(The description in the PR description is pretty good, but it needs to be in the commit message as well, and the DCO is also missing.)

jkmathews:
Just an FYI to all reviewers, this patch was previously on the EDK2 dev mailing list, but it was never merged. 
Link: https://edk2.groups.io/g/devel/message/106280

leiflindholm:
Commit message still needs updating.

leiflindholm (changes requested):
Commit message needs updating as per comments.

MadhavanJey:
> This change does not have a commit message. Please write one, following https://github.com/tianocore/tianocore.github.io/wiki/Commit-Message-Format.

Commit message has been updated.

NetworkPkg/IScsiDxe/IScsiProto.c
```diff
@@ -1917,8 +1916,17 @@ IScsiBuildKeyValueList (
 
     InsertTailList (ListHead, &KeyValuePair->List);
 
-    Data += AsciiStrLen (KeyValuePair->Value) + 1;
-    Len  -= (UINT32)AsciiStrLen (KeyValuePair->Value) + 1;
+    while ((Len > 0) && (*Data != '\0')) {
```
leiflindholm, line 1919: Could this be folded down to:

```
if (Len > 0) {
  UINT32 Remaining;
  Remaining = AsciiStrLen (KeyValuePair->Value) + 1;

  Data += Remaining;
  Len  -= Remaining;
} else {
  goto ON_ERROR;
}
```
?

There's nothing super clever going on here, so every bit of manual string parsing we can avoid feels like a future misunderstanding avoided.

mdkinney: This PR was closed and a new one opened here: https://github.com/tianocore/edk2/pull/11042

PR #10598, opened 2025-01-08
StandaloneMmgPkg: Core: Unload image if it fails to start

gmahadevan:
If a loaded image fails to start, then call the PeCoff Library to re-map the memory region.

Reviewed-by: Jeff Brasen <jbrasen@nvidia.com>

# Description

If a driver in StandaloneMm fails to load, we need to unload that image and reset the memory attributes for that region, else the next driver would fail to load

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Force a driver load to fail and check if the next driver can be loaded.

## Integration Instructions

N/A

### DIFF
--- a/StandaloneMmPkg/Core/Dispatcher.c
+++ b/StandaloneMmPkg/Core/Dispatcher.c
@@ -36,7 +36,9 @@
 
 **/
 
+#include "Library/DebugLib.h"
 #include "StandaloneMmCore.h"
+#include "Uefi/UefiBaseType.h"
 
 //
 // MM Dispatcher Data structures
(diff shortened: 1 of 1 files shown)


os-d:
This appears to be a duplicate of this: https://github.com/tianocore/edk2/pull/10586. I think my version is a bit more efficient, because we don't re-read the ImageContext. Am I missing anything else additional happening here?

gmahadevan:
> This appears to be a duplicate of this: #10586. I think my version is a bit more efficient, because we don't re-read the ImageContext. Am I missing anything else additional happening here?

You are correct, sorry I didn't look up the PRs before pushing this. Let me try your patch on my setup.

os-d:
> > This appears to be a duplicate of this: #10586. I think my version is a bit more efficient, because we don't re-read the ImageContext. Am I missing anything else additional happening here?
> 
> You are correct, sorry I didn't look up the PRs before pushing this. Let me try your patch on my setup.

No worries, please let me know if that fixes the issue you were seeing! I'm assuming the same one I did.

gmahadevan:
> > > This appears to be a duplicate of this: #10586. I think my version is a bit more efficient, because we don't re-read the ImageContext. Am I missing anything else additional happening here?
> > 
> > 
> > You are correct, sorry I didn't look up the PRs before pushing this. Let me try your patch on my setup.
> 
> No worries, please let me know if that fixes the issue you were seeing! I'm assuming the same one I did.

Your fix works. Thanks. I will withdraw this PR.

I had another change to the MmLoadImage() code.

if the MmAllocatePages() fails, then what ends up happening is that the Dispatcher silently exits and doesn't load any subsequent drivers, this happens on memory restricted systems.
I had added an ASSERT there to atleast stop StMM Boot. Are you thinking about this change as well ?

os-d:
> I had another change to the MmLoadImage() code.
> 
> if the MmAllocatePages() fails, then what ends up happening is that the Dispatcher silently exits and doesn't load any subsequent drivers, this happens on memory restricted systems. I had added an ASSERT there to atleast stop StMM Boot. Are you thinking about this change as well ?

I do not have that change, but an ASSERT there seems fine to me if you want to put up a PR for that.

gmahadevan:
Closing this PR as https://github.com/tianocore/edk2/pull/10586 already fixes this issue.

PR #10600, opened 2025-01-09
Add pip-hashes to complement pip-requirements.

deeglaze:
For stronger build integrity, the build tools as downloaded should be verified against hashes that this repo's maintainers also concur are authentic.

Note the "unsafe" qualifier for setuptools is not troublesome [1], but still pinning pip itself could be a problem [2]

Not pinning pip means that container definitions that pip install --upgrade pip -r requiremenst.txt will fail. The base container's pip package should be sufficient to install the edk2 build dependencies.

There are many hashes this tool adds for single release versions because one tool version can get released on many platforms and Python versions.

The tool stores hashes in alphabetical order rather than in a release list order. I have not culled any since there is no "official" development platform to weed out unnecessary hashes. If we decide that only the CI toolchain containers matter, then we should cull after first showing that they can handle --require-hashes in the pip-install step.

[1] https://github.com/jazzband/pip-tools/issues/806#issuecomment-493591664
[2] https://github.com/pypa/pip/issues/6459

# Description

- [ ] Breaking change? No, the file is not used by anything in this commit.
- [X] Impacts security? Yes, it improves build integrity.
- [ ] Includes tests? No, this will be followed-up with patches to the container definitions.

## How This Was Tested

I copied in the pip-hashes.txt instead of downloading the "master" pip-requirements in the tianocore ubuntu22 container definition:

```
COPY ./pip-hashes.txt /tmp/pip-hashes.txt                                                                                                                                                                                                                                                                                                                     
RUN pip install --require-hashes -r /tmp/pip-hashes.txt 
```

This was successful. Note this drops the --upgrade pip added in https://github.com/tianocore/containers/commit/46802aa8ff4afc3242ac29ee4372fb88e97d71f0.

I recommend that any reviewer install `pip-tools` themselves and run

```
pip-compile --allow-unsafe --generate-hashes --output-file=pip-hashes.txt pip-requirements.txt
```

To compare digests to show there is no funny business.

## Integration Instructions

As tested.


deeglaze:
PR #10592 may need to get merged first. Please advise.

PR #10603, opened 2025-01-09
MdeModulePkg: OvmfPkg: SecurityPkg: ShellPkg: fix [-Werror=maybe-uninitialized]

ghbaccount:
# Description

Sometimes I compile projects using old gcc (gcc-8.1.0)
This time I've tried to build ArmVirtPkg/ArmVirtQemu.dsc  and OvmfPkg/OvmfPkgIa{32,64}.dsc
and found a bunch of similar warnings preventing from complete project build.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

The changes look trivial. 

## Integration Instructions
N/A

### DIFF
--- a/SecurityPkg/RandomNumberGenerator/RngDxe/ArmTrng.c
+++ b/SecurityPkg/RandomNumberGenerator/RngDxe/ArmTrng.c
@@ -44,6 +44,10 @@ GenerateEntropy (
   UINTN       Index;
   UINTN       MaxBits;
 
+  if ((Length == 0) || (Entropy == NULL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
   ZeroMem (Entropy, Length);
 
   RequiredEntropyBits  = (Length << 3);
@@ -67,5 +71,5 @@ GenerateEntropy (
     Index                += (EntropyBits >> 3);
   } // while
 
-  return Status;
+  return EFI_SUCCESS;
 }
--- a/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressMediaSanitize.c
+++ b/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressMediaSanitize.c
@@ -360,6 +360,8 @@ NvmExpressMediaClear (
     return EFI_INVALID_PARAMETER;
   }
 
+  Status = EFI_SUCCESS;
+
   //
   // Per NIST 800-88r1, one or more pass of writes may be alteratively used.
   //
--- a/MdeModulePkg/Core/Dxe/Hand/Handle.c
+++ b/MdeModulePkg/Core/Dxe/Hand/Handle.c
@@ -1107,6 +1107,8 @@ CoreOpenProtocol (
     return EFI_INVALID_PARAMETER;
   }
 
+  Prot = NULL;
+
   //
   // Lock the protocol database
   //
--- a/MdeModulePkg/Core/Pei/Dispatcher/Dispatcher.c
+++ b/MdeModulePkg/Core/Pei/Dispatcher/Dispatcher.c
@@ -1564,6 +1564,11 @@ EvacuateTempRam (
   DEBUG ((DEBUG_VERBOSE, "Beginning evacuation of content in temporary RAM.\n"));
 
   //
+  // By default migrate all FVs and copy raw data
+  //
+  FvMigrationFlags = FLAGS_FV_RAW_DATA_COPY;
+
+  //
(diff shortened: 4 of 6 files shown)


SecurityPkg/RandomNumberGenerator/RngDxe/ArmTrng.c
```diff
@@ -46,6 +46,7 @@ GenerateEntropy (
 
   ZeroMem (Entropy, Length);
 
+  Status               = EFI_INVALID_PARAMETER;
```
samimujawar, line 49: I think it may be better to validate the input arguments before line 47.

if ((Length ==0) || (Entropy == NULL)) {
  return EFI_INVALID_PARAMETER;
}

That way it will also fix the issue at line 47 if Entropy is NULL.

samimujawar (approved):
Thank you for this fix.
The patch "ArmPkg: ArmGic: fix warning about uninitialized variable" looks good to me.

samimujawar (changes requested):
For patch "SecurityPkg/RngDxe: fix warning about uninitialized variable", it would be better to validate the function input parameters.

pierregondois (approved):
The ShellPkg patch lgtm

ArmPkg/Drivers/ArmGic/ArmGicLib.c
```diff
@@ -192,6 +192,7 @@ ArmGicAcknowledgeInterrupt (
     // Report Spurious interrupt which is what the above controllers would
     // return if no interrupt was available
     Value = 1023;
+    IntId = Value;
```
leiflindholm, line 195: Actually, I had a look at this function. Hated it. Since it does two things, and it's just weird. Wrong level of abstraction.

So I went and had a look at who's actually using the insane interface. Turns out no one is. Last user disappeared with 96c8e756819de09d14770aedbd63916c3004e657 "ArmPlatformPkg/PrePeiCore: Drop MPCore variant".

So could we instead delete the function, and its prototype in ArmPkg/Include/Library/ArmGicLib.h?

ghbaccount: I searched for this function on github one more time and did not find any usage of it it in other edk2 based projects. The only project is mu_plus , but it has its own copy mu_silicon_arm_tiano/ArmPkg.
So I can drop this patch and create another PR removing this function.

SecurityPkg/RandomNumberGenerator/RngDxe/ArmTrng.c
ghbaccount: Yes, this a good fix too.
Unfortunately I doesn't fix the warning.
The compiler I'm using is not so smart.
But after adding proposed parameters validation, we can return EFI_SUCCESS unconditionally.
In case of Status gets error value an early return will happen.

ArmPkg/Drivers/ArmGic/ArmGicLib.c
ghbaccount: Removed patch "ArmPkg: ArmGic: fix warning about uninitialized variable"

SecurityPkg/RandomNumberGenerator/RngDxe/ArmTrng.c
ghbaccount: Reworked patch "SecurityPkg/RngDxe: fix warning about uninitialized variable"

lgao4:
The changes in MdeModulePkg and ShellPkg are good to me.

lgao4 (approved):

ghbaccount:
> For patch "SecurityPkg/RngDxe: fix warning about uninitialized variable", it would be better to validate the function input parameters.

Reworked this patch. After adding input parameters validation this function always returns success.

ghbaccount:
> For patch "SecurityPkg/RngDxe: fix warning about uninitialized variable", it would be better to validate the function input parameters.

Done.

ghbaccount:
Sorry!
I can not solve "1 change requested" condition.
I've pushed a fix, I've wrote a comment but I'm not able to find smth like "Resolve conversation"

leiflindholm:
> Sorry! I can not solve "1 change requested" condition. I've pushed a fix, I've wrote a comment but I'm not able to find smth like "Resolve conversation"

Hmm, not clear to me how I could do that either. I can "dismiss review", but let's give @samimujawar until tomorrow to follow up. I can see you've made all the requested changes.

PR #10607, opened 2025-01-10
StandaloneMmPkg/MmIpl: Correct FV HOB check founction

hongbin123:
- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?


jiaxinwu (approved):

jiaxinwu:
Can you also run the test on the QSP to check the MM infra is workable or not? I think the bug could be avoid if we do the test on QSP.

niruiyu (approved):

PR #10611, opened 2025-01-11
BaseTools: Remove linux os from build options

deeglaze:
EDK2 does not run under linux. These compiler triplets should be using unknown to not claim an incorrect platform.

# Description

Corrects target compiler triples for CLANGDWARF.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Built and ran boot tests for AARCH64 and X64 on Vanadium from our fork of EDK2, based on edk2-stable202302. I don't have machines for IA32, ARM, and RISCV64 to test.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Conf/tools_def.template
+++ b/BaseTools/Conf/tools_def.template
@@ -2026,8 +2026,8 @@ DEFINE CLANGDWARF_IA32_X64_DLINK_FLAGS    = DEF(CLANGDWARF_IA32_X64_DLINK_COMMON
 DEFINE CLANGDWARF_IA32_DLINK2_FLAGS       = -Wl,--defsym=PECOFF_HEADER_SIZE=0x220 DEF(CLANGDWARF_DLINK2_FLAGS_COMMON)
 DEFINE CLANGDWARF_X64_DLINK2_FLAGS        = -Wl,--defsym=PECOFF_HEADER_SIZE=0x228 DEF(CLANGDWARF_DLINK2_FLAGS_COMMON)
 
-DEFINE CLANGDWARF_IA32_TARGET             = -target i686-pc-linux-gnu
-DEFINE CLANGDWARF_X64_TARGET              = -target x86_64-pc-linux-gnu
+DEFINE CLANGDWARF_IA32_TARGET             = -target i686-pc-unknown-gnu
+DEFINE CLANGDWARF_X64_TARGET              = -target x86_64-pc-unknown-gnu
 
 DEFINE CLANGDWARF_WARNING_OVERRIDES    = -Wno-parentheses-equality -Wno-empty-body -Wno-unused-const-variable -Wno-varargs -Wno-unknown-warning-option -Wno-unused-but-set-variable -Wno-unused-const-variable -Wno-unaligned-access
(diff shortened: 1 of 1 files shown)


deeglaze:
CI green is deceiving since there is no testing for CLANGDWARF build targets.

I did build -a AARCH64 -t CLANGDWARF -p ArmVirtPkg/ArmVirtQemu.dsc and build -a X64 -t CLANGDWARF -p OvmfPkg/OvmfPkgX64.dsc inside ghcr.io/tianocore/containers/ubuntu-22-dev:61000ca successfully.

Build failures in CI are due to Qemu download failures given server at capacity.

lgao4:
What difference between linux and unknown? What negative impact with linux target?

ardbiesheuvel:
I have no objections to this change in principle, but I'd also be interested in understanding whether or not this solves an actual problem. Some compiler default settings might be hidden behind the compiler triplet, and the ELF-to-PECOFF conversion in EDK2 is not as robust as we would like.

Did you test boot those builds you created?

deeglaze:
@lgao4 the problem to solve is regarding BoringSSL support. There are build configurations that use `#ifdef __linux__` that then assume that Linux capabilities are available to use. EDK2 as a platform should have its own platform define, like `__edk2__` to make OS-specific behavior decisions. The CrtWrapper concept is a lie that will only become more difficult to support as OpenSSL or BoringSSL add more behaviors specific to Linux when they target a linux compiler triplet.

@ardbiesheuvel Yes, I tested AARCH64 and X64 but don't have a means of testing the others. I can revert those if required.

Given that I've heard from Ard that BoringSSL support in EDK2 is less of an asset to the EDK2 project and only benefits Google, I'm attempting to make general improvements without adding maintenance burden.

If you want to only see these changes to add BoringSSL support in its 100% complete form, I don't have a good answer to that, given we don't want to add TLS capabilities to our firmware build, and yet another crypto library in the main tree is not an appropriate solution in my mind. I'd be much happier with a build environment variable like `PACKAGES_PATH` but specifically for replacing cryptography libraries modularly. I don't know how feasible that will be.

ardbiesheuvel:
> @lgao4 the problem to solve is regarding BoringSSL support. There are build configurations that use `#ifdef __linux__` that then assume that Linux capabilities are available to use. EDK2 as a platform should have its own platform define, like `__edk2__` to make OS-specific behavior decisions. The CrtWrapper concept is a lie that will only become more difficult to support as OpenSSL or BoringSSL add more behaviors specific to Linux when they target a linux compiler triplet.
> 

Agreed. OpenSSL currently uses OPENSSL_SYS_UEFI for this purpose as well.

> Given that I've heard from Ard that BoringSSL support in EDK2 is less of an asset to the EDK2 project and only benefits Google, I'm attempting to make general improvements without adding maintenance burden.
> 
> If you want to only see these changes to add BoringSSL support in its 100% complete form, I don't have a good answer to that, given we don't want to add TLS capabilities to our firmware build, and yet another crypto library in the main tree is not an appropriate solution in my mind. I'd be much happier with a build environment variable like `PACKAGES_PATH` but specifically for replacing cryptography libraries modularly. I don't know how feasible that will be.

The CLANG and GCC toolchains are part of the same 'familiy' in EDK2. While any Clang build can support any target triplet, GCC is built for one specific target.

If we adopt your suggestion, and rely on `__linux__` meaning `not EDK2`, we will end up in a situation where you cannot even build EDK2 with the native compiler on Linux. IOW, on x86, you would need x86-none-gcc because the native GCC is x86-linux-gnu-gcc, and it will #define `__linux__` and potentially break the build.

In the long term, I would prefer to rely on `-ffreestanding` vs `-fhosted` if we can - this more accurately describes the difference, and could hopefully remove the need for file I/O stubs and other syscall related plumbing in CrtWrapper that is only needed in a hosted environment. It can be detected in source by testing the `__STDC_HOSTED__` macro.

PR #10615, opened 2025-01-12
MdePkg/Include: Use _Static_assert for clang and GNUC

mdkinney:
# Description

The clang compiler generates the following error

```error: use of 'static_assert' without inclusion of <assert.h>```

This is due to the use of the MSC Extension `static_assert`. Use `_Static_assert` instead for clang and GNUC compilers.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

EmulatorPkg CLANGDWARF builds under Linux can reproduce this error before this change, and are resolved after this change.

## Integration Instructions

N/A

### DIFF
--- a/MdePkg/Include/Base.h
+++ b/MdePkg/Include/Base.h
@@ -800,12 +800,14 @@ typedef UINTN *BASE_LIST;
   @param  Message     Raised compiler diagnostic message when expression is false.
 
 **/
-#ifdef MDE_CPU_EBC
-#define STATIC_ASSERT(Expression, Message)
-#elif defined (_MSC_EXTENSIONS) || defined (__cplusplus)
+#if defined (__cplusplus)
+#define STATIC_ASSERT  static_assert
+#elif defined (__GNUC__) || defined (__clang__)
+#define STATIC_ASSERT  _Static_assert
+#elif defined (_MSC_EXTENSIONS)
 #define STATIC_ASSERT  static_assert
 #else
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: MdePkg/Include/Base.h [hunk 1]
```diff
@@ -800,12 +800,14 @@ typedef UINTN *BASE_LIST;
   @param  Message     Raised compiler diagnostic message when expression is false.
 
 **/
-#ifdef MDE_CPU_EBC
-#define STATIC_ASSERT(Expression, Message)
-#elif defined (_MSC_EXTENSIONS) || defined (__cplusplus)
+#if defined (__cplusplus)
```
lgao4, line 803: So, don't need to specially handle MDE_CPU_EBC arch?

mdkinney: EBC support has already been removed from tools_def.txt.  This is just one is many places in C/H files that need to remove special cases for EBC.

lgao4 (approved):

mdkinney:
@merfigy refresh

mdkinney:
@mergify rebase

ardbiesheuvel:
This breaks some platforms in `edk2-platforms` which use the pre-processor to generate device tree binaries.

`Platform/AMD/OverdriveBoard/DeviceTree/OverdriveBoard.inf`

I think we should just drop the error case here.

PR #10623, opened 2025-01-14
BaseTools: build_rule.template wholearchive test

apop5:
# Description
When running the /wholearchive test build, generate the test .dll as a different filename to prevent the system holding onto the dll file too long and generating a build error that the actual dll cannot be found.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Local Builds using VS2022 and VS2019. 

## Integration Instructions
N/A

### DIFF
--- a/BaseTools/Conf/build_rule.template
+++ b/BaseTools/Conf/build_rule.template
@@ -287,7 +287,8 @@
         $(DEBUG_DIR)(+)$(MODULE_NAME).dll
 
     <Command.MSFT, Command.INTEL>
-        "$(DLINK)" /OUT:${dst} $(DLINK_FLAGS) $(DLINK2_FLAGS) $(DLINK_SPATH) @$(STATIC_LIBRARY_FILES_LIST)
+        "$(DLINK)" /OUT:${dst}.wtemp $(DLINK_FLAGS) $(DLINK2_FLAGS) $(DLINK_SPATH) @$(STATIC_LIBRARY_FILES_LIST)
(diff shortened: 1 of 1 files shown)


mdkinney:
I observed this same issue and was working on a proposed fix as well.  I like the idea of creating a different output filename for each of the 2 $(DLINK) actions.  We do not want to increase the size of the output dir for these 2 $(DLINK) actions, which is why the same output file was use originally.

I recommend adding a $(RM) step to remove the .wtemp file.  If multiply defined symbols are detected, the log will list the errors, and the .wtemp file has no benefit.

apop5:
> I observed this same issue and was working on a proposed fix as well.  I like the idea of creating a different output filename for each of the 2 `$(DLINK)` actions.  We do not want to increase the size of the output dir for these 2 `$(DLINK)` actions, which is why the same output file was use originally.
> 
> 
> 
> I recommend adding a `$(RM)` step to remove the `.wtemp` file.  If multiply defined symbols are detected, the log will list the errors, and the `.wtemp` file has no benefit.

I can do that

mdkinney (approved):

mdkinney:
@bexcran Can you please review. This failure is starting to be seen on developer systems.

os-d (approved):

lgao4 (approved):

lgao4:
I agree the fix can avoid the potential build failure.

AshrafAliS (approved):

Issue #10626, opened 2025-01-15
[Bug]: <title> DXE_ASSERT when issuing drvcfg command in UEFI Shell


zhurui22:
### Is there an existing issue for this?

- [X] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### Code first?

- [X] Yes

### What packages are impacted?

ShellPkg

### Which targets are impacted by this bug?

DEBUG

### Current Behavior

 DXE_ASSERT when issuing drvcfg command in UEFI Shell

### Expected Behavior

To display the list of devices that are available for configuration

### Steps To Reproduce

1. boot to shell
2. input comand dvrcfg and press enter.
3. system will halt.

### Build Environment

```markdown
- OS(s):Windows11
- Tool Chain(s):
```

### Version Information

```text
c69e5e647d69fcbb6d2a4e5da82e72122d41566f
```

### Urgency

Low

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

No maintainer feedback needed

PR #10627, opened 2025-01-15
ShellPkg: DXE_ASSERT when issuing drvcfg command in UEFI Shell

zhurui22:
Fixes https://github.com/tianocore/edk2/issues/10626

Issue command drvcfg derectly, system halt

Reviewed-by: Liming Gao <gaoliming@byosoft.com.cn>

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions

### DIFF
--- a/ShellPkg/Library/UefiShellDriver1CommandsLib/DrvCfg.c
+++ b/ShellPkg/Library/UefiShellDriver1CommandsLib/DrvCfg.c
@@ -1370,12 +1370,6 @@ ShellCommandRunDrvCfg (
       FileName = NULL;
     }
 
-    if (FileName == NULL) {
-      ASSERT (FileName != NULL);
-      ShellStatus = SHELL_INVALID_PARAMETER;
-      goto Done;
-    }
-
     if (InFromFile && EFI_ERROR (ShellFileExists (FileName))) {
       ShellPrintHiiEx (-1, -1, NULL, STRING_TOKEN (STR_GEN_FIND_FAIL), gShellDriver1HiiHandle, L"drvcfg", FileName);
       ShellStatus = SHELL_INVALID_PARAMETER;


lgao4:
With this change, drvcfg will output help message when no input is specified?

pierregondois:
> With this change, drvcfg will output help message when no input is specified?

It seems the command `drvcfg` alone is valid, so it might not be necessary to output help message.
However issuing `drvcfg -c` doesn't output anything for me even though all devices and child devices should be printed out.

```
To display the list of devices that are available for configuration:
Shell> drvcfg
To display the list of devices and child devices that are available for configuration:
Shell> drvcfg –c
```

lgao4:
> drvcfg

According to UEFI Shell spec, please make sure this command works.

To display the list of devices that are available for configuration:
Shell> drvcfg
To display the list of devices and child devices that are available for configuration:
Shell> drvcfg -c

pierregondois:
> > With this change, drvcfg will output help message when no input is specified?
> 
> It seems the command `drvcfg` alone is valid, so it might not be necessary to output help message. However issuing `drvcfg -c` doesn't output anything for me even though all devices and child devices should be printed out.
> 
> ```
> To display the list of devices that are available for configuration:
> Shell> drvcfg
> To display the list of devices and child devices that are available for configuration:
> Shell> drvcfg –c
> ```

Just in case, this patch is an improvement compared to the current state where `drvcfg -c` crashes,
it's just that it would be even better to have a command doing something usefull.

lgao4 (approved):

lgao4:
> > > With this change, drvcfg will output help message when no input is specified?
> > 
> > 
> > It seems the command `drvcfg` alone is valid, so it might not be necessary to output help message. However issuing `drvcfg -c` doesn't output anything for me even though all devices and child devices should be printed out.
> > ```
> > To display the list of devices that are available for configuration:
> > Shell> drvcfg
> > To display the list of devices and child devices that are available for configuration:
> > Shell> drvcfg –c
> > ```
> 
> Just in case, this patch is an improvement compared to the current state where `drvcfg -c` crashes, it's just that it would be even better to have a command doing something usefull.

Agree this change is better than before.

pierregondois:
```
 * First line of commit message (subject line) is too long (92 >= 76).
```

Otherwise lgtm aswell

PR #10634, opened 2025-01-16
Bug xxx crypto pkg openssl clang asm file fix

mdkinney:
# Description

Update the CryptoPkg to support CLANGPDB and CLANGDWARF from
    both Windows and Linux host environments.

* Add PcdOpensslLibAssemblySourceStyleNasm to select the correct
  optimized assembly source style for OpensslLib for IA32/X64.
  NASM style is for MSFT and CLANGPDB. GAS style is for GCC.
  Use this PCD in OpensslLibAccel.inf and OpensslLibFullAccel.inf
  to select between .nasm and .S files.
* Add missing intrinsic functions required by CLANG IA32/X64 builds.
  * __ashlti3
  * __lshrdi3
* Disable warning -Wno-error=unused-function for CLANG build
  compatibility
* Set -D OPENSSL_NO_INLINE_ASM for CLANG build compatibility
* Update TestBaseCryptLib to split out the implementation of
   main() into its own C file that is only use for host-based
   unit tests. This is due to CLANGPDB for host environments
   injecting a __main() call that can only be resolved in host
   based builds that link against host libraries.
* Update Configure.py to update IA32/X64 [Sources] sections
  with feature flag expressions using the new PCD
  PcdOpensslLibAssemblySourceStyleNasm.

Align CLANG and GCC linker scripts to not discard COMMON
sections for CLANG builds. If COMMON sections are discarded,
then global variables from optimized OpensslLib assembly
source files are removed and results in symbol not found
errors.

- [X] Breaking change?
  - Examples: Platform DSC files that use `OpesslLibAccel.inf` or `OpensslLibFullAccel.inf` must add the following statement to select the correct asm source files:

```  
  !include CryptoPkg\CryptoPkgFeatureFlagPcds.dsc.inc
```

- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Build CryptoPkg on Windows using VS2022, CLANGPDB and CLANGDWARF for IA32/X64 for DEBUG/RELEASE/NOOPT
Build CryptoPkg on Linux using GCC, GCC5, CLANGPDB and CLANGDWARF for IA32/X64 for DEBUG/RELEASE/NOOPT

## Integration Instructions

Platform DSC file should add the following to support all OpensslLib instances

```  
  !include CryptoPkg\CryptoPkgFeatureFlagPcds.dsc.inc
```


liyi77 (approved):

jyao1:
would you please fix the conflict?

mdkinney:
PR #10687 was merged that consolidated the CLANG and GCC linker scripts to the GccBase.lds that was not discarding the COMMON sections.  The commit in this PR to prevent COMMON section removal has been removed and this PR has been rebased.

mdkinney:
@jyao1 @liyi77 I pushed one more update that was required to build `CryptoPkg` unit tests for `CLANGPDB` `NOOPT` on both Windows and Linux.  The file `CryptoPkg/Test/UnitTest/Library/BaseCryptLib/UnitTestMain.c` containes `main()` to support host based unit test builds. This function is unused when building target based unit tests.  However, when `CLANGPDB` `NOOPT` profile is used, unused functions are not optimized away, and the clang compiler has special treatment for `main()`.  It always injects a call to `__main()` that is a function implemented in the clang library that handles global constructors.  That causes an unresolved external when building target based unit tests.

The fix was to move `main()` into a new files called `UnitTestMainHost.c` so the clang compiler does not see a function called `main()` when building target based unit tests.

liyi77:
This change looks fine to me, thanks.

PR #10635, opened 2025-01-16
Move GIC PCDs from ArmPkg to MedPkg

AjanZhong:
# Description

As discussed in this topic: https://edk2.groups.io/g/devel/topic/patch_v5_2_6/102725178, move GIC PCDs from ArmPkg to MdePkg.

Update usage of GIC PCDs usage to adapt changes.

Cherry-pick from:
https://github.com/leiflindholm/edk2/commit/ed2e0db
https://github.com/leiflindholm/edk2/commit/87fe6ff
https://github.com/leiflindholm/edk2/commit/91567fb

- [ ] Breaking change?
  - N/A
- [ ] Impacts security?
  - **Security** - N/A
- [ ] Includes tests?
  - **Tests** -N/A

### DIFF
--- a/MdePkg/MdePkg.dec
+++ b/MdePkg/MdePkg.dec
@@ -2483,6 +2483,15 @@
   # @Prompt Time-out for a request, internal
   gEfiMdePkgTokenSpaceGuid.PcdIpmiSerialRequestRetryInterval|60000|UINT32|0x00000055
 
+[PcdsFixedAtBuild.AARCH64, PcdsFixedAtBuild.ARM, PcdsDynamic.AARCH64, PcdsDynamic.ARM]
+  #
+  # ARM Generic Interrupt Controller
+  #
+  gEfiMdePkgTokenSpaceGuid.PcdGicDistributorBase|0|UINT64|0x00000056
+  # Base address for the GIC Redistributor region that contains the boot CPU
+  gEfiMdePkgTokenSpaceGuid.PcdGicRedistributorsBase|0|UINT64|0x00000057
+  gEfiMdePkgTokenSpaceGuid.PcdGicInterruptInterfaceBase|0|UINT64|0x00000058
+
 [PcdsFixedAtBuild.AARCH64, PcdsPatchableInModule.AARCH64]
   ## GUID identifying the Rng algorithm implemented by CPU instruction.
   # @Prompt CPU Rng algorithm's GUID.
--- a/ArmPkg/ArmPkg.dec
+++ b/ArmPkg/ArmPkg.dec
@@ -342,14 +342,6 @@
   gArmTokenSpaceGuid.PcdGenericWatchdogEl2IntrNum|93|UINT32|0x00000009
 
   #
-  # ARM Generic Interrupt Controller
-  #
-  gArmTokenSpaceGuid.PcdGicDistributorBase|0|UINT64|0x0000000C
-  # Base address for the GIC Redistributor region that contains the boot CPU
-  gArmTokenSpaceGuid.PcdGicRedistributorsBase|0|UINT64|0x0000000E
-  gArmTokenSpaceGuid.PcdGicInterruptInterfaceBase|0|UINT64|0x0000000D
-
-  #
   # Bases, sizes and translation offsets of IO and MMIO spaces, respectively.
   # Note that "IO" is just another MMIO range that simulates IO space; there
   # are no special instructions to access it.
--- a/ArmPkg/ArmPkg.dec
+++ b/ArmPkg/ArmPkg.dec
@@ -348,7 +348,6 @@
   # Base address for the GIC Redistributor region that contains the boot CPU
   gArmTokenSpaceGuid.PcdGicRedistributorsBase|0|UINT64|0x0000000E
   gArmTokenSpaceGuid.PcdGicInterruptInterfaceBase|0|UINT64|0x0000000D
-  gArmTokenSpaceGuid.PcdGicSgiIntId|0|UINT32|0x00000025
 
   #
   # Bases, sizes and translation offsets of IO and MMIO spaces, respectively.
--- a/ArmPkg/Drivers/ArmGicDxe/ArmGicDxe.inf
+++ b/ArmPkg/Drivers/ArmGicDxe/ArmGicDxe.inf
@@ -54,9 +54,9 @@
   gEfiCpuArchProtocolGuid         ## CONSUMES ## NOTIFY
 
 [Pcd.common]
-  gArmTokenSpaceGuid.PcdGicDistributorBase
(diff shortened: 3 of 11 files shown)


kraxel (approved):

MdePkg/MdePkg.dec
```diff
@@ -2483,6 +2483,19 @@
...
+  # ARM Generic Interrupt Controller
+  #
+  gEfiMdePkgTokenSpaceGuid.PcdGicDistributorBase|0|UINT64|0x00000056
+  # Base address for the GIC Redistributor region that contains the boot CPU
+  gEfiMdePkgTokenSpaceGuid.PcdGicRedistributorsBase|0|UINT64|0x00000057
+  gEfiMdePkgTokenSpaceGuid.PcdGicInterruptInterfaceBase|0|UINT64|0x00000058
+
+[PcdsFeatureFlag.AARCH64, PcdsFeatureFlag.ARM]
+  # Define if the GICv3 controller should use the GICv2 legacy
+  gEfiMdePkgTokenSpaceGuid.PcdArmGicV3WithV2Legacy|FALSE|BOOLEAN|0x00000059
```
ardbiesheuvel, line 2497: Can we just get rid of this one?

AjanZhong: ArmGicLib and ArmGicDxe are still using this PCDs:
https://github.com/tianocore/edk2/blob/master/ArmPkg/Drivers/ArmGic/ArmGicLib.c#L196
https://github.com/tianocore/edk2-archive/blob/master/ArmPkg/Drivers/ArmGic/GicV3/ArmGicV3Dxe.c#L257

Not sure whether any user is still using this GiC V2 Legacy.

ardbiesheuvel: This is 10 year old cruft that was added at the time because the GIC version detection is based on the presence of the CPU system register interface rather than the architecture the GIC itself claims to implement.

At the time, this was needed because TF-A was lagging behind, and non-secure can only use the GIC in v3 mode if the secure world does so too.

I'd actually like to some early spring cleaning here, and not only rip this out, but separate the v2 and v3 GixDxes as well - platforms rarely need both, and the V2Legacy behavior can be retained (if needed) by only including the v2 DXE.

I'll try to whip something up asap - in the mean time, can we just drop this V2Legacy for now?

leiflindholm: Since I'm not a repo admin, I can't push to the PR branch, so I've updated my branch
https://github.com/leiflindholm/edk2/tree/rework/GicPcd-to-MdePkg

The diff of ArmGicV3Dxe.c looks messy, but I did only delete the if-side, de-indent the else-side, and move a variable declaration to the top of the function.

AjanZhong: Thanks @leiflindholm, let me update this PR according to your latest changes.

AjanZhong: Done.

MdePkg/MdePkg.dec
```diff
@@ -2483,6 +2483,15 @@
   # @Prompt Time-out for a request, internal
   gEfiMdePkgTokenSpaceGuid.PcdIpmiSerialRequestRetryInterval|60000|UINT32|0x00000055
 
+[PcdsFixedAtBuild.common.AARCH64, PcdsFixedAtBuild.common.ARM, PcdsDynamic.common.AARCH64, PcdsDynamic.common.ARM]
```
lgao4, line 2486: Syntax should be PcdsFixedAtBuild.AARCH64

leiflindholm: Whoops, right you are.
Travelling today with shaky internet connection - if someone else could fix this (drop all of the `.common` from this line), that'd be great. If not, I'll do it when practical.

ardbiesheuvel: Actually, I don't think these belong in MdePkg at all. They are an implementation detail of the GIC driver, and belng wherever that driver lives.

I've pushed a series that cleans up the GIC and fixes this issue as well. This should remove the need for this hack, and also remove the need for any of the GIC-related definitions to be moved out of ArmPkg

https://github.com/tianocore/edk2/pull/10636/commits

AjanZhong: ".common" removed in MdePkg.dec.

AjanZhong:
Rebase to fix conflict since https://github.com/tianocore/edk2/pull/10636/  (Arm gicdxe cleanup) merged

AjanZhong:
> Do we still need this PR?

Yes, we need to expose these PCDs to Packages/Modules which cannot reference ArmPkg.

ardbiesheuvel:
> > Do we still need this PR?
> 
> Yes, we need to expose these PCDs to Packages/Modules which cannot reference ArmPkg.

Why?

AjanZhong:
For instance, UefiPayload needs to create mapping for distribution/redistribution base for GiC.

AjanZhong:
> > > Do we still need this PR?
> > 
> > 
> > Yes, we need to expose these PCDs to Packages/Modules which cannot reference ArmPkg.
> 
> Why?

For instance, UefiPayload needs to create mapping for distribution/redistribution base for GiC.

ardbiesheuvel:
> > > > Do we still need this PR?
> > > 
> > > 
> > > Yes, we need to expose these PCDs to Packages/Modules which cannot reference ArmPkg.
> > 
> > 
> > Why?
> 
> For instance, UefiPayload needs to create mapping for distribution/redistribution base for GiC.

OK, understood. But there are other MMIO regions that need to be mapped, right?

I would prefer a new abstraction for this, rather than teaching UefiPayloadPkg about all the various PCDs that a platform might use internally.

For instance, a GICv2 platform needs to map the CPU interface too. 

Alternatively, it might be feasible to map the entire PA space using device mappings, and only map the RAM cacheable.

AjanZhong:
Hi @lgao4 @ardbiesheuvel Seems all comments have been addressed. Do you have any more comments to this pull request? Thanks.

ardbiesheuvel:
As I said already, I don't think these PCDs belong in MdePkg at all. So in case it wasn't clear, NAK to this series.

If the GIC driver needs mappings of its MMIO regions, it should create them and not rely on the platform code to have done so by the time it executes.

If other components need to communicate information about the platform, e.g., to populate ACPI tables, we need a proper API for that, and not expose arbitrary PCDs to the world that are fundamentally implementation details of the driver. And note that the PCDs only give you a base address, and so your code still needs to guess the size of the region it maps.

I can have a stab at fixing ArmGicDxe to perform these mappings as needed later this week. But please stop pushing these changes.

AjanZhong:
Close this PR due to https://github.com/tianocore/edk2/pull/10690 merged

PR #10636, opened 2025-01-16
Arm gicdxe cleanup

ardbiesheuvel:
Refactor the ARM GIC DXE library and driver code, so that the v2 pieces can be phased out.


leiflindholm:
Very nice cleanup!

Now we don't have the mess of an ArmGicLib residing in the same directory as the Dxe driver, could we also rename the directory ArmGicDxe?

ardbiesheuvel:
> Very nice cleanup!
> 
> Now we don't have the mess of an ArmGicLib residing in the same directory as the Dxe driver, could we also rename the directory ArmGicDxe?

Good idea.

leiflindholm (approved):

Issue #10637, opened 2025-01-17
[Bug]: CLANGPDB linker warning /align without /driver


deeglaze:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [ ] Firmware
- [x] Tool
- [ ] Unit Test

### Code first?

- [ ] Yes

### What packages are impacted?

Build or CI Code

### Which targets are impacted by this bug?

DEBUG, RELEASE, NOOPT

### Current Behavior

When I build OvmfPkg with CLANGPDB, I see a linker warning

```
lld-link: warning: /align specified without /driver; image may not run
```

### Expected Behavior

No warnings during linking.

### Steps To Reproduce

build -p OvmfPkg/OvmfPkgX64.dsc -t CLANGPDB -a X64 -b DEBUG -r OVMF -D SECURE_BOOT_ENABLE=TRUE

### Build Environment

```markdown
- OS(s): Debian 12
- Tool Chain(s): CLANGPDB
```

### Version Information

```text
Tag: edk2-stable202302
```

### Urgency

Low

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

No maintainer feedback needed

lgao4:
PR has been merged. It can be closed.

PR #10638, opened 2025-01-17
BaseTools: Add /DRIVER to CLANGPDB link flags

deeglaze:
# Description

This quiets the warning reported in Issue #10637. /ALIGN is not expected to be overridden without /DRIVER, which is correctly passed in the MSVC linker flags.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Built before and after to see the warning disappear.

## Integration Instructions

N/A


mdkinney:
I don't think we want this at this scope because it can not be overriden.  When building apps like GoogleTest we do not want /DRIVER.

deeglaze:
Why is /DRIVER in all the VS build type DLINK_FLAGS then?

mdkinney:
I believe it is required for the UEFI module types.  I am not sure if that has changes with newer compilers/linkers or not.

lgao4:
> I believe it is required for the UEFI module types. I am not sure if that has changes with newer compilers/linkers or not.

Seemly, CLANGPDB tool chain has the same behavior with VS tool chain. I think to take the same change in CLANGPDB tool chain.

mdkinney (approved):

mdkinney:
@mergify rebase

PR #10640, opened 2025-01-17
MdePkg/Base: Don't error out on missing compiler CPP macros

ardbiesheuvel:


mdkinney:
I was not aware of that use case.

mdkinney (approved):

ardbiesheuvel:
@mergifyio requeue

PR #10642, opened 2025-01-17
StandaloneMmPkg: Introduce a PCD to disable shadow boot FV

LeviYeoReum:
On some Arm platforms the boot firmware volume passed in the HOB by the secure world firmware (TF-A) is never freed and is always mapped as read-only.

On such platforms the heap memory can be saved by disabling the shadow copy of the boot firmware volume. This is useful as on some platforms the amount of memory available is limited.

Therefore, introduce a PCD PcdShadowBfv that platforms can configure to disable the shadow boot FV. 
This PCD is set to TRUE by default.

### DIFF
--- a/StandaloneMmPkg/StandaloneMmPkg.dec
+++ b/StandaloneMmPkg/StandaloneMmPkg.dec
@@ -55,6 +55,13 @@
   # @Prompt Maximum permitted FwVol section nesting depth (exclusive) in MM.
   gStandaloneMmPkgTokenSpaceGuid.PcdFwVolMmMaxEncapsulationDepth|0x10|UINT32|0x00000001
 
+  ## Option to make shadow copy of boot firmware volume while loading drivers.
+  #  This options is useful when the BFV is not located in the Flash area
+  #  but is in the RAM instead and therefore no shadow copy is needed.
+  #   TRUE  - Default. Make a shadow copy of the boot firmware volume.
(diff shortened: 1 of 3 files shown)


StandaloneMmPkg/StandaloneMmPkg.dec
```diff
@@ -55,6 +55,13 @@
   # @Prompt Maximum permitted FwVol section nesting depth (exclusive) in MM.
   gStandaloneMmPkgTokenSpaceGuid.PcdFwVolMmMaxEncapsulationDepth|0x10|UINT32|0x00000001
 
+  ## Option to disable shadow copy of boot firmware volume while loading drivers.
+  #  This options is useful when the BFV is not located in the Flash area
+  #  but is in the RAM instead and therefore no shadow copy is needed.
+  #   TRUE  - Disable shadow copy of the boot firmware volume.
+  #   FALSE - Default. Make a shadow copy of the boot firmware volume.
+  gStandaloneMmPkgTokenSpaceGuid.PcdDisableShadowBfv|FALSE|BOOLEAN|0x00000003
```
lzeng14, line 63: instead of DISALBE+FALSE, maybe the PCD name could be PcdShadowBfv with default TRUE to be more user friendly ?

LeviYeoReum: Thanks. I'll fix it with `PcdShadowBfv`.

samimujawar (approved):

PR #10646, opened 2025-01-20
PcAtChipsetPkg: Remove redundant ASSERT in HPET driver

abdattar:
Eliminate the redundant ASSERT in the HPET driver
that triggers when HPET capabilities cannot be read.

The driver will return an error code and unload gracefully.

Thus, the unnecessary ASSERT is removed.

Cc: Ray Ni <ray.ni@intel.com>

# Description
Removes unnecessary ASSERT.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on AMD platform

## Integration Instructions

N/A

### DIFF
--- a/PcAtChipsetPkg/HpetTimerDxe/HpetTimer.c
+++ b/PcAtChipsetPkg/HpetTimerDxe/HpetTimer.c
@@ -2,6 +2,8 @@
   Timer Architectural Protocol module using High Precision Event Timer (HPET)
 
   Copyright (c) 2011 - 2018, Intel Corporation. All rights reserved.<BR>
+  Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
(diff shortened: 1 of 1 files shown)


mdkinney:
If the HPET driver is included, then the platform should have valid HPET hardware present.

In general, these ASSERT()s indicate ab invalid platform FW configuration and either the HPET HW is not enabled/configured correctly or the HPET driver should be removed.

Is there a use case where the HPET driver is present and should be unloaded and a different timer driver should succeed?

niruiyu (approved):

mdkinney:
@abdattar Any feedback on my comment above?

abdattar:
Hi @mdkinney,
  Sorry I missed the last comment.
The proposed patch handles two cases.
1) Platform to share common dsc files, one platform choose to have HPET device and another platform can have local apic timer.
    This makes it easier to have command platform dsc file, if the HPET device is not present the HPET driver will be unloaded, 
    instead of ASSERT.
 2) Platform can provide setup option to enable/disable HPET.
     if user choose to disable the HPET, then it will not ASSERT(), instead unload the driver.
 Thanks
AbduL

mdkinney:
These use cases have challenges because there can be at most 1 instance of the Timer Arch Protocol.  Many Timer Arch Protocol implementations ASSERT() if there is already a Timer Arch Protocol installed.   For example:

https://github.com/tianocore/edk2/blob/62cd7d338b389aa21c00f0fc35f14a6fa9ba23b0/PcAtChipsetPkg/HpetTimerDxe/HpetTimer.c#L783

Seems like this patch is incomplete to handle multiple timer drivers.

Also, even if these ASSERT()s for Timer Arch Protocol already installed are removed, having multiple timer drivers in a platform depends on the dispatch order.  Your case of disabling the HPET HW.  If the non HPET timer driver is always dispatched first, then the HPET timer driver will never be used.

There is a good DEBUG_ERROR message if the HPET HW is not present, so I agree that the ASSERT()s in this PR can be removed without negative side effects.  Just seems like a partial solution to your use cases.

abdattar:
Hi @mdkinney,
Yeah looks like order does matter, I can see the similar behavior in another timer.
https://github.com/tianocore/edk2/blob/master/OvmfPkg/LocalApicTimerDxe/LocalApicTimerDxe.c#L323

Let me revisit and come-up with full solution.
closing the PR for now.
Thanks
AbduL

PR #10650, opened 2025-01-21
OvmfPkg/RiscVVirtQemu.dsc: enable VGA support

xypron:
# Description

- QEMU enables CONFIG_VIRTIO_VGA by default.
- Libvirt falls back from virtio-gpu-pci to virtio-vga if made available by QEMU.

We need to enable VGA support to get video output for EDK II in virt-manager:
- Driver QemuVideoDxe is added to OvmfPkg/RiscVVirt/RiscVVirtQemu.dsc/.fdf

## How This Was Tested

On x86_64 Ubuntu 25.04 (Plucky) I ran the Debian net install image https://cdimage.debian.org/cdimage/trixie_di_alpha1/riscv64/iso-cd/debian-trixie-DI-alpha1-riscv64-netinst.iso.

Emulated GPU output for EDK II and GRUB was only shown after applying the patch.

Used software packages:

* EDK II git head with this patch applied
* libvirt 10.10.0-4ubuntu1
* virtmanager 1:5.0.0-3
* QEMU 1:9.0.2+ds-4ubuntu9

## Integration Instructions

N/A


kraxel (approved):

ardbiesheuvel:
So in addition to virtio-gpu and ramfb, mach-virt on RISC-V also needs VGA emulation (x86 specific obsolete tech) to be able to display a framebuffer? Shouldn't we fix this on the QEMU or libvirt side instead?

xypron:
> So in addition to virtio-gpu and ramfb, mach-virt on RISC-V also needs VGA emulation (x86 specific obsolete tech) to be able to display a framebuffer? Shouldn't we fix this on the QEMU or libvirt side instead?

QEMU with commit 1742054f0bd5 ("hw/riscv: Enable VIRTIO_VGA for RISC-V virt machine") explicitly decided that this GPU emulation should be supported. Users won't understand why QEMU supports virtio-vga, but they get no output from EDK II while Linux runs fine.

For LoongArch the EDK II project merged d6dcf621df22 ("OvmfPkg/LoongArchVirt: Add build file") in 2024 which enables virtio-vga support.

xypron:
> (x86 specific obsolete tech)

There is nothing x86 specific about PCI based GPUs using MMIO for configuration.

kraxel:
> So in addition to virtio-gpu and ramfb, mach-virt on RISC-V also needs VGA emulation (x86 specific obsolete tech) to be able to display a framebuffer? Shouldn't we fix this on the QEMU or libvirt side instead?

It's not that simple.  QemuVirtioDxe handles a wide range of qemu display devices:
* `-device cirrus-vga` -- obsolete even on x86
* `-device qxl` -- vga device with 2d accel (obsolete too).
* `-device VGA` -- qemu standard vga
* `-device bochs-display` -- qemu standard vga minus the vga emulation bits (no text mode, ...), so it essentially provides a simple linear frame buffer in a pci memory bar.
* `-device virtio-vga` -- virtio gpu with vga compatibility.
The last two make sense on riscv too IMHO.

VirtioGpuDxe handles only devices without vga compatibility (aka `-device virtio-gpu-pci`).

ardbiesheuvel:
Fair enough.

ardbiesheuvel (approved):

vlsunil (approved):

PR #10657, opened 2025-01-22
OvmfPkg/QemuKernelLoaderFsDxe: use SIZE_OF_EFI_FILE_INFO

kraxel:
- **OvmfPkg/QemuKernelLoaderFsDxe: use SIZE_OF_EFI_FILE_INFO**

### DIFF
--- a/OvmfPkg/QemuKernelLoaderFsDxe/QemuKernelLoaderFsDxe.c
+++ b/OvmfPkg/QemuKernelLoaderFsDxe/QemuKernelLoaderFsDxe.c
@@ -290,7 +290,7 @@ QemuKernelBlobTypeToFileInfo (
   }
 
   NameSize     = (StrLen (Name) + 1) * 2;
-  FileInfoSize = OFFSET_OF (EFI_FILE_INFO, FileName) + NameSize;
+  FileInfoSize = SIZE_OF_EFI_FILE_INFO + NameSize;
 
   OriginalBufferSize = *BufferSize;
   *BufferSize        = FileInfoSize;


kraxel:
@lersek

MdePkg/Include/Guid/FileInfo.h
```diff
@@ -16,7 +16,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
     0x9576e92, 0x6d3f, 0x11d2, {0x8e, 0x39, 0x0, 0xa0, 0xc9, 0x69, 0x72, 0x3b } \
   }
 
-typedef struct {
+typedef PACKED struct {
```
mdkinney, line 19: This is not the way byte packing is applied to structures in rest of edk2.  #pragma pack() is used and should only be used if the UEFI Spec states that it should be byte packed.

OvmfPkg/QemuKernelLoaderFsDxe/QemuKernelLoaderFsDxe.c
```diff
@@ -291,6 +291,7 @@ QemuKernelBlobTypeToFileInfo (
 
   NameSize     = (StrLen (Name) + 1) * 2;
   FileInfoSize = OFFSET_OF (EFI_FILE_INFO, FileName) + NameSize;
+  ASSERT (FileInfoSize >= sizeof *FileInfo);
```
mdkinney, line 294: use of sizeof() on a structure that uses an array of [1] at the end to indicate a flexible array member is not correct.  Instead, the OFFSET_OF() macro should be used to determine the size of a structure declared in this way.

Of course proper use of flexible array member would be much cleaner.  However, some of the UEFI Spec structures were defined before flexible array members would available in all compilers and C standards.

MdePkg/Include/Guid/FileInfo.h
lersek: @mdkinney Hi Mike, I agree about `#pragma pack (1)` of course. Regarding the spec, members of the structure are already naturally aligned (and the spec requires natural alignment by default), so packing should have no effect other than removing the trailing padding (which is the intent). The wording of the spec, and the comment at the end of `MdePkg/Include/Guid/FileInfo.h`, near `SIZE_OF_EFI_FILE_INFO`, suggest that trailing padding was never intended for this structure.

There are two problems with `EFI_FILE_INFO` having trailing padding (with either issue occurring when the `FileName` member terminates before the trailing padding of `EFI_FILE_INFO` does):
- The `Size` member may be set to a value smaller than `sizeof (EFI_FILE_INFO)`, which is minimally weird (`Size` should never indicate truncation).
- If `EFI_FILE_PROTOCOL.GetInfo(... &gEfiFileInfoGuid ...)` is invoked with a zero `BufferSize` at first, and the output `BufferSize` is less than `sizeof (EFI_FILE_INFO)`, then the subequent allocation will not provide enough space for a whole `EFI_FILE_INFO` object. Subsequently, accesses like `FileInfo->Attribute`, which is equivalent to `(*FileInfo).Attribute`, may violate the compiler's expectation that `FileInfo` point at an object with at least `sizeof *FileInfo` bytes.

OvmfPkg/QemuKernelLoaderFsDxe/QemuKernelLoaderFsDxe.c
lersek: @mdkinney Hi Mike, `sizeof *FileInfo` is intentional here. The ASSERT() expresses that the invariable parts of `EFI_FILE_INFO`, plus the variable/flexible ("struct hack") `FileName` mmber, are *not shorter* than the size of `EFI_FILE_INFO`. In other words, that `EFI_FILE_INFO` does not get truncated due to a short `FileName`.

> Of course proper use of flexible array member would be much cleaner. However, some of the UEFI Spec structures were defined before flexible array members would available in all compilers and C standards.

This is (independently) highly relevant, as the UEFI spec actually declares `EFI_FILE_INFO.FileName` as a (C99) flexible array member, but edk2 declares the same member as a 1-element array (C89 struct hack).

ardbiesheuvel:
Struct packing also has impact on alignment, and on some architectures, the codegen will be much worse because the compiler will have to assume that a EFI_FILE_INFO may appear misaligned in memory.

Could we simply bring EFI_FILE_INFO in line with the spec, and turn it into a true flex array?

mdkinney:
I see the updated commit to change FileName[] to a flexible array member, which makes sizeof (EFI_FILE_INFO) a smaller value.  The entire repo needs to be reviewed for use of sizeof (EFI_FILE_INFO).  There is a macro called SIZE_OF_EFI_FILE_INFO that is used in some places and that is correct.  With the change to a flexible array member, SIZE_OF_EFI_FILE_INFO can be changed from using OFFSET_OF to using sizeof().

kraxel:
> I see the updated commit to change FileName[] to a flexible array member, which makes sizeof (EFI_FILE_INFO) a smaller value. The entire repo needs to be reviewed for use of sizeof (EFI_FILE_INFO). There is a macro called SIZE_OF_EFI_FILE_INFO that is used in some places and that is correct. With the change to a flexible array member, SIZE_OF_EFI_FILE_INFO can be changed from using OFFSET_OF to using sizeof().

Makes sense, I'll have a look next week.

lersek:
The v2 patch set and the related comments silently assume that the offset of the flexible array member equals the size of the structure; in other words, that using a flexible array member in `EFI_FILE_INFO` satisfies the `ASSERT` in question, regardless of how short `FileName` is in practice.

However, ISO C does not guarantee this. (The wording of the standard has changed over time regarding this, but both C99 (which introduced the flexible array member) and C17 express this, albeit with different language. I've not checked C11.)  Namely, C99 writes (6.7.2.1p16):

> the size of the structure shall be equal to the offset of the last element of an otherwise identical structure that replaces the flexible array member with an array of unspecified length. <sup>106</sup>

> 106) The length is unspecified to allow for the fact that implementations may give array members different alignments according to their lengths.

and right after this passage, C99 gives an example (6.7.2.1p17) that clarifies that `sizeof(EFI_FILE_INFO)` equaling `offsetof(EFI_FILE_INFO, FileName)` assumes that "all array members are aligned the same".

In C17, the wording is different (6.7.2.1p18):

> the size of the structure is as if the flexible array member were omitted except that it may have more trailing padding than the omission would imply

and the example in 6.7.2.1p22 clarifies that `sizeof (EFI_FILE_INFO) >= offsetof (EFI_FILE_INFO, FileName)`. In other words, equality is allowed, but not guaranteed.

In total, unless we can force all the edk2 compilers to avoid extraneous padding, I suggest we combine the flexible array member with a static assertion:
~~~
STATIC_ASSERT (
  sizeof (EFI_FILE_INFO) == OFFSET_OF (EFI_FILE_INFO, FileName),
  "flexible array member introduces unwanted padding"
  );
~~~

ardbiesheuvel:
Hi Laszlo,

I see your point in principle, but EFI_FILE_INFO does not have any trailing padding, and so I'm not sure this is needed.

mdkinney:
Would an example where extra padding could be added be the following?  Size without flexible array member would be 1 byte.  Size with flexible array member with natural alignment enforced would be 8 bytes.

```c
struct {
  UINT8  Byte;
  UINT64  FlexibleArray[];
}
```

ardbiesheuvel:
> Would an example where extra padding could be added be the following? Size without flexible array member would be 1 byte. Size with flexible array member with natural alignment enforced would be 8 bytes.
> 
> ```c
> struct {
>   UINT8  Byte;
>   UINT64  FlexibleArray[];
> }
> ```

I haven't seen the example Laszlo mentions, but what I had in mind is the other way around:

```c
struct foo {
    long l;
    int s;
    char c[];
} s;
```

where the size of the structure is 16, due to the padding, but the offset of the flexarray is 12.

[Godbolt link](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:2,endLineNumber:9,positionColumn:2,positionLineNumber:9,selectionStartColumn:2,selectionStartLineNumber:9,startColumn:2,startLineNumber:9),source:'struct+foo+%7B%0A++++long+l%3B%0A++++int+s%3B%0A++++char+c%5B%5D%3B%0A%7D+s%3B%0A%0Aint+bar(void)+%7B%0A++++return+__builtin_offsetof(struct+foo,+c)%3B%0A%7D'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:49.880523255813955,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:cclang_trunk,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'1',libraryCode:'1',trim:'0',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'-O2+-fno-builtin+-ffreestanding',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+clang+(trunk)+(Editor+%231)',t:'0')),k:34.129844961240316,l:'4',m:56.412995320120615,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+clang+(trunk)+(Compiler+%231)',t:'0')),header:(),l:'4',m:43.587004679879385,n:'0',o:'',s:0,t:'0')),k:50.11947674418606,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

lersek:
On a tangent (since we're ultimately discussing `EFI_FILE_PROTOCOL.GetInfo()` information types): the same issue exists with [`EFI_FILE_SYSTEM_INFO`](https://uefi.org/specs/UEFI/2.11/13_Protocols_Media_Access.html#efi-file-system-info) and [`EFI_FILE_SYSTEM_VOLUME_LABEL`](https://uefi.org/specs/UEFI/2.11/13_Protocols_Media_Access.html#efi-file-system-volume-label). The spec declares either of those with flexible array members, but edk2 uses a 1-element trailing member array in each.

... And then there's an extra twist with `EFI_FILE_SYSTEM_VOLUME_LABEL`: unfortunately, it's unfixable (in standard C, anyway). A structure must have at least one *other* member than the flexible array member. Both C99 and C17 say,

> As a special case, the last element of a structure <ins>with more than one named member</ins> may have an incomplete array type; this is called a _flexible array member_.

(Emphasis mine.) Because the sole member of `EFI_FILE_SYSTEM_VOLUME_LABEL` is `VolumeLabel`, the [type definition](https://uefi.org/specs/UEFI/2.11/13_Protocols_Media_Access.html#efi-file-system-volume-label) in the UEFI spec isn't standard C.

lersek:
The example from C99 6.7.2.1p17 (underscore emphasis mine):

> EXAMPLE <ins>Assuming that all array members are aligned the same</ins>, after the declarations:
> ~~~
> struct s { int n; double d[]; };
> struct ss { int n; double d[1]; };
> ~~~
> the three expressions:
> ~~~
> sizeof (struct s)
> offsetof(struct s, d)
> offsetof(struct ss, d)
> ~~~
> have the same value. The structure `struct s` has a flexible array member `d`.

The example from C17 6.7.2.1p20 and p22:

> **EXAMPLE 2** After the declaration:
> ~~~
> struct s { int n; double d[]; };
> ~~~
> the structure `struct s` has a flexible array member `d`. [...]

> After the further declaration:
> ~~~
> struct ss { int n; };
> ~~~
> the expressions:
> ~~~
> sizeof (struct s) >= sizeof (struct ss)
> sizeof (struct s) >= offsetof(struct s, d)
> ~~~
> are always equal to 1.

lersek:
Hi Mike, Ard,

each of your examples introduces additional padding; the specialty of Ard's example is that the extra padding is so large that it can accommodate at least one element in the flexible array member.

C17 seems to have an example that's relevant for this special case too (C17 6.7.2.1p21); I am trimming the quote for relevance:

> Following the above declaration:
> ~~~
> struct s t1 = { 0 }; // valid
> [...]
> t1.n = 4;            // valid
> t1.d[0] = 4.2;       // might be undefined behavior
> ~~~
> [...] The assignment to `t1.d[0]` is probably undefined behavior, but it is possible that
> ~~~
> sizeof (struct s) >= offsetof(struct s, d) + sizeof (double)
> ~~~
> in which case the assignment would be legitimate. Nevertheless, it cannot appear in strictly conforming code.

ardbiesheuvel (approved):
Given that this is orthogonal to the discussion going on in parallel about whether or not the filename field should be a true flexarray as per the EFI spec, I'm happy to merge this now.

mdkinney:
So the current commit is really just a cosmetic change to use the following macro.  Correct?

```
#define SIZE_OF_EFI_FILE_INFO  OFFSET_OF (EFI_FILE_INFO, FileName)
```

kraxel:
> So the current commit is really just a cosmetic change to use the following macro. Correct?
> 
> ```
> #define SIZE_OF_EFI_FILE_INFO  OFFSET_OF (EFI_FILE_INFO, FileName)
> ```
> 
> And the change is in OvmfPkg only, so the PR title should be updated

Yes.  synced PR title with patch subject.

lersek (approved):

mdkinney (approved):

PR #10658, opened 2025-01-22
MdeModulePkg PeiMain: Remove return for PeiCoreBuildHobHandoffInfoTable()

lzeng14:
PeiCoreBuildHobHandoffInfoTable() always returns EFI_SUCCESS, and also its return is not consumed at all, so this patch removes @return for PeiCoreBuildHobHandoffInfoTable().


lzeng14:
@lgao4 @mdkinney
Liming and Mike, could you help review this patch ? Or may any other reviewers need to be also added ?

os-d (approved):

mdkinney (approved):

PR #10660, opened 2025-01-23
EmulatorPkg: Move magic page to first allocation

### LINKED ISSUE #10652: [Bug]: EmulatorPkg Page Allocation Failure

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### Code first?

- [ ] Yes

EmulatorPkg

RELEASE

### Current Behavior

The Visual Studio IA32 `EmulatorPkg` build fails to boot sometimes failing to allocate a specific range of virtual memory on the host system:

```
ERROR : Could not allocate PeiServicesTablePage @ 13000000
```

From CI run: [BUILDLOG_EmulatorPkg_With_Run.txt](https://github.com/user-attachments/files/18499263/BUILDLOG_EmulatorPkg_With_Run.txt)

### Expected Behavior

The build should boot without an allocation failure.

### Steps To Reproduce

1. Build the EmulatorPkg IA32 RELEASE image:
   `stuart_build -c EmulatorPkg/PlatformCI/PlatformBuild.py TOOL_CHAIN_TAG=VS2022 TARGET=RELEASE -a IA32`
2. Run to shell:
   `stuart_build -c EmulatorPkg/PlatformCI/PlatformBuild.py TOOL_CHAIN_TAG=VS2022 TARGET=RELEASE -a IA32   MAKE_STARTUP_NSH=TRUE --FlashOnly`

Note: It's been very difficult to reproduce locally, but it has happened. When it did, running the same executable would have different results (no compilation change).

### Build Environment

```markdown
- OS(s): Windows 11 (and Windows Server 2022)
- Tool Chain(s): VS2022
```

### Version Information

```text
edk2 master tip. Commit f6afd876631e2f7d1e0eeda70c2bf299c5d83deb.
```

### Urgency

Medium

Someone else needs to fix it

Maintainer feedback requested

### Anything else?

Similar issue in the past that lead to modifying the `PcdPeiServicesTablePage` PCD value https://github.com/tianocore/edk2/commit/868254ace3a0729c9ed6eb1a8646f64c77ee84f9.

---

The value needs to be FixedAtBuild because modules need the value before dynamic PCD services are available. We've discussed potentially using a global resource within the user mode app to hold a dynamically allocated address as opposed to the fixed PCD value.

mdkinney:
# Description

Fixes #10652 

The magic page is allocated from a fixed address specified by `PcdPeiServicesTablePage`. This allocation has been observed to sometimes fail. Move the allocation of this buffer to the very beginning of `main()` to minimize the chance that another allocation is allocated from the `PcdPeiServicesTablePage` address.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

The issue was reproduced with a build of EmulatorPkg VS2022 IA32 RELEASE. In order to minimize the time required to boot cycle the EmulatorPkg build, the boot manager front page timeout is set to 0 and the Shell default delay is also set to 0.

```
build -a IA32 -t VS2022 -p EmulatorPkg\EmulatorPkg.dsc -b RELEASE --pcd gEfiMdePkgTokenSpaceGuid.PcdPlatformBootTimeOut=0 --pcd gEfiShellPkgTokenSpaceGuid.PcdShellDefaultDelay=0
```

A cmd script was then used to boot the EmulatorPkg in a loop and log the results to a file.

```cmd
@echo off
:loop
WinHost.exe
goto loop
```

A `startup.nsh` file is used to automatically reset the EmulatorPkg when the boot to Shell completes

```
BOOT SUCCESS !!!
reset
```
Within 10 to 20 boot cycles the allocation of the magic page fails.

With the fix in this PR that moves the allocation of the magic page to very beginning of `main()` the issue was not reproduced in over 11000 boot cycles of the EmulatorPkg.

## Integration Instructions

N/A

### DIFF
--- a/EmulatorPkg/Win/Host/WinHost.c
+++ b/EmulatorPkg/Win/Host/WinHost.c
@@ -453,6 +453,28 @@ Returns:
   EMU_THUNK_PPI        *SecEmuThunkPpi;
 
   //
+  // If enabled use the magic page to communicate between modules
+  // This replaces the PI PeiServicesTable pointer mechanism that
+  // deos not work in the emulator. It also allows the removal of
+  // writable globals from SEC, PEI_CORE (libraries), PEIMs
+  //
+  EmuMagicPage = (VOID *)(UINTN)(FixedPcdGet64 (PcdPeiServicesTablePage) & MAX_UINTN);
+  if (EmuMagicPage != NULL) {
+    UINT64  Size;
+    Status = WinNtOpenFile (
+               NULL,
+               SIZE_4KB,
+               0,
+               &EmuMagicPage,
+               &Size
+               );
+    if (EFI_ERROR (Status)) {
+      SecPrint ("ERROR : Could not allocate PeiServicesTablePage @ %p\n\r", EmuMagicPage);
+      return EFI_DEVICE_ERROR;
+    }
+  }
+
+  //
   // Enable the privilege so that RTC driver can successfully run SetTime()
   //
   OpenProcessToken (GetCurrentProcess (), TOKEN_ADJUST_PRIVILEGES|TOKEN_QUERY, &Token);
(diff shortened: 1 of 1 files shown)


niruiyu (approved):
Mike, Thanks for fixing this.
I sometimes met this issue and just cannot boot. Then I have to change the PcdPeiServicesTablePage value to workaround it.

PR #10667, opened 2025-01-24
OvmfPkg/X64: add opt/org.tianocore/UninstallMemAttrProtocol support

kraxel:
Add support for opt/org.tianocore/UninstallMemAttrProtocol, to allow
turning off EFI_MEMORY_ATTRIBUTE_PROTOCOL, simliar to ArmVirtPkg.

# rant

I *hate* to do this, but things are breaking left and right.  This time
it is not shim, 15.8 works fine and almost everyone has rolled out
updates meanwhile.  But now grub gets the NX stuff wrong, and following
that the linux kernel efi stub pagefaults.

Apparently it happens due something in the huge pile of distro
downstream patches doing some stupid legacy x86 linux things.

Vanilla 2.12 grub is doing fine.
aarch64 is doing fine too.
Even `chainloader $vmlinuz` works.
Only `linux $vmlinux` on x64 distro builds explodes ...

### DIFF
--- a/OvmfPkg/Library/PlatformBootManagerLib/BdsPlatform.c
+++ b/OvmfPkg/Library/PlatformBootManagerLib/BdsPlatform.c
@@ -1837,6 +1837,49 @@ SaveS3BootScript (
 }
 
 /**
+  Uninstall the EFI memory attribute protocol if it exists.
+**/
+STATIC
+VOID
+UninstallEfiMemoryAttributesProtocol (
+  VOID
+  )
+{
+  EFI_STATUS  Status;
+  EFI_HANDLE  Handle;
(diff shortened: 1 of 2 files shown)


ardbiesheuvel:
Is there a bug report about this for GRUB?

kraxel:
> Is there a bug report about this for GRUB?

https://issues.redhat.com/browse/RHEL-74341

Note it's not limited to redhat, I've seen suse and debian builds fail too on a quick test.  Probably because the distros copy broken patches from each other due to grub upstreaming being slow.

ardbiesheuvel:
I'll consider this once this is fully diagnosed and fixed on the distro side.

Issue #10289, opened 2022-10-26
ArmPkg should be deleted (Bugzilla Bug 4121)


tianocore-issues:
This issue was created automatically with [bugzilla2github](https://github.com/makubacki/bugzilla2github/tree/bz2gh_tianocore)
# Bugzilla Bug 4121
Date: 2022-10-26T11:41:17+00:00
From: @leiflindholm
To:   unassigned &lt;<unassigned>&gt;
CC:   

Last updated: 2022-10-26T11:41:17+00:00

tianocore-issues:
## Comment 19845
Date: 2022-10-26 11:41:17 +0000
From: @leiflindholm

- Industry Specification: ---
- Release Observed: EDK II Master
- Releases to Fix: EDK II Master
- Target OS: ---
- Bugzilla Assignee(s): unassigned &lt;<unassigned>&gt;

ArmPkg was created as a separate package since it was originally a proof-of concept new architecture, and has remained in that structure ever since.

However, this frequently causes conflicts with the requirements of MdePkg and MdeModulePkg to not have external dependencies, making the addition of some new features on ARM* a higher effort than X86.

As a result, over time we want to migrate core-architectural modules out of ArmPkg to MdePkg/MdeModulePkg(/UefiCpuPkg?), and anything non-core-architectural to other new homes.

This bug serves as a central point for this effort, and will be closed when ArmPkg is deleted.

PR #10671, opened 2025-01-24
Decouple package dependencies of StandaloneMmPkg

kuqin12:
# Description

This change moves the `StandaloneMmPeCoffExtraActionLib` from StandaloneMmPkg to ArmPkg because this library is specific to ARM platforms and the only consumer is in the ArmPkg.

In addition, a null instance of the library was created to satisfy CI and pipeline builds for various architectures without coupled package dependencies

- [x] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This change does not impact any functionality.

## Integration Instructions

Platforms need to update their `StandaloneMmPeCoffExtraActionLib` instance to point to `ArmPkg/Library/StandaloneMmPeCoffExtraActionLib/StandaloneMmPeCoffExtraActionLib.inf`


ArmPkg/ArmPkg.dsc
```diff
@@ -169,6 +169,7 @@
 
 [Components.common.MM_CORE_STANALONE]
```
os-d, line 170: I'm not sure why this isn't caught by the parser, but I don't believe `Components.common.<MODULE_TYPE>` is valid. I don't know if the parser will ignore or still build. Can you remove the misspelled `MM_CORE_STANALONE` here and `MM_STANDALONE` below and just put these in the standard components section?

os-d (approved):

ArmPkg/ArmPkg.dsc
kuqin12: Removed both section headers.

leiflindholm (changes requested):
We're seeing a highly unfortunate flurry of cleaning up StandaloneMmPkg by chucking bits over to ArmPkg. As per https://github.com/tianocore/edk2/issues/10289, ArmPkg should be shrinking, not growing.

What is the problem that is being solved by this?

kuqin12:
> We're seeing a highly unfortunate flurry of cleaning up StandaloneMmPkg by chucking bits over to ArmPkg. As per #10289, ArmPkg should be shrinking, not growing.
> 
> What is the problem that is being solved by this?

Thanks for the comment, @leiflindholm. I do not disagree with your direction. But with the current case, the Standalone MM package only carries a single `PeCoffExtraActionLib` instance for AArch64 platforms, and its only consumer lives in the ArmPkg. The `StandaloneMmPeCoffExtraActionLib` seems more like an outlier and unnecessary dependency given the code layout.

In addition, the dsc is building the standalone mm core for AArch64 platform by pulling in the ArmPkg entrypoints. But component build check is already done at their corresponding package CI process. Instead of pulling in the package dependency, introducing a null library to suffice the CI check seems to be a more proper way to go.

I understand your position of not inflating the ArmPkg, but I think it would be worthy if moving this library instance can create a clean cut between these packages. Please let me know if any thoughts.

leiflindholm:
The only reason for moving this out of StandaloneMmPkg is some idea of purity of packages.

I'm 100% fine with not bothering with doing that. But if you really feel strongly about that, you should also care about not just moving the impurity around, but actually cleaning up the legacy infelicities. If you're interested in that, let's talk.

For this approach, NAK.

kuqin12:
> The only reason for moving this out of StandaloneMmPkg is some idea of purity of packages.
> 
> I'm 100% fine with not bothering with doing that. But if you really feel strongly about that, you should also care about not just moving the impurity around, but actually cleaning up the legacy infelicities. If you're interested in that, let's talk.
> 
> For this approach, NAK.

Okay, sure. I am all for being constructive :) I am good with moving pieces to MdeModulePkg, MdePkg, or UefiCpuPkg. When looking at the mmu lib, what I am thinking is to transform it into the ARM version of page table lib like in the Cpu package.

Then its arm specific dependencies need to move accordingly. CacheMaintenanceLib is a smaller potato. The hard pull will be ArmLib itself. I do not want it to be in the MdeModulePkg or MdePkg wholesale. But any suggestion on how to dissect it and where to move them to?

niruiyu (approved):

niruiyu:
Just saw discussion above, will not add "push" label until the discussion ends.

kuqin12:
> > The only reason for moving this out of StandaloneMmPkg is some idea of purity of packages.
> > I'm 100% fine with not bothering with doing that. But if you really feel strongly about that, you should also care about not just moving the impurity around, but actually cleaning up the legacy infelicities. If you're interested in that, let's talk.
> > For this approach, NAK.
> 
> Okay, sure. I am all for being constructive :) I am good with moving pieces to MdeModulePkg, MdePkg, or UefiCpuPkg. When looking at the mmu lib, what I am thinking is to transform it into the ARM version of page table lib like in the Cpu package.
> 
> Then its arm specific dependencies need to move accordingly. CacheMaintenanceLib is a smaller potato. The hard pull will be ArmLib itself. I do not want it to be in the MdeModulePkg or MdePkg wholesale. But any suggestion on how to dissect it and where to move them to?

@leiflindholm gentle ping on this. Please let me know how think on where to place/dismantle ArmLib.

PR #10679, opened 2025-01-27
Platform hook notify

dyap0:
# Description

Provide PlatformHook NotifyProtocolGuid before and after SMM dispatch. This replaces the SMM core platform hook implementation in the traditional SMM (Edk2\MdeModulePkg\Include\Library\SmmCorePlatformHookLib.h). Unlike traditional SMM, Standalone MM prefers a more generic method to perform platform-specific tasks before and after the SMI handler. Therefore, it introduces the gMmEntryNotifyProtocolGuid and gMmExitNotifyProtocolGuid protocols to notify the SMM entry and exit. This approach is more flexible for the platform, eliminating the need to place all tasks in PlatformHookBeforeSmmDispatch() and PlatformHookAfterSmmDispatch(), thus platform code can depend on the protocol notification to perform platform-specific tasks.

## How This Was Tested

Manual test case run from platform code by hook on NotifyProtocolGuid.

## Integration Instructions

N/A


ardbiesheuvel:
Hello Diat,

> # Description
> Provide PlatformHook NotifyProtocolGuid before and after SMM dispatch.
> 

We can tell from the code that this is _what_ your series does. You will also need to describe which problem it solves, and why this is the right way to solve it. In this case, you will also need to explain how these changes are supported by the PI specification, given that you are adding protocol definitions to MdePkg.

Thanks,

MdePkg/Include/Protocol/MmConfiguration.h
```diff
@@ -59,5 +59,7 @@ struct _EFI_MM_CONFIGURATION_PROTOCOL {
 };
 
 extern EFI_GUID  gEfiMmConfigurationProtocolGuid;
+extern EFI_GUID  gEfiMmEntryNotifyProtocolGuid;
+extern EFI_GUID  gEfiMmExitNotifyProtocolGuid;
```
jiaxinwu, line 63: No need define it in .h, move to Edk2\StandaloneMmPkg\StandaloneMmPkg.dec directly, refer to:
gEventMmDispatchGuid

jiaxinwu:
> Hello Diat,
> 
> > # Description
> > Provide PlatformHook NotifyProtocolGuid before and after SMM dispatch.
> 
> We can tell from the code that this is _what_ your series does. You will also need to describe which problem it solves, and why this is the right way to solve it. In this case, you will also need to explain how these changes are supported by the PI specification, given that you are adding protocol definitions to MdePkg.
> 
> Thanks,

This replaces the SMM core platform hook implementation in the traditional SMM (Edk2\MdeModulePkg\Include\Library\SmmCorePlatformHookLib.h). Unlike traditional SMM, Standalone MM prefers a more generic method to perform platform-specific tasks before and after the SMI handler. Therefore, it introduces the gMmEntryNotifyProtocolGuid and gMmExitNotifyProtocolGuid protocols to notify the  SMM entry and exit. This approach is more flexible for the platform, eliminating the need to place all tasks in PlatformHookBeforeSmmDispatch() and PlatformHookAfterSmmDispatch(), thus platform code can depend on the protocol notification to perform platform-specific tasks.

MdePkg/Include/Protocol/MmConfiguration.h
dyap0: done

StandaloneMmPkg/Core/StandaloneMmCore.c
```diff
@@ -505,6 +505,7 @@ MmEntryPoint (
   EFI_MM_COMMUNICATE_HEADER  *CommunicateHeader;
   MM_COMM_BUFFER_STATUS      *CommunicationStatus;
   UINTN                      BufferSize;
+  EFI_HANDLE                 MmDispatchHandle;
```
jiaxinwu, line 508: rename it to MmHandle.

jiaxinwu (commented):
"StandaloneMmPkg: Created StandaloneMM Entry/Exit Notify Protocol."

"Created StandaloneMM Entry/Exit Notify Protocol for PlatformHook."

==>

Use "Produce" to replace the "Created"

StandaloneMmPkg/Core/StandaloneMmCore.c
dyap0: Ok updated

dyap0:
> "StandaloneMmPkg: Created StandaloneMM Entry/Exit Notify Protocol."
> 
> "Created StandaloneMM Entry/Exit Notify Protocol for PlatformHook."
> 
> ==>
> 
> Use "Produce" to replace the "Created"

jiaxinwu (approved):

niruiyu (approved):

niruiyu (approved):

PR #10680, opened 2025-01-27
NetworkPkg:Resolved Coverity Issues in MnpDxe

santhoshkumarvenkatachalam:
Resolved INTEGER_OVERFLOW Coverity Issues in MNP Dxe 1.MnpStop,MnpInstanceDeliverPacket
Expression "MnpDeviceData->ConfiguredChildrenNumber--" and Instance->RcvdPacketQueueSize-- where Both variables are known to be equal to 0,underflows the type that receives it.

# Description

### DIFF
--- a/NetworkPkg/MnpDxe/MnpConfig.c
+++ b/NetworkPkg/MnpDxe/MnpConfig.c
@@ -1305,6 +1305,10 @@ MnpStop (
   MnpDeviceData = MnpServiceData->MnpDeviceData;
   ASSERT (MnpDeviceData->ConfiguredChildrenNumber > 0);
 
+  if (MnpDeviceData->ConfiguredChildrenNumber <= 0) {
+    return EFI_OUT_OF_RESOURCES;
+  }
+
   //
   // Configure the receive filters.
   //
--- a/NetworkPkg/MnpDxe/MnpIo.c
+++ b/NetworkPkg/MnpDxe/MnpIo.c
@@ -325,6 +325,9 @@ MnpInstanceDeliverPacket (
   }
 
   ASSERT (Instance->RcvdPacketQueueSize != 0);
+  if (Instance->RcvdPacketQueueSize <= 0) {
+    return EFI_OUT_OF_RESOURCES;
+  }
 
   RxDataWrap = NET_LIST_HEAD (&Instance->RcvdPacketQueue, MNP_RXDATA_WRAP, WrapEntry);
   if (RxDataWrap->Nbuf->RefCnt > 2) {


NetworkPkg/MnpDxe/MnpConfig.c
```diff
@@ -1305,6 +1305,10 @@ MnpStop (
   MnpDeviceData = MnpServiceData->MnpDeviceData;
   ASSERT (MnpDeviceData->ConfiguredChildrenNumber > 0);
 
+  if (MnpDeviceData->ConfiguredChildrenNumber <= 0) {
```
leiflindholm, line 1308: ConfiguredChildrenNumber is UINTN - does this change not generate a new Coverity warning?

NetworkPkg/MnpDxe/MnpConfig.c
```diff
@@ -1305,6 +1305,10 @@ MnpStop (
   MnpDeviceData = MnpServiceData->MnpDeviceData;
   ASSERT (MnpDeviceData->ConfiguredChildrenNumber > 0);
 
+  if (MnpDeviceData->ConfiguredChildrenNumber <= 0) {
+    return EFI_OUT_OF_RESOURCES;
```
leiflindholm, line 1309: This does not feel like an appropriate return code.

Also, if we look at the call path, if we return an error here, we will fall back all the way to MnpConfigure which will itself pass this return value on.

But the documentation for MnpConfigure says that for "Others" error codes (as well as for EFI_OUT_OF_RESOURCES), "The MNP child driver instance has been reset to startup defaults.".
But I don't see anything like that happening as a result of this patch.
And given this error could only ever be triggered if the driver state ended up inconcistent, that would feel like an important thing to do.

NetworkPkg/MnpDxe/MnpIo.c
```diff
@@ -325,6 +325,9 @@ MnpInstanceDeliverPacket (
   }
 
   ASSERT (Instance->RcvdPacketQueueSize != 0);
+  if (Instance->RcvdPacketQueueSize <= 0) {
```
leiflindholm, line 328: Also UINTN.

NetworkPkg/MnpDxe/MnpIo.c
```diff
@@ -325,6 +325,9 @@ MnpInstanceDeliverPacket (
   }
 
   ASSERT (Instance->RcvdPacketQueueSize != 0);
+  if (Instance->RcvdPacketQueueSize <= 0) {
+    return EFI_OUT_OF_RESOURCES;
```
leiflindholm, line 329: This does not feel like an appropriate return value on finding that the packet queue is empty and there is nothing to deliver.

PR #10685, opened 2025-01-27
Add 18 register support for SMC and SVC calls for AArch64 platforms

kuqin12:
# Description

Current SMC and SVC only supports up to 8 registers.

However, the FF-A spec has definition for FFA direct request 2 which could use up to 18 registers.

This change is added to support such use case.

Note:
https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/36018 must be paired to make sure SPMC at EL3 to work properly.

- [x] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested on QEMU SBSA and booted to Windows

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Include/Library/ArmSmcLib.h
+++ b/ArmPkg/Include/Library/ArmSmcLib.h
@@ -23,6 +23,16 @@ typedef struct {
   UINTN    Arg5;
   UINTN    Arg6;
   UINTN    Arg7;
+  UINTN    Arg8;
+  UINTN    Arg9;
+  UINTN    Arg10;
+  UINTN    Arg11;
+  UINTN    Arg12;
+  UINTN    Arg13;
+  UINTN    Arg14;
+  UINTN    Arg15;
+  UINTN    Arg16;
+  UINTN    Arg17;
 } ARM_SMC_ARGS;
 
 /**
--- a/ArmPkg/Include/Library/ArmSvcLib.h
+++ b/ArmPkg/Include/Library/ArmSvcLib.h
@@ -22,6 +22,16 @@ typedef struct {
   UINTN    Arg5;
   UINTN    Arg6;
   UINTN    Arg7;
+  UINTN    Arg8;
+  UINTN    Arg9;
+  UINTN    Arg10;
+  UINTN    Arg11;
+  UINTN    Arg12;
+  UINTN    Arg13;
+  UINTN    Arg14;
+  UINTN    Arg15;
+  UINTN    Arg16;
+  UINTN    Arg17;
 } ARM_SVC_ARGS;
 
 /**
--- a/ArmPkg/Library/ArmFfaLib/ArmFfaCommon.c
+++ b/ArmPkg/Library/ArmFfaLib/ArmFfaCommon.c
@@ -25,6 +25,7 @@
 
 #include <IndustryStandard/ArmFfaSvc.h>
 #include <IndustryStandard/ArmFfaPartInfo.h>
+#include <IndustryStandard/ArmStdSmc.h>
 
 #include "ArmFfaCommon.h"
 
@@ -695,14 +696,24 @@ ArmFfaLibMsgSendDirectReq2 (
 
   ZeroMem (&FfaArgs, sizeof (ARM_FFA_ARGS));
 
-  FfaArgs.Arg0 = ARM_FID_FFA_MSG_SEND_DIRECT_REQ2;
-  FfaArgs.Arg1 = PACK_PARTITION_ID_INFO (gPartId, DestPartId);
-  FfaArgs.Arg2 = Uuid[0];
-  FfaArgs.Arg3 = Uuid[1];
-  FfaArgs.Arg4 = ImpDefArgs->Arg0;
-  FfaArgs.Arg5 = ImpDefArgs->Arg1;
-  FfaArgs.Arg6 = ImpDefArgs->Arg2;
-  FfaArgs.Arg7 = ImpDefArgs->Arg3;
+  FfaArgs.Arg0  = ARM_FID_FFA_MSG_SEND_DIRECT_REQ2;
+  FfaArgs.Arg1  = PACK_PARTITION_ID_INFO (gPartId, DestPartId);
+  FfaArgs.Arg2  = Uuid[0];
+  FfaArgs.Arg3  = Uuid[1];
+  FfaArgs.Arg4  = ImpDefArgs->Arg0;
+  FfaArgs.Arg5  = ImpDefArgs->Arg1;
+  FfaArgs.Arg6  = ImpDefArgs->Arg2;
+  FfaArgs.Arg7  = ImpDefArgs->Arg3;
+  FfaArgs.Arg8  = ImpDefArgs->Arg4;
+  FfaArgs.Arg9  = ImpDefArgs->Arg5;
+  FfaArgs.Arg10 = ImpDefArgs->Arg6;
+  FfaArgs.Arg11 = ImpDefArgs->Arg7;
+  FfaArgs.Arg12 = ImpDefArgs->Arg8;
+  FfaArgs.Arg13 = ImpDefArgs->Arg9;
+  FfaArgs.Arg14 = ImpDefArgs->Arg10;
+  FfaArgs.Arg15 = ImpDefArgs->Arg11;
+  FfaArgs.Arg16 = ImpDefArgs->Arg12;
+  FfaArgs.Arg17 = ImpDefArgs->Arg13;
 
   ArmCallFfa (&FfaArgs);
 
@@ -711,10 +722,20 @@ ArmFfaLibMsgSendDirectReq2 (
     return Status;
   }
 
-  ImpDefArgs->Arg0 = FfaArgs.Arg4;
-  ImpDefArgs->Arg1 = FfaArgs.Arg5;
-  ImpDefArgs->Arg2 = FfaArgs.Arg6;
-  ImpDefArgs->Arg3 = FfaArgs.Arg7;
+  ImpDefArgs->Arg0  = FfaArgs.Arg4;
+  ImpDefArgs->Arg1  = FfaArgs.Arg5;
+  ImpDefArgs->Arg2  = FfaArgs.Arg6;
+  ImpDefArgs->Arg3  = FfaArgs.Arg7;
+  ImpDefArgs->Arg4  = FfaArgs.Arg8;
+  ImpDefArgs->Arg5  = FfaArgs.Arg9;
+  ImpDefArgs->Arg6  = FfaArgs.Arg10;
+  ImpDefArgs->Arg7  = FfaArgs.Arg11;
+  ImpDefArgs->Arg8  = FfaArgs.Arg12;
+  ImpDefArgs->Arg9  = FfaArgs.Arg13;
+  ImpDefArgs->Arg10 = FfaArgs.Arg14;
+  ImpDefArgs->Arg11 = FfaArgs.Arg15;
+  ImpDefArgs->Arg12 = FfaArgs.Arg16;
+  ImpDefArgs->Arg13 = FfaArgs.Arg17;
 
   return EFI_SUCCESS;
 }
@@ -733,11 +754,30 @@ ArmFfaLibCommonInit (
   IN VOID
   )
 {
-  EFI_STATUS  Status;
-  UINT16      CurrentMajorVersion;
-  UINT16      CurrentMinorVersion;
+  EFI_STATUS    Status;
+  UINT16        CurrentMajorVersion;
+  UINT16        CurrentMinorVersion;
+  ARM_FFA_ARGS  FfaArgs;
 
   gFfaSupported = FALSE;
+
+  ZeroMem (&FfaArgs, sizeof (ARM_SMC_ARGS));
+  FfaArgs.Arg0 = SMCCC_VERSION;
+  ArmCallFfa (&FfaArgs);
+  if ((INT32)FfaArgs.Arg0 < 0) {
+    DEBUG ((DEBUG_ERROR, "%a: SMCCC_VERSION not supported\n", __func__));
+    return EFI_UNSUPPORTED;
+  }
+
+  // According to SMCCC Specification v1.6 G BET0
+  // Table F0-1: Changelog: Starting from SMCCC_VERSION v1.2, the interface
+  // - Permits calls to use R4–R7 as return register
+  // - Permits calls to use X4–X17 as return registers
+  // - Permits calls to use X8–X17 as argument registers
+  if ((INT32)FfaArgs.Arg0 < 0x10002) {
+    DEBUG ((DEBUG_ERROR, "%a: SMCCC_VERSION %x < 1.2\n", __func__, (UINT32)FfaArgs.Arg0));
+    return EFI_UNSUPPORTED;
+  }
 
   Status = ArmFfaLibGetVersion (
              ARM_FFA_MAJOR_VERSION,
--- a/ArmPkg/Library/ArmSmcLib/AArch64/ArmSmc.S
+++ b/ArmPkg/Library/ArmSmcLib/AArch64/ArmSmc.S
@@ -8,10 +8,20 @@
 #include <AsmMacroLib.h>
 
 ASM_FUNC(ArmCallSmc)
-  // Push x0 on the stack - The stack must always be quad-word aligned
-  str   x0, [sp, #-16]!
+  // Push frame pointer and return address on the stack
+  stp   x29, x30, [sp, #-16]!
+  mov   x29, sp
+
+  // x0 is the ARM_SMC_ARGS structure address
+  // x30 is used as a scratch register upon returning from SMC
+  mov   x30, x0
 
   // Load the SMC arguments values into the appropriate registers
+  ldp   x16, x17, [x0, #128]
+  ldp   x14, x15, [x0, #112]
+  ldp   x12, x13, [x0, #96]
+  ldp   x10, x11, [x0, #80]
+  ldp   x8, x9, [x0, #64]
   ldp   x6, x7, [x0, #48]
   ldp   x4, x5, [x0, #32]
   ldp   x2, x3, [x0, #16]
@@ -19,14 +29,18 @@ ASM_FUNC(ArmCallSmc)
 
   smc   #0
 
-  // Pop the ARM_SMC_ARGS structure address from the stack into x9
-  ldr   x9, [sp], #16
-
   // Store the SMC returned values into the ARM_SMC_ARGS structure.
-  // A SMC call can return up to 4 values - we do not need to store back x4-x7.
-  stp   x2, x3, [x9, #16]
-  stp   x0, x1, [x9, #0]
+  // A SMC call can return up to 18 values.
+  stp   x16, x17, [x30, #128]
+  stp   x14, x15, [x30, #112]
+  stp   x12, x13, [x30, #96]
+  stp   x10, x11, [x30, #80]
+  stp   x8, x9, [x30, #64]
+  stp   x6, x7, [x30, #48]
+  stp   x4, x5, [x30, #32]
+  stp   x2, x3, [x30, #16]
+  stp   x0, x1, [x30, #0]
 
-  mov   x0, x9
+  ldp   x29, x30, [sp], #16
 
   ret
--- a/ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c
+++ b/ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c
@@ -726,10 +726,20 @@ SetEventCompleteSvcArgs (
         EventCompleteSvcArgs->Arg0 = ARM_FID_FFA_MSG_SEND_DIRECT_RESP2;
 
         if (FfaMsgInfo->ServiceType == ServiceTypeMisc) {
-          EventCompleteSvcArgs->Arg4 = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg0;
-          EventCompleteSvcArgs->Arg5 = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg1;
-          EventCompleteSvcArgs->Arg6 = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg2;
-          EventCompleteSvcArgs->Arg7 = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg3;
+          EventCompleteSvcArgs->Arg4  = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg0;
+          EventCompleteSvcArgs->Arg5  = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg1;
+          EventCompleteSvcArgs->Arg6  = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg2;
+          EventCompleteSvcArgs->Arg7  = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg3;
+          EventCompleteSvcArgs->Arg8  = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg4;
+          EventCompleteSvcArgs->Arg9  = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg5;
+          EventCompleteSvcArgs->Arg10 = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg6;
+          EventCompleteSvcArgs->Arg11 = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg7;
+          EventCompleteSvcArgs->Arg12 = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg8;
+          EventCompleteSvcArgs->Arg13 = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg9;
+          EventCompleteSvcArgs->Arg14 = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg10;
+          EventCompleteSvcArgs->Arg15 = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg11;
+          EventCompleteSvcArgs->Arg16 = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg12;
+          EventCompleteSvcArgs->Arg17 = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg13;
         }
       }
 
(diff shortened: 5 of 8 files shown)


ArmPkg/Include/Library/ArmSmcLib.h
```diff
@@ -23,6 +25,16 @@ typedef struct {
   UINTN    Arg5;
   UINTN    Arg6;
   UINTN    Arg7;
+  UINTN    Arg8;
```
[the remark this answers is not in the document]
samimujawar: You could also try #if FixedPcdGetBool (PcdSxcUse18Registers)

ArmPkg/Library/ArmSmcLib/AArch64/ArmSmc.S
```diff
@@ -8,25 +8,55 @@
 #include <AsmMacroLib.h>
 
 ASM_FUNC(ArmCallSmc)
-  // Push x0 on the stack - The stack must always be quad-word aligned
+  // Push x0 and x19 on the stack - The stack must always be quad-word aligned
+  str   x19, [sp, #-16]!
   str   x0, [sp, #-16]!
 
   // Load the SMC arguments values into the appropriate registers
+  mov   x19, FixedPcdGetBool (PcdSxcUse18Registers)
```
samimujawar, line 16: I think this code could be replaced with #if FixedPcdGetBool (PcdSxcUse18Registers) ...

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c
```diff
@@ -730,6 +730,18 @@ SetEventCompleteSvcArgs (
           EventCompleteSvcArgs->Arg5 = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg1;
           EventCompleteSvcArgs->Arg6 = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg2;
           EventCompleteSvcArgs->Arg7 = mMiscMmCommunicateBuffer->DirectMsgArgs.Arg3;
+          if (FixedPcdGetBool (PcdSxcUse18Registers)) {
```
samimujawar, line 733: I think it should be possible to use #if FixedPcdGetBool (PcdSxcUse18Registers) ...

samimujawar:
I have posted some comments on how PcdSxcUse18Registers could be used. Can you try if that works, please? 

Also, 
a) what would happen if we make ArmSmcLib and ArmSvcLib support 18 registers by default, and only use PcdSxcUse18Registers to enable/disable the 18 register support in ArmFfaLib / DIRECT_REQ2 implementation?  

b) Would any of this break the compatibility with the SPMC and SPMD being at EL3?

ArmPkg/Include/Library/ArmSmcLib.h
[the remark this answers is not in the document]
kuqin12: @samimujawar thanks for your suggestion. I started with the way you suggested and ended up abandoning that idea. The reason is because including it in the header file, will require all the consumers to include this PCD in the header file, because AutoGen.h does not automatically pull the PCDs from dependent libraries into itself. This will make the usage of this header file more impactful. So I ended up defining the structure to be 18 register anyway, but only use 8 registers in the code if the PCD is set to TRUE/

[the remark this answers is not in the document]
kuqin12: @samimujawar A gentle ping on this. Could you please share your thoughts on the reply above? Thanks.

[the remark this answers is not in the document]
samimujawar: Hi @kuqin12,

I think we need to keep the existing interfaces as is and instead introduce new functions that cover 18 registers.

e.g. Introduce a new function ArmCallSmcEx() and correspondingly a new structure ARM_SMC_ARGS_EX.

The ArmSmcLib implementation does not need to care about the PcdSxcUse18Registers. It is the caller’s responsibility to ensure the right interface function is invoked. The same would apply to ArmSvcLib.

So if a service interface requires use of 18 registers, it will check if PcdSxcUse18Registers == TRUE and if so call ArmFfaLibMsgSendDirectReq2Ex() [which calls ArmCallSmcEx()] otherwise it will return EFI_UNSUPPORTED. 

Following are some examples on how error handling can be done by the service implementation.
Option 1:
// Runtime failure
#ifdef PcdSxcUse18Registers == TRUE
   ArmFfaLibMsgSendDirectReq2Ex()
#else
  return EFI_UNSUPPORTED;
#endif

Option 2:
// Build time failure
#ifdef PcdSxcUse18Registers == TRUE
   ArmFfaLibMsgSendDirectReq2Ex()
#else
  #error “18 register support required for using this service”
#endif

Option 3:
// Fallback if service can work with 8 registers
#ifdef PcdSxcUse18Registers == TRUE
   ArmFfaLibMsgSendDirectReq2Ex()
#else
   ArmFfaLibMsgSendDirectReq2()
#endif

The Platform DSC file can specify if 18 register support is available.
Example 1: When using SPMC &. SPMD @ EL3
// SPMC&SPMD at EL3 
PcdSxcUse18Registers = FALSE

Example 2: When using Hafnium
// Hafnium
PcdSxcUse18Registers = TRUE

Alternatively, the PcdSxcUse18Registers can be specified at the command line.
build -a AARCH64 -p <Path>/Platform.dsc --pcd PcdSxcUse18Registers=1 

I think doing this will ensure better backward compatibility and will not hide errors.

ArmPkg/Library/ArmSmcLib/AArch64/ArmSmc.S
```diff
@@ -8,25 +8,55 @@
 #include <AsmMacroLib.h>
 
 ASM_FUNC(ArmCallSmc)
-  // Push x0 on the stack - The stack must always be quad-word aligned
+  // Push x0 and x19 on the stack - The stack must always be quad-word aligned
+  str   x19, [sp, #-16]!
```
leiflindholm, line 12: I feel the existing comment block here suffers from an element of "add 1 to i".

What the pre-existing code does is it pushes the pointer to the ARM_SMC_ARGS structure, decrementing the stack pointer by 16 to ensure 16-byte alignment (I find it unhelpful to refer to it in words, especially when that size is not the architecture's native size).

With that explanation in place: so, as we're storing two 8-byte values, we don't need any additional padding. `stp, x0, x19 [sp, #-16]!` achieves the alignment in one go.

And once that's done, it would be worth to update the comment to reflect that we're pushing x0 because it's the ARM_SMC_ARGS pointer and x19 because it's callee-saved.

ArmPkg/Include/Library/ArmSmcLib.h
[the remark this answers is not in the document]
leiflindholm: Hmm... I'm not sure I agree, @samimujawar.
Could we possibly break the Pcd, function expansion etc discussion out into a mailing-list thread? Ard will be back later this week and may have further relevant input.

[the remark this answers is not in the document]
kuqin12: @samimujawar I think the whole system needs to be consistent on using the number of registers from the level of SMC/SVC.

My main concern is still on the asymmetry usage of the SMC/SVC interface if we go after a combo approach. i.e. the stmm could hand off the control using an 8-register SVC interface, such as `ARM_FID_FFA_ERROR` or `ARM_FID_FFA_MSG_SEND_DIRECT_RESP`, it is entirely possible that the normal world comes back with a `ARM_FID_FFA_MSG_SEND_DIRECT_REQ2`, which needs 18 registers to work with. At that point, the SVC return into stmm will just chop off the 10 registers and not able to handle it properly.

I can spin up an email thread to discuss further.

hrw:
ArmMonitorLib allows to use 18 registers for both arguments and results by using ARM_MONITOR_ARGS structure.

I switched SbsaQemu to use it in previous year: https://github.com/tianocore/edk2-platforms/commit/ddd48c187500383c10519fb7b39b607c100e390c

So the other solution can be recommending switch from ArmSmcLib and its old implemenation of spec to ArmMonitorLib?

kuqin12:
> ArmMonitorLib allows to use 18 registers for both arguments and results by using ARM_MONITOR_ARGS structure.
> 
> I switched SbsaQemu to use it in previous year: [tianocore/edk2-platforms@ddd48c1](https://github.com/tianocore/edk2-platforms/commit/ddd48c187500383c10519fb7b39b607c100e390c)
> 
> So the other solution can be recommending switch from ArmSmcLib and its old implemenation of spec to ArmMonitorLib?

Thanks for the information. So it is using "MDE_CPU_AARCH64" as the preprocessor to guard off extended register usage? If that is generally accepted, I am fine switching to it. But I wonder if this will break the legacy SPMC in EL3, if we just go 18 register straight?

ArmPkg/Include/Library/ArmSmcLib.h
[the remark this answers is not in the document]
leiflindholm: Thanks!
Email thread can be found at https://edk2.groups.io/g/devel/message/121153

ArmPkg/Library/ArmSmcLib/AArch64/ArmSmc.S
kuqin12: Thanks! This is updated.

ArmPkg/Library/ArmSmcLib/AArch64/ArmSmc.S
kuqin12: As discussed on the mailing list, this is no longer relevant.

ArmPkg/Include/Library/ArmSmcLib.h
```diff
@@ -10,6 +10,8 @@
 #ifndef ARM_SMC_LIB_H_
 #define ARM_SMC_LIB_H_
 
+#include <Library/PcdLib.h>
```
leiflindholm, line 13: The change is good, but the commit message needs to be updated to match the new code.

ArmPkg/Include/Library/ArmSvcLib.h
```diff
@@ -9,6 +9,8 @@
 #ifndef ARM_SVC_LIB_H_
 #define ARM_SVC_LIB_H_
 
+#include <Library/PcdLib.h>
```
leiflindholm, line 12: Same as previous - commit message needs updating.

ArmPkg/Library/ArmFfaLib/ArmFfaStandaloneMmLib.inf
```diff
@@ -34,4 +34,3 @@
 
 [Pcd]
   gArmTokenSpaceGuid.PcdFfaLibConduitSmc
-
```
leiflindholm, line 37: I'm all for getting rid of trailing whitespace, but it doesn't belong in a functional change commit.

kuqin12: Thanks. This was a mistake when reverting the PCD change. It is reverted back to original state.

ArmPkg/Include/Library/ArmSvcLib.h
kuqin12: Removed the PCD inclusion and updated commit message.

ArmPkg/Include/Library/ArmSmcLib.h
kuqin12: Removed the PCD inclusion and updated commit message.

kuqin12:
@leiflindholm @samimujawar The PR is updated with feedback here as well as over the email thread. Could you please check and see if there is anything you would like to change? Otherwise, may we merge the PR? Thanks!

leiflindholm (approved):

kuqin12:
@samimujawar @leiflindholm Now the change is approved. Could you please add a push tag so that this change can be merged? Thanks in advance~

ArmPkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -733,12 +754,28 @@ ArmFfaLibCommonInit (
...
+  FfaArgs.Arg0 = SMCCC_VERSION;
+  ArmCallFfa (&FfaArgs);
+  if ((INT32)FfaArgs.Arg0 < 0) {
+    DEBUG ((DEBUG_ERROR, "%a: SMCCC_VERSION not supported\n", __func__));
+    return EFI_UNSUPPORTED;
+  }
+
+  // Cf [1] s2.1.3 'Caller responsibilities',
+  // SMCCC version must be greater or equal than 1.1
+  if ((INT32)FfaArgs.Arg0 < 0x10001) {
```
samimujawar, line 774: I think the check should be greater than or equal to 1.2 
if ((INT32)FfaArgs.Arg0 < 0x10002) {

The reason being the SMCCC v1.2 (0x10002) extends the support for
Argument/Result register set:
• Permits calls to use R4–R7 as return register ( Section 4.1 ).
• Permits calls to use X4–X17 as return registers ( Section 3.1 ).
• Permits calls to use X8–X17 as argument registers ( Section 3.1 ).

Please see 'Appendix F: Changelog' in the SMCCC Specification version G BET0 at https://developer.arm.com/documentation/den0028/latest/

kuqin12: > I think the check should be greater than or equal to 1.2 if ((INT32)FfaArgs.Arg0 < 0x10002) {
> 
> The reason being the SMCCC v1.2 (0x10002) extends the support for Argument/Result register set: • Permits calls to use R4–R7 as return register ( Section 4.1 ). • Permits calls to use X4–X17 as return registers ( Section 3.1 ). • Permits calls to use X8–X17 as argument registers ( Section 3.1 ).
> 
> Please see 'Appendix F: Changelog' in the SMCCC Specification version G BET0 at https://developer.arm.com/documentation/den0028/latest/

Thanks. I will update the version number and add some comments.

kuqin12: > Also linking the TF-A pull request at https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/36018 without which the StMM implementation when using SPMC+SPMD@EL3, i.e. the FF-A support for Arm FVP platform will be broken.

And will add this information into the PR description.

samimujawar:
I do not have any further comments.

However, for StMM in edk2, I think we will only support FF-A 1.2 and SMCCC version >= 1.2.
The main reason being without FF-A DirectReq2 it is not possible to target multiple services in StMM. 
i.e. if we want to implement services like FF-A TPM over CRB in StMM they require DirectReq2.

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
```diff
@@ -29,17 +36,24 @@ ASM_FUNC(ArmCallSvc)
   dsb   nsh
   isb
 
-  // Pop the ARM_SVC_ARGS structure address from the stack into x9
-  ldr   x9, [sp, #16]
+  // Pop the ARM_SVC_ARGS structure address from the stack into x19
+  ldr   x19, [sp], #8
```
odeprez, line 40: I believe this is wrong because making sp unaligned to 16 bytes

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
```diff
@@ -12,13 +12,20 @@
 
 ASM_FUNC(ArmCallSvc)
   // Push frame pointer and return address on the stack
-  stp   x29, x30, [sp, #-32]!
+  stp   x29, x30, [sp, #-48]!
   mov   x29, sp
```
odeprez, line 16: why reserving this much space on the stack?

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
```diff
@@ -29,17 +36,24 @@ ASM_FUNC(ArmCallSvc)
...
+  stp   x14, x15, [x19, #112]
+  stp   x12, x13, [x19, #96]
+  stp   x10, x11, [x19, #80]
+  stp   x8, x9, [x19, #64]
+  stp   x6, x7, [x19, #48]
+  stp   x4, x5, [x19, #32]
+  stp   x2, x3, [x19, #16]
+  stp   x0, x1, [x19, #0]
+
+  mov   x0, x19
```
odeprez, line 54: why is it required to restore x0?

kuqin12:
> I do not have any further comments.
> 
> However, for StMM in edk2, I think we will only support FF-A 1.2 and SMCCC version >= 1.2. The main reason being without FF-A DirectReq2 it is not possible to target multiple services in StMM. i.e. if we want to implement services like FF-A TPM over CRB in StMM they require DirectReq2.

I understand your point and already updated the code implementation to bail at the FF-A library entrypoint if the SMCCC version is < 0x10002. Maybe I am missing anything actionable from your comment?

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
kuqin12: This is because the x0 will contain the pointer to the arg structure. The interface takes one argument as the in/out parameter, so after placing all registers in [x19], it is put back to x0 as the return pointer value.

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
kuqin12: I think this is a leftover from my previous version of changes. I will revert this and test.

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
kuqin12: Thanks for catching this. This was a mistake during the last round of change. But oddly enough, the platforms I tested with did not have any issues running this routine. Anyway, the routine is updated to not do this 8-byte offset and will only increment the stack by 16 on the second call of pulling x19 value.

The change is tested on both hardware and virtual platforms.

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
kuqin12: This is updated. I actually went ahead and reduce the stack reserving size from the original 32 byte to 16 bytes.

The change is tested on both hardware and virtual platforms.

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
odeprez: this is unnecessary as the function signature returns VOID
```
VOID
ArmCallSvc (
  IN OUT ARM_SVC_ARGS  *Args
  );
```

The IN condition is done by LDP instructions lines 24-32
The OUT condition is done by STP instructions lines 44-52

x0 as return value is never consumed

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
```diff
@@ -29,17 +36,24 @@ ASM_FUNC(ArmCallSvc)
   dsb   nsh
   isb
 
-  // Pop the ARM_SVC_ARGS structure address from the stack into x9
-  ldr   x9, [sp, #16]
+  // Pop the ARM_SVC_ARGS structure address from the stack into x19
+  ldr   x19, [sp, #8]
```
odeprez, line 40: this is unnecessary, you should move x0 to x19 prior to SVC at line 22
x19 is preserved as per AAPCS64 when returning from SVC

samimujawar:
> I do not have any further comments.
> 
> However, for StMM in edk2, I think we will only support FF-A 1.2 and SMCCC version >= 1.2. The main reason being without FF-A DirectReq2 it is not possible to target multiple services in StMM. i.e. if we want to implement services like FF-A TPM over CRB in StMM they require DirectReq2.

Sorry this was in context for the TF-A patch and to record what is supported in StMM. 
@LeviYeoReum is preparing a StandaloneMmPkg/Readme.md where I think this can be documented.

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
kuqin12: Ah, good call. Will update.

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
kuqin12: Will remove and only save and restore x19

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
kuqin12: Oh, this is good to know. I updated the implementation and tested on hardware platform as well. There should not be unaligned stack anymore.

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
kuqin12: Thanks for sharing the code. I integrated the code along with your other comments in this PR. I like it better now.

Also added you as co-author of the SVC patch :)

kuqin12: But I did keep the 16-byte stack reserving for x29 and x30. The test results indicate there is no need for 32 bytes. But do you know why it started with 32-byte reservation in the first place?

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
kuqin12: Updated.

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
kuqin12: Updated.

kuqin12:
@samimujawar @odeprez Thank you for the feedback. Do you have any other comments on this PR? Is there any further concerns on merging this change? Any input is appreciated.

ArmPkg/Library/ArmSmcLib/AArch64/ArmSmc.S
```diff
@@ -8,25 +8,37 @@
 #include <AsmMacroLib.h>
 
 ASM_FUNC(ArmCallSmc)
-  // Push x0 on the stack - The stack must always be quad-word aligned
-  str   x0, [sp, #-16]!
+  // Push x19 on to the stack
+  // x0 is the ARM_SMC_ARGS structure address
+  // x19 is callee save and used as a scratch register upon returning from SMC
+  stp   x19, xzr, [sp, #-16]!
+  mov   x19, x0
```
ardbiesheuvel, line 15: This code should set up a frame pointer so we can get a proper stack strace when needed,

```
stp x29, x30, [sp, #-16]!
mov x29, sp
```

That also gives you `x30` as a scratch register, as we won't be doing any function calls.
```
mov x30, x0
```

ArmPkg/Library/ArmSmcLib/AArch64/ArmSmc.S
```diff
@@ -8,25 +8,37 @@
...
+  // A SMC call can return up to 18 values.
+  stp   x16, x17, [x19, #128]
+  stp   x14, x15, [x19, #112]
+  stp   x12, x13, [x19, #96]
+  stp   x10, x11, [x19, #80]
+  stp   x8, x9, [x19, #64]
+  stp   x6, x7, [x19, #48]
+  stp   x4, x5, [x19, #32]
+  stp   x2, x3, [x19, #16]
+  stp   x0, x1, [x19, #0]
```
ardbiesheuvel, line 40: Use `x30` here instead of `x19`

ArmPkg/Library/ArmSmcLib/AArch64/ArmSmc.S
```diff
@@ -8,25 +8,37 @@
...
+
+  ldp   x19, xzr, [sp], #16
```
ardbiesheuvel, line 42: `ldp x29, x30, [sp], #16`

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
```diff
@@ -12,13 +12,21 @@
...
+  stp   x29, x30, [sp, #-16]!
   mov   x29, sp
 
-  // Push x0 on the stack - The stack must always be quad-word aligned
-  str   x0, [sp, #16]
+  // Push x19 on to the stack
+  // x0 is the ARM_SVC_ARGS structure address
+  // x19 is callee save and used as a scratch register upon returning from SVC
+  stp   x19, xzr, [sp, #-16]!
+  mov   x19, x0
```
ardbiesheuvel, line 22: Same comment as in the other patch.

ArmPkg/Library/ArmSmcLib/AArch64/ArmSmc.S
kuqin12: Thanks for the suggestion! I removed x19 usage and moved to x30 as you suggested.

ArmPkg/Library/ArmSmcLib/AArch64/ArmSmc.S
kuqin12: Thanks. Updated.

ArmPkg/Library/ArmSmcLib/AArch64/ArmSmc.S
kuqin12: Thanks. This is updated.

ArmPkg/Library/ArmSvcLib/AArch64/ArmSvc.S
kuqin12: This is updated.

kuqin12:
@ardbiesheuvel @odeprez @samimujawar Do you have any other feedback on this PR? Otherwise can we merge it? :) Thanks!

samimujawar (approved):

sachinami:
Hi @kuqin12 @leiflindholm 

It would seem that the [TF-A change](https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/36018) shared in the description was abandoned in favour of https://android.googlesource.com/trusty/external/trusted-firmware-a/+/71d7b056bab4f868768740a283648f4a39188df4%5E%21/#F0.

Could you share if you have any information on the final conclusion of the change.

kuqin12:
> Hi @kuqin12 @leiflindholm
> 
> It would seem that the [TF-A change](https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/36018) shared in the description was abandoned in favour of https://android.googlesource.com/trusty/external/trusted-firmware-a/+/71d7b056bab4f868768740a283648f4a39188df4%5E%21/#F0.
> 
> Could you share if you have any information on the final conclusion of the change.

@sachinami I think @LeviYeoReum and @samimujawar are better situated to answer your question.

LeviYeoReum:
Hi @sachinami. 

> It would seem that the [TF-A change](https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/36018) shared in the description was abandoned in favour of https://android.googlesource.com/trusty/external/trusted-firmware-a/+/71d7b056bab4f868768740a283648f4a39188df4%5E%21/#F0.
> 
> Could you share if you have any information on the final conclusion of the change.

This PR is duplicated with https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/36862.
So, I've closed the my own.
And the above one is still going to review but seems still stuck?

PR #10688, opened 2025-01-28
MdePkg/BasePeCoffLib: Fix use of uninitialized var after SafeIntLib fix

ardbiesheuvel:
SafeIntLib is only safe if it is used correctly, and if it signals an overflow, the error must be handled, rather than limping on and using garbage data from the stack, as this may turn out to be more dangerous than using the overflowed value.

So initialize RelocBase to NULL as it will get tested even in case of overflow, and this will prevent it from being initialized as expected.


lersek (approved):

mdkinney:
This appears to be a duplicate of PR #10689 and should be closed

PR #10690, opened 2025-01-28
Arm gic mmio mapping

ardbiesheuvel:
Refactor the CPU and GIC DXE drivers so that the latter can invoke the CPU arch protocol exposed by the former to map the MMIO control registers explicitly.

### DIFF
--- a/ArmPkg/Drivers/ArmGicDxe/GicV2/ArmGicV2Dxe.c
+++ b/ArmPkg/Drivers/ArmGicDxe/GicV2/ArmGicV2Dxe.c
@@ -22,6 +22,9 @@ Abstract:
 
 #define ARM_GIC_DEFAULT_PRIORITY  0x80
 
+#define GICD_V2_SIZE  SIZE_4KB
+#define GICC_V2_SIZE  SIZE_8KB
+
 // Interrupts from 1020 to 1023 are considered as special interrupts
 // (eg: spurious interrupts)
 #define ARM_GIC_IS_SPECIAL_INTERRUPTS(Interrupt) \
@@ -529,9 +532,28 @@ GicV2DxeInitialize (
   ASSERT (PcdGet64 (PcdGicInterruptInterfaceBase) <= MAX_UINTN);
   ASSERT (PcdGet64 (PcdGicDistributorBase) <= MAX_UINTN);
 
+  // Locate the CPU arch protocol - cannot fail because of DEPEX
+  Status = gBS->LocateProtocol (&gEfiCpuArchProtocolGuid, NULL, (VOID **)&gCpuArch);
+  ASSERT_EFI_ERROR (Status);
+
   mGicInterruptInterfaceBase = (UINTN)PcdGet64 (PcdGicInterruptInterfaceBase);
   mGicDistributorBase        = (UINTN)PcdGet64 (PcdGicDistributorBase);
-  mGicNumInterrupts          = ArmGicGetMaxNumInterrupts (mGicDistributorBase);
+
+  Status = gCpuArch->SetMemoryAttributes (gCpuArch, mGicDistributorBase, GICD_V2_SIZE, EFI_MEMORY_UC | EFI_MEMORY_XP);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "%a: Failed to map GICv2 distributor MMIO interface: %r\n", __func__, Status));
+    ASSERT_EFI_ERROR (Status);
+    return Status;
+  }
+
+  mGicNumInterrupts = ArmGicGetMaxNumInterrupts (mGicDistributorBase);
+
+  Status = gCpuArch->SetMemoryAttributes (gCpuArch, mGicInterruptInterfaceBase, GICC_V2_SIZE, EFI_MEMORY_UC | EFI_MEMORY_XP);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "%a: Failed to map GICv2 CPU MMIO interface: %r\n", __func__, Status));
+    ASSERT_EFI_ERROR (Status);
+    return Status;
+  }
 
   for (Index = 0; Index < mGicNumInterrupts; Index++) {
     GicV2DisableInterruptSource (&gHardwareInterruptV2Protocol, Index);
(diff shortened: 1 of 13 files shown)


ardbiesheuvel:
@AjanZhong

ArmPkg/Drivers/CpuDxe/CpuDxe.c
```diff
@@ -300,6 +304,29 @@ RemapUnusedMemoryNx (
   }
 }
 
+STATIC VOID  *mHardwareInterruptProtocolNotifyEventRegistration;
```
leiflindholm, line 307: I don't feel strongly about this, but the coding style says to define global vars before the first function (coding standards 5.4).

ardbiesheuvel: I'll move it.

ArmPkg/Drivers/CpuDxe/CpuDxe.c
```diff
@@ -83,7 +85,9 @@ CpuEnableInterrupt (
   IN EFI_CPU_ARCH_PROTOCOL  *This
   )
 {
-  ArmEnableInterrupts ();
+  if (mCpuInterruptsEnabled) {
```
leiflindholm, line 88: Can you elaborate on why this can't cause a race condition resulting in interrupts not being enabled? (in comment or commit message)

ardbiesheuvel: After I convince myself that that is actually the case? Sure :-)

ArmPkg/Drivers/ArmGicDxe/GicV3/ArmGicV3Dxe.c
```diff
@@ -36,7 +36,7 @@ extern EFI_HARDWARE_INTERRUPT_PROTOCOL   gHardwareInterruptV3Protocol;
 extern EFI_HARDWARE_INTERRUPT2_PROTOCOL  gHardwareInterrupt2V3Protocol;
 
 STATIC UINTN  mGicDistributorBase;
-STATIC UINTN  mGicRedistributorsBase;
+STATIC UINTN  mGicRedistributorBase;
```
leiflindholm, line 41: I mean, it *is* still the RedistributorsBase, even after we only access the first one...

ardbiesheuvel: It is not. If the current CPU does not correspond with the first redistributor frame, we will iterate until we find the one that does, and record its address in this variable.

ArmVirtPkg/Library/QemuVirtMemInfoLib/QemuVirtMemInfoLib.c
```diff
@@ -18,13 +18,13 @@
 #define MAX_VIRTUAL_MEMORY_MAP_DESCRIPTORS  5
```
leiflindholm, line 18: Slightly nitpicky commit message comment:
The GIC doesn't map its MMIO registers, the GIC *driver* maps its MMIO registers.

ArmPkg/Drivers/CpuDxe/CpuDxe.c
```diff
@@ -206,8 +206,7 @@ IdleLoopEventCallback (
 //
 // Globals used to initialize the protocol
 //
-EFI_HANDLE             mCpuHandle = NULL;
-EFI_CPU_ARCH_PROTOCOL  mCpu       = {
+EFI_CPU_ARCH_PROTOCOL  mCpu = {
```
leiflindholm: If we're cleaning up, should we make this static as well? It's not used outside this file.

ardbiesheuvel: See the next patch

ArmPkg/Drivers/CpuDxe/CpuDxe.c
```diff
@@ -308,6 +307,7 @@ CpuDxeInitialize (
 {
   EFI_STATUS  Status;
   EFI_EVENT   IdleLoopEvent;
+  EFI_HANDLE  CpuHandle;
```
leiflindholm, line 405: Was the NULL initialization in the original code incorrect, or is it missing here?

ArmPkg/Drivers/CpuDxe/CpuDxe.c
leiflindholm: Ah, you do that in 2.

ArmPkg/Drivers/CpuDxe/CpuDxe.c
ardbiesheuvel: It is missing here - this is why the CI is failing as well.

ArmPkg/Drivers/CpuDxe/CpuDxe.c
ardbiesheuvel: I've adopted another approach to ensure that the interrupt en/disable state is guaranteed to remain consistent from the caller's PoV, by keeping a shadow state initially, and syncing the actual state with that once the GIC driver is up

ArmPkg/Drivers/ArmGicDxe/GicV2/ArmGicV2Dxe.c
```diff
@@ -529,9 +532,22 @@ GicV2DxeInitialize (
...
 
+  // Locate the CPU arch protocol - cannot fail because of DEPEX
+  Status = gBS->LocateProtocol (&gEfiCpuArchProtocolGuid, NULL, (VOID **)&gCpuArch);
+  ASSERT_EFI_ERROR (Status);
+
   mGicInterruptInterfaceBase = (UINTN)PcdGet64 (PcdGicInterruptInterfaceBase);
   mGicDistributorBase        = (UINTN)PcdGet64 (PcdGicDistributorBase);
-  mGicNumInterrupts          = ArmGicGetMaxNumInterrupts (mGicDistributorBase);
+
+  Status = gCpuArch->SetMemoryAttributes (gCpuArch, mGicDistributorBase, GICD_V2_SIZE, EFI_MEMORY_UC | EFI_MEMORY_XP) ?:
```
leiflindholm, line 542: Is ?: not a GCC extension? https://gcc.gnu.org/onlinedocs/gcc/Conditionals.html

kraxel:
Last patch in this series breaks my guests.
Exception after loading the TPM PEI.
Removing the TPM from the guest config makes the VM boot.
I guess the TPM needs to map MMIO address ranges itself too?

ardbiesheuvel:
If you send a revert pr I'll merge it right away - it's not that important

kraxel:
https://github.com/tianocore/edk2/pull/10708

PR #10691, opened 2025-01-28
Add rng ppi

### LINKED ISSUE #10529: Code First - PI Specification - Add RNG PPI (Bugzilla Bug 4814)
This issue was created automatically with [bugzilla2github](https://github.com/makubacki/bugzilla2github/tree/bz2gh_tianocore)
# Bugzilla Bug 4814
Date: 2024-07-17T20:32:31+00:00
From: @makubacki
To:   @makubacki
CC:   @lgao4

Last updated: 2024-09-04T20:49:38+00:00

makubacki:
# Description

Closes #10529 

The RNG PPI was a code first proposal under [BZ1814](https://github.com/tianocore/edk2/issues/10529) and published in the [PI Spec v1.9](https://uefi.org/specs/PI/1.9/).

- edk2-staging Branch: https://github.com/tianocore/edk2-staging/commits/BZ4814-AddRngPpi
- [PI Spec Change Draft](https://github.com/tianocore/edk2-staging/blob/BZ4814-AddRngPpi/CodeFirst/BZ4814-SpecChange.md)
- [PI Spec v1.9 Published Section](https://uefi.org/specs/PI/1.9/V1_Additional_PPIs.html#random-number-generator-ppi)

This change adds the `RngPei` module to SecurityPkg to produce the PPI and the `RngLib` library to MdePkg to consume the PPI.

---

**MdePkg: Add Random Number Generator (RNG) PPI**

REF:https://bugzilla.tianocore.org/show_bug.cgi?id=4814

Adds a new PPI that serves the same purpose as EFI_RNG_PROTOCOL in
DXE. This PPI can be produced by a PEIM to provide a dynamic interface
to RNG services in PEI.

---

**MdePkg: Add PeiRngLib**

REF:https://bugzilla.tianocore.org/show_bug.cgi?id=4814

Adds a new PEI library instance for RngLib that uses the RNG services
provided by the RNG PPI.

This library instance will add a DEPEX on gEfiRngPpiGuid on modules
it links against. It can be used to allow PEIMs to get RNG support
over a dynamic interface.

---

**SecurityPkg: Add RngPei**

REF:https://bugzilla.tianocore.org/show_bug.cgi?id=4814

The `RngPei` PEIM can be used if RNG should be provided over a
dynamic binary interface to other PEIMs on a platform.

---

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- SecurityPkg CI
- Verify RNG linked with RngLib is executed as expected
- Verify random numbers are generated successfully with a valid
  RngLib

## Integration Instructions

Use the `RngPei` module if a platform needs to produce
`gEfiRngPpiGuid`.

The platform should usually link a different `RngLib` instance to
`RngPei` than other PEIMs that may use the RNG PPI produced since
`RngPei` is responsible for producing the PPI.

For example, a `RngLib` instance that uses the rdrand instruction
may be linked against `RngPei` and a `RngLib` instance that uses
the RNG PPI may be linked against other PEIMs.


MdePkg/Include/Ppi/Rng.h
```diff
@@ -0,0 +1,26 @@
...
+  in PEIMs, or entropy for seeding other random number generators.
+
+  Copyright (c) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#ifndef RNG_PPI_H_
+#define RNG_PPI_H_
+
+#include <Protocol/Rng.h>
```
mdkinney, line 12: I understand that the UEFI Protocol and the PI Spec PPI have the same services, but they are 2 specs that are independently updated.  Should the RNG PPI have its own definition in this file instead of depending on a protocol?

makubacki: I was thinking that split could happen when/if needed. But I don't have a strong opinion and I'm happy to separate the interface now if you'd like.

mdkinney: The part that is most concerning is a PPI including a Protocol.  We usually consider that the PEI Phase can be used independent of the DXE Phase or the UEFI Phase.

A split would be cleaner, but you would still have the algorithm GUIDs that can not be duplicated.

Since the algorithm GUIDs are shared between the Protocol and the PPI.  Perhaps there could be a MdePkg/Include/Guid/Rng.h file with the algorithm GUIDs and a generic version of the structure with the APIs, and then the Protocol and the PPI can include the <Guid/Rng.h> and do typedef.  This way, a protocol does not depend on a PPI and a PPI does not depend on a protocol.

makubacki: @mdkinney, I made the updates, and I tried to do this in a way that consolidates and cleans up relevant RNG code.

Two questions:

1. [MdeModulePkg/Include/Guid/RngAlgorithm.h](https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Include/Guid/RngAlgorithm.h) has a GUID defined as `gEdkiiRngAlgorithmUnSafe`. I couldn't find this in the UEFI or PI Specification and the other RNG algorithm GUIDs are defined in [MdePkg/Include/Protocol/Rng.h](https://github.com/tianocore/edk2/blob/master/MdePkg/Include/Protocol/Rng.h) and defined in the UEFI Specification. The RNG interface is also in MdePkg. I moved the existing `RngAlgorithms.h` file into MdePkg and added there other algorithms from the protocol header there. Ultimately, this seems a little messy given `gEdkiiRngAlgorithmUnSafe` is not in the same specification as the others. Let me know what you'd like to do.
2. For (1), I made the move from MdeModulePkg to MdePkg in a single commit (the first commit in the current series). That was the cleanest to me to handle the move. If we keep that change, let me know if you want that handled differently.

mdkinney: `gEfkiiRngAlgorithmUnSafe` was introduced in PR #4807 to indicate if an unsafe RNG generator is used with the example in that PR being an RNG based on a free running timer.  I agree it is not part of any industry standard specification, and the use of an unsafe RNG would never be recommended.  I think that this EDK II specific GUID should stay in the `MdeModulePkg` because it is not part of an industry standard spec.  The unsafe timer based RNG implementations can have the extra dependency on the `MdeModulePkg`.  RNG implementations that follow the industry standard spec algorithms can have a dependency on only the `MdePkg` and would not require the `MdeModulePkg` dependency for a reference to a GUID that does not apply.

Moving interfaces from a non MdePkg to the MdePkg is usually safe because every module/lib depends on the MdePkg.

Moving module/library implementations between packages should be split across multiple commits to allow platform DSC to accommodate the change of location.

makubacki: Sounds good, I'll keep `gEfkiiRngAlgorithmUnSafe` in MdeModulePkg. That would stay in `MdeModulePkg/Include/Guid/RngAlgorithm.h`. In that case, I'll combine the algorithms currently in `MdePkg/Include/Guid/RngAlgorithm.h` (all but the unsafe algorithm) into `MdePkg/Include/Guid/Rng.h` (alongside the struct definition). Please confirm that sounds good and I'll make that update.

makubacki: Pushed that update.

mdkinney:
The commit messages reference BZ links.  Should those be updated to GitHub Issues links?

makubacki:
> The commit messages reference BZ links. Should those be updated to GitHub Issues links?

Updated the references to GitHub issues.

mdkinney (approved):

mdkinney:
@jyao1 Please review.  This was started before soft freeze and should be considered for next stable tag

@lgao4 Request adding edk2-stable202502 milestone

jyao1 (approved):

PR #6408, opened 2024-11-06
Unit test framework pkg enable sanitize

mdkinney:
# Description

UnitTestFrameworkPkg: Use /MTd and enable Address Sanitizers

* Update host based unit test VS20xx builds to use /MTd instead of
  /MT to enable use of debug libraries for host based unit tests.
* Enable /fsanitize=address for host based unit test VS2019 builds
* Enable /fsanitize=address for host based unit test VS2022 builds
* Enable -fsanitize=address for host based unit test GCC builds
* Add UNIT_TESTING_ADDRESS_SANITIZER_ENABLE define that is set to
  TRUE by default so it is always enabled, but can be set to FALSE
  to temporarily disable during development/debug of unit tests.
* Add HostMemoryAllocationBelowAddressLib to support allocating 
  buffers below a specified address in unit test code and mocks to
  support code under test that requires a buffer below a specific 
  address (e.g. 4GB). The services included are:
```
VOID * EFIAPI HostAllocatePoolBelowAddress (IN VOID   *MaximumAddress, IN UINTN  Length);
VOID EFIAPI HostFreePoolBelowAddress (IN VOID  *Address);
VOID * EFIAPI HostAllocateAlignedPagesBelowAddress (IN VOID   *MaximumAddress, IN UINTN  Pages, IN UINTN  Alignment );
VOID EFIAPI HostFreeAlignedPagesBelowAddress (IN VOID   *Buffer, IN UINTN  Pages);
```
Enabling the Address Sanitizer can detect double frees, buffer
overflow, buffer underflow, access to invalid addresses, and
various exceptions. These can be detected in both the unit test
case sources as well as the code under test.

- [ ] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [ ] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

## Integration Instructions

The environment variable `GTEST_CATCH_EXCEPTION` must be set to `0` for so all exceptions are handled by the address sanitizer and not GoogleTest. This allows stack back trace and other details to be logged by the address sanitizer so the source of the issue identified address sanitizer can be determined.

[...]

PR #10694, opened 2025-01-29
UnitTestFrameworkPkg: Fix false positives from address sanitizer

mdkinney:
# Description

PR #6408 introduced a host specific NORETURN function to address false positives from the address sanitizer after LongJump() is called and the stack location is modified.

This approach was discussed here:
https://github.com/tianocore/edk2/pull/6408/files#r1918810499

Some false positives are still being observed with this initial solution.

The address sanitizer provides __asan_handle_no_return() to clean up shadow memory before a NORETURN function is called and provides a simpler implementation for this issue without having to introduce a host specific
NORETURN function.

https://github.com/llvm/llvm-project/blob/main/compiler-rt/include/sanitizer/asan_interface.h

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verified that all current unit tests pass as well as ones that were generating false positives.

## Integration Instructions

N/A

### DIFF
--- a/UnitTestFrameworkPkg/Library/UnitTestDebugAssertLib/UnitTestDebugAssertLibHost.cpp
+++ b/UnitTestFrameworkPkg/Library/UnitTestDebugAssertLib/UnitTestDebugAssertLibHost.cpp
@@ -18,32 +18,22 @@ extern "C" {
   #include <Library/BaseLib.h>
   #include <Library/UnitTestLib.h>
 
+  //
+  // If address sanitizer is enabled, then declare the function that is used to
+  // handle custom long jump implementation.
+  //
+ #ifdef __SANITIZE_ADDRESS__
+  void
+  __asan_handle_no_return (
+    );
+
+ #endif
+
   ///
   /// Point to jump buffer used with SetJump()/LongJump() to test if a function
   /// under test generates an expected ASSERT() condition.
   ///
   BASE_LIBRARY_JUMP_BUFFER  *gUnitTestExpectAssertFailureJumpBuffer = NULL;
-
-  /**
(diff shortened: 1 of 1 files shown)


mdkinney:
@mhaeuser This is a better and more complete fix to the false positives that you commented on in the previous PR.

UnitTestFrameworkPkg/Library/UnitTestDebugAssertLib/UnitTestDebugAssertLibHost.cpp
```diff
@@ -68,8 +58,18 @@ extern "C" {
 
     if (gUnitTestExpectAssertFailureJumpBuffer != NULL) {
       UT_LOG_INFO ("Detected expected ASSERT: %a(%d): %a\n", FileName, LineNumber, Description);
-      HostLongJump (gUnitTestExpectAssertFailureJumpBuffer, 1);
-      UNREACHABLE ();
+
+      //
+      // If address sanitizer is enabled, then inform sanitizer that a no return
+      // function is being called that will reset to stack to a previous stack
```
os-d, line 64:
```suggestion
      // function is being called that will reset to a previous stack

```

os-d (approved):

UnitTestFrameworkPkg/Library/UnitTestDebugAssertLib/UnitTestDebugAssertLibHost.cpp
mdkinney: I have used your suggested change

makubacki (approved):

PR #10695, opened 2025-01-29
ovmv / armvirt: add support for qemu uefi variable store

kraxel:
- **OvmfPkg: add IndustryStandard/QemuUefiVars.h**
- **OvmfPkg: add new VirtMmCommunicationDxe driver**
- **OvmfPkg/OvmfPkgX64: add QEMU_VARS option**
- **OvmfPkg: add PcdQemuVarsRequire**
- **OvmfPkg/VirtMmCommunicationDxe: stop on init failure**
- **OvmfPkg/VirtMmCommunicationDxe: add arm support**
- **ArmVirtPkg/ArmVirtQemu: add QEMU_VARS option**
- **[fixup] PcdQemuVarsRequire clash**

### DIFF
--- a/OvmfPkg/OvmfPkgX64.dsc
+++ b/OvmfPkg/OvmfPkgX64.dsc
@@ -31,6 +31,7 @@
   #
   DEFINE SECURE_BOOT_ENABLE      = FALSE
   DEFINE SMM_REQUIRE             = FALSE
+  DEFINE QEMU_PV_VARS            = FALSE
   DEFINE SOURCE_DEBUG_ENABLE     = FALSE
   DEFINE CC_MEASUREMENT_ENABLE   = TRUE
 
@@ -375,7 +376,7 @@
   PciLib|OvmfPkg/Library/DxePciLibI440FxQ35/DxePciLibI440FxQ35.inf
   QemuFwCfgS3Lib|OvmfPkg/Library/QemuFwCfgS3Lib/DxeQemuFwCfgS3LibFwCfg.inf
   VariablePolicyLib|MdeModulePkg/Library/VariablePolicyLib/VariablePolicyLibRuntimeDxe.inf
-!if $(SMM_REQUIRE) == TRUE
+!if $(SMM_REQUIRE) == TRUE || $(QEMU_PV_VARS) == TRUE
   MmUnblockMemoryLib|MdePkg/Library/MmUnblockMemoryLib/MmUnblockMemoryLibNull.inf
 !endif
 
@@ -498,6 +499,9 @@
 !if $(SMM_REQUIRE) == TRUE
   gUefiOvmfPkgTokenSpaceGuid.PcdSmmSmramRequire|TRUE
   gUefiCpuPkgTokenSpaceGuid.PcdCpuHotPlugSupport|TRUE
+  gEfiMdeModulePkgTokenSpaceGuid.PcdEnableVariableRuntimeCache|FALSE
+!endif
+!if $(QEMU_PV_VARS) == TRUE
   gEfiMdeModulePkgTokenSpaceGuid.PcdEnableVariableRuntimeCache|FALSE
 !endif
 !if $(SECURE_BOOT_ENABLE) == TRUE
@@ -1016,6 +1020,15 @@
   MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.inf
 
 !else
+!if $(QEMU_PV_VARS) == TRUE
+
+  #
+  # Variable driver stack (qemu -device uefi-vars)
+  #
+  OvmfPkg/VirtMmCommunicationDxe/VirtMmCommunication.inf
+  MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.inf
+
+!else
 
   #
   # Variable driver stack (non-SMM)
@@ -1030,6 +1043,7 @@
     <LibraryClasses>
       NULL|MdeModulePkg/Library/VarCheckUefiLib/VarCheckUefiLib.inf
   }
+!endif
 !endif
 
   #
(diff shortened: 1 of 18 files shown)


kraxel:
@ardbiesheuvel, @lersek 

Question: Does FDT_CLIENT_PROTOCOL have some way to deal with the qemu platform bus?

Essentially I need some logic simliar to tpm discovery in ArmVirtPkg/Library/PlatformPeiLib/PlatformPeiLib.c to calculate the actual mmio address (see "add arm support" patch).  Suggestions welcome.  Thanks.,

lersek:
@kraxel I think you may have to open-code the range translation on top of the `FDT_CLIENT_PROTOCOL` member functions.

ardbiesheuvel:
> @kraxel I think you may have to open-code the range translation on top of the `FDT_CLIENT_PROTOCOL` member functions.

It would be nice if we could put some abstraction in FDT_CLIENT_PROTOCOL that presents a `"reg"` property as an array of host base,size tuples, walking the tree as appropriate to parse cell sizes and `ranges` properties.

Otherwise, we'll be relying on one particular path that QEMU happens to generate for the platform bus.

kraxel:
So, I got things working.  Had to code up a bunch of support code to read `#address-cells` + `#size-cells` + helpers to read values from `reg` and `ranges` properties.  It's all in the new driver now (see the 'wip' commit), but it feels like this is the wrong place for the helper functions.  Suggestions where to move it?  FdtClient?  Some library?

Specifically I think the TPM code in PlatformPei needs that too, it assumes cell size 2 which is not the case (not tested yet though).

kraxel:
Oops, the rules for the cell sizes are more complicated ...
I guess I see now why the edk2 code usually simply looks at `RegSize`.

kraxel:
> It would be nice if we could put some abstraction in FDT_CLIENT_PROTOCOL that presents a `"reg"` property as an array of host base,size tuples, walking the tree as appropriate to parse cell sizes and `ranges` properties.

Placing it in FDT_CLIENT_PROTOCOL wouldn't help for the TPM in PEI though.
I can have a look nevertheless, but will probably take some time.

> Otherwise, we'll be relying on one particular path that QEMU happens to generate for the platform bus.

Yea.  Nesting things like this seems to be pretty rare though, and the qemu platform bus hasn't changed in 10 years, so in practice it probably would not be too bad in practice.

ardbiesheuvel:
> > It would be nice if we could put some abstraction in FDT_CLIENT_PROTOCOL that presents a `"reg"` property as an array of host base,size tuples, walking the tree as appropriate to parse cell sizes and `ranges` properties.
> 
> Placing it in FDT_CLIENT_PROTOCOL wouldn't help for the TPM in PEI though. I can have a look nevertheless, but will probably take some time.
> 

Yeah so I guess it should be a library that operates on the FDT blob directly. Perhaps FDT_CLIENT_PROTOCOL should grow a method to return that address, so DXE phase users of the library don't have to use the PCD.


> > Otherwise, we'll be relying on one particular path that QEMU happens to generate for the platform bus.
> 
> Yea. Nesting things like this seems to be pretty rare though, and the qemu platform bus hasn't changed in 10 years, so it probably would not be too bad in practice.

Fingers crossed :-)

kraxel:
> > Placing it in FDT_CLIENT_PROTOCOL wouldn't help for the TPM in PEI though. I can have a look nevertheless, but will probably take some time.
> 
> Yeah so I guess it should be a library that operates on the FDT blob directly. Perhaps FDT_CLIENT_PROTOCOL should grow a method to return that address, so DXE phase users of the library don't have to use the PCD.

The FDT is in a HOB.  I'd rather go directly lookup ther HOB instead of using FdtClientDxe for that only.

I'm still trying to figure how to move this forward best.  We have two libfdt implementations in the tree, `EmbeddedPkg/Library/FdtLib` and `MdePkg/Library/BaseFdtLib`.  I guess for new code the latter is the better option?

> > Yea. Nesting things like this seems to be pretty rare though, and the qemu platform bus hasn't changed in 10 years, so it probably would not be too bad in practice.
> 
> Fingers crossed :-)

Ok, PR has this now, with a comment about the built-in assumtion.

Undrafting PR, even though I don't expect this go in during the freeze.

ardbiesheuvel:
> I have a branch ready but it is not merged yet.

Got a link?

OvmfPkg/OvmfPkgX64.dsc
```diff
@@ -31,6 +31,7 @@
   #
   DEFINE SECURE_BOOT_ENABLE      = FALSE
   DEFINE SMM_REQUIRE             = FALSE
+  DEFINE QEMU_VARS               = FALSE
```
ardbiesheuvel, line 34: Can we make this `QEMU_PV_VARS` or something like that?

ardbiesheuvel (approved):

OvmfPkg/OvmfPkgX64.dsc
kraxel: Sure thing.

ardbiesheuvel (approved):
This looks good to me, thanks for working on it.

I'll merge this once there is an ETA for the QEMU pieces, please add a link here to the QEMU mailing list discussion.

kraxel:
> I'll merge this once there is an ETA for the QEMU pieces, please add a link here to the QEMU mailing list discussion.

v3 posted today: https://mail.gnu.org/archive/html/qemu-devel/2025-02/msg02113.html

I've Cc'ed you.  Would be great if you can have a look, the number of people with efi background in the qemu community is not very big ...

kraxel:
> qemu PR got merged just in time for the upcoming 10.0 version (code freeze now, release early april).

ping @ardbiesheuvel

wanglmb:
Is there any work  doing about libvirt ?

PR #10698, opened 2025-01-30
ArmPkg: ArmExceptionLib cleanup

leiflindholm:
While playing around with GCC -pedantic, I found an incorrect function pointer comparison in code that deals with relocating reset vectors and keeping debugger support happy.

Looking around how to clean that up, I realised:
1) This code is 64-bit unsafe in the first place.
2) I can find no users for very many years.
So I propose we simply get rid of this functionality for now, and if someone complains, we look at how to implement it properly.

After trimming up the now-unused Pcds, I found some stray references lying around, so clean those up and delete.


ArmPkg/Library/ArmExceptionLib/ArmExceptionLib.c
```diff
@@ -90,107 +63,28 @@ InitializeCpuExceptionHandlers (
...
+  // for AArch64 Align=4K is required.  Align=Auto can be used but this
+  // is known to cause an issue with populating the reset vector area
+  // for encapsulated FVs.
+  ASSERT (((UINTN)ExceptionHandlersStart & gExceptionVectorAlignmentMask) == 0);
 
-    // We do not copy the Exception Table at PcdGet64(PcdCpuVectorBaseAddress). We just set Vector
-    // Base Address to point into CpuDxe code.
-    VectorBase = (UINT64)(UINTN)ExceptionHandlersStart;
+  // We do not copy the Exception Table at PcdGet64(PcdCpuVectorBaseAddress). We just set Vector
+  // Base Address to point into CpuDxe code.
```
ardbiesheuvel, line 80: Can we drop this comment please? Along with `PcdCpuVectorBaseAddress`, actually.

leiflindholm: Oh right. I missed that one. Will cook up a cleanup PR for edk2-platforms first, then update this one.

leiflindholm: https://github.com/tianocore/edk2-platforms/pull/734 raised.

leiflindholm: Dropped. And I smashed the two commits, which seemed to make more sense when I thought about it.

ardbiesheuvel (approved):

PR #10707, opened 2025-01-31
ArmVirtPkg: turn off debug logging for VirtioSerialDxe

kraxel:
When running the verbose builds on arm for development and testing
it is quite convenient to use virtio serial to interact with the
system because console and logging are separated then.  Except that
the virtio serial driver spams the debug log in that case.  Set the
debug level for the driver to zero to avoid that.


kraxel:
can i re-run ci somehow if tests are failing due to flaky internet connectivity?

ardbiesheuvel (approved):

PR #10711, opened 2025-01-31
MdeModulePkg/Bus/Pci: Add a PCI to control removal of devices

Sean-StarLabs:
edk2 will remove devices that have all read-write bits set, believing they are invalid. This is not the case with bootloaders such as coreboot.

Add a Pcd to control this, defaulting to true to not change the existing behavour, but allow these bootloaders to work as intended.

### DIFF
--- a/MdeModulePkg/Bus/Pci/PciBusDxe/PciEnumeratorSupport.c
+++ b/MdeModulePkg/Bus/Pci/PciBusDxe/PciEnumeratorSupport.c
@@ -2585,7 +2585,9 @@ PciEnumeratorLight (
       //
       // Remove those PCI devices which are rejected when full enumeration
       //
-      RemoveRejectedPciDevices (RootBridgeDev->Handle, RootBridgeDev);
+      if (FixedPcdGetBool (PcdRemoveRejectedDevices)) {
+        RemoveRejectedPciDevices (RootBridgeDev->Handle, RootBridgeDev);
+      }
 
       //
       // Process option rom light
--- a/MdeModulePkg/MdeModulePkg.uni
+++ b/MdeModulePkg/MdeModulePkg.uni
@@ -1361,6 +1361,8 @@
 
 #string STR_gEfiMdeModulePkgTokenSpaceGuid_PcdPcieResizableBarSupport_PROMPT #language en-US "Enable PCIe Resizable BAR Capability Supported"
 
+#string STR_gEfiMdeModulePkgTokenSpaceGuid.PcdRemoveRejectedDevices_PROMPT "Don't remove PCI devices due to all read-write bits set."
+
 #string STR_gEfiMdeModulePkgTokenSpaceGuid_PcdPcieResizableBarSupport_HELP #language en-US "Indicates if the PCIe Resizable BAR Capability Supported.<BR><BR>\n"
                                                                                             "TRUE  - PCIe Resizable BAR Capability is supported.<BR>\n"
                                                                                             "FALSE - PCIe Resizable BAR Capability is not supported.<BR>"
--- a/MdeModulePkg/Bus/Pci/PciBusDxe/PciBusDxe.inf
+++ b/MdeModulePkg/Bus/Pci/PciBusDxe/PciBusDxe.inf
@@ -107,6 +107,7 @@
   gEfiMdeModulePkgTokenSpaceGuid.PcdMrIovSupport                ## CONSUMES
   gEfiMdeModulePkgTokenSpaceGuid.PcdPciDisableBusEnumeration    ## SOMETIMES_CONSUMES
   gEfiMdeModulePkgTokenSpaceGuid.PcdPcieResizableBarSupport     ## CONSUMES
+  gEfiMdeModulePkgTokenSpaceGuid.PcdRemoveRejectedDevices       ## CONSUMES
 
 [UserExtensions.TianoCore."ExtraFiles"]
   PciBusDxeExtra.uni
--- a/MdeModulePkg/MdeModulePkg.dec
+++ b/MdeModulePkg/MdeModulePkg.dec
@@ -2278,6 +2278,10 @@
   # @Prompt The value is use for Usb Network rate limiting supported.
   gEfiMdeModulePkgTokenSpaceGuid.PcdUsbNetworkRateLimitingFactor|100|UINT32|0x10000028
 
+  ## Remove PCI devices that are deemed invalid.
+  # @Prompt The value will determine whether devices are removed or not.
+  gEfiMdeModulePkgTokenSpaceGuid.PcdRemoveRejectedDevices|FALSE|BOOLEAN|0x10000029
+
 [PcdsPatchableInModule]
   ## Specify memory size with page number for PEI code when
   #  Loading Module at Fixed Address feature is enabled.


mdkinney:
What do you mean by all read-write bits set?  This seems related to path for full enumeration or not.  Can that policy be used without introducing a new PCD?

Sean-StarLabs:
> Can that policy be used without introducing a new PCD?

Sorry, I can't see it - which one are you referring too?

mdkinney:
Wondering if the state of the PCI devices is related to not running the full enumeration algorithm that assigns resources to PCI BARs. 

```
  ## Indicates if full PCI enumeration is disabled.<BR><BR>
  #   TRUE  - Full PCI enumeration is disabled.<BR>
  #   FALSE - Full PCI enumeration is not disabled.<BR>
  # @Prompt Disable full PCI enumeration.
  gEfiMdeModulePkgTokenSpaceGuid.PcdPciDisableBusEnumeration|FALSE|BOOLEAN|0x10000048
```

Sean-StarLabs:
That is magic, fixes this and two other issue that there were with coreboot... so
1. THANK YOU
2. Are these "paths" documented anywhere? I'm wondering if there's other things that should be switched of when just using edk2 as a payload.

mdkinney:
@ChaselChiu may be able to provide a list and docs based on UefiPayloadPkg work.

Sean-StarLabs:
Sorry, that was wrong - it just flukely works on MTL, anything older still has the problem.

Also, it seems UefiPayloadPkg already controls PcdPciDisableBusEnumeration

MrChromebox:
IMO, this patch addresses a symptom rather than the root cause -- the code governing PCI device rejection is buggy.

The issue we're seeing on boards running coreboot with Intel iGPUs is that when the 'top down' resource allocation strategy is employed, the IO BAR at offset 0x20 is assigned 0xffc0 - 0xffff (size 0x40), and `IsPciDeviceRejected()` returns true. There are several bugs in this function IMO:

* the PMEM64 BAR at offset 0x18 should be detected first and cause the loop to break / the function to return false
* the IO BAR at offset 0x20 is perfectly valid, but failing a check

The former is caused by `BarExisted()` (should be `BarExists()` ?) returning EFI_NOT_FOUND for the BAR at offset 0x18.
The latter I'm not sure about, because I don't understand the logic being used to test the validity of the BAR.

benjamindoron:
I agree with @MrChromebox. When we first discovered this at https://github.com/9elements/edk2/commit/444afa8d097cd76e7f8f2d8e5772d9aac8af99e8, we removed it entirely on the basis that this doesn't actually determine validity. What even would be an invalid BAR, one with size 0? I thought that the best strategy was to drop it, but left the possibility open that this addresses an issue with some hardware that I'm unaware of.

So, I'd rather discuss if this code catches any bad hardware. Regardless, this isn't about bootloaders, it's about this edge case of 'all flippable bits set' in code that may or may not still be useful.

MrChromebox:
reviving this since it went stale -- the device removal is still an issue with the iGPU and coreboot's top-down MMIO assignment

MdeModulePkg/MdeModulePkg.uni
```diff
@@ -1361,6 +1361,8 @@
 
 #string STR_gEfiMdeModulePkgTokenSpaceGuid_PcdPcieResizableBarSupport_PROMPT #language en-US "Enable PCIe Resizable BAR Capability Supported"
 
+#string STR_gEfiMdeModulePkgTokenSpaceGuid.PcdRemoveRejectedDevices_PROMPT "Don't remove PCI devices due to all read-write bits set."
```
MrChromebox: suggestion: "Don't remove PCI devices with potentially invalid BARs"

MrChromebox (approved):

Sean-StarLabs:
@niruiyu Hey Ray, would you be able to review please?

MrChromebox:
> If the existing logic is not sound, we should fix it. In any case, adding another PCD for this is my least preferred option.

I agree, but no one seems to be able to explain the logic, why it was used originally, or what should replace it

MdeModulePkg/Bus/Pci/PciBusDxe/PciEnumeratorSupport.c
```diff
@@ -2596,7 +2596,9 @@ PciEnumeratorLight (
       //
       // Remove those PCI devices which are rejected when full enumeration
       //
-      RemoveRejectedPciDevices (RootBridgeDev->Handle, RootBridgeDev);
+      if (FixedPcdGetBool (PcdRemoveRejectedDevices)) {
+        RemoveRejectedPciDevices (RootBridgeDev->Handle, RootBridgeDev);
```
ardbiesheuvel, line 2600: Can you explain how this change right here is related to this logic that does not accept ranges that end with `0xffff`. Also, could you provide some examples of ranges that are correct, that are unambiguously incorrect, and that are misidentified by the logic in question?

MrChromebox: > Can you explain how this change right here is related to this logic that does not accept ranges that end with `0xffff`. 

This change prevents `RemoveRejectedDevices()` from being called. A subfunction of that, `IsPciDeviceRejected()` contains the faulty logic.

> Also, could you provide some examples of ranges that are correct, that are unambiguously incorrect, and that are misidentified by the logic in question?

When using the top-down resource assignment strategy, coreboot assigns the following resources for an Intel iGPU (on an Alderlake-N board, in this example):

BAR (0x10): 0xbf000000 - 0xbfffffff
BAR (0x18): 0xd0000000 - 0xdfffffff
BAR (0x20): 0xffc0 - 0xffff (IO)

The IO BAR is the one that causes the issue.

The problematic code is in `MdeModulePkg/Bus/Pci/PciBusDxe/PciEnumeratorSupport.c`, starting around ln 2771:

```
Status = BarExisted (PciIoDevice, BarOffset, &TestValue, &OldValue);
    if (EFI_ERROR (Status)) {
      continue;
    }

    if ((TestValue & 0x01) != 0) {
      //
      // IO Bar
      //
      Mask      = 0xFFFFFFFC;
      TestValue = TestValue & Mask;
      if ((TestValue != 0) && (TestValue == (OldValue & Mask))) {
        return TRUE;
      }
    }
```
For the I/O BAR at 0x20, after `BarExisted()` is called `TestValue` and `OldValue` are both 0xFFC1.
So `TestValue & Mask` == `OldValue & Mask` and the function returns TRUE (and the device is rejected). As a result, here is no pre-OS display output.

I'm not sure where this logic of applying the same mask to `TestValue` and `OldValue` and comparing comes from. AFAICT, the standard way to determine the validity of a BAR is to check if `TestValue` is non-zero. The masking is only used to calculate the BAR's length (`~(*TestValue & Mask) + 1`), but that isn't what is done here. Can someone else explain it?

Now, for a normal 32-bit BAR, one would expect that `TestValue` would be 0xFFFFFFC1, not 0x0000FFC1, and I suspect that difference is what is causing the issue with the comparison. I'm not sure if that's a consequence of how coreboot is programming the BAR, or something else. Either way, the logic used seems incorrect.

ardbiesheuvel: > > Can you explain how this change right here is related to this logic that does not accept ranges that end with `0xffff`.
> 
> This change prevents `RemoveRejectedDevices()` from being called. A subfunction of that, `IsPciDeviceRejected()` contains the faulty logic.
> 
> > Also, could you provide some examples of ranges that are correct, that are unambiguously incorrect, and that are misidentified by the logic in question?
> 
> When using the top-down resource assignment strategy, coreboot assigns the following resources for an Intel iGPU (on an Alderlake-N board, in this example):
> 
> BAR (0x10): 0xbf000000 - 0xbfffffff BAR (0x18): 0xd0000000 - 0xdfffffff BAR (0x20): 0xffc0 - 0xffff (IO)
> 
> The IO BAR is the one that causes the issue.
> 

I see. So the issue seems to be that the EDK2 code marks 'rejected' PCI devices by setting all programmable BAR bits to `0x1`. As it happens, the valid value chosen for this I/O BAR also has all programmable BAR bits set to `0x1`, given that it is at the top of the I/O space.

If `TestValue` equals `0x0000FFC1`, it might just mean that the I/O address decoder in the device only supports 16-bits of I/O space, and so the higher bits are stuck to `0x0` too.

A better approach here would be to invert the logic in `IsPciDeviceRejected()`, and consider a PCI device rejected only if all BARs are 'unprogrammed'. E.g.,

```diff
--- a/MdeModulePkg/Bus/Pci/PciBusDxe/PciEnumeratorSupport.c
+++ b/MdeModulePkg/Bus/Pci/PciBusDxe/PciEnumeratorSupport.c
@@ -2735,6 +2735,7 @@ IsPciDeviceRejected (
   UINT32      OldValue;
   UINT32      Mask;
   UINT8       BarOffset;
+  BOOLEAN     Rejected;
 
   //
   // PPB should be skip!
@@ -2743,6 +2744,7 @@ IsPciDeviceRejected (
     return FALSE;
   }
 
+  Rejected = FALSE;
   if (IS_CARDBUS_BRIDGE (&PciIoDevice->Pci)) {
     //
     // Only test base registers for P2C
@@ -2754,16 +2756,18 @@ IsPciDeviceRejected (
         continue;
       }
 
+      Rejected  = TRUE;
       TestValue = TestValue & Mask;
-      if ((TestValue != 0) && (TestValue == (OldValue & Mask))) {
+      if ((TestValue == 0) || (TestValue != (OldValue & Mask))) {
         //
-        // The bar isn't programed, so it should be rejected
+        // The BAR is programmed, so the PCI device was not rejected
         //
-        return TRUE;
+        Rejected = FALSE;
+        break;
       }
     }
 
-    return FALSE;
+    return Rejected;
   }
 
   for (BarOffset = 0x14; BarOffset <= 0x24; BarOffset += sizeof (UINT32)) {
@@ -2775,14 +2779,16 @@ IsPciDeviceRejected (
       continue;
     }
 
+    Rejected = TRUE;
     if ((TestValue & 0x01) != 0) {
       //
       // IO Bar
       //
       Mask      = 0xFFFFFFFC;
       TestValue = TestValue & Mask;
-      if ((TestValue != 0) && (TestValue == (OldValue & Mask))) {
-        return TRUE;
+      if ((TestValue == 0) || (TestValue != (OldValue & Mask))) {
+        Rejected = FALSE;
+        break;
       }
     } else {
       //
@@ -2791,32 +2797,29 @@ IsPciDeviceRejected (
       Mask      = 0xFFFFFFF0;
       TestValue = TestValue & Mask;
 
+      if ((TestValue == 0) || (TestValue != (OldValue & Mask))) {
+        Rejected = FALSE;
+        break;
+      }
+
       if ((TestValue & 0x07) == 0x04) {
         //
         // Mem64 or PMem64
         //
         BarOffset += sizeof (UINT32);
-        if ((TestValue != 0) && (TestValue == (OldValue & Mask))) {
-          //
-          // Test its high 32-Bit BAR
-          //
-          Status = BarExisted (PciIoDevice, BarOffset, &TestValue, &OldValue);
-          if (TestValue == OldValue) {
-            return TRUE;
-          }
-        }
-      } else {
         //
-        // Mem32 or PMem32
+        // Test its high 32-Bit BAR
         //
-        if ((TestValue != 0) && (TestValue == (OldValue & Mask))) {
-          return TRUE;
+        Status = BarExisted (PciIoDevice, BarOffset, &TestValue, &OldValue);
+        if (TestValue != OldValue) {
+          Rejected = FALSE;
+          break;
         }
       }
     }
   }
 
-  return FALSE;
+  return Rejected;
 }
 
 /**
```

MrChromebox: You suggested:
```
if ((TestValue == 0) || (TestValue != (OldValue & Mask))) {
    //
    // The BAR is programmed, so the PCI device was not rejected
    //
    Rejected = FALSE;
    break;
}
```
while I think conceptually that reversing the rejection logic makes better sense (ie, a device is not rejected as long as it has one valid BAR), I still don't think the conditions are correct.
```
if ((TestValue & Mask) != 0)
    return FALSE;
```
should be all that's needed. A BAR that is non-zero when masked has been programmed. I can't find any documentation in the PCI spec to support comparing the original value and test values as is currently done.

MrChromebox:
so reversing the logic doesn't work out, since the rejection code is only looking at secondary BARs (ie, it skips BAR0 at 0x10), which means that any device with a single BAR would be rejected.

I think the least-worst (and least intrusive) solution is to just fix the comparison logic for the case of a 16-bit decoded IO BAR, which I have done in #11339, and verified fixes the issue

PR #10712, opened 2025-02-01
UefiPayloadPkg: Update Elf segment/section syncup process

ChaselChiu:
Original ELF segment/section syncup process having layout assumption which has caused build failure when layout changing.

New process will parse both existing segments and sections firstly to support different layout.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions

### DIFF
--- a/UefiPayloadPkg/Tools/ElfFv.py
+++ b/UefiPayloadPkg/Tools/ElfFv.py
@@ -331,23 +331,68 @@ def ModifyPHSegmentOffset64(NewUPLEntry, ElfHeaderOffset, PHSegmentName):
         SHData = SHentry[(i * elf_header.e_shentsize): (i * elf_header.e_shentsize) + elf_header.e_shentsize]
         unpacked_SectionHeader = ElfSectionHeader64.unpack(SHData)
         SHdrs.append(unpacked_SectionHeader)
+
+    # retrieve the .shstrtab section for parsing rest of the sections
+    shstrtab_section = NewUPLEntry[SHdrs[elf_header.e_shstrndx].sh_offset:SHdrs[elf_header.e_shstrndx].sh_offset+SHdrs[elf_header.e_shstrndx].sh_size]
+    shstrtab_strings = shstrtab_section.decode('utf-8', errors='ignore')
     for i in range(elf_header.e_phnum):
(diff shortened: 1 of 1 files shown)


UefiPayloadPkg/Tools/ElfFv.py
```diff
@@ -331,23 +331,68 @@ def ModifyPHSegmentOffset64(NewUPLEntry, ElfHeaderOffset, PHSegmentName):
...
+            print ('sync data segment and section:')
+            for i in range (elf_header.e_phnum):
+                # data segment type = PT_LOAD(1) and flags = R(4) + W(2) = 6
+                if PHdrs[i].p_type == 1 and PHdrs[i].p_flags == 6:
+                    if PHdrs[i].p_offset != dynamic_Segment_offset_found:
+                        PHdrs[i].p_offset = section_offset_found
+                        PHdrs[i].p_paddr = section_addr_found
+                        print ("PHdrs[%d].p_offset %x" % (i, PHdrs[i].p_offset))
+                        print ("PHdrs[%d].p_paddr %x" % (i, PHdrs[i].p_paddr))
+
```
ardbiesheuvel, line 395: This still assumes that here is a 1:1 correspondence between the `.text`, `.data` and `.dynamic` sections and the program headers, but this is not guaranteed.

It might be better to use your own linker script for UefiPayloadPkg so that you can keep it in sync with the python script. And of course, it would be best to simply require LLVM to be installed - it really doesn't make sense to use python for this; it is perfectly reasonable to require an ELF toolchain when using the ELF payload type for UPL.

ChaselChiu: I thought text, data and dynamic segments are very common for the most of the use cases, is it ok to keep 3 of them there in PHDRS? Or is there a way we could override it by UefiPayloadPkg BuildOptions section? It may be too much to diverge GccBase.lds just for adding dynamic segment to PHDRS.

ardbiesheuvel: Just adding the `.dynamic` section is fine, although it isn't needed for the PE/COFF conversion by GenFw. But if more changes are needed to GccBase.lds, you should find a way to override it in the DLINK flags.

ardbiesheuvel (approved):

PR #10722, opened 2025-02-05
PcatRealTimeClockRuntimeDxe: Preserve VRT bit in StatusD

PatrickRudolph:
# Description

On recent AMD platforms the VRT bit in the StatusD register is read-writeable. Adjust the code to not clear it as it would RtcWaitToUpdate() to return an error.

On platforms where the VRT bit is read-only the introduced code does the same as before.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Booted on AMD/glinda and found PcRtc Dxe working.

## Integration Instructions

N/A


PcAtChipsetPkg/PcatRealTimeClockRuntimeDxe/PcRtc.c
```diff
@@ -305,7 +305,11 @@ PcRtcInit (
   //
   // Clear RTC register D
   //
-  RegisterD.Data = FixedPcdGet8 (PcdInitialValueRtcRegisterD);
+  // On most platforms VRT is read-only, but when it's not it must
+  // be preserved to prevent failing in RtcWaitToUpdate().
+  //
+  RegisterD.Data = RtcRead (RTC_ADDRESS_REGISTER_D);
+  RegisterD.Data = FixedPcdGet8 (PcdInitialValueRtcRegisterD) | (RegisterD.Data & BIT7);
```
mdkinney, line 312: If BIT7 is set in PCD, BIT7 will be set no matter what state RegisterD is in.  Is this the expected behavior?

PatrickRudolph: Not sure if or when BIT7 should be set. Updated the commit to not set BIT7 by PcdInitialValueRtcRegisterD.

PatrickRudolph:
Can this get merged?

PcAtChipsetPkg/PcatRealTimeClockRuntimeDxe/PcRtc.c
```diff
@@ -305,7 +305,11 @@ PcRtcInit (
   //
   // Clear RTC register D
   //
-  RegisterD.Data = FixedPcdGet8 (PcdInitialValueRtcRegisterD);
+  // On most platforms VRT is read-only, but when it's not it must
+  // be preserved to prevent failing in RtcWaitToUpdate().
+  //
+  RegisterD.Data = RtcRead (RTC_ADDRESS_REGISTER_D);
+  RegisterD.Data = (FixedPcdGet8 (PcdInitialValueRtcRegisterD) & 0x7f)| (RegisterD.Data & BIT7);
```
niruiyu, line 312: I checked the PCH spec (https://www.intel.com/content/dam/www/public/us/en/documents/datasheets/9-series-chipset-pch-datasheet.pdf) and it has following explanation of VRT bit in Register D:

Valid RAM and Time Bit VRT - R/W.
0 - This bit should always be written as a 0 for write cycle, however it will return a 1 for read cycles.
1 - This bit is hardwired to 1 in the RTC power well.


So that means on a 1 is written with your changes.
That's a behavior change!

PatrickRudolph: Thanks for the review. The old ICH and PCH datasheets seem to be the only source for RTC register 0xd. It's not defined on newer Intel platforms.
To not break the older platforms I'll find another way of fixing this for recent AMD chipsets.

PR #10725, opened 2025-02-06
ArmPkg: ArmFfaStandaloneMmLib: Fix non-FFA path

kuqin12:
# Description

ArmFfaLibCommonInit will return EFI_UNSUPPORTED when there is no FFA supported on the platform. This is expected behavior. However, the return of error code will incur program asserts.

This change fixed the non-FFA path for the Standalone MM instance.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This change was tested on QEMU SBSA platform that does not have FFA support.

## Integration Instructions

N/A


ArmPkg/Library/ArmFfaLib/ArmFfaStandaloneMmLib.c
```diff
@@ -68,5 +68,26 @@ ArmFfaStandaloneMmLibConstructor (
   IN EFI_MM_SYSTEM_TABLE  *MmSystemTable
   )
 {
-  return ArmFfaLibCommonInit ();
+  EFI_STATUS  Status;
+
+  Status = ArmFfaLibCommonInit ();
+  if (EFI_ERROR (Status)) {
```
samimujawar, line 74: Thank you for this fix. I have a minor suggestion to combine the conditions in a single if statement.
```
if ((EFI_ERROR (Status) && (Status == EFI_UNSUPPORTED)) {
 ...
```

samimujawar (changes requested):

ArmPkg/Library/ArmFfaLib/ArmFfaStandaloneMmLib.c
ardbiesheuvel: Agree with Sami but in that case, you can drop the `EFI_ERROR (Status)` test entirely.

However, maybe it would make sense to add a DEBUG() statement here that logs `Status` upon errors?

kuqin12: Thanks for the feedback. It makes sense. I updated the routine accordingly. Please let me know if you have any other feedback.

ArmPkg/Library/ArmFfaLib/ArmFfaStandaloneMmLib.c
```diff
@@ -68,5 +68,25 @@ ArmFfaStandaloneMmLibConstructor (
...
+  if (Status == EFI_UNSUPPORTED) {
+    /*
+     * EFI_UNSUPPORTED return from ArmFfaLibCommonInit() means
+     * FF-A interface isn't supported.
+     * However, It doesn't make failure of loading driver/library instance
+     * (i.e) ArmPkg's MmCommunication Dxe/PEI Driver uses as well as SpmMm.
+     * So If FF-A is not supported the the MmCommunication Dxe/PEI falls
+     * back to SpmMm.
+     * For this case, return EFI_SUCCESS.
+     */
```
ardbiesheuvel, line 83: Please fix the grammar and capitalization on this comment block (and the double `the`). Also `i.e.` does not mean 'for example'.

kuqin12: Agreed. Updated comment. Please let me know if any other suggestions.

samimujawar (approved):

ardbiesheuvel (approved):

kuqin12:
Thanks for approving the change, can we merge this change?

lgao4:
This is a bug fix to handle the wrong return status. I think it can be merged for the stable tag 202502.

mdkinney (approved):

makubacki (approved):

PR #10732, opened 2025-02-07
MdeModulePkg/Variable: Init var policy after SMM variable is ready

makubacki:
# Description

On a MM system, the main UEFI variable logic resides in MMRAM. In that case, the variable policy logic in `VarCheckPolicyLib`, such as `VarCheckPolicyLibStandaloneMm` is linked against the MM driver also in that case `VariableStandaloneMm`.

The MM variable driver indicates its presence to the RT DXE driver via `gEfiSmmVariableProtocolGuid` to indicate variable read support is available from MM. This triggers installation of the variable architectural protocol in DXE.

Today, variable policy is initialized by calling
`VariablePolicySmmDxeMain()` in `VariableSmmRuntimeInitialize()`. In turn, this installs `gEdkiiVariablePolicyProtocolGuid`. Functions in `gEdkiiVariablePolicyProtocolGuid` may trigger MMIs. However, it is possible that the MM variable driver which is linked against the code with the variable policy MMI handlers (i.e. `VarCheckPolicyLib`) is not loaded yet.

Therefore, this change moves invocation of
`VariablePolicySmmDxeMain()` to `SmmVariableReady()` which is called on installation of `gEfiSmmVariableProtocolGuid` indicating variable MM services are ready. `gEdkiiVariablePolicyProtocolGuid` is still installed prior to the variable architectural protocol being installed.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- Verify variable policy is initialized as expected on QEMU Q35/SBSA and a physical Intel system.
- Verify the image handle passed to `VariablePolicySmmDxeMain()`.

## Integration Instructions

Some drivers may dispatch differently depending on platform MM readiness if those drivers depend on `gEdkiiVariablePolicyProtocolGuid`. This is not considered breaking as that is an inherent expectation in dependency expression based dispatch.


makubacki:
@lgao4, please review when you have a moment.

lgao4 (approved):

PR #10734, opened 2025-02-07
MdeModulePkg/UefiBootManagerLib: Apply var policy to HDDP

makubacki:
# Description

Applies variable policy to the "HDDP" UEFI variable in the mBmHardDriveBootVariableGuid vendor namespace to ensure a minimum allowed size and expected attributes are present.

A protocol notify is used to handle different platform scenarios where this instance of UefiBootManagerLib may be linked.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- Verified `"HDDP"` policy is registered as requested.
- Verified the protocol notify function is invoked if the variable policy protocol is installed.
- Verified the notify gracefully handles the case when policy is already applied to the variable.

## Integration Instructions

This applies a variable policy so that the variable is set as expected in current implementation. No further work should be necessary if the current implementation is used.

### DIFF
--- a/MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
+++ b/MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
@@ -2668,3 +2668,80 @@ EfiBootManagerGetNextLoadOptionDevicePath (
 {
   return BmGetNextLoadOptionDevicePath (FilePath, FullPath);
 }
+
+/**
+  Variable policy protocol installation notification.
+
+  @param[in]    Event           The notification event.
+  @param[in]    Context         Pointer to the context registered when the event is created. Not used.
+
+**/
+VOID
+EFIAPI
+OnVariablePolicyNotification (
+  IN  EFI_EVENT  Event,
+  IN  VOID       *Context
+  )
+{
+  EFI_STATUS                      Status;
+  EDKII_VARIABLE_POLICY_PROTOCOL  *VariablePolicy = NULL;
+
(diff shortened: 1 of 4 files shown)


makubacki:
@lgao4 & @niruiyu, please review when you have a moment.

lgao4:
> @lgao4 & @niruiyu, please review when you have a moment.

Do you find the invalid usage for this variable? So, you think it is necessary to add this policy check for this variable.

makubacki:
> Do you find the invalid usage for this variable? So, you think it is necessary to add this policy check for this variable.

This is proactively setting policy to restrict variable properties based on how it is actually used. The most important restriction here is limiting the attributes to `EFI_VARIABLE_BOOTSERVICE_ACCESS | EFI_VARIABLE_NON_VOLATILE` to prevent what is called "variable pinning" where an attacker creates an authenticated variable with a cert that is unknown to the firmware and OS. An attacker can also delete an unprotected variable and recreate the variable as a "pinned" variable. Once "pinned", there is no standard provision for altering the variable.

This is a variant of [VU#758382 - Unauthorized modification of UEFI variables in UEFI systems (cert.org)](https://www.kb.cert.org/vuls/id/758382) but discussed more recently on Slide 33 in [Data-Only Attacks Against UEFI BIOS](https://github.com/flothrone/smm/blob/main/ZN2021%20Dataonly%20attacks%20BIOS%20-%20Ermolov.pdf).

For this reason, some implementations ensure 100% UEFI variable policy coverage. This is to support safer usage of the variable in the upstream implementation since it is the producer/consumer of this variable. I may post some similar PRs soon for other variables.

lgao4 (approved):

PR #10735, opened 2025-02-07
MdeModulePkg: Add the EFI_RESOURCE_ATTRIBUTE_SPECIAL_PURPOSE attribute

makubacki:
# Description

Add the `EFI_RESOURCE_ATTRIBUTE_SPECIAL_PURPOSE` resource attribute as per the PI 1.9 spec. This flag is used to indicate that the memory should be treated as special purpose memory (SPM).

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- Test in system that produces resource descriptor HOBs with `EFI_RESOURCE_ATTRIBUTE_SPECIAL_PURPOSE`

## Integration Instructions

Use the special purpose attribute based on platform need.

### DIFF
--- a/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
+++ b/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
@@ -26,7 +26,8 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
                                        EFI_RESOURCE_ATTRIBUTE_16_BIT_IO           | \
                                        EFI_RESOURCE_ATTRIBUTE_32_BIT_IO           | \
                                        EFI_RESOURCE_ATTRIBUTE_64_BIT_IO           | \
-                                       EFI_RESOURCE_ATTRIBUTE_PERSISTENT          )
+                                       EFI_RESOURCE_ATTRIBUTE_PERSISTENT          | \
+                                       EFI_RESOURCE_ATTRIBUTE_SPECIAL_PURPOSE     )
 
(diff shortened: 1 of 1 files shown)


lgao4 (approved):

MdeModulePkg/Core/Dxe/Gcd/Gcd.c
```diff
@@ -690,6 +691,8 @@ ConverToCpuArchAttributes (
     CpuArchAttributes |= EFI_MEMORY_UCE;
   } else if ((Attributes & EFI_MEMORY_WP) == EFI_MEMORY_WP) {
     CpuArchAttributes |= EFI_MEMORY_WP;
+  } else if ((Attributes & EFI_MEMORY_SP) == EFI_MEMORY_SP) {
```
os-d, line 694: This is not needed, right, because EFI_MEMORY_SP is already part of EFI_MEMORY_ATTRIBUTE_MASK

os-d (approved):

MdeModulePkg/Core/Dxe/Gcd/Gcd.c
makubacki: Removed

lgao4:
> /azp run

Can this command trig CI?

makubacki:
> > /azp run
> 
> Can this command trig CI?

Yes. Specifically, the pipelines. In this case, I saw the pipelines were stuck in a "pending" state, so I kicked them with that.

PR #10738, opened 2025-02-08
MdePkg/DynamicStackCookieEntryPointLib: Drop execute-in-place versions

ardbiesheuvel:
SEC, PEI_CORE and PEIM type modules generally execute in place from memory that is not writable. This means that it is not generally possible to use an entrypoint implementation that stores a dynamically generated stack cookie into a global variable. For PEIMs in particular, there may be other options, such as a DEPEX on the permanent memory PPI, but the current dynamic implementations of the stack checking library entrypoints for PEI_CORE and PEIM modules must not be used as-is.

So remove them, and update the readme accordingly.


os-d:
Thanks Ard for cleaning this up. A few thoughts:
- The *StackCheckEntryPointNull.c files definitely should be removed, those were from the v1 of this PR and now not referenced by any INF in the v2; I just missed deleting a few of them. There is also one in the UefiApplication folder, will you delete that as well?
- I agree with PEI_CORE not being able to use dynamic cookies and XIP PEIMs also not being able to (I think part of the earlier discussion asked for them to be included for completeness, but I don’t recall). For the PEIM entry point, as you noted, it could have a dependency on the memory installed protocol, that should guarantee that that PEIM is dispatched into RAM, not XIP, right? That then gives flexibility to a platform to decide that it’s post memory-init PEIMs can use that entrypoint (a bit more annoying to implement module by module but doable). If my assumption is incorrect here, then I am okay to drop PEIM as well, this is why I did not include these in MdeLibs.dsc.inc.

MdePkg/Library/StackCheckLib/Readme.md
```diff
@@ -195,20 +195,9 @@ This will cause `MdeLibs.dsc.inc` to not link `StackCheckLibNull` and instead li
 stack cookie updating versions of `DxeCoreEntryPoint`, `StandaloneMmDriverEntryPoint`,
 `UefiApplicationEntryPoint`, and `UefiDriverEntryPoint`.
 
-Because edk2 does not implement exception handling for `SEC` and `PEI_CORE`, `MdeLibs.dsc.inc`
-uses `StackCheckLibNull` for these phases always. As a result, dynamic stack cookies are also
```
os-d, line 199: I think we should keep the section on StackCheckLib vs StackCheckLibNull, that still applies to these phases, even without dynamic stack cookies, because they will still have static stack cookies and stack cookie checking is orthogonal to how the cookie value was chosen. The part about exception handling for SEC and PEI_CORE was also a requested addition from Mike Kinney who pointed out platforms could implement their own exception handling there.

os-d (approved):

ardbiesheuvel:
> Thanks Ard for cleaning this up. A few thoughts:
> 
> * The *StackCheckEntryPointNull.c files definitely should be removed, those were from the v1 of this PR and now not referenced by any INF in the v2; I just missed deleting a few of them. There is also one in the UefiApplication folder, will you delete that as well?

Will do.

> * I agree with PEI_CORE not being able to use dynamic cookies and XIP PEIMs also not being able to (I think part of the earlier discussion asked for them to be included for completeness, but I don’t recall). For the PEIM entry point, as you noted, it could have a dependency on the memory installed protocol, that should guarantee that that PEIM is dispatched into RAM, not XIP, right? That then gives flexibility to a platform to decide that it’s post memory-init PEIMs can use that entrypoint (a bit more annoying to implement module by module but doable). If my assumption is incorrect here, then I am okay to drop PEIM as well, this is why I did not include these in MdeLibs.dsc.inc.

The question is really whether we want to deviate from the general rule that PEIMs must not have global variables. I'd prefer to explore other ways of doing this that work for all PEIMs not just once that were dispatched straight into RAM (as opposed to being shadowed after dispatch - this distinction is important)

MdePkg/Library/StackCheckLib/Readme.md
ardbiesheuvel: I'll tweak this, but I'll note that the prose about exception handling is duplicated here so I'm not sure it needs to be repeated.

makubacki:
> The question is really whether we want to deviate from the general rule that PEIMs must not have global variables. I'd prefer to explore other ways of doing this that work for all PEIMs not just once that were dispatched straight into RAM (as opposed to being shadowed after dispatch - this distinction is important)

I'd prefer to stick with this as well. The potential for error is high - either new code uses writing to any global to justify doing so in pre-memory without being aware of the limitations or code is unknowingly linked to pre-mem modules during platform integration. This is further exacerbated by cache eviction subtly not retaining data leading to incorrect behavior that goes down the "default" data value path in many cases.

os-d:
> > The question is really whether we want to deviate from the general rule that PEIMs must not have global variables. I'd prefer to explore other ways of doing this that work for all PEIMs not just once that were dispatched straight into RAM (as opposed to being shadowed after dispatch - this distinction is important)
> 
> I'd prefer to stick with this as well. The potential for error is high - either new code uses writing to any global to justify doing so in pre-memory without being aware of the limitations or code is unknowingly linked to pre-mem modules during platform integration. This is further exacerbated by cache eviction subtly not retaining data leading to incorrect behavior that goes down the "default" data value path in many cases.

I agree with you both. I think that architectures that do not have the CAR restriction have the option of skipping PEI entirely and so leaving PEI as designed with no global variable writing makes sense.

lgao4 (approved):

lgao4:
I agree to merge this change for stable tag 202502.

mdkinney (approved):

leiflindholm (approved):

makubacki (approved):

PR #10751, opened 2025-02-12
NetworkPkg: Resolved Coverity Issues in DHCP4Dxe and DHCP6Dxe

santhoshkumarvenkatachalam:
# Description

1.DhcpChooseOffer(CHECKED_RETURN)
Calling "DhcpValidateOptions" without checking return value as is done elsewhere 6 out of 7 times. 2.DhcpSendMessage(FORWARD_NULL)
SeedHead can be dereferencing the null pointer when Seed is NULL. 3.Dhcp6GetMappingTimeOut(OVERFLOW_BEFORE_WIDEN)
Potentially overflowing expression "10000000U * DadXmits.DupAddrDetectTransmits" with type "unsigned int"(32 bits) 4.Dhcp6DepriveAddress(INTEGER_OVERFLOW)
Expression "AddressCount - 1U", where "AddressCount" is known to be equal to 0, underflows the type that receives it. 5.Dhcp6CalculateLeaseTime(INTEGER_OVERFLOW)
When IaCb->Ia->IaAddressCount is less than 0 ,Expression "MinLt * 8U", overflows the type that receives it

### DIFF
--- a/NetworkPkg/Dhcp4Dxe/Dhcp4Option.c
+++ b/NetworkPkg/Dhcp4Dxe/Dhcp4Option.c
@@ -184,6 +184,9 @@ DhcpOptionIsValid (
   }
 
   ASSERT (Unit != 0);
+  if (Unit == 0) {
+    return FALSE;
+  }
 
   //
   // Validate that the option appears in the full units.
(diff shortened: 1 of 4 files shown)


NetworkPkg/Dhcp4Dxe/Dhcp4Option.c
```diff
@@ -184,6 +184,9 @@ DhcpOptionIsValid (
   }
 
   ASSERT (Unit != 0);
+  if (Unit == 0) {
```
leiflindholm, line 187: If the exercise is improving the code, as opposed to making the static analysis happy, could we instead add a default case to the switch statement?
Then we could also get rid of the Unit = 0 initialization.

leiflindholm:
The only relation between these fixes is coverity. That definitely makes sense to group them in a PR, but it would improve both review process and resulting git history if they were further broken up into separate commits for separate errors.
(As in, the same error in multiple locations as one commit.)

PR #10752, opened 2025-02-12
NetworkPkg:Resolved Coverity Issues in UefiPxeBcDxe

santhoshkumarvenkatachalam:
# Description

1.PxeBcHandleDhcp4Offer
        Expression "Private->SelectIndex - 1", where "Private->SelectIndex" is known to be equal to 0, underflows the type that receives it. 

2.PxeBcDhcp4Discover
          Directly dereferencing pointer "Token.Packet" which might have NULL , As Token.Packet is checked against NULL in the function Exit line no 1587. 

3.PxeBcHandleDhcp6Offer
            Expression "Private->SelectIndex - 1", where "Private->SelectIndex" is known to be equal to 0,underflows the type that receives it.

 4.PxeBcDhcp6Sarr
            Potentially overflowing expression "10000000U*DadXmits.DupAddrDetectTransmits" with type "unsigned int" (32 bits, unsigned)

5.PxeBcDriver.c
          Private Data might be dereferencing the NULL pointer in some scenario, And also private is checked against null in function's Exit. For Example: below code checked private with NULL pointer in PxeBcStart()
  if (FirstStart && Private != NULL) {
            FreePool (Private):

6.PxeBcDhcp4CallBack:
             case Dhcp4SendRequest is not reachable ,because PxeBcDhcp4CallBack returned in the beginning if Dhcp4Event is Dhcp4SendRequest.

### DIFF
--- a/NetworkPkg/UefiPxeBcDxe/PxeBcDhcp4.c
+++ b/NetworkPkg/UefiPxeBcDxe/PxeBcDhcp4.c
@@ -1025,8 +1025,13 @@ PxeBcHandleDhcp4Offer (
   EFI_PXE_BASE_CODE_MODE    *Mode;
   EFI_DHCP4_PACKET          *Ack;
 
+  SelectIndex = 0;
+
   ASSERT (Private->SelectIndex > 0);
-  SelectIndex = (UINT32)(Private->SelectIndex - 1);
+  if (Private->SelectIndex > 0) {
+    SelectIndex = (UINT32)(Private->SelectIndex - 1);
+  }
+
   ASSERT (SelectIndex < PXEBC_OFFER_MAX_NUM);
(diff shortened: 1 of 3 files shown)


NetworkPkg/UefiPxeBcDxe/PxeBcDhcp4.c
```diff
@@ -1250,15 +1255,6 @@ PxeBcDhcp4CallBack (
       //
       CopyMem (&Mode->DhcpDiscover.Dhcpv4, &Packet->Dhcp4, Packet->Length);
 
-    case Dhcp4SendRequest:
```
leiflindholm, line 1253: I'm not seeing any preceding references to Dhcp4SendRequest.

PR #10753, opened 2025-02-12
NetworkPkg:Resolved Coverity Issues reported in SNP Dxe.

santhoshkumarvenkatachalam:
# Description
1.ComponentName.c (Overflow)
               Expression "OffSet--", which is equal to 18446744073709551615, where "OffSet" is known to be equal to 0, underflows the type that receives it.

2.SnpUndi32GetStatus,SnpUndi32Initialize,SnpUndi32Transmit (Deadcode)
               Execution is not reaching as always snp is not equal to NULL As This pointer is compared with NULL in function start Macro EFI_SIMPLE_NETWORK_DEV_FROM_THIS always return address in This pointer.

3.PxeInit(Overflow)
             Expression "Snp->TxRxBufferSize - 1U", where "Snp->TxRxBufferSize" is known to be equal to 0, underflows the type that receives it.

### DIFF
--- a/NetworkPkg/SnpDxe/ComponentName.c
+++ b/NetworkPkg/SnpDxe/ComponentName.c
@@ -267,7 +267,12 @@ UpdateName (
   //
   // Remove the last '-'
   //
-  OffSet--;
+  if (OffSet > 0) {
+    OffSet--;
+  } else {
+    OffSet = 0;
+  }
+
   OffSet += UnicodeSPrint (
               HandleName + OffSet,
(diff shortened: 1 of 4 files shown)


leiflindholm:
Same comment as for #10751 and #10752:

The only relation between these fixes is coverity. That definitely makes sense to group them in a PR, but it would improve both review process and resulting git history if they were further broken up into separate commits for separate errors.
(As in, the same error in multiple locations as one commit.)

On the whole, I think it would make sense to merge these three PRs into a single "fixing coverity reports for NetworkPkg" PR and close the redundant ones.

PR #10754, opened 2025-02-12
BaseTools: Enable GenFv Rebase function

YuweiChen1110:
# Description

This patch is used to enable GenFv Rebase function. 
Developer can add "Xip=TRUE" in FDF file Rule section to enable Rebase feature. After enabling, the enabled type module will be rebased while others' will not.

Rebase behavior:
[1]. If "Xip" is defined in FDF file Rule section, then the target module_type's module will be rebased when consumed by GenFV
(Added from this PR).
[2]. Else if "FvForceRebase" is defined in FDF FV section, then the all modules in this FV will be rebased when consumed by GenFV (designed in Spec).

```
Example:
[Rule.Common.PEI_CORE]
  FILE PEI_CORE = $(NAMED_GUID) {
    PE32     PE32   Align=Auto   Xip=TRUE   $(INF_OUTPUT)/$(MODULE_NAME).efi
    UI       STRING ="$(MODULE_NAME)" Optional
    VERSION  STRING ="$(INF_VERSION)" Optional BUILD_NUM=$(BUILD_NUMBER)
  }
```

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Enabled Rebase function and run the OVMFPKG in both windows and linux environment and boot in QEMU.

## Integration Instructions

No changes needed though code may use the "Xip=TRUE" to enable Rebase function as needed.

### DIFF
--- a/BaseTools/Source/C/GenFv/GenFvInternalLib.c
+++ b/BaseTools/Source/C/GenFv/GenFvInternalLib.c
@@ -31,6 +31,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <io.h>
 #endif
 #include <assert.h>
+#include <ctype.h>
 
 #include <Guid/FfsSectionAlignmentPadding.h>
 
@@ -131,6 +132,10 @@ BOOLEAN                     mIsLargeFfs = FALSE;
 
 EFI_PHYSICAL_ADDRESS mFvBaseAddress[0x10];
 UINT32               mFvBaseAddressNumber = 0;
+UINTN                ModuleTableCount = 0;
+UINTN                ModuleTypeCount = 0;
+STATIC MODULE_XIP_INFO      ModuleXipList[MAX_MODULES_TYPE];
+STATIC FILE_MODULE_TYPE     FileModuleTable[MAX_NUMBER_OF_FILES_IN_FV];
 
 EFI_STATUS
 ParseFvInf (
@@ -2904,6 +2909,7 @@ Returns:
     //
     // Add the file
     //
+    StoreModuleType(mFvDataInfo.FvFiles[Index]);
     Status = AddFile (&FvImageMemoryFile, &mFvDataInfo, Index, &VtfFileImage, FvMapFile, FvReportFile);
 
     //
@@ -3510,6 +3516,8 @@ Returns:
   UINT32                                FfsHeaderSize;
   UINT32                                CurSecHdrSize;
 
+  CONST CHAR8 *ModuleType;
+  BOOLEAN XipEnabled = FALSE;
   Index              = 0;
   MemoryImagePointer = NULL;
   TEImageHeader      = NULL;
@@ -3533,6 +3541,18 @@ Returns:
     return EFI_SUCCESS;
   }
 
+  ModuleType = GetFileModuleType(FileName);
+  if (FvInfo->ForceRebase == 1) {
+      for (UINTN i = 0; i < ModuleTypeCount; i++) {
+          if (strcmp(ModuleXipList[i].ModuleName, ModuleType) == 0) {
+              XipEnabled = ModuleXipList[i].XipEnabled;
+              break;
+          }
+      }
+      if (!XipEnabled) {
+          return EFI_SUCCESS;
+      }
+  }
 
   XipBase = FvInfo->BaseAddress + XipOffset;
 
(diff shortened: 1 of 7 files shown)


os-d:
Can the FDF spec also be updated to include this functionality with a greater description? E.g. if a build rule has this set for some module types, what happens to an FV if FvForceRebase = FALSE, does Xip=TRUE take precedence, etc.

os-d:
@YuweiChen1110 , can you share the OVMF changes you made to test this? I tried enabling this on one of my platforms to test, but I was not getting the expected behavior. I thought that if I only set SEC, PEIM, and PEI_CORE module rules to include `Xip = TRUE` that I could have one FV that contained both PEI and DXE modules (just to test this) however, I still fail with:

```
GenFv: ERROR 3000: Invalid
INFO -   PE image Section-Alignment and File-Alignment do not match 
```

Which looks like the DXE modules are still being attempted to be rebased. I tried setting `FvForceRebase = FALSE` for that FV, with the PEI types with `Xip=TRUE` set, but then no modules were rebased. Finally I tried setting `Xip=FALSE` explicitly for the DXE module types, but still got the above error.

Some guidance would be appreciated, thanks! Perhaps I am misunderstanding the problem this is trying to resolve?

mdkinney:
Yes.  The FDF Spec will need to be updated.  First, the functionality of this needs to be verified and the syntax used to specify the XIP requirement is specified needs to be reviewed and accepted.  Then the doc and the code can be updated to match.

There are also the `RELOCS_STRIPPED`, `RELOCS_RETAINED`, `FIXED`, `CHECKSUM`, and alignment attributes associated with an FFS file.  Here is an example of them used in a rule: 

```
[Rule.Common.SEC]
  FILE SEC = $(NAMED_GUID) RELOCS_STRIPPED FIXED {
    PE32     PE32    Align = Auto     $(INF_OUTPUT)/$(MODULE_NAME).efi
    UI       STRING ="$(MODULE_NAME)" Optional
    VERSION  STRING ="$(INF_VERSION)" Optional BUILD_NUM=$(BUILD_NUMBER)
  }
```

This proposed XIP syntax is in the scope of the PE32 section, but I a, wondering since `XIP` is related to relocations, should the `XIP` attribute be at the FFS scope with the `RELOCS_*` attributes.

os-d:
Ah, I figured it out. If `FvForceRebase = TRUE` is set and the Xip rules are there, then they are respected and you indeed can mix PEI and DXE modules in one FV (more generically modules that need to be rebased and those that don't have matching section and file alignment).

I think this is a good change and resolves a long standing BaseTools issue. It does seem a bit counterintuitive to me that if `FvForceRebase = TRUE` is set then the FV is not rebased as a whole, it only has the modules marked as XIP that are rebased (but I did not explore further, if no modules are marked XIP, does the FV then get rebased as a whole?). 

It seems to me that it would be a bit more intuitive to have `FvForceRebase = TRUE` still rebase all modules in an FV, as I believe the FDF spec envisions (though a quick search of the FDF spec reveals fairly few references to it at all). If FvForceRebase is not set or FALSE, then you would only rebase the modules marked as `Xip = TRUE`. I believe that still maintains compatibility with existing FDFs (if they have `FvForceRebase = TRUE`, the FV will be rebased, if not, it will not be).

Mainly I want to make sure we have the various combinations of options covered and there are not some odd combinations of options that do not produce the desired effect (as per my initial testing).

BaseTools/Source/C/GenFv/GenFvInternalLib.c
```diff
@@ -4375,3 +4395,143 @@ Routine Description:
...
+        if (Colon != NULL) {
+            *Colon = '\0';
+            CHAR8 *Module = Line;
+            CHAR8 *Value = Colon + 2;
+
+            Value[strcspn(Value, "\r\n")] = '\0';
+            if (ModuleTypeCount < MAX_MODULES_TYPE) {
+                strncpy(ModuleXipList[ModuleTypeCount].ModuleName, Module, sizeof(ModuleXipList[ModuleTypeCount].ModuleName) - 1);
+                ModuleXipList[ModuleTypeCount].ModuleName[sizeof(ModuleXipList[ModuleTypeCount].ModuleName) - 1] = '\0';
+                if ((_stricmp(Value, "True") == 0) || (_stricmp(Value, "true") == 0)) {
```
mhaeuser, line 4435: Why are there two cases for this case-insensitive comparison?

BaseTools/Source/C/GenFv/GenFvInternalLib.c
```diff
@@ -4375,3 +4395,143 @@ Routine Description:
...
+    CHAR8 XipFileName[MAX_LONG_FILE_PATH];
+    FILE  *InputFile;
+    CHAR8 Line[MAX_XIP_LINE_LENGTH];
+    CHAR8 *Colon;
+
+
+    strcpy(XipFileName, FilePath);
+    CHAR8 *LastSlash = strrchr(XipFileName, '\\');
+    if (LastSlash != NULL) {
+        strcpy(LastSlash + 1, "Module_Xip");
```
mhaeuser, line 4413: This appends to ``FilePath``, so may this exceed ``MAX_LONG_FILE_PATH``?

apop5:
@bexcran @lgao4 

Can you verify this is okay?

makubacki:
@YuweiChen1110, did you intend to abandon this change? It would be very helpful for us.

Issue #10727, opened 2025-02-06
[Bug]: <Unclosed BlockIo2 Protocol during DiskIoDriverBindingStart Error Handling>


kanagavels97:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### Code first?

- [ ] Yes

### What packages are impacted?

MdeModulePkg

### Current Behavior

During the DiskIoDriverBindingStart, both the BlockIo and BlockIo2 protocols are opened for the specified controller handle. The Instance->SharedWorkingBuffer is then allocated based on the BlockSize. If the BlockSize is zero, indicating that the handle does not support Read/Write operations, the buffer will not be allocated. A NULL check is performed on the SharedWorkingBuffer, and if it is NULL, the process jumps to the ErrorExit. In the ErrorExit section, only the BlockIo protocol is closed, while the BlockIo2 protocol remains open.

We are facing an issue with Uninstalling the BlockIo2 protocol during the DriverBindingStop process because it remains open.

**MdeModulePkg\Universal\DiskIoDxe\DiskIo.c**

_ErrorExit:
  if (EFI_ERROR (Status)) {
    if ((Instance != NULL) && (Instance->SharedWorkingBuffer != NULL)) {
      FreeAlignedPages (
        Instance->SharedWorkingBuffer,
        EFI_SIZE_TO_PAGES (PcdGet32 (PcdDiskIoDataBufferBlockNum) * Instance->BlockIo->Media->BlockSize)
        );
    }
    if (Instance != NULL) {
      FreePool (Instance);
    }
    gBS->CloseProtocol (
           ControllerHandle,
           &gEfiBlockIoProtocolGuid,
           This->DriverBindingHandle,
           ControllerHandle
           );
  }_

### Expected Behavior

The check needs to be added for the BlockIo2 protocol in the ErrorExit, and it should be closed properly if the BlockIo2 protocol is opened.

### Steps To Reproduce

1. Connect a UFS device that supports WLUN. This WLUN will be detected as a separate handle under the UFS host controller.
2. In my case, the UFS device supports WLUN (44h), which is used for the RPMB feature. According to the UFS specification, the RPMB 
    LUN does not support Read/Write operations.
3. Add UFS module to the bios
4. Flash the bios on SUT and boot to the shell.
5. Run the devtree command to identify the controller handle numbers.
6. In my case, the UFS Host controller number is 165.
[...]

PR #10755, opened 2025-02-12
MdeModulePkg: Add a check and close BlockIo2 protocol in DiskIoDriverBindingStart.

kanagavels97:
MdeModulePkg: Add a check and close BlockIo2 protocol in DiskIoDriverBindingStart.

The check needs to be added for the BlockIo2 protocol in the ErrorExit, and it should be closed properly if the BlockIo2 protocol is opened.

# Description

During the DiskIoDriverBindingStart, both the BlockIo and BlockIo2 protocols are opened for the specified controller handle. The Instance->SharedWorkingBuffer is then allocated based on the BlockSize. If the BlockSize is zero, indicating that the handle does not support Read/Write operations, the buffer will not be allocated. A NULL check is performed on the SharedWorkingBuffer, and if it is NULL, the process jumps to the ErrorExit. In the ErrorExit section, only the BlockIo protocol is closed, while the BlockIo2 protocol remains open.

### DIFF
--- a/MdeModulePkg/Universal/Disk/DiskIoDxe/DiskIo.c
+++ b/MdeModulePkg/Universal/Disk/DiskIoDxe/DiskIo.c
@@ -226,6 +226,15 @@ ErrorExit:
            This->DriverBindingHandle,
            ControllerHandle
            );
+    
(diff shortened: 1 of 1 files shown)


kanagavels97:
Closed

Issue #7000, opened 2017-05-18
Add -D NO_MSABI_VAARGS define to CC_FLAGS for X64 XCODE5 builds (Bugzilla Bug 561)


tianocore-issues:
This issue was created automatically with [bugzilla2github](https://github.com/makubacki/bugzilla2github/tree/bz2gh_tianocore)
# Bugzilla Bug 561
Date: 2017-05-18T20:26:21+00:00
From: @mdkinney
To:   @mdkinney
CC:   @ajfish, nobody

Last updated: 2017-05-24T20:19:05+00:00

tianocore-issues:
## Comment 2090
Date: 2017-05-18 20:26:21 +0000
From: @mdkinney

- Industry Specification: ---
- Release Observed: ['EDK II Master', 'UDK 2017']
- Releases to Fix: ['EDK II Master', 'UDK 2017']
- Target OS: ---
- Bugzilla Assignee(s): @mdkinney

Update BaseTools/Conf/tools_def.template to add the define

-D NO_MSABI_VAARGS 

To CC_FLAGS for X64 XCODE5 builds.

The llvm/clang compiler used in XCODE5 builds supports the
_ms_ versions of the vararg builtins, but the compiler 
generates build errors.

The recommendation from the XCODE5 exports is to never use
the _ms_ version of the vararg builtins.  This define
NO_MSABI_VARARGS is already supported in MdePkg/Include/Base.h
to uses the standard vararg builtins.

tianocore-issues:
## Comment 2102
Date: 2017-05-19 02:40:10 +0000
From: @mdkinney

Assign to owner

tianocore-issues:
## Comment 2103
Date: 2017-05-19 02:52:40 +0000
From: @mdkinney

Patch mail sent

tianocore-issues:
## Comment 2113
Date: 2017-05-19 18:08:01 +0000
From: @mdkinney

Commit

https://github.com/tianocore/edk2/commit/bdaced0bcf9a15817fbab4e7b98200cf6628b7a5

tianocore-issues:
## Comment 2164
Date: 2017-05-24 20:18:49 +0000
From: @mdkinney

Additional commit

https://github.com/tianocore/edk2/commit/3e1d93c32efc7d716baeaa0d5d0ef96528141060

tianocore-issues:
## Comment 2165
Date: 2017-05-24 20:19:05 +0000
From: @mdkinney

[...]

PR #10759, opened 2025-02-12
ArmTransferListLib added utility functions and support for FVP

PrachotanReddy:
# Description

This patch series primarily deals with support for the [Firmware Handoff ](https://firmwarehandoff.github.io/firmware_handoff/main/index.html)specification.
This includes:
1. Added utilities to ArmTransferListLib to check validity of a TL and dump its contents.
2. Define and add PcdTransferListBaseAddress to store the base address of the TL coming in on register x3.
3. Define and add a TlHobGuid to hold the base address.
4. Introduce TL to the FVP platform and extract the DT from the TL to demonstrate functionality.
5. Fixes to ShellPkg to display the installed DT address via the dmem utility.

## How This Was Tested

Tested in Base FVP RevC platform by checking Boot, dumping the TL contents and verifying the DT address using dmem.

## Integration Instructions

N/A


ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -19,6 +19,89 @@
...
+  This function verifies the checksum of the Transfer List.
+
+  @param [in]   Tlh       Pointer to the Transfer List Header
+
+  @retval FALSE      Invalid Checksum
+  @retval TRUE       Valid Checksum
+
+**/
+BOOLEAN
+TlVerifyChecksum (
```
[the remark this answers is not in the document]
PrachotanReddy: Done.

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -19,6 +19,89 @@
...
+  if (Tlh == NULL) {
+    return TL_OPS_INVALID;
+  }
+
+  if (Tlh->Signature != TRANSFER_LIST_SIGNATURE_64) {
+    DEBUG ((DEBUG_ERROR, "Bad transfer list signature 0x%x\n", Tlh->Signature));
+    return TL_OPS_INVALID;
+  }
+
+  if (Tlh->TotalSize <= 0) {
```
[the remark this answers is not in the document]
PrachotanReddy: Changed to check if TotalSize is non-zero.

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
```diff
@@ -11,13 +11,72 @@
...
+    } else {
+      DeviceTreeBase = TlGetEntryData (Te);
+      if ((DeviceTreeBase == NULL) || (fdt_check_header (DeviceTreeBase) != 0)) {
+        DEBUG ((DEBUG_ERROR, "%a: Failed to get Device Tree data\n", __func__));
+        return EFI_NOT_FOUND;
+      }
+
+      FdtSize  = fdt_totalsize ((VOID *)DeviceTreeBase) + 0x6f;
+      FdtPages = EFI_SIZE_TO_PAGES (FdtSize);
+      NewBase  = AllocatePages (FdtPages);
```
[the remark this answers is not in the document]
PrachotanReddy: Yes, this can be configured at build-time.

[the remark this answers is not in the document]
pierregondois: Ok yes indeed

ArmPlatformPkg/ArmFvpPlatformDxe/ArmFvpPlatformDxe.c
```diff
@@ -0,0 +1,68 @@
+/** @file
```
leiflindholm, line 1: This code belongs in edk2-platforms, not edk2.

ArmPlatformPkg/ArmFvpPlatformDxe/ArmFvpPlatformDxe.c
```diff
@@ -0,0 +1,68 @@
...
+
+#include <Library/BaseLib.h>
+#include <Library/DebugLib.h>
+#include <Library/HobLib.h>
+#include <Library/UefiBootServicesTableLib.h>
+#include <Protocol/FdtClient.h>
+#include <Library/ArmTransferListLib.h>
+#include <Guid/TlHob.h>
+#include <Guid/FdtHob.h>
+#include <libfdt.h>
```
leiflindholm, line 20: And even if in edk2-platforms, https://github.com/tianocore/edk2/tree/master/MdePkg/Library/BaseFdtLib wrapper should be used. The broken EmbeddedPkg version is way overdue deletion.

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
```diff
@@ -1,6 +1,6 @@
 /** @file
 
-  Copyright (c) 2011-2014, ARM Limited. All rights reserved.
+  Copyright (c) 2011-2024, ARM Limited. All rights reserved.
```
leiflindholm, line 3: This module is used by a whole lot of platforms. And this change breaks all of them.
Meanwhile, the change is not currently relevant for almost any of them.

While that may change in the future, this needs to be introduced in a way that lets platforms opt into it.

ShellPkg/Library/UefiShellDebug1CommandsLib/Dmem.c
```diff
@@ -491,6 +491,11 @@ ShellCommandRunDmem (
               continue;
```
leiflindholm, line 491: commit message: I don't think the "Fixes:" tag is appropriate here.

ArmPkg/Include/IndustryStandard/ArmTransferList.h
```diff
@@ -73,6 +73,16 @@
  */
 #define TRANSFER_LIST_FL_HAS_CHECKSUM  BIT0
 
+/*
+ * Operation codes indicating the validity of the Transfer List.
+ */
+typedef enum {
+  TL_OPS_INVALID,  /* invalid for any operation */
```
leiflindholm, line 80: A lot of the naming in this set clashes with both edk2 coding style and with existing code adjacent.

TL_ -> TRANSFER_LIST_

ArmPkg/Include/Library/ArmTransferListLib.h
```diff
@@ -106,4 +106,58 @@ TlGetEntryData (
...
 
+/**
+  Dump the transfer list to the debug output.
+
+  @param [in]   Tlh       TransferListHeader
+
+**/
+VOID
+EFIAPI
+TlDump (
```
leiflindholm, line 117: TransferListDump (and so on for other function names).

ArmPkg/Include/Library/ArmTransferListLib.h
```diff
@@ -106,4 +106,58 @@ TlGetEntryData (
   IN TRANSFER_ENTRY_HEADER  *Te
   );
 
+/**
+  Dump the transfer list to the debug output.
+
+  @param [in]   Tlh       TransferListHeader
```
leiflindholm, line 112: TransferListHeader, and the description "Transfer list header" (and so on for other arguments)

ArmPkg/Include/Library/ArmTransferListLib.h
```diff
@@ -106,4 +106,58 @@ TlGetEntryData (
...
+TRANSFER_LIST_OPS
+EFIAPI
+TlCheckHeader (
+  IN TRANSFER_LIST_HEADER  *Tlh
+  );
+
+/**
+  Find a Transfer Entry Node in the Transfer List matched with the given tag-id.
+
+  @param [in]   Tlh       Pointer to the Transfer List Header
```
leiflindholm, line 151: This one gets the description right, "Pointer to the Transfer List Header". Use that description throughout.

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -168,3 +252,73 @@ TlGetEntryData (
...
+  @return Pointer to the Transfer Entry Node if successful otherwise NULL
+**/
+TRANSFER_ENTRY_HEADER *
+EFIAPI
+TlFindEntry (
+  IN TRANSFER_LIST_HEADER  *Tlh,
+  IN UINT16                TagId
+  )
+{
+  TRANSFER_ENTRY_HEADER  *Te = NULL;
```
leiflindholm, line 271: OK, so I see this all slipped through previous code review already, but it's quite broken coding-style wise.
I'm going to create a new PR cleaning this up, and then I'd appreciate if you could rework this to comply with that.

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
PrachotanReddy: Thanks for the comments. Is there a specific reason you think this change breaks other platforms? This was written in a way to be backward compatible even on platforms that didn't support transfer list.

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
```diff
@@ -11,13 +11,71 @@
...
+  VOID                   *TransferListBase;
+  VOID                   *DeviceTreeBase;
+  TRANSFER_ENTRY_HEADER  *Te;
+  UINT64                 *TlHobData;
+  UINTN                  FdtSize;
+  UINTN                  FdtPages;
+  UINT64                 *FdtHobData;
+  VOID                   *NewBase;
+
+  TransferListBase = (VOID *)(UINTN)PcdGet64 (PcdTransferListBaseAddress);
```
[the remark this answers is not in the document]
samimujawar: @leiflindholm please let us know if the suggested approach would work, or you think it would be better to introduce a new PlatformPeiLib that handles the Transfer List.

samimujawar (changes requested):

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -165,3 +249,73 @@ TransferListGetEntryData (
...
+  @return Pointer to the Transfer Entry Node if successful otherwise NULL
+**/
+TRANSFER_ENTRY_HEADER *
+EFIAPI
+TransferListFindEntry (
+  IN TRANSFER_LIST_HEADER  *TransferListHeader,
+  IN UINT16                TagId
+  )
+{
+  TRANSFER_ENTRY_HEADER  *Te = NULL;
```
leiflindholm, line 268: In the existing code I reworked, I renamed this local variable "Entry" in all functions. Can you please do the same in order to maintain alignment across the file?

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
```diff
@@ -8,10 +8,22 @@
...
+  b.ne  _SetSVCMode          // Branch if x1 is not equal to 0x14a0fb10b (TL signature)
+
+  // Branch if x2 is not equal to 0
+  cbnz  x2, _SetSVCMode
+
+  // Set the TransferList Base Address from register x3
+  adr   x4, PcdGet64(PcdTransferListBaseAddress)
+  str   x3, [x4]
+
+_SetSVCMode:
```
leiflindholm, line 23: This is an ominous-looking change at first. But also kind of nonsensical (AArch64 doesn't have an SVC mode).

@ardbiesheuvel this is clearly a hangaround from the original of this file being based on the AArch32 version. But is that even a valid thing to do _there_ anymore, or is this a hangaround of "change out of Secure Monitor Mode"?

I feel like we should have a separate commit deleting this label, at least on AArch64, and figure out a new label name here.

ArmPkg/ArmPkg.dec
```diff
@@ -104,6 +104,7 @@
   gArmMpCoreInfoGuid = { 0xa4ee0728, 0xe5d7, 0x4ac5,  {0xb2, 0x1e, 0x65, 0x8e, 0xd8, 0x57, 0xe8, 0x34} }
 
   gArmMmuReplaceLiveTranslationEntryFuncGuid = { 0xa8b50ff3, 0x08ec, 0x4dd3, {0xbf, 0x04, 0x28, 0xbf, 0x71, 0x75, 0xc7, 0x4a} }
+  gTlHobGuid = { 0xebe7bae8, 0xfe18, 0x43c5, { 0xbf, 0x3f, 0xf2, 0xb1, 0xaf, 0xb2, 0xdf, 0xb8 } }
```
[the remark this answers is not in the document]
leiflindholm: (also in commit message)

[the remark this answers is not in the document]
leiflindholm: And really, the filename too.

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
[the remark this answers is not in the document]
leiflindholm: I don't really want to fork the boot process based on that, but I also don't know if it's reasonable to assume the transfer list cannot reside at physical address zero. Does the spec say that?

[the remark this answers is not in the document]
PrachotanReddy: Hi Leif
The spec doesn't mandate this. However, the TF-A implementation that sets the args does a NULL check too, see:
https://github.com/ARM-software/arm-trusted-firmware/blob/fa8ca8bcd0cb1d75747fa5baa14e2b496c30ec82/lib/transfer_list/transfer_list.c#L67
Let me know what you think.

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
PrachotanReddy: Done.

ArmPkg/Include/IndustryStandard/ArmTransferList.h
```diff
@@ -2,7 +2,7 @@
   Header file defining a Transfer List and Transfer Entry as specified by the
   A-profile Firmware Handoff Protocol specification.
 
-  Copyright (c) 2024, Arm Limited. All rights reserved.<BR>
+  Copyright (c) 2025, Arm Limited. All rights reserved.<BR>
```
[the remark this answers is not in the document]
PrachotanReddy: Done.

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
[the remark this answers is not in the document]
leiflindholm: If the spec doesn't mandate it, we can't rely on it. That's kind of the point of the spec. The use of TF-A is not mandatory, and alternative secure-world firmwares exist.

We'll need to store the fact that we were passed a transfer list (in ModuleEntryPoint.S) somehow and check that here instead.

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
leiflindholm: Now raised as #10880.

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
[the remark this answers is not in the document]
samimujawar: The Firmware Handoff specification has been updated with the following wording: "R1: The **tl_base_pa address must be non-zero** and at an 8-byte aligned boundary.", see https://firmwarehandoff.github.io/firmware_handoff/main/transfer_list.html#transfer-list-requirements

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
```diff
@@ -8,6 +8,24 @@
...
+
+  // Branch if x2 is not equal to 0
+  cbnz  x2, _SkipTransferList
+
+  // Set the TransferList Base Address from register x3
+  adr   x4, PcdGet64(PcdTransferListBaseAddress)
+  str   x3, [x4]
+
+  // Set PcdTransferListSignatureValid to TRUE
+  adr   x4, PcdGetBool(PcdTransferListSignatureValid)
```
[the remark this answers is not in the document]
samimujawar: The commit message mentions "Also set PcdTransferListSignatureValid to TRUE to help validate
in future use cases."

As mentioned above, I think we can avoid PcdTransferListSignatureValid. 

Also the commit message needs to be updated so that TL acronym is not used.

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
```diff
@@ -11,13 +11,33 @@
...
 EFIAPI
 PlatformPeim (
   VOID
   )
 {
+  VOID    *TransferListBase;
+  UINT64  *TransferListHobData;
+
+  TransferListBase = (VOID *)(UINTN)PcdGet64 (PcdTransferListBaseAddress);
+  if ((PcdGetBool (PcdTransferListSignatureValid)) && (TransferListCheckHeader (TransferListBase) != TRANSFER_LIST_OPS_INVALID)) {
```
samimujawar, line 28: if ((TransferListBase =! NULL)) && (TransferListCheckHeader (TransferListBase) != TRANSFER_LIST_OPS_INVALID)) {

Or this should really be split into 2 conditions.

```
 if (TransferListBase =! NULL) {
     if ( TransferListCheckHeader (TransferListBase) != TRANSFER_LIST_OPS_INVALID) {
     ...
     } else {
       // The error message at line 38 should go here.
     }
} else {
   // optional else case that prints a Debug info message indicating a transfer list was not found.
}
```

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
```diff
@@ -11,13 +11,33 @@
...
+    TransferListDump (TransferListBase);
```
samimujawar, line 29: DEBUG_CODE_BEGIN ();
TransferListDump (TransferListBase);
DEBUG_CODE_END ();

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
```diff
@@ -11,13 +11,33 @@
...
+    TransferListHobData = BuildGuidHob (&gTransferListHobGuid, sizeof (*TransferListHobData));
+    if (TransferListHobData == NULL) {
+      ASSERT (0);
```
samimujawar, line 32: You could add a DEBUG_ERROR message here.

ArmPkg/Include/IndustryStandard/ArmTransferList.h
```diff
@@ -73,6 +73,16 @@
  */
 #define TRANSFER_LIST_FL_HAS_CHECKSUM  BIT0
 
+/*
```
samimujawar, line 76: Can the commit message be updated to use Transferlist instead of the acronym TL, please?

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
```diff
@@ -8,6 +8,24 @@
 #include <AsmMacroLib.h>
```
samimujawar, line 8: Please add a references section with link to the Firmware handoff specification in the file header
It may also be helpful to add this to other files that implement the Fw handoff spec details.

//
//  @par Reference(s):
//   - Firmware Handoff specification [https://firmwarehandoff.github.io/firmware_handoff/main]
//

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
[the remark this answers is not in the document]
PrachotanReddy: Thanks for the suggestion Sami.
I agree with you, this was introduced when the spec was not updated to reflect the base address is non-zero.

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
```diff
@@ -1,13 +1,34 @@
...
+  // Branch if x1 is not equal to 0x14a0fb10b (TransferList signature)
+  b.ne  _SkipTransferList
+
+  // Branch if x2 is not equal to 0
+  cbnz  x2, _SkipTransferList
+
+  // Set the TransferList Base Address from register x3
+  adr   x4, PcdGet64(PcdTransferListBaseAddress)
+  str   x3, [x4]
+
```
samimujawar, line 25: I think we need another branch here to line 31

samimujawar: or alternatively can you check if the code below can be used?

`
ASM_FUNC(_ModuleEntryPoint)
 // Set PcdTransferListBaseAddress explicitly to 0
  adr   x5, PcdGet64(PcdTransferListBaseAddress)
  mov   x4, xzr
  str   x4, [x5]

  // Check if register x1 matches the firmware handoff spec
  MOV64 (x4, 0x14a0fb10b)
  cmp   x1, x4
  // Branch if x1 is not equal to 0x14a0fb10b (TransferList signature)
  b.ne  _SkipTransferList

  // Branch if x2 is not equal to 0
  cbnz  x2, _SkipTransferList

  // Set the TransferList Base Address from register x3
  str   x3, [x5]

_SkipTransferList:

// Do early platform specific actions
  bl    ASM_PFX(ArmPlatformPeiBootAction)	  bl    ASM_PFX(ArmPlatformPeiBootAction)
`

ArmPkg/Include/Guid/TransferListHob.h
```diff
@@ -0,0 +1,23 @@
...
+
+  @par Reference(s):
+    - https://github.com/FirmwareHandoff/firmware_handoff
+
+**/
+
+#ifndef TRANSFER_LIST_HOB_H__
+#define TRANSFER_LIST_HOB_H__
+
+#define TL_HOB_GUID  {\
```
samimujawar, line 16: TL_HOB_GUID => TRANSFER_LIST_HOB_GUID

Please also check if there are any instances of the acronym TL and replace it appropriately in this series.

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
```diff
@@ -1,23 +1,54 @@
...
+      TransferListHobData = BuildGuidHob (&gTransferListHobGuid, sizeof (*TransferListHobData));
+      if (TransferListHobData == NULL) {
+        DEBUG ((DEBUG_ERROR, "%a: TransferList BuildGuidHob failed\n", __func__));
+        ASSERT (0);
+        return EFI_OUT_OF_RESOURCES;
+      }
+
+      *TransferListHobData = (UINTN)TransferListBase;
+    } else {
+      DEBUG ((DEBUG_ERROR, "%a: No Transfer List or Transfer List signature found @ 0x%p\n", __func__, TransferListBase));
```
samimujawar, line 46: I think this error message should be as below:
DEBUG ((DEBUG_ERROR, "%a: Invalid Transfer Transfer List signature found @ 0x%p\n", __func__, TransferListBase));

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
```diff
@@ -1,23 +1,54 @@
...
+    }
+  } else {
+    DEBUG ((DEBUG_INFO, "%a: No valid TransferList signature found, continuing boot\n", __func__));
```
samimujawar, line 49: I think this debug info message should be 
DEBUG ((DEBUG_INFO, "%a: No TransferList found, continuing boot\n", __func__));

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
PrachotanReddy: I think this error is thrown when a valid signature is present, i.e. TransferListBase is not NULL, but the transfer list is corrupted or invalid.
Changing the debug message to instead say DEBUG ((DEBUG_ERROR, "%a: No valid operations possible on TransferList found @ 0x%p\n", __func__, TransferListBase));

samimujawar (approved):

LeviYeoReum (approved):

manish-pandey-arm:
FWIW Transfer List standalone library is now available https://review.trustedfirmware.org/admin/repos/shared/transfer-list-library,general

PrachotanReddy:
@leiflindholm Can you please review the PR when you get a chance?

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
```diff
@@ -1,13 +1,34 @@
...
 //  SPDX-License-Identifier: BSD-2-Clause-Patent
 //
+// @Par Reference(s):
+// - Firmware Handoff specification [https://firmwarehandoff.github.io/firmware_handoff/main]
 //
 
 #include <AsmMacroLib.h>
 
 ASM_FUNC(_ModuleEntryPoint)
+  // Set PcdTransferListBaseAddress explicitly to 0
```
LeviYeoReum, line 13: Hmm I think it would be better to handle this logic in each ArmPlatformLib's `ArmPlatformPeiBootAction()`
so that it works not only PeilessSec but also Sec & Pei phase

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
```diff
@@ -1,23 +1,54 @@
...
 EFI_STATUS
 EFIAPI
 PlatformPeim (
   VOID
   )
 {
+  VOID    *TransferListBase;
+  UINT64  *TransferListHobData;
+
+  TransferListBase = (VOID *)(UINTN)PcdGet64 (PcdTransferListBaseAddress);
```
LeviYeoReum, line 30: I think it wouldn't be work in platform where Sec & PEI phase have.
In that case, It should create "TransferList Hob" in SecMain because `PcdTransferListBaseAddress` wouldn't be shared between Sec and PeiMain.
But for PeilessSec platform, it should generate the TransferListHob in here.
Should we check the existence of TransferListHob before creating it?

ArmPkg/ArmPkg.dec
```diff
@@ -92,6 +92,7 @@
   gArmMpCoreInfoGuid = { 0xa4ee0728, 0xe5d7, 0x4ac5,  {0xb2, 0x1e, 0x65, 0x8e, 0xd8, 0x57, 0xe8, 0x34} }
 
   gArmMmuReplaceLiveTranslationEntryFuncGuid = { 0xa8b50ff3, 0x08ec, 0x4dd3, {0xbf, 0x04, 0x28, 0xbf, 0x71, 0x75, 0xc7, 0x4a} }
+  gTransferListHobGuid = { 0xebe7bae8, 0xfe18, 0x43c5, { 0xbf, 0x3f, 0xf2, 0xb1, 0xaf, 0xb2, 0xdf, 0xb8 } }
```
LeviYeoReum, line 95: What about rename to `gArmTransferListHobGuid`?

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
samimujawar: The purpose of ArmPlatformPeiBootAction() is to allow a platform to perform early initialisation and would be very platform specific. e.g. see the implementation at 
https://github.com/tianocore/edk2/blob/master/ArmVirtPkg/Library/ArmPlatformLibQemu/AArch64/ArmPlatformHelper.S#L47
or
https://github.com/tianocore/edk2-platforms/blob/master/Platform/RaspberryPi/Library/PlatformLib/AArch64/RaspberryPiHelper.S#L52

While Transfer List is a standard mechanism to handoff from one firmware star to another. So this is common/standard code across platforms. It could be argued that the early transfer list detection code can be moved to a common assembler function that can be used in both the Sec and Pei phase. 

However, I do not think we should move it to ArmPlatformPeiBootAction().

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
PrachotanReddy: Agreeing with Sami's previous comment.
I think extending transfer list features to platforms with SEC+PEI phase can be a standalone patch series. This PR is a non-breaking change, implying no adverse affects on SEC+PEI platforms.

ArmPkg/ArmPkg.dec
PrachotanReddy: Done.

ArmPlatformPkg/ArmPlatformPkg.dec
```diff
@@ -119,3 +119,9 @@
   gArmPlatformTokenSpaceGuid.PcdPL031RtcPpmAccuracy|300000000|UINT32|0x00000022
 
   gArmPlatformTokenSpaceGuid.PcdWatchdogCount|0x0|UINT32|0x00000033
+
+[PcdsPatchableInModule]
```
leiflindholm, line 123: From https://github.com/tianocore-docs/edk2-DecSpecification/blob/main/2_dec_file_overview/210_pcd_usage.md

`Generally, PCDs using the FixedAtBuild and PatchableInModule access methods are used to set static configuration elements that can be determined at build time (or modified prior to inserting a module into a flash image). PCDs using the Dynamic and DynamicEx access methods are used for configuration knobs that may be manipulated by setup screens (or other methods) during the boot process.`

Based on the other description, I would have expected PcdsDynamic?

ArmPlatformPkg/ArmPlatformPkg.dec
```diff
@@ -119,3 +119,9 @@
   gArmPlatformTokenSpaceGuid.PcdPL031RtcPpmAccuracy|300000000|UINT32|0x00000022
 
   gArmPlatformTokenSpaceGuid.PcdWatchdogCount|0x0|UINT32|0x00000033
+
+[PcdsPatchableInModule]
+  # This Pcd is populated at run-time and is 0 by default.
```
leiflindholm, line 124: The definition below explicitly states that the default value is 0. And populated at runtime ... the Pcd type used should be all the information required.

What *might* be useful would be to call out "A value of 0 indicates no transfer list present."

ArmPlatformPkg/ArmPlatformPkg.dec
```diff
@@ -119,3 +119,9 @@
   gArmPlatformTokenSpaceGuid.PcdPL031RtcPpmAccuracy|300000000|UINT32|0x00000022
 
   gArmPlatformTokenSpaceGuid.PcdWatchdogCount|0x0|UINT32|0x00000033
+
+[PcdsPatchableInModule]
+  # This Pcd is populated at run-time and is 0 by default.
+  # It is typically updated based on the presence of the TransferList signature.
```
leiflindholm, line 125: We don't care about typically, we care about what is. If it's a dynamic Pcd, it can be updated.

ArmPlatformPkg/ArmPlatformPkg.dec
```diff
@@ -119,3 +119,9 @@
   gArmPlatformTokenSpaceGuid.PcdPL031RtcPpmAccuracy|300000000|UINT32|0x00000022
 
   gArmPlatformTokenSpaceGuid.PcdWatchdogCount|0x0|UINT32|0x00000033
+
+[PcdsPatchableInModule]
+  # This Pcd is populated at run-time and is 0 by default.
+  # It is typically updated based on the presence of the TransferList signature.
+  # Platforms must not initialise/override this value in the platform DSC file.
```
leiflindholm, line 126: We can't randomly tell platforms to avoid initializing certain Pcds.
Presumably the intent is to convey "if you don't have a transfer list, don't set this to non-zero"? If so, a statement as suggested above of "A value of 0 indicates no transfer list is present.

So to summarize my series of comments:
All you need to say here is that this value indicates the start address of a transfer list, and 0 is treated as no transfer list present.

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
```diff
@@ -1,13 +1,34 @@
...
+  adr   x5, PcdGet64(PcdTransferListBaseAddress)
+  mov   x4, xzr
+  str   x4, [x5]
+
+  // Check if register x1 matches the firmware handoff spec
```
leiflindholm, line 18: Matches the spec in what way? Say what it's testing for, not where what it's testing for is defined.

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
leiflindholm, line 13: Why? The default is 0. It will be set to that unless someone explicitly sets it to something else for their platform. In which case, why are we trying to ignore the platform owner's wishes?

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
```diff
@@ -1,13 +1,34 @@
...
+  MOV64 (x4, 0x14a0fb10b)
+  cmp   x1, x4
+  // Branch if x1 is not equal to 0x14a0fb10b (TransferList signature)
```
leiflindholm, line 21: We don't need to repeat the number. Just say not equal to the TransferList signature.

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
leiflindholm: Actually, we could reword this to include the other registers being checked (if we care beyond x1) - // Check if register assignment at handoff matches spec

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
leiflindholm: Also, "Branch". Branch where? Better to be specific: "skip transferlist init if..."

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
```diff
@@ -1,13 +1,34 @@
...
+  b.ne  _SkipTransferList
+
+  // Branch if x2 is not equal to 0
```
leiflindholm, line 24: Why?
We can communicate the reasoning more clearly by saying:
// "Skip TransferList init if...", possibly with an appended "... as mandated by spec"

leiflindholm:
OK, sorry, I realise that was just a barrage of comments.
Basically all of it is just about comments and clarifying intent.

But we need to figure out the Pcd type and handling, because the current design doesn't work to expected patterns for tianocore, and I think only works by accident.

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
```diff
@@ -1,13 +1,34 @@
...
+  cbnz  x2, _SkipTransferList
+
+  // Set the TransferList Base Address from register x3
+  str   x3, [x5]
```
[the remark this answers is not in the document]
PrachotanReddy: Hi Leif
Thanks for the feedback. This design/implementation approach is based off of how similar situations are being handled in other modules.
see: 1. https://github.com/tianocore/edk2/blob/2e5b357a1f80e7b05560f4cf9b4edbfca9de5ab4/ArmVirtPkg/ArmVirtKvmTool.dsc#L176
2. https://github.com/tianocore/edk2/blob/2e5b357a1f80e7b05560f4cf9b4edbfca9de5ab4/ArmVirtPkg/PrePi/AArch64/ModuleEntryPoint.S#L18

Additionally, per https://github.com/tianocore-docs/edk2-DscSpecification/blob/main/2_dsc_overview/28_pcd_sections.md#2822-pcdspatchableinmodule
Pcdspatchableinmodule can be patched by tools that know the offset of the PCD into the binary file. Isn't that what we are doing in this assembly?

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
leiflindholm: That seems excessive (and strange) to me. If a platform sets a base address, that is because it wants to force use of that address. If we ignore and overwrite that, we are breaking the platform. It's not something that leads to an improved situation.

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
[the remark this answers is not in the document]
leiflindholm: > Pcdspatchableinmodule can be patched by tools that know the offset of the PCD into the binary file. Isn't that what we are doing in this assembly?

Tools. Outside of the running system, before the image is flashed.
It just happens that your image is loaded into memory. If it was XIP in flash it would break.

The vm images can take certain shortcuts because they *know* the platform they are running on. Platform-independent drivers cannot.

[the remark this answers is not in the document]
PrachotanReddy: I tried grepping for usage of PcdSet and of Pcds of type Pcds.Dynamic in *.S but couldn't find anything. Is there an example platform or usage I can look at?
Also going back to ArmVirtPkg, ArmVirtQemu and other modules where I found the usage of PcdsPatchableInModule in a similar fashion also have certain other Pcds that are declared PcdsDynamicDefault, is there a particular reasoning to this?

Also, on trying to declare PcdTransferListBaseAddress as PcdsDynamicDefault, I'm unable to use the PcdGet64 or PcdGetPtr macros:
/AArch64/ModuleEntryPoint.iiii:14: Error: unexpected characters following instruction at operand 2 -- `adr x5,LibPcdGet64(1)


Further investigating, the Pcd wiki says this about:
PCDs if type PatchableInModule are declared as ‘volatile’ variables in every the module that uses that PCD.
But no mention of their inability to be modified at run-time.
https://github.com/tianocore/tianocore.github.io/wiki/PCD

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
samimujawar: Hi Leif,

I had suggested this on the assumption that we get a transfer list only on a successful handoff. 

But I see your point, a platform can place a transfer list at a platform specific address and utilise it (in which case it probably does not need to follow the [architecture bindings for the TL exchange at the handoff boundary](https://firmwarehandoff.github.io/firmware_handoff/main/register_conventions.html#register-usage-at-handoff-boundary)). 

This would mean that as long as the TransferList header validation checks out, then the platform can utilise the transfer list.

Kindly let me know if this is the use case you are considering.

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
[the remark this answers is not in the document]
ardbiesheuvel: The PeilessSec version of ArmVirtQemu (ab)uses patchable PCDs for a limited set of variables, as the SEC module is guaranteed to run from writable DRAM, and the PCD value is used only inside the module.

Patching PCDs at runtime does not propagate to other modules, it only updates the copy visible to the current module. And given that SEC and PEI modules often run XIP from NOR flash, generic code should never assume that these patchable PCDs are still patchable at runtime.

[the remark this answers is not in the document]
PrachotanReddy: Thanks @ardbiesheuvel and @leiflindholm .
In the current implementation the Pcd is only used within the context of the current module, but however like mentioned, we cannot assume them to be patchable.
I did an earlier PoC where I used global variables instead before switching to Pcds, do you think that'll be a better approach?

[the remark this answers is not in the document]
leiflindholm: That sounds good to me.

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
jmarinho: Hi @leiflindholm, indeed the TL is similar in design and intent to the HOB.
Are you implying we should handle it in a similar way to a HOB? Or integrate the TL in the HOB Library?

PrachotanReddy: Hi Leif
If the problem is sharing the values with later PI stages, we don't really need that anyway. Even in the current PCD implementation too, the value is only used within the current module, the PCD is not shared with any other PI stage, instead a GUID HOB is built after validation.
I'll go ahead and send in patches with the global variable approach.

ArmPlatformPkg/PeilessSec/PeilessSec.c
```diff
@@ -13,6 +13,7 @@
 
 UINT64  mSystemMemoryEnd = FixedPcdGet64 (PcdSystemMemoryBase) +
                            FixedPcdGet64 (PcdSystemMemorySize) - 1;
+UINT64  TransferListBaseAddr = 0;
```
leiflindholm, line 16: Ah, global symbols should have m (module-global) or g(fully-global) prefix. In this case, 'm' as the above variable.

PrachotanReddy: Thanks for catching that @leiflindholm, fixed it.

PrachotanReddy:
Hi @leiflindholm @ardbiesheuvel @samimujawar 
The CI check keeps failing [Ubuntu GCC - PR (Build_GCC5 TARGET_ARM_ARMPLATFORM)](https://github.com/tianocore/edk2/pull/10759/checks?check_run_id=43065975277).
The Windows VS env and my local build show no issues. I tried adding volatile expecting compiler efficiency flags but that didn't help either.

leiflindholm:
> Hi @leiflindholm @ardbiesheuvel @samimujawar The CI check keeps failing [Ubuntu GCC - PR (Build_GCC5 TARGET_ARM_ARMPLATFORM)](https://github.com/tianocore/edk2/pull/10759/checks?check_run_id=43065975277). The Windows VS env and my local build show no issues. I tried adding volatile expecting compiler efficiency flags but that didn't help either.

```
PROGRESS - Running Build DEBUG
ERROR - Compiler #7000 from : Failed to execute command
ERROR - EDK2 #002 from : Failed to build module
CRITICAL - Build failed
```
@makubacki @mdkinney any ideas?

makubacki:
> > Hi @leiflindholm @ardbiesheuvel @samimujawar The CI check keeps failing [Ubuntu GCC - PR (Build_GCC5 TARGET_ARM_ARMPLATFORM)](https://github.com/tianocore/edk2/pull/10759/checks?check_run_id=43065975277). The Windows VS env and my local build show no issues. I tried adding volatile expecting compiler efficiency flags but that didn't help either.
> 
> ```
> PROGRESS - Running Build DEBUG
> ERROR - Compiler #7000 from : Failed to execute command
> ERROR - EDK2 #002 from : Failed to build module
> CRITICAL - Build failed
> ```
> 
> @makubacki @mdkinney any ideas?

This is the issue from the full build log:

```
INFO - /usr/bin/aarch64-linux-gnu-ld: /tmp/ccwoSIur.ltrans0.ltrans.o: in function `_ModuleEntryPoint':
INFO - /__w/1/s/ArmPlatformPkg/PlatformPei/PlatformPeiLib.c:31:(.text._ModuleEntryPoint+0x28): undefined reference to `mTransferListBaseAddr'
INFO - /usr/bin/aarch64-linux-gnu-ld: /tmp/ccwoSIur.ltrans0.ltrans.o:/__w/1/s/ArmPlatformPkg/PlatformPei/PlatformPeiLib.c:31:(.text._ModuleEntryPoint+0x2c): undefined reference to `mTransferListBaseAddr'
INFO - collect2: error: ld returned 1 exit status
INFO - make: *** [GNUmakefile:351: /__w/1/s/Build/ArmPlatform/RELEASE_GCC5/AARCH64/ArmPlatformPkg/PlatformPei/PlatformPeim/DEBUG/PlatformPei.dll] Error 1
```

PrachotanReddy:
Hi @leiflindholm
Fixed all the issues, please take a look and let me know what you think.

leiflindholm (approved):

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
```diff
@@ -1,23 +1,43 @@
...
 
 #include <PiPei.h>
 
 #include <Library/ArmPlatformLib.h>
 #include <Library/HobLib.h>
 #include <Library/PcdLib.h>
+#include <Library/ArmTransferListLib.h>
+#include <Library/DebugLib.h>
+
+UINT64  mTransferListBaseAddr = 0;
```
ardbiesheuvel: Writable global variables are not permitted in PEIMs: when `ArmPlatformPkg/PlatformPei/PlatformPeim.inf` incorporates this library and executes in place from NOR flash, setting this variable will either have no effect or crash the system.

ardbiesheuvel (changes requested):

ardbiesheuvel:
Apologies for chiming in late:

`PlatformPeiLib` is not the right place for this - this is a library that platforms are expected to replace. The right way to do this is to update both `Sec` and `PeilessSec` and update `CEntrypoint()` to take a TL pointer. Once running C code, the TL pointer should be stored into a HOB as usual.

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
PrachotanReddy: Hi @ardbiesheuvel 
Thanks for catching this.
Should we build the hob in SecMain?
Are we allowed to build a GuidHob in PeilessSec.c or Sec.c?

PrachotanReddy:
Hi @makubacki 
How do I check the detailed build logs. The Ubuntu GCC build is now failing again, but it's all good locally.

PrachotanReddy:
Hi @ardbiesheuvel 
Thanks for catching this. Let me know if this fix looks good.

ardbiesheuvel:
This looks ok now but please implement the same for the normal SEC module.

PrachotanReddy:
Hi @ardbiesheuvel 
I'm trying to implement this for SEC too, but since "HobTable" is created by PEI core, we might not be able to build the hob in the sec module. Please let me know if there's a "right" place to build the Guid Hob. An alternate approach, is to introduce a PpiGuid, which holds the address for Pei to build the hob.
Also, the upstream edk2-platforms for ArmVExpress-FVP-AArch64 have references to packages like Nt32 which might need cleanup, hence we might not have a platform to readily test these changes.
Thus this enablement might need patches to additional modules, do you think support for SEC platforms can be a separate patch-series/PR?

PrachotanReddy:
Hi @ardbiesheuvel 
Checking in to see if you had a chance to review.

ardbiesheuvel:
Yes, it appears the most appropriate way to achieve this is to provide a PPI in Sec.c that records the TL base address passed to CEntrypoint. Then you'll need a separate PEIM that DEPEXes on the PPI guid, and constructs the HOB. Alternatively, you might just handle this in the generic part of PlatformPei - the PPI should either be there from the start of PEI or it will never appear, and so there is no ordering concern between the PEIMs

PrachotanReddy:
> Yes, it appears the most appropriate way to achieve this is to provide a PPI in Sec.c that records the TL base address passed to CEntrypoint. Then you'll need a separate PEIM that DEPEXes on the PPI guid, and constructs the HOB. Alternatively, you might just handle this in the generic part of PlatformPei - the PPI should either be there from the start of PEI or it will never appear, and so there is no ordering concern between the PEIMs

Thanks @ardbiesheuvel , to my other question, do you think this can be another PR, especially since the implementation is significantly different?

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
```diff
@@ -1,13 +1,29 @@
...
+// @Par Reference(s):
+// - Firmware Handoff specification [https://firmwarehandoff.github.io/firmware_handoff/main]
 //
 
 #include <AsmMacroLib.h>
 
 ASM_FUNC(_ModuleEntryPoint)
+
+  // Check if register assignment at handoff matches spec
+  MOV64 (x4, 0x14a0fb10b)
```
jmarinho, line 15: nit: should 0x14a0fb10b be a #defined or a PCD?

leiflindholm: Not a Pcd, because the value is not platform dependent.

Since it's only used once, the value a #define would bring is semantic, and could also be covered by the comment. _However_, the comment should in that case be more explicit about what it's referring to.

To me, the following two are equivalent:
```
// Check if x1 holds TransferList signature
MOV64 (x4, 0x14a0fb10b)
```
and
```
MOV64 (x4, TRANSFER_LIST_SIGNATURE)
```

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
```diff
@@ -76,6 +92,8 @@ _GetStackBase:
   MOV32 (x2, FixedPcdGet32(PcdCPUCorePrimaryStackSize))
   sub   x1, x1, x2
 
+_GetTransferListBase:
```
ardbiesheuvel: Please drop this label unless it is used anywhere

ArmPlatformPkg/Sec/Sec.c
```diff
@@ -80,6 +80,14 @@ STATIC CONST EFI_PEI_PPI_DESCRIPTOR  gCommonPpiTable[] = {
   }
 };
 
+// The Transfer List PPI is used to pass the Transfer List to the PEI Core
+// and goes at the end of the PPI list.
+STATIC EFI_PEI_PPI_DESCRIPTOR  mTransferListBase = {
+  (EFI_PEI_PPI_DESCRIPTOR_PPI | EFI_PEI_PPI_DESCRIPTOR_TERMINATE_LIST),
+  &gArmTransferListPpiGuid,
+  NULL // Set later by SecMain
```
ardbiesheuvel: SEC and PEI do not permit writable global variables. You will need to allocate this PPI dynamically if you want to modify it at runtime. You may be able to steal some space from the base of the stack like the code below is already doing.

ardbiesheuvel (changes requested):

ArmPlatformPkg/Sec/Sec.c
```diff
@@ -86,18 +86,27 @@ STATIC CONST EFI_PEI_PPI_DESCRIPTOR  gCommonPpiTable[] = {
...
   )
 {
   EFI_PEI_PPI_DESCRIPTOR  *PlatformPpiList;
   UINTN                   PlatformPpiListSize;
   UINTN                   ListBase;
-  EFI_PEI_PPI_DESCRIPTOR  *LastPpi;
+
+  // The Transfer List PPI is used to pass the Transfer List to the PEI Core
+  // and goes at the end of the PPI list.
+  EFI_PEI_PPI_DESCRIPTOR  mTransferListBase = {
```
ardbiesheuvel: Please drop this definition entirely (note btw that the `m` prefix is not appropriate for stack variable).

Instead, use `LastPpi` - see below

ArmPlatformPkg/Sec/Sec.c
```diff
@@ -107,14 +116,15 @@ CreatePpiList (
   ListBase = PcdGet64 (PcdCPUCoresStackBase);
   CopyMem ((VOID *)ListBase, gCommonPpiTable, sizeof (gCommonPpiTable));
   CopyMem ((VOID *)(ListBase + sizeof (gCommonPpiTable)), PlatformPpiList, PlatformPpiListSize);
-
-  // Set the Terminate flag on the last PPI entry
-  LastPpi = (EFI_PEI_PPI_DESCRIPTOR *)ListBase +
-            ((sizeof (gCommonPpiTable) + PlatformPpiListSize) / sizeof (EFI_PEI_PPI_DESCRIPTOR)) - 1;
```
ardbiesheuvel, line 113: Here, you should be able to insert something like

```c
if (TransferListBase != NULL) {
  LastPpi++;
  LastPpi->Flags = EFI_PEI_PPI_DESCRIPTOR_PPI;
  LastPpi->Guid = &gArmTransferListPpiGuid;
  LastPpi->Ppi = TransferListBase;
}
```
and drop the CopyMem() below

ArmPlatformPkg/Sec/Sec.c
```diff
@@ -107,14 +116,15 @@ CreatePpiList (
...
+  // Copy the Transfer List PPI after the Common and Platform PPIs
+  CopyMem (
+    (VOID *)(ListBase + sizeof (gCommonPpiTable) + PlatformPpiListSize),
+    &mTransferListBase,
+    sizeof (mTransferListBase)
+    );
 
   *PpiList     = (EFI_PEI_PPI_DESCRIPTOR *)ListBase;
-  *PpiListSize = sizeof (gCommonPpiTable) + PlatformPpiListSize;
+  *PpiListSize = sizeof (gCommonPpiTable) + PlatformPpiListSize + sizeof (mTransferListBase);
```
ardbiesheuvel: Take care to only add `sizeof(EFI_PEI_PPI_DESCRIPTOR)` here if `TransferListBase != NULL`

ArmPlatformPkg/PlatformPei/PlatformPeim.inf
```diff
@@ -43,6 +43,7 @@
 [Ppis]
   gEfiPeiMasterBootModePpiGuid                  # PPI ALWAYS_PRODUCED
   gEfiPeiBootInRecoveryModePpiGuid              # PPI SOMETIMES_PRODUCED
+  gArmTransferListPpiGuid                       # PPI SOMETIMES_PRODUCED
```
ardbiesheuvel: This should be `SOMETIMES_CONSUMED`

ardbiesheuvel (approved):

ardbiesheuvel:
This looks good to me now.

@leiflindholm any thoughts?

PR #10760, opened 2025-02-13
ArmPkg/ArmStandaloneMmCoreEntryPoint: Fix clang build issues

makubacki:
# Description

**Issue 1:**

```
error: immediate must be an integer in range [0, 65535].
movz x9, (-1) >> 48, lsl #48
movk x9, ((-1) >> 32) & 0xffff, lsl #32
movk x9, ((-1) >> 16) & 0xffff, lsl #16
movk x9, (-1) & 0xffff
```

From this code in ModuleEntryPoint.S:

```
  MOV64 (x9, ARM_FFA_RET_NOT_SUPPORTED)
```

Resolved with the change in MdePkg/Include/AArch64/AsmMacroLib.h.
(first commit)

---

**Issue 2:**

```
ArmStandaloneMmCoreEntryPoint.c:1041:48: error: cast to smaller
  integer type 'unsigned long' from 'void *'
  [-Werror,-Wvoid-pointer-to-int-cast]

  1041 |   DEBUG ((DEBUG_INFO, "Start Dump Hob: %lx\n",
             (unsigned long)HobStart));
```

The `%p` specifer is now used to print the pointers.
(second commit)

---

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- Build with the `CLANGPDB` toolchain

## Integration Instructions

- N/A


ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/AArch64/ModuleEntryPoint.S
```diff
@@ -55,7 +55,8 @@ ASM_FUNC(CheckFfaSupport)
 
   svc #0
 
-  MOV64 (x9, ARM_FFA_RET_NOT_SUPPORTED)
+  // Set x9 as ARM_FFA_RET_NOT_SUPPORTED (-1)
```
leiflindholm, line 58: This one's slightly terrifying. Do we understand why this hasn't broken in the past (so we can verify other uses of this macro aren't similarly broken)?

ardbiesheuvel: There is nothing wrong with this code - the assembler is just being overly pedantic.

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/AArch64/ModuleEntryPoint.S
```diff
@@ -55,7 +55,8 @@ ASM_FUNC(CheckFfaSupport)
 
   svc #0
 
-  MOV64 (x9, ARM_FFA_RET_NOT_SUPPORTED)
+  // Set x9 as ARM_FFA_RET_NOT_SUPPORTED (-1)
+  mvn x9, xzr
```
ardbiesheuvel, line 59: You could simply use `mov x9, #ARM_FFA_RET_NOT_SUPPORTED` here if the immediate fits into a single instruction (which is the case for `-1`)

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c
```diff
@@ -1038,9 +1038,9 @@ CEntryPoint (
     goto finish;
   }
 
-  DEBUG ((DEBUG_INFO, "Start Dump Hob: %lx\n", (unsigned long)HobStart));
+  DEBUG ((DEBUG_INFO, "Start Dump Hob: %p\n", HobStart));
```
ardbiesheuvel, line 1041: This is not a Clang issue - it's a MS toolchain issue, and I suspect there will be many more such issues in the ARM platform code. (`unsigned long` is only 32 bits wide on 64-bit Windows). 

I'm not going to object to this change, but going forward, let's change `unsigned long` to `UINTN` in such cases, rather than reinterpret the type.

leiflindholm: Well, it's an LP64/LLP64 issue.

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/AArch64/ModuleEntryPoint.S
leiflindholm: Oh, right, it's freaking out over the sign. I'm not sure that's overly pedantic. Ideally we should handle that.

Well, FF-A says that "All values are considered to be 32-bit signed integers", so maybe a cleaner fix would be to define ARM_FFA_RET_NOT_SUPPORTED as UINT32_MAX?

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/AArch64/ModuleEntryPoint.S
leiflindholm: Yeah, but then we're kind of throwing away the point of having a symbolic name in the first place.

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/AArch64/ModuleEntryPoint.S
leiflindholm: Or update the MOV64 macro to explicitly cast Val to (UINT64) on each use?

ardbiesheuvel: This is asm code no?

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/AArch64/ModuleEntryPoint.S
makubacki: I went with Ard's suggestion as it removes the need for the comment.

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c
makubacki: It sounds like we're okay with this code as-is. Please confirm and I'll close this thread.

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/AArch64/ModuleEntryPoint.S
leiflindholm: Yes, brain fart. Will move to email.

ardbiesheuvel (approved):

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/AArch64/ModuleEntryPoint.S
makubacki: Cherry-picked [Leif's patch](https://github.com/leiflindholm/edk2/commit/a7ff8aebcad1a6621ac52007b6dc530972b26125) as the first commit in the series. Reverted the change in ModuleEntryPoint.S for now.

PR #10761, opened 2025-02-13
MdePkg, MdeModulePkg, ShellPkg: Drop efi 1.10 usb hc protocol

leiflindholm:
# Description

This PR deletes the EFI_USB_HC_PROTOCOL, as discussed in https://edk2.groups.io/g/devel/message/121115.

As stated there, EFI_USB_HC_PROTOCOL existed in the pre-public EFI 1.10 specification,
but never made it into a UEFI release.

This is the sledgehammer approach - if it turns out there are real-life users of the
protocol, we may need to do something with a bit more finesse. _Or_ this may be an
opportunity to nudge them to rewrite code to the protocol covered by a specification.

- [x] Breaking change?
  - This change deletes the definitions and implementations of EFI_USB_HC_PROTOCOL,
    and deletes.
  - The only user of this protocol I am aware of is the Ohci Dxe driver in QuarkSocPkg
    in edk2-platforms.

## How This Was Tested

Build tested only.

This needs to be tested on out-of-tree platforms, so if the approach is deemed acceptable, I would request merging shortly after stable tag is made. As always, reverts are cheap, but I
think getting data on whether the interface is used at all would be valuable.

## Integration Instructions

N/A


pierregondois:
It seems the platform `Silicon/Intel/QuarkSocPkg/QuarkSocPkg.dsc` uses this protocol in `Silicon/Intel/QuarkSocPkg/QuarkSouthCluster/Usb/Ohci/Dxe/Ohci.c`. It also seems that the protocol could be replaced by the `gEfiUsb2HcProtocolGuid` protocol without too many modifications.
However this platform's build is failing currently.

leiflindholm:
> It seems the platform `Silicon/Intel/QuarkSocPkg/QuarkSocPkg.dsc` uses this protocol in `Silicon/Intel/QuarkSocPkg/QuarkSouthCluster/Usb/Ohci/Dxe/Ohci.c`. It also seems that the protocol could be replaced by the `gEfiUsb2HcProtocolGuid` protocol without too many modifications. However this platform's build is failing currently.

I did point this out in the PR description.
This is a fairly old platform, so I'd say it's worth considering whether it should be dropped from edk2-platforms. Obviously updating the driver would be fine too, but I'm not going to :)

What I'm more interested in is whether there are tons of non-public platforms using this not-even-legacy protocol.

leiflindholm:
On that note, do the other QuarkPlatformPkg maintainers have any comments?
@nate-desimone @SaiChaganty.

nate-desimone:
To my knowledge, most x86 SoCs expose their USB host controllers as PCIe Root Complex Integrated Endpoints that implement the XHCI standard. For that reason, the standard XhciDxe from MdeModulePkg works out-of-box without any SoC specific programming.

As far as SoCs that need a bespoke USB host controller driver, that seems to be more common in the embedded space to me. Quark is an interesting exception as it is an x86 SoC designed for the embedded space. I think you will find more of these type of drivers in the ARM ecosystem.

ardbiesheuvel:
> As far as SoCs that need a bespoke USB host controller driver, that seems to be more common in the embedded space to me. Quark is an interesting exception as it is an x86 SoC designed for the embedded space. I think you will find more of these type of drivers in the ARM ecosystem.

AFAIK, the first ever ARM prototype to run EDK2 was the BeagleBoard, which implements a EHCI high speed controller that relies on transaction translation to support FS/LS support, and so it does not rely on UhciDxe, only on EhciDxe.

I think it is safe to say that we can drop support for EFI_USB_HC_PROTOCOL once we're willing to retire QuarkSoc from edk2-platforms.

mdkinney:
The Quark platform can be dropped from edk2-platforms.  Please move forward with these changes.

MdeModulePkg/Bus/Usb/UsbBusDxe/UsbUtility.c
```diff
@@ -63,20 +63,17 @@ UsbHcGetCapability (
...
-                               NumOfPort,
-                               Is64BitCapable
-                               );
-  } else {
-    Status = UsbBus->UsbHc->GetRootHubPortNumber (UsbBus->UsbHc, NumOfPort);
-
-    *MaxSpeed       = EFI_USB_SPEED_FULL;
-    *Is64BitCapable = (UINT8)FALSE;
+  if (UsbBus->Usb2Hc == NULL) {
+    return EFI_INVALID_PARAMETER;
```
ardbiesheuvel, line 67: Can we drop these NULL checks? This code should never be reached if the OpenProtocol() call that populated the Usb2Hc field failed.

leiflindholm: (will do this when back from holiday on Monday)

leiflindholm: Done.

ardbiesheuvel (approved):

PR #10762, opened 2025-02-13
MdeModulePkg: Add a check and close BlockIo2 protocol.

kanagavels97:

During the DiskIoDriverBindingStart, the BlockIo and BlockIo2 protocols are opened for the specified controller handle. The Instance->SharedWorkingBuffer is then allocated based on the BlockSize. If the BlockSize is zero, indicating that the handle does not support Read/Write operations, the buffer will not be allocated. A NULL check is performed on the SharedWorkingBuffer, and if it is NULL, the process jumps to the ErrorExit. In the ErrorExit section, only the BlockIo protocol is closed, while the BlockIo2 protocol remains open.

### DIFF
--- a/MdeModulePkg/Universal/Disk/DiskIoDxe/DiskIo.c
+++ b/MdeModulePkg/Universal/Disk/DiskIoDxe/DiskIo.c
@@ -226,6 +226,15 @@ ErrorExit:
            This->DriverBindingHandle,
            ControllerHandle
            );
+
+    if (gDiskIoPrivateDataTemplate.BlockIo2 != NULL) {
+      gBS->CloseProtocol (
+             ControllerHandle,
+             &gEfiBlockIo2ProtocolGuid,
+             This->DriverBindingHandle,
+             ControllerHandle
+             );
+    }
   }
 
 ErrorExit1:


MdeModulePkg/Universal/Disk/DiskIoDxe/DiskIo.c
```diff
@@ -226,6 +226,15 @@ DiskIoDriverBindingStart (
            This->DriverBindingHandle,
            ControllerHandle
            );
+
+    if (gDiskIoPrivateDataTemplate.BlockIo2 != NULL) {
```
mdkinney, line 230: It looks like there are scenarios where gDiskIoPrivateDataTemplate.BlockIo2 could be non NULL from a different Disk I/O Start call that fails and the current Disk I/O Start that fails would close an unrelated Block I/O 2 protocol.  Using this global variable in the OpenProtocol() calls at the beginning of this function is not a good idea.  The template global should be treated as a const and perhaps should be changed to const and instead add some local variables for the pointers to the Block I/O Protocol and Block I/O 2 Protocol and assign them to the allocated version after AllocateCopyPool().

mdkinney:
@kanagavels97 Any updates from my feedback?

kanagavels97:
I'm closing this PR for now, and will make the changes according to the suggestion, and will create a new PR.

PR #10609, opened 2025-01-10
FatPkg: Validate Reserved FAT Entries on Volume Open

os-d:
# Description

There are two reserved FAT entries in the FAT filesystem that are expected to have valid contents in them. Today the FAT drivers do not validate these entries when reading from a device for the first time. This can cause infinite loops in the FAT driver when trying to read corrupted disks as reported in
https://github.com/tianocore/edk2/issues/9679.

This PR follows the recommended update requested in that bug to check the two reserved FAT entries and validate their contents against the spec defined values in both FatPei and EnhancedFatDxe when opening a device for the first time.

- [ ] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [ ] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

## How This Was Tested

Tested on OVMF, EmulatorPkg, and ArmVirtPkg and confirmed that correctly formatted devices are able to boot.

## Integration Instructions

N/A.

PR #10763, opened 2025-02-13
FatPkg: Revert "FatPkg: Validate Reserved FAT Entries on Volume Open"

os-d:
# Description

This reverts commit https://github.com/tianocore/edk2/commit/58766a472932c485d41163b1746fb1d9e7984f07.

In edk2 commit 58766a4, validation of the two reserved FAT entries
was added.

However, it also checked the return of FatGetFatEntry to MAX_UINT32,
which is what FatGetFatEntry returns when it encounters an error,
e.g. not being able to read the disk. However, MAX_UINT32 is also a
valid value for the reserved FAT entries and under some conditions
these will be returned in the success case.

A FAT volume formatted with these valid values of the reserved FAT
entries will fail to boot an OS because the opening of the volume
will fail.

However, the reason FatGetFatEntry returns MAX_UINT32 is that most
other uses of the function are comparing it against the END_OF_CHAIN
mark, which MAX_UINT32 will trip and those functions will fail out.

Because this is a critical bug that can prevent OS booting and the
bug the original commit was solving was accounting for a bad FAT
filesystem formatting tool, this commit is reverted for now.

Future work will clean up FatGetFatEntry so that it returns an
EFI_STATUS, but that involves more work and this bug needs to be
resolved in the meantime.

Signed-off-by: Oliver Smith-Denny <osde@microsoft.com>

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested by reproing the failure to boot an OS on OVMF and then being able to boot with this revert.

## Integration Instructions

N/A.


os-d:
@mdkinney @makubacki @ardbiesheuvel @leiflindholm @ajfish as I am the only FatPkg maintainer, can you please review this critical bug fix?

@lgao4 , I intend for this to be included in stable-202502 as this fixes a bug that prevents an OS from booting.

@philnoh2 as the reporter of this bug, can you please confirm this fixes your issue before it is merged? I was able to repro your case and this resolved it for me.

mdkinney:
Agree this should be considered for edk2-stable202502.  Will wait for complete review and testing.

philnoh2:
> @philnoh2 as the reporter of this bug, can you please confirm this fixes your issue before it is merged? I was able to repro your case and this resolved it for me.

Like the email information, we confirmed that the PR fixes the issue. Thank you for the quick update.

FatPkg/EnhancedFatDxe/Init.c
```diff
@@ -426,12 +426,11 @@ FatOpenDevice (
   }
 
   //
-  // Check the reserved FAT entries to ensure they contain valid values
+  // Check the reserved FAT entries to ensure they contain valid values.
+  // FatGetFatEntry returns MAX_UINT32 on error, but that is also a valid value, so we do not check
+  // for that value here. Failures to read the disk will have been caught earlier.
   //
   ReservedFatEntry = FatGetFatEntry (Volume, 0);
-  if (Volume->FatEntryBuffer == MAX_UINT32) {
```
leiflindholm, line 432: I really don't like this. FatGetFatEntry has a broken error reporting mechanism.
Deciding to ignore that in some cases because we trust any errors that could happen here cannot be caused for certain input values completely breaks the abstraction the function provides. Documenting that is not a fix.

The correct fix is to change the function API to return status and data separately.

*If* we strongly feel that is too invasive and we decide that we should be cutting corners to squeeze this into the release instead of reverting the change that broke the code, we *could*:
Explicitly call out in the FatGetFatEntry documentation comment that the function is only permitted to return error on out of range situations, and put a simple comment here saying "// FatGetFatEntry cannot fail for offset 0" (and similarly for offset 1 below)

leiflindholm (changes requested):

FatPkg/EnhancedFatDxe/Init.c
os-d: @leiflindholm I completey agree. As I mentioned, I plan on fixing the broken API, but I wanted to get a fix out immediately to resolve boot failures.

If you would prefer to revert the original PR instead, that is fine with me. It is fixing a bug that was reported in 2018 and never fixed, so not a high urgency to have the resolution in this stable tag.

Otherwise I am fine documenting the API now and fixing post stable tag. Please let me know your preference.

leiflindholm: I am a huge fan of reverts. The only question is whether the revert would itself be considered invasive enough that others feel we should push the stable tag out.
To me it (#10609) looks pretty clear and self-contained, so I would vote it isn't.

os-d: I believe it is self-contained and there haven’t been any FatPkg changes since, so a revert should not be an issue. I can put up a PR for that soon, assuming others are okay with that direction

ardbiesheuvel: +1 for reverting and fixing it properly after the stable tag

os-d: I will change this PR to be a revert instead of workaround

os-d: I have updated this PR, including the description and title. Please review when you have a chance.

leiflindholm (approved):

mdkinney (approved):

MrChromebox (approved):

lgao4:
Agree to merge it for stable tag 202502

PR #10772, opened 2025-02-19
IntelFsp2Pkg/FspSecCore: initialize FspMode in FSP_GLOBAL_DATA

niruiyu:
FspSecCore only runs in API mode.
When it allocates memory in stack for FSP_GLOBAL_DATA, it does not initialize the FspMode to FSP_IN_API_MODE.

The patch enhances the code to initialize the FspMode as well.

This change does not have functionality impact as there should be some code elsewhere which sets the FspMode.
This change allows to remove the extra code logic that sets FspMode in API mode.

Cc: Chasel Chiu <chasel.chiu@intel.com>
Cc: Nate DeSimone <nathaniel.l.desimone@intel.com>
Cc: Duggapu Chinni B <chinni.b.duggapu@intel.com>
Cc: Star Zeng <star.zeng@intel.com>
Cc: Ted Kuo <ted.kuo@intel.com>
Cc: Ashraf Ali S <ashraf.ali.s@intel.com>

### DIFF
--- a/IntelFsp2Pkg/FspSecCore/SecFsp.c
+++ b/IntelFsp2Pkg/FspSecCore/SecFsp.c
@@ -133,7 +133,7 @@ FspGlobalDataInit (
   UINTN  Idx;
 
   //
-  // Set FSP Global Data pointer
+  // Set FSP Global Data pointer when FSP runs in API mode
   //
   SetFspGlobalDataPointer (PeiFspData);
   ZeroMem ((VOID *)PeiFspData, sizeof (FSP_GLOBAL_DATA));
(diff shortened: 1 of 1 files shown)


ChaselChiu:
FSP_IN_API_MODE = 0 which is the default FspGlobalData content.
But it should be no harmful to explicitly set it.

ChaselChiu (approved):

lzeng14:
> FSP_IN_API_MODE = 0 which is the default FspGlobalData content. But it should be no harmful to explicitly set it.

So, no other place to initialize it to FSP_IN_API_MODE explicitly but just because value has been 0 after ZeroMem() ?

ChaselChiu:
> > FSP_IN_API_MODE = 0 which is the default FspGlobalData content. But it should be no harmful to explicitly set it.
> 
> So, no other place to initialize it to FSP_IN_API_MODE explicitly but just because value has been 0 after ZeroMem() ?

my understanding was that we only need to override it to 1 when running in DISPATCH_MODE.

niruiyu:
> > > FSP_IN_API_MODE = 0 which is the default FspGlobalData content. But it should be no harmful to explicitly set it.
> > 
> > 
> > So, no other place to initialize it to FSP_IN_API_MODE explicitly but just because value has been 0 after ZeroMem() ?
> 
> my understanding was that we only need to override it to 1 when running in DISPATCH_MODE.

Background of this change: I found a close-source function that allocates FSP_GLOBAL_DATA in dispatch mode. It also sets the FspMode to API or Dispatch mode accordingly by locating the FspApiMode PPI. With the open-source change, the close-source function can only deal with the Dispatch mode logic.

PR #10775, opened 2025-02-20
Rename EDKII_PEI_MP_SERVICES2_PPI to EFI_PEI_MP_SERVICES2_PPI

sachinami:
# Description

EDKII_PEI_MP_SERVICES2_PPI was introduced in https://mantis.uefi.org/mantis/view.php?id=2160

As per the naming conventions of the PI specification, it should be renamed to EFI_PEI_MP_SERVICES2_PPI.

The PI specification correction has been submitted in this mantis ticket: https://mantis.uefi.org/mantis/view.php?id=2500

- [x] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behaviour?
- [ ] Impacts security?
- [ ] Includes tests?

## Integration Instructions
Multi Processor Services 2 PPI consumers have to use the PPI structure EFI_PEI_MP_SERVICES2_PPI present in MdePkg/Include/Ppi/MpServices2.h and PPI GUID gEfiPeiMpServices2PpiGuid from MdePkg.


sachinami:
> Thank you for starting this cleanup. I am wondering if we should keep an alias of the EDK II PPI structure and EDK II GUID to allow downstream consumers time to migrate from the EDK II names to the EFI names.

@mdkinney Sure, the header and the PPI GUID have now been retained for backwards compatibility.

UefiCpuPkg/CpuMpPei/CpuMp2.c
```diff
@@ -1,12 +1,12 @@
...
-  EDKII_PEI_MP_SERVICES2_PPI Implementation code.
+  EFI_PEI_MP_SERVICES2_PPI Implementation code.
 
   Copyright (c) 2019, Intel Corporation. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
 
-#include "CpuMpPei.h"
+#include "CpuMp2Pei.h"
```
lzeng14, line 9: Look like no need to add CpuMp2Pei.h for extra function declarations

UefiCpuPkg/UefiCpuPkg.dec
```diff
@@ -151,8 +151,6 @@
 #
 
 [Ppis]
-  gEdkiiPeiMpServices2PpiGuid =    { 0x5cb9cb3d, 0x31a4, 0x480c, { 0x94, 0x98, 0x29, 0xd2, 0x69, 0xba, 0xcf, 0xba}}
```
lzeng14, line 154: If it is removed, how could it maintain backwards compatibility for the code still consuming this GUID ?

sachinami: The problem is, since the PPI GUID has been moved to MdePkg.dec as gEfiPeiMpServices2PpiGuid.
Having the same GUID here with a diffrent name will lead to a build error.

I don't believe it is possible to have an alias or Conditional Processing in DEC.

Is there a way around this?

niruiyu: > If it is removed, how could it maintain backwards compatibility for the code still consuming this GUID ?

The PR changed the PPI name which already breaks the backward compatibility:)
I thought we would maintain it through typedef and two Guid Name pointing to same GUID value for a while.
+ @mdkinney for comments.

sachinami: The GUID has existed in PI Spec since the 1.8 version (Released on Mar 03, 2023) and in edk2 since the edk2-stable201908 tag.

We would have to sacrifice backwards compatibility in either PI Spec or edk2. Or find a way to have two GUID Names pointing to the same GUID value in DEC files.

CC: @felixpgithub

mdkinney: If we focus the code change on adding the PI Spec 1.8 content to MdePkg with single GUID value, then we can perhaps limit the compatibility to source compatibility and use typedef and #define to support the EDK II names. Update the file  UefiCpuPkg/Include/Ppi/MpServices2.h to include the MdePkg PPI and use typedef to define the EDK II PPI structure and use #define to define the EDK II guid name to the MdePkg PPI guid name.  This would reduce to a GUID declared in only MdePkg.

sachinami: UefiCpuPkg/Include/Ppi/MpServices2.h has been updated for source compatibility as suggested @mdkinney.

mdkinney: I did not realize that the include file name for the PPI was the exact same name in both the UefiCpuPkg and the new one in the MdePkg.  They need to have different names or the order that the packages are listed in a module INF will not resolve to the correct one.

sachinami: To maintain backwards compatibility the header name cannot be changed in UefiCpuPkg.
The new one in MdePkg could be renamed to something like EfiMpServices2.h, but are there any concerns for forward compatibility in case UefiCpuPkg/Include/Ppi/MpServices2.h is removed in the future.

mdkinney: @leiflindholm @ajfish @ardbiesheuvel @makubacki Please review this PR and discussion. It involves resolving conflicts between a EDK II specific PPI and a PPI added to the Platform Initialization specification.

This PPI is defined in the Platform Initialization spec, so perhaps a prefix of `Pi` would be better to match some of the other include files in `MdePkg`.  Suggest `PiMpServices2.h`.

No concerns if/when `UefiCpuPkg/Include/Ppi/MpServices2.h` is removed.  That include file is only for the EDK II specific version of the PPI.  If all components are updated to use the PI version of the PPI, then they would have already switched to the `#include <Ppi/PiMpServices2.h>` and removal of that file would not have any impact at that time.

ardbiesheuvel: I might be missing something here, but given that all users of UefiCpuPkg will also include MdePkg, can't we just move MpServices2.h into MdePkg, and do something like

```diff
--- a/UefiCpuPkg/Include/Ppi/MpServices2.h
+++ b/MdePkg/Include/Ppi/MpServices2.h
@@ -17,6 +17,7 @@
   }
 
 typedef struct _EDKII_PEI_MP_SERVICES2_PPI EDKII_PEI_MP_SERVICES2_PPI;
+typedef struct _EDKII_PEI_MP_SERVICES2_PPI EFI_PEI_MP_SERVICES2_PPI;
 
 /**
   Get the number of CPU's.
@@ -274,5 +275,6 @@ struct _EDKII_PEI_MP_SERVICES2_PPI {
 };
 
 extern EFI_GUID  gEdkiiPeiMpServices2PpiGuid;
+extern EFI_GUID  gEfiPeiMpServices2PpiGuid;
 
 #endif

--- a/MdePkg/MdePkg.dec
+++ b/MdePkg/MdePkg.dec
@@ -1098,6 +1098,8 @@ [Ppis]
   ## Include/Ppi/Rng.h
   gEfiRngPpiGuid                 = { 0xeaed0a7e, 0x1a70, 0x4c2b, { 0x85, 0x58, 0x37, 0x17, 0x74, 0x56, 0xd8, 0x06 }}
 
+  gEfiPeiMpServices2PpiGuid =    { 0x5cb9cb3d, 0x31a4, 0x480c, { 0x94, 0x98, 0x29, 0xd2, 0x69, 0xba, 0xcf, 0xba}}
+
 [Protocols]
   ## Include/Protocol/MemoryAccept.h
   gEdkiiMemoryAcceptProtocolGuid = { 0x38c74800, 0x5590, 0x4db4, { 0xa0, 0xf3, 0x67, 0x5d, 0x9b, 0x8e, 0x80, 0x26 }}

```

as a first step, and add a duplicate GUID exception to the CI? Then, we can update existing users at our leisure, including out-of-tree ones. And if/when we feel the need, we can drop the EDKII version.

mdkinney: What you suggest would be functional.

However, MdePkg should only contain content from ind std specs.  Adding an EDK II specific content to MdePkg would break that rule.

ardbiesheuvel: > What you suggest would be functional.
> 
> However, MdePkg should only contain content from ind std specs. Adding an EDK II specific content to MdePkg would break that rule.

We are dealing with the definition of a PPI that was blessed by the spec authors and incorporated into the PI spec. The GUID is the same and the protocol definition is the same, only the name has a EDKII prefix.

So I don't think this still qualifies as EDK II specific content, and making the migration seamless by retaining a EDKII alias for the PPI for a limited timeframe is reasonably IMO.

mdkinney: I agree the definition of the PPI Structure and the PPI GUID are the same.  My concern is about use of the EDK II prefixed type names and GUID names.  You suggestion is reasonable.  Is there a way to generate a build warning for builds that are using the EDK II prefixed names so all consumers know to update and we can remove the EDK II names as quickly as possible?

ardbiesheuvel: > Is there a way to generate a build warning for builds that are using the EDK II prefixed names so all consumers know to update and we can remove the EDK II names as quickly as possible?

Given that build warnings are fatal these days anyway, this will break the build for those consumers.

So what we might do instead is declare the EDKII aliases conditionally, based on a pre-processor #define. That gives out-of-tree platforms an easy way to fix the broken build (adding the `-D` option to the compiler command line) when they upgrade, and this also serves as a natural opt-in to the deprecation and subsequent removal of those definitions. (i.e., setting the `-D` option implies an acknowledgement of the deprecation).

[For the GUID in particular, we could force a build error if the `-D` option is not set by conditionally emitting a declaration of `gEdkiiPeiMpServices2PpiGuid` into the PPI header with a conflicting type, considering that #ifdefs don't work in .DEC files]

mdkinney: That matches what we have done in the past for the unsafe string functions using the define `DISABLE_NEW_DEPRECATED_INTERFACES`.  Though I notice that has been removed from code but is still in many DSC files.

So suggestion is to remove MpServices.h from UefiCpuPkg.  Add MpServices.h to MdePkg following PI spec and add something like #ifdef ENABLE_DEPRECATED_EDKII_MP_SERVICES around EDK II typedef and GUID.

ardbiesheuvel:
Please ensure that no platform gets broken even temporarily when only some of the patches are applied in the order you are presenting them here.

Generally, that means the following when renaming and/or moving something

- create the new version
- move all users to the new version
- remove the old version

UefiCpuPkg/UefiCpuPkg.dec
sachinami: @ardbiesheuvel @mdkinney Thank you for your comments.

MpServices2.h has been removed from UefiCpuPkg and the instance in MdePkg has been modified for conditional backwards compatibility.

Also I believe it will be worth it to leave DISABLE_NEW_DEPRECATED_INTERFACES in the DSC files for now.
In UEFI 2.10A specification there are a bunch of items which were removed. These are definitions and interfaces which were deprecated in earlier versions of the spec and have been removed now.

Some of them are :

- gEfiDeviceIoProtocolGuid
- gEfiUgaIoProtocolGuid
- gEfiUgaDrawProtocolGuid
- gEfiScsiPassThruProtocolGuid
- EFI_VARIABLE_AUTHENTICATION
- EFI_IP4_CONFIG_PROTOCOL
- EFI_DRIVER_CONFIGURATION_PROTOCOL
- EFI_DRIVER_DIAGNOSTICS_PROTOCOL
- EFI_COMPONENT_NAME_PROTOCOL
  
    
  
I was thinking of creating a PR which will remove some of these and wrap up the others in DISABLE_NEW_DEPRECATED_INTERFACES.

mdkinney: I would recommend a new define name be used for this and the define name should include the name/version of the spec the items were deprecated.  That way, a platform can choose which deprecated items to keep or not based in spec name/version. This also makes it easier to review platform DSC file for these types of defines.

For example, for UEFI 2.10, we could use a define name such as `ENABLE_UEFI_2_10_DEPRECATED_INTERFACES`

sachinami: Got it @mdkinney. Thank you!

sachinami:
@ardbiesheuvel The commits have been re-ordered.

MdePkg/Include/Ppi/MpServices2.h
```diff
@@ -280,4 +280,15 @@ struct _EFI_PEI_MP_SERVICES2_PPI {
...
+#ifdef ENABLE_DEPRECATED_EDKII_MP_SERVICES2
+
+#define EDKII_PEI_MP_SERVICES2_PPI_GUID  EFI_PEI_MP_SERVICES2_PPI_GUID
+
+typedef EFI_PEI_MP_SERVICES2_PPI EDKII_PEI_MP_SERVICES2_PPI;
+
+#define gEdkiiPeiMpServices2PpiGuid  gEfiPeiMpServices2PpiGuid
+
+#endif
+
```
ardbiesheuvel, line 293: In order for this series to be bisectable, these declarations should be combined with the first patch, and the same patch should remove the original header file from UefiCpuPkg. Only the #ifdef/#endif should be added at the very end of the series.

Otherwise, there will be ambiguity around which set of declarations are visible when compiling code that depends on both MdePkg and UefiCpuPkg.

Doing so will break code that relies on UefiCpuPkg but not on MdePkg, but the assumption is that such code does not exist.

sachinami: I agree that these declarations should be combined with the first patch. But if the same patch should remove the original header file from UefiCpuPkg, that would mean that MdePkg changes and UefiCpuPkg would be going into the same commit.

And I believe cross-package commits are not permitted in edk2.

We could avoid this, if we bank on the assumption there there would be no code that relies on UefiCpuPkg but not on MdePkg.

ardbiesheuvel: > I agree that these declarations should be combined with the first patch. But if the same patch should remove the original header file from UefiCpuPkg, that would mean that MdePkg changes and UefiCpuPkg would be going into the same commit.
> 
> And I believe cross-package commits are not permitted in edk2.
>

It is permitted in exceptional cases.

> We could avoid this, if we bank on the assumption there there would be no code that relies on UefiCpuPkg but not on MdePkg.

That still means that the new header file that you introduce, which has the same base name, must provide both the old and the new PPI, without relying on any #defines.

So 
- patch `#1` adds the new header in MdePkg, with backward-compatible aliases defined unconditionally
- patch `#2` removes the old header
- patches `#3` - `#(n-1)` update the in-tree users
- patch `#n` adds the #ifdef conditionals around the backward-compatible aliases.

How does that sound?

sachinami: Sounds Good @ardbiesheuvel. The commits have been changed accordingly.

sachinami: @ardbiesheuvel @mdkinney Do we have any other concerns on this patch series.

ardbiesheuvel: I think @mdkinney wanted a different name for the preprocessor macro?

sachinami: If you are talking about [this comment](https://github.com/tianocore/edk2/pull/10775#discussion_r2007872585), that pre-processor macro is for an upcoming patch series that has no relation to this one.

sachinami: @mdkinney Can this patch series be merged.

ardbiesheuvel (approved):

mdkinney:
@ardbiesheuvel This PR was merged before I was able to confirm compatibility with downstream use of EDK II PEI MP Services 2 PPI that needs to be migrated to PEI MP Service 2 PPI.

Unfortunately, this PR breaks some downstream use cases, with the hardest one to handle being an INF that
references `gEdkiiPeiMpServices2Ppi` in a `[Ppis]` section.  The define for backwards compatibility can not 
address that use case. I have opened PR #10989 to add back the definitions required and update code comments.

We can consider keeping this compatibility till the next stable tag and remove all references to the 
EDK II PEI MP Services 2 PPI after the next stable tag release.

ardbiesheuvel:
> @ardbiesheuvel This PR was merged before I was able to confirm compatibility with downstream use of EDK II PEI MP Services 2 PPI that needs to be migrated to PEI MP Service 2 PPI.
> 

Apologies, I did not realize this was still pending.

> We can consider keeping this compatibility till the next stable tag and remove all references to the EDK II PEI MP Services 2 PPI after the next stable tag release.

Ack.

sachinami:
> > @ardbiesheuvel This PR was merged before I was able to confirm compatibility with downstream use of EDK II PEI MP Services 2 PPI that needs to be migrated to PEI MP Service 2 PPI.
> 
> Apologies, I did not realize this was still pending.
> 
> > We can consider keeping this compatibility till the next stable tag and remove all references to the EDK II PEI MP Services 2 PPI after the next stable tag release.
> 
> Ack.

Apologies from my end. I also did not know that this was pending.

PR #10785, opened 2025-02-25
Prevent VS2022 Linker Issues

makubacki:
# Description

**MdePkg/BaseMemoryLib: Prevent potential VS2022 linker failure**

After updating between various VS2022 versions such as 17.4 to
17.5, , linker failures began to appear in several modules because
`memset` is an unresolved symbol.

The following functions in BaseMemoryLib/MemLibGeneric.c have their
loop pattern replaced with the `memset` intrinsic function:

- `InternalMemSetMem16()`
- `InternalMemSetMem32()`
- `InternalMemSetMem64()`

An example of an error related to `InternalMemSetMem64()` in
VariableSmmRuntimeDxe is shown below:

```
INFO - BaseMemoryLib.lib(MemLibGeneric.obj) : error LNK2001:
         unresolved external symbol memset
INFO - <...>\VariableSmmRuntimeDxe.dll : fatal error LNK1120:
         1 unresolved externals
```

This was reproduced in several environments including:

- Public VM image:
  https://github.com/actions/runner-images/blob/win22/20230226.1/images/win/Windows2022-Readme.md

- Locally when updating from 17.4.4 to 17.5.1

> Note: This image (with 17.4) does not have this issue
  https://github.com/actions/runner-images/blob/win22/20230219.1/images/win/Windows2022-Readme.md

This change updates the type cast for the destination buffer to be
a pointer to `volatile` data to prevent this optimization with a
relatively minimum delta to prior implementation.

---

**MdeModulePkg/HiiDatabaseDxe: Prevent linker error**

Prevent an issue where `memcpy()` instrinsic is being used after
recent MSVC linker update in windows-2022 VM image from 14.31.31103
to 14.32.31326.

---

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Built with and without the change on an affected version (e.g. Visual Studio 17.5.1)
locally and on CI agents.

## Integration Instructions

- N/A


mdkinney:
I do not disagree with the formal correctness of this change to use CopyMem() for structure assignments.

However, I am just concerned about the performance impacts of this change, because I assume the previous compiler was just doing a 32-bit load and 32-bit store of each pixel structure.  Calling CopyMem() will do a lot more work.

I think changing these functions to use `EFI_GRAPHICS_OUTPUT_BLT_PIXEL_UNION` instead of `EFI_GRAPHICS_OUTPUT_BLT_PIXEL` will allow the use of the `UINT32 Raw` field to be used to keep the current algorithm and preserve the 32-bit load/store behavior.

makubacki:
> I do not disagree with the formal correctness of this change to use CopyMem() for structure assignments.
> 
> However, I am just concerned about the performance impacts of this change, because I assume the previous compiler was just doing a 32-bit load and 32-bit store of each pixel structure. Calling CopyMem() will do a lot more work.
> 
> I think changing these functions to use `EFI_GRAPHICS_OUTPUT_BLT_PIXEL_UNION` instead of `EFI_GRAPHICS_OUTPUT_BLT_PIXEL` will allow the use of the `UINT32 Raw` field to be used to keep the current algorithm and preserve the 32-bit load/store behavior.

That's fair. I'll try that and report back.

makubacki:
@mdkinney, made the update, let me know if that's what you had in mind.

MdeModulePkg/Universal/HiiDatabaseDxe/Image.c
```diff
@@ -549,13 +549,13 @@ ImageToBlt (
     OffsetY1 = Width * Ypos;
     OffsetY2 = ImageOut->Width * (BltY + Ypos);
     for (Xpos = 0; Xpos < Width; Xpos++) {
-      SrcPixel = BltBuffer[OffsetY1 + Xpos];
+      SrcPixel.Raw = *(UINT32 *)(&BltBuffer[OffsetY1 + Xpos]);
```
mdkinney, line 552: Don't use UINT32 *.  Instead, declare a local:
```
  EFI_GRAPHICS_OUTPUT_BLT_PIXEL_UNION *BltBufferPixel;

  BltBufferPixel = (EFI_GRAPHICS_OUTPUT_BLT_PIXEL_UNION *)BltBuffer;

  SrcPixel.Raw = BltBufferPixel[OffsetY1 + Xpos].Raw;
```

MdeModulePkg/Universal/HiiDatabaseDxe/Image.c
```diff
@@ -549,13 +549,13 @@ ImageToBlt (
     OffsetY1 = Width * Ypos;
     OffsetY2 = ImageOut->Width * (BltY + Ypos);
     for (Xpos = 0; Xpos < Width; Xpos++) {
-      SrcPixel = BltBuffer[OffsetY1 + Xpos];
+      SrcPixel.Raw = *(UINT32 *)(&BltBuffer[OffsetY1 + Xpos]);
       if (Transparent) {
         if (CompareMem (&SrcPixel, &ZeroPixel, 3) != 0) {
-          ImageOut->Image.Bitmap[OffsetY2 + BltX + Xpos] = SrcPixel;
+          WriteUnaligned32 ((UINT32 *)&ImageOut->Image.Bitmap[OffsetY2 + BltX + Xpos], SrcPixel.Raw);
```
mdkinney, line 555: Why are `WriteUnaligned32()` calls required?  Can `ImageOut->Image.Bitmap` be unaligned?

Also do not use UINT32 *.  

```
EFI_GRAPHICS_OUTPUT_BLT_PIXEL_UNION *ImageBitmap;

ImageBitmap = (EFI_GRAPHICS_OUTPUT_BLT_PIXEL_UNION *)ImageOut->Image.Bitmap;

ImageBitmap[OffsetY2 + BltX + Xpos].Raw = SrcPixel.Raw;
```

MdeModulePkg/Universal/HiiDatabaseDxe/Image.c
```diff
@@ -549,13 +549,13 @@ ImageToBlt (
...
-      SrcPixel = BltBuffer[OffsetY1 + Xpos];
+      SrcPixel.Raw = *(UINT32 *)(&BltBuffer[OffsetY1 + Xpos]);
       if (Transparent) {
         if (CompareMem (&SrcPixel, &ZeroPixel, 3) != 0) {
-          ImageOut->Image.Bitmap[OffsetY2 + BltX + Xpos] = SrcPixel;
+          WriteUnaligned32 ((UINT32 *)&ImageOut->Image.Bitmap[OffsetY2 + BltX + Xpos], SrcPixel.Raw);
         }
       } else {
-        ImageOut->Image.Bitmap[OffsetY2 + BltX + Xpos] = SrcPixel;
+        WriteUnaligned32 ((UINT32 *)&ImageOut->Image.Bitmap[OffsetY2 + BltX + Xpos], SrcPixel.Raw);
```
mdkinney, line 558:
```
ImageBitmap[OffsetY2 + BltX + Xpos].Raw = SrcPixel.Raw;
```

MdeModulePkg/Universal/HiiDatabaseDxe/Image.c
makubacki: Good idea. Updated.

MdeModulePkg/Universal/HiiDatabaseDxe/Image.c
makubacki: I don't think it's necessary after looking at it again. Removed it.

mdkinney (approved):

PR #10789, opened 2025-02-26
Remove uhci peim and ppi

ardbiesheuvel:
Similar to @leiflindholm's #10761, which removes the obsolete, older USB host controller from the DXE phase, this series drops the obsolete, older PPI which is only produced by UhciPei, which itself is obsolete.


leiflindholm:
I don't know if this needs to be removed, but ... it would be good to see if anyone cares about it? If not, then yeah, we should drop it.

leiflindholm:
> I don't know if this needs to be removed, but ... it would be good to see if anyone cares about it? If not, then yeah, we should drop it.

No reaction for a month, I think that's an answer of sorts.
Happy for this to be merged post stable tag (gives us plenty of time to revert in case someone screams).

leiflindholm (approved):

mdkinney:
@mergify rebase

lgao4 (approved):

PR #10795, opened 2025-02-27
pip-requirements.txt: Update pip modules

makubacki:
# Description

Updates from dependabot for pytools have been accumulating over
time. This updates the modules to latest.

`edk2-pytool-library` changes:

https://github.com/tianocore/edk2-pytool-library/compare/v0.21.9...v0.23.0

`edk2-pytool-extensions` changes:

https://github.com/tianocore/edk2-pytool-extensions/compare/v0.27.10...v0.29.0

Other pip modules are updated to latest versions tested alongside these pytool versions.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- CI
- Pytools integration and unit tests

## Integration Instructions

Notes:

In these releases, Python 3.10 is a Minimum Supported Version while 3.11, 3.12, and 3.13 are official versions. For more details:

https://github.com/tianocore/edk2-pytool-extensions/blob/master/docs/contributor/python_msv.md

PR #10797, opened 2025-02-28
CI: Install a fresh python virtual env every time (Linux)

osteffenrh:
# Description

Install a fresh python virtual environment every time to ensure
correct permissions and compatibility of the packages.
This is more robust than relying on being able to upgrade
an existing one.

pip upgrades are currently failing due to incorrect file permissions
in the container image virtual environment.

Let's not make use of the venv in the container and set up a fresh one
each time.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Test CI runs, for example #10796

## Integration Instructions
 

--

### DIFF
--- a/.azurepipelines/templates/platform-build-run-steps.yml
+++ b/.azurepipelines/templates/platform-build-run-steps.yml
@@ -55,7 +55,12 @@ steps:
     architecture: "x64"
   condition: ne('${{ parameters.usePythonVersion }}', '')
 
-- script: pip install -r pip-requirements.txt --upgrade
+- bash: |
+    virtualenv ~/venv
+    . ~/venv/bin/activate
+    echo "##vso[task.setvariable variable=VIRTUAL_ENV]${VIRTUAL_ENV}"
+    echo "##vso[task.prependpath]${VIRTUAL_ENV}/bin"
+    pip install -r pip-requirements.txt --upgrade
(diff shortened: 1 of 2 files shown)


ardbiesheuvel (approved):

Javagedes (approved):
I'll preface with the fact I have no problem with this (and my approval of the PR shows that too); however, I just want to understand the problem a little more. What is the use case in which upgrading/installing dependencies is failing? Are there CI pipelines in which this is happening, or is this internal usage of the edk2 container? If so what's the different use case that it's failing there then here?

makubacki (approved):

makubacki:
> I'll preface with the fact I have no problem with this (and my approval of the PR shows that too); however, I just want to understand the problem a little more. What is the use case in which upgrading/installing dependencies is failing? Are there CI pipelines in which this is happening, or is this internal usage of the edk2 container? If so what's the different use case that it's failing there then here?

It is to unblock https://github.com/tianocore/edk2/pull/10795. Issue was originally filed in the container repo here but that was to track the issue with the container it appears in. The solution here in edk2 is fine to me because outside the pipeline local users should be running as root and/or can set up their own virtual env.

https://github.com/tianocore/containers/issues/107

PR #10799, opened 2025-02-28
DynamicTablesPkg: Tolerate NULL when freeing MADT table

jgarver:
# Description

In some failure legs of BuildMadtTable (), we could end up in FreeMadtTableResources () with (*Table == NULL).  That's a valid error state and should be tolerated by FreeMadtTableResources () instead of ASSERTing.

Reviewed-by: Jeff Brasen <jbrasen@nvidia.com>
Reviewed-by: Jeshua Smith <jeshuas@nvidia.com>

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

No longer asserting.

## Integration Instructions

N/A


pierregondois:
Hello Jake,
the patch looks good to me, but I was wondering if you did hit this ASSERT in other table generators. For instance the GTDT generator has the same pattern,
Regards,
Pierre

jgarver:
I did not personally.  Due to other circumstances, I ended up in this error leg and hit the ASSERT.

pierregondois (approved):
The patch lgtm,
NIT: I think it was advised to provide reviews directly instead of submitting patches with `Reviewed-by` tags.

samimujawar (approved):

PR #10801, opened 2025-02-28
StandaloneMmPkg: Remove Assert when gMmCommBufferHobGuid does not exist.

apop5:
# Description
When running StandaloneMmCore on a system, do not assert if gMmCommBufferHobGuid is not found. Output a debug message that details that the gMmCommBufferHobGuid was not found, that only root MmiHandlers will be dispatched.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Tested on virtual platform with top of tree TFA, with HOB_LIST=1 TRANSFER_LIST=1 used when building TFA.

## Integration Instructions
N/A

### DIFF
--- a/StandaloneMmPkg/Core/StandaloneMmCore.c
+++ b/StandaloneMmPkg/Core/StandaloneMmCore.c
@@ -446,8 +446,9 @@ MmCorePrepareCommunicationBuffer (
   mInternalCommBufferCopy = NULL;
 
   GuidHob = GetFirstGuidHob (&gMmCommBufferHobGuid);
-  ASSERT (GuidHob != NULL);
   if (GuidHob == NULL) {
+    DEBUG ((DEBUG_ERROR, "Failed to find MM Communication Buffer HOB\n"));
+    DEBUG ((DEBUG_ERROR, "Only Root MMI Handlers will be supported!\n"));
     return;
   }
 


LeviYeoReum:
I think this patch wouldn't work properly.

When some of request from NsWorld comes, at MmEntryPoint, It handles the request first and
after that the same operation will handle again in MmEntryPoint's asynchronous MMI handle.
(IOW, the same handled request would be handled again by RootMmihandler -- see:
https://github.com/tianocore/edk2/blob/master/ArmPkg/Drivers/StandaloneMmCpu/EventHandle.c#L174)

BTW, why does MmEntrryPoint calls asynchronous MMI handler again thou it was handled via synchronous way?
(https://github.com/tianocore/edk2/blob/master/StandaloneMmPkg/Core/StandaloneMmCore.c#L604)

apop5:
@LeviYeoReum 

I see your point. With the previous change, while the system will boot, it was still running the same handlers multiple times.

To get past the assert that I am encountering (MmCorePrepareCommunicationBuffer is asserting when gMmCommBufferHobGuid is not found for aarch64), I removed the assert and am allowing the StMmCore to operate when it does not have a valid gMmCommBufferHobGuid . This will leave systems without a valid gMmCommBufferHobGuid to only dispatch Mmi Root handlers.

For the AARCH64 scenario, when an MMI Occurs, 
DelegatedEventLoop() in ArmStandaloneMmCoreEntryPoint gets the communication buffer from passed parameter.
It calls into PiMmStandaloneMmCpuDriverEntry (StandaloneMmCpu), which allocates a buffer (global variable), fill it with Mm Communicate buffer, and then calls the MmEntryPoint (StandaloneMmCore)
The MmEntryPoint (StandaloneMmCore) is expecting that there will be a valid comm buffer, but for Aarch64 it will not find one, and it will call MmiManage on the way out.
MmiManage will call all the MMI root handlers (Handlers that were registered with a NULL for their Guid)
   StandaloneMmCpu's Root handler PiMmCpuTpFwRootMmiHandler, will get called
   PiMmCpuTpFwRootMmiHandler will use the global variable allocated in PiMmStandaloneMmCpuDriverEntry, and call MmiManage again with a valid buffer from the parameters passed to DelegatedEventLoop
   MmiManage will see that there is a valid comm buffer and it will dispatch the correct MmiHandler

 
This is leaving Aarch64 in a very complex state for handling Mmis.  StandaloneMmCpu => StandaloneMmCore => StandaloneMmCpu => StandaloneMmCore => Mmi Non Root Handler  is inefficient for handling a single MMI.

apop5:
@ardbiesheuvel 

Can you please review this?

PR #10816, opened 2025-03-03
MdePkg: fix fru device description string copy error.

thejaputta:
As per IPMI spec, fru string can be maximum of 16 bytes length and fru length field can take 5 bits (to hold values up to 16 - 10000b).

Current implementation allocates only 4 bits for Fru String Length and this is causing an error when 16 bytes length fru string needs to be copied as the first 4 bits in 0x10 is 0.

Fixed the structure IPMI_SDR_RECORD_DEV_ID_STR_TYPE_LENGTH, by allocating 5 bits for Length field as per the spec instead of 4.

### DIFF
--- a/MdePkg/Include/IndustryStandard/IpmiNetFnStorage.h
+++ b/MdePkg/Include/IndustryStandard/IpmiNetFnStorage.h
@@ -11,6 +11,7 @@
   and Appendix H, Sub-function Assignments.
 
   Copyright (c) 1999 - 2018, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2025 - Nvidia Corporation.  All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 **/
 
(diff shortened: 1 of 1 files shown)


nicklela (approved):
Looks good to me.

mdkinney (approved):
Good catch.  The IPMI  Specification is clear on layout of this structure and the updates look good.

changab (approved):

nicklela:
Below build error from CI is not related to this pull request. @mdkinney may I know how do we handle this case? Do we need separate pull request to fix CI failure first?
```
ERROR - Error Duplicate guid:  462CAA21-7614-4503-836E-8AB6F4662331 (2)
	GUID: 462CAA21-7614-4503-836E-8AB6F4662331 NAME: UiApp FILE: /__w/1/s/MdeModulePkg/Application/UiApp/UiApp.inf
	GUID: 462CAA21-7614-4503-836E-8AB6F4662331 NAME: gUiAppFileGuid FILE: /__w/1/s/OvmfPkg/OvmfPkg.dec
```

mdkinney:
@mergify rebase

PR #10818, opened 2025-03-04
RedfishPkg: Populate serial number in SMBIOS type 42 record

thejaputta:


RedfishPkg/Include/Library/RedfishHostInterfaceLib.h
```diff
@@ -74,4 +75,17 @@ RedfishPlatformHostInterfaceNotification (
...
+/**
+  Get USB Virtual Serial Number.
+
+  @param[OUT] SerialNumber      Pointer to retrieve complete serial number.
+
+  @retval EFI_SUCCESS         Serial number is returned.
+  @retval Others              Failed to get the serial number.
+**/
+EFI_STATUS
+RedfishPlatformHostInterfaceUSBSerialNumber (
```
changab, line 87: The funtction RedfishPlatformHostInterfaceUSBSerialNumber doen't follow ECC. Please update it to RedfishPlatformHostInterfaceUsbSerialNumber (), thanks.

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
```diff
@@ -82,6 +85,18 @@ RedfishCreateSmbiosTable42 (
...
   } else {
     DeviceDataLength = DeviceDescriptor->DeviceDescriptor.UsbDeviceV2.Length;
+    SerialNumStrLen  = 0;
+    SerialNumber     = NULL;
+    Status           = RedfishPlatformHostInterfaceUSBSerialNumber (&SerialNumber);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "%a: Fail to get USB virtual serial number, %r.", __func__, Status));
+      DeviceDescriptor->DeviceDescriptor.UsbDeviceV2.SerialNumberStr = 0;
+    } else {
+      if (SerialNumber) {
```
changab, line 95: Please update it to if (SerialNumber != NULL) for the BOOLEAN condition check.

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
```diff
@@ -54,6 +54,9 @@ RedfishCreateSmbiosTable42 (
   EFI_SMBIOS_PROTOCOL                *Smbios;
   EFI_SMBIOS_HANDLE                  MemArrayMappedAddrSmbiosHandle;
   EFI_HANDLE                         Handle;
+  CHAR8                              *SerialNumber;
+  UINT8                              SerialNumStrLen;
+  UINT8                              StringCount = 1;
```
changab, line 59: We should have the variable initial on a separate line.

UINT8                              StringCount;

StringCount = 1;

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
```diff
@@ -82,6 +85,18 @@ RedfishCreateSmbiosTable42 (
...
+    SerialNumStrLen  = 0;
+    SerialNumber     = NULL;
+    Status           = RedfishPlatformHostInterfaceUSBSerialNumber (&SerialNumber);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "%a: Fail to get USB virtual serial number, %r.", __func__, Status));
+      DeviceDescriptor->DeviceDescriptor.UsbDeviceV2.SerialNumberStr = 0;
+    } else {
+      if (SerialNumber) {
+        SerialNumStrLen                                                = AsciiStrLen (SerialNumber);
+        DeviceDescriptor->DeviceDescriptor.UsbDeviceV2.SerialNumberStr = StringCount++;
```
changab, line 97: The reference of StringCount looks to me not much meaningful as SerialNumber is not defined as a pointer array, and unlike we will have more than one serial number, right?

RedfishPkg/Include/Library/RedfishHostInterfaceLib.h
thejaputta: Done, thank you.

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
thejaputta: Done, thank you.

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
thejaputta: Done, thank you.

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
thejaputta: Agree. Removed the string count increment as there is no other strings to populate in SMBIOS type 42 record.
Thank you.

igorkulchytskyy (approved):
Looks good for me

nicklela (approved):

thejaputta:
Hi, all failing CI checks are related to "SpellCheck failing for 99 words". Those failing words are not related to this change. Any suggestions on how to handle this issue?

mdkinney:
The spelling checks are in log, but does not fail CI.  The real error is above those messages in the log:

```
PROGRESS - Running Build RELEASE
ERROR - Compiler #2220 from D:\a\1\s\RedfishPkg\RedfishHostInterfaceDxe\RedfishHostInterfaceDxe.c(97): the following warning is treated as an error
WARNING - Compiler #4244 from D:\a\1\s\RedfishPkg\RedfishHostInterfaceDxe\RedfishHostInterfaceDxe.c(97): '=': conversion from 'UINTN' to 'UINT8', possible loss of data
ERROR - Compiler #1077 from NMAKE : fatal '"C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.43.34808\bin\Hostx86\x86\cl.exe" /FoD:\a\1\s\Build\RedfishPkg\RELEASE_VS2022\IA32\RedfishPkg\RedfishHostInterfaceDxe\RedfishHostInterfaceDxe\OUTPUT\.\ /showIncludes /nologo /arch:IA32 /c /WX /GS /W4 /Gs32768 /D UNICODE /O1b2 /GL /FIAutoGen.h /EHs-c- /GR- /GF /Gw /ID:\a\1\s\RedfishPkg\RedfishHostInterfaceDxe  /ID:\a\1\s\Build\RedfishPkg\RELEASE_VS2022\IA32\RedfishPkg\RedfishHostInterfaceDxe\RedfishHostInterfaceDxe\DEBUG  /ID:\a\1\s\MdePkg  /ID:\a\1\s\MdePkg\Include  /ID:\a\1\s\MdePkg\Test\UnitTest\Include  /ID:\a\1\s\MdePkg\Test\Mock\Include  /ID:\a\1\s\MdePkg\Library\MipiSysTLib\mipisyst\library\include  /ID:\a\1\s\MdePkg\Include\Ia32  /ID:\a\1\s\MdeModulePkg  /ID:\a\1\s\MdeModulePkg\Include  /ID:\a\1\s\MdeModulePkg\Test\Mock\Include  /ID:\a\1\s\RedfishPkg  /ID:\a\1\s\RedfishPkg\Include  /ID:\a\1\s\RedfishPkg\PrivateInclude\Crt  /ID:\a\1\s\RedfishPkg\Library\JsonLib  /ID:\a\1\s\RedfishPkg\Library\JsonLib\jansson\src @D:\a\1\s\Build\RedfishPkg\RELEASE_VS2022\IA32\RedfishPkg\RedfishHostInterfaceDxe\RedfishHostInterfaceDxe\OUTPUT\cc_resp_1.txt' : return code '0x2'
ERROR - Compiler #7000 from : Failed to execute command
ERROR - EDK2 #002 from : Failed to build module
CRITICAL - Build failed
```

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
```diff
@@ -82,6 +86,18 @@ RedfishCreateSmbiosTable42 (
...
     DeviceDataLength = DeviceDescriptor->DeviceDescriptor.UsbDeviceV2.Length;
+    SerialNumStrLen  = 0;
+    SerialNumber     = NULL;
+    Status           = RedfishPlatformHostInterfaceUsbSerialNumber (&SerialNumber);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "%a: Fail to get USB virtual serial number, %r.", __func__, Status));
+      DeviceDescriptor->DeviceDescriptor.UsbDeviceV2.SerialNumberStr = 0;
+    } else {
+      if (SerialNumber != NULL) {
+        SerialNumStrLen                                                = AsciiStrLen (SerialNumber);
```
mdkinney, line 97: Needs to be:

```
SerialNumStrLen                                                = (UINT8)AsciiStrLen (SerialNumber);
```

Also may need to verify or guarantee that that the length of the SerialNumber string fits in a 8-bit field.

thejaputta: Got it, thanks for the clarification!
I updated the serial number check to make sure it is under UINT8_MAX size and typecasted before copying.

mdkinney: The code update makes it safe.  It does call the string size function twice.  Could add one more local to get the result once and then compare against MAX_UINT8.

thejaputta: Updated the patch by using local to store the serial number string length and using it to check if its size is less than uint8 max. Thank you.

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
```diff
@@ -24,6 +24,8 @@
 #include <Library/UefiBootServicesTableLib.h>
 #include <Library/UefiRuntimeServicesTableLib.h>
 
+#define UINT8_MAX  0xff
```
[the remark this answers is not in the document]
thejaputta: Agree, thanks for the inputs.

RedfishPkg/Library/PlatformHostInterfaceLibNull/PlatformHostInterfaceLibNull.c
```diff
@@ -75,3 +76,20 @@ RedfishPlatformHostInterfaceNotification (
...
+                              memory for serial number.
+  @retval EFI_SUCCESS         Serial number is returned.
+  @retval Others              Failed to get the serial number
+**/
+EFI_STATUS
+RedfishPlatformHostInterfaceUsbSerialNumber (
+  OUT CHAR8  **SerialNumber
+  )
+{
+  return EFI_UNSUPPORTED;
```
ghbaccount, line 94: Shouldn't be such stub added into
https://github.com/tianocore/edk2/tree/master/EmulatorPkg/Library/RedfishPlatformHostInterfaceLib
and
https://github.com/tianocore/edk2/tree/master/RedfishPkg/Library/PlatformHostInterfaceBmcUsbNicLib ?
At least these libraries are in the edk2 tree and now became out of sync with the header.
I guess RedfishPlatformHostInterfaceUsbSerialNumber() implementation is absent for PlatformHostInterfaceBmcUsbNicLib, it's not a secret code how to get USB device serial number from its descriptor. I think it should be added too.
Exactly same error I'm observing while running `build -p RedfishPkg/RedfishPkg.dsc -t GCC5 -a X64`

```
edk2/RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c:90: undefined reference to `RedfishPlatformHostInterfaceUSBSerialNumber'
collect2: error: ld returned 1 exit status
make: *** [edk2/Build/RedfishPkg/DEBUG_GCC5/X64/RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe/DEBUG/RedfishHostInterfaceDxe.dll] Error 1
```

Upd:
I found https://github.com/NVIDIA/edk2-nvidia/blob/437d36639b30b866490ca203e2037932ffff11c3/Silicon/NVIDIA/Library/RedfishPlatformHostInterfaceOemLib/RedfishPlatformHostInterfaceOemLib.c#L677

and see it is based on OEM IPMI command.

Ok, since we do not have EFI_USB_IO_PROTOCOL here I think a stub in a libraries mentioned above is enough.

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
```diff
@@ -82,6 +88,31 @@ RedfishCreateSmbiosTable42 (
     DeviceDataLength = DeviceDescriptor->DeviceDescriptor.PciPcieDeviceV2.Length;
   } else {
     DeviceDataLength = DeviceDescriptor->DeviceDescriptor.UsbDeviceV2.Length;
+    SerialNumStrLen  = 0;
+    SerialNumber     = NULL;
```
ghbaccount, line 92: `FreePool (SerialNumber)` is being called unconditionally.
So for the case `DeviceType == REDFISH_HOST_INTERFACE_DEVICE_TYPE_PCI_PCIE_V2` SerialNumber is uninitialized variable.

Same for SerialNumStrLen and code branch https://github.com/tianocore/edk2/pull/10818/files#diff-95147722bb3106c1a95c9f0f0c342231e1cc71b612c9f71d5a9b88909be0c180R246

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
```diff
@@ -82,6 +88,31 @@ RedfishCreateSmbiosTable42 (
...
+    Status           = RedfishPlatformHostInterfaceUsbSerialNumber (&SerialNumber);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "%a: Fail to get USB virtual serial number, %r.", __func__, Status));
+      DeviceDescriptor->DeviceDescriptor.UsbDeviceV2.SerialNumberStr = 0;
+    } else {
+      if (SerialNumber != NULL) {
+        // Check if the string length exceeds MAX_UINT8 (255)
+        SerialNumberLength = AsciiStrLen (SerialNumber);
+        if (SerialNumberLength > MAX_UINT8) {
```
ghbaccount, line 101: Didn't you think about more generic solution?
It's not a low level or say platform specific code to reference to MAX_UINT8  values.
Could it be better to call AsciiStrnLenS (SerialNumber, SOME_MAX_NUMBER_CONSTANT) ?
It's not clear why 255 is a maximum length.  You just add it to the end (currently) of some structure.
We can take some clear reasonable value.
Also, let's assume RedfishPlatformHostInterfaceUsbSerialNumber()is responsible for conversion serial number from  UTF16 to char *, so there are too much assumptions and limitations in this code.

Could we do not limit length of SerialNumber at all.  Let AllocateZeroPool() or  Smbios->Add()  to decide? 

Another small note: as a result EFI_BAD_BUFFER_SIZE  is returned from RedfishHostInterfaceDxeEntryPoint, i.e _ModuleEntryPoint. It could be that EFI_BAD_BUFFER_SIZE is not a good value for doing so... May be EFI_OUT_OF_RESOURCES is a more common code, we already returning above?

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
```diff
@@ -82,6 +88,31 @@ RedfishCreateSmbiosTable42 (
...
+          DEBUG ((
+            DEBUG_ERROR,
+            "%a: SerialNumber length (%d) exceeds maximum allowed length for UINT8 (%d)\n",
+            __func__,
+            AsciiStrLen (SerialNumber),
```
ghbaccount, line 106: SerialNumberLength is already calculated, why do we need to repeat AsciiStrLen (SerialNumber) ?

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
```diff
@@ -82,6 +88,31 @@ RedfishCreateSmbiosTable42 (
...
+            MAX_UINT8
+            ));
```
ghbaccount, line 108: SerialNumber should be released here.

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
thejaputta: Usually, USB serial numbers are of max 64 byte size. So, I thought UINT8_MAX would be a more conservative value. If you think, we shouldn't limit its size here, I will update the code to declare serial number length as UINTN and ignore its size checks.

thejaputta: Done, thanks.

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
thejaputta: Ack.

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
thejaputta: Done, thanks

RedfishPkg/Library/PlatformHostInterfaceLibNull/PlatformHostInterfaceLibNull.c
thejaputta: Agree, done. Thanks.

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
thejaputta: Make sense, done. Thanks.

RedfishPkg/Include/Library/RedfishHostInterfaceLib.h
```diff
@@ -74,4 +75,17 @@ RedfishPlatformHostInterfaceNotification (
   OUT EFI_GUID  **InformationReadinessGuid
   );
 
+/**
+  Get USB Virtual Serial Number.
```
changab, line 79: What does the USB virtual serial number mean? We can say Get USB Serial number here. Also, I think we don't have to specifically say "USB" in the function name, although only the USB Redfish Host interface has the serial number defined in the structure. We don't know if there is another type of Redfish host interface that also provides the serial number.

RedfishPkg/Include/Library/RedfishHostInterfaceLib.h
```diff
@@ -74,4 +75,17 @@ RedfishPlatformHostInterfaceNotification (
...
+/**
+  Get USB Virtual Serial Number.
+
+  @param[OUT] SerialNumber      Pointer to retrieve complete serial number.
+
+  @retval EFI_SUCCESS         Serial number is returned.
+  @retval Others              Failed to get the serial number.
+**/
+EFI_STATUS
+RedfishPlatformHostInterfaceUsbSerialNumber (
```
changab, line 87: What about RedfishPlatformHostInterfaceSerialNumber?

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
```diff
@@ -82,6 +89,16 @@ RedfishCreateSmbiosTable42 (
...
   } else {
     DeviceDataLength = DeviceDescriptor->DeviceDescriptor.UsbDeviceV2.Length;
+    Status           = RedfishPlatformHostInterfaceUsbSerialNumber (&SerialNumber);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "%a: Fail to get USB virtual serial number, %r.", __func__, Status));
+      DeviceDescriptor->DeviceDescriptor.UsbDeviceV2.SerialNumberStr = 0;
+    } else {
+      if (SerialNumber != NULL) {
+        SerialNumStrLen                                                = (UINTN)AsciiStrLen (SerialNumber);
+        DeviceDescriptor->DeviceDescriptor.UsbDeviceV2.SerialNumberStr = StringCount;
```
changab, line 99: We don't have to introduce a local variable if StringCount is only used here.

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
```diff
@@ -82,6 +89,16 @@ RedfishCreateSmbiosTable42 (
     DeviceDataLength = DeviceDescriptor->DeviceDescriptor.PciPcieDeviceV2.Length;
   } else {
     DeviceDataLength = DeviceDescriptor->DeviceDescriptor.UsbDeviceV2.Length;
+    Status           = RedfishPlatformHostInterfaceUsbSerialNumber (&SerialNumber);
```
changab, line 92: If we change the function name to RedfishplatformHostINterfaceSerialNumber, we can then move this to above *if (DeviceType == REDFISH_HOST_INTERFACE_DEVICE_TYPE_PCI_PCIE_V2)*. Then this code looks more generic to all types of Redfish Host interfaces.

RedfishPkg/Include/Library/RedfishHostInterfaceLib.h
thejaputta: Renamed the function.

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
thejaputta: Ack, done. thanks.

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
thejaputta: I see that this serial number is specific to the USB interface. For the PCIe interface, there is no serial number field in the PCI_OR_PCIE_INTERFACE_DEVICE_DESCRIPTOR_V2 structure. Therefore, I believe we do not need to move the function above.

RedfishPkg/Include/Library/RedfishHostInterfaceLib.h
thejaputta: Done, thanks.

RedfishPkg/RedfishHostInterfaceDxe/RedfishHostInterfaceDxe.c
changab: Thanks for renaming this to a generic function name.
The reason to move this function to the main code flow is because we now have a generic function to retrieve the serial number for all types of host interfaces,  we can return EFI_UNSUPPORTED in the PCI network host interface library. With this,  we can have a consistent implementation in the different host interface libraries.
@igorkulchytskyy  and @nicklela, what do you think?

nicklela: > Thanks for renaming this to a generic function name. The reason to move this function to the main code flow is because we now have a generic function to retrieve the serial number for all types of host interfaces, we can return EFI_UNSUPPORTED in the PCI network host interface library. With this, we can have a consistent implementation in the different host interface libraries. @igorkulchytskyy and @nicklela, what do you think?

Do we have other interface which requires serial number too? If no, why we want all other interface libraries to implement a function which is never supported?

changab: @nicklela The reason is to keep the library implementation the same for the different transport interfaces. I would withdraw this comment if you think this is unnecessary.

thejaputta: Thank you @changab!

ghbaccount:
@thejaputta , you are still trying to submit merge commit that can not pass PatchCheck.

I would suggest to squash all your commits into the single one, save its hash or save it as a text file via `git format-patch`
Then reset your changes, rebase branch and reapply saved commit via `git am` or `git cherrypick ${savedhash}` and force push your changes again.
Personally, I can not see new patches properly neither in this PR nor in your branch. Something strange happens and I'm getting "Unable to load page. Please reload page and try again"  while navigating this PR. 

I hope this help.

thejaputta:
> @thejaputta , you are still trying to submit merge commit that can not pass PatchCheck.
> 
> I would suggest to squash all your commits into the single one, save its hash or save it as a text file via `git format-patch` Then reset your changes, rebase branch and reapply saved commit via `git am` or `git cherrypick ${savedhash}` and force push your changes again. Personally, I can not see new patches properly neither in this PR nor in your branch. Something strange happens and I'm getting "Unable to load page. Please reload page and try again" while navigating this PR.
> 
> I hope this help.

@ghbaccount, thank you for the advice. Previously, I rebased the patches and force-pushed after running PatchCheck locally to ensure all CI checks passed. Today, I squashed all the commits and force-pushed the changes again. However, the PatchCheck test still fails.

ghbaccount:
@thejaputta  PatchCheck wants you to split the commit into components:
```
Checking git commit: d25c15d74f
RedfishPkg: Populate serial number in SMBIOS type 42 record
The commit message format passed all checks.
The code passed all checks.
 * EmulatorPkg/EmulatorPkg.dec
 * RedfishPkg/RedfishPkg.dec
```
I guess it expects that change in EmulatorPkg must be a single commit.
It seems to be some CI specific option, because the patch is passed check when running locally.

thejaputta:
> @thejaputta PatchCheck wants you to split the commit into components:
> 
> ```
> Checking git commit: d25c15d74f
> RedfishPkg: Populate serial number in SMBIOS type 42 record
> The commit message format passed all checks.
> The code passed all checks.
> The commit adds/modifies files in multiple packages:
>  * EmulatorPkg/EmulatorPkg.dec
>  * RedfishPkg/RedfishPkg.dec
> ```
> 
> I guess it expects that change in EmulatorPkg must be a single commit. It seems to be some CI specific option, because the patch is passed check when running locally.

Thank you. It might have been the issue. I have updated the merge request with 2 separate commits for each Pkg and this time PatchCheck passed.
[tianocore.PatchCheck](https://github.com/tianocore/edk2/pull/10818/checks?check_run_id=39812702634)Successful in 44s — Build #20250402.1 succeeded

changab (approved):

PR #10819, opened 2025-03-04
IntelFsp2Pkg: new module to run pre FSP SecCore to rebase FSP

LiuZhiguang001:
This new module is optional. Add it only if FSP need to be rebased in boot time. It will find and rebase the SecCore and PeiCore in current FSP binary before FSP SecCore runs.
Note this module only works for X64.


IntelFsp2Pkg/PreFspSec/PreFspSec.c
```diff
@@ -0,0 +1,111 @@
...
+  return Status;
+}
+
+/**
+  This function will patch the Sec Core and Pei Core in current FSP.
+**/
+VOID
+EFIAPI
+FspPatchItself (
+  )
```
AshrafAliS, line 57: VOID

IntelFsp2Pkg/PreFspSec/PreFspSec.c
```diff
@@ -0,0 +1,111 @@
...
+{
+  UINT64           FspBase;
+  UINT64           SecCoreImageBase;
+  UINT64           PeiCoreImageBase;
+  FSP_INFO_HEADER  *FspInfoHeader;
+  UINT64           Delta;
+  EFI_STATUS       Status;
+
+  FspBase       = AsmGetRuntimeFspBaseAddress ();
```
AshrafAliS, line 66: how about making FspBase as UINT32 which help avoiding multiple typecasting

PR #10822, opened 2025-03-04
ArmPkg/ArmTransferListLib: improve coding style compliance

leiflindholm:
The existing code was full of abbreviated function/variable names. Let's nip that in the bud before we add more code to this library.

Update the current only upstream user: ArmStandaloneMmCoreEntryPoint.

- [x] Breaking change?
  - Any downstream users of ArmTransferListLib will break (but that library was added
    quite recently.

## How This Was Tested

Build tested, renaming only.

## Integration Instructions

Change your Tl* function calls to TransferList*.


samimujawar (approved):

LeviYeoReum (approved):

leiflindholm:
@mergifyio requeue

PR #10829, opened 2025-03-06
MdeModulePkg/XhciDxe: Adjust out-of-range bInterval values

jackp780:
# Description

Some high/super-speed devices incorrectly report full-speed bInterval values in their interrupt endpoint descriptors which results in an assertion error due to being out of range of the spec-expected values.

Rather than asserting, try to adjust those assuming they were expressed in units of ms with an upper limit of 8ms.

Inspiration was taken from the Linux kernel which has long had similar workarounds in its USB core (links to git.kernel.org):
* [USB: Fix up full-speed bInterval values in high-speed interrupt descriptor](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=300871cd963e24a68aaa9b762f4a10403697d9be)  (2007)
* [USB: change bInterval default to 10 ms](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=08c5cd37480f59ea39682f4585d92269be6b1424) (Later refined in 2016)

An example device showing incorrect interval values is the Lenovo L32p-30 monitor, which includes an embedded USB hub with built-in peripherals. Attached is an output of the [UsbTreeView](https://www.uwe-sieber.de/usbtreeview_e.html) utility which shows all its descriptor information of the monitor's hub as well as its embedded devices. In particular it contains a "Realtek HID" device which connects as a high-speed USB peripheral and reports an interrupt endpoint descriptor with bInterval=0x16.
[lenovo_monitor_usbtreeview.txt](https://github.com/user-attachments/files/19115096/lenovo_monitor_usbtreeview.txt)

These changes were internally reviewed within Qualcomm by @leiflindholm who had suggested the additional refactoring.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested by connecting the host to the USB-C port of the aforementioned monitor (Lenovo L32p-30) device which was triggering the assertion error without this fix, and verified that the fix allowed for successful enumeration.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c
+++ b/MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c
@@ -6,6 +6,7 @@
 Copyright (c) 2011 - 2020, Intel Corporation. All rights reserved.<BR>
 Copyright (c) Microsoft Corporation.<BR>
 Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.<BR>
+Copyright (C) 2025 Qualcomm Technologies, Inc. All rights reserved.<BR>
 SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
@@ -2809,6 +2810,40 @@ XhcDisableSlotCmd64 (
 }
 
 /**
+  Get the bInterval from descriptor and calculate the value to be used to initialize
+  the interval field of the endpoint context.
+
+  Refer to XHCI 1.1 spec section 6.2.3.6, tables 61 & 65
+
+  @param EpType       The endpoint type (only USB_ENDPOINT_ISO or USB_ENDPOINT_INTERRUPT).
+  @param DeviceSpeed  The connected device's speed.
+  @param Interval     The interval value from the descriptor.
+
+  @retval The calculated interval value to be applied to the endpoint context.
+
+**/
+static UINT8
+CalculateInterval (
+  UINT8  EpType,
+  UINT8  DeviceSpeed,
+  UINT8  Interval
+  )
+{
+  if ((EpType == USB_ENDPOINT_ISO) && (DeviceSpeed == EFI_USB_SPEED_FULL)) {
+    ASSERT (Interval >= 1 && Interval <= 16);
+    Interval = Interval + 2;
+  } else if ((EpType == USB_ENDPOINT_INTERRUPT) && ((DeviceSpeed == EFI_USB_SPEED_FULL) || (DeviceSpeed == EFI_USB_SPEED_LOW))) {
+    ASSERT (Interval != 0);
+    Interval = (UINT8)HighBitSet32 ((UINT32)Interval) + 3;
+  } else if ((DeviceSpeed == EFI_USB_SPEED_HIGH) || (DeviceSpeed == EFI_USB_SPEED_SUPER)) {
+    ASSERT (Interval >= 1 && Interval <= 16);
+    Interval = Interval - 1;
+  }
+
+  return Interval;
+}
+
+/**
   Initialize endpoint context in input context.
 
   @param Xhc            The XHCI Instance.
@@ -2838,7 +2873,6 @@ XhcInitializeEndpointContext (
   UINT8                    Dci;
   UINT8                    MaxDci;
   EFI_PHYSICAL_ADDRESS     PhyAddr;
-  UINT8                    Interval;
   TRANSFER_RING            *EndpointTransferRing;
 
   MaxDci = 0;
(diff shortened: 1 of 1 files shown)


MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c
```diff
@@ -2836,7 +2836,26 @@ CalculateInterval (
...
   } else if ((DeviceSpeed == EFI_USB_SPEED_HIGH) || (DeviceSpeed == EFI_USB_SPEED_SUPER)) {
-    ASSERT (Interval >= 1 && Interval <= 16);
+    //
+    // Some devices incorrectly report full-speed bInterval values in
+    // their high/super-speed interrupt endpoint descriptors.  Try to
+    // adjust those assuming they were expressed in units of ms with an
+    // upper limit of 8ms.
+    //
+    if ((Interval < 1) || (Interval > 16)) {
+      UINT8 OriginalInterval = Interval;
```
jackp780, line 2846: sorry, will force-push again after uncrustifying

MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c
```diff
@@ -2836,7 +2836,28 @@ CalculateInterval (
...
   } else if ((DeviceSpeed == EFI_USB_SPEED_HIGH) || (DeviceSpeed == EFI_USB_SPEED_SUPER)) {
-    ASSERT (Interval >= 1 && Interval <= 16);
+    //
+    // Some devices incorrectly report full-speed bInterval values in
+    // their high/super-speed interrupt endpoint descriptors.  Try to
+    // adjust those assuming they were expressed in units of ms with an
+    // upper limit of 8ms.
+    //
+    if ((Interval < 1) || (Interval > 16)) {
+      UINT8  OriginalInterval = Interval;
```
ardbiesheuvel, line 2846: This declaration should be at function scope.

MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c
```diff
@@ -2836,7 +2836,28 @@ CalculateInterval (
...
+
+      Interval = (UINT8)HighBitSet32 ((UINT32)Interval*8);
+      if (Interval == 0) {
```
ardbiesheuvel, line 2849: How can this condition ever be `true`?

ardbiesheuvel (commented):
Maybe add some prose to the commit log explaining that `bInterval` is the polling rate requested by the device from the host?

MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c
jackp780: Ok, will move it up

MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c
jackp780: Good catch, thanks. This was meant to handle the case if OriginalInterval==0, but in that case HighBitSet32(0) would return -1. This check should instead be against -1.

jackp780: @ardbiesheuvel I did a little more math-checking and added back an ASSERT (Interval != 0) as a precondition. Thus the HighBitSet32() call should be guaranteed to always be called on non-zero values and hence removed the result check.

lgao4 (approved):

lgao4:
The change is good to me.

ardbiesheuvel (approved):

leiflindholm:
@mergify refresh

mdkinney:
@leiflindholm We made a Mergify config change end of last week that required one additional configuration action that was completed this morning.  That should be resolved now and I will monitor.  With this change, rebase operations should occur without having to monitor going forward.

leiflindholm:
> @leiflindholm We made a Mergify config change end of last week that required one additional configuration action that was completed this morning. That should be resolved now and I will monitor. With this change, rebase operations should occur without having to monitor going forward.

Thanks. I saw #10861 go through and figured I was good to requeue :)

ardbiesheuvel:
@jackp780 you will have to make your branch writable to the project's maintainer if you want us to merge this.

(I have always found this to be a rather dubious requirement, but it is how things work at the moment. Mergify merges the latest master into your branch to perform the CI checks, and this only works if your branch is writable for us)

leiflindholm:
> @jackp780 you will have to make your branch writable to the project's maintainer if you want us to merge this.
> 
> (I have always found this to be a rather dubious requirement, but it is how things work at the moment. Mergify merges the latest master into your branch to perform the CI checks, and this only works if your branch is writable for us)

It says "Maintainers are allowed to edit this pull request.", I think Mergify (or github) is having a bad day.

@mdkinney @makubacki

makubacki:
@jackp780, do you have branch permissions set on your fork?

makubacki:
@mdkinney, I believe the issue is that the bot account `tianocore-issues` is in the `edk-ii-maintainers` but that group has the GitHub `write` role not the `maintain` role.

jackp780:
> @jackp780 you will have to make your branch writable to the project's maintainer if you want us to merge this.
> @jackp780, do you have branch permissions set on your fork?

No, I currently don't have any permissions enabled. Which accounts do I need to grant access to?

makubacki:
> No, I currently don't have any permissions enabled. Which accounts do I need to grant access to?

You do not need to do anything further. I just wanted to confirm permission there are not more strict than the defaults.

PR #10833, opened 2025-03-07
DynamicTablesPkg: Add support for non-serial DBG2 devices

jbrasen:
# Description

This commit adds support for building non-serial DBG2 devices

1. Add new common namespace objects:
   - EArchCommonObjMemoryRangeDescriptor
   - EArchCommonObjDbg2DeviceInfo

2. Enhance DBG2 table generator:
   - Add support for non-serial DBG2 devices
   - Support multiple debug devices
   - Improve memory handling with proper allocation checks
   - Add better cleanup in error paths

3. Update configuration manager:
   - Add parsers for new common namespace objects

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Created config manager object for a PCIe NIC and validated the DBG2 node

## Integration Instructions
N/A


DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
```diff
@@ -266,33 +172,166 @@ FreeDbg2TableEx (
...
+  } else {
+    MemoryRanges      = MemoryRange;
+    MemoryRangesCount = 1;
+  }
+
+  Dbg2DeviceSize = (UINT16)(sizeof (EFI_ACPI_DBG2_DEBUG_DEVICE_INFORMATION_STRUCT) +
+                            ((sizeof (EFI_ACPI_6_3_GENERIC_ADDRESS_STRUCTURE) + sizeof (UINT32)) * (MemoryRangesCount)) +
+                            (sizeof (CHAR8) * DBG2_NAMESPACESTRING_FIELD_SIZE));
+
+  NewAcpiDbg2 = ReallocatePool (*AcpiDbg2Len, *AcpiDbg2Len + Dbg2DeviceSize, *AcpiDbg2);
```
pierregondois, line 288: I think it would be simpler to compute the overall DBG2 table size first and do one allocation rather than re-allocating the table N times.

It should be possible to save the MemoryRangeDescriptors in an array to avoid fetching them multiple times

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
```diff
@@ -373,29 +423,55 @@ BuildDbg2TableEx (
...
+    DEBUG ((
+      DEBUG_ERROR,
+      "ERROR: DBG2: No information found. Status = %r\n",
+      EFI_NOT_FOUND
+      ));
+    return EFI_NOT_FOUND;
+  }
+
+  if (SerialPortCount != 0) {
+    Status = ValidateSerialPortInfo (SerialPortInfo, SerialPortCount);
```
pierregondois, line 461: NIT: maybe there should be a similar function for non-serial ports. The PortType/SubType + BaseAddress might be populated incorrectly

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
```diff
@@ -266,33 +172,166 @@ FreeDbg2TableEx (
...
+
+  // Check string length before concatenation
+  if ((DeviceInfo->ObjectName[0] != '\0') &&
+      (AsciiStrLen (SB_SCOPE) + AsciiStrLen (DeviceInfo->ObjectName) >= DBG2_NAMESPACESTRING_FIELD_SIZE))
+  {
+    DEBUG ((DEBUG_ERROR, "ERROR: DBG2: Namespace string too long\n"));
+    return EFI_BUFFER_TOO_SMALL;
+  }
+
+  if (DeviceInfo->AddressResourceToken != CM_NULL_TOKEN) {
```
pierregondois, line 255: If for a non-serial device, DeviceInfo->AddressResourceToken is not populated, we fall in the `else` condition and `MemoryRanges = MemoryRange;`. So there might be a NULL pointer access.

Cf. the comment below, it might be easier to give to the function an array of MemoryRange descriptors for both serial/non serial cases. For non-serial cases, the array will only have one element.

This also goes in the direction of pre-computing the overall DBG2 table size and avoid re-allocations.

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
```diff
@@ -35,148 +39,31 @@
...
-      OFFSET_OF (DBG2_DEBUG_DEVICE_INFORMATION, AddressSize)          \
-    },                                                                \
-  /* EFI_ACPI_6_2_GENERIC_ADDRESS_STRUCTURE BaseAddressRegister */  \
-    ARM_GAS32 (UartBase),                                             \
-  /* UINT32  AddressSize */                                         \
-    UartAddrLen,                                                      \
-  /* UINT8   NameSpaceString[MAX_DBG2_NAME_LEN] */                  \
-    UartNameStr                                                       \
-  }
+#define UID_DBG_SPCR_SKIP  1
```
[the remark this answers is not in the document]
pierregondois: Actually what I am suggesting is incorrect, the `_UID` seems to have to be unique among all the devices having the same `_CID` or `_HID`. This is unrelated to the namespace location. So adding a layer in the AML namespace won't help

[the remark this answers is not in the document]
pierregondois: Hello Jeff,
it seems this issue is very similar to what Jeshua is describing at: https://github.com/tianocore/edk2/pull/10841
I.e. cross generators index generation. We will try to provide a library to do that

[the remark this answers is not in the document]
jbrasen: For this I think I am going to swap back to only supporting 1 serial port as I didn't realize we had a 3rd place where the devices are created. I'll reduce this to just a single one and we can look at adding that support in the future.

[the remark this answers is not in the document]
pierregondois: ok right, seems like a good plan

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
```diff
@@ -187,14 +74,32 @@ GET_OBJECT_LIST (
...
+    Memory Range Descriptor Array information from the Configuration Manager.
+*/
+GET_OBJECT_LIST (
+  EObjNameSpaceArchCommon,
+  EArchCommonObjMemoryRangeDescriptor,
+  CM_ARCH_COMMON_MEMORY_RANGE_DESCRIPTOR
+  );
+
 /** Initialize the DBG2 UART with the parameters obtained from
-    the Configuration Manager.
```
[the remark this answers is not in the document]
jbrasen: Should be fixed in update

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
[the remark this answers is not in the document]
jbrasen: New version only supports one serial port, should revisit when we have better way to manage names and UIDs

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
jbrasen: Added checks for NULL memory and size issues, did not check Port types as those could be added to the spec unless we want to require an update for this in the future

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
jbrasen: Should be fixed in update

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
```diff
@@ -404,15 +488,40 @@
       " Status = %r\n",
       Status
       ));
+    goto error_handler;
+  }
+
+  AcpiDbg2Len = sizeof (EFI_ACPI_DEBUG_PORT_2_DESCRIPTION_TABLE);
+  for (Index = 0; Index < Dbg2DeviceCount; Index++) {
+    AcpiDbg2Len += sizeof (EFI_ACPI_DBG2_DEBUG_DEVICE_INFORMATION_STRUCT) +
+                   (sizeof (EFI_ACPI_6_3_GENERIC_ADDRESS_STRUCTURE) + sizeof (UINT32)) * Dbg2DevicesMemoryRangeCount[Index] +
```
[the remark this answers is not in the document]
jbrasen: I don't understand this error as the allocation is done under an if (Dbg2DeviceCount != 0) on line 433 and if the allocation fails it goes to the error handler. Only way it should get to this point as NULL is if Dbg2DeviceCount is 0 and then the for loop shouldn't run anything.

jbrasen:
@pierregondois I think all the review feedback was updated. I also added a host based test for this.

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
```diff
@@ -266,33 +164,160 @@ FreeDbg2TableEx (
...
+  CurrentDbg2Device->NameSpaceStringOffset           = sizeof (EFI_ACPI_DBG2_DEBUG_DEVICE_INFORMATION_STRUCT);
+  CurrentDbg2Device->OemDataLength                   = 0;
+  CurrentDbg2Device->OemDataOffset                   = 0;
+  CurrentDbg2Device->PortType                        = DeviceInfo->PortType;
+  CurrentDbg2Device->PortSubtype                     = DeviceInfo->PortSubtype;
+  CurrentDbg2Device->BaseAddressRegisterOffset       = CurrentDbg2Device->NameSpaceStringOffset + DBG2_NAMESPACESTRING_FIELD_SIZE;
+  CurrentDbg2Device->AddressSizeOffset               = CurrentDbg2Device->BaseAddressRegisterOffset +
+                                                       ((UINT16)sizeof (EFI_ACPI_6_3_GENERIC_ADDRESS_STRUCTURE) * ((UINT8)MemoryRangesCount));
+  for (Index = 0; Index < MemoryRangesCount; Index++) {
+    BaseAddressRegister                    = (EFI_ACPI_6_3_GENERIC_ADDRESS_STRUCTURE *)((UINT8 *)CurrentDbg2Device + CurrentDbg2Device->BaseAddressRegisterOffset + (Index * sizeof (EFI_ACPI_6_3_GENERIC_ADDRESS_STRUCTURE)));
```
pierregondois, line 297: NIT: the start of the array could be computed out of the loop, then the new cell to populate is at:
```
BaseAddressRegister += sizeof (EFI_ACPI_6_3_GENERIC_ADDRESS_STRUCTURE); 
```

Same comment for the `AddressSize`

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
```diff
@@ -266,33 +164,160 @@ FreeDbg2TableEx (
...
+    BaseAddressRegister->RegisterBitWidth  = 32;
+    BaseAddressRegister->RegisterBitOffset = 0;
+    BaseAddressRegister->AccessSize        = DeviceInfo->AccessSize;
+    BaseAddressRegister->Address           = MemoryRanges[Index].BaseAddress;
+    *AddressSize                           = (UINT32)(MemoryRanges[Index].Length);
+  }
+
+  if (DeviceInfo->ObjectName[0] == '\0') {
+    // If device string is empty then use "." as the name per the DBG2 specification.
+    AsciiSPrint ((CHAR8 *)CurrentDbg2Device + CurrentDbg2Device->NameSpaceStringOffset, DBG2_NAMESPACESTRING_FIELD_SIZE, ".");
```
pierregondois, line 309: Nit: is it possible to create a variable containing the pointer to `CurrentDbg2Device + CurrentDbg2Device->NameSpaceStringOffset` and add newlines to the function call ? This makes really long lines in the code (the CI doesn't seem to have an issue with that though)

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
```diff
@@ -356,161 +394,318 @@ BuildDbg2TableEx (
...
+               Dbg2DevicesMemoryRange[Index],
+               Dbg2DevicesMemoryRangeCount[Index]
+               );
     if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "ERROR: DBG2: Failed to populate device %u. Status = %r\n", Index, Status));
+      goto error_handler;
+    }
+  }
+
+  // Free these as they are not needed anymore
```
pierregondois, line 584: NIT: would it be possible to put these just before the final `return Status`, just to have all the memory allocation freeing bits at the same place ?

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
```diff
@@ -356,161 +394,318 @@ BuildDbg2TableEx (
...
+    Dbg2DevicesMemoryRange = NULL;
+  }
+
+  if (Dbg2DevicesMemoryRangeCount != NULL) {
+    FreePool (Dbg2DevicesMemoryRangeCount);
+    Dbg2DevicesMemoryRangeCount = NULL;
+  }
+
+  // Currently only one serial port is supported
+  if (SerialPortCount > 0) {
```
pierregondois, line 596: Would it be possible to create a specific function containing what is inside this `if` statement ? This would allow to make `BuildDbg2TableEx()` a slightly shorter ?

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
```diff
@@ -356,161 +394,318 @@ BuildDbg2TableEx (
...
+    Index                                     = 0; // Always use the first serial port
+    SerialMemoryRange.BaseAddress             = SerialPortInfo[Index].BaseAddress;
+    SerialMemoryRange.Length                  = SerialPortInfo[Index].BaseAddressLength;
```
pierregondois, line 599: @samimujawar Just a note that now that the static table used to have the `CM_ARCH_COMMON_SERIAL_PORT_INFO.BaseAddressLength` field hard coded to `PL011_UART_LENGTH`. This allowed some `CM_ARCH_COMMON_SERIAL_PORT_INFO` to have a valid address length while not providing any value.

Now that we rely on the CM_OBJECT, some ConfigurationManager might need to be updated.

Not sure what is the best solution here: either add an error condition to prevent from generating a DBG2 table when no value is provided, or use `0x1000` by default if no value is provided ?

pierregondois (commented):
Hello Jeff,
Thanks for adding the test file.

@samimujawar  as this is new to the package, would it be ok to check it the GoogleTest file if you're ok with it ? 

Regards,
Pierre

samimujawar:
> Hello Jeff, Thanks for adding the test file.
> 
> @samimujawar as this is new to the package, would it be ok to check it the GoogleTest file if you're ok with it ?
> 
> Regards, Pierre

@jbrasen Thanks a lot for introducing these tests. This is indeed a good initiative. 
Unfortunately, I have not looked into detail on how these tests are supposed to be designed/created. 
If you can add a section to the DynamicTablesPkg/Readme.md on how these tests can be built & run, it would be of great help for us to explore further. 
For now I think we could go ahead with your implementation, but we would be keen to build on this further and improve the testing for this package.

DynamicTablesPkg/Test/DynamicTablesPkgHostTest.dsc
```diff
@@ -0,0 +1,34 @@
...
+#
+##
+
+[Defines]
+  PLATFORM_NAME           = DynamicTablesPkgHostTest
+  PLATFORM_GUID           = 47c66bc0-e9ad-4d83-ae31-945eb128beee
+  PLATFORM_VERSION        = 0.1
+  DSC_SPECIFICATION       = 0x00010005
+  OUTPUT_DIRECTORY        = Build/DynamicTablesPkg/HostTest
+  SUPPORTED_ARCHITECTURES = IA32|X64
```
abdattar, line 17: is it not supported on AARCH64?

DynamicTablesPkg/Library/Common/TableHelperLib/ConfigurationManagerObjectParser.c
```diff
@@ -757,6 +757,23 @@ STATIC CONST CM_OBJ_PARSER  CmArchCommonStaInfoParser[] = {
   { "DeviceStatus", 4, "0x%x", NULL }
 };
 
+/** A parser for EArchCommonObjMemoryRangeDescriptor.
+*/
+STATIC CONST CM_OBJ_PARSER  CmArchCommonObjMemoryRangeDescriptor[] = {
+  { "BaseAddress", 8, "0x%llx", NULL },
+  { "Length",      8, "0x%llx", NULL },
```
abdattar, line 764: sizeof (UINT64) is much better than '8'.

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/GoogleTest/Dbg2GeneratorGoogleTest.inf
```diff
@@ -0,0 +1,50 @@
+## @file
+#  Google Test application for DBG2 Generator.
+#
+#  Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved. <BR>
+#  Copyright (c) Microsoft Corporation.
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+[Defines]
+  INF_VERSION                    = 0x00010005
```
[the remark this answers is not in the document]
abdattar: decimal number is much better to read. like 1.29

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
```diff
@@ -266,33 +164,160 @@ FreeDbg2TableEx (
...
+
+  Dbg2DeviceSize    = (UINT16)TotalSize;
+  CurrentDbg2Device = *AcpiDbg2Device;
+
+  // Initialize device structure
+  CurrentDbg2Device->Revision                        = EFI_ACPI_DBG2_DEBUG_DEVICE_INFORMATION_STRUCT_REVISION;
+  CurrentDbg2Device->Length                          = Dbg2DeviceSize;
+  CurrentDbg2Device->NumberofGenericAddressRegisters = (UINT8)MemoryRangesCount;
+  CurrentDbg2Device->NameSpaceStringLength           = DBG2_NAMESPACESTRING_FIELD_SIZE;
+  CurrentDbg2Device->NameSpaceStringOffset           = sizeof (EFI_ACPI_DBG2_DEBUG_DEVICE_INFORMATION_STRUCT);
```
pierregondois, line 288: Hello Jeff,
I think there might be a mistake here, the NameSpaceStringOffset should point to the `NamespaceString` field location, i.e. after the last `AddressSize` field

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
```diff
@@ -266,33 +164,160 @@ FreeDbg2TableEx (
...
+  CurrentDbg2Device->OemDataLength                   = 0;
+  CurrentDbg2Device->OemDataOffset                   = 0;
+  CurrentDbg2Device->PortType                        = DeviceInfo->PortType;
+  CurrentDbg2Device->PortSubtype                     = DeviceInfo->PortSubtype;
+  CurrentDbg2Device->BaseAddressRegisterOffset       = CurrentDbg2Device->NameSpaceStringOffset + DBG2_NAMESPACESTRING_FIELD_SIZE;
```
pierregondois, line 293: similarly, the list of BaseAddressRegister fields should be before the namespace string

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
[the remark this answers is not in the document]
jbrasen: Ok, I think it is right now

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/AcpiDbg2Lib.inf
```diff
@@ -23,7 +23,7 @@
 [Sources.ARM, Sources.AARCH64]
   Arm/ArmDbg2Generator.c
 
-[Sources.IA32, Sources.X86]
+[Sources.IA32, Sources.X86, Sources.X64]
```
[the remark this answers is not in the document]
jbrasen: Split into another commit in this PR

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
```diff
@@ -266,33 +164,160 @@ FreeDbg2TableEx (
...
-      (TableList[1]->Signature !=
-       EFI_ACPI_6_3_SECONDARY_SYSTEM_DESCRIPTION_TABLE_SIGNATURE))
-  {
-    DEBUG ((DEBUG_ERROR, "ERROR: DBG2: Invalid SSDT table pointer.\n"));
-    return EFI_INVALID_PARAMETER;
+  if (TableCount != 0) {
+    FreePool (TableList[0]);
   }
 
-  // Only need to free the SSDT table at index 1. The DBG2 table is static.
```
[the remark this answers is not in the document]
jbrasen: Updated the comments. Entry 0 is freed and then we do a loop to free all the SSDT tables.

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/GoogleTest/Dbg2GeneratorGoogleTest.cpp
```diff
@@ -0,0 +1,1447 @@
...
+
+  void
+  SetupSerialPortInfo (
+    UINT32  Count
+    )
+  {
+    mSerialPortInfo.resize (Count);
+    for (UINT32 i = 0; i < Count; i++) {
+      CM_ARCH_COMMON_SERIAL_PORT_INFO  info = { 0 };
+      info.BaseAddress       = 0x1000 * (i + 1);
```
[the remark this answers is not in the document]
jbrasen: Done

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/GoogleTest/Dbg2GeneratorGoogleTest.cpp
```diff
@@ -0,0 +1,1447 @@
...
+
+// Test multiple serial ports where only first one is used
+TEST_F (Dbg2GeneratorTest, MultipleSerialPorts) {
+  // Create two serial ports
+  std::vector<CM_ARCH_COMMON_SERIAL_PORT_INFO>  serialPorts;
+
+  // First serial port
+  CM_ARCH_COMMON_SERIAL_PORT_INFO  port1 = { 0 };
+
+  port1.BaseAddress       = 0x1000;
```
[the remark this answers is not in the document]
jbrasen: Done

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/GoogleTest/Dbg2GeneratorGoogleTest.inf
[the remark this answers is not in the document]
jbrasen: Done

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/GoogleTest/Dbg2GeneratorGoogleTest.inf
```diff
@@ -0,0 +1,50 @@
...
+  BaseLib
+  BaseMemoryLib
+  DebugLib
+  MemoryAllocationLib
+  TableHelperLib
+
+[Protocols]
+  gEdkiiConfigurationManagerProtocolGuid  ## CONSUMES
+
+[BuildOptions]
```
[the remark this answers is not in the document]
jbrasen: I think the MSVC one is it is on other host based tests and I recall a request to add it on another test I wrote but don't recall the reason. Removed the GCC flag

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/GoogleTest/Dbg2GeneratorGoogleTest.inf
```diff
@@ -0,0 +1,50 @@
...
+  MSFT:*_*_*_CC_FLAGS = /EHsc
+  GCC:*_*_*_CC_FLAGS = -fno-exceptions
+
+[Guids]
+  gEfiAcpiTableGuid                        ## SOMETIMES_CONSUMES  ## SystemTable
+
+[Pcd]
```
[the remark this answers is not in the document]
jbrasen: Removed

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/GoogleTest/Dbg2GeneratorGoogleTest.inf
```diff
@@ -0,0 +1,50 @@
...
+  Dbg2GeneratorGoogleTest.cpp
+  ../../../../../Test/Mock/Library/GoogleTest/Protocol/MockConfigurationManagerProtocol.cpp
+
+[Packages]
+  EmbeddedPkg/EmbeddedPkg.dec
+  MdePkg/MdePkg.dec
+  UnitTestFrameworkPkg/UnitTestFrameworkPkg.dec
+  DynamicTablesPkg/DynamicTablesPkg.dec
+
+[LibraryClasses]
```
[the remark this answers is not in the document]
jbrasen: Removed

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
jbrasen: Done

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
jbrasen: Done

DynamicTablesPkg/Include/ArchCommonNameSpaceObjects.h
```diff
@@ -59,6 +60,8 @@ typedef enum ArchCommonObjectID {
   EArchCommonObjPssInfo,                        ///< 32 - P-State status (PSS) Info
   EArchCommonObjPpcInfo,                        ///< 33 - P-State control (PPC) Info
   EArchCommonObjStaInfo,                        ///< 34 - _STA (Device Status) Info
+  EArchCommonObjMemoryRangeDescriptor,          ///< 35 - Memory Range Descriptor
+  EArchCommonObjDbg2DeviceInfo,                 ///< 36 - Generic DBG2 Device Info
```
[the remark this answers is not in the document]
jbrasen: Done

DynamicTablesPkg/Test/DynamicTablesPkgHostTest.dsc
jbrasen: AARCH64 host based tests would actually try to init the serial port so I don't think we can run in this case right now

DynamicTablesPkg/Library/Common/TableHelperLib/ConfigurationManagerObjectParser.c
jbrasen: Done

DynamicTablesPkg/Library/Acpi/Common/AcpiDbg2Lib/Dbg2Generator.c
jbrasen: Fixed

DynamicTablesPkg/Test/DynamicTablesPkgHostTest.dsc
pierregondois: Also I think the infrastructure of the HostBasedTests is missing something to run on AARCH64 machines unfortunately

NicholasGraves:
> > Just one minor comment, otherwise lgtm. The serial port information generated for Juno/kvmtool didn't change. I didn't try to create non-serial ports but the code looked correct.
> > Just need to check with @samimujawar if this is ok to include GoogleTests
> 
> I think adding GoogleTests is good. We need to see how GoogleTests can be supported other generators as well.

I'm adding a DynamicTables generator for CEDT right now and I'm using the same pattern as this PR for gtests. Intercepting the table registration makes implementing the rest of the test mostly straightforward.

PR #10836, opened 2025-03-10
Move CpuDxe from ArmPkg to UefiCpuPkg

AjanZhong:
# Description

As continues work to this discussion thread:
https://edk2.groups.io/g/devel/topic/patch_v5_2_6/102725178

This PR moves CpuDxe from ArmPkg/Drivers/CpuDxe to UefiCpuPkg/CpuDxeArm.

UefiCpuPkg contains CpuDxe for I386/X64/LoongArch64/RISCV64 architectures, move CpuDxe for Arm/AArch64 architectures to UefiCpuPkg making CpuDxe for all architectures which supported by EDK2 are easy to maintain.

- [ ] Breaking change?
  - N/A
- [ ] Impacts security?
  - N/A
- [ ] Includes tests?
  - N/A

## How This Was Tested

Packages depends on ArmPkg/Drivers/CpuDxe.inf should pass.

## Integration Instructions

Need to create a new PR in edk2-platform for this PR.


ArmPkg/ArmPkg.dsc
```diff
@@ -119,7 +119,7 @@
   ArmPkg/Library/SemihostLib/SemihostLib.inf
   ArmPkg/Library/ArmExceptionLib/ArmExceptionLib.inf
 
-  ArmPkg/Drivers/CpuDxe/CpuDxe.inf
+  UefiCpuPkg/CpuDxeArm/CpuDxe.inf
```
os-d, line 122: This is odd to have ArmPkg build a component that lives in UefiCpuPkg. I would expect UefiCpuPkg.dsc to build all the components that live within its package. I understand that the library dependencies are still within ArmPkg, but it doesn't seem to me that there is much value in moving the driver if the libraries aren't correctly dealt with. Untangling the ArmPkg libraries is a monumental task, but it is also the right direction to go

MdePkg/MdePkg.dec
```diff
@@ -362,6 +362,10 @@
   #
   ArmMmuLib|Include/Library/ArmMmuLib.h
 
+  ##  @libraryclass  Provides a default exception handler.
+  #
+  DefaultExceptionHandlerLib|Include/Library/DefaultExceptionHandlerLib.h
```
os-d, line 367: This is gated under `[LibraryClasses.ARM, LibraryClasses.AARCH64]`, but still being in MdePkg.dec I would consider renaming this to be more specific or merge it with other existing libraries classes, if applicable

MdePkg/MdePkg.dec
```diff
@@ -935,6 +935,9 @@
   ## Include/Guid/Cper.h
   gEfiIa32X64ErrorTypeMsCheckGuid    = { 0x48AB7F57, 0xDC34, 0x4f6c, { 0xA7, 0xD3, 0xB0, 0xB5, 0xB0, 0xA7, 0x43, 0x14 }}
 
+[Guids.ARM, Guids.AARCH64]
+  gArmTokenSpaceGuid       = { 0xBB11ECFE, 0x820F, 0x4968, { 0xBB, 0xA6, 0xF7, 0x6A, 0xFE, 0x30, 0x25, 0x96 } }
```
os-d, line 939: This might be okay as an interim step, but I think only if we have a plan for removing gArmTokenSpaceGuid in the future

AjanZhong:
Thanks @os-d , let me try to address your comments one by one. And will update this PR once I get good process.

MdePkg/MdePkg.dec
leiflindholm: No, this is not okay even as an interim step.

AjanZhong: Any suggestion for Guids belongs to ArmPkg? There are plenty of Guids need to be moved or removed, for instance, ArmMmuLib depends on gArmMmuRelaceLiveTransaltaionEntryFuncGuid.

leiflindholm: A token space Guid defines a token space. The token space in question is ArmPkg's token space. It makes absolutely no sense to duplicate that token space definition in another package. If we're very lucky, we don't end up with conflicting Pcd definitions in that token space. But that's only luck.

So let's rewind.

Where is gArmMmuReplaceLiveTranslationEntryFuncGuid used?
It is used *only* in ArmMmuLib. Nothing else cares about that Pcd.
Hence, if the function moves to a different package, then the Pcd definition should move with it, to a token space defined by *that* package.

MdePkg/MdePkg.dec
AjanZhong: 3 commits has been created to address this commit:
1. [CpuExceptionHandlerLib: Move DumpImageAndCpuContent to library header file]
2. [ArmPkg: Merge DefaultExceptionHandlerLib with ArmExceptionLib]
3. [ArmPkg,UefiCpuPkg: Move ArmExceptionLib from ArmPkg to UefiCpuPkg]

MdePkg/MdePkg.dec
AjanZhong: 2 commits updated to fix Guid issue:
1. [ArmPkg,MdePkg: Move PCD PcdRemapUnusedMemoryNx from ArmPkg to MdePkg]
2. [ArmPkg,MdePkg: Move PCD PcdVFPEnabled from ArmPkg to MdePkg]

ArmPkg/ArmPkg.dsc
AjanZhong: Decoupled UefiCpuPkg/CpuDxeArm/CpuDxe.inf with ArmPkg/ArmPkg.dsc in latest commit.

AjanZhong:
Now we have one issue need to be fixed: ERROR - Dependency Check: UefiCpuPkg/CpuDxeArm/CpuDxe.inf depends on pkg EmbeddedPkg/EmbeddedPkg.dec but pkg is not listed in AcceptableDependencies. 
This is caused by Protocol: gHardwareInterruptProtocolGuid is created by Arm CpuDxe. Let me refine this GUID in MdePkg to fix this issue.

MdePkg/MdePkg.dec
AjanZhong: 2 commits updated to move ArmMmuLib from ArmPkg to UefiCpuPkg:
1. [ArmPkg,MdePkg: Move ArmMmuLib.h from ArmPkg to MdePkg]
2. [ArmPkg,UefiCpuPkg: Move ArmMmuLib from ArmPkg to UefiCpuPkg]

AjanZhong:
@leiflindholm @os-d @ardbiesheuvel Hi all, PR is ready for review now. Please help review this PR when you are available. Thanks.

AjanZhong:
@leiflindholm @os-d @ardbiesheuvel Hi, would you mind kindly help review this PR when you are available. Thanks. :)

MdePkg/MdePkg.dec
```diff
@@ -927,6 +931,9 @@
   ## Include/Guid/Cper.h
   gEfiIa32X64ErrorTypeMsCheckGuid    = { 0x48AB7F57, 0xDC34, 0x4f6c, { 0xA7, 0xD3, 0xB0, 0xB5, 0xB0, 0xA7, 0x43, 0x14 }}
 
+[Guids.ARM, Guids.AARCH64]
+  gEfiArmMmuReplaceLiveTranslationEntryFuncGuid = { 0x309BD835, 0xEF7F, 0x834D, {0xA6, 0x37, 0x90, 0xB2, 0x6E, 0x6B, 0xCC, 0x63 } }
```
leiflindholm, line 935: @mdkinney @lgao4 what's your opinion of the Efi prefix here?
It matches surrounding definitions, far from all actually covered by the specs, but it's misleading.

ArmPkg/Drivers/CpuDxe/CpuDxe.inf
```diff
@@ -64,7 +64,7 @@
   gEfiVectorHandoffTableGuid
 
 [Pcd.common]
-  gArmTokenSpaceGuid.PcdVFPEnabled
+  gEfiMdePkgTokenSpaceGuid.PcdVFPEnabled
```
leiflindholm, line 64: I'm wondering whether we should rename this Pcd if moving it out into the Arm token space. I.e. PcdArmVFPEnabled *or* a more generic PcdFloatingPointEnabled.

MdePkg/MdePkg.dec
```diff
@@ -2009,6 +2009,9 @@
   ## Include/Protocol/ShellDynamicCommand.h
   gEfiShellDynamicCommandProtocolGuid  = { 0x3c7200e9, 0x005f, 0x4ea4, {0x87, 0xde, 0xa3, 0xdf, 0xac, 0x8a, 0x27, 0xc3 }}
 
+[Protocols.ARM, Protocols.AARCH64]
+  gEfiHardwareInterruptProtocolGuid =  { 0x38121D9B, 0x4C4F, 0x3D40, { 0x90, 0xED, 0x47, 0xCB, 0xE4, 0x18, 0xD2, 0xF9 }}
```
leiflindholm, line 2013: Same comment as before: does the Efi prefix make sense here?

@ardbiesheuvel do we think this protocol would be of any use to other architectures?
If so, it probably shouldn't be under .ARM/.AARCH64.
if not, an added Arm prefix would make sense when moving it out of ArmPkg.

ArmVirtPkg/ArmVirtCloudHv.dsc
```diff
@@ -244,7 +244,7 @@
   #
   # Architectural Protocols
   #
-  ArmPkg/Drivers/CpuDxe/CpuDxe.inf
+  UefiCpuPkg/CpuDxeArm/CpuDxe.inf
```
leiflindholm, line 247: I really don't like making a separate directory here.
Please merge it into the existing CpuDxe like LoongArch is.

MdePkg/MdePkg.dec
lgao4: EFI prefix means it is from UEFI spec. If it is not defined in UEFI spec, it should not use EFI prefix.

AjanZhong: Fixed

MdePkg/MdePkg.dec
AjanZhong: Commend addressed. Use 'Arm' Prefix.

ArmVirtPkg/ArmVirtCloudHv.dsc
AjanZhong: Moved to UefiCpuPkg/CpuDxe with other architectures.

ArmPkg/Drivers/CpuDxe/CpuDxe.inf
leiflindholm: I thought a bit more about this, and dug around. And I'm not sure any architecture other than 32-bit Arm can be used without floating-point. Indeed, I found some inconsistencies in the AArch64 implementation that means the Pcd could be deleted for that - and raised #10931.

So given that, I'm leaning back in the direction of preferring renaming the Pcd PcdArmVFPEnabled when moving it.

MdePkg/Include/Protocol/HardwareInterrupt.h
```diff
@@ -23,10 +23,10 @@
 //
 // Protocol GUID
 //
-// EAB39028-3D05-4316-AD0C-D64808DA3FF1
+// 9B1D1238-4F4C-403D-90ED-47CBE418D2F9
 
 #define EFI_HARDWARE_INTERRUPT_PROTOCOL_GGUID \
```
leiflindholm, line 28: The fact that there is a typo of GUID in this macro name makes me question whether this is needed at all. But if we are keeping it, why are we changing the value?

AjanZhong: oops, let me fix this typo later.
Since we need to update name of items to be moved, I changed value to keep same logic. If we need to keep GUI value unchanged, I will revert back to the original value.

AjanZhong:
Thanks all. Since this PR involves to much ARM related definition changes, as mentioned by @lgao4 "needs to evaluate its impact for the consumer code", it will takes quite a long time for this task.
This PR will be separated into several new Pull Requests. Thanks again. :)

MdeModulePkg/Include/Library/CpuExceptionHandlerLib.h
niruiyu, line 1: Firstly, the API should have "EFIAPI".
Secondly, I do not think it follows "Single Responsibility Principle" with the new lib API.
The new API contains two responsibilities: DumpImageInfo, DumpCpuContext.
Thirdly, I do not think CpuExceptionHandlerLib should own the responsibility of dumping PE coff image information.

The logic to dump PE coff image information is quite simple. All key logics are implemented/provided by PeCoffGetEntryPointLib. Can we still mark it as an internal/local function?

PR #10841, opened 2025-03-11
DynamicTablesPkg: container ProcId for PPTT/SSDT

jeshuasmith:
# Description

The current PPTT and SSDT code only allows for
AcpiProcessorId to be set/valid for leaf nodes,
but the ACPI spec allows (and in some cases
requires) it to be set for processor containers in the hierarchy as well.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

In use at NVIDIA since the EDK2 August 2024 stable release. Not tested on the tip of EDK2.

## Integration Instructions

N/A


jeshuasmith:
Note: only the Override path has been tested; we don't use the auto-assigned path.

PR #10843, opened 2025-03-11
PrmPkg: Handle Non-Existent MMIO and Align Signature to Spec

os-d:
# Description

This commit consists of two commits:

PrmPkg: PRM support for non-existent MMIO ranges
--
This is regarding PRM modules that are describing MMIO ranges.

When PrmConfigDxe is calling GetMemorySpaceDescriptor() with a memory range that is visible to the boot processor but has not been added to the memory map GetMemorySpaceDescriptor() will return EFI_SUCCESS and then return a memory descriptor indicating that the region is non-existent. This causes SetRuntimeMemoryRangeAttributes() to believe that the region has already been added to the memory map and will eventually cause an ASSERT.

This PR allows for SetRuntimeMemoryRangeAttributes() to treat a non-existent MMIO range the same as a range that triggered a EFI_NOT_FOUND error response from GetMemorySpaceDescriptor().

PrmPkg: Align Data Buffer Signature to Spec
--
edk2's PRM Data Buffer Signature is 'PRMD', however PRM spec 1.0 section 4.2.1 Static Data Buffer indicates that the signature should be 'PRMS'.

This commit aligns edk2's signature definition with the spec.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on a platform that had an MMIO range for a PRM module in non-existent memory.

## Integration Instructions

N/A.

### DIFF
--- a/PrmPkg/PrmConfigDxe/PrmConfigDxe.c
+++ b/PrmPkg/PrmConfigDxe/PrmConfigDxe.c
@@ -100,12 +100,17 @@ SetRuntimeMemoryRangeAttributes (
 
     Status2 = EFI_NOT_FOUND;
     Status  = gDS->GetMemorySpaceDescriptor (RuntimeMmioRanges->Range[Index].PhysicalBaseAddress, &Descriptor);
-    if (!EFI_ERROR (Status) &&
-        (
-         ((Descriptor.GcdMemoryType != EfiGcdMemoryTypeMemoryMappedIo) && (Descriptor.GcdMemoryType != EfiGcdMemoryTypeReserved)) ||
-         ((Descriptor.Length & EFI_PAGE_MASK) != 0)
-        )
-        )
+    if (!EFI_ERROR (Status) && (Descriptor.GcdMemoryType == EfiGcdMemoryTypeNonExistent)) {
+      // The MMIO range was found in the GCD, but is non-existent. We need to add this MMIO range to the GCD so that
+      // the PRM module can use it. As such, mark it as not found, as the behavior we want is the same as in that
+      // case.
+      Status = EFI_NOT_FOUND;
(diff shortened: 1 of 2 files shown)


makubacki (approved):

PrmPkg/PrmConfigDxe/PrmConfigDxe.c
```diff
@@ -100,12 +100,15 @@ SetRuntimeMemoryRangeAttributes (
...
     Status2 = EFI_NOT_FOUND;
     Status  = gDS->GetMemorySpaceDescriptor (RuntimeMmioRanges->Range[Index].PhysicalBaseAddress, &Descriptor);
-    if (!EFI_ERROR (Status) &&
-        (
-         ((Descriptor.GcdMemoryType != EfiGcdMemoryTypeMemoryMappedIo) && (Descriptor.GcdMemoryType != EfiGcdMemoryTypeReserved)) ||
-         ((Descriptor.Length & EFI_PAGE_MASK) != 0)
-        )
-        )
+    if (!EFI_ERROR (Status) && (Descriptor.GcdMemoryType == EfiGcdMemoryTypeNonExistent)) {
+      // The MMIO range is visible but has not been added to the memory space treat the region as not found.
```
ardbiesheuvel, line 104: Can we improve this comment please? What does 'visible' mean in this context?

And punctuation marks are free ...

os-d: I cherry-picked this from a downstream commit. I will update the comments

os-d: Updated.

ardbiesheuvel (approved):

Li-Aaron (approved):

PR #10853, opened 2025-03-12
OvmfPkg: Add opt-in X-PciMmioClassicWindow knob

MitchellAugustin:
# Description

Introduce a boolean fw_cfg knob "opt/ovmf/X-PciMmioClassicWindow", which, when TRUE, would bypass the PlatformDynamicMmioWindow functionality in [3] and instead use the "classic" MMIO window. When set, users' guest VMs would need to set the `pci=nocrs pci=realloc` kernel options in order to use passed-through GPUs whose real MMIO window sizes exceed the default without PDMW, but boots would be significantly shorter.

Context:
Due to an inefficiency in the way older host kernels manage pfnmaps for guest VM memory ranges[0], guests with large-BAR GPUs passed-through have a very long (multiple minutes) initialization time when the MMIO window advertised by OVMF is sufficiently sized for the passed-through BARs (i.e., when OVMF is behaving correctly). However, on older distro series such as Ubuntu Jammy, users have benefited from fast guest boot times when OVMF advertised an MMIO window that was too small to accommodate the full BAR, since this resulted in the long PCI initialization process being skipped (and retried later, if pci=realloc pci=nocrs were set).

While the root cause is being fully addressed in the upstream kernel[1], the proper solution relies on huge pfnmap support, which is a substantial series with many ABI changes that is unlikely to land in many LTS and legacy distro kernels, including those of Ubuntu Noble. As a result, the only kernel improvement supported on those kernels is this patch[2], which reduces the extra boot time by about half. Unfortunately, that boot time is still an average of 1-3 minutes longer per-VM-boot than what can be achieved when the host is running a version of OVMF without PlatformDynamicMmioWindow (PDMW) support (introduced in [3])

Since there is no way to force the use of the classic MMIO window size[4] in any version of OVMF after [3], and since we have a use case for such functionality on legacy distro kernels that would yield significant, recurring compute time savings across all impacted VMs, providing this opt-in functionality could be broadly valuable across distros.

[0]: https://lore.kernel.org/all/CAHTA-uYp07FgM6T1OZQKqAdSA5JrZo0ReNEyZgQZub4mDRrV5w@mail.gmail.com/
[1]: https://lore.kernel.org/all/20250205231728.2527186-1-alex.williamson@redhat.com/
[2]: https://lore.kernel.org/all/20250111210652.402845-1-alex.williamson@redhat.com/
[3]: https://github.com/tianocore/edk2/commit/ecb778d0ac62560aa172786ba19521f27bc3f650
[4]: https://edk2.groups.io/g/devel/topic/109651206?p=Created,,,20,1,0,0

https://bugs.launchpad.net/bugs/2101903

Signed-off-by: Mitchell Augustin <mitchell.augustin@canonical.com>

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

I have verified that this knob works as expected when set on DGX H100 (results in desired fast boot time, with GPUs still being usable as long as `pci=nocrs pci=realloc` are set in the guest, even on legacy kernels.) I also observed no regressions, and no change in behavior when X-PciMmioClassicWindow is absent or FALSE.

## Integration Instructions

Users should specify opt/ovmf/X-PciMmioClassicWindow=TRUE as an fw_cfg option in order to use this functionality.

Ex: If you use libvirt+qemu to manage your VMs, you can do the following:
```
<qemu:commandline>
    <qemu:arg value="-fw_cfg"/>
    <qemu:arg value="opt/ovmf/X-PciMmioClassicWindow,string=true"/>
  </qemu:commandline>
```

### DIFF
--- a/OvmfPkg/Library/PlatformInitLib/MemDetect.c
+++ b/OvmfPkg/Library/PlatformInitLib/MemDetect.c
@@ -1090,7 +1090,39 @@ PlatformAddressWidthInitialization (
   PlatformAddressWidthFromCpuid (PlatformInfoHob, TRUE);
   if (PlatformInfoHob->PhysMemAddressWidth != 0) {
     // physical address width is known
-    PlatformDynamicMmioWindow (PlatformInfoHob);
+    BOOLEAN               FwCfgPciMmioClassicWindow = FALSE;
+    EFI_STATUS            ClassicWindowStatus;
+
+    ClassicWindowStatus = QemuFwCfgParseBool (
+             "opt/ovmf/X-PciMmioClassicWindow",
+             &FwCfgPciMmioClassicWindow
+             );
+    switch (ClassicWindowStatus) {
+      case EFI_UNSUPPORTED:
+      case EFI_NOT_FOUND:
+        break;
+      case EFI_SUCCESS:
+        if (FwCfgPciMmioClassicWindow > 0) {
+          DEBUG ((DEBUG_INFO, "%a: using classic mmio window (requested in fw_cfg)\n", __func__));
+          break;
+        }
+
+      //
+      // fall through
+      //
+      default:
+        DEBUG ((
+          DEBUG_WARN,
+          "%a: error in X-PciMmioClassicWindow value from fw_cfg\n",
+          __func__
+          ));
+        break;
+    }
+
+    if (!FwCfgPciMmioClassicWindow)
+    {
+      PlatformDynamicMmioWindow (PlatformInfoHob);
+    }
     return;
   }
 


MitchellAugustin:
I'm trying to see why the code style checks failed in the "Ubuntu GCC5 PR" and "Windows VS2019 PR" tests, per [this guide](https://github.com/tianocore/tianocore.github.io/wiki/EDK-II-Code-Formatting#how-to-run-uncrustify), but the "Attachments" section in AzureDevOps for the failed tests shows no attachments. 
image
(although there are still other tests_ running, so I'm not sure if those need to complete before _any_ attachment would be added to ADO.)

luigix25:
@MitchellAugustin https://dev.azure.com/tianocore/edk2-ci/_build/results?buildId=176125&view=logs&j=ec42d809-3c3b-54a9-276c-e54a8b9aaee9&t=596e0656-4def-5804-b10b-1585519aa2e8

To get there: 
- click on the failing CI (Ubuntu GCC5 PR) 
- Click on "2 errors / 1 warnings" -
- pick the failing pipeline (Build_GCC5 TARGET_PLATFORMS")
- pick the failing part. There you can see the errors.

You can also run that check locally using `PatchCheck.py`. You can find it inside `BaseTools/Scripts`

kraxel:
There is no need to add a new config switch, we can also make ovmf use size specified on the command line, unconditionally.  Right now the code considers the size from fw_cfg the /minimum/ size, so you can force the window being larger but not the window being smaller.  Alowing both ways totally makes sense though.

Something like this:
```
@@ -37,6 +37,7 @@ typedef struct {
 
   UINT64               PcdPciMmio64Base;
   UINT64               PcdPciMmio64Size;
+  BOOLEAN              PcdPciMmio64Override;
   UINT32               PcdPciMmio32Base;
   UINT32               PcdPciMmio32Size;
   UINT64               PcdPciIoBase;
diff --git a/OvmfPkg/Library/PlatformInitLib/MemDetect.c b/OvmfPkg/Library/PlatformInitLib/MemDetect.c
index 1f987f2efd91..44b2e0d24f77 100644
--- a/OvmfPkg/Library/PlatformInitLib/MemDetect.c
+++ b/OvmfPkg/Library/PlatformInitLib/MemDetect.c
@@ -556,7 +556,8 @@ PlatformGetFirstNonAddress (
       break;
     case EFI_SUCCESS:
       if (FwCfgPciMmio64Mb <= 0x1000000) {
-        PlatformInfoHob->PcdPciMmio64Size = LShiftU64 (FwCfgPciMmio64Mb, 20);
+        PlatformInfoHob->PcdPciMmio64Size     = LShiftU64 (FwCfgPciMmio64Mb, 20);
+        PlatformInfoHob->PcdPciMmio64Override = TRUE;
         break;
       }
 
@@ -795,8 +796,10 @@ PlatformDynamicMmioWindow (
   AddrSpace = LShiftU64 (1, PlatformInfoHob->PhysMemAddressWidth);
   MmioSpace = LShiftU64 (1, PlatformInfoHob->PhysMemAddressWidth - 3);
 
-  if ((PlatformInfoHob->PcdPciMmio64Size < MmioSpace) &&
-      (PlatformInfoHob->PcdPciMmio64Base + MmioSpace < AddrSpace))
+  if (PlatformInfoHob->PcdPciMmio64Override) {
+    DEBUG ((DEBUG_INFO, "%a: using fwcfg override for mmio window\n", __func__));
+  } else if ((PlatformInfoHob->PcdPciMmio64Size < MmioSpace) &&
+             (PlatformInfoHob->PcdPciMmio64Base + MmioSpace < AddrSpace))
   {
     DEBUG ((DEBUG_INFO, "%a: using dynamic mmio window\n", __func__));
     DEBUG ((DEBUG_INFO, "%a:   Addr Space 0x%Lx (%Ld GB)\n", __func__, AddrSpace, RShiftU64 (AddrSpace, 30)));
```

MitchellAugustin:
Thanks, I'll test that approach out

MitchellAugustin:
OK, I've opened a new PR for the proposed revision: https://github.com/tianocore/edk2/pull/10856

PR #10855, opened 2025-03-13
BaseTools: Fix Debug Macro Checking to Include Scanning Files

linjaso1:
# Description

- In the commit 42a141800c0c26a09d2344e84a89ce4097a263ae there was a misuse of "is_dir" method.

- Treating it as an object rather than function call, which caused if-condition to always as "false".

- No files would be added to scanning list due to incorrect usage.

- This patch corrects the issue by properly using "is_dir()".

- [ ] Breaking change?
  - No, this change is bug-fix only.

- [ ] Impacts security?
  - No, this change do not include any security-related change.

- [X] Includes tests?
  - This change passed the EDK-II CI build without any issue.

## How This Was Tested

- Check the pytool plugin could be executed correctly.

## Integration Instructions

- N/A.

### DIFF
--- a/BaseTools/Plugin/DebugMacroCheck/DebugMacroCheck.py
+++ b/BaseTools/Plugin/DebugMacroCheck/DebugMacroCheck.py
@@ -429,7 +429,7 @@ def check_macros_in_directory(directory: PurePath,
 
         files = []
         for file in root_directory.rglob('*'):
-            if file.suffix in extensions and not file.is_dir:
+            if file.suffix in extensions and not file.is_dir():
                 files.append(Path(file))
 
             # Give an indicator progress is being made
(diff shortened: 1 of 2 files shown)


Javagedes (approved):

makubacki (approved):

pierregondois (approved):

Javagedes:
I'd prefer to get this in quickly; Since we have approvals from the necessary maintainers, I will place the push label once the CI passes. Please let me know if anyone disagrees.

linjaso1:
Patch-1 is used to fix the found issue before enabling back the checking.
Patch-2 is used to fix the real checking issue to make it run in normal mode.

I think it would be better to separate them.
Since patch-1 is still have a valid issue, even if we do not have Debug Macro Checking.

Javagedes:
Awaiting further comments from Ard before merging. Thanks for the follow up Ard!

ardbiesheuvel (approved):

PR #10862, opened 2025-03-14
MdePkg: IndustryStandard: Rename ACPI SPCR revision 4 structure

sarah-walker-arm:
# Description

The structure for the SPCR revision 4 table was originally named EFI_ACPI_4_0_SERIAL_PORT_CONSOLE_REDIRECTION_TABLE. This prefix suggests it is an ACPI 4.0 structure, which it is not. This could cause confusion with genuine ACPI 4.0 structures and defines (eg EFI_ACPI_4_0_SERIAL_PORT_CONSOLE_REDIRECTION_TABLE_SIGNATURE, which is unrelated to SPCR revision 4).

Rename the structure to EFI_ACPI_SERIAL_PORT_CONSOLE_REDIRECTION_TABLE_4.

- [x] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

edk2 builds for ARM FVP and boots. There are no users of the affected structure in the current edk2 tree.

## Integration Instructions

N/A


mdkinney:
Looks like this content was added 10 months ago and could be used with the current names by many consumers.

This should be marked as a "breaking change" and it would be good to know how widely these names have been used.  If you know they are not in use yet, then the impact would be minimal.

sarah-walker-arm:
This is not used upstream in either edk2 or edk2-platforms.

samimujawar:
@praveensankarn We would like to get the definition in the deader file fixed. Apparently, the upstream code edk2 & edk2-platforms does not utilise this structure name. 
Can you let us know if this is being used downstream, and plan to migrate the naming convention recommended in this patch, please?

praveensankarn:
@samimujawar, 
      Yes, this patch seems valid. And yes, we are using this structure in our downstream source, but we can consume the new patch once the next stable tag is released.

Thanks,
Praveen

samimujawar:
@praveensankarn Thank you for confirming.
@mdkinney, @lgao4 If there are no further comments,  can we proceed to merge this patch, please? I think downstream repos will need to eventually catchup. 

Regards,

Sami Mujawar

mdkinney (approved):

PR #10873, opened 2025-03-18
UefiCpuPkg/MpInitLib: Fix SNP AP creation when using known APIC IDs

tlendacky:
#Description

A typical initial AP boot up will choose a CpuNumber based on the ApIndex value that it gets back after a locked increment of the ApIndex value. The ApIndex to APIC ID relationship is random, which is not an issue when a broadcast INIT-SIPI is performed.

With SNP and a hypervisor that supports retrieval of the known APIC IDs, the broadcast INIT-SIPI method is replaced by waking each individual vCPU. In this situation, a specific VMSA is associated with a specific APIC ID. However, random assignment of an ApIndex can break this association. This isn't typically an issue, because the AP bring-up finishes with the AP issuing a HLT instruction, which is intercepted by the hypervisor and the AP won't run again until the next INIT-SIPI. However, when HLT isn't intercepted by the hypervisor (Qemu '-overcommit cpu-pm=on' parameter), then the HLT does not exit to the hypervisor. On the next INIT-SIPI, it can happen that a VMRUN is executed with a different VMSA address than was originally used, and if that VMSA is still in a VMRUN on another AP, then the executing VMRUN will fail, crashing the guest.

To fix this issue, add a CPU exchange info field, SevSnpKnownInitApicId, that indicates the APs are starting with an already known initial APIC ID and set the initial APIC ID and APIC ID in the CPU_INFO_IN_HOB HOB. During AP boot, the SevSnpKnownInitApicId field will result in the CpuNumber being set to the index with a matching APIC ID (similar to AP booting when the InitFlag != ApInitConfig).

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested using the Qemu with and without the '-overcommit cpu-pm=on' parameter and verifying correct synchronization of APIC ID to VMSA.

## Integration Instructions

N/A


kraxel (approved):

tlendacky:
@niruiyu and @jiaxinwu, any comments?

tlendacky:
@niruiyu and @jiaxinwu, please review.

tlendacky:
@niruiyu and @jiaxinwu, please review.

tlendacky:
Please, please review and merge this @niruiyu or @jiaxinwu. This is taking way too long for a fix that needs to get in.

tlendacky:
@niruiyu and @jiaxinwu, please review. I'm going to start pinging you every day.

PR #10874, opened 2025-03-19
Adding support for TPM over FF-A on ARM platforms

kuqin12:
# Description

This change adds the support of TPM over FF-A. It covers the TPM device library to allow TCG2 modules to communicate with the TPM service in the secure world through FF-A commands. In addition, it also supports the publication of TPM2 and the corresponding SSDT table that is responsible for supporting the physical presence interface through ASL methods during OS runtime.

- [ ] Breaking change?
- [x] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This change is tested on QEMU SBSA virtual platform and proprietary hardware platforms and booted on Windows platform.

## Integration Instructions

Add the following modules to the platform DSC file:
`SecurityPkg/Tcg/Tcg2StandaloneMmArm/Tcg2StandaloneMmArm.inf`
`SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.inf`

Link the following TPM library for TCG2 modules:
`Tpm2DeviceLib|ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2DeviceLibFfa.inf`


LeviYeoReum (commented):
Why do you make it with Tpm2DeviceLib?
Would It better to make it Tpm2InstanceLib so that Tpm2DeviceLibRouter can choise this FF-A CRB library?

kuqin12:
> Why do you make it with Tpm2DeviceLib? Would It better to make it Tpm2InstanceLib so that Tpm2DeviceLibRouter can choise this FF-A CRB library?

I started with device lib because it is able to be tested by directly linking to the driver without jumping through hoops.

But on the other hand, the router library is just picking among instance libraries using `PcdTpmInstanceGuid`, which only has TPM_NONE, TPM_1.2 and TPM_2. I do not see the benefits of add an extra layer of abstraction in this case, assuming all ARM platforms from this point and on are TPM2 ready.

Can you please let me know your usage?

SecurityPkg/Tcg/Tcg2AcpiFfa/Tpm2Ffa.asl
```diff
@@ -0,0 +1,265 @@
...
+  Scope (_SB_) {
+    Device(TPM0) {
+      Name (_HID, "NNNN0000")
+      Name (_CID, "MSFT0101")
+
+      //
+      // Operational region for TPM access
+      //
+      OperationRegion (TPMR, SystemMemory, FixedPcdGet64 (PcdTpmBaseAddress), FixedPcdGet32 (PcdTpmCrbRegionSize))
+      Field (TPMR, AnyAcc, NoLock, Preserve)
```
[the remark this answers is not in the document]
kuqin12: Agreed. This is an overlook of residue. Removed in the latest update.

SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.c
```diff
@@ -0,0 +1,746 @@
...
+
+  //
+  // Update TPM2 HID after measuring it to PCR
+  //
+  Status = UpdateHID (Table);
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  if (PcdGet32 (PcdTpm2CurrentIrqNum) != 0) {
```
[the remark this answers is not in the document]
kuqin12: Yes, you are right. This is another piece of residue that was left behind. I remove it and the function involved in the latest update.

SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.c
```diff
@@ -0,0 +1,746 @@
+/** @file
```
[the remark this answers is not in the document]
kuqin12: The TPM2 dynamic table does not support FFA type. But you are right, once the support is in place, it should not be mixed with the one from this driver.

SecurityPkg/Tcg/Tcg2AcpiFfa/Tpm2Ffa.asl
```diff
@@ -0,0 +1,265 @@
...
+      CreateDWordField (BUFF, 56, UCRQ) //   Physical Presence request operation to Get User Confirmation Status
+      CreateDWordField (BUFF, 60, IRQN) //   IRQ Number for _CRS
+      CreateByteField (BUFF, 64, SFRB) //   Is shortformed Pkglength for resource buffer
+
+      //
+      // FFA Direct Req2 Wrapper
+      //
+      OperationRegion(AFFH, FFixedHw, 4, 144)
+      Field(AFFH, BufferAcc, NoLock, Preserve) { AccessAs(BufferAcc, 0x1), FFAC, 1152 }
+      Method (FDR2, 0, Serialized) {
```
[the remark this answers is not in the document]
kuqin12: OS stakeholders just had a conversation about this specific piece. I will update the code accordingly.

ardbiesheuvel (approved):

SecurityPkg/Tcg/Tcg2AcpiFfa/Tpm2Ffa.asl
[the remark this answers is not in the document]
LeviYeoReum: When you see the spec -- 3.3.1 FFH usage in Operation Region Space for Secure Monitor Calls (SVC) and
Hypervisor Calls (HVC) in [ARM Functional Fixed Hardware Specification (FFH)](https://developer.arm.com/documentation/den0048/c/?lang=en]),

``` 
Offset is used to identify the functionality offered by this FFH address space. It must be set to one of the
following values:
  o 0x0 to indicate usage of 32-bit calling convention.
  o 0x1 to indicate usage of 64-bit calling convention.
  o All other values are reserved.
```

So, This function seems against the specification...

LeviYeoReum:
> > Why do you make it with Tpm2DeviceLib? Would It better to make it Tpm2InstanceLib so that Tpm2DeviceLibRouter can choise this FF-A CRB library?
> 
> I started with device lib because it is able to be tested by directly linking to the driver without jumping through hoops.
> 
> But on the other hand, the router library is just picking among instance libraries using `PcdTpmInstanceGuid`, which only has TPM_NONE, TPM_1.2 and TPM_2. I do not see the benefits of add an extra layer of abstraction in this case, assuming all ARM platforms from this point and on are TPM2 ready.
> 
> Can you please let me know your usage?


> `PcdTpmInstanceGuid`, which only has TPM_NONE, TPM_1.2 and TPM_2.

But we can makes it with Tpm Service UUID right?
I'm just asking. But I've implemented with this with Tpm2InstanceLib.
Actually there is no reason to hinder your patch for this Library.

kuqin12:
> > > Why do you make it with Tpm2DeviceLib? Would It better to make it Tpm2InstanceLib so that Tpm2DeviceLibRouter can choise this FF-A CRB library?
> > 
> > 
> > I started with device lib because it is able to be tested by directly linking to the driver without jumping through hoops.
> > But on the other hand, the router library is just picking among instance libraries using `PcdTpmInstanceGuid`, which only has TPM_NONE, TPM_1.2 and TPM_2. I do not see the benefits of add an extra layer of abstraction in this case, assuming all ARM platforms from this point and on are TPM2 ready.
> > Can you please let me know your usage?
> 
> > `PcdTpmInstanceGuid`, which only has TPM_NONE, TPM_1.2 and TPM_2.
> 
> But we can makes it with Tpm Service UUID right? I'm just asking. But I've implemented with this with Tpm2InstanceLib. Actually there is no reason to hinder your patch for this Library.

I can add one library as instance lib if so desired. It should be minor update.

SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.c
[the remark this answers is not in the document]
kuqin12: Yes, agreed. Will do.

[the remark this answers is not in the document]
kuqin12: Updated the driver header section to mention this. And fixed up the original comment to reflect the true state of the functionality.

kuqin12:
> > > > Why do you make it with Tpm2DeviceLib? Would It better to make it Tpm2InstanceLib so that Tpm2DeviceLibRouter can choise this FF-A CRB library?
> > > 
> > > 
> > > I started with device lib because it is able to be tested by directly linking to the driver without jumping through hoops.
> > > But on the other hand, the router library is just picking among instance libraries using `PcdTpmInstanceGuid`, which only has TPM_NONE, TPM_1.2 and TPM_2. I do not see the benefits of add an extra layer of abstraction in this case, assuming all ARM platforms from this point and on are TPM2 ready.
> > > Can you please let me know your usage?
> > 
> > 
> > > `PcdTpmInstanceGuid`, which only has TPM_NONE, TPM_1.2 and TPM_2.
> > 
> > 
> > But we can makes it with Tpm Service UUID right? I'm just asking. But I've implemented with this with Tpm2InstanceLib. Actually there is no reason to hinder your patch for this Library.
> 
> I can add one library as instance lib if so desired. It should be minor update.

Actually, I went ahead and added it in a new commit. Please let me know if that looks reasonable to you.

SecurityPkg/Tcg/Tcg2AcpiFfa/Tpm2Ffa.asl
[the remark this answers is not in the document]
kuqin12: Sorry for the delay. We are working in the background to make an agreement on the spec. I will update the patch once I test this change with updated Windows according to the new spec.

jyao1:
Hello
Is there any special reason to put **Tpm2DeviceLibFfa** to ArmPkg, while **Tcg2AcpiFfa** and **Tcg2StandaloneMmArm** to SecurityPkg?

Shouldn't they go together? Either ArmPkg or SecurityPkg?

leiflindholm:
> Hello Is there any special reason to put **Tpm2DeviceLibFfa** to ArmPkg, while **Tcg2AcpiFfa** and **Tcg2StandaloneMmArm** to SecurityPkg?
> 
> Shouldn't they go together? Either ArmPkg or SecurityPkg?

Agreed. And ArmPkg is not it.

kuqin12:
> > Hello Is there any special reason to put **Tpm2DeviceLibFfa** to ArmPkg, while **Tcg2AcpiFfa** and **Tcg2StandaloneMmArm** to SecurityPkg?
> > Shouldn't they go together? Either ArmPkg or SecurityPkg?
> 
> Agreed. And ArmPkg is not it.

@leiflindholm @jyao1 I do not disagree, either. But the Tpm2DeviceLib will depend on FFA lib directly, which seems inverted dependency. Any suggestions to avoid such entanglement?

leiflindholm:
> @leiflindholm @jyao1 I do not disagree, either. But the Tpm2DeviceLib will depend on FFA lib directly, which seems inverted dependency. Any suggestions to avoid such entanglement?

ArmFfaLib is an implementation of an industry standard specification, so belongs in MdePkg.

kuqin12:
> > @leiflindholm @jyao1 I do not disagree, either. But the Tpm2DeviceLib will depend on FFA lib directly, which seems inverted dependency. Any suggestions to avoid such entanglement?
> 
> ArmFfaLib is an implementation of an industry standard specification, so belongs in MdePkg.

That sounds reasonable. It also means that Smc/Svc libraries () will move to MdePkg as well. I can move them in the next push.

kuqin12:
> > > @leiflindholm @jyao1 I do not disagree, either. But the Tpm2DeviceLib will depend on FFA lib directly, which seems inverted dependency. Any suggestions to avoid such entanglement?
> > 
> > 
> > ArmFfaLib is an implementation of an industry standard specification, so belongs in MdePkg.
> 
> That sounds reasonable. It also means that Smc/Svc libraries () will move to MdePkg as well. I can move them in the next push.

I ended up creating a separate PR due to the scope of that change on its own:
https://github.com/tianocore/edk2/pull/10929

I will mark this as a draft until the PR above is merged. (And I assume the new FFH spec will be ready by then as well..)

ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2DeviceLibFfa.inf
```diff
@@ -0,0 +1,54 @@
...
+  ArmPkg/ArmPkg.dec
+  SecurityPkg/SecurityPkg.dec
+
+[LibraryClasses]
+  BaseLib
+  DebugLib
+  TimerLib
+  BaseMemoryLib
+  IoLib
+  Tpm2DebugLib
```
azizidomar-arm, line 45: Can you please help me find where this library is defined? I cannot find it in the upstream sources.
Keeping this change will most probably raise a build error.

ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
```diff
@@ -0,0 +1,409 @@
...
+    Status = EFI_INVALID_PARAMETER;
+    goto Exit;
+  }
+
+  if (mFfaTpm2PartitionId == MAX_UINT32) {
+    GetTpmServicePartitionId (&mFfaTpm2PartitionId);
+  }
+
+  ZeroMem (&FfaDirectReq2Args, sizeof (DIRECT_MSG_ARGS));
+  FfaDirectReq2Args.Arg0 = TPM2_FFA_GET_FEATURE_INFO;
```
azizidomar-arm, line 254: According to the TPM Service CRB Interface over FF-A specification [1], x3 should not be used.
[1] https://developer.arm.com/documentation/den0138/latest/

ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2Ptp.c
```diff
@@ -0,0 +1,420 @@
...
+  IN      PTP_CRB_REGISTERS_PTR  CrbReg
+  )
+{
+  EFI_STATUS  Status;
+
+  if (!Tpm2IsPtpPresence (CrbReg)) {
+    return EFI_NOT_FOUND;
+  }
+
+  MmioWrite32 ((UINTN)&CrbReg->LocalityControl, PTP_CRB_LOCALITY_CONTROL_REQUEST_ACCESS);
```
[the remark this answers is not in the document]
kuqin12: This is a thorny question :)

First off, I do not think we want to do relinquish in the place of request function, nor in the library overall. The library should allow the upper layer driver to acquire the locality and perform transactions as needed. The relinquish can occur when the driver or even the software/firmware entity is done, otherwise it will be unnecessary overhead.

Next, our option can be to perform it in Tcg2Dxe at ExitBootService, because UEFI is done with TPM transactions and ready for the OS to take over. However, we cannot do it naively for all architectures or all TPMs. Due to the fact that there is no UEFI code in EDK2 that relinquishes the locality, I doubt there is any existing UEFI firmware that relinquishes the locality before handing off the control to OS. And I am not sure whether the OS vendors would expect to see a relinquished locality. If not, they will need to make changes to acquire the localities before using it, and the updated firmware will have a broken TPM when booting the older OS.

I cannot seem to find that locality relinquish a requirement from TCG spec. Do you know where that is stated? If it is not a requirement in the spec, it might be hard to convince all stakeholders to make changes for relinquishment.

ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2InstanceLibFfa.c
```diff
@@ -0,0 +1,56 @@
...
+  if ((Status == EFI_SUCCESS) || (Status == EFI_UNSUPPORTED)) {
+    //
+    // Unsupported means platform policy does not need this instance enabled.
+    //
+    if (Status == EFI_SUCCESS) {
+      Status = InternalTpm2DeviceLibFfaConstructor ();
+      DumpPtpInfo ((VOID *)(UINTN)PcdGet64 (PcdTpmBaseAddress));
+    }
+
+    return EFI_SUCCESS;
```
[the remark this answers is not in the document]
kuqin12: Yes, will update. Thanks

ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2DeviceLibFfaBase.c
```diff
@@ -0,0 +1,83 @@
...
+  VOID
+  )
+{
+  EFI_STATUS  Status;
+
+  mActiveTpmInterfaceType = PcdGet8 (PcdActiveTpmInterfaceType);
+  mCRBIdleByPass          = 0xFF;
+
+  if (PcdGet64 (PcdTpmBaseAddress) == 0) {
+    Status = EFI_NOT_STARTED;
```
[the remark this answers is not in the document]
kuqin12: Agreed. This was written originally with the intention of supporting dynamic address setting. But that point was nullified when we move to use it as fixed PCD. I can update the return code to `EFI_NO_MAPPING` instead.

ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
```diff
@@ -0,0 +1,409 @@
...
+
+  Status = ArmFfaLibPartitionInfoGet (
+             &gEfiTpm2ServiceFfaGuid,
+             FFA_PART_INFO_FLAG_TYPE_DESC,
+             &Count,
+             &Size
+             );
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "Failed to get Tpm2 partition info. Status: %r\n", Status));
+    goto Exit;
```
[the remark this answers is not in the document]
kuqin12: Agreed, I will plumb through the error handling logic to do `ArmFfaLibRxRelease`.

ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2DeviceLibFfa.inf
kuqin12: This is a mistake. Will remove.

ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
kuqin12: This is for FFA Direct Req2 transactions, thus Arg0 in this case is for x4.

ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2InstanceLibFfa.c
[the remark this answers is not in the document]
kuqin12: Updated in the latest push.

ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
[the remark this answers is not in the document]
kuqin12: Added a new label and will jump there to release RX buffer, in case we will add more routine here.

ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2DeviceLibFfa.inf
kuqin12: Removed.

SecurityPkg/Tcg/Tcg2AcpiFfa/Tpm2Ffa.asl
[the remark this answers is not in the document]
kuqin12: The PR is updated to adhere to FFH spec v1.3 ALP0, which should be published shortly.

ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
azizidomar-arm: Yeah, my bad! Resolving.

azizidomar-arm: My bad! Feel free to resolve this conversation.

azizidomar-arm: BTW any reason to use **_FFA_MSG_SEND_DIRECT_REQ2_** instead of **_FFA_MSG_SEND_DIRECT_REQ_**? Is the TPM service usually supporting multiple services (each with its own UUID)?
Not all SPMCs support FFA_MSG_SEND_DIRECT_REQ2 for now, so maybe it would be better to use FFA_MSG_SEND_DIRECT_REQ, which is widely supported.

ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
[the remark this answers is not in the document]
azizidomar-arm: Thanks! Feel free to resolve this conversation (seems like I don't have enough rights to do so)

ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2DeviceLibFfaBase.c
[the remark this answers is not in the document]
azizidomar-arm: Thanks! Feel free to resolve this conversation.

ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2Ptp.c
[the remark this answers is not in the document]
azizidomar-arm: From Section **_6.5 TPM’s Software Interaction_** in [TCG PC Client Platform TPM Profile Specification for TPM 2.0](https://trustedcomputinggroup.org/wp-content/uploads/PC-Client-Specific-Platform-TPM-Profile-for-TPM-2p0-vp04p34_pubrev.pdf), following statements caught my attention:

> When the currently active locality is finished with the TPM, it must relinquish locality.

> Each Software agent, when it wishes to use the TPM, must request use of the locality it wishes to access. If the TPM  is idle, the first agent that sets this field will become the user. The TPM must set active locality to the locality that gains access to TPM. All other localities that have requested use of the TPM must poll on the TPM_ACCESS_x or the TPM_LOC_STS_x register to determine when they are granted access to the TPM

However, I agree that there's no need to introduce locality relinquishment if it's not part of the existing implementation.
Feel free to resolve this conversation.

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
```diff
@@ -0,0 +1,410 @@
...
+    GetTpmServicePartitionId (&mFfaTpm2PartitionId);
+  }
+
+  ZeroMem (&FfaDirectReq2Args, sizeof (DIRECT_MSG_ARGS));
+  FfaDirectReq2Args.Arg0 = TPM2_FFA_START;
+  FfaDirectReq2Args.Arg1 = (FuncQualifier & 0xFF);
+  FfaDirectReq2Args.Arg2 = (LocalityQualifier & 0xFF);
+
+  Status = ArmFfaLibMsgSendDirectReq2 (mFfaTpm2PartitionId, &gEfiTpm2ServiceFfaGuid, &FfaDirectReq2Args);
+  if (EFI_ERROR (Status)) {
```
azizidomar-arm, line 298: What if the TPM service is interrupted? Wouldn't it make more sense to rerun the partition instead of exiting?

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2Ptp.c
```diff
@@ -0,0 +1,420 @@
...
+  if (!Tpm2IsPtpPresence (CrbReg)) {
+    return EFI_NOT_FOUND;
+  }
+
+  MmioWrite32 ((UINTN)&CrbReg->LocalityControl, PTP_CRB_LOCALITY_CONTROL_REQUEST_ACCESS);
+  Status = Tpm2ServiceStart (
+             TPM2_FFA_START_FUNC_QUALIFIER_LOCALITY,
+             0
+             );
+  return Status;
```
azizidomar-arm, line 147: Shouldn't we check that activeLocality matches our requested locality, to confirm that the TPM has successfully granted it?

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
kuqin12: My original intention was that the secure partition better mark the ns-interrupt to be queued. But I think it does make sense to resume the interrupted direct_req2 calls. I will add that change in the next push.

ArmPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
kuqin12: The current EDK2 FFA implementation in general is very attached to FFA v1.2 and above at this point. i.e. stmm support direct_req2 to dispatch to non-root mmi handlers.

The TPM PPI interface also relies on stmm + direct_req2 and direct_req2 invocation from asl code. So I directly move to direct_req2 route. Because there was no support for direct_req anyway.

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2Ptp.c
kuqin12: Good call. Will update.

kuqin12: This is updated.

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
kuqin12: This is updated.

jyao1:
@kuqin12 ,  thanks for the update. I am OK with this patch.

Since you are expert on ARM-FFA part, would you mind to add yourself to be the reviewer and help me maintain the ARM-FFA related module?

If you agree, please add a section to https://github.com/tianocore/edk2/blob/master/Maintainers.txt, with following

```
SecurityPkg: ARM-FFA related modules
F: SecurityPkg/*Arm
F: SecurityPkg/*Ffa
R: name <email> [github-id]
```

SecurityPkg/Include/Guid/Tpm2ServiceFfa.h
```diff
@@ -0,0 +1,43 @@
...
+
+  Copyright (c), Microsoft Corporation.
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#ifndef TPM2_SERVICE_FFA_H_
+#define TPM2_SERVICE_FFA_H_
+
+#define EFI_TPM2_SERVICE_FFA_GUID \
```
jyao1, line 15: This GUID is defined in TPM over FF-A spec section 3.1

Then you don't need to add `EFI_` prefix. I recommend to use `TPM2_FFA_SERVICE_UUID`.

SecurityPkg/Include/Guid/Tpm2ServiceFfa.h
```diff
@@ -0,0 +1,43 @@
...
+#define TPM2_FFA_ERROR_ALREADY            0x8e000009
+#define TPM2_FFA_ERROR_DENIED             0x8e00000a
+#define TPM2_FFA_ERROR_NOMEM              0x8e00000b
+
+#define TPM_SERVICE_FEATURE_SUPPORT_NOTIFICATION  0xfea70000
+
+#define TPM2_FFA_START_FUNC_QUALIFIER_COMMAND   0x0
+#define TPM2_FFA_START_FUNC_QUALIFIER_LOCALITY  0x1
+
+extern EFI_GUID  gEfiTpm2ServiceFfaGuid;
```
jyao1, line 41: Ditto. You don't add "Efi" prefix.

SecurityPkg/Include/Guid/Tpm2ServiceFfa.h
kuqin12: Thanks, will update.

SecurityPkg/Include/Guid/Tpm2ServiceFfa.h
kuqin12: Thanks, will update.

kuqin12:
> @kuqin12 , thanks for the update. I am OK with this patch.
> 
> Since you are expert on ARM-FFA part, would you mind to add yourself to be the reviewer and help me maintain the ARM-FFA related module?
> 
> If you agree, please add a section to https://github.com/tianocore/edk2/blob/master/Maintainers.txt, with following
> 
> ```
> SecurityPkg: ARM-FFA related modules
> F: SecurityPkg/*Arm
> F: SecurityPkg/*Ffa
> R: name <email> [github-id]
> ```

Thanks, @jyao1. I can't claim to be an expert but I'd be willing to help review the incoming changes to these functional components :) I will add change in the next push.

SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.c
```diff
@@ -0,0 +1,468 @@
...
+  }
+
+  //
+  // Patch HID in ASL code before loading the SSDT.
+  //
+  for (DataPtr  = (UINT8 *)(Table + 1);
+       DataPtr <= (UINT8 *)((UINT8 *)Table + Table->Length - TPM_HID_PNP_SIZE);
+       DataPtr += 1)
+  {
+    if (AsciiStrCmp ((CHAR8 *)DataPtr, TPM_HID_TAG) == 0) {
```
samimujawar, line 221: Can we use AmlLib to parse the ACPI table and fixup these values, please?

samimujawar: Probably this could be done later as part of a separate PR.

SecurityPkg/Tcg/Tcg2AcpiFfa/Tpm2Ffa.asl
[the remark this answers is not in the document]
kuqin12: In that case, the stmm partition does not need to include the module that registers the handler, nor indicate the partition supports this service ID. The ASL call will end up as a framework error.

PPI is treated as part of the HLK test for Windows. So we expect it to always be present for Windows systems.

SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.c
kuqin12: Hmm, AmlLib is from DynamicsPkg, it will introduce tangled dependency. It will be less ideal to that regard. Maybe we can move AmlLib to somewhere else in a later PR?

SecurityPkg/Tcg/Tcg2StandaloneMmArm/Tcg2StandaloneMmArm.c
```diff
@@ -0,0 +1,171 @@
...
+#include <Library/Tcg2PhysicalPresenceLib.h>
+
+/**
+  This function checks if the required DTPM instance is TPM 2.0.
+
+  @retval TRUE  The required DTPM instance is equal to gEfiTpmDeviceInstanceTpm20DtpmGuid.
+  @retval FALSE The required DTPM instance is not equal to gEfiTpmDeviceInstanceTpm20DtpmGuid.
+**/
+BOOLEAN
+IsTpm20Dtpm (
```
rahul1-kumar, line 34: Just a dump question , Any reason why we have ARM specific Tcg2StandaloneMM, this should be agnostic to the processor ? 

We are currently using  TCG2 Standalone MM driver  ( Tcg2StandaloneMm ) for TCG defined features.

kuqin12: Actually a good question. That is route I started with, but it ended up with a lot of hooks we need to add to cover the difference between ARM and x86 platforms.

One example is that the ARM uses the `CommBuffer`, which is actually sent through MM communicate, as the NVS region instead of the region allocated during boot. Otherwise, we will need to create FFA memory protocol library, have the ACPI driver donate memory during boot, communicate the memory handle, SMM driver accepts the donation, then we will be able to use it. I tested that route and it works. But I think this separate driver is a cleaner solution and less prone to errors. (this entire driver is less than 200 lines)

SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.c
jyao1: I recommend to use different PR to resolve this.

jyao1 (approved):

SecurityPkg/Tcg/Tcg2AcpiFfa/Tpm2Ffa.asl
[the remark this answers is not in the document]
jyao1: I think this current patch is good enough.
If we have new requirement to use PCD control, feel free to file another issue and follow up later, please.

PR #10878, opened 2025-03-19
BaseTools/tools_def: Always link with -Wl,-z,notext for BFD/LLD

chewi:
# Description

Firstly, I had to drop some unused GCC IA32/X64 flag variables. These haven't been used since before 2d07607d8b1a0710ba7379f8ee6c11dae1, when UNIXGCC support was dropped. The recent change in 14cb48b0a053b44c5a6bcc89cbbbf86ac78c7820 to make linker warnings fatal was therefore ineffective for these architectures.

Commit a257988f590ba90dd8394dd6bc7014ae9d814a08 added `-Wl,-z,notext`, but only when linking for IA32/X64 with LLD.
    
BFD can also be configured to either warn or error when text relocations are detected. It does not check at all by default, but Gentoo Linux tells it to warn in its regular configuration and tells it to error in its hardened configuration.

At least the AARCH64 and IA32/X64 code does include text relocations. With linker warnings fatal, this now fails to build on Gentoo Linux. We should therefore always use `-Wl,-z,notext`.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

I have built (but not run) these changes for X64, AARCH64, RISCV64, and LOONGARCH64 with GCC/BFD.

## Integration Instructions

N/A


chewi:
CC @os-d

os-d:
This makes sense to me, but I'd like to ask @ardbiesheuvel to comment on whether expanding his `-Wl,-z,notext` is appropriate, as I am not a GCC linking expert, though it seems reasonable, since we convert the ELF images to PE/COFF.

chewi:
Thanks. I'm now wondering whether RISCV and LOONGARCH64 should be covered by `--Wl,--fatal-warnings` too. Let me know if you'd like me to add that.

chewi:
> Did you test the changes to the GCC48 and GCC49 macros with GCC 4.8 and GCC 4.9, respectively?

Admittedly not. I wasn't sure whether CI covered that.

> In any case, I would prefer to leave those alone - either they are unused, and so it doesn't matter, and if they are, we shouldn't break them.

Okay, I've reworked this a fair bit, adding a variable to reduce duplication. It now leaves GCC48/GCC49 alone and adds all the aforementioned flags to GCC5 for all architectures. It also adds `-Wl,-z,notext` to all CLANGDWARF architectures for consistency.

I've redone the same testing as before, plus CLANGDWARF for X64 and AARCH64.

ardbiesheuvel (approved):

chewi:
Please let's get this in now before things shift around again.

ardbiesheuvel:
@lgao4 @bexcran

lgao4 (approved):

PR #10880, opened 2025-03-20
ArmPlatformPkg/PeilessSec: get rid of _SetSVCMode

leiflindholm:
First spotted the _SetSVCMode symbol in the ModuleEntryPoint.S during
review of an unrelated PR. This is clearly bogus since there is no SVC
mode in AArch64, and there is indeed no actual code associated with the
symbol.

Then looked at the AArch32 version, which does update the CPSR to
- Set SVC mode
- Mask FIQ and IRQ exceptions
but this is the expected mode for entering edk2 anyway, so my *guess*
is this change is irrelevant for any current users.
So delete it and see if anyone notices.

- [x] Breaking change?
  - Potentially, if there are any 32-bit Arm users relying on the existing behaviour.

## How This Was Tested

Compile-tested only, we know it's a NOP in normal use on AArch32.


ardbiesheuvel (approved):

LeviYeoReum:
Sorry to late.

LeviYeoReum (approved):

samimujawar:
@LeviYeoReum Thanks for confirming.

@leiflindholm Levi has tested this patch on Juno and was able to successfully boot to the UEFI shell with AArch32 UEFI @EL1

samimujawar (approved):

leiflindholm:
> @LeviYeoReum Thanks for confirming.
> 
> @leiflindholm Levi has tested this patch on Juno and was able to successfully boot to the UEFI shell with AArch32 UEFI @el1
> 
> Apparently, there is no option in the TF-A Juno builds to select landing in the Normal world at AArch32 EL1. Levi has made some local changes to test this patch. We will need to check if the TF-A project would accept handing off to the normal world in AArch32 EL1 as an additional build option.

@samimujawar That feels like a (non-critical) shortcoming in TF-A, but in reality is probably mostly restricted to validation exercises like this?

Are you happy enough for this PR to go in anyway?

PR #10886, opened 2025-03-24
BaseTools: Dump library dependency chain on build failure

niruiyu:
When a module M depends on L1, which depends on L2, which depends on L3, the build fails when the library instance of L3 cannot be found according to the library class-instance mapping configuration specified in the DSC file.
When such failure happens, the build tool only prints that the instance of L3 required by module M cannot be found. But it does not tell how L3 is required by M.

The change enhances build tool to print the entire dependency chain when such failure happens.
With the change, the new error message will be as follows:

<dsc-path>(...): error 4000: Instance of library class [L3] is not found for module [M], [L3] is:
  consumed by <instance of L2>
    consumed by <instance of L1>

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions

### DIFF
--- a/BaseTools/Source/Python/Workspace/WorkspaceCommon.py
+++ b/BaseTools/Source/Python/Workspace/WorkspaceCommon.py
@@ -90,6 +90,16 @@ def GetDeclaredPcd(Platform, BuildDatabase, Arch, Target, Toolchain, additionalP
 def GetLibraryInstances(Module, Platform, BuildDatabase, Arch, Target, Toolchain):
     return GetModuleLibInstances(Module, Platform, BuildDatabase, Arch, Target, Toolchain,Platform.MetaFile,EdkLogger)
 
+def GenerateDependencyDump(ConsumedByList, M, Level, Visited):
+    if M in Visited:
+        return []
+    Visited.add(M)
+    Indentation = "\t" * Level
+    DependencyDump = [f"{Indentation}consumed by {M}"]
+    for m in ConsumedByList[M]:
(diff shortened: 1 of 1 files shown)


lgao4:
I agree this is a good idea to show the real consumed library instance. Could you give one example to demostrate this change?

niruiyu:
> I agree this is a good idea to show the real consumed library instance. Could you give one example to demostrate this change?

e.g.: Comment BaseMemoryLib in UefiCpuPkg.dsc as following:

```
[LibraryClasses]
  BaseLib|MdePkg/Library/BaseLib/BaseLib.inf
  # BaseMemoryLib|MdePkg/Library/BaseMemoryLib/BaseMemoryLib.inf
```

Invoke build:
```
build -p UefiCpuPkg/UefiCpuPkg.dsc -a X64
```


The error will be like the following:
```
/home/ray/work/edk2/UefiCpuPkg/UefiCpuPkg.dsc(...): error 4000: Instance of library class [BaseMemoryLib] is not found for module [/home/ray/work/edk2/UefiCpuPkg/CpuIo2Dxe/CpuIo2Dxe.inf], [BaseMemoryLib] is:
        consumed by /home/ray/work/edk2/MdePkg/Library/BaseLib/BaseLib.inf
                consumed by /home/ray/work/edk2/MdePkg/Library/BaseIoLibIntrinsic/BaseIoLibIntrinsic.inf
```

lgao4 (approved):

PR #6527, opened 2024-12-11
[v2] Add support for a SVSM vTPM

osteffenrh:
# Description

This is version 2 of #5770.

This series adds a DTpm driver to communicate with a virtual TPM running in the Secure VM Service Module (SVSM), enabling OVMF to do measured boot in SEV-SNP confidential VMs.

SEV-SNP provides a feature known as VM Privilege Level (VMPL), which allows for services to be run in the guest at different privilege levels. By running at VMPL0 (most priviledged VM level), the SVSM can be used to provide privileged services, e.g. virtual TPM, for the guest rather than trust such services from the hypervisor.

As described in the [SVSM specification](https://www.amd.com/content/dam/amd/en/documents/epyc-technical-docs/specifications/58019.pdf), guest components can call to the SVSM vTPM through the vTPM protocol (protocol-id 2).

The vTPM protocol follows the [Microsoft TPM Simulator interface (MSSIM)](https://github.com/microsoft/ms-tpm-20-ref/blob/main/TPMCmd/Simulator/include/prototypes/Simulator_fp.h) and it supports two services:

    SVSM_VTPM_QUERY (call-id 0): query MSSIM commands and vTPM features supported
    SVSM_VTPM_CMD (call-id 1): send a MSSIM command to be run by the vTPM and get the result

The SVSM vTPM protocol is also added in this series.

- [ ] Breaking change?
- [x] Impacts security?
  - This PR touches existing TPM driver code and implements new TPM related functionality.
- [ ] Includes tests?

## How This Was Tested

This was tested with the latest COCONUT-SVSM upstream code (commit 0b8feaf9), which provides vTPM service. The instructions to build and run a guest under a SVSM can be found in the [INSTALL.md](https://github.com/coconut-svsm/svsm/blob/main/Documentation/docs/installation/INSTALL.md).

The OVMF boot message log below shows that it was able to find the SVSM vTPM and also use it for measured boot.
```
[...]
Creating MpInformation2 HOB...
Loading PEIM F12F698A-E506-4A1B-B32E-6920E55DA1C4
Loading PEIM at 0x0007FB30000 EntryPoint=0x0007FB31399 TpmMmioSevDecryptPei.efi
TpmMmioSevDecryptPeimEntryPoint
TpmMmioSevDecryptPeimEntryPoint: mapping TPM MMIO address range unencrypted
Install PPI: 35C84FF2-7BFE-453D-845F-683A492CF7B7
[...]

PR #10891, opened 2025-03-25
SecurityPkg-Tpm2DeviceLibDTpm: Check SNP enabled prior to using AmdSvsmLib

jxu023:
# Description

This patch fixes boot failures observed on SEV and SEV-ES VMs using OvmfPkg broken by https://github.com/tianocore/edk2/pull/6527.

AmdSvsmLib currently doesn't check if SNP enabled, thus using AmdSvsmLib may errantly cause the caller code to believe SVSM is present. This can lead to boot failure on non-SNP enabled VMs.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Compiled OVMF target (but not tested) on edk2 upstream code following instructions https://github.com/tianocore/tianocore.github.io/wiki/Common-instructions

Compiled and tested on internal edk2 fork on SEV VM boot-up using internal hypervisor.

## Integration Instructions

N/A

### DIFF
--- a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2DeviceLibDTpmSvsm.inf
+++ b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2DeviceLibDTpmSvsm.inf
@@ -64,3 +64,4 @@
   gEfiSecurityPkgTokenSpaceGuid.PcdCRBIdleByPass             ## PRODUCES
   gEfiSecurityPkgTokenSpaceGuid.PcdSvsmVTpmPresence          ## PRODUCES
   gEfiSecurityPkgTokenSpaceGuid.PcdSvsmVTpmBufferPtr         ## PRODUCES
+  gEfiMdePkgTokenSpaceGuid.PcdConfidentialComputingGuestAttr ## CONSUMES
--- a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c
+++ b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c
@@ -15,6 +15,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Library/Tpm2DeviceLib.h>
 #include <Library/PcdLib.h>
 #include <Library/DebugLib.h>
+#include <ConfidentialComputingGuestAttr.h>
 #include "Tpm2Ptp.h"
 #include "Tpm2Svsm.h"
 
@@ -26,6 +27,80 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 /// @}
 
 static BOOLEAN  mUseSvsmVTpm = FALSE;
+
+/**
+  The function check if the specified Attr is set.
+
+  @param[in]  CurrentAttr   The current attribute.
+  @param[in]  Attr          The attribute to check.
+
+  @retval  TRUE      The specified Attr is set.
+  @retval  FALSE     The specified Attr is not set.
+
+**/
+STATIC
+BOOLEAN
+AmdMemEncryptionAttrCheck (
+  IN  UINT64                             CurrentAttr,
+  IN  CONFIDENTIAL_COMPUTING_GUEST_ATTR  Attr
+  )
+{
+  UINT64  CurrentLevel;
+
+  CurrentLevel = CurrentAttr & CCAttrTypeMask;
+
+  switch (Attr) {
+    case CCAttrAmdSev:
+      //
+      // SEV is automatically enabled if SEV-ES or SEV-SNP is active.
+      //
+      return CurrentLevel >= CCAttrAmdSev;
+    case CCAttrAmdSevEs:
+      //
+      // SEV-ES is automatically enabled if SEV-SNP is active.
+      //
+      return CurrentLevel >= CCAttrAmdSevEs;
+    case CCAttrAmdSevSnp:
+      return CurrentLevel == CCAttrAmdSevSnp;
+    case CCAttrFeatureAmdSevEsDebugVirtualization:
+      return !!(CurrentAttr & CCAttrFeatureAmdSevEsDebugVirtualization);
+    default:
+      return FALSE;
+  }
+}
+
+/**
+  Check if the specified confidential computing attribute is active.
+
+  @param[in]  Attr          The attribute to check.
+
+  @retval TRUE   The specified Attr is active.
+  @retval FALSE  The specified Attr is not active.
+
+**/
+STATIC
+BOOLEAN
+EFIAPI
+ConfidentialComputingGuestHas (
+  IN  CONFIDENTIAL_COMPUTING_GUEST_ATTR  Attr
+  )
+{
+  UINT64  CurrentAttr;
+
+  //
+  // Get the current CC attribute.
+  //
+  CurrentAttr = PcdGet64 (PcdConfidentialComputingGuestAttr);
+
+  //
+  // If attr is for the AMD group then call AMD specific checks.
+  //
+  if (((RShiftU64 (CurrentAttr, 8)) & 0xff) == 1) {
+    return AmdMemEncryptionAttrCheck (CurrentAttr, Attr);
+  }
+
+  return (CurrentAttr == Attr);
+}
 
 /**
  Initializes SVSM vTPM if present, or otherwise uses TCG PTP method.
@@ -44,6 +119,10 @@ EFIAPI
 TryUseSvsmVTpm (
   )
 {
+  if (!ConfidentialComputingGuestHas (CCAttrAmdSevSnp)) {
+    return FALSE;
+  }
+
   UINT8  SvsmVTpmPresence = (UINT8)PcdGet8 (PcdSvsmVTpmPresence);
 
   if (SvsmVTpmPresence == SVSM_VTPM_PRESENCE_UNKNOWN) {
--- a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2InstanceLibDTpmSvsm.inf
+++ b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2InstanceLibDTpmSvsm.inf
@@ -58,3 +58,4 @@
   gEfiSecurityPkgTokenSpaceGuid.PcdCRBIdleByPass           ## PRODUCES
   gEfiSecurityPkgTokenSpaceGuid.PcdSvsmVTpmPresence        ## PRODUCES
   gEfiSecurityPkgTokenSpaceGuid.PcdSvsmVTpmBufferPtr       ## PRODUCES
+  gEfiMdePkgTokenSpaceGuid.PcdConfidentialComputingGuestAttr ## CONSUMES


osteffenrh (approved):

osteffenrh:
LGTM!

Thanks for finding and fixing.
I think we can go with this fix for now.

```diff
diff --git a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c
index 3e44f6e12a35..613582d70883 100644
--- a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c
+++ b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c
@@ -47,8 +47,9 @@ TryUseSvsmVTpm (
 {
   UINT8  SvsmVTpmPresence = (UINT8)PcdGet8 (PcdSvsmVTpmPresence);
 
-  if (!MemEncryptSevSnpIsEnabled())
+  if (!MemEncryptSevSnpIsEnabled ()) {
     return FALSE;
+  }
 
   if (SvsmVTpmPresence == SVSM_VTPM_PRESENCE_UNKNOWN) {
     SvsmVTpmPresence = Tpm2SvsmQueryTpmSendCmd () ? SVSM_VTPM_PRESENT : SVSM_VTPM_ABSENT;

```
To make the CI happier.

SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2DeviceLibDTpmSvsm.inf
```diff
@@ -48,6 +48,7 @@
   MdePkg/MdePkg.dec
   SecurityPkg/SecurityPkg.dec
   UefiCpuPkg/UefiCpuPkg.dec
+  OvmfPkg/OvmfPkg.dec
```
kuqin12, line 51: This seems inverted dependency.

jxu023: Hmm, it does look circular. I guess it doesn't break compilation. I wonder: are we kind of already in a circular dependency situation w.r.t higher level "Pkg.dec" organization?

Reviewing our header files around AMD SVSM  and the SVSM e-vTPM support shows - MdePkg (cpuid and msr defs), UefiCpuPkg (contains header for AmdSvsmLib), OvmfPkg (contains code for non-null version of AmdSvsmLib, header for MemEncSevLib.h), and SecurityPkg (Tpm2DeviceLib.h - consumes AmdSvsmLib).

Rephrased as follows:
Some code in OvmfPkg consumes Tpm2DeviceLib.h from Security Pkg.
Some code in SecurityPkg consumes AmdSvsmLib.h from UefiCpuPkg and AmdSvsmLib implementation from OvmfPkg.

It seems like we kind of already have a circular dependency here between the higher level Pkgs, but not between their subcomponents. I think adding the header dependency here is a similar case.

Could we perhaps bundle fixing this into a future cleanup?

jxu023:
It seems like the SevEsWorkArea used to cache the SEV_STATUS MSR is re-used in MpInitLib for the SEV-ES jump table boot up code and becomes not valid after the MpInitLib PEIM runs. Is this correct? If that's the case then it seems like the SnpIsEnabled check in PeiMemEncryptSevLibInternal.c becomes wrong when it's called from the TPM discovery module which loads after MpInitLib according to the boot logs (if we're doing sev-es ap jump table style booting). DxeMemEncryptSevLib deals with this by using PcdConfidentialComputingGuestAttr instead.

```
BIOS: Loading PEIM at 0x000BFBBD000 EntryPoint=0x000BFBC63ED Loading PEIM at 0x000BFBBD000 EntryPoint=0x000BFBC63ED CpuMpPei.efi
...
...
...
BIOS: Loading PEIM B1517C78-F518-42E5-B270-F4B1F402E53C
BIOS: Loading PEIM at 0x000BFB7D000 EntryPoint=0x000BFB82C8C Loading PEIM at 0x000BFB7D000 EntryPoint=0x000BFB82C8C PvUefiPei.efi
BIOS: Install PPI: 2AB86EF5-ECB5-4134-B556-3854CA1FE1B4
BIOS: Loading PEIM F12F698A-E506-4A1B-B32E-6920E55DA1C4
BIOS: Loading PEIM at 0x000BFB77000 EntryPoint=0x000BFB7B701 Loading PEIM at 0x000BFB77000 EntryPoint=0x000BFB7B701 TpmMmioSevDecryptPei.efi
BIOS: TpmMmioSevDecryptPeimEntryPoint
BIOS: TpmMmioSevDecryptPeimEntryPoint: mapping TPM MMIO address range unencrypted
BIOS: Install PPI: 35C84FF2-7BFE-453D-845F-683A492CF7B7
BIOS: Loading PEIM BF7F2B0C-9F2F-4889-AB5C-12460022BE87
BIOS: Loading PEIM at 0x000BF7FB000 EntryPoint=0x000BF7FE284 Loading PEIM at 0x000BF7FB000 EntryPoint=0x000BF7FE284 Tcg2ConfigPei.efi
BIOS:  Found SVSM vTPM
BIOS: Tcg2ConfigPeimEntryPoint
BIOS: Tcg2ConfigPeimEntryPoint: TPM2 detected

```

I wonder if this means the initial fix here is still broken for SEV-ES VM boot.

tlendacky:
Depending on when this is running, you can do something like what MpInitLib does in UefiCpuPkg. It gets the SEV_STATUS MSR value from a PCD and then checks its value. See the ConfidentialComputingGuestHas() implementation in MpLib.c

tlendacky:
> BIOS: TpmMmioSevDecryptPeimEntryPoint
> BIOS: TpmMmioSevDecryptPeimEntryPoint: mapping TPM MMIO address range unencrypted

And this may need to be visited to ensure this doesn't occur for the SVSM vTPM. It's ok for SEV and SEV-ES.

SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2DeviceLibDTpmSvsm.inf
osteffenrh: I agree that it seems wrong to have something in SecurityPkg consume `OvmfPkg.dec`, but `Tpm2DeviceLibDTpmSvsm.inf` is intended only for use in Ovmf anyway.

kuqin12: Would it be better to move this instance to Ovmf in that case?

osteffenrh: That is hard to do, since it shares `.c` files with the regular lib.

tlendacky: Can you see if you can read the PcdConfidentialComputingGuestAttr PCD value and use that? See UefiCpuPkg/Library/MpInitLib/MpLib.c and how the ConfidentialComputingGuestHas() function is used to check for SNP enablement.

osteffenrh: Tried it, seems to work and is much more minimal.

osteffenrh: @jxu023 WDYT?

```diff
diff --git a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2DeviceLibDTpmSvsm.inf b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2DeviceLibDTpmSvsm.inf
index da48bd3c60b5..0efcd4fca7b9 100644
--- a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2DeviceLibDTpmSvsm.inf
+++ b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2DeviceLibDTpmSvsm.inf
@@ -64,3 +64,4 @@ [Pcd]
   gEfiSecurityPkgTokenSpaceGuid.PcdCRBIdleByPass             ## PRODUCES
   gEfiSecurityPkgTokenSpaceGuid.PcdSvsmVTpmPresence          ## PRODUCES
   gEfiSecurityPkgTokenSpaceGuid.PcdSvsmVTpmBufferPtr         ## PRODUCES
+  gEfiMdePkgTokenSpaceGuid.PcdConfidentialComputingGuestAttr ## CONSUMES
diff --git a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2InstanceLibDTpmSvsm.inf b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2InstanceLibDTpmSvsm.inf
index 4baf363c1137..3396030e68c1 100644
--- a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2InstanceLibDTpmSvsm.inf
+++ b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2InstanceLibDTpmSvsm.inf
@@ -53,8 +53,9 @@ [LibraryClasses]
   AmdSvsmLib
 
 [Pcd]
-  gEfiSecurityPkgTokenSpaceGuid.PcdTpmBaseAddress          ## CONSUMES
-  gEfiSecurityPkgTokenSpaceGuid.PcdActiveTpmInterfaceType  ## PRODUCES
-  gEfiSecurityPkgTokenSpaceGuid.PcdCRBIdleByPass           ## PRODUCES
-  gEfiSecurityPkgTokenSpaceGuid.PcdSvsmVTpmPresence        ## PRODUCES
-  gEfiSecurityPkgTokenSpaceGuid.PcdSvsmVTpmBufferPtr       ## PRODUCES
+  gEfiSecurityPkgTokenSpaceGuid.PcdTpmBaseAddress            ## CONSUMES
+  gEfiSecurityPkgTokenSpaceGuid.PcdActiveTpmInterfaceType    ## PRODUCES
+  gEfiSecurityPkgTokenSpaceGuid.PcdCRBIdleByPass             ## PRODUCES
+  gEfiSecurityPkgTokenSpaceGuid.PcdSvsmVTpmPresence          ## PRODUCES
+  gEfiSecurityPkgTokenSpaceGuid.PcdSvsmVTpmBufferPtr         ## PRODUCES
+  gEfiMdePkgTokenSpaceGuid.PcdConfidentialComputingGuestAttr ## CONSUMES
diff --git a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c
index 8a49fe936f12..7fd47c397fc2 100644
--- a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c
+++ b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c
@@ -27,6 +27,30 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 
 static BOOLEAN  mUseSvsmVTpm = FALSE;
 
+#define CCAttrAmdSevSnp  0x102
+#define CCAttrTypeMask   0x000000000000ffff
+
+BOOLEAN
+IsAmdSevSnpActive (
+  )
+{
+  UINT64  CurrentAttr;
+
+  //
+  // Get the current CC attribute.
+  //
+  CurrentAttr = PcdGet64 (PcdConfidentialComputingGuestAttr);
+
+  //
+  // If attr is for the AMD group then check for SNP
+  //
+  if (((RShiftU64 (CurrentAttr, 8)) & 0xff) == 1) {
+    return (CurrentAttr & CCAttrTypeMask) == CCAttrAmdSevSnp;
+  }
+
+  return FALSE;
+}
+
 /**
  Initializes SVSM vTPM if present, or otherwise uses TCG PTP method.
 
@@ -44,6 +68,10 @@ EFIAPI
 TryUseSvsmVTpm (
   )
 {
+  if (!IsAmdSevSnpActive ()) {
+    return FALSE;
+  }
+
   UINT8  SvsmVTpmPresence = (UINT8)PcdGet8 (PcdSvsmVTpmPresence);
 
   if (SvsmVTpmPresence == SVSM_VTPM_PRESENCE_UNKNOWN) {
```

tlendacky: > ```diff
> +#define CCAttrAmdSevSnp  0x102
> +#define CCAttrTypeMask   0x000000000000ffff
> ```

These values are already defined in MdePkg/Include/ConfidentialComputingGuestAttr.h

I suggest you do just what is done in MpLib.c. Then the "IsAmdSevSnpAcitve ()" call becomes "ConfidentialComputingGuestHas (CCAttrAmdSevSnp)"

tlendacky: And it should be moved down below the UINT8 variable declaration and then the SvsmVTpmPresence only assigned after that.

osteffenrh:
```diff
diff --git a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2DeviceLibDTpmSvsm.inf b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2DeviceLibDTpmSvsm.inf
index da48bd3c60b5..42ee16135a3b 100644
--- a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2DeviceLibDTpmSvsm.inf
+++ b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2DeviceLibDTpmSvsm.inf
@@ -57,6 +57,7 @@ [LibraryClasses]
   DebugLib
   PcdLib
   AmdSvsmLib
+  MpInitLib
 
 [Pcd]
   gEfiSecurityPkgTokenSpaceGuid.PcdTpmBaseAddress            ## CONSUMES
diff --git a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2InstanceLibDTpmSvsm.inf b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2InstanceLibDTpmSvsm.inf
index 4baf363c1137..c82575342c58 100644
--- a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2InstanceLibDTpmSvsm.inf
+++ b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2InstanceLibDTpmSvsm.inf
@@ -51,6 +51,7 @@ [LibraryClasses]
   DebugLib
   PcdLib
   AmdSvsmLib
+  MpInitLib
 
 [Pcd]
   gEfiSecurityPkgTokenSpaceGuid.PcdTpmBaseAddress          ## CONSUMES
diff --git a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c
index 8a49fe936f12..e7b3d4c3c2be 100644
--- a/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c
+++ b/SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c
@@ -15,6 +15,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Library/Tpm2DeviceLib.h>
 #include <Library/PcdLib.h>
 #include <Library/DebugLib.h>
+#include <Library/MpInitLib/MpLib.h>
 #include "Tpm2Ptp.h"
 #include "Tpm2Svsm.h"
 
@@ -46,6 +47,10 @@ TryUseSvsmVTpm (
 {
   UINT8  SvsmVTpmPresence = (UINT8)PcdGet8 (PcdSvsmVTpmPresence);
 
+  if (!ConfidentialComputingGuestHas (CCAttrAmdSevSnp)) {
+    return FALSE;
+  }
+
   if (SvsmVTpmPresence == SVSM_VTPM_PRESENCE_UNKNOWN) {
     SvsmVTpmPresence = Tpm2SvsmQueryTpmSendCmd () ? SVSM_VTPM_PRESENT : SVSM_VTPM_ABSENT;
     PcdSet8S (PcdSvsmVTpmPresence, SvsmVTpmPresence);
```

tlendacky: > ```diff
> @@ -57,6 +57,7 @@ [LibraryClasses]
>    DebugLib
>    PcdLib
>    AmdSvsmLib
> +  MpInitLib
> ```

No, you can't include MpInitLib. You'll have to copy the code.

At some point a ConfidentialCompute library should be created in the MdePkg or such, that can consolidate the code and remove the duplication.

jxu023:
Updated PR to use the CC guest attr PCD according to Oliver's diff in the review comment, and re-tested on SEV-ES and SVSM successfully with this patch on internal edk2 fork and ovmf build.

SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c
```diff
@@ -44,6 +66,10 @@ EFIAPI
 TryUseSvsmVTpm (
   )
 {
+  if (!IsAmdSevSnpActive ()) {
```
osteffenrh, line 69: `if (!ConfidentialComputingGuestHas (CCAttrAmdSevSnp)) {`

SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c
```diff
@@ -27,6 +28,27 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
...
+  //
+  // If attr is for the AMD group then check for SNP
+  //
+  if (((RShiftU64 (CurrentAttr, 8)) & 0xff) == 1) {
+    return (CurrentAttr & CCAttrTypeMask) == CCAttrAmdSevSnp;
+  }
+
+  return FALSE;
+}
+
```
osteffenrh, line 51: I think what Tom was suggesting is: Replace this function with a 1:1 copy of the one from MpInitLib.c:

```
/**
  Check if the specified confidential computing attribute is active.

  @param[in]  Attr          The attribute to check.

  @retval TRUE   The specified Attr is active.
  @retval FALSE  The specified Attr is not active.

**/
BOOLEAN
EFIAPI
ConfidentialComputingGuestHas (
  IN  CONFIDENTIAL_COMPUTING_GUEST_ATTR  Attr
  )
{
  UINT64  CurrentAttr;

  //
  // Get the current CC attribute.
  //
  CurrentAttr = PcdGet64 (PcdConfidentialComputingGuestAttr);

  //
  // If attr is for the AMD group then call AMD specific checks.
  //
  if (((RShiftU64 (CurrentAttr, 8)) & 0xff) == 1) {
    return AmdMemEncryptionAttrCheck (CurrentAttr, Attr);
  }

  return (CurrentAttr == Attr);
}
```
and then call that below.

osteffenrh: Also the suggested-by line in the commit message needs a space before `<`.

osteffenrh (changes requested):

SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c
jxu023: updated PR to replace with a 1:1 copy (but marked as static) and re-tested

SecurityPkg/Library/Tpm2DeviceLibDTpm/Tpm2PtpSvsmShim.c
jxu023: replaced

jxu023:
Updated to use 1:1 copy of ConfidentialComputingGuestHas() from MpInit.c to check if snp is enabled

osteffenrh (approved):

tlendacky (approved):

jyao1 (approved):

PR #10895, opened 2025-03-26
RedfishPkg/RedfishHttpDxe : Fix the incorrect length of the Basic Auth

changab:
Use AsciiStrLen function instead of AsciiStrSize to determine the length of Basic Auth string.

# Description
Use AsciiStrSize, which leads to the incorrect length of the Basic Auth string.
Use AsciiStrLen function instead of AsciiStrSize.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
 The change was tested on Redfish simulator.

## Integration Instructions
None


igorkulchytskyy (approved):
I was about to propose the same fixes))

nicklela (approved):
Thanks for fixing my bug 😁

PR #10828, opened 2025-03-06
BaseTools/GenFv: Ensure the minimum pad file size for the FV with VTF

philnoh2:
In case of the FV with VTF, the left size should be enough to add the minimum pad file size (EFI_FFS_FILE_HEADER, 0x18). It prevents the build error, "GenFv: ERROR 0006: invalid FFS file header checksum" caused by the pad file overwriting some header data in VTF. This includes these updates for CalculateFvSize() function.

1. If NumBlocks is not defined, ensure the minimum pad file size for the left size (if the pad file is required as VTF is not bottom aligned at end of block, insert EFI_FFS_FILE_HEADER to ensure the pad file size)
2. If NumBlocks is defined, report more clear error message (the required fv image size = 0x%x. the set fv image size = 0x%x. Free space left is not enough to add a pad file (0x18))
3. Remove MaxPadFileSize, which is reported when the taken size is same as the total size. It can not be the actual left size to add an FFS file. It causes confusion when referring to the build log (FV Space Information)

How to reproduce the case/issue on EDK2 codebase: 
1. Clone the TianoCore master codebase (e.g. 3/5 master).
2. Build any one of the OVMF's arch projects. (e.g. OvmfPkg/OvmfPkgX64.dsc).
3. Take the free space size value of SECFV (e.g. 0x12345 free from the build messages of FV Space Information).
```
FV Space Information
SECFV [34%Full] 212992 (0x34000) total, 74480 (0x122f0) used, 138512 (0x21d10) free
```
4. Add a line to OvmfPkg\ResetVector\Ia16\ResetVectorVtf0.asm to fill up that FV's free space up to "0x10" bytes left. (e.g. TIMES (0x12345-0x10) DB 0).
```
TIMES (0x21d10-0x10) DB 0
```
5. Rebuild the OVMF project. Then the build error is raised.
```
Generating SECFV FV
#### ['GenFv', '-a', 'C:\\Develop\\OvmfPkgX64.0305\\Build\\OvmfX64\\DEBUG_VS2019\\FV\\Ffs\\SECFV.inf', '-o', 'C:\\Develop\\OvmfPkgX64.0305\\Build\\OvmfX64\\DEBUG_VS2019\\FV\\SECFV.Fv', '-i', 'C:\\Develop\\OvmfPkgX64.0305\\Build\\OvmfX64\\DEBUG_VS2019\\FV\\SECFV.inf']
Return Value = 2
GenFv: ERROR 0006: invalid FFS file header checksum
[...]

PR #10903, opened 2025-03-31
BaseTools/build: Reduce special FV full to warning

mdkinney:
# Description

If an FV_SPARE_SPACE_THRESHOLD is enabled and an FV is 100% full with 0 bytes free, then this is likely a special FV that may have alignment requirements
for the FFS file for both the start address and the length and the FFS file consumes all the available FV space.

Reduce FV_SPARE_SPACE_THRESHOLD from an error to a warning if this FV 0 bytes free condition is
detected.

PR #10828 introduced the generation of these error conditions for an FV with large alignment requirements. The pad region before the aligned FFS file used to be counted as free space even though it could never be used due to the alignment requirements. There was actually no free space available. PR #10828 fixed the free space calculation to properly show it as 0 bytes free, and this change then caused build error when FV_SPARE_SPACE_THRESHOLD feature was enabled. The
reduction to a warning for this condition allows the build to complete with errors and also provides a
build log warning message for review.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Build breaks with an error message for special FV with FFS file with large start address and length alignment requirements when FV_SPARE_SPACE_THRESHOLD  is enabled. The FFS file consumes all the remaining space of the FV and this is the expected behavior and no error should be generated.

Disabling FV_SPARE_SPACE_THRESHOLD avoids the error.

Applied this code change with FV_SPARE_SPACE_THRESHOLD enabled and the build completes with any errors and generates a warning for the FV that is completely full.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Source/Python/build/build.py
+++ b/BaseTools/Source/Python/build/build.py
@@ -2431,9 +2431,18 @@ class Build():
                         if len(NameValue) == 2 and NameValue[0].strip() == 'EFI_FV_SPACE_SIZE':
                             FreeSizeValue = int(NameValue[1].strip(), 0)
                             if FreeSizeValue < Threshold:
-                                EdkLogger.error("build", FV_FREESIZE_ERROR,
-                                                '%s FV free space %d is not enough to meet with the required spare space %d set by -D FV_SPARE_SPACE_THRESHOLD option.' % (
-                                                    FvName, FreeSizeValue, Threshold))
+                                if FreeSizeValue == 0:
+                                    # A free size of 0 means the FV is exactly 100% full which usually indicates a special
+                                    # FV for a region that contains a fixed size image with special alignment requirements
+                                    # with potentiaily a fixed address. Log a warning for review, but do not generate an
+                                    # error.
+                                    EdkLogger.warn("build", FV_FREESIZE_ERROR,
+                                                    '%s FV free space %d is not enough to meet with the required spare space %d set by -D FV_SPARE_SPACE_THRESHOLD option.' % (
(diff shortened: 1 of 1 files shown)


mdkinney:
@philnoh2 Please review this change.  I have seen platform build that break from PR #10828.  The changes in PR #10828 look correct, but this additional change is needed to make sure special case FVs do not generate an error message that breaks the build when size threshold checks are enabled.

lgao4 (approved):

philnoh2:
> @philnoh2 Please review this change. I have seen platform build that break from PR #10828. The changes in PR #10828 look correct, but this additional change is needed to make sure special case FVs do not generate an error message that breaks the build when size threshold checks are enabled.

I have reviewed the change. It looks good. A review point could be whether user can notice the warning message. I expect that user could notice it when seeing FV Space Information at end of build process.

mdkinney:
> > @philnoh2 Please review this change. I have seen platform build that break from PR #10828. The changes in PR #10828 look correct, but this additional change is needed to make sure special case FVs do not generate an error message that breaks the build when size threshold checks are enabled.
> 
> I have reviewed the change. It looks good. A review point could be whether user can notice the warning message. I expect that user could notice it when seeing FV Space Information at end of build process.

Yes.  The warning message does show up at the end of the build. If the content does not fit in the FV, then the build will break with an error message.  If there is at least 1 byte of free space, then the threshold error message will break the build if the free space is less than the threshold. This warning message will only be shown if the FFS files fit and there are exactly 0 bytes remaining  which should be a special condition that can be reviewed.

PR #10915, opened 2025-04-02
Add platform dependent entries to HMAT

SophiaWang-Google:
# Description

 Add platform dependent entries to HMAT

- [ ] Breaking change?
  N/A
- [ ] Impacts security?
  N/A
- [ ] Includes tests?
  N/A

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/MdePkg/Include/IndustryStandard/Acpi64.h
+++ b/MdePkg/Include/IndustryStandard/Acpi64.h
@@ -2205,6 +2205,7 @@ typedef struct {
   UINT32                                                                          NumberOfTargetProximityDomains;
   UINT8                                                                           Reserved2[4];
   UINT64                                                                          EntryBaseUnit;
+  UINT32                                                                          LatBwInfo[];
 } EFI_ACPI_6_4_HMAT_STRUCTURE_SYSTEM_LOCALITY_LATENCY_AND_BANDWIDTH_INFO;
 
 ///
@@ -2231,6 +2232,7 @@ typedef struct {
   EFI_ACPI_6_4_HMAT_STRUCTURE_MEMORY_SIDE_CACHE_INFO_CACHE_ATTRIBUTES    CacheAttributes;
   UINT8                                                                  Reserved2[2];
   UINT16                                                                 NumberOfSmbiosHandles;
+  UINT16                                                                 SmbiosHandles[];
 } EFI_ACPI_6_4_HMAT_STRUCTURE_MEMORY_SIDE_CACHE_INFO;
 
 ///


mdkinney:
Commit message is missing description and Signed-off-by.

MdePkg/Include/IndustryStandard/Acpi64.h
```diff
@@ -2205,6 +2205,7 @@ typedef struct {
   UINT32                                                                          NumberOfTargetProximityDomains;
   UINT8                                                                           Reserved2[4];
   UINT64                                                                          EntryBaseUnit;
+  UINT32                                                                          LatBwInfo[];
```
mdkinney, line 2208: From the ACPI Spec, this is a complex structure.  Not just an array of UINT32 values.  The first set of value are 32-bit values for the Initiator, then 32-bit values for the Target.  Then Latency Bandwidth values that are all 16-bit values.  And since this is a set of variable length array.  This make using this flexible array member complex.

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
```diff
@@ -723,6 +727,59 @@ typedef struct CmArmEtInfo {
...
+/** A structure that describes the Memory Proximity Domain Attribute.
+
+    ID: EArmObjMemoryProxDomainAttrInfo
+*/
+
+typedef struct CmArmMemoryProxDomainAttrInfo {
+  UINT16 Type;
+  UINT16 Flags;
+  UINT32 ProcessorProximityDomain;
+  UINT32 MemoryProximityDomain;
```
pierregondois, line 749: Same comment, this should be a token referencing a `CM_ARCH_COMMON_MEMORY_AFFINITY_INFO` struct

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
```diff
@@ -723,6 +727,59 @@ typedef struct CmArmEtInfo {
...
+} CM_ARM_MEMORY_PROX_DOMAIN_ATTR_INFO;
+
+/** A structure that describes the Memory Latency Bandwidth Info.
+
+    ID: EArmObjMemoryLatBwInfo
+*/
+
+typedef struct CmArmMemoryLatBwInfo {
+  UINT16 Type;
```
pierregondois, line 758: I think you can remove the `Type` field from all the structures, this can be inferred from the structure itself, i.e. all the `struct CmArmMemoryLatBwInfo` will have the same value

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
```diff
@@ -723,6 +727,59 @@ typedef struct CmArmEtInfo {
...
+} CM_ARM_MEMORY_LAT_BW_INFO;
+
+/** A structure that describes the Memory Cache Info.
+
+    ID: EArmObjMemoryCacheInfo
+*/
+
+typedef struct CmArmMemoryCacheInfo {
+  UINT16 Type;
+  UINT32 MemoryProximityDomain;
```
pierregondois, line 777: Similar comment, this should be a token I think

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
```diff
@@ -723,6 +727,59 @@ typedef struct CmArmEtInfo {
...
+  UINT64 MemorySideCacheSize;
+  UINT32 CacheAttributes;
+  UINT16 NumSmbiosHandles;
+  UINT16 SmbiosHandles [8];
```
pierregondois, line 781: The max number of handles would be 8. I think this should be a Token referencing a CM_ARCH_COMMON_OBJ_REF. There is currently no structure describing smbios tables, but we should not limit the number of handles to 8 IMO.

MdePkg/Include/IndustryStandard/Acpi64.h
SophiaWang-Google: Actually in HmatGenerator.c (upload for review in this same PR)

I use LatBwInfo[] to store all the 
// UINT32 InitiatorProximityDomainList[];
// UINT32 TargetProximityDomainList[];
// UINT16 LatencyEntry[][];

Please see the code below:
  void *MemLatBwInfoLocal = (void*) MemLatBwInfo;

  MemLatBwInfoLocal += sizeof(EFI_ACPI_6_4_HMAT_STRUCTURE_SYSTEM_LOCALITY_LATENCY_AND_BANDWIDTH_INFO);

  CopyMem(MemLatBwInfoLocal, CmMemLatBwInfo->InitiatorProximityDomainList, sizeof(UINT32) * CmMemInitTargetInfo->NumInitiator);
  MemLatBwInfoLocal += sizeof(UINT32) * CmMemInitTargetInfo->NumInitiator;
  CopyMem(MemLatBwInfoLocal, CmMemLatBwInfo->TargetProximityDomainList, sizeof(UINT32) * CmMemInitTargetInfo->NumTarget);
  MemLatBwInfoLocal += sizeof(UINT32) * CmMemInitTargetInfo->NumTarget;
  CopyMem(MemLatBwInfoLocal, CmMemLatBwInfo->RelativeDistanceEntry, sizeof(UINT16) * CmMemInitTargetInfo->NumTarget * CmMemInitTargetInfo->NumInitiator);

I tried different method to implement this flexible size array:
1. If we define 
 UINT32 InitiatorProximityDomainList[];
 UINT32 TargetProximityDomainList[];
 UINT16 LatencyEntry[][];

inside EFI_ACPI_6_4_HMAT_STRUCTURE_SYSTEM_LOCALITY_LATENCY_AND_BANDWIDTH_INFO, we will get the compilation error:
../../../../MdePkg/Include/IndustryStandard/Acpi64.h:2203:10: error: flexible array member not at end of struct
2203 | UINT32 InitiatorProximityDomainList[];

2. If I use 3 pointers inside EFI_ACPI_6_4_HMAT_STRUCTURE_SYSTEM_LOCALITY_LATENCY_AND_BANDWIDTH_INFO
The pointers need memory to store, so won't work

So I add this flexible size array at the end of EFI_ACPI_6_4_HMAT_STRUCTURE_SYSTEM_LOCALITY_LATENCY_AND_BANDWIDTH_INFO, build is fine, and in the implementation, we can fill in the detailed contents in this LatBwInfo[].

MdePkg/Include/IndustryStandard/Acpi64.h
```diff
@@ -2231,6 +2232,7 @@ typedef struct {
   EFI_ACPI_6_4_HMAT_STRUCTURE_MEMORY_SIDE_CACHE_INFO_CACHE_ATTRIBUTES    CacheAttributes;
   UINT8                                                                  Reserved2[2];
   UINT16                                                                 NumberOfSmbiosHandles;
+  UINT16                                                                 SmbiosHandles[];
```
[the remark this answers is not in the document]
SophiaWang-Google: If we do not use flexible size array here, maybe we add those in implementation?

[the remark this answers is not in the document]
SophiaWang-Google: Actually for the implementation to use this array, please refer to  HmatGenerator.c (uploaded in this review) ln127

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
```diff
@@ -723,6 +727,59 @@ typedef struct CmArmEtInfo {
...
 
+/** A structure that describes the Memory Initiator and Target number.
+
+    ID: EArmObjMemoryInitTargetInfo
+*/
+
+typedef struct CmArmMemoryInitTargetInfo {
+  UINT32 NumInitiator;
+  UINT32 NumTarget;
+} CM_ARM_MEMORY_INIT_TARGET_INFO;
```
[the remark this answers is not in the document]
SophiaWang-Google: Indeed the number of initiators/targets might vary.

The current implementation is get the number of initiator and target from Configuration manager. 
Then dynamically allocate the memory for System Locality Latency and Bandwidth Information Structure.

Please refer to HmatGenerator.c ln369. Current implementation supports variable number of initiator and target. 
HmatGenerator.c ln268 get # of initiator and # of target from Configuration manager.

Please let me know whether this solution is acceptable.

MdePkg/Include/IndustryStandard/Acpi64.h
mdkinney: Adding flexible array member is a good idea.  Adding comments above the flexible array member to describe the contents of the flexible array member would be helpful.  Also, since the layout of the data within the flexible array member is complex, it could just be type UINT8 LatBwInfo[], and the code the processes this data and follow the rules described in the ACPI spec to get to the start of each region and access using the correct size.

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
SophiaWang-Google: Do you have any idea how to populate this number if we do not want to hard code?

MdePkg/Include/IndustryStandard/Acpi64.h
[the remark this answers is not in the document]
mdkinney: For this use case, the flexible array member is correct and the `UINT16` type is correct because the data is an array of handles with length specified by `NumberOfSmbiosHandles`.  To get the size of the entire record, a computation has to be done: sizeof(struct) + NumberOfSmbiosHandles * sizeof(UINT16)

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
```diff
@@ -723,6 +727,59 @@ typedef struct CmArmEtInfo {
...
+
+/** A structure that describes the Memory Proximity Domain Attribute.
+
+    ID: EArmObjMemoryProxDomainAttrInfo
+*/
+
+typedef struct CmArmMemoryProxDomainAttrInfo {
+  UINT16 Type;
+  UINT16 Flags;
+  UINT32 ProcessorProximityDomain;
```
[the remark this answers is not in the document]
SophiaWang-Google: Status = GetEArmObjMemoryProxDomainAttrInfo (
             CfgMgrProtocol,
             CM_NULL_TOKEN,
             &MemProxDomainAttrInfo,
             &MemProxDomainAttrCount
             );

I use MemProxDomainAttrCount to expand the table to NumOfInitiator * NumOfTarget tables, please see:

[028h 0040   2]               Structure Type : 0000 [Memory Proximity Domain Attributes]
[02Ah 0042   2]                     Reserved : 0000
[02Ch 0044   4]                       Length : 00000028
[030h 0048   2]        Flags (decoded below) : 0002
            Processor Proximity Domain Valid : 0
[032h 0050   2]                    Reserved1 : 0000
[034h 0052   4] Attached Initiator Proximity Domain : 00000000
[038h 0056   4]      Memory Proximity Domain : 00000000
[03Ch 0060   4]                    Reserved2 : 00000000
[040h 0064   8]                    Reserved3 : 0000000000000000
[048h 0072   8]                    Reserved4 : 0000000000000000

[050h 0080   2]               Structure Type : 0000 [Memory Proximity Domain Attributes]
[052h 0082   2]                     Reserved : 0000
[054h 0084   4]                       Length : 00000028
[058h 0088   2]        Flags (decoded below) : 0002
            Processor Proximity Domain Valid : 0
[05Ah 0090   2]                    Reserved1 : 0000
[05Ch 0092   4] Attached Initiator Proximity Domain : 00000000
[060h 0096   4]      Memory Proximity Domain : 00000001
[064h 0100   4]                    Reserved2 : 00000000
[068h 0104   8]                    Reserved3 : 0000000000000000
[070h 0112   8]                    Reserved4 : 0000000000000000

[078h 0120   2]               Structure Type : 0000 [Memory Proximity Domain Attributes]
[07Ah 0122   2]                     Reserved : 0000
[07Ch 0124   4]                       Length : 00000028
[080h 0128   2]        Flags (decoded below) : 0002
            Processor Proximity Domain Valid : 0
[082h 0130   2]                    Reserved1 : 0000
[084h 0132   4] Attached Initiator Proximity Domain : 00000000
[088h 0136   4]      Memory Proximity Domain : 00000002
[08Ch 0140   4]                    Reserved2 : 00000000
[090h 0144   8]                    Reserved3 : 0000000000000000
[098h 0152   8]                    Reserved4 : 0000000000000000

Given we do not have any hard-coded array in this data structure, I am not sure whether we can use MemProxDomainAttrCount to create the tables. 

But using Token for CmArmMemoryLatBwInfo is a great idea, I will modify the code accordingly.

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
```diff
@@ -723,6 +727,59 @@ typedef struct CmArmEtInfo {
...
+
+typedef struct CmArmMemoryLatBwInfo {
+  UINT16 Type;
+  UINT8  Flags;
+  UINT8  DataType;
+  UINT8  MinTransferSize;
+  UINT32 InitiatorProximityDomainsNumber;
+  UINT32 TargetProximityDomainsNumber;
+  UINT64 EntryBaseUnit;
+  UINT32 InitiatorProximityDomainList[8];
```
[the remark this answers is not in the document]
SophiaWang-Google: The have one System Locality Latency and Bandwidth Information Structure table:
Initiator Proximity Domain List[0]
Initiator Proximity Domain List[1]
Initiator Proximity Domain List[s-1]
If we want to use the token which is a good idea, we need to define the data structures as:
typedef struct CmArmMemoryLatBwInfo {
  UINT8  Flags;
  UINT8  DataType;
  UINT8  MinTransferSize;
  UINT32 InitiatorProximityDomainsNumber;
  UINT32 TargetProximityDomainsNumber;
  UINT64 EntryBaseUnit;
}

typedef struct CmArmInitiatorProximityDomainList {
CM_OBJECT_TOKEN    InitiatorProximityDomainListToken;
 UINT32 InitiatorProximityDomainList;
}

..

Please let me know whether it is what we want to do.

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
```diff
@@ -723,6 +732,70 @@ typedef struct CmArmEtInfo {
...
+/** A structure that describes the Memory Proximity Domain Attribute.
+
+    ID: EArmObjMemoryProxDomainAttrInfo
+*/
+
+typedef struct CmArmMemoryProxDomainAttrInfo {
+  UINT16 Flags;
+  UINT32 InitiatorProximityDomain;
+  UINT32 MemoryProximityDomain;
+} CM_ARM_MEMORY_PROX_DOMAIN_ATTR_INFO;
```
abdattar, line 744: This needs to be CM_ARCH_MEMORY_PROX_DOMAIN_ATTR_INFO and moved to ArchCommonNamespaceObjects.h.
X64 also supports the HMAT table.
Same comment applies to all structure info.

abdattar (commented):
make the HMAT table generator as generic library rather than Arm specific library.
X64 also supports the HMAT table.

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
pierregondois: If you want to have an arbitrary number of elements, you should use the a token. You can:
- Add a `CM_ARM_SMBIOS_HANDLES` structure as you did. When populating the information, you can create an array of this structure an use as many elements as desired.
- Add an `SmbiosHandlesToken` field in this structure. This Token will be used to retrieve/identify the previously defined array.

You can check how the`CM_ARM_GICC_INFO.CpcToken` is used for instance.

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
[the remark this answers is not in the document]
pierregondois: I was thinking about something like that:

```
typedef struct CmArmMemoryLatBwInfo {
UINT8 Flags;
UINT8 DataType;
UINT8 MinTransferSize;
UINT32 InitiatorProximityDomainsNumber;
UINT32 TargetProximityDomainsNumber;
UINT64 EntryBaseUnit;
// Token referencing a CM_ARCH_COMMON_OBJ_REF
CM_OBJECT_TOKEN InitiatorProximityDomainList;
// Token referencing a CM_ARCH_COMMON_OBJ_REF
CM_OBJECT_TOKEN TargetProximityDomainList;
// Token referencing a CM_ARCH_COMMON_WORD
CM_OBJECT_TOKEN RelativeDistanceEntry
}

struct CmArchCommonWord {
UINT16 Value;
} CM_ARCH_COMMON_WORD;
```

Then the `CM_ARCH_COMMON_OBJ_REF` referenced by `InitiatorProximityDomainList` will look like:
```
{
// Could be a GicC object, I,e, CPU0
Token0,
// Could be a GicC object, I,e, CPU1
Token1,
// Could be a Gic Its object (if they can be initiators)
Token2,
}
```

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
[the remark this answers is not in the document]
pierregondois: Here it is assumed that you have only one `Memory Latency Bandwidth` ACPI sub-table. If you want to describe 2 sub-tables, it won't be possible anymore as there is no way to identify which `CM_ARM_MEMORY_INIT_TARGET_INFO` should be used.
I think the `CM_ARM_MEMORY_INIT_TARGET_INFO` structure should be removed. You can use tokens to reference the objects you want in the other structures.

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
```diff
@@ -723,6 +732,70 @@ typedef struct CmArmEtInfo {
...
+typedef struct CmArmMemoryCacheInfo {
+  UINT32 MemoryProximityDomain;
+  UINT64 MemorySideCacheSize;
+  UINT32 CacheAttributes;
+  UINT16 NumSmbiosHandles;
+} CM_ARM_MEMORY_CACHE_INFO;
+
+typedef struct CmArmSmbioHandles {
+  UINT16 SmbiosHandles;
+} CM_ARM_SMBIOS_HANDLES;
```
samimujawar, line 798: I think this needs to be an arch common object. 

Also, there is a PR at https://github.com/tianocore/edk2-staging/pull/510/files which adds Dynamic SMBIOS support and is very close to being merged in the staging branch. Once merged, it should be possible to implement SMBIOS generators as well. The initial plan would be to have 1 or 2 SMBIOS generators on the staging branch and their corresponding implementation in edk2-platforms staging branch to demonstrate the usage. I think we are very close to getting there. 
This means we will have Dynamic SMBIOS support merged in edk2 mainline next. 

@gmahadevan Any thoughts on how this would work with Dynamic SMBIOS?

DynamicTablesPkg/Include/ArmNameSpaceObjects.h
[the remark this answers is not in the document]
SophiaWang-Google: CM_ARM_MEMORY_INIT_TARGET_INFO is removed since we will use token to get the same info.

DynamicTablesPkg/Include/ArchCommonNameSpaceObjects.h
```diff
@@ -145,6 +152,21 @@ typedef struct CmArchCommonObjRef {
...
+    can then query the configuration manager for these arrays using the
+    object ID EArchCommonObjCmRef.
+
+    This can be used is to represent one-to-many relationships between objects.
+
+    ID: EArchCommonWord
+*/
+typedef struct CmArchCommonWord {
+  UINT16 Value;
+} CM_ARCH_COMMON_WORD;
```
SophiaWang-Google, line 168: Please do not review this yet. 

Still debating whether we need to use CmArchCommonWord as suggested in the comment or CM_ARCH_COMMON_OBJ_REF should be good enough.

Also HmatGenerator.c is still under dev, please do not review yet.

SophiaWang-Google, line 168: Please do not review the code yet. Still debating whether we can use CM_OBJECT_TOKEN for 16 bits data array.

More testing and debugging now.

DynamicTablesPkg/Include/ArchCommonNameSpaceObjects.h
```diff
@@ -812,6 +834,87 @@ typedef struct CmArchCommonStaInfo {
...
+
+typedef struct CmArchCommonMemoryLatBwInfo {
+  UINT8  Flags;
+  UINT8  DataType;
+  UINT8  MinTransferSize;
+  UINT32 InitiatorProximityDomainsNumber;
+  UINT32 TargetProximityDomainsNumber;
+  UINT64 EntryBaseUnit;
+  // Token referencing CM_ARCH_COMMON_INITIATOR_DOMAIN_LIST
+  CM_OBJECT_TOKEN InitiatorProximityDomainListToken;
```
SophiaWang-Google, line 863: According to ACPI spec, InitiatorProximityDomainList and TargetProximityDomainList are just numbers/indexes of proximity domain.Please refer to **https://stevescargall.com/blog/2022/11/linux-numa-distances-explained/**

If we use CM_OBJECT_TOKEN to refer to every proximity domain, that means we need to create the big array of  CM_ARCH_COMMON_INITIATOR_DOMAIN_LIST, for example: if the system have 400 CPUs, and only have 4 proximity domains, we need to create array of  CM_ARCH_COMMON_INITIATOR_DOMAIN_LIST with size of 400, and each cpu points to one element inside the array.

It turns out to be:
cpu 0 ---> proximity domain 0
cpu 1 ---> proximity domain 0
...
cpu 99 ---> proximity domain 0
cpu 100 ---> proximity domain 1
...
cpu 199 ---> proximity domain 1
...
cpu 399 ---> proximity domain 3

Do we really want to create that big size of CM_ARCH_COMMON_TARGET_DOMAIN_LIST array, or we can rely on edk2-platform to fill in proximity numbers (in above example: it is 4)

To avoid hard-coded array size in the original version, we can use separate data structures as shown in current version:
typedef struct CmArchCommonInitiatorProximityDomainList {
  UINT32 InitiatorProximityDomainList;
} CM_ARCH_COMMON_INITIATOR_DOMAIN_LIST;

/** A structure that describes the Target Proximity Domain List Info.
    ID: EArchCommonObjTargetDomainList
*/
typedef struct CmArchCommonTargetProximityDomainList {
  UINT32 TargetProximityDomainList;
} CM_ARCH_COMMON_TARGET_DOMAIN_LIST;

/** A structure that describes the Relative Distance Entry Info.
    ID: EArchCommonObjRelativeDistanceEntry
*/
typedef struct CmArchCommonRelativeDistanceEntry
  UINT16 RelativeDistanceEntry;
} CM_ARCH_COMMON_RELATIVE_DISTANCE_ENTRY;

And rely on the InitiatorProximityDomainCount 
    Status = GetEArmObjInitiatorProximityDomain (
             CfgMgrProtocol,
             CM_NULL_TOKEN,
             &InitiatorProximityDomainInfo,
             &InitiatorProximityDomainCount
             );
to fill in multiple entries of CM_ARCH_COMMON_INITIATOR_DOMAIN_LIST

Please let me know your preference.

pierregondois: Hello Sophia,
I think it would be good to use a `CM_ARCH_COMMON_OBJ_REF` to store the Initiator/Target.
Indeed, we should ideally avoid duplication of the information:
- If the index of a domain changes (for any reason), we would have to find back where this index was used and modify it aswell.
- If the index need to be generated one day, it will be difficult to patch all the places where the index was hand-written
- Ideally, we should actually not need a `ProximityDomain` field in the `CM_ARM_GICC_INFO`, we should store a reference to the proximity domain the CPU belongs to. Unfortunately this design was too complex to start with, but IMO we should avoid hand-writing values that can be generated

pierregondois: Hello Sophia,
when reading your message, it seems I might have missed/misunderstood something.
Please let me come back to you tomorrow to be sure about the suggestion.

pierregondois: Hello Sophia,
The best solution would be to deprecate the CM_ARM_GICC_INFO.ProximityDomain
and CM_ARM_GICC_INFO.ClockDomain fields I think. Same for the CM_X64_LOCAL_APIC_X2APIC_AFFINITY_INFO struct.

Ideally we should:
1-
Add a new CM_ARCH_COMMON_DOMAIN object:
```
typedef struct CmArchCommonDomain {
  /// Generate Domain Id
  /// TRUE if the DynamicTablesPkg framework should
  /// generate the Domain Id
  BOOLEAN GenerateDomainId;

  /// If GenerateDomainId is FALSE, use this value
  /// for the DomainId;
  UINT32 DomainId;
} CM_ARCH_COMMON_DOMAIN;
```
2-
Add new fields in the CM_ARM_GICC_INFO structure:
- ProximityDomainToken: Token referencing a CM_ARCH_COMMON_DOMAIN
- ClockDomainToken: Token referencing a CM_ARCH_COMMON_DOMAIN

3-
Deprecate the following fields in CM_ARM_GICC_INFO:
- ProximityDomain
- ClockDomain
The SratGenerator is the only consumer of these fields. We will have to either:
- Check if the `ProximityDomainToken|ClockDomainToken` are populated. If yes, use them to generate the Srat table, otherwise use the previous `ProximityDomain|ClockDomain` fields
- Create a second SratGenerator for the new users and don't modify the initial SratGenerator

4-
The diagram would look like that:

```
CM_ARCH_COMMON_MEMORY_LAT_BW_INFO.TargetProximityDomainListToken-------+
CM_ARCH_COMMON_MEMORY_LAT_BW_INFO.InitiatorProximityDomainListToken-+  |
                                                                    |  |
                                                                    |  |
              +-----------------------------------------------------+  |
              |                                                        |
              v                                                        |
+-------------------------+                                            |
|  CM_ARCH_COMMON_OBJ_REF |                                            |
|  (this is an array)     |                                            |
|-------------------------|                                            |
|          Token[0]-------+----------------+                           |
|          Token[1]-+     |                |                           |
+-------------------|-----+                |                           |
                    |                      |                           |
                    v                      v                           v
+-------------------------+   +-------------------------+   +-------------------------+
|   CM_ARCH_COMMON_DOMAIN |   |   CM_ARCH_COMMON_DOMAIN |   |   CM_ARCH_COMMON_DOMAIN |
|     (Proximity Domain)  |   |     (Proximity Domain)  |   |     (Proximity Domain)  |
|-------------------------|   |-------------------------|   |-------------------------|
| GenerateDomainId::FALSE |   | GenerateDomainId::FALSE |   | GenerateDomainId::FALSE |
| DomainId::0             |   | DomainId::1             |   | DomainId::2             |
+-------------------------+   +-------------------------+   +-------------------------+
         ^ ^                                  ^ ^
         | |                                  | |
         | +--------------------------------+ | |
         +--------------------------------+ | | |
                                          | | | |
+---------------------------------+       | | | |
|        CM_ARM_GICC_INFO         |       | | | |
|  (this is an array)             |       | | | |
|---------------------------------|       | | | |
| GicC[0].ProximityDomainToken----+-------+ | | |
| GicC[1].ProximityDomainToken----+---------+ | |
| GicC[2].ProximityDomainToken----+-----------+ |
| GicC[3].ProximityDomainToken----+-------------+
+---------------------------------+
```
-------------------------------------------------------------------------------------------
-------------------------------------------------------------------------------------------
-------------------------------------------------------------------------------------------


And to depict the relative distance between domains:
1-
Add a strucure:
```
typedef struct CmArchCommonRelativeDistanceEntry {
  CM_OBJECT_TOKEN FirstDomain;
  CM_OBJECT_TOKEN SecondDomain;
  UINT16          RelativeDistance;
} CM_ARCH_COMMON_RELATIVE_DISTANCE_ENTRY;
```
4-
The diagram would look like that:
```
+-------------------------+   +-------------------------+   +-------------------------+
|   CM_ARCH_COMMON_DOMAIN |   |   CM_ARCH_COMMON_DOMAIN |   |   CM_ARCH_COMMON_DOMAIN |
|     (Proximity Domain)  |   |     (Proximity Domain)  |   |     (Proximity Domain)  |
|-------------------------|   |-------------------------|   |-------------------------|
| GenerateDomainId::FALSE |   | GenerateDomainId::FALSE |   | GenerateDomainId::FALSE |
| DomainId::0             |   | DomainId::1             |   | DomainId::2             |
+-------------------------+   +-------------------------+   +-------------------------+
            ^                                    ^                      ^   ^
            |                                    |                      |   |
            |                                    |                      |   |
            +-------------------------------+    |                      |   |
                                            |    |                      |   |
+----------------------------------------+  |    |                      |   |
| CM_ARCH_COMMON_RELATIVE_DISTANCE_ENTRY |  |    |                      |   |
|    (This is an array)                  |  |    |                      |   |
+----------------------------------------+  |    |                      |   |
|  (Index0)                              |  |    |                      |   |
|  FirstDomain---------------------------+--+    |                      |   |
|  SecondDomain--------------------------+-------|----------------------+   |
|  RelativeDistance::4 (distance of 4)   |       |                          |
+----------------------------------------+       |                          |
|  (Index1)                              |       |                          |
|  FirstDomain---------------------------+-------+                          |
|  SecondDomain--------------------------+----------------------------------+
|  RelativeDistance::6 (distance of 6)   |
+----------------------------------------+
                ^
                |
                +---------------------------------------------+
                                                              |
CM_ARCH_COMMON_MEMORY_LAT_BW_INFO.RelativeDistanceEntryToken--+
```

@samimujawar , would it be possible to also have a look if you're ok with this design ?
@SophiaWang-Google If the design seems ok to you and Sami, I can help regarding the rework, as you are mainly interested in the HmatGenerator I assume, 

Regards,
Pierre

SophiaWang-Google: Pierre,

Thanks for proposing the solution on creating the new CM_ARCH_COMMON_DOMAIN structure.

For this new solution, my original concern still exists: 
If we use GicC token for the reference of ProximityDomain (see below diagram copied from your reply), for example the system has 400 CPUS (4 ProximityDomains), we end up with creating 400 tokens.  

 CM_ARM_GICC_INFO         |       | | | |
|  (this is an array)             |       | | | |
|---------------------------------|       | | | |
| GicC[0].ProximityDomainToken----+-------+ | | |
| GicC[1].ProximityDomainToken----+---------+ | |
| GicC[2].ProximityDomainToken----+-----------+ |
| GicC[3].ProximityDomainToken

Another concern is does DynamicTablesPkg framework know how to generate ProximityDomain ID? I assume it is supposed to be platform dependent, please correct me if I am wrong.

pierregondois: Hello Sophia,
To keep the formatting of text in you reply, I think you should use '`' or '```'. You can click on 'Preview' to see what the answer will look like.
- If there are 400 CPUs, there will be effectively be 400 GicC structures with each containing one token pointing to a `CM_ARCH_COMMON_DOMAIN` structure. If the 400 CPUs are partitioned in 4 Proximity Domains, and if there are 2 Memory Domains (for instance), the latency BW table will be an array of 4x2. Or am I missing something ?
- The DynamicTablesPkg is not capable of generating values that can be re-used accross generators right now, but there is ongoing work to enable that. 
- IIUC, the ProximityDomain values are not meaningful, so either the DynamicTablesPkg is creating all the IDs and there will be no issue, either the use wants specific IDs to be used and the used can provide it by setting `GenerateDomainId:FALSE` and `DomainId:4` (for instance)

```
typedef struct CmArchCommonDomain {
  /// Generate Domain Id
  /// TRUE if the DynamicTablesPkg framework should
  /// generate the Domain Id
  BOOLEAN GenerateDomainId;

  /// If GenerateDomainId is FALSE, use this value
  /// for the DomainId;
  UINT32 DomainId;
} CM_ARCH_COMMON_DOMAIN;
```

SophiaWang-Google: Pierre,

Can you please give rough estimate when this re-architecture work will be done?

Given HMAT generator code is ready with current code base, can we review and get the code merged for now and change the code later with your newly re-architecture work?

pierregondois: Hello Sophia,
I think it would be better (at least for me) to use the correct objects in the first place. Would it be possible to rebase your work on this branch and use the CmObj that are there ?

Normally there should not need more information than that. This will require a bit of juggling though...

https://github.com/pierregondois/edk2/tree/pg/proximity_domain_rework

SophiaWang-Google: Hello Pierre,

Thanks for working on proximity_domain_rework, I appreciate the speed.

I just updated ArchCommonNameSpaceObjects.h by using CmObj.

For now, I will pause the update of this CL until https://github.com/pierregondois/edk2/tree/pg/proximity_domain_rework is merged.

Please kindly let me know when HMAT generator work can be resumed.

Best,
Sophia

pierregondois: Hello Sophia,
ok yes understood

PR #10931, opened 2025-04-08
Arm*Pkg, UefiPayloadPkg: Clean up ARM/AArch64 FP behaviour

leiflindholm:
There is a Pcd defined in ArmPkg to indicate whether a given platform should enable
floating-point support (and permit it in codegen).

However, for AArch64 this is in fact mandated by the UEFI specification, so does
not make sense to be configurable. And there is no need to probe for FP support at
runtime.

Additionally, since this affects codegen, the architectural configuration required
to support floating-point (and NEON) must be performed before we start executing
C code.

So:
- Move invocation of ArmEnableVFP to asm.
- Unconditionally enable FP/NEON for AArch64.
- Drop a stray dependency declaration on PcdVFPEnabled from ArmPkg CpuDxe.
- Make PcdVFPEnabled ARM-only and drop setting it for AArch64 platforms.

- [x] Breaking change?
  - Out-of-tree (including edk2-platforms) AArch64 platforms that explicitly set PcdVFPEnabled will fail at build time.

## How This Was Tested

Build tested.

## Integration Instructions

AArch64 platforms should stop setting PcdVFPEnabled.
Platforms buildable for both Arm and AArch64 should move PcdVFPEnabled to a [PcdsFixedAtBuild.ARM] section in their .dsc.

### DIFF
--- a/ArmVirtPkg/ArmVirtQemu.dsc
+++ b/ArmVirtPkg/ArmVirtQemu.dsc
@@ -246,8 +246,6 @@
   # point only, for entry point versions >= 3.0.
   gEfiMdeModulePkgTokenSpaceGuid.PcdSmbiosEntryPointProvideMethod|0x2
 
-  gArmTokenSpaceGuid.PcdVFPEnabled|1
-
 [PcdsDynamicDefault.common]
   gEfiMdePkgTokenSpaceGuid.PcdPlatformBootTimeOut|3
 
--- a/ArmPkg/ArmPkg.dec
+++ b/ArmPkg/ArmPkg.dec
@@ -130,10 +130,6 @@
 [PcdsFixedAtBuild.common]
   gArmTokenSpaceGuid.PcdTrustzoneSupport|FALSE|BOOLEAN|0x00000006
 
-  # This PCD should be a FeaturePcd. But we used this PCD as an '#if' in an ASM file.
-  # Using a FeaturePcd make a '(BOOLEAN) casting for its value which is not understood by the preprocessor.
-  gArmTokenSpaceGuid.PcdVFPEnabled|0|UINT32|0x00000024
-
   gArmTokenSpaceGuid.PcdCpuResetAddress|0x00000000|UINT32|0x00000005
 
   #
@@ -212,6 +208,10 @@
   gArmTokenSpaceGuid.PcdFvBaseAddress|0|UINT64|0x0000002D
 
 [PcdsFixedAtBuild.ARM]
+  # This PCD should be a FeaturePcd. But we used this PCD as an '#if' in an ASM file.
+  # Using a FeaturePcd make a '(BOOLEAN) casting for its value which is not understood by the preprocessor.
+  gArmTokenSpaceGuid.PcdVFPEnabled|0|UINT32|0x00000024
+
   #
   # ARM Security Extension
   #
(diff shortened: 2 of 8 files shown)


ArmVirtPkg/Library/ArmPlatformLibQemu/AArch64/ArmPlatformHelper.S
```diff
@@ -112,4 +112,4 @@ ASM_FUNC(ArmPlatformPeiBootAction)
   msr    sctlr_el1, x3           // enable MMU and caches
   isb
 
-0:b      ArmEnableVFP            // enable SIMD before entering C code
```
ardbiesheuvel, line 115: The user of this library does not use PrePi, so moving this is not going to work AFAICT.

ArmVirtPkg/ArmVirtQemu.dsc
```diff
@@ -246,8 +246,6 @@
   # point only, for entry point versions >= 3.0.
   gEfiMdeModulePkgTokenSpaceGuid.PcdSmbiosEntryPointProvideMethod|0x2
 
-  gArmTokenSpaceGuid.PcdVFPEnabled|1
-
```
ardbiesheuvel, line 250: Why is this being disabled for 32-bit ARM?

leiflindholm: It was never set for 32-bit ARM, this is in section [PcdsFixedAtBuild.AARCH64].

ArmVirtPkg/Library/ArmPlatformLibQemu/AArch64/ArmPlatformHelper.S
leiflindholm: Ah, whoops. In that case, I should be able to just drop this hunk, right?

ardbiesheuvel: I think so yes

leiflindholm: Done.

ArmVirtPkg/PrePi/Arm/ModuleEntryPoint.S
```diff
@@ -12,6 +12,12 @@ ASM_FUNC(_ModuleEntryPoint)
   // Do early platform specific actions
   bl    ASM_PFX(ArmPlatformPeiBootAction)
 
+#if (FixedPcdGet32 (PcdVFPEnabled))
+   // Enable Floating Point. This needs to be done before entering C code, which
+   // may use FP/SIMD registers.
+   bl    ArmEnableVFP
+#endif
+
```
ardbiesheuvel, line 20: This is actually not the case on ARM, since we explicitly use the softfloat ABI.

I don't mind changing this for symmetry, but let's call that out here.

ardbiesheuvel (approved):

ArmVirtPkg/PrePi/Arm/ModuleEntryPoint.S
leiflindholm: Fair point, comment updated.

PR #10937, opened 2025-04-09
SecurityPkg: Prevent invalid DBX from being set

Flickdm:
# Description

This PR prevents setting the  Dbx variable the Default being provided is less than the size of the EFI_SIGNATURE_LIST structure. This is to prevent the setting of ab invalid DBX which would cause the system to fail to boot.

Additionally, this can be used to signal that setting the DBX should leave DBX undefined for Platforms that want to let the OS be the sole servicer of the DBX.

Breakdown of the math is as follows:

1. **`sizeof(EFI_SIGNATURE_LIST)`**:
   - This is the size of the `EFI_SIGNATURE_LIST` structure itself, which includes: - `EFI_GUID SignatureType` (16 bytes) - `UINT32 SignatureListSize` (4 bytes) - `UINT32 SignatureHeaderSize` (4 bytes) - `UINT32 SignatureSize` (4 bytes)
   - Total: `16 + 4 + 4 + 4 = 28 bytes`

2. **`SignatureHeaderSize`**:
   - This is the size of the optional signature header. If no header is provided, this value is `0`.

3. **`SignatureSize`**:
   - This is the size of each `EFI_SIGNATURE_DATA` entry. For an empty list, this value is `0`.

The total size of an empty `EFI_SIGNATURE_LIST` is:
```c
sizeof(EFI_SIGNATURE_LIST) + SignatureHeaderSize
```

1. **No Signature Header**:
   - If `SignatureHeaderSize = 0`, the size is: ```c 28 + 0 = 28 bytes ```

2. **With a Signature Header**:
   - If `SignatureHeaderSize = 16` (example size for a header), the size is: ```28 + 16 = 44 bytes ```

- **Minimum Size**: `28 bytes` (if `SignatureHeaderSize = 0`).
- **Additional Size**: Add the value of `SignatureHeaderSize` if a header is included.

- [ ] Breaking change?
- [X] Impacts security?
  - Allows platforms to leave the DBX undefined. While this isnt a security concern if the OS services the DBX this should be understood by platform integrators.
- [ ] Includes tests?

## How This Was Tested

On a physical platform using defaults from https://github.com/microsoft/secureboot_objects/pull/174

## Integration Instructions

N/A

### DIFF
--- a/SecurityPkg/Library/SecureBootVariableLib/SecureBootVariableLib.c
+++ b/SecurityPkg/Library/SecureBootVariableLib/SecureBootVariableLib.c
@@ -23,6 +23,21 @@
 #include <Library/SecureBootVariableLib.h>
 #include <Library/PlatformPKProtectionLib.h>
 
+//
+// This is the minimum size of a EFI_SIGNATURE_LIST and data and still be valid.
+//
+// The UEFI specification defines this as:
+//  Each signature list is a list of signatures of one type, identified by SignatureType.
+//  The signature list contains a header and then an array of zero or more signatures in
+//  the format specified by the header. The size of each signature in the signature list
+//  is specified by SignatureSize.
+//
+// This is the size of a signature list with some data in it. Effectively 44 bytes are
+// required to set the DBX. However in reality, the size of a real payload will be larger
+// since this miniumum size does not consider the SIGNATURE_LIST
+//
+#define MINIMUM_VALID_SIGNATURE_LIST  (sizeof(EFI_SIGNATURE_LIST) + sizeof(EFI_SIGNATURE_DATA) - 1)
+
 // This time can be used when deleting variables, as it should be greater than any variable time.
 EFI_TIME  mMaxTimestamp = {
   0xFFFF,     // Year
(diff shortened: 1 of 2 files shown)


SecurityPkg/Library/SecureBootVariableLib/SecureBootVariableLib.c
```diff
@@ -816,12 +831,24 @@ SetSecureBootVariablesToDefault (
...
                &gEfiImageSecurityDatabaseGuid,
                DataSize,
                Data
                );
+  } else {
+    //
+    // The DBX is allowed to be empty by default
+    //
+    DEBUG ((DEBUG_INFO, "%a - Skipping Dbx - DataSize(%u)\n", __func__, DataSize));
+    Status = EFI_SUCCESS;
```
jyao1, line 850: Ummmm. Maybe user inputs an invalid DBX e.g. DataSize is 16 bytes only, instead of "empty" DBX where DataSize is 0.

Should be distinguish them?

jyao1:
@Flickdm , build failure?

SecurityPkg/Library/SecureBootVariableLib/SecureBootVariableLib.c
Flickdm: Honestly I was hoping for a bit of discussion here!

Originally, I wanted to use a DataSize of 0 to signal (and maybe that is the correct answer) that the DBX should not be set. However [EnrollFromInput(..)](https://github.com/tianocore/edk2/blob/fdede2e1eaa44119aabb4d29e2c6812d758fb953/SecurityPkg/Library/SecureBootVariableLib/SecureBootVariableLib.c#L712) does not allow for a DataSize to be 0. Which means in market devices without a firmware upgrade may reject those 0 byte payloads. Additionally, Platforms that use a PCD array to set SECURE_BOOT_PAYLOAD_INFO cannot set 0 due to limitations in autogen. 

Given this, I rethought how we structure EFI_SIGNATURE_LISTS and it is possible to create a EFI_SIGNATURE_LIST with no data effectively a Empty Signature List. This is a valid EFI_SIGNATURE_LIST that will allow a system to boot with simply no data. Additionaly, this is spec allowed

>  Each signature list is a list of signatures of one type, identified by SignatureType.
>  The signature list contains a header and then an array of zero or more signatures in
>  the format specified by the header. The size of each signature in the signature list
>  is specified by SignatureSize.

[source](https://uefi.org/specs/UEFI/2.9_A/32_Secure_Boot_and_Driver_Signing.html#efi-signature-data)

So given this there are 4 cases that should be handled

1. DataSize = 0
1.1 Special case where no data is provided - effectively empty
2. DataSize = 1-27
2.1 Invalid DataSize - system will not boot
3. DataSize = 28
3.1 Special case where Signature List has no data - Should I allow this to be set? I'm effectively using this case as a backwards compatible special case to leave the DBX empty and still have the platform boot even if it is set. 
4. DataSize = 29-43
4.1 Could be Valid signature list but invalid data

This leaves me with 2 questions,
Should I distinguish 0,  28, and anything else as Invalid"?
Should I prevent a 28 byte valid Empty EFI Signature list from being set, or can I use this as a backwards compatible solution?

Flickdm: Additionally, this minimum size restraint is also valid for the other Secure Boot Variables if you think we should handle those cases? It should definitely result in INVALID_PARAMETER if a user accidentally puts data in there that will prevent the machine from booting

jyao1 (approved):

PR #10948, opened 2025-04-11
Rename Azure-Pipeline for Windows

Damien-Chen:
Rename it to VS2022 because it now run on VS2022 to avoid confusion.

### DIFF
--- a/.azurepipelines/Windows-VS2022.yml
+++ b/.azurepipelines/Windows-VS2022.yml
@@ -1,0 +1,29 @@
+## @file
+# Azure Pipeline build file for a build using Windows and VS2022
+#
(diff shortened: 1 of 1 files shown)


makubacki:
@Damien-Chen, thanks for the PR but the YAML file has not been renamed because more extensive changes are required in Azure DevOps to use the new file name and change other pipeline information. I started to do that in https://github.com/tianocore/edk2/pull/8637/ and I can revisit completing those changes.

However, this PR is incomplete and not aligned with the direction to be VS version agnostic and also address the GCC YAML files. For that reason, I will take this as a reminder to reopen that PR and make the necessary changes.

PR #10950, opened 2025-04-11
Add GHSA PR Checks

makubacki:
# Description

Performs simple CI checks against a common set of pacakges and
targets for PRs in private tianocore GitHub Security Advisory
(GHSA) forks. This is meant to provide relatively simple sanity
checks there before the contributions are run against public
CI infrastructure.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- [Tested that all builds currently pass](https://github.com/tianocore/edk2/actions/runs/14411844399)
- [Tested that the workflow does not run in the normal edk2 repo](https://github.com/tianocore/edk2/actions/runs/14412026629)
- Testing that it does run as expected in GHSA fork PRs (WIP)

## Integration Instructions

- N/A

### DIFF
--- a/.github/workflows/ghsa-pr.yml
+++ b/.github/workflows/ghsa-pr.yml
@@ -1,0 +1,260 @@
+# A simple PR gate workflow that runs on GitHub Security Advisory Pull Requests.
+#
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+
+name: "GHSA Pull Request Check"
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  ghsa_pr_checks:
+    if: >
+      github.event.pull_request.base.repo.fork == true &&
+      github.event.pull_request.base.repo.owner.login == 'tianocore' &&
+      startsWith(github.event.pull_request.base.repo.name, 'edk2-ghsa-')
+
+    name: GHSA Pull Request (CI) Checks
+    runs-on: ubuntu-24.04
+    container:
+      image: ghcr.io/tianocore/containers/fedora-40-test:c98ff99
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    permissions:
+      contents: read
+
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/ghsa-pr.yml
```diff
@@ -0,0 +1,260 @@
...
+        path: |
+          **/UPDATE_LOG.txt
+        retention-days: 7
+        if-no-files-found: ignore
+
+    - name: Build Tools From Source
+      run: python BaseTools/Edk2ToolsBuild.py -t GCC5
+
+    - name: CI Build
+      run: stuart_ci_build -c .pytool/CISettings.py -t DEBUG -p ${{ matrix.Package }} -a ${{ matrix.ArchList }} TOOL_CHAIN_TAG=GCC5
```
Javagedes, line 222: If this is simple sanity checks, do we need to run all CI checks? Could we get away with a subset? Maybe just CompilerPlugin? or CompilerPlugin and HostBasedUnitTest?

makubacki: Simple really meant a customized set of packages/checks. Given how few updates are made to GHSA PRs, it's okay to run everything.

makubacki:
I should have read https://github.com/github/roadmap/issues/627 before starting this. The workflow does work but GitHub currently does not support running actions from GHSA private forks. I'm going to leave this PR up and rework it for normal CI to help provide a path to move our infrastructure (even if only partially) more into GitHub.

.github/workflows/ghsa-pr.yml
```diff
@@ -0,0 +1,260 @@
...
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Set PATH
+      run: |
+        echo "${HOME}/.local/bin" >> $GITHUB_PATH
+        echo "new PATH=${PATH}"
+
+    - name: Configure Git Safe Directory
```
Javagedes, line 105: Do we care to run stuart_pr_eval to skip any packages?

makubacki: For the original GHSA scenario it was not necessary. For the pivot to being a normal workflow, yes. There will also be some other changes like that made too.

PR #10951, opened 2025-04-12
SecurityPkg/FvReportPei: Add explicit handling for the case where there is only one FV.

AnsenHuangMSFT:
# Description
there is a potential issue in the code when handling a single Firmware Volume (FV) report. Specifically, the problem lies in the logic for calculating and verifying the hash values for the FVs.

Key Issue:
The code assumes that there are multiple FVs when calculating the hash values. If there is only one FV, the logic for handling FvHashValue and HashValue might not work as intended.

FvHashValue == HashValue: This condition is true only if no hash values were calculated (e.g., no FVs were processed). For a single FV, this condition might incorrectly pass or fail depending on how FvHashValue is updated.
AlgInfo->HashAll: This function calculates the hash for all FVs. If there is only one FV, the range FvHashValue - HashValue might not be handled correctly, leading to incorrect hash verification.
CompareMem: The comparison might fail if the calculated hash (FvHashValue) is not properly updated or initialized for a single FV.

Potential Bug:
If there is only one FV, the code might:

Skip the hash calculation for the single FV due to incorrect pointer arithmetic or logic.
Fail to verify the hash because FvHashValue is not updated correctly.
Incorrectly report EFI_VOLUME_CORRUPTED even if the single FV is valid.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Only report 1 FV info and the code able to return success result. 

## Integration Instructions
N/A

### DIFF
--- a/SecurityPkg/FvReportPei/FvReportPei.c
+++ b/SecurityPkg/FvReportPei/FvReportPei.c
@@ -252,16 +252,27 @@ VerifyHashedFv (
     FvInfo[FvIndex].Base = (UINT64)(UINTN)FvBuffer;
   }
 
-  //
-  // Check final hash for all FVs.
-  //
-  if ((FvHashValue == HashValue) ||
-      (AlgInfo->HashAll (HashValue, FvHashValue - HashValue, FvHashValue) &&
-       (CompareMem (HashInfo->Hash, FvHashValue, AlgInfo->HashSize) == 0)))
-  {
-    Status = EFI_SUCCESS;
+  if (FvNumber == 1) {
+    //
+    // Directly compare the single FV's hash
+    //
+    if (CompareMem (HashInfo->Hash, HashValue, AlgInfo->HashSize) == 0) {
+      Status = EFI_SUCCESS;
+    } else {
+      Status = EFI_VOLUME_CORRUPTED;
+    }
   } else {
-    Status = EFI_VOLUME_CORRUPTED;
+    //
+    // Check final hash for all FVs. Handle multiple FVs
+    //
+    if ((FvHashValue == HashValue) ||
+        (AlgInfo->HashAll (HashValue, FvHashValue - HashValue, FvHashValue) &&
+         (CompareMem (HashInfo->Hash, FvHashValue, AlgInfo->HashSize) == 0)))
+    {
+      Status = EFI_SUCCESS;
+    } else {
+      Status = EFI_VOLUME_CORRUPTED;
+    }
   }
 
 Done:


jyao1:
@AnsenHuangMSFT ,
Would you please clarify, what you mean `the logic for handling FvHashValue and HashValue **might not work** as intended.`?
What do you observed in this case?

In the sentence, `For a single FV, this condition **might incorrectly pass or fail** depending on how FvHashValue is updated.`, do you find the issue when you do code review? or do you run into issue in your environment?
If it is a real case, what do you observed in your environment?

To me, the fix seems a work around - just handle a special case in a special way.
I am not sure why we cannot merge the 1 vol case into multiple vol case?

Last but not least, would you please let me know how you validate the fix?

AnsenHuangMSFT:
1. In my BIOS, I only report 1 FV info to gEdkiiPeiFirmwareVolumeInfoStoredHashFvPpiGuid.    
    With only 1 FV report,  The VerifyHashedFv also return EFI_VOLUME_CORRUPTED

2.I am run into issue in my environment. It is a real case,  In my system, I only have one FV info to report for OBB. 
3.When only one FV info report.  below code won't work, it assume there is always multi-FV info report in gEdkiiPeiFirmwareVolumeInfoStoredHashFvPpiGuid.   The code is not a workaround. it is a real fix to handle if there is only one FV report to gEdkiiPeiFirmwareVolumeInfoStoredHashFvPpiGuid

  //
  // Check final hash for all FVs.
  //
  if ((FvHashValue == HashValue) ||
      (AlgInfo->HashAll (HashValue, FvHashValue - HashValue, FvHashValue) &&
       (CompareMem (HashInfo->Hash, FvHashValue, AlgInfo->HashSize) == 0)))
  {
    Status = EFI_SUCCESS;
  } else {
    Status = EFI_VOLUME_CORRUPTED;
  }

4.  I test the code with only report 1 FV info to gEdkiiPeiFirmwareVolumeInfoStoredHashFvPpiGuid. and the VerifyHashedFv able to return EFI_SUCCESS with this PR change.

jyao1:
> 1. In my BIOS, I only report 1 FV info to gEdkiiPeiFirmwareVolumeInfoStoredHashFvPpiGuid.
>    With only 1 FV report,  The VerifyHashedFv always return EFI_VOLUME_CORRUPTED
>
> 2.I am run into issue in my environment. It is a real case, In my system, I only have one FV info to report for OBB.

OK, from this sentence, I only know it does not work in your environment. But what is the root-case? I still do not understand. 

Your description is very vague - `For a single FV, this condition **might incorrectly pass or fail** depending on how FvHashValue is updated`.

So, how FvHashValue is updated in your environment?
Could it be caused by your tool to calculate the hash value, because your tool is using different logic with the current code?
I just guess, because I have no clue.

Would you please educate me on what the root case is?

3.When only one FV info report. below code won't work, it assume there is always multi-FV info report in gEdkiiPeiFirmwareVolumeInfoStoredHashFvPpiGuid. The code is not a workaround. it is a real fix to handle if there is only one FV report to gEdkiiPeiFirmwareVolumeInfoStoredHashFvPpiGuid

I feel we might use "workaround" for different meaning. Let me clarify without using this term.

I know this fix can pass the check in your environment. What I am not sure is that if this is the best way to fix it.

To me, it makes more sense to provide a fix to merge one FV case into multiple FVs case, instead of just write another branch to handle a specific case. I am not sure why it cannot be achieved. Would you please enlighten me on that?

> 4. I test the code with only report 1 FV info to gEdkiiPeiFirmwareVolumeInfoStoredHashFvPpiGuid. and the VerifyHashedFv able to return EFI_SUCCESS with this PR change.

At least, you need to validate the failure case... But we can hold this.
The first thing we need to do is to understand the root cause. See my comment in 2 and 3.

PR #10958, opened 2025-04-15
Add setup variable save callback GUID

mark-skum:
When saving variables in the setup menu, a callback event is required to handle project-specific operations. For example, on ARM platforms, we need to save DDR parameters to SPI flash. During the next boot, the system processor will read these parameters to initialize the DDR.


mark-skum:
update for a new branch

PR #10960, opened 2025-04-15
ArmPkg/ArmScmiDxe: Fix SCMI message parameter override in multi-transaction cases

JohnnyFan-CIX:
Fix an issue where SCMI message parameters could be incorrectly overwritten by return values when multiple transactions are needed to retrieve data.

This happens in scenarios such as:
  1. ClockDescribeRates where the number of clock rates exceeds the transfer limit.
  2. PerformanceDescribeLevels where the number of performance levels is too large.

In these cases, part of the input message buffer may be reused or overwritten during each iteration, leading to incorrect behavior.

Ensure the message parameters are preserved properly between transactions.


JohnnyFan-CIX:
@leiflindholm @ardbiesheuvel @samimujawar Could you please help review this PR? Thank you!

ArmPkg/Drivers/ArmScmiDxe/ScmiClockProtocol.c
```diff
@@ -199,18 +200,19 @@ ClockDescribeRates (
...
+  Status = ScmiCommandGetPayload (&MessageParams1);
   if (EFI_ERROR (Status)) {
     return Status;
   }
 
   Cmd.ProtocolId = ScmiProtocolIdClock;
   Cmd.MessageId  = ScmiMessageIdClockDescribeRates;
 
-  *MessageParams++ = ClockId;
+  MessageParams2 = MessageParams1+1;
```
ardbiesheuvel, line 211: What is this supposed to achieve? What are you expecting `MessageParams1+1` to refer to?

JohnnyFan-CIX: According to SCMI Specification Section 4.6.2.5, when an agent sends a message to the PMU (Power Management Unit), MessageParams1 points to the input parameter clock_id, and MessageParams2 points to the input parameter rate_index. Since SCMI uses a shared memory communication mechanism, when the agent receives a response from the PMU, MessageParams1 then points to the return parameter status, and MessageParams2 points to num_rate_flags.

When multiple communications are required to retrieve all supported clock rates, both MessageParams1 and MessageParams2 must be properly re-initialized before each request. This ensures that the return values do not override the input parameters, maintaining correctness across all message exchanges.

JohnnyFan-CIX: According to SCMI Specification Section 4.6.2.5, when an agent sends a message to the PMU (Power Management Unit), MessageParams1 points to the input parameter clock_id, and MessageParams2 points to the input parameter rate_index. Since SCMI uses a shared memory communication mechanism, when the agent receives a response from the PMU, MessageParams1 then points to the return parameter status, and MessageParams2 points to num_rate_flags.

When multiple communications are required to retrieve all supported clock rates, both MessageParams1 and MessageParams2 must be properly re-initialized before each request. This ensures that the return values do not override the input parameters, maintaining correctness across all message exchanges.

samimujawar: @JohnnyFan-CIX thanks for the above explanation. Is it possible to update the commit message to reflect the above clarification, please?

Also if you could add a space between the '+' operator, please? 
i.e. 
- MessageParams2 = MessageParams1+1;
+MessageParams2 = MessageParams1 + 1;

JohnnyFan-CIX: @samimujawar Sure, thanks for the feedback!

I've updated the commit message to include the clarification, and also added the space around the + operator as suggested.

Please let me know if there's anything else needed.

samimujawar (approved):

PR #10967, opened 2025-04-15
MdeModulePkg: DxeCore: Set Image Memory Protections Through GCD

### LINKED ISSUE #10771: [Bug]: DXE core memory protection feature might cause inconsistency between GCD internal data structure with CPU page table

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### Code first?

- [ ] Yes

MdeModulePkg

DEBUG, RELEASE

### Current Behavior

When we enable DXE core memory protection feature, if we set some GCD irrelevent attributes like EFI_MEMORY_SP or EFI_MEMORY_CPU_CRYPTO via GCD SetMemorySpaceAttributes (), it will cause page fault except.

### Expected Behavior

We can still invoke SetMemorySpaceAttributes () successfully to **append** EFI_MEMORY_SP or EFI_MEMORY_CPU_CRYPTO 

### Steps To Reproduce

1. Set the following two platform PCD to enable memory protection.
PcdSetNxForStack| TRUE
PcdDxeNxMemoryProtectionPolicy| 0x7FD5

2. Invoke GCD SetMemorySpaceAttributes () API in later DXE phase to append EFI_MEMORY_SP or EFI_MEMORY_CPU_CRYPTO (keep the original attributes) 
3. System will have page fault exception

### Build Environment

```markdown
- OS(s):Windows 11
- Tool Chain(s): VS2019
```

### Version Information

```text
Commit: f979f51157c5560c658537ea2a488b7a8439e463
```

### Urgency

High

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

The root cause of this issue is that Edk2\MdeModulePkg\Core\Dxe\Misc\MemoryProtection.c (line 1294) directly uses CPU arch protocol to clear XP in page table when loading DXE driver for execution, but it fails to update the GCD internal data structure according (the memory range still makes as XP attribute):
gCpu->SetMemoryAttributes (gCpu, Memory, Length, NewAttributes);

Later invoking of GCD SetMemorySpaceAttributes () will incorrectly re-apply the incorrect XP attribute and make it take effect in the final page table. When CPU executes the memory, it triggers the page fault exception.

os-d:
# Description

Today, SetUefiImageMemoryAttributes calls directly to the CPU Arch protocol to set EFI_MEMORY_XP or EFI_MEMORY_RO on image memory. However, this bypasses the GCD and so the GCD is out of sync with the actual state of memory.

This can cause an issue in the scenario where a new attribute is being set (whether a virtual attribute or a real HW attribute), if the GCD attributes are queried for a region and the new attribute is appended to the existing GCD attributes (which are incorrect), then the incorrect attributes can get applied. This can result in setting EFI_MEMORY_XP on code sections of images and causing an execution fault.

This patch updates SetUefiImageMemoryAttributes to call into the GCD to update the attributes there and let the GCD code call into the CPU Arch protocol to update the page table.

While here, do a couple small updates in the function to improve error handling and logging.

Closes #10771 .

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Booted OVMF and ArmVirt to Windows and Linux with the change; used the repro scenario to confirm a boot to Windows occurs.

## Integration Instructions

N/A.

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
@@ -200,7 +200,7 @@ SetUefiImageMemoryAttributes (
 
   FinalAttributes = (Descriptor.Attributes & EFI_CACHE_ATTRIBUTE_MASK) | (Attributes & EFI_MEMORY_ATTRIBUTE_MASK);
 
-  DEBUG ((DEBUG_INFO, "SetUefiImageMemoryAttributes - 0x%016lx - 0x%016lx (0x%016lx)\n", BaseAddress, Length, FinalAttributes));
+  DEBUG ((DEBUG_VERBOSE, "SetUefiImageMemoryAttributes - 0x%016lx - 0x%016lx (0x%016lx)\n", BaseAddress, Length, FinalAttributes));
 
   ASSERT (gCpu != NULL);
   gCpu->SetMemoryAttributes (gCpu, BaseAddress, Length, FinalAttributes);
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
@@ -194,16 +194,108 @@ SetUefiImageMemoryAttributes (
   EFI_STATUS                       Status;
   EFI_GCD_MEMORY_SPACE_DESCRIPTOR  Descriptor;
   UINT64                           FinalAttributes;
+  UINT64                           CurrentAddress;
+  UINT64                           CurrentLength;
 
-  Status = CoreGetMemorySpaceDescriptor (BaseAddress, &Descriptor);
-  ASSERT_EFI_ERROR (Status);
+  CurrentAddress = BaseAddress;
 
-  FinalAttributes = (Descriptor.Attributes & EFI_CACHE_ATTRIBUTE_MASK) | (Attributes & EFI_MEMORY_ATTRIBUTE_MASK);
+  // we loop here because we may have multiple memory space descriptors that overlap the requested range
+  // this will definitely be the case for unprotecting an image, because that calls this function for the entire image,
+  // which we split into different GCD descriptors when we protected it.
+  while (CurrentAddress < BaseAddress + Length) {
+    Status = CoreGetMemorySpaceDescriptor (CurrentAddress, &Descriptor);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((
+        DEBUG_ERROR,
+        "%a - Failed to get memory space descriptor for address %llx with status %r. Cannot protect image.\n",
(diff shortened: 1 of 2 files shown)


os-d:
@qhuang8 , can you confirm that this patch resolves the crash you see?

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
```diff
@@ -196,14 +196,41 @@ SetUefiImageMemoryAttributes (
   UINT64                           FinalAttributes;
 
   Status = CoreGetMemorySpaceDescriptor (BaseAddress, &Descriptor);
-  ASSERT_EFI_ERROR (Status);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a - Failed to get memory space descriptor for address %llx with status %r. Cannot protect image.\n",
```
makubacki, line 202: Missing `__func__`?

os-d: Thanks, fixed.

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
```diff
@@ -196,14 +196,41 @@ SetUefiImageMemoryAttributes (
...
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a failed on %llx of length %llx with attributes %llx - %r\n",
+      Status,
+      __func__,
+      BaseAddress,
+      Length,
+      FinalAttributes
+      ));
```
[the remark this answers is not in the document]
os-d: Thanks, fixed.

lgao4 (approved):

lgao4:
@os-d , the code change looks good. But, I don't understand what's wrong in CI.

os-d:
> @os-d , the code change looks good. But, I don't understand what's wrong in CI.

@lgao4 thanks for the review! This was my bad, I didn't set the capabilities first if required. I have updated the PR to do this and tested one of the failing cases locally and it now passes. Confirming with the original reproer that this fixes their issue.

os-d:
@lgao4 can you please re-review the PR? I have made significant changes since your initial review to handle the non-aligned image case. The original reproer has confirmed it fixes the OVMF repro, though was having issues integrating into 2402 to test on a physical platform. I am testing on a physical system now.

lgao4:
> @lgao4 can you please re-review the PR? I have made significant changes since your initial review to handle the non-aligned image case. The original reproer has confirmed it fixes the OVMF repro, though was having issues integrating into 2402 to test on a physical platform. I am testing on a physical system now.

Thanksf for your update. New version patch is good to me.

PR #10968, opened 2025-04-15
Multiple packages: stop using EmbeddedPkg libfdt

leiflindholm:
Many, many, orbits of the sun, ago - an ancient version of libfdt was haphazardly smashed sideways into edk2 EmbeddedPkg. It was never intended for general purpose use, but it
escaped (as software does) and here we are.

Quite some time ago now, a proper, maintainable integration of libfdt was added as MdePkg BaseFdtLib. However, as these things go, people found old code, and used that for guidance, and kept using the thing that should not be.

This does not yet delete the old version, giving a (short) grace period under which platform ports can override which library is used on a per-module basis.

The non-bisectable changes across DynamicTablesPkg/EmbeddedPkg/ArmVirtPkg/OvmfPkg have now been squashed into a single commit.

- [x] Breaking change?
  - This breaks all existing users of EmbeddedPkg FdtLib.
- [x] Impacts security?
  - It stops people from using a library that does not receive security updates.

## How This Was Tested

Compile-tested only.

## Integration Instructions

For proper migration:

```
-  FdtLib|EmbeddedPkg/Library/FdtLib/FdtLib.inf
+  FdtLib|MdePkg/Library/BaseFdtLib/BaseFdtLib.inf

```

```
-#include <libfdt.h>
+#include <Library/FdtLib.h>

```

and then update all the `fdt_` identifiers to the corresponding `Fdt` one.

If your experience matches mine at all, you may find that that the old libfdt was pulling in certain dependencies existing modules had missed to declare explicitly, and will now make themselves known.

For a short stop-gap, you can enable support for the deprecated variant on idividual modules:

```
[Components]
  MyPackage/MyModule/MyModule.inf {
    FdtLib|EmbeddedPkg/Library/FdtLib/FdtLib.inf
  }
```


ardbiesheuvel (approved):

samimujawar (approved):

leiflindholm:
As I stated in the description, this was compile-tested only and testing shows it is currently broken. Working on the fix, and will flip out of draft when done.

vlsunil (approved):
Thanks!

DynamicTablesPkg/Library/FdtHwInfoParserLib/Arm/ArmFdtHwInfoParser.c
```diff
@@ -5,6 +5,7 @@
   SPDX-License-Identifier: BSD-2-Clause-Patent
```
ardbiesheuvel, line 5: Typo in the patch title

gguo11837463 (approved):

Benny3345678 (approved):

shuoliu0 (approved):

shuoliu0 (commented):
It would be helpful if add some comment on fdt_get_property and fdt_get_prop's usage in commit message, e.g.

 * fdt_get_property() retrieves a pointer to the fdt_property
 * structure within the device tree blob corresponding to the property
 * named 'name' of the node at offset nodeoffset.
 
 * fdt_getprop() retrieves a pointer to the value of the property
 * named 'name' of the node at offset nodeoffset (this will be a
 * pointer to within the device blob itself, not a copy of the value).

leiflindholm:
> It would be helpful if add some comment on fdt_get_property and fdt_get_prop's usage in commit message, e.g.
> 
> ```
>  * fdt_get_property() retrieves a pointer to the fdt_property
>  * structure within the device tree blob corresponding to the property
>  * named 'name' of the node at offset nodeoffset.
>  
>  * fdt_getprop() retrieves a pointer to the value of the property
>  * named 'name' of the node at offset nodeoffset (this will be a
>  * pointer to within the device blob itself, not a copy of the value).
> ```

I sort of don't think the documentation of this belongs in the commit message, but I agree the situation is a bit confusing (it already tripped me up), so I added a brief summary of the differences to the commit message.

shuoliu0:
> > It would be helpful if add some comment on fdt_get_property and fdt_get_prop's usage in commit message, e.g.
> > ```
> >  * fdt_get_property() retrieves a pointer to the fdt_property
> >  * structure within the device tree blob corresponding to the property
> >  * named 'name' of the node at offset nodeoffset.
> >  
> >  * fdt_getprop() retrieves a pointer to the value of the property
> >  * named 'name' of the node at offset nodeoffset (this will be a
> >  * pointer to within the device blob itself, not a copy of the value).
> > ```
> 
> I sort of don't think the documentation of this belongs in the commit message, but I agree the situation is a bit confusing (it already tripped me up), so I added a brief summary of the differences to the commit message.

Thanks

PR #10974, opened 2025-04-16
ArmFfaLib: Replace SMCCC_VERSION check with FF-A version check 

mgamal:
# Description

Patch 8d03c42d38d1 ("ArmPkg: ArmFfaLib: Update FF-A direct message to support 18 registers") calls SMCCC_VERSION by default during FF-A initialization even if the SMC conduit is not enabled. This leads to sending malformed FF-A versions if the SVC conduit is used instead.

This is observed for example when using an SPMC@EL-1 (e.g. rust-spmc) which receives the FF-A message via an SVC call and interprets SMCCC_VERSION as an FF-A Function ID and subsequently fails to handle it.

Since 18-register support is only support with FF-A >= 1.2, replace
SMCCC_VERSION check with FF-A version check

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested locally with rust-spmc as an EL-1 SPMC

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
+++ b/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
@@ -754,30 +754,11 @@ ArmFfaLibCommonInit (
   IN VOID
   )
 {
-  EFI_STATUS    Status;
-  UINT16        CurrentMajorVersion;
-  UINT16        CurrentMinorVersion;
-  ARM_FFA_ARGS  FfaArgs;
+  EFI_STATUS  Status;
+  UINT16      CurrentMajorVersion;
+  UINT16      CurrentMinorVersion;
 
   gFfaSupported = FALSE;
-
-  ZeroMem (&FfaArgs, sizeof (ARM_SMC_ARGS));
-  FfaArgs.Arg0 = SMCCC_VERSION;
-  ArmCallFfa (&FfaArgs);
-  if ((INT32)FfaArgs.Arg0 < 0) {
-    DEBUG ((DEBUG_ERROR, "%a: SMCCC_VERSION not supported\n", __func__));
-    return EFI_UNSUPPORTED;
-  }
-
-  // According to SMCCC Specification v1.6 G BET0
-  // Table F0-1: Changelog: Starting from SMCCC_VERSION v1.2, the interface
-  // - Permits calls to use R4–R7 as return register
-  // - Permits calls to use X4–X17 as return registers
-  // - Permits calls to use X8–X17 as argument registers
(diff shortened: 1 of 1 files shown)


kuqin12:
Is there any way to query SMCCC version even through SVC conduit? Not checking the version will have the rest of the program bear the assumption that 18 registers can be used for FFA. Is EL-1 SPMC compatible with 18 register usage?

kuqin12:
> > Is there any way to query SMCCC version even through SVC conduit? Not checking the version will have the rest of the program bear the assumption that 18 registers can be used for FFA. Is EL-1 SPMC compatible with 18 register usage?
> 
> As far as I know, SMCCC version querying can be done through either SMC or HVC calls, but not SVC calls. I think 18 register usage is supported by default in FF-A >= 1.2, so if there was no special reason to specifically query the SMCCC version, I guess we can replace it with an FF-A version check.
> 
> An SPMC that supports FF-A >= 1.2 should support 18 register usage.
> 
> @samimujawar Any thoughts on this?

18 register is needed in FF-A >= 1.2. Moving to query version and require it to be higher than v1.2 will be a tighter restriction than just checking on the SMCCC version. I do not have concern about it. But I am not sure about other SPMC configurations at large.

mgamal:
In the meantime there is no way to check SMCCC Version with SVC calls. Only SMC and HVC calls are supported for this prupose, which means that a SPMC @ EL-1 can not use this calls AFAIK. So we still need to make sure that the call is done when the SMC conduit is used.

mgamal:
Looking back at this email thread
https://edk2.groups.io/g/devel/topic/111543575#msg121176

it seems that we shouldn't have a use case for 18-register support with FF-A < 1.2, even if SMCCC_VERSION >= 1.2. So I guess we can replace the SMCCC version check with an FF-A version check.

@kuqin12 @samimujawar Any thoughts on this?

kuqin12:
> Looking back at this email thread https://edk2.groups.io/g/devel/topic/111543575#msg121176
> 
> it seems that we shouldn't have a use case for 18-register support with FF-A < 1.2, even if SMCCC_VERSION >= 1.2. So I guess we can replace the SMCCC version check with an FF-A version check.
> 
> @kuqin12 @samimujawar Any thoughts on this?

I do not have concerns about using FFA version v1.2 as the criteria for proceeding.

mgamal:
@kuqin12 @samimujawar I've updated the patch to replace SMCCC version checking with FF-A version. Please take a look at it and let me know if you have any comments.

kuqin12 (approved):

samimujawar (approved):

PR #10976, opened 2025-04-16
Fix smm boot

gdong1:
# Description

This PR adds missing libraries required by SMM (SMM_SUPPORT=TRUE). and it also fixed a typo in SMRR.
It also adds memory type information HOB build for ELF format universal payload to make it consistent
with other UEFI payloads (using FD and FIT format).

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Tested build and boot on different platforms.

With this change, SMM variable works as expected from UEFI shell. and also OS boot is fine.

## Integration Instructions

### DIFF
--- a/UefiPayloadPkg/UefiPayloadEntry/UniversalPayloadEntry.c
+++ b/UefiPayloadPkg/UefiPayloadEntry/UniversalPayloadEntry.c
@@ -4,7 +4,7 @@
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
-
+#include <Guid/MemoryTypeInformation.h>
 #include "UefiPayloadEntry.h"
 
 #define MEMORY_ATTRIBUTE_MASK  (EFI_RESOURCE_ATTRIBUTE_PRESENT             |        \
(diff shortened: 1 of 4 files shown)


shuoliu0 (approved):

ChaselChiu (approved):

UefiPayloadPkg/UefiPayloadEntry/UniversalPayloadEntry.c
```diff
@@ -482,6 +491,15 @@ _ModuleEntryPoint (
   Status = BuildHobs (BootloaderParameter, &DxeFv);
   ASSERT_EFI_ERROR (Status);
 
+  //
+  // Create Memory Type Information HOB
+  //
+  BuildGuidDataHob (
```
ChaselChiu, line 497: I think we should check if this HOB has been passed by bootloader before creating the default one:
  //
  // To create Memory Type Information HOB
  //
  GuidHob = GetFirstGuidHob (&gEfiMemoryTypeInformationGuid);
  if (GuidHob == NULL) {
    BuildGuidDataHob (
      &gEfiMemoryTypeInformationGuid,
      mDefaultMemoryTypeInformation,
      sizeof (mDefaultMemoryTypeInformation)
      );
  }

gdong1: Thanks for the comments. it was updated.

PR #10977, opened 2025-04-17
MdeModulePkg: Shortcircuit GCD Dumping Logic if Not Printing

os-d:
# Description

CoreDumpGcdMemorySpaceMap() gets called on every update to the GCD, but it only prints if DEBUG_GCD is set. However, the compiler is not smart enough to remove all of this logic if we are not printing anything, so we end up needlessly allocating memory for the copy of the map and spending many cycles looping through each entry, only to not print anything.

This code is compiled out on release builds, but slows down debug builds that aren't printing at DEBUG_GCD level.

This patch updates CoreDumpGcdMemorySpaceMap() to shortcircuit and immediately exit if DEBUG_GCD is not set. It also adds the same logic to CoreDumpGcdIoSpaceMap(), which is called less frequently, but has the same issue.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested by booting to shell.

## Integration Instructions

N/A.

### DIFF
--- a/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
+++ b/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
@@ -153,6 +153,13 @@ CoreDumpGcdMemorySpaceMap (
   EFI_GCD_MEMORY_SPACE_DESCRIPTOR  *MemorySpaceMap;
   UINTN                            Index;
 
+  // The compiler is not smart enough to compile out the whole function if DEBUG_GCD is not enabled, so we end up
+  // looping through the GCD every time it gets updated, which wastes a lot of needless cycles if we aren't going to
+  // print it. So shortcircuit and jump out if we don't need to print it.
+  if (!DebugPrintLevelEnabled (DEBUG_GCD)) {
+    return;
+  }
+
   Status = CoreGetMemorySpaceMap (&NumberOfDescriptors, &MemorySpaceMap);
   ASSERT (Status == EFI_SUCCESS && MemorySpaceMap != NULL);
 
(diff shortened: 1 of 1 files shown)


mdkinney:
This function is wrapped by DEBUG_CODE_BEGIN() and DEBUG_CODE_END().  Disabling DEBUG_CODE should have the same effect.

os-d:
> This function is wrapped by DEBUG_CODE_BEGIN() and DEBUG_CODE_END(). Disabling DEBUG_CODE should have the same effect.

@mdkinney, correct in a release build this logic is skipped. However, in a debug build where you aren't printing DEBUG_GCD (which is the usual configuration for development) you end up wasting a lot of time in this function getting the GCD (which has to go through every entry in the GCD to copy it) then going through each copied entry, only to end up deciding not to print it.

So this patch is intended to save time on DEBUG builds where you still want DEBUG_CODE to execute, but you aren't printing the GCD (which again is the majority of cases).

os-d:
@lgao4 can you review this PR, please?

mdkinney:
> > This function is wrapped by DEBUG_CODE_BEGIN() and DEBUG_CODE_END(). Disabling DEBUG_CODE should have the same effect.
> 
> @mdkinney, correct in a release build this logic is skipped. However, in a debug build where you aren't printing DEBUG_GCD (which is the usual configuration for development) you end up wasting a lot of time in this function getting the GCD (which has to go through every entry in the GCD to copy it) then going through each copied entry, only to end up deciding not to print it.
> 
> So this patch is intended to save time on DEBUG builds where you still want DEBUG_CODE to execute, but you aren't printing the GCD (which again is the majority of cases).

That make sense to me.  This is a good change

mdkinney (approved):

lgao4 (approved):

PR #10978, opened 2025-04-18
MdeModulePkg: Add setup variable save callback GUID

mark-skum:
When saving variables in the setup menu, a callback event is required to handle project-specific operations.

For example:
On ARM platforms, we need to save DDR parameters to SPI flash. During the next boot, the system processor will read these parameters to initialize the DDR.


mark-skum:
@dandanbi @lgao4 Can you please help to review this PR? Thanks you!

lgao4:
Can you use EFI_BROWSER_ACTION_SUBMITTED action type for this purpose? This action type will be trigged when EfiVarstore setup variable is submitted.

mark-skum:
> Can you use EFI_BROWSER_ACTION_SUBMITTED action type for this purpose? This action type will be trigged when EfiVarstore setup variable is submitted.
Hi lgao4,
Thanks for you reply.
The function of this callback is to ensure that after the variable is written into nvram, it can be read back through getvariable,But,EFI_BROWSER_ACTION_SUBMIT cannot be achieved.

lgao4:
> > Can you use EFI_BROWSER_ACTION_SUBMITTED action type for this purpose? This action type will be trigged when EfiVarstore setup variable is submitted.
> > Hi lgao4,
> > Thanks for you reply.
> > The function of this callback is to ensure that after the variable is written into nvram, it can be read back through getvariable,But,EFI_BROWSER_ACTION_SUBMIT cannot be achieved.

HII CallBack function can be trigged after the efi variable is submitted. The callback function can read the efi variable.

PR #10981, opened 2025-04-18
MdeModulePkg: PiSmmIpl: Fix physical address dereferencing

kuqin12:
# Description

With current implemenation, all 3 SmmCommunication* functions go through the same routine, which will dereference the incoming pointer to inspect whether this is a V3 buffer or not.

However, the caller always pass in the physical addresses, which could cause the system to page fault after OS take over the runtime control.

This change refactored the common routine to accept both virtual and physical address.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This change is tested on OVMF package and booted to Windows.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
+++ b/MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
@@ -550,12 +550,10 @@ SmmCommunicationCommunicate (
   IN OUT UINTN                             *CommSize OPTIONAL
   )
 {
-  EFI_STATUS                    Status;
-  EFI_SMM_COMMUNICATE_HEADER    *CommunicateHeader;
-  EFI_MM_COMMUNICATE_HEADER_V3  *CommunicateHeaderV3 = NULL;
-  BOOLEAN                       OldInSmm;
-  UINTN                         TempCommSize;
-  UINTN                         CommHeaderSize;
+  EFI_STATUS                  Status;
+  EFI_SMM_COMMUNICATE_HEADER  *CommunicateHeader;
+  BOOLEAN                     OldInSmm;
+  UINTN                       TempCommSize;
 
   //
   // Check parameters
@@ -565,28 +563,17 @@ SmmCommunicationCommunicate (
   }
 
   CommunicateHeader = (EFI_SMM_COMMUNICATE_HEADER *)CommBuffer;
-  if (CompareGuid (&CommunicateHeader->HeaderGuid, &gEfiMmCommunicateHeaderV3Guid)) {
-    CommunicateHeaderV3 = (EFI_MM_COMMUNICATE_HEADER_V3 *)CommBuffer;
-    if (CommunicateHeaderV3->BufferSize < sizeof (EFI_MM_COMMUNICATE_HEADER_V3) + CommunicateHeaderV3->MessageSize) {
+
+  if (CommSize == NULL) {
+    TempCommSize = OFFSET_OF (EFI_SMM_COMMUNICATE_HEADER, Data) + CommunicateHeader->MessageLength;
+  } else {
+    TempCommSize = *CommSize;
+    //
+    // CommSize must hold HeaderGuid and MessageLength
+    //
+    if (TempCommSize < OFFSET_OF (EFI_SMM_COMMUNICATE_HEADER, Data)) {
       return EFI_INVALID_PARAMETER;
     }
-
-    TempCommSize   = (UINTN)CommunicateHeaderV3->BufferSize;
-    CommHeaderSize = sizeof (EFI_MM_COMMUNICATE_HEADER_V3);
-  } else {
-    if (CommSize == NULL) {
-      TempCommSize = OFFSET_OF (EFI_SMM_COMMUNICATE_HEADER, Data) + CommunicateHeader->MessageLength;
-    } else {
-      TempCommSize = *CommSize;
-      //
-      // CommSize must hold HeaderGuid and MessageLength
-      //
-      if (TempCommSize < OFFSET_OF (EFI_SMM_COMMUNICATE_HEADER, Data)) {
-        return EFI_INVALID_PARAMETER;
-      }
-    }
-
-    CommHeaderSize = OFFSET_OF (EFI_SMM_COMMUNICATE_HEADER, Data);
   }
 
   //
@@ -643,20 +630,16 @@ SmmCommunicationCommunicate (
   //
   // Before SetVirtualAddressMap(), we are in SMM or SMRAM is open and unlocked, call SmiManage() directly.
   //
-  TempCommSize -= CommHeaderSize;
+  TempCommSize -= OFFSET_OF (EFI_SMM_COMMUNICATE_HEADER, Data);
   Status        = gSmmCorePrivate->Smst->SmiManage (
                                            &CommunicateHeader->HeaderGuid,
                                            NULL,
                                            CommunicateHeader->Data,
                                            &TempCommSize
                                            );
-  TempCommSize += CommHeaderSize;
+  TempCommSize += OFFSET_OF (EFI_SMM_COMMUNICATE_HEADER, Data);
   if (CommSize != NULL) {
     *CommSize = TempCommSize;
-  }
-
-  if (CommunicateHeaderV3 != NULL) {
-    CommunicateHeaderV3->BufferSize = TempCommSize;
   }
 
   //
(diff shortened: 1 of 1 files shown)


mdkinney:
How much as the V3 path been tested?  I see an issues with 

```
  //
  // Before SetVirtualAddressMap(), we are in SMM or SMRAM is open and unlocked, call SmiManage() directly.
  //
  TempCommSize -= CommHeaderSize;
  Status        = gSmmCorePrivate->Smst->SmiManage (
                                           &CommunicateHeader->HeaderGuid,
                                           NULL,
                                           CommunicateHeader->Data,
                                           &TempCommSize
                                           );
```'

When V3 is detected, I think it needs to pass in the MessageGuid and MessageData from the V3 comm buffer.

File: MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c [hunk 1]
```diff
@@ -560,13 +562,13 @@ SmmCommunicationCommunicate (
   //
   // Check parameters
   //
-  if (CommBuffer == NULL) {
+  if (CommBufferVirtual == NULL) {
```
mdkinney, line 565: Should check CommonBufferPhysical too.

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
```diff
@@ -667,6 +669,48 @@ SmmCommunicationCommunicate (
...
+SmmCommunicationCommunicate (
+  IN CONST EFI_SMM_COMMUNICATION_PROTOCOL  *This,
+  IN OUT VOID                              *CommBuffer,
+  IN OUT UINTN                             *CommSize OPTIONAL
+  )
+{
+  return SmmCommunicationCommunicateCommon (
+            &mSmmCommunication,
+            CommBuffer,
+            CommBuffer,
```
mdkinney, line 709: This does not resolve this issue.  CommBuffer is always a physical address.  By passing the physical address both the physical and virtual address, the use of the physical address to check the V3 GUID is still used.

kuqin12: Yup.. I just replied with an email. This change will not resolve the issue. I do have one proposal through email. Please let me know how you think about it.

kuqin12:
> How much as the V3 path been tested? I see an issues with
> 
> ```
>   //
>   // Before SetVirtualAddressMap(), we are in SMM or SMRAM is open and unlocked, call SmiManage() directly.
>   //
>   TempCommSize -= CommHeaderSize;
>   Status        = gSmmCorePrivate->Smst->SmiManage (
>                                            &CommunicateHeader->HeaderGuid,
>                                            NULL,
>                                            CommunicateHeader->Data,
>                                            &TempCommSize
>                                            );
> ```'
> 
> When V3 is detected, I think it needs to pass in the MessageGuid and MessageData from the V3 comm buffer.
> ```

I did test v3 protocol. But I think the logic you mentioned never hit during my test.. My pass always went through MM control trigger. I will see if there is some change I can do here.

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
kuqin12: Updated.

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
```diff
@@ -525,7 +525,7 @@ SmmBase2GetSmstLocation (
   after SetVirtualAddressMap().
 
   @param[in] This                The EFI_SMM_COMMUNICATION_PROTOCOL instance.
-  @param[in, out] CommBuffer     A pointer to the buffer to convey into SMRAM.
+  @param[in, out] CommBuffer     A pointer to the virtual address of the buffer to convey into SMRAM.
```
mdkinney, line 528: This change should be removed to match original.  CommBuffer has to be physical.

mdkinney:
I have tested this change on systems that were failing and this change has resolved the issues,

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
kuqin12: Right, forgot to push the last change. Should be updated now.

mdkinney (approved):

mdkinney:
@niruiyu and @lgao4 Please review. This is a critical bug fix that is required to address a regression.

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
```diff
@@ -740,11 +723,98 @@ MmCommunicationMmCommunicate3 (
...
+  UINT64                        TempCommSize;
+
+  //
+  // Check parameters
+  //
+  if ((CommBufferPhysical == NULL) || (CommBufferVirtual == NULL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  CommunicateHeader = (EFI_MM_COMMUNICATE_HEADER_V3 *)CommBufferVirtual;
```
jiaxinwu, line 739: *1. This patch confuses me the design/implementation of Communicate3 and Communicate2, looks the usage of those two Communication ways are different:
For V3, CommunicateHeader uses a virtual address, while gSmmCorePrivate->CommunicationBuffer uses a physical address.
For V2, the entire process operates is to use the physical address space.
Why has such implementation difference? can you explain in detail in the commit message?


*2. This patch looks has 2 kind of fix:
First one is about the input parament of gSmmCorePrivate->Smst->SmiManage() fix, which is incorrect for the V3.
Second one is about the virtual address/physical address usage difference for the V3 and V2/V1.
Is it possible to make it clear? which one make the issue happen? why it happens? how it fix the issue? please document in the commit message clearly.

niruiyu: Jiaxin,
Answer to question 1: because v2 logic only accesses the MessageLength field in the comm buffer when the CommSize is NULL. Luckily all runtime calls to V1 and V2 supply the !NULL CommSize.

mdkinney: Yes.  2 changes.

One is to revert the V3 change so V1 and V2 are exactly as they were before V3 was introduced.  Yes, the V1 and V2 implementations are confusing and will need further review and either code changes or spec changes or both to fix.  This PR is not fixing V1 or V2.  Just restoring their previous behavior.  OS boot failures were observed, and putting V1 and V2 back fixes those issues.

The 2nd change was in V3 only for the SmiManage() call.  It was incorrect.  Passing in the wrong GUID and the wrong message data.  The SmiManage() path with V3 has never been tested which is why any bugs from it had not been observed yet.

mdkinney: I agree the commit message is incorrect,  The common routine approach was set aside and a new implementation for V3 was selected instead.

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
```diff
@@ -565,28 +563,17 @@ SmmCommunicationCommunicate (
...
-  if (CompareGuid (&CommunicateHeader->HeaderGuid, &gEfiMmCommunicateHeaderV3Guid)) {
-    CommunicateHeaderV3 = (EFI_MM_COMMUNICATE_HEADER_V3 *)CommBuffer;
-    if (CommunicateHeaderV3->BufferSize < sizeof (EFI_MM_COMMUNICATE_HEADER_V3) + CommunicateHeaderV3->MessageSize) {
-      return EFI_INVALID_PARAMETER;
-    }
 
-    TempCommSize   = (UINTN)CommunicateHeaderV3->BufferSize;
-    CommHeaderSize = sizeof (EFI_MM_COMMUNICATE_HEADER_V3);
+  if (CommSize == NULL) {
+    TempCommSize = OFFSET_OF (EFI_SMM_COMMUNICATE_HEADER, Data) + CommunicateHeader->MessageLength;
```
niruiyu, line 568: bug here but it's an old bug. Agree not to fix in this patch.

niruiyu (approved):
Can you please submit a bug fix to the runtime communication driver in StandaloneMmPkg also?

updated: The StandaloneMmPkg does not have the bug.

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
jiaxinwu: thanks Ray and Mike, i got why the issue happen. I'm ok with the fix in V2 in another new PR. Then, please update the commit message to make sure the cause and effect and the corresponding changes are clear.

kuqin12:
> Can you please submit a bug fix to the runtime communication driver in StandaloneMmPkg also?
> 
> updated: The StandaloneMmPkg does not have the bug.

@niruiyu I tested the standalone mm version and observed no issue there. That instance of MM communicate (v1/2/3) is always using virtual address. That driver has its own issue for v1 communicate as it always touches the incoming buffer, but as Mike mentioned on the email thread, I think we should instead deprecate v1 or rework all instances to not touch incoming buffers if we need to stick closely to the spec.

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
kuqin12: I will update the commit message shortly.

kuqin12: The commit message is updated in the latest push.

mdkinney (approved):

jiaxinwu (approved):

PR #10992, opened 2025-04-22
MdePkg/Library/BaseRngLib/Riscv: use CPU RNG instructions only

0n41rd4:
Generate random number using risc-v CPU RNG instructions only, as for the other architectures. This is safer than using Mersenne Twister from a seed, which is the current implementation, as the latter is known to be cryptographically unsafe (e.g. see https://www.schutzwerk.com/en/blog/attacking-a-rng/)

# Description

See above

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Will run CI

## Integration Instructions

N/A

### DIFF
--- a/MdePkg/Library/BaseRngLib/Riscv/Rng.c
+++ b/MdePkg/Library/BaseRngLib/Riscv/Rng.c
@@ -18,75 +18,7 @@
 
 #define SEED_RETRY_LOOPS  100
 
-// 64-bit Mersenne Twister implementation
-// A widely used pseudo random number generator. It performs bit shifts etc to
-// achieve the random number. It's output is determined by SEED value generated
-// by RISC-V SEED CSR"
-
-#define STATE_SIZE  312
-#define MIDDLE      156
-#define INIT_SHIFT  62
-#define TWIST_MASK  0xb5026f5aa96619e9ULL
-#define INIT_FACT   6364136223846793005ULL
-#define SHIFT1      29
-#define MASK1       0x5555555555555555ULL
-#define SHIFT2      17
-#define MASK2       0x71d67fffeda60000ULL
-#define SHIFT3      37
-#define MASK3       0xfff7eee000000000ULL
-#define SHIFT4      43
-
-#define LOWER_MASK  0x7fffffff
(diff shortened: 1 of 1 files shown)


0n41rd4:
Hi there, did you have a chance to look at this?

lgao4:
@vlsunil , can you review this change?

vlsunil:
Thanks Liming. Somehow original PR didn't copy me.

It looks good to me.  @dhaval-rivos : Could you please review this as well?

dhaval-rivos:
I am not sure completely removing DRBG and solely relying on CSR based entropy is what the spec recommends. " Expected usage is to condition (typically with SHA-2/3) the output from an entropy source and use it to seed a cryptographically secure Deterministic Random Bit Generator (DRBG) such as AES-based CTR_DRBG cite:[BaKe15]. The combination of an Entropy Source, Conditioning, and a DRBG can be used to create random bits securely cite:[BaKeRo:21]. ". If at all there is a concern with DRBG implementation namely Mersenne Twister, it should be replaced with something more secure, would be the right approach.

0n41rd4:
@dhaval-rivos I agree that CSR based entropy can be improved, but this library is supposed to provide that:

#  BaseRng Library that uses CPU RNG instructions (e.g. RdRand) to
#  provide random numbers.

And this is how this library works for arm and amd too. If an algorithm is desired to generate random numbers on top of what this library provides, then a cryptographically safe one could be used insead. E.g. ChaCha20, but using a vulnerable algorithm is not safe

xypron:
MdePkg/Library/BaseRngLib/AArch64 provides fresh entropy from the CPU RNG for every invocation of ArchGetRandomNumber64. I would not know why RISC-V should handle this differently. Using a PNRG here is unexpected.

What stopped the merge request?

vlsunil:
Sorry I missed tracking it and the bot closed it after inactivity. I agree with @0n41rd4.

 @dhaval-rivos : I hope you are fine with the explanation from @0n41rd4.

vlsunil (approved):

PR #10996, opened 2025-04-23
Ovmf memory debug logging2

ajyoung-oracle:
# Description

OVMF Memory Debug Logging Support

Overview:

This PR adds support to log OVMF debug messages to a memory buffer. The memory buffer can be extracted from a VM's QEMU process (or a core file) to debug issues.

Background:

OVMF currently offers 2 methods to collect debug messages - via the "debug IO port" (logged by QEMU to a host disk file) and via the serial port (-D DEBUG_ON_SERIAL_PORT). Both of these methods are impractical under normal VM operation. They increase VM boot time, clutter the serial port, and use host disk resources. In contrast, logging the debug messages to memory is fast and takes no disk resources allowing it to be enabled by default for customer environments. This allows for "first-time" issue analysis, negating the need to reproduce issues with debug messages enabled, which makes for a much better customer experience should an issue occur.

Code Overview:

The code introduces 3 new libraries:

  * OvmfPkg/Library/PlatformDebugLibMem: New DebugLib class library which logs debug messages to memory. This library was derived from PlatformDebugLibIoPort - simply altered to pass the debug message to MemDebugLogWrite() instead of writing the debug message to the debug IO port.
   
  * OvmfPkg/Library/MemDebugLogCommonLib: Library to manage access to the memory debug log circular buffer.
    
  * OvmfPkg/Library/MemDebugLogLib: Library which implements the key MemDebugLogWrite() function (called by PlatformDebugLibMem). There is a different implementation of this library for the various boot phases/module types (i.e. SEC, PEIM, DXE, Runtime). See more detail below.

The code also introduces a new PEIM:

  * OvmfPkg/MemDebugLogPei: When memory becomes available, this PEIM allocates the main memory debug log buffer and installs a PPI containing a function to write to the main memory buffer.

A bit of complication arises from the the fact that main memory is not available until the PEI phase and several debug messages are logged before then. To remedy this, the code makes use of a "early" memory buffer (taken from PeiMemFvBase) which is used to log the initial (pre-memory) messages. Once memory becomes available, the PEIM code allocates the main memory debug log buffer, copies the messages from the early buffer and then switches to use the main memory buffer (by installing a PPI/HOB).

Thus, the SEC version of MemDebugLogLib writes debug messages to the "early" debug log buffer. The PEIM version of MemDebugLogLib also writes initial messages to the early buffer until the MemDebugLogPei PPI is installed. Once the PPI is installed the PEI version of MemDebugLogLib switches to use the PPI. The DXE and runtime versions of the MemDebugLogLib write to the main memory debug log buffer (obtaining the address of the buffer via a HOB). The runtime version of MemDebugLogLib (used by DXE_RUNTIME_DRIVER type drivers) will allow debug writes until the OS makes the SetVirtualAddressMap() BS call.

All calls to MemDebugLogWrite() eventually end up calling MemDebugLogWriteCommon() (from MemDebugLogCommonLib) - which handles the details of maintaining the circular debug log buffer and header (with head/tail pointers, etc.). Since it is theoretically possible for multiple vcpus to attempt to write debug messages simultaneously (during vcpu init), the library uses a spinlock to protect the critical sections when accessing the buffer.

The feature (i.e. the new libraries and PEIM) are disabled by default and are enabled via the new "DEBUG_TO_MEM" build flag (which can be enabled on the build command line - similar to DEBUG_ON_SERIAL_PORT).

Notes:

  The debug log memory buffer size can be configured via FwCfg and is circular - i.e. only the most recent debug messages are retained if the memory buffer overflows. This is appropriate as typically only the most recent debug messages are relevant when debugging an issue. The code currently defaults to a 32 page (128K) sized memory debug log buffer (the default is configured via a PCD).

  A host-side tool/utility can be easily implemented to search the VM's QEMU memory regions to locate the OVMF memory debug log buffer (located on a page boundary), decipher the buffer header and display the circular buffer contents (debug messages). We (Oracle) already have such a utility which can extract the OVMF memory debug log from a live QEMU process or a QEMU core file.
    
  This feature doesn't work with Confidential Computing VMs as the guest memory is encrypted and thus not readable from the host. A future enhancement could possibly be made to mark the OVMF memory debug log buffer as unencrypted (?) TBD
    
  This PR only covers OVMF (x86_64) but was written to be easily extended to AAVMF (Arm) in a future PR. TBD

- [x] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

OVMF builds for both OvmfPkgIa32X64.dsc and OvmfPkgX64.dsc (both with and without -D DEBUG_TO_MEM) were built/tested and the OVMF Memory Debug Log extracted from the VM's QEMU process (by a custom utility) and assessed for correctness.

## Integration Instructions

N/A

### DIFF
--- a/OvmfPkg/Include/Library/MemDebugLogCommonLib.h
+++ b/OvmfPkg/Include/Library/MemDebugLogCommonLib.h
@@ -1,0 +1,155 @@
+/** @file
+  Memory Debug Log common defs/funcs to access the memory buffer.
+
+  Copyright (C) 2025, Oracle and/or its affiliates.
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#ifndef __MEM_DEBUG_LOG_COMMON_H_
+#define __MEM_DEBUG_LOG_COMMON_H_
+
+#include <Uefi/UefiBaseType.h>
+
+//
+// Cap max buffer at 4MB (0x400 4K pages)
+//
+#define MAX_MEM_DEBUG_LOG_PAGES  0x400
+
+#define MEM_DEBUG_LOG_MAGIC1  0x646d656d666d766f  // "ovmfmemd"
+#define MEM_DEBUG_LOG_MAGIC2  0x6367616d67756265  // "ebugmagc"
+
+#pragma pack(1)
+//
+// Mem Debug Log buffer header.
+// The Log buffer is circular. Only the most
+// recent messages are retained. Older messages
+// will be discarded if the buffer overflows.
+// The Debug Log starts just after the header.
+// Versioning the structure allows for future expansion.
+//
+typedef struct {
+  //
+  // Magic values used by tools to locate the buffer in memory
+  // These MUST be the first two fields of the structure.
+  // Use a 128 bit Magic to make sure to avoid any possible
+  // collision with random data in memory.
+  UINT64             Magic1;
+  UINT64             Magic2;
+  //
+  // Structure version- MUST be third field and set to 1
+  //
+  UINT8              Version;
+  //
+  // Debug log size minus header
+  //
+  UINT32             DebugLogSize;
+  //
+  // Debug log head offset
+  //
+  UINT32             DebugLogHeadOffset;
+  //
+  //  Debug log tail offset
+  //
+  UINT32             DebugLogTailOffset;
+  //
+  // Protect the log from MP access
+  //
+  volatile UINT64    MemDebugLogLock;
+  //
+  // Flag to indicate if the buffer wrapped and was thus truncated.
+  //
+  UINT8              Truncated;
+  //
+  // Firmware Build Version (PcdFirmwareVersionString)
+  //
+  CHAR8              FirmwareVersion[32];
+} MEM_DEBUG_LOG_HDR_V1;
+
+//
+// HOB used to pass the mem debug log buffer addr from PEI to DXE
+//
+typedef struct {
+  EFI_PHYSICAL_ADDRESS    MemDebugLogBufAddr;
+} MEM_DEBUG_LOG_HOB_DATA;
+
+#pragma pack()
+
+/**
+  Write a CHAR8 string to the memory debug log circular buffer.
+
+  @param MemDebugLogBufAddr       Address of the memory debug log buffer.
+
+  @param Buffer                   Pointer to a CHAR8 string to write to the
+                                  debug log buffer.
+
+  @param Length                   Length of the CHAR8 string to write to the
+                                  debug log buffer. Not including NULL terminator
+                                  byte.
+
+  @retval RETURN_SUCCESS          String succcessfully written to the memory log buffer.
+
+  @retval RETURN_NOT_FOUND        Memory log buffer is not properly initialized.
+
+  @retval EFI_INVALID_PARAMETER   Invalid input parameters.
+**/
+EFI_STATUS
+EFIAPI
+MemDebugLogWriteCommon (
+  IN EFI_PHYSICAL_ADDRESS  MemDebugLogBufAddr,
+  IN CHAR8                 *Buffer,
+  IN UINTN                 Length
+  );
+
+/**
+  Initialize the memory debug log buffer header
+
+  @param MemDebugLogBufAddr       Address of the memory debug log buffer.
+
+  @param MemDebugLogBufSize       Size of the memory debug log buffer.
+
+  @retval RETURN_SUCCESS          Log buffer successfully initialized.
+
+  @retval EFI_INVALID_PARAMETER   Invalid input parameters.
+**/
+EFI_STATUS
+EFIAPI
+MemDebugLogInitCommon (
+  IN EFI_PHYSICAL_ADDRESS  MemDebugLogBufAddr,
+  UINT32                   MemDebugLogBufSize
+  );
+
+/**
+  Copy the memory debug log buffer
+
+  @param MemDebugLogBufDestAddr   Address of destination memory debug log buffer.
+
+  @param MemDebugLogBufSrcAddr    Address of source memory debug log buffer.
+
+  @retval RETURN_SUCCESS          Log buffer successfuly copied.
+
+  @retval EFI_INVALID_PARAMETER   Invalid input parameters.
+**/
+EFI_STATUS
+EFIAPI
+MemDebugLogCopyCommon (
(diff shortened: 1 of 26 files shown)


kraxel:
Any specific reason why this is splitted into three different libraries?  I don't think this is needed.  You can put all this into one directory and create three slightly different versions of the DebugLib library for SEC, PEI and DXE.  Look at `OvmfPkg/Library/QemuFwCfgLib` for example which has `QemuFwCfg{Sec,Pei,Dxe}Lib.inf`.  Some code is in sec/pei/dxe specific source file but most code is in shared source files which is linked into all library variants.

I think it makes sense to install an EFI configuration table describing the log buffer, so the operating system can find the log buffer too.  The linux kernel could easily export it via sysfs then.

I think it makes sense to support logging to *both* debugcon and memory.  OVMF detects whenever a debugcon device is present or not; most of the logging overhead goes away in case the device is not preset.  Having debugcon enabled in builds make sense IMHO.  Having memory logging clearly is useful too.

Do you have pointers to the log extraction utilities mentioned?

OvmfPkg/Include/Library/MemDebugLogCommonLib.h
```diff
@@ -0,0 +1,155 @@
...
+  // Magic values used by tools to locate the buffer in memory
+  // These MUST be the first two fields of the structure.
+  // Use a 128 bit Magic to make sure to avoid any possible
+  // collision with random data in memory.
+  UINT64             Magic1;
+  UINT64             Magic2;
+  //
+  // Structure version- MUST be third field and set to 1
+  //
+  UINT8              Version;
```
kraxel, line 42: I'd suggest to use a headersize field instead.  When extending the structure just append fields to the header.  Existing old software will be able to parse the parts of the structure they know about, providing better backward compatibility.

I'd also suggest to make sure all struct fields are aligned.  Unaligned field access is a PITA on !x86 architectures.

OvmfPkg/Include/Library/MemDebugLogCommonLib.h
```diff
@@ -0,0 +1,155 @@
...
+  //
+  UINT32             DebugLogHeadOffset;
+  //
+  //  Debug log tail offset
+  //
+  UINT32             DebugLogTailOffset;
+  //
+  // Protect the log from MP access
+  //
+  volatile UINT64    MemDebugLogLock;
```
kraxel, line 58: Needs a more verbose comment how this works.

OvmfPkg/Include/Library/MemDebugLogCommonLib.h
```diff
@@ -0,0 +1,155 @@
...
+  //
+  // Flag to indicate if the buffer wrapped and was thus truncated.
+  //
+  UINT8              Truncated;
+  //
+  // Firmware Build Version (PcdFirmwareVersionString)
+  //
+  CHAR8              FirmwareVersion[32];
```
kraxel, line 66: This is rather short, I'd suggest at least 64 bytes.  Especially test builds (with bug identifier and timestamp included in the version string) are easily longer than that.

ajyoung-oracle:
> Any specific reason why this is splitted into three different libraries? I don't think this is needed. You can put all this into one directory and create three slightly different versions of the DebugLib library for SEC, PEI and DXE. Look at `OvmfPkg/Library/QemuFwCfgLib` for example which has `QemuFwCfg{Sec,Pei,Dxe}Lib.inf`. Some code is in sec/pei/dxe specific source file but most code is in shared source files which is linked into all library variants.
> 
> 
> I think it makes sense to support logging to _both_ debugcon and memory. OVMF detects whenever a debugcon device is present or not; most of the logging overhead goes away in case the device is not preset. Having debugcon enabled in builds make sense IMHO. Having memory logging clearly is useful too.
> 
 
  Thanks for the review. Yes, I can combine the Libraries into one. Will do.
  
  I also like the idea to make it possible to write debug messages to both debugcon *and* memory simultaneously. How about serial too? Should I create a new OvmfPkg specific DebugLib that can write to any combination of debugcon, mem and serial? With debugcon as the default and mem and serial optionally selectable (via -D DEBUG_TO_MEM and -D DEBUG_ON_SERIAL_PORT respectively)? Seems that should be easy to do. What do you think?

OvmfPkg/Include/Library/MemDebugLogCommonLib.h
ajyoung-oracle: OK. Will do. Thanks

OvmfPkg/Include/Library/MemDebugLogCommonLib.h
ajyoung-oracle: Will do. Thanks

OvmfPkg/Include/Library/MemDebugLogCommonLib.h
ajyoung-oracle: Sure. Will do

ajyoung-oracle:
> I think it makes sense to install an EFI configuration table describing the log buffer, so the operating system can find the log buffer too. The linux kernel could easily export it via sysfs then.
> 

 OK. I think I'd prefer to address this in a separate subsequent PR if that's OK?

kraxel:
> I also like the idea to make it possible to write debug messages to both debugcon _and_ memory simultaneously. How about serial too? Should I create a new OvmfPkg specific DebugLib that can write to any combination of debugcon, mem and serial? With debugcon as the default and mem and serial optionally selectable (via -D DEBUG_TO_MEM and -D DEBUG_ON_SERIAL_PORT respectively)? Seems that should be easy to do. What do you think?

I think the interesting combinations are serial+memory (for non-qemu platforms) and debugcon+memory (qemu platforms).  Logging to both serial+debugcon looks pointless to me.  debugcon+memory is probably easiest to implement because debugcon is a OvmfPkg library anyway so we only need to change OvmfPkg for that.  The memory logging feature could be implemented as `MemoryLog` library and `PlatformDebugLibIoPort` could simply call into the libary to store a copy of the log message in the memory log buffer too.

Doing the same for serial+memory requires the `MemoryLog` library API (and Null implementation) being in MdePkg, so the same logic can be added to `BaseDebugLibSerialPort` without adding a OvmfPkg dependency to MdePkg.

kraxel:
> > I think it makes sense to install an EFI configuration table describing the log buffer, so the operating system can find the log buffer too. The linux kernel could easily export it via sysfs then.
> 
> OK. I think I'd prefer to address this in a separate subsequent PR if that's OK?

Fine with me.

PR #10998, opened 2025-04-23
ArmPlatformPkg: Restore LzmaDecompressLib to PeilessSec

os-d:
# Description

PeilessSec was created when PrePiUniCore was deprecated in these two commits:
[91117d70d89fbef4d7cac2d57854fa613716ca4d](https://github.com/tianocore/edk2/commit/91117d70d89fbef4d7cac2d57854fa613716ca4d) and
[5749b70b5a892e1cf0f2bfa37865a552ba8f96e5](https://github.com/tianocore/edk2/commit/5749b70b5a892e1cf0f2bfa37865a552ba8f96e5)

However, when PrePiUniCore was copied, it dropped LzmaDecompressLib from the library class list in the inf. This breaks platforms using this INF, as they are structured to have PeilessSec.inf in an FV with an LZMA compressed FV that contains DxeMain.inf, which is then executed next. Without LzmaDecompressLib's constructor run, no guided section extract handlers are registered for these sections and executes halts in SEC with an ASSERT.

This patch adds LzmaDecompressLib back to PeilessSec.inf and ArmPlatformPkg's DSC. With this change, platforms using PeilessSec.inf can boot again.

I also confirmed that this is not needed in ArmPlatformPkg/Sec/Sec.inf.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on a platform using PeilessSec.inf that was formerly using PrePiUniCore.inf that was failing to boot. Boot now succeeds.

## Integration Instructions

N/A.

### DIFF
--- a/ArmPlatformPkg/PeilessSec/PeilessSec.inf
+++ b/ArmPlatformPkg/PeilessSec/PeilessSec.inf
@@ -43,6 +43,7 @@
   DebugAgentLib
   DebugLib
   HobLib
+  LzmaDecompressLib
   MemoryInitPeiLib
   PerformanceLib
   PlatformPeiLib
--- a/ArmPlatformPkg/ArmPlatformPkg.dsc
+++ b/ArmPlatformPkg/ArmPlatformPkg.dsc
@@ -49,6 +49,7 @@
   IoLib|MdePkg/Library/BaseIoLibIntrinsic/BaseIoLibIntrinsic.inf
   LcdHwLib|ArmPlatformPkg/Library/LcdHwNullLib/LcdHwNullLib.inf
   LcdPlatformLib|ArmPlatformPkg/Library/LcdPlatformNullLib/LcdPlatformNullLib.inf
+  LzmaDecompressLib|MdeModulePkg/Library/LzmaCustomDecompressLib/LzmaCustomDecompressLib.inf
   MemoryAllocationLib|MdePkg/Library/UefiMemoryAllocationLib/UefiMemoryAllocationLib.inf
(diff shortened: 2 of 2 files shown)


ArmPlatformPkg/PeilessSec/PeilessSec.inf
```diff
@@ -43,6 +43,7 @@
   DebugAgentLib
   DebugLib
   HobLib
+  LzmaDecompressLib
```
ardbiesheuvel, line 46: This change was deliberate. LzmaDecompressLib is not a library class, and the module does not strictly depend on it.

In the integration, it may be the case that LzmaDecompressLib is required, when the payload is compressed using LZMA. So LzmaDecompressLib may be incorporated into the module using NULL library class resolution.

E.g.,
```
Platform/ARM/JunoPkg/ArmJuno.dsc-  ArmPlatformPkg/PeilessSec/PeilessSec.inf {
Platform/ARM/JunoPkg/ArmJuno.dsc-    <LibraryClasses>
Platform/ARM/JunoPkg/ArmJuno.dsc:      NULL|MdeModulePkg/Library/LzmaCustomDecompressLib/LzmaCustomDecompressLib.inf
Platform/ARM/JunoPkg/ArmJuno.dsc-  }
```

os-d: Sure, I was going to go that approach, but I didn't see mention of that in the PR (I may well have just missed it), so I put this up to restore it.

I agree however, and I was surprised that it relied directly on LZMA. I'll drop this PR and update the platform, thanks.

os-d:
Closing and will update platforms.

PR #11000, opened 2025-04-24
ArmPkg/ArmScmiDxe: Avoid 64-bit access from compiler optimization

JohnnyFan-CIX:
In ScmiClockProtocol.c, the ClockRateSet function contains the following assignments:
    ClockRateSetAttributes->Rate.Low  = (UINT32)Rate;
    ClockRateSetAttributes->Rate.High = (UINT32)(Rate >> 32);

In some cases, the compiler may optimize these two 32-bit accesses into a single 64-bit access. This behavior is problematic when the SCMI message payload only supports 32-bit accesses, leading to hardware faults.

This patch ensures the compiler performs two distinct 32-bit writes, preventing unsafe 64-bit memory access.

### DIFF
--- a/ArmPkg/Drivers/ArmScmiDxe/ArmScmiClockProtocolPrivate.h
+++ b/ArmPkg/Drivers/ArmScmiDxe/ArmScmiClockProtocolPrivate.h
@@ -16,8 +16,8 @@
 
 // Clock rate in two 32bit words.
 typedef struct {
-  UINT32    Low;
-  UINT32    High;
+  volatile UINT32    Low;
+  volatile UINT32    High;
 } CLOCK_RATE_DWORD;
 
 // Format of the returned rate array. Linear or Non-linear,.RatesFlag Bit[12]


ArmPkg/Drivers/ArmScmiDxe/ArmScmiClockProtocolPrivate.h
```diff
@@ -16,8 +16,8 @@
 
 // Clock rate in two 32bit words.
 typedef struct {
-  UINT32    Low;
-  UINT32    High;
+  volatile UINT32    Low;
+  volatile UINT32    High;
```
ardbiesheuvel, line 20: Nack.

If this struct is backed by anything other than ordinary memory, the code should be using MMIO accessors not ordinary loads and stores.

JohnnyFan-CIX: Thanks for the clarification.

You're right — if the backing memory is MMIO, then using ordinary C loads/stores can indeed lead to issues, and proper MMIO accessors should be used instead.

That said, from what I can tell, the current SCMI transport layer (via `ScmiCommandGetPayload`) does not clearly define whether the payload buffer is always regular memory or potentially MMIO. In our use case, it's backed by MMIO, but I understand other platforms might implement it differently.

Would it make sense to update the SCMI transport interface to explicitly document or enforce how payload access should be handled? Or alternatively, should we defensively treat all SCMI payload accesses as MMIO by default?

I'm happy to revise the patch either way — just want to make sure it's aligned with the intended design.

ardbiesheuvel: The code is poorly designed, and as you said, the documentation does not give any guidance for the clock protocols regarding to memory attributes.

However, the ARM MtlLib implementation for Juno (in edk2-platforms) clearly distinguishes between mailbox accesses (using ordinary loads and stores) and doorbell accesses (using MMIO). There are also some places in the SCMI spec where the use of normal non-cached attributes is required for shared memory buffers.

So my take on this is that the mailbox should behave like normal memory, and shouldn't require special handling such as volatile or MMIO accessors. If your implementation does not tolerate that, you should fix it in the MtlLib layer.

JohnnyFan-CIX: Thanks for the feedback.

Unless the message payload is changed from being register-backed to memory-backed, I'm currently unsure how to fix this issue within the MtlLib layer. I'd prefer to avoid changing the implementation of the message payload if possible.

After reviewing the SCMI spec again, I found that section 4.1.2 (Message format) clearly defines all message parameters as 32-bit values. Based on this, I believe these fields should be written using separate 32-bit accesses.

The original code:
  ClockRateSetAttributes->Rate.Low  = (UINT32)Rate;
  ClockRateSetAttributes->Rate.High = (UINT32)(Rate >> 32);
may be optimized by the compiler into a single 64-bit write, which could violate the spec's 32-bit granularity requirement and cause issues on certain platforms.

To ensure better compliance with the spec and minimize the risk of platform-specific issues, it might be worth considering a fix here.

ardbiesheuvel: > Thanks for the feedback.
> 
> Unless the SCMI mailbox message payload is changed to normal memory, I'm currently unsure how to fix this issue within the MtlLib layer. I'd prefer to avoid changing the implementation of the message payload if possible.
> 

Couldn't you use a bounce buffer?

> After reviewing the SCMI spec again, I found that section 4.1.2 (Message format) clearly defines all message parameters as 32-bit values. Based on this, I believe these fields should be written using separate 32-bit accesses.
> 

I disagree - even if the types of the parameters are 32-bit, it certainly does not imply that each parameter should be stored into memory individually.

> The original code: ClockRateSetAttributes->Rate.Low = (UINT32)Rate; ClockRateSetAttributes->Rate.High = (UINT32)(Rate >> 32); may be optimized by the compiler into a single 64-bit write, which could violate the spec's 32-bit granularity requirement and cause issues on certain platforms.
> 

There is no 32-bit granularity requirement in the spec.

JohnnyFan-CIX: Thanks! I'll move the fix to MtlLib using a bounce buffer and close this PR.

PR #11001, opened 2025-04-24
ArmPkg/ArmScmiDxe: Use array indexing to access adjacent buffers

ardbiesheuvel:
MessageParams1 and MessageParams2 always point to adjacent slots in the payload buffer, so no need to keep two separate pointers.


ardbiesheuvel:
cc @JohnnyFan-CIX

samimujawar (approved):

PR #11004, opened 2025-04-25
Add UUID-GUID conversion interfaces in ArmFfaLib

kuqin12:
# Description

UUID to EFI GUID conversion is a common operation that is needed for processing FF-A parameters.

This change adds 2 new interfaces to support such conversion and removes an implementation from Standalone MM entrypoint to leverage the new interfaces.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This change was tested on QEMU SBSA platform and booted to UEFI shell.

## Integration Instructions

N/A

### DIFF
--- a/MdePkg/Include/Library/BaseLib.h
+++ b/MdePkg/Include/Library/BaseLib.h
@@ -4715,6 +4715,101 @@ BitFieldCountOnes64 (
   IN       UINTN   EndBit
   );
 
+/*******************************************************************************
+
+  UUID (Universally Unique IDentifier), as defined in RFC4122
+  (https://datatracker.ietf.org/doc/html/rfc4122#section-4.1), is a 128-bit number
+  used to uniquely identify information in computer systems.
+
+  UUIDs contains 5 fields:
+  - time_low: 32 bits
+  - time_mid: 16 bits
+  - time_hi_and_version: 16 bits
+  - clock_seq_hi_and_reserved: 8 bits
+  - clock_seq_low: 8 bits
+  - node: 8 bits * 6
+
+  Each field encoded with the Most Significant Byte first (known as network byte
+  order, or big-endian).
+
+  GUID (Globally Unique Identifier), on the other hand, is a 128-bit number used
+  in UEFI environments, which is similar to UUID but has a different byte order
+  in memory. See https://uefi.org/specs/UEFI/2.11/Apx_A_GUID_and_Time_Formats.html
+
+  GUID also contains 5 fields:
+  - TimeLow: 32 bits
+  - TimeMid: 16 bits
+  - TimeHiAndVersion: 16 bits
+  - ClockSeqHighAndReserved: 16 bits
+  - ClockSeqLow: 8 bits
+  - Node: 8 bits * 6
+
+  TimeLow, TimeMid, TimeHighAndVersion fields in the EFI are encoded with the Least
+  Significant Byte first (also known as little-endian).
+
+  Example:
+  Consider the same string representation/registry format for MM communication v2:
+  "378daedc-f06b-4446-8314-40ab933c87a3"
+
+  In UUID format, it is represented as:
+  - Data fields:
+    - time_low: 0x37 0x8d 0xae 0xdc (0x378daedc in big-endian)
+    - time_mid: 0xf0 0x6b (0xf06b in big-endian)
+    - time_hi_and_version: 0x44 0x46 (0x4446 in big-endian)
+    - clock_seq_hi_and_reserved: 0x83
+    - clock_seq_low: 0x14
+    - node: 0x00, 0xab, 0x93, 0x3c, 0x87, 0xa3
+  - Byte representation in memory:
+    - 37 8d ae dc f0 6b 44 46 83 14 40 ab 93 3c 87 a3
+
+  However, in GUID format, it is represented as:
+  - Data fields:
+    - TimeLow: 0xdc 0xae 0x8d 0x37 (0x378daedc in little-endian)
+    - TimeMid: 0x6b 0xf0 (0xf06b in little-endian)
+    - TimeHiAndVersion: 0x46 0x44 (0x4446 in little-endian)
+    - ClockSeqHighAndReserved: 0x83
+    - ClockSeqLow: 0x14
+    - Node: 0x00, 0xab, 0x93, 0x3c, 0x87, 0xa3
+  - Byte representation in memory:
+    - dc ae 8d 37 6b f0 46 44 83 14 40 ab 93 3c 87 a3
+
+*******************************************************************************/
+
+/**
+  This function converts a GUID in UEFI format to a UUID in RFC4122 format.
+
+  The conversion is done by swapping the byte order of the TimeLow, TimeMid, and
+  TimeHiAndVersion fields, while keeping the ClockSeq and Node fields unchanged.
+
+  @param [in] FromGuid  GUID in format to be converted to UUID RFC4122 format.
+  @param [out] ToUuid   Pointer to a GUID structure that will hold the converted
+                        UUID in RFC4122 format.
+**/
+VOID
+EFIAPI
+ConvertGuidToUuid (
+  IN   GUID  *FromGuid,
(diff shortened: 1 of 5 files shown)


kuqin12:
@samimujawar @leiflindholm @ardbiesheuvel Do you have any suggestions on this change?

LeviYeoReum (approved):

leiflindholm (changes requested):
Well, there is nothing Arm architecture specific about MS-GUID to UUID conversion, so if we're breaking it out it should move to MdePkg or MdeModulePkg.

samimujawar (approved):

kuqin12:
@leiflindholm @samimujawar could you please let me know if there are any further comments on this PR? If not, would you mind merging the change? Thanks in advance!

leiflindholm (approved):
I'm OK with this, but ideally we'd want to hear from MdePkg maintainers.

lgao4 (approved):

lgao4:
> @lgao4 @mdkinney could you please review and provide some feedback? Thanks in advance.

I agree to add this API in BaseLib.

kuqin12:
@lgao4 @leiflindholm I know we are in soft freeze, but any opinions on merging this change as part of the 2505 tag? Thanks!

lgao4:
> @lgao4 @mdkinney could you please review and provide some feedback? Thanks in advance.

lgao4:
> @lgao4 @leiflindholm I know we are in soft freeze, but any opinions on merging this change as part of the 2505 tag? Thanks!

Seemly, these changes are not the bug fix. They refine the current implementation. So, I suggest to merge them after the stable tag 202505.

lgao4:
If no other comment, I will merge it this week.

kuqin12:
> I had always though incorrectly that UUID and GUID were the same. From RFC4122, UUID stores values in network byte order (big endian) and the UEFI Specification requires little endian, so EFI_GUID is in little endian byte order. GUID and EFI_GUID are the same type in edk2. So GUID and EFI_GUID are both little endian.
> 
> This distinction is not explicitly called out in the comments for the BaseLib functions being added. This is very similar to the SwapBytesxx() APIs that are used for example in the network stack to convert values larger than 8-bits between network byte order and little endian byte order.
> 
> Do we really need 2 APIs? Could we define a single SwapBytesGuid() or SwapBytesUuid() function that switches byte order of all fields in a GUID/UUID that are larger than 8-bits?

Indeed, as you mentioned, the key difference between these two types is endianness. However, explicitly distinguishing between UUID and GUID in the function name enhances clarity more effectively than a swap function. By specifying the types of both input and output data, it improves code readability and makes the flow more intuitive.

kuqin12:
> > > I had always though incorrectly that UUID and GUID were the same. From RFC4122, UUID stores values in network byte order (big endian) and the UEFI Specification requires little endian, so EFI_GUID is in little endian byte order. GUID and EFI_GUID are the same type in edk2. So GUID and EFI_GUID are both little endian.
> > > This distinction is not explicitly called out in the comments for the BaseLib functions being added. This is very similar to the SwapBytesxx() APIs that are used for example in the network stack to convert values larger than 8-bits between network byte order and little endian byte order.
> > > Do we really need 2 APIs? Could we define a single SwapBytesGuid() or SwapBytesUuid() function that switches byte order of all fields in a GUID/UUID that are larger than 8-bits?
> > 
> > 
> > Indeed, as you mentioned, the key difference between these two types is endianness. However, explicitly distinguishing between UUID and GUID in the function name enhances clarity more effectively than a swap function. By specifying the types of both input and output data, it improves code readability and makes the flow more intuitive.
> 
> I think that depends on how often it gets used and if the code that calls the function clearly identifies the need to switch from network byte order to little endian for from little endian to network byte order. We do not have 2 APIs to convert 64-bit values.
> 
> It is also confusing because both APIs have an identical implementation.

Although the conversion is symmetric, I would prefer that the input and output parameters explicitly include descriptions and types that indicate the endianness/format being passed. This would eliminate the need to manually trace through the entire software stack or keep track of details on paper. Would it be better if the functions are renamed to something like `ConvertLittleEndianGuidToBigEndian`?

leiflindholm:
> I had always though incorrectly that UUID and GUID were the same. From RFC4122, UUID stores values in network byte order (big endian) and the UEFI Specification requires little endian, so EFI_GUID is in little endian byte order. GUID and EFI_GUID are the same type in edk2. So GUID and EFI_GUID are both little endian.

No, it's actually worse than that. From https://uefi.org/specs/UEFI/2.11/Apx_A_GUID_and_Time_Formats.html
`It should also be noted that TimeLow, TimeMid, TimeHighAndVersion fields in the EFI are encoded as little endian.`

MdePkg/Library/BaseLib/ConvertGuidUuid.c
```diff
@@ -0,0 +1,131 @@
...
+  IN   GUID    *FromGuid,
+  OUT  UINT64  *ToUuid
+  )
+{
+  UINT32  *Data32;
+  UINT16  *Data16;
+
+  CopyGuid ((GUID *)ToUuid, FromGuid);
+  Data32    = (UINT32 *)ToUuid;
+  Data32[0] = SwapBytes32 (Data32[0]);
```
mdkinney, line 96: Why not call SwapBytesxx() on the fields of GUID without using additional local variables?

Also, why is UUID a UINT64*?  Can't is also be type GUID?  The field names are the same for UUID and GUID.  Only the endianness is different.

mdkinney: I do not agree.  UINT64* is not a UUID by type or size.  Look at the python uuid module as an example.  It uses the same object class for both forms and has a single method to return the little endian version of an uuid object.

MdePkg/Library/BaseLib/ConvertGuidUuid.c
```diff
@@ -0,0 +1,131 @@
...
+  UUIDs contains 5 fields:
+  - time_low: 32 bits
+  - time_mid: 16 bits
+  - time_hi_and_version: 16 bits
+  - clock_seq_hi_and_reserved: 8 bits
+  - clock_seq_low: 8 bits
+  - node: 48 bits
+
+  Each field encoded with the Most Significant Byte first (known as network byte
+  order).
```
leiflindholm, line 26: If we use little-endian bellw, we should also mention big-endian here for symmetry,

MdePkg/Library/BaseLib/ConvertGuidUuid.c
mdkinney: LGTM.  Thanks for the quick updates.

mdkinney (approved):

MdePkg/Include/Library/BaseLib.h
```diff
@@ -4715,6 +4715,50 @@ BitFieldCountOnes64 (
   IN       UINTN   EndBit
   );
 
+/**
+  This function converts a GUID to a UUID in RFC4122 format.
+
+  The conversion is done by swapping the byte order of the TimeLow, TimeMid, and
```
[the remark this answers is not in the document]
leiflindholm: Sorry, what I meant was: you have added the very detailed documentation of the difference between UUID and GUID. There is no need to rehash this in every function comment header.

```
@param [in] FromGuid  GUID to be converted to UUID format.
@param [out] ToUuid   Pointer to a UINT64 array that will hold the converted UUID.
```

would be sufficient.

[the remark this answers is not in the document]
leiflindholm: The situation is confusing enough that I think it's worth having it in both.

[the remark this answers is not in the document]
kuqin12: Thanks for clarifying. Will update shortly.

leiflindholm (approved):

kuqin12:
@lgao4 We have been through another round of updates. Would you mind helping reviewing and/or merging the PR? Thanks in advance!

lgao4 (approved):

Issue #9401, opened 2023-07-20
NetworkPkg: HTTP protocol throughput too small (Bugzilla Bug 4505)


tianocore-issues:
This issue was created automatically with [bugzilla2github](https://github.com/makubacki/bugzilla2github/tree/bz2gh_tianocore)
# Bugzilla Bug 4505
Date: 2023-07-20T22:50:38+00:00
From: @liyi77
To:   @liyi77
CC:   @lgao4, @lersek

Last updated: 2023-10-26T16:37:23+00:00

tianocore-issues:
## Comment 21756
Date: 2023-07-20 22:50:38 +0000
From: @liyi77

- Industry Specification: ---
- Release Observed: EDK II Master
- Releases to Fix: EDK II Master
- Target OS: ---
- Bugzilla Assignee(s): @liyi77

In edk2’s HTTP protocol implementation that uses too small tcp windows, we need to increase HTTP beffer size and enable TCP window scaling to improve performance. 
I tested it locally in virtual environment and it improves the throughput almost by 300%:

-#define HTTP_BOOT_BLOCK_SIZE                   1500
+#define HTTP_BOOT_BLOCK_SIZE                   32000

+  Tcp4Option-&gt;EnableWindowScaling = TRUE;
+  Tcp6Option-&gt;EnableWindowScaling = TRUE;

-#define HTTP_BUFFER_SIZE_DEAULT   65535
+#define HTTP_BUFFER_SIZE_DEAULT   0x200000

tianocore-issues:
## Comment 21771
Date: 2023-07-25 05:38:27 +0000
From: @lgao4

yi1.li@intel.com is working on it.

tianocore-issues:
## Comment 22071
Date: 2023-10-26 16:37:23 +0000
From: @lersek

Fixed in commit b240eab03530f063ef5438497d70a731b19a201e.

PR #11007, opened 2025-04-25
NetworkPkg:  Add PCD for HTTP transfer buffer size

baranee:
# Description

Current HTTP transfer size have a default value of 2M. 
This PCD change allows the platform code to change the transfer buffer size.
Based on the network bandwidth, the buffer size can be adjusted to improve performance.
Large buffer size provides better performance in high-bandwidth connections.
However, small buffer size helps to reduce packet loss in low-bandwidth scenarios.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verified basic HTTP file download still works.

## Integration Instructions

N/A

### DIFF
--- a/NetworkPkg/HttpDxe/HttpDriver.h
+++ b/NetworkPkg/HttpDxe/HttpDriver.h
@@ -3,6 +3,7 @@
 
   Copyright (c) 2015 - 2021, Intel Corporation. All rights reserved.<BR>
   (C) Copyright 2016 Hewlett Packard Enterprise Development LP<BR>
+  (c) Copyright 2025 HP Development Company, L.P.
 
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
@@ -27,6 +28,7 @@
 #include <Library/HttpLib.h>
 #include <Library/DpcLib.h>
 #include <Library/PrintLib.h>
+#include <Library/PcdLib.h>
 
 //
 // UEFI Driver Model Protocols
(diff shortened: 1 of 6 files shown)


SaloniKasbekar (approved):

mdkinney:
Please add to the PR description and Commit Message and to the PCD description how a platform integrator would use this new PCD setting.  Why is 2MB a good default?  What would be the impact of making it larger or making it smaller?  Also, are there constraints on the values that can be set?  Can be be any byte value or are there only specific aligned values that must be used?

baranee:
Initially it was 64K and changed to 2M as part of this commit - https://github.com/tianocore/edk2/commit/b240eab03530f063ef5438497d70a731b19a201e which is tested on virtual environment based on the comments in https://github.com/tianocore/edk2/issues/9401

However, testing has shown that 32K provides better download times in certain environments. Intent is to let platform override this default value using PCD based on discussion in https://edk2.groups.io/g/devel/topic/patch_networkpkg_increase/100271527 and retain the existing EDK2 behavior.

I couldn't find any documentation on acceptable values.

baranee:
Will update the descriptions.

mdkinney:
Updated look good.  Please combine to 1 commit.

mdkinney:
@baranee You can do update in this PR.  Just do forced push to this PR with the updated commit.

baranee:
> @baranee You can do update in this PR. Just do forced push to this PR with the updated commit.

Attempted `git push --force` to combine the commits.

mdkinney (approved):

PR #11008, opened 2025-04-25
Add comment for function in QemuVideoDxe

Damien-Chen:
Add the comment of `outb` and `ontw` function in QemuVideoDxe driver.

## How This Was Tested

N/A

## Integration Instructions

N/A


Damien-Chen:
During the investigation of QEMU Virtio-gpu related, I found that some functions are missing comment.
I try my best to understand it and add comment for it.
Could any expert that can help me check it ?

PR #11011, opened 2025-04-25
Mde/MdeModule: Extend incompatible Pci Devcies to account for OptionRoms.

apop5:
# Description
The existing incompatible PciDeviceSupport allows for addressing BARs that are incompatible.
Extend the support to allow for reporting that a device's OptionRom is incompatible as well.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Ran on a platform using a [gEfiIncompatiblePciDeviceSupportProtocolGuid](https://github.com/microsoft/mu_plus/blob/1d5fda37f1b83265a3aebb1e629127301b420417/MsCorePkg/IncompatiblePciDevices/NoOptionRomsAllowed/NoOptionRomsAllowed.inf) which prevented any option roms from executing. 

## Integration Instructions
Platforms will have the capability of installing their own instance of gEfiIncompatiblePciDeviceSupportProtocolGuid which reports that a device option rom is not able to be supported.

### DIFF
--- a/MdePkg/Include/Protocol/IncompatiblePciDeviceSupport.h
+++ b/MdePkg/Include/Protocol/IncompatiblePciDeviceSupport.h
@@ -29,7 +29,7 @@
   devices still depends on the implementation of the PCI bus driver. The PCI bus
   driver may fully, partially, or not even support these incompatible devices.
 
-  During PCI bus enumeration, the PCI bus driver will probe the PCI Base Address
+  During PCI bus enumeration, the PCI bus driver will probe the PCI Base Addressa
   Registers (BARs) for each PCI device regardless of whether the PCI device is
   incompatible or not to determine the resource requirements so that the PCI bus
   driver can invoke the proper PCI resources for them.  Generally, this resource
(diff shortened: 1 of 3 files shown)


MdePkg/Include/Protocol/IncompatiblePciDeviceSupport.h
```diff
@@ -78,6 +78,15 @@
...
   }
 
+///
+/// Extension to Acpi10.h for describing the ROM BAR as the incompatible BAR.
+///
+/// Specifying the ROM bar as incompatible indicates to the PciBusDxe driver to
+/// ignore the ROM BAR altogether.  Currently, ACPI only defines 0x01-0x03 for
+/// MEM, IO, BUS.
+///
+#define INCOMPATIBLE_ACPI_ADDRESS_SPACE_TYPE_ROM  0x0F
```
mdkinney, line 88: This has not been added to the UEFI PI Specifications for the Incompatible PCI Device Support Protocol.  Also, the latest ACPI Specification says that values 3..191 are reserved, which means this value of 0x0F could conflict with a future version of the ACPI Specification.  

This would require changes to both the ACPI Spec to allocate a value for this use case and an update to the UEFI PI Specification.  This PR should be dropped and changed to a code first spec/code change approach.  Unless an alternate method of indicating a PCI Option ROM BAR override is used.

PR #11013, opened 2025-04-25
MdeModulePkg: DebugImageInfoTable: Fix Array Maintenance

os-d:
# Description

The DebugImageInfoTable contains an array of image info structures. The current implementation removes an entry by freeing the info structure and putting NULL in that entry of the array. It then decrements the table size tracked in the table. However, the array is invalid at this point, it contains a NULL entry, which the UEFI spec does not envision and it contains a valid entry past the end of the array as tracked in the spec defined config table. If the table is consumed at this point it can lead to an invalid assessment of the image state, which defeats the purpose of the table.

When a new info structure is added, it then scans for the first NULL entry adds a pointer to the new info structure there and increments the table size to cover the entrythat was formerly past the end of the array.

The current implementation requires that once an unload happens, more loads happen than unloads and that the last operation is not an unload (which won't be true in the shell, e.g.). This is needlessly complex, as the order of the table doesn't matter (and in fact this implementation doesn't preserve image loading order either).

This patch updates the removal function to free the desired info structure, move the last entry of the array to this freed spot, mark the last entry as NULL, and decrement the table count. The entry addition function then just always puts a new entry at the end of the array, expanding it as necessary. This simplifies the logic and covers the gaps that were present.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested by running a shell app that uses the DebugImageInfoTable and confirmed it could access all expected entries.

## Integration Instructions

N/A.

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/DebugImageInfo.c
+++ b/MdeModulePkg/Core/Dxe/Misc/DebugImageInfo.c
@@ -177,20 +177,7 @@ CoreNewDebugImageInfoEntry (
 
   Table = mDebugInfoTableHeader.EfiDebugImageInfoTable;
 
-  if (mDebugInfoTableHeader.TableSize < mMaxTableEntries) {
-    //
-    // We still have empty entires in the Table, find the first empty entry.
-    //
-    Index = 0;
-    while (Table[Index].NormalImage != NULL) {
-      Index++;
-    }
-
-    //
-    // There must be an empty entry in the in the table.
-    //
-    ASSERT (Index < mMaxTableEntries);
-  } else {
+  if (mDebugInfoTableHeader.TableSize >= mMaxTableEntries) {
     //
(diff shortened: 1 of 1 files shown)


cfernald (approved):

os-d:
> Having the entries in load order would be convenient for debugging, but this feature was neither provided by the previous implementation nor is it required by the UEFI specification, cf #11019.

Yeah, my reasoning was that image load order is fairly easy to get from the log and as you said, neither the spec nor the existing implementation provided for it, so I thought speed was more important here (especially as on most boots this table will not be used, but still produced).

os-d:
@lgao4 can you please review this PR as it was put up before the soft freeze?

ardbiesheuvel (approved):

lgao4 (approved):

PR #11017, opened 2025-04-26
UniversalPayload: Add AARCH64 support on BlSupporDxe

AjanZhong:
# Description

UniversalPayload: Add AARCH64 support on BlSupporDxe.

1. Refactor BlSupportDxe to support multiple architectures.
2. Add Add AARCH64 support on BlSupporDxe.
3. Move ArmMmuLib from ArmPkg to UefiCpuPkg.

- [ ] Breaking change?
  - N/A
- [ ] Impacts security?
  - N/A
- [ ] Includes tests?
  - N/A

## How This Was Tested

Both X64 and Arm64 platform.

## Integration Instructions
N/A

Corresponding EDK2-platform PR to adapt change "3. Move ArmMmuLib from ArmPkg to UefiCpuPkg.":
https://github.com/tianocore/edk2-platforms/pull/769


ArmPkg/ArmPkg.dec
```diff
@@ -115,12 +115,6 @@
   # attributes initially, and can be disabled otherwise.
   gArmTokenSpaceGuid.PcdRemapUnusedMemoryNx|TRUE|BOOLEAN|0x00000048
 
-[PcdsFeatureFlag.ARM]
```
leiflindholm, line 124: @ardbiesheuvel @samimujawar Is this even a feature we want to support anymore? Or could we get rid of the Pcd, and the PreferNonshareableMemory() function?

There are no upstream platforms setting the Pcd (last upstream user was ArmVExpress-CTA15-A7) and the only downstream platform I can find in github is in the EFIDroid repo where it was added for MSM8610, which is also a Cortex-A7, and then seems to have been cloned onto modern platforms (where it gets ignored, because it's ARM-only).

ardbiesheuvel: Yeah let's get rid of it.

ArmPkg/ArmPkg.dec
```diff
@@ -41,10 +41,6 @@
   #
   ArmHvcLib|Include/Library/ArmHvcLib.h
 
-  ##  @libraryclass  Provides a Mmu interface.
-  #
-  ArmMmuLib|Include/Library/ArmMmuLib.h
-
```
ardbiesheuvel, line 47: This library class could live in UefiCpuPkg too, no?

ArmPkg/ArmPkg.dec
```diff
@@ -91,8 +87,6 @@
   # Include/Guid/ArmMpCoreInfo.h
   gArmMpCoreInfoGuid = { 0xa4ee0728, 0xe5d7, 0x4ac5,  {0xb2, 0x1e, 0x65, 0x8e, 0xd8, 0x57, 0xe8, 0x34} }
 
-  gArmMmuReplaceLiveTranslationEntryFuncGuid = { 0xa8b50ff3, 0x08ec, 0x4dd3, {0xbf, 0x04, 0x28, 0xbf, 0x71, 0x75, 0xc7, 0x4a} }
-
```
ardbiesheuvel, line 95: This definitely belongs with the implementation, not with the API definition. So this should be moved to UefiCpuPkg.

AjanZhong: Done.

ArmPkg/ArmPkg.dec
AjanZhong: yes, it has been moved to UefiCpuPkg in this PR.

ArmPkg/Library/ArmLib/ArmBaseLib.inf
```diff
@@ -47,4 +47,4 @@
   MdePkg/MdePkg.dec
 
 [FeaturePcd.ARM]
-  gArmTokenSpaceGuid.PcdNormalMemoryNonshareableOverride
+  gEfiMdePkgTokenSpaceGuid.PcdNormalMemoryNonshareableOverride
```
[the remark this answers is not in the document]
AjanZhong: Done.

ArmPkg/ArmPkg.dec
AjanZhong: Dropped support to Pcd: PcdNormalMemoryNonshareableOverride. However, file level static function PreferNonshareableMemory is stilled invoked by other functions. So I kept function PreferNonshareableMemory.

gguo11837463 (approved):

AjanZhong:
Separated commit "Move ArmMmuLib to UefiCpuPkg", since these commit involves six different packages. To meet PatchCheck policy, use two commits to implement it.

ArmPkg/ArmPkg.dec
AjanZhong: @leiflindholm @ardbiesheuvel  Hi, Leif and Ard, I have addressed most comments, would you mind help check whether this solution fulfill comments in "PreferNonshareableMemory()" fuction"
Thanks.

ArmPlatformPkg/ArmPlatformPkg.dsc
```diff
@@ -38,7 +38,7 @@
 [LibraryClasses.common]
   ArmLib|ArmPkg/Library/ArmLib/ArmBaseLib.inf
   ArmPlatformLib|ArmPlatformPkg/Library/ArmPlatformLibNull/ArmPlatformLibNull.inf
-  ArmMmuLib|ArmPkg/Library/ArmMmuLib/ArmMmuBaseLib.inf
+  ArmMmuLib|UefiCpuPkg/Library/ArmMmuLib/ArmMmuBaseLib.inf
```
leiflindholm, line 41: This commit needs to be squashed into the previous one to prevent breaking bisect.
Also, there are 17 .dsc* files in edk2-platforms that will need updating when this PR goes in.

ArmPkg/ArmPkg.dec
leiflindholm: We can nuke the function completely.

The first two invocations are trivially deletable.
The third one ... well, if that was a problem we already broke at least as far back as 1f3b1eb308220, 5 years ago, if not earlier.

So you can delete both the if statements from
https://github.com/tianocore/edk2/blob/master/ArmPkg/Library/ArmMmuLib/Arm/ArmMmuLibCore.c#L373 (TTBR_SHAREABLE is unconditionally set by the immediately preceding line) keeping only the contents of the else clause.

ArmPlatformPkg/ArmPlatformPkg.dsc
AjanZhong: Thanks for your quick response. :)
Previously this commit is squashed with previous one. 
Due to this commit: https://github.com/tianocore/edk2/commit/c44efa0fb , we need to list all modified packages in Commit Subject Line, however, we have totally 6 modules need to be listed in Commit Subject Line, it makes Subject Line too long.  So I suggest to use 2 commits. What's your opinion?

To *.dsc* files in edk2-platform, I will create a PR to match changes in this PR.

ArmPkg/ArmPkg.dec
AjanZhong: Thanks, let me do it.

ArmPlatformPkg/ArmPlatformPkg.dsc
leiflindholm: Ah, Arm architecture being spread over three packages strikes again.
Hmm.

No, splitting the patch is not the answer. The subject line limit is meant to force people to keep the short-log ... short. But inserting a bisect breakage is not an appropriate workaround.

I've raised #11241 to address this situation.

AjanZhong: PR created in edk2-platforms:
https://github.com/tianocore/edk2-platforms/pull/809

AjanZhong: Two commits which are used to "update ArmMmuLib" has been squashed into one commit, it will lead to patch check failure, I will rebase this PR again after https://github.com/tianocore/edk2/pull/11241 merged.

ArmPkg/Library/ArmMmuLib/Arm/ArmMmuLibCore.c
```diff
@@ -371,19 +326,15 @@ ArmConfigureMmu (
   TTBRAttributes = ArmHasMpExtensions () ? TTBR_MP_WRITE_BACK_ALLOC
                                          : TTBR_WRITE_BACK_ALLOC;
   if (TTBRAttributes & TTBR_SHAREABLE) {
```
leiflindholm: Drop this if-statement too: TTBR_SHAREABLE is always set.

ArmPlatformPkg/ArmPlatformPkg.dsc
leiflindholm: @AjanZhong #11241 now merged!

leiflindholm:
OK, it's very late for this comment, and I apologise, but it just occurred to me to look where other MmuLibs might reside ... and it turns out RiscV and LoongArch both keep theirs in UefiCpuPkg (as RiscVMmuLib and CpuMmuLib respectively).

Since I'd like to really only do a move this invasive once ... could you possibly rework to do the same?

AjanZhong:
> OK, it's very late for this comment, and I apologise, but it just occurred to me to look where other MmuLibs might reside ... and it turns out RiscV and LoongArch both keep theirs in UefiCpuPkg (as RiscVMmuLib and CpuMmuLib respectively).
> 
> Since I'd like to really only do a move this invasive once ... could you possibly rework to do the same?

No problem Lief, you have helped a lot for patches reviewing. :)

So we should move ArmMmuLib.h from MdePkg/Include/Library/ to UefiCpuPkg/Include/Library/, and keep ArmMmuLib implementation in UefiCpuPkg/Library/ArmMmuLib, right?

leiflindholm:
> So we should move ArmMmuLib.h from MdePkg/Include/Library/ to UefiCpuPkg/Include/Library/, and keep ArmMmuLib implementation in UefiCpuPkg/Library/ArmMmuLib, right?

Yes please. Many thanks.

AjanZhong:
> > So we should move ArmMmuLib.h from MdePkg/Include/Library/ to UefiCpuPkg/Include/Library/, and keep ArmMmuLib implementation in UefiCpuPkg/Library/ArmMmuLib, right?
> 
> Yes please. Many thanks.

Sure, let me do it. Since it might trigger Github CI failures, I will convert this PR to draft first, and convert back when it's ready. :)

ardbiesheuvel (approved):

leiflindholm:
@mergify requeue

leiflindholm (approved):

PR #11019, opened 2025-04-28
MdeModulePkg: Correct CoreRemoveDebugImageInfoEntry()

xypron:
# Description

The UEFI specification does not allow NULL pointers in the EFI Debug Info Table. The TableSize field in the EFI Debug Info Table Header must always match the array size of the EFI Debug Info Table.

Instead of creating a NULL entry when removing an entry, move the tail of the EFI Debug Info Table one slot to the front.

# Breaking change?
no

# Impacts security?
no

# Includes tests?
no

## How This Was Tested
https://lore.kernel.org/u-boot/20250428080112.108658-1-heinrich.schuchardt@canonical.com/T/#u
has a patch for U-Boot to create a test binary.

## Integration Instructions
N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/DebugImageInfo.c
+++ b/MdeModulePkg/Core/Dxe/Misc/DebugImageInfo.c
@@ -7,6 +7,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
 
+#include <Library/BaseMemoryLib.h>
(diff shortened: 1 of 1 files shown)


os-d:
Ironically, I put up a PR to fix this same issue last week: https://github.com/tianocore/edk2/pull/11013. Can you test with that and confirm it fixes the issue you are seeing? I also updated the Add function to clean it up

PR #11025, opened 2025-04-30
IntelFsp2Pkg/CacheLib: Make integer width consistent in loop condition

makubacki:
# Description

Explicitly casts `mFixedMtrrTable[MsrNum].Length` to resolve a `comparison-with-wider-type` CodeQL error.

https://codeql.github.com/codeql-query-help/cpp/cpp-comparison-with-wider-type/

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- CI
- CodeQL against IntelFsp2Pkg

## Integration Instructions

N/A


IntelFsp2Pkg/Library/BaseCacheLib/CacheLib.c
```diff
@@ -328,7 +328,7 @@ ProgramFixedMtrr (
     return EFI_DEVICE_ERROR;
   }
 
-  for ( ; ((ByteShift < 8) && (*Len >= mFixedMtrrTable[MsrNum].Length)); ByteShift++) {
+  for ( ; ((ByteShift < 8) && (*Len >= (UINT64)mFixedMtrrTable[MsrNum].Length)); ByteShift++) {
```
ChaselChiu, line 331: enlarging the data type might bring some "unwanted data" into the "(UINT64) mFixedMtrrTable[MsrNum].Length"?
How about changing it like below which looks safer?
```
mFixedMtrrTable[MsrNum].Length <= (UINT32) *Len
```

makubacki: @ChaselChiu, since this is upcasting a 32-bit unsigned integer (`mFixedMtrrTable[MsrNum].Length`) to a 64-bit unsigned integer, the value should be zero-extended predictably with zeros. That prevents truncating the 64-bit `Len` value with `(UINT32)*Len`.

Sorry for the delay. Let me know if you still have any concerns.

ChaselChiu (approved):

IntelFsp2Pkg/Library/BaseCacheLib/CacheLib.c
ChaselChiu: thanks for clarification!

PR #11026, opened 2025-04-30
MdeModulePkg: Fix Image Memory Protection Applying

os-d:
# Description

Commit 5ccb5fff02a66b21898bd57f48bbd7c3cd6f4e8d updated the image memory protection code to set the protection attributes through the GCD instead of directly to the page table. However, this code had an implicit assumption that each base address passed to it was the beginning of a GCD descriptor. On the virtual platforms tested, this was the case. However, on a physical platform, a scenario was encountered where the base address was not the beginning of a GCD descriptor, thus causing memory attributes to be applied incorrectly.

This assumption does not need to be made and this patch updates the code to handle the case where the base address is not the beginning of a GCD descriptor.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on a physical Intel platform and on the original repro with OVMF VS2019 and VS2022.

## Integration Instructions

N/A.

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
@@ -196,8 +196,11 @@ SetUefiImageMemoryAttributes (
   UINT64                           FinalAttributes;
   UINT64                           CurrentAddress;
   UINT64                           CurrentLength;
+  UINT64                           ImageEnd;
+  UINT64                           DescEnd;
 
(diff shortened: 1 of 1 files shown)


os-d:
@lgao4 sorry, I was unclear on the PR that added this that I was still testing on a physical platform with the latest change. This bug was discovered during that testing and now I have completed physical platform testing.

apop5 (approved):

lgao4 (approved):

PR #11027, opened 2025-05-01
ArmPkg: StMM: Use x24 register to store stack address

mgamal:
# Description

Now that the use of 18 registers for FF-A is supported. Registers from x0-x17 can be used by FF-A to return call results. StMM also uses registers x19-x23 to store SPMC boot information.

Use the x24 register to store the stack address instead of x12 to avoid it being clobbered by FF-A calls.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested running rust-spmc as an EL-1 SPMC

## Integration Instructions

N/A


LeviYeoReum (approved):

samimujawar (approved):

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/AArch64/ModuleEntryPoint.S
```diff
@@ -119,7 +119,7 @@ ASM_FUNC(SetStackPermissionFfa)
   mov x9, xzr
   mov x10, xzr
   mov x11, xzr
-  mov x12, xzr
```
leiflindholm, line 122: Is this not zeroing out return values? If so I think it should still be x12.

leiflindholm: No, it's just zeroing scratch registers, and x24 is now one. Ignore me.

leiflindholm (approved):

PR #11030, opened 2025-05-02
MdeModulePkg: Correct Usb Mouse Z for absolute pointer.

apop5:
# Description
https://www.usb.org/sites/default/files/hid1_11.pdf Appendix B, B.2 Protocol 2 (Mouse) details the information returned by a usb mouse HID.

bytes 3..n are specific to the device, but are used by absolute pointer devices to return a z axis.
Prior to this change, the existing code was reusing the X value for the Z axis, which was incorrect.
For usb devices which do not return enough data for a z axis, this change will be a no-op.

for reference, the dxemouse can also be viewed to show that for "z axis" is it referencing the 3rd byte of returned data. 

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
N/A

## Integration Instructions
No integration necessary.

### DIFF
--- a/MdeModulePkg/Bus/Usb/UsbMouseAbsolutePointerDxe/UsbMouseAbsolutePointer.c
+++ b/MdeModulePkg/Bus/Usb/UsbMouseAbsolutePointerDxe/UsbMouseAbsolutePointer.c
@@ -847,7 +847,7 @@ OnMouseInterruptComplete (
     UsbMouseAbsolutePointerDevice->State.CurrentZ =
       MIN (
         MAX (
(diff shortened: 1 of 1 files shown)


mdkinney:
How was this tested?

mdkinney (approved):

apop5:
> How was this tested?

The details of the testing of this have been lost, but its been used since 2018 in platforms with absolute pointers (usb touch screens) since then.

mdkinney:
@lgao4 This PR was submitted before soft freeze and has passed CI.  Please review.  I approve this for edk2-stable202505.

lgao4 (approved):

PR #11039, opened 2025-05-05
WifiConnectionManager: PreScan Check for Connect

Zclarkwilliams:
# Description

When the Wifi network is enabled the connection
manager will trigger a network scan without a
profile to use. If there is a connected network
or attempting a connection, the scan will
interrupt and break the connection.

Fix - The Wifi Connection Manager will check for
SSID length longer than the minimum SSID length,
and if the state is not in connecting or
connected. Only of SSID string length is over the
defined minimum and state are not connected then
it will attempt the network scan as normal.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This change was tested on platforms with Wifi connections established and tested if this fixes the break in the current connect as well if this change breaks anything when the platform is supposed to scan correctly.

## Integration Instructions

N/A

### DIFF
--- a/NetworkPkg/WifiConnectionManagerDxe/WifiConnectionMgrImpl.c
+++ b/NetworkPkg/WifiConnectionManagerDxe/WifiConnectionMgrImpl.c
@@ -1489,7 +1489,25 @@ WifiMgrOnTimerTick (
   }
 
   Nic = (WIFI_MGR_DEVICE_DATA *)Context;
+  if (Nic->ConnectPendingNetwork == NULL) {
+    DEBUG ((DEBUG_INFO, "[WiFi Connection Manager] No profile for connection, no scan triggered!\n"));
+    gBS->CloseEvent (Event);
+    return;
+  }
+
+  if (StrLen (Nic->ConnectPendingNetwork->SSId) < SSID_MIN_LEN) {
+    DEBUG ((DEBUG_INFO, "[WiFi Connection Manager] Invalid SSID length for connection, no scan triggered!\n"));
+    gBS->CloseEvent (Event);
+    return;
+  }
+
   NET_CHECK_SIGNATURE (Nic, WIFI_MGR_DEVICE_DATA_SIGNATURE);
+
+  if ((Nic->ConnectState == WifiMgrConnectedToAp) || (Nic->ConnectState == WifiMgrConnectingToAp)) {
+    DEBUG ((DEBUG_INFO, "[WiFi Connection Manager] Already connecting to AP, no scan triggered!\n"));
+    gBS->CloseEvent (Event);
+    return;
+  }
 
   Status = WifiMgrGetLinkState (Nic, &LinkState);
   if (EFI_ERROR (Status)) {


NetworkPkg/WifiConnectionManagerDxe/WifiConnectionMgrImpl.c
```diff
@@ -1491,6 +1491,12 @@ WifiMgrOnTimerTick (
   Nic = (WIFI_MGR_DEVICE_DATA *)Context;
   NET_CHECK_SIGNATURE (Nic, WIFI_MGR_DEVICE_DATA_SIGNATURE);
 
+  if (StrLen (Nic->ConnectPendingNetwork->SSId) < SSID_MIN_LEN || Nic->ConnectState == WifiMgrConnectedToAp || Nic->ConnectState == WifiMgrConnectingToAp) {
```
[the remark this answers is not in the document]
Zclarkwilliams: added

mdkinney:
Is this considered a critical bug fix for the edk2-stable202505 release?

Zclarkwilliams:
> Is this considered a critical bug fix for the edk2-stable202505 release?

Yes, This prevents network breaking due to scanning while in connection processing state or no profile provided for connection causing noise and instability.

NetworkPkg/WifiConnectionManagerDxe/WifiConnectionMgrImpl.c
```diff
@@ -1491,6 +1491,14 @@ WifiMgrOnTimerTick (
   Nic = (WIFI_MGR_DEVICE_DATA *)Context;
   NET_CHECK_SIGNATURE (Nic, WIFI_MGR_DEVICE_DATA_SIGNATURE);
 
+  if ((Nic->ConnectPendingNetwork == NULL) || (StrLen (Nic->ConnectPendingNetwork->SSId) < SSID_MIN_LEN) || \
```
mdkinney, line 1494: Would prefer to see NULL check of ConnectPendingNetwork in its own if and only deref in side that if statement.  Also, there are 2 conditions being checked here with a debug message that says one or the other.  It would be better to check each condition independent with a clear debug message for each condition.

Zclarkwilliams: Broke this into 2 checks, valid profile and current connection status.

NetworkPkg/WifiConnectionManagerDxe/WifiConnectionMgrImpl.c
```diff
@@ -1491,6 +1491,18 @@ WifiMgrOnTimerTick (
   Nic = (WIFI_MGR_DEVICE_DATA *)Context;
   NET_CHECK_SIGNATURE (Nic, WIFI_MGR_DEVICE_DATA_SIGNATURE);
```
mdkinney, line 1492: Another bug here.  If Context is NULL, then Nic is NUL, then Signature check will dereference NULL.  Should really add a check for Context NULL first.  Then the Nic NULL check below can be skipped

NetworkPkg/WifiConnectionManagerDxe/WifiConnectionMgrImpl.c
```diff
@@ -1491,6 +1491,18 @@ WifiMgrOnTimerTick (
   Nic = (WIFI_MGR_DEVICE_DATA *)Context;
   NET_CHECK_SIGNATURE (Nic, WIFI_MGR_DEVICE_DATA_SIGNATURE);
 
+  if ((Nic->ConnectPendingNetwork == NULL) || (StrLen (Nic->ConnectPendingNetwork->SSId) < SSID_MIN_LEN)) {
+    DEBUG ((DEBUG_INFO, "[WiFi Connection Manager] No valid profile for connection, no scan triggered!\n"));
+    gBS->CloseEvent (Event);
```
mdkinney, line 1496: Why do these error return paths close the event and others in this function do not?

Zclarkwilliams: These failure conditions indicate that the scan is invalid and should not be executed, so we close the event. Others are possible valid scans where there was a hiccup in the system where the scan is still valid but didn't execute fully this round, but should still scan.

NetworkPkg/WifiConnectionManagerDxe/WifiConnectionMgrImpl.c
```diff
@@ -1491,6 +1491,18 @@ WifiMgrOnTimerTick (
   Nic = (WIFI_MGR_DEVICE_DATA *)Context;
   NET_CHECK_SIGNATURE (Nic, WIFI_MGR_DEVICE_DATA_SIGNATURE);
 
+  if ((Nic->ConnectPendingNetwork == NULL) || (StrLen (Nic->ConnectPendingNetwork->SSId) < SSID_MIN_LEN)) {
```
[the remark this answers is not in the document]
Zclarkwilliams: fixed

NetworkPkg/WifiConnectionManagerDxe/WifiConnectionMgrImpl.c
Zclarkwilliams: fixed

NetworkPkg/WifiConnectionManagerDxe/WifiConnectionMgrImpl.c
mdkinney: Perhaps some comments to explain why the event is being closed.

ardbiesheuvel: `WifiMgrOnTimerTick()` is called by a timer event, and IIUC, closing it will prevent it from being called again until the driver unbinds and binds again. Is that really the right thing to do here?

NetworkPkg/WifiConnectionManagerDxe/WifiConnectionMgrImpl.c
```diff
@@ -1489,8 +1489,26 @@ WifiMgrOnTimerTick (
...
+  }
+
   NET_CHECK_SIGNATURE (Nic, WIFI_MGR_DEVICE_DATA_SIGNATURE);
 
+  if ((Nic->ConnectState == WifiMgrConnectedToAp) || (Nic->ConnectState == WifiMgrConnectingToAp)) {
+    DEBUG ((DEBUG_INFO, "[WiFi Connection Manager] Already connecting to AP, no scan triggered!\n"));
+    gBS->CloseEvent (Event);
+    return;
+  }
+
```
ardbiesheuvel, line 1511: These early returns from the function make some of the code in the `switch` below unreachable, so it seems to me that the logic should be combined in some way.

Zclarkwilliams:
On review of the event closure causing possible future issues if a user wants to connect and scan a new profile, this might cause issues. Drafted updated changes to set timer control through timer setting control in HII. 
https://github.com/tianocore/edk2/pull/11078

PR #11042, opened 2025-05-06
NetworkPkg/IScsiDxe:Fix for out of bound memory access for bz4207 (CVE-2024-38805)

santhoshVenkat25:
# Description

A malicious iSCSI target could reply to the iSCSI initiator with a malformed packet, causing out-of-bounds memory reads and writes. This most likely leads to a denial of service, as the write primitive should not be exploitable.

To Fix this,
In IScsiBuildKeyValueList, check if we have any data left (Len > 0) before advancing the Data pointer and reducing Len.
Avoids wrapping Len. Also Used SafeUint32SubSafeUint32Sub call to reduce the Len .

### DIFF
--- a/NetworkPkg/IScsiDxe/IScsiProto.c
+++ b/NetworkPkg/IScsiDxe/IScsiProto.c
@@ -1880,6 +1880,8 @@ IScsiBuildKeyValueList (
 {
   LIST_ENTRY            *ListHead;
   ISCSI_KEY_VALUE_PAIR  *KeyValuePair;
+  EFI_STATUS            Status;
+  UINT32                Result;
 
   ListHead = AllocatePool (sizeof (LIST_ENTRY));
   if (ListHead == NULL) {
@@ -1903,9 +1905,14 @@ IScsiBuildKeyValueList (
       Data++;
     }
 
-    if (*Data == '=') {
+    // Here Len must not be zero.
+    // The value of Len is size of data buffer. Actually, Data is make up of strings.
+    // AuthMethod=None\0TargetAlias=LIO Target\0 TargetPortalGroupTag=1\0
+    // (1) Len == 0, *Data != '=' goto ON_ERROR
+    // (2) *Data == '=', Len != 0 normal case.
+    // (3) *Data == '=', Len == 0, Between Data and Len are mismatch, Len isn't all size of data, as error.
+    if ((Len > 0) && (*Data == '=')) {
       *Data = '\0';
-
       Data++;
       Len--;
     } else {
@@ -1915,10 +1922,22 @@ IScsiBuildKeyValueList (
 
     KeyValuePair->Value = Data;
 
-    InsertTailList (ListHead, &KeyValuePair->List);
+    Status = SafeUint32Add ((UINT32)AsciiStrLen (KeyValuePair->Value), 1, &Result);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "%a Memory Overflow is Detected.\n", __func__));
+      FreePool (KeyValuePair);
+      goto ON_ERROR;
+    }
 
-    Data += AsciiStrLen (KeyValuePair->Value) + 1;
-    Len  -= (UINT32)AsciiStrLen (KeyValuePair->Value) + 1;
+    Status = SafeUint32Sub (Len, Result, &Len);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "%a Out of bound memory access Detected.\n", __func__));
+      FreePool (KeyValuePair);
+      goto ON_ERROR;
+    }
+
+    InsertTailList (ListHead, &KeyValuePair->List);
+    Data += Result;
   }
 
   return ListHead;


santhoshVenkat25:
Hi, @lgao4 

can you please review this commit and add this changes in to EDK202505 Tag

Thanks,
santhosh

NetworkPkg/IScsiDxe/IScsiProto.c
```diff
@@ -1917,8 +1922,12 @@ IScsiBuildKeyValueList (
 
     InsertTailList (ListHead, &KeyValuePair->List);
 
+    if (EFI_ERROR (SafeUint32Sub (Len, (UINT32)AsciiStrLen (KeyValuePair->Value) + 1, &Len))) {
```
mdkinney, line 1925: Do call to SafeUnit32Sub() on its own line and capture return status.  Then use return status in EFI_ERROR() macro.

Also update the DEBUG() message to more clearly indicate the actual error condition.  "Failure to do subtraction" does not describe the error.

Also, the call to AsciiStrLen (KeyValuePair->Value) +1 is computed more than one time and could overflow.  Should safe int lib be used to addition operation too and done once and the result used to update Len and Data?

mdkinney (changes requested):

NetworkPkg/IScsiDxe/IScsiProto.c
santhoshVenkat25: hi @mdkinney 

https://github.com/santhoshVenkat25/edk2/commit/80dda6c0a4751fd584e012d86d03b2343a0a5a36

I have addressed your review comment,Can you please review this commit?

Thanks,
santhosh

mdkinney: I have left more comments in that commit.  

This PR is in a strange state.  After addressing the latest comments, please combine all changes to a single commit and rebase this PR with no merge commits and do a forced push.

mdkinney (changes requested):

santhoshVenkat25:
@mdkinney 

I have made changes for debug message review comments.
and force pushed the changes

NetworkPkg/IScsiDxe/IScsiProto.c
```diff
@@ -1916,9 +1923,19 @@ IScsiBuildKeyValueList (
     KeyValuePair->Value = Data;
 
     InsertTailList (ListHead, &KeyValuePair->List);
+    Status = SafeUint32Add (AsciiStrLen (KeyValuePair->Value), 1, &Result);
```
mdkinney, line 1926: The logic that manages the allocated buffers in this function is confusing.  When ON_ERROR path is taken, all KeyValuePair buffers actually added to the linked list are freed.  Line 1919 has to do an independent FreePool() call because the KeyValuePair allocated has not been added to the linked list yet.  The new boundary check are done after KeyValuePair is inserted into the linked list, so ON_ERROR path will free the KeyValuePair being added.

It would be easier to follow if either
a) KeyValuePair is always added to the linked list before any condition check are made, and ON_ERROR does all free operations.
b) KeyValuePair is not added to the linked list until all condition checks are made and any condition check failure must free KeyValuePair buffer and then goto ON_ERROR to free all other linked list items.

santhoshVenkat25: @mdkinney 

```
// Here Len must not be zero.
    // The value of Len is size of data buffer. Actually, Data is make up of strings.
    // AuthMethod=None\0TargetAlias=LIO Target\0 TargetPortalGroupTag=1\0
    // (1) Len == 0, *Data != '=' goto ON_ERROR
    // (2) *Data == '=', Len != 0 normal case.
    // (3) *Data == '=', Len == 0, Between Data and Len are mismatch, Len isn't all size of data, as error.
    if ((Len > 0) && (*Data == '=')) {
      *Data = '\0';
      Data++;
      Len--;
    } else {
      FreePool (KeyValuePair);
      goto ON_ERROR;
    }

    KeyValuePair->Value = Data;

    
    Status = SafeUint32Add (AsciiStrLen (KeyValuePair->Value), 1, &Result);
    if (EFI_ERROR (Status)) {
      DEBUG ((DEBUG_ERROR, "%a Memory Overflow is Detected.\n", __func__));
      FreePool (KeyValuePair);
      goto ON_ERROR;
    }

    Status = SafeUint32Sub (Len, Result, &Len);
    if (EFI_ERROR (Status)) {
      DEBUG ((DEBUG_ERROR, "%a Out of bound memory access Detected.\n", __func__));
      FreePool (KeyValuePair);
      goto ON_ERROR;
    }
    InsertTailList (ListHead, &KeyValuePair->List);
    Data += Result;
  }
```


Followed B) suggestion and made changes ,kindly review the code in comment section.

Thanks,
santhosh

mdkinney: Looks good to me

lgao4:
This is a security issue. It can still be merged in hard feature freeze phase.

NetworkPkg/IScsiDxe/IScsiProto.c
santhoshVenkat25: @lgao4 
Patchcheck.py checks is working locally for this commit.

 it seems i added the signoff by in this commit message.can you please let me know why it is failing?, Am I missing some thing?
 
 @mdkinney 

I updated changes.can you pass the review?

mdkinney: There are 2 commits.  One is a merge commit that is not allowed.  Always do "update with rebase" to avoid this issue.

santhoshVenkat25: @mdkinney Thank you for the help.

mdkinney (approved):

mdkinney:
@lgao4 I agree that this can be merged for edk2-stable202505

PR #11048, opened 2025-05-08
Smm cpu wait all aps fix

jiaxinwu:
# Description

Set AllApArrivedWithException to TRUE if the 1st timeout sync is skipped but all APs have arrived, then set AllApArrivedWithException to TRUE.

This PR also refines the debug log in the SmmWaitForApArrival.

## How This Was Tested

Pass normal boot.

### DIFF
--- a/UefiCpuPkg/PiSmmCpuDxeSmm/MpService.c
+++ b/UefiCpuPkg/PiSmmCpuDxeSmm/MpService.c
@@ -236,12 +236,14 @@ SmmWaitForApArrival (
   UINTN    Index;
   UINT32   DelayedCount;
   UINT32   BlockedCount;
+  UINT32   DisabledCount;
   BOOLEAN  SyncNeeded;
 
   PERF_FUNCTION_BEGIN ();
 
-  DelayedCount = 0;
-  BlockedCount = 0;
+  DelayedCount  = 0;
+  BlockedCount  = 0;
+  DisabledCount = 0;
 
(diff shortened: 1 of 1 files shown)


jiaxinwu:
hi Ray, @niruiyu, can you help review this change?

UefiCpuPkg/PiSmmCpuDxeSmm/MpService.c
```diff
@@ -327,11 +329,13 @@ SmmWaitForApArrival (
   }
 
   if (!mSmmMpSyncData->AllApArrivedWithException) {
+    DEBUG ((DEBUG_ERROR, "SmmWaitForApArrival: Failed to wait all APs enter SMI!\n"));
```
lzeng14, line 332: Is there valid case to enter this situation ?

jiaxinwu: yes, it's the another SMI source (UCNA MCA) combined with LMCE causes all APs to enter SMI.

lzeng14: And when code runs to line 332, that is a valid case or not ? If valid case, may not use DEBUG_ERROR.

jiaxinwu: i don't think it's valid, failed to wait all APs enter SMI will lead system run into unexpected behavior since the user expect all Aps arrival, so, i add the error message here.

jiaxinwu:
replaced by the https://github.com/tianocore/edk2/pull/11317

PR #11056, opened 2025-05-09
MdeModulePkg/Library: make ArmFfaPeiLib available early PEIM stage

LeviYeoReum:
commit 26fb5edff397 ("MdeModulePkg/ArmFfaLib: Add depex on gEfiPeiMemoryDiscoveredPpiGuid") restricts ArmFfaPeiLib usage only after PEI phase can be used permanent memory. However, This would be problem for PEIM which runs before gEfiPeiMemoryDiscoveredPpiGuid Ppi installed. (i.e) Tcg2Pei PEIM.

To resolve this, remove dependency on gEfiPeiMemoryDiscoveredPpiGuid and let ArmFfaPeiLib remap the Rx/Tx buffer after
gEfiPeiMemoryDiscoveredPpiGuid is installed so that ArmFfaPeiLib can be used with temporary memory.

Fixes: 26fb5edff397 ("MdeModulePkg/ArmFfaLib: Add depex on gEfiPeiMemoryDiscoveredPpiGuid")

## How This Was Tested

Run FVP-RevC with Tcg2Pei, VariableService and MmcommunicationPei.

### DIFF
--- a/MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
+++ b/MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
@@ -35,6 +35,89 @@
 #include "ArmFfaRxTxMap.h"
 
 /**
+  Update Rx/TX buffer information.
+
+  @param  BufferInfo            Rx/Tx buffer information.
+
+**/
+STATIC
+VOID
+EFIAPI
+UpdateRxTxBufferInfo (
+  OUT ARM_FFA_RX_TX_BUFFER_INFO  *BufferInfo
+  )
+{
+  BufferInfo->TxBufferAddr = (VOID *)(UINTN)PcdGet64 (PcdFfaTxBuffer);
+  BufferInfo->TxBufferSize = PcdGet64 (PcdFfaTxRxPageCount) * EFI_PAGE_SIZE;
+  BufferInfo->RxBufferAddr = (VOID *)(UINTN)PcdGet64 (PcdFfaRxBuffer);
+  BufferInfo->RxBufferSize = PcdGet64 (PcdFfaTxRxPageCount) * EFI_PAGE_SIZE;
+}
+
+/**
+  Notification service to be called when gEfiPeiMemoryDiscoveredPpiGuid is installed.
+  This function change reamp Rx/Tx buffer with permanent memory from
+  temporary Rx/Tx buffer.
+
+  Since, the Rx/Tx buffer is chanaged after gEfiPeiMemoryDiscoveredPpiGuid is installed,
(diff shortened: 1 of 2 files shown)


LeviYeoReum:
@samimujawar @kuqin12 

could you review this patch please?

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
```diff
@@ -74,6 +138,15 @@ ArmFfaPeiLibConstructor (
...
+  Status = (*PeiServices)->LocatePpi (
+                             PeiServices,
+                             &gEfiPeiMemoryDiscoveredPpiGuid,
+                             0,
+                             NULL,
+                             &Dummy
+                             );
+  TempMem = (EFI_ERROR (Status)) ? TRUE : FALSE;
+
   RxTxBufferHob = GetFirstGuidHob (&gArmFfaRxTxBufferInfoGuid);
```
kuqin12, line 150: This seems to be 85% duplicated functionality as in `PeiServicesMemoryDiscoveredNotifyCallback`? Can we please define a common routine for both places?

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
```diff
@@ -34,6 +34,84 @@
...
+
+  RxTxBufferHob = GetFirstGuidHob (&gArmFfaRxTxBufferInfoGuid);
+  ASSERT (RxTxBufferHob != NULL);
+  BufferInfo = GET_GUID_HOB_DATA (RxTxBufferHob);
+
+  /*
+   * temporary memory doesn't need to be free...
+   * otherwise PEI memory manager using permanent memory will be confused...
+   */
+  PcdSet64S (PcdFfaTxBuffer, 0x00);
```
kuqin12, line 91: I think there needs to be a callout that consumers should not cache Rx/Tx buffer pointers before permanent memory is ready.

kuqin12 (approved):

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
LeviYeoReum: Agree. and this case is for only PEIM which calls `RegisterForShadow()`.
Unfortunately, I couldn't find any good idea for this do automatically but It think it's enough to specifiy in this comment.

Am I missing?

Thanks!

kuqin12 (approved):

lgao4 (approved):

lgao4:
@LeviYeoReum , please fix CI failure.

PR #11057, opened 2025-05-09
OvmfPkg: Clarify Is800155Event

deeglaze:
The Event3 memory comparison is technically correct since the definitions of the struct types are the same. The extended bodies of the events are different. The Event2 size guard for the Event3 comparison should be split to use the Event3 in its sizeof for better clarity.

The large single condition makes the function difficult to understand, so the combined logic is split into different conditional statements.

There is no dedicated library for event Tcg event predicates to deduplicate this logic into for reuse in SecurityPkg and OvmfPkg, but I'm open to suggestion.

# Description

- [ ] Breaking change? No
- [ ] Impacts security? No
- [ ] Includes tests? No

## How This Was Tested

Internal TPM attestation tests.

## Integration Instructions

N/A


ardbiesheuvel (approved):

jyao1:
I am OK with patch.

Could you please fix CI failure?

PR #11063, opened 2025-05-12
MdeModulePkg: Acpi cumulative codeql issues.

apop5:
# Description

Running Codeql on MdeModulePkg/Universal/Acpi drivers results in codeql errors stemming from missing null tests.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Local CI, booting Virutal platform with no ill effects.

## Integration Instructions
No integration necessary.

### DIFF
--- a/MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiSdt.c
+++ b/MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiSdt.c
@@ -336,6 +336,11 @@ SdtRegisterNotify (
   CurrentNotifyList = AllocatePool (sizeof (EFI_ACPI_NOTIFY_LIST));
   ASSERT (CurrentNotifyList != NULL);
 
+  if (CurrentNotifyList == NULL) {
+    DEBUG ((DEBUG_ERROR, "%a Failed to allocate pool\n", __func__));
+    return;
+  }
+
   //
   // Initialize the table contents
   //
@@ -484,6 +489,10 @@ SdtOpenSdtTable (
 
   AmlHandle = AllocatePool (sizeof (*AmlHandle));
   ASSERT (AmlHandle != NULL);
+  if (AmlHandle == NULL) {
+    return EFI_NOT_FOUND;
+  }
+
   AmlHandle->Signature       = EFI_AML_ROOT_HANDLE_SIGNATURE;
   AmlHandle->Buffer          = (VOID *)((UINTN)Table->Table + sizeof (EFI_ACPI_SDT_HEADER));
   AmlHandle->Size            = Table->Table->Length - sizeof (EFI_ACPI_SDT_HEADER);
@@ -560,6 +569,9 @@ SdtOpenEx (
   //
   AmlHandle = AllocatePool (sizeof (*AmlHandle));
   ASSERT (AmlHandle != NULL);
+  if (AmlHandle == NULL) {
+    return EFI_INVALID_PARAMETER;
+  }
 
   AmlHandle->Signature       = EFI_AML_HANDLE_SIGNATURE;
   AmlHandle->Buffer          = Buffer;
@@ -928,6 +940,10 @@ SdtDuplicateHandle (
 
   DstAmlHandle = AllocatePool (sizeof (*DstAmlHandle));
   ASSERT (DstAmlHandle != NULL);
+  if (DstAmlHandle == NULL) {
+    return NULL;
+  }
+
   CopyMem (DstAmlHandle, (VOID *)AmlHandle, sizeof (*DstAmlHandle));
 
   return DstAmlHandle;
@@ -967,6 +983,10 @@ SdtFindPathFromRoot (
     // Duplicate RootHandle
     //
     *HandleOut = (EFI_ACPI_HANDLE)SdtDuplicateHandle (AmlHandle);
+    if (*HandleOut == NULL) {
+      return EFI_INVALID_PARAMETER;
+    }
+
     return EFI_SUCCESS;
   }
 
--- a/MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiTableProtocol.c
+++ b/MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiTableProtocol.c
@@ -538,6 +538,9 @@ AddTableToList (
   //
   CurrentTableList = AllocatePool (sizeof (EFI_ACPI_TABLE_LIST));
   ASSERT (CurrentTableList);
+  if (CurrentTableList == NULL) {
+    return EFI_OUT_OF_RESOURCES;
+  }
 
   //
   // Determine table type and size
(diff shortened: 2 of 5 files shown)


MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiSdt.c
```diff
@@ -336,6 +336,11 @@ SdtRegisterNotify (
   CurrentNotifyList = AllocatePool (sizeof (EFI_ACPI_NOTIFY_LIST));
   ASSERT (CurrentNotifyList != NULL);
 
+  if (CurrentNotifyList == NULL) {
```
samimujawar, line 337: I think the ASSERT should be moved inside the if condition to achieve minor optimisation. This would apply for the other changes in this patch as well.

Also in the current case, would it be good to return an error code from this function EFI_OUT_OF_RESOURCES? However this would mean that the caller would also need to deal with the error code, e.g. line 443 - 444 in this file, and probably elsewhere.

MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiSdt.c
```diff
@@ -484,6 +489,10 @@ SdtOpenSdtTable (
 
   AmlHandle = AllocatePool (sizeof (*AmlHandle));
   ASSERT (AmlHandle != NULL);
+  if (AmlHandle == NULL) {
+    return EFI_NOT_FOUND;
```
samimujawar, line 492: I can see why EFI_NOT_FOUND is being returned, but I think it would be better to return EFI_OUT_OF_RESOURCES. This would mean that the return type of the API would change and would possibly require the callers to fix the code. 

I would prefer this to fix this properly. However, I would leave it to the MdeModulePkg maintainers to decide.

MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiSdt.c
```diff
@@ -560,6 +569,9 @@ SdtOpenEx (
   //
   AmlHandle = AllocatePool (sizeof (*AmlHandle));
   ASSERT (AmlHandle != NULL);
+  if (AmlHandle == NULL) {
+    return EFI_INVALID_PARAMETER;
```
samimujawar: Here gain I think EFI_OUT_OF_RESOURCES should be returned.

MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiSdt.c
```diff
@@ -967,6 +983,10 @@ SdtFindPathFromRoot (
     // Duplicate RootHandle
     //
     *HandleOut = (EFI_ACPI_HANDLE)SdtDuplicateHandle (AmlHandle);
+    if (*HandleOut == NULL) {
+      return EFI_INVALID_PARAMETER;
```
samimujawar: EFI_OUT_OF_RESOURCES?

MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiTableProtocol.c
```diff
@@ -538,6 +538,9 @@ AddTableToList (
   //
   CurrentTableList = AllocatePool (sizeof (EFI_ACPI_TABLE_LIST));
   ASSERT (CurrentTableList);
+  if (CurrentTableList == NULL) {
```
samimujawar, line 541: ASSERT can be moved in the if condition here.

MdeModulePkg/Universal/Acpi/AcpiTableDxe/AmlNamespace.c
```diff
@@ -538,6 +545,9 @@ AmlFindPath (
   RootNameSeg[0]  = AML_ROOT_CHAR;
   RootNameSeg[1]  = 0;
   AmlRootNodeList = AmlCreateNode (RootNameSeg, NULL, AmlHandle->AmlByteEncoding);
+  if (AmlRootNodeList == NULL) {
+    return EFI_OUT_OF_RESOURCES;
```
samimujawar, line 550: It would be good to update the function documentation as well to reflect the new return code.

samimujawar (commented):
@apop5 Thank you for these fixes. I have some minor suggestions that you may want to take a look at. Otherwise, this patch looks good to me.

Also, can you let me know how I can run codeql for a package, please? e.g. I would like to run codeql on say DynamicTablesPkg, etc. If you could point me to the relevant documentation it would be of great help.

MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiSdt.c
apop5: I moved all asserts inside of the conditionals.


Also, thank you for reviewing the code.

MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiSdt.c
apop5: I am hesitant to start changing the return types for public interfaces, mainly because while the process to update the interfaces would take time where this code would continue to susceptible to null dereferences.

MdeModulePkg/Universal/Acpi/AcpiTableDxe/AmlNamespace.c
apop5: I updated all the functions to detail the return statements (internal functions) and to specify that null can be returned.

apop5:
> @apop5 Thank you for these fixes. I have some minor suggestions that you may want to take a look at. Otherwise, this patch looks good to me.
> 
> 
> 
> Also, can you let me know how I can run codeql for a package, please? e.g. I would like to run codeql on say DynamicTablesPkg, etc. If you could point me to the relevant documentation it would be of great help.

The codeQL workflow file (https://github.com/tianocore/edk2/blob/master/.github/workflows/codeql.yml) is where CI runs codeQL.

The important part is that --codeQL parameter is passed during the stuart_update and stuart_ci_build steps. This will download the codeQL cli and it will run it during the build process.

samimujawar (approved):

samimujawar:
> > @apop5 Thank you for these fixes. I have some minor suggestions that you may want to take a look at. Otherwise, this patch looks good to me.
> > Also, can you let me know how I can run codeql for a package, please? e.g. I would like to run codeql on say DynamicTablesPkg, etc. If you could point me to the relevant documentation it would be of great help.
> 
> The codeQL workflow file (https://github.com/tianocore/edk2/blob/master/.github/workflows/codeql.yml) is where CI runs codeQL.
> 
> The important part is that --codeQL parameter is passed during the stuart_update and stuart_ci_build steps. This will download the codeQL cli and it will run it during the build process.

Thank you for the pointers.

apop5:
@lgao4 

These have approvals, are they good to merge?

PR #11066, opened 2025-05-12
MdeModulePkg/HiiDatabaseDxe: Fix BlockSize length

CodingVoid:
# Description

The BlockSize calculation was missing the 0 terminator which caused the string block to shrink by 1 every time the string was processed. Therefore causing memory corruptions, because the string took more memory space as was allocated for the string block therefore corrupting the memory pool at the end (which caused an ASSERT upon trying to free it).

- [ ] Breaking change?
- [x] Impacts security?
- [ ] Includes tests?

## How This Was Tested

No more ASSERTS from the memory allocator.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Universal/HiiDatabaseDxe/String.c
+++ b/MdeModulePkg/Universal/HiiDatabaseDxe/String.c
@@ -1000,7 +1000,7 @@ SetStringWorker (
     case EFI_HII_SIBT_STRING_SCSU_FONT:
     case EFI_HII_SIBT_STRINGS_SCSU:
     case EFI_HII_SIBT_STRINGS_SCSU_FONT:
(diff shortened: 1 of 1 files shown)


lgao4 (approved):

lgao4:
> One minor code formatting issues to resolve to pass CI:
> 
> ```
> ERROR - -      BlockSize  = OldBlockSize + StrSize(String);
> ERROR - +      BlockSize  = OldBlockSize + StrSize (String);
> ```
> 
> @lgao4 This looks like it may candidate for edk2-stable202505

I agree.

lgao4:
@CodingVoid , could you help check what's wrong in CI?

PR #11071, opened 2025-05-13
MdePkg/IndustryStandard: update Tpm2Acpi table to revision 5

LeviYeoReum:
Tpm2AcpiTable revision 5 adds new START_METHOD with Arm FF-A transfer. 
add related information and update its generator.


samimujawar:
I think this patch needs to be split into 2. The MdePkg/Include/IndustryStandard/Tpm2Acpi.h change should be an independent patch.

LeviYeoReum:
> I think this patch needs to be split into 2. The MdePkg/Include/IndustryStandard/Tpm2Acpi.h change should be an independent patch.

Okay. I'll split this. Thanks!

lgao4 (approved):

lgao4:

mdkinney:
Please add the name and version of the specification this content was added.

PR #11075, opened 2025-05-13
MdeModulePkg: SmmCommunicationBuffer cumulative codeql issues.

apop5:
# Description
Running Codeql on MdeModulePkg/Universal/SmmCommunicationBuffer drivers results in codeql errors stemming from missing null tests.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Local CI and booting virtual platform with no ill effects.

## Integration Instructions
No integration necessary.


lgao4:
@apop5 , could you create one PR to include all changes for codeql issue?

apop5:
> @apop5 , could you create one PR to include all changes for codeql issue?

I previously had this:
https://github.com/tianocore/edk2/pull/6266

But from the time that I put it up, new issues were being introduced. I figured there would be a better chance of getting these in by splitting them up so it's less to review.

If you would prefer, I can go back to a monolithic version, but it will be a moving target to get it in while the code continues to progress.

niruiyu (approved):

Issue #11074, opened 2025-05-13
[Bug]: default OVMF build overwrites memory on FreePages() causing page fault if pages are not writable


bluca:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

MdePkg

### Which targets are impacted by this bug?

RELEASE

### Current Behavior

The default value of gEfiMdePkgTokenSpaceGuid.PcdDebugPropertyMask in the OvmfPKgX64.dsc config enables the 0x08 bit, which causes DebugClearMemory() to attempt to overwrite memory freed by FreePages() with a fixed pattern.

In shim we have started marking memory used to load executable images as executable and non-writable, according to the NX workflow requirement. We use AllocatePages() to allocate memory for the image, and FreePages to free it.

When an image loading/starting goes wrong, we attempt to free it. Given it's an executable image, it's marked as non-writable. DebugClearMemory causes a page fault when it attempts to overwrite it:

```
systemd-stub@0x72c64000 v999-bluca
FSOpen: Open '\loader\addons' Success
FSOpen: Open '\loader\addons\1.addon.efi' Success
InstallProtocolInterface: 6E6BAEB8-7108-4179-949D-A3493415EC97 7D368D18
InstallProtocolInterface: 5B1B31A1-9562-11D2-8E3F-00A0C969723B 7D368D18
InstallProtocolInterface: BC62157E-3E33-4FEC-9920-2D3B36D750DF 7D376D18
!!!! X64 Exception Type - 0E(#PF - Page-Fault)  CPU Apic ID - 00000000
!!!! ExceptionData - 0000000000000003  I:0 R:0 U:0 W:1 P:1 PK:0 SS:0 SGX:0
RIP  - 000000007EF0931A, CS  - 0000000000000038, RFLAGS - 0000000000010206
RAX  - 00000000000000AF, RCX - 0000000000005000, RDX - 000000007D352000
RBX  - 000000007D352000, RSP - 000000007EEEAAF0, RBP - 000000007EEEAB80
RSI  - 000000007EF10F20, RDI - 000000007D353000
R8   - 00000000000000AF, R9  - 000000007D357FFF, R10 - 0000000072C7C2F7
R11  - 0000000001CBD9D0, R12 - 000000007D357FFF, R13 - 000000007D358000
R14  - 000000007D35B300, R15 - 000000007D357FFF
[...]

PR #11077, opened 2025-05-14
UefiCpuPkg: Apply ReadOnly on Ap loop buffers.

apop5:
# Description

In the MpInitLib, pages are allocated for
the APs, and data is copied into them. Depending
on the allocation type, Nx needs to be removed
from the pages to allow the processors to execute.

Additionally, ReadOnly needs to be applied to the
pages after they have been filled with the
instructions that the APs need to execute.

Found by running PagingAudit test application on system and found a single section as RWX.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Booted virtual platform to windows with no page faults.
Passed local CI.
Ran a PagingAudit report for system and after changes, no memory was marked RWX.

## Integration Instructions
no integration necessary.

### DIFF
--- a/MdeModulePkg/MdeModulePkg.dec
+++ b/MdeModulePkg/MdeModulePkg.dec
@@ -1508,8 +1508,7 @@
   #  OEM Reserved       0x4000000000000000<BR>
   #  OS Reserved        0x8000000000000000<BR>
   #
-  # NOTE: User must NOT set NX protection for EfiLoaderCode / EfiBootServicesCode / EfiRuntimeServicesCode. <BR>
-  #       User MUST set the same NX protection for EfiBootServicesData and EfiConventionalMemory. <BR>
+  # NOTE: User MUST set the same NX protection for EfiBootServicesData and EfiConventionalMemory. <BR>
   #
   # e.g. 0x7FD5 can be used for all memory except Code. <BR>
   # e.g. 0x7BD4 can be used for all memory except Code and ACPINVS/Reserved. <BR>
--- a/UefiCpuPkg/Library/MpInitLib/DxeMpLib.c
+++ b/UefiCpuPkg/Library/MpInitLib/DxeMpLib.c
@@ -160,7 +160,7 @@ GetWakeupBuffer (
   @retval 0       Cannot find free memory below 4GB.
 **/
 UINTN
-AllocateCodeBuffer (
+AllocateCodePage (
   IN UINTN  BufferSize
   )
 {
@@ -417,14 +417,11 @@ AllocateApLoopCodeBuffer (
 /**
   Remove Nx protection for the range specific by BaseAddress and Length.
 
-  The PEI implementation uses CpuPageTableLib to change the attribute.
-  The DXE implementation uses gDS to change the attribute.
-
   @param[in] BaseAddress  BaseAddress of the range.
   @param[in] Length       Length of the range.
 **/
 VOID
-RemoveNxprotection (
+RemoveNxProtection (
   IN EFI_PHYSICAL_ADDRESS  BaseAddress,
   IN UINTN                 Length
   )
@@ -432,17 +429,58 @@ RemoveNxprotection (
   EFI_STATUS                       Status;
   EFI_GCD_MEMORY_SPACE_DESCRIPTOR  MemDesc;
 
-  //
-  // TODO: Check EFI_MEMORY_XP bit set or not once it's available in DXE GCD
-  //       service.
-  //
   Status = gDS->GetMemorySpaceDescriptor (BaseAddress, &MemDesc);
   if (!EFI_ERROR (Status)) {
-    gDS->SetMemorySpaceAttributes (
-           BaseAddress,
-           Length,
-           MemDesc.Attributes & (~EFI_MEMORY_XP)
-           );
+    if (((MemDesc.Capabilities & EFI_MEMORY_XP) == EFI_MEMORY_XP) && ((MemDesc.Attributes & EFI_MEMORY_XP) == EFI_MEMORY_XP)) {
+      Status = gDS->SetMemorySpaceAttributes (
+                      BaseAddress,
+                      Length,
+                      MemDesc.Attributes & (~EFI_MEMORY_XP)
+                      );
+
+      if (EFI_ERROR (Status)) {
+        DEBUG ((DEBUG_ERROR, "%a - Setting Nx on 0x%p returned %r\n", __func__, BaseAddress, Status));
+        ASSERT_EFI_ERROR (Status);
+      }
+    }
+  } else {
+    DEBUG ((DEBUG_ERROR, "%a - Memory Address was not found in Memory map! %lp %r\n", __func__, BaseAddress, Status));
+    ASSERT_EFI_ERROR (Status);
+  }
+}
+
+/**
+  Add ReadOnly protection to the range specified by BaseAddress and Length.
+
+  @param[in] BaseAddress  BaseAddress of the range.
+  @param[in] Length       Length of the range.
+**/
+VOID
+ApplyRoProtection (
+  IN EFI_PHYSICAL_ADDRESS  BaseAddress,
+  IN UINTN                 Length
+  )
+{
+  EFI_STATUS                       Status;
+  EFI_GCD_MEMORY_SPACE_DESCRIPTOR  MemDesc;
+
+  Status = gDS->GetMemorySpaceDescriptor (BaseAddress, &MemDesc);
+  if (!EFI_ERROR (Status)) {
+    if (((MemDesc.Capabilities & EFI_MEMORY_RO) == EFI_MEMORY_RO) && ((MemDesc.Attributes & EFI_MEMORY_RO) != EFI_MEMORY_RO)) {
+      Status = gDS->SetMemorySpaceAttributes (
+                      BaseAddress,
+                      Length,
+                      MemDesc.Attributes | EFI_MEMORY_RO
+                      );
+
+      if (EFI_ERROR (Status)) {
+        DEBUG ((DEBUG_ERROR, "%a - Setting Ro on 0x%p returned %r\n", __func__, BaseAddress, Status));
+        ASSERT_EFI_ERROR (Status);
+      }
+    }
+  } else {
+    DEBUG ((DEBUG_ERROR, "%a - Memory Address was not found in Memory map! %lp %r\n", __func__, BaseAddress, Status));
+    ASSERT_EFI_ERROR (Status);
   }
 }
 
--- a/UefiCpuPkg/Library/MpInitLib/MpLib.c
+++ b/UefiCpuPkg/Library/MpInitLib/MpLib.c
@@ -2072,6 +2072,7 @@ MpInitLibInitialize (
   UINTN                    ApResetVectorSizeAbove1Mb;
   UINTN                    BackupBufferAddr;
   UINTN                    ApIdtBase;
+  IA32_CR0                 Cr0;
 
   FirstMpHandOff = GetNextMpHandOffHob (NULL);
   if (FirstMpHandOff != NULL) {
@@ -2248,7 +2249,13 @@ MpInitLibInitialize (
   // Copy all 32-bit code and 64-bit code into memory with type of
   // EfiBootServicesCode to avoid page fault if NX memory protection is enabled.
   //
-  CpuMpData->WakeupBufferHigh = AllocateCodeBuffer (ApResetVectorSizeAbove1Mb);
+  CpuMpData->WakeupBufferHigh = AllocateCodePage (ApResetVectorSizeAbove1Mb);
+
+  Cr0.UintN = AsmReadCr0 ();
+  if (Cr0.Bits.PG != 0) {
+    RemoveNxProtection ((EFI_PHYSICAL_ADDRESS)(UINTN)CpuMpData->WakeupBufferHigh, ALIGN_VALUE (ApResetVectorSizeAbove1Mb, EFI_PAGE_SIZE));
+  }
+
   CopyMem (
     (VOID *)CpuMpData->WakeupBufferHigh,
     CpuMpData->AddressMap.RendezvousFunnelAddress +
@@ -2256,6 +2263,9 @@ MpInitLibInitialize (
     ApResetVectorSizeAbove1Mb
     );
   DEBUG ((DEBUG_INFO, "AP Vector: non-16-bit = %p/%x\n", CpuMpData->WakeupBufferHigh, ApResetVectorSizeAbove1Mb));
+  if (Cr0.Bits.PG != 0) {
+    ApplyRoProtection ((EFI_PHYSICAL_ADDRESS)(UINTN)CpuMpData->WakeupBufferHigh, ALIGN_VALUE (ApResetVectorSizeAbove1Mb, EFI_PAGE_SIZE));
+  }
 
   //
   // Save APIC mode for AP to sync
@@ -3484,7 +3494,7 @@ PrepareApLoopCode (
     // Make sure that the buffer memory is executable if NX protection is enabled
     // for EfiReservedMemoryType.
     //
-    RemoveNxprotection (Address, EFI_PAGES_TO_SIZE (FuncPages));
+    RemoveNxProtection (Address, EFI_PAGES_TO_SIZE (FuncPages));
   }
 
   mReservedTopOfApStack = (UINTN)Address + EFI_PAGES_TO_SIZE (StackPages+FuncPages);
@@ -3492,6 +3502,10 @@ PrepareApLoopCode (
   mReservedApLoop.Data = (VOID *)(UINTN)Address;
   ASSERT (mReservedApLoop.Data != NULL);
   CopyMem (mReservedApLoop.Data, ApLoopFunc, ApLoopFuncSize);
+  if (Cr0.Bits.PG != 0) {
+    ApplyRoProtection (Address, EFI_PAGES_TO_SIZE (FuncPages));
+  }
+
   if (!CpuMpData->UseSevEsAPMethod) {
     //
     // processors without SEV-ES and paging is enabled
--- a/UefiCpuPkg/Library/MpInitLib/PeiMpLib.c
+++ b/UefiCpuPkg/Library/MpInitLib/PeiMpLib.c
@@ -290,7 +290,7 @@ GetWakeupBuffer (
   @retval 0       Cannot find free memory below 4GB.
 **/
 UINTN
-AllocateCodeBuffer (
+AllocateCodePage (
   IN UINTN  BufferSize
   )
 {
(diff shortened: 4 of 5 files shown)


UefiCpuPkg/Library/MpInitLib/DxeMpLib.c
```diff
@@ -432,17 +433,55 @@ RemoveNxprotection (
...
+  IN EFI_PHYSICAL_ADDRESS  BaseAddress,
+  IN UINTN                 Length
+  )
+{
+  EFI_STATUS                       Status;
+  EFI_GCD_MEMORY_SPACE_DESCRIPTOR  MemDesc;
+
+  Status = gDS->GetMemorySpaceDescriptor (BaseAddress, &MemDesc);
+  if (!EFI_ERROR (Status)) {
+    if ((MemDesc.Attributes & EFI_MEMORY_RO) != EFI_MEMORY_RO) {
```
os-d, line 473: We should check the capabilities first and set them if required.

VivianNK (approved):

UefiCpuPkg/Library/MpInitLib/PeiMpLib.c
```diff
@@ -903,3 +900,19 @@ RemoveNxprotection (
...
+**/
+VOID
+ApplyReadOnlyMemoryProtection (
+  IN EFI_PHYSICAL_ADDRESS  BaseAddress,
+  IN UINTN                 Length
+  )
+{
+  // stub for now, page tables are limited in PEI
+  return;
+}
```
niruiyu: PEI version of ApplyReadOnlyMemoryProtection can be implemented through PageTbaleLib.

UefiCpuPkg/Library/MpInitLib/DxeMpLib.c
```diff
@@ -417,32 +418,70 @@ AllocateApLoopCodeBuffer (
...
+                      );
+
+      if (EFI_ERROR (Status)) {
+        DEBUG ((DEBUG_ERROR, "%a - Setting Ro on 0x%p returned %r\n", __func__, BaseAddress, Status));
+        ASSERT_EFI_ERROR (Status);
+      }
+    } else {
+      DEBUG ((DEBUG_ERROR, "%a - Memory Address was not found in Memory map! %lp %r\n", __func__, BaseAddress, Status));
+      ASSERT_EFI_ERROR (Status);
+    }
```
niruiyu: this ASSERT_EFI_ERROR() is meaningless as the entire code runs under Status == EFI_SUCCESS branch.

UefiCpuPkg/Library/MpInitLib/DxeMpLib.c
```diff
@@ -417,32 +418,70 @@ AllocateApLoopCodeBuffer (
...
+                      Length,
+                      MemDesc.Attributes & (~EFI_MEMORY_XP)
+                      );
+
+      if (EFI_ERROR (Status)) {
+        DEBUG ((DEBUG_ERROR, "%a - Setting Ro on 0x%p returned %r\n", __func__, BaseAddress, Status));
+        ASSERT_EFI_ERROR (Status);
+      }
+    } else {
+      DEBUG ((DEBUG_ERROR, "%a - Memory Address was not found in Memory map! %lp %r\n", __func__, BaseAddress, Status));
```
niruiyu: The debug message is wrong.

UefiCpuPkg/Library/MpInitLib/DxeMpLib.c
apop5: should be fixed in the latest push (fixed where the debug messages to the correct status checks)

UefiCpuPkg/Library/MpInitLib/DxeMpLib.c
apop5: should be fixed in the latest push (fixed where the debug messages to the correct status checks)

UefiCpuPkg/Library/MpInitLib/PeiMpLib.c
niruiyu: Are you going to implement real logic here as what's been done in RemoveNxProtection()?
If not, at least please update the comments to NOT say "page tables are limited in PEI".

apop5: Yes, I plan on looking at it tomorrow.

Do you know if there are any platforms making real use of paging in the PEI phase? The two platforms I checked did not have the PE bit set in CR0.

UefiCpuPkg/Library/MpInitLib/MpLib.c
```diff
@@ -2244,6 +2244,7 @@ MpInitLibInitialize (
   // EfiBootServicesCode to avoid page fault if NX memory protection is enabled.
   //
   CpuMpData->WakeupBufferHigh = AllocateCodeBuffer (ApResetVectorSizeAbove1Mb);
+  RemoveNxProtection ((EFI_PHYSICAL_ADDRESS)(UINTN)CpuMpData->WakeupBufferHigh, ALIGN_VALUE (ApResetVectorSizeAbove1Mb, EFI_PAGE_SIZE));
```
LiuZhiguang001: Here, the buffer is allocated by Allocate**Code**Buffer. Don't need to remove NX protection anymore.

UefiCpuPkg/Library/MpInitLib/PeiMpLib.c
apop5: It should be updated now if you want to review it.

UefiCpuPkg/Library/MpInitLib/MpLib.c
apop5: @os-d 

I believe that this still requires nx removal, and RO applied after the allocation occurs. Otherwise how would the code ever be able to copy something in the allocated pages?

os-d: Correct, all `AllocateCodeBuffer` does is ensure the buffer is below 4GB: https://github.com/tianocore/edk2/blob/a808062890de96d8f816aa920d9b5cf13151a40c/UefiCpuPkg/Library/MpInitLib/DxeMpLib.c#L163. With NX set it will come back as NX, so NX must be removed to write to it, then RO set.

UefiCpuPkg/Library/MpInitLib/DxeMpLib.c
```diff
@@ -468,6 +507,15 @@ MpInitChangeApLoopCallback (
   CpuMpData->Pm16CodeSegment = GetProtectedMode16CS ();
   CpuMpData->ApLoopMode      = PcdGet8 (PcdCpuApLoopMode);
   mNumberToFinish            = CpuMpData->CpuCount - 1;
+  RemoveNxProtection (
```
LiuZhiguang001: Have you observed any issue here without this PR?
Per my understanding, this mReservedApLoop buffer is already removed NX limitation when it is allocated in MpLib.c.
Your change here is not needed.

Also, why use RelocateApLoopFuncSizeAmdSev instead of RelocateApLoopFuncSizeGeneric as buffer size?

UefiCpuPkg/Library/MpInitLib/MpLib.c
LiuZhiguang001: no, AllocateCodeBuffer will make sure the buffer's nx bit is removed.

image

Please notice the allocated buffer type is EfiBootServices**Code**

os-d: That is incorrect. It is up to the platform whether EfiBootServicesCode gets NX applied to it via PcdDxeNxMemoryProtectionPolicy. A typical platform decision will be to set NX on code regions and code sections must have NX cleared such as here in the image loader: https://github.com/tianocore/edk2/blob/e83a694c0ff800820b21e5f6ecd3fa8d944674d4/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c#L351-L360, then data sections have NX set.

The core does not do anything special for code regions, it is up to the platform policy set in the PCD.

UefiCpuPkg/Library/MpInitLib/DxeMpLib.c
apop5: I retested after we moved the nx removal to the MpInitLibInitialize, the hang is gone for the aps at exit boot services.

I backed out the changes to the MpInitChangeApLoopCallback.

The second question for RelocateApLoopFuncSizeGeneric vs RelocateApLoopFuncSizeAmdSev  is no longer relevant without the backed out change.

RelocateApLoopFuncSizeAmdSev is a larger size than RelocateApLoopFuncSizeGeneric, but since both of them are being upsized to a page of memory, it didn't matter which one was used because both were easily fitting inside of a page.

UefiCpuPkg/Library/MpInitLib/MpLib.c
apop5: @LiuZhiguang001 

Do you have any other concerns?

LiuZhiguang001: I disagree
image

Platform must NOT set NX protection for EfiBootServicesCode

os-d: That is an outdate comment from 8 years ago

image

Platforms do set NX for code types, which is valid to be secure by default, requiring . Some bootloaders for instance will allocate everything as EfiLoaderCode, including standard allocations. These should be non-executable.

Beyond this, the DXE Core does not guarantee that it is handing out RWX pages (see discussion here, e.g. https://github.com/tianocore/edk2/issues/11074). If some other entity had this page previously and had set NX on it and then freed it, it is no guarantee that DXE Core will clear that when allocating it to MpLib.

Your proposal to not touch the memory protections leaves open the possibility of crash (and we see it crashing on platforms today). @apop5 's proposal to correctly set the memory protections both fixes the existing crash and even if a platform is running without NX protections at all, then this is effectively a no-op, no harm done. So it is far safer to be explicit about what is needed, which this needs to be executable. It should be marked as such.

LiuZhiguang001: Thanks for the information. I am not aware of such usage. 
However, if you think such usage is correct and the comments is out of date, can you please send another PR to update the comments?

Before the new PR is merged, I can only assume the comment is correct. Based on this assumption, the possibility of crash is caused by the issue from DXE core. We should fix the DXE core to make sure it can clear the NX bit when allocating EfiLoaderCode type buffer. Current change is more like a WA for me.

UefiCpuPkg/Library/MpInitLib/MpLib.c
```diff
@@ -2249,13 +2250,22 @@ MpInitLibInitialize (
   // EfiBootServicesCode to avoid page fault if NX memory protection is enabled.
   //
   CpuMpData->WakeupBufferHigh = AllocateCodeBuffer (ApResetVectorSizeAbove1Mb);
+
+  Cr0.UintN = AsmReadCr0 ();
+  if (Cr0.Bits.PG != 0) {
+    RemoveNxProtection ((EFI_PHYSICAL_ADDRESS)(UINTN)CpuMpData->WakeupBufferHigh, ALIGN_VALUE (ApResetVectorSizeAbove1Mb, EFI_PAGE_SIZE));
+  }
```
niruiyu, line 2257: I do not think it's needed as Code page is executable.
If here RemoveNxProtection() is needed, how about other code that allocates code pages?
e.g.: https://github.com/intel-restricted/firmware.boot.uefi.iafw.externalmirror.edk2/blob/bd994a11b406ef02aacefad42b7a870c30b84b31/MdeModulePkg/Core/Dxe/Image/Image.c#L695

apop5: Hi Ray,

It doesn't work exactly like that. Lets take the example of the most secure system possible, a system that has memory protections set to their highest level (Ro on Code Buffers, Nx on all other buffers, Wp on read only data, etc)

When an allocation occurs, expecting the memory protections to be set to their final state is not possible. Take the code buffer allocation occurring here. For a code allocation, this would imply that Ro was already set for the memory region when the address was returned, which would prevent the system from copying code into the new allocation. The GCD/Page tables expect the caller to make sure that memory protections are set upon the buffers after they have initialized the allocated memory.

As @os-d mentioned [here](https://github.com/tianocore/edk2/pull/11077#discussion_r2162020014) we have already seen cases where memory contains attributes other than expected. We need to be able to handle these edge cases correctly to prevent an exception from occurring. As it stands today, if this driver is run and memory protections are configured for their highest setting, the system will hang/generate an exception. 

Having Edk2 drivers perform code allocations is rare in edk2, most of the code allocations are handled in the core through load image/start image, which allows the dxe core to be able to configure memory protections correctly. This is handling one of those edge cases were a driver is directly allocating code regions and configuring them to have maximum security.

UefiCpuPkg/Library/MpInitLib/MpLib.c
apop5: I added the commit to remove the out of date comments in MdeModulePkg.dec.

UefiCpuPkg/Library/MpInitLib/MpLib.c
apop5: The function you are pointing to (CoreLoadPeImage) https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Core/Dxe/Image/Image.c#L686

is called from CoreLoadImageCommon, and down at the bottom of that function, you can see a call to ProtectUefiImage
https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Core/Dxe/Image/Image.c#L1463


As I mentioned above, the DxeCore already has the capabilities built into it to ensure memory protections are applied to the appropriate sections of the PECoff images. This PR is just to clean up the case where the AP wake up buffers are being manually allocated and not going through the existing DxeCore infrastructure.

UefiCpuPkg/Library/MpInitLib/MpLib.c
LiuZhiguang001: Thanks. 
hi @lgao4 , please help review the change in MdeModulePkg.dec

MdeModulePkg/Universal/RegularExpressionDxe/oniguruma
lgao4, line 1: Seemly, this change for onigurma is not related to this PR. Please confirm.

MdeModulePkg/MdeModulePkg.dec
```diff
@@ -1515,8 +1515,7 @@
   #  OEM Reserved       0x4000000000000000<BR>
   #  OS Reserved        0x8000000000000000<BR>
   #
-  # NOTE: User must NOT set NX protection for EfiLoaderCode / EfiBootServicesCode / EfiRuntimeServicesCode. <BR>
-  #       User MUST set the same NX protection for EfiBootServicesData and EfiConventionalMemory. <BR>
```
lgao4, line 1512: OK. The same NX protection is not requried any more. Right?

apop5: Correct, this comment is outdated, and it was exposed through this PR that there was an outdated comment.

MdeModulePkg/Universal/RegularExpressionDxe/oniguruma
apop5: correct, it was incorrectly added. 

Removed in latest.

apop5:
@niruiyu @LiuZhiguang001 @lgao4 

Any other concerns/questions for this PR?

lgao4:
> @niruiyu @LiuZhiguang001 @lgao4
> 
> Any other concerns/questions for this PR?

I have no other comments.

lgao4 (approved):

LiuZhiguang001 (approved):

LiuZhiguang001:
> @niruiyu @LiuZhiguang001 @lgao4
> 
> Any other concerns/questions for this PR?

I have no concerns now. Thanks for the contribution.

UefiCpuPkg/Library/MpInitLib/DxeMpLib.c
```diff
@@ -417,32 +417,70 @@ AllocateApLoopCodeBuffer (
...
+  if (!EFI_ERROR (Status)) {
+    if (((MemDesc.Capabilities & EFI_MEMORY_XP) == EFI_MEMORY_XP) && ((MemDesc.Attributes & EFI_MEMORY_XP) == EFI_MEMORY_XP)) {
+      Status = gDS->SetMemorySpaceAttributes (
+                      BaseAddress,
+                      Length,
+                      MemDesc.Attributes & (~EFI_MEMORY_XP)
+                      );
+
+      if (EFI_ERROR (Status)) {
+        DEBUG ((DEBUG_ERROR, "%a - Setting Ro on 0x%p returned %r\n", __func__, BaseAddress, Status));
```
niruiyu: The code here is to remove XP, not set RO. Can you update the debug messages?

UefiCpuPkg/Library/MpInitLib/MpLib.c
niruiyu, line 2257: Can you rename AllocateCodeBuffer() to AllocateCodePage()?
So that we guarantee that RemoveNxProtection() does not remove NX for memory region after "AP Reset Vector Above 1MB".

UefiCpuPkg/Library/MpInitLib/MpLib.c
```diff
@@ -2249,13 +2250,22 @@ MpInitLibInitialize (
...
+
   CopyMem (
     (VOID *)CpuMpData->WakeupBufferHigh,
     CpuMpData->AddressMap.RendezvousFunnelAddress +
     CpuMpData->AddressMap.ModeTransitionOffset,
     ApResetVectorSizeAbove1Mb
     );
   DEBUG ((DEBUG_INFO, "AP Vector: non-16-bit = %p/%x\n", CpuMpData->WakeupBufferHigh, ApResetVectorSizeAbove1Mb));
+  if (Cr0.Bits.PG != 0) {
+    ApplyReadOnlyMemoryProtection ((EFI_PHYSICAL_ADDRESS)(UINTN)CpuMpData->WakeupBufferHigh, ALIGN_VALUE (ApResetVectorSizeAbove1Mb, EFI_PAGE_SIZE));
```
niruiyu: Can you please name the two functions in a consistent way?
* RemoveNxProtection
* ApplyReadOnlyMemoryProtection

Either they can be:
* RemoveNxProtection
* ApplyRoProtection
or:
* RemoveNonExecutableMemoryProtection
* ApplyReadOnlyMemoryProtection

niruiyu (commented):
2 minor comments.

UefiCpuPkg/Library/MpInitLib/MpLib.c
apop5: Fixed in latest.

Went with your first suggestion, RemoveNxProtection/ApplyRoProtection

UefiCpuPkg/Library/MpInitLib/MpLib.c
apop5: Done in latest.

UefiCpuPkg/Library/MpInitLib/DxeMpLib.c
apop5: Yes, thanks for catching that.

niruiyu (approved):

Issue #10105, opened 2021-06-02
EFI_MM_COMMUNICATE_HEADER has architecture dependent sized field (Bugzilla Bug 3430)


tianocore-issues:
This issue was created automatically with [bugzilla2github](https://github.com/makubacki/bugzilla2github/tree/bz2gh_tianocore)
# Bugzilla Bug 3430
Date: 2021-06-02T21:41:04+00:00
From: @kuqin12
To:   @kuqin12
CC:   @lgao4

Last updated: 2022-01-14T15:21:30+00:00

tianocore-issues:
## Comment 16254
Date: 2021-06-02 21:41:04 +0000
From: @kuqin12

- Industry Specification: UEFI Platform Initialization Specification Specification
- Target OS: ---
- Bugzilla Assignee(s): @kuqin12

In PI Spec v1.7 Errata A, Vol.4, Sec 5.7 MM Communication Protocol, the MessageLength field of EFI_MM_COMMUNICATE_HEADER (also defined as EFI_SMM_COMMUNICATE_HEADER) is currently defined as type UINTN.

But this structure, as a generic definition, could be used for both PEI and DXE MM communication. Thus for a system that supports PEI MM launch, but operates PEI in 32bit mode and MM foundation in 64bit, the current EFI_MM_COMMUNICATE_HEADER definition will cause structure parse error due to UINTN used.

The suggested change is to make the MessageLength field defined with definitive size as below:
```
typedef struct {
EFI_GUID  HeaderGuid;
UINT64    MessageLength;
UINT8     Data[ANYSIZE_ARRAY];
} EFI_MM_COMMUNICATE_HEADER;
```

This change will impact the structure consumers including:
MdeModulePkg/Core/PiSmmCore
MdeModulePkg/Application/SmiHandlerProfileInfo
MdeModulePkg/Application/MemoryProfileInfo

Note: MdeModulePkg/Library/SmmLockBoxLib/SmmLockBoxPeiLib is also consuming this structure, but it handled this size discrepancy internally. (Without the spec change in this ticket, all applicable PEI MM communicate callers will need to engage the same routine)

The reference implementation is tracked by ticket: https://bugzilla.tianocore.org/show_bug.cgi?id=3398.

tianocore-issues:
## Comment 16307
Date: 2021-06-08 21:44:37 +0000
From: @lgao4

Kun is working on it.

tianocore-issues:
## Comment 16325
[...]

Issue #10094, opened 2021-05-17
EFI_MM_COMMUNICATE_HEADER has architecture dependent sized field (Bugzilla Bug 3398)


tianocore-issues:
This issue was created automatically with [bugzilla2github](https://github.com/makubacki/bugzilla2github/tree/bz2gh_tianocore)
# Bugzilla Bug 3398
Date: 2021-05-17T23:19:36+00:00
From: @kuqin12
To:   @kuqin12
CC:   @lgao4

Last updated: 2022-01-14T15:21:08+00:00

tianocore-issues:
## Comment 16087
Date: 2021-05-17 23:19:36 +0000
From: @kuqin12

- Industry Specification: ---
- Release Observed: EDK II Master
- Releases to Fix: EDK II Master
- Target OS: ---
- Bugzilla Assignee(s): @kuqin12

The MessageLength field of EFI_MM_COMMUNICATE_HEADER (also defined as EFI_SMM_COMMUNICATE_HEADER) is currently defined as type UINTN per https://github.com/tianocore/edk2/blob/1fbf5e30ae8eb725f4e10984f7b0a208f78abbd0/MdePkg/Include/Protocol/MmCommunication.h#L29.

But this structure, as a generic definition, could be used for both PEI and DXE MM communication. Thus for a system that supports PEI MM launch, but operates PEI in 32bit mode and MM foundation in 64bit, the current EFI_MM_COMMUNICATE_HEADER definition will cause structure parse error due to UINTN used.

The suggested fix is to change this field defined as UINT64: https://github.com/kuqin12/mu_basecore/commit/b396dfe54b5f7556a12622cd94a5fa869086c4d3

tianocore-issues:
## Comment 16101
Date: 2021-05-19 04:21:36 +0000
From: @kuqin12

Note that this change could cause some modules to fail to build. So far identified:
https://github.com/tianocore/edk2/blob/15ee7b76891a78141e6e30ef3f8572e8d6b326d2/MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c#L530

https://github.com/tianocore/edk2/blob/15ee7b76891a78141e6e30ef3f8572e8d6b326d2/MdeModulePkg/Application/SmiHandlerProfileInfo/SmiHandlerProfileInfo.c#L119

[...]

PR #11079, opened 2025-05-14
Extend MM communication v3 to ARM and UEFI variable modules

kuqin12:
# Description

This change adds support of MM communicate v3 for ARM based drivers and moved variable driver to use MM communicate v3.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This change was tested on QEMU Q35 and SBSA machines and booted to UEFI shell.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
+++ b/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
@@ -18,6 +18,7 @@
 #include <Library/UefiRuntimeServicesTableLib.h>
 
 #include <Protocol/MmCommunication2.h>
+#include <Protocol/MmCommunication3.h>
 
 #include <IndustryStandard/ArmStdSmc.h>
 #include <IndustryStandard/ArmFfaSvc.h>
@@ -148,6 +149,181 @@ SendSpmMmCommunicate (
 
   This function provides a service to send and receive messages from a registered UEFI service.
 
+  @param[in, out] CommBufferPhysical  Physical address of the MM communication buffer
+  @param[in, out] CommBufferVirtual   Virtual address of the MM communication buffer
+  @param[in, out] CommSize            The size of the data buffer being passed in. On input,
+                                      when not omitted, the buffer should cover EFI_MM_COMMUNICATE_HEADER
+                                      and the value of MessageLength field. On exit, the size
+                                      of data being returned. Zero if the handler does not
+                                      wish to reply with any data. This parameter is optional
+                                      and may be NULL.
+
+  @retval EFI_SUCCESS            The message was successfully posted.
+  @retval EFI_INVALID_PARAMETER  CommBufferPhysical or CommBufferVirtual was NULL, or
+                                 integer value pointed by CommSize does not cover
+                                 EFI_MM_COMMUNICATE_HEADER and the value of MessageLength
+                                 field.
+  @retval EFI_BAD_BUFFER_SIZE    The buffer is too large for the MM implementation.
+                                 If this error is returned, the MessageLength field
+                                 in the CommBuffer header or the integer pointed by
+                                 CommSize, are updated to reflect the maximum payload
+                                 size the implementation can accommodate.
+  @retval EFI_ACCESS_DENIED      The CommunicateBuffer parameter or CommSize parameter,
+                                 if not omitted, are in address range that cannot be
+                                 accessed by the MM environment.
+
+**/
+STATIC
+EFI_STATUS
+EFIAPI
+MmCommunicationCommon (
+  IN OUT VOID   *CommBufferPhysical,
+  IN OUT VOID   *CommBufferVirtual,
+  IN OUT UINTN  *CommSize OPTIONAL
+  )
+{
+  EFI_MM_COMMUNICATE_HEADER     *CommunicateHeader;
+  EFI_MM_COMMUNICATE_HEADER_V3  *CommunicateHeaderV3;
+  UINTN                         BufferSize;
+  UINTN                         *MessageSize;
+  UINTN                         HeaderSize;
+  EFI_STATUS                    Status;
+
+  Status     = EFI_ACCESS_DENIED;
+  BufferSize = 0;
+
+  //
+  // Check parameters
+  //
+  if ((CommBufferVirtual == NULL) || (CommBufferPhysical == NULL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  Status              = EFI_SUCCESS;
+  CommunicateHeader   = CommBufferVirtual;
+  CommunicateHeaderV3 = NULL;
+  // CommBuffer is a mandatory parameter. Hence, Rely on
+  // MessageLength + Header to ascertain the
+  // total size of the communication payload rather than
+  // rely on optional CommSize parameter
+  if (CompareGuid (
+        &CommunicateHeader->HeaderGuid,
+        &gEfiMmCommunicateHeaderV3Guid
+        ))
+  {
+    CommunicateHeaderV3 = (EFI_MM_COMMUNICATE_HEADER_V3 *)CommBufferVirtual;
+    BufferSize          = CommunicateHeaderV3->BufferSize;
+    MessageSize         = &CommunicateHeaderV3->MessageSize;
+    HeaderSize          = sizeof (EFI_MM_COMMUNICATE_HEADER_V3);
+  } else {
+    BufferSize = CommunicateHeader->MessageLength +
+                 sizeof (CommunicateHeader->HeaderGuid) +
+                 sizeof (CommunicateHeader->MessageLength);
+    MessageSize = &CommunicateHeader->MessageLength;
+    HeaderSize  = sizeof (CommunicateHeader->HeaderGuid) +
+                  sizeof (CommunicateHeader->MessageLength);
+  }
+
+  // If CommSize is not omitted, perform size inspection before proceeding.
+  if (CommSize != NULL) {
+    // This case can be used by the consumer of this driver to find out the
+    // max size that can be used for allocating CommBuffer.
+    if ((*CommSize == 0) ||
+        (*CommSize > mNsCommBuffMemRegion.Length))
+    {
+      *CommSize = mNsCommBuffMemRegion.Length;
+      Status    = EFI_BAD_BUFFER_SIZE;
+    }
+
+    //
+    // CommSize should cover at least MessageLength + sizeof (EFI_MM_COMMUNICATE_HEADER);
+    //
+    if (*CommSize < BufferSize) {
+      Status = EFI_INVALID_PARAMETER;
+    }
+  }
+
+  //
+  // If the message length is 0 or greater than what can be tolerated by the MM
+  // environment then return the expected size.
+  //
+  if ((*MessageSize == 0) ||
+      (BufferSize > mNsCommBuffMemRegion.Length))
+  {
+    *MessageSize = mNsCommBuffMemRegion.Length - HeaderSize;
+    Status       = EFI_BAD_BUFFER_SIZE;
+  }
+
+  // MessageLength or CommSize check has failed, return here.
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  // Copy Communication Payload
+  CopyMem ((VOID *)mNsCommBuffMemRegion.VirtualBase, CommBufferVirtual, BufferSize);
+
+  if (IsFfaSupported ()) {
+    Status = SendFfaMmCommunicate ();
+  } else {
+    Status = SendSpmMmCommunicate ();
+  }
+
+  if (!EFI_ERROR (Status)) {
+    ZeroMem (CommBufferVirtual, BufferSize);
+    // On successful return, the size of data being returned is inferred from
+    // MessageLength + Header.
+    CommunicateHeader = (EFI_MM_COMMUNICATE_HEADER *)mNsCommBuffMemRegion.VirtualBase;
+    if ((CommunicateHeaderV3 != NULL) && !CompareGuid (
+                                            &CommunicateHeader->HeaderGuid,
+                                            &gEfiMmCommunicateHeaderV3Guid
+                                            ))
+    {
+      // Sanity check to make sure we are still using v3
+      DEBUG ((
+        DEBUG_ERROR,
+        "%a Expected v3 header, but got v1 header! %g\n",
+        __func__,
+        &CommunicateHeader->HeaderGuid
+        ));
+      return EFI_INVALID_PARAMETER;
+    }
+
+    if (CommunicateHeaderV3 != NULL) {
+      CommunicateHeaderV3 = (EFI_MM_COMMUNICATE_HEADER_V3 *)CommunicateHeader;
+      BufferSize          = CommunicateHeaderV3->BufferSize;
+    } else {
+      BufferSize = CommunicateHeader->MessageLength +
+                   sizeof (CommunicateHeader->HeaderGuid) +
+                   sizeof (CommunicateHeader->MessageLength);
+    }
+
+    if (BufferSize > mNsCommBuffMemRegion.Length) {
+      // Something bad has happened, we should have landed in ARM_SMC_MM_RET_NO_MEMORY
+      Status = EFI_BAD_BUFFER_SIZE;
+      DEBUG ((
+        DEBUG_ERROR,
+        "%a Returned buffer exceeds communication buffer limit. Has: 0x%llx vs. max: 0x%llx!\n",
+        __func__,
+        BufferSize,
+        (UINTN)mNsCommBuffMemRegion.Length
+        ));
+    } else {
+      CopyMem (
+        CommBufferVirtual,
+        (VOID *)mNsCommBuffMemRegion.VirtualBase,
+        BufferSize
+        );
+    }
+  }
+
+  return Status;
+}
+
+/**
+  Communicates with a registered handler.
+
+  This function provides a service to send and receive messages from a registered UEFI service.
+
   @param[in] This                     The EFI_MM_COMMUNICATION_PROTOCOL instance.
   @param[in, out] CommBufferPhysical  Physical address of the MM communication buffer
   @param[in, out] CommBufferVirtual   Virtual address of the MM communication buffer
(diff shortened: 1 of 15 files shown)


kraxel:
The changes to VirtMmCommunication are not going to work.  qemu can not parse v3 messages, so doing a hard switch is not an option.  The v3 protocol implementation must send v2 messages to qemu.

Even if qemu gets support for v3 messages in the future the driver needs to check whenever qemu actually supports v3 messages.

Is there any documentation explaining the motivation for EFI_MM_COMMUNICATION3_PROTOCOL?  Commit 11cc0d639e1518619d0af56b8e2abd58024a2294 which introduces the struct has links to bugzilla.tianocore.org, but they are dead now ...

mdkinney:
> The changes to VirtMmCommunication are not going to work. qemu can not parse v3 messages, so doing a hard switch is not an option. The v3 protocol implementation must send v2 messages to qemu.
> 
> Even if qemu gets support for v3 messages in the future the driver needs to check whenever qemu actually supports v3 messages.
> 
> Is there any documentation explaining the motivation for EFI_MM_COMMUNICATION3_PROTOCOL? Commit [11cc0d6](https://github.com/tianocore/edk2/commit/11cc0d639e1518619d0af56b8e2abd58024a2294) which introduces the struct has links to bugzilla.tianocore.org, but they are dead now ...

The Bugzillas were migrated to GitHub Issues.
* https://github.com/tianocore/edk2/issues/10094
* https://github.com/tianocore/edk2/issues/10105

kuqin12:
> The changes to VirtMmCommunication are not going to work. qemu can not parse v3 messages, so doing a hard switch is not an option. The v3 protocol implementation must send v2 messages to qemu.
> 
> Even if qemu gets support for v3 messages in the future the driver needs to check whenever qemu actually supports v3 messages.
> 
> Is there any documentation explaining the motivation for EFI_MM_COMMUNICATION3_PROTOCOL? Commit [11cc0d6](https://github.com/tianocore/edk2/commit/11cc0d639e1518619d0af56b8e2abd58024a2294) which introduces the struct has links to bugzilla.tianocore.org, but they are dead now ...

So QEMU has its own MM core? Is there anything else the QEMU backed MM core would do other than UEFI var?

V3 eliminates the usage of UINT in the original header and used variable length struct. The former causes trouble when communication occurs between different modes (i.e. 32bit PEI and 64bit MM). The latter sprinkles the `OFFSET_OF` pattern all over the codebase and there are still some missed instances in the past years (i.e. the changes I made in MmCommuncation DXE driver in this PR).

The v3 header is larger than v1 header. So technically I can cache the content mangle/move the buffer header to make this a v1 header. But it seems wrong. May I know what the long-term plan is for QEMU to support this interface? Knowing it will help to form a solution to sustain both sides better. Thanks.

kraxel:
> So QEMU has its own MM core? Is there anything else the QEMU backed MM core would do other than UEFI var?

qemu can process EfiSmmVariableProtocol + VarCheckPolicyLibMmiHandler messages and implements UEFI variable store that way.  Which allows secure boot support for virtual machines without depending on SMM (x64) or el3 (aarch64) emulation.

> V3 eliminates the usage of UINT in the original header and used variable length struct. The former causes trouble when communication occurs between different modes (i.e. 32bit PEI and 64bit MM). The latter sprinkles the `OFFSET_OF` pattern all over the codebase and there are still some missed instances in the past years (i.e. the changes I made in MmCommuncation DXE driver in this PR).

Ok.  That makes sense.  qemu supports only the 64-bit version of the struct, i.e. there is no 32-bit guest support right now.

> The v3 header is larger than v1 header. So technically I can cache the content mangle/move the buffer header to make this a v1 header.

There is a buffer copy anyway.  And the v3 header must be translated into the 64bit version of the v1 header, which should simplify things because I think this comes down to simply dropping the first three fields (`HeaderGuid`, `BufferSize` and `Reserved`) of of the v3 header.

> But it seems wrong. May I know what the long-term plan is for QEMU to support this interface? Knowing it will help to form a solution to sustain both sides better. Thanks.

Well, I'm asking questions because I'm trying to make plans ...

I think essentially it comes down to two options:
 * simply continue with the 64-bit variant of the v1 struct, or
 * implement v3 struct support in addition to v1.

But even when qemu gets v3 support in the future the `VirtMmCommunicationDxe` driver must support both v1 and v3 because we will have qemu versions with and without v3 support.

kuqin12:
> > So QEMU has its own MM core? Is there anything else the QEMU backed MM core would do other than UEFI var?
> 
> qemu can process EfiSmmVariableProtocol + VarCheckPolicyLibMmiHandler messages and implements UEFI variable store that way. Which allows secure boot support for virtual machines without depending on SMM (x64) or el3 (aarch64) emulation.
> 
> > V3 eliminates the usage of UINT in the original header and used variable length struct. The former causes trouble when communication occurs between different modes (i.e. 32bit PEI and 64bit MM). The latter sprinkles the `OFFSET_OF` pattern all over the codebase and there are still some missed instances in the past years (i.e. the changes I made in MmCommuncation DXE driver in this PR).
> 
> Ok. That makes sense. qemu supports only the 64-bit version of the struct, i.e. there is no 32-bit guest support right now.
> 
> > The v3 header is larger than v1 header. So technically I can cache the content mangle/move the buffer header to make this a v1 header.
> 
> There is a buffer copy anyway. And the v3 header must be translated into the 64bit version of the v1 header, which should simplify things because I think this comes down to simply dropping the first three fields (`HeaderGuid`, `BufferSize` and `Reserved`) of of the v3 header.
> 
> > But it seems wrong. May I know what the long-term plan is for QEMU to support this interface? Knowing it will help to form a solution to sustain both sides better. Thanks.
> 
> Well, I'm asking questions because I'm trying to make plans ...
> 
> I think essentially it comes down to two options:
> 
> * simply continue with the 64-bit variant of the v1 struct, or
> * implement v3 struct support in addition to v1.
> 
> But even when qemu gets v3 support in the future the `VirtMmCommunicationDxe` driver must support both v1 and v3 because we will have qemu versions with and without v3 support.

Would it make sense to have QEMU take the variable payload (maybe also the size as well) directly? Intaking the MM communicate buffer wholesale seems like the traffic is intercepted one abstraction layer too early.

Alternatively, can we expand the "UEFI_VARS_CMD*" and add a check for command availability? Then we can send payload from UEFI with new commands if v3 support is indicated available.

kraxel:
> > But even when qemu gets v3 support in the future the `VirtMmCommunicationDxe` driver must support both v1 and v3 because we will have qemu versions with and without v3 support.
> 
> Would it make sense to have QEMU take the variable payload (maybe also the size as well) directly? Intaking the MM communicate buffer wholesale seems like the traffic is intercepted one abstraction layer too early.

I also want get the events (end-of-dxe, ready-to-boot, exit-boot-service) and variable policy updates.  So it's more than only the variable protocol, and I need the message guid to dispatch on the qemu side.

> Alternatively, can we expand the "UEFI_VARS_CMD*" and add a check for command availability?
> Then we can send payload from UEFI with new commands if v3 support is indicated available.

Sure, that part isn't much of a problem.  We have that FLAGS register returned from qemu, and we can allocate a bit there to signal v3 support to the firmware.

The IMHO more interesting question is whenever this actually improves things.  Given that the 64-bit version of the v1 protocol is used between VirtMmCommunicationDxe and qemu the problems caused by UINTN size not being fixed do not exist in the first place.  I think the v3 protocol implementation could essentially be this (untested):
```
EFI_STATUS
EFIAPI
VirtMmCommunication3Communicate (
  IN CONST EFI_MM_COMMUNICATION3_PROTOCOL  *This,
  IN OUT VOID                              *CommBufferPhysical,
  IN OUT VOID                              *CommBufferVirtual
  )
{
  EFI_MM_COMMUNICATE_HEADER_V3  *Header  = CommBufferVirtual;
  UINTN                         HdrSize  = OFFSET_OF (EFI_MM_COMMUNICATE_HEADER_V3, MessageGuid);
  UINTN                         CommSize = Header->BufferSize - HdrSize;
  EFI_STATUS                    Status;

  Status = VirtMmCommunication2Communicate (
             NULL,
             CommBufferPhysical + HdrSize,
             CommBufferVirtual + HdrSize,
             &CommSize
             );

  return Status;
}
```

leiflindholm:
> > But even when qemu gets v3 support in the future the `VirtMmCommunicationDxe` driver must support both v1 and v3 because we will have qemu versions with and without v3 support.
> 
> Would it make sense to have QEMU take the variable payload (maybe also the size as well) directly? Intaking the MM communicate buffer wholesale seems like the traffic is intercepted one abstraction layer too early.
> 
> Alternatively, can we expand the "UEFI_VARS_CMD*" and add a check for command availability? Then we can send payload from UEFI with new commands if v3 support is indicated available.

I don't think we should be looking at a QEMU solution here. Surely this change would affect any current users? Are we saying that there are no current users other than QEMU?

mdkinney:
> > > > qemu compatibility affects VirtMmCommunicationDxe (one patch in the PR) only.
> > > 
> > > 
> > > But is it necessary if we need to solve the migration path anyway for other users?
> > 
> > 
> > Well, there is no migration path. The patches do a hard switch from v2 to v3 protocol.
> 
> Yeah. I don't like it. That's not how standards are supposed to move.

From a producer perspective, the V3 could be incrementally added so both V2 and V3 are supported
From a consumer perspective, support for V3 could be incrementally added with priority, so V3 is preferred if it is available and falls back to V2.

At a later date, production of V2 and consu,ption of V2 could be removed if all downstream implementations have moved to V3.

kuqin12:
> > > > > qemu compatibility affects VirtMmCommunicationDxe (one patch in the PR) only.
> > > > 
> > > > 
> > > > But is it necessary if we need to solve the migration path anyway for other users?
> > > 
> > > 
> > > Well, there is no migration path. The patches do a hard switch from v2 to v3 protocol.
> > 
> > 
> > Yeah. I don't like it. That's not how standards are supposed to move.
> 
> From a producer perspective, the V3 could be incrementally added so both V2 and V3 are supported From a consumer perspective, support for V3 could be incrementally added with priority, so V3 is preferred if it is available and falls back to V2.
> 
> At a later date, production of V2 and consu,ption of V2 could be removed if all downstream implementations have moved to V3.

Okay, I will modify the variable module patches to fall back to the v2 and leave out the patch for OVMF for now. OVMF change can be added once QEMU is updated.

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
```diff
@@ -737,6 +737,14 @@ MmCommunicationMmCommunicate3 (
   }
 
   CommunicateHeader = (EFI_MM_COMMUNICATE_HEADER_V3 *)CommBufferVirtual;
+  if (CompareMem (
```
leiflindholm: Could use CompareGuid?

leiflindholm:
With this update, we should be able to drop the breaking-change label, right?

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
```diff
@@ -175,16 +175,18 @@ SendSpmMmCommunicate (
 **/
 EFI_STATUS
 EFIAPI
-MmCommunication2Communicate (
-  IN CONST EFI_MM_COMMUNICATION2_PROTOCOL  *This,
-  IN OUT VOID                              *CommBufferPhysical,
-  IN OUT VOID                              *CommBufferVirtual,
-  IN OUT UINTN                             *CommSize OPTIONAL
+MmCommunicationCommon (
```
samimujawar, line 179: I think we can make this function STATIC.

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c
```diff
@@ -493,11 +502,42 @@ ValidateMmCommBufferAddr (
...
+  } else {
+    BufferSize = ((EFI_MM_COMMUNICATE_HEADER *)CommBufferAddr)->MessageLength +
+                 OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, Data);
+    Status = SafeUint64Add (
+               CommBufferAddr,
+               BufferSize,
+               &BufferEnd
+               );
+  }
+
```
samimujawar, line 532: I think line 521 to 528 can be moved here and line 532 to 536 can be removed, right?

ArmPkg/Drivers/StandaloneMmCpu/EventHandle.c
```diff
@@ -188,19 +200,41 @@ PiMmCpuTpFwRootMmiHandler (
...
+      gGuidedEventContextV3,
+      gGuidedEventContextV3->BufferSize,
+      gGuidedEventContextV3->MessageSize
+      ));
+
+    Status = mMmst->MmiManage (
+                      &gGuidedEventContextV3->MessageGuid,
+                      NULL,
+                      gGuidedEventContextV3->MessageData,
+                      (UINTN *)(&gGuidedEventContextV3->MessageSize) // MU_CHANGE: BZ3398 Make MessageLength the same size in EFI_MM_COMMUNICATE_HEADER for both 32 and 64
```
samimujawar: I think we can drop the comment "// MU_CHANGE: BZ3398 Make MessageLength the same size in EFI_MM_COMMUNICATE_HEADER for both 32 and 64"

samimujawar: same for line 235

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
```diff
@@ -163,17 +165,33 @@ InitCommunicateBuffer (
...
-  }
+  if (mMmCommunication3 != NULL) {
+    if (DataSize + SMM_COMMUNICATE_HEADER_SIZE_V3 + SMM_VARIABLE_COMMUNICATE_HEADER_SIZE > mVariableBufferSize) {
+      return EFI_INVALID_PARAMETER;
+    }
 
-  SmmCommunicateHeader = (EFI_MM_COMMUNICATE_HEADER *)mVariableBuffer;
-  CopyGuid (&SmmCommunicateHeader->HeaderGuid, &gEfiSmmVariableProtocolGuid);
-  SmmCommunicateHeader->MessageLength = DataSize + SMM_VARIABLE_COMMUNICATE_HEADER_SIZE;
+    ZeroMem (mVariableBuffer, mVariableBufferSize);
```
samimujawar: Should line 176 be moved at line 170 ?

kuqin12: Agreed. This is updated.

ArmPkg/Drivers/StandaloneMmCpu/EventHandle.c
kuqin12: Yes. Sorry for creeping in odd comments. This is removed.

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c
kuqin12: You are correct. This is updated so that the safe int math is using the same routine.

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
kuqin12: Sure. This is marked as static.

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
kuqin12: Agreed. This is updated.

kuqin12:
> With this update, we should be able to drop the breaking-change label, right?

Yes. I removed the breaking-change label.

leiflindholm (approved):

leiflindholm:
> > With this update, we should be able to drop the breaking-change label, right?
> 
> Yes. I removed the breaking-change label.

Hmm, if you only removed the tick, that doesn't seem to do anything for already set labels. But I removed it manually.

kuqin12:
> > > With this update, we should be able to drop the breaking-change label, right?
> > 
> > 
> > Yes. I removed the breaking-change label.
> 
> Hmm, if you only removed the tick, that doesn't seem to do anything for already set labels. But I removed it manually.

The labels field seem to be read-only from my end. Probably because I do not have write access. Thanks for removing it!

kuqin12:
@lgao4 @niruiyu @jiaxinwu Could you please review the PR and let me know if you have any feedback? Thanks in advance!

MdeModulePkg/Include/Guid/SmmVariableCommon.h
```diff
@@ -9,6 +9,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #ifndef _SMM_VARIABLE_COMMON_H_
 #define _SMM_VARIABLE_COMMON_H_
 
+#include <Pi/PiMmCis.h>
```
lgao4: This header is not required to be included. It can be removed. Please try.

lgao4:
> @lgao4 @niruiyu @jiaxinwu Could you please review the PR and let me know if you have any feedback? Thanks in advance!

The change in MdeModulePkg is good to me.

MdeModulePkg/Include/Guid/SmmVariableCommon.h
kuqin12: Thanks, this is updated.

kuqin12:
@leiflindholm @lgao4 @samimujawar this PR is reviewed. May I have help in merging the change when you have a chance? Thanks in advance!

Javagedes:
@mergifyio refresh

Javagedes:
@mergifyio requeue

Javagedes:
@Mergifyio refresh

Javagedes:
The PR is not being added the queue because it is not meeting the requirement that all conversations be resolved. However I don't see any conversations to resolve.

image

Javagedes:
https://github.com/Mergifyio queue

kuqin12:
> The PR is not being added the queue because it is not meeting the requirement that all conversations be resolved. However I don't see any conversations to resolve.
> 
> image

@Javagedes @leiflindholm Thanks for the help. But I think this ^ status remains even after the rebase.. The only outstanding comment is likely this one below. Is "fixed" treated as "resolved" during evaluation?

image

makubacki:
@mergifyio queue

makubacki:
Looking at this PR from the GraphQL API, this thread is indeed showing as unresolved.

```json
      "pullRequest": {
        "reviewThreads": {
          "nodes": [
            {
              "isResolved": false,
              "comments": {
                "nodes": [
                  {
                    "body": "## Returned pointer not checked\n\nValue may be null; it should be checked before dereferencing.\n\n[Show more details](https://github.com/tianocore/edk2/security/code-scanning/1677)",
                    "author": {
                      "login": "github-advanced-security"
                    },
                    "createdAt": "2025-06-12T21:48:22Z"
                  }
                ]
              }
            },
```

makubacki:
I tried resolving it over the API and got:

```json
{
  "data": {
    "resolveReviewThread": null
  },
  "errors": [
    {
      "type": "UNPROCESSABLE",
      "path": [
        "resolveReviewThread"
      ],
      "locations": [
        {
          "line": 3,
          "column": 3
        }
      ],
      "message": "The thread is not a conversation and cannot be resolved"
    }
  ]
}
```

Looking further.

makubacki:
I also tried resolving the security alert from the REST API. That successfully dismissed the alert but something seems to not be connecting the "isResolved" value that mergify might be looking at.

makubacki:
@mergifyio refresh

makubacki:
@mdkinney, would you prefer @kuqin12 to create a new PR or we push this one directly?

leiflindholm:
> @mdkinney, would you prefer @kuqin12 to create a new PR or we push this one directly?

I'm not @mdkinney but my preference would be pushing it directly. The only reason I see for doing anything else would be if you wanted to keep trying to untangle whatever went wrong.

makubacki:
> I'm not @mdkinney but my preference would be pushing it directly. The only reason I see for doing anything else would be if you wanted to keep trying to untangle whatever went wrong.

I asked Mike since he set up and primarily manages the Mergify config. I agree since it has passed all GitHub checks, I think it makes sense to move forward with the PR being merged including all original reviewer approvals and comments as opposed to make a new one for Mergify.

mdkinney:
I am looking.  This is likely due to a CodeQL comment

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -86,30 +88,63 @@
...
+    // Program the MM header size
+    MmCommunicateHeader->MessageLength = CommunicateBufferSize - SMM_COMMUNICATE_HEADER_SIZE;
+
+    MmVarCommsHeader = (SMM_VARIABLE_COMMUNICATE_HEADER *)(CommunicateBuffer + SMM_COMMUNICATE_HEADER_SIZE);
+
+    // We are only supporting GetVariable and GetNextVariableName
+    MmVarCommsHeader->Function = Function;
+
+    // Send the MM request using MmCommunicationPei
+    Status = MmCommunicationPpi->Communicate (MmCommunicationPpi, CommunicateBuffer, &CommunicateBufferSize);
```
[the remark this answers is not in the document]
mdkinney: Was this issue resolved?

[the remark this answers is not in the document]
kuqin12: Yes. CodeQL cannot figure out that `MmCommunicationPpi` will be populated in the second `PeiServicesLocatePpi` call, which is why this comment is left in the first place. So I added an explicit check at line 127 to pacify the bot.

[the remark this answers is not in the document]
kuqin12: Oh, we only get to add a comment to be able to resolve the thread? This is good to know. Took long enough to figure out.

mdkinney:
There is one unresolved issues reported by a security-bot.  Was that resolved?

kuqin12:
> There is one unresolved issues reported by a security-bot. Was that resolved?

Yes, it was fixed by explicitly checking the pointer before using it. The bot comment is marked as fixed. We cannot resolve it until you commented on it :)

mdkinney:
> > There is one unresolved issues reported by a security-bot. Was that resolved?
> 
> Yes, it was fixed by explicitly checking the pointer before using it. The bot comment is marked as fixed. We cannot resolve it until you commented on it :)

Correct.  We have seen this issue before when a bot comments on a PR.  The status for resolved conversations is not cleared when the reported issue is resolved.  Requires a person to make an additional comment and then resolve the conversation for the status to be reflected correctly to Mergify.

This issue was reported to Mergify in Nov 2024.  Their evaluation determined it was an issue with the status returned by GitHub API and they opened a GitHub support ticket on the issue at that time.

PR #11082, opened 2025-05-15
UefiCpuPkg/MpInitLib: Fix split-lock violation from MP_CPU_EXCHANGE_INFO

ajyoung-oracle:
# Description

UefiCpuPkg/MpInitLib: Fix split-lock violation from MP_CPU_EXCHANGE_INFO

With latest x86_64 processors and Linux kernel which support split-lock detection, OVMF was implicated
in a split-lock violation.  It was determined that the NumApsExecuting field of the MP_CPU_EXCHANGE_INFO
struct in MpInitLib (which is used atomically by the AP Reset Vector assembly code) was the culprit resulting in
warning from the kernel, like so:

    [1035537.770691] x86/split lock detection: #AC: qemu-kvm/785650 took a split_lock trap at address: 0x7fe40050

which aligns with associated OVMF debug output:

    AP Vector: non-16-bit = 7FE40000/DF
    WakeupBufferStart = 9F000, WakeupBufferSize = 0
    AP Vector: 16-bit = 9F000/41, ExchangeInfo = 9F041/5C

Therefore, align the MP_CPU_EXCHANGE_INFO structure and move the NumApsExecuting field
to the beginning to ensure it resides within a single cache-line avoiding the split-lock. Do the same
for the ApIndex field as it is also used atomically and thus subject to a potential split-lock violation.

Results of the fix (with added OVMF debug to output the address of NumApsExecuting field):

Before fix:

    AP Vector: non-16-bit = 7FE40000/DF
    WakeupBufferStart = 9F000, WakeupBufferSize = 0
    AP Vector: 16-bit = 9F000/41, ExchangeInfo = 9F041/5C
    MpCpuExchangeInfo->NumApsExecuting Addr = 9F07D

A 4 byte field at address 9F07D crosses a cache-line boundary resulting in a lock-split violation. Also, notice the ExchangeInfo struct is unaligned at addr 9F041
    
With fix:

    AP Vector: non-16-bit = 7FE40000/E5
    WakeupBufferStart = 9F000, WakeupBufferSize = 0
    AP Vector: 16-bit = 9F000/41, ExchangeInfo = 9F048/5C
    MpCpuExchangeInfo->NumApsExecuting Addr = 9F054

Aligned address 9F054 does not cross a cache-line avoiding the lock-split violation. And, notice the ExchangeInfo struct is now aligned at addr 9F048.

NOTE: The OVMF split-lock violations result in a non-trivial performance degradation (due to bus locking). Here is some
profiling data with and without this fix:

    VCPUS:  AP Finish Time (microseconds):          # kernel split-lock detections:
    (max)   (without fix)   (with fix)              (without fix)                   (with fix)
    2       22557           30                      1                               0
    4       44738           46                      3                               0
    8       76974           60                      6                               0
    16      124207          170                     10                              0
    32      128673          241                     ~15 (callbacks suppressed)      0
    64      233840          424                     ~30 (callbacks suppressed)      0
    128     417998          1161                    ~58 (callbacks suppressed)      0
    254     820083          7395                    ~114 (callbacks suppressed)     0

Cc: Ray Ni <ray.ni@intel.com>
Cc: Jiaxin Wu <jiaxin.wu@intel.com>
Cc: Zhiguang Liu <zhiguang.liu@intel.com>
Cc: Dun Tan <dun.tan@intel.com>
Cc: Rahul Kumar <rahul1.kumar@intel.com>
Cc: Gerd Hoffmann <kraxel@redhat.com>
Cc: Star Zeng <star.zeng@intel.com>

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

OVMF Ia32X64 and X64 builds were tested on latest Intel X9 hardware (which supports split-lock detection). With the fix applied, the split-lock violations no longer occur.

## Integration Instructions

N/A


UefiCpuPkg/Library/MpInitLib/MpLib.c
```diff
@@ -960,7 +960,7 @@ GetApResetVectorSize (
   )
 {
   if (SizeBelow1Mb != NULL) {
-    *SizeBelow1Mb = AddressMap->ModeTransitionOffset + sizeof (MP_CPU_EXCHANGE_INFO);
+    *SizeBelow1Mb = ALIGN_VALUE (AddressMap->ModeTransitionOffset, 8) + sizeof (MP_CPU_EXCHANGE_INFO);
```
niruiyu, line 963: Can you use ALIGN_VARIABLE? or replace 8 with sizeof (UINTN)?

niruiyu (changes requested):
The code change looks generally good to me.
I added one minor comment.

But the commit message is too short when comparing the description put in the PR. Can you please include more details that you put in the PR description in the commit message? At least please highlight two things:
1. the MP_CPU_EXCHANGE_INFO structure is not allocated in an aligned address.
2. NumApsExecuting/ApIndex are type of UINTN. GdtrProfile/IdtrProfile are 6 (32bit) or 10(64bit) bytes which may cause NumApsExecuting/ApIndex across cacheline boundary (assuming cache line size is 64byte). That's the reason we need to put the two fields before GdtrProfile/IdtrProfile.

ajyoung-oracle:
> The code change looks generally good to me. I added one minor comment.
> 
> But the commit message is too short when comparing the description put in the PR. Can you please include more details that you put in the PR description in the commit message? At least please highlight two things:
> 
>     1. the MP_CPU_EXCHANGE_INFO structure is not allocated in an aligned address.
> 
>     2. NumApsExecuting/ApIndex are type of UINTN. GdtrProfile/IdtrProfile are 6 (32bit) or 10(64bit) bytes which may cause NumApsExecuting/ApIndex across cacheline boundary (assuming cache line size is 64byte). That's the reason we need to put the two fields before GdtrProfile/IdtrProfile.

Thanks Ray for the review. 

I updated the code to align the MP_CPU_EXCHANGE_INFO struct on a UINTN boundary (instead of explicit 8 bytes).

I also made the commit  message more verbose as you suggested above.

Thanks again.

niruiyu (approved):

PR #11086, opened 2025-05-16
MdeModulePkg: Leak Memory if Not RW on FreePages

### LINKED ISSUE #11074: [Bug]: default OVMF build overwrites memory on FreePages() causing page fault if pages are not writable

os-d:
# Description

Currently, if the DebugClearMemory bit is set in the PcdDebugPropertyMask, CoreConvertPagesEx will attempt to write a pattern to the pages being freed. However, it does not check that the page is writeable, which will cause a page fault if not.
Furthermore, if NX protections are not enabled, the core does not ensure that any freed pages are RW, which is the state expected when they are allocated next. If they are not RW, the allocating driver will crash trying to use them.

This patch updates the page freeing code to query the memory attributes protocol, if present, for the attributes. If this call fails or the attributes are not RW at a minimum, the core leaks the memory (returning success to the caller). If the memory attribute protocol is not present (either because a platform doesn't produce it or it is before the protocol has been produced, the core
continues with freeing memory. This is either before the CPU Arch protocol is available (so drivers can't change memory attributes) or otherwise matches existing behavior. This was deemed the best approach to let memory that can't be guaranteed to be RW leak instead of letting a driver crash when allocating it. It was deemed less brittle to simply leak the memory instead of attempting to change the attributes.

Closes #11074.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested by the original reproer using Linux shim v16 and attempting to FreePages that had attributes changed and confirmed that this path now works. Also tested on OVMF booting to shell.

## Integration Instructions

N/A.


bluca (approved):
Tested again, fixes the issue with shim 16, thanks!

os-d:
@lgao4 and @mdkinney please consider merging this for edk2-stable202505. This addresses a crash for the Linux shim v16 on debug builds. When DebugClearMemory is set, DXE Core will currently attempt to write to all freed pages, even if they are not writeable. This patch ensures that DXE Core will change the memory attributes of the free pages before writing to them, or will not write to them if it fails to.

Without this, users of shim v16 will experience failures when using builds of edk2 that have DebugClearMemory set when loading certain Linux distributions.

mdkinney:
@ardbiesheuvel @leiflindholm @makubacki @ajfish Please review for edk2-stable202505

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -1648,11 +1650,55 @@ CoreInternalFreePages (
...
+
+  // The free pages code will attempt to zero the freed pages if DebugClearMemory is enabled. However, there is no
+  // guarantee that the memory is writeable, which can cause the core to crash. So, we need to check if the memory is
+  // writeable before we attempt to zero it. We must do this outside of the memory lock to avoid deadlocks going
+  // through the GCD, which may cause page allocations to occur. It is safe to call the GCD from here, because the
+  // GCD only uses pool backed memory and CoreFreePoolPages() does not call through this path, so we won't hit a
+  // deadlock on the GCD lock. We don't attempt to call this before gCpu is initialized, because before that point,
+  // page attributes can't be set anyway and we may be very early in GCD init and not have the GCD populated yet.
+  if (DebugClearMemoryEnabled () && (gCpu != NULL)) {
+    Status = CoreGetMemorySpaceDescriptor (
```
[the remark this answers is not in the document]
os-d: I have pushed a fix for these two edge cases. @bluca can you please re-test and confirm this still works for your use case? I do not anticipate any change, it did work for regular OVMF for me.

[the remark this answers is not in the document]
bluca: Can confirm this still works with shim 16, thanks

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -1706,14 +1720,84 @@ CoreInternalFreePages (
 
   IsGuarded = IsPageTypeToGuard (Entry->Type, AllocateAnyPages) &&
               IsMemoryGuarded (Memory);
+
```
mdkinney, line 1723: I recommend the logic added here be moved to an internal helper function to compute MemoryWrittable.

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -1706,14 +1720,84 @@ CoreInternalFreePages (
...
+        break;
+      }
+
+      // we may have started in the middle of a descriptor, so we need to move to the beginning of the next descriptor,
+      // or the end of the range, whichever is smaller
+      CurrentAddress += CurrentLength;
+    }
+
+    // if any step failed, we will skip the debug clear memory
+    if (EFI_ERROR (Status)) {
```
mdkinney, line 1776: Isn't this a more critical failure.  If the pages are not writable, but they are freed to memory type EfiConventionalMemory, then the range can be allocated in the future using AllocatePages().  DEBUG_CLEAR_MEM() will be attempted again on allocation and the allocated pages may need to be writable based on memory type of the later allocation.

MdeModulePkg/Core/Dxe/Mem/Page.c
os-d: Sure, I'll move that.

MdeModulePkg/Core/Dxe/Mem/Page.c
os-d: Well, actually I think the problem is even worse than that. In our platforms, we always have the NX protection policy set, so I was relying on ApplyMemoryProtectionPolicy in CoreFreePages to mark the free pages back to RW (setting EFI_MEMORY_XP): https://github.com/tianocore/edk2/blob/2e5b357a1f80e7b05560f4cf9b4edbfca9de5ab4/MdeModulePkg/Core/Dxe/Mem/Page.c#L1757.

However, if a platform does not have the NX protection policy set, these pages could just be allocated again as RO, regardless of DebugClearMemory (yes you are right DebugClearMemory would crash, but if it was disabled then the consumer would crash). A driver would expect that any pages allocated via AllocatePages should be RW and would crash writing to them.

One approach is that this logic could not be gated behind if DebugClearMemoryEnabled() is true. If this logic is done on every CoreFreePages call, we ensure that no matter what a caller has done with pages of memory, it is safe to use for future allocations. In practice, anything changing attributes tends to set them back before freeing, which is a good pattern to follow.

Another approach is to use the memory attribute protocol, which is actually what the Mu version did, but it existed there for a while before edk2 had it for x86. The advantage of using the memory attribute protocol is that it has a GetMemoryAttributes function, so you can query the range for each FreePages call, if, as it will be on most deallocations, the access attributes are RW, do nothing else. If the attributes are not RW, then the memory attribute protocol (or GCD API or CPU arch protocol) can be used to set the attributes to RW. I didn't go with the memory attribute protocol because its intention was not to use from the core and I'm not sure I like that paradigm. But, it does provide the Get API that is useful. It also bypasses the GCD, which means the GCD remains out of sync, which is not good, but is something that happens because of the architecture of the memory management subsystem.

In either of these cases, this error is more critical and can assert and return a failure (i.e. not safe to reallocate again in the future if we couldn't change the attributes).

However, this greater problem is something that has existed in edk2 for a long time and seemingly any callers who changed page attributes changed them back, a la https://github.com/tianocore/edk2/blob/2e5b357a1f80e7b05560f4cf9b4edbfca9de5ab4/MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceIo.c#L1056-L1069.

I can go with either approach, I'm not sure the memory attribute protocol would really be any more performant, it would still be a page table walk. Or we can force the NX protection policy :). Because this is trying to go in for the stable tag, I can do a version here and come back and implement a larger approach, e.g. this just fixes the known bug (but leaves open the bug for systems without the NX protection policy set) and closes that for the stable tag, then we go for the more complete bug. Let me know your thoughts.

mdkinney: How about this.  If a request to FreePages() is unable put some of the pages back into RW state for future EfiConventionalMemory allocations, then only free (and clear memory) for the pages that are in the RW state. This means that callers that do not restore the attributes will at worse generate a memory leak when pages are freed.  If callers perform proper attribute management, then no memory leak.  This also removes the need for the core to make any policy decisions.

os-d: I think that's a reasonable approach. So my thought then would be to do this logic unconditionally (i.e. just drop the if DebugClearMemoryEnabled() check) and then as you said, if we failed to set the attributes on a given range, don't free that range (but try to continue to the next range if it exists). I can then also drop the piping of MemoryWriteable throughout, as that is no longer needed, the internal free pages code can assume it has RW memory.

Let me rework this.

MdeModulePkg/Core/Dxe/Mem/Page.c
os-d: Resolving as this is no longer relevant with the current code.

MdeModulePkg/Core/Dxe/Mem/Page.c
os-d: @mdkinney I implemented this, please take a close look (I am asking my team to review also to ensure I didn't make a mistake). I had to move some of the logic around in order to handle freeing just part of the pages and leaking the rest. Again, this could move to the memory attribute protocol if desired (query the range, if attributes are RW, skip of all this logic) or it can be done unconditionally like this implementation.

This is also now a larger impact change, so I'll defer to you whether this should be targeted for the stable tag or not. I have tested on OVMF booting to shell and once we decide on a path forward, I'll ask the original reproer to confirm this still works.

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -1765,6 +1772,157 @@ CoreFreePages (
...
+    // we may also be in the middle of a descriptor, so ensure our length is not larger than the descriptor length
+    if (EndAddress > DescEnd) {
+      CurrentLength = DescEnd - CurrentAddress;
+    } else {
+      CurrentLength = EndAddress - CurrentAddress;
+    }
+
+    // we always set the attributes here, don't check the descriptor attributes, because they may be out of sync
+    // if the attributes were changed with the CPU arch protocol or the memory attribute protocol
+    Status = CoreSetMemorySpaceAttributes (
```
apop5, line 1879: At the end of this, this memory will be left as RWX, right?

os-d: It depends on what is set there, but RWX is likely. However, inside of CoreInternalFreePages now, the NX policy is applied, so it will get NX applied to it if the platform has that configured. If not, then all memory will be RWX (except whatever drivers change). So at the end of the free call, attributes will be set as the platform has configured.

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -1765,6 +1772,157 @@ CoreFreePages (
...
+  CurrentAddress = Memory;
+  EndAddress     = Memory + EFI_PAGES_TO_SIZE (NumberOfPagesAligned);
+  while (CurrentAddress < EndAddress) {
+    Status = CoreGetMemorySpaceDescriptor (
+               CurrentAddress,
+               &Descriptor
+               );
+
+    if (EFI_ERROR (Status)) {
+      DEBUG ((
```
mdkinney, line 1852: Is this really an error condition and an ASSERT() condition?  Since it will continue in release builds, should this be a WARN message with no ASSERT() so the log will should some memory pages were not actually freed?

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -1765,6 +1772,157 @@ CoreFreePages (
...
+    // we always set the attributes here, don't check the descriptor attributes, because they may be out of sync
+    // if the attributes were changed with the CPU arch protocol or the memory attribute protocol
+    Status = CoreSetMemorySpaceAttributes (
+               CurrentAddress,
+               CurrentLength,
+               (Descriptor.Attributes & (~(EFI_MEMORY_RO | EFI_MEMORY_RP)))
+               );
+
+    if (EFI_ERROR (Status)) {
+      DEBUG ((
```
mdkinney, line 1886: Same here.  Is this really an error condition and an ASSERT() condition?  Since it will continue in release builds, should this be a WARN message with no ASSERT() so the log will should some memory pages were not actually freed?

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -1765,6 +1772,157 @@ CoreFreePages (
...
+      ASSERT_EFI_ERROR (Status);
+
+      // we can't free this range, so we skip it and move to the next descriptor
+      CurrentAddress += CurrentLength;
+      continue;
+    }
+
+    Status = CoreInternalFreePages (CurrentAddress, EFI_SIZE_TO_PAGES ((UINTN)CurrentLength), MemoryType, IsGuarded);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((
```
mdkinney, line 1903: And here.  Is this really an error condition and an ASSERT() condition?  Since it will continue in release builds, should this be a WARN message with no ASSERT() so the log will should some memory pages were not actually freed?

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -1765,6 +1772,157 @@ CoreFreePages (
...
+  EFI_STATUS                       OperationStatus;
+
+  NumberOfPagesAligned = NumberOfPages;
+  OperationStatus      = EFI_SUCCESS;
+
+  if (NumberOfPages == 0) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  CoreAcquireMemoryLock ();
```
mdkinney, line 1812: Should the acquire/release lock just be moved into CoreFindEntryToFree() and this API be free of locks.

os-d: I’m fine to move the locks. The only reason I didn’t put them in CoreFindEntryToFree() was to allow future use of that API from locked functions, but since there is no usage today and it is easy to change, I’ll move it.

MdeModulePkg/Core/Dxe/Mem/Page.c
os-d: I’m fine to change this, my rationale is I do consider it an error if a caller passed in a memory address to free that didn’t exist in the GCD, that means the caller either messed up the pointer it had to point to some place in memory that doesn’t exist or the core incorrectly gave out some pointer that doesn’t exist, in either case the existence of pointers to non-existent memory is an error case to me that I would want to catch with an assert. In release builds, you would see the error message and know some pages were not freed (same with if it were a warn).

But, if you prefer warn and no assert, I can switch to that.

MdeModulePkg/Core/Dxe/Mem/Page.c
os-d: This one I had similar logic to making it an error and assert, in that if we are leaking memory, that is an error to me and something I would want to catch with an assert (especially because failing to set attributes may mean something larger with the core is happening that is wrong). But again, in release builds having an error vs a warning doesn’t make a difference (unless the platform is not printing one of the error levels), the log would show leaked pages.

If you prefer to change to warn and no assert, I’ll do that.

MdeModulePkg/Core/Dxe/Mem/Page.c
os-d: This one probably doesn’t need to assert, because there are many asserts within the internal free logic, so either we would hit one of those or we can bubble the error up, i.e. I don’t think we should assert here because the failure of this function is the next level down, not at this level. I would still keep the message at error level, again leaking memory I consider an error (particularly if it is runtime memory, since that memory leak will persist into the OS).

But, if you want to change the other logs to warn, I’ll change this one, too.

In general, I tend to prefer to not assert in the core code and let a caller assert if some operation fails so it is easier to debug, but in my experience, a lot of drivers don’t check the FreePages status (particularly shim, where this bug came from, doesn’t check a lot of EFI status codes), so without the core asserting, the log messages would likely get lost (if you booted, why check the log).

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -1765,6 +1772,157 @@ CoreFreePages (
...
+      CurrentAddress += CurrentLength;
+      continue;
+    }
+
+    // we may have started in the middle of a descriptor, so we need to move to the beginning of the next descriptor,
+    // or the end of the range, whichever is smaller
+    CurrentAddress += CurrentLength;
+  }
+
+  return OperationStatus;
```
mdkinney, line 1923: We need to agree on the return status from this API

Here is what I propose, and it is why I am asking above if some of the conditions are actually error/assert conditions or not.

* If pages requested to free were not allocated using AllocatePages(), then return an error
* If pages requested to free were allocated with AllocatePages(), and all of the pages can be freed, then return success
* If pages requested to free were allocated with AllocatePages(), but one or more or even all of the pages can not be freed due to memory protection conflicts, then still return success.

os-d: I agree I was being too loose with the return status. I agree with your proposal, with a slight tweak, let me know what you think:
- If pages requested to free were not allocated using AllocatePages(), then return an error. => The GetMemorySpaceDescriptor() case failing is this case to me, the caller has passed in an address that does not exist in the GCD, so therefore cannot have been allocated with AllocatePages(), so return an error. In this case we should consider whether we attempt to free from that address forward, as I do currently, or just return and say the caller has passed in something invalid, don't attempt to free at all, because we may be freeing something we shouldn't.
- If pages requested to free were allocated with AllocatePages(), and all of the pages can be freed, then return success
- If pages requested to free were allocated with AllocatePages(), but one or more or even all of the pages can not be freed due to memory protection conflicts, then still return success => But assert in this case so that the memory leak can be caught in debug builds. We don't expect this to occur, but does indicate something is wrong in the core so we should catch this.

mdkinney: > * If pages requested to free were allocated with AllocatePages(), but one or more or even all of the pages can not be freed due to memory protection conflicts, then still return success => But assert in this case so that the memory leak can be caught in debug builds. We don't expect this to occur, but does indicate something is wrong in the core so we should catch this.

If ASSERT() is generated, then production builds that do not remove ASSERT()s and attempt to boot the shim discussed here will halt.  I think this should be a DEBUG_WARN message and continue without an ASSERT().  It is a tradeoff between finding platform FW bugs that are not maintaining memory properties correctly and providing compatibility with known OS loader behaviors.

os-d: Are there production builds that don't disable asserts? I was under the impression that any productions builds would have asserts disabled. There are a number of other asserts in the free pages path, e.g. https://github.com/tianocore/edk2/blob/2e5b357a1f80e7b05560f4cf9b4edbfca9de5ab4/MdeModulePkg/Core/Dxe/Mem/Page.c#L853-L857 (and in fact shim on a debug build of OVMF was tripping the NumberOfPages assert as well, once this was resolved because they were trying to free a null pointer with 0 pages, but I encouraged them to fix that).

os-d: Another approach would be to wrap the asserts in `DEBUG_CODE()` blocks if we are concerned about production platforms having asserts enabled. I do think it is very valuable to have the asserts for core development because we have seen many issues with the GCD and so going through the GCD routines is the right thing to do, but possibly will expose other issues.

os-d: I'm fine to move these to DEBUG_WARN

mdkinney: I am aware of platform validation tests that boot many different OS types with DEBUG builds with ASSERT() enabled.  If a shim boot loader with this behavior discussed here is used in that scenario, then it would halt.  Since the exposed behavior is a memory leak with no other side effects, it does not prevent the system from continuing to function.  There are many sources of memory leaks that are not caught with DEBUG_WARM or ASSERT() conditions today.  Seems odd to add ASSERT() for a memory leak now.  Review of DEBUG_WARN messages would be required to observe these conditions in platform FW.

I think ASSERT() conditions should be used for when continued execution of code past the ASSERT() will cause exceptions or unexpected/undefined behavior.

os-d: Well, it would only halt if the core logic was incorrect. Shim with this logic boots with asserts enabled with this change (confirmed by original reproer offline). I don't see the asserts as necessarily to catch the memory leak (though I think useful to find those cases), I see it more to catch the case that the GCD/memory protections feature has entered a bad state, which could lead to who knows what errors down the line (and we have seen many such errors arise because of the GCD being in a bad state).

There are already existing GCD asserts for losing sanity, e.g. https://github.com/tianocore/edk2/blob/2e5b357a1f80e7b05560f4cf9b4edbfca9de5ab4/MdeModulePkg/Core/Dxe/Gcd/Gcd.c#L785 and memory protections asserts for very similar operations, e.g. https://github.com/tianocore/edk2/blob/2e5b357a1f80e7b05560f4cf9b4edbfca9de5ab4/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c#L1023. So I don't really see this as a new paradigm. In fact, I did not trace this through, it is entirely possible the GCD code would assert in these cases and never hit the top level assert.

My main concern with changing this logic is that the GCD gets out of sync somewhere else (which happens a lot, unfortunately) and causes odd behavior down the line.
 
However, if the asserts are still a point of contention, I will drop them. Just let me know.

Regardless, I will move the error messages to warnings, change the failure path from GetMemorySpaceDescriptor() to an return an error (same as these pages were not allocated with AllocatePages), and in the other cases return success.

mdkinney: I only did not want to see ASSERT() for the memory leak.  The reset make sense.

os-d: To clarify, you are asking that the failure path from the call to `CoreSetMemorySpaceAttributes()` be reduced to WARN and have no ASSERT? Or are you saying to leave that as is, because failures in that function indicate a bad state in the GCD?

mdkinney: I think that is right.  How about update the code based on the latest info from these discussions and get these conversations resolved and start the review fresh again.

os-d: @mdkinney I have made these changes and resolved the existing conversations, please re-review.

MdeModulePkg/Core/Dxe/Mem/Page.c
os-d: I dropped the assert here as there are already asserts in the CoreInternalFreePages code, but kept the debug error and returning any failing status from the internal freeing code (because if just that range was passed to FreePages, even without the memory attribute logic, it would have failed and returned an error).

MdeModulePkg/Core/Dxe/Mem/Page.c
os-d: I dropped this to a WARN, removed the assert, and did not capture any status from this case (i.e. caller would see EFI_SUCCESS if other parts succeed and this fails).

MdeModulePkg/Core/Dxe/Mem/Page.c
os-d: I moved the locks to CoreFindEntryToFree.

os-d:
@mdkinney @lersek @lgao4 I have updated this PR (and title/description) per offline discussion. Can you please review? This now simply queries the memory attributes and leaks the memory if it is not RW at a minimum. It does not attempt to set the attributes correctly.

@bluca can you please confirm that this resolves your repro? It boots to OVMF shell for me.

bluca (approved):
Works for me, thanks!

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -1741,6 +1741,36 @@ CoreFreePages (
...
+  if (gMemoryAttributeProtocol != NULL) {
+    Status = gMemoryAttributeProtocol->GetMemoryAttributes (
+                                         gMemoryAttributeProtocol,
+                                         Memory,
+                                         EFI_PAGES_TO_SIZE (NumberOfPages),
+                                         &Attributes
+                                         );
+
+    // if we failed to get the attributes, or if the memory is read-only or read-protected,
+    // then we leak the memory and return success. This is to avoid a known bootloader bug where it free memory that is
```
ardbiesheuvel: How is this a bootloader bug? Does the memory attribute protocol definition include any guidance as to whether a user of the protocol should restore a region's permission attributes before freeing it? This seems like the responsibility of the DXE core to me.

os-d: :). There has been a lot of discussion about this and going back and forth. FWIW, I agree that this is a responsibility of the DXE Core and that the PI/UEFI specs don't specify any state for memory at rest to be in or to be returned in these APIs. However, the real way to fix this would be to remove CpuDxe (at least the paging aspect) and have DXE Core control the page table, then we eliminate all the lack of synchronization between the GCD and the page table and the various ways to change it. That is obviously a major change and high risk, so not on the table at the moment.

Failing that, offline conversation and conversation on the associated issue (https://github.com/tianocore/edk2/issues/11074) have led to this compromise which is considered lower risk than the core trying to change attributes it doesn't know about (the GCD being out of sync).

My original PR did change the attributes here when they were incorrect, which again I agree should be the responsibility of the core, but I can agree that this has always been the behavior of the core and to reduce risk, we can simply leak the memory.

I can change the wording of the comment, the bug that was filed was DXE Core writing to memory it didn't know if it was writeable or not, this was because of a released Linux shim's actions, but agreed that it isn't specified what behavior it should take.

os-d:
@mdkinney @lgao4 , I need to change the wording in the comment, but please review the functional changes.

lgao4:
> @mdkinney @lgao4 , I need to change the wording in the comment, but please review the functional changes.

The change is good to me.

MdeModulePkg/Core/Dxe/Mem/Page.c
ardbiesheuvel: Yes, please update the comment. In general, referring to a 'known bootloader bug' without providing any details is highly likely to cause confusion in the future, and so it would be much better to add an annotation here that we failed to specify the expected behavior when we defined the protocol, and as a result, we cannot make any assumption whatsoever about the permission attributes of pages being handed back to the page allocator.

os-d: @ardbiesheuvel agreed. I have updated the comment, please re-review.

PR #11092, opened 2025-05-20
BaseTools: Add gcov_info to GCC linker script

garybeihl:
# Description

GCC is capable of adding code coverage sections that are filled in at runtime. This change collects those sections into the final binary so that the code coverage data are available for dumping and post-processing using gcov.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Test shell app compiled with gcc --coverage and linked with this GccBase.lds was run under QEMU.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Scripts/GccBase.lds
+++ b/BaseTools/Scripts/GccBase.lds
@@ -5,6 +5,7 @@
   Copyright (c) 2010 - 2021, Intel Corporation. All rights reserved.<BR>
   Copyright (c) 2015, Linaro Ltd. All rights reserved.<BR>
   (C) Copyright 2016 Hewlett Packard Enterprise Development LP<BR>
+  Copyright (C) 2025, Microsoft Corporation. All rights reserved.<BR>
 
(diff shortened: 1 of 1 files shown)


mdkinney:
Doesn't this increase the size of all FW components built with GCC.  Shouldn't this be something that can be opt-in choice on a per component basis?

garybeihl:
It only adds .gcov_info sections if they are present in the compiler output. It is opt-in per translation unit by using the gcc --coverage flag. Without using --coverage in CC_FLAGS, there should be no change in the final binaries.

garybeihl:
Could someone please review this PR? Please? Thank you!

mdkinney:
This PR is in a strange state.  It has 400+ commits and 1 file change. Please rebase against master.

garybeihl:
Closing the PR and starting over as I can't get this accepted by the PatchCheck tool. I will start over with a fresh branch.

PR #11093, opened 2025-05-21
ArmPkg: ArmPsciResetSystemLib: PSCI warm reset fix

kuqin12:
# Description

This resolves an issue in the ResetWarm() routine where the warm reset was being attempted twice instead of querying for PSCI SYSTEM_RESET2 support. In addition, it ensures that the monitor args are zero'd prior to monitor calls.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

The change was tested warm reset flow with this change, confirmed that SYSTEM_RESET2 works as expected.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Library/ArmPsciResetSystemLib/ArmPsciResetSystemLib.c
+++ b/ArmPkg/Library/ArmPsciResetSystemLib/ArmPsciResetSystemLib.c
@@ -71,13 +71,16 @@ ResetWarm (
 {
   ARM_MONITOR_ARGS  Args;
 
-  Args.Arg0 = ARM_SMC_ID_PSCI_SYSTEM_RESET2_AARCH64;
+  Args.Arg0 = ARM_SMC_ID_PSCI_FEATURES;
+  Args.Arg1 = ARM_SMC_ID_PSCI_SYSTEM_RESET2_AARCH64;
 
   // Is SYSTEM_RESET2 supported?
   ArmMonitorCall (&Args);
   if (Args.Arg0 == ARM_SMC_PSCI_RET_SUCCESS) {
     // Send PSCI SYSTEM_RESET2 command
     Args.Arg0 = ARM_SMC_ID_PSCI_SYSTEM_RESET2_AARCH64;
+    Args.Arg1 = 0; // Reset type
+    // cookie does not matter with reset type 0
 
     ArmMonitorCall (&Args);
   } else {


ArmPkg/Library/ArmPsciResetSystemLib/ArmPsciResetSystemLib.c
```diff
@@ -51,6 +52,7 @@ ResetCold (
 {
   ARM_MONITOR_ARGS  Args;
 
+  ZeroMem (&Args, sizeof (ARM_MONITOR_ARGS));
```
ardbiesheuvel, line 55: If this is important, please do it in a separate patch.

kuqin12: Okay, will update shortly.

ArmPkg/Library/ArmPsciResetSystemLib/ArmPsciResetSystemLib.c
```diff
@@ -71,12 +72,15 @@ ResetWarm (
 {
   ARM_MONITOR_ARGS  Args;
 
-  Args.Arg0 = ARM_SMC_ID_PSCI_SYSTEM_RESET2_AARCH64;
+  ZeroMem (&Args, sizeof (ARM_MONITOR_ARGS));
```
ardbiesheuvel, line 75: Same for this ZeroMem() and all the others. Please move them to a different patch, and explain why it is important that these structures are zeroed before use.

ArmPkg/Library/ArmPsciResetSystemLib/ArmPsciResetSystemLib.c
kuqin12: I ended up removing it. It was meant to prevent the garbage on the stack being penetrated to secure world. But given neither PSCI nor SMCCC clarifies this, the change does not seem as necessary.

ArmPkg/Library/ArmPsciResetSystemLib/ArmPsciResetSystemLib.c
kuqin12: This is for the simplicity of zeroing all the arguments. I do not want the garbage to get into TZ.

leiflindholm: Yes, so say that in the commit message for the separate patch. The zeroing is not part of the "implement querying" change.

kuqin12: Updated, along with zeroing calls used in other reset functions.

ArmPkg/Library/ArmPsciResetSystemLib/ArmPsciResetSystemLib.c
kuqin12: I added this to a separate commit.

leiflindholm (approved):

samimujawar (approved):

ardbiesheuvel (approved):

kuqin12:
@ardbiesheuvel @leiflindholm @samimujawar Is there any chance to carry this as part of the 202505 tag? This is breaking the warm reset due to the dirty stack variable content.

leiflindholm:
> @ardbiesheuvel @leiflindholm @samimujawar Is there any chance to carry this as part of the 202505 tag? This is breaking the warm reset due to the dirty stack variable content.

It's breaking the warm reset where? This code is unmodified since July 2024 (and that was a rejig). Does that justify pushing the stable tag out by two weeks?

If a platform is breaking because the unclean stack, that sounds like a platform port issue, but it also isn't what this PR was said to be about. The zeroing of registers was presented as cleanup while implementing the probing improvement - not as a bug that needed fixing.

kuqin12:
> > @ardbiesheuvel @leiflindholm @samimujawar Is there any chance to carry this as part of the 202505 tag? This is breaking the warm reset due to the dirty stack variable content.
> 
> It's breaking the warm reset where? This code is unmodified since July 2024 (and that was a rejig). Does that justify pushing the stable tag out by two weeks?
> 
> If a platform is breaking because the unclean stack, that sounds like a platform port issue, but it also isn't what this PR was said to be about. The zeroing of registers was presented as cleanup while implementing the probing improvement - not as a bug that needed fixing.

I think it comes down to whether the platform cares about the reset type or not. The code is functional on the surface because there is a cold reset as the fall back. My interpretation is that this is always broken on various platforms, just no one bothers checking the reset type until now.

kuqin12:
> > > @ardbiesheuvel @leiflindholm @samimujawar Is there any chance to carry this as part of the 202505 tag? This is breaking the warm reset due to the dirty stack variable content.
> > 
> > 
> > It's breaking the warm reset where? This code is unmodified since July 2024 (and that was a rejig). Does that justify pushing the stable tag out by two weeks?
> > If a platform is breaking because the unclean stack, that sounds like a platform port issue, but it also isn't what this PR was said to be about. The zeroing of registers was presented as cleanup while implementing the probing improvement - not as a bug that needed fixing.
> 
> I think it comes down to whether the platform cares about the reset type or not. The code is functional on the surface because there is a cold reset as the fall back. My interpretation is that this is always broken on various platforms, just no one bothers checking the reset type until now.

But, it is probably not worthy to wait for another 2 weeks..

ardbiesheuvel:
OK, so there are three separate issues being addressed here:
- the first SMC call was supposed to check the availability of SYSTEM_RESET2, but invoked it directly instead
- both calls to SYSTEM_RESET2 were passing a bogus reset type
- SMC parameters that are unused are being loaded with garbage values from the uninitialized ARM_MONITOR_ARGS struct

The first two are clear bugs, which need to be fixed. Beyond that, I don't see the need to do anything, including setting `cookie` to `0x0`, which is documented as a dont-care if the reset type is `0x0`

kuqin12:
> OK, so there are three separate issues being addressed here:
> 
> * the first SMC call was supposed to check the availability of SYSTEM_RESET2, but invoked it directly instead
> * both calls to SYSTEM_RESET2 were passing a bogus reset type
> * SMC parameters that are unused are being loaded with garbage values from the uninitialized ARM_MONITOR_ARGS struct
> 
> The first two are clear bugs, which need to be fixed. Beyond that, I don't see the need to do anything, including setting `cookie` to `0x0`, which is documented as a dont-care if the reset type is `0x0`

Okay, reworked the patch and updated the commit message to reflect the fixes. The second commit was dropped for simplicity.

PR #11095, opened 2025-05-21
Maintainers.txt: restrict Arm architectures wildcard entries

leiflindholm:
Back in the day, we added wildcard maintainer entries for */Arm/ and */AArch64/, to make sure Arm maintainers got cc:d on patches What this was really meant for was core packages - MdeModulePkg and MdePkg - which were the only ones containing Arm-specific code outside of Arm*Pkg at the time.

However, this is no longer true, so we're increasingly finding ourselves cc:d on PRs where we don't necessarily have any opinion from an arch-maintainer perspective.

So convert the block to an include-list applying to MdeModulePkg, MdePkg, and UefiCpuPkg.

### DIFF
--- a/Maintainers.txt
+++ b/Maintainers.txt
@@ -84,8 +84,12 @@ M: Liming Gao <gaoliming@byosoft.com.cn> [lgao4]
 EDK II Architectures:
 ---------------------
 ARM, AARCH64
-F: */AArch64/
-F: */Arm/
+F: MdeModulePkg/*/AArch64/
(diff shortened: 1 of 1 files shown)


ardbiesheuvel (approved):

makubacki (approved):

leiflindholm:
Lol, patchcheck is failing on part of the commit message starting a line with "cc:d", but it's just part of the description. I can't be bothered to fix the script, so I'll just update the message.

samimujawar (approved):

PR #11096, opened 2025-05-22
NetworkPkg/HttpBootDxe: Update HTTP Boot to wait for IPv6 duplicate address detection to finish

samtsai0121:
# Description

Align HTTP Boot behavior with PXE by inserting a delay to wait for IPv6 Duplicate Address Detection (DAD) to complete before issuing DHCPv6 requests. This avoids EFI_NO_MAPPING errors caused by early DHCP attempts before a valid IPv6 address is ready.

Problem:
On some platforms, HTTP boot over IPv6 fails with EFI_NO_MAPPING during initial DHCPv6 attempts. The failure is due to the system trying to send Solicit messages before IPv6 DAD finishes, resulting in no usable IP address at that time.

Solution:
Insert a retry mechanism to poll DAD completion when the initial call to Dhcp6->Start() fails with EFI_NO_MAPPING. This behavior mirrors PXE's handling, where it waits for a valid IPv6 address to be assigned before retrying the DHCPv6 flow.

Reference: PXE DAD wait logic implemented in "PxeBcDhcp6.c", see commit "[ed2bfec](https://github.com/tianocore/edk2/commit/ed2bfecbcd6eed998be628b310bf89705fdb586c)".

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- Set up an HTTP IPv6 boot environment.
- Confirmed `EFI_NO_MAPPING` no longer occurs after inserting the delay.
- Verified successful HTTP boot.

## Integration Instructions

N/A

### DIFF
--- a/NetworkPkg/HttpBootDxe/HttpBootDhcp6.c
+++ b/NetworkPkg/HttpBootDxe/HttpBootDhcp6.c
@@ -931,6 +931,82 @@ ON_EXIT:
 }
 
 /**
+  Retry DHCPv6 when EFI_NO_MAPPING is returned, by waiting for Duplicate Address
+  Detection (DAD) to complete and restarting the DHCPv6 process.
+
+  This function stops the current DHCP6 session and waits for DAD to complete
+  based on the value returned by Ip6Config->GetData(), then retries Dhcp6->Start().
+
+  @param[in]  Dhcp6     The pointer to the EFI_DHCP6_PROTOCOL instance.
+  @param[in]  Ip6Cfg    The pointer to the EFI_IP6_CONFIG_PROTOCOL instance.
+
+  @retval EFI_SUCCESS           DHCPv6 process restarted successfully after DAD.
+  @retval EFI_DEVICE_ERROR      Failed during DAD timer wait or DHCP restart.
+  @retval EFI_OUT_OF_RESOURCES  Failed to allocate timer resources.
+  @retval Others                Underlying protocol errors.
+
+**/
+EFI_STATUS
+HandleDhcp6NoMappingRetry (
+  IN EFI_DHCP6_PROTOCOL       *Dhcp6,
+  IN EFI_IP6_CONFIG_PROTOCOL  *Ip6Cfg
+  )
+{
+  EFI_STATUS                                Status;
+  EFI_STATUS                                TimerStatus;
+  EFI_EVENT                                 Timer = NULL;
+  UINT64                                    GetMappingTimeOut;
+  UINTN                                     DataSize;
+  EFI_IP6_CONFIG_DUP_ADDR_DETECT_TRANSMITS  DadXmits;
+
+  //
+  // IP6 Linklocal address is not available for use, so stop current Dhcp process
+  // and wait for duplicate address detection to finish.
+  //
+  Dhcp6->Stop (Dhcp6);
+
+  //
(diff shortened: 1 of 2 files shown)


NetworkPkg/HttpBootDxe/HttpBootDhcp6.c
```diff
@@ -1013,6 +1022,52 @@ HttpBootDhcp6Sarr (
...
+      return Status;
+    }
+
+    Status = gBS->CreateEvent (EVT_TIMER, TPL_CALLBACK, NULL, NULL, &Timer);
+    if (EFI_ERROR (Status)) {
+      Dhcp6->Configure (Dhcp6, NULL);
+      return Status;
+    }
+
+    GetMappingTimeOut = TICKS_PER_SECOND * DadXmits.DupAddrDetectTransmits + HTTP_BOOT_DAD_ADDITIONAL_DELAY;
```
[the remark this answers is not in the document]
samtsai0121: @SaloniKasbekar 
Thanks for bringing this up!

The 3-second HTTP_BOOT_DAD_ADDITIONAL_DELAY follows the same approach as PXE, which uses PXEBC_DAD_ADDITIONAL_DELAY in UefiPxeBcDxe/PxeBcImpl.h to provide a similar delay buffer.

In PXE, this delay is added after retrieving DupAddrDetectTransmits from Ip6Config->GetData() to ensure that Duplicate Address Detection (DAD) has sufficient time to complete before DHCPv6 starts, preventing premature EFI_NO_MAPPING errors.

A similar constant, DHCP6_DAD_ADDITIONAL_DELAY, is also used in Dhcp6Dxe/Dhcp6Utility.h for internal DHCPv6 timeout calculations. These modules each define their own delay locally because they implement and control the DHCPv6 flow independently.

From analyzing the DAD implementation (Ip6NdFasterTimerTicking), we found:
- A random initial delay (0–1 second) occurs before DAD starts.
- Neighbor Solicitation (NS) packets are retransmitted every 1 second (via RetransTimer).
- The timer event that triggers DAD state progression runs every 100ms.
- This means even with one DAD transmit, actual completion time can range from 1.1 to 2+ seconds.

PXE’s use of a 3-second buffer provides a conservative safety margin. Since HTTP Boot follows a similar independent DHCPv6 process, we applied the same principle for consistency and reliability.

NetworkPkg/HttpBootDxe/HttpBootDhcp6.c
```diff
@@ -1013,6 +1023,52 @@ HttpBootDhcp6Sarr (
   // Start DHCPv6 S.A.R.R. process to acquire IPv6 address.
   //
   Status = Dhcp6->Start (Dhcp6);
+  if (Status == EFI_NO_MAPPING) {
+    //
```
leiflindholm, line 1027: NetworkPkg is especially plagued by way-too-long functions.
Can we pre-empt this expansion of one by turning this block into a static helper function?

samtsai0121: @leiflindholm 
I'll refactor this block into a static helper to keep the main function from growing too long.

samtsai0121:
Hi @mdkinney, this PR has been reviewed and approved by both @leiflindholm and @SaloniKasbekar. If there are no further concerns, could you please help merge it at your convenience? Thank you!

Issue #10089, opened 2021-05-10
CryptoPkg: Resource leaks in RsaSetKey() (Bugzilla Bug 3386)


tianocore-issues:
This issue was created automatically with [bugzilla2github](https://github.com/makubacki/bugzilla2github/tree/bz2gh_tianocore)
# Bugzilla Bug 3386
Date: 2021-05-10T11:59:41+00:00
From: Allen_Wynn
To:   unassigned &lt;<unassigned>&gt;
CC:   @lgao4, jian.j.wang

Last updated: 2021-05-11T21:42:38+00:00

tianocore-issues:
## Comment 15979
Date: 2021-05-10 11:59:41 +0000
From: Allen_Wynn

- Industry Specification: ---
- Release Observed: EDK II Master
- Releases to Fix: EDK II Master
- Target OS: ---
- Bugzilla Assignee(s): unassigned &lt;<unassigned>&gt;

In file CryptoPkg\Library\BaseCryptLib\Pk\CryptRsaBasic.c procedure RsaSetKey() allocates memory several times by calling BN_new() and BN_dup() but does not free the memory, leading to resource leaks.

Within each of the &quot;case&quot; blocks, before returning FALSE the code needs to ensure that any memory allocated by BN_new() gets deallocated by calling BN_free() or BN_clear_free()

Additionally, before returning TRUE the procedure should check to see if any memory was allocated by BN_new() and deallcoate by calling BN_free() or BN_clear_free().

tianocore-issues:
## Comment 16012
Date: 2021-05-11 21:42:38 +0000
From: @lgao4

Include CrytpoPkg maintainer to review this issue.

PR #11105, opened 2025-05-23
CryptoPkg: Fix memory leak in RsaSetKey

baranee:
# Description

Memory allocated by BN_new calls isn't deallocated within RsaSetKey.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

RSA signature verification still works after this change and confirmed thru debug memory allocated by BN_new are de-allocated.

## Integration Instructions

N/A

### DIFF
--- a/CryptoPkg/Library/BaseCryptLib/Pk/CryptRsaBasic.c
+++ b/CryptoPkg/Library/BaseCryptLib/Pk/CryptRsaBasic.c
@@ -8,6 +8,7 @@
   4) RsaPkcs1Verify
 
 Copyright (c) 2009 - 2020, Intel Corporation. All rights reserved.<BR>
+(c) Copyright 2025 HP Development Company, L.P.
 SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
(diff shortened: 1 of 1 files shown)


baranee:
Attempt to fix the issue https://github.com/tianocore/edk2/issues/10089

CryptoPkg/Library/BaseCryptLib/Pk/CryptRsaBasic.c
```diff
@@ -112,6 +114,9 @@ RsaSetKey (
   BnDq   = NULL;
   BnQInv = NULL;
 
+  AllocatedBn[0] = NULL;
```
liyi77, line 117: Why add new param here? I think
if (BnX != NULL) {
  BN_free (BnX);
}
is good enough.

liyi77: > BnX is not always allocated since it can be retrieved from `RSA_get0_key` \ `RSA_get0_factors` \ `RSA_get0_crt_params`. Have to deallocate only if it gets initialized thru `BN_new` interface.

Agree, approved.

liyi77 (approved):

PR #11115, opened 2025-05-23
[QemuLoadImageLib] Prefix initrd=initrd in cmdline

miguelfrde:
# Description

The current implementation causes issues when the command line contains arguments to be passed to `/init` after `"--"`. Instead we can prepend that at the beginning of the cmdline.

From: https://www.kernel.org/doc/html/v6.12/admin-guide/kernel-parameters.html

> Everything after “--” is passed as an argument to init.

Problem:

```
[    0.000000] Kernel command line: panic=0 init=/init console=ttyS0 earlyprintk=ttyS0 -- echo hello | nc -l -p 8080 initrd=initrd
...
[init] running command  echo hello | nc -l -p 8080 initrd=initrd
nc: exec initrd=initrd: No such file or directory
```

With this fix:

```
[    0.000000] Kernel command line: initrd=initrd panic=0 init=/init  -- echo hello | nc -l -p 8080
...
[init] running command  echo hello | nc -l -p 8080

```

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verified this fixed my problem in internal test using qemu.

## Integration Instructions

N/A


ardbiesheuvel:
The patch looks fine (but please fix the subject)

However, I wonder why we still need to pass `initrd=initrd` to begin with, given that even x86 has supported the initrd LoadFile2 protocol implementation since Linux v5.7.

@kraxel ?

PR #11116, opened 2025-05-23
Prevent memcpy intrinsics in VS22 (17.14.2)

makubacki:
# Description

The latest VS2022 update replaces some code patterns with struct
assignments with `memcpy`. This change convert the code to
explicitly use `CopyMem`.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- Build before and after with 17.14.2 VS2022 toolchain

## Integration Instructions

- Use this commit if you encounter `memcpy` linking issues in these files.

### DIFF
--- a/ShellPkg/Library/UefiShellDebug1CommandsLib/HexEdit/BufferImage.c
+++ b/ShellPkg/Library/UefiShellDebug1CommandsLib/HexEdit/BufferImage.c
@@ -142,9 +142,8 @@ HBufferImageBackup (
   VOID
   )
 {
-  HBufferImageBackupVar.MousePosition = HBufferImage.MousePosition;
-
-  HBufferImageBackupVar.BufferPosition = HBufferImage.BufferPosition;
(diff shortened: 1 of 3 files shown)


apop5 (approved):

ShellPkg/Library/UefiShellDebug1CommandsLib/HexEdit/BufferImage.c
```diff
@@ -2069,16 +2064,12 @@ HBufferImageAddCharacterToBuffer (
   //
   // get a place to add
   //
-  for (Index = (INTN)(Size + Count - 1); Index >= (INTN)Pos; Index--) {
```
pierregondois, line 2072: Maybe a dumb question, but shouldn't this be instead:
```
CopyMem (&BufferPtr[Pos], &BufferPtr[Pos - Count], Size - Pos + Count);
```

makubacki: Sorry it took me a while to get back. Yes, I think you're right and I updated it.

PR #11123, opened 2025-05-27
ArmPlatformPkg/Sec: Deal with entry at EL2 with VHE enabled

ardbiesheuvel:
The architecture now permits HCR_EL2.E2H to be RES1, and so even though the UEFI spec is silent on the matter, entering at EL2 with VHE enabled is a condition that needs to be dealt with. In particular, virtual machines running under nested virtualization on Apple M2 or newer will enter in this manner

In this case, the CPTR_EL2 system register needs to be treated as if its layout is identical to CPACR_EL1. Not doing so may result in all kinds of surprises, but the most noticeable is an early crash on the use of FP/SIMD registers, which get disabled inadvertently by this code.

### DIFF
--- a/ArmPlatformPkg/Sec/AArch64/Helper.S
+++ b/ArmPlatformPkg/Sec/AArch64/Helper.S
@@ -8,6 +8,8 @@
 #include <AsmMacroLib.h>
 #include <AArch64/AArch64.h>
 
+.set    HCR_EL2_E2H,           0x1 << 34
+
 // Setup EL1 while in EL1
 ASM_FUNC(SetupExceptionLevel1)
    mov  x5, x30                   // Save LR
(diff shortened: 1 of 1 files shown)


ardbiesheuvel:
@jrtc27

jrtc27:
ArmVirtPkg/Library/ArmPlatformLibQemu/AArch64/ArmPlatformHelper.S also defines this. Can we centralise the define in somewhere like MdePkg/Include/AArch64/AArch64.h (where CPACR_EL2 defines are)?

ardbiesheuvel:
> ArmVirtPkg/Library/ArmPlatformLibQemu/AArch64/ArmPlatformHelper.S also defines this. Can we centralise the define in somewhere like MdePkg/Include/AArch64/AArch64.h (where CPACR_EL1/CPTR_EL2 defines are)?

Yeah good point. I'll follow up with some cleanup on this file.

PR #11127, opened 2025-05-28
BaseTools: fix cross-compilation of antlr & dlg

Mic92:
antlr and dlg need to run on the build system, so we need to allow users to provide a way to specify a compiler

# Description

- [ ] ~~Breaking change?~~
- [ ] ~~Impacts security?~~
- [ ] ~~Includes tests?~~

## How This Was Tested

By cross-compilling

## Integration Instructions

### DIFF
--- a/BaseTools/Source/C/VfrCompile/GNUmakefile
+++ b/BaseTools/Source/C/VfrCompile/GNUmakefile
@@ -32,6 +32,9 @@ LINKER = $(CXX)
 
 EXTRA_CLEAN_OBJECTS = EfiVfrParser.cpp EfiVfrParser.h VfrParser.dlg VfrTokens.h VfrLexer.cpp VfrLexer.h VfrSyntax.cpp tokens.h
 
+CC_FOR_BUILD ?= $(CC)
+CXX_FOR_BUILD ?= $(CXX)
+
 MAKEROOT ?= ../..
 
 include $(MAKEROOT)/Makefiles/header.makefile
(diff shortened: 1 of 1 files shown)


Mic92:
cc @bexcran
cc @mdkinney

Mic92:
cc @avnik

Mic92:
> @Mic92 I tend to inject `CC = $(CC_FOR_BUILD)` etc into BaseTools/Source/C/Makefiles/header.makefile (looks like all BaseTools is a host tools)

Feel free to propose a different PR.

bexcran:
Is there a reason I'm not understanding why you can't use CC and CXX but need an additional variable?

nate-desimone:
Agreed with @bexcran, this doesn't seem nessesary.

Mic92:
> We don't need that. A few years ago I updated the build system so it should be possible to just use the standard variables CC and CXX. e.g.:
> 
> ```
> CC=aarch64-linux-gnu-gcc CXX=aaarch64-linux-gnu-g++ make -C BaseTools
> ```

I guess this works if we rely on make not trying to rebuild those binaries. Not the nicest user experience if you have a configured cross environment but this works as a work-around for nixpkgs.

avnik:
@bexcran 
> We don't need that. A few years ago I updated the build system so it should be possible to just use the standard variables CC and CXX. e.g.:
> 
> ```
> CC=aarch64-linux-gnu-gcc CXX=aaarch64-linux-gnu-g++ make -C BaseTools
> ```

With `CC=aarch64-linux-gnu-gcc CXX=aaarch64-linux-gnu-g++ make -C BaseTools` it would build antlr and dlg for aarch64, where they need to be build for host platform (these tools using for compiling other parts BaseTools)

bexcran:
> With `CC=aarch64-linux-gnu-gcc CXX=aaarch64-linux-gnu-g++ make -C BaseTools` it would build antlr and dlg for aarch64, where they need to be build for host platform (these tools using for compiling other parts BaseTools)

You're assuming the host platform isn't AArch64. :)

PR #11129, opened 2025-05-28
CLANG 20.1.0 build bug fixes

mdkinney:
# Description

Fix issues detected with builds using CLANG 20.1.0.
- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Perform package builds using CLANGPDB and CLANGDWARF in Windows and Linux using CLANG 20.1.0.

## Integration Instructions

N/A


MdeModulePkg/Bus/Spi/SpiNorFlashJedecSfdp/SpiNorFlash.c
```diff
@@ -63,6 +63,7 @@ FillWriteBuffer (
   // Copy Opcode into Write Buffer
   Instance->SpiTransactionWriteBuffer[0] = Opcode;
   Index                                  = 1;
+  AddressSize                            = 0;
```
leiflindholm, line 66: Does this need a corresponding check if AddressSize == 0 further down?

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
```diff
@@ -408,6 +408,7 @@ CreatMmHobList (
   //
   // Get platform HOBs
   //
+  PlatformHobList = NULL;
   PlatformHobSize = 0;
   Status          = CreateMmPlatformHob (NULL, &PlatformHobSize);
   if (Status == RETURN_BUFFER_TOO_SMALL) {
```
leiflindholm, line 414: Could we address this bit of success handling instead?
We could return if any other than this condition happens and drop indentation of the rest of the block. Then we already have a NULL check on PlatformHobList.

MdeModulePkg/Bus/Spi/SpiNorFlashJedecSfdp/SpiNorFlash.c
mdkinney: Unless I am missing something all other paths will either assign a non-zero value to AddressSize or it will ASSERT().  If ASSERT()s are disabled, then AddressSize will be 0 and it will call CopyMem() with a length of 0 (which is not an error condition) and add 0 to Index and complete the function.  I do not know if there are other side effects.  If there are, then all the paths that ASSERT() should be evaluated to determine if they should return early.

EmbeddedPkg/Library/FdtLib/fdt_strtoul.c
```diff
@@ -30,8 +30,8 @@ strtoul (
   }
 
   if (RETURN_ERROR (Status)) {
-    return MAX_UINTN;
+    return (unsigned long)MAX_UINTN;
```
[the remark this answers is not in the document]
mdkinney: I am happy to see that change go in first.

RedfishPkg/Library/JsonLib/JsonLib.inf
```diff
@@ -84,5 +84,5 @@
   #
   MSFT:*_*_X64_CC_FLAGS = /wd4204 /wd4244 /wd4090 /wd4334 /wd4706 /DHAVE_CONFIG_H=1 /U_WIN32 /UWIN64 /U_MSC_VER
   MSFT:*_*_IA32_CC_FLAGS = /wd4204 /wd4244 /wd4090 /wd4706 /DHAVE_CONFIG_H=1 /U_WIN32 /UWIN64 /U_MSC_VER
-  GCC:*_*_*_CC_FLAGS = -Wno-unused-function -Wno-unused-but-set-variable
+  GCC:*_*_*_CC_FLAGS = -Wno-unused-function -Wno-unused-but-set-variable -DHAVE_CONFIG_H=1 -UWIN32 -U_WIN32 -UWIN64 -U_MSC_VER
```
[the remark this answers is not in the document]
mdkinney: It is not setting.  It is unsetting something that would never be set in non-Windows envs.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
mdkinney: I agree that the logic should be changed here.  It is possible for EFI_SUCCESS to be returned if the required size of the HOB List is 0.  So have to check for EFI_SUCCESS and EFI_BUFFER_TOO_SMALL and for an invalid size of 0 for EFI_BUFFER_TO_SMALL.  Something like the following.  Not sure if I like the use of CpuDeadLoop().  Prefer ASSERT() configured as dead loop to detect this condition make sure it can never occur in release builds.

```
  //
  // Get platform HOBs
  //
  PlatformHobSize = 0;
  Status          = CreateMmPlatformHob (NULL, &PlatformHobSize);
  if (Status == EFI_SUCCESS) {
    DEBUG ((DEBUG_INFO, "%a: The size of platform MM HOBs to create is 0\n", __func__));
    return NULL;
  }
  if (Status != RETURN_BUFFER_TOO_SMALL) {
    DEBUG ((DEBUG_ERROR, "%a: Unable to determine size of platform MM HOBs to create\n", __func__));
    ASSERT (Status == RETURN_BUFFER_TOO_SMALL);
    return NULL;
  }
  if (PlatformHobSize == 0) {
    DEBUG ((DEBUG_ERROR, "%a: The size of platform MM HOBs to allocate is 0\n", __func__));
    ASSERT (PlatformHobSize != 0);
    return NULL;
  }

  //
  // Create platform HOBs for MM foundation to get MMIO HOB data.
  //
  PlatformHobList = AllocatePages (EFI_SIZE_TO_PAGES (PlatformHobSize));
  if (PlatformHobList == NULL) {
    DEBUG ((DEBUG_ERROR, "%a: Out of resource to create platform MM HOBs\n", __func__));
    ASSERT (PlatformHobList != NULL);
    CpuDeadLoop ();
  }

  BufferSize = PlatformHobSize;
  Status     = CreateMmPlatformHob (PlatformHobList, &PlatformHobSize);
  ASSERT_EFI_ERROR (Status);
  ASSERT (BufferSize == PlatformHobSize);
```

EmbeddedPkg/Library/FdtLib/fdt_strtoul.c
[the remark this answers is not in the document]
mdkinney: I will drop EmbeddedPkg FdtLib change from this patch series.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
```diff
@@ -410,25 +410,36 @@ CreatMmHobList (
...
+    return NULL;
+  }
+
+  if (Status != RETURN_BUFFER_TOO_SMALL) {
+    DEBUG ((DEBUG_ERROR, "%a: Unable to determine size of platform MM HOBs to create\n", __func__));
+    ASSERT (Status == RETURN_BUFFER_TOO_SMALL);
+    return NULL;
+  }
+
+  if (PlatformHobSize == 0) {
```
sweeaun, line 424: Is this checking needed? Shouldn't this already taken care by line 413, PlatformHobSize ==0 when status == EFI_SUCCESS?

sweeaun: Got it. Thanks @niruiyu

MdeModulePkg/Bus/Spi/SpiNorFlashJedecSfdp/SpiNorFlash.c
leiflindholm: I will confess I ask because I once had a look at this code before (when I was playing with building a platform with -Wpedantic) and it made me sad.

There's also the `BigEndianAddress >>= ((sizeof (UINT32) - AddressSize) * 8);` which may or may not be programmatically fine, but is an offset of zero ever correct? And if we're doing a CopyMem of 0 bytes to offset 0 from something, we're wasting both cycles and neurons.

Basically what I'm saying is I wish someone would rewrite this whole function, because I see many future "fix build" patches in its future if we don't.

mdkinney: Do you want to open a new GitHub issue for the owner to address these additional concerns?

EmbeddedPkg/Library/FdtLib/fdt_strtoul.c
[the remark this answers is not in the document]
leiflindholm: https://github.com/tianocore/edk2/pull/11139 raised

leiflindholm (approved):

os-d:
@mdkinney did you build ArmPkg and ArmPlatformPkg with clang? Last time we had looked at ARM64 clang support, there were some GCC ASM directives that clang didn't support, which would crash the build. This was awhile ago, though, so that might have been added to clang since then.

leiflindholm:
> @mdkinney did you build ArmPkg and ArmPlatformPkg with clang? Last time we had looked at ARM64 clang support, there were some GCC ASM directives that clang didn't support, which would crash the build. This was awhile ago, though, so that might have been added to clang since then.

Me and Ard generally default to clang these days. If you run into any specific failures, raise an issue.

os-d:
> Me and Ard generally default to clang these days. If you run into any specific failures, raise an issue.

That’s good to hear, thanks! We are looking at moving some of our partners to clang, so glad to see that build fixes are being made and CI is being looked at for it.

lgao4 (approved):

mdkinney:
@niruiyu Can you please complete the review of this PR for the packages you maintain?

mdkinney:
@AbnerChang Can you please review the RedfishPkg changes?

mdkinney:
@jyao1 Can you please review the SecurityPkg changes?

jyao1 (approved):

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
```diff
@@ -410,25 +410,36 @@ CreatMmHobList (
   //
   PlatformHobSize = 0;
   Status          = CreateMmPlatformHob (NULL, &PlatformHobSize);
-  if (Status == RETURN_BUFFER_TOO_SMALL) {
+  if (Status == EFI_SUCCESS) {
+    DEBUG ((DEBUG_INFO, "%a: The size of platform MM HOBs to create is 0\n", __func__));
+    return NULL;
```
niruiyu: This function creates a hob list that contains both foundation hobs and platform hobs.

Refer to the original code as below:
```c
  //
  // Final result includes: PHIT HOB, Platform HOBs, Foundation HOBs and an END node.
  //
  *HobSize = PhitHobSize + PlatformHobSize + FoundationHobSize + sizeof (EFI_HOB_GENERIC_HEADER);
```

So, if `CreateMmPlatformHob()` returns NULL hob list, this function still needs to create the foundation hob list.

mdkinney: Thanks for catching the logic change.  I have pushed an update that follows the original logic, but cleans up the error handling and resolves the clang error.  It also fixes a potential memory leak.

The clang error that is resolved is the following:
```
C:\work\GitHub\mdkinney\edk2\StandaloneMmPkg\Drivers\StandaloneMmIplPei\StandaloneMmIplPei.c:413:7: error: variable 'PlatformHobList' is used uninitialized whenever 'if' condition is false [-Werror,-Wsometimes-uninitialized]
  413 |   if (Status == RETURN_BUFFER_TOO_SMALL) {
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:\work\GitHub\mdkinney\edk2\StandaloneMmPkg\Drivers\StandaloneMmIplPei\StandaloneMmIplPei.c:448:25: note: uninitialized use occurs here
  448 |                         PlatformHobList,
      |                         ^~~~~~~~~~~~~~~
C:\work\GitHub\mdkinney\edk2\StandaloneMmPkg\Drivers\StandaloneMmIplPei\StandaloneMmIplPei.c:413:3: note: remove the 'if' if its condition is always true
  413 |   if (Status == RETURN_BUFFER_TOO_SMALL) {
      |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:\work\GitHub\mdkinney\edk2\StandaloneMmPkg\Drivers\StandaloneMmIplPei\StandaloneMmIplPei.c:400:46: note: initialize the variable 'PlatformHobList' to silence this warning
  400 |   VOID                       *PlatformHobList;
      |                                              ^
      |                                               = NULL
1 error generated.
```

niruiyu (approved):

PR #11130, opened 2025-05-28
MdeModulePkg: Refactor MM Services Tables linked in MM Core Perf Lib

makubacki:
# Description

The code used a `MmServicesTableLib` dependency for both the Standalone MM and Traditional MM instances and shared code between those. There is not a readily available `MmServicesTable` lib instance for Traditional SMM (that can link to `PiSmmCore`). To ease integration and prevent creating an instance just for this case, this change uses `SmmServicesTableLib` in the Traditional SMM instance and `MmServicesTableLib` in the Standalone MM instance and moves code as necessary to accommodate this.

For general reference, there are two available instances of
`MmServicesTableLib`:

- MdePkg/MmServicesTableLib - only supports DXE_SMM drivers.
- MdePkg/StandaloneMmServicesTableLib - supports both MM drivers and
  MM cores.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- Build and boot on a Traditional MM and Standalone MM X64 platform

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Library/SmmCorePerformanceLib/MmCorePerformanceLib.c
+++ b/MdeModulePkg/Library/SmmCorePerformanceLib/MmCorePerformanceLib.c
@@ -42,13 +42,6 @@ UINT32     mCachedLength   = 0;
 UINT32     mBootRecordSize = 0;
 BOOLEAN    mPerformanceMeasurementEnabled;
 
-//
-// Interfaces for SMM PerformanceMeasurement Protocol.
-//
-EDKII_PERFORMANCE_MEASUREMENT_PROTOCOL  mPerformanceMeasurementInterface = {
-  CreatePerformanceMeasurement,
-};
-
 /**
   Return the module name and optionally module GUID for a given handle.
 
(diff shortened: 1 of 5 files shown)


makubacki:
@jiaxinwu, @niruiyu, @lgao4, this PR has been up for nearly 2 weeks. Please review it.

benjamindoron (approved):
UefiPayload runs into this build error, I agree that we should get this in as soon as possible. I elected to solve it by adding SMM_CORE to MmServicesTableLib's list of supported library classes, and I was going to push that and my MmServicesTableLib instance for the standalone MM core until I saw this. But I agree that this approach is easier for integration.

makubacki:
> I agree with your approach to avoid adding a new MmServicesTableLib instance just for PiSmmCore.
> 
> One minor comment: Can you kindly add more description in the commit message? e.g.: Two available instances of MmServicesTableLib: * MdePkg/MmServicesTableLib only supports DXE_SMM driver * MdePkg/StandaloneMmServicesTableLib supports both MM driver and MM Core.

Thanks for the feedback. I've updated both the PR description and the commit message to mention that.

PR #11137, opened 2025-05-29
MdeModulePkg: Fix ValidateFmpCapsule in DxeCapsuleLibFmp

bexcran:
# Description

ValidateFmpCapsule was only checking for versions 1 and 2 of the EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER, while FmpCapsuleHeader.py in BaseTools/Source/Python/Common/Uefi/Capsule/ generates capsules with version 3.

Update ValidateFmpCapsule to also check for version 3, and fix the fields which are checked: According to UEFI Specification 2.11 (https://uefi.org/sites/default/files/resources/UEFI_Spec_Final_2.11.pdf) UpdateHardwareSupport was introduced in v2 and ImageCapsuleSupport in v3, so the version 1 and 2 offsets needed updated.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Ran `CapsuleApp capsule-file.cap` in the UEFI Shell and verified that it no longer fails verification.


LeviYeoReum (approved):

bexcran:
@lgao4 Could you review this please?

MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleLib.c
```diff
@@ -301,8 +301,10 @@ ValidateFmpCapsule (
     }
 
     if (ImageHeader->Version == 1) {
-      FmpImageHeaderSize = OFFSET_OF (EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER, UpdateHardwareInstance);
+      FmpImageHeaderSize = OFFSET_OF (EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER, UpdateVendorCodeSize);
```
lgao4: Original logic is correct. Version 1 EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER has UpdateVendorCodeSize field. So, its structure size is the offset of next field to structure header.

bexcran: Oh, of course. So the only change needed is the addition of version 3?

MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleLib.c
```diff
@@ -304,6 +304,8 @@ ValidateFmpCapsule (
       FmpImageHeaderSize = OFFSET_OF (EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER, UpdateHardwareInstance);
     } else if (ImageHeader->Version == 2) {
       FmpImageHeaderSize = OFFSET_OF (EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER, ImageCapsuleSupport);
+    } else if (ImageHeader->Version == 3) {
+      FmpImageHeaderSize = sizeof (EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER);
```
lgao4, line 308: this change is not required, because line 295 has set FmpImageHeaderSize = sizeof (EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER);

bexcran: Thanks, I see that now. The code is correct, but somewhere something is going wrong if IMAGE_HEADER_INIT_VERSION is set to 0x03. It ends up with a mismatch with the calculated size being 8 bytes less than the actual size.

PR #11138, opened 2025-05-30
Ray/mm core fix infinite loop

niruiyu:
# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions


StandaloneMmPkg/Core/FwVol.c
```diff
@@ -285,14 +275,25 @@ MmDispatchFvs (
...
     if (!FixedPcdGetBool (PcdShadowBfv)) {
       Fv = (EFI_FIRMWARE_VOLUME_HEADER *)((UINTN)FvHob.FirmwareVolume->BaseAddress);
     } else {
       Fv = AllocateCopyPool (FvHob.FirmwareVolume->Length, (VOID *)(UINTN)FvHob.FirmwareVolume->BaseAddress);
+      ASSERT (Fv != NULL);
       if (Fv == NULL) {
         DEBUG ((DEBUG_ERROR, "%a: Fail to allocate MM memory for Fv!\n", __func__));
         CpuDeadLoop ();
-        return;
+        continue;
```
samimujawar, line 296: Is there an implementation of CpuDeadLoop() that can break out? From what I see at https://github.com/tianocore/edk2/blob/master/MdePkg/Library/BaseLib/CpuDeadLoop.c#L14-L34 the loop is infinite.
Or this change is only applicable when a debugger intervenes and allows the code flow to continue?

StandaloneMmPkg/Core/FwVol.c
```diff
@@ -285,14 +275,25 @@ MmDispatchFvs (
...
     }
 
+    if (Index == ARRAY_SIZE (mMmFv)) {
+      DEBUG ((
+        DEBUG_INFO,
+        "%a: The number of FV Hobs exceeds the max supported FVs (%d) in StandaloneMmCore, skip it.\n",
+        __func__,
+        ARRAY_SIZE (mMmFv)
+        ));
+      continue;
```
samimujawar, line 285: Should we break from the for loop here?
Otherwise the log at line 264 will print the wrong Index for the FvHob.FirmwareVolume->BaseAddress & Length, and may lead to confusion.
Or Should printing of the Index be dropped from the log at line 264?

samimujawar (commented):
I have some questions about patch 3/3 otherwise the remaining changes look good to me.

StandaloneMmPkg/Core/FwVol.c
niruiyu: The CpuDeadLoop() is to halt the boot. "continue" is to help the debugging experience to support the continue running when debugger skips the CpuDeadLoop().

StandaloneMmPkg/Core/FwVol.c
niruiyu: I had similar thoughts when implementing the change because if a FV is skipped, the log would be:
  Fv[1] address = 0xxxx, size = 0x0
  Skip zero-length FV.
  Fv[1] address = 0xxxxx, size = 0xxxx
  Fv[2] address = 0xxxxx, size = 0xxxx
  The number of FV Hobs exceeds the max supported FVs (2) in StandaloneMmCore, skip it.

But if removing the Index from the log and only printing the Index for a valid FV, I need to add another line of debug message to show the FV index.

Comparing between, I think it's not too bad to have Index in the log even we haven't decided to drop the FV.

samimujawar (approved):

xuweiintel (approved):

PR #11143, opened 2025-05-30
DynamicTablesPkg: Support for generic SSDT devices

jbrasen:
# Description

Support for creating generic SSDT devices.
These are devices that are of the following format

Device(DEV0) {
  Name (_HID, "TEST1080")
  Name (_UID, 0)
  Name (_CCA, ZERO)

  Name(_CRS, ResourceTemplate() {
    Memory32Fixed(ReadWrite, Base1, Size1)
    Memory32Fixed(ReadWrite, Base2, Size2)
    Interrupt(ResourceConsumer, Level, ActiveHigh, Exclusive) { Int1 }
  })
}

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Generated entries with platform code.

## Integration Instructions

N/A

### DIFF
--- a/DynamicTablesPkg/Include/ArchCommonNameSpaceObjects.h
+++ b/DynamicTablesPkg/Include/ArchCommonNameSpaceObjects.h
@@ -64,6 +64,8 @@ typedef enum ArchCommonObjectID {
   EArchCommonObjGenericDbg2DeviceInfo,          ///< 36 - Generic DBG2 Device Info
   EArchCommonObjCxlHostBridgeInfo,              ///< 37 - CXL Host Bridge Info
   EArchCommonObjCxlFixedMemoryWindowInfo,       ///< 38 - CXL Fixed Memory Window Info
+  EArchCommonObjGenericDeviceInfo,              ///< 39 - Generic Device Info
+  EArchCommonObjGenericInterrupt,               ///< 40 - Generic Interrupt
   EArchCommonObjMax
 } EARCH_COMMON_OBJECT_ID;
 
@@ -206,6 +208,8 @@ typedef struct CmArchCommonPciAddressMapInfo {
 
 /** A structure that describes the
     Generic Interrupts.
+
+    ID: EArchCommonObjGenericInterrupt
 */
 typedef struct CmArchCommonGenericInterrupt {
   /// Interrupt number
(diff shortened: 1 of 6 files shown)


DynamicTablesPkg/Include/ArchCommonNameSpaceObjects.h
```diff
@@ -907,6 +911,42 @@ typedef struct CmArchCommonCxlFixedMemoryWindowInfo {
...
+
+  /// Unique Id
+  UINT32             Uid;
+
+  /// Hardware revision
+  BOOLEAN            HrvValid;
+  UINT32             Hrv;
+
+  /// Cache Coherency Attribute
+  BOOLEAN            Cca;
```
pierregondois, line 937: The ASL function `QWordMemory()` takes a `Cacheable` parameter which can be one of:
```
- 0 cacheable (Cacheable)
- 1 cacheable and write-combining (WriteCombining)
- 2 cacheable and prefetchable (Prefetchable)
- 3 uncacheable (NonCacheable)
```

Would it make more sense to define an enum for these values an have a corresponding field in `CM_ARCH_COMMON_MEMORY_RANGE_DESCRIPTOR` ?
If this works for you, maybe `CM_ARCH_COMMON_MEMORY_RANGE_DESCRIPTOR`  would also need a `CacheableValid` field as you did for `Hrv`. Indeed, the default value in `19.6.110 QWordMemory (QWord Memory Resource Descriptor Macro)` for the cacheable field is `uncacheable`, which matches the enum number 3, so the `CacheableValid` field would allow to remove any confusion

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtGenericDeviceLib/SsdtGenericDeviceGenerator.c
```diff
@@ -0,0 +1,519 @@
...
+        goto exit_handler;
+      }
+
+      for (Index = 0; Index < MemoryRangesCount; Index++) {
+        Status = AmlCodeGenRdQWordMemory (
+                   TRUE,
+                   TRUE,
+                   FALSE,
+                   FALSE,
+                   0,
```
pierregondois, line 242: Cf. the other comment about `cacheable`, I think this field should be dynamically populated if there is a need for the `_CCA` field to be populated.

PR #11144, opened 2025-05-31
PrePiLib missing header used in file

Flickdm:
# Description

PrePiLib.h is missing the header file that defines the structures used in the file.

Namely
 - EFI_PEI_FV_HANDLE
 - EFI_PEI_FILE_HANDLE
 
- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Without linking magic, using this file causes the linker to explode.

## Integration Instructions

N/A

### DIFF
--- a/EmbeddedPkg/Include/Library/PrePiLib.h
+++ b/EmbeddedPkg/Include/Library/PrePiLib.h
@@ -11,6 +11,7 @@
 #define __PRE_PI_LIB_H__
 
 #include <Guid/ExtractSection.h>
+#include <Pi/PiPeiCis.h>
 
 /**
   This service enables discovery of additional firmware volumes.


EmbeddedPkg/Include/Library/PrePiLib.h
```diff
@@ -11,6 +11,7 @@
 #define __PRE_PI_LIB_H__
 
 #include <Guid/ExtractSection.h>
+#include <Pi/PiPeiCis.h>
```
leiflindholm, line 14: The change is good, thanks!

The commit message could use a couple of tweaks:
1) The package should be called out in the shortlog (see https://github.com/tianocore/tianocore.github.io/wiki/Commit-Message-Format) - i.e., please prepend "EmbeddedPkg:"
2) Actually, I think there are more than those two ... maybe say "for example" instead of "namely"?

PR #11145, opened 2025-05-31
EmbeddedPkg: introduce GBL protocols

dimorinny:
[GBL (Generic Bootloader)](https://source.android.com/docs/core/architecture/bootloader/generic-bootloader) is Google's UEFI-based bootloader for Android. It runs as an EFI application on top of the platform firmware and delegates board- and platform-specific operations back to that firmware through a set of GBL-defined EFI protocols.

This PR introduces the header definitions (GUIDs, revisions and interface structures) for those protocols into EmbeddedPkg, so that EDK2-based firmware can implement them.

 Protocols added:

| Protocol | Purpose |
| --- | --- |
| `GblEfiAvb` | Android Verified Boot board-specific logic (device lock state, key validation, boot state). |
| `GblEfiAvf` | Vendor DICE handover and Secret Keeper public key for the Android Virtualization Framework. |
| `GblEfiBootControl` | A/B slot / boot target selection. |
| `GblEfiBootMemory` | Retrieving and synchronizing boot and partition buffers. |
| `GblEfiDebug` | Reporting fatal errors and debug events to firmware. |
| `GblEfiFastboot` | Platform-specific Android Fastboot operations. |
| `GblEfiFastbootTransport` | Firmware-provided transport interface for Android Fastboot. |
| `GblEfiOsConfiguration` | Kernel bootconfig fixups and device-tree component selection at boot time. |

GBL upstream [protocols signatures](https://cs.android.com/android/kernel/superproject/+/common-android-mainline:bootable/libbootloader/gbl/libefi_types/include/uefi/protocols/).

## How This Was Tested

Used by GBL on top of QCOM EDK2-based FW.

## Integration Instructions

Implementation [requirements](https://source.android.com/docs/core/architecture/bootloader/generic-bootloader/gbl-dev#implementation).

GBL protocols [specification](https://cs.android.com/android/kernel/superproject/+/common-android-mainline:bootable/libbootloader/gbl/docs/).


EmbeddedPkg/Include/Protocol/DtFixup.h
```diff
@@ -0,0 +1,50 @@
+/** @file
+
+  Copyright (c) 2025, The Android Open Source Project.
```
dimorinny, line 3: WIP: figuring out license requirements with author.

leiflindholm:
(Oops, wrong tab.)

dimorinny:
Hello @leiflindholm. Uploaded work-in-progress GBL protocols, along with a few open questions we currently have. We’d also appreciate any early feedback to help ensure the protocols are in good shape ahead of the first release planned for Oct 2025.

Thank you!

EmbeddedPkg/Include/Protocol/GblEfiAbSlot.h
```diff
@@ -0,0 +1,194 @@
+/** @file
+
+  Copyright (c) 2025, The Android Open Source Project.
+
+  SPDX-License-Identifier: Apache-2
```
dimorinny: @leiflindholm EDK2 CI checks are flagging `Apache-2`, mentioning only `BSD-2` is allowed. I’ve noticed some EDK2 code (e.g., MbedTLS) uses `Apache-2`, but I assume that might not be permitted for protocol headers.

Could you please confirm whether `Apache-2` is acceptable in this context, or should we switch to `BSD-2`?

leiflindholm: Short answer: BSD-2-Clause-Patent is the default and anything else requries justification. So if you could switch to that, my life (and possibly yours) would be made easier.

Slightly longer: the list of "acceptable licenses" really refers to "if we need to integrate external code into edk2 _and_ it's considered critical enough, here are licenses we could even consider". What would then follow would be lengthy discussions with everyone involved (who may have legal departments implementing hard requirements about anything other than the default license) having to agree that it's fine.

EmbeddedPkg/Include/Protocol/GblEfiAvb.h
```diff
@@ -0,0 +1,174 @@
...
+  CONST CHAR8    *VendorVersion;
+  CONST CHAR8    *VendorSecurityPatch;
+};
+
+/*
+  Extra partition name requested for verification.
+*/
+struct _GBL_EFI_AVB_PARTITION {
+  UINTN    NameLen;  // in/out
+  CHAR8    *Name;    // caller-allocated
```
dimorinny: @leiflindholm (CC: @dpursell)

[Here](https://cs.android.com/android/kernel/superproject/+/common-android-mainline:bootable/libbootloader/gbl/docs/gbl_efi_avb_protocol.md;l=108) we pass a buffer for the firmware to fill with the partition name of an extra partition to be verified by `libavb`.

We’ve noticed that standard UEFI protocols often follow the GPT partition name format — UTF-16 with a 72-byte (36-character) limit. However, `libavb` uses UTF-8 for partition names and enforces a different size limit (29 bytes/symbols). After some [discussion](https://android-review.googlesource.com/c/platform/bootable/libbootloader/+/3643383/comment/db5c47e7_79161c92/), we decided to use UTF-8 for this case to align with `libavb` on the protocol level.

Do you think this is acceptable, or would you recommend sticking with GPT-style UTF-16 everywhere?

dimorinny: @leiflindholm 

Another related question. As I see `CHAR8` EDK2 type is a simple `char` which can be both signed or unsignder depending on the underlying arch and tooling configuration. GBL's `char8_t` [is always](https://cs.android.com/android/kernel/superproject/+/common-android-mainline:bootable/libbootloader/gbl/libefi_types/defs/types.h;l=22) `unint8_t`, so always unsigned which should works well for ASCII, but could cause issues with signed char on UEFI side for UTF-8.

According to what I see `-funsigned-char` [is used](https://github.com/tianocore/edk2/blob/master/BaseTools/Conf/tools_def.template#L928) for the archs we care about. Could you please help us understand how the char compatibility issue is usually handled with UEFI. Is it safe to always use unsigned char on GBL side?

leiflindholm: For historical reasons, Unicode in UEFI is UCS-2, not UTF-8. (Sources are UTF-8, but in-memory representation is UCS-2.) So overlap of Ascii and Unicode isn't really a thing we have.

See commit 3f7f2871 for just how confusing (yet still predictable) this got.
Ultimately, the spec (https://uefi.org/specs/UEFI/2.11/02_Overview.html#common-uefi-data-types) states that CHAR is meant to hold ISO-Latin-1 encodings, which I choose to interpret as CHAR _needing_ to be unsigned.

leiflindholm: It's not unheard of for edk2 code interfacing with other standards to handle UTF-8.
However, I would ask that the code (via comments) is *very* explicit where CHAR* is specifically intended to point to UTF-8. Basically, grep MdePkg for UTF-8 and see how that's handled elsewhere.

The only real concern I have is that libavb has a smaller symbol limit than the GPT format. Is there any chance to grow it?

EmbeddedPkg/Include/Protocol/GblEfiAbSlot.h
```diff
@@ -0,0 +1,194 @@
+/** @file
```
leiflindholm: Comment on commit message:
Could we have a bit of a description please? And a link to documentation would be ideal.

EmbeddedPkg/Include/Protocol/GblEfiAvb.h
leiflindholm: @dimorinny appreciate the license update - could you have a look at this comment too?

EmbeddedPkg/Include/Protocol/GblEfiAbSlot.h
dimorinny: Updated the commit message to include a high-level description and a link to GBL's README. Also added links to protocol documentation in the headers. Does this address your request?

EmbeddedPkg/Include/Protocol/GblEfiAvb.h
dimorinny: Yes, very useful comment. I'm currently busy with other protocols-related stuff, going to get back to it next week.

dimorinny: Sorry for a long delay.

> It's not unheard of for edk2 code interfacing with other standards to handle UTF-8.
However, I would ask that the code (via comments) is very explicit where CHAR* is specifically intended to point to UTF-8. Basically, grep MdePkg for UTF-8 and see how that's handled elsewhere.

Just to make sure I understand correctly — isn't this conflict with the UEFI spec stating that `CHAR` (so I assume `CHAR8` as well) is meant to hold ISO-Latin-1 encodings, as you mentioned in the other thread?

> The only real concern I have is that libavb has a smaller symbol limit than the GPT format. Is there any chance to grow it?

Yes, we can technically grow the limit from 32 to 36 bytes on the `libavb` side since it's an artificial [limitation](https://cs.android.com/android/platform/superproject/main/+/main:external/avb/libavb/avb_cmdline.h;l=42), I will try to do so. However, we still can't allow the FW to use all 36 bytes to provide a partition name. The reason is `libavb` appends the slot suffix and null terminator (like `_a\0`), which will extend it with 3 extra symbols (ASCII characters, so strictly 1 byte per symbol). Which means, we can only pass a 33-bytes buffer at max to stay within GPT limits and avoid runtimes errors when the provided buffer is too big. My plan is to rename `Name` to `BaseName` and update the doc to make this situation clear.

Please let me know if this is not sufficient.

leiflindholm: > Sorry for a long delay.

No problem, thanks for getting back to it.

> > It's not unheard of for edk2 code interfacing with other standards to handle UTF-8.
> > However, I would ask that the code (via comments) is very explicit where CHAR* is specifically intended to point to UTF-8. Basically, grep MdePkg for UTF-8 and see how that's handled elsewhere.
> 
> Just to make sure I understand correctly — isn't this conflict with the UEFI spec stating that `CHAR` (so I assume `CHAR8` as well) is meant to hold ISO-Latin-1 encodings, as you mentioned in the other thread?

Yes. Which is why the request for heavy commenting where affected.

It also wouldn't be wrong to use UINT8*, but I'm not sure whether that would actually be less confusing?

leiflindholm: > > The only real concern I have is that libavb has a smaller symbol limit than the GPT format. Is there any chance to grow it?
> 
> Yes, we can technically grow the limit from 32 to 36 bytes on the `libavb` side since it's an artificial [limitation](https://cs.android.com/android/platform/superproject/main/+/main:external/avb/libavb/avb_cmdline.h;l=42), I will try to do so. However, we still can't allow the FW to use all 36 bytes to provide a partition name. The reason is `libavb` appends the slot suffix and null terminator (like `_a\0`), which will extend it with 3 extra symbols (ASCII characters, so strictly 1 byte per symbol). Which means, we can only pass a 33-bytes buffer at max to stay within GPT limits and avoid runtimes errors when the provided buffer is too big. My plan is to rename `Name` to `BaseName` and update the doc to make this situation clear.
> 
> Please let me know if this is not sufficient.

I think that's an entirely fine solution. Turning it into `BaseName` makes it pretty clear what's going on.

dimorinny: Thanks for the feedback, we've [done](https://android-review.googlesource.com/c/platform/bootable/libbootloader/+/3712526) so, will update the protocol signature a bit later along with other changes we're planning.

EmbeddedPkg/Include/Protocol/GblEfiAvf.h
```diff
@@ -0,0 +1,59 @@
...
+// {e7f1c4a6-0a52-4f61-bd98-9e60b559452a}
+//
+#define GBL_EFI_AVF_PROTOCOL_GUID \
+  { 0xe7f1c4a6, 0x0a52, 0x4f61, { 0xbd, 0x98, 0x9e, 0x60, 0xb5, 0x59, 0x45, 0x2a } }
+
+// Still in progress
+#define GBL_EFI_AVF_PROTOCOL_REVISION  0x00000000
+
+typedef struct _GBL_EFI_AVF_PROTOCOL GBL_EFI_AVF_PROTOCOL;
+
```
dimorinny, line 28: Hello @leiflindholm, we have another topic we want to hear your opinion about, which may be also useful for our future protocols design.

At this moment AVF loading (so this protocol as well) is optional for GBL and we use its availability (via EFI_BOOT_SERVICES.OpenProtocol) as an indecation whether we need to do AVF related stuff on GBL side. But recently we received a request to allow disabling AVF at the bootloader run-time via oem fastboot command handled by FW side through GBL's fastboot protocol and cosidering couple of ways to do so:

1. The option we're leaning towards is to request FW to uninstall AVF protocol via [EFI_BOOT_SERVICES.UninstallProtocolInterface()](https://uefi.org/specs/UEFI/2.10/07_Services_Boot_Services.html#efi-boot-services-uninstallprotocolinterface) in case FW decides AVF should be disabled.

2. An alternative approach is to extend this protocol, so GBL will explicitly ask FW whether AVF is supposed to be loaded:

```
enum AVF_CONFIGURATION_FLAG {
  // AVF has been disabled for this boot.
  AVF_CONFIGURATION_FLAG_DISABLED = (1 << 0);
};

typedef EFI_STATUS (EFIAPI *GBL_EFI_AVF_READ_CONFIGURATION)(
  IN GBL_EFI_AVF_PROTOCOL *This,
  OUT UINT32                            *Flags,
);
```

---

The first approach seems more UEFI ideomatic from our perspective, so we can utilize UEFI framework and optinal nature of this protocol to make the desceision. From GBL perspective we guarantee we never try to open this protocol before its enabled state is finalized (we open it only after fastboot), so it feels safe from [EFI_BOOT_SERVICES.UninstallProtocolInterface()](https://uefi.org/specs/UEFI/2.10/07_Services_Boot_Services.html#efi-boot-services-uninstallprotocolinterface) documentation perspective, but I'm afraid we may overlooking some potential problems we may hit in the future with that?

Summarizing all of that, do you think it's safe to rely on `UninstallProtocolInterface` to disable AVF or you prefer an explicit approach with dedicated method to check that state?

Thank you for your input!

leiflindholm: Hi @dimorinny. I agree 1 feels more idiomatic, and hence it would be my preference.
It may require some wording in your protocol specifications regarding how "If present, <someone> needs to ensure AVF protocol is ready for use after <event> handlers have completed."
Maybe look to UEFI spec and EFI_EVENT_GROUP_READY_TO_BOOT for inspiration? (that particular event also makes sense to me, but I may not have the full picture)

dimorinny: Thanks, very useful

leiflindholm (approved):
Hi @dimorinny. Back from holiday - it looks to me like all my comments have been addressed. Are you ready for this to be merged?

dimorinny:
We currently have a large number of changes still in progress, so I prefer to merge this closer to the release date, which is now planned for mid-November.

leiflindholm:
As per a conversation with @dimorinny in December, this is still in progress.

dimorinny:
Hey @leiflindholm, thanks for bumping the PR to prevent it from being closed.

As discussed, we have pre-frozen the GBL protocols with the release of the 0x00000100 revision. This means we are committed to maintaining backward compatibility for all subsequent pre-release revisions until the official 0x00010000 release (planned for around the end of May 2026), at which point final breaking changes may occur to finalize the specification. I have added a warning to each header to ensure this remains clear.

We have agreed with the corresponding QCOM team that merging this into `EmbeddedPkg` now would be beneficial for the integration process.

I would greatly appreciate any comments or notes on the proposed signatures. As mentioned, we cannot break the ABI at this stage, but we will definitely address such feedback once 1.0 is released. For all other matters, I will be happy to address your comments now.

ardbiesheuvel:
> Hey @leiflindholm, thanks for bumping the PR to prevent it from being closed.
> 
> As discussed, we have pre-frozen the GBL protocols with the release of the 0x00000100 revision. This means we are committed to maintaining backward compatibility for all subsequent pre-release revisions until the official 0x00010000 release (planned for around the end of May 2026), at which point final breaking changes may occur to finalize the specification. I have added a warning to each header to ensure this remains clear.
> 
> We have agreed with the corresponding QCOM team that merging this into `EmbeddedPkg` now would be beneficial for the integration process.
> 

If there is an expectation that these protocols might change within the next ~3 months, they should definitely not be merged before that time. Tianocore is not a vehicle for sharing pre-preproduction code between Google and Qualcomm.

Adding warnings to the headers does not actually change anything in that respect: the current versions will likely make the next stable release, but the final versions will not, resulting in two published versions, both of which will need to remain supported going forward. It is really much better to pretend (from upstream pov) that these pre-production versions never existed once they are finalized.

ardbiesheuvel:
> Work in progress PR with GBL-related protocols.
> 
> GBL [upstream headers](https://cs.android.com/android/kernel/superproject/+/common-android-mainline:bootable/libbootloader/gbl/libefi_types/defs/protocols/).
> 

Mind updating this link?

leiflindholm:
> > Hey @leiflindholm, thanks for bumping the PR to prevent it from being closed.
> > As discussed, we have pre-frozen the GBL protocols with the release of the 0x00000100 revision. This means we are committed to maintaining backward compatibility for all subsequent pre-release revisions until the official 0x00010000 release (planned for around the end of May 2026), at which point final breaking changes may occur to finalize the specification. I have added a warning to each header to ensure this remains clear.
> > We have agreed with the corresponding QCOM team that merging this into `EmbeddedPkg` now would be beneficial for the integration process.
> 
> If there is an expectation that these protocols might change within the next ~3 months, they should definitely not be merged before that time.

I don't think that's what @dimorinny meant. The stat of things is:
- It'd be good if we could merge code that will actually be used in production Android devices, based on the protocol already supported by u-boot (but ignoring the fragile/incorrect memreserve operations). Currently, that sounds like an EmbeddedPkg thing to me.
- We want to bring the sanitised version of the protocol into the UEFI spec, at which point we can add that into MdePkg and deprecate the EmbeddedPkg version.

> Tianocore is not a vehicle for sharing pre-preproduction code between Google and Qualcomm.

Absolutely.

> Adding warnings to the headers does not actually change anything in that respect: the current versions will likely make the next stable release, but the final versions will not, resulting in two published versions, both of which will need to remain supported going forward. It is really much better to pretend (from upstream pov) that these pre-production versions never existed once they are finalized.

Agreed on the warnings. Do you feel the same about the rest looking through the goggles I present above?

leiflindholm:
OK, Ard has pinged me on the side and told me I should have my coffee _before_ I start commenting online. This isn't about DT_FIXUP_PROTOCOL.

At which point ... yeah, even for EmbeddedPkg we shouldn't be merging non-stabilised APIs.

If the thing is in flux, but there is a clear intent to merge it once it stabilises, we could always indicate that by creating a staging branch and holding it there.

dimorinny:
Thanks, Ard and Leif, for the clarifications. Let's hold off on that until we're ready to release 1.0. I've moved the PR back to a draft state to reflect this.

dimorinny:
Not stale and going to be updated soon.

EmbeddedPkg/Include/Protocol/GblEfiAvb.h
```diff
@@ -0,0 +1,193 @@
...
+#define GBL_EFI_AVB_BOOT_COLOR_RED_EIO  BIT4
+
+typedef UINT64 GBL_EFI_AVB_PARTITION_FLAGS;
+#define GBL_EFI_AVB_PARTITION_FLAG_VERIFY            BIT0
+#define GBL_EFI_AVB_PARTITION_FLAG_VERIFY_IF_EXISTS  BIT1
+#define GBL_EFI_AVB_PARTITION_FLAG_FLASH_CRITICAL    BIT2
+#define GBL_EFI_AVB_PARTITION_FLAG_FDR               BIT3
+
+typedef enum {
+  GBL_EFI_AVB_KEY_VALIDATION_STATUS_INVALID = 0,
```
leiflindholm: One of the things CI is complaining about is that enum members are supposed to be CamelCase.

EmbeddedPkg/Include/Protocol/GblEfiAvb.h
```diff
@@ -0,0 +1,193 @@
...
+
+/*
+  GBL EFI AVB Protocol.
+  Delegates Android Verified Boot (AVB) board-specific logic to firmware.
+
+  Related docs:
+  https://cs.android.com/android/kernel/superproject/+/common-android-mainline:bootable/libbootloader/gbl/docs/gbl_efi_avb_protocol.md
+*/
+
+#ifndef __GBL_EFI_AVB_PROTOCOL_H__
```
leiflindholm: Since the PR was raised, we have switched the repo to use #pragma once instead of include guards.

EmbeddedPkg/Include/Protocol/GblEfiAvb.h
```diff
@@ -0,0 +1,193 @@
+/** @file
+
+  Copyright (c) 2025, The Android Open Source Project.
```
leiflindholm: Add a trailing `<BR>` to copyright lines to make them more readable (once more are added) in doxygen:d html.

leiflindholm (changes requested):
I've only highlighted the first instance of every pattern, but there are a few style changes needed throughout.

EmbeddedPkg/Include/Protocol/GblEfiAvb.h
dimorinny: Done it. Also moved the protocol description under the same Doxygen-discoverable block, following the other protocols pattern.

EmbeddedPkg/Include/Protocol/GblEfiAvb.h
dimorinny: Done

EmbeddedPkg/Include/Protocol/GblEfiAvb.h
dimorinny: Done

leiflindholm (commented):
@dimorinny I think the code itself looks good (and passes CI).

The PR description needs updating, and it would be nice if we could have a (very brief) description in the commit message of what GBL is and hence what (very high-level) the added protocols are for, instead of just a link to somewhere.

dimorinny:
@leiflindholm thanks for the pointer. Done.

leiflindholm (approved):

PR #11146, opened 2025-05-31
Add edksetup.fish for fish shell users

VoltagedDebunked:
# Description

Added `edksetup.fish` file for fish shell users, so people don't have to use bash (which might not be convenient and will exclude features of the fish shell)

## How This Was Tested

I am a fish shell user, and I've tested this repeatedly and built a test with the new fish configuration in my fish shell.

## Integration Instructions

Simply place the new `edksetup.fish` file into the root of the project.

### DIFF
--- a/edksetup.fish
+++ b/edksetup.fish
@@ -1,0 +1,114 @@
+#
+# Copyright (c) 2006 - 2016, Intel Corporation. All rights reserved.<BR>
+# Copyright (c) 2016, Linaro Ltd. All rights reserved.<BR>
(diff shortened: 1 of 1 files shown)


leiflindholm:
I'm ok with the contribution as a concept. A few things I'd like to see changed:
- Add a commit message.
- Try to align the output with the existing edk2setup.sh

Basically, in order to be able to maintain alignment over time, I would like the output of `diff -u edksetup.sh edksetup.fish` to be as small as possible.

PR #11147, opened 2025-05-31
BaseTools: Add support for mingw-w64

nate-desimone:
# BaseTools: Add support for mingw-w64

Adds support for building the C language BaseTools for Windows using toolchains based on mingw-w64.

Mingw-w64 is a collection of header files, libraries, and tools that when combined with a compiler enable development of Windows software. Mingw-w64 is a fork of the original MinGW (Minimalist GNU for Windows). Most active development on MinGW has ceased and mingw-w64 is now the actively maintained successor.  Mingw-w64 provides a libc implementation built on top of Microsoft's UCRT (Universal C Runtime) with all nessesary compiler bindings needed to support the C++11 feature set.

If one wishes to use EDK II to build UEFI firmware on the Windows platform, then the C BaseTools need to be compiled as Windows applications. This includes the PcdValueInit.exe program, which needs to be recompiled every time a firmware build is run in order to regenerate the initial values for structured PCDs. Currently, BaseTools only supports the Visual C++ toolchain on the Windows platform. The following new features have been added to enable usage of the toolchains derived from mingw-w64:

 - Fixes to the BaseTools C source code to support the use of a GCC-style compiler on the Windows OS.
 - The GNU Make-style Makefiles for the C BaseTools have been modified to support Windows.
 - BaseTools now supports generating GNU Make-style Makefiles on the Windows platform for the purpose of building firmware.
 - `edksetup.bat` has been modified to optionally build BaseTools via mingw-w64. There is no impact to the existing support for Visual C++ and Visual C++ remains the default toolchain.

Modern mingw-w64 development appears to have coalesced around MSYS2, which produces a distributions of both GCC and LLVM/Clang that use mingw-w64 to target the Windows OS. This MSYS2 Clang distribution has a UNIX-like directory layout and includes Windows binaries of GNU Make. Combined with the open source licensing, MSYS2's Clang distribution is a highly attractive choice as an alternative Windows SDK for open source projects such as TianoCore.

- [ ] Breaking change?
- [ ] Impacts security?
- [x] Includes tests?

## How This Was Tested

 - Tested both GCC + mingw-w64 and Clang + mingw-w64 and confirmed that both flavors of mingw-w64 build a working BaseTools.
 - Confirmed that `edksetup.bat` configures the environment correctly for building BaseTools with mingw-w64 when the `Mingw-w64` command line argument is passed.
 - Confirmed that `edksetup.bat Rebuild Mingw-w64` compiles BaseTools using mingw-w64.
  - Tested building BaseTools using mingw-w64 for the IA32, X64, and AARCH64 architectures and confirmed that BaseTools built with mingw-w64 works on all these architectures.
 - Tested the MSYS2 Clang distribution against the official LLVM distribution and can confirm that if the compiler version is the same the emitted machine code is identical between the two. Interestingly, the MSYS2 Clang distribution emits the path to the PDB file using `"/"` as the path seperator instead of `"\"`. That appears to be the only difference in output. Therefore, using the MSYS2 Clang distribution to compile firmware seems a reasonable choice.
 - Tested building `OvmfPkgIa32X64` using the MSYS2 Clang distribution and confirmed it builds and boots.
 - On an Windows ARM64 installation, tested building the `OvmfPkgIa32X64` using the clangarm64 MSYS2 environment. Confirmed that this generates native AARCH64 BaseTools, and confirmed that this toolchain can build OVMF without the need for a cross-compiler.
 - Tested building several Intel proprietary codebases with the MSYS2 Clang distribution and confimed that they build and boot.

## Integration Instructions

For the vast majority of users, the only system setup change nessesary to use a mingw-w64 toolchain is to set the `BASETOOLS_MINGW_PATH` to the directory containing the desired mingw-w64 based toolchain.

A new command line argument has been added to edksetup.bat: Mingw-w64. If this command line argument is set, then the script will set the `BASETOOLS_MINGW_BUILD` environment variable. The user can also opt to set this environment variable manually before running edksetup.bat

If `BASETOOLS_MINGW_BUILD` is defined, then the `BASETOOLS_MINGW_PATH` environment variable must point to the directory containing the mingw-w64 toolchain.

If `CLANG_BIN` is not defined and `%BASETOOLS_MINGW_PATH%\bin\clang.exe` exists, then `edksetup.bat` will set `CLANG_BIN=%BASETOOLS_MINGW_PATH%\bin\`. This removes the requirement to configure the `CLANG_BIN` environment variable manually in order to run a `CLANGPDB` or `CLANGDWARF` build if one has the MSYS2 Clang distribution installed. If one wishes to use a different copy of Clang (for example official LLVM binaries) to build firmware and only use the MSYS2 Clang to build BaseTools, then one can continue to set the `CLANG_BIN` environment variable, same as before.

If `CLANG_HOST_BIN` is not defined and `BASETOOLS_MINGW_BUILD` is defined and `%BASETOOLS_MINGW_PATH%\bin\mingw32-make.exe` exists, then `edksetup.bat` will add `%BASETOOLS_MINGW_PATH%\bin\` to the PATH and set `CLANG_HOST_BIN=mingw32-`. This enable usage of the GNU Make included in the mingw-w64 toolchain to build firmware in addition to BaseTools. if `BASETOOLS_MINGW_BUILD` is not defined, `edksetup.bat` will continue to set `CLANG_HOST_BIN=n`, which uses `nmake` to build firmware. This behavior can be overridden by manually setting the value of `CLANG_HOST_BIN` before executing `edksetup.bat` if one wishes to use a specific Make utility for the `CLANGPDB`/`CLANGDWARF` toolchains.


ardbiesheuvel:
> If one wishes to use EDK II to build UEFI firmware on the Windows platform, then the C BaseTools need to be compiled as Windows applications. This includes the PcdValueInit.exe program, which needs to be recompiled every time a firmware build is run in order to regenerate the initial values for structured PCDs.

This should be fixed - having to rebuild a host build tool for every target platform build is not an acceptable situation.

@mdkinney @leiflindholm @makubacki

nate-desimone:
> This should be fixed - having to rebuild a host build tool for every target platform build is not an acceptable situation.

@ardbiesheuvel - I agree that is not ideal and I am open to discussing solutions for that, but it is also out of scope for the purposes of this PR.

ardbiesheuvel:
> > This should be fixed - having to rebuild a host build tool for every target platform build is not an acceptable situation.
> 
> @ardbiesheuvel - I agree that is not ideal and I am open to discussing solutions for that, but it is also out of scope for the purposes of this PR.

Yeah fair point.

I don't  have any objections to this PR but I don't use Windows so I'll leave it to others to comment.

BaseTools/Source/C/Makefiles/header.makefile
```diff
@@ -5,48 +5,125 @@
...
-  ifndef HOST_ARCH
-    $(info Could not detected HOST_ARCH from uname results)
-    $(error HOST_ARCH is not defined!)
+endif
+ifndef HOST_ARCH
+  $(error HOST_ARCH is not defined!)
+endif
+
+#Set up BaseTools binary path for Windows builds
+ifeq (Windows, $(findstring Windows,$(MAKE_HOST)))
```
ajgrier, line 93: Could everybody use a BIN_PATH instead of only windows?

BaseTools/Source/C/VfrCompile/GNUmakefile
```diff
@@ -38,11 +38,17 @@ include $(MAKEROOT)/Makefiles/header.makefile
...
+BIN_DIR=.
+export BIN_DIR
+
 .PHONY:all
 all: $(MAKEROOT)/bin $(APPLICATION)
 
 $(APPLICATION): $(OBJECTS)
 	$(LINKER) -o $(APPLICATION) $(VFR_LFLAGS) $(OBJECTS) -L$(MAKEROOT)/libs $(LIBS)
+ifeq (Windows, $(findstring Windows,$(MAKE_HOST)))
+	$(CP) $(APPLICATION).exe $(BIN_PATH)
```
ajgrier, line 50: Why doesn't APPLICATION include BIN_PATH? The copy seems like an unnecessary step.

BaseTools/Source/Python/Workspace/DscBuildData.py
```diff
@@ -105,10 +106,16 @@ def _IsFieldValueAnArray (Value):
 %s
 '''
 
-PcdGccMakefile = '''
-MAKEROOT ?= $(EDK_TOOLS_PATH)/Source/C
-LIBS = -lCommon
-'''
+if sys.platform == "win32":
+    PcdGccMakefile = '''
```
ajgrier: use a formatted string and os.sep here?

PcdGccMakefile = f'\nMAKEROOT ?= $(EDK_TOOLS_PATH){os.sep}Source{os.sep}C\nLIBS = -lCommon\n'

BaseTools/Source/Python/Workspace/DscBuildData.py
```diff
@@ -2957,12 +2979,16 @@ def GenerateByteArrayValue (self, StructuredPcds):
...
         MakeApp = PcdMakefileHeader
-        if sys.platform == "win32":
+        if sys.platform == "win32" and not self._MingwBaseToolsBuild:
             MakeApp = MakeApp + 'APPFILE = %s\\%s.exe\n' % (self.OutputPath, PcdValueInitName) + 'APPNAME = %s\n' % (PcdValueInitName) + 'OBJECTS = %s\\%s.obj %s.obj\n' % (self.OutputPath, PcdValueInitName, os.path.join(self.OutputPath, PcdValueCommonName)) + 'INC = '
         else:
             MakeApp = MakeApp + PcdGccMakefile
-            MakeApp = MakeApp + 'APPFILE = %s/%s\n' % (self.OutputPath, PcdValueInitName) + 'APPNAME = %s\n' % (PcdValueInitName) + 'OBJECTS = %s/%s.o %s.o\n' % (self.OutputPath, PcdValueInitName, os.path.join(self.OutputPath, PcdValueCommonName)) + \
-                      'include $(MAKEROOT)/Makefiles/app.makefile\n' + 'TOOL_INCLUDE +='
+            if sys.platform == "win32":
+                MakeApp = MakeApp + 'APPFILE = %s\\%s.exe\n' % (self.OutputPath, PcdValueInitName) + 'APPNAME = %s\n' % (PcdValueInitName) + 'OBJECTS = %s/%s.o %s.o\n' % (self.OutputPath, PcdValueInitName, os.path.join(self.OutputPath, PcdValueCommonName)) + \
```
ajgrier: single parameterized formatted string here, as above?

BaseTools/Source/Python/Workspace/DscBuildData.py
```diff
@@ -3010,11 +3036,14 @@ def GenerateByteArrayValue (self, StructuredPcds):
...
         MakeApp += CC_FLAGS
 
-        if sys.platform == "win32":
+        if sys.platform == "win32" and not self._MingwBaseToolsBuild:
             MakeApp = MakeApp + PcdMakefileEnd
             MakeApp = MakeApp + AppTarget % ("""\tcopy $(APPLICATION) $(APPFILE) /y """)
         else:
-            MakeApp = MakeApp + AppTarget % ("""\tcp -p $(APPLICATION) $(APPFILE) """)
+            if sys.platform == "win32":
+                MakeApp = MakeApp + AppTarget % ("""\t$(CP) $(APPLICATION).exe $(APPFILE) /y""")
```
ajgrier: seems like this could be collapsed down to a single parameterized line as well. non-windows should be able to cope with `$(CP)`, `$(APPSUFFIX)` would be .exe on windows, empty on non-windows, and `$(CP_ARGS)` or something like it would be /y on windows but empty on non-windows.

BaseTools/Tests/GNUmakefile
```diff
@@ -1,15 +1,57 @@
...
+# Copyright (c) 2009 - 2025, Intel Corporation. All rights reserved.<BR>
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 #
 
 all: test
 
-test:
-	@if command -v "${PYTHON_COMMAND}" >/dev/null 2>&1; then ${PYTHON_COMMAND} RunTests.py; else python RunTests.py; fi
+ifndef PYTHON_COMMAND
+  ifeq (Windows, $(findstring Windows,$(MAKE_HOST)))
```
ajgrier, line 11: can we reuse header.makefile here?

BaseTools/Source/C/Makefiles/GnuMakeUtils.py
ajgrier, line 1: get_host_arch seems overwrought -- under what conditions does native python platform.machine() produce an incorrect host architecture?

I understand wanting to abstract file operations (cp, mv, rm, etc) but its use seems to be conditional on MAKE_HOST instead of embracing the abstraction across windows and non-windows usages. I support abstracting operations into variables, (`$(CP)` instead of `cp` or `copy`) but seems like that logic could be implemented in makefile logic instead of calling out into a python re-implementation of unix-style operations.

if you want to go down the python abstraction layer route, everybody needs to use it, with no if/else based on MAKE_HOST where it's used. common `$(CP)` used for everybody.

ajgrier (approved):
Ideally OS-specific details can be concentrated in makefiles or scripts without need for if/else based on OS anywhere else, but the first step is the hardest.

BaseTools/Source/C/Makefiles/header.makefile
nate-desimone: Reason for BIN_PATH is because of differences with how the C BaseTools is set up between Windows vs. UNIX systems. On Windows systems, all of the .exe files get copied to BaseTools\Bin\Win32. On UNIX systems, binaries are found in BaseTools/Source/C/bin. There are several scripts that expect them to be present in the BaseTools\Bin, so extra copy is to provide backward compatibility with that expectation on Windows systems.

BaseTools/Source/C/VfrCompile/GNUmakefile
nate-desimone: Reason for BIN_PATH is because of differences with how the C BaseTools is set up between Windows vs. UNIX systems. On Windows systems, all of the .exe files get copied to BaseTools\Bin\Win32. On UNIX systems, binaries are found in BaseTools/Source/C/bin. There are several scripts that expect them to be present in the BaseTools\Bin, so extra copy is to provide backward compatibility with that expectation on Windows systems.

BaseTools/Source/Python/Workspace/DscBuildData.py
nate-desimone: Done

BaseTools/Source/Python/Workspace/DscBuildData.py
nate-desimone: Done

BaseTools/Source/Python/Workspace/DscBuildData.py
nate-desimone: Done

BaseTools/Tests/GNUmakefile
nate-desimone: header.makefile has several relative paths based on $(MAKEROOT) that won't work from BaseTools/Tests

BaseTools/Source/C/Makefiles/GnuMakeUtils.py
nate-desimone: > get_host_arch seems overwrought -- under what conditions does native python platform.machine() produce an incorrect host architecture?

On UNIX systems, platform.machine() is a wrapper around `uname -m`. On Windows systems, it uses WMI to determine the CPU architecture. Both of these methods have some corner cases which the get_host_arch function handles.

On Windows systems, HOST_ARCH must be set based on which machine architecture mingw-w64 was compiled for, not the kernel's architecture.

> I understand wanting to abstract file operations (cp, mv, rm, etc) but its use seems to be conditional on MAKE_HOST instead of embracing the abstraction across windows and non-windows usages. I support abstracting operations into variables, (`$(CP)` instead of `cp` or `copy`) but seems like that logic could be implemented in makefile logic instead of calling out into a python re-implementation of unix-style operations.
> 
> if you want to go down the python abstraction layer route, everybody needs to use it, with no if/else based on MAKE_HOST where it's used. common `$(CP)` used for everybody.

Done

gapalomi (approved):

makubacki:
> > If one wishes to use EDK II to build UEFI firmware on the Windows platform, then the C BaseTools need to be compiled as Windows applications. This includes the PcdValueInit.exe program, which needs to be recompiled every time a firmware build is run in order to regenerate the initial values for structured PCDs.
> 
> This should be fixed - having to rebuild a host build tool for every target platform build is not an acceptable situation.
> 
> @mdkinney @leiflindholm @makubacki

Responding late as I was out of office last week. There was an effort to move basetools to Python in https://github.com/tianocore/edk2-basetools, however that has stalled. I think this would be a good topic for the Tianocore Tools & CI meeting (@spbrogan).

bexcran:
@makubacki Personally I was glad the effort to rewrite the BaseTools in Python stalled, because the existing Python code is really quite bad. I don't think it was a good use of time/effort.

makubacki:
> @makubacki Personally I was glad the effort to rewrite the BaseTools in Python stalled, because the existing Python code is really quite bad. I don't think it was a good use of time/effort.

I agree it looked like a lot of work was needed. Details aside, I just wanted to mention that solution path is no longer an active option.

nate-desimone:
@makubacki, @bexcran - On the topic of pure Python BaseTools, there are some very difficult to break dependencies. As I somewhat alluded to in the cover letter for this PR, generating the initial values for Structured PCDs requires one to serialize c-structs.

I don't see a pure Python BaseTools being possible without the existence of a pure Python C compiler. If such a thing exists, we would then need to be able integrate that pure Python C compiler into BaseTools and use its lexer to generate the PCD c-structs in binary format.

All that said, I would be happy to present the work done for this PR in the Tools meeting if there is interest.

mdkinney:
There are some good Python modules that parse C code.  I have used pyclibrary to parse library class .h files to help with generation of GoogleMocks.

https://github.com/mdkinney/edk2/blob/Bug_xxx_GenerateMockLibraries/BaseTools/Scripts/GenerateMocksFromLibClass.py

pyclibrary may be able to parse the PCD c-structs for this use case.  Would need further investigation.

PR #11153, opened 2025-06-04
NetworkPkg: Add PCD to control http boot enable or disable.

mark-skum:
Add PCD to control http boot enable or disable follow PXE boot.

### DIFF
--- a/ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c
+++ b/ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c
@@ -797,52 +797,6 @@ InitializeMiscMmCommunicateBuffer (
 }
 
 /**
-  Convert UUID to EFI_GUID format.
-  for example, If there is EFI_GUID named
-  "378daedc-f06b-4446-8314-40ab933c87a3",
-
-  EFI_GUID is saved in memory like:
-     dc ae 8d 37
-     6b f0 46 44
-     83 14 40 ab
-     93 3c 87 a3
-
-  However, UUID should be saved like:
-     37 8d ae dc
-     f0 6b 44 46
-     83 14 40 ab
-     93 3c 87 a3
-
-  FF-A and other software components (i.e. linux-kernel)
-  uses below format.
-
-  To patch mm-service properly, the passed uuid should be converted to
-  EFI_GUID format.
-
-  @param [in]  Uuid            Uuid
-  @param [out] Guid            EFI_GUID
-
-**/
-STATIC
-VOID
-EFIAPI
-ConvertUuidToEfiGuid (
-  IN  UINT64    *Uuid,
-  OUT EFI_GUID  *Guid
-  )
-{
-  UINT32  *Data32;
-  UINT16  *Data16;
-
-  Data32    = (UINT32 *)Uuid;
-  Data32[0] = SwapBytes32 (Data32[0]);
-  Data16    = (UINT16 *)&Data32[1];
-  Data16[0] = SwapBytes16 (Data16[0]);
-  Data16[1] = SwapBytes16 (Data16[1]);
-  CopyGuid (Guid, (EFI_GUID *)Uuid);
-}
-
-/**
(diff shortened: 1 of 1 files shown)


mark-skum:
@ajfish Please help to have a review, thanks

NetworkPkg/NetworkDynamicPcds.dsc.inc
```diff
@@ -14,7 +14,10 @@
 #
 # IPv4 and IPv6 PXE Boot support.
 #
-!if ($(NETWORK_ENABLE) == TRUE) AND ($(NETWORK_PXE_BOOT_ENABLE) == TRUE)
```
leiflindholm: I don't know why this change is here, but it has nothing to do with the stated purpose of the commit or the PR in general.

mark-skum: Thanks for your review
If the PCD is referenced in the INF file, attempting to disable NETWORK_PXE_BOOT_ENABLE will result in a build error.
 I think, it is not a good practice to define PCD with an if statement.So, I took the opportunity to make some modifications.

mark-skum: for example,
If PcdIPv4PXESupport has already been used in edk2-Platform, if some projects do not require the Pxe boot function, then NETWORK_PXE_BOOT_ENABLE should be disabled. At the same time, PcdIPv4PXESupport in edk2-Platform should also be removed. This makes code maintenance very inconvenient.

mark-skum: The definition of PCD is controlled by if statements, which has had a negative impact on the development of the code.

leiflindholm: That sounds to me like the subject of a PR called "NetworkPkg: rework conditional PXEBOOT handling". Not something to try to latch onto one called "NetworkPkg: Add PCD to control httpboot".

mark-skum: What you mean is that I need to restore this part and then re-upload the related content about this PXEBOOT.

mark-skum: I'll make some revisions again.

leiflindholm: Yes please. Leave this logic unchanged as part of this PR and raise a separate PR for the changes you want to make to the conditionality.

xpahos: > I think, it is not a good practice to define PCD with an if statement.So, I took the opportunity to make some modifications.

Just a quick note on how this worked before adding the option to disable PXE. When the network boot option was disabled, PXE was also disabled. Is it really necessary to remove the conditional variable settings?

NetworkPkg/NetworkDynamicPcds.dsc.inc
```diff
@@ -14,7 +14,10 @@
...
 # IPv4 and IPv6 PXE Boot support.
 #
-!if ($(NETWORK_ENABLE) == TRUE) AND ($(NETWORK_PXE_BOOT_ENABLE) == TRUE)
   gEfiNetworkPkgTokenSpaceGuid.PcdIPv4PXESupport|0x01
   gEfiNetworkPkgTokenSpaceGuid.PcdIPv6PXESupport|0x01
-!endif
+#
+# IPv4 and IPv6 HTTP Boot support.
+#
+  gEfiNetworkPkgTokenSpaceGuid.PcdIPv4HttpSupport|0x01
```
SaloniKasbekar: What is the benefit of adding a separate PCD for IPv4 and IPv6? Wouldnt a single PcdHttpBootSupport be enough?

NetworkPkg/NetworkDynamicPcds.dsc.inc
mark-skum: There is a need to dynamically modify the PXE boot through the setup option.

NetworkPkg/NetworkDynamicPcds.dsc.inc
mark-skum: Because there are 4 boot options, we want to control each one separately.
image

PR #11158, opened 2025-06-05
BaseTools\Plugin\HostBasedUnitTestRunner: Reduce the unit test duplicate execution

GeorgeLiao012:
# Description

The HostBasedUnitTestRunner.py is part of unit test that used to execute unit test application and generated code coverage report. Code coverage report generation is different in the different operation system. Please refer it as below.

[Windows]
	1.Run unit test application 
	2.Adopt OpenCppCoverage tool to run unit test application to generate coverage report.
[Linux]
	1.Run unit test application to generate .gcda and .gcno data.
	2.Adopt Lcov tool to collect .gcda and .gcno data then convert it to the coverage report. 
	
According to the above description, in the Windows the unit test application will be executed twice, But it is not necessary. 
So adopt variable CODE_COVERAGE to control it, unit test application only execute once in the windows for save time.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Local unit test build under windows and Linux.

## Integration Instructions

No integration required.
Just reduce unit test execution in the Windows.

### DIFF
--- a/BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
+++ b/BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
@@ -110,6 +110,14 @@ class HostBasedUnitTestRunner(IUefiBuildPlugin):
                 return 0
 
             for test in testList:
+                #
+                # The gen_code_coverage_msvc() function uses the OpenCppCoverage tool to generate code coverage data,
+                # so unit test execution for the VS20XX toolchain can be skipped in this loop.
+                #
+                if (thebuilder.env.GetValue("CODE_COVERAGE") != "FALSE") and \
+                    (thebuilder.env.GetValue("TOOL_CHAIN_TAG").startswith("VS")):
+                    continue
+
                 # Configure output name if test uses cmocka.
                 shell_env.set_shell_var(
                     'CMOCKA_XML_FILE', test + ".CMOCKA.%g." + arch + ".result.xml")


BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -106,36 +106,37 @@ def do_post_build(self, thebuilder):
...
-                            for case in suite:
-                                for result in case:
-                                    if result.tag == 'failure':
-                                        logging.warning(
-                                            "%s Test Failed" % os.path.basename(test))
-                                        logging.warning(
-                                            "  %s - %s" % (case.attrib['name'], result.text))
-                                        failure_count += 1
+                # If "CODE_COVERAGE" equal "FALSE" or used GCC5 then run unit test.
+                if (thebuilder.env.GetValue("CODE_COVERAGE") == "FALSE") or (thebuilder.env.GetValue("TOOL_CHAIN_TAG") == "GCC5"):
```
mdkinney: Why run for GCC5 here?  That would mean GCC5 builds would still run unit tests twice.

GeorgeLiao012: The coverage data generation is different between windows (OpenCoverage) and Linux (Gcov/LCOV)
 
Windows  :
1.The HostBasedUnitTestRunner used OpenCppCoverage to execute unit test app for generate coverage data (*.cov)
2.The coverage data will be convert to the coverage.xml by HostBasedUnitTestRunner.py
 
In the windows OS, the coverage data generation doesn’t rely on the Unit Test execution, so we can remove it from windows OS.
 
Linux  :
1. Added --coverage parameter in the unit test build process for generate *.gcno file.
2. Executing unit test app which generates the *.gcda file.
3. The HostBasedUnitTestRunner used locv to digest *.gcno and *.gcda for generate  coverage.xml
 
As *.gcda generation depends on executing unit test app, so executing unit test app need to be reserved for Linux environment. 
By the way, the unit test app will be execute only one time in the Linux environment.

mdkinney: The logic here is hard to follow.  There needs to be a more complete description in the PR and the commit message for this change to help with the reviews.

The unit tests are executed in 2 places in Windows.  One is in this loop.  The other is in the method gen_code_coverage_msvc() that runs each unit test through OpenCppCoverage.

The goal of this change is to make sure the call to gen_code_coverage_msvc() is the only time the unit test is executed when code coverage information is being collected.  If code coverage information is not being collected then the call to gen_code_coverage_msvc() will not be made and the unit tests must be executed in this loop.

Also, if the unit tests are being executed in a Linux env, then gen_code_coverage_msvc() will never be called and OpenCppCoverage will never be called.  Unfortunately, the check for GCC5 is being used as a check for Linux env which is also confusing.  The check for the OS type is used above to collect the set of Unit Test Apps, but GCC5 is used elsewhere.  Also, there is both GCC and GCC5 tool chain tags now, so the GCC5 check is incomplete.  It would be better to use OS type check consistently instead of the tool chain tag checks.  But I think these clean ups can be done in a different PR.

So the logic change here is to skip the direct execution of the unit test apps if the unit test apps will also be executed with OpenCppCoverage.  Or the unit test apps must be executed in this loop if the OS type is Linux or code coverage information is not being collected.  The check for GCC5 is a proxy for the OS is Linux check.

The logic in this code change looks correct to me.

Please update the PR description and add details to the commit message to describe the logic change and why running the unit tests apps in this loop is correct when CODE_COVERAGE == FALSE or TOOL_CHAN_TAG == GCC5 and how that guarantees that unit tests are executed only one for all conditions.  May need to add a small logic table to make it clear.

CODE_COVERAGE  |  TOOL_CHAIN_TAG |  Where Unit Tests Run
-------------------- | --------------------- |  ----------------
FALSE | GCC5 | do_post_build() loop
FALSE | VSxxxx | do_post_build() loop
TRUE | GCC5 | do_post_build() loop 
TRUE | VSxxxx | gen_code_coverage_msvc()

mdkinney (approved):

mdkinney:
@Javagedes Can you please review this plugin optimization for Windows unit test builds.

Javagedes:
@mdkinney yes I can review it. I'm currently out of town on vacation until Wednesday, so if we can hold this until then, I'd be happy to!

GeorgeLiao012:
> @mdkinney yes I can review it. I'm currently out of town on vacation until Wednesday, so if we can hold this until then, I'd be happy to!

@mdkinney and @Javagedes Thank you for your help.

spbrogan:
A few things that i would suggest be addressed. 

- [ ] Please fill in the PR template and clean up the content you don't plan to use
- [ ] I still see GCC5 as a proxy for all GCC and Linux.  Since GCC5 was depreciated 2 years ago it should have included cleaning up any basetool assumptions on GCC5.  All CI infrastructure should be using GCC and clean up any dependencies on GCC5.  Please don't add more debt.
- [ ] Mike's PR comment is far more descriptive than the comment left in the code.  Please add the additional context to the code.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -106,36 +106,48 @@ def do_post_build(self, thebuilder):
...
+                # ________________________________________________________________
+                # | CODE_COVERAGE   | TOOL_CHAIN_TAG  | Where Unit Tests Run      |
+                # ________________________________________________________________
+                # | FALSE           | GCC5            | do_post_build() loop      |
+                # | FALSE           | VSxxxx          | do_post_build() loop      |
+                # | TRUE            | GCC5            | do_post_build() loop      |
+                # | TRUE            | VSxxxx          | gen_code_coverage_msvc()  |
+                # |_________________|_________________|___________________________|
+                #
+                if (thebuilder.env.GetValue("CODE_COVERAGE") == "FALSE") or (thebuilder.env.GetValue("TOOL_CHAIN_TAG") == "GCC5"):
```
Javagedes: 1. Please update this GCC toolchain to search be `thebuilder.env.GetValue("TOOL_CHAIN_TAG").startswith("GCC")`
2. We already have a lot of nesting in this bit of code, which is not your fault, but I'd like to prevent further nesting. Please update this if check to check the negative of this, and just `continue`, instead of shifting all of this code to the right with another nest.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -106,36 +106,48 @@ def do_post_build(self, thebuilder):
...
+                                    for result in case:
+                                        if result.tag == 'failure':
+                                            logging.warning(
+                                                "%s Test Failed" % os.path.basename(test))
+                                            logging.warning(
+                                                "  %s - %s" % (case.attrib['name'], result.text))
+                                            failure_count += 1
 
             if thebuilder.env.GetValue("CODE_COVERAGE") != "FALSE":
                 if thebuilder.env.GetValue("TOOL_CHAIN_TAG") == "GCC5":
```
Javagedes: This actually looks like another bug not necessarily related to your PR. Can you update this to be

```suggestion
                if thebuilder.env.GetValue("TOOL_CHAIN_TAG").startswith("GCC"):
```

mdkinney: @GeorgeLiao012 Please implement this suggestion so this works for GCC and GCC5 and GCCNOLTO tool chain tags

mdkinney:
@GeorgeLiao012 please review and update for all feedback in this PR.  For example, the PR summary is not filled in correctly and the table you added is too brief.  Please add a more complete description of the change.

I also see the comment block added has very long lines.  Please format comments to 80 columns.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
GeorgeLiao012: Done, I have followed your feedback to change PR.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
GeorgeLiao012: Done, I have updated it.

GeorgeLiao012:
> @GeorgeLiao012 please review and update for all feedback in this PR. For example, the PR summary is not filled in correctly and the table you added is too brief. Please add a more complete description of the change.
> 
> I also see the comment block added has very long lines. Please format comments to 80 columns.

I have completed it. Thank you very much.

mdkinney:
reopen.

mdkinney:
reopen

mikebeaton:
Dear @GeorgeLiao012 - I believe this PR looks good and worth merging.

Could you possibly update it to handle CLANGDWARF/CLANGPDB by adding CLANG* logic as well?

Note in your summary table it should say GCC* not GCC5 (in fact, GCC is the current full-support toolchain) and then similar logic and table entries for CLANG* (which will handle CLANGDWARF and CLANGPDB).

GeorgeLiao012:
> Dear @GeorgeLiao012 - I believe this PR looks good and worth merging.
> 
> Could you possibly update it to handle CLANGDWARF/CLANGPDB by adding CLANG* logic as well?
> 
> Note in your summary table it should say GCC* not GCC5 (in fact, GCC is the current full-support toolchain) and then similar logic and table entries for CLANG* (which will handle CLANGDWARF and CLANGPDB).

Sure, but please allow me more time to do it, I will update this pr as soon as possible.

mdkinney:
@GeorgeLiao012 Please sync to latest and address comments

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -110,6 +110,10 @@ def do_post_build(self, thebuilder):
                 return 0
 
             for test in testList:
+                if (thebuilder.env.GetValue("CODE_COVERAGE") != "FALSE") and \
```
mdkinney, line 117: Add a comment that gen_code_coverage_msvc() executes unit tests to collect code coverage information, so the execution of unit tests for VS20xx tool chain tags can be skipped in this loop if code coverage information is being collected.

GeorgeLiao012: I have added the comment. Thanks.

mdkinney:
@Javagedes Please review the updated version of this PR.  LGTM

GeorgeLiao012:
@Javagedes @mikebeaton 
Please review the updated version of this PR. Thank you.

apop5:
@GeorgeLiao012 

Could you please rebase this PR to the latest edk2 code to get clang checks to run?

PR #11160, opened 2025-06-05
Update SnpDxe for alignment fault issue

eeshanl:
# Description

Update SnpDxe for alignment fault issue
- Updates SNP_DRIVER allocation to use AllocatePages()
- Allocates the PXE_CDB struct as a pointer instead, using PciIo->AllocateBuffer()

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested with PXE boot on physical arm device

## Integration Instructions

N/A


NetworkPkg/SnpDxe/Snp.c
```diff
@@ -361,24 +361,31 @@ SimpleNetworkDriverStart (
...
+    DEBUG ((DEBUG_NET, "\nCould not allocate CDB structures.\n"));
+    goto Error_DeleteSNP;
   }
 
-  Snp = (SNP_DRIVER *)(UINTN)Address;
-
   ZeroMem (Snp, sizeof (SNP_DRIVER));
 
+  Snp->Cdb = (PXE_CDB *)(UINTN)Address;
+  ZeroMem (Snp->Cdb, sizeof (PXE_CDB));
```
ardbiesheuvel: Can we combine this allocation with the one a few lines down? The CPB and DB fields are already allocated together.

eeshanl: Done

ardbiesheuvel: You are still calling AllocateBuffer twice

eeshanl: Oh I misunderstood, you are asking to use one AllocateBuffer call for both the inner structures as adjacent. Will update this. Thanks.

eeshanl: Updated to 1 AllocateBuffer() call. I have CDB first then CPB adjacently, since I explicitly know the sizeof(PXE_CDB) and not CPB, CDB is first.

ardbiesheuvel: AFAICT `sizeof(PXE_CDB)` equals 32 bytes, and taking those from the 2048 byte CPB allocation seems fine to me, as 2016 bytes is still ample for all the PXE_CPB types I can see in the code. I am by no means a PXE/UNDI expert, though :-)

eeshanl: Thanks for the review!

eeshanl:
@ardbiesheuvel thanks for the comments/reviews. 

@SaloniKasbekar @Zclarkwilliams please take a look for review. Thank you!

NetworkPkg/SnpDxe/Snp.c
```diff
@@ -361,22 +361,12 @@ SimpleNetworkDriverStart (
...
-                    PciIo,
-                    AllocateAnyPages,
-                    EfiBootServicesData,
-                    SNP_MEM_PAGES (sizeof (SNP_DRIVER)),
-                    &Address,
-                    0
-                    );
-
-  if (Status != EFI_SUCCESS) {
+  Snp = AllocatePages (SNP_MEM_PAGES (sizeof (SNP_DRIVER)));
```
mdkinney: If this is a normal buffer with no DMA, then can this be changed to AllocatePool()?

NetworkPkg/SnpDxe/Snp.c
```diff
@@ -463,8 +453,10 @@ SimpleNetworkDriverStart (
     goto Error_DeleteSNP;
   }
 
-  Snp->Cpb = (VOID *)(UINTN)Address;
-  Snp->Db  = (VOID *)((UINTN)Address + 2048);
+  Snp->Cdb = (VOID *)(UINTN)Address;
```
mdkinney: With this change, why use hard coded 4096?  Also, before Db could be aligned at a 2048 boundary.  Now base of Db depends on sizeof (PXE_CDB).  Does Db have alignment requirements for DMA?

NetworkPkg/SnpDxe/Snp.c
```diff
@@ -463,8 +453,10 @@ SimpleNetworkDriverStart (
     goto Error_DeleteSNP;
   }
 
-  Snp->Cpb = (VOID *)(UINTN)Address;
-  Snp->Db  = (VOID *)((UINTN)Address + 2048);
+  Snp->Cdb = (VOID *)(UINTN)Address;
+  Snp->Cpb = (VOID *)((UINTN)Address + (UINTN)sizeof (PXE_CDB));
+  Snp->Db  = (VOID *)((UINTN)Snp->Cpb + 2048);
+  ZeroMem ((VOID *)(UINTN)Address, 4096);
```
mdkinney: The commit message only mentions making sure only CDB is in DMA capable memory.  However, this DMA memory allocation is used for CDB, CPB, and DB.  Do all 3 need to be in DMA capable memory and the commit message is incorrect.  Or do CPB and DB need to be moved into non DMA memory.

NetworkPkg/SnpDxe/Snp.c
```diff
@@ -693,11 +685,7 @@ SimpleNetworkDriverStart (
     FreePool (Snp->RecycledTxBuf);
   }
 
-  PciIo->FreeBuffer (
-           PciIo,
-           SNP_MEM_PAGES (sizeof (SNP_DRIVER)),
-           Snp
-           );
+  FreePages (Snp, SNP_MEM_PAGES (sizeof (SNP_DRIVER)));
```
mdkinney: Should this be FreePool()?

NetworkPkg/SnpDxe/Snp.c
```diff
@@ -813,14 +801,9 @@ SimpleNetworkDriverStop (
...
-           Snp->Cpb
-           );
-
-  PciIo->FreeBuffer (
-           PciIo,
-           SNP_MEM_PAGES (sizeof (SNP_DRIVER)),
-           Snp
+           Snp->Cdb
            );
+  FreePages (Snp, SNP_MEM_PAGES (sizeof (SNP_DRIVER)));
```
mdkinney: Should this be FreePool()?

mdkinney (changes requested):

NetworkPkg/SnpDxe/Snp.c
ardbiesheuvel: CPB and DB were already being allocated together in DMA capable memory. It was I who suggested to combine CDB with this allocation rather than performing yet another PciIo->AllocateBuffer() call.

Re alignment, I would not expect this to make any difference - the size and alignment appear to be rather arbitrary. However, given that CDB is just 32 bytes, we could perhaps swap these, and but CDB at PAGE_SIZE-32 in the allocation.

NetworkPkg/SnpDxe/Snp.c
os-d: Agreed, these (and the associated free calls) should be moved to the pool APIs.

NetworkPkg/SnpDxe/Snp.c
os-d: The 4096 should be moved to EFI_PAGE_SIZE.

The reason that DB was set at an offset of 2048 previously, I believe, was just to split the page evenly between CPB and DB, not that there was a specific 2048 requirement. Alignment of its members should be met, but this gets a little weird because there are a handful of CPB and DB structures defined by the PXE spec and it depends on what operation is happening. At a minimum, 8 byte alignment should be maintained for the start of the structures. These structures are all fairly small (<100 bytes IIRC), so it would also be easy to align DB back at the 2048 boundary without issue. This weirdness is why the original PR allocated a new buffer for CDB and left CPB and DB alone, but Ard asked to move CDB to this structure to save the extra page (which makes sense but does require this odd dance).

For ARM64, there were very strict alignment requirements if NonDiscoverablePciDxe was being used (say a USB NIC) but we've had a discussion with Ard on the mailing list and a PR is going in to fix NonDiscoverablePciDxe so that it doesn't mark host buffers as device memory (which triggers the alignment requirements).

NetworkPkg/SnpDxe/Snp.c
os-d: Yeah, putting CDB at the end is probably a better path forward, leave the current structures alone, though agreed that it shouldn't make a difference.

kuqin12: @ardbiesheuvel can you please reiterate why combining all these fields would be a good idea, other than saving some space? It makes the code very hacky. Unless we define them in a structure with explicit paddings, the implementation will be very fragile and could break in the next code change.

Also, would this be boot service memory that will get reclaimed once OS takes over? Why would it matter that much for saving transient memory usage? Any input is appreciated.

NetworkPkg/SnpDxe/Snp.c
eeshanl: yes, changed to FreePool()

NetworkPkg/SnpDxe/Snp.c
eeshanl: yes, changed to FreePool()

NetworkPkg/SnpDxe/Snp.c
eeshanl: Updated to use EFI_PAGE_SIZE and (EFI_PAGE_SIZE / 2)

NetworkPkg/SnpDxe/Snp.c
eeshanl: Changed to AllocatePool()

NetworkPkg/SnpDxe/Snp.c
ardbiesheuvel: > @ardbiesheuvel can you please reiterate why combining all these fields would be a good idea, other than saving some space? It makes the code very hacky. Unless we define them in a structure with explicit paddings, the implementation will be very fragile and could break in the next code change.
> 

Two of those fields are already combined, and we only need 32 bytes of additional space, which we have. 

Allocating a buffer like this involves allocating a page, and then descending into the MMU logic to update the mapping attributes of that page. This could trigger a split of a 1G page into 2M blocks, and subsequently, splitting such a 2M block into 4k pages so that the attributes can be updated on a single page.

Doing this 2 or 3 times for a couple of buffers that can all comfortably share 1 single page of memory makes no sense at all.

> Also, would this be boot service memory that will get reclaimed once OS takes over? Why would it matter that much for saving transient memory usage? Any input is appreciated.

It is not about preserving memory at all.

kuqin12: How often do we use this routine? Given this is in `SimpleNetworkDriverStart`, I assume it will only happen when there is a nic controller connected? The MMU logic you mentioned will happen on all allocate page calls if memory protection is enabled anyway (which they should for the sake of security). I think the saving will be minimal compared to other bazillion allocations during the boot process.

However, take one step back, using the page allocated for CPB as a scratch pad is not a good pattern to follow. It sprinkles magic numbers in the implementation details. Can we define an internal structure for this instead? So that we can move away from the scratch pad buffer usage and better assure the alignment without manual calculations?

```
typedef union {
    PXE_CPB_START_30 cpb_start_30;
    PXE_CPB_START_31 cpb_start_31;
    // More CPB definitions here...
} PXE_CPB;

typedef union {
    PXE_DB_INITIALIZE db_init;
    PXE_DB_RECEIVE_FILTERS db_rec_fltr;
    // More DB definitions here...
} PXE_DB;

typedef struct {
  PXE_CPB Cpb;
  UINT8 Padding[SIZE_2KB - sizeof (PXE_CPB)];
  PXE_DB DB;
  UINT8 Padding2[SIZE_2KB - sizeof (PXE_DB) - sizeof (PXE_CDB)];
  PXE_CDB CDB;
} SNP_DRIVER_2;

STATIC_ASSERT (OFFSET_OF (DB, SNP_DRIVER_2) == EFI_PAGE_SIZE / 2);
STATIC_ASSERT (sizeof (SNP_DRIVER_2) == EFI_PAGE_SIZE);
// More checks if needed for alignment assurance.
```

eeshanl: Hi @ardbiesheuvel updated to move CDB to the very end of the page, at EFI_PAGE_SIZE - 32

eeshanl:
Hi @ardbiesheuvel @mdkinney @os-d , addressed all your comments. Please take another look for re-review. Thanks!

NetworkPkg/SnpDxe/Snp.c
ardbiesheuvel: > The MMU logic you mentioned will happen on all allocate page calls if memory protection is enabled anyway (which they should for the sake of security).

Not quite. The common case is still that all memory is RW- and is transitioned to R-X only for executable pages. Also, memory type changes require more elaborate handling on ARM, whereas permission attribute change can usually be made without break-before-make (BBM) [unless a split is needed]

But you are right that this is a stone cold code path and so correctness should be prioritized over efficiency.


> I think the saving will be minimal compared to other bazillion allocations during the boot process.
>

Fair enough.
 
> However, take one step back, using the page allocated for CPB as a scratch pad is not a good pattern to follow. It sprinkles magic numbers in the implementation details. Can we define an internal structure for this instead? So that we can move away from the scratch pad buffer usage and better assure the alignment without manual calculations?
> 
> ```
> typedef union {
>     PXE_CPB_START_30 cpb_start_30;
>     PXE_CPB_START_31 cpb_start_31;
>     // More CPB definitions here...
> } PXE_CPB;
> 
> typedef union {
>     PXE_DB_INITIALIZE db_init;
>     PXE_DB_RECEIVE_FILTERS db_rec_fltr;
>     // More DB definitions here...
> } PXE_DB;
> 
> typedef struct {
>   PXE_CPB Cpb;
>   UINT8 Padding[SIZE_2KB - sizeof (PXE_CPB)];
>   PXE_DB DB;
>   UINT8 Padding2[SIZE_2KB - sizeof (PXE_DB) - sizeof (PXE_CDB)];
>   PXE_CDB CDB;
> } SNP_DRIVER_2;
> 
> STATIC_ASSERT (OFFSET_OF (DB, SNP_DRIVER_2) == EFI_PAGE_SIZE / 2);
> STATIC_ASSERT (sizeof (SNP_DRIVER_2) == EFI_PAGE_SIZE);
> // More checks if needed for alignment assurance.
> ```

If you are volunteering to clean up this driver, I won't stand in your way :-)

But my point stands: either do it properly, or just add 1 additional buffer to the PCI memory allocation that already covers two similar buffers in the same context.

ardbiesheuvel:
Please fix the patch subject line to it starts with `NetworkPkg/SnpDxe:`

eeshanl:
Hi @ardbiesheuvel @mdkinney @os-d @kuqin12 sorry was oof last week.
I updated to use individual AllocateBuffer calls for the 3 structures.
Please take another look for re-review. Thanks!

NetworkPkg/SnpDxe/Snp.c
```diff
@@ -361,22 +363,12 @@ SimpleNetworkDriverStart (
...
-  if (Status != EFI_SUCCESS) {
+  Snp = AllocatePool (sizeof (SNP_DRIVER));
+  if (Snp == NULL) {
     DEBUG ((DEBUG_NET, "\nCould not allocate SNP_DRIVER structure.\n"));
     goto NiiError;
   }
 
-  Snp = (SNP_DRIVER *)(UINTN)Address;
-
   ZeroMem (Snp, sizeof (SNP_DRIVER));
```
mdkinney: Can combine with above by using AllocateZeroPool().

NetworkPkg/SnpDxe/Snp.c
```diff
@@ -453,18 +445,50 @@ SimpleNetworkDriverStart (
                     PciIo,
                     AllocateAnyPages,
                     EfiBootServicesData,
-                    SNP_MEM_PAGES (4096),
-                    &Address,
+                    SNP_MEM_PAGES (EFI_PAGE_SIZE),
+                    &CpbAddress,
```
mdkinney: Since PCI I/O AllocateBuffer() returns a VOID*, can't &Snp->Cpb be used directly and remove the CpbAddress local?

NetworkPkg/SnpDxe/Snp.c
```diff
@@ -453,18 +445,50 @@ SimpleNetworkDriverStart (
...
+    DEBUG ((DEBUG_NET, "\nCould not allocate CPB structure.\n"));
+    goto Error_DeleteSNP;
+  }
+
+  Status = PciIo->AllocateBuffer (
+                    PciIo,
+                    AllocateAnyPages,
+                    EfiBootServicesData,
+                    SNP_MEM_PAGES (EFI_PAGE_SIZE),
+                    &DbAddress,
```
mdkinney: Same comment to remove local

NetworkPkg/SnpDxe/Snp.c
```diff
@@ -453,18 +445,50 @@ SimpleNetworkDriverStart (
...
+    DEBUG ((DEBUG_NET, "\nCould not allocate DB structure.\n"));
+    goto Error_DeleteSNP;
+  }
+
+  Status = PciIo->AllocateBuffer (
+                    PciIo,
+                    AllocateAnyPages,
+                    EfiBootServicesData,
+                    SNP_MEM_PAGES (sizeof (PXE_CDB)),
+                    &CdbAddress,
```
mdkinney: Same comment to remove local

NetworkPkg/SnpDxe/Snp.c
```diff
@@ -453,18 +445,50 @@ SimpleNetworkDriverStart (
...
-    DEBUG ((DEBUG_NET, "\nCould not allocate CPB and DB structures.\n"));
+    DEBUG ((DEBUG_NET, "\nCould not allocate CDB structure.\n"));
     goto Error_DeleteSNP;
   }
 
-  Snp->Cpb = (VOID *)(UINTN)Address;
-  Snp->Db  = (VOID *)((UINTN)Address + 2048);
+  Snp->Cpb = (VOID *)(UINTN)CpbAddress;
+  Snp->Db  = (VOID *)(UINTN)DbAddress;
+  Snp->Cdb = (PXE_CDB *)(UINTN)CdbAddress;
```
mdkinney: Same comment to remove locals and remove these assignments

NetworkPkg/SnpDxe/Snp.c
```diff
@@ -453,18 +445,50 @@ SimpleNetworkDriverStart (
...
+  ZeroMem ((VOID *)(UINTN)CpbAddress, EFI_PAGE_SIZE);
+  ZeroMem ((VOID *)(UINTN)DbAddress, EFI_PAGE_SIZE);
+  ZeroMem ((VOID *)(UINTN)CdbAddress, sizeof (PXE_CDB));
```
mdkinney: Use Snp fields and remove unneeded typecasts

NetworkPkg/SnpDxe/Snp.c
```diff
@@ -812,15 +849,20 @@ SimpleNetworkDriverStop (
   PciIo = Snp->PciIo;
   PciIo->FreeBuffer (
            PciIo,
-           SNP_MEM_PAGES (4096),
+           SNP_MEM_PAGES (EFI_PAGE_SIZE),
            Snp->Cpb
```
mdkinney: This looks like a double free of Snp->Cbp can occur with calls added above

NetworkPkg/SnpDxe/Snp.c
```diff
@@ -812,15 +849,20 @@ SimpleNetworkDriverStop (
...
+           SNP_MEM_PAGES (EFI_PAGE_SIZE),
            Snp->Cpb
            );
-
   PciIo->FreeBuffer (
            PciIo,
-           SNP_MEM_PAGES (sizeof (SNP_DRIVER)),
-           Snp
+           SNP_MEM_PAGES (EFI_PAGE_SIZE),
+           Snp->Db
```
mdkinney: This looks like a double free of Snp->Db can occur with calls added above

NetworkPkg/SnpDxe/Snp.c
```diff
@@ -812,15 +849,20 @@ SimpleNetworkDriverStop (
...
+           );
+  PciIo->FreeBuffer (
+           PciIo,
+           SNP_MEM_PAGES (sizeof (PXE_CDB)),
+           Snp->Cdb
```
mdkinney: This looks like a double free of Snp->Dcb can occur with calls added above

NetworkPkg/SnpDxe/Snp.c
eeshanl: done

NetworkPkg/SnpDxe/Snp.c
eeshanl: removed all locals.

NetworkPkg/SnpDxe/Snp.c
eeshanl: removed all locals

NetworkPkg/SnpDxe/Snp.c
eeshanl: removed this local, but since the SNP_DRIVER definition has PXE_CDB explicitily, it would make the deferences to this struct more ugly as it would require typecasts before de reference. As a result, i have just casted it to (VOID **) here.

NetworkPkg/SnpDxe/Snp.c
eeshanl: removed locals and removed these assignments.

NetworkPkg/SnpDxe/Snp.c
eeshanl: done

NetworkPkg/SnpDxe/Snp.c
eeshanl: added if check for NULL pointer before free.

NetworkPkg/SnpDxe/Snp.c
eeshanl: added if check for NULL pointer before free.

NetworkPkg/SnpDxe/Snp.c
eeshanl: added if check for NULL pointer before free.

NetworkPkg/SnpDxe/Snp.c
```diff
@@ -810,17 +839,31 @@ SimpleNetworkDriverStop (
...
+
+  if (Snp->Cdb != NULL) {
+    PciIo->FreeBuffer (
+             PciIo,
+             SNP_MEM_PAGES (sizeof (PXE_CDB)),
+             (VOID *)(UINTN)Snp->Cdb
+             );
+  }
+
+  FreePool (Snp);
```
mdkinney: Potential double free of Snp here.  The action in NiiError and Error_DeleteSNP are almost the same.  They should be combined to remove duplicate actions.

eeshanl: Would removing NiiError be the best approach here? We goto NiiError when failing before any of the Allocate calls occur, thus skipping all of the Free's. If we combine these 2, then we will have to wrap the entire free section with a if (Snp != NULL) . Seems these 2 goto's already account for appropriately freeing the structures.

Also in the SimpleNetworkDriverStop function, Snp is dereferenced many times already without a NULL check. This would be a problem if SimpleNetworkDriverStop gets called after SimpleNetworkDriverStart fails due to an allocation failure with the SNP_DRIVER Snp structure.

mdkinney: It is a normal pattern to for drivers in error path to wrap actions around the structure like Snp being NULL.

The Start() function has to succeed before the Stop() function can be called at a later time.  These constraints are listed in the UEFI Specification.  If you think there are issues in the Stop() function, then I suggest that you enter a different issue for that so we can focus on completing this PR first.

eeshanl: Updated to add NULL checks and removed NiiError and combined with Error_DeleteSNP.

NetworkPkg/SnpDxe/Snp.c
ardbiesheuvel: So why did you change this to 3 separate PciIo->AllocateBuffer() calls?

eeshanl: Hi @ardbiesheuvel from the previous discussion, I was under the impression that the original way of allocating the buffer and then doing pointer arithmetic to assign the structure addresses was very hacky. This solution allows allocation and free of each of the individual SNP_DRIVER sub-structures in a clean manner.

eeshanl:
Hi @mdkinney updated the PR per your comments, could you please re-review? Also, there is no NetworkPkg maintainer, however @SaloniKasbekar and @Zclarkwilliams, if you want to review, please do.

eeshanl:
Hi @mdkinney @ardbiesheuvel @SaloniKasbekar @Zclarkwilliams please take a look for review as this has been open for a few weeks now and would like to get it in for the stable tag. Thanks

lgao4:
> Hi @mdkinney @ardbiesheuvel @SaloniKasbekar @Zclarkwilliams please take a look for review as this has been open for a few weeks now and would like to get it in for the stable tag. Thanks

The change is clear. It looks good to me.

lgao4:
@ardbiesheuvel , @mdkinney , @leiflindholm , @ajfish , @makubacki , this patch requests to catch the stable tag 202508. This is a clear bug fix. I am OK to merge it. If you have comments, please raise.

mdkinney:
A few remaining items that need to be address for this PR to be merged:

1) The goto label `Error_DeleteSNP` does not follow naming conventions.  Recommend `ErrorDeleteSnp'
2) The error path that frees the allocated buffers duplicates the logic to free the DMA capable buffers.  The error path logic should be simplified so that logic is only present one time
3) The NetworkPkg reviewers @SaloniKasbekar and @Zclarkwilliams have not reviewed this PR.  These reviews are required before the stewards acting as maintainers for the NetworkPkg can approve

I recommend delaying merging this PR until after edk2-stable202508.

os-d:
> A few remaining items that need to be address for this PR to be merged:
> 
> 1. The goto label `Error_DeleteSNP` does not follow naming conventions.  Recommend `ErrorDeleteSnp'
> 2. The error path that frees the allocated buffers duplicates the logic to free the DMA capable buffers.  The error path logic should be simplified so that logic is only present one time
> 3. The NetworkPkg reviewers @SaloniKasbekar and @Zclarkwilliams have not reviewed this PR.  These reviews are required before the stewards acting as maintainers for the NetworkPkg can approve
> 
> I recommend delaying merging this PR until after edk2-stable202508.

I don't have a problem with delaying this until after the stable tag or points 1 and 2, but for point 3, @eeshanl has done tried to get review from @SaloniKasbekar and @Zclarkwilliams , but both have been unresponsive. 

This PR was opened on June 5, no reply from the reviewers. Then:

image

No reply.

image

No reply.

image

No reply.

It would be great if they would review, but at a certain point I think the stewards need to merge this (especially as both you and Ard have already reviewed). Eeshan can continue to give them weekly pings until after the stable tag, but I would propose if they haven't reviewed by then that we need help from a steward.

mdkinney:
I agree on need for prompt reviews from the NetworkPkg reviews.  I will work on that aspect.  If they are no longer able to perform the review responsibilities, then this package would fall into the unmaintained state and the stewards would be responsible for the reviews.

eeshanl:
> A few remaining items that need to be address for this PR to be merged:
> 
> 1. The goto label `Error_DeleteSNP` does not follow naming conventions.  Recommend `ErrorDeleteSnp'
> 2. The error path that frees the allocated buffers duplicates the logic to free the DMA capable buffers.  The error path logic should be simplified so that logic is only present one time
> 3. The NetworkPkg reviewers @SaloniKasbekar and @Zclarkwilliams have not reviewed this PR.  These reviews are required before the stewards acting as maintainers for the NetworkPkg can approve
> 
> I recommend delaying merging this PR until after edk2-stable202508.

Fixed and updated the PR for both points 1 and 2. Switched to `ErrorDeleteSnp' and created a 'SimpleNetworkDriverFree' function that gets called on error in 'SimpleNetworkDriverStart' and is also called in 'SimpleNetworkDriverStop'.

Zclarkwilliams (changes requested):
Since this change modifies Cdb from a structure member to a pointer (for UEFI spec alignment), we should implement defensive programming by adding NULL pointer checks in all functions that access Snp->Cdb.

eeshanl:
> Since this change modifies Cdb from a structure member to a pointer (for UEFI spec alignment), we should implement defensive programming by adding NULL pointer checks in all functions that access Snp->Cdb.

Added NULL checks on Cdb in every function that accesses Cdb-> within NetworkPkg/SnpDxe

Zclarkwilliams (approved):

NetworkPkg/SnpDxe/Snp.c
mdkinney: @ardbiesheuvel Is this topic resolved now?

ardbiesheuvel: I still don't understand why we need 3 separate PciIo->AllocateBuffer() calls to allocate 3 DMA'able regions that are all logically attached to the same device/driver instance, and can just be carved out of a single allocation. It makes the logic much simpler, and if the pointer arithmetic is considered to be an issue (but given that this is C, I wonder why it would), we could define a struct with the 3 regions as fields (with the appropriate alignment) and take the addresses of those after allocating a single DMA'able region for the whole structure.

eeshanl: The reason for the three buffers is because two of the structures are dynamic in length. I can add two unions (with padding), one for cpb and one for db, then define the structure for the allocated page as one cdb, one cpb union, and one db union, then reference the fields like that. It seems simpler to waste a little memory and avoid confusion. Maybe I misunderstood, but it seemed this was an ok path forward from previous discussions in this PR. Would you prefer I change this?

mdkinney:
@ardbiesheuvel There are unresolved comments in this PR that need to be resolved before Mergify can merge.

PR #11162, opened 2025-06-06
NetworkPkg: Add PCD to control http boot enable or disable.

mark-skum:
Add PCD to control http boot enable or disable follow PXE boot.

### DIFF
--- a/MdeModulePkg/Universal/HiiDatabaseDxe/Database.c
+++ b/MdeModulePkg/Universal/HiiDatabaseDxe/Database.c
@@ -687,6 +687,11 @@ FindQuestionDefaultSetting (
   PCD_DATA_DELTA                 *DeltaData;
   BOOLEAN                        VarCheck;
 
+  if (EfiVarStore == NULL) {
+    DEBUG ((DEBUG_ERROR, "EfiVarStore is null\n"));
+    return EFI_NOT_FOUND;
+  }
+
   if (gSkuId == 0xFFFFFFFFFFFFFFFF) {
     gSkuId = LibPcdGetSku ();
   }


NetworkPkg/HttpBootDxe/HttpBootDxe.c
```diff
@@ -1296,6 +1304,10 @@ HttpBootDxeDriverEntryPoint (
 {
   EFI_STATUS  Status;
 
+  if ((PcdGet8 (PcdIPv4HttpSupport) == HTTP_DISABLED) && (PcdGet8 (PcdIPv6HttpSupport) == HTTP_DISABLED)) {
```
xpahos: Would it make sense to consider using `PcdGetBool` here?

Also, just to clarify — what behavior is expected from `HttpBootDxeDriverEntryPoint` if only one of these parameters is disabled? Do we actually need to call `EfiLibInstallDriverBindingComponentName2` for both protocols in that case?

NetworkPkg/HttpBootDxe/HttpBootDxe.c
```diff
@@ -305,6 +305,10 @@ HttpBootIp4DxeDriverBindingSupported (
 {
   EFI_STATUS  Status;
 
+  if (PcdGet8 (PcdIPv4HttpSupport) == HTTP_DISABLED) {
```
xpahos: Is this check really necessary here if the protocol wasn't called for initialization?

mark-skum:
@leiflindholm I have resubmitted a PR. please help to review. thanks

NetworkPkg/HttpBootDxe/HttpBootDxe.c
mark-skum: Follow the policy of PXE Boot

NetworkPkg/HttpBootDxe/HttpBootDxe.c
mark-skum: it's necessary for disable HttpBootIPv4 or HttpBootIPv6.

NetworkPkg/HttpBootDxe/HttpBootDxe.c
xpahos: Yes, I understand that this is copied from PXE. But perhaps it would be better to rework both places. `CoreConnectSingleController` has a list of installed protocols, and every time DXE connects a new driver, it results in a new call to the `Supported` method. This code is inconsistent:

1. Both protocols are disabled: there is no need to add an additional check for `HttpBootIpXDxeDriverBindingSupported`, since the protocols are not installed.
2. One of the protocols is disabled: it is necessary to check every time `CoreConnectSingleController` is called.

This seems strange to me.

mark-skum: I'm resubmitting a PR to remove this.
if ((PcdGet8 (PcdIPv4HttpSupport) == HTTP_DISABLED) && (PcdGet8 (PcdIPv6HttpSupport) == HTTP_DISABLED))
and use PcdGetBool

PR #11080, opened 2025-05-14
Adding FF-A memory management library

kuqin12:
# Description

This change adds the FF-A memory management protocol implementation according to https://developer.arm.com/documentation/den0140/latest/

- [ ] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [ ] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

## How This Was Tested

This change was tested on QEMU SBSA and proprietary hardware with Hafnium as the SPMC.

## Integration Instructions

Add `ArmFfaMemMgmtLib|MdePkg/Library/ArmFfaMemMgmtLib/ArmFfaMemMgmtLib.inf` to platform DSC.

PR #11166, opened 2025-06-06
MdeModulePkg/Library: Fix memory leak for ArmFfaLib's Rx/Tx buffer and add ArmFfaLib for PeilessSec

LeviYeoReum:
# Description

This patchset adds:
    - Fix memory leak on Rx/Tx buffer in ArmFfaPeiLib.
    - Add ArmFfaSecLib for PeilessSec.

Patch 1 fixes the memory leak on RxTx buffer when ArmFfaPeiLib used.
Patch 2 makes PcdFixedAtBuildPcd as PcdsFixedAtBuildPCD.
Patch 3 adds ArmFfaSecLib used in PeilessSec
Patch 4 cleanup duplication code on ArmFfaLib.
Patch 5 change type of buffer address in ArmFFaRxTxBufferInfo structure.


MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
```diff
@@ -176,17 +138,54 @@ ArmFfaPeiLibConstructor (
...
+     * This Rx/Tx buffer is mapped by ArmFfaSecLib.
+     */
+    RxTxBufferAllocationHob = FindRxTxBufferAllocationHob (TRUE);
+
+    /*
+     * Below case Rx/Tx buffer mapped by ArmPeiLib but in temporary memory.
+     */
+    if (RxTxBufferAllocationHob == NULL) {
+      RxTxBufferAllocationHob = FindRxTxBufferAllocationHob (FALSE);
+      ASSERT (RxTxBufferAllocationHob != NULL);
```
samimujawar, line 170: Please handle the error when RxTxBufferAllocationHob == NULL.

LeviYeoReum:
@kuqin12, @samimujawar 
Could you review this please?

Thanks!

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
```diff
@@ -176,17 +138,54 @@ ArmFfaPeiLibConstructor (
...
+        &RxTxBufferAllocationHob->AllocDescriptor.Name,
+        &gArmFfaRxTxBufferInfoGuid
+        );
     }
+
+    Status = (*PeiServices)->NotifyPpi (PeiServices, &mNotifyOnPeiMemoryDiscovered);
+    ASSERT_EFI_ERROR (Status);
+
+    /*
+     * Change RemapRequired to here to prevent other PEIM from registering
```
kuqin12:
```suggestion
     * Change RemapRequired to FALSE here to prevent other PEIM from registering
```

MdeModulePkg/Library/ArmFfaLib/ArmFfaRxTxMap.c
```diff
@@ -270,3 +303,127 @@ ArmFfaLibRxTxUnmap (
...
+
+  RxTxBufferAllocationHob = FindRxTxBufferAllocationHob (TRUE);
+  if (RxTxBufferAllocationHob == NULL) {
+    return EFI_NOT_FOUND;
+  }
+
+  Status = GetRxTxBufferMinSizeAndAlign (&MinSizeAndAlign);
+
+  ZeroMem (&FfaArgs, sizeof (ARM_FFA_ARGS));
+  FfaArgs.Arg0 = ARM_FID_FFA_RXTX_UNMAP;
```
kuqin12, line 295: Would you do unmap -> map routine using the corresponding FFA primitive functions rather than directly bang on the FFA calls?

LeviYeoReum: It couldn't the Remap function's point is "not allocate new memory" but use "already allocated one"
current ArmRxTx(Un)Map function is including the "allocate & free" memory.
Unfortunately, We couldn't use the ArmFfaLib primitive function.

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
LeviYeoReum: > FALSE

Thanks. I'll add it :)

LeviYeoReum:
@kuqin12 @samimujawar 
Gentle ping in case of forgotten.

kuqin12:
@LeviYeoReum I was thinking about this change in general. Would it make sense to let SEC phase to handle the memory movement? So that DXE phase does not need to know about this? I think leaving the movement to DXE phase and perform operations based on a flag in data is like logic leakage.

Or is there any reason that it cannot be done in SEC phase? In our use case, by the time SEC hands off to DXE, the permanent memory is already available.

LeviYeoReum:
Hi Kun, 

Thanks for reviewing this patch :) 

> @LeviYeoReum I was thinking about this change in general. Would it make sense to let SEC phase to handle the memory movement? So that DXE phase does not need to know about this? I think leaving the movement to DXE phase and perform operations based on a flag in data is like logic leakage.
> 
> Or is there any reason that it cannot be done in SEC phase? In our use case, by the time SEC hands off to DXE, the permanent memory is already available.

Actually for PeilessSec as you said, this is true. it don't need to do memory movement.
However, what I thought is when SEC + PEI phase is enabled and SEC allocates the Rx/Tx buffer.
In this sitaution, SEC couldn't use the permanent memory but it could be used by PEI second phase.
But, Suppose no PEIM uses the ArmFfaLib.
Then the Rx/Tx buffer allocated by SEC still allocated in temporary memory, 
so If the DXE doesn't handle the memory movement, It wouldn't use the Rx/Tx buffer properly.

Thanks.

kuqin12:
> Hi Kun,
> 
> Thanks for reviewing this patch :)
> 
> > @LeviYeoReum I was thinking about this change in general. Would it make sense to let SEC phase to handle the memory movement? So that DXE phase does not need to know about this? I think leaving the movement to DXE phase and perform operations based on a flag in data is like logic leakage.
> > Or is there any reason that it cannot be done in SEC phase? In our use case, by the time SEC hands off to DXE, the permanent memory is already available.
> 
> Actually for PeilessSec as you said, this is true. it don't need to do memory movement. However, what I thought is when SEC + PEI phase is enabled and SEC allocates the Rx/Tx buffer. In this sitaution, SEC couldn't use the permanent memory but it could be used by PEI second phase. But, Suppose no PEIM uses the ArmFfaLib. Then the Rx/Tx buffer allocated by SEC still allocated in temporary memory, so If the DXE doesn't handle the memory movement, It wouldn't use the Rx/Tx buffer properly.
> 
> Thanks.

I see. Please correct me if I misunderstood, we mainly have 2 scenarios:

| Boot Process | Notes |
|-|-|
| SEC -> DXE | No memory movement |
| SEC -> PEI -> DXE | Need movement in PEI |

The concern is mainly around when no PEIM uses FFA, correct? Wondering if you have considered using a separate PEIM to handle the movement? We are currently imposing all FFA consuming PEIMs to register a callback, is that necessary?

A less-than-optimal, but simpler solution is to remap the buffer regardless in DXE phase. This is trading the complicated logic of carrying on the information with a little overhead of remapping the buffer, unless you can think of any reason why DXE has to use the same Rx/Tx buffer as PEI.

LeviYeoReum:
Hi Kun,

> 
> I see. Please correct me if I misunderstood, we mainly have 2 scenarios:
> 
> Boot Process	Notes
> SEC -> DXE	No memory movement

This case is PeilessSec. So Yes.


> SEC -> PEI -> DXE	Need movement in PEI

Here is tow case.
    - In PEI phase no one use ArmFfaPeiLib so, movement occurs in PEI so Dxe do it.
    -  PEI do memory movement.

> The concern is mainly around when no PEIM uses FFA, correct? 

Yes.

> Wondering if you have considered using a separate PEIM to handle the movement? We are currently imposing all FFA >consuming PEIMs to register a callback, is that necessary?

Yes. if PEIM use ArmFfaPeiLib. it's done by ArmFfaLib.
But in case of doing no PEIM using ArmFfaPeiLib, makes a PEIM seems too much.

> A less-than-optimal, but simpler solution is to remap the buffer regardless in DXE phase. This is trading the complicated logic of carrying on the information with a little overhead of remapping the buffer, unless you can think of any reason why DXE has to use the same Rx/Tx buffer as PEI.

Hmm.. This seems not good for me.
If Rx/Tx buffer is allocated in ealry PEI phase (before permenant memory use), It's allocated as "Reserved Memory".
But If SPMC requires minimum alignment as 64KB. It's too big to take it as overhead (That's why I think it's a "LEAK").
As you said, There's no reason we must use "the same Rx/Tx Buffer". But we care the "LEAK".
Also, access the Temporary memory after PEIM can use permanent memory seems the UAF. 

Thanks ;)

kuqin12:
> > Wondering if you have considered using a separate PEIM to handle the movement? We are currently imposing all FFA >consuming PEIMs to register a callback, is that necessary?
> 
> Yes. if PEIM use ArmFfaPeiLib. it's done by ArmFfaLib. But in case of doing no PEIM using ArmFfaPeiLib, makes a PEIM seems too much.
> 
Is a PEIM really considered too much? I think it might be better than sprinkle the callbacks around all PEIMs that _do_ use FFA?
>
> > A less-than-optimal, but simpler solution is to remap the buffer regardless in DXE phase. This is trading the complicated logic of carrying on the information with a little overhead of remapping the buffer, unless you can think of any reason why DXE has to use the same Rx/Tx buffer as PEI.
> 
> Hmm.. This seems not good for me. If Rx/Tx buffer is allocated in ealry PEI phase (before permenant memory use), It's allocated as "Reserved Memory". But If SPMC requires minimum alignment as 64KB. It's too big to take it as overhead (That's why I think it's a "LEAK"). As you said, There's no reason we must use "the same Rx/Tx Buffer". But we care the "LEAK". Also, access the Temporary memory after PEIM can use permanent memory seems the UAF.
> 
Sorry, maybe I missed your point of "LEAK". But even if we stick to the same buffer throughout the boot phase, OS/hypervisor will remap their Rx/Tx buffer anyway. Would that not be a leakage?

LeviYeoReum:
Hi Kun, 

> Is a PEIM really considered too much? I think it might be better than sprinkle the callbacks around all PEIMs that _do_ use FFA?

Hmm, the "Remap" is done only once. I don't think the callback is sprinkled by all PEIM?
Secondly, though Rx/Tx buffer is allocated by Library but remaping is done by PEIM seems much srpinkle for me since
the FF-A user for PEI, it always add one more line in there .dsc file to use ArmFfaPeiLib..

> > Hmm.. This seems not good for me. If Rx/Tx buffer is allocated in ealry PEI phase (before permenant memory use), It's allocated as "Reserved Memory". But If SPMC requires minimum alignment as 64KB. It's too big to take it as overhead (That's why I think it's a "LEAK"). As you said, There's no reason we must use "the same Rx/Tx Buffer". But we care the "LEAK". Also, access the Temporary memory after PEIM can use permanent memory seems the UAF.
> 
> Sorry, maybe I missed your point of "LEAK". But even if we stick to the same buffer throughout the boot phase, OS/hypervisor will remap their Rx/Tx buffer anyway. Would that not be a leakage?

No. When you see the before this patch, we skip the "FreeMemory()" in case of PEIM allocate Rx/Tx buffer 
(in PeiServiceMemoryDiscoveredNotifyCallback()).
This means if Rx/Tx Buffer is allocated by PEIM, the former Rx/Tx buffer is still allocated as a "Reserved Memory"
after PEI intialized the permenat memory (the allocated at temporary memory wiill be mirgrated in peremnant memory to by calculating offest) -- for detail, Please see my commit message:
https://github.com/tianocore/edk2/pull/11166/commits/a5d19c5d106a5b02c50f8680a8fa1b57c781e2cf

So, this unused Rx/Tx buffer memory area still reserved in OS/Hypervisor, It's a leak.

Thanks :)

kuqin12:
> Hmm, the "Remap" is done only once. I don't think the callback is sprinkled by all PEIM? Secondly, though Rx/Tx buffer is allocated by Library but remaping is done by PEIM seems much srpinkle for me....
> 
We are imposing all FFA aware PEI modules to carry the check of permanent memory availability. Additionally, the callbacks are registered for each of the PEIMs that will use FFA lib in early PEI phase, correct? I think this will increase the footprint on flash for each involved module. And early PEIMs normally do not have compression.
> 
> No. When you see the before this patch, we skip the "FreeMemory()" in case of PEIM allocate Rx/Tx buffer (in PeiServiceMemoryDiscoveredNotifyCallback()). This means if Rx/Tx Buffer is allocated by PEIM, the former Rx/Tx buffer is still allocated as a "Reserved Memory" after PEI intialized the permenat memory (the allocated at temporary memory wiill be allocated in peremnant memory to by calculating offest).
> 
> So, this unused Rx/Tx buffer memory area still reserved in OS/Hypervisor, It's a leak.
> 
Are you confirming that, if OS/Hypervisor will not use the same Rx/Tx buffer region, they will "leak" the memory region because it is marked as reserved in UEFI, even after this change?

LeviYeoReum:
Hi Kun,

> We are imposing all FFA aware PEI modules to carry the check of permanent memory availability. Additionally, the callbacks are registered for each of the PEIMs that will use FFA lib in early PEI phase, correct? I think this will increase the footprint on flash for each involved module. And early PEIMs normally do not have compression.

No, callback is registered by the first PEIM who allocates the Rx/Tx buffer if Rx/Tx buffer are allocated in Temporary memory.
So, It doesn't impose entry PEIM module to take of it.
I afraid of whether I understand your concern correct or not, But for me separate handling for Remap (migrate) in separate PEIM module makes user mistake easily who want to use ArmFfaPeiLib. 
May I ask what is benefit to have separate PEIM to handle Remap only for Rx/Tx buffer taking over my worry?
The increase size for this patch for code area seems ignorable I think compared to have separate PEIM module for this (Including basic library for ths PEIM and etc..?).


> Are you confirming that, if OS/Hypervisor will not use the same Rx/Tx buffer region, they will "leak" the memory region because it is marked as reserved in UEFI, even after this change?

Before this patch, Yes.

1. Allocate Rx/Tx buffer in temporary memory.
2. while Rx/Tx buffer is mirgrated after permanent my use, the Rx/Tx buffer allocated in temporary memory is migrated too.
3. at the Callback for PEIM memory, It allocates New "Rx/Tx" buffer, So (2) buffer is leaked and still remain as "Reserved"
    (NOTE: It saves this information in the "HOB", so Dxe reserves this memory as "Reserved Memory")

After this Patch, No.
Since:
    1. Rx/Tx buffer is remapped with migrated address, in this progress, there is no allocation.
    2. At the ExitBootService, this area is freed via FreeMemory() this means the area is remarked as Conventional memory.
        So no memory leak for this.

Thanks!

kuqin12:
> Before this patch, Yes.
> 
> 1. Allocate Rx/Tx buffer in temporary memory.
> 2. while Rx/Tx buffer is mirgrated after permanent my use, the Rx/Tx buffer allocated in temporary memory is migrated too.
> 3. at the Callback for PEIM memory, It allocates New "Rx/Tx" buffer, So (2) buffer is leaked and still remain as "Reserved"
>    (NOTE: It saves this information in the "HOB", so Dxe reserves this memory as "Reserved Memory")
> 
> After this Patch, No. Since: 1. Rx/Tx buffer is remapped with migrated address, in this progress, there is no allocation. 2. At the ExitBootService, this area is freed via FreeMemory() this means the area is remarked as Conventional memory. So no memory leak for this.
> 
Great. The migration part makes sense.

We have been all over the places during the conversation. My overall point was that I hope the PEI being the only one that needs to handle the logic. A separate PEIM will cover the edge case if no PEIM uses FFA. The single module probably will save the footprint more scalablely. But you can decide how to proceed on that point.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -0,0 +1,334 @@
...
+EFI_STATUS
+EFIAPI
+ArmFfaLibGetRxTxBuffers (
+  OUT VOID    **TxBuffer OPTIONAL,
+  OUT UINT64  *TxBufferSize OPTIONAL,
+  OUT VOID    **RxBuffer OPTIONAL,
+  OUT UINT64  *RxBufferSize OPTIONAL
+  )
+{
+  if ((mTxBuffer == NULL) || (mRxBuffer == 0x00)) {
```
kuqin12: Why wouldn't these 2 be the same style?

MdeModulePkg/Include/Guid/ArmFfaRxTxBufferInfo.h
```diff
@@ -14,16 +14,22 @@
  */
 typedef struct ArmFfaRxTxBuffersInfo {
   /// Tx Buffer Address.
-  VOID      *TxBufferAddr;
+  VOID       *TxBufferAddr;
```
kuqin12: Since we are touching these fields, I think it would make more sense to change them over to EFI_PHYSICAL_ADDRESS. It will save a lot hassle below that involves double pointers.

LeviYeoReum:
> > Before this patch, Yes.
> > 
> > 1. Allocate Rx/Tx buffer in temporary memory.
> > 2. while Rx/Tx buffer is mirgrated after permanent my use, the Rx/Tx buffer allocated in temporary memory is migrated too.
> > 3. at the Callback for PEIM memory, It allocates New "Rx/Tx" buffer, So (2) buffer is leaked and still remain as "Reserved"
> >    (NOTE: It saves this information in the "HOB", so Dxe reserves this memory as "Reserved Memory")
> > 
> > After this Patch, No. Since: 1. Rx/Tx buffer is remapped with migrated address, in this progress, there is no allocation. 2. At the ExitBootService, this area is freed via FreeMemory() this means the area is remarked as Conventional memory. So no memory leak for this.
> 
> Great. The migration part makes sense.

Thanks ;)

> 
> We have been all over the places during the conversation. My overall point was that I hope the PEI being the only one that needs to handle the logic. A separate PEIM will cover the edge case if no PEIM uses FFA. The single module probably will save the footprint more scalablely. But you can decide how to proceed on that point

I understood what is your point. But, IMHO, I'm not sure the add PEIM really saves the footprint taking the duty to add the PEIM on every platform which want to use ArmFfaPeiLib and a risk -- not add the the PEIM for handling this.

Also, I'm not sure about the feature "edge case" like this problem -- handling something over the Phases.
if this happen in near future, I think It would be better to add at that time since for handling single problem to add one more PEIM seems to increase footprint for me.

If you agree, I want to remain as it is right now.

Thanks.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
LeviYeoReum: Thanks. I'll fix it :)

MdeModulePkg/Include/Guid/ArmFfaRxTxBufferInfo.h
LeviYeoReum: Cool. Let me cange it.

Thanks!

MdeModulePkg/Include/Guid/ArmFfaRxTxBufferInfo.h
```diff
@@ -9,21 +9,29 @@
...
 
   /// Rx Buffer Size.
-  UINT64    RxBufferSize;
+  UINT64                  RxBufferSize;
+
+  /// Rx/Tx buffer should be remapped to permanent memory.
+  BOOLEAN                 RemapRequired;
+
+  /// Rx/Tx buffer offset from its allocation base.
+  UINTN                   RemapOffset;
```
kuqin12: nit: Would you want to move this to UINT64? I do not think we need the natural width for this field?

MdeModulePkg/Library/ArmFfaLib/ArmFfaStandaloneMmRxTxMap.c
```diff
@@ -1,18 +1,17 @@
...
 
-  Copyright (c) 2024, Arm Limited. All rights reserved.<BR>
+  Copyright (c) 2024-2025, Arm Limited. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
    @par Glossary:
      - FF-A - Firmware Framework for Arm A-profile
 
    @par Reference(s):
-     - Arm Firmware Framework for Arm A-Profile v1.3 ALP1: [https://developer.arm.com/documentation/den0077/l]
```
kuqin12, line 11: Can you please use the specific version of the spec instead? I agree with Leif and Sami's suggestion here: https://github.com/tianocore/edk2/pull/11080#discussion_r2131666812. The implementation can only stick to a specific version it was written against.

MdeModulePkg/Include/Guid/ArmFfaRxTxBufferInfo.h
LeviYeoReum: Thanks. I'll fix it.

MdeModulePkg/Library/ArmFfaLib/ArmFfaStandaloneMmRxTxMap.c
LeviYeoReum: Thanks. I'll fix it.

kuqin12 (approved):

LeviYeoReum:
Hi @samerhaj  and @lgao4,

If there is no problem, Would you merge this patch series please?

Thanks.

lgao4:
> Hi @samerhaj and @lgao4,
> 
> If there is no problem, Would you merge this patch series please?
> 
> Thanks.

If no more comments, I will merge it next week.

MdeModulePkg/MdeModulePkg.dec
```diff
@@ -1218,6 +1218,10 @@
   # @Prompt Defines the page allocation for the MM communication buffer; default is 128 pages (512KB).
   gEfiMdeModulePkgTokenSpaceGuid.PcdMmCommBufferPages|128|UINT32|0x30001061
 
+  ## This dynamic PCD holds the number of pages for the FFA TX/RX buffer.
```
samimujawar: Please fix the comment, it still says dynamic PCD.

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
```diff
@@ -176,17 +138,54 @@ ArmFfaPeiLibConstructor (
...
+                RxTxBufferAllocationHob->AllocDescriptor.MemoryBaseAddress);
+
+      CopyGuid (
+        &RxTxBufferAllocationHob->AllocDescriptor.Name,
+        &gArmFfaRxTxBufferInfoGuid
+        );
     }
+
+    Status = (*PeiServices)->NotifyPpi (PeiServices, &mNotifyOnPeiMemoryDiscovered);
+    ASSERT_EFI_ERROR (Status);
```
samimujawar: Return appropriate error code, please.

MdeModulePkg/Library/ArmFfaLib/ArmFfaRxTxMap.c
```diff
@@ -270,3 +273,197 @@ ArmFfaLibRxTxUnmap (
...
+  BufferInfo->TxBufferAddr = (VOID *)(UINTN)PcdGet64 (PcdFfaTxBuffer);
+  BufferInfo->TxBufferSize = PcdGet64 (PcdFfaTxRxPageCount) * EFI_PAGE_SIZE;
+  BufferInfo->RxBufferAddr = (VOID *)(UINTN)PcdGet64 (PcdFfaRxBuffer);
+  BufferInfo->RxBufferSize = PcdGet64 (PcdFfaTxRxPageCount) * EFI_PAGE_SIZE;
+}
+
+/**
+  Find Rx/TX buffer memory allocation hob.
+
+  @param  UseGuid             Find MemoryAllocationHob using Guid.
```
samimujawar: @param  UseGuid             Find MemoryAllocationHob using gArmFfaRxTxBufferInfoGuid.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.c
```diff
@@ -0,0 +1,107 @@
...
+                 sizeof (ARM_FFA_RX_TX_BUFFER_INFO)
+                 );
+  if (BufferInfo == NULL) {
+    DEBUG ((DEBUG_ERROR, "%a: Failed to create Rx/Tx Buffer Info Hob\n", __func__));
+    ArmFfaLibRxTxUnmap ();
+    return EFI_OUT_OF_RESOURCES;
+  }
+
+  RxTxBufferAllocationHob = FindRxTxBufferAllocationHob (FALSE);
+  ASSERT (RxTxBufferAllocationHob != NULL);
```
samimujawar, line 89: Please handle the error condition when RxTxBufferAllocationHob == NULL.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -0,0 +1,368 @@
...
+  UINTN         Property1;
+  UINTN         Property2;
+  UINTN         MinSizeAndAlign;
+  UINTN         MaxSize;
+  VOID          *Buffers;
+  VOID          *TxBuffer;
+  VOID          *RxBuffer;
+  UINT64        BufferSize;
+
+  BufferSize = PcdGet64 (PcdFfaTxRxPageCount) * EFI_PAGE_SIZE;
```
samimujawar: I think we can get rid of BufferSize and update line 185 with (PcdGet64 (PcdFfaTxRxPageCount) * EFI_PAGE_SIZE)

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -0,0 +1,368 @@
...
+  }
+
+  mTxBuffer = TxBuffer;
+  mRxBuffer = RxBuffer;
+
+  return EFI_SUCCESS;
+
+ErrorHandler:
+  FreePages (Buffers, (PcdGet64 (PcdFfaTxRxPageCount) * 2));
+  TxBuffer = NULL;
```
samimujawar: Line 218 and 219 are probably not needed.

MdeModulePkg/Library/ArmFfaLib/ArmFfaStandaloneMmRxTxMap.c
```diff
@@ -197,6 +142,7 @@ ArmFfaLibRxTxMap (
   TxBuffer   = Buffers;
   RxBuffer   = Buffers + BufferSize;
 
+  ZeroMem (&FfaArgs, sizeof (ARM_FFA_ARGS));
```
samimujawar, line 145: This should be part of the original change.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -752,3 +755,146 @@ ArmFfaLibCommonInit (
 
   return EFI_SUCCESS;
 }
+
```
samimujawar, line 758: I think you should reorder your patches so that this patch introduces the common functionality which is then used in subsequent patches.

MdeModulePkg/Library/ArmFfaLib/ArmFfaRxTxMap.c
```diff
@@ -189,6 +132,7 @@ ArmFfaLibRxTxMap (
   TxBuffer = Buffers;
   RxBuffer = Buffers + BufferSize;
 
+  ZeroMem (&FfaArgs, sizeof (ARM_FFA_ARGS));
```
samimujawar, line 135: This should be part of the original change. This has nothing to do with common functionality that this patch intends to provide.

MdeModulePkg/Include/Guid/ArmFfaRxTxBufferInfo.h
```diff
@@ -9,27 +9,29 @@
...
 
+#include <Uefi/UefiBaseType.h>
+
 /**
  * Guid Hob Data for gArmFfaRxTxBufferInfoGuid Guid Hob.
  */
 typedef struct ArmFfaRxTxBuffersInfo {
   /// Tx Buffer Address.
-  VOID       *TxBufferAddr;
+  EFI_PHYSICAL_ADDRESS    TxBufferAddr;
```
samimujawar, line 19: I think this change should be the first patch in this series.

LeviYeoReum: I don't this is matter of order.
It was a part of refector only...

MdeModulePkg/Library/ArmFfaLib/ArmFfaRxTxMap.c
LeviYeoReum: No, Please see the line 141 which is deleted.
The first one is for ArmFfaFeature() but it merged to GetRxTxBufferMinSizeAndAlign() so it removed.
this ZeroMem() for FF-A mapping operation. 
So this is part of fix.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: Nit:
before refactor something. bugfix should be first and make it refactor
otherwise, "refactor" code has the same bug!

MdeModulePkg/Library/ArmFfaLib/ArmFfaStandaloneMmRxTxMap.c
LeviYeoReum: As I mentioned In other comments No.
LINE 150 which is delated is for FF-A get feature.
This is for the ARM_FID_FFA_RXTX_MAP operation init.
So it should be here

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
LeviYeoReum: Okay. I'll remove. Thanks!

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.c
LeviYeoReum: It doesn't need! Since above gArmFfaRxTxBufferInfoGuid is created
This is for delvelopment one. if we add your check it becomes deadcode.
So, I don't want to add.

MdeModulePkg/Library/ArmFfaLib/ArmFfaRxTxMap.c
LeviYeoReum: Thanks. I'll change it.

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
LeviYeoReum: Thanks. I'll change it.

MdeModulePkg/MdeModulePkg.dec
LeviYeoReum: Thanks for catching this!

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
LeviYeoReum: As I said on the other comment. In this code, the "RxTxBufferAllocationHob" always should be found.
This is debug purpose and if we add check. That'll be "DEAD CODE"
So, I don't want to add.

Thanks.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
LeviYeoReum: Okay. I'll remove it.
Thanks

LeviYeoReum:
Hi @samimujawar  and @lgao4 

Could you merge this patch series please?

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
samimujawar: Looks like the comment update was lost. Can you check, please?

LeviYeoReum: Thanks. I've missed it.

LeviYeoReum:
> Looks like one minor comment change was lost. Otherwise this PR looks good to me.

Thanks. I've updated it !

PR #11173, opened 2025-06-10
edksetup.sh: Update "inux" to "Unix-like"

Damien-Chen:
*inux -> Unix-like

### DIFF
--- a/edksetup.sh
+++ b/edksetup.sh
@@ -3,7 +3,7 @@
 # Copyright (c) 2016, Linaro Ltd. All rights reserved.<BR>
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 #
-# In *inux environment, the build tools's source is required and need to be compiled
+# In Unix-like environment, the build tools's source is required and need to be compiled
 # firstly, please reference https://github.com/tianocore/tianocore.github.io/wiki/SourceForge-to-Github-Quick-Start
 # to get how to setup build tool.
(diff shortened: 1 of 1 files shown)


leiflindholm:
That's not really a typo, but more of a short-hand for unix-like system. Which would traditionally be written something more like *NIX. See https://en.wikipedia.org/wiki/Unix-like

*inux is a pretty poor form of that though, so I'd take a patch that changes it to a more common variant.

But please follow the instruction at https://github.com/tianocore/tianocore.github.io/wiki/Commit-Message-Format if you want to submit an updated patch.

Damien-Chen:
Ok, so maybe change it to 
*inux -> Unix-like 
would be better.

leiflindholm:
I'm happy with that, but could you squash the two commits into a single one please?

leiflindholm:
This now turned into three commits. I'm not asking you to merge additional changes from the main branch into the topic branch. I'm asking you to create a topic branch with only the total sum of the changes you are wanting to include in a single commit.

leiflindholm:
> This now turned into three commits. I'm not asking you to merge additional changes from the main branch into the topic branch. I'm asking you to create a topic branch with only the total sum of the changes you are wanting to include in a single commit.

I just tried to do that for you through the github UI, which may or may not have done the right thing.

leiflindholm (approved):

PR #11174, opened 2025-06-10
CloudHv: bump PcdCpuMaxLogicalProcessorNumber to 254

tpressure:
# Description

When using `CLOUDHV.fd` on cloud-hypervisor, only a total of 64 vCPUs can be used. This is because cloud-hypervisor does not (yet) implement a fw_cfg device that tells us the maximum number of logical processors and the default cloud-hypervisor config sets this value to 64. Thus, the firmware is only allocating stacks for 64 cpus. When more than 64 vcpus are in use, random vcpus crash due to stack collisions during bootup.

We should bump the default here to at least what cloud-hypervisor currently supports, that is, 254 vcpus.

I'm also open to bump this to 1024 or 8192 (the maximum number of CPUs Linux supports) because I'm working on >254 vcpu support in CHV anyway.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Without this change, using more than 64 vCPUs with cloud-hypervisor and EDK2 resulted in a stuck boot. With this PR, the VM comes up successfully.

## Integration Instructions

N/A


kraxel:
PR description says CloudHW supports 254 CPUs max.  So why not use that?
Supporting 256 or more CPUs will most likely need more firmware changes, specifically turn on x2apic mode.

tpressure:
> PR description says CloudHW supports 254 CPUs max. So why not use that? Supporting 256 or more CPUs will most likely need more firmware changes, specifically turn on x2apic mode.

I can change this to 254 of course. But like I said, I'm also open to bump this to 1024 (or any other large number) since I am working on >254 vCPU support for cloud-hypervisor anyway. I already have a working prototype, and there's actually nothing more to be done on the EDK2 side besides bumping this number.

tpressure:
@kraxel I've changed it to 254 and updated the commit message accordingly. We can always bump this again when support for >254 vCPUs has landed in cloud-hypervisor.

kraxel (approved):

tpressure:
bump 

@kraxel can you tell me what I need to do here to get this over the finish line?

ardbiesheuvel:
> bump
> 
> @kraxel can you tell me what I need to do here to get this over the finish line?

Please fix the PatchCheck error reported by CI

PR #11175, opened 2025-06-10
BaseTools/PatchCheck.py: add package name checks in subject lines

leiflindholm:
This PR adds a test to PatchCheck.py that packages modified by a commit are listed
in the commit subject line, as set out in

Suggested by @ardbiesheuvel.


BaseTools/Scripts/PatchCheck.py
```diff
@@ -209,6 +210,14 @@ def find_signatures(self, sig):
 
         return sigs
 
+    def check_subject(self, updated_packages):
+        if updated_packages:
+            for package in updated_packages:
+                current_package_re = r".*" + re.escape(package) + r".*:.*"
+                if not re.search(current_package_re, self.subject):
+                    self.error("Subject line does not follow required format.")
```
ardbiesheuvel: Thanks for hacking this up. 

Might we be a bit more helpful here in describing how the subject line does not follow the required format?

leiflindholm: Yeah, I did think about that, but it gets messy real quick, since technically it's not just packages (also BaseTools and a bunch of .-directories). Plus, when this fails the script does output a link to the wiki documentation on commit format.

Very open to suggestions.

leiflindholm: Hmm, so if we vague it up a bit, something that isn't an absolute lie ... how about:
`Subject line not in "package/component: description" format`?

ardbiesheuvel: Yeah that would work nicely

BaseTools/Scripts/PatchCheck.py
```diff
@@ -695,7 +704,8 @@ def __init__(self, rev_spec, max_count):
             email = self.read_committer_email_address_from_git(commit)
             self.ok &= EmailAddressCheck(email, 'Committer').ok
             patch = self.read_patch_from_git(commit)
-            check_patch = CheckOnePatch(commit, patch)
+            updated_packages = self.get_parent_packages (dec_files, commit, 'ADM')
+            check_patch = CheckOnePatch(commit, patch, updated_packages)
```
pierregondois, line 708: Ideally I think the list of modified packages should be extracted from the patch itself, otherwise the following path breaks:
```
CheckOnePatchFile()
\-CheckOnePatch() # updated_packages is not provided here
```
This can be triggered by calling PatchCheck on a patch file:
```
python3 BaseTools/Scripts/PatchCheck.py 0001-BaseTools[...].patch
```

Finding out which files have been modified in a .patch file could be done using `lsdiff`, but this requires installing `patchutils`. Otherwise just greping for '^+++' and '^---' maybe

leiflindholm: OK, good catch.
But jumping out into tempfiles when we're already extracting the necessary information without them feels less than ideal. I'll see how I can replumb.

leiflindholm: Ah, is it possible that you and I were both unaware of the other's normal workflow?
I _never_ call PatchCheck.py on files. Hadn't even noticed that was an option.

But when you say "breaks", you mean "doesn't flag", right?

leiflindholm: Presumably that means the calling-on-patch-file flow already doesn't trap for example multi-package without `Continuous-integration-options: PatchCheck.ignore-multi-package`.
Hmm, yes it does (mean that).

Given that, would you settle for the following
```
diff --git a/BaseTools/Scripts/PatchCheck.py b/BaseTools/Scripts/PatchCheck.py
index 846f9449499a..581e3a8bb33f 100755
--- a/BaseTools/Scripts/PatchCheck.py
+++ b/BaseTools/Scripts/PatchCheck.py
@@ -570,7 +570,7 @@ class CheckOnePatch:
     patch content.
     """
 
-    def __init__(self, name, patch, updated_packages):
+    def __init__(self, name, patch, updated_packages=None):
         self.patch = patch
         self.find_patch_pieces()
 
```
which fixes my breaking of the file-based checking without addressing the fact that this flow already only checks a subset of the commit-based flow?

pierregondois: Yes ok,
TBH, I am not using `PatchCheck` on `.patch` files either. I was just concerned about having the command line crashing if someone was using it

leiflindholm:
Any feedback from BaseTools maintainers?
@bexcran @gapalomi @lgao4 @YuweiChen1110

BaseTools/Scripts/PatchCheck.py
```diff
@@ -209,6 +210,14 @@ def find_signatures(self, sig):
 
         return sigs
 
+    def check_subject(self, updated_packages):
+        if updated_packages:
+            for package in updated_packages:
+                current_package_re = r".*" + re.escape(package) + r".*:.*"
```
gapalomi: This regex will allow typos in the package path. For example:

`xxxBaseToolsxxx/PatchCheck.py: verify commit message lists package(s)` would be a valid commit title

leiflindholm: Yes. It will also permit multiple packages enumerated, as required for commits that touches multiple packages. And any outside-of-packages top-level files.

gapalomi: What about a more robust regex like: `r"(^|, ?)" + re.escape(package) + r"([ ,:\/])"`
for example:  `(^|, ?)BaseTools([ ,:\/])` will have the following result:

```
xxxBaseToolsxxx/PatchCheck.py: some description ❌
BaseTools,UefiPayloadPkg: some description  ✅
ArmPkg,ArmVirtPkg, BaseTools: some description ✅
ArmPkg,ArmVirtPkg,BaseTools: some description ✅
BaseTools/IndustryStandard: some description ✅
```
In this way we can prevent typos and still support multiple packages.

leiflindholm: Sure, that makes sense. Updated.

leiflindholm:
Ping? @bexcran @gapalomi @lgao4 @YuweiChen1110

gapalomi (approved):

leiflindholm:
@makubacki @mdkinney hmm, it seems CI is failing due to timeout in the OVMF_X64_MM_DEBUG test. But this PR only modifies PatchCheck.py.

I suspect bbee92c9af59e877f9a5028140e600571faea181 from #11210.

leiflindholm:
> @makubacki @mdkinney hmm, it seems CI is failing due to timeout in the OVMF_X64_MM_DEBUG test. But this PR only modifies PatchCheck.py.
> 
> I suspect [bbee92c](https://github.com/tianocore/edk2/commit/bbee92c9af59e877f9a5028140e600571faea181) from #11210.

Ah, I note #11219 has already been raised.

leiflindholm:
Hmm, the current dilemma makes me realise we also want to permit reverts,
which come with a subject line of the form:
Revert "<Subject line of commit to revert>".
@gapalomi thoughts?

gapalomi:
> Hmm, the current dilemma makes me realise we also want to permit reverts, which come with a subject line of the form: Revert "". @gapalomi thoughts?

Ah, good catch. I don't see documentation about reverts in https://github.com/tianocore/tianocore.github.io/wiki/Commit-Message-Format

Here are some of the reverts I found in the log:

- `ArmVirtPkg/ArmVirtQemu: Revert "Permit the use of dynamic PCDs in PEI"`
- `ArmVirtPkg: Revert "Use dynamic PCD to set the SMCCC conduit"`
- `Revert ae1079b386a597108a8070652bf7cdaa4ec3dda3`
- `Revert "FatPkg: Validate Reserved FAT Entries on Volume Open"`
- `revert/update of a change this morning.`
- `Faint!change messages cause error, this is an revert`

I agree revert should follow: `Revert "<original title>"` format as `Revert "FatPkg: Validate Reserved FAT Entries on Volume Open"`. I suggest adding the check for the "Revert" keyword in the prefix regex as `(Revert "|^|, ?)` and update the documentation

@lgao4, @mdkinney for comments

leiflindholm:
@gapalomi done.

lgao4:
> > Hmm, the current dilemma makes me realise we also want to permit reverts, which come with a subject line of the form: Revert "". @gapalomi thoughts?
> 
> Ah, good catch. I don't see documentation about reverts in https://github.com/tianocore/tianocore.github.io/wiki/Commit-Message-Format
> 
> Here are some of the reverts I found in the log:
> 
> * `ArmVirtPkg/ArmVirtQemu: Revert "Permit the use of dynamic PCDs in PEI"`
> * `ArmVirtPkg: Revert "Use dynamic PCD to set the SMCCC conduit"`
> * `Revert ae1079b386a597108a8070652bf7cdaa4ec3dda3`
> * `Revert "FatPkg: Validate Reserved FAT Entries on Volume Open"`
> * `revert/update of a change this morning.`
> * `Faint!change messages cause error, this is an revert`
> 
> I agree revert should follow: `Revert "<original title>"` format as `Revert "FatPkg: Validate Reserved FAT Entries on Volume Open"`. I suggest adding the check for the "Revert" keyword in the prefix regex as `(Revert "|^|, ?)` and update the documentation
> 
> @lgao4, @mdkinney for comments


> > Hmm, the current dilemma makes me realise we also want to permit reverts, which come with a subject line of the form: Revert "". @gapalomi thoughts?
> 
> Ah, good catch. I don't see documentation about reverts in https://github.com/tianocore/tianocore.github.io/wiki/Commit-Message-Format
> 
> Here are some of the reverts I found in the log:
> 
> * `ArmVirtPkg/ArmVirtQemu: Revert "Permit the use of dynamic PCDs in PEI"`
> * `ArmVirtPkg: Revert "Use dynamic PCD to set the SMCCC conduit"`
> * `Revert ae1079b386a597108a8070652bf7cdaa4ec3dda3`
> * `Revert "FatPkg: Validate Reserved FAT Entries on Volume Open"`
> * `revert/update of a change this morning.`
> * `Faint!change messages cause error, this is an revert`
> 
> I agree revert should follow: `Revert "<original title>"` format as `Revert "FatPkg: Validate Reserved FAT Entries on Volume Open"`. I suggest adding the check for the "Revert" keyword in the prefix regex as `(Revert "|^|, ?)` and update the documentation
> 
> @lgao4, @mdkinney for comments

It is OK to me. Thanks!

lgao4 (approved):

leiflindholm:
@mergify requeue

leiflindholm:
@mergify dequeue

leiflindholm:
@mergify requeue

kuqin12:
@leiflindholm I think this change nullified the `Continuous-integration-options: PatchCheck.ignore-multi-package` notation inside commit message. Is this intentional? See this failure for example: https://github.com/tianocore/edk2/pull/11080/checks?check_run_id=44815508364.

leiflindholm:
> @leiflindholm I think this change nullified the `Continuous-integration-options: PatchCheck.ignore-multi-package` notation inside commit message. Is this intentional? See this failure for example: https://github.com/tianocore/edk2/pull/11080/checks?check_run_id=44815508364.

No, I did specifically take that into account - the test is failing on subject line formatting.
The regexp expects packages enumerated as `OnePkg, AnotherPkg:`.
Whereas these committs do `OnePkg: AnotherPkg:`.

Fwiw, my initial lax approach would have permitted the `OnePkg: AnotherPkg:` format :)

I'm happy to reopen the conversation on how strict the checker should be, but if you swap the first comma for a colon in those subject lines they should go through.

kuqin12:
> > @leiflindholm I think this change nullified the `Continuous-integration-options: PatchCheck.ignore-multi-package` notation inside commit message. Is this intentional? See this failure for example: https://github.com/tianocore/edk2/pull/11080/checks?check_run_id=44815508364.
> 
> No, I did specifically take that into account - the test is failing on subject line formatting. The regexp expects packages enumerated as `OnePkg, AnotherPkg:`. Whereas these committs do `OnePkg: AnotherPkg:`.
> 
> Fwiw, my initial lax approach would have permitted the `OnePkg: AnotherPkg:` format :)
> 
> I'm happy to reopen the conversation on how strict the checker should be, but if you swap the first comma for a colon in those subject lines they should go through.

Ah, thanks for the info! I will give that a try.

gapalomi:
> > @leiflindholm I think this change nullified the `Continuous-integration-options: PatchCheck.ignore-multi-package` notation inside commit message. Is this intentional? See this failure for example: https://github.com/tianocore/edk2/pull/11080/checks?check_run_id=44815508364.
> 
> No, I did specifically take that into account - the test is failing on subject line formatting. The regexp expects packages enumerated as `OnePkg, AnotherPkg:`. Whereas these committs do `OnePkg: AnotherPkg:`.
> 
> Fwiw, my initial lax approach would have permitted the `OnePkg: AnotherPkg:` format :)
> 
> I'm happy to reopen the conversation on how strict the checker should be, but if you swap the first comma for a colon in those subject lines they should go through.

I see both patterns have been used, comma(`,`) separated, and colon(`:`) separated packages:
- `OvmfPkg: PlatformPei: reserve SEC/PEI temp RAM for S3 resume`
- `ArmPkg, ArmVirtPkg: put SpellCheck in AuditOnly mode`

My preference it's to keep only the comma separated title as that makes the colon to be the separation character between packages affected and title.

@makubacki, can you provide some feedback on this?

makubacki:
> 
> I see both patterns have been used, comma(`,`) separated, and colon(`:`) separated packages:
> 
> * `OvmfPkg: PlatformPei: reserve SEC/PEI temp RAM for S3 resume`
> * `ArmPkg, ArmVirtPkg: put SpellCheck in AuditOnly mode`
> 
> My preference it's to keep only the comma separated title as that makes the colon to be the separation character between packages affected and title.
> 
> @makubacki, can you provide some feedback on this?

A few comments, please help clarify my understanding of the current state of conversation.

1. These are not packages but a package and a module: `OvmfPkg: PlatformPei:`, I would expect that to be `OvmfPkg/PlatformPei:`.
2. We normally don't allow multiple packages in a single patch. If there are exceptions, I think both the exception criteria and the format used to specify the multiple packages (whatever that may be) should be described in https://github.com/tianocore/tianocore.github.io/wiki/Commit-Message-Format
   - If multiple packages are allowed, I agree with this reasoning so it is simpler for parsers to determine where a description line in the subject begins:
     > My preference it's to keep only the comma separated title as that makes the colon to be the separation character between packages affected and title.

PR #11179, opened 2025-06-11
Free handle buffers after LocateHandleBuffer

MarsDoge:
Free handle buffers after LocateHandleBuffer

Fix memory leaks by adding missing FreePool calls:

- SpiBusDxe: release SpiHcHandles in SpiBusEntry exit path
- HttpUtilitiesDxe: free HandleBuffer in HttpUtilitiesDxeUnload
- DxePrmContextBufferLib: cleanup HandleBuffer in GetModuleContextBuffers
- RedfishConfigHandler: free handle buffers in Stop and initialization routines

REF: https://uefi.org/sites/default/files/resources/UEFI_Spec_Final_2.11.pdf
Chapter 7.3.15: "Services - Boot Services.LocateHandleBuffer":
It is the caller's responsibility to call the Boot Service.FreePool when
the caller no longer requires the contents of Buffer.


niruiyu (approved):

Li-Aaron (approved):

NetworkPkg/HttpUtilitiesDxe/HttpUtilitiesDxe.c
```diff
@@ -72,10 +72,15 @@ HttpUtilitiesDxeUnload (
                     NULL
                     );
     if (EFI_ERROR (Status)) {
-      return Status;
+      goto Exit;
     }
   }
 
+Exit:
+  if (HandleBuffer != NULL) {
```
leiflindholm: This test does nothing useful. If the LocateHandleBuffer call failed in the first place, we already returned. Please drop it and call FreePool unconditionally.

MdeModulePkg/Bus/Spi/SpiBus/SpiBusDxe.c
```diff
@@ -193,6 +193,13 @@ SpiBusEntry (
   }
 
 Exit:
+  //
+  // Free the buffer containing the list of handles from the handle database
+  //
+  if (SpiHcHandles != NULL) {
```
leiflindholm: It's not clear to me that SpiHcHandles *will* be NULL if the LocateHandleBuffer call fails, so this test isn't safe.

Meanwhile, the function could return directly after LocateHandleBuffer if that was the case. Then this test could be dropped and FreePool be called unconditionally.

PrmPkg/Library/DxePrmContextBufferLib/DxePrmContextBufferLib.c
```diff
@@ -147,7 +149,14 @@ GetModuleContextBuffers (
     Guid
     ));
 
-  return EFI_NOT_FOUND;
+  Status = EFI_NOT_FOUND;
+
+Exit:
+  if (HandleBuffer != NULL) {
```
leiflindholm: Could we drop the success handling on line 107 instead and return immediately if LocateHandleBuffer fails?

We could then drop this redundant test, which is probably unsafe anyway, since I think the value of HandleBuffer would be undefined at that point?

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.c
```diff
@@ -272,4 +277,8 @@ RedfishConfigHandlerInitialization (
                     );
     ASSERT_EFI_ERROR (Status);
   }
+
+  if (HandleBuffer != NULL) {
```
leiflindholm: We already returned on line 244 in the situation where HandleBuffer does not hold a valid pointer.

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.c
```diff
@@ -189,7 +189,7 @@ RedfishConfigCommonStop (
                   &HandleBuffer
                   );
   if (EFI_ERROR (Status) && (Status != EFI_NOT_FOUND)) {
-    return Status;
+    goto Exit;
```
leiflindholm: It seems to me the existing code holds a bug-in-waiting, since in the case of Status == EFI_NOT_FOUND the code will happily progress beyond where anything relevant could be done.

Meanwhile, after this patch it will attempt to free whatever HandleBuffer randomly points to (in the EFI_NOT_FOUND case).

Better to do a separate
```
if (Status == EFI_NOT_FOUND) {
  return EFI_SUCCESS;
} else if (EFI_ERROR (Status)) {
  return Status;
}
```

and then unconditionally calling FreePool at the end of the function.

NetworkPkg/HttpUtilitiesDxe/HttpUtilitiesDxe.c
MarsDoge: Okay, I will drop it, and return correct status.

MdeModulePkg/Bus/Spi/SpiBus/SpiBusDxe.c
MarsDoge: To be sure, the allocated buffer must be released after LocateHandleBuffer returns successfully, so I agree with your suggestion: release it directly later, and check for errors after the first call, and return early if there are any problems.

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.c
MarsDoge: According to the original logic, when `Status == EFI_NOT_FOUND` occurs, an error may occur when accessing HandleBuffer, I will separate it.

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.c
```diff
@@ -208,6 +209,8 @@ RedfishConfigCommonStop (
     }
   }
 
+  FreePool (HandleBuffer);
```
leiflindholm: For consistency, could you make this one gBs->FreePool as well, like the other instances?

NetworkPkg/HttpUtilitiesDxe/HttpUtilitiesDxe.c
```diff
@@ -72,11 +72,15 @@ HttpUtilitiesDxeUnload (
                     NULL
                     );
     if (EFI_ERROR (Status)) {
-      return Status;
+      goto Exit;
     }
   }
+  Status = EFI_SUCCESS;
```
leiflindholm, line 79: There is a CI failure here because uncrustify wants a blank line after }

MarsDoge: Okay, it has been corrected.

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.c
MarsDoge: Okay, it has been changed.

leiflindholm:
@changab @igorkulchytskyy @nicklela could we get a review for RedfishPkg please?

igorkulchytskyy (approved):
Approved the changes in RedfishPkg

PR #11196, opened 2025-06-15
UI tweaks

bexcran:
# Description

I noticed a couple of simple UI tweaks that might improve the user experience.
Rename 'Reset' to 'Reboot' to make it clear it won't reset the settings, but restart the computer.

Also, change 'Tls Auth Configuration' to 'TLS Auth Configuration'.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Booted into UiApp and looked at the pages.

## Integration Instructions

N/A


ardbiesheuvel (approved):

MdeModulePkg/Application/UiApp/FrontPageStrings.uni
```diff
@@ -50,10 +50,10 @@
                                        #language fr-FR  "Missing String"
 #string STR_EMPTY_STRING               #language en-US  ""
                                        #language fr-FR  ""
-#string STR_RESET_STRING               #language en-US  "Reset"
-                                       #language fr-FR  "Reset"
-#string STR_RESET_STRING_HELP          #language en-US  "Reset the current setting."
-                                       #language fr-FR  "Reset the current setting."
```
leiflindholm, line 56: This isn't French. Which seems to be a problem in several locations in this file.

kraxel (approved):

MdeModulePkg/Application/UiApp/FrontPageStrings.uni
```diff
@@ -50,10 +50,10 @@
                                        #language fr-FR  "Missing String"
 #string STR_EMPTY_STRING               #language en-US  ""
                                        #language fr-FR  ""
-#string STR_RESET_STRING               #language en-US  "Reset"
-                                       #language fr-FR  "Reset"
-#string STR_RESET_STRING_HELP          #language en-US  "Reset the current setting."
-                                       #language fr-FR  "Reset the current setting."
+#string STR_RESET_STRING               #language en-US  "Reboot"
```
leiflindholm: I agree the previous state was confusing, but the result of selecting this option _is_ calling gRt->ResetSystem(). Specifically with the EfiResetCold argument.

And technically speaking, we haven't finished booting yet since we're in the Boot-Device Selection (BDS) phase ;)

Anyway: this clearly needs fixing, but my preference would be aligning with UEFI terminology, if that's possible without being confusing.

Would it be enough to keep the Reset as the menu item, but change the help string (which with your change is actually used) to "Cold reset (reboots the computer)."?

bexcran: I like that idea. I wasn't sure about diverging from UEFI terminology, so I think just updating the help text makes sense.

leiflindholm (approved):
With the only caveat that I don't know French very well...

PR #11197, opened 2025-06-16
CryptoPkg: Add API to set the TLS Security level.

kanagavels97:
Add a fixed PCD support to adjust the OpensslTlsSecurity level. The existing 0 permits everything as per the Documentation 3.4

# Description

Add a fixed PCD support to adjust the OpensslTlsSecurity level. As per the [https://docs.openssl.org/3.4/man3/SSL_CTX_set_security_level/#default-callback-behaviour](url) openssl documentation 3.4 set 0 level security permits everything. So, we have implemented adjusting the security level 2 as the default through the fixed PCD.


liyi77:
New PCD will break platforms build, new API would be a better choice.

kanagavels97:
> New PCD will break platforms build, new API would be a better choice.

@liyi77 Do you suggest any specific conditions for returning the security level from the API?
Because, if it is a PCD, we can provide an option to the user to change the security level.

liyi77:
Such as:

TlsSetSecurityLevel (
  IN     VOID   *Tls,
  IN     UINTN  Level
  )
{
  TlsConn = (TLS_CONNECTION *)Tls;
  SSL_set_security_level (TlsConn->Ssl, Level);

kanagavels97:
> Such as:
> 
> TlsSetSecurityLevel ( IN VOID *Tls, IN UINTN Level ) { TlsConn = (TLS_CONNECTION *)Tls; SSL_set_security_level (TlsConn->Ssl, Level);

@liyi77  This is function ok.

I noticed that the above implementation does not provide an option for the user to modify the security level.
However, If it is a fixed PCD, we can allow the user to change the security level as their requirement through PCD.
The ultimate aim in my case is to allow the user to change the security level; otherwise, we can just keep it as passing the 0 to SSL_set_security_level.

Could you please let me know how we can handle this better than PCD?

liyi77:
If the user wants to change the security level to N, they can call TlsSetSecurityLevel ( Tls, N);.
If not, the API can be ignored.
I believe the result is the same as the PCD implementation. New API is more flexible than PCD at least for me.

kanagavels97:
> If the user wants to change the security level to N, they can call TlsSetSecurityLevel ( Tls, N);. If not, the API can be ignored. I believe the result is the same as the PCD implementation. New API is more flexible than PCD at least for me.

@liyi77 The user prefers not to modify the EDK2 file everytime for each project. Since this is an EDK2 file, using an API would require maintaining those changes privately for each project.

kanagavels97:
> > If the user wants to change the security level to N, they can call TlsSetSecurityLevel ( Tls, N);. If not, the API can be ignored. I believe the result is the same as the PCD implementation. New API is more flexible than PCD at least for me.
> 
> @liyi77 The user prefers not to modify the EDK2 file everytime for each project. Since this is an EDK2 file, using an API would require maintaining those changes privately for each project.
@liyi77 Do you have any update for the above comment?

liyi77:
I stand by my point that the new API will be more flexible, allowing multiple TLS instances to have different levels as needed, rather than fixed at build time.
PCD will be an invisible change not noticeable to users, and any new functionality should be obvious. This would be a bad start.

kanagavels97:
Closing this PR as I have created a new PR.

Please find a new PR https://github.com/tianocore/edk2/pull/11331

Issue #11002, opened 2025-04-24
Default Seed used for random number generator in CryptoPkg.


kanagavels97:
Hi @mdkinney 

The random number generators must be seeded with unpredictable values before performing any cryptographic operations. Operations using a default random seed can be compromised/bypassed by an attacker.

The implementation for RandomSeed is as follows, where a NULL value for the Seed parameter results in the usage of DefaultSeed.

![Image](https://github.com/user-attachments/assets/8b791638-5484-4c6e-b3b4-a36f555239bb)

The following functionality is present in CryptPkcs7Sign.c file, which is an OpenSSL wrapper where the default seed is used before generating a private key used for signing.

![Image](https://github.com/user-attachments/assets/aa0bc2d7-7db1-4aa4-a84c-e9b7a28b162c)
![Image](https://github.com/user-attachments/assets/085a93e5-65bc-4b75-b6c6-6b3bec970ec0)

The following code snippet is in TlsInit.c file for initializing TLS functionality, which uses the default seed value for the random number generator.

![Image](https://github.com/user-attachments/assets/53140035-8446-4a33-a8e2-b81a014ec2be)

Using a known or predictable seed value weakens or fully compromises the cryptographic operation, allowing it to be exploited. An attacker can bypass TLS session authentication or mechanisms validating PKCS #7 signatures.

Could you explain why the RandomSeed() function is called with a seed value of NULL and a seed size of 0?

Could you let us know if there are plans to implement a Cryptographically Secure Pseudo-Random Number Generator (CSPRNG)?

Thanks,
Kanagavel

cc: srinivasanmani@ami.com gayathrit@ami.com prarthanasagar@ami.com kanagavels@ami.com

liyi77:
For X86, we default to use tsc as random seed (RandTsc.c).

Even so its entropy is not good enough for cryptography.
Using RngLib instead might be a better approach.
@jyao1

kanagavels97:
Hi @liyi77 @mdkinney 

In the CryptRandTsc.file, the RandomSeed function uses the current system time as a seed when it is called with a NULL seed value and a seed size of 0. However, since the current time is predictable, it does not provide strong randomness.

Based on your recommendation, we have decided to use RngLib as the random number generator for both x64 and AArch64 architectures.

Once we’ve completed the implementation using RngLib, would it be appropriate to submit a patch to EDK II? We’d appreciate your guidance or suggestions on this.

Thanks,
Kanagavel

[...]

PR #11198, opened 2025-06-16
CryptoPkg: Use RngLib to get Random number.

kanagavels97:
REF: https://github.com/tianocore/edk2/issues/11002

Use RngLib to get the Random number instead of passing the Default seed when the seed value is NULL and Seed size is zero.

# Description

Add an implementation to use the RngLib to get the RandomNumber instead of passing the predictable default seed value when the RandomSeed() is called with seed value NULL and Seed size zero.

### DIFF
--- a/CryptoPkg/Library/BaseCryptLib/Rand/CryptRand.c
+++ b/CryptoPkg/Library/BaseCryptLib/Rand/CryptRand.c
@@ -9,11 +9,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include "InternalCryptLib.h"
 #include <openssl/rand.h>
 #include <openssl/evp.h>
-
-//
-// Default seed for UEFI Crypto Library
-//
-CONST UINT8  DefaultSeed[] = "UEFI Crypto Library default seed";
+#include <Library/RngLib.h>
 
 /**
   Sets up the seed value for the pseudorandom number generator.
@@ -38,6 +34,8 @@ RandomSeed (
   IN  UINTN         SeedSize
   )
 {
+  UINT32  RandomNumber;
+
   if (SeedSize > INT_MAX) {
     return FALSE;
   }
@@ -49,7 +47,11 @@ RandomSeed (
   if (Seed != NULL) {
     RAND_seed (Seed, (UINT32)SeedSize);
   } else {
-    RAND_seed (DefaultSeed, sizeof (DefaultSeed));
+    if (!GetRandomNumber32 (&RandomNumber)) {
+      return FALSE;
+    }
+
+    RAND_seed (&RandomNumber, sizeof (RandomNumber));
   }
 
   if (RAND_status () == 1) {
--- a/CryptoPkg/Library/BaseCryptLib/Rand/CryptRandTsc.c
+++ b/CryptoPkg/Library/BaseCryptLib/Rand/CryptRandTsc.c
@@ -10,6 +10,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <openssl/rand.h>
 #include <openssl/evp.h>
 #include <Library/PrintLib.h>
+#include <Library/RngLib.h>
 
 /**
   Sets up the seed value for the pseudorandom number generator.
@@ -34,7 +35,7 @@ RandomSeed (
   IN  UINTN         SeedSize
   )
 {
-  CHAR8  DefaultSeed[128];
+  UINT32  RandomNumber;
 
   if (SeedSize > INT_MAX) {
     return FALSE;
@@ -47,17 +48,12 @@ RandomSeed (
   if (Seed != NULL) {
     RAND_seed (Seed, (UINT32)SeedSize);
   } else {
-    //
-    // Retrieve current time.
-    //
-    AsciiSPrint (
-      DefaultSeed,
-      sizeof (DefaultSeed),
-      "UEFI Crypto Library default seed (%ld)",
-      AsmReadTsc ()
-      );
+    // Use RngLib to get the RandomNumber.
+    if (!GetRandomNumber32 (&RandomNumber)) {
+      return FALSE;
+    }
 
-    RAND_seed (DefaultSeed, sizeof (DefaultSeed));
+    RAND_seed (&RandomNumber, sizeof (RandomNumber));
   }
 
   if (RAND_status () == 1) {
--- a/CryptoPkg/Library/BaseCryptLib/BaseCryptLib.inf
+++ b/CryptoPkg/Library/BaseCryptLib/BaseCryptLib.inf
@@ -99,6 +99,7 @@
   PrintLib
   UefiBootServicesTableLib
   SynchronizationLib
+  RngLib
 
 [Protocols]
   gEfiMpServiceProtocolGuid
--- a/CryptoPkg/Library/BaseCryptLib/RuntimeCryptLib.inf
+++ b/CryptoPkg/Library/BaseCryptLib/RuntimeCryptLib.inf
@@ -98,6 +98,7 @@
   OpensslLib
   IntrinsicLib
   PrintLib
+  RngLib
 
 #
 # Remove these [BuildOptions] after this library is cleaned up
--- a/CryptoPkg/Library/BaseCryptLib/SmmCryptLib.inf
+++ b/CryptoPkg/Library/BaseCryptLib/SmmCryptLib.inf
@@ -94,6 +94,7 @@
   PrintLib
   MmServicesTableLib
   SynchronizationLib
+  RngLib
 
 #
 # Remove these [BuildOptions] after this library is cleaned up


pierregondois:
The patch looks good to me. As a side comment, shouldn't we have a valid implementation of `RandomSeed()` in `CryptoPkg/Library/BaseCryptLibMbedTls/Rand` ?

```
BOOLEAN
EFIAPI
RandomSeed (
  IN  CONST  UINT8  *Seed  OPTIONAL,
  IN  UINTN         SeedSize
  )
{
  return TRUE;
}
```

kanagavels97:
> The patch looks good to me. As a side comment, shouldn't we have a valid implementation of `RandomSeed()` in `CryptoPkg/Library/BaseCryptLibMbedTls/Rand` ?
> 
> ```
> BOOLEAN
> EFIAPI
> RandomSeed (
>   IN  CONST  UINT8  *Seed  OPTIONAL,
>   IN  UINTN         SeedSize
>   )
> {
>   return TRUE;
> }
> ```

Do you want the same implementation for MbedTls library as well?

pierregondois:
> > The patch looks good to me. As a side comment, shouldn't we have a valid implementation of `RandomSeed()` in `CryptoPkg/Library/BaseCryptLibMbedTls/Rand` ?
> > ```
> > BOOLEAN
> > EFIAPI
> > RandomSeed (
> >   IN  CONST  UINT8  *Seed  OPTIONAL,
> >   IN  UINTN         SeedSize
> >   )
> > {
> >   return TRUE;
> > }
> > ```
> 
> Do you want the same implementation for MbedTls library as well?

It would be a good think IMO.
Another thing is that adding a dependency of `CryptoPkg/Library/BaseCryptLib/BaseCryptLib.inf` over the RngLib might break some platform builds. For instance:
`Platform/Sophgo/SG2042_EVB_Board/SG2042.dsc`
will break.

Note that the build seems already failing. I had to to the following to see the unresolved dependency:
```
diff --git a/Platform/Sophgo/SG2042_EVB_Board/SG2042.dsc b/Platform/Sophgo/SG2042_EVB_Board/SG2042.dsc
index 5f15e3a621a1..03dd084fdb9b 100644
--- a/Platform/Sophgo/SG2042_EVB_Board/SG2042.dsc
+++ b/Platform/Sophgo/SG2042_EVB_Board/SG2042.dsc
@@ -29,7 +29,7 @@
   # Defines for default states.  These can be changed on the command line.
   # -D FLAG=VALUE
   #
-  DEFINE SECURE_BOOT_ENABLE      = FALSE
+  DEFINE SECURE_BOOT_ENABLE      = TRUE
   DEFINE DEBUG_ON_SERIAL_PORT    = TRUE
 
   #
@@ -99,7 +99,7 @@
   SortLib|MdeModulePkg/Library/BaseSortLib/BaseSortLib.inf
   ShellLib|ShellPkg/Library/UefiShellLib/UefiShellLib.inf
   UefiBootManagerLib|MdeModulePkg/Library/UefiBootManagerLib/UefiBootManagerLib.inf
-  FdtLib|EmbeddedPkg/Library/FdtLib/FdtLib.inf
+  FdtLib|MdePkg/Library/BaseFdtLib/BaseFdtLib.inf
   VariableFlashInfoLib|MdeModulePkg/Library/BaseVariableFlashInfoLib/BaseVariableFlashInfoLib.inf
   VariablePolicyHelperLib|MdeModulePkg/Library/VariablePolicyHelperLib/VariablePolicyHelperLib.inf
   ImagePropertiesRecordLib|MdeModulePkg/Library/ImagePropertiesRecordLib/ImagePropertiesRecordLib.inf
@@ -153,7 +153,7 @@
   RealTimeClockLib|EmbeddedPkg//Library/VirtualRealTimeClockLib/VirtualRealTimeClockLib.inf
 
   # Flattened Device Tree (FDT) access library
-  FdtLib|EmbeddedPkg/Library/FdtLib/FdtLib.inf
+  FdtLib|MdePkg/Library/BaseFdtLib/BaseFdtLib.inf
 
 [LibraryClasses.common.SEC]
 !ifdef $(DEBUG_ON_SERIAL_PORT)
```

and the command:
`build -p Platform/Sophgo/SG2042_EVB_Board/SG2042.dsc -t GCC5 -a RISCV64`

CryptoPkg/Library/BaseCryptLib/Rand/CryptRandTsc.c
```diff
@@ -47,17 +48,12 @@ RandomSeed (
   if (Seed != NULL) {
     RAND_seed (Seed, (UINT32)SeedSize);
   } else {
-    //
```
liyi77, line 50: I think we don't need to keep two files with exactly the same content. CryptRanTsc.c can be removed.

liyi77:
> It would be a good think IMO. Another thing is that adding a dependency of `CryptoPkg/Library/BaseCryptLib/BaseCryptLib.inf` over the RngLib might break some platform builds. For instance: `Platform/Sophgo/SG2042_EVB_Board/SG2042.dsc` will break.
> 

Sounds strange, RngLib is a dependency of OpensslLib: https://github.com/tianocore/edk2/blob/92c714f8b7482f69f6f66bf78b884feacec66fa1/CryptoPkg/Library/OpensslLib/OpensslLib.inf#L706-L710

Generally as long as you use Openssl CryptoLib, this dependency is already satisfied.

pierregondois:
> Sounds strange, RngLib is a dependency of OpensslLib:
> 
> https://github.com/tianocore/edk2/blob/92c714f8b7482f69f6f66bf78b884feacec66fa1/CryptoPkg/Library/OpensslLib/OpensslLib.inf#L706-L710
> 
> Generally as long as you use Openssl CryptoLib, this dependency is already satisfied.

Yes right, the RngLib dependency is already broken for the platform.
While checking some other platforms, it seems that the RngLib is present for all of them. However I didn't try to build all of them.

CryptoPkg/Library/BaseCryptLib/Rand/CryptRandTsc.c
kanagavels97: @liyi77 You mean we can use CryptRand.c file for IA32, X64, ARM, and AARCH64?

liyi77: > @liyi77 You mean we can use CryptRand.c file for IA32, X64, ARM, and AARCH64?

Yes, they are same things anyway

kanagavels97: Ok @liyi77 , I will remove CryptRandTsc.c file

kanagavels97:
> > Sounds strange, RngLib is a dependency of OpensslLib:
> > https://github.com/tianocore/edk2/blob/92c714f8b7482f69f6f66bf78b884feacec66fa1/CryptoPkg/Library/OpensslLib/OpensslLib.inf#L706-L710
> > 
> > Generally as long as you use Openssl CryptoLib, this dependency is already satisfied.
> 
> Yes right, the RngLib dependency is already broken for the platform. While checking some other platforms, it seems that the RngLib is present for all of them. However I didn't try to build all of them.

@pierregondois 

Could you please confirm whether the RngLib dependency is still required for BaseCryptLib?

pierregondois:
> > > Sounds strange, RngLib is a dependency of OpensslLib:
> > > https://github.com/tianocore/edk2/blob/92c714f8b7482f69f6f66bf78b884feacec66fa1/CryptoPkg/Library/OpensslLib/OpensslLib.inf#L706-L710
> > > 
> > > Generally as long as you use Openssl CryptoLib, this dependency is already satisfied.
> > 
> > 
> > Yes right, the RngLib dependency is already broken for the platform. While checking some other platforms, it seems that the RngLib is present for all of them. However I didn't try to build all of them.
> 
> @pierregondois
> 
> Could you please confirm whether the RngLib dependency is still required for BaseCryptLib?

I think you patch introduces the dependency of BaseCryptLib over the RngLib. Some modules/platforms might not satisfy this dependency.
However I did personally face such breakage. The `Platform/Sophgo/SG2042_EVB_Board/SG2042.dsc` platform doesn't build before applying you change.
So I can confirm I didn't face any new breakage due to your patch. I cannot guarantee all platforms/modules will satisfy the dependency by default though.

CryptoPkg/Library/BaseCryptLib/Rand/CryptRand.c
```diff
@@ -49,7 +47,11 @@ RandomSeed (
   if (Seed != NULL) {
     RAND_seed (Seed, (UINT32)SeedSize);
   } else {
-    RAND_seed (DefaultSeed, sizeof (DefaultSeed));
+    if (!GetRandomNumber32 (&RandomNumber)) {
+      return FALSE;
+    }
```
jyao1, line 52: Will this cause compatibility problem?

What if the random pool does not have enough entropy at that time?

liyi77: > What if the random pool does not have enough entropy at that time?

OpenSSL will call into here when entropy not enough, which use RngLib to acquire requested amount of entropy:
https://github.com/tianocore/edk2/blob/1bbd68755c9dec61662e27b30ef71a45a33331f2/CryptoPkg/Library/OpensslLib/OpensslStub/rand_pool.c#L77-L98

> Will this cause compatibility problem?

As answered in the previous question, RngLib is actually already a dependency of OpensslLib and will not introduce new compatibility issues.

PR #6169, opened 2024-09-07
IntrinsicLib: Move from CryptoPkg to MdePkg

mikebeaton:
# Description

This library is already used by one client other than CryptoPkg, specifically `OvmfPkg/Library/BhyveFwCtlLib`.

It seems likely that certain other third party or internal packages built against EDK 2 may also wish to use this functionality, therefore this commit suggests moving the library to the standard shared location of MdePkg.

I have split this into multiple commits, one per package, as requested by PatchCheck.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

CI

## Integration Instructions

N/A

PR #11201, opened 2025-06-17
MdePkg CompilerIntrinsicsLib: Unify the definitions of size_t in xxx_ms.c

YangGangUEFI:
Previous `typedef unsigned __int32 size_t;` would result in `"error C2371: 'size_t': redefinition; different basic types"` when using `CompilerIntrinsicsLib` in X64 builds.

Use `UINTN` in `ProcessorBind.h` instead.

# Description

- [ ] Breaking change?

- [ ] Impacts security?

- [ ] Includes tests?

## How This Was Tested

Use `-a X64` to build some modules that use `CompilerIntrinsicsLib`

```
  XXX.inf {
    <LibraryClasses>
      NULL|MdePkg/Library/CompilerIntrinsicsLib/CompilerIntrinsicsLib.inf
  }
```

## Integration Instructions

N/A

### DIFF
--- a/MdePkg/Library/CompilerIntrinsicsLib/memcmp_ms.c
+++ b/MdePkg/Library/CompilerIntrinsicsLib/memcmp_ms.c
@@ -7,11 +7,7 @@
 //
 // ------------------------------------------------------------------------------
 
-#if defined (_M_ARM64)
-typedef unsigned __int64 size_t;
-#else
-typedef unsigned __int32 size_t;
-#endif
+typedef UINTN size_t;
 
 int
 memcmp (
--- a/MdePkg/Library/CompilerIntrinsicsLib/memcpy_ms.c
+++ b/MdePkg/Library/CompilerIntrinsicsLib/memcpy_ms.c
@@ -7,11 +7,7 @@
 //
 // ------------------------------------------------------------------------------
 
-#if defined (_M_ARM64)
-typedef unsigned __int64 size_t;
-#else
-typedef unsigned __int32 size_t;
-#endif
+typedef UINTN size_t;
 
 void *
 memcpy (
--- a/MdePkg/Library/CompilerIntrinsicsLib/memmove_ms.c
+++ b/MdePkg/Library/CompilerIntrinsicsLib/memmove_ms.c
@@ -7,6 +7,13 @@
 //
 // ------------------------------------------------------------------------------
 
+//
+// Starting from VS2019, the compiler recognizes memmove() as an intrinsic function,
+// so we need to use _MSC_VER to control the following logic.
+//
+
+#if defined (_MSC_VER) && (_MSC_VER >= 1920)
+
 typedef UINTN size_t;
 
 void *
@@ -45,3 +52,5 @@ memmove (
 
   return dest;
 }
+
+#endif
--- a/MdePkg/Library/CompilerIntrinsicsLib/memmove_ms.c
+++ b/MdePkg/Library/CompilerIntrinsicsLib/memmove_ms.c
@@ -7,11 +7,7 @@
 //
 // ------------------------------------------------------------------------------
 
-#if defined (_M_ARM64)
-typedef unsigned __int64 size_t;
-#else
-typedef unsigned __int32 size_t;
-#endif
+typedef UINTN size_t;
 
 void *
 memmove (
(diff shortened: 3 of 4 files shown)


lgao4 (approved):

YangGangUEFI:
` warning C4163: 'memmove': not available as an intrinsic function` occurs in VS2015, but VS2019/2022 is ok.

Since VS2015 is still supported, it seems we need to handle this situation.

YangGangUEFI:
According to https://devblogs.microsoft.com/cppblog/msvc-backend-updates-in-visual-studio-2019-preview-2/ and https://learn.microsoft.com/en-us/visualstudio/releases/2019/release-notes-v16.0 , we can find:

> The compiler recognizes memmove() as an intrinsic function and optimizes accordingly. This improves code generation for operations built on memmove() including std::copy() and other higher level library code such as std::vector and std::string construction

> Improved optimization of code using memmove, such as std::copy or std::vector and std::string construction

So, we can use `_MSC_VER` to control MdePkg\Library\CompilerIntrinsicsLib\memmove_ms.c

lgao4:
@ardbiesheuvel , do you think this change is OK for ARM arch?

ardbiesheuvel:
No objections from me but this is VS specific so better ask someone like @os-d

os-d:
> ` warning C4163: 'memmove': not available as an intrinsic function` occurs in VS2015, but VS2019/2022 is ok.
> 
> Since VS2015 is still supported, it seems we need to handle this situation.

VS2015 is not still supported. edk2 has not removed definitions for building with it, though that has been planned and desired by BaseTools maintainers for a while now, see https://edk2.groups.io/g/devel/topic/drop_vs2015_support/107129562. It has not been run in CI for a long time and there are no guarantees it works.

VS2015 ended mainstream support 5 years ago and is fully end of life in a couple months: https://learn.microsoft.com/en-us/lifecycle/products/visual-studio-2015. I would prefer to not add the logic for the _MSC_VER check. It has worked for VS2015 until now and any existing users who run into an issue with it should move forward to a toolchain that is not a decade old. In general, it is highly brittle to add compiler version checks in code and we should avoid it unless there is no other choice.

os-d:
Taking a step back, I notice you are trying to use CompilerIntrinsicsLib for X64. This is an ARM and AARCH64 library (for now). Any intrinsics you are hitting in X64 should be worked around by using the BaseLib provided functions, e.g. things like RShiftU64().

For now, edk2 has a policy against using compiler intrinsics, except where it can’t be avoided such as in third party code (OpenSSL) and ARM/AARCH64. See https://github.com/tianocore/edk2/pull/6169 for one of the many discussions on this.

You should not use CompilerIntrinsicsLib for X64 without a greater change made to support X64 and that should go through an RFC process, but you could start with a GH issue.

This change, other than the _MSC_VER part seems okay to me, but I may want to leave that speedbump in using it for X64 there, because it is not made to work there, only a small subset of the intrinsics would work.

There is certainly room to adding logic to CompilerIntrinsicsLib.inf to only build for ARM/AARCH64 (many files already are) to help clarify this.

I’ll defer to @mdkinney for any updated thoughts on intrinsics.

YangGangUEFI:
> I would prefer to not add the logic for the _MSC_VER check.

If we ever decide to remove VS2015 (and possibly VS2017) from BaseTools (so that consumers have to use a newer toolchain), I'm OK with that(remove the second commit).

> Taking a step back, I notice you are trying to use CompilerIntrinsicsLib for X64. This is an ARM and AARCH64 library (for now). Any intrinsics you are hitting in X64 should be worked around by using the BaseLib provided functions, e.g. things like RShiftU64().

For the first commit it's more like https://github.com/tianocore/edk2/pull/176, but also brings some minimum support for X64(If user use it).
I think the commit information can be updated.

> There is certainly room to adding logic to CompilerIntrinsicsLib.inf to only build for ARM/AARCH64 (many files already are) to help clarify this.

Since we have already restricted linking `CompilerIntrinsicsLib` to ARM/AARCH64 only in `MdePkg\MdeLibs.dsc.inc` , this does not seem to be necessary.

os-d:
> > Taking a step back, I notice you are trying to use CompilerIntrinsicsLib for X64. This is an ARM and AARCH64 library (for now). Any intrinsics you are hitting in X64 should be worked around by using the BaseLib provided functions, e.g. things like RShiftU64().
> 
> For the first commit it's more like #176, but also brings some minimum support for X64(If user use it). I think the commit information can be updated.
> 

What I'm saying is that I don't think we should add X64 support to this library in a partial way, going against the edk2 policy. If we are going to change the policy, that should be an RFC and then the library should be completely updated to support other architectures.

PR #11204, opened 2025-06-17
Don't Allocate Page 0

os-d:
# Description

Currently DxeIpl attempts to set page 0 to all 0's and to create a memory allocation HOB for it. However, DxeIpl will also unmap the page when mapping page tables and if null detection is not enabled, DxeCore will set the page to 0, regardless of allocation status.

Because no consumers are using the memory allocation HOB for page 0, drop it. Instead, ensure that PeiCore and DxeCore do not allow allocating page 0; it should always be reserved for null pointer detection. It also complicates the story for platforms that are attempting to audit the system and ensure that no modules are using page 0. With these memory allocation HOBs in place, it is difficult to tell if it is simply DxeIpl who has allocated the memory or another module.

This commit drops the memory allocation HOB publishing and ensures that DxeCore and PeiCore do not allocate page 0. DxeCore already will not allocate page 0 to callers of AllocatePages who call with a type other than AllocateAddress, this just changes so that AllocateAddress cannot allocate at page 0 (which if null detection is enabled will cause a page fault). PeiCore does not have AllocateAddress and so this ensures standard allocations do not receive page 0.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested booting Ovmf and ArmVirt to shell.

## Integration Instructions

Any modules attempting to allocate page 0 by address must change what address they are allocating.

### DIFF
--- a/BaseTools/Source/Python/build/BuildReport.py
+++ b/BaseTools/Source/Python/build/BuildReport.py
@@ -1814,13 +1814,21 @@ class FdRegionReport(object):
             for Ffs in Wa.FdfProfile.FvDict[FvName.upper()].FfsList:
                 for Section in Ffs.SectionList:
                     try:
-                        for FvSection in Section.SectionList:
-                            if FvSection.FvName in self.FvList:
-                                continue
-                            self._GuidsDb[Ffs.NameGuid.upper()] = FvSection.FvName
-                            self.FvList.append(FvSection.FvName)
-                            self.FvInfo[FvSection.FvName] = ("Nested FV", 0, 0)
-                            self._DiscoverNestedFvList(FvSection.FvName, Wa)
+                        # Handle the case where an entire FFS is a FV, and not
+                        # a sub-section of the FFS.
+                        if getattr(Section, 'FvFileName', None) is None:
+                            for FvSection in Section.SectionList:
+                                if FvSection.FvName in self.FvList:
+                                    continue
+                                self._GuidsDb[Ffs.NameGuid.upper()] = FvSection.FvName
+                                self.FvList.append(FvSection.FvName)
+                                self.FvInfo[FvSection.FvName] = ("Nested FV", 0, 0)
+                                self._DiscoverNestedFvList(FvSection.FvName, Wa)
+                        else:
+                            self._GuidsDb[Ffs.NameGuid.upper()] = Section.FvFileName
+                            self.FvList.append(Section.FvName)
+                            self.FvInfo[Section.FvName] = ("Nested FV", 0, 0)
+                            self._DiscoverNestedFvList(Section.FvName, Wa)
                     except AttributeError:
                         pass
 


ardbiesheuvel (approved):

lgao4:
I know X86 Legacy Boot needs to allocate zero address. This change will not allow this boot mode. Right?

kraxel:
> I know X86 Legacy Boot needs to allocate zero address. This change will not allow this boot mode. Right?

That is correct.  Legacy boot support has been removed from the edk2 repo though.
CSM support was removed in 2023, VbeShim (which needed the zero page too for int10h hacks) followed in 2024.

olegilyasov:
> I know X86 Legacy Boot needs to allocate zero address. This change will not allow this boot mode. Right?

Yes, indeed. Legacy boot will be broken after this change.
In general, NULL pointer detection is a debugging feature that should not be generic. NULL pointer related problems are as "good" as any other memory corruptions.

os-d:
> > I know X86 Legacy Boot needs to allocate zero address. This change will not allow this boot mode. Right?
> 
> Yes, indeed. Legacy boot will be broken after this change. In general, NULL pointer detection is a debugging feature that should not be generic. NULL pointer related problems are as "good" as any other memory corruptions.

As Gerd noted, edk2 no longer supports legacy boot. However, this change doesn’t break legacy boot. Legacy boot can still use page 0. It just can’t allocate it. But that’s fine, nothing else can either, so it is free for legacy boot code to use.

I agree all memory corruption is important to catch and I am mid upstreaming more memory subsystem fixes, but that also means we should catch the most common kind of corruption, null pointer usage. This is an age old mitigation and we should support it always. That being said, this PR doesn’t do that. It just makes page 0 non-allocatable.

olegilyasov:
> > > I know X86 Legacy Boot needs to allocate zero address. This change will not allow this boot mode. Right?
> > 
> > 
> > Yes, indeed. Legacy boot will be broken after this change. In general, NULL pointer detection is a debugging feature that should not be generic. NULL pointer related problems are as "good" as any other memory corruptions.
> 
> As Gerd noted, edk2 no longer supports legacy boot. However, this change doesn’t break legacy boot. Legacy boot can still use page 0. It just can’t allocate it. But that’s fine, nothing else can either, so it is free for legacy boot code to use.
> 
> I agree all memory corruption is important to catch and I am mid upstreaming more memory subsystem fixes, but that also means we should catch the most common kind of corruption, null pointer usage. This is an age old mitigation and we should support it always. That being said, this PR doesn’t do that. It just makes page 0 non-allocatable.

Null pointer detection implies the CPU exception handler that catches it. If legacy code starts using page 0 without allocating it, it will need its own exception handler for this, right?

os-d:
> Null pointer detection implies the CPU exception handler that catches it. If legacy code starts using page 0 without allocating it, it will need its own exception handler for this, right?

This PR does not change null pointer detection. That is still controlled through the PCD

https://github.com/tianocore/edk2/blob/51d273d8c3dbc36f25f3d2af27bef6e01604d90c/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c#L786-L799.

Legacy boot platforms would set this PCD to 0 and not enable null pointer detection. Then it is mapped RW and it can be written to, no exception handlers.

All this PR does is not let page 0 be allocated, which again is not a prerequisite to writing to it.

olegilyasov:
> > Null pointer detection implies the CPU exception handler that catches it. If legacy code starts using page 0 without allocating it, it will need its own exception handler for this, right?
> 
> This PR does not change null pointer detection. That is still controlled through the PCD
> 
> https://github.com/tianocore/edk2/blob/51d273d8c3dbc36f25f3d2af27bef6e01604d90c/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c#L786-L799
> .
> 
> Legacy boot platforms would set this PCD to 0 and not enable null pointer detection. Then it is mapped RW and it can be written to, no exception handlers.
> 
> All this PR does is not let page 0 be allocated, which again is not a prerequisite to writing to it.

Got it, thank you.

os-d:
@lgao4 the open conversations have been resolved, please re-review.

lgao4:
> @lgao4 the open conversations have been resolved, please re-review.

Thanks for your detail. I have no other comments.

lgao4 (approved):

os-d:
@benjamindoron @ChaselChiu @gdong1 @gguo11837463 @jameslu8 @LinusvPelt @Sean-StarLabs @shuoliu0 any review/comment from the UefiPayloadPkg side? The MdeModulePkg and OvmfPkg maintainers have signed off and the UefiPayloadPkg change is just updating the copied driver from those packages, so hopefully should not be controversial.

PR #11207, opened 2025-06-18
NetworkPkg/WifiConnectionManager: net list scan ui

Zclarkwilliams:
# Description

Wifi Connection Manager no longer scans for available networks when the driver in loaded. Instead the driver must start a scan when appropriate to not block other processes. This adds a one time scan for networks when the user opens the network settings page in the UI, enabled the WiFi Network and then opens the network list page to view available networks.

- [ ] Breaking change?
  - This adds a one time network scan to the UI page change to network list page.
- [ ] Impacts security?
  - No
- [ ] Includes tests?
  - No

## How This Was Tested

This change was testing on a platform for which had the
WifiConnectionManagerDxe driver loaded, Wifi Network enabled
in the BIOS menu. The network list page was opened and the
network list was populated with available networks. The scan
was not performed when the driver was loaded, but only when
the user navigated to the network list page.

## Integration Instructions

Flash new firmware with change included.


SaloniKasbekar (approved):

NetworkPkg/WifiConnectionManagerDxe/WifiConnectionMgrHiiConfigAccess.c
```diff
@@ -1558,6 +1558,10 @@ WifiMgrDxeHiiConfigAccessCallback (
         // User triggered a scan process.
         //
         Private->CurrentNic->OneTimeScanRequest = TRUE;
+        Status = WifiMgrStartScan (Private->CurrentNic);
+        if (EFI_ERROR (Status)) {
+          DEBUG ((DEBUG_WARN, "[WiFi Connection Manager] Error: Failed in start scan for network list with status %r", Status));
+        }
```
SaloniKasbekar, line 1565: leave an empty line between } and break to avoid the uncrustify failure

Zclarkwilliams: added

NetworkPkg/WifiConnectionManagerDxe/WifiConnectionMgrHiiConfigAccess.c
```diff
@@ -1558,6 +1558,11 @@ WifiMgrDxeHiiConfigAccessCallback (
         // User triggered a scan process.
         //
         Private->CurrentNic->OneTimeScanRequest = TRUE;
+        Status = WifiMgrStartScan (Private->CurrentNic);
```
leiflindholm: Commit message:

- missing DCO (Signed-off-by)
- "no longer scans"
  - Why? Because of this commit or something else? (avoid passive voice)
    - If because of this commit, please rewrite without passive voice.
    - If not explain why, since when.

Zclarkwilliams: fixed

SaloniKasbekar:
Update commit message to include missing Signed-off-by (instead of signed-off-by)

NetworkPkg/WifiConnectionManagerDxe/WifiConnectionMgrHiiConfigAccess.c
```diff
@@ -1664,7 +1669,8 @@ WifiMgrDxeHiiConfigAccessCallback (
 
         Status = gBS->SetTimer (Private->CurrentNic->TickTimer, TimerPeriodic, EFI_TIMER_PERIOD_MILLISECONDS (500));
         if (EFI_ERROR (Status)) {
-          gBS->CloseEvent (Private->CurrentNic->TickTimer);
```
leiflindholm: This change, and the identical one below, appears to be a separate bugfix: calling CloseEvent to stop a Timer. (Which I guess actually ended up doing the same thing, but obscurely.) That belongs in a separate commit.
The added debug really ought to also be a separate commit.

Zclarkwilliams:
> Update commit message to include missing Signed-off-by (instead of signed-off-by)

fixed

NetworkPkg/WifiConnectionManagerDxe/WifiConnectionMgrHiiConfigAccess.c
Zclarkwilliams: I've removed the change to use the timer set rather than the cancel event. I will submit another PR bug for the change that will fix the event removal and and ad debugs as that is in the same pocket for the best way to handle the scan mitigation on a timer set issue.

mdkinney (approved):

leiflindholm (approved):

leiflindholm:
@mergify dequeue

leiflindholm:
@mergify requeue

PR #10854, opened 2025-03-13
Adding support for MM Communicate v3

kuqin12:
# Description

As PI spec v1.9 introduced MM communicate v3, this change adds the support for MM communicate v3 in MdePkg, MdeModulePkg, and StandaloneMmPkg.

- [ ] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [x] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

## How This Was Tested

This change is tested on QEMU virtual platform and verified variable service can operate through both v1 and v3 mm communicate protocols.

## Integration Instructions

Locate MM communicate v3 PPI/Protocol, and populate the communication buffer as defined in the spec.

PR #11208, opened 2025-06-18
MdeModulePkg/SmmCore: Modify check for Mm Communicate Buffer without data

sachinami:
# Description

The Mm communication buffer validation logic was slightly modified in #10854 when MM Communicate v3 was introduced.

If the buffer size is compared against the size of EFI_SMM_COMMUNICATE_HEADER it does not take into consideration that the buffer might not have data.

The check has now been changed back to checking for only the header size as it was before.

- [ ] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behaviour?
- [ ] Impacts security?
- [ ] Includes tests?


MdeModulePkg/Core/PiSmmCore/PiSmmCore.c
```diff
@@ -730,12 +731,31 @@ SmmEntryPoint (
...
+        LegacyCommunicateHeader = (EFI_SMM_COMMUNICATE_HEADER *)CommunicationBuffer;
+        CommGuid                = &LegacyCommunicateHeader->HeaderGuid;
+        CommData                = LegacyCommunicateHeader->Data;
+        CommHeaderSize          = OFFSET_OF (EFI_SMM_COMMUNICATE_HEADER, Data);
+      }
+
       //
       // Check for over or underflows
       //
+      IsOverUnderflow = EFI_ERROR (SafeUintnSub (BufferSize, CommHeaderSize, &BufferSize));
```
kuqin12: This line will change the `BufferSize` in place and makes the `SmmIsBufferOutsideSmmValid` only validate the truncated size.

MdeModulePkg/Core/PiSmmCore/PiSmmCore.c
```diff
@@ -730,12 +731,31 @@ SmmEntryPoint (
                        (UINT8 *)gSmmCorePrivate,
                        sizeof (*gSmmCorePrivate)
                        );
+
+      // Find Mm Communicate Header type
+      CommGuid = &((EFI_MM_COMMUNICATE_HEADER_V3 *)CommunicationBuffer)->HeaderGuid;
```
kuqin12: The point of validating the size first it to make sure we at least have 16 bytes to read.

sachinami:
@kuqin12 Thank you for your comments. I have simplified the changes to modify only the buffer size check.
Could you please share your thoughts.

lgao4 (approved):

kuqin12 (approved):

PR #11210, opened 2025-06-19
UefiCpuPkg/PiSmmCpuDxeSmm: Add sync barrier before BSP invokes SmmCoreEntry

xuweiintel:
This patch introduces a synchronization point between BSP and APs immediately after each AP completes SmmCpuPlatformHookBeforeMmiHandler(). Previously, BSP might proceed to the next stage while some APs were still executing the platform hook, leading to a potential race condition if the BSP invokes inter-processor coordination mechanisms (e.g., SmmStartupThisAp ()) too early.

By adding SmmCpuSyncWaitForAPs() in the BSP path and SmmCpuSyncReleaseBsp() in the AP path after the platform hook, this change ensures BSP waits for all APs to complete their initialization before continuing.

The debug comment tags (e.g., /// #6: Signal BSP) are updated accordingly to reflect the correct sequence steps. This improves the determinism and correctness of SMM entry synchronization logic in multi-processor scenarios.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions


xuweiintel:
> according the discussion, we still need one round sync for mtrr confiuration.

The patch is updated, please help review. Thanks a lot

PR #11213, opened 2025-06-20
Linux/GCC host-based unit test fixes

mdkinney:
# Description

Update UnitTestFrameworkPkg and BaseTools to improve support for host-based unit tests in Linux/GCC environments
* Add `-pie` to `ASLCC_FLAGS` for host-based unit tests to address link failure when ACPI table `.aslc`/`.act` files are used.
* Update all uses of `lcov` in the `HostBasedUnitTestRunner` plugin to ignore mismatch errors for `lcov` version >=2.0. This extends an earlier change to all `lcov` invocations.
* Update `GenFw` to skip `R_X86_64_NONE` relocations for unknown symbols and not generate and error message and exit. Linux/GCC host-based unit tests may use shared libraries for features such as `lcov` and `asan` and links to shared libs use relocations of this type that are not an error.
* Update `GoogleTestLib.h` to redefine `NULL` to C++ standard `nullptr`.  `<gtest.h>` uses standard C includes that defines `NULL` to the non-standard `__null` and the GoogleTest matchers see `nullptr` and `_null` as different and simple checks using `NULL` fail. With this redefine, all uses of `NULL` are mapped to `nullptr` and `NULL` can be used in unit test checks and mocks without addition typecasts.
* Update `MdePkg`, `NetworkPkg`, and `DynamicTablesPkg` to use `<Library/GoogleTestLib.h>` instead of `<gtest/gtest.h>` so the edk2 google test extensions are available to unit tests.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

* Build unit tests with `.aslc`/`.act` files that fail to build and verify build success after these changes.
* Build complex unit tests that cause `lcov` to fail with mismatch errors and verify the build succeeds after these changes.
* Verify that `NULL` can be used without extra typecasts in `EXPECT_EQ()` and return values from `EXPECT_CALL()` function mocks.

### DIFF
--- a/NetworkPkg/Dhcp6Dxe/GoogleTest/Dhcp6DxeGoogleTest.cpp
+++ b/NetworkPkg/Dhcp6Dxe/GoogleTest/Dhcp6DxeGoogleTest.cpp
@@ -4,7 +4,7 @@
   Copyright (c) Microsoft Corporation
   SPDX-License-Identifier: BSD-2-Clause-Patent
 **/
-#include <gtest/gtest.h>
+#include <Library/GoogleTestLib.h>
 
 ////////////////////////////////////////////////////////////////////////////////
 // Run the tests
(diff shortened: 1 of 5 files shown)


makubacki (approved):

SaloniKasbekar (approved):

mdkinney:
@mergify rebase

PR #11215, opened 2025-06-20
UefiCpuPkg/MpInitLib: Ensure AP wake up on WakeUpByInitSipiSipi mode

philnoh2:
If AP loop mode is not Halt loop, it is found that the AP in ApInitDone state exits the loop mode immediately when WAKEUP_AP_SIGNAL is signaled at the first step of AP wake up. When WakeUpByInitSipiSipi is enabled, it is possible that the AP could clear StartupApSignal buffer before the BSP sends Init-Startup IPI to the AP. In the case, it causes that the BSP frees AP reset vector buffer even though AP wake up is in process. This results in an exception error. To ensure AP wake up, it is expected that the AP is in the known non-executable state before the AP wakeup signal.

The issue happens when the AP is disabled through the MP service of gEfiMpServiceProtocolGuid->EnableDisableAP(FALSE) and enabled later through gEfiMpServiceProtocolGuid->EnableDisableAP(TRUE). In the flow, the AP that in MINOTOR-MWAIT-Loop or Run-Loop can respond to the WAKEUP_AP_SIGNAL before the BSP sends the INIT-SIPI-SIPI.

 For example, the related code in WakeUpAp function is as follow.

```c
WakeUpAP ()
{
  ...
  } else {
    ...
    //
    // Wakeup specified AP
    //
    ASSERT (CpuMpData->InitFlag == ApInitDone);
    *(UINT32 *)CpuData->StartupApSignal = WAKEUP_AP_SIGNAL;
    if (ResetVectorRequired) {
      ...
      } else {
        SendInitSipiSipi (
          CpuInfoInHob[ProcessorNumber].ApicId,
          (UINT32)ExchangeInfo->BufferStart
          );
      }
    }

    //
    // Wait specified AP waken up
    //
    WaitApWakeup (CpuData->StartupApSignal);
  }

  if (ResetVectorRequired) {
    FreeResetVector (CpuMpData);
  }
  ...
}
```
- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

The update has been verified on AMD platforms. We met the case/issue when testing an MP test tool for EFI shell. The tool enables and disables APs sequentially through gEfiMpServiceProtocolGuid->EnableDisableAP service. The error was detected by the stress test (e.g. 10000 times) in the tool. It fixes the issue.

## Integration Instructions

N/A

### DIFF
--- a/UefiCpuPkg/Library/MpInitLib/MpLib.c
+++ b/UefiCpuPkg/Library/MpInitLib/MpLib.c
@@ -2,7 +2,7 @@
   CPU MP Initialize Library common functions.
 
   Copyright (c) 2016 - 2024, Intel Corporation. All rights reserved.<BR>
-  Copyright (c) 2020 - 2024, AMD Inc. All rights reserved.<BR>
+  Copyright (C) 2020 - 2025 Advanced Micro Devices, Inc. All rights reserved.<BR>
 
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
@@ -1663,10 +1663,24 @@ ResetProcessorToIdleState (
 
   CpuMpData = GetCpuMpData ();
 
-  CpuMpData->WakeUpByInitSipiSipi = TRUE;
   if (CpuMpData == NULL) {
     DEBUG ((DEBUG_ERROR, "[%a] - Failed to get CpuMpData.  Aborting the AP reset to idle.\n", __func__));
     return;
+  }
+
+  CpuMpData->WakeUpByInitSipiSipi = TRUE;
+
+  //
+  // "CpuMpData->WakeUpByInitSipiSipi = TRUE" requires that the AP must be in Halt loop.
+  // If AP loop mode is not Halt loop, make sure that the AP is in the known non-executable state.
+  //
+  // AP could be in MONITOR/MWAIT state that could wake up the AP when setting WAKEUP_AP_SIGNAL in WakeUpAp().
+  // And then SendInitSipiSipi() in WakeupAp() could wake up the AP again to run the AP Wakeup vector.
+  // But the AP wakeup vector could be freed by the BSP while the AP is running it.
+  //
+  if (CpuMpData->ApLoopMode != ApInHltLoop) {
+    SendInitIpi (((CPU_INFO_IN_HOB *)(UINTN)CpuMpData->CpuInfoInHob)[ProcessorNumber].ApicId);
+    MicroSecondDelay (PcdGet32 (PcdCpuInitIpiDelayInMicroSeconds));
   }
 
   WakeUpAP (CpuMpData, FALSE, ProcessorNumber, NULL, NULL, TRUE);
--- a/UefiCpuPkg/Library/MpInitLib/DxeMpInitLib.inf
+++ b/UefiCpuPkg/Library/MpInitLib/DxeMpInitLib.inf
@@ -3,6 +3,7 @@
 #
 #  Copyright (c) 2016 - 2023, Intel Corporation. All rights reserved.<BR>
 #  Copyright (c) 2024, Loongson Technology Corporation Limited. All rights reserved.<BR>
+#  Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.<BR>
 #  SPDX-License-Identifier: BSD-2-Clause-Patent
 #
 ##
@@ -100,3 +101,4 @@
   gUefiCpuPkgTokenSpaceGuid.PcdGhcbHypervisorFeatures                  ## CONSUMES
   gUefiCpuPkgTokenSpaceGuid.PcdSevEsWorkAreaBase                       ## SOMETIMES_CONSUMES
   gUefiCpuPkgTokenSpaceGuid.PcdFirstTimeWakeUpAPsBySipi                ## CONSUMES
+  gUefiCpuPkgTokenSpaceGuid.PcdCpuInitIpiDelayInMicroSeconds           ## SOMETIMES_CONSUMES
--- a/UefiCpuPkg/Library/MpInitLib/PeiMpInitLib.inf
+++ b/UefiCpuPkg/Library/MpInitLib/PeiMpInitLib.inf
@@ -3,6 +3,7 @@
 #
 #  Copyright (c) 2016 - 2024, Intel Corporation. All rights reserved.<BR>
 #  Copyright (c) 2024, Loongson Technology Corporation Limited. All rights reserved.<BR>
+#  Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.<BR>
 #  SPDX-License-Identifier: BSD-2-Clause-Patent
 #
 ##
@@ -83,6 +84,7 @@
   gUefiCpuPkgTokenSpaceGuid.PcdSevEsWorkAreaBase                   ## SOMETIMES_CONSUMES
   gUefiCpuPkgTokenSpaceGuid.PcdGhcbHypervisorFeatures              ## CONSUMES
   gUefiCpuPkgTokenSpaceGuid.PcdFirstTimeWakeUpAPsBySipi            ## CONSUMES
+  gUefiCpuPkgTokenSpaceGuid.PcdCpuInitIpiDelayInMicroSeconds       ## SOMETIMES_CONSUMES
 
 [Ppis]
   gEdkiiPeiShadowMicrocodePpiGuid        ## SOMETIMES_CONSUMES


LiuZhiguang001:
I agree this is an issue, thanks for identifying it.
However, sending Init to AP is too heavy. 
How about moving the line  "*(UINT32 *)CpuData->StartupApSignal = WAKEUP_AP_SIGNAL;" after InitSipiSipi and before "WaitApWakeup (CpuData->StartupApSignal);"

philnoh2:
> I agree this is an issue, thanks for identifying it. However, sending Init to AP is too heavy. How about moving the line "*(UINT32 *)CpuData->StartupApSignal = WAKEUP_AP_SIGNAL;" after InitSipiSipi and before "WaitApWakeup (CpuData->StartupApSignal);"

Thank you for your review and suggestion. Previously we reviewed the suggestion. At that time, we thought that the way wouldn't be applicable because ApWakeupFunction function on the AP (InitSipiSipi) would reach the code that clears StartupApSignal buffer, before the BSP sets WAKEUP_AP_SIGNAL to the buffer. It is found that the function clears the buffer in the middle where is before invoking the AP function. In the case, the AP would not take a chance to clear the buffer in the function. Then, would result in an infinite loop in WaitApWakeup function. It looks like setting WAKEUP_AP_SIGNAL before InitSipiSipi is required on the current code flow for AP wake up. 

For the PR update, we considered that currently ResetProcessorToIdleState function is limitedly called in the following cases. We thought that sending Init IPI to AP in the function could have the limited impact.

CheckThisAP/CheckAllAPs functions : In case timeout is raised in the functions (as a failure)
EnableDisableApWorker : In case EnableDisableAP service (PPI, Protocol) is called (e.g. UefiCpuPkg UnitTest calls it)

LiuZhiguang001:
> > I agree this is an issue, thanks for identifying it. However, sending Init to AP is too heavy. How about moving the line "*(UINT32 *)CpuData->StartupApSignal = WAKEUP_AP_SIGNAL;" after InitSipiSipi and before "WaitApWakeup (CpuData->StartupApSignal);"
> 
> Thank you for your review and suggestion. Previously we reviewed the suggestion. At that time, we thought that the way wouldn't be applicable because ApWakeupFunction function on the AP (InitSipiSipi) would reach the code that clears StartupApSignal buffer, before the BSP sets WAKEUP_AP_SIGNAL to the buffer. It is found that the function clears the buffer in the middle where is before invoking the AP function. In the case, the AP would not take a chance to clear the buffer in the function. Then, would result in an infinite loop in WaitApWakeup function. It looks like setting WAKEUP_AP_SIGNAL before InitSipiSipi is required on the current code flow for AP wake up.
> 
> For the PR update, we considered that currently ResetProcessorToIdleState function is limitedly called in the following cases. We thought that sending Init IPI to AP in the function could have the limited impact.
> 
> CheckThisAP/CheckAllAPs functions : In case timeout is raised in the functions (as a failure) EnableDisableApWorker : In case EnableDisableAP service (PPI, Protocol) is called (e.g. UefiCpuPkg UnitTest calls it)

Thanks for the detailed explanation. You are right; my suggestion also has issue. :)

philnoh2:
Hi EDK2 Maintainers (@jiaxinwu, @kraxel, @LiuZhiguang001, @lzeng14, @niruiyu, @rahul1-kumar, @td36),

In addition to the first review, if possible, could you please help to review the PR additionally that was summitted on June 20 ?
If it looks good for merge, let me expect its approval together.

Thank you.

UefiCpuPkg/Library/MpInitLib/MpLib.c
```diff
@@ -1660,13 +1660,22 @@ ResetProcessorToIdleState (
...
     return;
   }
 
+  CpuMpData->WakeUpByInitSipiSipi = TRUE;
+
+  //
+  // "CpuMpData->WakeUpByInitSipiSipi = TRUE" requires that the AP must be in Halt loop.
+  // If AP loop mode is not Halt loop, make sure that the AP is in the known non-executable state.
+  //
+  if ((CpuMpData->InitFlag == ApInitDone) && (CpuMpData->ApLoopMode != ApInHltLoop)) {
```
niruiyu: 1. Can you assert (InitFlag == ApInitDone) instead of adding a check?
2. Can you update the comments to explain in more detail? such as AP could be in MONITOR/MWAIT state that setting WakeUpSignal in WakeUpAp() could wake up the AP to clear the WakeupSignal and then SendInitSipi() in WakeUpAp() could wake up AP again to run the AP Wakeup vector but the AP wakeup vector could be freed by BSP during AP is running it.

philnoh2: Thank you for your review.

For 1>, I understand. In the current EDK2, I can check that ResetProcessorToIdleState() is called from 3 functions (CheckAllAPs, CheckThisAP, EnableDisableApWorker - protocol service) after ApInitDone state. Based on that, I think that basically there would not be the ASSERT failure condition. But as it is a library function, ASSERT would make ResetProcessorToIdleState() not callable in the init stage, which may not the case. In case it is called in the init stage, the updated code below is not necessary and cause delay. Therefore, if () might be better than ASSERT and it is not changing the original logic. 

_SendInitIpi (((CPU_INFO_IN_HOB *)(UINTN)CpuMpData->CpuInfoInHob)[ProcessorNumber].ApicId);
MicroSecondDelay (PcdGet32 (PcdCpuInitIpiDelayInMicroSeconds));_

For 2>,  I agree. Let me update the PR to add more detail to the comment. Before that, let me take some time to check whether you have additional feedback regarding above comment on 1>.

philnoh2: The PR has been updated for the comment. I followed up the example comment.

philnoh2:
Hi EDK2 Maintainers (@jiaxinwu, @kraxel, @LiuZhiguang001, @lzeng14, @niruiyu, @rahul1-kumar, @td36),

Recently (7/10) the PR has been updated to support the latest review. 
Could you please help to review and approve it ?

Thank you.

UefiCpuPkg/Library/MpInitLib/MpLib.c
LiuZhiguang001: This if condition is a little confusing
`if ((CpuMpData->InitFlag == ApInitDone) && (CpuMpData->ApLoopMode != ApInHltLoop))`
It makes people think what if the InitFlag is not ApInitDone. Actually, we should not call it when InitFlag is not ApInitDone.
You said, "In case it is called in the init stage,", but I don't think we should consider such case

philnoh2: I understand. As ResetProcessorToIdleState() is a library function, I thought that it could better for the update to focus on the error condition more specifically. But if the code (InitFlag check) causes a confusion, I think that removing it would be better, based on "we should not call it when InitFlag is not ApInitDone". Let me follow up the point. I have updated the PR again to remove the code. Could you review it again ? Thank you.

philnoh2:
Hi EDK2 Maintainers (@jiaxinwu, @kraxel, @LiuZhiguang001, @lzeng14, @niruiyu, @rahul1-kumar, @td36),

Recently (8/1), the PR has been updated to support the latest review.
Could you please help to review it again ?

Thank you.

philnoh2:
Hi EDK2 Maintainers (+ Liming (@lgao4) who updated the issue report, [11394]( https://github.com/tianocore/edk2/issues/11394))

Since the PR was lastly updated for the latest feedback on 8/1, it looks like that the PR has been a pending status.
Could you please help to review the PR again ? If it is good, let me expect getting it approved together.

Thank you.

lgao4:
> Hi EDK2 Maintainers (+ Liming (@lgao4) who updated the issue report, [11394](https://github.com/tianocore/edk2/issues/11394))
> 
> Since the PR was lastly updated for the latest feedback on 8/1, it looks like that the PR has been a pending status. Could you please help to review the PR again ? If it is good, let me expect getting it approved together.
> 
> Thank you.

There is no other feedback. I also have no comment. Seemly, this change is the good fix. If UefiPkg maitainer has no objection, I agree to merge it.

lgao4 (approved):

philnoh2:
> Can you refine the commit message to make it clearer? As the code change is in the Re-enable AP path but the commit message does not mention the DisableAp/EnableAp at all.
> 
> Can I assume that the issue only happens when AP is disabled through EnableDisableAP(FALSE) and enabled later through EnableDisableAP(TRUE)? In that flow, AP is in MINOTOR-MWAIT-Loop or Run-Loop which can respond to the WAKEUP_AP_SIGNAL before BSP sending INIT-SIPI-SIPI.

I understand. Your assumption is right. I have updated the commit message based on that. Please review the latest update.

niruiyu (approved):

PR #11225, opened 2025-06-24
Revert "UefiCpuPkg/PiSmmCpuDxeSmm: Add sync barrier before BSP invokes SmmCoreEntry"

leiflindholm:
PR #11210 introduced Commit bbee92c9af59,
("UefiCpuPkg/PiSmmCpuDxeSmm: Add sync barrier before BSP invokes SmmCoreEntry") which causes a boot failure in OVMF_X64_MMDEBUG targets, breaking upstream CI.

Revert it.


mdkinney:
I agree with the need to revert this change for edk2 CI failures in OVMF.

@xuweiintel and @niruiyu please review the OVMF failures from this commit and submit a new PR to address the race condition.

PR #11232, opened 2025-06-25
Enable EDK2 to respect hypervisor PCI SBDF assignments

myrade:
# Enable EDK2 to respect hypervisor PCI SBDF assignments

This adds additional logic to relevant PCI enumeration and root bridge configuration functions to enable firmware to respect hypervisor's PCI SBDF assignments. Without this, firmware will always override prior assignments with its own algorithm. This makes no changes to the outcome of PCI root bridge setup and enumeration in the event that hypervisor does not make any PCI SBDF assignments.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested both with and without a custom GCP hypervisor that makes SBDF assignments on VMs with multiple root bridges.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Bus/Pci/PciBusDxe/PciLib.c
+++ b/MdeModulePkg/Bus/Pci/PciBusDxe/PciLib.c
@@ -1096,6 +1096,7 @@ PciScanBus (
   UINT16                             Register;
   UINTN                              HpIndex;
   PCI_IO_DEVICE                      *PciDevice;
+  EFI_PCI_IO_PROTOCOL                *PciDeviceIo;
   EFI_EVENT                          Event;
   EFI_HPC_STATE                      State;
   UINT64                             PciAddress;
@@ -1293,9 +1294,22 @@ PciScanBus (
           }
         }
 
-        Status = PciAllocateBusNumber (Bridge, *SubBusNumber, 1, SubBusNumber);
-        if (EFI_ERROR (Status)) {
-          return Status;
+        PciDeviceIo = &PciDevice->PciIo;
+        Status      = PciDeviceIo->Pci.Read (
+                                         PciDeviceIo,
+                                         EfiPciIoWidthUint8,
+                                         PCI_BRIDGE_SECONDARY_BUS_REGISTER_OFFSET,
+                                         1,
+                                         &SecondBus
+                                         );
+
+        if (!EFI_ERROR (Status) && (SecondBus != 0)) {
+          *SubBusNumber = SecondBus;
+        } else {
+          Status = PciAllocateBusNumber (Bridge, *SubBusNumber, 1, SubBusNumber);
+          if (EFI_ERROR (Status)) {
+            return Status;
+          }
         }
 
         SecondBus = *SubBusNumber;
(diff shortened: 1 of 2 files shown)


OvmfPkg/Library/PciHostBridgeUtilityLib/PciHostBridgeUtilityLib.c
```diff
@@ -366,6 +408,14 @@ PciHostBridgeUtilityGetRootBridgesBusScan (
       ++Initialized;
       LastRootBridgeNumber = RootBridgeNumber;
     }
+
+    //
+    // If there is already a subordinate bus, defer to it.
+    //
+    SubordinateBus = GetSubordinateBusOfPciBridge (RootBridgeNumber, Device);
+    if ((SubordinateBus != 0) && (SubordinateBus > RootBridgeNumber)) {
```
kraxel, line 416: This looks wrong, the function scans for additional /root/ busses (`qemu -device pxb-pcie`)..

MdeModulePkg/Bus/Pci/PciBusDxe/PciLib.c
```diff
@@ -1293,9 +1294,22 @@ PciScanBus (
...
+                                         );
+
+        if (!EFI_ERROR (Status) && (SecondBus != 0)) {
+          *SubBusNumber = SecondBus;
+        } else {
+          Status = PciAllocateBusNumber (Bridge, *SubBusNumber, 1, SubBusNumber);
+          if (EFI_ERROR (Status)) {
+            return Status;
+          }
         }
```
kraxel, line 1313: Hmm, I think that will only work if either all or none of the pci bridges have the SubordinateBus configured.
To make a mixed setup work properly I think you have to do an extra pass which takes the SubordinateBus-configured bus numbers out of the pool used by PciAllocateBusNumber().

OvmfPkg/Library/PciHostBridgeUtilityLib/PciHostBridgeUtilityLib.c
myrade: Ah, I see the error. Looks like I failed to notice this because I actually did have a root bus on bus 1.

myrade: Scratch my last comment. I'm not familiar with how QEMU works here, do you specifically anticipate this change being disruptive? When we don't set up any PCI devices in the hypervisor before executing firmware this works the same as before, ie. only root bridges respond with a Vendor ID and get identified by `PciHostBridgeUtilityGetRootBridgesBusScan` as root bridges.

When we instead use a custom hypervisor to set up the PCI topology (in the interest of pre-assigning SBDFs) then `PciHostBridgeUtilityGetRootBridgesBusScan` will need to actually poll every device for a subordinate bus setting since all PCI devices are now responding with Vendor ID. If that's incompatible with the parameters of OvmfPkg I understand, but I'm not sure it actually disrupts QEMU's pxb pattern.

kraxel: Secondary busses are handled with standard pci driver enumeration (in MdeModulePkg/Bus/Pci/...).  This code will find additional root busses.  I don't think you have to change anything here.

ardbiesheuvel:
Could this platform use the light enumerator instead?

ardbiesheuvel:
Closing for now - please reopen/resubmit when you have the bandwidth to revisit this issue.

PR #11233, opened 2025-06-26
EmbeddedPkg/TimeBaseLib: aligning year with UEFI specification

psingh3086:
EmbeddedPkg/TimeBaseLib: aligning year with UEFI specification
According to the UEFI Specification version 2.11, the valid range for the Year field in the EFI_TIME structure is from 1900 to 9999. Currently IsTimeValid() checks a restricted range 2000 - 2099. 
 
Update range in TimeBaseLib.c to match UEFI specification.

## How This Was Tested
 
Run EDK2-test(SCT) suite to verify code change.
 
https://github.com/tianocore/edk2-test/pull/267

### DIFF
--- a/EmbeddedPkg/Library/TimeBaseLib/TimeBaseLib.c
+++ b/EmbeddedPkg/Library/TimeBaseLib/TimeBaseLib.c
@@ -271,8 +271,8 @@ IsTimeValid (
   )
 {
   // Check the input parameters are within the range specified by UEFI
-  if ((Time->Year  < 2000)              ||
-      (Time->Year   > 2099)              ||
+  if ((Time->Year  < 1900)              ||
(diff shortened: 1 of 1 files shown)


leiflindholm (approved):
We seem to have ended up here because TimeBaseLib was originally broken out of the PL031 driver and so these helper functions were working around the constraints of that design.

However, the hardware constraints should be worked around in the device-specific code, not the helper library.

Since this is a relaxation compared to previous behaviour and since behaviour will be identical for years 2000-2099, I don't think we need to panic about updating existing RealTimeClockLib implementations first.

@ardbiesheuvel seems to be out this week, but we discussed this last week, so I'll merge this one shortly unless someone screams.

PR #11238, opened 2025-06-30
UefiCpuPkg/PiSmmCpuDxeSmm: Add sync barrier before BSP invokes SmmCor…

xuweiintel:
…eEntry

This patch introduces a synchronization point between the BSP and APs to ensure all APs have entered their SMM wait-loop (while (TRUE) in APHandler ()) before the BSP calls into the SMI handler logic via gSmmCpuPrivate ->SmmCoreEntry().

Previously, the BSP would invoke ReleaseAllAPs() and immediately proceed to SmmCoreEntry() without confirming whether APs had reached the stable waiting state. If SmmStartupThisAp() was called inside the SMI handler shortly after ReleaseAllAPs(), it might lead to a race condition: APs are issued two consecutive wait signals (SmmCpuSyncWaitForBsp()). BSP sends two consecutive releases (ReleaseAllAPs() + SmmStartupThisAp()) If an AP has not yet responded to the first release, the second release may overwrite the semaphore state, and the AP might miss the notification, causing it to hang or behave unpredictably.

To address this:
A SmmCpuSyncWaitForAPs() is added in BSP after
mmCpuPlatformHookBeforeMmiHandler() and before entering SmmCoreEntry(). A matching SmmCpuSyncReleaseBsp() is added in AP immediately after its own SmmCpuPlatformHookBeforeMmiHandler()

This ensures that BSP does not enter SMI handler logic or dispatch any AP-related requests before all APs are confirmed to be idle and ready. Debug sync point markers (e.g., /// #6, #7) are updated accordingly. This change eliminates a subtle but critical race condition in multi-processor/multi-socket systems during SMM entry and improves overall synchronization safety.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions


jiaxinwu (approved):

lzeng14 (approved):

mdkinney:
Can you provide an analysis of why the OvmfPkg failed with the previous version of this change and what was modified in this version to address that failure?

jiaxinwu:
> Can you provide an analysis of why the OvmfPkg failed with the previous version of this change and what was modified in this version to address that failure?

let me explain the reason:

Relax mode is default enabled in ovmf. In relax mode, when waiting for all APs for the first time(#7: Wait APs), the apcount is invalid, causing the code to continue running. In SmmCoreEntry, when starting all APs, the APs have not yet reached the present state when setting busy (line 1326 in MpService.c), so busy is not set. Later, when ReleaseAllAPs is called (line 1362 in MpService.c), the APs become present, and as they continue running, they find that busy is not set, which leads to an assert (line 892 in MpService.c).
image

according the SmmCpuSyncWaitForAPs(), the caller shall make sure the NumberOfAPs have already checked-in to avoid the undefined behavior.
image

So, the fix is to move the new added sync flow into the if condition to make sure NumberOfAPs is correct:
if ((SyncMode == MmCpuSyncModeTradition) || SmmCpuFeaturesNeedConfigureMtrrs ()) {
}

lzeng14:
@mdkinney Could you help add the push tag ?

PR #11239, opened 2025-06-30
OvmfPkg/PlatformDebugLibIoPort: Add check for MemDebugLogWrite

luigix25:
This check is present for every call to `MemDebugLogWrite` but it is missing here. This may cause an unwanted write to the buffer, when the buffer's address is not null and MemDebug is disabled.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions

### DIFF
--- a/OvmfPkg/Library/PlatformDebugLibIoPort/DebugLib.c
+++ b/OvmfPkg/Library/PlatformDebugLibIoPort/DebugLib.c
@@ -235,7 +235,9 @@ DebugAssert (
   //
   // Send the string to Memory Debug Log
   //
-  MemDebugLogWrite (Buffer, Length);
+  if (MemDebugLogEnabled ()) {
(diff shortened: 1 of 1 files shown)


kraxel (approved):

ajyoung-oracle:
This change is technically not needed (i.e. it's not a BUG) because the MemDebugLogNull implementation of MemDebugLogWrite() is a NOP. The only reason we added MemDebugLogEnabled() is it to allow DebugPrintMarker() to be optimized to bypass the string formatting. But, it doesn't hurt to add the extra call in DebugAssert() as well. Just FYI...

luigix25:
Thanks for the information! I guess I have to change the commit message. I'll leave it up to the maintainers to decide whether or not to merge this.

PR #29, opened 2015-11-04
BaseTools/Source/C/VfrCompile/Pccts/dlg honor CC if already set

AelMalinka:

Issue #28, opened 2015-10-30
Win32 Directory is missing for BaseTools Bin for the nt32 switch


PaulSelles:
Hi,

The github edk2 repository is missing the BaseTools/Bin/Win32 directory. This directory exists int he svn repository and need to be ported over to github in order for edksetup.bat to work with the nt32 swtich.

A edksetup.bat call with the nt32 switch is called, the nt32 switch is fowarded to BaseTools/toolsetup.bat script. Within toolsetup.bat, the script check for the existence of  %EDK_TOOLS_PATH%\Bin\Win32 (BaseTools/Bin/Win32) to set the EDK_TOOLS_BIN. Since the BaseTools/Bin/Win32 directory is not present in the github repository the setup script fails. Copying over the Win32 directory from the svn repository fixes this issue.

Thanks,
Paul

PaulSelles:
Hi,

Since the BaseTools/Bin/Win32 files are available in [tianocore](https://github.com/tianocore)/[https://github.com/tianocore/edk2-BaseTools-win32](https://github.com/tianocore/edk2-BaseTools-win32), they were added as a submodule to [VS2015 Support #27](https://github.com/tianocore/edk2/pull/27).

Closing issue.


Thanks,
Paul

PR #11240, opened 2025-06-30
Add support for ARM GICv5

sarah-walker-arm:
# Description

Add support for the ARM GICv5 interrupt controller. This pull request includes the changes to the ArmGicDxe driver, supporting ArmLib changes to detect GICv5 presence, and an update to the TimerDxe driver to support the GICv5 architected PPI numbers.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Boot FVP VersatileExpress model with both GICv3 and GICv5 configurations. Confirmed boot to Linux with appropriate DTB.

## Integration Instructions

N/A


sarah-walker-arm:
The VersatileExpress platform integration is at https://github.com/tianocore/edk2-platforms/pull/807

leiflindholm:
Note: This PR depends on features only documented in https://developer.arm.com/documentation/aes0070/latest/

MdePkg/Include/AArch64/AArch64.h
```diff
@@ -134,6 +137,10 @@
 // so we need to define it here
 #define RNDR  S3_3_C2_C4_0
 
+// The ID_AA64PFR2_EL1 register was added in ARMv8.7 (TODO verify). Since we
```
leiflindholm: Arguably, the longer explanations for which version they were added no longer matter, so we could probably group all of the ID_AA64xxFR_ #defines under a single comment header saying "Definitions for ID registers introduced post ARMv8.0 and not given symbolic
names in all relevant assemblers".

ArmPkg/Drivers/ArmGicDxe/GicV5/AArch64/ArmGicV5.S
```diff
@@ -0,0 +1,191 @@
+#
+#  Copyright (c) 2025, ARM Limited. All rights reserved.
+#
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+#
+
+#include <AsmMacroLib.h>
+
+#if !defined(__clang__)
```
leiflindholm: I don't think we need to worry about clang 3.5 building a GICv5 driver. Could we drop this whole ifdef block?
(But if we did, I'd expect it to not know the symbolic register names anyway.)

ArmPkg/Drivers/ArmGicDxe/GicV5/AArch64/ArmGicV5.S
```diff
@@ -0,0 +1,191 @@
+#
+#  Copyright (c) 2025, ARM Limited. All rights reserved.
```
leiflindholm, line 2: Maybe take inventory that copyright lines for all added files match?

ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c
```diff
@@ -0,0 +1,602 @@
...
+#include "ArmGicDxe.h"
+#include "ArmGicV5.h"
+#include <Library/TimerLib.h>
+
+STATIC UINTN                       mGicIrsConfigFrameBase;
+STATIC UINT32                      SpiRange;
+STATIC UINTN                       mGicNumInterrupts;
+STATIC HARDWARE_INTERRUPT_HANDLER  *mRegisteredInterruptHandlers;
+
+#define GIC_V5_IRS_CONFIG_FRAME_SIZE  SIZE_64KB
```
leiflindholm: Could we move all these #defines to ArmGicV5.h?

ArmPkg/Drivers/ArmGicDxe/ArmGicDxe.c
```diff
@@ -74,7 +74,14 @@ InterruptDxeInitialize (
 {
   EFI_STATUS  Status;
 
-  if (!GicV3Supported ()) {
+  if (ArmHasGicV5SystemRegisters ()) {
+ #ifdef MDE_CPU_AARCH64
```
leiflindholm: I think we should trust ArmHasGicV5SystemRegisters(). No need for additional filtering.

ArmPkg/Drivers/ArmGicDxe/ArmGicDxe.inf
```diff
@@ -47,6 +49,7 @@
   IoLib
   PcdLib
   UefiLib
+  TimerLib
```
leiflindholm, line 52: If you move that up one line we at least have four lines in alphabetical order.

ArmPkg/Drivers/ArmGicDxe/ArmGicV5Dxe.inf
```diff
@@ -0,0 +1,55 @@
...
+  ArmGicCommonDxe.c
+  ArmGicDxe.c
+  ArmGicDxe.h
+  GicV5/ArmGicV5Dxe.c
+
+[Sources.AARCH64]
+  GicV5/AArch64/ArmGicV5.S
+
+[Packages]
+  MdePkg/MdePkg.dec
```
leiflindholm: Please sort packages alphabetically.

ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c
```diff
@@ -0,0 +1,602 @@
...
+
+Abstract:
+
+  Driver implementing the GICv5 interrupt controller protocol
+
+--*/
+
+#include "ArmGicDxe.h"
+#include "ArmGicV5.h"
+#include <Library/TimerLib.h>
```
leiflindholm: There seems to be some includes missing. (For clarity, source files should include the headers they depend on.)

ArmPkg/Drivers/ArmGicDxe/GicV5/AArch64/ArmGicV5.S
sarah-walker-arm: What exactly is mismatched?

leiflindholm: Within this commit, I see
```
#  Copyright (c) 2025, ARM Ltd. All rights reserved.<BR>
#  Copyright (c) 2025, ARM Limited. All rights reserved.     (also missing <BR>)
*  Copyright (c) 2025, Arm Limited. All rights reserved.<BR>

```
I suspect the last one would be the current preferred one by Arm marketing, but my knowledge is a few years out of date.

samimujawar: I can confirm the last one is preferred.

ArmPkg/Drivers/ArmGicDxe/ArmGicDxe.c
sarah-walker-arm: I restored this check, as otherwise AArch32 builds fail with unresolved symbols.

leiflindholm: OK, but then it makes no sense to implement the function for AArch32 in the previous commit.

I think we may have some legacy to untangle.
Hmm, I wonder what the cleanest way to slice this would be.

So, scanning through the previous commits:
Right, I see ArmHasGicV5SystemRegisters () is added adjacent to
ArmHasGicSystemRegisters (), which is why you added it for AArch32.
But actually, the former means "do we support GicV5" whereas the
latter means "please tell me we're not an acient crippled implementation from around the early days of GIC supporting ".

Whereas here, rather than "do we have GicV5 system registers" what we really want to know is "is our Gic a Gicv5?", i.e. GicV5Supported ().

So how about we:
- Drop the ArmHasGicV5SystemRegisters () functions from the preceding commit.
- Move GicV3Supported () to ArmGicCommonDxe.c.
- Clone this file (ArmGicV5Dxe.c?) and leave the original untouched.
- For the cloned file
  - Implement a GicV5Supported function (identical to the dropped AArch64
    ArmHasGicV5SystemRegisters ()), like the GicV3Supported () function.
  - Drop the GicV2DxeInitialize call.
- Use ArmGicV5Dxe.c instead of ArmGicDxe.c in ArmGicV5Dxe.inf.
?

sarah-walker-arm: Reading your suggestions, that would seem to end up with the result of ArmGicDxe, which to my knowledge is intended to support _any_ GIC, not supporting v5, with v5 support being limited to the v5-only ArmGicV5Dxe. Is that what you want?

leiflindholm: Fair point. (Although as I meant it, it would still be supporting GICv2-v4.)

I guess my take is that ArmGicDxe was created to support all GICs available at the point it was created. I.e. v2-v4. Which was very useful - especially for development: being able to build a single image and then test it against qemu with different modelled GICs with a simple command line change.

But ultimately GICv2, and the memory mapped cpu register interfaces, are very much legacy technology.

I don't know if the single v2-v4 driver is outdated enough that it needs deleting, but I don't see a value in turning it into a v2-v5 driver. A separate v3-v5 could well also make sense.

Of course, others might feel differently. (@ardbiesheuvel should be back tomorrow)

MdePkg/Include/Library/ArmLib.h
```diff
@@ -699,6 +699,18 @@ ArmHasGicSystemRegisters (
   VOID
   );
 
+/**
```
mgamal, line 702: I'd suggest moving this function declaration to the next patch (ArmPkg: ArmLib: Add functions to read AA64PFR2 and check GICv5 support)

sarah-walker-arm: The reason it's here is to avoid altering two packages in a single commit.

ArmPkg/Include/Library/ArmGicLib.h
```diff
@@ -166,11 +166,24 @@
 // Bit Mask for
 #define ARM_GIC_ICCIAR_ACKINTID  0x3FF
 
+//
+// GIC SPI and extended SPI ranges
```
mgamal, line 170: Are these defines GICv5-specific or generic for all GIC versions? If it's the former then I guess it'd be better to name them with ARM_GICV5_* prefix?

sarah-walker-arm: These are for GIC pre-v5.

MdePkg/Include/Library/ArmLib.h
mgamal: I see. Fair enough.

ArmPkg/ArmPkg.dec
```diff
@@ -299,6 +299,13 @@
   gArmTokenSpaceGuid.PcdArmArchTimerVirtIntrNum|27|UINT32|0x00000041
   gArmTokenSpaceGuid.PcdArmArchTimerHypVirtIntrNum|28|UINT32|0x0000004A
 
+  # ARM Architectural Timer Interrupt(GICv5 PPI) numbers
+  gArmTokenSpaceGuid.PcdArmArchTimerSecIntrNumV5|0x2000001d|UINT32|0x000004B
+  gArmTokenSpaceGuid.PcdArmArchTimerIntrNumV5|0x2000001e|UINT32|0x0000004C
+  gArmTokenSpaceGuid.PcdArmArchTimerHypIntrNumV5|0x2000001a|UINT32|0x0000004D
+  gArmTokenSpaceGuid.PcdArmArchTimerVirtIntrNumV5|0x2000001b|UINT32|0x0000004E
+  gArmTokenSpaceGuid.PcdArmArchTimerHypVirtIntrNumV5|0x2000001c|UINT32|0x0000004F
+
```
ardbiesheuvel: This needs an explanation why it is needed. I would expect that a given system either supports GICv5 or it does not, and so it is not at all clear to me why we need an entire set of PCDs to describe the exact same thing.

If this is needed to accommodate virtual platforms where GICv5 support is configurable, and the PPI assignment changes as a result, I would expect the platform code to take care of this, so it does not clutter up the generic support code.

ArmPkg/Drivers/ArmGicDxe/ArmGicDxe.c
ardbiesheuvel: Why do we need a separate driver for this? Can we just incorporate these changes into the GICv3 driver instead? Then, whether or not the combined v2/v3 driver also gets this functionality is up for debate, and I'd lean towards whichever is the easiest to accomplish (and leaving AArch32 behind in this sense is perfectly fine)

ArmPkg/ArmPkg.dec
sarah-walker-arm: The ARM architectural timer has specific interrupt numbers that are architecturally defined by the GIC. GICv5 uses slightly different numbers to previous GICs, as the GICv5 interrupt ID scheme now explicitly includes the interrupt type.

There should be no need for anything adding to the platform code for this; either the platform is using GICv5, and therefore should use the v5 PPIs, or it isn't, and it should use the older ones.

ArmPkg/Drivers/ArmGicDxe/ArmGicDxe.c
sarah-walker-arm: This change already provides a combine v2-v5 driver, and a separate v5 only driver. Why is a v3-v5 driver needed?

ArmPkg/ArmPkg.dec
ardbiesheuvel: > The ARM architectural timer has specific interrupt numbers that are architecturally defined by the GIC. GICv5 uses slightly different numbers to previous GICs, as the GICv5 interrupt ID scheme now explicitly includes the interrupt type.
> 

If the numbers are different, those new values can be assigned to the existing PCDs, no?

sarah-walker-arm: Which module would do the assignment?

ArmPkg/Drivers/ArmGicDxe/ArmGicDxe.c
ardbiesheuvel: A v5 only driver seems unnecessary, and I'd prefer to leave the v2/v3 combined driver alone, if at all possible.

Is there a straight-forward way to add this new functionality to the v3 DXE driver only, and leave the combined one alone?

sarah-walker-arm: This can be done, but I'd like to know why this is the desired outcome?

ArmPkg/ArmPkg.dec
ardbiesheuvel: > Which module would do the assignment?

The platform's DSC typically populates these PCDs with appropriate values for the platform in question. If that platform has a GICv5, it would presumably use the GICv5 architected values (assuming that the core in question follows the architecture).

For virtual platforms, we use dynamic PCDs where the values are assigned at boot time, e.g., based on the device tree provided by the VMM.

In either case, I fail to see the need for a separate set of GICv5 specific PCDs.

sarah-walker-arm: Placing these values in the platform DSC seems unnecessary, given they will be the same for every platform with a specific GIC. Making them the responsibility of the platform would seem to cause considerable duplication of PCDs (along with the associated risk of error) for no real gain.

ArmPkg/Drivers/ArmGicDxe/ArmGicDxe.c
ardbiesheuvel: Actually, that is a good question.

My main concern is having too many combinations, that will never get tested properly in practice. But thinking about this a bit more, I see that not adding GICv5 support to the combined driver could also create issues down the road.

Could you clarify the reason for proposing a GICv5-only driver in addition to the combined one? Could we add GICv5 support to the GICv3 driver, and make the combined driver inherit that simply by incorporating the GICv3 driver (which it already does)?

Apologies if I am missing some obvious constraints here - feel free to spell them out.

ArmPkg/ArmPkg.dec
ardbiesheuvel: > Placing these values in the platform DSC seems unnecessary, given they will be the same for every platform with a specific GIC. Making them the responsibility of the platform would seem to cause considerable duplication of PCDs (along with the associated risk of error) for no real gain.

Then avoid the PCDs altogether, and just use `#define`s in the header files. Adding a new set of PCDs just to gain access to an alternate set of architected default values is not the right way to go about this.

MdePkg/Include/AArch64/AArch64.h
```diff
@@ -33,6 +33,9 @@
 #define AARCH64_PFR0_FP   (0xF << 16)
 #define AARCH64_PFR0_GIC  (0xF << 24)
 
+// ID_AA64PFR2 - AArch64 Processor Feature Register 2 definitions
+#define AARCH64_PFR2_GCIE  (0xF << 12)
+
```
ardbiesheuvel, line 38: Is there a publicly accessible reference for this?

ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c
```diff
@@ -0,0 +1,582 @@
...
+GicV5GetInterruptSourceState (
+  IN EFI_HARDWARE_INTERRUPT_PROTOCOL  *This,
+  IN HARDWARE_INTERRUPT_SOURCE        Source,
+  IN BOOLEAN                          *InterruptState
+  )
+{
+  UINT32  InterruptType = Source & GICV5_INTERRUPT_TYPE_MASK;
+  UINT32  InterruptId   = Source & GICV5_INTERRUPT_ID_MASK;
+
+  UINT64  mask;
```
ardbiesheuvel: `Mask`

ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c
```diff
@@ -0,0 +1,582 @@
...
+  return GicCommonRegisterInterruptSource (
+           This,
+           Source,
+           Handler,
+           &mRegisteredInterruptHandlers[IntId]
+           );
+}
+
+// The protocol instance produced by this driver
+EFI_HARDWARE_INTERRUPT_PROTOCOL  gHardwareInterruptV5Protocol = {
```
ardbiesheuvel: Can this be made `STATIC`?

ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c
```diff
@@ -0,0 +1,582 @@
...
+{
+  UINT32  InterruptType = Source & GICV5_INTERRUPT_TYPE_MASK;
+  UINT32  InterruptId   = Source & GICV5_INTERRUPT_ID_MASK;
+
+  if (InterruptType == GICV5_INTERRUPT_TYPE_SPI) {
+    ArmGicV5SpiEnable (Source);
+  } else if (InterruptType == GICV5_INTERRUPT_TYPE_PPI) {
+    if (InterruptId >= GICV5_NUM_PPI_INTERRUPTS) {
+      DEBUG ((DEBUG_ERROR, "GicV5EnableInterruptSource: invalid PPI number\n"));
+      ASSERT (FALSE);
```
ardbiesheuvel: `ASSERT (FALSE)` does not produce a very helpful diagnostic in the DEBUG log. Better to assert on the actual condition being checked here: `ASSERT (InterruptId < GICV5_NUM_PPI_INTERRUPTS)`

ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c
```diff
@@ -0,0 +1,582 @@
...
+      mask |= (1ull << (InterruptId - 64));
+      ArmGicV5SetPpiEnabler1 (mask);
+    } else {
+      UINT64  mask = ArmGicV5GetPpiEnabler0 ();
+      mask |= (1ull << InterruptId);
+      ArmGicV5SetPpiEnabler0 (mask);
+    }
+  } else {
+    DEBUG ((DEBUG_ERROR, "GicV5EnableInterruptSource: invalid interrupt type\n"));
+    ASSERT (FALSE);
```
ardbiesheuvel: Same here: `ASSERT (InterruptType == GICV5_INTERRUPT_TYPE_SPI || InterruptType == GICV5_INTERRUPT_TYPE_PPI)`

ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c
```diff
@@ -0,0 +1,582 @@
...
+  UINT32  InterruptId   = Source & GICV5_INTERRUPT_ID_MASK;
+
+  UINT64  mask;
+  UINT64  Status;
+
+  if (InterruptType == GICV5_INTERRUPT_TYPE_SPI) {
+    Status = ArmGicV5ReadInterruptConfig (Source);
+
+    if (Status & 1) {
+      ASSERT (FALSE);
```
ardbiesheuvel: `ASSERT ((Status & 1) == 0)`

ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c
```diff
@@ -0,0 +1,582 @@
...
+  )
+{
+  UINT32  InterruptType = Source & GICV5_INTERRUPT_TYPE_MASK;
+  UINT32  InterruptId   = Source & GICV5_INTERRUPT_ID_MASK;
+  UINT32  GicTriggerType;
+
+  if (InterruptType == GICV5_INTERRUPT_TYPE_SPI) {
+    if (!WaitForIrsSpiIdle ()) {
+      ASSERT (FALSE);
+      return EFI_UNSUPPORTED;
```
ardbiesheuvel: Please return something other than `EFI_UNSUPPORTED` here, as that suggests an issue with the caller provided function arguments rather than an internal issue in the hardware or driver.

ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c
```diff
@@ -0,0 +1,582 @@
...
+    }
+
+    MmioWrite32 (mGicIrsConfigFrameBase + GICV5_IRS_SPI_SELR, InterruptId);
+    if (!WaitForIrsSpiIdle ()) {
```
ardbiesheuvel: Can we combine these configuration MMIO accesses and preceding/subsequent waits for idle into helper functions?

ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c
```diff
@@ -0,0 +1,582 @@
...
+  } else {
+    DEBUG ((DEBUG_ERROR, "GicV5SetTriggerType: invalid interrupt type\n"));
+    ASSERT (FALSE);
+    return EFI_UNSUPPORTED;
+  }
+
+  return EFI_SUCCESS;
+}
+
+EFI_HARDWARE_INTERRUPT2_PROTOCOL  gHardwareInterrupt2V5Protocol = {
```
ardbiesheuvel: Can this be `STATIC`?

ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c
```diff
@@ -0,0 +1,582 @@
...
+  }
+
+  Status = GicCommonInstallAndRegisterInterruptService (
+             &gHardwareInterruptV5Protocol,
+             &gHardwareInterrupt2V5Protocol,
+             GicV5IrqInterruptHandler,
+             GicV5ExitBootServicesEvent
+             );
+
+  if (Status == EFI_SUCCESS) {
```
ardbiesheuvel: Please don't use success handling. Instead, do something like

```
if (EFI_ERROR (Status)) {
  FreePool();
  return Status;
}

ArmGicV5EnableInterruptInterface();
return EFI_SUCCESS;
```

ArmPkg/Drivers/TimerDxe/TimerDxe.c
```diff
@@ -377,27 +382,38 @@ TimerInitialize (
   Status = TimerDriverSetTimerPeriod (&gTimer, 0);
   ASSERT_EFI_ERROR (Status);
 
+  if (ArmHasGicV5SystemRegisters ()) {
+    TimerSecIntrNum  = GICV5_ARCH_TIMER_SEC_INTR_NUM;
+    TimerIntrNum     = GICV5_ARCH_TIMER_INTR_NUM;
+    TimerHypIntrNum  = GICV5_ARCH_TIMER_HYP_INTR_NUM;
+    TimerVirtIntrNum = GICV5_ARCH_TIMER_VIRT_INTR_NUM;
+  } else {
```
ardbiesheuvel, line 390: This looks fine now. But please provide a commit log for this patch.

MdePkg/Include/AArch64/AArch64.h
sarah-walker-arm: Yes, https://developer.arm.com/documentation/aes0070/latest/, 9.2.14.

ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c
leiflindholm: If it can, it should be m, not g.

ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c
leiflindholm: If it can, it should be m, not g.

ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c
```diff
@@ -0,0 +1,626 @@
...
+  IN EFI_HARDWARE_INTERRUPT_PROTOCOL  *This,
+  IN HARDWARE_INTERRUPT_SOURCE        Source,
+  IN HARDWARE_INTERRUPT_HANDLER       Handler
+  )
+{
+  // The GICv5 driver supports both PPIs and SPIs. To support both in a single
+  // interrupt handler array, map IDs 0 to NUM_PPI_INTERRUPTS-1 as PPIs, and
+  // NUM_PPI_INTERRUPTS and above as SPIs.
+  UINTN  Type = Source & GICV5_INTERRUPT_TYPE_MASK;
+  UINTN  IntId;
```
ardbiesheuvel, line 327: This breaks the CLANG build:

```
/home/ardb/build/edk2/ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c:337:14: error: variable 'IntId' is used uninitialized whenever 'if' condition is false [-Werror,-Wsometimes-uninitialized]
  337 |   } else if (Type == GICV5_INTERRUPT_TYPE_SPI) {
      |              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/ardb/build/edk2/ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c:342:7: note: uninitialized use occurs here
  342 |   if (IntId >= mGicNumInterrupts) {
      |       ^~~~~
/home/ardb/build/edk2/ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c:337:10: note: remove the 'if' if its condition is always true
  337 |   } else if (Type == GICV5_INTERRUPT_TYPE_SPI) {
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/ardb/build/edk2/ArmPkg/Drivers/ArmGicDxe/GicV5/ArmGicV5Dxe.c:327:15: note: initialize the variable 'IntId' to silence this warning
  327 |   UINTN  IntId;
      |               ^
      |                = 0
```

ArmPkg/ArmPkg.dsc
```diff
@@ -124,6 +124,7 @@
   ArmPkg/Drivers/ArmGicDxe/ArmGicDxe.inf
   ArmPkg/Drivers/ArmGicDxe/ArmGicV2Dxe.inf
   ArmPkg/Drivers/ArmGicDxe/ArmGicV3Dxe.inf
+  ArmPkg/Drivers/ArmGicDxe/ArmGicV5Dxe.inf
```
ardbiesheuvel: Again, I don't think we need a separate GICv5 driver.

What I would like to see is:
- drop ArmGicV5Dxe
- drop AArch32 support from GicV3Dxe (i.e., remove the [Sources.ARM] block and move the .inf to [Components.AARCH64] in ArmPkg.dsc)
- add v5 functionality to GicV3Dxe, so that GicV3DxeInitialize() invokes GicV5DxeInitialize under the hood, if the hardware is v5 capable
- add a dummy GicV5DxeInitialize() to the combined v2/v3 driver for AArch32 builds

This means that 32-bit platforms relying on GicV3Dxe directly should move to the combined v2/v3 driver, but this is fine, as no such platforms should exist to begin with.

ArmPkg/Drivers/ArmGicDxe/ArmGicDxe.c
sarah-walker-arm: Apologies, missed this last comment.

The only reason for a v5-only driver was to match the existing v2-only and v3-only drivers. There is to my knowledge no existing use case for it.

ArmPkg/Drivers/TimerDxe/TimerDxe.c
ardbiesheuvel: Having looked at the spec now, I am confused about these changes: these assignments are identical to the PCD defaults (which carry the architected PPIs for GICv3 and earlier), except for the fact these these appear to have the type field set to LPI?

These are PPIs, and so AIUI, the type should be 0x0 no? In which case these architected defaults are identical to the existing PCDs if not overridden, and we don't need these changes to begin with? Please clarify.

sarah-walker-arm: The type field is 001b for PPI, so the new values are correct.

ardbiesheuvel: OK, I got confused by the fact that the type starts at bit #29 not bit #28. 

So what we end up here is an INTID not an interrupt number, right? If my understanding is correct, would you mind updating the variable names and replace IntrNum with IntId. This may seem pedantic, but the fact that the type now becomes part of the identifier is a significant change, and using clear variable names does help a great deal, especially for people not as familiar with the GIC architecture as you are.

PR #11241, opened 2025-06-30
BaseTools/PatchCheck.py: use dynamic subject length when many packages enumerated

leiflindholm:
#11175 added a check for BaseTools that all packages being modified are enumerated before the :. However, when something very invasive happens, 75 characters run out quickly.

Tidy the subject line handling up a little, before adding special handling for the case where the first ':' character in the subject is more than 55 characters in, always permitting 20 characters beyond that.


leiflindholm:
(Fixed typo: second "zero" -> "one" in 1/3.)

leiflindholm:
@bexcran @gapalomi @lgao4 @YuweiChen1110
This is needed to unblock https://github.com/tianocore/edk2/pull/11017.

gapalomi (approved):

PR #11246, opened 2025-07-01
OvmfPkg/PlatformBootManagerCommonLib: Move Uninstall MemAttrProtocol check

ajyoung-oracle:
# Description
Commit efaa102d006f added EFI memory attribute Protocol support to UefiCpuPkg/CpuDxe which is used by OVMF. This breaks some older shims which use the protocol improperly.

There is an existing PCD and fw_cfg option to workaround this issue by forcing the uninstallation of the protocol prior to booting a boot option.

So, move the checks for the PCD/fw_cfg option from PlatformBootManagerLibLight to PlatformBootManagerCommonLib to allow OVMF to boot older/broken shims.

Cc: Gerd Hoffmann <kraxel@redhat.com>
Cc: Ard Biesheuvel <ardb+tianocore@kernel.org>
Cc: Jiewen Yao <jiewen.yao@intel.com>

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

With both OVMF (x86_64) and AAVMF (Arm), both of which consume PlatformBootManagerCommonLib,
boot older shims which fail when MemAttrProtocol is present. With this new fix included, shims can boot properly. Tested both the PCD and the fw_cfg option method to uninstall the protocol.

## Integration Instructions

N/A


kraxel:
Phew.  Been there, tried that, got no T-Shirt.  See https://github.com/tianocore/edk2/pull/10667

shim actually isn't a big problem.  version 15.8+ is fine, and it is a security release, you should certainly not run something older.  And it is old enough that distros have rolled out updates (and new install media) meanwhile.

grub is alot more messy.  upstream 2.12 is actually fine, but distros carry tons of downstream patches, things are breaking left and right and apparently distros do not get that sorted in a timely fashion.  The issue is known for almost half a year meanwhile ...

Adding UninstallMemAttrProtocol support to OVMF indeed serves as workaround.  Problem with that approach is that it works for virtual machines only.  RHEL / Fedora builds have that nevertheless.  But at the end of the day that only buys some time, it is clearly not a proper solution for the problem.

Hardware vendors (I know about one specifically, but most likely it is not the only one) started to update their firmware to 2025 edk2 versions, and now they are seeing the NX-related page faults too when testing linux distros.

ajyoung-oracle:
> Phew. Been there, tried that, got no T-Shirt. See #10667

 Ah.. gotcha. In my world, an OVMF upgrade causing VMs (with broken shims/grub/etc.) to stop booting would be perceived as an OVMF BUG (right or wrong). So, I need to at least have a configurable (i.e. fw-cfg) method to allow the VMs to continue booting.

ardbiesheuvel:
> > Phew. Been there, tried that, got no T-Shirt. See #10667
> 
> Ah.. gotcha. In my world, an OVMF upgrade causing VMs (with broken shims/grub/etc.) to stop booting would be perceived as an OVMF BUG (right or wrong). So, I need to at least have a configurable (i.e. fw-cfg) method to allow the VMs to continue booting.

This is not about whether or not OVMF is buggy.

The problem is with distros having hacked up downstream forks of boot components that have bugs that don't exist in the upstream. So if distros are happy to ship broken GRUBs etc, surely they can find the motivation to carry the workarounds that are needed as a result in their downstream forks of OVMF.

ajyoung-oracle:
OK, Got it... I will close this PR.

PR #11248, opened 2025-07-02
MdePkg: Expand max debug message length to 0x200

EricGao2015:
If string length is longer than 0x100, debug output will be truncated. Change MAX_DEBUG_MESSAGE_LENGTH to 0x200. This value can meet most scenarios.
Same commit(https://github.com/tianocore/edk2/commit/b5d1dc94d08acf4bb8b95fe50baff3d605714015) is submitted in OvmfPkg.

### DIFF
--- a/MdePkg/Library/BaseDebugLibSerialPort/DebugLib.c
+++ b/MdePkg/Library/BaseDebugLibSerialPort/DebugLib.c
@@ -24,7 +24,7 @@
 //
 // Define the maximum debug and assert message length that this library supports
 //
-#define MAX_DEBUG_MESSAGE_LENGTH  0x100
+#define MAX_DEBUG_MESSAGE_LENGTH  0x200
 
 //
 // VA_LIST can not initialize to NULL for all compiler, so we use this to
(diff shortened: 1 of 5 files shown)


mdkinney:
What debug messages are longer than 0x100.  We should look into making those shorter.

EricGao2015:
> What debug messages are longer than 0x100. We should look into making those shorter.

I want to print out parameter Result string of  `EFI_HII_CONFIG_ACCESS_PROTOCOL.ExtractConfig()`, but output is truncated. Because the Results string length is longer than 0x100. 
```
typedef
EFI_STATUS
  (EFIAPI * EFI_HII_ACCESS_EXTRACT_CONFIG ) (
  IN CONST EFI_HII_CONFIG_ACCESS_PROTOCOL    *This,
  IN CONST EFI_STRING                        Request,
  OUT EFI_STRING                             *Progress,
  OUT EFI_STRING                             *Results
  );
```
As mentioned in https://github.com/tianocore/edk2/commit/b5d1dc94d08acf4bb8b95fe50baff3d605714015, DevicePath string length may also exceed 0x100.

mdkinney:
Loops can be used to break those up into multiple DEBUG() invocations.  Especially HII extract config data.  Because that can be must larger than the 0x200 length this change suggests, so this is not a fix for HII extract config.

EricGao2015:
> Loops can be used to break those up into multiple DEBUG() invocations. Especially HII extract config data. Because that can be must larger than the 0x200 length this change suggests, so this is not a fix for HII extract config.

Maybe a new debug print function for super long string is needed. I'll close this PR.

Issue #11245, opened 2025-07-01
CryptoPkg: KMIP connection works even after the certificates expired.


kanagavels97:
Hi @liyi77 

In the TlsNew function, the X509_V_FLAG_NO_CHECK_TIME flag is enabled. This configuration bypasses the validation of certificate and CRL expiration dates against the current system time. As a result, the KMIP connection remains functional even when the certificates are expired.

![Image](https://github.com/user-attachments/assets/a9437d14-5c93-407b-899a-3efa94dc3edb)

Could you please confirm whether we should avoid setting the X509_V_FLAG_NO_CHECK_TIME flag in order to enforce certificate and CRL validity checks against the current date and time?

Thanks,
Kanagavel

@PrarthanaSagarV @arvinrajp

liyi77:
There is a bug in the time check in openssl 1.1 that will break tls: https://github.com/tianocore/edk2/issues/9223

I believe this issue has been fixed when upgraded to openssl 3.0. You can turn on this flag if you need it.

kanagavels97:
@liyi77 Can I submit a patch by enabling flag for time check?

liyi77:
Of course, please go ahead.

PR #11250, opened 2025-07-02
CryptoPkg: Enable the time check flag.

kanagavels97:
REF:https://github.com/tianocore/edk2/issues/11245

Enable the time check flag to verify the current time during the TLS certificate validation.


liyi77:
Looks good to me, have you tested it with expired cert and report the correct error?

kanagavels97:
> Looks good to me, have you tested it with expired cert and report the correct error?

Yes, we have tested by setting the system date 3 years ahead of the validity of the certificate.
It fails with SSL error code L14:R86, which indicates there was a problem with the certificate.
The connection is successful using the same configuration, when the X509_V_FLAG_NO_CHECK_TIME flag is enabled.

PR #11253, opened 2025-07-03
BaseTools,EmbeddedPkg: Set SOURCE_DATE_EPOCH if not used and use that for VirtualRealTimeClockLib

leiflindholm:
The first commit is unrelated really, but is a bit of cleanup I did when
starting a different approach than I ended up with.

Second commit makes sure SOURCE_DATE_EPOCH is set in the environment,
setting it to the value of time.time() if not already present. Then uses this
value to print the "Build start time:" message.

Third commit modifies EmbeddedPkg VirtualRealTimeClockLib to use this variable
instead of invoking shell commands, making the module buildable also in Windows
environments.

### DIFF
--- a/BaseTools/Source/Python/build/build.py
+++ b/BaseTools/Source/Python/build/build.py
@@ -82,7 +82,7 @@ TmpTableDict = {}
 #   If any of above environment variable is not set or has error, the build
 #   will be broken.
 #
-def CheckEnvVariable():
+def CheckEnvVariables():
     # check WORKSPACE
     if "WORKSPACE" not in os.environ:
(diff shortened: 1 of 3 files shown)


mdkinney:
@leiflindholm Should this PR be part of edk2-stable202508?

leiflindholm:
> @leiflindholm Should this PR be part of edk2-stable202508?

Whoops, I thought this was already merged, but I see we have no review from the BaseTools maintainers.

Ideally, yes, and it was at least partially reviewed (well) before soft-freeze happened.

mdkinney:
Not stale

leiflindholm:
@bexcran @lgao4 @gapalomi @YuweiChen1110 @ponchofigueroa ?

lgao4 (approved):

PR #11257, opened 2025-07-07
DynamicTablesPkg: Add alias for EArchCommonObjPciConfigSpaceInfo

abdattar:
Introduce EArchCommonObjMcfgPciConfigSpaceInfo as an alias for EArchCommonObjPciConfigSpaceInfo.

The CM_ARCH_COMMON_PCI_CONFIG_SPACE_INFO object is utilized by both the MCFG and ACPI SSDT PCIE generators. Assigning a unique configuration object ID ensures the platform configuration driver supplies the correct data.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on AMD platform.

## Integration Instructions

N/A

### DIFF
--- a/DynamicTablesPkg/Library/Acpi/Common/AcpiMcfgLib/McfgGenerator.c
+++ b/DynamicTablesPkg/Library/Acpi/Common/AcpiMcfgLib/McfgGenerator.c
@@ -2,6 +2,7 @@
   MCFG Table Generator
 
   Copyright (c) 2017 - 2019, ARM Limited. All rights reserved.
+  Copyright (C) 2025, Advanced Micro Devices, Inc. All rights reserved.
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
   @par Reference(s):
@@ -25,9 +26,9 @@
 /** ARM standard MCFG Generator
 
 Requirements:
-  The following Configuration Manager Object(s) are required by
-  this Generator:
-  - EArchCommonObjPciConfigSpaceInfo
+  This Generator requires the following Configuration Manager Object(s):
+  - EArchCommonObjMcfgPciConfigSpaceInfo (primary, preferred)
+  - EArchCommonObjPciConfigSpaceInfo (secondary, for backward compatibility)
 */
 
 #pragma pack(1)
@@ -53,6 +54,14 @@ typedef
 GET_OBJECT_LIST (
   EObjNameSpaceArchCommon,
   EArchCommonObjPciConfigSpaceInfo,
+  CM_ARCH_COMMON_PCI_CONFIG_SPACE_INFO
+  );
+
+/** Retrieve the MCFG PCI Configuration Space Information.
+*/
+GET_OBJECT_LIST (
+  EObjNameSpaceArchCommon,
+  EArchCommonObjMcfgPciConfigSpaceInfo,
   CM_ARCH_COMMON_PCI_CONFIG_SPACE_INFO
   );
 
@@ -154,12 +163,22 @@ BuildMcfgTable (
   }
 
   *Table = NULL;
-  Status = GetEArchCommonObjPciConfigSpaceInfo (
+  Status = GetEArchCommonObjMcfgPciConfigSpaceInfo (
              CfgMgrProtocol,
              CM_NULL_TOKEN,
              &PciConfigSpaceInfoList,
              &ConfigurationSpaceCount
              );
+  if (Status == EFI_NOT_FOUND) {
+    // Fallback to the older object if the new one is not found.
+    Status = GetEArchCommonObjPciConfigSpaceInfo (
+               CfgMgrProtocol,
+               CM_NULL_TOKEN,
+               &PciConfigSpaceInfoList,
+               &ConfigurationSpaceCount
+               );
+  }
+
   if (EFI_ERROR (Status)) {
     DEBUG ((
       DEBUG_ERROR,
--- a/DynamicTablesPkg/Library/Common/TableHelperLib/ConfigurationManagerObjectParser.c
+++ b/DynamicTablesPkg/Library/Common/TableHelperLib/ConfigurationManagerObjectParser.c
@@ -184,7 +184,8 @@ STATIC CONST CM_OBJ_PARSER  CmArmGenericWatchdogInfoParser[] = {
   { "Flags",               4, "0x%x",   NULL }
 };
 
-/** A parser for EArchCommonObjPciConfigSpaceInfo.
+/** A parser for EArchCommonObjPciConfigSpaceInfo or
+    EArchCommonObjMcfgPciConfigSpaceInfo.
 */
 STATIC CONST CM_OBJ_PARSER  CmArchCommonPciConfigSpaceInfoParser[] = {
   { "BaseAddress",           8,                        "0x%llx", NULL },
@@ -930,6 +931,7 @@ STATIC CONST CM_OBJ_PARSER_ARRAY  ArchCommonNamespaceObjectParser[] = {
   CM_PARSER_ADD_OBJECT (EArchCommonObjMemoryLatBwInfo,              CmArchCommonMemoryLatBwInfo),
   CM_PARSER_ADD_OBJECT (EArchCommonObjMemoryCacheInfo,              CmArchCommonMemoryCacheInfo),
   CM_PARSER_ADD_OBJECT (EArchCommonObjSpcrInfo,                     CmArchCommonObjSpcrInfoParser),
+  CM_PARSER_ADD_OBJECT (EArchCommonObjMcfgPciConfigSpaceInfo,       CmArchCommonPciConfigSpaceInfoParser),
   CM_PARSER_ADD_OBJECT_RESERVED (EArchCommonObjMax)
 };
 
--- a/DynamicTablesPkg/Include/ArchCommonNameSpaceObjects.h
+++ b/DynamicTablesPkg/Include/ArchCommonNameSpaceObjects.h
@@ -72,6 +72,7 @@ typedef enum ArchCommonObjectID {
   EArchCommonObjMemoryCacheInfo,                ///< 44 - Memory Cache Info
   EArchCommonObjSpcrInfo,                       ///< 45 - Serial Terminal and Interrupt Info
   EArchCommonObjTpm2DeviceInfo,                 ///< 46 - TPM2 Device Info
+  EArchCommonObjMcfgPciConfigSpaceInfo,         ///< 47 - MCFG PCI Configuration Space Info
   EArchCommonObjMax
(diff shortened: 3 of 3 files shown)


pierregondois:
Hello Abdul,
I am not sure I understand why this is needed, would it be possible to explain the issue/case you need to handle ?

abdattar:
Hi @pierregondois ,
The `CM_ARCH_COMMON_PCI_CONFIG_SPACE_INFO` object is used to obtain PCIe configuration information from the platform via the configuration manager driver.

This object is utilized to generate both the MCFG and PCIE SSDT tables.
However, the platform (configuration manager) cannot differentiate between them, so an alias is needed to generate the MCFG table using this alias.

Why are the MCFG table data and the PCIE SSDT table `CM_ARCH_COMMON_PCI_CONFIG_SPACE_INFO` different?
On x64/IA-PC compatible systems, it is possible to provide base address mappings of PCIe address space for each "Segment" in the MCFG table.
Each segment may contain multiple root bridges.

In contrast, the PCIE SSDT table expects a `CM_ARCH_COMMON_PCI_CONFIG_SPACE_INFO` entry for each root bridge.

pierregondois:
Just to be sure I understand correctly:
A PCI segment contains one or more PCI root bridge. If there are at least 2 root bridges in a PCI segment, it is not possible to correctly populate the MCFG table.

From the MCFG spec,  for the field `Configuration space base address allocation structure [n]`:
```
A list of the memory mapped configuration base
address allocation structures. This list will
contain at least one entry corresponding to each
PCI Segment Group present in the platform. The
structure of this entry is defined in Table 4-3.
```
So I am not sure each PCI segment should have a unique Configuration space sub-structures. I.e. if there are 2 root bridges, shouldn't they be both described in the MCFG with the same PCI segment ?

The knowledge I have on PCI is relatively low, so please tell me if I didn't understand the issue or anything else

abdattar:
Hi @pierregondois 
  The current implementation of the MCFG table is correct and functions as intended.
  I am working on generating (porting) the ACPI SSDT PCI table for the X64 architecture, where some implementation conflicts arise.
  The SSDT PCI table generation requires a CM_ARCH_COMMON_PCI_CONFIG_SPACE_INFO object for each root bridge, whereas the MCFG table is designed for segment-level configuration, not for individual root bridges.
  

> Just to be sure I understand correctly: A PCI segment contains one or more PCI root bridge. If there are at least 2 root bridges in a PCI segment, it is not possible to correctly populate the MCFG table.
Nope, we can populate the MCFG table but it will contain information related to segment only.
it will not have information about each root bridge, like which root bridge(1 or 2) contains which address mapping.
> 
> From the MCFG spec, for the field `Configuration space base address allocation structure [n]`:
> 
> ```
> A list of the memory mapped configuration base
> address allocation structures. This list will
> contain at least one entry corresponding to each
> PCI Segment Group present in the platform. The
> structure of this entry is defined in Table 4-3.
> ```
> 
> So I am not sure each PCI segment should have a unique Configuration space sub-structures. I.e. if there are 2 root bridges, shouldn't they be both described in the MCFG with the same PCI segment ?
Yes, each PCI segment should have a unique configuration space sub-structures.
comment from spec, section 4.1.2
The MCFG table is only used to communicate the base addresses corresponding to the PCI Segment Groups available to the system at boot. This table directly refers to PCI Segment Groups defined in the system via the _SEG object in the ACPI name space for the applicable host bridge device. For systems containing only a single PCI Segment Group, the default PCI Segment Group number, namely, PCI Segment Group 0, is implied. In such a case, the default PCI Segment Group need not be represented in the ACPI Nname Sspace (i.e., no _SEG method is required in such a hierarchy).

another document from picsig, title "Enabling Multiple Base Addresses per PCI Segment Group"
_The changes enable the MCFG table format to allow for multiple memory mapped base address entries, instances of Table 4-3, per single PCI Segment Group.
Currently, the MCFG table format allows for more than one memory mapped base address entry, instance of Table 4-3, provided each entry corresponds to a unique PCI Segment Group consisting of 256 PCI buses. Multiple entries corresponding to a single PCI Segment Group are not allowed._
Ref : 
> 
> The knowledge I have on PCI is relatively low, so please tell me if I didn't understand the issue or anything else

DynamicTablesPkg/Library/Acpi/Common/AcpiMcfgLib/McfgGenerator.c
```diff
@@ -52,7 +53,7 @@ typedef
 */
 GET_OBJECT_LIST (
   EObjNameSpaceArchCommon,
-  EArchCommonObjPciConfigSpaceInfo,
+  EArchCommonObjMcfgPciConfigSpaceInfo,
```
pierregondois, line 64: The issue with that is that it will break platforms that don't have/need the new `EArchCommonObjMcfgPciConfigSpaceInfo`. Would it make sense to do the following instead ?

(pseudo-code)
```
  Status = GetEArchCommonObjMcfgPciConfigSpaceInfo()
  if (Status == EFI_NOT_FOUND) {
    // If not found, default to PciConfigSpaceInfo
    Status = GetEArchCommonObjPciConfigSpaceInfo ()
  }
```

pierregondois (commented):
Ok I think I finally understood, just one small comment

abdattar:
Hi @pierregondois , apologies for the delayed reply.
I've addressed the review comments—could you please review them again?

DynamicTablesPkg/Library/Acpi/Common/AcpiMcfgLib/McfgGenerator.c
abdattar: Done.

DynamicTablesPkg/Library/Common/TableHelperLib/ConfigurationManagerObjectParser.c
```diff
@@ -893,6 +894,7 @@ STATIC CONST CM_OBJ_PARSER_ARRAY  ArchCommonNamespaceObjectParser[] = {
   CM_PARSER_ADD_OBJECT (EArchCommonObjFixedFeatureFlags,            CmArchCommonFixedFeatureFlagsParser),
   CM_PARSER_ADD_OBJECT (EArchCommonObjCmRef,                        CmArchCommonObjRefParser),
   CM_PARSER_ADD_OBJECT (EArchCommonObjPciConfigSpaceInfo,           CmArchCommonPciConfigSpaceInfoParser),
+  CM_PARSER_ADD_OBJECT (EArchCommonObjMcfgPciConfigSpaceInfo,       CmArchCommonPciConfigSpaceInfoParser),
```
pierregondois: The structure is an array, so the location is important. I think the new entry should be added at the end.
A parser for the `EArchCommonObjTpm2DeviceInfo` object is missing, but it can be added afterwards I guess.

pierregondois (approved):

abdattar:
Looks like merge failed with below message, do I need to rebase the branch?
"The branch protection setting Require branches to be up to date before merging is not compatible with max_parallel_checks>1, queue_conditions != merge_conditions and must be unset."

pierregondois:
Hello @abdattar,
I am re-reading this conversation, and I am wondering whether introducing EArchCommonObjMcfgPciConfigSpaceInfo was correct.

In "PCI Firmware Specification Revision 3.2":
`Multiple entries corresponding to a single PCI Segment Group is not allowed.`

The spec. you quoted "Enabling Multiple Base Addresses per PCI Segment Group", seems to actually describe:
- what the MCFG table requested in PCI Firmware spec 3.2
- what the MCFG table will request in PCI Firmware spec 3.3

In "PCI Firmware Specification Revision 3.3":
```
Multiple entries corresponding to a single PCI Segment Group are also allowed provided the Field (PCI
Segment Group Number, Start Bus Number, and End Bus Number) uniquely identifies each PCI
Host Bridge and the bus number values do not overlap.
```

So I am wondering whether the EArchCommonObjPciConfigSpaceInfo object was actually enough to describe host bridges for SSDT and MCFG spec:
- in 3.2, it would have been possible to gather EArchCommonObjPciConfigSpaceInfo objects with the same segment ID and advertise one MCFG entry per segment
- in 3.3, it should be even simpler as 2 host bridges with the same segment can be described as 2 MCFG entries

Please let me know what you think.

abdattar:
> Hello @abdattar, I am re-reading this conversation, and I am wondering whether introducing EArchCommonObjMcfgPciConfigSpaceInfo was correct.
> 
> In "PCI Firmware Specification Revision 3.2": `Multiple entries corresponding to a single PCI Segment Group is not allowed.`
> 
> The spec. you quoted "Enabling Multiple Base Addresses per PCI Segment Group", seems to actually describe:
> 
> * what the MCFG table requested in PCI Firmware spec 3.2
> * what the MCFG table will request in PCI Firmware spec 3.3
> 
> In "PCI Firmware Specification Revision 3.3":
> 
> ```
> Multiple entries corresponding to a single PCI Segment Group are also allowed provided the Field (PCI
> Segment Group Number, Start Bus Number, and End Bus Number) uniquely identifies each PCI
> Host Bridge and the bus number values do not overlap.
> ```
> 
> So I am wondering whether the EArchCommonObjPciConfigSpaceInfo object was actually enough to describe host bridges for SSDT and MCFG spec:
> 
> * in 3.2, it would have been possible to gather EArchCommonObjPciConfigSpaceInfo objects with the same segment ID and advertise one MCFG entry per segment
> * in 3.3, it should be even simpler as 2 host bridges with the same segment can be described as 2 MCFG entries
> 
> Please let me know what you think.

Let me recall why I introduced the alias:
  1. SSDT generator consumes CM_ARCH_COMMON_PCI_CONFIG_SPACE_INFO per root bridge. This is per-root-bridge granularity.
     Generates one ACPI device node per root bridge and ECAM MMIO reservation ReserveEcamSpace().
 2. PCI FW 3.3 allows multiple MCFG entries sharing the same segment number (with unique (Segment, StartBus, EndBus) combination). If the same object list fed both generators, the SSDT generator would emit multiple ACPI device nodes with the same  _SEG value — which may or may not be the platform's intent.
3. The MCFG revision byte has never been incremented — it is always 0x01 in acpi spec. There is no easy way in genertor library detect pci3.2 or pci3.3 and adjust its behavior accordingly.

   Without the alias, the generator library would need to implement spec-version detection and entry aggregation/splitting logic — which can be easily handler by platform through configuration manager, not in a common library.

One more aspect which X64 dont use but worth checking (Its just my assumption)
  if PcdPciUseSegmentAsUid = TRUE and two root bridges share the same segment number (valid under PCI FW 3.3 with different bus ranges), they would both get the same _UID — which is an ACPI namespace collision. This is
  another concrete reason why MCFG and SSDT data cannot blindly share the same object list under 3.3 semantics.

if you have any alternative solution, i am happy to adapt it.

PR #11265, opened 2025-07-08
BaseTools/toolsetup.bat: Set IASL_PREFIX when using Mingw-w64 on Windows

nate-desimone:
# Description

If IASL_PREFIX is not already defined, then edksetup.bat should set it. This unifies the behavior between Visual Studio and Mingw-w64.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- Confirmed that if the IASL_PREFIX environment variable does not exist, then edksetup.bat will set it when using Mingw-w64.

## Integration Instructions

N/A


BaseTools/toolsetup.bat
```diff
@@ -333,6 +333,12 @@ if defined VS2022 (
   call %EDK_TOOLS_PATH%\set_vsprefix_envs.bat
   call %EDK_TOOLS_PATH%\get_vsvars.bat
 )
+if defined BASETOOLS_MINGW_BUILD (
+  if not defined IASL_PREFIX (
+    REM For VS builds this environment variable is set via set_vsprefix_envs.bat
+    set IASL_PREFIX=C:\ASL\
```
gapalomi: Can you check if the path exists? Then raise an error if not to stop execution

nate-desimone: Done.

PR #11267, opened 2025-07-09
BaseTools: Fix FMMT FvHandler Padding operation issue

YuweiChen1110:
When using the FMMT FvHandler function, new padding size should be calculated correctly comparing with origin ffs and new ffs, else it will cause the binary size changes. This patch is used to fix the bug.

- [ ] Breaking change?
  No
- [ ] Impacts security?
  No
- [ ] Includes tests?
  No

## How This Was Tested

Have tested it with the issue finding image binary. After deploying, the binary size changes ignored.

## Integration Instructions

N/A


AshrafAliS (approved):

AshrafAliS:
@lgao4 @bexcran @gapalomi help to review

lgao4 (approved):

PR #11271, opened 2025-07-09
ShellPkg/Library: Remove unnecessary error check

Nanjinmoorthy2002:
Remove the error handling from the connect.c file as it is redundant.

# Description

<_Remove the error handling from connect.c as it is redundant. The same condition is validated, making this check unnecessary.._>

## Integration Instructions

<_ N/A_>

### DIFF
--- a/ShellPkg/Library/UefiShellDriver1CommandsLib/Connect.c
+++ b/ShellPkg/Library/UefiShellDriver1CommandsLib/Connect.c
@@ -520,10 +520,6 @@ ShellCommandRunConnect (
         }
 
         Handle1 = ConvertHandleIndexToHandle ((UINTN)Intermediate);
-        if (EFI_ERROR (Status)) {
-          ShellPrintHiiEx (-1, -1, NULL, STRING_TOKEN (STR_GEN_INV_HANDLE), gShellDriver1HiiHandle, L"connect", Param1);
-          ShellStatus = SHELL_INVALID_PARAMETER;
-        }
       } else {
         Handle1 = NULL;
       }


ShellPkg/Library/UefiShellDriver1CommandsLib/Connect.c
```diff
@@ -520,10 +520,6 @@ ShellCommandRunConnect (
         }
 
         Handle1 = ConvertHandleIndexToHandle ((UINTN)Intermediate);
```
pierregondois, line 522: Checking the Status is effectively redundant, but checking the following seems more appropriate instead:
```
if (Handle1 == NULL) {
  ...
}
```

Nanjinmoorthy2002: But after this change also we faced a Coverity warning as this handle check happens in the below code.

if ((Param1 != NULL) && (Handle1 == NULL)) {
...
}

pierregondois: `Handle1 = ConvertHandleIndexToHandle ((UINTN)Intermediate);`
If `Handle1` is NULL, it seems that `ShellStatus = SHELL_INVALID_PARAMETER`, so we should not enter the following code IIUC:
```
      if (ShellStatus == SHELL_SUCCESS) {
        if ((Param1 != NULL) && (Handle1 == NULL)) {
```

Nanjinmoorthy2002: Anyway, the Handle1 check happens later in the code, so why do we need this part?

if (Handle1 == NULL) {
  ...
}

pierregondois: yes right indeed, sorry for the bad comment

pierregondois (approved):

PR #11273, opened 2025-07-09
MdeModulePkg: DxeMain: Honor memory bucket during memory map merging

kuqin12:
# Description

The current memory map merging logic in `CoreAddRange` only checks the memory type without considering the bucket type.

However, the `CoreGetMemoryMap` routine will only report the entries that are entirely inside a specific bucket type. This will confuse the consumer and report false free memory that belongs to special memory buckets.

This change fixes the issue by gating the merging logic by checking the bucket type first. The merging is only allowed either if the bucket type is not one of the special types, or if the bucket type is the same for both entries.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This is tested on proprietary hardware platform as well as QEMU virtual platforms.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Mem/Imem.h
+++ b/MdeModulePkg/Core/Dxe/Mem/Imem.h
@@ -146,6 +146,21 @@ CoreInternalAllocatePages (
   IN BOOLEAN                   NeedGuard
   );
 
+/**
+  Get the memory bucket type for a given memory range.
+
+  @param  PhysicalStart  The address of the first byte in the memory region.
+  @param  PhysicalEnd    The address of the last byte in the memory region.
+
+  @return The memory type for the bucket that contains the given physical address range.
+          If the address range does not match any special bucket, it returns EfiMaxMemoryType.
+**/
+EFI_MEMORY_TYPE
+GetBucketMemoryType (
+  IN EFI_PHYSICAL_ADDRESS  PhysicalStart,
+  IN EFI_PHYSICAL_ADDRESS  PhysicalEnd
+  );
+
 //
 // Internal Global data
 //
--- a/MdeModulePkg/Core/Dxe/Mem/Page.c
+++ b/MdeModulePkg/Core/Dxe/Mem/Page.c
@@ -96,6 +96,16 @@ EFI_MEMORY_TYPE_INFORMATION  gMemoryTypeInformation[EfiMaxMemoryType + 1] = {
 GLOBAL_REMOVE_IF_UNREFERENCED   BOOLEAN  gLoadFixedAddressCodeMemoryReady = FALSE;
 
 /**
+  Internal function.  Moves any memory descriptors that are on the
+  temporary descriptor stack to heap.
+
+**/
+VOID
+CoreFreeMemoryMapStack (
+  VOID
+  );
+
+/**
   Enter critical section by gaining lock on gMemoryLock.
 
 **/
@@ -142,6 +152,98 @@ RemoveMemoryMapEntry (
 }
 
 /**
+  Helper function to evaluate if memory regions intersect.
+
+  @param  Start1     The address of the first byte in the first memory region.
+  @param  End1       The address of the last byte in the first memory region.
+  @param  Start2     The address of the first byte in the second memory region.
+  @param  End2       The address of the last byte in the second memory region.
+
+  @return TRUE if the memory regions intersect, FALSE otherwise.
+**/
+STATIC
+BOOLEAN
+MemoryRegionsIntersect (
+  IN EFI_PHYSICAL_ADDRESS  Start1,
+  IN EFI_PHYSICAL_ADDRESS  End1,
+  IN EFI_PHYSICAL_ADDRESS  Start2,
+  IN EFI_PHYSICAL_ADDRESS  End2
+  )
+{
+  return (((Start1 <= End2) && (Start2 <= Start1)) ||
+          ((Start2 <= End1) && (Start1 <= Start2)));
+}
+
+/**
+  Get the memory bucket type for a given memory range.
+
+  @param  PhysicalStart  The address of the first byte in the memory region.
+  @param  PhysicalEnd    The address of the last byte in the memory region.
+
+  @return The memory type for the bucket that contains the given physical address range.
+          If the address range does not match any special bucket, it returns EfiMaxMemoryType.
+**/
+EFI_MEMORY_TYPE
+GetBucketMemoryType (
+  IN EFI_PHYSICAL_ADDRESS  PhysicalStart,
+  IN EFI_PHYSICAL_ADDRESS  PhysicalEnd
+  )
+{
+  EFI_MEMORY_TYPE  BucketType;
+
+  // Find the bucket type for the incoming memory region.
+  for (BucketType = (EFI_MEMORY_TYPE)0; BucketType < EfiMaxMemoryType; BucketType++) {
+    //
+    // If the number of pages for this memory type is not zero, the input region
+    // better be within the same bucket. We only care about the special memory type
+    // here because we need these buckets to remain consistent so that the OS resume
+    // logic can work properly. The same applies to the memory allocation logic.
+    //
+    if (mMemoryTypeStatistics[BucketType].Special && (mMemoryTypeStatistics[BucketType].NumberOfPages != 0)) {
+      if ((PhysicalStart >= mMemoryTypeStatistics[BucketType].BaseAddress) &&
+          (PhysicalEnd <= mMemoryTypeStatistics[BucketType].MaximumAddress))
+      {
+        break;
+      } else if (MemoryRegionsIntersect (
+                   PhysicalStart,
+                   PhysicalEnd,
+                   mMemoryTypeStatistics[BucketType].BaseAddress,
+                   mMemoryTypeStatistics[BucketType].MaximumAddress
+                   ))
+      {
+        // The start and end overlap the bucket, but not fully inclusive. We should not allow this.
+        DEBUG ((
+          DEBUG_ERROR,
+          "%a: %lx-%lx intersects bucket type %d (%lx-%lx)\n",
+          __func__,
+          PhysicalStart,
+          PhysicalEnd,
+          BucketType,
+          mMemoryTypeStatistics[BucketType].BaseAddress,
+          mMemoryTypeStatistics[BucketType].MaximumAddress
+          ));
+
+        ASSERT (FALSE);
+      }
+    }
+  }
+
+  // If we can find the bucket type, use it to guide the merging logic below.
+  // Otherwise, we will not care about the bucket type.
+  if (BucketType >= EfiMaxMemoryType) {
+    DEBUG ((
+      DEBUG_PAGE,
+      "%a: defaulting to max for %lx -%lx\n",
+      __func__,
+      PhysicalStart,
+      PhysicalEnd
+      ));
+  }
+
+  return BucketType;
+}
+
+/**
   Internal function.  Adds a ranges to the memory map.
   The range must not already exist in the map.
 
@@ -161,13 +263,74 @@ CoreAddRange (
   IN UINT64                Attribute
   )
 {
-  LIST_ENTRY  *Link;
-  MEMORY_MAP  *Entry;
+  LIST_ENTRY       *Link;
+  MEMORY_MAP       *Entry;
+  EFI_MEMORY_TYPE  BucketType;
+  EFI_MEMORY_TYPE  MergeType;
+  BOOLEAN          Break;
 
   ASSERT ((Start & EFI_PAGE_MASK) == 0);
   ASSERT (End > Start);
 
   ASSERT_LOCKED (&gMemoryLock);
+
+  // Find the bucket type for the incoming memory region.
+  Break = FALSE;
+  for (BucketType = (EFI_MEMORY_TYPE)0; BucketType < EfiMaxMemoryType; BucketType++) {
+    //
+    // If the number of pages for this memory type is not zero, the input region better
+    // be within the same bucket. Otherwise, we will handle the ones we care about,
+    // the special memory types, in chunks.
+    //
+    if (mMemoryTypeStatistics[BucketType].Special && (mMemoryTypeStatistics[BucketType].NumberOfPages != 0)) {
+      if ((Start <= mMemoryTypeStatistics[BucketType].MaximumAddress) &&
+          (End > mMemoryTypeStatistics[BucketType].MaximumAddress))
+      {
+        //
+        // The start overlaps the bucket, so we let self-recursion handle the tail, and we
+        // handle the head.
+        //
+        // |----------|---Special Memory Bucket---|----------|
+        // |--------------^---------------------------^------|
+        // |------------Start------------------------End-----|
+        //
+        CoreAddRange (
+          Type,
+          mMemoryTypeStatistics[BucketType].MaximumAddress + 1,
+          End,
+          Attribute
+          );
+        CoreFreeMemoryMapStack ();
+        End   = mMemoryTypeStatistics[BucketType].MaximumAddress;
+        Break = TRUE;
+      }
+
+      if ((Start < mMemoryTypeStatistics[BucketType].BaseAddress) &&
+          (End >= mMemoryTypeStatistics[BucketType].BaseAddress))
+      {
+        // The end overlaps the bucket, so we let self-recursion handle the head, and we
+        // handle the tail.
+        //
+        // |----------|---Special Memory Bucket---|----------|
+        // |------^-------------------^----------------------|
+        // |----Start----------------End---------------------|
+        //
+        CoreAddRange (
+          Type,
+          Start,
+          mMemoryTypeStatistics[BucketType].BaseAddress - 1,
+          Attribute
+          );
+        CoreFreeMemoryMapStack ();
+        Start = mMemoryTypeStatistics[BucketType].BaseAddress;
+        Break = TRUE;
+      }
+
+      if (Break) {
+        break;
+      }
+    }
+  }
 
   DEBUG ((DEBUG_PAGE, "AddRange: %lx-%lx to %d\n", Start, End, Type));
 
@@ -210,7 +373,8 @@ CoreAddRange (
   // and the same Attribute
   //
 
-  Link = gMemoryMap.ForwardLink;
+  MergeType = GetBucketMemoryType (Start, End);
+  Link      = gMemoryMap.ForwardLink;
   while (Link != &gMemoryMap) {
     Entry = CR (Link, MEMORY_MAP, Link, MEMORY_MAP_SIGNATURE);
     Link  = Link->ForwardLink;
@@ -221,6 +385,14 @@ CoreAddRange (
 
     if (Entry->Attribute != Attribute) {
       continue;
+    }
+
+    if (MergeType != EfiMaxMemoryType) {
+      // We are in the midst of merging memory descriptors, so we can only merge
+      // with the same type as the merge type.
+      if (MergeType != GetBucketMemoryType (Entry->Start, Entry->End)) {
+        continue;
+      }
     }
 
     if (Entry->End + 1 == Start) {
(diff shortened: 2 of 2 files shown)


kuqin12:
@lgao4 @mdkinney @ardbiesheuvel It would be helpful if you could provide some feedback on this change. Thanks.

ardbiesheuvel:
Happy to review but please explain what a bucket is in this context.

kuqin12:
> Happy to review but please explain what a bucket is in this context.

Thanks. I think the bucket is also called bin? It is an estimate for the usage of special memory types, such as reserved memory, ACPI reclaim, runtime, etc. So that the memory manager can use this as a reference to try to allocate the corresponding type of memory around the same address. This could potentially avoid fragmented memory, and "enhance the success rate" of OS hibernation resume.

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -779,12 +948,12 @@ CoreAddMemoryDescriptor (
     }
 
     if (gMemoryTypeInformation[Index].NumberOfPages != 0) {
+      mMemoryTypeStatistics[Type].NumberOfPages   = gMemoryTypeInformation[Index].NumberOfPages;
```
lgao4: Why make this change? Seemly, there is no difference.

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -161,14 +263,75 @@ CoreAddRange (
...
+      {
+        //
+        // The start overlaps the bucket, so we let self-recursion handle the tail, and we
+        // handle the head.
+        //
+        // |----------|---Special Memory Bucket---|----------|
+        // |--------------^---------------------------^------|
+        // |------------Start------------------------End-----|
+        //
+        CoreAddRange (
```
lgao4: Is there the real case to enter this condition?

kuqin12: Yes, this function call could happen when regions are added to the memory map list through GCD entries, i.e. promoting memory regions to be allocable through `PromoteMemoryResource`.

I think instead of manipulating the logic in other places, this CoreAddRange is a bottom layer that can handle both the breakups and merging in one place.

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: The main change here is the order. This ensures the number of pages for `mMemoryTypeStatistics` to be set when the free pages is called below, which will avoid unintended merging of memory regions during these calls.

kuqin12:
@lgao4 Do you have any further comments on this change?

lgao4 (approved):

lgao4:
> @lgao4 Do you have any further comments on this change?

I have no other comments. I just approve this change.

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -141,6 +151,98 @@ RemoveMemoryMapEntry (
...
+BOOLEAN
+MemoryRegionsIntersect (
+  IN EFI_PHYSICAL_ADDRESS  Start1,
+  IN EFI_PHYSICAL_ADDRESS  End1,
+  IN EFI_PHYSICAL_ADDRESS  Start2,
+  IN EFI_PHYSICAL_ADDRESS  End2
+  )
+{
+  return (((Start1 <= End2) && (Start2 <= Start1)) ||
+          ((Start2 <= End1) && (Start1 <= Start2)));
```
ardbiesheuvel: This can be simplified to `Start1 < End2 && Start2 < End1` but we should add an ASSERT() that Start1 <= End1 and Start2 <= End2

ardbiesheuvel (commented):
This is still lacking an explanation of what buckets are, and what the expected behavior is.

AIUI, memory buckets are reserved regions from which memory of a certain type can be allocated, right? So why is it an error for memory belonging to a special memory bucket to be reported as free if it can be allocated?

Without a good overview (and consensus) of what memory buckets are and how they are supposed to work, I don't think we should merge this patch.

kuqin12:
> AIUI, memory buckets are reserved regions from which memory of a certain type can be allocated, right? So why is it an error for memory belonging to a special memory bucket to be reported as free if it can be allocated?
>
Yes, that's correct. If the memory map reports a region belonging to a special memory bucket as free, it may appear to be available for general allocation. However, if an allocation request in that region is made for a different memory type, the allocator will return EFI_NOT_FOUND, because that region is reserved exclusively for its designated memory type.

In essence, bucket memory is allocatable—but only when the requested memory type matches the bucket's assigned type. Allowing the memory map to merge free regions across different types would obscure the boundaries of these memory buckets. This could lead to unpredictable allocation failures, depending on the system's memory layout.
> 
> Without a good overview (and consensus) of what memory buckets are and how they are supposed to work, I don't think we should merge this patch.
>
Memory buckets are used in the EFI DXE Core to ensure that sufficient memory is provided for specific EFI memory types. This mechanism helps maintain a consistent and predictable memory layout across boots, which is especially important for features like OS hibernation that rely on memory layout stability.

During memory service initialization, the DXE Core checks for the presence of a `gMemoryTypeInformation` HOB. If found, it uses the specified number of pages for each memory type to pre-allocate memory regions accordingly. These regions are then marked in the memory map as special "bucket" types.

Once established, the memory bucket mechanism prioritizes allocation requests for those memory types from their respective pre-"allocated" regions—provided there is still available space in the bucket. This helps prevent fragmentation and ensures that memory types remain in consistent physical locations across reboots.

When GetMemoryMap() is called, the memory type of each bucketed region is overridden with its bucket type. This allows consumers—typically the OS—to apply special handling, such as enforcing memory attributes, checking alignment, or caching content.

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: I agree that the two comparisons follow the same logic as the original change. However, adding two ASSERTs to ensure the assumption seems to go against the goal of simplifying the code. Instead, that might make it harder to read and follow?

kuqin12:
@ardbiesheuvel @mdkinney Do you happen to have further comments on this PR? Thanks in advance!

kuqin12:
@ardbiesheuvel @mdkinney A gentle ping on this PR. Is there anything else I can clarify on the change or the memory bin/bucket feature itself?

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: Updated.

kuqin12:
@mdkinney @ardbiesheuvel Another gentle ping on this PR. Now that we are out of freeze, can you please provide feedback or approval on this change? Thanks in advance.

mdkinney:
I also was confused by the term "bucket".  The DXE Core refers to these as "Memory Type bins".

I recommend we use the same terminology consistently in the code and the PR summary and the commit messages.

It is also no clear to me the use case that has the incorrect behavior that this PR is attempting to fix.  Please provide a complete example that shows the initial state of the bins, the requested operation, the result from the current logic and why it is incorrect, and the state with the updated logic and why it is correct.  That should be part of the PR summary.

kuqin12:
> I also was confused by the term "bucket". The DXE Core refers to these as "Memory Type bins".
> 
> I recommend we use the same terminology consistently in the code and the PR summary and the commit messages.
> 
> It is also no clear to me the use case that has the incorrect behavior that this PR is attempting to fix. Please provide a complete example that shows the initial state of the bins, the requested operation, the result from the current logic and why it is incorrect, and the state with the updated logic and why it is correct. That should be part of the PR summary.

@mdkinney thanks for the suggestion. The `bucket` term is replaced with `bin`s as suggested.

I also updated the PR description with an example and how the change will update the behavior in this case. Please let me know if there is any further clarification you would prefer to add. Thanks in advance!

kuqin12:
@mdkinney would you happen to have further feedback on this change and the PR description? Thanks.

kuqin12:
@ardbiesheuvel @mdkinney a gentle ping on this PR. Do you have any other feedback? Hopefully the PR description provides better context on the issue this change is trying to fix. Thanks in advance!

mdkinney:
@kuqin12 Thank you for the detailed example.  I understand the corner case that needs to accommodated now.  I will review the logic changes with this example in mind.

Please update the PR title and PR description replacing "bucket" with "bin"

kuqin12:
> @kuqin12 Thank you for the detailed example. I understand the corner case that needs to accommodated now. I will review the logic changes with this example in mind.
> 
> Please update the PR title and PR description replacing "bucket" with "bin"

PR title is updated. Thanks in advance!

mdkinney:
@kuqin12 The main issue that this PR raises is available memory between the bins. The original algorithm for the bins never generated this case because alignment of all bins was 4KB.  It seems that the system requirements for some of the bins to have 64KB alignment and some 4KB has introduced this new scenario.

No DXE Core memory manager changes would be required of there were no gaps between the bins.  I can see 2 ways to preserve this no gap condition:
* All bins are required to use the same alignment (e.g. 64 KB)
* Sort the order of bins from largest alignment to smallest alignment.

Were these options considered from the platform configuration of the bin type/size/alignment perspective?

Could these platform configuration requirements for the bins allow this PR to be dropped?

kuqin12:
> @kuqin12 The main issue that this PR raises is available memory between the bins. The original algorithm for the bins never generated this case because alignment of all bins was 4KB. It seems that the system requirements for some of the bins to have 64KB alignment and some 4KB has introduced this new scenario.
> 
> No DXE Core memory manager changes would be required of there were no gaps between the bins. I can see 2 ways to preserve this no gap condition:
> 
> * All bins are required to use the same alignment (e.g. 64 KB)
> * Sort the order of bins from largest alignment to smallest alignment.
> 
> Were these options considered from the platform configuration of the bin type/size/alignment perspective?
> 
> Could these platform configuration requirements for the bins allow this PR to be dropped?

@mdkinney Thanks for the feedback. 64KB requirements for runtime related memory comes from his commit: https://github.com/tianocore/edk2/commit/08855193cafebaa2e071ffb5f083bcb3cb6d4df0. Maybe @ardbiesheuvel or @os-d can chime in on the requirement.

But I think requiring the alignment seems a bit overboard to work around the merging logic. Sorting the bins might be a feasible alternative to solve the issue.

On the other hand, I would recommend keeping the check of overlapped memory regions and bins, at least the one inside CoreGetMemoryMap, to ensure the returned memory map is sane.

mdkinney:
Checks for good memory maps seems like a good idea.  Including checks for gaps between the bins.

mdkinney:
Also, the purpose of the MemoryTypeInformationHob is to provide the information to the DXE Core to produce a memory map that is not fragmented to support ACPI S4.  By allowing regions of available memory that can be allocated for BsData/BsCode/LoaderCode/LoaderData between the bins allows the memory map to be fragmented.  The simplest possible memory map is one where only the memory types that are required to be consistently located are described by the MemoryTypeInformationHob and the allocation of those memory type bins are in a single contiguous region of memory with no chance of gaps.

os-d:
Per UEFI spec (https://uefi.org/specs/UEFI/2.10_A/02_Overview.html#aarch64-platforms), AARCH64 cannot have mixed mappings in runtime 64 KB pages (because the arch supports a base page size of 16KB and 64KB, which Linux does have some distros that use it). Ard (and I'll let him speak deeper to this) chose to implement that as runtime types must have a 64KB runtime allocation granularity, which then bubbles into the bins.

image

I'm forgetting all of the nuances between the existing bin algorithm and this proposed new one since the PRs been up for a couple months now, but I think the proposal of ordering the bins and still checking sanity later makes sense.

Will there still be issues with the bins on the outside of the range? e.g. if the first memory range in the first bin gets freed, will it get merged with EfiConventionalMemory prior to the bin or will the existing logic reserve it for the bin? Same with the last memory region in the last bin. I recall when we were working on the new algorithm there were issues like that popping up.

kuqin12:
@mdkinney I took another look and implemented a naive version earlier Wednesday. The sorting change is not pushed because there are 2 issues that prevent the fix from resolving the original flaw with this approach:

1. The `CoreAllocatePages` does the alignment logic internally, which means that it potentially allocates a larger memory block than the bin logic requested. But the statistics will honor the NumberOfPages from the hob, rather than the really allocated bin size. In order to pass the no-gap check and ensure memory map does not sneak in other memory into padded regions, we could patch the same logic inside the memory initialization routine, and make the statistic structure to use the aligned number of pages. Then the check for gaps could pass. But this feels like logic leakage and does not seem ideal.
2. Another issue is that even if we fix the above, a freed memory right next to the last allocated bin will cause the merge logic to engage. This will then trip on memory bin and another memory type overlap issue. i.e. reusing the example we have, if a memory region at `0x857FF000 - 0x857FFFFF` is freed, and memory type 5 bin is completely empty. The freed memory region will be merged to memory type 5 bin region and creating a `0x857FF000 - 0x85EFFFFF` entry.

Please let me know if you have any suggestions

kuqin12:
@mdkinney could you please let me know if you have other suggestions? Otherwise the current version seems to be _the_ solution that can resolve the original issue.

kuqin12:
@mdkinney would you happen to have any feedback on the 2 issues I mentioned above?

https://github.com/tianocore/edk2/pull/11273#issuecomment-3279249551

mdkinney:
My only other idea is to add guard bins of type EfiUnusableMemory or an OEM Reserved Memory Type at the top and bottom.  That would guarantee no incorrect merging condition would ever need to be handled.

My main concern with this patch is the complexity of the changes and how do we know these changes do not introduce any regressions.  That is why I am making sure we work through all options here.

kuqin12:
> My only other idea is to add guard bins of type EfiUnusableMemory or an OEM Reserved Memory Type at the top and bottom. That would guarantee no incorrect merging condition would ever need to be handled.
> 
@mdkinney thanks for the suggestion. I tried to look at this option earlier. I had to consume a combined 65KB for both the before and after guard bins. (because the before guard needs to be contiguous with the first bin, so the before-bin could need 64KB). It is less than ideal, but there is a path forward.

But the next issue that I observed was the freed special bins themselves. Once they are freed at the end of the initialization routine, they are already merged back into a large conventional region. If a GetMemoryMap query comes in before any page for a given special bin (that is 0 pages allocated for that type), the empty bin will still belong to a large conventional bin across multiple memory types. Would the next option be to allocate a separator bin between every special bins?
> My main concern with this patch is the complexity of the changes and how do we know these changes do not introduce any regressions. That is why I am making sure we work through all options here.
>
I totally understand your concern. But with all the options we have been through, is it really worthy to keep down the path in order to keep using the current questionable merge logic?

I will keep on the path for separator bin option in the meantime.

kuqin12:
@mdkinney here is a draft of the logic that summarized what was discussed + additional changes to encompass the edge cases: https://github.com/tianocore/edk2/pull/11544.

As I mentioned earlier, the new draft includes a separator between each bin to ensure the regions are contiguous. But this would be some memory that is "reserved" for the corresponding bin. Please let me know your thoughts. I can move that change over here if we agree to proceed with this path.

Thanks in advance.

kuqin12:
@mdkinney just checking in to see if you have any thoughts on the other draft? It does not seem simple either and wastes one page (plus alignment) for each involved memory type. Please let me know your thoughts.

mdkinney:
@kuqin12  I do not like the idea of wasting pages.  So the concept of the original PR seems like the best approach to honor the bins allocated and prevent merging of free pages the cause a bin boundary to be in the middle of a freed region.  I need to go back and review all the logic changes and touch points in your original proposal and see if there is a way to make it simpler.  Perhaps some helper functions, or perhaps some additional meta data in the internal memory manager structures.

kuqin12:
> @kuqin12 I do not like the idea of wasting pages. So the concept of the original PR seems like the best approach to honor the bins allocated and prevent merging of free pages the cause a bin boundary to be in the middle of a freed region. I need to go back and review all the logic changes and touch points in your original proposal and see if there is a way to make it simpler. Perhaps some helper functions, or perhaps some additional meta data in the internal memory manager structures.

@mdkinney Thanks. I will close the other PR. Please let me know if you have any feedback on the original change.

kuqin12:
@mdkinney the other PR is closed. Please let me know if you have feedback/thoughts on this change. Thanks in advance.

MdeModulePkg/Core/Dxe/Mem/Imem.h
```diff
@@ -146,6 +146,21 @@ CoreInternalAllocatePages (
...
+  Get the memory bin type for a given memory range.
+
+  @param  PhysicalStart  The address of the first byte in the memory region.
+  @param  PhysicalEnd    The address of the last byte in the memory region.
+
+  @return The memory type for the bin that contains the given physical address range.
+          If the address range does not match any special bin, it returns EfiMaxMemoryType.
+**/
+EFI_MEMORY_TYPE
+GetMemoryBinType (
```
mdkinney: Since this function is only used with Page.c, it can be removed from Imem.h and this function can be declared as static in Page.c

MdeModulePkg/Core/Dxe/Mem/Page.c
mdkinney: This changes CoreAddRange() to be recursive.  Can the logic be implemented without recursion?

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -161,14 +266,75 @@ CoreAddRange (
...
+          (End >= mMemoryTypeStatistics[BinType].BaseAddress))
+      {
+        // The end overlaps the bin, so we let self-recursion handle the head, and we
+        // handle the tail.
+        //
+        // |----------|--- Special Memory Bin ----|----------|
+        // |------^-------------------^----------------------|
+        // |----Start----------------End---------------------|
+        //
+        CoreAddRange (
```
mdkinney: This changes CoreAddRange() to be recursive.  Can the logic be implemented without recursion?

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -2028,6 +2200,30 @@ CoreGetMemoryMap (
...
+        } else if (mMemoryTypeStatistics[Type].Special &&
+                   (mMemoryTypeStatistics[Type].NumberOfPages > 0) &&
+                   MemoryRegionsIntersect (
+                     Entry->Start,
+                     Entry->End,
+                     mMemoryTypeStatistics[Type].BaseAddress,
+                     mMemoryTypeStatistics[Type].MaximumAddress
+                     ))
+        {
+          // There is partial overlap with a special memory type bin.
```
mdkinney, line 2495: How would this condition occur of the rest of the logic in this PR is correct?  Would this only occur of the internal memory map structures are corrupt?

mdkinney:
Have any GoogleTest based unit tests been implemented to verify that the logic changes here resolve all the possible bin layout and adjacent memory alloc/free conditions?

MdeModulePkg/Core/Dxe/Mem/Imem.h
kuqin12: Thanks, this is removed.

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: Changed it to be local variable-based functions and non-recursive.

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: This check is in place to ensure the returned memory is sane, because this is what the non-core consumers will be consuming.

This overlap should not occur, if all memory map manipulation goes through the `CoreAddRange` function. But I did notice that whenever the memory map is being "converted" during allocation or free, that routine does not go through this same function.

kuqin12:
> Have any GoogleTest based unit tests been implemented to verify that the logic changes here resolve all the possible bin layout and adjacent memory alloc/free conditions?

@mdkinney A google test was added to test the added function `SplitIncomingRange`, and existing `CoreAddRange` as well as `CoreGetMemoryMap`.

kuqin12:
@mdkinney hope you had a great holiday. Please let me know if you have any feedback for the updated patches. Thanks in advance.

kuqin12:
This change is superseded by https://github.com/tianocore/edk2/pull/11972. A follow up PR will be created to check the overlaps shortly.

kuqin12:
> This change is superseded by #11972. A follow up PR will be created to check the overlaps shortly.

A new PR was created here: https://github.com/tianocore/edk2/pull/12076/changes

PR #11274, opened 2025-07-09
EmbeddedPkg: Ensure variables initialized.

apop5:
# Description
Ensure that variables are initialized prior to their use. There is a codepath that would result in an uninitialized variable being used. Caught by ARM64 compiler.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Local CI and running on virtual platform.

## Integration Instructions
No integration necessary.

### DIFF
--- a/EmbeddedPkg/Library/PrePiLib/FwVol.c
+++ b/EmbeddedPkg/Library/PrePiLib/FwVol.c
@@ -296,10 +296,13 @@ FfsProcessSection (
   UINT32                    CompressedDataLength;
   BOOLEAN                   Found;
 
-  Found         = FALSE;
-  *OutputBuffer = NULL;
-  ParsedLength  = 0;
-  Status        = EFI_NOT_FOUND;
+  Found             = FALSE;
+  *OutputBuffer     = NULL;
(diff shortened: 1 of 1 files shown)


EmbeddedPkg/Library/PrePiLib/FwVol.c
```diff
@@ -296,10 +296,13 @@ FfsProcessSection (
   UINT32                    CompressedDataLength;
   BOOLEAN                   Found;
 
-  Found         = FALSE;
```
leiflindholm, line 299: Subject missing affected module.

"There is a codepath that would result in an uninitialized
variable being used."
Please point it out.

"Caught by ARM64 compiler."
Which one, what version, what was the error message produced? In all build targets or just a specific one?

This function is an unreadable maze, and this patch may well make some error messages go away for a confused compiler. But if all we're doing is making error messages go away, we should at least leave a clear record of that.

A better fix would be breaking the function down into several helper functions so that neither programmers nor compilers get confused by reading it.

apop5:
Closed in favor of #11279

Issue #9256, opened 2022-11-21
Deferring EOI until after RestoreTPL() causes system time to stop (Bugzilla Bug 4162)


tianocore-issues:
This issue was created automatically with [bugzilla2github](https://github.com/makubacki/bugzilla2github/tree/bz2gh_tianocore)
# Bugzilla Bug 4162
Date: 2022-11-21T10:47:36+00:00
From: Michael Brown &lt;<mcb30>&gt;
To:   unassigned &lt;<unassigned>&gt;
CC:   @ajfish, igor.druzhinin, @lersek, @mdkinney, pbonzini

See also:     https://bugzilla.tianocore.org/show_bug.cgi?id=2815
Last updated: 2022-12-23T12:00:59+00:00

tianocore-issues:
## Comment 20070
Date: 2022-11-21 10:47:36 +0000
From: Michael Brown &lt;<mcb30>&gt;

- Industry Specification: ---
- Release Observed: ['EDK II Master', 'edk2-stable202008', 'edk2-stable202011', 'edk2-stable202102', 'edk2-stable202105']
- Releases to Fix: EDK II Master
- Target OS: ---
- Bugzilla Assignee(s): unassigned &lt;<unassigned>&gt;

Commit 239b50a86 (&quot;OvmfPkg: End timer interrupt later to avoid stack overflow under load&quot;) deferred the call to SendApicEoi() until after the call to RestoreTPL(), in order to avoid the potential stack underflow as described in bug 2815.

This has the unwanted side effect that any callbacks invoked as a result of RestoreTPL() will now run before the EOI is sent to the APIC.  Until the callbacks return, no further timer interrupts will be raised and the system time value in mEfiSystemTime will remain unchanging.

If any callbacks invoked by RestoreTPL() include a timeout loop based on mEfiSystemTime (e.g. via SetTimer() and CheckEvent()), this loop will never time out and the system will appear to hang.

This situation can be reproduced by e.g. attaching a USB network card connected via an xHCI USB controller.  The &quot;system poll timer&quot; callback to MnpSystemPoll() will eventually cause the following sequence of events:

- Hardware timer interrupt occurs
- TimerInterruptHandler() in LocalApicTimerDxe is called
- TimerInterruptHandler() calls RaiseTPL(TPL_HIGH_LEVEL)
- TimerInterruptHandler() calls RestoreTPL() before sending EOI
- RestoreTPL() triggers a call to MnpSystemPoll()
[...]

PR #11276, opened 2025-07-10
Fix stack overflow issue due to nested interrupts

xuweiintel:
# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions


kraxel:
Filling the description with some actual content would be very nice, especially on testing.

The patches look good to me.

@mcb30 FYI

mcb30:
Looks plausible to me.  Has it been tested in the two use cases that originally caused problems?
* Using a USB network device.
* Executing an EFI application that illegally enables interrupts (via a direct `sti` instruction) while executing at `TPL_HIGH_LEVEL`.

xuweiintel:
> Looks plausible to me. Has it been tested in the two use cases that originally caused problems?
> 
> * Using a USB network device.
> * Executing an EFI application that illegally enables interrupts (via a direct `sti` instruction) while executing at `TPL_HIGH_LEVEL`.

I will try the above two cases and update the result here later.

ardbiesheuvel:
@leiflindholm @samimujawar

The generic changes in this PR affect all architectures, not just X86.

MdeModulePkg/Core/Dxe/Event/Tpl.c
```diff
@@ -9,6 +9,16 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include "DxeMain.h"
 #include "Event.h"
 
+//
+// Bit mask of TPLs that were interrupted (typically during RestoreTPL's
+// event dispatching, though there are reports that the Windows boot loader
+// executes stray STIs at TPL_HIGH_LEVEL).  CoreRaiseTpl() sets the
```
ardbiesheuvel, line 15: Could we please avoid X86-specific lingo here? This code is shared by all architectures, and terms like `STI` and `IRET` may not be familiar to everyone.

MdeModulePkg/Core/Dxe/Event/Tpl.c
```diff
@@ -9,6 +9,16 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
...
 
+//
+// Bit mask of TPLs that were interrupted (typically during RestoreTPL's
+// event dispatching, though there are reports that the Windows boot loader
+// executes stray STIs at TPL_HIGH_LEVEL).  CoreRaiseTpl() sets the
+// OldTpl-th bit when it detects it was called from an interrupt handler,
+// because the corresponding CoreRestoreTpl() needs different semantics for
+// the CPU interrupt state.  See CoreRaiseTpl() and CoreRestoreTpl() below.
+//
+static UINTN  mInterruptedTplMask = 0;
```
ardbiesheuvel, line 20: This should be either `UINT32` or `UINT64`  - if `UINT32` is not large enough, nor will `UINTN` on 32-bit architectures.

MdeModulePkg/Core/Dxe/Event/Tpl.c
```diff
@@ -72,7 +83,31 @@ CoreRaiseTpl (
   // If raising to high level, disable interrupts
   //
   if ((NewTpl >= TPL_HIGH_LEVEL) &&  (OldTpl < TPL_HIGH_LEVEL)) {
-    CoreSetInterruptState (FALSE);
+    //
+    // When gCpu is NULL, assume we're not called from an interrupt handler.
```
ardbiesheuvel, line 87: Why is this assumption safe?

MdeModulePkg/Core/Dxe/Event/Tpl.c
```diff
@@ -72,7 +83,31 @@ CoreRaiseTpl (
...
+      // Interrupts are currently enabled.
+      // Keep them disabled while at TPL_HIGH_LEVEL.
+      //
+      CoreSetInterruptState (FALSE);
+    } else {
+      //
+      // Within an interrupt handler.  Save the TPL that was interrupted;
+      // It must be higher than the previously interrupted TPL, since
+      // CoreRestoreTpl reset all bits up to and including the requested TPL.
+      //
```
ardbiesheuvel, line 107: We are conflating `interrupts are disabled` with `running within an interrupt handler` here. This needs justification, because not all CPU arch protocol implementations may behave the same in this regard.

UefiCpuPkg/CpuDxe/CpuDxe.c
niruiyu, line 1: When I reread the commit message, I found it could be a bit easier to understand if reworded as follows:

X86 CpuDxe driver uses a global C variable to record the interrupt state.
The global C variable is updated every time CpuArch.EnableInterrupt() or
CpuArch.DisableInterrupt() is called.
CpuArch.GetInterruptState() returns the global C variable value.

But CpuArch.GetInterruptState() incorrectly returns TRUE when it is called
in the interrupt context.
The reason is that the C global variable indicating the interrupt state
remains TRUE after the interrupt is disabled by CLI from the CpuExceptionHandlerLib.

The commit removes the C global variable and always reads the
interrupt state from the CPU register (EFLAGS).

MdeModulePkg/Core/Dxe/Event/Tpl.c
niruiyu: agree. we could just say "Windows boot loader disables CPU interrupt at TPL_HIGH_LEVEL"

MdeModulePkg/Core/Dxe/Event/Tpl.c
niruiyu: The type chooses "UINTN" as the existing flag "gEventPending" uses "UINTN".

MdeModulePkg/Core/Dxe/Event/Tpl.c
niruiyu: gCpu (CpuArch protocol) is produced by the one who sets up the interrupt handler.
When gCpu is NULL, the interrupt handler is not set up.

MdeModulePkg/Core/Dxe/Event/Tpl.c
niruiyu: The code assumes that the CPU interrupt is disabled as soon as the interrupt handler runs.
So, when interrupts are disabled, the code assumes running in the interrupt context.
Can you explain what else the implementation could be? Or suggest a better change?

eblipp:
This pull request (PR) makes substantial changes to how TPL (Task Priority Level) and interrupt nesting are handled in the EDK2 UEFI firmware codebase, particularly targeting the DXE event system and related OVMF modules. Here’s a detailed review of the modifications:

---

## Summary of Changes

### 1. Core TPL/Interrupt Handling (MdeModulePkg/Core/Dxe/Event/Tpl.c)
- **New Mechanism for Nested Interrupts:**  
  - Introduces a static variable mInterruptedTplMask to keep track of which TPLs were interrupted, improving the handling of nested interrupts.
  - CoreRaiseTpl() and CoreRestoreTpl() are updated to set and clear bits in this mask based on interrupt entry/exit, allowing for more precise and robust restoration of the CPU’s interrupt state.
  - Adds extensive comments explaining the reasoning and edge cases, such as handling stray STIs and nested interrupt behavior.
- **More Accurate Interrupt State Handling:**  
  - The previous implementation assumed a straightforward enable/disable logic, but the new version queries the actual interrupt state when possible.
  - CoreSetInterruptState() is called with more carefully determined values, based on whether the code is running in an interrupt context.

### 2. Removal of NestedInterruptTplLib
- **Deleted Files:**  
  - OvmfPkg/Library/NestedInterruptTplLib/Iret.c, Iret.h, Tpl.c, and NestedInterruptTplLib.inf are removed.
- **Removed Usage:**  
  - All references to NestedInterruptTplLib are deleted from LocalApicTimerDxe and DSC/INF configuration files.
  - The timer interrupt handler in LocalApicTimerDxe switches from using NestedInterruptRaiseTPL/NestedInterruptRestoreTPL to the standard gBS->RaiseTPL/gBS->RestoreTPL.

### 3. Clean-up and Refactoring
- **UefiCpuPkg/CpuDxe/CpuDxe.c:**  
  - Removes the global InterruptState variable, instead directly querying the hardware with GetInterruptState().
  - CpuEnableInterrupt and CpuDisableInterrupt now simply enable/disable interrupts without tracking state in software.
  - CpuGetInterruptState returns the queried hardware state.
- **Build Configuration:**  
  - Removes NestedInterruptTplLib from OvmfPkg/IntelTdx/IntelTdxX64.dsc and OvmfPkg/OvmfPkgIa32X64.dsc.

---

## Strengths

- **Correctness and Robustness:**  
  The new implementation is more robust in handling nested interrupts, reducing the risk of stack overflows and subtle bugs due to incorrect TPL/interrupt state transitions.
- **Clarity:**  
  The added comments are thorough, explaining the rationale and edge cases in detail. This is very helpful for future maintainers.
- **Code Simplification:**  
  By removing the NestedInterruptTplLib and centralizing logic in CoreRaiseTpl/CoreRestoreTpl, the codebase is easier to follow and maintain.
- **Consistency:**  
  The change unifies the way TPL and interrupts are managed across the codebase, using standard UEFI interfaces instead of custom library functions.

---

## Areas to Consider / Suggestions

- **Testing Coverage:**  
  Given the sensitivity of interrupt and TPL management in firmware, extensive regression and stress testing on affected platforms (especially OVMF and TDX) is critical. Bugs here can cause hard-to-debug firmware hangs or security issues.
- **Documentation:**  
  The inline comments are excellent, but consider updating any external documentation or developer guides that reference NestedInterruptTplLib or describe interrupt/TPL handling.
- **Edge Cases:**  
  While the code appears to handle nested interrupts and interrupted TPLs properly, reviewers should pay special attention to platform-specific interrupt controller quirks or unusual OS loader behaviors (some are mentioned in the comments).
- **Performance:**  
  There may be a slight overhead in querying the actual interrupt state rather than relying on a cached value, but this is likely negligible compared to the correctness gains.

---

## Conclusion

This PR is a significant improvement to the EDK2 codebase’s handling of TPLs and nested interrupts, especially for the OVMF platform. The changes are well-motivated, thoroughly commented, and remove unnecessary complexity. As long as regression testing passes and no platform-specific issues are uncovered, this should be a very positive change.

Let me know if you want a line-by-line breakdown or have specific concerns about any part of the diff!

ardbiesheuvel:
@eblipp please stop pasting LLM generated drivel into random PRs

MdeModulePkg/Core/Dxe/Event/Tpl.c
ardbiesheuvel: Fair enough

sarah-walker-arm:
Looks okay for Arm, besides the already raised issue around x86isms in comments.

MdeModulePkg/Core/Dxe/Event/Tpl.c
mcb30: > agree. we could just say "Windows boot loader disables CPU interrupt at TPL_HIGH_LEVEL"

**Enables**, not disables.  Interrupts are supposed to be disabled at TPL_HIGH_LEVEL, and EDK2 assumes that this invariant will always hold.  The Windows bug is that it first raises to TPL_HIGH_LEVEL (which is already illegal: that level is specified to be "for use exclusively by the firmware") and then enables interrupts (which breaks the firmware's invariant assumption).

mdkinney:
@xuweiintel Please fix CI issues.

@niruiyu are any more changes required to merge?

xunfengd:
> > Looks plausible to me. Has it been tested in the two use cases that originally caused problems?
> > 
> > * Using a USB network device.
> > * Executing an EFI application that illegally enables interrupts (via a direct `sti` instruction) while executing at `TPL_HIGH_LEVEL`.
> 
> I will try the above two cases and update the result here later.

For the test2,
we have validated pass,
``` C
  OldTpl = gBS->RaiseTPL (TPL_HIGH_LEVEL);

  EFlags.UintN = AsmReadEflags ();
  DEBUG((DEBUG_ERROR,"[PXE][Function:%a]\n",__FUNCTION__));
  DEBUG((DEBUG_ERROR,"[PXE] EFlags.UintN = %x\n",EFlags.UintN));
  DEBUG((DEBUG_ERROR,"[PXE] EFlags.Bits.IF = %x Tpl: %x\n",EFlags.Bits.IF, OldTpl));

 
  EnableInterrupts ();

  EFlags.UintN = 0x0;
  EFlags.UintN = AsmReadEflags ();
  DEBUG((DEBUG_ERROR,"[PXE][After] [Function:%a] Enable IF by STI \n",__FUNCTION__));
  DEBUG((DEBUG_ERROR,"[PXE][After] EFlags.UintN = %x\n",EFlags.UintN));
  DEBUG((DEBUG_ERROR,"[PXE][After] EFlags.Bits.IF = %x Tpl: %x\n",EFlags.Bits.IF, OldTpl));
  
  gBS->RestoreTPL (OldTpl);
```
image

For test1, we're still reproducing it.

mcb30:
> For the test2, we have validated pass,
> 
> ```c
>   OldTpl = gBS->RaiseTPL (TPL_HIGH_LEVEL);
> 
>   EFlags.UintN = AsmReadEflags ();
>   DEBUG((DEBUG_ERROR,"[PXE][Function:%a]\n",__FUNCTION__));
>   DEBUG((DEBUG_ERROR,"[PXE] EFlags.UintN = %x\n",EFlags.UintN));
>   DEBUG((DEBUG_ERROR,"[PXE] EFlags.Bits.IF = %x Tpl: %x\n",EFlags.Bits.IF, OldTpl));
> 
>  
>   EnableInterrupts ();
> 
>   EFlags.UintN = 0x0;
>   EFlags.UintN = AsmReadEflags ();
>   DEBUG((DEBUG_ERROR,"[PXE][After] [Function:%a] Enable IF by STI \n",__FUNCTION__));
>   DEBUG((DEBUG_ERROR,"[PXE][After] EFlags.UintN = %x\n",EFlags.UintN));
>   DEBUG((DEBUG_ERROR,"[PXE][After] EFlags.Bits.IF = %x Tpl: %x\n",EFlags.Bits.IF, OldTpl));
>   
>   gBS->RestoreTPL (OldTpl);
> ```

Thank you.  How do we tell in this test that an interrupt actually occurred between `EnableInterrupts()` and `RestoreTPL()`?

gumingyun:
> > > Looks plausible to me. Has it been tested in the two use cases that originally caused problems?
> > > 
> > > * Using a USB network device.
> > > * Executing an EFI application that illegally enables interrupts (via a direct `sti` instruction) while executing at `TPL_HIGH_LEVEL`.
> > 
> > 
> > I will try the above two cases and update the result here later.
> 
> For the test2, we have validated pass,
> 
> ```c
>   OldTpl = gBS->RaiseTPL (TPL_HIGH_LEVEL);
> 
>   EFlags.UintN = AsmReadEflags ();
>   DEBUG((DEBUG_ERROR,"[PXE][Function:%a]\n",__FUNCTION__));
>   DEBUG((DEBUG_ERROR,"[PXE] EFlags.UintN = %x\n",EFlags.UintN));
>   DEBUG((DEBUG_ERROR,"[PXE] EFlags.Bits.IF = %x Tpl: %x\n",EFlags.Bits.IF, OldTpl));
> 
>  
>   EnableInterrupts ();
> 
>   EFlags.UintN = 0x0;
>   EFlags.UintN = AsmReadEflags ();
>   DEBUG((DEBUG_ERROR,"[PXE][After] [Function:%a] Enable IF by STI \n",__FUNCTION__));
>   DEBUG((DEBUG_ERROR,"[PXE][After] EFlags.UintN = %x\n",EFlags.UintN));
>   DEBUG((DEBUG_ERROR,"[PXE][After] EFlags.Bits.IF = %x Tpl: %x\n",EFlags.Bits.IF, OldTpl));
>   
>   gBS->RestoreTPL (OldTpl);
> ```
> 
> image
> For test1, we're still reproducing it.

For the test1, we have tested default BIOS and BIOS with patch, from the result test1 validated pass too.
USB network device MAC: 0C3796CC6DDA
image
The result as below:
image
image

mcb30:
The testing seems quite sketchy to me, compared to the testing that was done for NestedInterruptTplLib.  I would like to have some confidence that:
1. The testers have **understood** the original two problem cases that I mentioned.
2. The testers have managed to **reproduce** the original problem cases, with 100% repeatability.  (Note that reproducing the problem will necessarily require a patch to disable the use of NestedInterruptTplLib, and the details of this patch should ideally be provided as part of the test results.)
3. The testers have verified that the precise same tests that **fail** when deliberately reproducing the problem will pass when this PR is applied, with absolutely no other changes.

Otherwise, the testing documented so far is just at the level of "it seems to work ok", which is really not sufficient for something with this level of subtlety and complexity.

mdkinney:
@mcb30 Do you have a link to the testing that was done for the NestedInterruptTplLib so we can rerun the same tests?

I think we should also consider some unit tests to make sure the scenarios in Case 2 are always covered so we never introduce a regression.

For Case 1, USB Network Stack, is there a link to the root cause analysis for that scenario, so we can potentially implement unit tests for that general case as well?

mcb30:
> @mcb30 Do you have a link to the testing that was done for the NestedInterruptTplLib so we can rerun the same tests?

It's unfortunately all handwritten on paper.  I'm happy to share copies of the notebooks, if it would help.

> I think we should also consider some unit tests to make sure the scenarios in Case 2 are always covered so we never introduce a regression.
> 
> For Case 1, USB Network Stack, is there a link to the root cause analysis for that scenario, so we can potentially implement unit tests for that general case as well?

There's a very long conversation about it in the original bugzilla, which is now at https://github.com/tianocore/edk2/issues/9256

gumingyun:
For the test2, trigger software GPE by setting register SWGPE_CTRL bit17 to 1. And read register GPE0_STS_127_96, check bit2 SWGPE_STS is 1.
image
image

xunfengd:
> > For the test2, we have validated pass,

> 
> Thank you. How do we tell in this test that an interrupt actually occurred between `EnableInterrupts()` and `RestoreTPL()`?

Hi Michael,
For the test two,  we have tried to trigger the software GPE by setting register SWGPE_CTRL bit17 to 1. And read register GPE0_STS_127_96, then check  bit2 SWGPE_STS is 1 which means that the SCI is triggered successfully.   Do you think this experimental data meets the requirements?
https://github.com/tianocore/edk2/pull/11276#issuecomment-3519619743

panyongg:
@mcb30 , @ardbiesheuvel , as you see in the comments, all the concerns, requestings and tests are already done. from our side, there is no risk to merge these changes. **could you please help to check and approve it? thanks. if it's still pending something, please let us know. thanks.** It takes very long and impacts our program now. Thanks for help.
And another problom is that, the owner of this PR (branch) has not worked on this task. And the PR Branch is conflicted now. So we are going to close this PR and create a new one to merge the code changes. is it ok? any concerns?

panyongg:
@mcb30 , @ardbiesheuvel @kraxel @sarah-walker-arm @niruiyu @mdkinney As checked, we also need to update code for previous comments. But the problom is that, the PR owner has left and not worked on this PR. So seems we have to submit new PR for all the changes, agree? or any concern?

mdkinney:
Please open a new PR and mention this PR in the new PR, so the 2 PRs are linked.

ardbiesheuvel:
> Please open a new PR and mention this PR in the new PR, so the 2 PRs are linked.

... and please address the review comments in the old PR.

mdkinney:
> > Please open a new PR and mention this PR in the new PR, so the 2 PRs are linked.
> 
> ... and please address the review comments in the old PR.

I agree.  And the new PR should start with the exact same version of the code in this PR and then add in the unresolved conversations in the new PR and then add the code changes that address each of the opens.

MdeModulePkg/Core/Dxe/Event/Tpl.c
panyongg: @ardbiesheuvel , any futher comments to resolve this concern? Thanks.

ardbiesheuvel: I am only trying to make sure that the assumptions are sound. There is existing code that may execute in DXE context that uses the `SetInterruptState()` APIs. If such code manipulates the TPL as well, is there a risk that we may misidentify `running with interrupts disabled` as `running within an interrupt handler`? And if so, is that a problem?

panyongg:
> > > Please open a new PR and mention this PR in the new PR, so the 2 PRs are linked.
> > 
> > 
> > ... and please address the review comments in the old PR.
> 
> I agree. And the new PR should start with the exact same version of the code in this PR and then add in the unresolved conversations in the new PR and then add the code changes that address each of the opens.

Great! Thanks for the confirmation. So we will submit new PR and linked to this one. And then add all the unresolved conversations in new PR. Thanks.

panyongg:
> > > For the test2, we have validated pass,
> 
> > Thank you. How do we tell in this test that an interrupt actually occurred between `EnableInterrupts()` and `RestoreTPL()`?
> 
> Hi Michael, For the test two, we have tried to trigger the software GPE by setting register SWGPE_CTRL bit17 to 1. And read register GPE0_STS_127_96, then check bit2 SWGPE_STS is 1 which means that the SCI is triggered successfully. Do you think this experimental data meets the requirements? [#11276 (comment)](https://github.com/tianocore/edk2/pull/11276#issuecomment-3519619743)

@mcb30 , Thanks for your help on the tests. could you please help to check the results for test2? do we still need other test coverage? thanks a lot!

PR #11277, opened 2025-07-10
MdeModulePkg/Bus/Usb/UsbBusDxe: USB Device Enumeration Compatibility for Non-Compliant Devices

MarlboroChuang:
This patch enhances the USB enumeration process in EDK2 to improve compatibility with non-standards-compliant devices that may fail during standard enumeration sequences. The solution is based on USB specifications and references implementations from both Linux and Windows environments.

[Suggested Solution]
Enumeration Script Optimization:
Integrated a retry mechanism to sequentially execute enumeration scripts, inspired by the enumeration flows of Windows, Linux, and EDK2. This improves robustness when handling edge-case devices.

BOS Descriptor Check for SuperSpeed Devices:
Some SuperSpeed-capable devices may fall back to High-Speed mode and cause subsequent commands to fail. We now check the BOS descriptor to verify SuperSpeed support and trigger a port reset if needed to re-enumerate the device properly.

Manufacturer String Descriptor Caching:
Certain devices require immediate follow-up commands after reading the LANGID string to fetch Manufacturer, Product, or SerialNumber strings. These strings are now cached after initial retrieval to allow UsbIoGetStringDescriptor() to return them directly, improving efficiency and stability.

TPL Handling in UsbBusRecursivelyConnectWantedUsbIo: Refactored to adopt the TPL handling mechanism from UsbConnectDriver, enabling ConnectController() to operate correctly at TPL_CALLBACK level for safer driver connection.

### DIFF
--- a/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbBus.h
+++ b/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbBus.h
@@ -123,7 +123,7 @@ typedef struct _USB_HUB_API    USB_HUB_API;
 //
 // Send clear feature request timeout, set by experience
 //
-#define USB_CLEAR_FEATURE_REQUEST_TIMEOUT  10
+#define USB_CLEAR_FEATURE_REQUEST_TIMEOUT  10000    // EDK uses 10.  Change for AMD
 
 //
 // Bus raises TPL to TPL_NOTIFY to serialize all its operations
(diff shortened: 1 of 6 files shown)


ardbiesheuvel:
Given that you are making 4 separate changes, please split this up into [at least] 4 patches.

MarlboroChuang:
Thanks for your suggestion. I will drop this PR and then prepare new seperated PRs.

ardbiesheuvel:
Please don't open new PRs when respinning - this loses the review history. Instead, use git force-push to replace the patches with new versions.

MarlboroChuang:
OK, I see your point. Thanks a lot.

PR #11279, opened 2025-07-10
EmbeddedPkg: clean up Prepilib FfsProcessSection ()

leiflindholm:
# Description

#11274 was raised to get rid of an incorrect error message from some unnamed compiler that got confused. But it was quite right to be confused, the underlying code was quite convoluted.

That bugged me enough that I decided to rewrite the function into something readable by both humans and compilers.

## How This Was Tested

Tested on ArmVirtQemuKernel.


leiflindholm:
Spotted a mistake, pushed an update.

EmbeddedPkg/Library/PrePiLib/FwVol.c
```diff
@@ -260,6 +260,91 @@ FindFileEx (
...
+STATIC
+UINT32
+GetSectionNSize (
+  IN EFI_COMMON_SECTION_HEADER  *Section
+  )
+{
+  UINT32  SectionSize;
+
+  if (IS_SECTION2 (Section)) {
+    ASSERT (SECTION2_SIZE (Section) > 0x00FFFFFF);
```
apop5, line 279: Instead of relying on just asserts (which can be disabled) can this return a section size of 0?

apop5 (approved):

EmbeddedPkg/Library/PrePiLib/FwVol.c
```diff
@@ -260,6 +260,91 @@ FindFileEx (
...
+
+  @retval EFI_SECTION_TYPE
+ **/
+STATIC
+EFI_SECTION_TYPE
+GetSectionNCompressionType (
+  IN EFI_COMMON_SECTION_HEADER  *Section
+  )
+{
+  if (IS_SECTION2 (Section)) {
```
ardbiesheuvel, line 345: Maybe add an ASSERT() ~~here~~ above this line that the section type equals EFI_SECTION_COMPRESSION?

EmbeddedPkg/Library/PrePiLib/FwVol.c
leiflindholm: I mean, this is refactoring existing behaviour.

If we want to change behaviour, I think that should be a separate pr.

To me the assert makes sense, because part of what made the original code so convoluted is that SECTION2 is an extension of SECTION by an added 32-bit Size field if the 24-bit one is insufficient. So a size that would fit in 24 bits effectively means the header is corrupted.

EmbeddedPkg/Library/PrePiLib/FwVol.c
leiflindholm: And the same for GetCompressionSectionNHeaderSize above?

Yeah, fair point, and a halfway decent compiler should be able to fold any duplicated tests out.

EmbeddedPkg/Library/PrePiLib/FwVol.c
apop5: Looking at this further, it runs so early that this would be before any recovery of the system would be possible. 
The asserts should be fine.

leiflindholm:
Also folding in this change since we're still confusing clang 19.1.7 without it.

```
@@ -418,7 +418,9 @@ FfsProcessSection (
                    &DstBufferSize,
                    &ScratchBufferSize
                    );
-      } else if (Section->Type == EFI_SECTION_GUID_DEFINED) {
+      } else {  // Section->Type == EFI_SECTION_GUID_DEFINED)

```

Arguably, the contents of the
`    } else if ((Section->Type == EFI_SECTION_COMPRESSION) || (Section->Type == EFI_SECTION_GUID_DEFINED)) {`
should be broken out into a few more helper functions, but I've kind of run out of steam.

EmbeddedPkg/Library/PrePiLib/FwVol.c
leiflindholm: Yeah, generally I'm very much in favour of useful return values over asserts. Just unclear what to do about it here.

PR #11286, opened 2025-07-13
MdePkg/MdeLibs: provide CompilerIntrinsicsLib on all architectures

xypron:
# Description

Compilers like GCC insert calls to library functions like memcpy() and memset() to implement initialization or copying of structures. Freestanding binaries need to be compiled with a library providing these.

CompilerIntrinsicsLib has already been used for this purpose on the Arm and Aarch64 architectures. EDK2-Test fails to build against edk2-stable202505 at least on LoongArch64 and Riscv64 without this library. Make it available for all architectures.

Remove conflicting and obsoleted CryptoPkg/Library/IntrinsicLib/CopyMem.c.

- [ ] Breaking change?
no
- [ ] Impacts security?
no
- [ ] Includes tests?
no
## How This Was Tested
Build test
## Integration Instructions
n/a


os-d:
@mdkinney this has been a popular topic of late :)

@xypron this PR goes against current edk2 policy, which is not to rely on compiler intrinsics except when it cannot be avoided, namely on ARM/AARCH64 and from external code (e.g. OpenSSL). See this for one of the many discussions on it: https://github.com/tianocore/edk2/pull/6169.

Instead, the functions from BaseLib should be used to work around the compiler inserting intrinsics, e.g. RShiftU64, etc.

There have been other asks to make CompilerIntrinsicsLib available to other architectures, but because that goes against current edk2 policy, I believe it should be an RFC first. There are issues doing that, such as compiler intrinsics not being a stable interface, etc.

That aside, your PR doesn’t really make a CompilerIntrinsicsLib for all architectures. It just makes it link for all architectures. Most of the intrinsics in there are still ARM/AARCH64 only. If we are going to change this policy, I think we should completely convert this lib to support edk2’s support architectures.

TL;DR: I believe the way to fix your problem is to use the BaseLib intrinsic workarounds wherever they have cropped up, either in edk2-test or edk2. If you want to push forward with using compiler intrinsics globally, I think that needs to be an RFC and the code change much more flushed out. But, I defer to the stewards in this.

PR #11297, opened 2025-07-15
SecurityPkg/Library: Tpm2DeviceSecLibFfa for PeilessSec

LeviYeoReum:
To support measurement and extend PCR in PeilessSec with
TPM device using FF-A over CRB, add Tpm2DeviceSecLibFfa for PeilessSec.
Since it couldn't use the DynamicPcd in SEC phase, refactor
some routines uses DynamicPcd:
  - get tpm service partition id.
  - get tpm interface type.


ardbiesheuvel:
Please split this up.

LeviYeoReum:
> Please split this up.

Thanks. I'll split this.

LeviYeoReum:
@kuqin12.
Could you review this patch? 

Thanks!

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2InfoFfa.c
```diff
@@ -0,0 +1,87 @@
...
+
+  if (PcdGet16 (PcdTpmServiceFfaPartitionId) != TPM2_FFA_PARTITION_ID_INVALID) {
+    *PartitionId = PcdGet16 (PcdTpmServiceFfaPartitionId);
+    return EFI_SUCCESS;
+  }
+
+  Status = FfaTpm2GetServicePartitionId (PartitionId);
+  if (!EFI_ERROR (Status)) {
+    PcdSet16S (PcdTpmServiceFfaPartitionId, *PartitionId);
+  }
```
kuqin12, line 84: We probably also want to ensure the return is not the same as the invalid ID, which is why I just used 32bit value in the first place :)

kuqin12 (approved):

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2InfoFfa.c
LeviYeoReum: > which is why I just used 32bit value in the first place :)

Thanks for explanation!

> We probably also want to ensure the return is not the same as the invalid ID
Okay. But I think it would be better to make `FfaTpm2GetServicePartitionId()` return ERORR when the PartitionId as `TPM2_FFA_PARTITION_ID_INVALID` without checking the PartitionId from caller.

Thanks for review!

jyao1 (approved):

PR #11304, opened 2025-07-15
Fix random SEV-ES boot crash

tlendacky:
# Description

The SevSnpKnownInitApicId field within the MP_CPU_EXCHANGE_INFO structure is not guaranteed to be zero when booting an SEV-ES guest (or any guest). While the check in SevSnpGetInitCpuNumber() is conditionally guarded by verifying if the guest is an SEV-SNP guest, the check in SevEsGetApicId() is not similarly guarded.

This lack of protection can cause SevEsGetApicId() to return to the wrong location. If the value of the SevSnpKnownInitApicId field contains the exact random value of 1, the guest will be treated as an SEV-SNP guest rather than an SEV-ES guest and return to the wrong location in the code which will lead to a crash.

Ensure that all SEV related fields in MP_CPU_EXCHANGE_INFO structure are properly initialized, thus removing the need for guarding access to the SevSnpKnownInitApicId field.

Additionally, the FillExchangeInfoDataSevEs() is only called for SEV-SNP guests, but the name (and comment in MpLib.c) implies that it is also called for SEV-ES guests. Rename the function to FillExchangeInfoDataSevSnp() to avoid confusion.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This issue was recreated in two ways:
1. Attempting 500 SEV-ES guest boots. Typically, the guest failed somewhere within the first 100 attempts.
2. Manually setting the SevSnpKnownInitApicId field to the value 1, which recreated the boot failure.

This fix was then tested by performing 500 SEV-ES guest boots without a single failure.

## Integration Instructions

N/A


kraxel (approved):

ardbiesheuvel (approved):

ardbiesheuvel:
@tlendacky remind me to merge this if there is no response from the UefiCpuPkg maintainers

bssrikanth:
Thank you @tlendacky , I have tested the patch on top of edk2 master (commit 8be9a344d353) and it is resolving the issue reported.


Tested-by: Srikanth Aithal <sraithal@amd.com>

PR #11305, opened 2025-07-15
ArmPkg,MdePkg: ArmGixDxe Gicv3/v5 tweaks

leiflindholm:
# Description
FEAT_GCIE, which introduces the GICv5 system register interfaces to the Arm architecture
complicates things slightly since:
- The old FEAT_GIC is decribed as informing as to whether "GIC CPU interface system registers" are implemented or not.
- FEAT_GCIE is mutually exclusive with FEAT_GIC.

Existing code used `ArmHasGicSystemRegisters ()` as an indicator of whether the GICv3 driver (also compatible with GICv4) was supported. But now we have two incompatible sets of GIC System Registers.

This PR does a few things to try to decrease the architectural confusion:
- It aligns GICv3 and GICv5 support probing in ArmGicDxe.c
- It renames the ArmHasGicSystemRegisters () function ArmHasGicV3SystemRegisters ()
- It wraps the GicV3DxeInitialize () invokation under a common GicDxeInitialize () shared with
  GicV5DxeInitialize () instead of chaining the latter behind the former.

Yes, this is all cosmetic, but it makes the situation a little bit less confusing for me.

## How This Was Tested

Non-functional. Compile tested armvirtqemu ARM/AARCH64.

## Integration Instructions

Any out-of-tree users will need to change any invocations of `ArmHasGicSystemRegisters ()` to `ArmHasGicV3SystemRegisters ()`.


leiflindholm:
@sarah-walker-arm

sarah-walker-arm:
Does GicV3Dxe now not need to support ARM in addition to AARCH64?

leiflindholm:
> Does GicV3Dxe now not need to support ARM in addition to AARCH64?

This PR does not introduce any functional changes. Unless I've made a mistake that's slipped through my testing, in which case could you point it out for me?

sarah-walker-arm:
> > Does GicV3Dxe now not need to support ARM in addition to AARCH64?
> 
> This PR does not introduce any functional changes. Unless I've made a mistake that's slipped through my testing, in which case could you point it out for me?

You've removed GicV5/Arm/ArmGicV5.c, which removes the dummy GicV5DxeInitialize() for 32-bit ARM. Hence this build error : https://dev.azure.com/tianocore/edk2-ci/_build/results?buildId=192588&view=logs&j=38f904c1-7e1c-5c40-1518-68f022fd0dfb&t=a72a1ef1-a2bd-5040-4c95-6449d6f38f47&l=2475

ardbiesheuvel:
> > > Does GicV3Dxe now not need to support ARM in addition to AARCH64?
> > 
> > 
> > This PR does not introduce any functional changes. Unless I've made a mistake that's slipped through my testing, in which case could you point it out for me?
> 
> You've removed GicV5/Arm/ArmGicV5.c, which removes the dummy GicV5DxeInitialize() for 32-bit ARM. Hence this build error : https://dev.azure.com/tianocore/edk2-ci/_build/results?buildId=192588&view=logs&j=38f904c1-7e1c-5c40-1518-68f022fd0dfb&t=a72a1ef1-a2bd-5040-4c95-6449d6f38f47&l=2475

Yeah, so in order to build the v2/v3 combined driver for 32-bit ARM, it needs a dummy implementation of GicV5DxeInitialize() that always returns EFI_UNSUPPORTED. This confines the v2 and ARM legacy to the combined driver, leaving a single, legacy free GicV3+ driver for AArch64 going forward.

leiflindholm:
Right, I failed to run a NOOPT build. Gah.

Which found another issue in ArmPkg when I tried it (new 1/4, unrelated to GIC). @kuqin12, mind having a look at 1/5?

So, since GICv5Initialize on ARM is a workaround in a single source file, I've folded that in as an ifdef'd dummy function in that file, immediately before the function that calls it.

ArmPkg/Drivers/ArmGicDxe/GicV3/ArmGicV3Dxe.c
```diff
@@ -739,6 +739,22 @@ GicV3ExitBootServicesEvent (
...
+EFI_STATUS
+GicV5DxeInitialize (
+  VOID
+  )
+{
+  return EFI_UNSUPPORTED;
+}
+
+#endif
+
```
ardbiesheuvel, line 757: I intentionally kept this confined to the combined v2/v3 driver so that the GicV3+ driver is legacy free.

In spite of the #ifdef, this just creates technical debt by adding ARM specific code to a source file that was generic before.

leiflindholm: Whereas the other puts GicV5 code, and GicV5 source files, into an architecture that cannot support it.

One way clutters and confuses the source tree.
The other way clutters a corner of a single source file, at the only point it's needed.

The only debt that is carried is that which the driver has intentionally opted into, only more obviously. I'm happy for that chunk to be dropped outright at any point, and there is no plausible situation in which it could be expanded.

ardbiesheuvel: > Whereas the other puts GicV5 code, and GicV5 source files, into an architecture that cannot support it.
> 

Yes, but under an `Arm/` sub-directory that never gets accessed by the build unless building for ARM.

> One way clutters and confuses the source tree. The other way clutters a corner of a single source file, at the only point it's needed.
> 
> The only debt that is carried is that which the driver has intentionally opted into, only more obviously. I'm happy for that chunk to be dropped outright at any point, and there is no plausible situation in which it could be expanded.

I strongly feel that it should not exist to begin with. I'd rather start removing ARM code entirely than adding new `#ifdef MDE_CPU_ARM` checks to AArch64 code.

ardbiesheuvel: ... but the file does not need to be called `Arm/GicV5/GicV5.c` though - we can rename it to something more appropriate.

kuqin12:
> Right, I failed to run a NOOPT build. Gah.
> 
> Which found another issue in ArmPkg when I tried it (new 1/4, unrelated to GIC). @kuqin12, mind having a look at 1/5?

Thanks for the fix. Looks good to me. But I am not sure why it was not caught earlier.

ArmPkg/Drivers/ArmGicDxe/GicV3/ArmGicV3Dxe.c
leiflindholm: > I strongly feel that it should not exist to begin with. I'd rather start removing ARM code entirely than adding new #ifdef MDE_CPU_ARM checks to AArch64 code.

Well, that situation seems to be resolving itself ... so I'll close this.

PR #11307, opened 2025-07-15
SecurityPkg: CodeQL Fixes

Flickdm:
# Description

Makes changes to comply with alerts raised by CodeQL.

Most of the issues here fall into the following two categories:

1. Potential use of uninitialized pointer.
2. Inconsistent integer width in comparison.

- [ ] Breaking change?
- [X] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Ships on a real platforms both client and server
## Integration Instructions

N/A


jyao1 (approved):

mdkinney:
This PR introduced a regression in PEI FV verification.  VerifyHashedFv () can handle a HashInfo of NULL and returns EFI_SUCCESS.  This PR change the behavior to return EFI_NOT_FOUND instead when HashInfo is NULL.

Flickdm:
> This PR introduced a regression in PEI FV verification. VerifyHashedFv () can handle a HashInfo of NULL and returns EFI_SUCCESS. This PR change the behavior to return EFI_NOT_FOUND instead when HashInfo is NULL.

Good catch, Making the revert here!
https://github.com/tianocore/edk2/pull/11403

mdkinney:
Instead of a revert, I have opened PR that should preserve the CodeQL fix and preserve the required behavior.

https://github.com/tianocore/edk2/pull/11402

PR #11310, opened 2025-07-16
Dev/ray/fsp preserve gdtr

niruiyu:
# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions


AshrafAliS (commented):
How about IA32 support. as IntelFsp2Pkg/Library/BaseFspCommonLib/FspCommonLib.c is common for IA32 and X64

AshrafAliS (approved):
changes only on CONTEXT_STACK_64 so X64 support only.

IntelFsp2Pkg/Library/BaseFspSwitchStackLib/X64/Stack.nasm
```diff
@@ -106,6 +125,25 @@ SkipPagetableSave:
     lidt    [rsp]
     add     rsp, 16
 
+    ; Restore GDTR
+    lgdt    [rsp]
+    add     rsp, 16
+
+    ; Restore Segment registers
+    pop     rdx
+    ; mov     cs, dx    ; CS cannot be directly loaded, it requires a far return or far jump
```
ChaselChiu: should we skip cs when pushing to stack?

tedkuo1: If so, do we still need to preserve space for cs?

IntelFsp2Pkg/FspSecCore/X64/Fsp24ApiEntryM.nasm
```diff
@@ -173,6 +173,24 @@ NotMultiPhaseMemoryInitApi:
   push   rdx
 SkipPagetableSave:
 
+  ; Save Segment registers
```
lzeng14, line 176: Does IntelFsp2Pkg/FspSecCore/X64/FspApiEntryM.nasm also need to be updated ? Or it will be deprecated ?

IntelFsp2Pkg/Library/BaseFspSwitchStackLib/X64/Stack.nasm
niruiyu: Good comments @ChaselChiu  and @tedkuo1. Will prepare patch to restore CS also.
The saved CS selector should correspond to the correct descriptor in the restored GDT
The CPU only validates CS when executing far jumps/calls, not during normal instruction execution
The short window between GDT restore and CS restore is safe as long as no far jumps occur

IntelFsp2Pkg/FspSecCore/X64/Fsp24ApiEntryM.nasm
niruiyu: @lzeng14 , good comments. As my patch updated the C structure but did not update FspApiEntryM.nasm, using FspSecCoreM.inf would cause issues.
I will update the PR to include FspApiEntryM.nasm also.

IntelFsp2Pkg/Library/BaseFspSwitchStackLib/X64/Stack.nasm
niruiyu: I've updated the code to restore CS also.

IntelFsp2Pkg/Library/BaseFspSwitchStackLib/X64/Stack.nasm
```diff
@@ -92,6 +92,25 @@ ASM_PFX(FspSwitchStack):
...
+    push    rdx
+    mov     rdx, gs
+    push    rdx
+    mov     rdx, fs
+    push    rdx
+    mov     rdx, es
+    push    rdx
+    mov     rdx, ds
+    push    rdx
+    mov     rdx, cs
```
ChaselChiu, line 106: still not clear about pushing/poping cs inside FSP is necessary, I think when FSP code is executing the cs is FSP cs, not bootloader cs, right? the bootloader cs will be the "API return address" before switching control to FSP API entry point and that should be already handled by call/ret instructions.

niruiyu: CS is not changing when bootloader calls FSP API entry point.
CS may change when FSP implementation does a far jump to change CS.

ChaselChiu: when FSP switching control back to bootloader, the ret instruction will return to bootloader cs, so I think cs saving/restoring is already handled.

what will happen if we do not push/pop FSP cs at all in this PR?

ChaselChiu: OK. I think I got your point, if the platforms have overridden GDT inside FSP to have different cs, then they will require this cs save/restore.
Original switch stack does not handle GDT overridden scenario.
sorry for confusing.

ChaselChiu (approved):

lzeng14 (approved):

PR #11312, opened 2025-07-16
EmbeddedPkg: Fix incorrect define for hardware interrupt2 protocol

cfernald:
# Description

In HardwareInterrupt2.h the define for the protocol does not match the definition of the protocol in the dec file. The definition in the dec file is 32898322-2da1-474a-baaa-f3f7cf569470 where in the definition the define is 32898322-2d1a-474a-baaa-f3f7cf569470 and the string is even less correct with various bit swaps.

This commit changes all the strings and definitions to be consistent with the dec definition.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on QEMU. No code found that uses the define, does not affect the open source implementation.

## Integration Instructions

N/A


EmbeddedPkg/Include/Protocol/HardwareInterrupt2.h
```diff
@@ -11,10 +11,10 @@
 
 #include <Protocol/HardwareInterrupt.h>
 
-// 22838932-1a2d-4a47-aaba-f3f7cf569470
+// 32898322-2da1-474a-baaa-f3f7cf569470
```
ardbiesheuvel, line 14: Can we just drop this comment?

cfernald: It seems to be more or less the standard to have the human readable formatted guid in a comment above the define. This also helps when users search the string in the codebase.

That said, I don't have a strong preference and am happy to remove it if that's preferred.

ardbiesheuvel: Fair enough - let's leave it in.

ardbiesheuvel (approved):

ardbiesheuvel:
Please fix the commit log `more correct` -> `less correct`.

cfernald:
> Please fix the commit log `more correct` -> `less correct`.

Done

PR #11315, opened 2025-07-17
MdeModulePkg: Change PCD type to support dynamic

pranavvv1:
Move PcdSerialRegisterBase from [PcdsFixedAtBuild, PcdsPatchableInModule] section to [PcdsFixedAtBuild, PcdsPatchableInModule, PcdsDynamic, PcdsDynamicEx] section to enable dynamic configuration.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested with internal platform and it is working

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/MdeModulePkg.dec
+++ b/MdeModulePkg/MdeModulePkg.dec
@@ -1352,10 +1352,6 @@
   # @Prompt Enable serial port cable detetion.
   gEfiMdeModulePkgTokenSpaceGuid.PcdSerialDetectCable|FALSE|BOOLEAN|0x00020006
 
-  ## Base address of 16550 serial port registers in MMIO or I/O space. Default is 0x3F8.
-  # @Prompt Base address of serial port registers.
-  gEfiMdeModulePkgTokenSpaceGuid.PcdSerialRegisterBase|0x03F8|UINT64|0x00020002
-
   ## Baud rate for the 16550 serial port.  Default is 115200 baud.
   # @Prompt Baud rate for serial port.
   # @ValidList  0x80000001 | 921600, 460800, 230400, 115200, 57600, 38400, 19200, 9600, 7200, 4800, 3600, 2400, 2000, 1800, 1200, 600, 300, 150, 134, 110, 75, 50
@@ -1790,6 +1786,11 @@
   gEfiMdeModulePkgTokenSpaceGuid.PcdAcpiExposedTableVersions|0x20|UINT32|0x0001004c
 
 [PcdsFixedAtBuild, PcdsPatchableInModule, PcdsDynamic, PcdsDynamicEx]
+
+  ## Base address of 16550 serial port registers in MMIO or I/O space. Default is 0x3F8.
+  # @Prompt Base address of serial port registers.
+  gEfiMdeModulePkgTokenSpaceGuid.PcdSerialRegisterBase|0x03F8|UINT64|0x00020002
+
   ## UART clock frequency is for the baud rate configuration.
   # @Prompt Serial Port Clock Rate.
   gEfiMdeModulePkgTokenSpaceGuid.PcdSerialClockRate|1843200|UINT32|0x00010066


apop5:
I don't think that this will work. 

the base address PCD was PcdsFixedAtBuild a lot of the consumers of the PCD are of type BASE, meaning they will work in SEC,PEI,DXE,STMM, SMM etc because they do not have external dependencies on other protocols/libraries.

changing this pcd type to dynamic introduces a dependency on having the PCD database available, where as fixed at build resulted in the value being a #define, with out a dependency on PCD services being available. .

This PR breaks the existing BASE libraries that were consuming it.

ChaselChiu:
> I don't think that this will work.
> 
> the base address PCD was PcdsFixedAtBuild a lot of the consumers of the PCD are of type BASE, meaning they will work in SEC,PEI,DXE,STMM, SMM etc because they do not have external dependencies on other protocols/libraries.
> 
> changing this pcd type to dynamic introduces a dependency on having the PCD database available, where as fixed at build resulted in the value being a #define, with out a dependency on PCD services being available. .
> 
> This PR breaks the existing BASE libraries that were consuming it.

This PR just extends the PCD type capabilities, it does not "define" the PCD type, so nothing broken.
Platform DSC is the decision maker for PCD type and still FixedAtBuild can be defined by platforms if they want.
Some platforms only execute  BASE SerialPortLib after PcdDatabase available in PEI and DXE phases which should be ok to define PCD as Dynamic.
Do you see any reason that we have to limit platforms to only use FixedAtBuild PCD type?

ChaselChiu:
@lgao4 , how about you? could you also share your thoughts about this PR?

lgao4:
> @lgao4 , how about you? could you also share your thoughts about this PR?

This change doesn't break anything. But, if the platform sets this PCD as Dyanmic, the platform may not work. I think the comment  asks this question that PcdSerialRegisterBase can't be confiured as DynamicPcd.

ChaselChiu:
> > @lgao4 , how about you? could you also share your thoughts about this PR?
> 
> This change doesn't break anything. But, if the platform sets this PCD as Dyanmic, the platform may not work. I think the comment asks this question that PcdSerialRegisterBase can't be confiured as DynamicPcd.

SerialPortLib is not only used by sending BIOS debug message, it also can be used by communicating with peripheral devices. And consuming those BASE lib in which BIOS phase is platform decision and if my platform only consume such BASE lib after PcdDatabase ready (for example, adding PcdPpi or PcdProtocol in the driver depex), what will be the problem consuming Dynamic PCD?

lgao4:
> > > @lgao4 , how about you? could you also share your thoughts about this PR?
> > 
> > 
> > This change doesn't break anything. But, if the platform sets this PCD as Dyanmic, the platform may not work. I think the comment asks this question that PcdSerialRegisterBase can't be confiured as DynamicPcd.
> 
> SerialPortLib is not only used by sending BIOS debug message, it also can be used by communicating with peripheral devices. And consuming those BASE lib in which BIOS phase is platform decision and if my platform only consume such BASE lib after PcdDatabase ready (for example, adding PcdPpi or PcdProtocol in the driver depex), what will be the problem consuming Dynamic PCD?

Right. If your platform consumes SerialPortLib when PCD service is ready, this PCD will allow to be set as Dynamic.

ChaselChiu:
> > > > @lgao4 , how about you? could you also share your thoughts about this PR?
> > > 
> > > 
> > > This change doesn't break anything. But, if the platform sets this PCD as Dyanmic, the platform may not work. I think the comment asks this question that PcdSerialRegisterBase can't be confiured as DynamicPcd.
> > 
> > 
> > SerialPortLib is not only used by sending BIOS debug message, it also can be used by communicating with peripheral devices. And consuming those BASE lib in which BIOS phase is platform decision and if my platform only consume such BASE lib after PcdDatabase ready (for example, adding PcdPpi or PcdProtocol in the driver depex), what will be the problem consuming Dynamic PCD?
> 
> Right. If your platform consumes SerialPortLib when PCD service is ready, this PCD will allow to be set as Dynamic.

Thanks Liming! If we are on the same page, could you approve this PR for merging?

lgao4:
> [apop5](/apop5)

@apop5 , does it clarify your comments?

apop5:
> > [apop5](/apop5)
> 
> @apop5 , does it clarify your comments?

Yes, its a good explanation of the desired use case, and I can understand how this would be used.

It still complicates the use case for the debugging (i.e. making sure the platform overrides the pcd as a fixed at build).

ChaselChiu:
> > > [apop5](/apop5)
> > 
> > 
> > @apop5 , does it clarify your comments?
> 
> Yes, its a good explanation of the desired use case, and I can understand how this would be used.
> 
> It still complicates the use case for the debugging (i.e. making sure the platform overrides the pcd as a fixed at build).



> > > [apop5](/apop5)
> > 
> > 
> > @apop5 , does it clarify your comments?
> 
> Yes, its a good explanation of the desired use case, and I can understand how this would be used.
> 
> It still complicates the use case for the debugging (i.e. making sure the platform overrides the pcd as a fixed at build).

I do not think this PR will cause any impact to existing platform use cases - you should not need to make any change to your platform when this PR merged:
Case 1: If platform already defined PcdSerialRegisterBase type as FixedAtBuild (Or PatchableInModule) in its own DSC file, then PCD will be FixAtBuild (or PatchableInModule), same as before this PR merging.
Case 2: If platform did not define  PcdSerialRegisterBase type in its own DSC file, then BaseTools by default will define it as FixedAtBuild.

This PR just "allows" platform DSC to define this PCD type as Dynamic or DynamicEx (previously only FixedAtBuild or PatchAbleInModule can be defined). This PR does not "enforce" PCD type as Dynamic.

Please let me know if you observing any platform code change needed due to this PR.

ChaselChiu:
@apop5 , @lgao4 , are we good with this change now? could we get approval to merge it?

lgao4:
> @apop5 , @lgao4 , are we good with this change now? could we get approval to merge it?

I agree it has no impact for the existing platform.

ChaselChiu:
> > @apop5 , @lgao4 , are we good with this change now? could we get approval to merge it?
> 
> I agree it has no impact for the existing platform.

Great! could you approve this PR for merging?

ChaselChiu:
Thanks, @apop5 , @lgao4

PR #11316, opened 2025-07-17
MdeModulePkg/StandaloneMmPkg: Always Print Driver Load Messages

os-d:
# Description

Today, DXE/PEI/StandaloneMM/SMM Core's image loaders only print driver load messages if debug code is enabled. However, these are some of the most important prints in the codebase: on a given system even if you have nothing else to debug with, you can see the last driver executed.

Debug code blocks are used to skip logic that only exists for debug purposes and wastes time on a release build. However, the logic to print a line and determine the filename from the PDB is not extensive and provides critical information, so it is inappropriate to wrap in a debug code section.

Platforms can still choose to disable logging at DEBUG_INFO/DEBUG_LOAD and will not see the error messages.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Shipping in platforms, simple print change.

## Integration Instructions

N/A.


ardbiesheuvel (approved):
The overhead of doing the lookup unconditionally would seem negligible to me, given that the total number of drivers loaded is never more than a few hundred at the most, even on an unusual configuration.

niruiyu (approved):
I agree.

MdeModulePkg/Core/Pei/Image/Image.c
```diff
@@ -623,6 +623,10 @@ PeiLoadImageLoadImage (
   UINT16                Machine;
   EFI_SECTION_TYPE      SearchType1;
   EFI_SECTION_TYPE      SearchType2;
+  CHAR8                 *AsciiString;
+  CHAR8                 EfiFileName[512];
```
samimujawar, line 627: The EfiFileName buffer size seems to be used inconsistently. Is it possible to address this, please? I think we can make the buffer size 512 bytes.

lgao4: Right. One additional commit can be made for it.

os-d: Np, I'll make another commit.

os-d:
@samimujawar @lgao4 I have pushed two more commits to address your feedback (one for MdeModulePkg, one for StandaloneMmPkg). Please re-review.

lgao4 (approved):

samimujawar (approved):

PR #11317, opened 2025-07-18
Smm cpu wait all aps fix

jiaxinwu:
update AllApArrivedWithException if the 1st timeout sync is skipped but all APs have arrived

This PR also refines the debug log in the SmmWaitForApArrival.


niruiyu (commented):
it seems the commit contains lots of changes that add extra spaces in code.
Why?

jiaxinwu:
> it seems the commit contains lots of changes that add extra spaces in code. Why?

i don't know, but i will check & fix it.

niruiyu (approved):

PR #11320, opened 2025-07-18
SecurityPkg: Add Tpm2DumpLib

os-d:
# Description

This PR adds Tpm2DumpLib to SecurityPkg. It is used to share code between different Tpm2DeviceLibs that need to dump out common TPM transactions for debugging purposes.

This is not a revamp of Tpm2DebugLib, which was rejected. Tpm2DebugLib was attempting to control debug prints by switching out libraries classes. This is not the purpose of Tpm2DumpLib. This new library is only attempting to share common code between different modules. Control of the debug prints is left to the platform on whether they set DEBUG_SECURITY level or not.

- [x] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

CI passes. This simply reuses the logic in another lib.

## Integration Instructions

This is marked as a breaking change because a new library class is added. Platforms must add Tpm2DumpLib to their DSCs if they depend on Tpm2DeviceLibDTpm or Tpm2DeviceLibFfa.


kuqin12 (approved):

jyao1:
@os-d, thanks for the follow up.

I see two things in this patch set:
1) add a new Tpm2DumpLib.
2) change the debug level DEBUG_VERBOSE to DEBUG_SECURITY.

They are independent feature to resolve two independent problem.
As such, I recommend using two different PRs.

First, I have no concern on 2). So please submit a dedicate PR for that.

However, my concern for 1) is that every platforms must be updated because we introduce a mandatory lib class - Tpm2DumpLib. I don't understand the benefit so far.
If possible, please submit a patch without any other platform DSC change.

osteffenrh (approved):
From the SVSM vTPM side: All good.
Tested it with the current Coconut SVSM main branch.

os-d:
> @os-d, thanks for the follow up.
> 
> I see two things in this patch set:
> 
> 1. add a new Tpm2DumpLib.
> 2. change the debug level DEBUG_VERBOSE to DEBUG_SECURITY.
> 
> They are independent feature to resolve two independent problem. As such, I recommend using two different PRs.
> 
> First, I have no concern on 2). So please submit a dedicate PR for that.
> 
> However, my concern for 1) is that every platforms must be updated because we introduce a mandatory lib class - Tpm2DumpLib. I don't understand the benefit so far. If possible, please submit a patch without any other platform DSC change.

@jyao1 , sure np I have moved the `DEBUG_VERBOSE` to `DEBUG_SECURITY` change to a separate PR.

The reason I added Tpm2DumpLib is to share code between different components (Tpm2DeviceLibDTpm, Tpm2DeviceLibFfa, and platform code), which is a standard use case for creating library classes in edk2.

The alternatives would be:
- Carry multiple versions of the code. This is not ideal because it inevitably means one will go stale and it is extra unneeded files. This is the current state of the code.
- Share the C file that implements this logic. This avoids a new library class, but means that the INFs need to have logic to include relative paths, which is a bit odd, but definitely used in many other components in edk2. This approach doesn't allow platforms to use the logic, which is unfortunate as it means the platform needs to duplicate the logic. Also, the current tree layout would make this weird, Tpm2DeviceLibFfa.inf would have to include `../Tpm2DeviceLibDTpmDump/Tpm2DeviceLibDTpmDump.c`, which is not a usual practice to include a file from another library class, which leads to the next alternative below.
- The SecurityPkg code tree could be changed such that you have a structure below which would allow for a more natural inclusion of `../Tpm2DeviceLibDump.c`. However, this approach still doesn't let platforms use the shared logic and is still a breaking change because every DSC would need to update the path to the Tpm2DeviceLib they are using.
  - SecurityPkg/Library/Tpm2DeviceLib/
    - Tpm2DeviceLibDTpm/
    - Tpm2DeviceLibFfa/
    - Tpm2DeviceLibDump.c
 
So my preference would be to do the solution I have in the PR, which does require DSCs to update but in a very trivial way: `Tpm2DumpLib|SecurityPkg/Library/Tpm2DumpLib/Tpm2DumpLib.inf`. This is the kind of change that is expected whenever new edk2 stable tags are pulled, so this is really not a problem. This allows platform code to leverage the same code instead of having to copy it.

Please let me know your thoughts.

os-d:
@jyao1 what are your thoughts to approach? I'd like to get this in before the hard freeze, since this was opened before the soft freeze started and consumers are asking for the dumping functionality in the FFA lib.

mikebeaton:
@jyao1 - This keeps being marked stale, but I believe looks sane and @os-d's argument for the approach taken likewise?

os-d:
Thanks @mikebeaton , I've been leaving this untouched because I think this would fit easily in with https://github.com/tianocore/edk2/pull/11634 instead of becoming a new library class on its own, but I've been waiting for that PR. I'll work with @Raymond-MS to get that in.

PR #11321, opened 2025-07-18
BaseTools: PatchCheck.py: Allow MultiPkg Commits For Subject Check

os-d:
# Description

Subject Check was added to PatchCheck.py to enforce that if a package was touched in a commit that it be referenced in the subject line.

However, this is impractical for multipackage commits with many packages, e.g. when stack cookies were added, every package was touched, but in a rote way, and it is not reasonable to put every package in a subject line.

This updates PatchCheck.py to check if ignore_multi_package is set and if so only require that package names be included in the subject if there are fewer than 3 packages touched. Otherwise, PatchCheck will require the message to start with `Global:` to indicate it touches more than 3 packages.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Running PatchCheck.py on the following commits:
- Single package commit with package name in subject -> pass
- 3 packages changed, not in subject -> fail
- 3 packages changed, in subject with multipkg ignore set -> pass
- 4 packages changed, 3 of them in subject with multipkg ignore set -> fail
- 4 packages changed, Global in commit message with multipkg ignore set -> pass

## Integration Instructions

N/A.


os-d:
@leiflindholm @bexcran I ran into this issue working on a commit that touched a lot of packages. I changed it so that PatchCheck will only enforce one package name if you are ignoring multi package, but this could be handled other ways (we already have the number of packages, so I guess we could just check if that is greater than one). This also could only enforce the package name is in the subject when the package count is less than 3 or something. I don't have a big preference; I just don't want to block the sometimes necessary large number of package commits.

os-d:
@bexcran @lgao4 @leiflindholm can you please review?

lgao4:
> @bexcran @lgao4 @leiflindholm can you please review?

What's your commit message? Can you give one example with multiple commits?

leiflindholm:
Whoops, sorry, this was lost in the chaos of my inbox after I came back off holiday.

My priority is for the shortlog (subject) to be useful, and the check was added because we were seeing an increase of frankly sloppy submissions. And letting patches through with a shortlog indicating they modify only a single (or worse, a few random ones out of a larger set) package would be actively misleading.

But I agree once you start making _truly_ global changes (as opposed to touching 2-3) it becomes less useful to enumerate the specific packages. (See also #11241.)

Not a properly thought out proposal at all, but could we agree on a definition of a "Global" change which could be used for multi-package commits? I.e. permit a subject of `Global: ....` if that tag is set?

BaseTools/Scripts/PatchCheck.py
```diff
@@ -117,13 +117,14 @@ def __init__(self, subject, message, author_email, updated_packages):
 
         self.check_contributed_under()
         if not MergifyMerge:
+            # Get the CI options first so that future checks can use the configuration
```
leiflindholm: It may seem silly, but I would prefer this broken out as a separate change.
Then the reason for the move can be described in the commit message instead of adding residual comment verbosity to the codebase.

os-d:
> Whoops, sorry, this was lost in the chaos of my inbox after I came back off holiday.
> 
> My priority is for the shortlog (subject) to be useful, and the check was added because we were seeing an increase of frankly sloppy submissions. And letting patches through with a shortlog indicating they modify only a single (or worse, a few random ones out of a larger set) package would be actively misleading.
> 
> But I agree once you start making _truly_ global changes (as opposed to touching 2-3) it becomes less useful to enumerate the specific packages. (See also #11241.)
> 
> Not a properly thought out proposal at all, but could we agree on a definition of a "Global" change which could be used for multi-package commits? I.e. permit a subject of `Global: ....` if that tag is set?

@leiflindholm I think the Global tag is reasonable. I agree that when you are updating 2 - 3 packages it is useful to have the names. I think 3 packages is the longest we could support or else there will not be enough room for the actual subject line (keeping under 75 characters). I can amend this to allow `Global: ...` if the package count is over 3 and multi package is enabled (or just drop the plumbing for checking multipackage here since the multipackage check will fail on its own anyway).

leiflindholm:
> @leiflindholm I think the Global tag is reasonable. I agree that when you are updating 2 - 3 packages it is useful to have the names. I think 3 packages is the longest we could support or else there will not be enough room for the actual subject line (keeping under 75 characters). I can amend this to allow Global: ... if the package count is over 3 and multi package is enabled 

I'm happy with that.
(Note that after #11241 the script does let you always get in 20 chars after the `:`.)

> (or just drop the plumbing for checking multipackage here since the multipackage check will fail on its own anyway).

You mean skipping multi-package check if subject line starts with `Global:`?

My only concern then would be if anyone uses the tag for any other purposes - effectively, we might then want to _ensure_ Global is only used for packages with the ignore-multi-package tag.

I don't feel strongly about it. Kind of happy either way.

os-d:
> > (or just drop the plumbing for checking multipackage here since the multipackage check will fail on its own anyway).
> 
> You mean skipping multi-package check if subject line starts with `Global:`?
> 
> My only concern then would be if anyone uses the tag for any other purposes - effectively, we might then want to _ensure_ Global is only used for packages with the ignore-multi-package tag.
> 
> I don't feel strongly about it. Kind of happy either way.

I also don't feel strongly about it. I'll keep it as is (where we check that ignore-multi-package is set) but change it to require Global if over 3 packages.

BaseTools/Scripts/PatchCheck.py
makubacki: Moving the change to a separate commit seems reasonable to isolate it as a pre-requisite to the other more substantial change. But I think that this particular comment is reasonable in the code since it calls out an ordering assumption made in these calls and could prevent a future regression not otherwise discovered via git blame alone.

On the other hand, this is partly compensating for poor naming in `check_ci_options_format()` which actually does more than "set", it has the important side effect of assigning `self.ignore_multi_package = True` which makes the non-intuitive and necessitating the comment in the first place. If it was named something like `gather_ci_configuration_for_checks()`, that would make it clearer that configuration should be gathered before it is used in subsequent checks.

os-d: Because this is PatchCheck, I don't care too much. I changed this to be two commits and dropped the comment. I am going to resolve this comment for now, but can make other adjustments if folks want.

os-d:
@leiflindholm I have updated the script, please take a look when you have a chance. I updated the PR description with the test cases I tried.

@lgao4 as mentioned in the description, one example commit that PatchCheck would now fail is this one: https://github.com/tianocore/edk2/commit/ef4f3aa3f7e3c28c7f0e1a3c35711f1a85becd71. PatchCheck would now require that all packages names be in the subject line, which obviously does not add value. With this change, PatchCheck would now enforce that the subject message be: `Global: <description>`.

lgao4:
> @leiflindholm I have updated the script, please take a look when you have a chance. I updated the PR description with the test cases I tried.
> 
> @lgao4 as mentioned in the description, one example commit that PatchCheck would now fail is this one: [ef4f3aa](https://github.com/tianocore/edk2/commit/ef4f3aa3f7e3c28c7f0e1a3c35711f1a85becd71). PatchCheck would now require that all packages names be in the subject line, which obviously does not add value. With this change, PatchCheck would now enforce that the subject message be: `Global: <description>`.

Yes. This change makes sence. You also need to update wiki https://github.com/tianocore/tianocore.github.io/wiki/Commit-Message-Format for this new rule.

os-d:
> Yes. This change makes sence. You also need to update wiki https://github.com/tianocore/tianocore.github.io/wiki/Commit-Message-Format for this new rule.

@lgao4 I have edited the wiki here: https://github.com/tianocore/tianocore.github.io/wiki/Commit-Message-Format. Can you confirm if this is appropriate? Feel free to edit in place.

lgao4:
> > Yes. This change makes sence. You also need to update wiki https://github.com/tianocore/tianocore.github.io/wiki/Commit-Message-Format for this new rule.
> 
> @lgao4 I have edited the wiki here: https://github.com/tianocore/tianocore.github.io/wiki/Commit-Message-Format. Can you confirm if this is appropriate? Feel free to edit in place.

The update in wiki is good.

os-d:
@bexcran @lgao4 now that the freeze has ended, can we merge this?

PR #11328, opened 2025-07-21
ShellPkg: Shell Validate parameter before use.

apop5:
# Description

When FvSimpleFileSystem is included in a firmware image, the FV is accessible as a simple file system.

Shell contained a bad assumption that the FileDevicepath, the path where the efi shell was loaded from, was always a Media device path/media vendor device path. It would make a blind cast of the device path node.

Add a check to verify device path type/subtype before casting the node to a FILEPATH_DEVICE_PATH.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
A system was hanging when entering shell. Tracked to this bad cast.
After adding checks, verified system no longer hung.

Verified that startup.nsh FFS file was still located and executed.

## Integration Instructions
No integration necessary.


apop5:
> Please fix the patch subject - summarize what the patch does, rather than why

@ardbiesheuvel 

Its been updated. Thanks.

ShellPkg/Application/Shell/Shell.c
```diff
@@ -1262,9 +1262,11 @@ LocateStartupScript (
 
     InternalEfiShellSetEnv (L"homefilesystem", StartupScriptPath, TRUE);
 
-    StartupScriptPath = StrnCatGrow (&StartupScriptPath, &Size, ((FILEPATH_DEVICE_PATH *)FileDevicePath)->PathName, 0);
-    PathRemoveLastItem (StartupScriptPath);
-    StartupScriptPath = StrnCatGrow (&StartupScriptPath, &Size, mStartupScript, 0);
+    if ((DevicePathType (FileDevicePath) == MEDIA_DEVICE_PATH) && (DevicePathSubType (FileDevicePath) == MEDIA_VENDOR_DP)) {
```
ardbiesheuvel: Shouldn't this be `MEDIA_FILEPATH_DP`?

apop5: yes, you are right, I copied the wrong value.

I reran the tests with the updated code and they pass.

ardbiesheuvel (approved):

PR #11331, opened 2025-07-22
CryptoPkg: Add API to set TLS security level.

kanagavels97:
Add an API to set the user preferred TLS security level.

# Description

According to the [OpenSSL 3.5 documentation](https://docs.openssl.org/3.5/man3/SSL_CTX_set_security_level/), setting the TLS security level to 0 permits all protocols and cipher suites, effectively disabling security restrictions. we have set the TLS security level to 2 by default. Additionally, we have provided an API that allows users to customize the security level based on their requirements.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## Integration Instructions

<N/A>

### DIFF
--- a/CryptoPkg/Library/TlsLib/TlsInit.c
+++ b/CryptoPkg/Library/TlsLib/TlsInit.c
@@ -188,7 +188,7 @@ TlsNew (
   //
   // This retains compatibility with previous version of OpenSSL.
   //
-  SSL_set_security_level (TlsConn->Ssl, 0);
+  SSL_set_security_level (TlsConn->Ssl, 3);
 
   //
   // Initialize the created SSL Object
--- a/CryptoPkg/Private/Protocol/Crypto.h
+++ b/CryptoPkg/Private/Protocol/Crypto.h
@@ -21,7 +21,7 @@
 /// the EDK II Crypto Protocol is extended, this version define must be
 /// increased.
 ///
-#define EDKII_CRYPTO_VERSION  17
+#define EDKII_CRYPTO_VERSION  19
 
 ///
 /// EDK II Crypto Protocol forward declaration
@@ -3952,6 +3952,27 @@ EFI_STATUS
   );
 
 /**
+  Set the Tls security level.
+
+  This function Set the Tls security level.
+  If Tls is NULL, nothing is done.
+
+  @param[in]  Tls                Pointer to the TLS object.
+  @param[in]  Level              Pointer to the Tls Security level need to set.
+  @param[in]  DataSize           Size of level, it should be sizeof (UINT8)
+
+  @retval  EFI_SUCCESS           The Tls security level was set successfully.
+  @retval  EFI_INVALID_PARAMETER The parameters are invalid.
+
+**/
+typedef
+EFI_STATUS
+(EFIAPI *EDKII_CRYPTO_TLS_SET_SECURITY_LEVEL)(
+  IN VOID    *Tls,
+  IN UINT8   Level
+  );
+
+/**
   Gets the protocol version used by the specified TLS connection.
 
   This function returns the protocol version used by the specified TLS
@@ -5710,6 +5731,9 @@ struct _EDKII_CRYPTO_PROTOCOL {
   EDKII_CRYPTO_PKCS1V2_DECRYPT                        Pkcs1v2Decrypt;
   EDKII_CRYPTO_RSA_OAEP_ENCRYPT                       RsaOaepEncrypt;
   EDKII_CRYPTO_RSA_OAEP_DECRYPT                       RsaOaepDecrypt;
+  /// TLS Set(continued)
+  EDKII_CRYPTO_TLS_SET_SERVER_NAME                    TlsSetServerName;
+  EDKII_CRYPTO_TLS_SET_SECURITY_LEVEL                 TlsSetSecurityLevel;
 };
 
 extern GUID  gEdkiiCryptoProtocolGuid;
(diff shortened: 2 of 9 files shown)


CryptoPkg/Private/Protocol/Crypto.h
```diff
@@ -5710,6 +5731,8 @@ struct _EDKII_CRYPTO_PROTOCOL {
   EDKII_CRYPTO_PKCS1V2_DECRYPT                        Pkcs1v2Decrypt;
   EDKII_CRYPTO_RSA_OAEP_ENCRYPT                       RsaOaepEncrypt;
   EDKII_CRYPTO_RSA_OAEP_DECRYPT                       RsaOaepDecrypt;
+  /// TLS (continued)
```
liyi77: need to update '''#define EDKII_CRYPTO_VERSION''' after protocol change

CryptoPkg/Library/TlsLib/TlsInit.c
```diff
@@ -188,7 +188,7 @@ TlsNew (
   //
   // This retains compatibility with previous version of OpenSSL.
   //
-  SSL_set_security_level (TlsConn->Ssl, 0);
```
liyi77, line 191: please dont change default level for compatibility

kanagavels97: While a minimum security level of 2 is recommended, the API provides flexibility for users to configure the security level according to their specific requirements.

liyi77: Not sure if other users are still using the old algorithm, which may break their tests.

@jyao1 Could you please comment on whether we should update the default secure level?

Level 0
Everything is permitted. This retains compatibility with previous versions of OpenSSL.
Level 1
The security level corresponds to a minimum of 80 bits of security. Any parameters offering below 80 bits of security are excluded. As a result RSA, DSA and DH keys shorter than 1024 bits and ECC keys shorter than 160 bits are prohibited. Any cipher suite using MD5 for the MAC is also prohibited. Any cipher suites using CCM with a 64 bit authentication tag are prohibited. Note that signatures using SHA1 and MD5 are also forbidden at this level as they have less than 80 security bits. Additionally, SSLv3, TLS 1.0, TLS 1.1 and DTLS 1.0 are all disabled at this level.
Level 2
Security level set to 112 bits of security. As a result RSA, DSA and DH keys shorter than 2048 bits and ECC keys shorter than 224 bits are prohibited. In addition to the level 1 exclusions any cipher suite using RC4 is also prohibited. Compression is disabled.

kanagavels97: @jyao1

Could you please provide your suggestion on the previous comment?

kanagavels97: @liyi77 @jyao1 Any update on this?

CryptoPkg/Private/Protocol/Crypto.h
kanagavels97: Updated

CryptoPkg/Library/TlsLib/TlsInit.c
kanagavels97: @liyi77 @jyao1 Please provide your suggestion on this.

liyi77: We need more feedback from community, I will approve it if no objections edk2 devel group.

jyao1:
I think it is a good idea to set to Level 2.

The question is if we want to expose an API.

Is there any real use case to use Level 0 or level 1?

If no, can we just to set Level 2 directly?

liyi77:
Hi @kanagavels97 , I'm thinking setting default to 3 would be a better choice for now.
https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-57pt1r5.pdf
image


 What about your idea?

kanagavels97:
> Hi @kanagavels97 , I'm thinking setting default to 3 would be a better choice for now. https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-57pt1r5.pdf image
> 
> What about your idea?

I'm ok with setting the security level to 3 by deafault. I implemented an API to change the security level as per your suggestion and let me know whether the API is still needed or we can just skip the API and will proceed with setting secuirty level to 3.

> Hi @kanagavels97 , I'm thinking setting default to 3 would be a better choice for now. https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-57pt1r5.pdf image
> 
> What about your idea?

@liyi77 

I’m okay with setting the security level to 3 by default. I’ve also implemented the API to allow changing the security level as per your earlier suggestion.

Could you please confirm whether we should:
Proceed with setting the default security level to 3 and skip the API, or
Keep the API in place for flexibility?

liyi77:
I would like to keep the API in place for flexibility and set default level to 3.
A worthwhile discussion is whether we should set a minimum security level, such as only allowing users to choose from levels 2 - 5. Or allow complete flexibility.

@jyao1 Could you share your idea for default security level, thanks!

jyao1:
Yes, I think setting to 3 is better idea. I also agree to only allow level 2-5 in the new API.
Let's forbid level 0/1, if no push back from community.

liyi77:
Hi @kanagavels97 , as discussed above, let's set the default value to 3 and set the lower limit of TlsSetSecurityLevel to 2.

I have sent email to EDK2 devel group about this PR and will approve it in a week if no further concerns be raised.

kanagavels97:
Created new PR for this [https://github.com/tianocore/edk2/pull/11534](url).

PR #11341, opened 2025-07-24
MdeModulePkg/PciBusDxe: Keep track of rejected state explicitly

ardbiesheuvel:
The full PCI enumerator may reject a PCI device to deal with a resource conflict, and does so by programming the device BARs to all 0x1 bits. Later on, it will detect this rejected state by reprobing the BARs, and checking whether all modifiable bits have been set to 0x1.

This subtly relies on the fact that the resource allocation logic rarely (or never) assigns resources all the way up to the end of the available window, and so this all 0x1 state will never occur in practice.

However, the situation is different for the light enumerator, which inherits the PCI resource allocations from a preceding firmware stage, which may decide to allocate top-down rather than bottom-up.

In this case, we may encounter an existing BAR assignment that is valid, but indistinguishable from the all 0x1 assignment that signifies rejection. E.g., a 16-bit I/O BAR being allocated top-down from a 16-bit I/O resource window that ends at 0xffff will end up in this situation, and the whole device will be rejected as a result.

Given that reprobing the BARs (which involves a read/write/read/write MMIO sequence for each BAR of each device) in order to detect this rejected state is a dubious practice to begin with, let's keep track of this state explicitly, through a BOOLEAN flag in the PCI I/O device struct.

Note that simply keeping a 'rejected' flag may regress platforms that rely on the light enumerator to disable devices that have all BARs set to 0x1 when entering. But we can keep a device wide flag that indicates that the device has at least one programmed BAR, and manage this flag when parsing and assigning BARs.


MrChromebox:
at first glance this seems like a much cleaner way to handle the issue of rejected devices - I'll give it a test tomorrow

MrChromebox:
no issues with PCI device rejection, but somehow breaks the PS2 keyboard in edk2. Nothing jumps out in the serial log, but the keyboard is non-functional/can't enter setup.

ardbiesheuvel:
> no issues with PCI device rejection, but somehow breaks the PS2 keyboard in edk2. Nothing jumps out in the serial log, but the keyboard is non-functional/can't enter setup.

Weird. And this doesn't happen with the other fix? Although unlikely, I suppose this might be a port IO conflict with another PCI device that was rejected before - any chance you could find out?

MrChromebox:
> Weird. And this doesn't happen with the other fix? Although unlikely, I suppose this might be a port IO conflict with another PCI device that was rejected before - any chance you could find out?

correct, no issue with #11339. I'll have to add some debug logging to see what's different between the two. Will  try to get to this before I go on travel mid next week

MrChromebox:
spent some time on this over the weekend and couldn't find anything to explain the keyboard failure, but it's 100% reproducible.

ardbiesheuvel:
> spent some time on this over the weekend and couldn't find anything to explain the keyboard failure, but it's 100% reproducible.

Thanks.

I'll try to dig a bit deeper but it won't be before September

ardbiesheuvel:
@MrChromebox is there any way to run this in QEMU?

MrChromebox:
@ardbiesheuvel I don't see why not, I'm able to boot coreboot + edk2 using the qemu q35 target (-M q35)

PR #11342, opened 2025-07-24
pip-requirements: Add pylibfdt and swig

cap2k4-rivos:
# Description

pip-requirements: Add pylibfdt and PE tool and swig

1. Tool supportUPL requires libfdt support
2. Also added swig which is required by UefiPayloadPkg.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
N /A

## Integration Instructions
N/A

### DIFF
--- a/pip-requirements.txt
+++ b/pip-requirements.txt
@@ -17,3 +17,5 @@ edk2-pytool-extensions~=0.29.11
 antlr4-python3-runtime==4.9
 lcov-cobertura==2.1.1
 regex==2024.11.6
+pylibfdt==1.7.2
+pefile


pip-requirements.txt
```diff
@@ -17,3 +17,5 @@ edk2-pytool-extensions~=0.29.11
 antlr4-python3-runtime==4.9
 lcov-cobertura==2.1.1
 regex==2024.11.6
+swig==4.3.1
```
ChaselChiu: swig should not be needed, could you double check?

pip-requirements.txt
```diff
@@ -17,3 +17,5 @@ edk2-pytool-extensions~=0.29.11
 antlr4-python3-runtime==4.9
 lcov-cobertura==2.1.1
 regex==2024.11.6
+swig==4.3.1
+pylibfdt==1.7.2
```
ChaselChiu, line 20: by the way, pefile will be needed when building FIT UPL. (for adding relocation data to FIT header)

ChaselChiu (approved):

pip-requirements.txt
dhaval-rivos: @ChaselChiu https://github.com/dgibson/dtc/pull/154 Isn't pylibfdt using swig? I think without this is not compiling.

cap2k4-rivos:
@ChaselChiu @gdong1, I have added swig to the requirements as it is required by pyfdtlib. kindly review the changes again. Thanks

gdong1 (approved):

ChaselChiu:
> @ChaselChiu @gdong1, I have added swig to the requirements as it is required by pyfdtlib. kindly review the changes again. Thanks

I thought we have removed the swig dependency by below PR: https://github.com/devicetree-org/pylibfdt/pull/3
Could you share what error you encountered when "pip install pylibfdt" that must install swig?

cap2k4-rivos:
Hi @ChaselChiu, it seems I mistook the swig dependency for pylibfdt. It's only required when building from source, not when using wheels from PyPi (pip). I've removed it again. Apologies for the confusion.

ChaselChiu (approved):

PR #11343, opened 2025-07-24
DynamicTablesPkg: add Tpm2DeviceTableLib to generate Tpm2 device table

LeviYeoReum:
Introduce Tpm2DeviceTableLib to generate SSDT table which describes
Tpm2 devices.
For this, add field TpmDeviceSize in CmArchCommonTpm2InterfaceInfo
to describe TPM2 device area in the SSDT table.
    
This dynamic table generation is controlled by PcdGenTpm2DeviceTable
which default value is FALSE.
When it is TRUE, the TPM2 device ssdt table is generated when TPM2 ACPI
table is generated.
    

DynamicTablesPkg/Library/Common/Tpm2DeviceTableLib/Tpm2DeviceTableLib.c
```diff
@@ -0,0 +1,329 @@
+/** @file
+  Tpm2 device table generating Library
+
+  Copyright (c) 2025, Arm Limited. All rights reserved.<BR>
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+  @par Reference(s):
+  - TCG ACPI specification.
+    (https://developer.arm.com/documentation/ihi0069/)
```
pierregondois: Wrong spec I think

DynamicTablesPkg/Library/Common/Tpm2DeviceTableLib/Tpm2DeviceTableLib.c
```diff
@@ -0,0 +1,329 @@
...
+/** C array containing the compiled AML template.
+    This symbol is defined in the auto generated C file
+    containing the AML bytecode array.
+*/
+extern CHAR8  tpm2devicetabletemplate_aml_code[];
+
+/** Fixup the Tpm2 device UID (_UID).
+
+  @param  [in]  RootNodeHandle  Pointer to the root of an AML tree.
+  @param  [in]  Uid             UID for the Serial Port.
```
pierregondois: Should search for `serial` word in the file, I think some comments should be updated

DynamicTablesPkg/Library/Common/Tpm2DeviceTableLib/Tpm2DeviceTableLib.c
```diff
@@ -0,0 +1,329 @@
...
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "ERROR: TPM2-DEVICE-FIXUP: Failed to Serialize SSDT Table Data."
+      " Status = %r\n",
+      Status
+      ));
+  }
+
+ErrorHandler:
```
pierregondois: NIT: as this path is also used when normally exiting the function, maybe the name should be `error` -> `exit`

DynamicTablesPkg/Include/ArchCommonNameSpaceObjects.h
```diff
@@ -722,6 +722,11 @@ typedef struct CmArchCommonTpm2InterfaceInfo {
 
   /** Log Area Start Address */
   UINT64    Lasa;
+
+  /** Tpm Device Size.
+   *  This information is used for Tpm2 Device Table using DynamicTableGenerator.
```
pierregondois: NIT: Just the following should be enough:
```
/** Tpm Device Size. */
```

DynamicTablesPkg/Include/ArchCommonNameSpaceObjects.h
```diff
@@ -722,6 +722,11 @@ typedef struct CmArchCommonTpm2InterfaceInfo {
 
   /** Log Area Start Address */
   UINT64    Lasa;
+
+  /** Tpm Device Size.
+   *  This information is used for Tpm2 Device Table using DynamicTableGenerator.
+   */
+  UINT64    TpmDeviceSize;
```
pierregondois: I think it might be better to create a separate CmObj for TPM devices. This might be arguable from a platform PoV, but if there are multiple TPM devices, their size might differ (and their address also surely).

A TPM device object with:
- a base address (populated by `FixedPcdGet64 (PcdTpmBaseAddress)` in the configuration manager)
- an address length

should fit right ?

LeviYeoReum: That's sound resonable.
In PoV, having a TPM devices more than one seems strange,
But, Having separate object to describe TPM device seems reasonable and it would be more scalable to describe other feature in feature.

I'll make a new object for this.

Thanks!

DynamicTablesPkg/Include/ArchCommonNameSpaceObjects.h
LeviYeoReum: Okay. I'll fix it!

DynamicTablesPkg/Library/Common/Tpm2DeviceTableLib/Tpm2DeviceTableLib.c
LeviYeoReum: Sorry for my bad. I'll fix it :)

DynamicTablesPkg/Library/Common/Tpm2DeviceTableLib/Tpm2DeviceTableLib.c
LeviYeoReum: Right.. I'll fix

DynamicTablesPkg/Library/Common/Tpm2DeviceTableLib/Tpm2DeviceTableLib.c
LeviYeoReum: Okay. I'll change it.

PR #11344, opened 2025-07-24
ArmPlatformPkg: add measurement on PeilessSec

LeviYeoReum:
This patches add support of measurement on PeilessSec.
Since PeilessSec doesn't have PEI phase, it couldn't do measurement
done by Tcg2Pei.
For doing those kind of measurement done by Tcg2Pei in PeilessSec,
add PeilessSecMeasureLib in PeilessSec.

Signed-off-by: Yeoreum Yun <yeoreum.yun@arm.com>

PR #11345, opened 2025-07-24
ArmPlatformPkg/PlatformPei: generates TPM event log hobs

LeviYeoReum:
When measured boot is enabled, TF-A passes
the early measurement event log to the BL33 firmware (i.e., UEFI)
through a transfer list entry with
the tag ID TRANSFER_ENTRY_TAG_ID_TPM_EVENT_LOG.

Therefore, generate TPM event log HOBs from
the early measurement event log so they can be
unified with the Tcg2Dxe event logs.

### DIFF
--- a/ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
+++ b/ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
@@ -9,8 +9,25 @@
 #include <PiPei.h>
 
 #include <Library/ArmPlatformLib.h>
+#include <Library/ArmTransferListLib.h>
+#include <Library/DebugLib.h>
 #include <Library/HobLib.h>
 #include <Library/PcdLib.h>
+
+/**
+  Build event log hob consumed by Tcg2Dxe.
+
+  @param[in]  TransferList  Transfer list header
+
+  @return EFI_SUCCESS       Success to generate HOBs from passed event log.
+  @return Others            Failed to create event log hob
+
+**/
+EFI_STATUS
+EFIAPI
+BuildPeiEventLogHobs (
+  IN TRANSFER_LIST_HEADER  *TransferList
+  );
 
 EFI_STATUS
 EFIAPI
@@ -18,7 +35,30 @@ PlatformPeim (
   VOID
   )
 {
+  EFI_STATUS            Status;
+  TRANSFER_LIST_HEADER  *TransferList;
+
   BuildFvHob (PcdGet64 (PcdFvBaseAddress), PcdGet32 (PcdFvSize));
+
+  Status = TransferListGetFromHobList (&TransferList);
+  if (EFI_ERROR (Status)) {
+    // Transfer list is not found, this means
+    // early event log is not available and
+    // measured boot may not be enabled.
+    return EFI_SUCCESS;
+  }
+
+  if (TransferListCheckHeader (TransferList) == TRANSFER_LIST_OPS_INVALID) {
+    DEBUG ((DEBUG_ERROR, "%a: Invalid transfer list... Stop boot...\n", __func__));
+    return EFI_INVALID_PARAMETER;
+  }
+
+  // Create HOBs for the early measurement Event logs.
+  Status = BuildPeiEventLogHobs (TransferList);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "%a: Failed to build event log Pei Hobs...: %r\n", __func__, Status));
+    return Status;
+  }
 
   return EFI_SUCCESS;
 }
--- a/ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
+++ b/ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
@@ -1,0 +1,611 @@
+/** @file
+
+  Copyright (c) 2025, ARM Limited. All rights reserved.
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include <PiPei.h>
+
+#include <Library/BaseLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/DebugLib.h>
+#include <Library/ArmTransferListLib.h>
+#include <Library/HobLib.h>
+
+#include <IndustryStandard/Tpm20.h>
+#include <IndustryStandard/UefiTcgPlatform.h>
+
+typedef struct {
+  TPMI_ALG_HASH    HashAlgo;
+  UINT32           HashMask;
+  UINT16           HashSize;
+  CHAR8            *HashName;
+} INTERNAL_HASH_INFO;
+
+STATIC INTERNAL_HASH_INFO  mHashInfo[] = {
+  { TPM_ALG_SHA1,   HASH_ALG_SHA1,   SHA1_DIGEST_SIZE,   "SHA1"   },
+  { TPM_ALG_SHA256, HASH_ALG_SHA256, SHA256_DIGEST_SIZE, "SHA256" },
+  { TPM_ALG_SHA384, HASH_ALG_SHA384, SHA384_DIGEST_SIZE, "SHA384" },
+  { TPM_ALG_SHA512, HASH_ALG_SHA512, SHA512_DIGEST_SIZE, "SHA512" },
+};
+
+typedef struct {
+  UINTN    EventLogBase;
+  UINTN    EventLogSize;
+  VOID     *Event;
+  UINTN    EventSize;
+} EVENT_LOG_ITERATOR;
+
+/**
+  Get hash size based on Algo
+
+  @param[in]     HashAlgo           Hash Algorithm Id.
+  @param[out]    HashInfo           Hash Algorithm Information.
+
+  @return Size of the hash.
+
+**/
+STATIC
+EFI_STATUS
+EFIAPI
+GetHashInfo (
+  IN TPMI_ALG_HASH        HashAlgo,
+  OUT INTERNAL_HASH_INFO  **HashInfo
+  )
+{
+  UINTN  Idx;
+
+  for (Idx = 0; Idx < ARRAY_SIZE (mHashInfo); Idx++) {
+    if (mHashInfo[Idx].HashAlgo == HashAlgo) {
+      *HashInfo = &mHashInfo[Idx];
+      return EFI_SUCCESS;
+    }
+  }
+
+  return EFI_UNSUPPORTED;
+}
+
+/**
+  Check Event is SpecId Event.
+
+  @param[in]   Event                  Event
+
+  @return      TRUE             Event is SpecId event.
+  @return      FALSE            Event isn't SpecId event.
+
+ **/
+STATIC
+BOOLEAN
+EFIAPI
+IsSpecIdEvent (
+  IN VOID  *Event
+  )
+{
+  TCG_PCR_EVENT             *TcgPcrEvent;
+  TCG_EfiSpecIDEventStruct  *TcgEfiSpecIdEventStruct;
+
+  TcgPcrEvent             = (TCG_PCR_EVENT *)Event;
+  TcgEfiSpecIdEventStruct = (TCG_EfiSpecIDEventStruct *)
+                            (Event + OFFSET_OF (TCG_PCR_EVENT, Event));
+
+  if (!((TcgPcrEvent->EventType == EV_NO_ACTION) &&
+        ((CompareMem (
+            TcgEfiSpecIdEventStruct->signature,
+            TCG_EfiSpecIDEventStruct_SIGNATURE_02,
+            sizeof (TcgEfiSpecIdEventStruct->signature)
+            ) == 0) ||
+         (CompareMem (
+            TcgEfiSpecIdEventStruct->signature,
+            TCG_EfiSpecIDEventStruct_SIGNATURE_03,
+            sizeof (TcgEfiSpecIdEventStruct->signature)
+            ) == 0))))
+  {
+    return FALSE;
+  }
+
+  return TRUE;
+}
+
+/**
+  Get size of SpecId Event.
+
+  @param[in]   SpecIdEvent                  SpecId event
+
+  @return      Size of SpecId event.
+
+ **/
+STATIC
+UINTN
+EFIAPI
+GetSpecIdEventSize (
+  IN  VOID  *SpecIdEvent
+  )
+{
+  TCG_PCR_EVENT  *TcgPcrEvent;
+
+  TcgPcrEvent = (TCG_PCR_EVENT *)SpecIdEvent;
+
+  return OFFSET_OF (TCG_PCR_EVENT, Event) + TcgPcrEvent->EventSize;
+}
+
+/**
+  Get size of digest list in eventlog.
+
+  @param[in]   TcgPcrEvent2        eventlog.
+
+  @return      Size of digest list.
+
+ **/
+STATIC
+UINTN
+EFIAPI
+GetDigestListSize (
+  IN  TCG_PCR_EVENT2  *TcgPcrEvent2
+  )
+{
+  EFI_STATUS          Status;
+  UINTN               Idx;
+  VOID                *Digest;
+  UINT32              DigestCount;
+  TPMI_ALG_HASH       HashAlgo;
+  INTERNAL_HASH_INFO  *HashInfo;
+
+  /*
+   * The Digest list is a packed structure composed of a list of
+   * { {Hash Algorithm, Digest} x DigestCount },
+   * wherein the Digest length varies based on the Hash algorithm.
+   * Therefore, parse to the end of the digest list.
+   *
+   * See the TCG EFI Protocol Specification's 5.2 Crypto Agile Log Entry Format.
+   */
+  DigestCount = TcgPcrEvent2->Digest.count;
+  HashAlgo    = TcgPcrEvent2->Digest.digests[0].hashAlg;
+  Digest      = (VOID *)&TcgPcrEvent2->Digest.digests[0].digest;
+
+  for (Idx = 0; Idx < DigestCount; Idx++) {
+    Status = GetHashInfo (HashAlgo, &HashInfo);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "%a: Can't get HashInfo for %d\n", __func__, HashAlgo));
+      return 0;
+    }
+
+    Digest += HashInfo->HashSize;
+    CopyMem (&HashAlgo, Digest, sizeof (TPMI_ALG_HASH));
+    Digest += sizeof (TPMI_ALG_HASH);
+  }
+
+  Digest -= sizeof (TPMI_ALG_HASH);
+
+  return (UINTN)Digest - (UINTN)&TcgPcrEvent2->Digest.digests[0];
+}
+
+/**
+  Initialize event log iterator
+
+  @param[out]    Iter               Iterator.
+  @param[in]     EventLogBase       Event log base address.
+  @param[in]     EventLogSize       Total event log size.
+
+**/
+STATIC
+VOID
+EFIAPI
+EventLogIteratorInit (
+  OUT EVENT_LOG_ITERATOR  *Iter,
+  IN VOID                 *EventLogBase,
+  IN UINTN                EventLogSize
+  )
+{
+  Iter->EventLogBase = (UINTN)EventLogBase;
+  Iter->EventLogSize = EventLogSize;
+  Iter->Event        = NULL;
+  Iter->EventSize    = 0;
+}
+
+/**
+  Get next event form event log iterator
+
+  @param[in,out]    Iter               Iterator
+  @param[out]       NextEvent          Next event
+  @param[out]       NextEventSize      Size of next event
+
+  @return EFI_SUCCESS             Success to get next event.
+  @return EFI_NOT_FOUND           No more event.
+  @return EFI_INVALID_PARAMETER   Invalid event log was found.
+
+**/
+STATIC
+EFI_STATUS
+EFIAPI
+EventLogIteratorGetNext (
+  IN OUT EVENT_LOG_ITERATOR  *Iter,
+  OUT    VOID                **NextEvent,
+  OUT    UINTN               *NextEventSize
+  )
+{
+  EFI_STATUS      Status;
+  TCG_PCR_EVENT2  *TcgPcrEvent2;
+  VOID            *Event;
+  UINT32          EventSize;
+  UINTN           DigestListSize;
+
+  if (Iter->Event == NULL) {
+    Iter->Event = (VOID *)Iter->EventLogBase;
+  } else {
+    Iter->Event += Iter->EventSize;
+  }
+
+  if ((UINTN)Iter->Event >= (Iter->EventLogBase + Iter->EventLogSize)) {
+    Status          = EFI_NOT_FOUND;
+    Iter->Event     = NULL;
+    Iter->EventSize = 0;
+    goto ExitHandler;
+  }
+
+  if (IsSpecIdEvent (Iter->Event)) {
+    Status          = EFI_SUCCESS;
+    Iter->EventSize = GetSpecIdEventSize (Iter->Event);
+    goto ExitHandler;
+  }
+
+  TcgPcrEvent2   = ((TCG_PCR_EVENT2 *)Iter->Event);
+  Event          = (VOID *)&TcgPcrEvent2->Digest.digests[0];
+  DigestListSize = GetDigestListSize (TcgPcrEvent2);
+  if (DigestListSize == 0) {
+    Status          = EFI_INVALID_PARAMETER;
+    Iter->Event     = NULL;
+    Iter->EventSize = 0;
+    goto ExitHandler;
+  }
+
+  Event += DigestListSize;
+  CopyMem (&EventSize, Event, sizeof (TcgPcrEvent2->EventSize));
+  Event += sizeof (TcgPcrEvent2->EventSize);
+  Event += EventSize;
+
+  Iter->EventSize = (UINTN)Event - (UINTN)Iter->Event;
+  Status          = EFI_SUCCESS;
+
+ExitHandler:
+  *NextEvent     = Iter->Event;
+  *NextEventSize = Iter->EventSize;
+  return Status;
+}
+
+/**
+  This function dump TCG_EfiSpecIDEventStruct.
+
+  @param[in]   Event                  Event
+
+**/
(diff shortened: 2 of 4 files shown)


LeviYeoReum:
@ardbiesheuvel, @leiflindholm and @samimujawar.

Could I ask you to see this patch when you're available?

Thanks.

ArmPlatformPkg/ArmPlatformPkg.ci.yaml
```diff
@@ -50,7 +50,8 @@
             "EmbeddedPkg/EmbeddedPkg.dec",
             "MdeModulePkg/MdeModulePkg.dec",
             "MdePkg/MdePkg.dec",
-            "UefiCpuPkg/UefiCpuPkg.dec"
+            "UefiCpuPkg/UefiCpuPkg.dec",
```
samimujawar: Would it be better to reword the commit message as below?

ArmPlatformPkg/PlatformPei: generate TPM event log hobs

When measured boot is enabled, TF-A hands-off the early
measurement event log to the BL33 firmware (i.e. UEFI) via
a transfer list with tag_id
TRANSFER_ENTRY_TAG_ID_TPM_EVENT_LOG.

Therefore, generate TPM event log HOBs from the
early measurement event log, for unifying with the
Tcg2Dxe event logs.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
```diff
@@ -0,0 +1,598 @@
...
+  )
+{
+  TCG_PCR_EVENT             *TcgPcrEvent;
+  TCG_EfiSpecIDEventStruct  *TcgEfiSpecIdEventStruct;
+
+  TcgPcrEvent             = (TCG_PCR_EVENT *)Event;
+  TcgEfiSpecIdEventStruct = (TCG_EfiSpecIDEventStruct *)
+                            (Event + OFFSET_OF (TCG_PCR_EVENT, Event));
+
+  /*
```
samimujawar: I think this comment is incorrect there and needs to be removed.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
```diff
@@ -0,0 +1,598 @@
...
+    /*
+     * If event log is invalid, skip creating EventLogHobs from TransferList.
+     */
+    DEBUG ((DEBUG_WARN, "%a: Invalid TPM event log...", __func__));
+    return EFI_SUCCESS;
+  }
+
+  if (!IsSpecIdEvent (NextEvent)) {
+    /*
+     * First event should be SpecId event otherwise it's invalid event log.
```
samimujawar: The first event must be a SpecId event otherwise the event log is invalid. Therefore, skip creating HOBs for the early measurements.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
```diff
@@ -0,0 +1,598 @@
...
+     * skip creating EventlogHobs from TransferList.
+     */
+    DEBUG ((DEBUG_WARN, "%a: Invalid TPM event log...", __func__));
+    return EFI_SUCCESS;
```
samimujawar: If we do not have a proper early measurement event log, is it meaningful to continue? Should we just return an error instead?

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
```diff
@@ -0,0 +1,598 @@
...
+  }
+
+  /*
```
samimujawar, line 595: Tcg2Dxe generates a SpecId Event when it generates the final event log. Therefore, skip creating a HOB for the TF-A SpecIdEvent log.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
```diff
@@ -0,0 +1,598 @@
...
+  }
+
+  Event -= sizeof (TPMI_ALG_HASH);
+  CopyMem (&EventSize, Event, sizeof (TcgPcrEvent2->EventSize));
+  Event += sizeof (TcgPcrEvent2->EventSize);
+  Event += EventSize;
+
+  Iter->EventSize = (UINTN)Event - (UINTN)Iter->Event;
+
+ErrorHandler:
```
samimujawar: I think this label should be renamed to ExitHandler as it is used in both error and success states to initialise the out parameters.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
```diff
@@ -0,0 +1,598 @@
...
+  @param[out]  EventLogSize           Total size of TPM event log
+
+  @return      EFI_SUCCESS
+  @return      EFI_NOT_FOUND
+
+**/
+STATIC
+EFI_STATUS
+EFIAPI
+GetTpmEventLog (
```
samimujawar: Should this function be renamed to GetEarlyMesurementLog() ?

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
```diff
@@ -0,0 +1,598 @@
...
+  TRANSFER_ENTRY_HEADER  *TransferEntry;
+
+  TransferEntry = TransferListFindFirstEntry (TransferList, TRANSFER_ENTRY_TAG_ID_TPM_EVENT_LOG);
+  if (TransferEntry == NULL) {
+    DEBUG ((DEBUG_ERROR, "%a: No TPM event log entry found...\n", __func__));
+    return EFI_NOT_FOUND;
+  }
+
+  /*
+   * The first 4 bytes are flags.
```
samimujawar: Please see my comments at https://github.com/tianocore/edk2/pull/11344/files/f99a6a07febf356e0980b0eab35913d6ff8968d6#r2249172083
I think similar changes are needed here too.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
```diff
@@ -0,0 +1,598 @@
...
+   * See TPM event log table entry layout (XFERLIST_EVLOG) in
+   * https://github.com/FirmwareHandoff/firmware_handoff/blob/main/source/transfer_list.rst#id25
+   */
+  *EventLog = TransferListGetEntryData (TransferEntry) + sizeof (UINT32);
```
samimujawar: It would be good to have a local variable of type XFERLIST_EVLOG used here.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
```diff
@@ -0,0 +1,598 @@
...
+    Event += sizeof (TPMI_ALG_HASH);
+  }
+
+  Event -= sizeof (TPMI_ALG_HASH);
+  CopyMem (&EventDataSize, Event, sizeof (TcgPcrEvent2->EventSize));
+  Event               += sizeof (TcgPcrEvent2->EventSize);
+  StartupLocalityEvent = (TCG_EfiStartupLocalityEvent *)Event;
+
+  if ((EventDataSize == sizeof (TCG_EfiStartupLocalityEvent)) &&
+      (CompareMem (&StartupLocalityEvent->Signature, TCG_EfiStartupLocalityEvent_SIGNATURE, 16) == 0))
```
samimujawar: It may be good to define a macro for the signature size.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
```diff
@@ -0,0 +1,598 @@
+/** @file
+
+  Copyright (c) 2025, ARM Limited. All rights reserved.
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
```
samimujawar, line 5: Please add a reference to the TransferList spec.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
```diff
@@ -0,0 +1,598 @@
...
+
+  DEBUG ((DEBUG_INFO, "\n"));
+
+  DigestBuffer = DigestBuffer - sizeof (TPMI_ALG_HASH);
+  CopyMem (&EventSize, DigestBuffer, sizeof (TcgPcrEvent2->EventSize));
+  DEBUG ((DEBUG_INFO, "    EventSize - 0x%08x\n", EventSize));
+  EventBuffer = DigestBuffer + sizeof (TcgPcrEvent2->EventSize);
+
+  if ((EventSize == sizeof (TCG_EfiStartupLocalityEvent)) &&
+      (CompareMem (EventBuffer, TCG_EfiStartupLocalityEvent_SIGNATURE, 16) == 0))
```
samimujawar: Use macro instead of magic numbers, please.

ArmPlatformPkg/PlatformPei/PlatformPeim.inf
```diff
@@ -54,6 +57,8 @@
 
 [Guids]
   gArmTransferListHobGuid
+  gTcgEvent2EntryHobGuid
+  gTpm2StartupLocalityHobGuid
```
samimujawar: Looks like gTpm2StartupLocalityHobGuid is nit used here, can you check, please?

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
```diff
@@ -8,17 +8,97 @@
...
-  BuildFvHob (PcdGet64 (PcdFvBaseAddress), PcdGet32 (PcdFvSize));
+  EFI_STATUS            Status;
+  TRANSFER_LIST_HEADER  *TransferList;
+
+  BuildFvHob (
+    PcdGet64 (PcdFvBaseAddress),
+    PcdGet32 (PcdFvSize)
+    );
+
+  Status = GetTransferList (&TransferList);
```
samimujawar: Should this code block be something like:
```

Status = GetTransferList();
if (EFI_ERROR (Status)) {
  if (Status != EFI_NOT_FOUND) {
    // Log an error and return
    return Status;
  }
  // Transfer list is not found, this means
  // early event log is not available and
  // measured boot may not be enabled.
  return EFI_SUCESS;
}

// Transfer list found, check if it is valid.
if (TransferListCheckHeader (TransferList) == TRANSFER_LIST_OPS_INVALID) {
       DEBUG ((DEBUG_ERROR, "%a: Invalid transfer list... Stop boot...\n", __func__));
       return EFI_INVALID_PARAMETER;
     }

// Create HOBs for the early measurement Event logs.
Status = BuildPeiEventLogHobs (TransferList);
   if (EFI_ERROR (Status)) {
     DEBUG ((DEBUG_ERROR, "%a: Failed to build event log Pei Hobs...: %r\n", __func__, Status));
     return Status;
   }

```

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
```diff
@@ -8,17 +8,97 @@
...
+  if (!EFI_ERROR (Status)) {
+    if (TransferListCheckHeader (TransferList) == TRANSFER_LIST_OPS_INVALID) {
+      DEBUG ((DEBUG_ERROR, "%a: Invalid transfer list... Stop boot...\n", __func__));
+      return EFI_INVALID_PARAMETER;
+    }
+  } else {
+    return EFI_SUCCESS;
+  }
+
+  Status = BuildPeiEventLogHobs (TransferList);
```
samimujawar, line 57: Do we need to add a way to specify/detect that measured boot support is enabled? Does this patch change the existing behaviour? What is the impact on existing platforms that did not support measured boot?

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
LeviYeoReum: Note:
   I'll use some helper for transfer list in https://github.com/tianocore/edk2/pull/11381.

And I'll add your suggested comments.

Thanks ;)

ArmPlatformPkg/ArmPlatformPkg.ci.yaml
LeviYeoReum: That would be better :) 

Thanks!

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
LeviYeoReum: > Do we need to add a way to specify/detect that measured boot support is enabled? Does this patch change the existing behaviour? What is the impact on existing platforms that did not support measured boot?

I don't think so since the support for TF-A passes transfer list is supported since April, this year
(https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/36207)
And the edk2's support for receving "TransferList" is quire recent.

Also, If the platform doesn't support a measurement boot, It wouldn't use the there own Tcg2Dxe driver.
It means it doesn't change the existing behaviour but, If some platform passes transfer list, 
It just uses more HOB for event log.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
LeviYeoReum: As I said in other comment, 
I'll use some helper for transfer list in https://github.com/tianocore/edk2/pull/11381.

So, I don't think it would need to add the spec here.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
LeviYeoReum: Right. I'll remove.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
LeviYeoReum: Okay. I'll modify it.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
LeviYeoReum: It would be much better with `sizeof`. I'll change it.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
LeviYeoReum: I'll change it with `sizeof` if some event has different signature size on its event in the future.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
LeviYeoReum: As I said, It'll be replace with https://github.com/tianocore/edk2/pull/11381.
So I'll remove it. Thanks!

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
LeviYeoReum: Since I'm going to use https://github.com/tianocore/edk2/pull/11381,
I'll remove it. Thanks!

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
LeviYeoReum: Since I'm going to use https://github.com/tianocore/edk2/pull/11381,
I'll remove it. Thanks!

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
LeviYeoReum: Thanks. I'll change with your suggetion :)

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
LeviYeoReum: Right. I would be better for Invalid event log. Thanks!

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
LeviYeoReum: Thanks. I'll change it as your suggestion :D

LeviYeoReum:
@samimujawar,
This'll be rebased after https://github.com/tianocore/edk2/pull/11381 is merged.
Base on this, I've updated according to your suggestion.

Thanks!

ArmPlatformPkg/PlatformPei/PlatformPeim.inf
LeviYeoReum: Thanks. I'll remove it.

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
```diff
@@ -8,17 +8,62 @@
...
+  IN TRANSFER_LIST_HEADER  *TransferList
+  );
+
 EFI_STATUS
 EFIAPI
 PlatformPeim (
   VOID
   )
 {
-  BuildFvHob (PcdGet64 (PcdFvBaseAddress), PcdGet32 (PcdFvSize));
```
leiflindholm, line 21: Unrelated change.
If you do want to move it below the local variable definitions, please do it in a separate commit and without adding unnecessary line breaks.

ArmPlatformPkg/PlatformPei/PlatformPeim.inf
```diff
@@ -53,7 +56,7 @@
   gArmTokenSpaceGuid.PcdFvSize
 
 [Guids]
-  gArmTransferListHobGuid
```
leiflindholm, line 50: Why drop this declaration?

LeviYeoReum: Since via the ArmTransferListLib and via header file (Guid/ArmTransferList.h) the gArmTransferListHobGuid is included.

Although the compile has no problem, It would be better to remain as it is. since BuildHob uses this GUID directly.

Thanks.

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
LeviYeoReum: Grrr. This is awful this of uncrustify.sh .... 
I'll restore this one. Sorry!

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
```diff
@@ -0,0 +1,562 @@
...
+
+  EventLogIteratorInit (&Iter, EventLog, EventLogSize);
+
+  Status = EventLogIteratorGetNext (&Iter, &NextEvent, &NextEventSize);
+  if (EFI_ERROR (Status)) {
+    /*
+     * If event log is invalid, skip creating EventLogHobs from TransferList.
+     */
+    DEBUG ((DEBUG_WARN, "%a: Invalid TPM event log...", __func__));
+    return EFI_SUCCESS;
```
samimujawar: Why are we returning success here? If we have reached here, we expect that the event log is valid right? 
If the check at Line 520 succeeds, it indicates that we have an Event Log. Now if the event log contents/parsing fails we should return an error, right?

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
```diff
@@ -0,0 +1,562 @@
...
+   * Therefore, skip creating a HOB for the TF-A SpecIdEvent log.
+   */
+  while (!EFI_ERROR (EventLogIteratorGetNext (&Iter, &NextEvent, &NextEventSize))) {
+    Status = BuildOnePeiEventLogHob (NextEvent, NextEventSize);
+    if (EFI_ERROR (Status)) {
+      return Status;
+    }
+  }
+
+  return EFI_SUCCESS;
```
samimujawar: I think this should be changed to
return Status;

That way if you break out of the while loop at line 554, we return the correct error code.

Also do we need to do cleanup in case of an error?

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
```diff
@@ -0,0 +1,562 @@
...
+      DEBUG ((DEBUG_ERROR, "%a: Can't get HashInfo for %d\n", __func__, HashAlgo));
+      return Status;
+    }
+
+    Event += HashInfo->HashSize;
+    CopyMem (&HashAlgo, Event, sizeof (TPMI_ALG_HASH));
+    Event += sizeof (TPMI_ALG_HASH);
+  }
+
+  Event -= sizeof (TPMI_ALG_HASH);
```
samimujawar, line 504: Can we avoid what is happening at line 460 and line 463, please?
I think implementing GetDigestListSize() should solve this issue as well.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
```diff
@@ -0,0 +1,562 @@
...
+  VOID                         *GuidHob;
+  UINT32                       EventDataSize;
+  TCG_EfiStartupLocalityEvent  *StartupLocalityEvent;
+
+  TcgPcrEvent2 = ((TCG_PCR_EVENT2 *)Event);
+  DigestCount  = TcgPcrEvent2->Digest.count;
+  HashAlgo     = TcgPcrEvent2->Digest.digests[0].hashAlg;
+  Event        = (VOID *)&TcgPcrEvent2->Digest.digests[0].digest;
+
+  for (Idx = 0; Idx < DigestCount; Idx++) {
```
samimujawar, line 492: I think a comment would make things clearer.
/*Although the TCG_PCR_EVENT2.Digest.digests[x].digest type TPMU_HA is defined as a union of all digest sizes, the event log contains a digest list packed as a list of { {Hash Algorithm Type, DigestValue} x DigestCount }, wherein the digest length varies based on the Hash algorithm. Therefore, parse to the end of the digest list. */

In fact, since this same loop also appears at line 211 it would be much better to implement a function GetDigestListSize(TPML_DIGEST_VALUES), with appropriate documentation header.

Can we do this, please?

samimujawar: Also adding a reference to https://trustedcomputinggroup.org/resource/tcg-efi-protocol-specification/ in the file header and referring section 5.2 Crypto Agile Log Entry Format here would be useful.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
```diff
@@ -0,0 +1,562 @@
...
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include <PiPei.h>
+
+#include <Library/BaseLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/DebugLib.h>
+#include <Library/ArmPlatformLib.h>
```
samimujawar: I think line 14 and 17 can be dropped.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
LeviYeoReum: Right. I'll return the error on this.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
LeviYeoReum: Not need to. Since the "EventLogIteratorGetNext()" return when it meets the "end of log".
(though event log has invalid hash algorithm but that's out of specification for event log).

But the error while creating the eventlog hob when it meet the OOM and there's no way to cleanup the hob
(since hob free is not allowed).

So no cleanup but it's the best to return with error log.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
LeviYeoReum: Okay. I'll add this.

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
LeviYeoReum: Okay. I'll add

ArmPlatformPkg/PlatformPei/PlatformPeiTpmEventLog.c
LeviYeoReum: Okay. I'll remove them.

ArmPlatformPkg/PlatformPei/PlatformPeim.inf
leiflindholm: OK, that's a reasonable argument.

But ... if we consider the GUID to be implicitly imported by ArmTransferListLib.h, then I think PlatformPeiLib.c should not also need to directly include `Guid/TransferListHob.h`.

LeviYeoReum: Sorry for my mistaken word. the ArmTransferListLib.h doesn't include "Guid/TransferListHob.h"
In its header. first I mentioned `Guid/ArmTransferList.h` is for `Guid/TransferListHob.h`.
So, including `Guid/TransferListHob.h` is needed for PlatformPeim.

But, PlatformPeiLib doesn't use `gArmTransferListHobGuid` directly so, I'll remove 
"Guid/TransferListHob.h" in here.

Since in PlatformPeim uses "gArmTransferListHobGuid" to build TransferlistHob,
I'll restore to declaration of gArmTransferListHobGuid  in .inf

Thanks!

ArmPlatformPkg/PlatformPei/PlatformPeiLib.c
```diff
@@ -9,16 +9,56 @@
...
   VOID
   )
 {
+  EFI_STATUS            Status;
+  TRANSFER_LIST_HEADER  *TransferList;
+
   BuildFvHob (PcdGet64 (PcdFvBaseAddress), PcdGet32 (PcdFvSize));
 
+  Status = TransferListGetFromHobList (&TransferList);
+  if (!EFI_ERROR (Status)) {
```
ardbiesheuvel: Please invert the condition, so that the non-TPM case returns early, and the rest of the code can deal with the TPM case unconditionally.

LeviYeoReum: Thanks. I'll change it.

ardbiesheuvel (approved):

PR #11350, opened 2025-07-28
UefiPayloadPkg: Fdtparserlib: Updates to handle above 4G DMA and reserved memory ranges

cap2k4-rivos:
# Description

Adds new PCD to control the DMA above 4G support, default is set to TRUE

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

N/A

## Integration Instructions

N/A


UefiPayloadPkg/Library/FdtParserLib/FdtParserLib.c
```diff
@@ -774,6 +774,7 @@ ParsePciRootBridge (
       mPciRootBridgeInfo->RootBridge[RbIndex].PMem.Limit           = PcdGet32 (PcdPciReservedPMemLimit);
       mPciRootBridgeInfo->RootBridge[RbIndex].UID                  = RbIndex;
       mPciRootBridgeInfo->RootBridge[RbIndex].HID                  = EISA_PNP_ID (0x0A03);
+      mPciRootBridgeInfo->RootBridge[RbIndex].DmaAbove4G           = PcdGetBool (PcdPciReservedDmaAbove4G);
```
gdong1: As the comments mentioned, we should not avoid hard-coded value in payload. Do you think we could check 'dma-ranges' to decided DmaAbove4G?  If dma-range sub-node is not found, or only below 4GB regions, then set DmaAbove4G to FALSE, else set it to true. 
dma-ranges | O | prop-encoded-array or empty | Refer to here for generic definition: 4.1.2.4 dma-rangesDefine the DMA address width capability of the root bridge. If only below 4GB region was defined, the payload may assume above 4GB DMA is not supported.

dhaval-rivos: I think checking range might work too. It is more dynamic in that sense.

dhaval-rivos: @cap2k4-rivos The commit title needs to change as you are addressing 2 distinct items here. You could make it update/enhancements to handle above 4G DMA and reserver memory ranges or something of that sort.

Also we could add more description to above 4G support commit (both commit and implementation with some comments)

gdong1 (approved):

PR #11356, opened 2025-07-29
UefiCpuPkg: Use DT based boot hartid

cap2k4-rivos:
Add a fall back to fetch the boot-hartid directly from the device tree.

# Description

This patch modifies UefiCpuPkg/CpuDxeRiscV64 to fetch the boot hartid directly from the device tree in case of RISCV_SEC Hob being null.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?
## How This Was Tested

N/A

## Integration Instructions

N?A

### DIFF
--- a/UefiCpuPkg/CpuDxeRiscV64/CpuDxe.c
+++ b/UefiCpuPkg/CpuDxeRiscV64/CpuDxe.c
@@ -9,6 +9,10 @@
 **/
 
 #include "CpuDxe.h"
+#include <Pi/PiBootMode.h>
+#include <Pi/PiHob.h>
+#include <Library/HobLib.h>
+#include <Library/FdtLib.h>
 
 //
 // Global Variables
@@ -337,10 +341,47 @@ InitializeCpu (
   const EFI_GUID          SecHobDataGuid = RISCV_SEC_HANDOFF_HOB_GUID;
 
   Hob = GetFirstGuidHob (&SecHobDataGuid);
-  ASSERT (Hob != NULL);
+  if (Hob != NULL) {
+    SecData     = GET_GUID_HOB_DATA (Hob);
+    mBootHartId = SecData->BootHartId;
+  } else {
+    CONST EFI_HOB_GUID_TYPE  *FdtHob = GetFirstGuidHob (&gFdtHobGuid);
 
-  SecData     = GET_GUID_HOB_DATA (Hob);
-  mBootHartId = SecData->BootHartId;
+    ASSERT (FdtHob != NULL);
+
+    CONST VOID  *DeviceTreeBase =
+      (CONST VOID *)(UINTN)*(CONST UINT64 *)GET_GUID_HOB_DATA (FdtHob);
+
+    ASSERT (FdtCheckHeader (DeviceTreeBase) == 0);
+
+    //
+    // /chosen node
+    //
+    INT32  Node = FdtSubnodeOffsetNameLen (
+                    DeviceTreeBase,
+                    0,
+                    "chosen",
+                    sizeof ("chosen") - 1
+                    );
+
+    ASSERT (Node >= 0);
+
+    //
+    // boot-hartid
+    //
+    INT32               Len;
+    CONST FDT_PROPERTY  *Prop =
+      FdtGetProperty (DeviceTreeBase, Node, "boot-hartid", &Len);
+
+    ASSERT (Prop != NULL && Len == sizeof (UINT32));
+
+    //
+    // Device-tree cells are big-endian
+    //
+    UINTN  BootHartId = SwapBytes32 (*(CONST UINT32 *)Prop->Data);
+
+    mBootHartId = BootHartId;
+  }
 
   DEBUG ((DEBUG_INFO, " %a: mBootHartId = 0x%x.\n", __func__, mBootHartId));
 


ardbiesheuvel (changes requested):
Please put the more elaborate description from the PR into the patch commit log.

cap2k4-rivos:
> Please put the more elaborate description from the PR into the patch commit log.

Added the description to the commit message, please take a look. Thanks

cap2k4-rivos:
@kraxel @jiaxinwu @niruiyu @ardbiesheuvel @lzeng14 @rahul1-kumar Kindly take a look at this PR.

dhaval-rivos:
@vlsunil

cap2k4-rivos:
@kraxel, @jiaxinwu, @niruiyu, @ardbiesheuvel, @lzeng14, @rahul1-kumar, @vlsunil, kindly take a look at this PR.

UefiCpuPkg/CpuDxeRiscV64/CpuDxe.c
```diff
@@ -337,10 +341,47 @@ InitializeCpu (
...
-
-  SecData     = GET_GUID_HOB_DATA (Hob);
-  mBootHartId = SecData->BootHartId;
+  if (Hob != NULL) {
+    SecData     = GET_GUID_HOB_DATA (Hob);
+    mBootHartId = SecData->BootHartId;
+  } else {
+    CONST EFI_HOB_GUID_TYPE  *FdtHob = GetFirstGuidHob (&gFdtHobGuid);
+
+    ASSERT (FdtHob != NULL);
```
vlsunil: I would prefer to keep the initialization of the local separate and also validate instead of simple assert here.

UefiCpuPkg/CpuDxeRiscV64/CpuDxe.c
```diff
@@ -337,10 +341,47 @@ InitializeCpu (
...
+
+    CONST VOID  *DeviceTreeBase =
+      (CONST VOID *)(UINTN)*(CONST UINT64 *)GET_GUID_HOB_DATA (FdtHob);
```
vlsunil: Why do we need all these typecasting?

UefiCpuPkg/CpuDxeRiscV64/CpuDxe.c
```diff
@@ -337,10 +341,47 @@ InitializeCpu (
...
+      FdtGetProperty (DeviceTreeBase, Node, "boot-hartid", &Len);
+
+    ASSERT (Prop != NULL && Len == sizeof (UINT32));
+
+    //
+    // Device-tree cells are big-endian
+    //
+    UINTN  BootHartId = SwapBytes32 (*(CONST UINT32 *)Prop->Data);
+
+    mBootHartId = BootHartId;
```
vlsunil: Why can't we directly use mBootHartId?

UefiCpuPkg/CpuDxeRiscV64/CpuDxe.c
```diff
@@ -337,10 +341,47 @@ InitializeCpu (
...
+    INT32               Len;
+    CONST FDT_PROPERTY  *Prop =
+      FdtGetProperty (DeviceTreeBase, Node, "boot-hartid", &Len);
+
+    ASSERT (Prop != NULL && Len == sizeof (UINT32));
+
+    //
+    // Device-tree cells are big-endian
+    //
+    UINTN  BootHartId = SwapBytes32 (*(CONST UINT32 *)Prop->Data);
```
vlsunil: Please don't rely on ASSERTs for the validity of DT nodes/properties. They may not exist.

UefiCpuPkg/CpuDxeRiscV64/CpuDxe.c
```diff
@@ -337,10 +341,47 @@ InitializeCpu (
...
+                    sizeof ("chosen") - 1
+                    );
+
+    ASSERT (Node >= 0);
+
+    //
+    // boot-hartid
+    //
+    INT32               Len;
+    CONST FDT_PROPERTY  *Prop =
```
vlsunil: I can't find now but I vaguely remember edk2 coding standards expect variable initialization to be separate. I thought checkpatch/uncrustify would fail otherwise.

UefiCpuPkg/CpuDxeRiscV64/CpuDxe.c
cap2k4-rivos: Hey @vlsunil, the typecasting here is actually needed. Without dereferencing, the pointer will point to the HOB's data section, not the device tree. This is what this is doing, we first get a VOID * pointing to the start of the HOB's data section then casts this pointer to a "pointer to a constant 64-bit integer" assuming that the data inside the HOB is a 64-bit address then dereferences the pointer to read the 64-bit address value that is stored inside the HOB and then final cast to make it a void *.

vlsunil: (CONST VOID *) *(UINTN *)GET_GUID_HOB_DATA (FdtHob) - Isn't it sufficient?

UefiCpuPkg/CpuDxeRiscV64/CpuDxe.c
```diff
@@ -337,10 +341,60 @@ InitializeCpu (
...
+      DEBUG ((DEBUG_WARN, "/chosen node not found in FDT.\n"));
+      return EFI_NOT_FOUND;
+    }
+
+    //
+    // boot-hartid
+    //
+    Prop = FdtGetProperty (DeviceTreeBase, Node, "boot-hartid", &Len);
+
+    if ((Prop == NULL) || (Len != sizeof (UINT32))) {
```
vlsunil: Why this check Len != sizeof (UINT32? What if hartid is a 64-bit value?

UefiCpuPkg/CpuDxeRiscV64/CpuDxe.c
cap2k4-rivos: Yes, actually, it would be sufficient. Thanks

UefiCpuPkg/CpuDxeRiscV64/CpuDxe.c
cap2k4-rivos: I have modified the check. I now check for both 32-bit and 64-bit values. Along with that, in case of a 64-bit value on a 32-bit system, I return an EFI_UNSUPPORTED status, as it might lead to loss of data.

UefiCpuPkg/CpuDxeRiscV64/CpuDxe.c
```diff
@@ -337,10 +341,73 @@ InitializeCpu (
...
+    }
+
+    //
+    // Device-tree cells are big-endian
+    //
+    if (Len == sizeof (UINT32)) {
+      mBootHartId = SwapBytes32 (*(CONST UINT32 *)Prop->Data);
+    } else if (Len == sizeof (UINT64)) {
+      UINT64  HartId64 = SwapBytes64 (*(CONST UINT64 *)Prop->Data);
+      if (HartId64 > MAX_UINTN) {
```
vlsunil: Since Len is already checked, I think this is not necessary. Simpler way would be to just return error if Len > sizeof (UINTN).

UefiCpuPkg/CpuDxeRiscV64/CpuDxe.c
```diff
@@ -337,10 +341,72 @@ InitializeCpu (
...
   Hob = GetFirstGuidHob (&SecHobDataGuid);
-  ASSERT (Hob != NULL);
-
-  SecData     = GET_GUID_HOB_DATA (Hob);
-  mBootHartId = SecData->BootHartId;
+  if (Hob != NULL) {
+    SecData     = GET_GUID_HOB_DATA (Hob);
+    mBootHartId = SecData->BootHartId;
+  } else {
+    const EFI_HOB_GUID_TYPE  *FdtHob;
```
lzeng14, line 348: Should the code use const and CONST consistently ?

PR #11357, opened 2025-07-29
Fixes for `GenerateCapsule.py --decode`

SergiiDmytruk:
# Description

This is a follow-up on https://github.com/tianocore/edk2/pull/5807, addressing some more issues with unpacking capsules.

See individual commits for explanation of what they are fixing.

- [ ] Breaking change?
  - No, unless somebody relied on the ability to specify only some OpenSSL-related options for `--decode` and still get signature verification.
- [ ] Impacts security?
  - No.
- [ ] Includes tests?
  - No.

## How This Was Tested

Ran `--decode` on a capsule, then packed it back using `--encode` with no modifications.  This didn't work previously due to `"Dependencies": "None"`, it would also built a capsule without DXEs because those weren't mentioned in the JSON made by `--decode`.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Source/Python/Capsule/GenerateCapsule.py
+++ b/BaseTools/Source/Python/Capsule/GenerateCapsule.py
@@ -433,8 +433,13 @@ if __name__ == '__main__':
                                             DepexExp
                                             ))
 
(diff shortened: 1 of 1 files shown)


bexcran:
Ping to keep this alive.

PR #11362, opened 2025-07-30
Move RealTimeClockLib header from EmbeddedPkg to MdeModulePkg

sachinami:
# Description

RealTimeClockLib is a generic architectural implementation that is not restricted to the embedded segment or any platform in particular.
So it can be moved to MdeModulePkg.

This change can also help reduce fragmentation between x86 and ARM implementations as it is a de facto standard for ARM platforms, but rarely used on x86 platforms.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?


sachinami:
@mdkinney Patch Check still fails despite adding the "Continuous-integration-options: PatchCheck.ignore-multi-package" tag to the commit.
Am I missing something here.

leiflindholm:
> @mdkinney Patch Check still fails despite adding the "Continuous-integration-options: PatchCheck.ignore-multi-package" tag to the commit. Am I missing something here.

That bit is working, but is lowering the severity to a "warning" rather than an error.

The error is probably the ` * Subject line not in "package/component: description" format!` a few lines above. I.e., the subject line does not enumerate the modified packages.

leiflindholm:
I agree with moving this out of EmbeddedPkg, but as it's not an industry standard, I think MdeModulePkg would be a better fit.

EmbeddedPkg/EmbeddedPkg.dec
```diff
@@ -32,7 +32,6 @@
```
leiflindholm, line 32: Comment on commit message:
RealTimeClock is misspelled (and should really be RealTimeClockLib).

But please drop the second paragraph, the justification bit doesn't belong in the changelog. (Could make sense in the overall PR description.)

sachinami:
> > @mdkinney Patch Check still fails despite adding the "Continuous-integration-options: PatchCheck.ignore-multi-package" tag to the commit. Am I missing something here.
> 
> That bit is working, but is lowering the severity to a "warning" rather than an error.
> 
> The error is probably the ` * Subject line not in "package/component: description" format!` a few lines above. I.e., the subject line does not enumerate the modified packages.

Got it. I understand that this PR would solve the issue https://github.com/tianocore/edk2/pull/11321

EmbeddedPkg/EmbeddedPkg.dec
sachinami: Thank you ! The commit message has been modified.

sachinami:
> I agree with moving this out of EmbeddedPkg, but as it's not an industry standard, I think MdeModulePkg would be a better fit.

I chose MdePkg instead of MdeModulePkg because:
1) The Real Time Clock protocol header already resides there:

```
  ## Include/Protocol/RealTimeClock.h
  gEfiRealTimeClockArchProtocolGuid = { 0x27CFAC87, 0x46CC, 0x11D4, { 0x9A, 0x38, 0x00, 0x90, 0x27, 0x3F, 0xC1, 0x4D }}
```
 
2) The components which use RealTimeClockLib already have MdePkg included but not MdeModulePkg.
Making this a minimal change.

leiflindholm:
> > I agree with moving this out of EmbeddedPkg, but as it's not an industry standard, I think MdeModulePkg would be a better fit.
> 
> I chose MdePkg instead of MdeModulePkg because:
> 
>     1. The Real Time Clock protocol header already resides there:
> 
> 
> ```
>   ## Include/Protocol/RealTimeClock.h
>   gEfiRealTimeClockArchProtocolGuid = { 0x27CFAC87, 0x46CC, 0x11D4, { 0x9A, 0x38, 0x00, 0x90, 0x27, 0x3F, 0xC1, 0x4D }}
> ```

Yes, but this is exactly what the MdePkg/MdeModulePkg split is for: 
EFI_REAL_TIME_CLOCK_ARCH_PROTOCOL_GUID is defined in the PI specification
RealTimeClockLib.h defines the interfaces of some generic helper functions someone wrote.

>     2. The components which use RealTimeClockLib already have MdePkg included but not MdeModulePkg.
>        Making this a minimal change.

I approve of the line of reasoning, but I think adhering to the established design philosophy needs to take priority.

(That said, I'm not a maintainer of either MdePkg or MdeModulePkg, so they should probably speak up too.)

leiflindholm:
> Got it. I understand that this PR would solve the issue #11321

It would, if we agreed that is the correct way to go...

Meanwhile, if you updated the patch subject line to say
`ArmPlatformPkg,EmbeddedPkg,MdeModulePkg: ...` it would go through now.

leiflindholm:
> > Got it. I understand that this PR would solve the issue #11321
> 
> It would, if we agreed that is the correct way to go...
> 
> Meanwhile, if you updated the patch subject line to say `ArmPlatformPkg,EmbeddedPkg,MdeModulePkg: ...` it would go through now.

Ah, I see you split up the patches to get around the issue.
That however breaks bisect. So I'd much prefer to have them squashed back together with the above tweak to the subject line.

sachinami:
> > > Got it. I understand that this PR would solve the issue #11321
> > 
> > 
> > It would, if we agreed that is the correct way to go...
> > Meanwhile, if you updated the patch subject line to say `ArmPlatformPkg,EmbeddedPkg,MdeModulePkg: ...` it would go through now.
> 
> Ah, I see you split up the patches to get around the issue. That however breaks bisect. So I'd much prefer to have them squashed back together with the above tweak to the subject line.

I believe the way the patches are split right now won't break bisect.
And the subject line has been tweaked.

leiflindholm:
> I believe the way the patches are split right now won't break bisect.
And the subject line has been tweaked.

The problem is that after the first commit, PL031RealTimeClockLib, TemplateRealTimeClockLib, VirtualRealTimeClockLib and RealTimeClockRuntimeDxe have an undeclared dependency on MdeModulePkg, which means they will fail to build until that declaration is introduced by the subsequent commits.

sachinami:
> > I believe the way the patches are split right now won't break bisect.
> > And the subject line has been tweaked.
> 
> The problem is that after the first commit, PL031RealTimeClockLib, TemplateRealTimeClockLib, VirtualRealTimeClockLib and RealTimeClockRuntimeDxe have an undeclared dependency on MdeModulePkg, which means they will fail to build until that declaration is introduced by the subsequent commits.

Thank you for the clarification Leif. The commits have been squashed.

sachinami:
Opened complimentary PR in edk2-platforms: https://github.com/tianocore/edk2-platforms/pull/840

sachinami:
Hi @leiflindholm Could you help merge the changes as edk2 stable tag has been released.

sachinami:
@lgao4 Could you please help merge this approved PR. Thank you in advance.

lgao4:
@sachinami , done

sachinami:
@lgao4 Appreciate your constant support!

leiflindholm:
> Hi @leiflindholm Could you help merge the changes as edk2 stable tag has been released.

Apologies, you caught me just as I went off on holiday. Happy to see it was merged in my absence.

PR #11367, opened 2025-07-31
Move gFdtTableGuid from EmbeddedPkg to MdePkg

PrachotanReddy:
Moved the gFdtTableGuid from EmbeddedPkg to MdePkg.
This also fixes commit dcde1480bf9f42441925fa4196f6879ec9044d81, which introduced an unintended dependency between ShellPkg and EmbeddedPkg.

Cc: Hema Anmisha <hema.anmisha.kalavakolanu@intel.com>

# Description
ShellPkg was updated as part of a previous PR to allow the dmem utility to show the DTB address. However, this introduced a dependency on EmbeddedPkg.
hema.anmisha.kalavakolanu@intel.com reported:
ShellPkg is not expected to have dependency with EmbeddedPkg, as it is specific to Embedded segment.
ShellPkg is a generic package likes MdeModulePkg or NetworkPkg.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Built and run on FVP.

## Integration Instructions

N/A


ardbiesheuvel (approved):
Please move the GUID definition to MdePkg instead (assuming that would solve the dependency issue)

leiflindholm:
Based on Ard's comment, I expect he set the push label by accident.
The fix is clearly what he suggested: moving gFdtTableGuid to MdePkg.

PrachotanReddy:
Hi @leiflindholm and @ardbiesheuvel 
I reviewed this, one thing that stands out is the gFdtTableGuid is not specified in the UEFI specification, I found a [thread](https://github.com/ARM-software/ebbr/issues/45) which calls this vendor specific,  I'm not sure if it should go in the MdePkg, . Additionally this guid was introduced to EmbeddedPkg about 10 years ago and since been used as a dependency with other packages, I'm not very confident if this will be a breaking change affecting the dependency chain. What do you think? 
Also, this commit was introduced in the original [PR](https://github.com/tianocore/edk2/pull/10759) since it had initial commits which were handling DT and they have been replaced after review. This ShellPkg commit is not an immediate requirement at the moment, but yet it's apparently blocking builds as reported by folks at Intel, reverting seems like a clean approach for now while we figure out the mechanics of where gFdtTableGuid must go.

leiflindholm:
> Hi @leiflindholm and @ardbiesheuvel I reviewed this, one thing that stands out is the gFdtTableGuid is not specified in the UEFI specification, I found a [thread](https://github.com/ARM-software/ebbr/issues/45) which calls this vendor specific, I'm not sure if it should go in the MdePkg, .

A good many things that are industry standards but not part of UEFI live in MdePkg.

This is a de-facto standard GUID used across several open source codebases (and probably some closed ones) for a decade and a half. I don't think the use of one specific word in a single conversation 5 years ago trumps that.

> Additionally this guid was introduced to EmbeddedPkg about 10 years ago and since been used as a dependency with other packages, I'm not very confident if this will be a breaking change affecting the dependency chain. What do you think?

Moving a GUID is a simple mechanical change, and if it is the right thing to do then it should be done. It has a very minimal one-off overhead in any but the very most broken configuration management processes. That applies for any GUID. (Clearly we should try to avoid pointless churn.)

*This* particular move could only cause a problem for a module that currently depended on EmbeddedPkg but _not_ on MdePkg. I can't see how that would be possible.

Moreover: *not* moving the GUID would seem to suggest the Shell should never be able to use it and this functionality is being permanently removed.

> Also, this commit was introduced in the original [PR](https://github.com/tianocore/edk2/pull/10759) since it had initial commits which were handling DT and they have been replaced after review. This ShellPkg commit is not an immediate requirement at the moment, but yet it's apparently blocking builds as reported by folks at Intel, reverting seems like a clean approach for now while we figure out the mechanics of where gFdtTableGuid must go.

While we shouldn't be completely flippant about breaking downstream workflows, this GUID residing in EmbeddedPkg can only cause a problem if one of two things is taking place:
- Some Intel-internal process is discarding parts of the upstream repository before building.
- A test looking for cross-package dependencies is run and flags an error.

A downstream-only failure should not cause a revert-until-we-figure-this-out unless there are strong reasons for doing so. When there is a very simple actual fix available, which doesn't break other people's workflows, that reason does not exist.

EmbeddedPkg/EmbeddedPkg.dec
```diff
@@ -50,10 +50,6 @@
 [Guids.common]
   gEmbeddedTokenSpaceGuid       = { 0xe0d8ca17, 0x4276, 0x4386, { 0xbb, 0x79, 0x48, 0xcb, 0x81, 0x3d, 0x3c, 0x4f }}
 
-  ## FDT Configuration Table
-  # Include/Guid/Fdt.h
-  gFdtTableGuid = { 0xb1b621d5, 0xf19c, 0x41a5, { 0x83, 0x0b, 0xd9, 0x15, 0x2c, 0x69, 0xaa, 0xe0 } }
-  gFdtVariableGuid = { 0x25a4fd4a, 0x9703, 0x4ba9, { 0xa1, 0x90, 0xb7, 0xc8, 0x4e, 0xfb, 0x3e, 0x57 } }
```
leiflindholm, line 56: gFdtTableGuid is an industry standard thing.
But gFdtVariableGuid is as far as I can tell in fact only used by Arm Ltd. platforms,
in edk2-platforms/Platform/ARM/Drivers/FdtPlatformDxe/FdtPlatform.c

Can we drop moving gFdtTableGuid from this change? I think that merits a (longer) discussion on its own merits. And regardless of outcome, I don't think MdePkg is where it should move.

leiflindholm:
Why bother with the revert and reintroduction?

EmbeddedPkg/EmbeddedPkg.dec
PrachotanReddy: > Can we drop moving gFdtTableGuid from this change? I think that merits a (longer) discussion on its own merits. And regardless of outcome, I don't think MdePkg is where it should move.

Do you mean gFdtVariableGuid? That would mean, I probably have to create a seperate Guid/FdtVariable.h in EmbeddedPkg?

leiflindholm: Gah, yes, sorry - gFdtVariableGuid.
Oh, right, that's why you're moving both... Urgh.

@samimujawar how would you like to deal with this? Are there downstream users of gFdtVariableGuid that you know of? Could it move across into ARM.dec?

mdkinney:
Should the push label be removed from this PR?

Should this revert PR be closed and a new PR created to address the package that declares the GUIDs?

leiflindholm:
> Should the push label be removed from this PR?
> 
> Should this revert PR be closed and a new PR created to address the package that declares the GUIDs?

I think that would make the most sense.

mdkinney:
> Why bother with the revert and reintroduction?

I agree.  The revert commit should be removed.  The title should be changed to indicate that the GUIDs declarations are being moved from `EmbeddedPkg` to `MdePkg`.

I do not see any use of the `gFdtVariableGuid` in edk2 repo.  I do see it used in `edk2-platforms` repo, and that GUID is only used if the Feature PCD `PcdOverridePlatformFdt` is set and `PcdOverridePlatformFdt` is declared in `edk2-platforms` repo.  So I think the declaration of `gFdtVariableGuid` in edk2 repo is incorrect and `gFdtVariableGuid`  should be moved to `edk2-platforms` repo in the same DEC file that `PcdOverridePlatformFdt` is declared which is `Platform/ARM/Drivers/FdtPlatformDxe/FdtPlatformDxe.dec`.

mdkinney:
Removed push label until this PR is updated

leiflindholm (approved):

leiflindholm:
The first commit needs a
tag in the commit message.

ardbiesheuvel:
> Although gFdtTableGuid is not in the UEFI spec, it is a de-facto standard GUID used across projects. 

This is not true, actually - it does appear in the UEFI spec, and it is called `EFI_DTB_TABLE_GUID` there.

leiflindholm:
> > Although gFdtTableGuid is not in the UEFI spec, it is a de-facto standard GUID used across projects.
> 
> This is not true, actually - it does appear in the UEFI spec, and it is called `EFI_DTB_TABLE_GUID` there.

Oh, right! I forgot we eventually got around to add it.

*makes mental note to follow up post stable tag with a spec alignment cleanup*

mdkinney:
This PR removes `gFdtVariableGuid`.  This will break edk2-platforms builds.  Is there an edk2-platforms PR to add `gFdtVariableGuid`.  Seems like the edk2-platforms PR would need to be merged before this one.

leiflindholm:
> This PR removes `gFdtVariableGuid`. This will break edk2-platforms builds. Is there an edk2-platforms PR to add `gFdtVariableGuid`. Seems like the edk2-platforms PR would need to be merged before this one.

It only breaks Platform/ARM/Drivers/FdtPlatformDxe/, which in turn Dis only used by ArmJuno.
I don't want to be too flippant about breaking platforms, but that platform is ancient and an Arm Ltd platform. I don't think it needs to gate this PR being merged.

Ard's comment from when it was moved into edk2-platforms from edk2 in 2017 was
```
   Import FdtPlatformDxe from EmbeddedPkg into Platform/ARM, given that it
   is not used anywhere else, nor should it be.
```

@samimujawar is this in fact desirable to keep even for Juno?

mdkinney:
@samerhaj I will open an issue for `gFdtVariableGuid` in `edk2-platforms` and assign to you.  

We will move forward with merging this PR for the `edk2-stable202508` release.

mdkinney (approved):

EmbeddedPkg/EmbeddedPkg.dec
mdkinney: Will move to an edk2-platforms issue.

leiflindholm:
> @samerhaj I will open an issue for `gFdtVariableGuid` in `edk2-platforms` and assign to you.
> 
> We will move forward with merging this PR for the `edk2-stable202508` release.

Wrong target. Github autocompleted the same for me, I just happened to spot it before pressing submit :)

@samimujawar

mdkinney:
@mergify rebase

PR #11369, opened 2025-07-31
{UefiCpuPkg,UefiPayloadPkg}: Include what you use

benjamindoron:
# Description

It's apparent that this wasn't even build-tested until now. Declare the necessary packages and library classes to link against to allow this to compile.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Build UefiPayload for AARCH64 (with the CLANGDWARF compiler) before and after, and see that it now compiles successfully. Clang and GCC fail to find lto-aarch64 too, and it's unclear at this time whether it's only a local issue. To be looked into later. Consider dropping this from your tools_def.txt if you experience problems too.

## Integration Instructions

N/A

### DIFF
--- a/UefiCpuPkg/Library/BaseArchSupportLib/BaseArchSupportLib.inf
+++ b/UefiCpuPkg/Library/BaseArchSupportLib/BaseArchSupportLib.inf
@@ -28,3 +28,6 @@
 
 [LibraryClasses]
   BaseLib
+
+[LibraryClasses.AARCH64]
+  ArmLib
--- a/UefiPayloadPkg/UefiPayloadPkg.dsc
+++ b/UefiPayloadPkg/UefiPayloadPkg.dsc
@@ -426,11 +426,12 @@
   ArmHvcLib|ArmPkg/Library/ArmHvcLib/ArmHvcLib.inf
   ArmLib|MdePkg/Library/ArmLib/ArmBaseLib.inf
   ArmMmuLib|UefiCpuPkg/Library/ArmMmuLib/ArmMmuBaseLib.inf
+  ArmMonitorLib|ArmPkg/Library/ArmMonitorLib/ArmMonitorLib.inf
   ArmSmcLib|MdePkg/Library/ArmSmcLib/ArmSmcLib.inf
 
   BaseMemoryLib|MdePkg/Library/BaseMemoryLib/BaseMemoryLib.inf
   CacheMaintenanceLib|ArmPkg/Library/ArmCacheMaintenanceLib/ArmCacheMaintenanceLib.inf
-  EfiResetSystemLib|ArmPkg/Library/ArmPsciResetSystemLib/ArmPsciResetSystemLib.inf
+  ResetSystemLib|ArmPkg/Library/ArmPsciResetSystemLib/ArmPsciResetSystemLib.inf
   PL011UartLib|ArmPlatformPkg/Library/PL011UartLib/PL011UartLib.inf
   PL011UartClockLib|ArmPlatformPkg/Library/PL011UartClockLib/PL011UartClockLib.inf
   SerialPortLib|ArmPlatformPkg/Library/PL011SerialPortLib/PL011SerialPortLib.inf


UefiPayloadPkg/BlSupportDxe/BlSupportDxe.inf
```diff
@@ -39,6 +39,9 @@
   UefiPayloadPkg/UefiPayloadPkg.dec
   UefiCpuPkg/UefiCpuPkg.dec
 
+[Packages]
```
ChaselChiu: should this be [Packages.AARCH64]?

benjamindoron: Yes, it should. Done. Thanks!

gdong1 (approved):
It looks good to me.

ChaselChiu (approved):

benjamindoron:
This throws an error for:
> Dependency Check: UefiCpuPkg/Library/BaseArchSupportLib/BaseArchSupportLib.inf depends on pkg ArmPkg/ArmPkg.dec but pkg is not listed in AcceptableDependencies

Do we know why this is the case? The AARCH64 instance of this library uses ArmLib, that seems to be where all the actual ARM functionality is located, that seems fine to me. Can we update the rules?

makubacki:
> This throws an error for:
> 
> > Dependency Check: UefiCpuPkg/Library/BaseArchSupportLib/BaseArchSupportLib.inf depends on pkg ArmPkg/ArmPkg.dec but pkg is not listed in AcceptableDependencies
> 
> Do we know why this is the case? The AARCH64 instance of this library uses ArmLib, that seems to be where all the actual ARM functionality is located, that seems fine to me. Can we update the rules?

ArmPkg/ArmPkg.dec is not listed as an allowed dependency in https://github.com/tianocore/edk2/blob/d46aa46c8361194521391aa581593e556c707c6e/UefiCpuPkg/UefiCpuPkg.ci.yaml#L46-L50

benjamindoron:
> > This throws an error for:
> > > Dependency Check: UefiCpuPkg/Library/BaseArchSupportLib/BaseArchSupportLib.inf depends on pkg ArmPkg/ArmPkg.dec but pkg is not listed in AcceptableDependencies
> > 
> > 
> > Do we know why this is the case? The AARCH64 instance of this library uses ArmLib, that seems to be where all the actual ARM functionality is located, that seems fine to me. Can we update the rules?
> 
> ArmPkg/ArmPkg.dec is not listed as an allowed dependency in
> 
> https://github.com/tianocore/edk2/blob/d46aa46c8361194521391aa581593e556c707c6e/UefiCpuPkg/UefiCpuPkg.ci.yaml#L46-L50

Okay, thanks. I've added a commit to this PR to add ArmPkg as an acceptable dependency. While I think that it may be better to split ArmLib or migrate at least some of its functionality out, I'm not maintaining Arm in EDK2, so I'll let them handle that.

mdkinney:
What needs to change for the content needed from ArmPkg to be moved to either the UefiCpuPkg or the MdePkg?

benjamindoron:
The issue is ArmLib. It's my understanding that that library is a grab-bag of Arm-specific functionality and assorted assembly, sometimes overlapping BaseLib, which is a design I don't like or understand. In this case, we'd just have to move the definition of `ArmGetPhysicalAddressBits()` out of ArmLib and into another library, possibly this one (and handle all consumers that used the function through ArmLib), or duplicate it here. I don't like the idea of duplicating the code, and to break ArmLib up (which is the solution I think is best), we'd want to plan how it's going to be split up, which may need some planning and input from Arm's maintainers.

This is why I thought I'd just allow ArmPkg as a dependency, but if that's unacceptable, then we probably want to pull ArmPkg maintainers in. And until then, there may be a linker error for AARCH64's variant of BaseArchSupportLib if no other library linked into the module declares ArmLib.

benjamindoron:
Another idea is to accept that ArmLib is BaseLib-esque (it even calls itself ArmBaseLib in the INF), and move the whole library to UefiCpuPkg (probably, since I believe that MdePkg is more where fundamental and/or "base" code goes)

gguo11837463 (approved):
Approve from UefiPayloadPkg

niruiyu (approved):

gdong1 (approved):

lzeng14:
> Another idea is to accept that ArmLib is BaseLib-esque (it even calls itself ArmBaseLib in the INF), and move the whole library to UefiCpuPkg (probably, since I believe that MdePkg is more where fundamental and/or "base" code goes)

The dependency is about library class, or library instance ?
ArmLib.h has been in MdePkg, MdePkg/Include/Library/ArmLib.h.

benjamindoron:
> The dependency is about library class, or library instance ? ArmLib.h has been in MdePkg, MdePkg/Include/Library/ArmLib.h.

Thanks for this. Done: I've dropped the unneeded ArmPkg.dec package reference, and so this now doesn't need to change the CI rules.

mikebeaton:
Not stale

benjamindoron:
Can we merge this? I think it has all the approvals it needs.

makubacki:
> Can we merge this? I think it has all the approvals it needs.

Attn: @niruiyu, @jiaxinwu, @gdong1, @Sean-StarLabs, @jameslu8

Also, @benjamindoron, as a UefiPayloadPkg maintainer with approval for both packages, you can also add the push label.

PR #11370, opened 2025-07-31
StandaloneMmPkg: Implement IPL in DXE and support platform-specific launches

benjamindoron:
# Description

coreboot's UefiPayload is preparing to use the standalone MM environment for supporting payload features that run in an SMM/MM environment, such as authenticated variables. This requires some of the following changes, some of which enhance common code too:

UefiPayload skips the PEI phase, so factor out the PEI-specific logic into a PEI-specific files, and implement a DXE variant of the MM IPL. An MM using this IPL has the same TCB as SMM, so the new IPL is presumed okay, but platforms should consider which IPL they need because this grows their TCB. As a side-effect, this makes integrating an MM environment into a platform with a 32-bit PEI easier: they can simply defer loading MM until DXE. If platforms with 32-bit PEIs should actually be officially supported, however, it would be better to have a module to do a mode switch instead, or implement this in the MM core entrypoint library.

coreboot's use of standalone MM (called "payload MM," and found at the patch series containing https://review.coreboot.org/c/coreboot/+/79870) is similar to Arm's usage in that a trusted environment is already initialised, and standalone MM is being allowed in to SMRAM too. Since SMRAM is already locked down, we need a platform-specific mechanism to get inside. coreboot exposes an SMI handler for this purpose, that allows one core/module to be copied in and called. As standalone MM may service other such environments, we implement a new library for the IPL that hooks the calling of the MM core, allowing platforms to provide their own way for this, and falling back to a plain function call otherwise. An example of this can be found at https://github.com/benjamindoron/edk2/commit/ece725ab15bd173ad2d48cbc480d89e93f084f08, which will be upstreamed in time.

As neither coreboot's payload MM nor Arm's usage of standalone MM use PiSmmCpuStandaloneMm, this means that even HOBs copied by common code might not actually be needed, and so, their absence might not be a problem. Therefore, drop the assertion in MmIplCopyGuidHob.

- [x] Breaking change?
  - **Breaking change**: MmIplPlatformHookLib must be defined. Platforms intending to preserve the current behaviour should use MmIplPlatformHookLibNull.
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested against coreboot, with the patch-series that includes https://review.coreboot.org/c/coreboot/+/79870 applied, and a UefiPayload with the additional payload MM support. As mentioned above, there is no PR for that yet, one will be prepared in time when ready (which also depends on this going in). https://github.com/benjamindoron/coreboot and https://github.com/benjamindoron/edk2/tree/dev/uefipayload are the tested trees, with all patches applied.

## Integration Instructions

Standalone MM platforms must now define MmIplPlatformHookLib.

### DIFF
--- a/StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
+++ b/StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
@@ -246,6 +246,46 @@ MmIplBuildMmStatusCodeUseSerialHob (
 }
 
 /**
+  Builds MMRAM descriptor HOB.
+
+  This function builds MMRAM descriptor HOB.
+  It can only be invoked during PEI phase;
+  If new HOB buffer is NULL, then ASSERT().
+
+  @param[in]       Hob            The pointer of new HOB buffer.
+  @param[in]       MmramBlocks    The MMRAM descriptor blocks to insert to HOB list.
+
+**/
+VOID
+MmIplBuildMmramDescriptorHob (
+  IN UINT8                           *HobBuffer,
+  IN OUT UINTN                       *HobBufferSize,
+  IN EFI_MMRAM_HOB_DESCRIPTOR_BLOCK  *MmramBlocks
+  )
+{
+  EFI_MMRAM_HOB_DESCRIPTOR_BLOCK  *MmramDescriptorBlock;
+  EFI_HOB_GUID_TYPE               *GuidHob;
+  UINT16                          HobLength;
+  UINT16                          DataSize;
+
+  DataSize = (UINT16)(sizeof (EFI_MMRAM_HOB_DESCRIPTOR_BLOCK) + (MmramBlocks->NumberOfMmReservedRegions - 1) * sizeof (EFI_MMRAM_DESCRIPTOR));
+
+  HobLength = ALIGN_VALUE (sizeof (EFI_HOB_GUID_TYPE) + DataSize, 8);
+  if (*HobBufferSize >= HobLength) {
+    ASSERT (HobBuffer != NULL);
+    MmIplCreateHob (HobBuffer, EFI_HOB_TYPE_GUID_EXTENSION, HobLength);
+
+    GuidHob = (EFI_HOB_GUID_TYPE *)HobBuffer;
+    CopyGuid (&GuidHob->Name, &gEfiSmmSmramMemoryGuid);
+
+    MmramDescriptorBlock = (EFI_MMRAM_HOB_DESCRIPTOR_BLOCK *)(GuidHob + 1);
+    CopyMem (MmramDescriptorBlock, MmramBlocks, DataSize);
+  }
+
+  *HobBufferSize = HobLength;
+}
+
+/**
   Copies a data buffer to a newly-built HOB for GUID HOB
 
   This function builds a customized HOB tagged with a GUID for identification, copies the
@@ -1013,7 +1053,7 @@ CreateMmFoundationHobList (
   // Build SMRAM memory Hob in MM HOB list
   //
   HobLength = GetRemainingHobSize (*FoundationHobSize, UsedSize);
-  MmIplCopyGuidHob (FoundationHobList + UsedSize, &HobLength, &gEfiSmmSmramMemoryGuid, FALSE);
+  MmIplBuildMmramDescriptorHob (FoundationHobList + UsedSize, &HobLength, Block);
   UsedSize += HobLength;
 
   //
--- a/StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
+++ b/StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
@@ -1067,6 +1067,16 @@ CreateMmFoundationHobList (
   UsedSize += HobLength;
 
   //
+  // Platforms may use the DXE IPL when CpuMpPei isn't or cannot be used.
+  // In this case, there is no Mp Information2 HOB in the primary list, so, build it here.
+  //
+  if (HobLength == 0) {
+    HobLength = GetRemainingHobSize (*FoundationHobSize, UsedSize);
+    MmIplBuildMpInformationHob (FoundationHobList + UsedSize, &HobLength);
+    UsedSize += HobLength;
+  }
+
+  //
   // Build ACPI variable HOB
   //
   HobLength = GetRemainingHobSize (*FoundationHobSize, UsedSize);
--- a/StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
+++ b/StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
@@ -290,7 +290,7 @@ MmIplBuildMmramDescriptorHob (
 
   This function builds a customized HOB tagged with a GUID for identification, copies the
   input data to the HOB data field and returns the start address of the GUID HOB data.
-  If new HOB buffer is NULL or the GUID HOB could not found, then ASSERT().
+  If new HOB buffer is NULL, then ASSERT().
 
   @param[in]       HobBuffer            The pointer of HOB buffer.
   @param[in, out]  HobBufferSize        The available size of the HOB buffer when as input.
@@ -311,7 +311,10 @@ MmIplCopyGuidHob (
 
   UsedSize = 0;
   GuidHob  = GetFirstGuidHob (Guid);
-  ASSERT (GuidHob != NULL);
+
+  if (GuidHob == NULL) {
+    DEBUG ((DEBUG_WARN, "GUIDed HOB %g could not be found\n", Guid));
+  }
 
   while (GuidHob != NULL) {
     if (*HobBufferSize >= UsedSize + GuidHob->HobLength) {
--- a/StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
+++ b/StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
@@ -267,6 +267,9 @@ MmIplAllocateMmramPage (
                                                             sizeof (EFI_MMRAM_HOB_DESCRIPTOR_BLOCK) + ((FullMmramRangeCount - 1) * sizeof (EFI_MMRAM_DESCRIPTOR))
                                                             );
   ASSERT (NewDescriptorBlock != NULL);
+  if (NewDescriptorBlock == NULL) {
+    return 0;
+  }
 
   NewDescriptorBlock->NumberOfMmReservedRegions = FullMmramRangeCount;
   FullMmramRanges                               = NewDescriptorBlock->Descriptor;
--- a/StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
+++ b/StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
@@ -224,7 +224,7 @@ FindLargestMmramRange (
   Allocate available MMRAM for MM core image.
 
   @param[in]  Pages                     Page count of MM core image.
-  @param[out] NewBlock                  Pointer of new mmram block HOB.
+  @param[out] NewBlock                  Pointer to new MMRAM blocks.
 
   @return  EFI_PHYSICAL_ADDRESS         Address for MM core image to be loaded in MMRAM.
 **/
@@ -262,8 +262,7 @@ MmIplAllocateMmramPage (
   // 2. Split the largest region and mark the allocated region as ALLOCATED
   //
   FullMmramRangeCount = CurrentBlock->NumberOfMmReservedRegions + 1;
-  NewDescriptorBlock  = (EFI_MMRAM_HOB_DESCRIPTOR_BLOCK *)BuildGuidHob (
-                                                            &gEfiSmmSmramMemoryGuid,
+  NewDescriptorBlock  = (EFI_MMRAM_HOB_DESCRIPTOR_BLOCK *)AllocatePool (
                                                             sizeof (EFI_MMRAM_HOB_DESCRIPTOR_BLOCK) + ((FullMmramRangeCount - 1) * sizeof (EFI_MMRAM_DESCRIPTOR))
                                                             );
   ASSERT (NewDescriptorBlock != NULL);
@@ -293,11 +292,6 @@ MmIplAllocateMmramPage (
   Allocated->PhysicalStart = Largest->PhysicalStart + Largest->PhysicalSize;
   Allocated->RegionState   = Largest->RegionState | EFI_ALLOCATED;
   Allocated->PhysicalSize  = EFI_PAGES_TO_SIZE (Pages);
-
-  //
-  // Scrub old one
-  //
-  ZeroMem (&MmramInfoHob->Name, sizeof (MmramInfoHob->Name));
 
   //
   // New MMRAM descriptor block
@@ -427,6 +421,7 @@ ExecuteMmCoreFromMmram (
       Entry = (MM_FOUNDATION_ENTRY_POINT)(UINTN)ImageContext.EntryPoint;
       Entry (MmHobList);
       FreePages (MmHobList, EFI_SIZE_TO_PAGES (MmHobSize));
+      FreePool (Block);
     }
   }
 
--- a/StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
+++ b/StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
@@ -315,6 +315,7 @@ ExecuteMmCoreFromMmram (
   )
 {
   EFI_STATUS                      Status;
+  EFI_STATUS                      Status2;
   UINTN                           PageCount;
   VOID                            *MmHobList;
   UINTN                           MmHobSize;
@@ -372,6 +373,11 @@ ExecuteMmCoreFromMmram (
   ImageContext.ImageAddress &= ~((EFI_PHYSICAL_ADDRESS)ImageContext.SectionAlignment - 1);
 
   //
+  // Call platform hook now that image info is known
+  //
+  PlatformHookBeforeMmLoad (&ImageContext);
+
+  //
   // Print debug message showing MM Core load address.
   //
   DEBUG ((DEBUG_INFO, "StandaloneMM IPL loading MM Core at MMRAM address %p\n", (VOID *)(UINTN)ImageContext.ImageAddress));
@@ -404,7 +410,7 @@ ExecuteMmCoreFromMmram (
                     MmFvBase,
                     MmFvSize,
                     &MmCoreFileName,
-                    ImageContext.ImageAddress,
+                    ImageContext.DestinationAddress != 0 ? ImageContext.DestinationAddress : ImageContext.ImageAddress,
                     EFI_PAGES_TO_SIZE (PageCount),
                     ImageContext.EntryPoint,
                     Block
@@ -418,8 +424,12 @@ ExecuteMmCoreFromMmram (
       //
       // Execute image
       //
-      Entry = (MM_FOUNDATION_ENTRY_POINT)(UINTN)ImageContext.EntryPoint;
-      Entry (MmHobList);
+      Status2 = PlatformHookCallMmCore (&ImageContext, MmHobList, &Status);
+      if (Status2 == EFI_UNSUPPORTED) {
+        Entry = (MM_FOUNDATION_ENTRY_POINT)(UINTN)ImageContext.EntryPoint;
+        Entry (MmHobList);
+      }
+
       FreePages (MmHobList, EFI_SIZE_TO_PAGES (MmHobSize));
       FreePool (Block);
     }
--- a/StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
+++ b/StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
@@ -327,10 +327,6 @@ ExecuteMmCoreFromMmram (
   PE_COFF_LOADER_IMAGE_CONTEXT    ImageContext;
   MM_FOUNDATION_ENTRY_POINT       Entry;
   EFI_MMRAM_HOB_DESCRIPTOR_BLOCK  *Block;
-  EFI_PEI_MM_ACCESS_PPI           *MmAccess;
-  UINTN                           Size;
-  UINTN                           Index;
-  UINTN                           MmramRangeCount;
 
   MmFvBase = 0;
   MmFvSize = 0;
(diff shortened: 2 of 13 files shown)


StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
niruiyu, line 1: I found the commit not only moves some functions to Common.c but also changes some function implementations.
We should separate the move and change to 2 commits.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
```diff
@@ -262,8 +262,7 @@ MmIplAllocateMmramPage (
   // 2. Split the largest region and mark the allocated region as ALLOCATED
   //
   FullMmramRangeCount = CurrentBlock->NumberOfMmReservedRegions + 1;
-  NewDescriptorBlock  = (EFI_MMRAM_HOB_DESCRIPTOR_BLOCK *)BuildGuidHob (
-                                                            &gEfiSmmSmramMemoryGuid,
+  NewDescriptorBlock  = (EFI_MMRAM_HOB_DESCRIPTOR_BLOCK *)AllocatePool (
```
niruiyu, line 265: Please update the commit message that changes this logic to be more specific.
The current commit message does not clearly explain the commit.

niruiyu:
please mark the PR as draft if you are WIP on it still.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
benjamindoron: I thought this would be fine, since I mention it in the commit message and it's also in scope of extracting phase-specific operations to another file. But sure, if it makes it easier to see changes that simply move functions around, and those that define something new, sure, done. It'll be in the next push.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
benjamindoron: I've updated the wording to elaborate on how DXE can't create HOBs, let me know if this is better

benjamindoron:
> please mark the PR as draft if you are WIP on it still.

Fair, noted for next time. I've implemented the final piece and migrated the commits I'd like comments on to https://github.com/tianocore/edk2/pull/11393, so this is properly ready now, and that will hopefully get comments as part of review.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
```diff
@@ -0,0 +1,478 @@
...
+  //
+  // 2. Split the largest region and mark the allocated region as ALLOCATED
+  //
+  FullMmramRangeCount = CurrentBlock->NumberOfMmReservedRegions + 1;
+  NewDescriptorBlock  = (EFI_MMRAM_HOB_DESCRIPTOR_BLOCK *)AllocatePool (
+                                                            sizeof (EFI_MMRAM_HOB_DESCRIPTOR_BLOCK) + ((FullMmramRangeCount - 1) * sizeof (EFI_MMRAM_DESCRIPTOR))
+                                                            );
+  ASSERT (NewDescriptorBlock != NULL);
+
+  NewDescriptorBlock->NumberOfMmReservedRegions = FullMmramRangeCount;
```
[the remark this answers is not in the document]
benjamindoron: This is unlikely, and was present before, but I can fix it here. Both BuildGuidHob and AllocatePool have similar semantics, so the fix remains the same.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
niruiyu, line 265: Since the function behavior has changed to not create HOB, the function comments of MmIplAllocateMmramPage should be updated to not state that NewBlock is a pointer to a new mmram block HOB.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
```diff
@@ -1063,6 +1063,12 @@ CreateMmFoundationHobList (
   MmIplCopyGuidHob (FoundationHobList + UsedSize, &HobLength, &gMpInformation2HobGuid, TRUE);
   UsedSize += HobLength;
 
+  if (HobLength == 0) {
+    HobLength = GetRemainingHobSize (*FoundationHobSize, UsedSize);
+    MmIplBuildMpInformationHob (FoundationHobList + UsedSize, &HobLength);
+    UsedSize += HobLength;
+  }
+
```
niruiyu, line 1078: Can you explain why the MpInformation HOB is not available in the HOB list?

Even when the bootloader is coreboot, the FSP that's launched by coreboot should produce the MpInformation HOB.

niruiyu: Also, if the reason makes sense, can you add comments to explain why "build" is needed and "copy" gets no such hob?

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
```diff
@@ -624,6 +624,23 @@ MmIplBuildCommBufferHob (
...
+**/
+VOID
+MmIplBuildMpInformationHob (
+  IN UINT8      *HobBuffer,
+  IN OUT UINTN  *HobBufferSize
+  )
+{
+  //
+  // DEPEX guarantees that the HOB was already produced.
+  //
```
niruiyu, line 641: Also please explicitly explain that this function should NEVER be called by MmIpl PEIM.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
```diff
@@ -371,6 +372,11 @@ ExecuteMmCoreFromMmram (
   ImageContext.ImageAddress += ImageContext.SectionAlignment - 1;
   ImageContext.ImageAddress &= ~((EFI_PHYSICAL_ADDRESS)ImageContext.SectionAlignment - 1);
 
+  //
+  // Call platform hook now that image info is known
+  //
+  PlatformHookBeforeMmLoad (&ImageContext);
+
```
niruiyu, line 379: I prefer we not add the hooks.

If SMRAM is already locked when MmIplDxe runs, how is it possible to load the MmCore into the SMRAM?
I do not understand the whole flow that coreboot tries to enable. Can you explain in detail?

StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
```diff
@@ -311,7 +311,10 @@ MmIplCopyGuidHob (
 
   UsedSize = 0;
   GuidHob  = GetFirstGuidHob (Guid);
-  ASSERT (GuidHob != NULL);
+
+  if (GuidHob == NULL) {
+    DEBUG ((DEBUG_WARN, "GUIDed HOB %g could not be found\n", Guid));
+  }
```
niruiyu, line 317: We should not just replace the assertion with a debug log.
Some HOBs should exist, and the assertion is to catch the missing HOBs.
If you feel some HOBs are optional, you can let MmIplCopyGuidHob return something, and the caller can choose to print a debug log or assert.

niruiyu (changes requested):
I've reviewed all code. Big concern is the new MmIplPlatformHookLib. I prefer not to add it.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
benjamindoron: Yes, this is true. I've fixed this locally, and the change will be in the next push.

Another thing: I've realised that this will misbehave if `MmIplAllocateMmramPage` is called twice: it will leak memory, which is tolerable, but worse, since we're no longer scrubbing the old HOB, the second call will start with the original SMRAM ranges and not know that something else allocated some, leading to memory corruption. Do we need to support this, and even if we don't, do we want to, just in case some library that's linked in makes a call here? We could stash `NewDescriptorBlock` to a global variable, and if it's set, the second allocation will start with these ranges instead of the HOB.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
benjamindoron: I suspect that the FSP will only produce the MpInformation HOB if its MP services are enabled (i.e., that `SkipMpInit` is not set), although I only skimmed the way the FSP integrates CpuMpPei once, and it may have changed in more recent platforms (since I'm aware that now, coreboot always produces an MP PPI, even when setting `SkipMpInit`). Regardless, this would only mean that coreboot has this HOB. UefiPayload wouldn't have it, since - unlike Slim Bootloader, which passes HOBs - coreboot passes its own table format/a Universal Payload FDT to the payload. Its the UefiPayloadEntry/*UniversalPayloadEntry module's job to create HOBs out of this. As far as I know, coreboot does not tell the payload how many APs there are, it's assumed that it can discover this easily through `cpuid`.

I'll add a comment.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
benjamindoron: Done

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
benjamindoron: Briefly, it's a 'memcpy and function call' design. More specifically, our implementation of `PlatformHookBeforeMmLoad` makes use of PeCoffLib's `DestinationAddress` feature, loading the MM core into a temporary buffer in DXE, but relocating it as usual to match its final destination in MMRAM.

Here is our implementation of [PlatformHookCallMmCore()](https://github.com/benjamindoron/edk2/blob/cae184b/UefiPayloadPkg/Library/MmIplPlatformHookLibPayloadMm/PayloadMmInit.c#L211). Essentially, we prepare a struct that describes what to copy, where to copy it, and where to call with which arguments to pass. We then end up [in this SMI handler](https://review.coreboot.org/c/coreboot/+/79870/6/src/drivers/payload_mm_interface/smi.c#44) I've written for coreboot.

coreboot is calling the standalone MM core's PE entrypoint. Since the ABI and expected CPU state expected by each environment may differ (32-bit vs 64-bit, whether SIMD spec extensions are enabled or not, etc), we swap the library that provides the entrypoint out [for our own](https://github.com/benjamindoron/edk2/blob/0b5d5f7/UefiPayloadPkg/Library/MmCorePayloadMmEntryPoint/X64/TransferExec.nasm#L58), and finally, we're in EDK2's standalone MM.

I've written some documentation for coreboot that can be found [here](https://review.coreboot.org/c/coreboot/+/79870/6/Documentation/drivers/payload_mm_interface/index.md). If you have any questions, or ideas for where in EDK2 we can put it once when ready, let me know!

> I prefer we not add the hooks.

I hope this explanation helps explain things and helps us get this in? The alternative solution is something I was using before: leaving SMRAM closed, but unlocked, and opening SMRAM up again in order to do IPL's regular function call into the core. We abandoned this because we didn't like it though, it's silicon-dependent whether it can work.

benjamindoron:
> I've reviewed all code. Big concern is the new MmIplPlatformHookLib. I prefer not to add it.

Thanks! I've described the flow we're enabling in more detail above, and we can continue there if there's more I can explain and we can discuss.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
benjamindoron: The problem is that it's implementation-dependent whether some HOBs added around here are required or not.
- `gMmCommBufferHobGuid`: Universally required, it's an aspect of the standalone MM *core* design.
- `gSmmBaseHobGuid`: Strictly required for what I've begun calling "PI SMM," but not present with payload MM. I don't think it makes sense to begin building it in coreboot's setup either: we're able to largely use standalone MM as a software environment (and not deal with CPU SMBASEs), since the true hardware initialisation is handled by coreboot. Therefore, we only need the HOB used by MM Access.
- `gMpInformation2HobGuid`: Basically, I think this one is recommended for a module that looks like PiSmmCpuDxeSmm (and our BlSmmCpuPayloadMm does look like it). I don't know if Arm provides it though. The DXE instance of the IPL will now build it, if necessary. This is a good example where a return value is useful, but `HobBufferSize` as output already tells this.
- `gEfiAcpiVariableGuid`: Again, only required by a PI SMM implementation, where (from memory), this is used to assist S3Resume2 in calling to PiSmmCpuDxeSmm. We don't need it in coreboot, since it will handle the process of locking down SMM and everything else that PiSmmCpuDxeSmm does on S3 resume by itself.

How about we drop the debug message too, and then:
- When a HOB isn't found, `HobBufferSize` is 0 on output.
- When it's 0, the caller will call some new function in MmPlatformHobProducerLib, `MmPlatformIsHobNeeded(EFI_GUID *Guid)` that can decide if this an error, or missing by design.

I'm still not sure this is the best design, so I'd welcome ideas.

benjamindoron:
@niruiyu What do you think of the design I described above? Will we be able to upstream this, or something similar, to allow coreboot-based platforms to implement authenticated variables securely?

benjamindoron:
@niruiyu Could you read what I wrote at https://github.com/tianocore/edk2/pull/11370#discussion_r2298836145 and let me know what you think about the design? We want this for a better Secure Boot in our customer's next coreboot project, so we want to finalise the design and get this into coreboot and EDK2. We're going to be merging the coreboot part soon

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
niruiyu: We can update MmIplAllocateMmramPage() to following
```c
EFI_PHYSICAL_ADDRESS
MmIplAllocateMmramPage (
  IN  UINTN                             Pages,
  IN  UINTN                             Capability,
  IN OUT UINTN                          *Count,
  IN OUT EFI_MMRAM_DESCRIPTOR           *Descriptors
  )
```

The function asserts when Descriptors needs to split but there is no free descriptor that can be used: `ASSERT (*Count < Capability);`

The caller is responsible for allocating an array of EFI_MMRAM_DESCRIPTOR (could be in the stack using local variables) and copies the GUIDed HOB content to the local array.

This is like your global-variable method but I think it's better to avoid global variables in this case.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplDxe.c
```diff
@@ -0,0 +1,371 @@
...
+  )
+{
+  EFI_MM_COMMUNICATION_PROTOCOL  *MmCommunication;
+  EFI_STATUS                     Status;
+
+  Status = gBS->LocateProtocol (&gEfiMmCommunicationProtocolGuid, NULL, (VOID **)&MmCommunication);
+  ASSERT_EFI_ERROR (Status);
+
+  return MmCommunication->Communicate (MmCommunication, CommBuffer, CommSize);
+}
```
niruiyu, line 50: Is it possible to share the Communicate##() function in StandaloneMmIplPei.c?

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
niruiyu: So the PlatformHookBeforeMmLoad () allocates a normal under-4G memory and updates ImageAddress field so that MmIplDxe loads/relocates MmCore to the normal under-4G memory.
Then PlatformHookCallMmCore () triggers the software SMI and passes the parameters through general purpose registers. SMI handler is implemented in coreboot side, which loads the MmCore from normal memory to SMRAM.

niruiyu: I still prefer not to change the IPL common logic to support the tricky loading process.

Can you adjust the MMRAM descriptor HOB to report a normal range so that Ipl loads the MMCore to a normal memory as what you expect? The original HOB can be saved to another GUIDed HOB (such as gOriginalSmRamRegionHob).

Then the MmCore entrypoint where you implement a special entrypoint lib code can trigger the SMI.
The entrypoint can adjust the HOB list to replace the MmRAM descriptor HOB with the one saved in gOriginalSmRamRegionHob.

Through this, all the customization is done outside of IPL and MmCore supports the self-loading through a customized entrypoint lib.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplDxe.c
benjamindoron: It uses the MM Control PPI, so we'd need some additional function either way. It would be a kind of 'do it' function that uses the MM Control PPI/protocol to trigger the SMI in hardware. I thought that continuing the pattern that the PEI IPL had was simpler: there's an IPL module, and DXE has a communicate module too.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
benjamindoron: I don't understand what you said is tricky about it: I thought it was a pretty elegant use of the `DestinationAddress` field in the ImageContext that has PeCoffLib perform the relocation we want. It keeps everything neatly encapsulated: the IPL loads, and the core takes control (from the IPL, from coreboot) and begins to execute. I thought that Arm's standalone MM implementation might start using this in time too.

With your approach, we'd have to leak memory (the MM core located outside of MMRAM). We could use MmPlatformHobProducerLib, drop some event on the system, and run a callback to reclaim the memory, but this is more effort. And it means more HOB list manipulations.

I believe that this is what standalone MM should be for: supporting MM environments that aren't just UEFI's PI SMM implementation for x86, and the modifications both aren't invasive to the general flow of the IPL, and also support what other platforms, like Arm's Trusted Firmware + standalone MM, would most likely need for the IPL to load their MM.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
benjamindoron: Yeah, I think that's better and I can write that when I'm next working on MM.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplDxe.c
```diff
@@ -0,0 +1,371 @@
...
+**/
+
+#include "StandaloneMmIplPei.h"
+#include <Library/UefiBootServicesTableLib.h>
+#include <Protocol/FirmwareVolume2.h>
+#include <Protocol/FirmwareVolumeBlock.h>
+#include <Protocol/MmAccess.h>
+#include <Protocol/MmCommunication.h>
+#include <Protocol/MpService.h>
+#include <Guid/MpInformation2.h>
```
niruiyu, line 16: can we get rid of the global variable?

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplDxe.c
niruiyu: I misunderstood the logic. It's an internal function. ok to me now.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
niruiyu: But the hook APIs are needed in your approach. That's an incompatible change.

memory leak is not a problem as the memory is a BS type memory and will be reclaimed once OS takes control. Also this leak is not from a loop that may run multi-times.
Do you treat the BS memory usages from DXE drivers are leaks?

niruiyu:
Can you rebase your PR on the latest master branch?

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplDxe.c
benjamindoron: I stash it to free the buffer allocated by `Fv->ReadSection` once we're done, as a comment lower down says. It's an unneeded copy: PeCoffLib copies from here (which could be an in-memory FV, or possibly a SPI flash-based FV) to MMRAM, at which point this copy doesn't matter.

(I believe that `Fv->ReadSection` allocates memory so that the backend doesn't matter. If it's not memory mapped, the FV2 protocol would be required to communicate with some storage controller to retrieve it, but the caller just wants a buffer)

benjamindoron: `LocateMmCoreFv`, which sets this variable, is called by `ExecuteMmCoreFromMmram`, which stores the pointer to the buffer in the ImageContext describing the MM core's PE. ImageContext is just on the stack, and we don't have visibility to it externally.

I'm realising that little of this is obvious if you don't start following the variable and its users around. How about I just add a comment about what the global is for?

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplCommon.c
benjamindoron: That's true, about how it's not a real leak. I prefer being clean about my code, but you're right that the design guarantees it's all reclaimed later.

I think I understand: your concern is that I'm adding a new library dependency to the IPL?

niruiyu: yes my concern is the hook lib. In general, I think the hook lib design pattern is the very last choice as once the hook lib is born, more hook apis will be added. Also people do not understand clearly what each hook API should do. It may end up with lots of tricky code implemented in the hook APIs and hurt greatly to the initial goal to develop a common IPL.

benjamindoron: Can we make an exception for standalone MM? There's already another platform that uses it and needs to load it in a special way (Arm TrustZone's standalone MM) and it's a good environment for existing MM implementations (like TrustZone, and coreboot SMM) to wrap for implementing privileged EDK2 code. Like I said, I thought that Arm would migrate to this approach in time. I'll look into migrating Arm's implementation myself if that'll help get this merged.

Otherwise, I don't like the idea of self-loading, because I think that it makes things messy: we have to lie to MM Access so that it makes the IPL load into DXE memory. But if I can't convince you, I'll just do that. Our customer wants something upstreamed to reduce the maintenance burden. Let me know what you think of making an exception for standalone MM loading, and we'll go from there.

PR #11375, opened 2025-08-03
ShellPkg: Fix cd .. & Tab key in External shell

Nanjinmoorthy2002:
# Description

Resolved issue where 'cd ..' did not navigate backward due to the current directory not being updated properly in the external shell. Also fixed Tab key behavior in the external shell, which displayed files inside FS0:\EFI\Boot as current directory is not updated in FileInterfaceStdInRead.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Copy the Shell.efi into USB.
In shell boot the efi from Efi/boot.
In external shell, give cd .. command and tab key.

## Integration Instructions

NA

### DIFF
--- a/ShellPkg/Application/Shell/ShellParametersProtocol.c
+++ b/ShellPkg/Application/Shell/ShellParametersProtocol.c
@@ -403,29 +403,26 @@ CreatePopulateInstallShellParametersProtocol (
   //
   // Populate the 3 faked file systems...
   //
+  (*NewShellParameters)->StdIn  = &FileInterfaceStdIn;
+  (*NewShellParameters)->StdOut = &FileInterfaceStdOut;
+  (*NewShellParameters)->StdErr = &FileInterfaceStdErr;
   if (*RootShellInstance) {
-    (*NewShellParameters)->StdIn  = &FileInterfaceStdIn;
-    (*NewShellParameters)->StdOut = &FileInterfaceStdOut;
-    (*NewShellParameters)->StdErr = &FileInterfaceStdErr;
-    Status                        = gBS->InstallProtocolInterface (
-                                           &gImageHandle,
-                                           &gEfiShellParametersProtocolGuid,
-                                           EFI_NATIVE_INTERFACE,
-                                           (VOID *)(*NewShellParameters)
-                                           );
+    Status = gBS->InstallProtocolInterface (
+                    &gImageHandle,
+                    &gEfiShellParametersProtocolGuid,
+                    EFI_NATIVE_INTERFACE,
+                    (VOID *)(*NewShellParameters)
+                    );
   } else {
     //
     // copy from the existing ones
     //
-    (*NewShellParameters)->StdIn  = ShellInfoObject.OldShellParameters->StdIn;
-    (*NewShellParameters)->StdOut = ShellInfoObject.OldShellParameters->StdOut;
-    (*NewShellParameters)->StdErr = ShellInfoObject.OldShellParameters->StdErr;
-    Status                        = gBS->ReinstallProtocolInterface (
-                                           gImageHandle,
-                                           &gEfiShellParametersProtocolGuid,
-                                           (VOID *)ShellInfoObject.OldShellParameters,
-                                           (VOID *)(*NewShellParameters)
-                                           );
+    Status = gBS->ReinstallProtocolInterface (
+                    gImageHandle,
+                    &gEfiShellParametersProtocolGuid,
+                    (VOID *)ShellInfoObject.OldShellParameters,
+                    (VOID *)(*NewShellParameters)
+                    );
   }
 
   return (Status);
--- a/ShellPkg/Application/Shell/ShellProtocol.c
+++ b/ShellPkg/Application/Shell/ShellProtocol.c
@@ -3251,7 +3251,7 @@ EfiShellSetCurDir (
       return (EFI_INVALID_PARAMETER);
     }
 
-    //    gShellCurMapping = MapListItem;
+    //  gShellCurMapping = MapListItem;
     if (DirectoryName != NULL) {
       //
       // change current dir on that file system
--- a/ShellPkg/Library/UefiShellCommandLib/UefiShellCommandLib.c
+++ b/ShellPkg/Library/UefiShellCommandLib/UefiShellCommandLib.c
@@ -1533,27 +1533,28 @@ ShellCommandCreateInitialMappingsAndPaths (
     //
     if (gShellCurMapping != NULL) {
       gShellCurMapping = NULL;
-      CurDir           = gEfiShellProtocol->GetEnv (L"cwd");
-      if (CurDir != NULL) {
-        MapName = AllocateCopyPool (StrSize (CurDir), CurDir);
-        if (MapName == NULL) {
-          return EFI_OUT_OF_RESOURCES;
-        }
+    }
 
-        SplitCurDir = StrStr (MapName, L":");
-        if (SplitCurDir == NULL) {
-          SHELL_FREE_NON_NULL (MapName);
-          return EFI_UNSUPPORTED;
-        }
-
-        *(SplitCurDir + 1) = CHAR_NULL;
-        MapListItem        = ShellCommandFindMapItem (MapName);
-        if (MapListItem != NULL) {
-          gShellCurMapping = MapListItem;
-        }
-
-        SHELL_FREE_NON_NULL (MapName);
+    CurDir = gEfiShellProtocol->GetEnv (L"cwd");
+    if (CurDir != NULL) {
+      MapName = AllocateCopyPool (StrSize (CurDir), CurDir);
+      if (MapName == NULL) {
+        return EFI_OUT_OF_RESOURCES;
       }
+
+      SplitCurDir = StrStr (MapName, L":");
+      if (SplitCurDir == NULL) {
+        SHELL_FREE_NON_NULL (MapName);
+        return EFI_UNSUPPORTED;
+      }
+
+      *(SplitCurDir + 1) = CHAR_NULL;
+      MapListItem        = ShellCommandFindMapItem (MapName);
+      if (MapListItem != NULL) {
+        gShellCurMapping = MapListItem;
+      }
+
+      SHELL_FREE_NON_NULL (MapName);
     }
   } else {
     Count = (UINTN)-1;


ardbiesheuvel (changes requested):
Please split this up into 2 patches

Nanjinmoorthy2002:
@ardbiesheuvel 

Split into two separate commits. Please help to review.

ShellPkg/Application/Shell/ShellParametersProtocol.c
```diff
@@ -403,29 +403,26 @@ CreatePopulateInstallShellParametersProtocol (
...
   if (*RootShellInstance) {
-    (*NewShellParameters)->StdIn  = &FileInterfaceStdIn;
-    (*NewShellParameters)->StdOut = &FileInterfaceStdOut;
-    (*NewShellParameters)->StdErr = &FileInterfaceStdErr;
-    Status                        = gBS->InstallProtocolInterface (
-                                           &gImageHandle,
-                                           &gEfiShellParametersProtocolGuid,
-                                           EFI_NATIVE_INTERFACE,
-                                           (VOID *)(*NewShellParameters)
-                                           );
```
ardbiesheuvel, line 415: Please explain how this change is related to the symptom you are addressing, and why this change in behavior (not copying the existing ones) is guaranteed to be correct in all cases.

ShellPkg/Application/Shell/ShellParametersProtocol.c
```diff
@@ -403,29 +403,26 @@ CreatePopulateInstallShellParametersProtocol (
...
+    Status = gBS->InstallProtocolInterface (
+                    &gImageHandle,
+                    &gEfiShellParametersProtocolGuid,
+                    EFI_NATIVE_INTERFACE,
+                    (VOID *)(*NewShellParameters)
+                    );
   } else {
     //
     // copy from the existing ones
     //
```
ardbiesheuvel, line 419: Is this comment still appropriate?

ShellPkg/Application/Shell/ShellProtocol.c
```diff
@@ -3251,7 +3251,7 @@ EfiShellSetCurDir (
       return (EFI_INVALID_PARAMETER);
     }
 
-    //    gShellCurMapping = MapListItem;
+    gShellCurMapping = MapListItem;
```
ardbiesheuvel: Please explain in the commit log how this change is related to the problem you are addressing.

ShellPkg/Application/Shell/ShellParametersProtocol.c
Nanjinmoorthy2002: When the Tab key was pressed in the external shell, it only displayed files inside FS0:\EFI\Boot. This issue occurred because the old ShellParameters were being reinstalled in the external shell, preventing the current directory from being updated correctly in the FileInterfaceStdIn function. As a result, the shell always pointed to FS0:\EFI\Boot.

To resolve this, I updated both the internal and external shells to assign the NewShellParameters. After this change, the current directory is updated correctly, and the Tab key functionality works as expected.

ShellPkg/Application/Shell/ShellParametersProtocol.c
Nanjinmoorthy2002: No need to remove it.

ShellPkg/Application/Shell/ShellProtocol.c
Nanjinmoorthy2002: The gShellCurMapping variable is updated when the map -r command is executed in the shell and in the EfiShellSetCurDir function.

In the EfiShellSetCurDir function, gShellCurMapping is compared with MapListItem. Since gShellCurMapping is not updated, the environment variable is also not set correctly, and as a result, the current directory is not updated. 

To resolve this, I assigned gShellCurMapping to MapListItem in EfiShellSetCurDir. I also verified the map -r command, it is also work properly in the external shell.

ardbiesheuvel: Please explain **in the commit log** not in a GitHub conversation that will be untraceable once this PR is merged.

Nanjinmoorthy2002:
@ardbiesheuvel 
Updated the commit messages with proper description please check with it.

ShellPkg/Application/Shell/ShellProtocol.c
```diff
@@ -3251,7 +3251,7 @@ EfiShellSetCurDir (
       return (EFI_INVALID_PARAMETER);
     }
 
-    //    gShellCurMapping = MapListItem;
+    gShellCurMapping = MapListItem;
     if (DirectoryName != NULL) {
```
pierregondois, line 3255: `gShellCurMapping` is corrected only if we change the current directory.
When launching a child shell, if `cwd` is not NULL, `gShellCurMapping` should be set directly.
I think it would be more correct to update `gShellCurMapping` in `ShellCommandCreateInitialMappingsAndPaths()`

The following seems to work for me:
```
diff --git a/ShellPkg/Library/UefiShellCommandLib/UefiShellCommandLib.c b/ShellPkg/Library/UefiShellCommandLib/UefiShellCommandLib.c
index ba85efcb7fa1..5e9b833d5ffb 100644
--- a/ShellPkg/Library/UefiShellCommandLib/UefiShellCommandLib.c
+++ b/ShellPkg/Library/UefiShellCommandLib/UefiShellCommandLib.c
@@ -1533,27 +1533,28 @@ ShellCommandCreateInitialMappingsAndPaths (
     //
     if (gShellCurMapping != NULL) {
       gShellCurMapping = NULL;
-      CurDir           = gEfiShellProtocol->GetEnv (L"cwd");
-      if (CurDir != NULL) {
-        MapName = AllocateCopyPool (StrSize (CurDir), CurDir);
-        if (MapName == NULL) {
-          return EFI_OUT_OF_RESOURCES;
-        }
+    }
 
-        SplitCurDir = StrStr (MapName, L":");
-        if (SplitCurDir == NULL) {
-          SHELL_FREE_NON_NULL (MapName);
-          return EFI_UNSUPPORTED;
-        }
-
-        *(SplitCurDir + 1) = CHAR_NULL;
-        MapListItem        = ShellCommandFindMapItem (MapName);
-        if (MapListItem != NULL) {
-          gShellCurMapping = MapListItem;
-        }
+    CurDir           = gEfiShellProtocol->GetEnv (L"cwd");
+    if (CurDir != NULL) {
+      MapName = AllocateCopyPool (StrSize (CurDir), CurDir);
+      if (MapName == NULL) {
+      }
 
+      SplitCurDir = StrStr (MapName, L":");
+      if (SplitCurDir == NULL) {
         SHELL_FREE_NON_NULL (MapName);
+        return EFI_UNSUPPORTED;
       }
+
+      *(SplitCurDir + 1) = CHAR_NULL;
+      MapListItem        = ShellCommandFindMapItem (MapName);
+      if (MapListItem != NULL) {
+        gShellCurMapping = MapListItem;
+      }
+
+      SHELL_FREE_NON_NULL (MapName);
     }
   } else {
     Count = (UINTN)-1;
```

NIT:
Also for the commit messages, I think it is more conventional to have impersonal sentences:
`I assigned gShellCurMapping to MapListItem in EfiShellSetCurDir.` -> `Assign gShellCurMapping to MapListItem ...`

ShellPkg/Application/Shell/ShellParametersProtocol.c
pierregondois: Maybe:
```
Upon launching a child Shell, StdIn/Out/Err file handles are copied from the parent Shell. These file handles rely on the parent Shell environment. The Current Working Directory of the parent shell is not updated when navigating in the child Shell. This results in having a tab completion consistently iterating over the files in the Current Working Directory of the parent Shell.

Always update StdIn/Out/Err file handles when launching a child Shell.
```

Also I could not trigger any meaningful bug so far, but there might be a risk of not inheriting from the parent Shell redirection. For instance if we do `Shell.efi < input.txt` or `Shell.efi > output.txt`, the child shell should be launched with a modified StdIn or StdOut.

ShellPkg/Application/Shell/ShellProtocol.c
Nanjinmoorthy2002: With your changes issue is fixed.

ShellPkg/Application/Shell/ShellParametersProtocol.c
Nanjinmoorthy2002: I tried running the commands you shared `Shell.efi < input.txt or Shell.efi > output.txt`, but I’m facing a hang issue in the shell. I also tried using the default source, but the issue persists. Could you please explain how to use this command in the shell to verify the changes?

Additionally, could you please explain why this command is being suggested to verify the impact of the changes?

ShellPkg/Application/Shell/ShellProtocol.c
```diff
@@ -3251,7 +3251,7 @@ EfiShellSetCurDir (
       return (EFI_INVALID_PARAMETER);
     }
 
-    //    gShellCurMapping = MapListItem;
+    //  gShellCurMapping = MapListItem;
```
pierregondois, line 3254: Should it be removed ?

ShellPkg/Application/Shell/ShellProtocol.c
pierregondois: Would it be possible to re-write the commit message as the patch changed:
- in a more impersonal way (i.e. with using `I`)
- by describing exactly how to reproduce the issue, what is happening and how it is fixed

For instance:
```
    When cd.. command given in external shell
    it did not navigate backward
    because of current directory not updating.
```
I don't think the issue is triggered if we launch the external shell from the `root` directory (i.e. when we are not in any filesystem and the Shell prompt is `Shell>`). So it would be nice to describe exactly what needs to be done.

ShellPkg/Application/Shell/ShellParametersProtocol.c
pierregondois: Since StdIn/Out/Err are not inherited anymore from the parent Shell but just replaced by the default implementation, I was wondering if this would trigger any issue if the external Shell is supposed to inherit from the parent StdIn/Out/Err implementation, i.e. if we run `Shell.efi < input.txt` or `Shell.efi > out.txt`.

It also hangs for me in some cases, but I assume this is normal if `input.txt` is empty for instance. I will spend a bit more time tomorrow to check.

PR #11379, opened 2025-08-04
ArmPkg/ArmMmuLib: Add support for LPA2

sarah-walker-arm:
# Description

Add support for LPA2 (52-bit physical addressing).

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Booted to UEFI shell on FVP Versatile Express model with and without LPA2 enabled.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/CpuDxe/AArch64/Mmu.c
+++ b/ArmPkg/Drivers/CpuDxe/AArch64/Mmu.c
@@ -2,7 +2,7 @@
 
 Copyright (c) 2009, Hewlett-Packard Company. All rights reserved.<BR>
 Portions copyright (c) 2010, Apple Inc. All rights reserved.<BR>
-Portions copyright (c) 2011-2021, Arm Limited. All rights reserved.<BR>
+Portions copyright (c) 2011-2025, Arm Limited. All rights reserved.<BR>
 Copyright (c) 2017, Intel Corporation. All rights reserved.<BR>
 
 SPDX-License-Identifier: BSD-2-Clause-Patent
@@ -30,12 +30,12 @@ STATIC
 VOID
 GetRootTranslationTableInfo (
   IN  UINTN  T0SZ,
-  OUT UINTN  *RootTableLevel,
+  OUT  INTN  *RootTableLevel,
   OUT UINTN  *RootTableEntryCount
   )
 {
-  *RootTableLevel      = (T0SZ - MIN_T0SZ) / BITS_PER_LEVEL;
-  *RootTableEntryCount = TT_ENTRY_COUNT >> (T0SZ - MIN_T0SZ) % BITS_PER_LEVEL;
+  *RootTableLevel      = (INTN)(T0SZ - MIN_T0SZ) / BITS_PER_LEVEL;
+  *RootTableEntryCount = TT_ENTRY_COUNT >> (INTN)(T0SZ - MIN_T0SZ) % BITS_PER_LEVEL;
 }
 
 /**
@@ -167,7 +167,7 @@ UINT64
 GetNextEntryAttribute (
   IN     UINT64  *TableAddress,
   IN     UINTN   EntryCount,
-  IN     UINTN   TableLevel,
+  IN     INTN    TableLevel,
   IN     UINT64  BaseAddress,
   IN OUT UINT64  *PrevEntryAttribute,
   IN OUT UINT64  *StartGcdRegion
@@ -273,7 +273,7 @@ SyncCacheConfig (
   EFI_STATUS                       Status;
   UINT64                           PageAttribute;
   UINT64                           *FirstLevelTableAddress;
-  UINTN                            TableLevel;
+  INTN                             TableLevel;
   UINTN                            TableCount;
   UINTN                            NumberOfDescriptors;
   EFI_GCD_MEMORY_SPACE_DESCRIPTOR  *MemorySpaceMap;
@@ -419,7 +419,7 @@ EfiAttributeToArmAttribute (
 EFI_STATUS
 GetMemoryRegionRec (
   IN     UINT64  *TranslationTable,
-  IN     UINTN   TableLevel,
+  IN     INTN    TableLevel,
   IN     UINT64  *LastBlockEntry,
   IN OUT UINTN   *BaseAddress,
   OUT    UINTN   *RegionLength,
@@ -520,7 +520,7 @@ GetMemoryRegion (
 {
   EFI_STATUS  Status;
   UINT64      *TranslationTable;
-  UINTN       TableLevel;
+  INTN        TableLevel;
   UINTN       EntryCount;
   UINTN       T0SZ;
 
--- a/ArmPkg/Drivers/CpuDxe/AArch64/Mmu.c
+++ b/ArmPkg/Drivers/CpuDxe/AArch64/Mmu.c
@@ -18,6 +18,20 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #define MIN_T0SZ        16
 #define BITS_PER_LEVEL  9
 
+STATIC
+UINT64
+GetOutputAddress (
+  IN  UINT64   Entry,
+  IN  BOOLEAN  Lpa2Enabled
+  )
+{
+  if (Lpa2Enabled) {
+    return (Entry & TT_ADDRESS_MASK_DESCRIPTION_TABLE_LPA2) | ((Entry & TT_UPPER_ADDRESS_MASK) << (50 - 8));
+  } else {
+    return Entry & TT_ADDRESS_MASK_DESCRIPTION_TABLE;
+  }
+}
+
 /**
   Parses T0SZ to determine the level and number of entries at the root
   of the translation table.
@@ -126,8 +140,9 @@ RegionAttributeToGcdAttribute (
 STATIC
 UINT64
 GetFirstPageAttribute (
-  IN UINT64  *FirstLevelTableAddress,
-  IN UINTN   TableLevel
+  IN UINT64   *FirstLevelTableAddress,
+  IN UINTN    TableLevel,
+  IN BOOLEAN  Lpa2Enabled
   )
 {
   UINT64  FirstEntry;
@@ -139,7 +154,7 @@ GetFirstPageAttribute (
     // Only valid for Levels 0, 1 and 2
 
     // Get the attribute of the subsequent table
-    return GetFirstPageAttribute ((UINT64 *)(FirstEntry & TT_ADDRESS_MASK_DESCRIPTION_TABLE), TableLevel + 1);
+    return GetFirstPageAttribute ((UINT64 *)GetOutputAddress (FirstEntry, Lpa2Enabled), TableLevel + 1, Lpa2Enabled);
   } else if (((FirstEntry & TT_TYPE_MASK) == TT_TYPE_BLOCK_ENTRY) ||
              ((TableLevel == 3) && ((FirstEntry & TT_TYPE_MASK) == TT_TYPE_BLOCK_ENTRY_LEVEL3)))
   {
@@ -226,7 +241,7 @@ GetNextEntryAttribute (
 
       // Increase the level number and scan the sub-level table
       GetNextEntryAttribute (
-        (UINT64 *)(Entry & TT_ADDRESS_MASK_DESCRIPTION_TABLE),
+        (UINT64 *)GetOutputAddress (Entry, ArmGetTCR () & TCR_DS),
         TT_ENTRY_COUNT,
         TableLevel + 1,
         (BaseAddress + (Index * TT_ADDRESS_AT_LEVEL (TableLevel))),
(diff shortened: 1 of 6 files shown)


LeviYeoReum (approved):

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -22,6 +22,12 @@
 #include <Library/HobLib.h>
 #include "ArmMmuLibInternal.h"
 
+UINTN
```
leiflindholm: This function is intentionally not exported via ArmLib.h.

Please replace this (and its callers) with new exported functions in ArmLib checking for specific feature support. The MMFR macros from 1/3 can then move to MdePkg/Include/AArch64/AArch64.h where other similar things are also defined..

(continued below)

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -486,6 +525,27 @@ GcdAttributeToPageAttribute (
...
+  UINTN   MaxAddressBits;
+  UINT64  MaxAddress;
+  UINT64  MMFR0;
+
+  MaxAddressBits = MIN (ArmGetPhysicalAddressBits (), MAX_VA_BITS);
+  MaxAddress     = LShiftU64 (1ULL, MaxAddressBits) - 1;
+
+  MMFR0 = ArmReadIdAA64Mmfr0 ();
+
+  if ((MaxAddress > SIZE_256TB) && ((MMFR0 & MMFR0_TGRAN4_MASK) == MMFR0_TGRAN4_52BITS)) {
```
leiflindholm: (continued re MMFR0)
i.e. instead of doing this, do
`if ((MaxAddress > SIZE_256TB) && ArmHas52BitTgran4 ())`

And implementing ArmHas52BitTgran4 () in ArmLib.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -222,8 +238,10 @@ UpdateRegionMappingRecursive (
   IN  UINT64   AttributeSetMask,
   IN  UINT64   AttributeClearMask,
   IN  UINT64   *PageTable,
-  IN  UINTN    Level,
-  IN  BOOLEAN  TableIsLive
+  IN  INTN     Level,
+  IN  BOOLEAN  IsRootTable,
+  IN  BOOLEAN  TableIsLive,
+  IN  BOOLEAN  Is52Bit
```
leiflindholm: Can we make this a UINTN VaBits instead?

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -91,9 +97,19 @@ ArmMemoryAttributeToPageAttribute (
...
 
+// T0SZ can be below MIN_T0SZ when LPA2 is in use, meaning the page table starts at level -1
 #define MIN_T0SZ        16
 #define BITS_PER_LEVEL  9
-#define MAX_VA_BITS     48
+#define MAX_VA_BITS     52
+
+STATIC
+UINT64
+Get52BitPageTableAddress (
```
leiflindholm: Can we make this a generic GetPageTableAddress that also takes a VaBits (see below) option, returning the unmodified input if not 52?

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -363,7 +385,12 @@ UpdateRegionMappingRecursive (
       }
 
       if (!IsTableEntry (*Entry, Level)) {
-        EntryValue = (UINTN)TranslationTable | TT_TYPE_TABLE_ENTRY;
+        if (Is52Bit) {
```
leiflindholm: Then we can fold out this test, always calling GetPageTableAddress...

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -373,10 +400,17 @@ UpdateRegionMappingRecursive (
           );
       }
     } else {
-      EntryValue  = (*Entry & AttributeClearMask) | AttributeSetMask;
-      EntryValue |= RegionStart;
-      EntryValue |= (Level == 3) ? TT_TYPE_BLOCK_ENTRY_LEVEL3
-                                 : TT_TYPE_BLOCK_ENTRY;
+      if (Is52Bit) {
+        EntryValue  = ((*Entry & AttributeClearMask) | AttributeSetMask) & ~(3 << 8);
```
leiflindholm: Can we have a macro or two for that live-coded bitmask?

Then, if using a generic GetPageTableAddress below, we can drop the if statement and add a line here something like
`EntryValue &= (VaBits == 52) ? <macro> : MAX_UINTN;`

leiflindholm (changes requested):

MdePkg/Include/AArch64/AArch64Mmu.h
```diff
@@ -186,11 +194,16 @@
 #define TCR_PASZ_42BITS_4TB    (0x3UL)
 #define TCR_PASZ_44BITS_16TB   (0x4UL)
 #define TCR_PASZ_48BITS_256TB  (0x5UL)
+#define TCR_PASZ_52BITS_4PB    (0x6UL)
 
 // The value written to the T*SZ fields are defined as 2^(64-T*SZ). So a 39Bit
 // Virtual address range for 512GB of virtual space sets T*SZ to 25
 #define INPUT_ADDRESS_SIZE_TO_TXSZ(a)  (64 - a)
 
+#define MMFR0_TGRAN4_52BITS  (0x1UL << 28)
```
leiflindholm: Could we move these to sit with the other ID register definitions in MdePkg/Include/AArch64/AArch64.h ?

leiflindholm: Other than that, I'm very happy with this set.

I _would_ prefer for Ard to give his feedback as well, and he's out for another couple of weeks I think. But then we're in hard freeze for stable tag now anyway.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -91,9 +91,24 @@ ArmMemoryAttributeToPageAttribute (
...
 #define MIN_T0SZ        16
 #define BITS_PER_LEVEL  9
-#define MAX_VA_BITS     48
+#define MAX_VA_BITS     52
+
+STATIC
+UINT64
+GetPageTableAddress (
+  IN  UINTN  Address,
+  IN  UINTN  VaBits
```
ardbiesheuvel: The only purpose of `VaBits` appears to be comparing it to `52`, in which case LPA2 is assumed to be enabled.

It is conceivable, though, that the MMU might be configured for a smaller VA space, while still enabling LPA2 via `TCR.DS`, and it is the latter setting that decides how the page tables are interpreted. (Note that the UEFI spec only requires system memory to be mapped 1:1 but MMIO regions that are outside of the 48-bit addressable region are permitted to be remapped to a lower address)

So instead, please use a `BOOL Lpa2Enabled` and base its value on `TCR.DS`

ArmPkg/Drivers/CpuDxe/AArch64/Mmu.c
```diff
@@ -30,12 +30,12 @@ STATIC
 VOID
 GetRootTranslationTableInfo (
   IN  UINTN  T0SZ,
-  OUT UINTN  *RootTableLevel,
+  OUT  INTN  *RootTableLevel,
```
ardbiesheuvel, line 47: Please fix the indentation here.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -91,9 +91,24 @@ ArmMemoryAttributeToPageAttribute (
...
 
+// T0SZ can be below MIN_T0SZ when LPA2 is in use, meaning the page table starts at level -1
 #define MIN_T0SZ        16
 #define BITS_PER_LEVEL  9
-#define MAX_VA_BITS     48
+#define MAX_VA_BITS     52
+
+STATIC
+UINT64
+GetPageTableAddress (
```
ardbiesheuvel: Please call this `EncodeOutputAddress` to clarify that the address in question is not the address of a page table, but the output address inserted into a page table descriptor.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -91,9 +91,24 @@ ArmMemoryAttributeToPageAttribute (
...
+  IN  UINTN  Address,
+  IN  UINTN  VaBits
+  )
+{
+  if (VaBits == 52) {
+    return ((UINTN)Address & 0x0003fffffffff000) | (((UINTN)Address >> 50) << 8);
```
ardbiesheuvel: This masks off the low 12 bits for LPA2 but returns the entire value for !LPA2. Please make this consistent.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -373,8 +395,9 @@ UpdateRegionMappingRecursive (
           );
       }
     } else {
-      EntryValue  = (*Entry & AttributeClearMask) | AttributeSetMask;
-      EntryValue |= RegionStart;
+      EntryValue  = ((*Entry & AttributeClearMask) | AttributeSetMask);
```
ardbiesheuvel: Are the extra parentheses really needed?

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -373,8 +395,9 @@ UpdateRegionMappingRecursive (
           );
       }
     } else {
-      EntryValue  = (*Entry & AttributeClearMask) | AttributeSetMask;
-      EntryValue |= RegionStart;
+      EntryValue  = ((*Entry & AttributeClearMask) | AttributeSetMask);
+      EntryValue &= (VaBits == 52) ? (~TT_UPPER_ADDRESS_MASK) : MAX_UINT64;
```
ardbiesheuvel: This just clears any shareability bits that happen to be set as a result of applying the masks, right? I don't mind that for now, but please add a comment here that highlights that. In the future, we may have to tighten the API and fail a call when it attempts to program a shareability setting that is not supported under LPA2.

Also, could we move the masking into `EncodeOutputAddress`?

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -654,13 +707,16 @@ ArmConfigureMmu (
       TCR |= TCR_IPS_16TB;
     } else if (MaxAddress < SIZE_256TB) {
       TCR |= TCR_IPS_256TB;
+    } else if ((MaxAddress < SIZE_4PB) && ArmHas52BitTgran4 ()) {
+      TCR   |= TCR_IPS_4PB | TCR_DS;
```
ardbiesheuvel: Setting `TCR.IPS` to `0b110` affects the way `TTBR0_ELx` is interpreted, regardless of whether or not `TCR.DS` is set. So setting both is fine, but the code handling the assignment of `TTBR0` needs to be updated too.

ArmPkg/Drivers/CpuDxe/AArch64/Mmu.c
sarah-walker-arm: This is the indentation uncrustify produces.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -730,7 +790,15 @@ ArmConfigureMmu (
     MAIR_ATTR (TT_ATTR_INDX_MEMORY_WRITE_BACK, MAIR_ATTR_NORMAL_MEMORY_WRITE_BACK)
     );
 
-  ArmSetTTBR0 (TranslationTable);
+  if (TCR & TCR_DS) {
```
ardbiesheuvel: This should depend on IPS not DS.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -583,6 +634,7 @@ ArmConfigureMmu (
   UINTN       T0SZ;
   UINTN       RootTableEntryCount;
   UINT64      TCR;
+  UINTN       VaBits;
```
ardbiesheuvel: This can be dropped now, right?

ardbiesheuvel (approved):

ardbiesheuvel:
Looking all good now, thanks!

PR #11386, opened 2025-08-05
BaseTools: Align Pre-Processor Macros for CLANGPDB and CLANGDWARF

nate-desimone:
# Description

Update the CLANGPDB toolchain to define the __GNUC__, __GNUC_MINOR__, __GNUC_PATCHLEVEL__, and __MINGW32__ macros with values that match the pre-defined values seen when using the x86_64-unknown-windows-gnu target triple. This minimizes the differences in pre-processor macros between the CLANGPDB and CLANGDWARF targets. This fixes compilation issues with CryptoPkg/Library/OpensslLib/OpensslGen.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- Confirmed that CryptoPkg/Library/OpensslLib/OpensslGen is now able to build.
- Confirmed that all edk2 packages that support building with CLANGPDB continue to build.
- Confirmed that the Intel platforms in edk2-platforms that support building with CLANGPDB continue to build.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Conf/tools_def.template
+++ b/BaseTools/Conf/tools_def.template
@@ -1933,7 +1933,7 @@ DEFINE CLANGPDB_IA32_TARGET          = -target i686-pc-windows-msvc
 DEFINE CLANGPDB_X64_TARGET           = -target x86_64-pc-windows-msvc
 
 DEFINE CLANGPDB_WARNING_OVERRIDES    = -Wno-parentheses-equality -Wno-tautological-compare -Wno-tautological-constant-out-of-range-compare -Wno-empty-body -Wno-unused-const-variable -Wno-varargs -Wno-unknown-warning-option -Wno-unused-but-set-variable -Wno-unused-const-variable -Wno-unaligned-access -Wno-microsoft-enum-forward-reference
-DEFINE CLANGPDB_ALL_CC_FLAGS         = DEF(GCC48_ALL_CC_FLAGS) DEF(CLANGPDB_WARNING_OVERRIDES) -fno-stack-protector -funsigned-char -ftrap-function=undefined_behavior_has_been_optimized_away_by_clang -Wno-address -Wno-shift-negative-value -Wno-unknown-pragmas -Wno-incompatible-library-redeclaration -Wno-null-dereference -mno-implicit-float -mms-bitfields -mno-stack-arg-probe -fno-omit-frame-pointer -U _MSC_VER -D __GNUC__
+DEFINE CLANGPDB_ALL_CC_FLAGS         = DEF(GCC48_ALL_CC_FLAGS) DEF(CLANGPDB_WARNING_OVERRIDES) -fno-stack-protector -funsigned-char -ftrap-function=undefined_behavior_has_been_optimized_away_by_clang -Wno-address -Wno-shift-negative-value -Wno-unknown-pragmas -Wno-incompatible-library-redeclaration -Wno-null-dereference -mno-implicit-float -mms-bitfields -mno-stack-arg-probe -fno-omit-frame-pointer -U _MSC_VER -D __GNUC__=4 -D __GNUC_MINOR__=2 -D __GNUC_PATCHLEVEL__=1 -D __MINGW32__=1
 
 ###########################
 # CLANGPDB IA32 definitions


bexcran:
I don't understand why this is needed, since the GNUC macros should already be defined by clang. And the MINGW32 macro shouldn't be globally defined like that, but only for Windows?

On my Ubuntu 25.04 system with clang 20.1.2, the  code:
```c
#include <stdio.h>

int main(int argc, char **argv)
{
    (void)argc;
    (void)argv;
    printf("__GNUC__\t\t%d\n", __GNUC__);
    printf("__GNUC_MINOR__\t\t%d\n", __GNUC_MINOR__);
    printf("__GNUC_PATCHLEVEL__\t%d\n", __GNUC_PATCHLEVEL__);
    return 0;
}
```
Outputs:
```c
__GNUC__		4
__GNUC_MINOR__		2
__GNUC_PATCHLEVEL__	1
```

nate-desimone:
> I don't understand why this is needed, since the GNUC macros should already be defined by clang. And the MINGW32 macro shouldn't be globally defined like that, but only for Windows?

Hi Rebecca, try compiling again with `-target x86_64-pc-windows-msvc` added to the clang command line. You will see that those macros go away, hence why I added them back so that edk2 code can mostly treat CLANGPDB and CLANGDWARF the same (at least from a pre-processor macro standpoint.)

bexcran:
> Hi Rebecca, try compiling again with `-target x86_64-pc-windows-msvc` added to the clang command line. You will see that those macros go away, hence why I added them back so that edk2 code can mostly treat CLANGPDB and CLANGDWARF the same (at least from a pre-processor macro standpoint.)

Ah, I see this is when using the version of Clang from Visual Studio. In that case it seems clang tries to be compatible with MSVC instead of GNU C hence why `__GNUC__` goes away (https://blog.conan.io/2022/10/13/Different-flavors-Clang-compiler-Windows.html).

Does Clang in MSCV compatibility mode really implement the GNU C extensions implied by `__GNUC__`? Also, would it be possible to limit this change to the case of ClangCL only?

nate-desimone:
Hi @bexcran, per your questions:

> Ah, I see this is when using the version of Clang from Visual Studio.

Actually no. This behavior can be seen with any copy of Clang. Including the vanilla llvm.org binaries. Unlike GCC, Clang supports dynamic cross-compilation. With GCC, if one wishes to change the instruction set, ABI, or executable format, one must re-compile GCC itself; resulting in a different “cross-compiler” that supports a different instruction set, ABI, binary format, etc. That is not the case with LLVM/Clang. All supported instruction sets, ABIs, and binary formats are included. Instead of building a cross-compiler, one simply changes the LLVM target triple.

> In that case it seems clang tries to be compatible with MSVC instead of GNU C hence why `__GNUC__` goes away (https://blog.conan.io/2022/10/13/Different-flavors-Clang-compiler-Windows.html).

Correct. Clang will re-configure its pre-defined pre-processor macros to make it appear as if you are using MSVC when you give the command line argument `-target x86_64-pc-windows-msvc` instead of `-target x86_64-unknown-windows-gnu`.

> Does Clang in MSCV compatibility mode really implement the GNU C extensions implied by `__GNUC__`?

Yes, it does. Moreover, it doesn't support the MSVC style `#pragma` statements either.

> Also, would it be possible to limit this change to the case of ClangCL only?

This is not just applicable to ClangCL, it is applicable to vanilla clang. I am exclusively using vanilla clang on Linux (or Clang + Mingw-w64 on Windows) to run builds. To use the exact verbiage from that blog post you referenced, I'm using "Msys2 Clang" but I see identical behavior in "LLVM/ Clang".

lgao4 (approved):

leiflindholm:
While I understand the reasoning behind this change, I find it fainly horrifying (the GCC ones, less concerned about the `-D __MINGW32__=1`), and I believe it could cause more harm than good. And I would very nervous about it going in before the stable tag:

These are preprocessor macros to let the code make build-time decisions based on knowing which exact GCC patch version it is being compiled with.

As a result, this change is creating an ABI that says that the decisions that would make sense for GCC 4.2.1 (released July 18, 2007) are the appropriate ones for CLANGPDB (regardless of version). And I simply cannot believe that would be universally true.

But it's doing this not just for OpensslLib where it would be a localised maintenance hazard, but across the entire codebase, including any future additions.

Could we look into a stop-gap fix for Openssllib only and break out a side discussion on the best long-term solution to bridge the gap?

gapalomi (approved):

nate-desimone:
> As a result, this change is creating an ABI that says that the decisions that would make sense for GCC 4.2.1 (released July 18, 2007) are the appropriate ones for CLANGPDB (regardless of version). And I simply cannot believe that would be universally true.

I fully agree that not all decisions made for GCC 4.2.1 will universally apply to clang. But it is a well-known clang-ism that clang always identifies itself as GCC 4.2.1. There is plenty of code out there with pre-processor macros are explicitly coded to handle the fact that clang identifies as an ancient GCC version. For example: https://github.com/welovegit/LZMA-SDK/blob/main/C/CpuArch.h#L299

mdkinney:
I agree that if the only known impact is to OpenSSL, then this flags should only be set in the OpenSSL lib INF files.

In general, the use of these macros to change behavior would be in include files associated with standard C libs.  Most of the edk2 components do not use those, and the components that do are typically added as submodule with INF that handles compatibility with edk2 along with custom libs to support the subset of C lib APIs used by the submodule.

leiflindholm:
> But it is a well-known clang-ism that clang always identifies itself as GCC 4.2.1. 

Clearly not well-known enough ;)

lgao4:
@ardbiesheuvel , @mdkinney , @leiflindholm , @ajfish , @makubacki , this patch requests to catch the stable tag 202508. This is a critical bug fix. If you have comments, please raise.

leiflindholm:
> @ardbiesheuvel , @mdkinney , @leiflindholm , @ajfish , @makubacki , this patch requests to catch the stable tag 202508. This is a critical bug fix. If you have comments, please raise.

@lgao4 For clarity: my opinion is that this change should _not_ go into stable tag.

The change enables a case that to my understanding has never worked before, so I would argue it's not a fix. If that case is important enough, or if I have misunderstood, it can be addressed locally in CryptoPkg/OpensslLib instead.

mdkinney:
@leiflindholm I agree it is not for `edk2-stable202508`.

The issue that this PR is attempting to resolve is only encountered if `-Wundef` is added.  This flag breaks the build if there is a reference to a macro in `#if` that is not defined.  It would be a new feature for edk2 `tools_def.txt` to add `-Wundef` to catch references to macros in `#if` that are not defined.  This may be a valuable feature to consider for edk2 for all supported toolchains.  If that feature is requested, then the macros definitions suggested in this PR would be required for `CLANGPDB` builds.

From some further analysis, it look like a feature request to add `-Wundef` would impact more than just OpenSSL.  It would impact other submodules that use standard C includes such as LZMA.  This would imply the need to add the required defines to `tools_def.txt`.

I recommend that a feature request created to add `-Wundef` to GCC/CLANG toolchains and add `/wd4668` to VS20xx toolchains. If that feature request is accepted, then this PR can be updated with the full set of changes to support that feature.

PR #11393, opened 2025-08-06
StandaloneMmPkg: Add APRIORI support

benjamindoron:
# Description

APRIORI support is practically required for an RSC infrastructure-based DebugLib stack, or the handlers might not be installed until late in the dispatcher's runs.

I also want feedback on some changes needed for the Standalone MM core to be able to use the RSC infrastructure itself: do we want this enough to make such changes?

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested as part of https://github.com/tianocore/edk2/pull/11370.

## Integration Instructions

N/A


benjamindoron:
Is this able to go in? Is the ability to output debug messages to the RSC stack in the core worth the very real risk that other memory allocations, which may be less tolerant of failures than DebugLibReportStatusCode, fail?

ardbiesheuvel:
`APRIORI` is in the spec, so I guess we need to support it. But it is a horrid hack, given that it completely ignores `DEPEX` relations between drivers. So for individual downstream platforms, we should support it but I would *really* like to avoid it in order to be able to use common debug infrastructure.

niruiyu:
As we continue the discussion regarding the need of APRIRI support in MM Core. Can we maybe create a different PR which contains the following commits from this PR? I hope the following 2 commits can be merged:
- StandaloneMmPkg/Core: Improve the readability of debug statement
- StandaloneMmPkg/Core: Print discovered drivers that weren't dispatched

benjamindoron:
@niruiyu Done, I've extracted those to https://github.com/tianocore/edk2/pull/12201

niruiyu (commented):
Can you evaluate if the APRIORI support is not necessary with this commit https://github.com/tianocore/edk2/commit/37db80097dbe4d1487191422240bdc0a9ad8d963?

The idea is that we can put some MM drivers to the very beginning of the FV and force them to not include a depex section through build rule, so that they will be dispatched in order before any other MM drivers.

benjamindoron:
@niruiyu That commit won't change anything. The standard is, or at least, was, that all drivers have a DEPEX section with at least `TRUE`. The RSC modules, which is what I want this for, have this section already, per their INF files. We want RSC to be loaded first, as the dispatcher will otherwise load modules in another order, meaning that we don't receive debug messages for much too long.

ardbiesheuvel:
> @niruiyu That commit won't change anything. The standard is, or at least, was, that all drivers have a DEPEX section with at least `TRUE`. The RSC modules, which is what I want this for, have this section already, per their INF files. We want RSC to be loaded first, as the dispatcher will otherwise load modules in another order, meaning that we don't receive debug messages for much too long.

Why isn't it sufficient to put the RSC module first in the FV? The dispatcher does a linear pass over the contents,
and so it will dispatch those modules first if their DEPEXes are satisfied. This seems flaky, but APRIORI is very flaky to begin with, and this is just a DEBUG feature, right?

PR #11396, opened 2025-08-07
MdeModulePkg: Fix UEFI runtime driver loading after EndOfDxe

### LINKED ISSUE #11374: [Bug]: UEFI runtime drivers loaded after EndOfDxe are not executable by OS

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

MdeModulePkg, OvmfPkg

DEBUG, NOOPT, RELEASE

### Current Behavior

We load a runtime Driver after EndOfDxe event and then call it from the operating system (OS).

As per [this source code reference](https://github.com/tianocore/edk2/blob/964c22b/MdeModulePkg/Core/Dxe/Misc/MemoryAttributesTable.c#L594-L597) UEFI runtime Drivers loaded after EndOfDxe are not added to `MemoryAttribytesTable`.

* In the past this meant they would get RWX permissions in the OS in some firmware, this was reported as #9906.
* Now https://github.com/tianocore/edk2/commit/bb248a95091ab542440053d9c289a97e80eb6630 introduced a regression, making these drivers get R/O permissions.

As a result of this change (https://github.com/tianocore/edk2/commit/bb248a95091ab542440053d9c289a97e80eb6630) attempting to execute runtime Driver (loaded after EndOfDxe) code from the OS results in a page fault.

### Expected Behavior

As per UEFI specification it is allowed for the UEFI application to load the runtime drivers and they should be available after a UEFI OS loader has taken control ([specification reference](https://uefi.org/specs/UEFI/2.11/02_Overview.html#uefi-drivers)).

### Steps To Reproduce

Our use case is a bit more complicated, but it is possible to reproduce the issue the following way:
 
1. Take OVMF (x64)
2. Load a simple runtime driver that will override any Runtime Service, e.g. SetVariable.
3. Try to load Linux, e.g. Debian 12 installer.

The following crash will be observed with console=ttyS0:

```
[    0.757347] kernel tried to execute NX-protected page - exploit attempt? (uid: 0)
[    0.758306] BUG: unable to handle page fault for address: fffffffeff7f5faa
[    0.758607] #PF: supervisor instruction fetch in kernel mode
[    0.758764] #PF: error_code(0x0011) - permissions violation
[    0.759045] PGD 50704067 P4D 50704067 PUD 50700063 PMD 11c4063 PTE 800000007f5f5061
[    0.759884] Oops: 0011 [#1] PREEMPT SMP PTI
[    0.759946] CPU: 0 PID: 0 Comm: swapper/0 Not tainted 6.1.0-35-amd64 #1  Debian 6.1.137-1
[    0.759946] Hardware name: Acidanthera Macmini8,1/Mac-7BA5B2DFE22DDD8C, BIOS 2069.0.0.0.0 08/07/2024
[    0.759946] RIP: 0010:0xfffffffeff7f5faa
[    0.759946] Code: 00 5b 5f 5e 41 5c 41 5d 41 5e 41 5f 5d c3 66 83 a5 c0 fd ff ff 00 48 8d 8d b0 fd ff ff 6a 10 5a e8 58 f1 ff ff e9 72 fd ff ff <55> 48 89 e5 41 57 41 56 41 55 41 54 56 57 53 48 81 ec 48 02 00 00
[    0.759946] RSP: 0000:ffffffffbce03e08 EFLAGS: 00010282
[    0.759946] RAX: 0000000000000202 RBX: 8000000000000006 RCX: ffffffffbc40a2b8
[    0.759946] RDX: ffffffffbce03e90 RSI: ffffffffbc40a2b8 RDI: fffffffeff7f5faa
[    0.759946] RBP: ffffffffbce03e40 R08: 0000000000000007 R09: 0000000000000000
[    0.759946] R10: 0000000000000048 R11: ffffffffbced4400 R12: ffffffffbce03e90
[    0.759946] R13: 0000000000000007 R14: 0000000000000000 R15: 0000000000000000
[    0.759946] FS:  0000000000000000(0000) GS:ffff9cc0bb600000(0000) knlGS:0000000000000000
[    0.759946] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[    0.759946] CR2: fffffffeff7f5faa CR3: 00000000011ca000 CR4: 00000000001506f0
[    0.759946] Call Trace:
[    0.759946]  <TASK>
[    0.759946]  ? __efi_call+0x28/0x30
[    0.759946]  virt_efi_set_variable_nonblocking+0xb0/0x120
[    0.759946]  efi_delete_dummy_variable+0x5c/0x80
[    0.759946]  efi_enter_virtual_mode+0x429/0x44b
[    0.759946]  start_kernel+0x6a1/0x733
[    0.759946]  secondary_startup_64_no_verify+0xe5/0xeb
[    0.759946]  </TASK>
[    0.759946] Modules linked in:
[    0.759946] CR2: fffffffeff7f5faa
[    0.759946] ---[ end trace 0000000000000000 ]---
[    0.759946] RIP: 0010:0xfffffffeff7f5faa
[    0.759946] Code: 00 5b 5f 5e 41 5c 41 5d 41 5e 41 5f 5d c3 66 83 a5 c0 fd ff ff 00 48 8d 8d b0 fd ff ff 6a 10 5a e8 58 f1 ff ff e9 72 fd ff ff <55> 48 89 e5 41 57 41 56 41 55 41 54 56 57 53 48 81 ec 48 02 00 00
[    0.759946] RSP: 0000:ffffffffbce03e08 EFLAGS: 00010282
[    0.759946] RAX: 0000000000000202 RBX: 8000000000000006 RCX: ffffffffbc40a2b8
[    0.759946] RDX: ffffffffbce03e90 RSI: ffffffffbc40a2b8 RDI: fffffffeff7f5faa
[    0.759946] RBP: ffffffffbce03e40 R08: 0000000000000007 R09: 0000000000000000
[    0.759946] R10: 0000000000000048 R11: ffffffffbced4400 R12: ffffffffbce03e90
[    0.759946] R13: 0000000000000007 R14: 0000000000000000 R15: 0000000000000000
[    0.759946] FS:  0000000000000000(0000) GS:ffff9cc0bb600000(0000) knlGS:0000000000000000
[    0.759946] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033
[    0.759946] CR2: fffffffeff7f5faa CR3: 00000000011ca000 CR4: 00000000001506f0
[    0.759946] Kernel panic - not syncing: Attempted to kill the idle task!
[    0.759946] ---[ end Kernel panic - not syncing: Attempted to kill the idle task! ]—-
```

### Build Environment

```markdown
- OS(s): macOS 15, but also reproducible on Ubuntu 24.04.
- Tool Chain(s): XCODE5, but also reproducible with GCC
```

### Version Information

```text
edk2/master (7fe3609022ce7f52190bd6384d117f1510e15afe)

Bug introduced in https://github.com/tianocore/edk2/commit/bb248a95091ab542440053d9c289a97e80eb6630.
```

### Urgency

Medium

I will fix it

Maintainer feedback requested

### Anything else?

I suggest to remove [MemoryAttributesTable.c#L594-L597 ](https://github.com/tianocore/edk2/blob/964c22b/MdeModulePkg/Core/Dxe/Misc/MemoryAttributesTable.c#L594-L597) lines to fix the issue. Can submit a patch if there are no objections.

vit9696:
Memory Attributes Table needs to be updated to contain executable permissions for UEFI runtime drivers loaded after EndOfDxe. Fixes a regression introduced by bb248a9.

# Description

Fixes #11374

## How This Was Tested

I built OVMF before and after this change and tested Debian 12 ISO booting in QEMU after a custom UEFI runtime driver was loaded.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryAttributesTable.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryAttributesTable.c
@@ -87,7 +87,6 @@ STATIC IMAGE_PROPERTIES_PRIVATE_DATA  mImagePropertiesPrivateData = {
 STATIC EFI_LOCK  mMemoryAttributesTableLock = EFI_INITIALIZE_LOCK_VARIABLE (TPL_NOTIFY);
 
 BOOLEAN                      mMemoryAttributesTableEnable      = TRUE;
-BOOLEAN                      mMemoryAttributesTableEndOfDxe    = FALSE;
 EFI_MEMORY_ATTRIBUTES_TABLE  *mMemoryAttributesTable           = NULL;
 BOOLEAN                      mMemoryAttributesTableReadyToBoot = FALSE;
 BOOLEAN                      gMemoryAttributesTableForwardCfi  = TRUE;
@@ -282,7 +281,6 @@ InstallMemoryAttributesTableOnEndOfDxe (
   IN VOID       *Context
   )
 {
-  mMemoryAttributesTableEndOfDxe = TRUE;
   InstallMemoryAttributesTable ();
 
   DEBUG_CODE_BEGIN ();
@@ -591,11 +589,6 @@ InsertImageRecord (
 
   DEBUG ((DEBUG_VERBOSE, "InsertImageRecord - 0x%x\n", RuntimeImage));
 
-  if (mMemoryAttributesTableEndOfDxe) {
-    DEBUG ((DEBUG_INFO, "Do not insert runtime image record after EndOfDxe\n"));
-    return;
-  }
-
   ImageRecord = AllocatePool (sizeof (*ImageRecord));
   if (ImageRecord == NULL) {
     return;
@@ -677,11 +670,6 @@ RemoveImageRecord (
 
   DEBUG ((DEBUG_VERBOSE, "RemoveImageRecord - 0x%x\n", RuntimeImage));
   DEBUG ((DEBUG_VERBOSE, "RemoveImageRecord - 0x%016lx - 0x%016lx\n", (EFI_PHYSICAL_ADDRESS)(UINTN)RuntimeImage->ImageBase, RuntimeImage->ImageSize));
-
-  if (mMemoryAttributesTableEndOfDxe) {
-    DEBUG ((DEBUG_INFO, "Do not remove runtime image record after EndOfDxe\n"));
-    return;
-  }
 
   ImageRecord = FindImageRecord ((EFI_PHYSICAL_ADDRESS)(UINTN)RuntimeImage->ImageBase, RuntimeImage->ImageSize, &mImagePropertiesPrivateData.ImageRecordList);
   if (ImageRecord == NULL) {


lgao4:
@os-d , can you help review this change?

os-d (approved):
As noted in the associated issue, I agree that the MAT should always be fully reflective of loaded runtime drivers. Some platforms do use end of DXE as a security boundary (and there are simple safety concerns about loading drivers after end of DXE) but if edk2 wants to make this a security boundary, then this should happen in the loader, not here. I.e. if a security policy is going to be in effect, it should block loading the images, not where we report them. I’m not advocated for or against this, just noting why this PR is acceptable to me; this was never the right place for this check.

mdkinney:
Is this required for edk2-stable202508?

ardbiesheuvel:
I agree with @os-d that this restriction seems arbitrary, rather than a fundamental policy, and this makes me suspicious that it might have been added to work around a practical issue. Perhaps it would be better to relax this restriction to `ReadyToBoot` instead, rather than remove it completely.

> Is this required for edk2-stable202508?

I think it is important that we get this fixed, and rushing this through now seems premature, given that the regression appears in a couple of preceding stable tags as well. Let's keep this broken until we decide on a permanent fix.

os-d:
> I agree with @os-d that this restriction seems arbitrary, rather than a fundamental policy, and this makes me suspicious that it might have been added to work around a practical issue. Perhaps it would be better to relax this restriction to `ReadyToBoot` instead, rather than remove it completely.
> 

@ardbiesheuvel this came from the old properties table: https://github.com/tianocore/edk2/commit/4a723ed258367471eac8b4ae32558f09ef65672e. This was before my time in edk2, so I don't have the context on the differences between the properties table and MAT (as I see you copied most of the properties table logic to the MAT). Looks like it was to keep SMM in sync with the OS in terms of MAT.

As I noted, I still think this is the wrong approach. Leaving a driver hanging in the wind so the OS MAT and SMM MAT are the same is dangerous. If SMM needs to be kept up to date, it should have a notification mechanism.

Which again goes back to if edk2 wants to have a boundary here, the boundary should be in the loader not in the MAT code.

vit9696:
@lgao4, @os-d, @ardbiesheuvel, @mdkinney, I believe we are generally settled the change by itself is reasonable. Can we merge this please?

os-d:
> @lgao4, @os-d, @ardbiesheuvel, @mdkinney, I believe we are generally settled the change by itself is reasonable. Can we merge this please?

@lgao4, as MdeModulePkg maintainer, needs to review in order to merge. I don't have authority over this package :).

Summarizing from above: I think this change on its own is appropriate. edk2 should evaluate whether they want a security boundary here and if so prevent loading drivers at this time instead of simply not reporting them in the MAT.

lgao4:
> > @lgao4, @os-d, @ardbiesheuvel, @mdkinney, I believe we are generally settled the change by itself is reasonable. Can we merge this please?
> 
> @lgao4, as MdeModulePkg maintainer, needs to review in order to merge. I don't have authority over this package :).
> 
> Summarizing from above: I think this change on its own is appropriate. edk2 should evaluate whether they want a security boundary here and if so prevent loading drivers at this time instead of simply not reporting them in the MAT.

I understand this change. I am OK to always report the information. The consumer can decide how to handle them.

lgao4 (approved):

vit9696:
Thank you for your review! Can we merge this please?

PR #11401, opened 2025-08-08
MdeModulePkg: consider UNSUPPORTED return as valid in ArmFfaStandaloneMmLib

LeviYeoReum:
commit b534cabbdac4 ("ArmFfaLib: Add Rx/Tx support for Stmm secure partition") makes ArmFfaStnadlaoneMm(Core)Lib map Rx/Tx buffer in its constructor.

This makes a failure of loading StandaloneMm in legacy platform which doesn't implements Rx/Tx buffer related API since it doesn't need to.

StandaloneMm could be only service provider not cosumer in some platform
where doesn't need to map RxTx buffer.
Therefore, Considier EFI_UNSUPPORTED return in ArmFfaStandaloneMmLib's
constructor.

Fixes: b534cabbdac4 ("ArmFfaLib: Add Rx/Tx support for Stmm secure partition")

### DIFF
--- a/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
+++ b/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
@@ -73,7 +73,7 @@ SendFfaMmCommunicate (
 
   while (Status == EFI_INTERRUPT_PENDING) {
     // We are assuming vCPU0 of the StMM SP since it is UP.
-    Status = ArmFfaLibRun (mStMmPartId, 0x00);
+    Status = ArmFfaLibRun (mStMmPartId, 0x00, NULL);
   }
 
   return Status;
--- a/ArmPkg/Drivers/MmCommunicationPei/MmCommunicationPei.c
+++ b/ArmPkg/Drivers/MmCommunicationPei/MmCommunicationPei.c
@@ -275,7 +275,7 @@ SendFfaMmCommunicate (
 
   while (Status == EFI_INTERRUPT_PENDING) {
     // We are assuming vCPU0 of the StMM SP since it is UP.
-    Status = ArmFfaLibRun (mStMmPartId, 0x00);
+    Status = ArmFfaLibRun (mStMmPartId, 0x00, NULL);
   }
 
   return Status;
--- a/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
+++ b/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
@@ -535,6 +535,7 @@ ErrorHandler:
 
   @param [in]   PartId       Partition id
   @param [in]   CpuNumber    Cpu number in partition
+  @param [out]  DirectMsgArg return arguments for direct msg resp/resp2
 
   @retval EFI_SUCCESS
   @retval Other              Error
@@ -543,10 +544,12 @@ ErrorHandler:
 EFI_STATUS
 EFIAPI
 ArmFfaLibRun (
-  IN  UINT16  PartId,
-  IN  UINT16  CpuNumber
+  IN  UINT16           PartId,
+  IN  UINT16           CpuNumber,
+  OUT DIRECT_MSG_ARGS  *DirectMsgArg OPTIONAL
   )
 {
+  EFI_STATUS    Status;
   ARM_FFA_ARGS  FfaArgs;
 
   ZeroMem (&FfaArgs, sizeof (ARM_FFA_ARGS));
@@ -556,7 +559,39 @@ ArmFfaLibRun (
 
   ArmCallFfa (&FfaArgs);
 
-  return FfaArgsToEfiStatus (&FfaArgs);
+  Status = FfaArgsToEfiStatus (&FfaArgs);
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  if (DirectMsgArg != NULL) {
+    ZeroMem (DirectMsgArg, sizeof (DIRECT_MSG_ARGS));
+
+    if (FfaArgs.Arg0 == ARM_FID_FFA_MSG_SEND_DIRECT_RESP) {
+      DirectMsgArg->Arg0 = FfaArgs.Arg3;
+      DirectMsgArg->Arg1 = FfaArgs.Arg4;
+      DirectMsgArg->Arg2 = FfaArgs.Arg5;
+      DirectMsgArg->Arg3 = FfaArgs.Arg6;
+      DirectMsgArg->Arg4 = FfaArgs.Arg7;
+    } else if (FfaArgs.Arg0 == ARM_FID_FFA_MSG_SEND_DIRECT_RESP2) {
+      DirectMsgArg->Arg0  = FfaArgs.Arg4;
+      DirectMsgArg->Arg1  = FfaArgs.Arg5;
+      DirectMsgArg->Arg2  = FfaArgs.Arg6;
+      DirectMsgArg->Arg3  = FfaArgs.Arg7;
+      DirectMsgArg->Arg4  = FfaArgs.Arg8;
+      DirectMsgArg->Arg5  = FfaArgs.Arg9;
+      DirectMsgArg->Arg6  = FfaArgs.Arg10;
+      DirectMsgArg->Arg7  = FfaArgs.Arg11;
+      DirectMsgArg->Arg8  = FfaArgs.Arg12;
+      DirectMsgArg->Arg9  = FfaArgs.Arg13;
+      DirectMsgArg->Arg10 = FfaArgs.Arg14;
+      DirectMsgArg->Arg11 = FfaArgs.Arg15;
+      DirectMsgArg->Arg12 = FfaArgs.Arg16;
+      DirectMsgArg->Arg13 = FfaArgs.Arg17;
+    }
+  }
+
+  return Status;
 }
 
 /**
--- a/MdePkg/Include/Library/ArmFfaLib.h
+++ b/MdePkg/Include/Library/ArmFfaLib.h
@@ -301,6 +301,7 @@ ArmFfaLibSpmIdGet (
 
   @param [in]   PartId       Partition id
   @param [in]   CpuNumber    Cpu number in partition
+  @param [out]  DirectMsgArg return arguments for direct msg resp/resp2
 
   @retval EFI_SUCCESS
   @retval Other              Error
@@ -309,8 +310,9 @@ ArmFfaLibSpmIdGet (
 EFI_STATUS
 EFIAPI
 ArmFfaLibRun (
-  IN  UINT16  PartId,
-  IN  UINT16  CpuNumber
+  IN  UINT16           PartId,
+  IN  UINT16           CpuNumber,
+  OUT DIRECT_MSG_ARGS  *DirectMsgArg OPTIONAL
   );
 
 /**
--- a/SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
+++ b/SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
@@ -202,7 +202,7 @@ Tpm2GetInterfaceVersion (
   Status = ArmFfaLibMsgSendDirectReq2 (mFfaTpm2PartitionId, &gTpm2ServiceFfaGuid, &FfaDirectReq2Args);
   while (Status == EFI_INTERRUPT_PENDING) {
     // We are assuming vCPU0 of the TPM SP since it is UP.
-    Status = ArmFfaLibRun (mFfaTpm2PartitionId, 0x00);
+    Status = ArmFfaLibRun (mFfaTpm2PartitionId, 0x00, &FfaDirectReq2Args);
   }
 
   if (EFI_ERROR (Status)) {
@@ -254,7 +254,7 @@ Tpm2GetFeatureInfo (
   Status = ArmFfaLibMsgSendDirectReq2 (mFfaTpm2PartitionId, &gTpm2ServiceFfaGuid, &FfaDirectReq2Args);
   while (Status == EFI_INTERRUPT_PENDING) {
     // We are assuming vCPU0 of the TPM SP since it is UP.
-    Status = ArmFfaLibRun (mFfaTpm2PartitionId, 0x00);
+    Status = ArmFfaLibRun (mFfaTpm2PartitionId, 0x00, &FfaDirectReq2Args);
   }
 
   if (EFI_ERROR (Status)) {
@@ -298,7 +298,7 @@ Tpm2ServiceStart (
   Status = ArmFfaLibMsgSendDirectReq2 (mFfaTpm2PartitionId, &gTpm2ServiceFfaGuid, &FfaDirectReq2Args);
   while (Status == EFI_INTERRUPT_PENDING) {
     // We are assuming vCPU0 of the TPM SP since it is UP.
-    Status = ArmFfaLibRun (mFfaTpm2PartitionId, 0x00);
+    Status = ArmFfaLibRun (mFfaTpm2PartitionId, 0x00, &FfaDirectReq2Args);
   }
 
   if (EFI_ERROR (Status)) {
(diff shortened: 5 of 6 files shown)


MdeModulePkg/Library/ArmFfaLib/ArmFfaStandaloneMmLib.c
```diff
@@ -58,6 +59,14 @@ ArmFfaStandaloneMmLibConstructor (
     DEBUG ((DEBUG_ERROR, "%a failed. Status = %r\n", __func__, Status));
   }
 
+  /*
+   * Some legacy platform doesn't need to map Rx/Tx buffer (i.e) RPMB platform.
+   * In this case, Skip mapping Rx/Tx buffer.
+   */
+  if (PcdGetBool (PcdFfaSkipRxTxMapping)) {
+    return Status;
```
apalos: It's a bit more readable if you explicitly return EFI_SUCCESS.

Other than that lgtm

LeviYeoReum: That would be better. I'll change it.

MdeModulePkg/Library/ArmFfaLib/ArmFfaStandaloneMmLib.c
```diff
@@ -56,6 +57,15 @@ ArmFfaStandaloneMmLibConstructor (
 
   if (EFI_ERROR (Status)) {
     DEBUG ((DEBUG_ERROR, "%a failed. Status = %r\n", __func__, Status));
+    return Status;
+  }
+
+  /*
+   * Some legacy platform doesn't need to map Rx/Tx buffer (i.e) RPMB platform.
```
kuqin12: Is this due to not needed? Or not supported by legacy platform?

Rx/Tx buffer is introduced in ALP2, according to the spec history. I assume this is not the case?

If this is just not needed? Do you want to add the PCD check in other constructors as well?

LeviYeoReum: Hi Kun.
It’s both of.

they are NOT IMPLEMENTED 
in legacy platform — RPMB (optee’s stmm_sp as TA) since it doesn’t need to use “RX/TX buffer“.

This platform is bit of special since FF-A operation from StMM TA comes to syscall handler not to SPMC directly
And this syscall handler doens’t implement RX/TX related since it doesn’t need.

In case of normal world side, it doesn’t need to use this PCD since the SPMC is implemented all.

TBH, this patch to work pre-existing platform as it is.
Anyway if FFA_FEATURE for Rx/Tx returns UNSPPORTED, without this Pcd, it always failed though that’s not need it.

mikkorapeli-linaro:
With https://github.com/tianocore/edk2-platforms/pull/842, https://github.com/tianocore/edk2/pull/11401 and https://github.com/tianocore/edk2/pull/11401 applied we have our rockpi4b tests passing: https://ledge.validation.linaro.org/scheduler/job/118103
Using edk2 base commit d68f418300cd13005d2c78fc8d918b61698f5d00 and edk2-platforms commit https://github.com/tianocore/edk2-platforms/commit/d6eb4de31701a50b26da061919565c3a3460db88, u-boot 3526f990b77fca1c933f1d8b24eb9385010a05bf, optee 61150e5450956dec327375ec120e31210774c991

MdeModulePkg/Library/ArmFfaLib/ArmFfaStandaloneMmLib.c
samimujawar: I also think we should avoid introducing a new PCD if possible. 
I can see that ArmFfaLibRxTxMap() calls get features to see if ARM_FID_FFA_RXTX_MAP is supported. Also the fid  ARM_FID_FFA_RXTX_MAP invocation will return UNSUPPORTED if the feature is absent. 
This means if the SPMC does not implement the RX/TX mapping functionality it will return unsupported.

Please let me know if I have missed something, but I think this optional feature should be handled by the code appropriately and the new PCD should be avoided.

>This platform is bit of special since FF-A >operation from StMM TA comes to syscall >handler not to SPMC directly
>And this syscall handler doens’t implement >RX/TX related since it doesn’t need.

If the invocation does not return UNSUPPORTED then it means the syscall handler or SPMC is not spec compliant and that needs to be fixed.

LeviYeoReum: > I can see that ArmFfaLibRxTxMap() calls get features to see if ARM_FID_FFA_RXTX_MAP is supported. Also the fid ARM_FID_FFA_RXTX_MAP invocation will return UNSUPPORTED if the feature is absent.
This means if the SPMC does not implement the RX/TX mapping functionality it will return unsupported.

Yes. But in Constructor, what should return the finally?
In some of platform it would be fine but in other it would definitely fail. 

> If the invocation does not return UNSUPPORTED then it means the syscall handler or SPMC is not spec compliant and that needs to be fixed.

I'm not sure the RX/TX ABIs is "MANDATORRY" at least when I see the FF-A memory mangement specification.
IOW, where Rx/Tx buffer doesn't required it seems to FF-A feaure allows to return "UNSUPPORTED" for that interface.

So if above is true, I think this PCD is need for the platform where it doesn't need to use Rx/Tx buffer
as its origin purpose.

Am I missing something?

jenswikl: In 7.2.2.2 Buffer setup, it says:
> Allocation of a buffer pair for an endpoint can be done by the endpoint or its partition manager.

It's the latter case for StMM managed by OP-TEE. If the buffers are already mapped, why do we care about FFA_RXTX_MAP?

kuqin12: > Yes. But in Constructor, what should return the finally? In some of platform it would be fine but in other it would definitely fail.
> 
I think this is a valid point. But would it be better to query the interface support through "GET_FEATURE" and try to map it if this is supported?

kuqin12: > It's the latter case for StMM managed by OP-TEE. If the buffers are already mapped, why do we care about FFA_RXTX_MAP?

Would you please let me know how the partition will get where the Rx/Tx buffer is located in this case?

LeviYeoReum: > I think this is a valid point. But would it be better to query the interface support through "GET_FEATURE" and try to map it if this is supported?

I think it's enough FFA_FEATURE called in ArmFfaLibRxTxMap() to check the Valid Rx/Tx buffer size.
Whether call with Function ID or to get Rx/Tx size info, it'll return UNSUPPORTED when Rx/Tx buffer ABI is not implemented.

LeviYeoReum: > Would you please let me know how the partition will get where the Rx/Tx buffer is located in this case?

I think in this case, SPMC can use pre-allocated memory as Rx/Tx buffer and 
this can be passed via Hob or other way to StandaloneMm.

But AFAIK, If platform SPMC doesn't need to implment Rx/Tx buffer
    - Get partition info via parition info reg.
    - running model with FFA_DIRECT_MSG
    - No memory share/donate between partition
 
Then, it doesn't need to implement Rx/Tx buffer ABIs.
(Again when I see the spec it doesn't seem to MANDATORY).

Am I missing something?

kuqin12: > I think it's enough FFA_FEATURE called in ArmFfaLibRxTxMap() to check the Valid Rx/Tx buffer size. Whether call with Function ID or to get Rx/Tx size info, it'll return UNSUPPORTED when Rx/Tx buffer ABI is not implemented.
>
What I meant is why do we not take the UNSUPPORTED as an accepted error code and proceed with the rest of the routine? I agree that we need some fix if this function ID is not implemented.

kuqin12: > Then, it doesn't need to implement Rx/Tx buffer ABIs. (Again when I see the spec it doesn't seem to MANDATORY).
> 
If it is not mandatory, do you think we need to patch the same for DXE/PEI/SEC as well? Is the SPMC replying with inconsistent error codes?

LeviYeoReum: Hi @kuqin12 ,

> What I meant is why do we not take the UNSUPPORTED as an accepted error code and proceed with the rest of the routine? I agree that we need some fix if this function ID is not implemented.

> If it is not mandatory, do you think we need to patch the same for DXE/PEI/SEC as well? Is the SPMC replying with inconsistent error codes?

I think we can do like:
  - StandaloneMm: since it's definitely service provider. it can be consumer of other or not.
                                 Therefore, EFI_UNSUPPORTED could be valid return value.
                                 But, at least need to remain log "No Rx/Tx buffer support".
                                 
  - DXE/PEI/SEC: Normal world is always a "cosumer" of the service
                             Therefore, there must be a way to get Partition-Id.
                             If there is no support for "partition-info-get with REG", Rx/Tx buffer must be mapped to use
                             "partition-info-get via Rx buffer". In this case EFI_UNSUPPORTED is definitely ERROR
                             else "partition-info-get with REG" is implemented, we can consider EFI_UNSUPPORTED
                             as a valid return.

I think in this patch to fix it seems enough to change consider EFI_UNSUPPORTED while getting Rx/Tx map is enough

But for later, I think it's time to implements the ArmFfaLibPartitionInfoReg() and some helper function
so that Normal world's ArmFfaLib() can separate when it considers EFI_UNSUPPROTED while mapping Rx/Tx buffer as valid or not.

Am I missing?

Thanks.

kuqin12: > I think in this patch to fix it seems enough to change consider EFI_UNSUPPORTED while getting Rx/Tx map is enough
> 
Okay, sure, can we do this in this PR and do not introduce the PCD?

We have a function in Project MU for `ArmFfaLibPartitionInfoReg`: https://github.com/microsoft/mu_feature_ffa/blob/a336c75fca12bbba0f48d966e8b626c11d7434c2/FfaFeaturePkg/Library/ArmFfaLibEx/ArmFfaLibEx.c#L447C1-L447C25

I think we can port that function (and maybe other notification-related functions) over to EDK2 sometime later.

LeviYeoReum: > Okay, sure, can we do this in this PR and do not introduce the PCD?

Sure.

>  We have a function in Project MU for ArmFfaLibPartitionInfoReg: https://github.com/microsoft/mu_feature_ffa/blob/a336c75fca12bbba0f48d966e8b626c11d7434c2/FfaFeaturePkg/Library/ArmFfaLibEx/ArmFfaLibEx.c#L447C1-L447C25

Thanks for reference.

> I think we can port that function (and maybe other notification-related functions) over to EDK2 sometime later.

Sure.
At first let me change this patch. and then port ArmFfaLibPartitionInfoReg() first to resolve this return issue :)
and about notification let do it in another patchset.

Thanks!

LeviYeoReum:
@mikkorapeli-linaro and @apalos.
Could you make a test with this patch only without edk2-platform's patch?
(https://github.com/tianocore/edk2-platforms/pull/842).

Thanks.

MdeModulePkg/Library/ArmFfaLib/ArmFfaStandaloneMmLib.c
LeviYeoReum: Hi @kuqin12,

> We have a function in Project MU for ArmFfaLibPartitionInfoReg: https://github.com/microsoft/mu_feature_ffa/blob/a336c75fca12bbba0f48d966e8b626c11d7434c2/FfaFeaturePkg/Library/ArmFfaLibEx/ArmFfaLibEx.c#L447C1-L447C25

I've upload pr -- https://github.com/tianocore/edk2/pull/11420

mikkorapeli-linaro:
With this patch alone, https://github.com/tianocore/edk2/commit/86b2318edaf2e59dadf672e752e4d1e8950dd159.patch, the result is a panic from OP-TEE.

https://ledge.validation.linaro.org/scheduler/job/118555

```
E/TC:? 0 spm_handle_scall:979 Undefined syscall 0x84000064
E/TC:? 0 
E/TC:? 0 User mode undef-abort at address 0x0
E/TC:? 0  esr 0x00000000  ttbr0 0x2000030122000   ttbr1 0x00000000   cidr 0x0
E/TC:? 0  cpu #0          cpsr 0x00000000
```

OP-TEE panics on undefined syscall by default. This has been the case for a long time so I don't think thie new code should be asking for any supported FFA features.

Cc: @jenswi-linaro

LeviYeoReum:
> With this patch alone, https://github.com/tianocore/edk2/commit/86b2318edaf2e59dadf672e752e4d1e8950dd159.patch, the result is a panic from OP-TEE.
> 
> https://ledge.validation.linaro.org/scheduler/job/118555
> 
> ```
> E/TC:? 0 spm_handle_scall:979 Undefined syscall 0x84000064
> E/TC:? 0 
> E/TC:? 0 User mode undef-abort at address 0x0
> E/TC:? 0  esr 0x00000000  ttbr0 0x2000030122000   ttbr1 0x00000000   cidr 0x0
> E/TC:? 0  cpu #0          cpsr 0x00000000
> ```
> 
> OP-TEE panics on undefined syscall by default. This has been the case for a long time so I don't think thie new code should be asking for any supported FFA features.
> 
> Cc: @jenswi-linaro

Hi @mikkorapeli-linaro and @apalos,

Should we change the optee code to return StandaloneMm with `UNSUPPORTED` instead of panic? like:

```
--- a/core/arch/arm/kernel/stmm_sp.c
+++ b/core/arch/arm/kernel/stmm_sp.c
@@ -975,10 +975,14 @@ static bool spm_handle_scall(struct thread_scall_regs *regs)
                DMSG("Received FFA mem perm set");
                spm_handle_set_mem_attr(regs);
                return true;
-       default:
-               EMSG("Undefined syscall %#"PRIx32, (uint32_t)*a0);
+       case FFA_ERROR:
+               EMSG("Received FFA error");
                return_from_sp_helper(true /*panic*/, 0xabcd, regs);
                return false;
+       default:
+               DMSG("Undefined syscall %#"PRIx32, (uint32_t)*a0);
+               spm_eret_error(STMM_RET_NOT_SUPPORTED, regs);
+               return true;
        }
 }
```

Or is any other reason should be paniced when receive unsupported function-id?

jenswikl:
> Should we change the optee code to return StandaloneMm with `UNSUPPORTED` instead of panic?

Yes, something like that makes more sense.

LeviYeoReum:
Thanks @jenswi-linaro.

I've upload one PR for this:
  - https://github.com/OP-TEE/optee_os/pull/7508
  
@jenswi-linaro, @mikkorapeli-linaro and @apalos 

If you don't mind, could you review this small change too?

Thanks!

kuqin12 (approved):

LeviYeoReum:
> There a typo in the commit message for 'ArmFfaStnadlaoneMm'.
> 
> Other than that the patch looks reasonable

Thanks. I've fixed it!

mikkorapeli-linaro:
Tested https://github.com/tianocore/edk2/pull/11401, https://github.com/OP-TEE/optee_os/pull/7508 and https://github.com/tianocore/edk2-platforms/pull/842 on top of latest edk2, edk2-platforms, optee_os, u-boot, tf-a etc git master branches.

On rockpi4b with stmm config and our tests pass:
https://ledge.validation.linaro.org/scheduler/job/118692

`Tested-by: Mikko Rapeli <mikko.rapeli@linaro.org>`

LeviYeoReum:
I don't know why the megebot's failure is happend (it seems totally unrelated to this patch).
Is that ignorable??

ardbiesheuvel:
@mergifyio requeue

ardbiesheuvel:
@mergifyio rebase

ardbiesheuvel:
I did not close this

ardbiesheuvel:
@mergifyio requeue

ardbiesheuvel:
Reopening ...

ardbiesheuvel:
Merged via #11473

PR #11402, opened 2025-08-08
[edk2-stable202508] SecurityPkg/FvReportPei: Fix regression from CodeQL fixes

mdkinney:
# Description

PR https://github.com/tianocore/edk2/pull/11307 fixed a number of CodeQL related issues in the SecurityPkg. An update to FvReportPei introduced a behavior change that causes some FVs to fail verification.

If HashInfo is NULL, then the FV verification should be bypassed and the return status should be EFI_SUCCESS. Change the return status from EFI_NOT_FOUND to EFI_SUCCESS to match the behavior before the CodeQL related changes.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

FV verification in PEI on some platforms was failing.  With this change the FV verification in PEI worked as expected.

## Integration Instructions

N/A

### DIFF
--- a/SecurityPkg/FvReportPei/FvReportPei.c
+++ b/SecurityPkg/FvReportPei/FvReportPei.c
@@ -397,7 +397,8 @@ CheckStoredHashFv (
                  BootMode
                  );
     } else {
-      Status = EFI_NOT_FOUND;
+      DEBUG ((DEBUG_INFO, "Bypass FV hash verification\r\n"));
+      Status = EFI_SUCCESS;
     }
 
     if (!EFI_ERROR (Status)) {


mdkinney:
This fix addresses a regression and is required for edk2-stable202508.

apop5 (approved):

SecurityPkg/FvReportPei/FvReportPei.c
```diff
@@ -397,7 +397,8 @@ CheckStoredHashFv (
                  BootMode
                  );
     } else {
-      Status = EFI_NOT_FOUND;
+      DEBUG ((DEBUG_INFO, "Bypass FV hash verification\r\n"));
+      Status = EFI_SUCCESS;
```
makubacki, line 401: It's a good opportunity to update the function description to clarify this behavior.

mdkinney: Suggest opening an issue for this to be addressed after stable tag.  The focus here is to address the change in behavior that introduced a regression.

makubacki: I think poor documentation was partly a cause of the regression. I'll open the issue.

mdkinney: I agree.  Core review also missed the behavior change.  First few lines of VerifyHashedFv() check for HashInfo NULL and return EFI_SUCCESS for that condition.

makubacki: https://github.com/tianocore/edk2/issues/11404

mdkinney: Also not clear what issue CodeQL flagged.  Perhaps the fix for CodeQL should have been in VerifyHashedFv() instead of CheckStoredHashFv().  CodeQL should have had visibility to both since they are in the same file.

Flickdm: https://github.com/tianocore/edk2/security/code-scanning/1678

For reference in my closed PR - Codeql flagged the revert as well since I was doing a pure revert

apop5: > Also not clear what issue CodeQL flagged. Perhaps the fix for CodeQL should have been in VerifyHashedFv() instead of CheckStoredHashFv(). CodeQL should have had visibility to both since they are in the same file.

CodeQl flagged a missing null check, which was that the function has the possibility of returning NULL, and that the result was being used before it was checked, CodeQl doesn't make a distinction between functions which allow a NULL value, it just finds the code path where a protentional null value was passed in.

mdkinney:
@leiflindholm can you provide an example from the original commit that was not correct and what you would like to see instead for these CodeQL issue resolutions?  I would like to make sure we all agree on the patterns to use for these types of fixes.

mdkinney:
> I disagree. This fixes one caught issue with an absolute mess of a commit that should never have been merged in the first place.
> 
> The right solution is to revert the broken commit and introduce the changes in a legible fashion that can be usefully reviewed.

Doe anyone disagree with reverting PR #11307 for edk2-stable202508 and revisiting these changes after the stable tag?

leiflindholm:
> @leiflindholm can you provide an example from the original commit that was not correct and what you would like to see instead for these CodeQL issue resolutions? I would like to make sure we all agree on the patterns to use for these types of fixes.

It's the basic "one change, one commit". "Change all issues flagged by codeQL in SecurityPkg" is not one thing. Which could be gleaned simply from the commit message, which listed four authors, had an itemised list, and _still_ added "among others".

I think it's fine to group all occurrences _of a specific pattern_ within a single package, but other than that, code quality changes should not be treated differently from other changes.

mdkinney:
I agree that the changes should be grouped based on package and a specific CodeQL issue type to make them easier to review.  Once the reviewer(s) understands the specific CodeQL issue type being addressed, then evaluating each change to make sure that specific issue type is resolved with no behavioral changes is much easier.

Multiple authors across multiple commits seems ok to me.  Splitting up the work can be faster.  However, the patterns used to address the same type of issue type should be consistent.

ardbiesheuvel:
> I agree that the changes should be grouped based on package and a specific CodeQL issue type to make them easier to review.

+1

ardbiesheuvel:
> Does anyone disagree with reverting PR #11307 for edk2-stable202508 and revisiting these changes after the stable tag?

No objection.

leiflindholm:
> I agree that the changes should be grouped based on package and a specific CodeQL issue type to make them easier to review. Once the reviewer(s) understands the specific CodeQL issue type being addressed, then evaluating each change to make sure that specific issue type is resolved with no behavioral changes is much easier.
> 
> Multiple authors across multiple commits seems ok to me. Splitting up the work can be faster. 

Absolutely. Strongly agree.

And I don't mind a single larger commit having several named authors, if it's a case of trying to ensure that attribution is given; for example when upstreaming code that has been developed downstream over some time.

In this case though, it rather served as an indicator that individual and separate mostly (but not exclusively) mechanical efforts had been squashed together at the submission stage. Complicating review.

> However, the patterns used to address the same type of issue type should be consistent.

mdkinney:
Based on feedback, this PR is being closed and a new PR to revert PR #11307 will be opened

PR #11406, opened 2025-08-11
Revert "SecurityPkg: CodeQL Fixes."

mdkinney:
# Description

This reverts commit ba6a8eb045aededbc1472cce7314ab1911f0ea1c.

PR https://github.com/tianocore/edk2/pull/11307 introduced a logic change that caused regressions in FV verification on some platforms. This PR is being reverted to restore the prior logic.

The Code QL fixes in https://github.com/tianocore/edk2/pull/11307 can be resubmitted without the logic change along with one commit for each type of Code QL issue being addressed.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## Integration Instructions

N/A

### DIFF
--- a/SecurityPkg/FvReportPei/FvReportPei.c
+++ b/SecurityPkg/FvReportPei/FvReportPei.c
@@ -71,10 +71,7 @@ InstallPreHashFvPpi (
             + HashSize;
 
   PreHashedFvPpi = AllocatePool (PpiSize);
-  if (PreHashedFvPpi == NULL) {
-    ASSERT (PreHashedFvPpi != NULL);
-    return;
-  }
+  ASSERT (PreHashedFvPpi != NULL);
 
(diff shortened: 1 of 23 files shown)


leiflindholm (approved):
I agree to include this in edk2-stable202508.

mdkinney:
@makubacki Given there are CodeQL failures with this revert, are there CodeQL config changes also required for this revert?

Or do I dismiss these issues as will not fix for this revert commit?

makubacki:
> @makubacki Given there are CodeQL failures with this revert, are there CodeQL config changes also required for this revert?
> 
> Or do I dismiss these issues as will not fix for this revert commit?

I think dismissing makes sense here. That will allow the code scanning check to pass but still produce the failures locally which we'd want so they can be fixed properly in the future.

PR #11410, opened 2025-08-12
OVMF ResetVector + #VC exceptions

kraxel:
# Description

The AMD SEV code in the OVMF ResetVector carefully installs and uninstalls the #VC exception handler, which is needed to use the CPUID instruction (traps with SEV-ES and SEV-SNP).  I'm wondering whenever it would be better to simply install the exception handler unconditionally.  Specifically this would allow the TDX detection code (which uses CPUID too) work properly when running in SEV-ES or SEV-SNP mode.

Background: the reset vector can handle being started in 32bit flat mode, which is how TDX defines the VM entry.  One (unintentional?) side effect is that this also assists in SEV/TDX detection.  Only when started in 32bit mode the reset vector checks for TDX, otherwise the check is skipped and only the SEV check runs.

Starting non-confidential VMs in 32-bit mode works too.
Starting SEV-SNP VMs in 32-bit mode does not work:  CPUID -> #VC (without handler) -> double-fault -> boom

The somewhat hackish test patch in this PR simply installs the #VC handler in the IsTdx function.  This makes 32-bit entry work for SEV-SNP.

Comments?

@tlendacky


tlendacky:
> The somewhat hackish test patch in this PR simply installs the #VC handler in the IsTdx function. This makes 32-bit entry work for SEV-SNP.
> 
> Comments?

I suppose it should be fine to do something along that line. Not sure the best way to approach it, maybe just a new function that you call before calling InitTdx at Main32?

kraxel:
> I suppose it should be fine to do something along that line.

Good.

> Not sure the best way to approach it, maybe just a new function that you call before calling InitTdx at Main32?

I'll have a look how to do that less hacky, and maybe cleanup the reset vector code a bit along the way.
(probably after kvm forum, i.e. september).

ardbiesheuvel:
No objections from me, but could you explain why the ability to boot SEV-SNP guests in 32-bit mode is a useful thing to have?

kraxel:
> could you explain why the ability to boot SEV-SNP guests in 32-bit mode is a useful thing to have?

It allows to boot OVMF without needing the mapping below 1M for the real mode entry point.  Also nice for consistency (32-bit entry works in all modes).

ardbiesheuvel:
> > could you explain why the ability to boot SEV-SNP guests in 32-bit mode is a useful thing to have?
> 
> It allows to boot OVMF without needing the mapping below 1M for the real mode entry point. Also nice for consistency (32-bit entry works in all modes).

I don't understand entirely what this means, but if you and Tom agree that this is a useful thing to have, I'll happily merge it.

tlendacky:
> > > could you explain why the ability to boot SEV-SNP guests in 32-bit mode is a useful thing to have?
> > 
> > 
> > It allows to boot OVMF without needing the mapping below 1M for the real mode entry point. Also nice for consistency (32-bit entry works in all modes).
> 
> I don't understand entirely what this means, but if you and Tom agree that this is a useful thing to have, I'll happily merge it.

Looks good to me. Not sure if it should be commented a bit more on the fact that setting up of the #VC handler is called from Main.asm but teardown is called from Flat32ToFlat64.asm, but that could possibly become stale in the future, so...

Booted all 4 types of VMs successfully: SVM, SEV, SEV-ES and SEV-SNP

PR #11411, opened 2025-08-12
Add SNP cache coherency vulnerability mitigation

tlendacky:
# Description

An SNP cache coherency vulnerability may require a mitigation to evict cache lines after memory has been validated. This series addresses that issue. The series consists of the mitigation, as well as support to detect if the mitigation is required (via CPUID) and not performing it if it is not necessary.

This addresses CVE-2024-36331.

- [ ] Breaking change?
- [x] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested by ensuring that the mitigation code was executed whenever an SNP guest was validating private memory. Additionally, the CPUID support was tested to ensure the mitigation code was not executed when the CPUID information indicated the mitigation was not necessary.

## Integration Instructions

N/A


kraxel (approved):

OvmfPkg/Library/BaseMemEncryptSevLib/X64/SnpPageStateChangeInternal.c
```diff
@@ -281,6 +314,14 @@ InternalSetPageState (
     //
     if (State == SevSnpPagePrivate) {
       AmdSvsmSnpPvalidate (Info);
+
+      if (MemEncryptSevSnpDoCoherencyMitigation ()) {
+        for (Index = 0; Index <= Info->Header.EndEntry; Index++) {
+          Address     = ((EFI_PHYSICAL_ADDRESS)Info->Entry[Index].GuestFrameNumber) << EFI_PAGE_SHIFT;
```
AshrafAliS: do we need to use LShiftU64

tlendacky: > do we need to use LShiftU64

Yes, it probably should. Let me update the patches.

OvmfPkg/Include/WorkArea.h
```diff
@@ -53,9 +53,12 @@ typedef struct _SEC_SEV_ES_WORK_AREA {
...
   // - ReceivedVc: Indicator that the VC handler was called. It is used during
   //   the SevFeature detection in OvmfPkg/ResetVector/Ia32/AmdSev.asm
+  // - CoherencySfwNo: Indicator that the SEV-SNP cache line evication
+  //   mitigation is not needed.
   //
-  UINT8     ReceivedVc : 1;
-  UINT8     Reserved1  : 7;
+  UINT8     ReceivedVc     : 1;
+  UINT8     CoherencySfwNo : 1;
+  UINT8     Reserved1      : 6;
```
ardbiesheuvel: Can we avoid the bit fields here? AFAIK the C spec does not define how these are mapped onto the underlying bits, and so we shouldn't be using them for fields that are accessed from assembler.

tlendacky: Ok, let me change over to using bitmasks.

OvmfPkg/Library/BaseMemEncryptSevLib/X64/SnpPageStateChangeInternal.c
```diff
@@ -213,6 +215,33 @@ PageStateChange (
...
+
+  /*
+   * For SEV guests, a read from the first/last cache-lines of a 4K page
+   * using the guest key is sufficient to cause a flush of all cache-lines
+   * associated with that 4K page without incurring all the overhead of a
+   * full CLFLUSH sequence.
+   */
+  for (PageIdx = 0; PageIdx < NumPages; PageIdx++) {
+    Val = Bytes[PageIdx * SIZE_4KB];
+    Val = Bytes[PageIdx * SIZE_4KB + SIZE_4KB - 1];
```
ardbiesheuvel, line 241: Could we avoid the kludge with `#ifdef` and `volatile` if we did something like

```
CompareMem (Bytes[PageIdx * SIZE_4KB], Bytes[PageIdx * SIZE_4KB + SIZE_4KB - 1], 1);
```
here? (Just a question, no need to change anything right away)

tlendacky: That should work as long as we can guarantee that the compiler doesn't optimize out the compare if we never check the return value. As long as the first and last byte of the page is read, it should be ok.

ardbiesheuvel: Yeah, I guess `volatile` is better in that regard. Do you still need the `#ifdef` and the attribute if you make the variable static?

tlendacky: Since the library is used in SEC and PEI, I don't believe I can make it static.

tlendacky: I could probably move the #define within the header file, e.g.:

In header:
```
#undef __UNUSED__
#if defined (__GNUC__) || defined (__clang__)
#define __UNUSED__ __attribute__ ((__unused__))
#else
#define __UNUSED__
#endif
```

In code
```
volatile UINT8 Val __UNUSED__;
```
If you think that looks better, I can give that a try.

ardbiesheuvel: Let's just go with what you have - we can move things around when more occurrences of this pattern emerge

PR #11413, opened 2025-08-13
Global/ArmFfaLib: Fix ArmFfaLibRun() caller couldn't get return arguments

LeviYeoReum:
When ArmFfaLibDirectMsgReq(2) is preempted, caller of these functions should resume it works via ArmFfaLibRun() and the secure partition will be return with FFA_DIRECT_MSG_RESP(2) with return arguments.

However, since ArmFfaLibRun() gets its return in its stack variable, So caller of ArmFfaLibRun() doesn't get the return arguments from secure partition.

To resolve this, add output parameter to ArmFfaLibRun() to receive return arguments.

Fixes: 5d1b38dd07c4 ("ArmPkg: Add ArmFfaLib used in Dxe driver")
Continuous-integration-options: PatchCheck.ignore-multi-package
Reported-by: Mariam Elshakfy <Mariam.Elshakfy@arm.com>


leiflindholm:
So to clarify: this is not causing a problem in current upstream code, but is problematic for out-of-tree code?

Are you wanting this to go into the stable tag? (we are currently in hard freeze, and that would delay the release)

LeviYeoReum:
> So to clarify: this is not causing a problem in current upstream code, but is problematic for out-of-tree code?
This is found from out-of-tree first, but it could happen upstream code 

if some TPM device using Tpm2DeviceLibFfa preempted by SPMC while it's operating.

> Are you wanting this to go into the stable tag? (we are currently in hard freeze, and that would delay the release)

Untill it was found, it have passed long time since ArmFfaLib and Tpm2DeviceLibFfa was merged.
So I think we can delay merge this after finish realese.

Thanks.

LeviYeoReum:
I don't know why PatchCheck.py return Error with:

```
MdeModulePkg/Library: fix ArmFfaLibRun() caller couldn't get ret-args
 * Subject line not in "package/component: description" format!
```

@leiflindholm, Would you let me know what I'm missing please?

Thanks

leiflindholm:
> @leiflindholm, Would you let me know what I'm missing please?

The enumeration of all modified packages in the subject line.
(Until #11321 gets merged and lets you substitute the keyword 'Global'.)

LeviYeoReum:
> > @leiflindholm, Would you let me know what I'm missing please?
> 
> The enumeration of all modified packages in the subject line. (Until #11321 gets merged and lets you substitute the keyword 'Global'.)

Thanks for sharing. at first, I'll change it with "Global" then.

LeviYeoReum:
@leiflindholm.
This is really weird, Though I try with:
"Global: xxx" and "Global/ArmFfaLib: xxx", The PatchCheck returns error message below:
```
 * Subject line not in "package/component: description" format!
 ```
 
 Any idea?

leiflindholm:
> @leiflindholm. This is really weird, Though I try with: "Global: xxx" and "Global/ArmFfaLib: xxx", The PatchCheck returns error message below:
> 
> ```
> The commit message format is not valid:
>  * Subject line not in "package/component: description" format!

Well, as I said, it will work once #11321 has been merged. But we're currently in hard freeze.

PR #11414, opened 2025-08-13
MdePkg/DxeIoLibCpuIo2: allow use from UEFI_DRIVER

MarsDoge:
# Description
On architectures without native port I/O instructions (e.g. non-x86 such as LoongArch), BaseIoLibIntrinsic cannot be used for in/out operations. Ps2KeyboardDxe depends on IoLib to access the PS/2 controller and must fall back to a protocol-based implementation.

Current DxeIoLibCpuIo2 uses:
  LIBRARY_CLASS = IoLib|DXE_DRIVER
This blocks UEFI_DRIVER modules from linking against it, even though they run in the DXE phase and can safely consume EFI_CPU_IO2_PROTOCOL.

Call flow comparison
--------------------
(x86 and other I/O-port capable arch):
  Ps2KeyboardDxe
    -> BaseIoLibIntrinsic (IoLib)
        ... -> native CPU I/O instructions (in/out)

(I/O-port-less arch):
  Ps2KeyboardDxe
    -> DxeIoLibCpuIo2 (IoLib, protocol-backed)
         ...-> CpuIo2Dxe (EFI_CPU_IO2_PROTOCOL provider)
            ......  -> MMIO / SoC-specific bus access

Solution (following DxeSmbusLib pattern)
--------
Relax LIBRARY_CLASS in DxeIoLibCpuIo2.inf:
This enables UEFI drivers to use protocol-based IoLib.

Reported-by: Chen Zhang <zhangchen@loongson.cn>
Signed-off-by: Dongyan Qian <qiandongyan@loongson.cn>
Tested-by: Yanbing Lv <lvyanbing@loongson.cn>

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions

### DIFF
--- a/MdePkg/Library/DxeIoLibCpuIo2/DxeIoLibCpuIo2.inf
+++ b/MdePkg/Library/DxeIoLibCpuIo2/DxeIoLibCpuIo2.inf
@@ -16,7 +16,7 @@
   FILE_GUID                      = 33D33BF3-349E-4768-9459-836A9F7558FB
   MODULE_TYPE                    = DXE_DRIVER
   VERSION_STRING                 = 1.0
-  LIBRARY_CLASS                  = IoLib|DXE_DRIVER
+  LIBRARY_CLASS                  = IoLib|DXE_DRIVER UEFI_DRIVER UEFI_APPLICATION
   CONSTRUCTOR                    = IoLibConstructor
 
 #


mdkinney:
One issue is that CPU I/O Protocol is not defined in UEFI Specification.  Any UEFI Driver that depends on that protocol is not UEFI conformant.  The UEFI Specification protocol that can provide access to I/O ports is the PCI Root Bridge I/O Protocol.  If that protocol is available in your platforms, then an IoLib instance that layers on top of PCI Root Bridge I/O Protocol could be type UEFI_DRIVER.

MarsDoge:
Thanks for your review, I understand your point about following the UEFI specification and using the PCI Root Bridge I/O Protocol whenever possible.
However, in our case, some UEFI_DRIVER modules are not attached to the PCI bus at all.
For example, **Ps2KeyboardDxe** is mounted on the ISA bus, and the driver only supports accessing I/O ports directly from the CPU's perspective (e.g., calling [`IoRead8`](https://github.com/tianocore/edk2/blob/d46aa46c8361194521391aa581593e556c707c6e/MdeModulePkg/Bus/Isa/Ps2KeyboardDxe/Ps2KbdCtrller.c#L724) directly).
In such scenarios, there is no PCI Root Bridge device to obtain the protocol from, and the only practical way to perform I/O is through the CPU I/O protocol.
This is why we want to allow UEFI_DRIVER to link against DxeIoLibCpuIo2, so that such non-PCI bus devices can still work properly on architectures without native port I/O instructions.

mdkinney:
If a CPU I/O protocol is available, then is there a reason an IoLib has not been implemented.

MarsDoge:
I understand that `IoLib` [BaseIoLibIntrinsic](https://github.com/tianocore/edk2/blame/d46aa46c8361194521391aa581593e556c707c6e/MdePkg/Library/BaseIoLibIntrinsic/BaseIoLibIntrinsic.inf#L60) is a basic library that performs port access from the CPU’s perspective using either Mmio or native I/O instructions. On unified-address architectures that do not support separate I/O instructions (such as AARCH64, RISC-V, and LoongArch), this library is typically not implemented.

In contrast, [CpuMmio2Dxe](https://github.com/tianocore/edk2/blame/d46aa46c8361194521391aa581593e556c707c6e/UefiCpuPkg/CpuMmio2Dxe/CpuMmio2Dxe.inf#L42) provides an implementation of EFI_CPU_IO2_PROTOCOL for such architectures, enabling port I/O access through MMIO. It uses `gEfiMdePkgTokenSpaceGuid.PcdPciIoTranslation` to translate the port address into the corresponding MMIO region, thereby achieving compatible port access.

Therefore, on non-x86 platforms, a series of mapped I/O ports can be accessed through MMIO in the DXE stage to drive a series of drivers, such as SIO, PS/2, EC and other chips.

mdkinney:
Nothing you describe would prevent an IoLib instance for your platform from being implemented. In fact, that same PCD could be used from the IoLib implementation.

MarsDoge:
I have a question. As you mentioned, is it necessary to implement the "IoLib (BaseIoLibIntrinsic)" operation?
If we ignore platform differences, does the IoLib link in the UEFI_DRIVER driver Ps2KeyboardDxe only support basic access? Shouldn't it also support protocol access to IoLib?

MarsDoge:
Hi @mdkinney , is there anything wrong with my understanding? 
Shouldn't UEFI_DRIVER also support the protocol when calling IoLib?
Looking forward to your reply, Tks !

MarsDoge:
@lgao4 Hi, Li ming, Do you have any comments? looking forward to your review.

lgao4:
> @lgao4 Hi, Li ming, Do you have any comments? looking forward to your review.

Mike mentions UEFI driver that are out of board, such as option rom uefi driver. It can't consume IO protocol. Your case is the driver built-into the platform bios image. So, they can consume IO protocol. But, your case can be DXE_DRIVER. Right?

MarsDoge:
@lgao4 Thanks your very much for your review.

What I expect is that UEFI_DRIVER can link IO Protocol `DxeIoLibCpuIo2 (IoLib) ` to call, rather than just supporting static linking of `BaseIoLibIntrinsic (IoLib)`; 
If this, UEFI_DRIVER does not need to consider the implementation of the platform IOLib and can be decoupled from the platform static library.

Example: MdePkg/Library/DxeSmbusLib/DxeSmbusLib.inf 
The UEFI_DIRVER driver only manages device-side communication and does not need to consider platform Smbus implementation.

So, I actually want UEFI_DRIVER can link IO Protocol `DxeIoLibCpuIo2 (IoLib) ` to call.

mdkinney:
The issue is that a diver that has a module type UEFI_DRIVER that uses dynamic interfaces (Protocol) is only allowed to use UEFI defined protocols.  Mapping to a protocol like UEFI PCI I/O Protocol is allowed.  Mapping to PI CPU I/O Protocols is not.  This is required so a UEFI Driver is it built independently can be integrated or loaded into an UEFI conformant platform.

If a mapping to the PI CPU I/O Protocols is allowed for UEFI Drivers, then any UEFI Driver that uses that mapping can not be integrated/loaded into a UEFI conformant platform that does not implement the PI Protocols.

My objection to this change is that is opens up UEFI Drivers to be build in a non UEFI conformant manner with no errors/warnings.

MarsDoge:
I understand your concern. I added the Mmio interface to the PS2 driver to solve this problem. https://github.com/tianocore/edk2/pull/11550
So I am going to close this PR, thanks for your review.

PR #11415, opened 2025-08-14
ArmVirtPkg: Add SATA driver to AArch64 QEMU Virt OVMF

JasonBrave:
# Description

Add SATA AHCI driver would make the QEMU ARM Virt virtual machine able to boot from SATA hard drive or CD-ROM attached to a PCIe SATA Controller.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Booted Debian 13 installer from a emulated SATA CD-ROM connected to AHCI controller with following command:

`qemu-system-aarch64 -M virt -cpu cortex-a72 -bios QEMU_EFI.fd -device qemu-xhci -device usb-kbd -device usb-tablet -m 16G -smp 8 -serial stdio -device ich9-ahci -drive file=debian-13.0.0-arm64-netinst.iso,id=abcd,media=cdrom,if=none -device ide-cd,drive=abcd -device ramfb`

## Integration Instructions

N/A


leiflindholm:
I'll let Ard comment once he gets back from holidays, but I'd just like to ask why?

The image already contains virtio block device as well as NVMe support, both of which take a lot less runtime effort from qemu than SATA.

JasonBrave:
It would be useful for two use cases,
1. VFIO PCIe SATA controller passthrough into VM, which one of the connected hard drive is the boot disk
2. Boot OS installer ISO if the guest OS does not have a Virtio SCSI driver to read from CD-ROM

ardbiesheuvel:
@kraxel any thoughts?

kraxel:
> @kraxel any thoughts?

I'm also wondering what the actual use case is ...

> It would be useful for two use cases,
>     1. VFIO PCIe SATA controller passthrough into VM, which one of the connected hard drive is the boot disk

Is this a theoretical thing or do you have a  real case needing this?

>     2. Boot OS installer ISO if the guest OS does not have a Virtio SCSI driver to read from CD-ROM

Do such guests exist in the first place?  Note that virtio-scsi is older than aarch64.

PR #11418, opened 2025-08-15
RedfishPkg/ConfigHandler: Implement the disconnection of Redfish tran…

changab:
…sport interface

Provide an event for the Redfish client to signal Redfish configuration handler driver to disconnect from the Redfish transport interface.

# Description

Redfish interface remains open when it is no longer needed. For example, the HTTP session remains open when Redfish communication is no longer needed. This results in the Redfish server having too many sessions open. Redfish server (e.g. BMC) throws out the failure of too many HTTP sessions.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

1. Boot the system with Redfish enabled.
2. Either boot to EFI shell or OS
3. Reboot the system
4. Check the Redfish server and the session remains open.

## Integration Instructions
 N/A

### DIFF
--- a/RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.c
+++ b/RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.c
@@ -13,9 +13,10 @@
 REDFISH_CONFIG_DRIVER_DATA  gRedfishConfigData;     // Only one Redfish service supported
                                                     // on platform for the BIOS
                                                     // Redfish configuration.
-EFI_EVENT                          gEndOfDxeEvent        = NULL;
-EFI_EVENT                          gExitBootServiceEvent = NULL;
-EDKII_REDFISH_CREDENTIAL_PROTOCOL  *gCredential          = NULL;
+EFI_EVENT                          gEndOfDxeEvent                 = NULL;
+EFI_EVENT                          gExitBootServiceEvent          = NULL;
+EFI_EVENT                          mgConfigHandlerDisconnectEvent = NULL;
+EDKII_REDFISH_CREDENTIAL_PROTOCOL  *gCredential                   = NULL;
 
 /**
   Callback function executed when the EndOfDxe event group is signaled.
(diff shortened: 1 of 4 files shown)


RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.c
```diff
@@ -67,6 +68,40 @@ RedfishConfigOnExitBootService (
...
+EFIAPI
+RedfishConfigDisconnection (
+  IN  EFI_EVENT  Event,
+  OUT VOID       *Context
+  )
+{
+  EFI_STATUS            Status;
+  EFI_REST_EX_PROTOCOL  *RestEx;
+
+  Status = gBS->HandleProtocol (
```
nicklela, line 88: I have two questions:

1. Is it possible that we have multiple RestEx instances? If yes, we may want to close all of them.
2. This event can be signaled many times since we don't close event. If this is expected, please check the debug level from DEBUG_ERROR to DEBUG_INFO.

igorkulchytskyy: I think overall we can have several RestEx instances.
But in this file we have that variable declared
REDFISH_CONFIG_DRIVER_DATA  gRedfishConfigData;     // Only one Redfish service supported
                                                    // on platform for the BIOS
                                                    // Redfish configuration.
And it said just one Redfish service (we have just one  RedfishServiceRestExHandleto install RestEx on)

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.c
```diff
@@ -160,6 +195,26 @@ RedfishConfigCommonInit (
...
+                  );
+  if (EFI_ERROR (Status)) {
+    gBS->CloseEvent (gEndOfDxeEvent);
+    gBS->CloseEvent (gExitBootServiceEvent);
+    gEndOfDxeEvent        = NULL;
+    gExitBootServiceEvent = NULL;
+    DEBUG ((DEBUG_ERROR, "%a: Fail to register Redfish transport interface disconnection event.", __func__));
+    return Status;
+  }
+
```
igorkulchytskyy, line 218: I think we should close mgConfigHandlerDisconnectEvent event in Unload function

changab: mgConfigHandlerDisconnectEvent is undefined when CreateEventEx returns failure. I set it to NULL for the failed condition.

PR #11420, opened 2025-08-16
MdeModulePkg/ArmFfaLib: add ArmFfaLibPartitionInfoGetRegs()

LeviYeoReum:
This patch adds ArmFfaLibPartitionInfoGetRegs() in ArmFfaLib
As ArmFfaLibPartitionInfoGetRegs() is added, Normal world ArmFfLib Rx/Tx buffer is not a madatory.
That's why return EFI_UNSUPPORTED while Rx/Tx buffer mapping in ArmFfaLib constructor 
can be consider as valid return If ARM_FFA_PARTITION_INFO_GET_REGS is supported by SPMC.

Patch #1 adds ArmFfaLibPartitionInfoGetRegs()
Patch #2 considers EFI_UNSUPPORTED while Rx/Tx buffer mapping in constrcutor as valid return
Patch #3 add helper function named ArmFfaLibGetPartitionInfo().
Patch #4-#6 applies ArmFfaLibGetPartitionInfo() in each modules.


MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -530,6 +530,126 @@ ArmFfaLibPartitionInfoGet (
...
+**/
+EFI_STATUS
+EFIAPI
+ArmFfaLibPartitionInfoGetRegs (
+  IN EFI_GUID                 *ServiceGuid,
+  IN OUT UINT32               *PartDescCount,
+  OUT EFI_FFA_PART_INFO_DESC  *PartDesc OPTIONAL
+  )
+{
+ #if defined (MDE_CPU_AARCH64)
```
kuqin12: May I know what this preprocessor is for?

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -530,6 +530,126 @@ ArmFfaLibPartitionInfoGet (
...
+    }
+
+    CurIdx = ((FfaArgs.Arg2 >> FFA_PART_INFO_METADATA_CURRENT_IDX_SHIFT) &
+              FFA_PART_INFO_IDX_MASK);
+    Tag = ((FfaArgs.Arg2 >> FFA_PART_INFO_METADATA_TAG_SHIFT) &
+           FFA_PART_INFO_TAG_MASK);
+    DescSize = ((FfaArgs.Arg2 >> FFA_PART_INFO_METADATA_DESC_SIZE_SHIFT) &
+                FFA_PART_INFO_DESC_SIZE_MASK);
+
+    if (DescSize > sizeof (EFI_FFA_PART_INFO_DESC)) {
```
kuqin12: Is this right? We are truncating the descriptor regardlessly?

LeviYeoReum: No problem for truncate In compatible version of FF-A, 
Partition information's structure seems not over the size boundary 3 regsiters.
In this boundary. when some new field added which `EFI_FFA_PART_INFO_DESC` doesn't knwo,
by copying the as much as `EFI_FFA_PART_INFO_DESC` I think it can use without problem while using the information declared in `EFI_FFA_PART_INFO_DESC`.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: This ABI is only supported for 64 bits not 32 bits.
Without this preprocessor, while we manipulate bit operation over 32 bits with FfaArgs directly.
CI targeted with 32bit architecture complains.
(I think this is better than add simple temporary UINT64 while manipulate or extract the imformation from bit over 32bits -- seems more dirty).

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
```diff
@@ -124,6 +126,26 @@ ArmFfaPeiLibConstructor (
   if (RxTxBufferHob == NULL) {
     Status = ArmFfaLibRxTxMap ();
     if (EFI_ERROR (Status)) {
+      if (Status != EFI_UNSUPPORTED) {
+        return Status;
```
kuqin12, line 130: nit: Would you need some debug prints here?

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.c
```diff
@@ -72,6 +74,26 @@ ArmFfaSecLibConstructor (
 
   Status = ArmFfaLibRxTxMap ();
   if (EFI_ERROR (Status)) {
+    if (Status != EFI_UNSUPPORTED) {
+      return Status;
```
kuqin12, line 78: Same here?

MdePkg/Include/Library/ArmFfaLib.h
```diff
@@ -378,4 +378,27 @@ ArmFfaLibMsgSendDirectReq2 (
...
+
+  @retval EFI_SUCCESS
+  @retval EFI_UNSUPPORTED
+  @retval EFI_INVALID_PARAMETER
+  @retval Other                       Error
+
+**/
+EFI_STATUS
+EFIAPI
+ArmFfaLibGetPartitionInfo (
```
kuqin12, line 423: It really seems that this commit should be broken up to separate commits, with each covering a different component.

LeviYeoReum: Hmm. I think adding a new interface and applies into one commit seems to be better for bisect...
Am I missing something?

Thanks

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
LeviYeoReum: I think ArmFfaLibRxTxMap() already prints the error. So it seems fine without any additional error log in here.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.c
LeviYeoReum: I think ArmFfaLibRxTxMap() already prints the error. So it seems fine without any additional error log in here.

MdePkg/Include/Library/ArmFfaLib.h
kuqin12: Hmm, I think it will be better to add the new interface in one commit. Then onboard the components one by one.

LeviYeoReum: Okay. I've separated the patch. Thanks.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
samimujawar: Should DescSize be used for bounds checking in the for loop below?

shenghuliu: This DescSize is from the other party which should also follow the Specs, if wrong size is returned from smc call, it should be logged or asserted. Quietly change this size and keep going may cause the undefined behavior.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -530,6 +530,126 @@ ArmFfaLibPartitionInfoGet (
...
+EFIAPI
+ArmFfaLibPartitionInfoGetRegs (
+  IN EFI_GUID                 *ServiceGuid,
+  IN OUT UINT32               *PartDescCount,
+  OUT EFI_FFA_PART_INFO_DESC  *PartDesc OPTIONAL
+  )
+{
+ #if !defined (MDE_CPU_AARCH64)
+  return EFI_UNSUPPORTED;
+ #else
```
shenghuliu: for better readability, can #else be changed to #endif and remove #endif on line 650 ?

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -530,6 +530,126 @@ ArmFfaLibPartitionInfoGet (
...
+    }
+
+    CurIdx = ((FfaArgs.Arg2 >> FFA_PART_INFO_METADATA_CURRENT_IDX_SHIFT) &
+              FFA_PART_INFO_IDX_MASK);
+    Tag = ((FfaArgs.Arg2 >> FFA_PART_INFO_METADATA_TAG_SHIFT) &
+           FFA_PART_INFO_TAG_MASK);
+    DescSize = ((FfaArgs.Arg2 >> FFA_PART_INFO_METADATA_DESC_SIZE_SHIFT) &
+                FFA_PART_INFO_DESC_SIZE_MASK);
+
+    if (DescSize > sizeof (EFI_FFA_PART_INFO_DESC)) {
```
shenghuliu: This DescSize is from the other party which should also follow the Specs, if wrong size is returned from smc call, it should be logged or asserted. Quietly change this size and keep going may cause the undefined behavior.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -530,6 +530,126 @@ ArmFfaLibPartitionInfoGet (
...
+
+    Count = ((FfaArgs.Arg2 >> FFA_PART_INFO_METADATA_LAST_IDX_SHIFT) &
+             FFA_PART_INFO_IDX_MASK) + 1;
+
+    /*
+     * Return Total number of partitions related ServiceGuid.
+     */
+    if ((PartDesc == NULL) || (*PartDescCount == 0)) {
+      *PartDescCount = Count;
+      return EFI_SUCCESS;
```
shenghuliu: return EFI_BUFFER_TOO_SMALL seems to be the common practice though EFI_SUCCESS also works as commented in function header.

if return EFI_BUFFER_TOO_SMALL, please modify comment in function header.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: No. otherwise becuase of UINTN size, build failed for bit operation.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: This is not good for according to FF-A spec where there is no correspond return "BUFFER_TOO_SMALL".
So, I remain as it is.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: As I said other comments, This is for scalarablity. as sepc updated, "EFI_FFA_PART_INFO_DESC" can be increased in the range of 3 registers size.
So for the compatiblity the partition which compiled with size of lower-version can only be used the
"old version" of partition info.
So, it doesn't need to change ASSERT.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
shenghuliu: can you help me to understand why the proposal will impact UINTN size?  Here is the proposal:

#if !defined (MDE_CPU_AARCH64)
  return EFI_UNSUPPORTED;
 #endif

// now we are in the MDE_CPU_AARCH64 world.
......sizeof(UINTN) = 8

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
shenghuliu: assume in future, vm old EFI_FFA_PART_INFO_DESC still be 3x8, and spmc new EFI_FFA_PART_INFO_DESC could be increased to 6x8. when Regs +=3 gets executed, the next DESC copy source will start with FfaArgs.Arg6 which is still in the previous spmc new EFI_FFA_PART_INFO_DESC, do I miss anything here?

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
shenghuliu: how does this have anything to do with FF-A Specs?  EFI_SUCCESS is hardcoded here. if you can hardcode EFI_SUCCESS here, why not EFI_BUFFER_TOO_SMALL?   At this point, FF-A return conversion has already been done before on line 600.  Exercising UEFI common practice will keep such APIs consistent.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: Please See line 596, `FFA_PART_INFO_METADATA_TAG_SHIFT` is 32 bit
and If UINTN with `FFA_PART_INFO_METADATA_TAG_SHIFT` shift, it makes a build error on 32 bit architecture.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: Please check a `EfiStatusToFfaStatus()` or `FfaStatusToEfiStatus()` for convertsion of EFI_STATUS and FFA_Status. The ArmFfaLibXXX try to make a correspondant return with the specification and
EFI_BUFFER_TOO_SMALL doesn't have some proper return for FFA_Status.

Also, This is implementation based what should do when user pass a PartCount == 0.
So I think it would be better to remain it as it is.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: > assume in future, vm old EFI_FFA_PART_INFO_DESC still be 3x8, and spmc new EFI_FFA_PART_INFO_DESC could be increased to 6x8. when Regs +=3 gets executed, the next DESC copy source will start with FfaArgs.Arg6 which is still in the previous spmc new EFI_FFA_PART_INFO_DESC, do I miss anything here?

In this case, we coudln't make a compatible with SPMC FF-A version.
This is common case, even some FF-A version doesn't coherent to v1.2 , we need to make workable with later version (maybe we need to patch for later when the later spec update).

However, Consider the "Compatible" spec update and EFI_FFA_PART_INFO is increased in bondary of 3 Regs size.
Then we don't need to recompile the UEFI since it's "compatible" with later SPMC version.

That's why It's not a problem for this truncation and doesn't need to check the exact size match for each version.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -530,6 +530,126 @@ ArmFfaLibPartitionInfoGet (
   return Status;
 }
 
+/**
+  Get Partition info via registers.
+  This function is supported by aarch64 only.
+
+  @param [in]       ServiceGuid       Service guid.
+  @param [in, out]  PartDescCount     Return number of partition info realted to
```
vishalo: There is a typo here - "realted" ...

MdePkg/Include/Library/ArmFfaLib.h
```diff
@@ -265,6 +265,31 @@ ArmFfaLibPartitionInfoGet (
   OUT UINT32    *Size OPTIONAL
   );
 
+/**
+  Get Partition info via registers.
+  This function is supported by aarch64 only.
+
+  @param [in]       ServiceGuid       Service guid.
+  @param [in, out]  PartDescCount     Return number of partition info realted to
```
vishalo: There is a typo here "realted".

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: Thanks! I'll fix it.

MdePkg/Include/Library/ArmFfaLib.h
LeviYeoReum: Thanks! I'll fix it.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: > This DescSize is from the other party which should also follow the Specs, if wrong size is returned from smc call, it should be logged or asserted. Quietly change this size and keep going may cause the undefined behavior.

About this, Please check about other my comments.

Thanks

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
shenghuliu: current EFI_FFA_PART_INFO size is already 24 which is equal 3 Regs size, how EFI_FFA_PART_INFO could be increased and still be "in boundary of 3 Regs size"?

LeviYeoReum: Ah you're right.
I've missed that one. I'll change it to exact comparsion.

Thanks

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
vishalo: I don't think this has any bearing on FFA Spec - returning EFI_BUFFER_TOO_SMALL allows callers to allocate memory, and handle error case.  Could you please have interface return EFI_BUFFER_TOO_SMALL ?

vishalo (changes requested):

kuqin12:
Not entirely related to this PR, but I was thinking about the usage of this library earlier. And it seems that the main churn for this library class comes from the Rx/Tx buffer, i.e. whether is supported, whether we have it in the right environment, whether we can have the correct framework to support the memory allocation, whether we need to unregister it when handing off to the OS, etc..

I would think for a large collection of basic FFA functionalities, it is really monitor calls + interpretation. We probably can keep it simple, then we can use it anywhere, link to all modules. But for Tx/Rx, which is like a global, and will need memory usage, and pass on the information of those buffers, it might be better to be handled separately.

TL;DR
Would it make better sense to create a Rx/Tx buffer library to specifically support `RXTX_` related function IDs? To take one step further, would it be necessary to control the framework related information (Rx/Tx location, discovered Partition Info, etc.) through protocol/PPI in a module (i.e. PEI, DXE, STMM, etc.)

LeviYeoReum:
> Not entirely related to this PR, but I was thinking about the usage of this library earlier. And it seems that the main churn for this library class comes from the Rx/Tx buffer, i.e. whether is supported, whether we have it in the right environment, whether we can have the correct framework to support the memory allocation, whether we need to unregister it when handing off to the OS, etc..
> 
> I would think for a large collection of basic FFA functionalities, it is really monitor calls + interpretation. We probably can keep it simple, then we can use it anywhere, link to all modules. But for Tx/Rx, which is like a global, and will need memory usage, and pass on the information of those buffers, it might be better to be handled separately.
> 
> TL;DR Would it make better sense to create a Rx/Tx buffer library to specifically support `RXTX_` related function IDs? To take one step further, would it be necessary to control the framework related information (Rx/Tx location, discovered Partition Info, etc.) through protocol/PPI in a module (i.e. PEI, DXE, STMM, etc.)

Hmm. I'm not sure about "How to gather Rx/Tx related API".
In case of GetPartitionInfo ABI, yes that's clear but What about direct message kind of API?
Actaully these kind of API can be differ from the platform even the services whether it uses Rx/Tx buffer together or not.
That's why at the first time, It seems too much to have separate library for Rx/Tx buffer I've separated it just with file.

Well. If there'll be addition of ABI in FF-A using Rx/Tx buffer (not related MEMORY management), it seems meaningful to have separation However, I don't think it's valuable to have it now.

Thanks.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: > I don't think this has any bearing on FFA Spec - returning EFI_BUFFER_TOO_SMALL allows callers to allocate memory, and handle error case. Could you please have interface return EFI_BUFFER_TOO_SMALL ?

No. In here intension is returns "EFI_SUCCESS" for getting count only.
If not the user always check the with (EFI_ERROR(Status) && Status == EFI_BUFFER_TOO_SMALL).
Since the user specifiy "NULL", it's much better to return. this kind of pattern is usally used to get the "output buffer size"

If you said about only PartDesc != NULL && DescCount == 0,
It depends on the implementation.
Someone can do not anything since it doesn't copy anything or
Someone can said it's a error DescCount couldn't be 0 or
Someone can define this method as a way of "Get Partition Count number"

I've chosed the thrid stragy since the ff-a driver in kernel side accept the same straegy
and the keep the return value compatible with EFI_STATUS and FFA_STATUS in this API that is good.

kuqin12:
> Hmm. I'm not sure about "How to gather Rx/Tx related API". In case of GetPartitionInfo ABI, yes that's clear but What about direct message kind of API? Actaully these kind of API can be differ from the platform even the services whether it uses Rx/Tx buffer together or not. That's why at the first time, It seems too much to have separate library for Rx/Tx buffer I've separated it just with file.
> 
> Well. If there'll be addition of ABI in FF-A using Rx/Tx buffer (not related MEMORY management), it seems meaningful to have separation However, I don't think it's valuable to have it now.
> 
> Thanks.

Your response essentially means we can already do the separation :)

The implicit usage of Rx/Tx will still need to get Rx/Tx buffer information, which I think better fits into its own component. Whether that component is a set of modules (PEIM/DXE driver/Stmm) or a set of libraries for different phases, that is debatable.

I will not derail this PR further. But I might take a crack at it to see how much effort we will need to get that separated.

kuqin12:
> Patch #1 adds ArmFfaLibPartitionInfoGetRegs() Patch #2 considers EFI_UNSUPPORTED while Rx/Tx buffer mapping in constrcutor as valid return Patch #3 add helper function named ArmFfaLibGetPartitionInfo().
> 
The PR description needs to be updated.

LeviYeoReum:
> > Patch #1 adds ArmFfaLibPartitionInfoGetRegs() Patch #2 considers EFI_UNSUPPORTED while Rx/Tx buffer mapping in constrcutor as valid return Patch #3 add helper function named ArmFfaLibGetPartitionInfo().
> 
> The PR description needs to be updated.

Updated. Thanks!

LeviYeoReum:
> > Hmm. I'm not sure about "How to gather Rx/Tx related API". In case of GetPartitionInfo ABI, yes that's clear but What about direct message kind of API? Actaully these kind of API can be differ from the platform even the services whether it uses Rx/Tx buffer together or not. That's why at the first time, It seems too much to have separate library for Rx/Tx buffer I've separated it just with file.
> > Well. If there'll be addition of ABI in FF-A using Rx/Tx buffer (not related MEMORY management), it seems meaningful to have separation However, I don't think it's valuable to have it now.
> > Thanks.
> 
> Your response essentially means we can already do the separation :)
> 
> The implicit usage of Rx/Tx will still need to get Rx/Tx buffer information, which I think better fits into its own component. Whether that component is a set of modules (PEIM/DXE driver/Stmm) or a set of libraries for different phases, that is debatable.
> 
> I will not derail this PR further. But I might take a crack at it to see how much effort we will need to get that separated.

Well. I'll also take a time to think about it more (perhaps with Library? since for use in SEC phase too?)
Thanks to point this!

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -530,6 +530,127 @@ ArmFfaLibPartitionInfoGet (
...
+                FFA_PART_INFO_DESC_SIZE_MASK);
+
+    if (DescSize != sizeof (EFI_FFA_PART_INFO_DESC)) {
+      return EFI_ACCESS_DENIED;
+    }
+
+    Regs = &FfaArgs.Arg3;
+    for (Idx = 0; Idx < (CurIdx - StartIdx) + 1; Idx++) {
+      CopyMem (PartDesc, Regs, DescSize);
+      Regs += 3;
```
shenghuliu: if use sizeof (EFI_FFA_PART_INFO_DESC)/sizeof( UINTN) to replace the hardcoded 3, it will avoid modifying this function in case EFI_FFA_PART_INFO_DESC expands in future.

LeviYeoReum: That's good idea.
Thanks.

LeviYeoReum:
@samimujawar  and @kuqin12.

If you don't have anymore comment, Could you review and merge this patch series?

Thanks

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
samimujawar: Now that the support for AArch32 is dropped, I think we no longer need the check at line 559.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
samimujawar: I think EFI_BUFFER_TOO_SMALL should be returned when *PartDescCount == 0. The reason being the name of this function is ArmFfaLibPartitionInfoGetRegs(), which means the intention is to get the Partition Info. So when this function is used to identify the size of the buffer we should return EFI_BUFFER_TOO_SMALL. The return value of EFI_SUCCESS means that the Partition Info has been retrieved successfully, which is not really the case in this scenario.

Had this function name been ArmFfaLibGetPartitionCount() then EFI_SUCCESS would have been acceptable.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -530,6 +530,127 @@ ArmFfaLibPartitionInfoGet (
...
+      CopyMem (PartDesc, Regs, DescSize);
+      Regs += (sizeof (EFI_FFA_PART_INFO_DESC) / sizeof (FfaArgs.Arg0));
+      PartDesc++;
+      if (--DescCount == 0) {
+        break;
+      }
+    }
+
+    PrevIdx = CurIdx;
+  } while (CurIdx < (Count -1) && (DescCount != 0));
```
samimujawar: Please run uncrustify.
There should have been a space between 'Count - 1'.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -530,6 +530,127 @@ ArmFfaLibPartitionInfoGet (
...
+
+    CurIdx = ((FfaArgs.Arg2 >> FFA_PART_INFO_METADATA_CURRENT_IDX_SHIFT) &
+              FFA_PART_INFO_IDX_MASK);
+    Tag = ((FfaArgs.Arg2 >> FFA_PART_INFO_METADATA_TAG_SHIFT) &
+           FFA_PART_INFO_TAG_MASK);
+    DescSize = ((FfaArgs.Arg2 >> FFA_PART_INFO_METADATA_DESC_SIZE_SHIFT) &
+                FFA_PART_INFO_DESC_SIZE_MASK);
+
+    if (DescSize != sizeof (EFI_FFA_PART_INFO_DESC)) {
+      return EFI_ACCESS_DENIED;
```
samimujawar: Should this be EFI_BAD_BUFFER_SIZE or EFI_INVALID_PARAMETER. In either case, please document the error type in the function header.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: This is already result of uncrustify. But I'm bit of suprised that wasn't caught.
Anyway I'll fix this. Thanks!

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: Okay. I'll change it.
Thanks

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: Okay. I'll fix this. Thanks!

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
ardbiesheuvel: Unresolving as this has not been changed.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
ardbiesheuvel: Same

MdePkg/Include/Library/ArmFfaLib.h
ardbiesheuvel: Same. Please don't resolve these items if you are not posting the fixes.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.c
```diff
@@ -72,6 +74,26 @@ ArmFfaSecLibConstructor (
...
+    Status = ArmFfaLibGetFeatures (
+               ARM_FID_FFA_PARTITION_INFO_GET_REGS,
+               0x00,
+               &Property1,
+               &Property2
+               );
+    if (!EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_INFO, "%a Rx/Tx buffer doesn't support.\n", __func__));
+    }
+
```
ardbiesheuvel, line 96: You are adding the exact same code 3 times. Please refactor this.

MdeModulePkg/Library/ArmFfaLib/ArmFfaDxeLib.c
```diff
@@ -75,6 +75,8 @@ ArmFfaDxeLibConstructor (
   EFI_STATUS                 Status;
   EFI_HOB_GUID_TYPE          *RxTxBufferHob;
   ARM_FFA_RX_TX_BUFFER_INFO  *BufferInfo;
+  UINTN                      Property1;
```
ardbiesheuvel, line 78: Please fix the commit log:

- inconstructor is not a word
- 'doesn't supported' is incorrect grammar.
-

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -950,6 +950,85 @@ ArmFfaLibCommonInit (
   return EFI_SUCCESS;
 }
 
+/**
```
ardbiesheuvel, line 953: Please fix the commit log. I don't understand the first sentence.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.c
LeviYeoReum: I don't think such refactoring is necessary.
This is because the behavior of the ArmFfaLib constructor differs for each phase,
and in the case of the Rx/Tx buffer check, it only involves calling two APIs.
Moreover, the modified part specifically handles the case where
ArmFfaLibRxTxMap() is not supported.
As described in the comment, it simply checks whether
FFA_GET_PART_INFO_REGS() is supported,
so rewriting or introducing a common function for this purpose
doesn't seem necessary.

MdeModulePkg/Library/ArmFfaLib/ArmFfaDxeLib.c
LeviYeoReum: Thanks. I'll fix.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: Thanks. I'll fix.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.c
LeviYeoReum: Anyway I have a plan to refactor ArmFfaLib overall.
At that time. let me do refactor for this not in this commit.

Thanks.

LeviYeoReum:
@samimujawar, @ardbiesheuvel 

Gentle ping for this.

MdePkg/Include/Library/ArmFfaLib.h
ardbiesheuvel: Fix the typo please

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
ardbiesheuvel: Again, please don't resolve conversations unless you actually fix the issue.

ardbiesheuvel (changes requested):

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: Sorry to make a noise.

while I updating the branch, there was a mistake with the old version.

MdePkg/Include/Library/ArmFfaLib.h
LeviYeoReum: Sorry to make a noise.
I've updated. 

Thanks

LeviYeoReum:
Gentle ping in case of forgotten:
@samimujawar, @ardbiesheuvel

ardbiesheuvel (approved):

PR #11430, opened 2025-08-20
Add CLANGPDB DLink flags configuration to build options

deepak5x:
# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions


CryptoPkg/CryptoPkg.dsc
```diff
@@ -329,6 +329,7 @@
       TlsLib|CryptoPkg/Library/TlsLib/TlsLib.inf
     <BuildOptions>
       MSFT:*_*_*_DLINK_FLAGS     = /ALIGN:4096 /FILEALIGN:4096 /SUBSYSTEM:CONSOLE
+      CLANGPDB: *_CLANGPDB_*_DLINK_FLAGS = /ALIGN:4096 /FILEALIGN:4096 /SUBSYSTEM:CONSOLE
```
os-d: This is not defined correctly, `CLANGPDB` is not a family: https://github.com/tianocore/edk2/blob/d46aa46c8361194521391aa581593e556c707c6e/BaseTools/Conf/tools_def.template#L233-L235. I'm not sure how the build system will actually treat this.

CLANGPDB's family is GCC: https://github.com/tianocore/edk2/blob/d46aa46c8361194521391aa581593e556c707c6e/BaseTools/Conf/tools_def.template#L1917. The GCC rules with `*_*` in the front should cover CLANGPDB and you would instead specify `*_CLANGPDB` to cover that.

Now, this is actually more complicated than that because it depends on whether you are building with CLANGPDB on Linux (where you would want the GCC flags) or on Windows (where you'd want the MSFT flags). But the way the flags are currently defined in tools_def.template are the MSFT flags. So I think CLANGPDB will only work on Windows today, unless I'm missing something?

@mdkinney @ardbiesheuvel do you have thoughts on having the CLANGPDB family be MSFT? Or, should CLANGPDB (and so maybe CLANGDWARF, too) have its own family now that it is seeing greater usage?

os-d:
Please include a PR description for what you are trying to do

CryptoPkg/CryptoPkg.dsc
mdkinney: CLANGPDB works under Linux when building firmware modules to generate a UEFI conformant PE/COFF image with debug symbol information in a PDB format.

CLANGPDB can not be used with a build that needs to generate a Linux ELF formatted executable such as EmulatorPkg and Unit Test applications.

os-d: CLANGPDB works on Linux even though the DLINK flags have /ALIGN and other MSVC linker flags?

mdkinney: > CLANGPDB works on Linux even though the DLINK flags have /ALIGN and other MSVC linker flags?

Yes

ardbiesheuvel: If CLANGPDB understands all the `/FOO` style options, it makes sense to treat it as `MSFT` rather than `GCC`, considering that it produces PE/COFF natively. The fact that it is Clang based should be immaterial in that case, and CLANGDWARF should remain in the `GCC` family IMO.

mdkinney: It is more complex than that.  clang compiler uses GCC style options.  clang linker that can produce PDB symbols uses MSFT style options.  The clang-cl compiler uses MSFT style options, but the edk2 does not use that version of the clang compiler.

ardbiesheuvel: Ok, then I'd argue that this puts CLANGPDB into a distinct family of its own.

mdkinney: CLANGPDB already has its own BUILDRULEFAMILY declared in tools_def.template to handle these scenarios.  I do not think CLANGPDB need to have its own FAMILY.  There is already use of CLANGPDB BUILDRULEFAMILY in other components.

```
*_CLANGPDB_*_*_FAMILY                = GCC
*_CLANGPDB_*_*_BUILDRULEFAMILY       = CLANGPDB
```

CLANGDWARF and XCODE5 do as well

```
*_CLANGDWARF_*_*_FAMILY             = GCC
*_CLANGDWARF_*_*_BUILDRULEFAMILY    = CLANGDWARF

*_XCODE5_*_*_FAMILY            = GCC
*_XCODE5_*_*_BUILDRULEFAMILY   = XCODE
```

ardbiesheuvel: Ah yes, I had forgotten about `BUILDRUILEFAMILY` - that should be sufficient in principle, it only makes it a bit tedious to specify overrides for `DLINK` flags as they will always have to be specified separately for CLANGPDB, but I guess that is simply a consequence of its hybrid nature.

nate-desimone (commented):
@deepak5x - Please run the patch check tool and make sure that all style/formatting issues are fixed before submitting a PR. Also per @os-d's comment, please add a description to the PR.

The actual technical content here looks good to me. Essentially whenever a module chooses a specific PE/COFF file alignment option when building with Visual C++ it makes sense to replicate that build option for CLANGPDB since the link phase for those two compilers follow the same high-level design.

CryptoPkg/CryptoPkg.dsc
mdkinney: The one change I suggest is to simplify the lines added.

Instead of listing CLANGPDB in both family and tool chain tag fields:

```
CLANGPDB: *_CLANGPDB_*_DLINK_FLAGS = /ALIGN:256 /FILEALIGN:4096
```

Simplify to:

```
CLANGPDB: *_*_*_DLINK_FLAGS = /ALIGN:256 /FILEALIGN:4096
```

This comment applies to all the changes in this PR

deepak5x: Updated

deepak5x:
> @deepak5x - Please run the patch check tool and make sure that all style/formatting issues are fixed before submitting a PR. Also per @os-d's comment, please add a description to the PR.
> 
> The actual technical content here looks good to me. Essentially whenever a module chooses a specific PE/COFF file alignment option when building with Visual C++ it makes sense to replicate that build option for CLANGPDB since the link phase for those two compilers follow the same high-level design.

As changes are in multiple packages and if tried to add in "package,package: description" format, it cross description length.
We get below error:
"
CryptoPkg,SecurityPkg,EmulatorPkg,OvmfPkg,StandaloneMmPkg/BuildOptions : Add CLANGPDB DLINK_FLAGS flags to build options
 * First line of commit message (subject line) is too long (120 > 91)."

SecurityPkg/SecurityPkg.dsc
```diff
@@ -433,5 +433,6 @@
 [BuildOptions]
    MSFT:*_*_IA32_DLINK_FLAGS = /ALIGN:256
    MSFT:*_*_IA32_DLINK_XIPFLAGS = /ALIGN:256
+   CLANGPDB: *_*_*_DLINK_FLAGS = /ALIGN:256
```
mdkinney: The MSFT lines are only for IA32.  Why is CLANGPDB using * instead of IA32 to match MSFT?

EmulatorPkg/EmulatorPkg.dsc
```diff
@@ -532,6 +532,7 @@
   GCC:DEBUG_CLANGPDB_*_CC_FLAGS =-O0 -Wno-unused-command-line-argument -Wno-incompatible-pointer-types -Wno-enum-conversion -Wno-incompatible-pointer-types -Wno-sometimes-uninitialized -Wno-constant-conversion -Wno-main-return-type
 
   MSFT:*_*_*_DLINK_FLAGS     = /ALIGN:4096 /FILEALIGN:4096
+  CLANGPDB: *_*_*_DLINK_FLAGS = /ALIGN:256 /FILEALIGN:4096
```
mdkinney: CLANGPDB should only be set within WIN_HOST_BUILD.  EmulatorPkg does not support CLANGPDB with in non-Windows OS build envs.

EmulatorPkg/EmulatorPkg.dsc
```diff
@@ -540,6 +541,7 @@
   # CLANGPDB tool chain depends on WIN_HOST_BUILD flag to generate the windows application.
   #
   GCC:*_CLANGPDB_*_DLINK_FLAGS     = /ALIGN:4096 /FILEALIGN:4096
+  CLANGPDB: *_*_*_DLINK_FLAGS = /ALIGN:256 /FILEALIGN:4096
```
mdkinney: This is a clean up of line 543, so this new line should replace line 543.

CryptoPkg/CryptoPkg.dsc
```diff
@@ -641,6 +649,7 @@
     <BuildOptions>
       MSFT:*_*_IA32_DLINK_FLAGS = /ALIGN:64
       MSFT:*_*_X64_DLINK_FLAGS  = /ALIGN:256
+      CLANGPDB: *_*_*_DLINK_FLAGS = /ALIGN:256
```
mdkinney: Should add 2 lines.  One for CLANGPDB for IA32 and one for CLANGPDB  for X64.

CryptoPkg/CryptoPkg.dsc
```diff
@@ -654,6 +663,7 @@
     <BuildOptions>
       MSFT:*_*_IA32_DLINK_FLAGS = /ALIGN:4096
       MSFT:*_*_X64_DLINK_FLAGS  = /ALIGN:4096
+      CLANGPDB: *_*_*_DLINK_FLAGS = /ALIGN:256
```
mdkinney: Should be 2 lines.  One for IA32 and one for X64 and the alignment should be 4096 to match MSFT

CryptoPkg/CryptoPkg.dsc
```diff
@@ -601,6 +608,7 @@
       MSFT:*_*_IA32_DLINK_FLAGS = /ALIGN:4096
       MSFT:*_*_X64_DLINK_FLAGS  = /ALIGN:4096
       GCC:*_*_AARCH64_DLINK_XIPFLAGS = -z common-page-size=0x1000
+      CLANGPDB: *_*_*_DLINK_FLAGS = /ALIGN:4096
```
mdkinney: Should break up into 2 lines for IA32 and X64 to match MSFT settings

CryptoPkg/CryptoPkg.dsc
```diff
@@ -587,6 +593,7 @@
     <BuildOptions>
       MSFT:*_*_IA32_DLINK_FLAGS = /ALIGN:64
       MSFT:*_*_X64_DLINK_FLAGS  = /ALIGN:256
+      CLANGPDB: *_*_*_DLINK_FLAGS = /ALIGN:256
```
mdkinney: Split into 2 lines with different alignments for IA32 and X64

CryptoPkg/CryptoPkg.dsc
```diff
@@ -547,6 +552,7 @@
       MSFT:*_*_IA32_DLINK_FLAGS = /ALIGN:4096
       MSFT:*_*_X64_DLINK_FLAGS  = /ALIGN:4096
       GCC:*_*_AARCH64_DLINK_XIPFLAGS = -z common-page-size=0x1000
+      CLANGPDB: *_*_*_DLINK_FLAGS = /ALIGN:4096
```
mdkinney: Split into 2 lines for IA32 and X64 to match MSFT

CryptoPkg/CryptoPkg.dsc
```diff
@@ -533,6 +537,7 @@
     <BuildOptions>
       MSFT:*_*_IA32_DLINK_FLAGS = /ALIGN:64
       MSFT:*_*_X64_DLINK_FLAGS  = /ALIGN:256
+      CLANGPDB: *_*_*_DLINK_FLAGS = /ALIGN:256
```
mdkinney: Split into 2 lines with different alignments for IA32 and X64

CryptoPkg/CryptoPkg.dsc
```diff
@@ -480,6 +483,7 @@
       MSFT:*_*_X64_DLINK_FLAGS  = /ALIGN:4096
       MSFT:*_*_X64_DLINK_XIPFLAGS  = /ALIGN:4096
       GCC:*_*_AARCH64_DLINK_XIPFLAGS = -z common-page-size=0x1000
+      CLANGPDB: *_*_*_DLINK_FLAGS = /ALIGN:4096
```
mdkinney: Split into 2 lines for IA32 and X64 to match MSFT

CryptoPkg/CryptoPkg.dsc
```diff
@@ -463,6 +465,7 @@
       MSFT:*_*_IA32_DLINK_XIPFLAGS = /ALIGN:64
       MSFT:*_*_X64_DLINK_FLAGS  = /ALIGN:256
       MSFT:*_*_X64_DLINK_XIPFLAGS  = /ALIGN:256
+      CLANGPDB: *_*_*_DLINK_FLAGS = /ALIGN:256
```
mdkinney: Split into 2 lines with different alignments for IA32 and X64

os-d:
> > @deepak5x - Please run the patch check tool and make sure that all style/formatting issues are fixed before submitting a PR. Also per @os-d's comment, please add a description to the PR.
> > The actual technical content here looks good to me. Essentially whenever a module chooses a specific PE/COFF file alignment option when building with Visual C++ it makes sense to replicate that build option for CLANGPDB since the link phase for those two compilers follow the same high-level design.
> 
> As changes are in multiple packages and if tried to add in "package,package: description" format, it cross description length. We get below error: " CryptoPkg,SecurityPkg,EmulatorPkg,OvmfPkg,StandaloneMmPkg/BuildOptions : Add CLANGPDB DLINK_FLAGS flags to build options The commit message format is not valid:
> 
> * First line of commit message (subject line) is too long (120 > 91)."

@deepak5x in this case, there is no reason all of these different packages need to be merged into one commit. Please split each package out into a separate commit, in which case you will not run into too long of a commit message.

When there is a need for a commit to span many packages, Global: <Subject> can be used.

EmulatorPkg/EmulatorPkg.dsc
deepak5x: Will be raising new PR

SecurityPkg/SecurityPkg.dsc
deepak5x: Raised new PR https://github.com/tianocore/edk2/pull/11594 as commit title issue comes for the multiple pkg

os-d: @deepak5x see my comment below from last week. You don't need separate PRs, you just need separate commits in this package, one per package.

deepak5x: > @deepak5x see my comment below from last week. You don't need separate PRs, you just need separate commits in this package, one per package.

Hi @os-d , sorry for the confusions, addressed all the comments and added commits into same PR. Please help to review.

EmulatorPkg/EmulatorPkg.dsc
```diff
@@ -539,7 +539,7 @@
   #
   # CLANGPDB tool chain depends on WIN_HOST_BUILD flag to generate the windows application.
   #
-  GCC:*_CLANGPDB_*_DLINK_FLAGS     = /ALIGN:4096 /FILEALIGN:4096
+  CLANGPDB: *_*_*_DLINK_FLAGS = /ALIGN:256 /FILEALIGN:4096
```
os-d: Why did this change to 256? For EmulatorPkg, it doesn't matter so much, but general rule of thumb should be 4096. Also, you replaced one instance of GCC:*_CLANGPDB_*_DLINK_FLAGS for the WinHost build, but there are two more below, for setting /BASE. Do those need to be updated?

deepak5x: _DLINK_FLAGS was changed for the /ALIGN, i am not sure if we need to change the /BASE

os-d: If the GCC syntax did not work for the use case and CLANGPDB does, then we should change all of the statements here.

liyi77 (approved):
CryptoPkg Part looks good to me

kraxel (approved):

PR #11431, opened 2025-08-20
ArmPkg: mark ARMv9 cpus as such in SMBIOS Type4

hrw:
SMBIOS Type4.ProcessorFamily2 field has separate value for ARMv9 cpu cores. This change reads ID_AA64ZFR0_EL1 register to check for SVE2 support (mandatory on ARMv9) and if it is present then system is marked as ARMv9 one.

# Description

BSA ACS checks for SVE2 on ARMv9 cpus.

Check is done by comparing SMBIOS Type4 ProcessorFamily2 field against 0x102 (ARMv9) value.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

I ran BSA ACS test 16 and it passed.

## Integration Instructions

N/A


ArmPkg/Library/ArmLib/AArch64/AArch64Support.S
```diff
@@ -393,4 +393,8 @@ ASM_FUNC(ArmReadIdAA64Pfr2)
   mrs   x0, ID_AA64PFR2_EL1
   ret
 
+ASM_FUNC(ArmReadIdAA64Zfr0)
```
hrw, line 396: I used "s3_0_c0_c4_4" here so it builds even for default armv8.0 cpu. Otherwise it complains that ID_AA64ZFR0_EL1 register is not supported.

leiflindholm: We tend to use this pattern: https://github.com/tianocore/edk2/blob/master/MdePkg/Include/AArch64/AArch64.h#L128
to deal with Arm's broken and always late mechanism for bringing in symbolic system register names.

ArmPkg/Universal/Smbios/ProcessorSubClassDxe/SmbiosProcessorArmCommon.c
```diff
@@ -207,6 +208,7 @@ SmbiosGetProcessorFamily2 (
 {
   UINTN   MainIdRegister;
   UINT16  ProcessorFamily2;
+  UINT64  Zfr0Register;
```
leiflindholm, line 211: For checking features, we have ArmHas* helpers in ArmPkg/Library/ArmLib/AArch64/AArch64Lib.c, could you please add an ArmHasSve2()?

However, from looking at the Arm ARM (0487L.b) what I'm seeing is "FEAT_SVE2 is OPTIONAL from Armv9.0", so I'm not sure that's the right test. In fact, I'm not seeing a single feature added by Armv9.0 that is Mandatory.

leiflindholm: I'm not convinced there is a way to achieve this programmatically (which is basically how Arm wants it). In which case, the pragmatic approach would be to change it to a Pcd.

hrw: On systems like SbsaQemu it can be v8 or v9 so static Pcd does not make sense.

hrw:
Abandoning - may come with better idea other time.

ArmPkg/Universal/Smbios/ProcessorSubClassDxe/SmbiosProcessorArmCommon.c
leiflindholm: No, but it could be a Static/Dynamic Pcd and we could figure out a platform-specific way to detect. Still, it would get messy.

PR #11436, opened 2025-08-22
BaseTools: Add gcov_info to GCC linker script

garybeihl:
GCC is capable of adding code coverage sections that are filled in at runtime. This change collects those sections into the final binary so that the code coverage data are available for dumping and post-processing using gcov.

# Description

GCC is capable of adding code coverage sections that are filled in at runtime. This change collects those sections into the final binary so that the code coverage data are available for dumping and post-processing using gcov.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Test shell app compiled with gcc --coverage and linked with this GccBase.lds was run under QEMU.

## Integration Instructions

N/A


AshrafAliS:
Do we have any BKM to generate the cov

garybeihl:
> Do we have any BKM to generate the cov

Nothing yet. I could put an example into a future PR is there's interest. A recent version of GCC and matching gcov library/utils is required. This PR is just to enable further target-based coverage work. Not intended as a replacement for host-based (googletest) coverage, but rather as a complement.

garybeihl:
Hi, it looks like the Analyze (IntelFsp2WrapperPkg, IA32,X64) job failed due to a transient error (possibly network-related). All checks were passing prior to my rebase. Could someone with access please re-run that job? Thanks!

garybeihl:
Could I please get a review or at least some reviewer feedback on this small change? Thank you!

mikebeaton:
Would it make more sense to implement this PR when there is known tooling, or a method for using it, also included in the codebase? I am happy to have it explained why not; but are you not currently having to make loads of other changes to the build in order to collect this coverage info - so why does this change need to go into master without the rest? Again, happy to be given context which helps clarify why this is indeed a good decision!

garybeihl:
Hi - Thanks for looking at this! This PR is a very small change that enables experimentation with target-based coverage. I have some additional changes (the test shell app mentioned in the testing section) that implement an example usage, but they are pretty messy right now and probably not optimized for presentation in a PR. The changes in this first PR should not break anything and they make it easier to come up with a better usage PR later on. You are right that there are lots of other changes needed to support the target-based coverage example, but it might be cleaner to separate those changes from the basic PR that collects the .gcov info. Hope this helps!

mikebeaton:
Okay, thank you. But in that case I am still not convinced that this is better to commit separately.

I agree that it _appears_ harmless, indeed useful in some future state of the codebase, and also somewhat (but seemingly minimally) useful to someone currently working on this, as you are.

But given that - as I understand it - you would have to be making significant other changes in the repo anyway, each time, in order to actually use it - I still don't see the logic of 'storing' this commit on its own, in master, here, before more is ready. It seems much more appropriate to store in one of your own branches, for now?

garybeihl:
I guess it's also that the particular way I might use the .gcov capability might be significantly different from the way someone else might choose to use it. Enabling and using the feature seem orthogonal enough to deserve separate PRs. But if not, then I understand. Thank you for looking at this!

PR #11437, opened 2025-08-24
RedfishPkg/ConfigHandler: Redfish transport interface disconnection

changab:
Provide an event for the Redfish client to signal Redfish configuration handler driver to disconnect from the Redfish transport interface.

# Description

Provide an event for the Redfish client to signal Redfish
configuration handler driver to disconnect from the Redfish
transport interface.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

While Redfish is enabled, reboot the system to see if the HTTP session is closed on BMC side.

## Integration Instructions

N/A

### DIFF
--- a/RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.h
+++ b/RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.h
@@ -30,6 +30,7 @@
 //
 #include <Protocol/EdkIIRedfishCredential.h>
 #include <Protocol/EdkIIRedfishConfigHandler.h>
+#include <Protocol/RestEx.h>
 
 //
 // Driver Version
--- a/RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.c
+++ b/RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.c
@@ -13,9 +13,10 @@
 REDFISH_CONFIG_DRIVER_DATA  gRedfishConfigData;     // Only one Redfish service supported
                                                     // on platform for the BIOS
                                                     // Redfish configuration.
-EFI_EVENT                          gEndOfDxeEvent        = NULL;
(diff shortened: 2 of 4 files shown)


nicklela (approved):

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.h
```diff
@@ -30,6 +30,7 @@
 //
 #include <Protocol/EdkIIRedfishCredential.h>
 #include <Protocol/EdkIIRedfishConfigHandler.h>
+#include <Protocol/RestEx.h>
```
leiflindholm, line 33: This would seem to violate https://tianocore-docs.github.io/edk2-CCodingStandardsSpecification/draft/5_source_files/53_include_files.html#534-include-files-may-include-only-those-headers-that-it-directly-depends-upon

Does that also apply to the two include statements already present pefore this commit?

changab: > This would seem to violate https://tianocore-docs.github.io/edk2-CCodingStandardsSpecification/draft/5_source_files/53_include_files.html#534-include-files-may-include-only-those-headers-that-it-directly-depends-upon
> 
> Does that also apply to the two include statements already present pefore this commit?

This driver has dependencies on the above two header files.  But, thanks for catching this, I think we could leverage  EFI_REDFISH_DISCOVER_PROTOCOL.ReleaseRedfishService() to disconnect the transport interface. I will check the feasibility.

leiflindholm: The driver, yes - but the header file itself?

changab:
I am following up with this.

changab:
@leiflindholm, this one was stale. PR: https://github.com/tianocore/edk2/pull/12550 is the successor.

PR #11442, opened 2025-08-26
[RFC] Remove OvmfPkgIA32 and ArmVirtPkg ARM Platforms

os-d:
# Description

Per RFC https://edk2.groups.io/g/devel/topic/rfc_remove_ovmf_ia32_and/114152215, this PR removes the OvmfPkgIA32 and ArmVirtPkg ARM virtual platforms from edk2. The RFC has stopped generating comments and the stable tag period has ended, so I have moved the RFC to this PR to get additional comments and move forward with the deprecation. The full justifications and discussions are in that RFC, but bringing the high level summary here.

Summary:
--
These are fully 32-bit virtual platforms that do not represent platforms in the wild any longer. They have special considerations
that they enforce on common code, which leads to more configuration options and a worse state for common code to accommodate these virtual platforms.

edk2 hosts virtual platforms with the goal for a quick development vehicle and to be representative of real platforms. These two
virtual platforms no longer meet the goal of being representative of physical platforms. They should be removed.

Keeping the IA32 only virtual platforms complicates work such as strengthening memory protections because IA32 was not built with the same standards of memory protections that X64 was (and has been iterated on). As a result, attempting to increase the security state for physical platforms (which are exclusively not IA32 in DXE) becomes complicated. Hedges must be made to support these virtual platforms, just for the sake of them, not for any real platform they used to represent.

Furthermore, these platforms take up CI time. In a recent pipeline run, we had these times:

| Pipeline | Time |
|----------|-------|
| OVMF IA32 VS2022 DEBUG | 8m 33s |
| OVMF IA32 VS2022  RELEASE | 10m 39s |
| OVMF IA32 VS2022 NOOPT | 14m 18s |
| OVMF IA32 GCC DEBUG | 8m 6s |
| OVMF IA32 GCC RELEASE | 7m 49s |
| OVMF IA32 GCC NOOPT | 6m 48s |
| ArmVirtPkg QEMU ARM DEBUG | 7m 19s |
| ArmVirtPkg QEMU ARM RELEASE | 7m 38s |
| ArmVirtPkg QEMU ARM NOOPT | 8m 42s |
| ArmVirtPkg QEMU KERNEL ARM DEBUG | 8m 42s |
| ArmVirtPkg QEMU KERNEL ARM RELEASE | 8m 39s |
| ArmVirtPkg QEMU KERNEL ARM NOOPT | 6m 45s |
| ArmVirtPkg KVMTOOL ARM DEBUG | 6m 37s |
| ArmVirtPkg KVMTOOL ARM RELEASE | 6m 13s |

These builds can be parallelized, but when they are, they take away the opportunity for other pipelines to be run in parallel, which then extends the PR gate CI time. In addition, it is being discussed to add clang pipelines, which then further magnifies this problem.

In short, removing these virtual platforms allows us to simplify the codebase, remove configuration options, strengthen security
guarantees, and improve CI times greatly. In return, we lose development and validation on platforms that do not exist any longer. I believe this is far and away a good trade. Again, for real platforms that were recently manufactured, namely 32-bit SEC and PEI platforms, we will retain OVMF IA32/X64 for development and validation.

Proposal:
--
1) Have edk2-stable202508 be the final release with IA32 OVMF and ARM ArmVirtPkg included. IA32/X64 OVMF will be kept for now and have a separate deprecation timeline.

2) Fork edk2-stable202508 for these virtual platforms to be consumed in perpetuity and any build breaking changes can be applied (e.g. when a submodule remote changes, etc.) and no other fixes be applied. If any consumers want to maintain these 32 bit platforms, they are of course welcome to fork it and independently maintain it with whatever features they
desire.

3) After this change completes, remove all 32 bit (IA32 and ARM) code from DXE/MM. At this time, all ARM32 code could be removed also.

4) After consultation with Intel, plan for a deprecation story for 32 bit code from SEC/PEI (e.g. the rest of the 32 bit code in edk2). This is dependent on the story for Intel platforms moving to 64 bit PEI (and if this includes SEC or not). This would also include deprecating and eventually removing OVMF IA32/X64.

- [x] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Running the virtual platforms.

## Integration Instructions

This removes OVMF IA32 and ArmVirtPkg ARM virtual platforms from edk2. They can be consumed from a branch forked from edk2-stable202508 here: TBD


mdkinney:
This is a very large PR and it is mixing different types of changes.

For CI execution time, could we consider moving some of these platforms from PR review checks to a periodic check (perhaps weekly).  That would reduce the CI time without requiring all these other changes.

I also see removal of EBC mixed into this PR.  Can we make EBC removal from modules that do not make sense for EBC as its own PR as a clean up we do first?

os-d:
> This is a very large PR and it is mixing different types of changes.

It is a very large PR, but the changes are of the same type: remove a deprecated virtual platform in each commit. It is removed wholesale. If you want, I can split each commit into a separate PR. I can also split these commits into smaller commits, if you want, I think the benefit of having one commit that removes the platform is that it is easy to bring back if that ever becomes the case, simply revert the removal, and also it is easier when looking at the history: this is the commit where OVMF IA32 was removed. Simple, done.

> For CI execution time, could we consider moving some of these platforms from PR review checks to a periodic check (perhaps weekly). That would reduce the CI time without requiring all these other changes.
> 

For these deprecated platforms, I believe we should just remove them. I don't think there is value in continuing to run them at all (they'll just start to fail and then people won't fix them, because they are deprecated platforms). They will continue to force us to incorporate fixes for 32-bit code in DXE, which is what my main goal here is. CI time improvement is a great added benefit, but these platforms require us to keep DXE/MM in the past, only supporting the minimum set of features (or having lots of untested configurability). We should remove them. On the mailing thread, there was general agreement about this path. Linux distros just wanted a place to be able to consume OVMF IA32 from, which is why Leif suggested a branch from the stable tag that I proposed only will get build breaking updates (such as an upstream remote changing, etc.).

> I also see removal of EBC mixed into this PR. Can we make EBC removal from modules that do not make sense for EBC as its own PR as a clean up we do first?

I can drop removing EBC from comments. When removing deprecated architectures from the INFs, I couldn't help but remove the deprecated EBC, since it is just a comment anyway that isn't used for anything and is often inaccurate. But, I can definitely remove that.

mdkinney:
The process defined today supports stable tags. A branch can be created from a stable tag for critical bug fixes, but critical bug fixes can only be applied until the next stable tag is created which is typically about 3 months.

A long lived supported branch is new concept for the edk2 repo.  We need to see a complete proposal on how that will be named, who is responsible for maintenance, and how GitHub issues and PRs are targeted to a long lived supported branch.

os-d:
> The process defined today supports stable tags. A branch can be created from a stable tag for critical bug fixes, but critical bug fixes can only be applied until the next stable tag is created which is typically about 3 months.
> 
> A long lived supported branch is new concept for the edk2 repo. We need to see a complete proposal on how that will be named, who is responsible for maintenance, and how GitHub issues and PRs are targeted to a long lived supported branch.

Well, to be honest, I think that edk2 will have satisfied any responsibility to the community who wants to consume these virtual platforms by creating a branch from the stable tag. I thought it was a nice to have to put any future build breaking changes there, but it absolutely doesn't have to be. If any consumer wants to take OVMF IA32, e.g., and put it in a new home and host it, fixing up whatever build breaks have occurred, they can certainly do that. edk2 provides a reference firmware for core code. It is not a production platform code repository.

So, from my perspective, the stable branch process is just fine.

niruiyu (dismissed):

mdkinney:
I have confirmed there are downstream use cases for EmulatorPkg IA32, so please drop the 3rd commit from this PR and update the PR title/description.

We can revisit this at a later date if those use cases are removed.

os-d:
> I have confirmed there are downstream use cases for EmulatorPkg IA32, so please drop the 3rd commit from this PR and update the PR title/description.
> 
> We can revisit this at a later date if those use cases are removed.

@mdkinney I'm willing to do that to move the removal of the other two forward, but what about my suggestion via the mailing list?

E.g. that EmulatorPkg IA32's use case would be for IA32 SEC/PEI, which is fine, but not DXE. So I can create an EmulatorPkg IA32 DXE Core that it boots into, which can just print to the serial port "You made it" or dump the hob list or some system state, whatever. But that allows issues in IA32 SEC/PEI to be debugged via EmulatorPkg, but if there is some issue in an X64 DXE using the IA32 SEC/PEI, EmulatorPkg IA32 can't be used to debug that anyway, either OVMF IA32X64 or a physical platform has to be moved to.

Another option would be to make an EmulatorPkg override of each of the DXE components it is considered needed to keep for IA32 (which again, I don't think should be any). This of course causes a larger code base to duplicate all of this.

The main point is that if we leave any platforms with 32-bit DXE/MM components, all of those components must keep supporting this deprecated architecture. This makes new changes challenging, because they can still break the expectations of IA32. I really want to find a path where mainstream edk2 is free of 32-bit in DXE/MM. There are a number of changes I have been wanting to upstream around memory protections, in particular, that have been annoying to do so because of 32-bit arch support, architectures that were not built with modern standards of memory protections.

If neither of these alternatives is acceptable, can you share more of the downstream use case so we can brainstorm if there is a mutually agreeable proposal? Regardless, I will remove EmulatorPkg from this PR and we can deal with it separately.

ardbiesheuvel:
> > I have confirmed there are downstream use cases for EmulatorPkg IA32, so please drop the 3rd commit from this PR and update the PR title/description.
> > We can revisit this at a later date if those use cases are removed.
> 
> @mdkinney I'm willing to do that to move the removal of the other two forward, but what about my suggestion via the mailing list?
> 

Is there a reason we cannot run EmulatorPkg with IA32 SEC/PEI and 64-bit DXE like the mixed OVMF? That would resemble actual hardware platforms much more closely than pure IA32.

mdkinney:
> > > I have confirmed there are downstream use cases for EmulatorPkg IA32, so please drop the 3rd commit from this PR and update the PR title/description.
> > > We can revisit this at a later date if those use cases are removed.
> > 
> > 
> > @mdkinney I'm willing to do that to move the removal of the other two forward, but what about my suggestion via the mailing list?
> 
> Is there a reason we cannot run EmulatorPkg with IA32 SEC/PEI and 64-bit DXE like the mixed OVMF? That would resemble actual hardware platforms much more closely than pure IA32.

The mixed mode is not possible for an OS application.

mdkinney:
> > I have confirmed there are downstream use cases for EmulatorPkg IA32, so please drop the 3rd commit from this PR and update the PR title/description.
> > We can revisit this at a later date if those use cases are removed.
> 
> @mdkinney I'm willing to do that to move the removal of the other two forward, but what about my suggestion via the mailing list?
> 
> E.g. that EmulatorPkg IA32's use case would be for IA32 SEC/PEI, which is fine, but not DXE. So I can create an EmulatorPkg IA32 DXE Core that it boots into, which can just print to the serial port "You made it" or dump the hob list or some system state, whatever. But that allows issues in IA32 SEC/PEI to be debugged via EmulatorPkg, but if there is some issue in an X64 DXE using the IA32 SEC/PEI, EmulatorPkg IA32 can't be used to debug that anyway, either OVMF IA32X64 or a physical platform has to be moved to.
> 
> Another option would be to make an EmulatorPkg override of each of the DXE components it is considered needed to keep for IA32 (which again, I don't think should be any). This of course causes a larger code base to duplicate all of this.
> 
> The main point is that if we leave any platforms with 32-bit DXE/MM components, all of those components must keep supporting this deprecated architecture. This makes new changes challenging, because they can still break the expectations of IA32. I really want to find a path where mainstream edk2 is free of 32-bit in DXE/MM. There are a number of changes I have been wanting to upstream around memory protections, in particular, that have been annoying to do so because of 32-bit arch support, architectures that were not built with modern standards of memory protections.
> 
> If neither of these alternatives is acceptable, can you share more of the downstream use case so we can brainstorm if there is a mutually agreeable proposal? Regardless, I will remove EmulatorPkg from this PR and we can deal with it separately.

The use case involves booting to UEFI Shell and running test/app in the UEFI Shell env.

mdkinney:
> > I have confirmed there are downstream use cases for EmulatorPkg IA32, so please drop the 3rd commit from this PR and update the PR title/description.
> > We can revisit this at a later date if those use cases are removed.
> 
> @mdkinney I'm willing to do that to move the removal of the other two forward, but what about my suggestion via the mailing list?
> 
> E.g. that EmulatorPkg IA32's use case would be for IA32 SEC/PEI, which is fine, but not DXE. So I can create an EmulatorPkg IA32 DXE Core that it boots into, which can just print to the serial port "You made it" or dump the hob list or some system state, whatever. But that allows issues in IA32 SEC/PEI to be debugged via EmulatorPkg, but if there is some issue in an X64 DXE using the IA32 SEC/PEI, EmulatorPkg IA32 can't be used to debug that anyway, either OVMF IA32X64 or a physical platform has to be moved to.
> 
> Another option would be to make an EmulatorPkg override of each of the DXE components it is considered needed to keep for IA32 (which again, I don't think should be any). This of course causes a larger code base to duplicate all of this.
> 
> The main point is that if we leave any platforms with 32-bit DXE/MM components, all of those components must keep supporting this deprecated architecture. This makes new changes challenging, because they can still break the expectations of IA32. I really want to find a path where mainstream edk2 is free of 32-bit in DXE/MM. There are a number of changes I have been wanting to upstream around memory protections, in particular, that have been annoying to do so because of 32-bit arch support, architectures that were not built with modern standards of memory protections.
> 
> If neither of these alternatives is acceptable, can you share more of the downstream use case so we can brainstorm if there is a mutually agreeable proposal? Regardless, I will remove EmulatorPkg from this PR and we can deal with it separately.

I would also like to highlight that the industry standards that the edk2 project provides an implementation of has not deprecated the IA32 architecture.

If there are challenges implementing core features for one CPU arch vs another CPU arch, then that may indicate a need for a set of abstraction services that allow the core implementation to support CPU features being present or absent.

os-d:
> I would also like to highlight that the industry standards that the edk2 project provides an implementation of has not deprecated the IA32 architecture.
> 

Sure, but there will also never be another IA32 platform. The UEFI spec also still describes Itanium, e.g.

> If there are challenges implementing core features for one CPU arch vs another CPU arch, then that may indicate a need for a set of abstraction services that allow the core implementation to support CPU features being present or absent.

Right, but this then requires a lot of backbending to support one virtual platform that does not represent modern real hardware. And the kind of issues I'm talking about are buggy things with IA32 DXE that has been left to bit rot because the only usage of it is these virtual platforms that test a very small set of functionality and so when anything new comes in, these bugs get exposed. So then it becomes trying to fix a can of worms for a virtual platform that does not represent modern real hardware. And that becomes a much larger battle.

> The use case involves booting to UEFI Shell and running test/app in the UEFI Shell env.

Why can this not be migrated to use OVMF IA32/X64, which as Ard pointed out is much more representative of modern HW? QEMU also runs as a host application, obviously edk2 runs it in pipelines. It may be on the order of 30 seconds slower (but also runs much more code). I think it is reasonable to discuss whether edk2 should continue support this downstream usage when it has other appropriate avenues.

But, as I mentioned, I don't want to derail the removal of OVMF IA32 and ArmVirtPkg ARM, so I will drop EmulatorPkg from this PR and we can have a separate discussion.

os-d:
I have removed the third commit from this PR (EmulatorPkg IA32) and updated the title/description. I have also added back the EBC comments (e.g. only IA32/ARM are removed from Supported Architectures, not EBC).

OvmfPkg/Include/IndustryStandard/QemuFwCfg.h
```diff
@@ -38,7 +38,7 @@
 
 //
 // The fw_cfg registers can be found at these IO Ports, on the IO-mapped
-// platforms (Ia32 and X64).
+// platform (X64).
```
mdkinney: These registers are in the chipset and are not related to IA32 or X64.  I think the platforms part of this comment can be removed completely.

os-d: Updated.

mdkinney:
The PR Description needs to remove the EmulatorPkg references.

os-d:
> The PR Description needs to remove the EmulatorPkg references.

I've removed them. I had put them there to put this PR in context, but I've removed the additional language.

mdkinney (approved):

os-d:
It seems there is general agreement about this approach (at least comments on the mail thread and PR have died down and approvals have started coming in).

Can the OvmfPkg/ArmVirtPkg maintainers/reviewers please review? @ardbiesheuvel @leiflindholm @jyao1 @kraxel @samimujawar

ardbiesheuvel:
> It seems there is general agreement about this approach (at least comments on the mail thread and PR have died down and approvals have started coming in).
> 
> Can the OvmfPkg/ArmVirtPkg maintainers/reviewers please review? @ardbiesheuvel @leiflindholm @jyao1 @kraxel @samimujawar

If we remove ArmVirtQemu for 32-bit ARM, we should just rip out all of the ARM code entirely: it will just bit rot otherwise, given that this is the only platform that uses this code in CI. The only remaining platforms in edk2-platforms that might conceivably be used in 32-bit mode are BeagleBoard, Juno and Raspberry Pi, and I have no reason to assume that anyone has bothered with those in a long time, with the exception of @bexcran who asked about this a while ago.

os-d:
> If we remove ArmVirtQemu for 32-bit ARM, we should just rip out all of the ARM code entirely: it will just bit rot otherwise, given that this is the only platform that uses this code in CI.

I agree, but I wanted to do this piecewise because it is a large set of changes, which can spook reviewers. I didn't want to make it even larger and then have this sit for longer. When this goes in, I'll put up a PR to remove ARM entirely. Is that acceptable? I'll start working on that PR now.

ardbiesheuvel:
> > If we remove ArmVirtQemu for 32-bit ARM, we should just rip out all of the ARM code entirely: it will just bit rot otherwise, given that this is the only platform that uses this code in CI.
> 
> I agree, but I wanted to do this piecewise because it is a large set of changes, which can spook reviewers. I didn't want to make it even larger and then have this sit for longer. When this goes in, I'll put up a PR to remove ARM entirely. Is that acceptable? I'll start working on that PR now.

Let's give @leiflindholm some time to chime in - he might feel radically differently about this.

leiflindholm:
> Let's give @leiflindholm some time to chime in - he might feel radically differently about this.

I think there is already enough Arm code that isn't actually used by any current platform that it's not problematic to delete those bits in a separate PR.

Also, we'll need to prune some edk2-platforms ports that currently enable building for ARM|AARCH64.

os-d:
> > Let's give @leiflindholm some time to chime in - he might feel radically differently about this.
> 
> I think there is already enough Arm code that isn't actually used by any current platform that it's not problematic to delete those bits in a separate PR.
> 
> Also, we'll need to prune some edk2-platforms ports that currently enable building for ARM|AARCH64.

Yeah, I can also put up an edk2-platforms PR removing ARM support (and the ARM only platforms).

samimujawar (approved):
I have no issues with discontinuing ARM (AArch32) support.

ardbiesheuvel (approved):

bexcran:
@os-d I'm going to be working on removing IPF and EBC support from edk2. So our work doesn't clash, what's your timeline looking like to deorbiting ARM and IA32 support?

os-d:
> @os-d I'm going to be working on removing IPF and EBC support from edk2. So our work doesn't clash, what's your timeline looking like to deorbiting ARM and IA32 support?

@bexcran great! I should have my PR up to remove ARM32 support from edk2 by the end of the week, I am part way through, now.

IA32 won’t be touched for awhile, platforms shipped with IA32 SEC/PEI as recently as 2023, so it will be years before the community is ready to drop that support. IA32 DXE/MM can be dropped sooner, but honestly it isn’t worthwhile to do piecemeal, so I’ll wait until we are ready to drop IA32 entirely.

So if you want to wait until I put up the PR to drop ARM32 support, that probably makes the most sense, thanks!

PR #11446, opened 2025-08-27
UefiCpuPkg: X64 MpLib: Spinlock RestoreVolatileRegisters

os-d:
# Description

All APs share the same GDT and segment descriptors. Each time an AP is started up, the wake up function restores the GDT and other system registers with the RestoreVolatileRegisters function.

A General Protection fault has recently been observed on multiple generations of Intel silicon when the APs have been left in an mwait loop by CpuMpPei and the initial wake up state in DXE is not init-sipi-sipi. The cores all come up at the same time, where they attempt to restore the GDT from the same address. However, the TSS is particular cannot be loaded simultaneously: when the busy bit is set in the TSS (which occurs when loaded by the ltr instruction in RestoreVolatileRegisters) the ltr instruction will fail with a GP fault with an exception context of 0x40, the entry in the GDT corresponding to the TSS.

In other scenarios, the cores will not come up at exactly the same time, either because they've been serialized through an init-sipi-sipi or the timing is right enough that they load the TSS separately. RestoreVolatileRegisters does attempt to clear the busy bit of the TSS before setting it via ltr, but this is not protected by a lock and so the other APs can execute ltr inbetween an AP setting the busy bit to 0 and then doing ltr.

This commit fixes this behavior by using the MpLock, shared among all APs (and the BSP) to serialize execution of RestoreVolatileRegisters. This resolves the bug seen. On systems where the timing was not hitting this, it is essentially a no-op.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on multiple generations of Intel physical platforms that were hitting this issue, it resolved the issue.

## Integration Instructions

N/A.

### DIFF
--- a/UefiCpuPkg/Library/MpInitLib/MpLib.c
+++ b/UefiCpuPkg/Library/MpInitLib/MpLib.c
@@ -36,12 +36,12 @@ SaveVolatileRegisters (
   Restore the volatile registers following INIT IPI.
 
   @param[in]  VolatileRegisters   Pointer to volatile resisters
-  @param[in]  IsRestoreDr         TRUE:  Restore DRx if supported
-                                  FALSE: Do not restore DRx
+  @param[in]  MpLock              Pointer to spin lock to synchronize access to GDT
 **/
 VOID
 RestoreVolatileRegisters (
-  IN CPU_VOLATILE_REGISTERS  *VolatileRegisters
+  IN CPU_VOLATILE_REGISTERS  *VolatileRegisters,
+  IN SPIN_LOCK               *MpLock
   );
 
 /**
@@ -117,7 +117,7 @@ FutureBSPProc (
   //
   SaveVolatileRegisters (&DataInHob->APInfo.VolatileRegisters);
   AsmExchangeRole (&DataInHob->APInfo, &DataInHob->BSPInfo);
-  RestoreVolatileRegisters (&DataInHob->APInfo.VolatileRegisters);
+  RestoreVolatileRegisters (&DataInHob->APInfo.VolatileRegisters, &DataInHob->MpLock);
 }
 
 /**
@@ -247,11 +247,17 @@ SaveVolatileRegisters (
 **/
 VOID
 RestoreVolatileRegisters (
-  IN CPU_VOLATILE_REGISTERS  *VolatileRegisters
+  IN CPU_VOLATILE_REGISTERS  *VolatileRegisters,
+  IN SPIN_LOCK               *MpLock
   )
 {
   CPUID_VERSION_INFO_EDX  VersionInfoEdx;
   IA32_TSS_DESCRIPTOR     *Tss;
+
+  // The GDT and segment descriptors are shared by all APs and cannot be accessed
+  // concurrently. Acquire the MP lock to prevent other APs from changing them
+  // while we are restoring them.
+  AcquireSpinLock (MpLock);
 
   AsmWriteCr3 (VolatileRegisters->Cr3);
   AsmWriteCr4 (VolatileRegisters->Cr4);
@@ -283,6 +289,8 @@ RestoreVolatileRegisters (
       AsmWriteTr (VolatileRegisters->Tr);
     }
   }
+
+  ReleaseSpinLock (MpLock);
 }
 
 /**
@@ -777,7 +785,7 @@ ApWakeupFunction (
       //   to initialize AP in InitConfig path.
       // NOTE: IDTR.BASE stored in CpuMpData->CpuData[ProcessorNumber].VolatileRegisters points to a different IDT shared by all APs.
       //
-      RestoreVolatileRegisters (&CpuMpData->CpuData[ProcessorNumber].VolatileRegisters);
+      RestoreVolatileRegisters (&CpuMpData->CpuData[ProcessorNumber].VolatileRegisters, &CpuMpData->MpLock);
       InitializeApData (CpuMpData, ProcessorNumber, BistData, ApTopOfStack);
       ApStartupSignalBuffer = CpuMpData->CpuData[ProcessorNumber].StartupApSignal;
     } else {
@@ -804,7 +812,7 @@ ApWakeupFunction (
         0
         );
 
-      RestoreVolatileRegisters (&CpuMpData->CpuData[ProcessorNumber].VolatileRegisters);
+      RestoreVolatileRegisters (&CpuMpData->CpuData[ProcessorNumber].VolatileRegisters, &CpuMpData->MpLock);
 
       if (GetApState (&CpuMpData->CpuData[ProcessorNumber]) == CpuStateReady) {
         Procedure = (EFI_AP_PROCEDURE)CpuMpData->CpuData[ProcessorNumber].ApFunction;
@@ -910,7 +918,7 @@ DxeApEntryPoint (
     AsmWriteMsr64 (MSR_IA32_EFER, EferMsr.Uint64);
   }
 
-  RestoreVolatileRegisters (&CpuMpData->CpuData[ProcessorNumber].VolatileRegisters);
+  RestoreVolatileRegisters (&CpuMpData->CpuData[ProcessorNumber].VolatileRegisters, &CpuMpData->MpLock);
   InterlockedIncrement ((UINT32 *)&CpuMpData->FinishedCount);
   PlaceAPInMwaitLoopOrRunLoop (
     CpuMpData->ApLoopMode,
@@ -2701,7 +2709,7 @@ SwitchBSPWorker (
   //
   SaveVolatileRegisters (&CpuMpData->BSPInfo.VolatileRegisters);
   AsmExchangeRole (&CpuMpData->BSPInfo, &CpuMpData->APInfo);
-  RestoreVolatileRegisters (&CpuMpData->BSPInfo.VolatileRegisters);
+  RestoreVolatileRegisters (&CpuMpData->BSPInfo.VolatileRegisters, &CpuMpData->MpLock);
   //
   // Set the BSP bit of MSR_IA32_APIC_BASE on new BSP
   //


os-d:
There were several other approaches considered here, including creating separate copies of the GDT and descriptors for each AP, but this was a heavier lift and considered riskier than this approach because that affects many more areas. This scoped approach is easy to implement and understand and resolves the issue, but may leave open other such races between APs (though not in this function).

x64 owners input would be appreciated.

niruiyu (commented):
if (Tss->Bits.P == 1) {
      Tss->Bits.Type &= 0xD;  // 1101 - Clear busy bit just in case
      AsmWriteTr (VolatileRegisters->Tr);
    }

I guess the above logic was to resolve the issue as all APs share one TR.
But I have some concerns regarding the addition of spin-lock. In a many-core system, waking up all APs simultaneously may become slower with it.

I think not sharing the TSS might be a better approach. That's the GDTR points to a GDT containing all normal segments shared by cores and NumberOfCpu TSS for each core.

os-d:
> ```
> if (Tss->Bits.P == 1) {
>   Tss->Bits.Type &= 0xD;  // 1101 - Clear busy bit just in case
>   AsmWriteTr (VolatileRegisters->Tr);
> }
> ```
> 
> I guess the above logic was to resolve the issue as all APs share one TR. But I have some concerns regarding the addition of spin-lock. In a many-core system, waking up all APs simultaneously may become slower with it.
> 

Yeah, this logic was intended to fix this issue, but it created a race condition where one AP would clear the busy bit, but before executing AsmWriteTr, another AP could execute AsmWriteTr, which set the busy bit again, generating the GP fault.

> I think not sharing the TSS might be a better approach. That's the GDTR points to a GDT containing all normal segments shared by cores and NumberOfCpu TSS for each core.

I agree that a separate TSS is a better approach. My concern is if it is too risky changing this pattern which has existed for a long time. I'm not sure if other code has been depending on a shared TSS. I'm also not sure if there are other GDT descriptors that may have a similar issue.

However, I agree the separate structures would be better. I will try with a separate TSS and can update this PR.

niruiyu:
> > ```
> > if (Tss->Bits.P == 1) {
> >   Tss->Bits.Type &= 0xD;  // 1101 - Clear busy bit just in case
> >   AsmWriteTr (VolatileRegisters->Tr);
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I guess the above logic was to resolve the issue as all APs share one TR. But I have some concerns regarding the addition of spin-lock. In a many-core system, waking up all APs simultaneously may become slower with it.
> 
> Yeah, this logic was intended to fix this issue, but it created a race condition where one AP would clear the busy bit, but before executing AsmWriteTr, another AP could execute AsmWriteTr, which set the busy bit again, generating the GP fault.
> 
> > I think not sharing the TSS might be a better approach. That's the GDTR points to a GDT containing all normal segments shared by cores and NumberOfCpu TSS for each core.
> 
> I agree that a separate TSS is a better approach. My concern is if it is too risky changing this pattern which has existed for a long time. I'm not sure if other code has been depending on a shared TSS. I'm also not sure if there are other GDT descriptors that may have a similar issue.
> 
> However, I agree the separate structures would be better. I will try with a separate TSS and can update this PR.

I reviewed the logic in UefiCpuPkg/Library/CpuExceptionHandlerLib/Ia32|X64/ArchExceptionHandler.c:ArchSetupExceptionStack().
The GDTR.base already points to different GDT for each AP. TSS is also separated for each AP.
Then the TSS/TSS_DESC are already separated. The separation is when PcdStackGuard is TRUE.

When PcdStackGuard is FALSE, TR for every AP should be 0.

Why does the bug still exist?

os-d:
>I reviewed the logic in UefiCpuPkg/Library/CpuExceptionHandlerLib/Ia32|X64/ArchExceptionHandler.c:ArchSetupExceptionStack(). The GDTR.base already >points to different GDT for each AP. TSS is also separated for each AP. Then the TSS/TSS_DESC are already separated. The >separation is when PcdStackGuard is TRUE.
> 
> When PcdStackGuard is FALSE, TR for every AP should be 0.
> 
> Why does the bug still exist?

This is a good point, thanks for calling this out. I did some more research.

There are a two different scenarios here. In the case of PcdStackGuard is false, ArchSetupExceptionStack doesn't get executed on the APs: https://github.com/tianocore/edk2/blob/d14e964692bafcae129ba91e7537a41f16e18a12/UefiCpuPkg/CpuDxe/CpuMp.c#L766-L768

However, even when PcdStackGuard is true, `InitializeMpExceptionHandlers` gets called after `MpInitLibInitialize`: https://github.com/tianocore/edk2/blob/d14e964692bafcae129ba91e7537a41f16e18a12/UefiCpuPkg/CpuDxe/CpuMp.c#L787-L799.

`MpInitLibInitialize` is where the BSP's GDT + TSS get copied for the APs: https://github.com/tianocore/edk2/blob/d14e964692bafcae129ba91e7537a41f16e18a12/UefiCpuPkg/Library/MpInitLib/MpLib.c#L2114

https://github.com/tianocore/edk2/blob/d14e964692bafcae129ba91e7537a41f16e18a12/UefiCpuPkg/Library/MpInitLib/MpLib.c#L237-L239

You are correct that for the first invocation, TR = 0, so this bug cannot be hit. However, after the APs run the first time, they `SaveVolatileRegisters`: https://github.com/tianocore/edk2/blob/d14e964692bafcae129ba91e7537a41f16e18a12/UefiCpuPkg/Library/MpInitLib/MpLib.c#L858

Which is going to set TR: https://github.com/tianocore/edk2/blob/d14e964692bafcae129ba91e7537a41f16e18a12/UefiCpuPkg/Library/MpInitLib/MpLib.c#L239

Then, we have our two cases again, however the same bug can come up in either place (but if PcdCpuStackGuard is disabled, it could happen on every startup of the APs, if PcdCpuStackGuard is enabled, it can only come in the next startup).

Focusing on the PcdCpuStackGuard is enabled case, the next call to wake up the APs is the one you referenced to do ArchSetupExceptionStack, which will create a new GDT/TSS for each AP. However, before it gets to executing that function, `ApWakeupFunction` does a `RestoreVolatileRegisters` which at this point is still pointing to the BSP's GDT/TSS: https://github.com/tianocore/edk2/blob/d14e964692bafcae129ba91e7537a41f16e18a12/UefiCpuPkg/Library/MpInitLib/MpLib.c#L807

This occurs before the function we are intended to call (to switch GDTs): https://github.com/tianocore/edk2/blob/d14e964692bafcae129ba91e7537a41f16e18a12/UefiCpuPkg/Library/MpInitLib/MpLib.c#L821

And then we have TR != 0 and so we execute this, which is where we GP fault: https://github.com/tianocore/edk2/blob/d14e964692bafcae129ba91e7537a41f16e18a12/UefiCpuPkg/Library/MpInitLib/MpLib.c#L283

So, I think there are two fixes needed:
- I think we need the separate exception stacks set up for the APs regardless of PcdCpuStackGuard being set or not. That will create a separate GDT/TSS for each AP and the separate exception stack is important for sane exception handling, regardless of whether the exception was a stack overflow or not (there are other stack corruptions that can occur).
- I think we should set TR back to 0 at the end of `MpInitLibInitialize` after we execute the APs the first time: https://github.com/tianocore/edk2/blob/35a3ceb882b57da0964c8b4a038e8808b3dc2b13/UefiCpuPkg/Library/MpInitLib/MpLib.c#L2421. That will ensure that the next time the APs are woken up (to create their own GDT) they won't try to restore the TSS, so they'll avoid the GP fault. After that point, they'll restore the TSS, but they'll be using their own, so it won't conflict with other APs

I need to test that this works still, but what are your thoughts?

os-d:
> I need to test that this works still, but what are your thoughts?

@niruiyu this approach worked on a physical Intel client system that was experiencing the issue. The only concern I have with this approach is that it requires the consumer of MpInitLib to do a specific action (create a new GDT for the APs), otherwise you can hit the bug. It would be better to pull that into MpInitLibInitialize itself so that it is in control of this flow and can ensure the proper steps.

What do you think about this approach?

niruiyu:
> And then we have TR != 0 and so we execute this, which is where we GP fault

Why is TR non-zero?

The AP's initial TR in VolatileRegisters[ProcessorNumber] is set to 0 in https://github.com/tianocore/edk2/blob/9c150aa63c60c61bb0c62164e8f493650581c314/UefiCpuPkg/Library/MpInitLib/MpLib.c#L2206.

So whenever AP wakes up, the TR is 0.

The TR value in VolatileRegisters[ProcessNumber] is changed to non-zero in SaveVolatileRegisters() after the MP routine that sets the separate GDT/TSS/TR  returns to ApWakeupFunction().

os-d:
> > And then we have TR != 0 and so we execute this, which is where we GP fault
> 
> Why is TR non-zero?
> 
> The AP's initial TR in VolatileRegisters[ProcessorNumber] is set to 0 in
> 
> https://github.com/tianocore/edk2/blob/9c150aa63c60c61bb0c62164e8f493650581c314/UefiCpuPkg/Library/MpInitLib/MpLib.c#L2206
> 
> .
> So whenever AP wakes up, the TR is 0.
> 
> The TR value in VolatileRegisters[ProcessNumber] is changed to non-zero in SaveVolatileRegisters() after the MP routine that sets the separate GDT/TSS/TR returns to ApWakeupFunction().

The Routine that sets the separate GDT/TSS/TR is not the first thing ran on the APs. This is run first: https://github.com/tianocore/edk2/blob/65485e195f37653bbbc06db2001c3a189d468666/UefiCpuPkg/Library/MpInitLib/MpLib.c#L2433.

After executing that first procedure (there are a couple other paths that could wake up APs in the initialize function, too, it looks like), it does a SaveVolatileRegisters: https://github.com/tianocore/edk2/blob/65485e195f37653bbbc06db2001c3a189d468666/UefiCpuPkg/Library/MpInitLib/MpLib.c#L858.

SaveVolatileRegisters sets the TR field

https://github.com/tianocore/edk2/blob/65485e195f37653bbbc06db2001c3a189d468666/UefiCpuPkg/Library/MpInitLib/MpLib.c#L239.

Which, per SDM is a non-zero value on power up/reset (or the PEI MP code could have set it):

image

So now when CpuDxe tries to execute the code to set the new GDT/TSS/TR, it will do RestoreVolatileRegisters in ApWakeupFunction:

https://github.com/tianocore/edk2/blob/65485e195f37653bbbc06db2001c3a189d468666/UefiCpuPkg/Library/MpInitLib/MpLib.c#L807

TR is now non-zero, so the code will attempt to do an ltr and hits the GP fault:
https://github.com/tianocore/edk2/blob/65485e195f37653bbbc06db2001c3a189d468666/UefiCpuPkg/Library/MpInitLib/MpLib.c#L276.

So switching to a separate GDT/TSS/TR is the right thing, but the code never gets there to do it. However, in this flow if we set VolatileRegisters->TR = 0 after that first wakeup, where VolatileRegisters->TR got set to a non-zero value, it works and we get to set a separate GDT/TSS/TR and from that point forward we can't hit this GP fault. But, we have to be able to get far enough to set the separate GDT/TSS/TR.

niruiyu:
> After executing that first procedure (there are a couple other paths that could wake up APs in the initialize function, too, it looks like), it does a SaveVolatileRegisters:

The brief flow is as follows:
```c
807:  RestoreVolatileRegisters (&CpuMpData->CpuData[ProcessorNumber].VolatileRegisters); // AP.TR is 0
...
821:  Procedure (Parameter);  // TR is changed if Procedure is to enable stack guard in APs
...
858:  SaveVolatileRegisters (&CpuMpData->CpuData[ProcessorNumber].VolatileRegisters); // AP.TR is not zero. AP.GDTR points to its own GDT.
```

The above flow won't cause a shared TSS issue. TR's initial value is 0 per SDM "Table 11-1. IA-32 and Intel® 64 Processor States Following Power-up, Reset, or INIT".

Does your firmware use CpuMpPei in PEI phase? Is stack guard enabled in PEI phase? Is PEI phase 64bit mode? When PEI and DXE use the same execution mode, a different init path in DXE is taken:
```c
2364:  SwitchApContext (MpHandOffConfig, FirstMpHandOff);
```

A complex combination of all the above factors may create an environment that can cause the issue.

os-d:
>> After executing that first procedure (there are a couple other paths that could wake up APs in the initialize function, too, it looks like), it does a SaveVolatileRegisters:
> 
> The brief flow is as follows:
> 

AP.TR is non-zero here, not zero. This is either because CpuMpPei set it up

> ```c
> 807:  RestoreVolatileRegisters (&CpuMpData->CpuData[ProcessorNumber].VolatileRegisters); // AP.TR is 0
> ...
> 821:  Procedure (Parameter);  // TR is changed if Procedure is to enable stack guard in APs
> ...
> 858:  SaveVolatileRegisters (&CpuMpData->CpuData[ProcessorNumber].VolatileRegisters); // AP.TR is not zero. AP.GDTR points to its own GDT.
> ```
> 
> The above flow won't cause a shared TSS issue. TR's initial value is 0 per SDM "Table 11-1. IA-32 and Intel® 64 Processor States Following Power-up, Reset, or INIT".
> 

You are correct that the TR's initial value is 0, I was missing that STR and LTR only affect the visible part of the TR:

image

image

So that's why I thought that TR would be non-zero on reset, since the invisible part is, but from the software side, it will be zero.

> Does your firmware use CpuMpPei in PEI phase? Is stack guard enabled in PEI phase? Is PEI phase 64bit mode? When PEI and DXE use the same execution mode, a different init path in DXE is taken:

Yes, we use CpuMpPei and have stack guard enabled in PEI. CpuMpPei is leaving the APs in mwait and the PCD is set to not do an initial wakeup by init-sipi-sipi. This is 64 bit PEI.

> 
> ```c
> 2364:  SwitchApContext (MpHandOffConfig, FirstMpHandOff);
> ```
> 
> A complex combination of all the above factors may create an environment that can cause the issue.

Yeah, I believe that is correct, where this is a relatively new scenario and is a race condition, which is why it isn't seen widespread. I believe that CpuMpPei is leaving the APs in a state such that TR != 0 and then we hit the flow I describe.

Another possible answer (though I'm not sure the implications of this) would be to have MpInitLibInitialize have the APs reset their TR to 0 on startup. I think that would also work. But, as noted, the other option to change the saved registers to 0 at the end of initialize also works.

os-d:
@niruiyu ping on this thread. Any thoughts about my comments above?

niruiyu:
> @niruiyu ping on this thread. Any thoughts about my comments above?

I am still trying to figure out a clear path that leads to the bug. Without understanding the whole path, I am reluctant to accept the changes.

os-d:
> > @niruiyu ping on this thread. Any thoughts about my comments above?
> 
> I am still trying to figure out a clear path that leads to the bug. Without understanding the whole path, I am reluctant to accept the changes.

@niruiyu let me try to lay out the full path in more detail.

The two Intel client platforms where we have seen the bug occur have X64 PEI, gUefiCpuPkgTokenSpaceGuid.PcdCpuApLoopMode set to 2 (mwait-loop), have PcdCpuStackGuard set to TRUE, and are using CpuMpPei.

When CpuMpPei comes up, it calls `MpInitLibInitialize`: https://github.com/tianocore/edk2/blob/2a8d98d0a5e8d662726b79e632658919cc8321f7/UefiCpuPkg/CpuMpPei/CpuMpPei.c#L352

`MpInitLibInitialize` will wake up all the APs with `VolatileRegisters.TR == 0`: https://github.com/tianocore/edk2/blob/2a8d98d0a5e8d662726b79e632658919cc8321f7/UefiCpuPkg/Library/MpInitLib/MpLib.c#L2220.

When the APs wake up and run, they will not try to restore the task register, because `VolatileRegisters.TR == 0`: https://github.com/tianocore/edk2/blob/2a8d98d0a5e8d662726b79e632658919cc8321f7/UefiCpuPkg/Library/MpInitLib/MpLib.c#L276.

Next, CpuMpPei will initialize the exception handlers on the APs, which sets their own GDT/TSS/TR: https://github.com/tianocore/edk2/blob/2a8d98d0a5e8d662726b79e632658919cc8321f7/UefiCpuPkg/CpuMpPei/CpuMpPei.c#L360

https://github.com/tianocore/edk2/blob/2a8d98d0a5e8d662726b79e632658919cc8321f7/UefiCpuPkg/Library/CpuExceptionHandlerLib/X64/ArchExceptionHandler.c#L260

So, after this procedure completes, we have the following state:

|BSP|AP0|AP..N|
|BSP.GDT|AP0.GDT|AP..N.GDT|
|BSP.TSS|AP0.TSS|AP..N.TSS|
|BSP.TR != 0|AP0.TR != 0|AP..N.TR != 0|

Finally, the APs will be placed in the mwait loop: https://github.com/tianocore/edk2/blob/2a8d98d0a5e8d662726b79e632658919cc8321f7/UefiCpuPkg/Library/MpInitLib/MpLib.c#L885

PEI continues to use the APs with whatever procedures being run, always using their own GDT/TSS/TR. No problem.

Now we transition to DXE. The APs are in mwait, with the table above showing their register state.

CpuDxe launches and runs `MpInitLibInitialize`: https://github.com/tianocore/edk2/blob/2a8d98d0a5e8d662726b79e632658919cc8321f7/UefiCpuPkg/CpuDxe/CpuMp.c#L787

However, it doesn't know that the APs already have their own GDT/TSS/TR set and so it writes over them with the BSP's GDT/TSS but not TR, as `VolatileRegisters.TR == 0`: 

https://github.com/tianocore/edk2/blob/2a8d98d0a5e8d662726b79e632658919cc8321f7/UefiCpuPkg/Library/MpInitLib/MpLib.c#L2125-L2128

https://github.com/tianocore/edk2/blob/2a8d98d0a5e8d662726b79e632658919cc8321f7/UefiCpuPkg/Library/MpInitLib/MpLib.c#L274 

So now we have this state:

|BSP|AP0|AP..N|
|BSP.GDT|BSP.GDT|BSP.GDT|
|BSP.TSS|BSP.TSS|BSP.TSS|
|BSP.TR != 0|AP0.TR != 0|AP..N.TR != 0|

When that initial wakeup procedure is completed (but still running on the APs in `ApWakeupFunction`, we call `SaveVolatileRegisters`: https://github.com/tianocore/edk2/blob/2a8d98d0a5e8d662726b79e632658919cc8321f7/UefiCpuPkg/Library/MpInitLib/MpLib.c#L858

This will set `VolatileRegisters.TR != 0`, because, as shown in the table above, the APs have non-zero task registers: https://github.com/tianocore/edk2/blob/2a8d98d0a5e8d662726b79e632658919cc8321f7/UefiCpuPkg/Library/MpInitLib/MpLib.c#L239. 

Then CpuDxe runs the procedure to set separate exception handlers, which will cause the APs to wakeup in `ApWakeupFunction`, where they will run `RestoreVolatileRegisters` _before_ running the procedure on the AP (which is going to switch back to their own GDT/TSS/TR): https://github.com/tianocore/edk2/blob/2a8d98d0a5e8d662726b79e632658919cc8321f7/UefiCpuPkg/Library/MpInitLib/MpLib.c#L807.

This is where the race condition happens. As shown above, `VolatileRegisters.TR !=0` so `RestoreVolatileRegisters` will attempt to  write the task register: https://github.com/tianocore/edk2/blob/2a8d98d0a5e8d662726b79e632658919cc8321f7/UefiCpuPkg/Library/MpInitLib/MpLib.c#L283

It tries to clear the busy bit before writing the TR, but not under a lock, so that's where the race condition is. We hit this situation:

AP X: Clear busy bit in shared BSP TSS

                                  AP Y: Clear busy bit in shared BSP TSS
                                  AP Y: Write TR, which sets the busy bit in shared BSP TSS

AP X: Write TSS -> GP Fault, busy bit is set

So, there are many different ways to handle this. One way is the spin lock method that I have in this PR. That protected the critical section of clearing the busy bit and writing the TR. However, I agree this is heavy handed and it can be done lockless.

Another, simple way is to do this: https://github.com/os-d/mu_basecore/commit/ab658c622c57c6aafc421d831a4a00ad0c6b0694. When we set `VolatileRegisters.TR = 0` after executing the APs for the first time in CpuDxe, we will not attempt to write the task register in `RestoreVolatileRegisters` on the next AP wakeup. This will run the procedure to set their own GDT/TSS/TR and then this race condition cannot happen. It will be the same as when CpuMpPei came up for the first time.

However, this solution has two downsides:
- MpInitLib now has an expectation for its consumers: they must run the procedure to set the APs to use their own GDT/TSS/TR as the next step after calling `MpInitLibInitialize`, otherwise they face this race condition
  - This could be mitigated by calling the call to set up exception handling into `MpInitLibInitialize` itself, so the consumers get the APs in a good state after initializing the lib
- The APs go from BSP GDT/TSS/TR -> their own -> Back to BSP's -> Back to their own (different than original)

Yet another option to go, though more changes, would be to pass state between CpuMpPei and CpuDxe to indicate that CpuMpPei left the APs in a state where they have their own exception stack set up, so they don't need to go back to the BSP's registers temporarily. I haven't tested or traced through this entirely, so I'm not sure if there are any issues on this path.

My inclination is to go with setting `VolatileRegisters.TR = 0` after running `MpInitLibInitialize` because it is the smallest delta to the current code, which is complex, with many, many configurable paths to go through. It would be best to also pull in setting up the separate exception stacks into MpInitLib to reduce the contract with consumers.

What are your thoughts?

os-d:
@niruiyu we are seeing more devices affected by this. We now have two generations of Intel servers hitting this bug, in addition to the two generations of Intel client.

We have this resolved on our side, but I assume others will be hitting this as well.

os-d:
@niruiyu I have pushed the proposed changes here, please re-review.

Other UefiCpuPkg maintainers/reviewers, please also review. @ecuevasf @jiaxinwu @kraxel @LiuZhiguang001 @lzeng14 @td36 @arunsbaskaran as this is a critical bug fix affecting multiple generations of Intel platforms.

niruiyu:
@os-d Thanks for the detailed explanation. I have now understood why the issue could happen.
Let me rephrase the issue in my way, and please review if my understanding matches your thoughts.

There are several places where Save/RestoreVolatileRegisters() is called.
The following 2 places are relevant for us to understand the bug.

1. In `ApWakeupFunction()`

```c
    While (TRUE) {
      RestoreVolatileRegisters ();
      ApProcedure();
      SaveVolatileRegisters ();
      Monitor-Mwait()
    }
```

2. In `DxeApEntryPoint()`

```c
     RestoreVolatileRegisters ();
     Monitor-Mwait()
     ApWakeupFunction();
```

When PEI and DXE both run in 64-bit mode, DXE MP code takes over APs by letting the PEI MP code call `DxeApEntryPoint()`.
But the `RestoreVolatileRegisters()` call in `DxeApEntryPoint()` only restores GDTR (to the same one used by BSP) and does NOT restore TR (leaving TR in a non-zero value).
 > TR is not restored because `MpInitLibInitialize()` manually resets VolatileRegister.TR to 0 and logic in `RestoreVolatielRegisters()` skips to restore TR when the value to be restored is 0.

This causes all APs to be in an inconsistent state: all AP's TR point to the same entry in the shared GDT. Usually, any bug may appear in an inconsistent state.

In our case (as you described patiently, thank you!!), When all APs wake up from `DxeApEntryPoint()::Monitor-Mwait()` (due to the 1st MP service call from DXE code), they execute `DxeApEntryPoint()::ApWakeupFunction()`.
Inside `ApWakeupFunction()`, `SaveVolatileRegisters()` is called, which saves the shared GDTR and non-zero TR to the VolatileRegister structure. The next time APs wake up (due to the 2nd MP service call from DXE code), `RestoreVolatileRegisters ()` is called inside `ApWakeupFunction()`, which writes the shared GDTR to all APs and non-zero to all APs. This is the first time `LTR` instruction is called in the DXE phase. When two APs execute `LTR` at the same time, GP may happen as the TSS is in a busy state.


Even though `RestoreVolatileRegisters()` contains the following check when restoring TR:

```c
  if ((VolatileRegisters->Tr != 0) &&
      (VolatileRegisters->Tr < VolatileRegisters->Gdtr.Limit))
  {
    Tss = (IA32_TSS_DESCRIPTOR *)(VolatileRegisters->Gdtr.Base +
                                  VolatileRegisters->Tr);
    if (Tss->Bits.P == 1) {
      Tss->Bits.Type &= 0xD;  // 1101 - Clear busy bit just in case
      AsmWriteTr (VolatileRegisters->Tr);
    }
  }
```
The shared GDT is inherited from the PEI phase, and it contains the TSS entry. So `VolatileRegisters->Tr < VolatileRegisters->Gdtr.Limit` is TRUE and non-zero TR is restored.

UefiCpuPkg/Library/MpInitLib/MpLib.c
```diff
@@ -2443,6 +2443,22 @@ MpInitLibInitialize (
...
+  // wakeup to switch contexts, we must ensure the APs do not attempt to set the TR as it can fault before they are
+  // able to execute the function to switch GDT/TSS.
+  for (Index = 0; Index < CpuMpData->CpuCount; Index++) {
+    if (Index == CpuMpData->BspNumber) {
+      continue;
+    }
+
+    CpuMpData->CpuData[Index].VolatileRegisters.Tr = 0;
+  }
+
```
niruiyu: This change does not work because the code may return early in line 2398.

niruiyu:
How about this change? Can it fix the issue?

```diff
diff --git a/UefiCpuPkg/Library/MpInitLib/MpLib.c b/UefiCpuPkg/Library/MpInitLib/MpLib.c
index 63b846445d..58d1cb537e 100644
--- a/UefiCpuPkg/Library/MpInitLib/MpLib.c
+++ b/UefiCpuPkg/Library/MpInitLib/MpLib.c
@@ -279,8 +279,27 @@ RestoreVolatileRegisters (
     Tss = (IA32_TSS_DESCRIPTOR *)(VolatileRegisters->Gdtr.Base +
                                   VolatileRegisters->Tr);
     if (Tss->Bits.P == 1) {
-      Tss->Bits.Type &= 0xD;  // 1101 - Clear busy bit just in case
-      AsmWriteTr (VolatileRegisters->Tr);
+      if ((Tss->Bits.Type & BIT1) == 0) {^M
+        //^M
+        // TR is used when stack guard feature is enabled to use a separate safe stack when^M
+        // stack overflow happens.^M
+        // When the stack guard is enabled in PEI phase, TR is not zero and each processor^M
+        // has its own GDT so TR (offset to GDT) points to different TSS entry in each processor.^M
+        //^M
+        // There is a small window in early DXE phase after MpInitLibInitialize() is called:^M
+        //   TR is inherited from PEI which is non-zero.^M
+        //     TR is not restored to 0 as this function skips restoring 0 to TR.^M
+        //   GDT is switched to a common GDT for all processors.^M
+        // SaveVolatileRegisters() is called from ApWakeupFunction() before AP goes to sleep, which^M
+        // saves the non-zero TR value to CpuMpData->CpuData[].VolatileRegisters.Tr, while the GDT^M
+        // is still common for all processors.^M
+        // The next time AP is woken up, RestoreVolatileRegisters() is called from ApWakeupFunction(),^M
+        // which restores the non-zero TR value from CpuMpData->CpuData[].VolatileRegisters.Tr, causing^M
+        // TR points to the same TSS entry in the common GDT.^M
+        // We should skip restoring TR in this case.^M
+        //^M
+        AsmWriteTr (VolatileRegisters->Tr);^M
+      }^M
     }
   }
 }
```

os-d:
> @os-d Thanks for the detailed explanation. I have now understood why the issue could happen. Let me rephrase the issue in my way, and please review if my understanding matches your thoughts.
> 
> There are several places where Save/RestoreVolatileRegisters() is called. The following 2 places are relevant for us to understand the bug.
> 
> 1. In `ApWakeupFunction()`
> 
> ```c
>     While (TRUE) {
>       RestoreVolatileRegisters ();
>       ApProcedure();
>       SaveVolatileRegisters ();
>       Monitor-Mwait()
>     }
> ```
> 
> 2. In `DxeApEntryPoint()`
> 
> ```c
>      RestoreVolatileRegisters ();
>      Monitor-Mwait()
>      ApWakeupFunction();
> ```
> 
> When PEI and DXE both run in 64-bit mode, DXE MP code takes over APs by letting the PEI MP code call `DxeApEntryPoint()`. But the `RestoreVolatileRegisters()` call in `DxeApEntryPoint()` only restores GDTR (to the same one used by BSP) and does NOT restore TR (leaving TR in a non-zero value).
> 
> > TR is not restored because `MpInitLibInitialize()` manually resets VolatileRegister.TR to 0 and logic in `RestoreVolatielRegisters()` skips to restore TR when the value to be restored is 0.
> 
> This causes all APs to be in an inconsistent state: all AP's TR point to the same entry in the shared GDT. Usually, any bug may appear in an inconsistent state.
> 
> In our case (as you described patiently, thank you!!), When all APs wake up from `DxeApEntryPoint()::Monitor-Mwait()` (due to the 1st MP service call from DXE code), they execute `DxeApEntryPoint()::ApWakeupFunction()`. Inside `ApWakeupFunction()`, `SaveVolatileRegisters()` is called, which saves the shared GDTR and non-zero TR to the VolatileRegister structure. The next time APs wake up (due to the 2nd MP service call from DXE code), `RestoreVolatileRegisters ()` is called inside `ApWakeupFunction()`, which writes the shared GDTR to all APs and non-zero to all APs. This is the first time `LTR` instruction is called in the DXE phase. When two APs execute `LTR` at the same time, GP may happen as the TSS is in a busy state.
> 
> Even though `RestoreVolatileRegisters()` contains the following check when restoring TR:
> 
> ```c
>   if ((VolatileRegisters->Tr != 0) &&
>       (VolatileRegisters->Tr < VolatileRegisters->Gdtr.Limit))
>   {
>     Tss = (IA32_TSS_DESCRIPTOR *)(VolatileRegisters->Gdtr.Base +
>                                   VolatileRegisters->Tr);
>     if (Tss->Bits.P == 1) {
>       Tss->Bits.Type &= 0xD;  // 1101 - Clear busy bit just in case
>       AsmWriteTr (VolatileRegisters->Tr);
>     }
>   }
> ```
> 
> The shared GDT is inherited from the PEI phase, and it contains the TSS entry. So `VolatileRegisters->Tr < VolatileRegisters->Gdtr.Limit` is TRUE and non-zero TR is restored.

Yes that matches my understanding, thanks!

os-d:
> How about this change? Can it fix the issue?
> 
> ```diff
> diff --git a/UefiCpuPkg/Library/MpInitLib/MpLib.c b/UefiCpuPkg/Library/MpInitLib/MpLib.c
> index 63b846445d..58d1cb537e 100644
> --- a/UefiCpuPkg/Library/MpInitLib/MpLib.c
> +++ b/UefiCpuPkg/Library/MpInitLib/MpLib.c
> @@ -279,8 +279,27 @@ RestoreVolatileRegisters (
>      Tss = (IA32_TSS_DESCRIPTOR *)(VolatileRegisters->Gdtr.Base +
>                                    VolatileRegisters->Tr);
>      if (Tss->Bits.P == 1) {
> -      Tss->Bits.Type &= 0xD;  // 1101 - Clear busy bit just in case
> -      AsmWriteTr (VolatileRegisters->Tr);
> +      if ((Tss->Bits.Type & BIT1) == 0) {^M
> +        //^M
> +        // TR is used when stack guard feature is enabled to use a separate safe stack when^M
> +        // stack overflow happens.^M
> +        // When the stack guard is enabled in PEI phase, TR is not zero and each processor^M
> +        // has its own GDT so TR (offset to GDT) points to different TSS entry in each processor.^M
> +        //^M
> +        // There is a small window in early DXE phase after MpInitLibInitialize() is called:^M
> +        //   TR is inherited from PEI which is non-zero.^M
> +        //     TR is not restored to 0 as this function skips restoring 0 to TR.^M
> +        //   GDT is switched to a common GDT for all processors.^M
> +        // SaveVolatileRegisters() is called from ApWakeupFunction() before AP goes to sleep, which^M
> +        // saves the non-zero TR value to CpuMpData->CpuData[].VolatileRegisters.Tr, while the GDT^M
> +        // is still common for all processors.^M
> +        // The next time AP is woken up, RestoreVolatileRegisters() is called from ApWakeupFunction(),^M
> +        // which restores the non-zero TR value from CpuMpData->CpuData[].VolatileRegisters.Tr, causing^M
> +        // TR points to the same TSS entry in the common GDT.^M
> +        // We should skip restoring TR in this case.^M
> +        //^M
> +        AsmWriteTr (VolatileRegisters->Tr);^M
> +      }^M
>      }
>    }
>  }
> ```

@niruiyu This may work. It relies on the fact that the BSP's TSS's busy bit will always be set by the time the APs run. I'm not familiar enough with the use of the busy bit to assert this, but if that is the case, then this should be safe to do.

I do still want to pair this with always setting up the separate GDT/TSS for each AP (which currently is only done when PcdCpuStackGuard is enabled) as you pointed out, this is the safer behavior and could prevent many such issues.

I will test this on the platforms experiencing the issue and let you know if this resolves it.

UefiCpuPkg/Library/MpInitLib/MpLib.c
os-d: You are correct. If we keep this approach, this would be moved to a goto Exit sort of thing.

os-d:
@niruiyu your approach booted successfully on the Intel client and server systems that were experiencing this GP fault. If you are confident the BSP's TSS's busy bit will always stay set to 1, then I am okay going with that approach and I can either update this PR to do that or have you open a PR to do it. Whatever you prefer.

Otherwise, if we are not confident, my approach (amended so that MpInitLibInitialize() has a goto to ensure this logic is always applied on exit) does not rely on the state of the BSP's TSS's busy bit, so I can update this PR to have the goto.

Let me know what you prefer.

niruiyu:
> @niruiyu your approach booted successfully on the Intel client and server systems that were experiencing this GP fault. If you are confident the BSP's TSS's busy bit will always stay set to 1, then I am okay going with that approach and I can either update this PR to do that or have you open a PR to do it. Whatever you prefer.
> 
> Otherwise, if we are not confident, my approach (amended so that MpInitLibInitialize() has a goto to ensure this logic is always applied on exit) does not rely on the state of the BSP's TSS's busy bit, so I can update this PR to have the goto.
> 
> Let me know what you prefer.


The busy bit in the TSS descriptor is set by hardware when "LTR" executes.
But hardware never clears the busy bit. Software is responsible for clearing the busy bit when it uses "LTR" to switch to a new TSS descriptor.

So I prefer my approach. However, please review the long comments there and adjust as you need to make them easier to understand.

os-d:
> > @niruiyu your approach booted successfully on the Intel client and server systems that were experiencing this GP fault. If you are confident the BSP's TSS's busy bit will always stay set to 1, then I am okay going with that approach and I can either update this PR to do that or have you open a PR to do it. Whatever you prefer.
> > Otherwise, if we are not confident, my approach (amended so that MpInitLibInitialize() has a goto to ensure this logic is always applied on exit) does not rely on the state of the BSP's TSS's busy bit, so I can update this PR to have the goto.
> > Let me know what you prefer.
> 
> The busy bit in the TSS descriptor is set by hardware when "LTR" executes. But hardware never clears the busy bit. Software is responsible for clearing the busy bit when it uses "LTR" to switch to a new TSS descriptor.
> 
> So I prefer my approach. However, please review the long comments there and adjust as you need to make them easier to understand.

@niruiyu I have edited this PR to go with your approach, in addition to always creating the separate GDT as we have discussed that is the safer path. I slightly edited your long comment, please feel free to make other suggestions.

niruiyu (commented):
I do not quite understand the intention of the first commit ("Always initialize separate AP exception stacks").

With 2nd commit, do we still need the fist commit?

os-d:
> I do not quite understand the intention of the first commit ("Always initialize separate AP exception stacks").
> 
> With 2nd commit, do we still need the fist commit?

@niruiyu the reason is twofold:
- As you said here https://github.com/tianocore/edk2/pull/11446#pullrequestreview-3162202309, I agree that the APs should always have their own GDTs (etc.). This is the safest option and prevents any other such issues from arising. It is already done with PcdCpuStackGuard enabled, so there is no behavior change there. For systems that didn't have PcdCpuStackGuard enabled, this still provides safety in having their own GDT and if they do take a double fault they can use the separate exception stack.
- The second reason is that in the PcdCpuStackGuard case, if a double fault were to occur (which can still occur even if the stack guard is not enabled) then the task register would be 0, which says GDT offset 0 is the TSS, which is not correct and we'd probably triple fault at that point.

To put it simply, it is safer and more correct for each AP to always have their own GDT, etc. To flip the question, why would we not want the APs to have their own control structures?

os-d:
@niruiyu friendly ping, if we are agreed, can we merge this?

niruiyu:
> @niruiyu friendly ping, if we are agreed, can we merge this?

I am not fully convinced about the first patch being necessary.

What do you think about what the StackGuard PCD controls? With your first patch, it seems to me that the stack guard feature should be turned on always, and no need to control it with a PCD.

My understanding of the PCD is that the PCD could be set to false if the platform owner has confidence in no occurrence of stack overflow. In theory, the confidence makes no sense, so the PCD should be set to TRUE at least in the DEBUG build to catch the stack overflow issues, assuming that the DEBUG build can expose all stack overflow issues.
With your first patch, I feel that we will have trouble defining what the PCD controls.

@mdkinney can you kindly review the first patch of this PR and comment?

os-d:
> > @niruiyu friendly ping, if we are agreed, can we merge this?
> 
> I am not fully convinced about the first patch being necessary.
> 
> What do you think about what the StackGuard PCD controls? With your first patch, it seems to me that the stack guard feature should be turned on always, and no need to control it with a PCD.
> 
> My understanding of the PCD is that the PCD could be set to false if the platform owner has confidence in no occurrence of stack overflow. In theory, the confidence makes no sense, so the PCD should be set to TRUE at least in the DEBUG build to catch the stack overflow issues, assuming that the DEBUG build can expose all stack overflow issues. With your first patch, I feel that we will have trouble defining what the PCD controls.
> 
> @mdkinney can you kindly review the first patch of this PR and comment?

@niruiyu I believe PcdCpuStackGuard should not exist and the stack should always have a guard page. I think that is a critical and decades old security/safety feature. I'm happy to take that up in a separate PR to drop it. No platform should ship without it, it leaves open a classic attack surface.

However, this aside, the first commit in this PR does not change applying the guard page. That is still done based on the PCD:

BSP:
https://github.com/tianocore/edk2/blob/1e7a83cbb6f26f02f743c06aeafee164a3ec59bf/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c#L814 

APs:
https://github.com/tianocore/edk2/blob/1e7a83cbb6f26f02f743c06aeafee164a3ec59bf/UefiCpuPkg/Library/MpInitLib/DxeMpLib.c#L584

All the first commit does is ensure we have separate IDT/GDT for the APs and a separate stack for exceptions, all of which is reasonable regardless of PcdCpuStackGuard: https://github.com/tianocore/edk2/blob/1e7a83cbb6f26f02f743c06aeafee164a3ec59bf/UefiCpuPkg/Library/CpuExceptionHandlerLib/X64/ArchExceptionHandler.c#L125.

So I don't think this adds confusion to what PcdCpuStackGuard controls. PcdCpuStackGuard controls whether or not a guard page is placed at the end of the stack. That remains unchanged.

niruiyu:
Ok. I now understand your point. The PCD only controls the presence of the guard page in the stack bottom.
But no matter whether the PCD is true or false, #PF and #DF use a separate stack.

But, you missed to change the code in DxeMain.c:
```c
  if (PcdGetBool (PcdCpuStackGuard)) {
    Status = InitializeSeparateExceptionStacks (NULL, NULL);
    ASSERT_EFI_ERROR (Status);
  }
```

os-d:
> Ok. I now understand your point. The PCD only controls the presence of the guard page in the stack bottom. But no matter whether the PCD is true or false, #PF and #DF use a separate stack.
> 
> But, you missed to change the code in DxeMain.c:
> 
> ```c
>   if (PcdGetBool (PcdCpuStackGuard)) {
>     Status = InitializeSeparateExceptionStacks (NULL, NULL);
>     ASSERT_EFI_ERROR (Status);
>   }
> ```

@niruiyu , yep exactly. Thanks, I have added the commit, please re-review.

@lgao4 , this PR now contains a change to MdeModulePkg, mirroring the UefiCpuPkg change. Please review.

lgao4:
> > Ok. I now understand your point. The PCD only controls the presence of the guard page in the stack bottom. But no matter whether the PCD is true or false, #PF and #DF use a separate stack.
> > But, you missed to change the code in DxeMain.c:
> > ```c
> >   if (PcdGetBool (PcdCpuStackGuard)) {
> >     Status = InitializeSeparateExceptionStacks (NULL, NULL);
> >     ASSERT_EFI_ERROR (Status);
> >   }
> > ```
> 
> @niruiyu , yep exactly. Thanks, I have added the commit, please re-review.
> 
> @lgao4 , this PR now contains a change to MdeModulePkg, mirroring the UefiCpuPkg change. Please review.

The change in MdeModulePkg is good to me.

os-d:
@lgao4 @niruiyu I fixed the CI break on EmulatorPkg. @lgao4 I added another commit changing CpuExceptionHandlerLibNull to return EFI_SUCCESS for InitializeSeparateExceptionStacks instead of EFI_UNSUPPORTED. Now that DxeMain.c calls it unconditionally, EmulatorPkg would hit the assert on that function failing and die. When it returns success, we succeed.

If there are no other comments, can we merge this? Hoping to fix downstream broken platforms.

PR #11448, opened 2025-08-28
OvmfPkg: RiscVVirt: Add Secure Boot and TPM support

pttuan:
# Description

This patchset adds Secure Boot and TPM support for the RISC-V architecture and the RiscVVirt platform.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- Started a TPM emulator using swtpm:
`swtpm socket --tpmstate dir=/tmp/mytpm1 --ctrl type=unixio,path=/tmp/mytpm1/swtpm-sock --tpm2 --log level=20 --daemon`
- Launched QEMU with TPM support enabled:
`-chardev socket,id=chrtpm,path=/tmp/mytpm1/swtpm-sock -tpmdev emulator,id=tpm0,chardev=chrtpm -device tpm-tis-device,tpmdev=tpm0`

## Integration Instructions
N/A

### DIFF
--- a/OvmfPkg/RiscVVirt/Feature/SecureBoot/SecureBootDefaultKeysInit/SecureBootDefaultKeysInit.c
+++ b/OvmfPkg/RiscVVirt/Feature/SecureBoot/SecureBootDefaultKeysInit/SecureBootDefaultKeysInit.c
@@ -1,0 +1,643 @@
+/** @file
+  This driver init default Secure Boot variables
+
+  Copyright (c) 2011 - 2018, Intel Corporation. All rights reserved.<BR>
+  (C) Copyright 2018 Hewlett Packard Enterprise Development LP<BR>
+  Copyright (c) 2021, ARM Ltd. All rights reserved.<BR>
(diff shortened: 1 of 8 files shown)


kraxel:
Noticed this because it breaks my build.  A few notes on the secure boot changes:

Adding `SecureBootDefaultKeysInit` is a valid approach.  It is not riscv specific though, so IMHO it should be moved out of the riscv subdirectory.  It should get a config option, because it does not build out-of-the box (needs certificates added to the tree).  Also this is not the only way to enroll keys, so it is not a hard dependency for secure boot support.

Alternative approaches to enroll keys:
 * create some boot media with `Shell.efi` and `EnrollDefaultKeys.efi`.
 * use https://gitlab.com/kraxel/virt-firmware to create a json file for the qemu variable store.

Additional notes:
 * Adding the x64 dbx database to riscv64 is pointless :)
 * Not sure what microsofts plans for signing riscv efi binaries are, but I'd expect you will not see any signed with the old (2011) signing keys, only with the new (2023) keys.

PR #11454, opened 2025-08-29
BaseTools: Enable 4k alignment for x64 tool chains.

apop5:
# Description

Found while compiling with ClangDwarf. 
Binaries were not producing 4k aligned executables. 
Modified the tool chains to generate 4k alignment. 
Tool chains modified are CLANGDWARF and GCC.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Local platform build.

## Integration Instructions
No integration required.
Size increases in binaries with specific tool chains may occur.


niruiyu (commented):
VS** toolchain uses 32 Byte alignment.
Why the GCC uses 4K?

niruiyu:
> VS** toolchain uses 32 Byte alignment. Why the GCC uses 4K?

And in x64 PEI system, 4K alignment for XIP code wastes SPI space.

os-d:
> VS** toolchain uses 32 Byte alignment. Why the GCC uses 4K?

@niruiyu VS2022 uses 4K alignment (/ALIGN:4096): https://github.com/tianocore/edk2/blob/65485e195f37653bbbc06db2001c3a189d468666/BaseTools/Conf/tools_def.template#L833-L835

> And in x64 PEI system, 4K alignment for XIP code wastes SPI space.

I agree. @apop5 , we should add XIPFLAGS like we did for VS2022 and ARM64: https://github.com/tianocore/edk2/blob/65485e195f37653bbbc06db2001c3a189d468666/BaseTools/Conf/tools_def.template#L838

https://github.com/tianocore/edk2/blob/65485e195f37653bbbc06db2001c3a189d468666/BaseTools/Conf/tools_def.template#L1323

That resolves the flash size problem for XIP modules, which aren't getting memory protections anyway.

BaseTools/Conf/tools_def.template
```diff
@@ -5,7 +5,7 @@
 #  Copyright (c) 2015, Hewlett-Packard Development Company, L.P.<BR>
 #  (C) Copyright 2020, Hewlett Packard Enterprise Development LP<BR>
 #  Copyright (c) 2022, Loongson Technology Corporation Limited. All rights reserved.<BR>
-#  Copyright (c) Microsoft Corporation
+#  Copyright (c) Microsoft Corporation.<BR>
```
os-d: nit: Not needed

niruiyu:
I just found that 4096 is used for VS2022. But I didn't notice that because /ALIGN:32 is in the DLINK flag for VS2022 as well. Because my monitor is small I only see the first /ALIGN:32 but not the second /ALIGN:4096.
Maybe we should remove /ALIGN:32 with /ALIGN:4096.

And how does following work?
```
# BaseTools cannot handle XIP modules with different section and file alignment
*_VS2022_X64_DLINK_XIPFLAGS     = /ALIGN:32
```

niruiyu:
> XIPFLAGS

I saw the following build rule, which uses DLINK_XIPFLAGS.
But it applies to all SEC/PEI_CORE/PEIM. right?
```
[Static-Library-File.SEC, Static-Library-File.PEI_CORE, Static-Library-File.PEIM]
```

os-d:
> I just found that 4096 is used for VS2022. But I didn't notice that because /ALIGN:32 is in the DLINK flag for VS2022 as well. Because my monitor is small I only see the first /ALIGN:32 but not the second /ALIGN:4096. Maybe we should remove /ALIGN:32 with /ALIGN:4096.

Yeah, there is no reason for both and `/ALIGN:32` can be removed.





> > XIPFLAGS
> 
> I saw the following build rule, which uses DLINK_XIPFLAGS. But it applies to all SEC/PEI_CORE/PEIM. right?
> 
> ```
> [Static-Library-File.SEC, Static-Library-File.PEI_CORE, Static-Library-File.PEIM]
> ```

Correct. The build system has no way of knowing whether a given file is XIP or not (there was a PR to add this knowledge but it was dropped; we are trying to get it resurrected https://github.com/tianocore/edk2/pull/10754), so XIPFLAGS applies to all of SEC, PEI_CORE, and PEIM. Which is fine since we don't apply memory protections in SEC/PEI anyway, so we don't need 4K sections.

os-d:
@apop5 this PR should still be valid, reopening. Can you add the XIPFLAGS and get BaseTools maintainers review?

apop5:
> @apop5 this PR should still be valid, reopening. Can you add the XIPFLAGS and get BaseTools maintainers review?

@os-d 
I remember something in the past where common-page-size does not do well with an override? Am I remembering this correctly?

apop5:
> > I just found that 4096 is used for VS2022. But I didn't notice that because /ALIGN:32 is in the DLINK flag for VS2022 as well. Because my monitor is small I only see the first /ALIGN:32 but not the second /ALIGN:4096. Maybe we should remove /ALIGN:32 with /ALIGN:4096.
> 
> Yeah, there is no reason for both and `/ALIGN:32` can be removed.
> 
> > > XIPFLAGS
> > 
> > 
> > I saw the following build rule, which uses DLINK_XIPFLAGS. But it applies to all SEC/PEI_CORE/PEIM. right?
> > ```
> > [Static-Library-File.SEC, Static-Library-File.PEI_CORE, Static-Library-File.PEIM]
> > ```
> 
> Correct. The build system has no way of knowing whether a given file is XIP or not (there was a PR to add this knowledge but it was dropped; we are trying to get it resurrected #10754), so XIPFLAGS applies to all of SEC, PEI_CORE, and PEIM. Which is fine since we don't apply memory protections in SEC/PEI anyway, so we don't need 4K sections.

Duplicated /ALIGN:32 is removed.

os-d:
> > @apop5 this PR should still be valid, reopening. Can you add the XIPFLAGS and get BaseTools maintainers review?
> 
> @os-d I remember something in the past where common-page-size does not do well with an override? Am I remembering this correctly?

We had one crazy summer where GCC was parsing arguments in a different order than last is what is taken, but that depended on the phase of the moon and we never saw it again. Edk2 already depends on overriding args by passing them again on the cmd line, so we are okay to do it.

apop5:
@ardbiesheuvel @lgao4 @ponchofigueroa @mikebeaton

When you get a chance, can you please review this PR?

The goal is to get x64 images (Dxe, Smm/standalone Mm) to have 4k image section alignment to match MSVC's 4k alignment. This is to allow memory protections, etc, to be applied to images loaded during this timeframe.

lgao4:
For VS tool chain, I am OK for this change.

BaseTools/Conf/build_rule.template
```diff
@@ -338,7 +338,7 @@
         $(RM) ${dst}.wtemp
         "$(DLINK)" /OUT:${dst} $(DLINK_FLAGS) $(DLINK_XIPFLAGS) $(DLINK_SPATH) @$(STATIC_LIBRARY_FILES_LIST)
 
-    <Command.GCC>
+    <Command.GCC, Command.CLANGDWARF>
```
mdkinney, line 341: So this is required to get XIPFLAGS into CLANGDWARF.

However, CLANGPDB is not listed in this rule, so XIPFLAGS will not be applied to CLANGPDB build.  I think CLANGPDB should be added to <Command.MSFT> section of this rule.

BaseTools/Conf/tools_def.template
```diff
@@ -1978,14 +1985,17 @@ NOOPT_CLANGDWARF_X64_NASM_FLAGS       = -O0 -f elf64 -g
...
 DEBUG_CLANGDWARF_X64_DLINK2_FLAGS     = DEF(CLANGDWARF_X64_DLINK2_FLAGS) -O3 -fuse-ld=lld
 
 RELEASE_CLANGDWARF_X64_CC_FLAGS       = DEF(CLANGDWARF_ALL_CC_FLAGS) -m64 "-DEFIAPI=__attribute__((ms_abi))" -mno-red-zone -mcmodel=small -fpie -fdirect-access-external-data -Oz -flto DEF(CLANGDWARF_X64_TARGET)
 RELEASE_CLANGDWARF_X64_DLINK_FLAGS    = DEF(CLANGDWARF_X64_TARGET) DEF(CLANGDWARF_IA32_X64_DLINK_FLAGS) -flto -Wl,-O3 -Wl,-melf_x86_64 -Wl,--oformat,elf64-x86-64 -Wl,-pie -mcmodel=small -Wl,--apply-dynamic-relocs
+RELEASE_CLANGDWARF_X64_DLINK_XIPFLAGS = -z common-page-size=0x40
 RELEASE_CLANGDWARF_X64_DLINK2_FLAGS   = DEF(CLANGDWARF_X64_DLINK2_FLAGS) -O3 -fuse-ld=lld
 
 NOOPT_CLANGDWARF_X64_CC_FLAGS         = DEF(CLANGDWARF_ALL_CC_FLAGS) -m64 "-DEFIAPI=__attribute__((ms_abi))" -mno-red-zone -mcmodel=small -fpie -fdirect-access-external-data -O0 DEF(CLANGDWARF_X64_TARGET) -g
 NOOPT_CLANGDWARF_X64_DLINK_FLAGS      = DEF(CLANGDWARF_X64_TARGET) DEF(CLANGDWARF_IA32_X64_DLINK_FLAGS) -Wl,-O0 -Wl,-melf_x86_64 -Wl,--oformat,elf64-x86-64 -Wl,-pie -mcmodel=small -Wl,--apply-dynamic-relocs
+NOOPT_CLANGDWARF_X64_DLINK_XIPFLAGS   = -z common-page-size=0x40
```
mdkinney: Why are these X64 settings 0x40 when other X64 settings above are 0x20?

BaseTools/Conf/build_rule.template
apop5: I only ran a test with GCC and ClangDwarf. let me run a test for clangpdb.

BaseTools/Conf/tools_def.template
apop5: DEFINE CLANGDWARF_IA32_X64_DLINK_COMMON   = -nostdlib -Wl,-q,--gc-sections -z common-page-size=0x40 

Clangdwarf has been using a common-page-size of 0x40

GCC48 is using 0x20
GCC49, GCC5, GCC are using 0x40
I don't know if this is an oversight, or desired behavior, but it has been set this way for the past 10 years. 

I was attempting to have XIP match the current settings that exist today, but I am not opposed to getting a matching set.

BaseTools/Conf/build_rule.template
apop5: you are correct, clangpdb needed updated for IA32/x64 as well. The alignment has been changed to 4k for x64 non XIP modules.

BaseTools/Conf/tools_def.template
mdkinney: My guess is that at some point the alignment had to be increased from 0x20 to 0x40 due to build failures, but GCC48 was not updated at that time and perhaps has not been used in a case that fails.  

Probably best to just leave GCC48 at 0x20 for now and look into this topic further when we review PE/COFF alignment  topic from UEFI Plugfest https://github.com/tianocore/edk2/discussions/11573

ardbiesheuvel: This is explained here:

```
commit dafe0fedc5085b6f117068ef1faacb6176512d87
Author: Jordan Justen <jordan.l.justen@intel.com>
Date:   Mon Jul 28 17:37:10 2014 +0000

    BaseTools: Add GCC49 toolchain; align data sections to 0x40

    GCC 4.9 may use 64-byte (0x40) alignment for data sections.
    
    Therefore we use a different link script for GCC 4.9. The only
    difference from the gcc4.4-ld-script is the alignment for data
    sections.
    
    When using the GCC48 toolchain with GCC 4.9, this error would be
    encountered by GenFw:
    > GenFw: ERROR 3000: Invalid
    >   Unsupported section alignment.
```

os-d:
@apop5 pinging on this one, I think we should get this in. See https://github.com/tianocore/edk2/discussions/12280 for another angle to this.

apop5:
> @apop5 pinging on this one, I think we should get this in. See #12280 for another angle to this.

Thanks for reminding me. Moved changes on top of latest tools_def and build_rule and ran builds with following results:

| Component | SectionAlignment | FileAlignment | Path |
|---|---:|---:|---|
| PeiCore | 64 | 64 | Build/MdeModule/DEBUG_CLANGDWARF/IA32/PeiCore.efi |
| DxeCore | 64 | 64 | Build/MdeModule/DEBUG_CLANGDWARF/IA32/DxeCore.efi |
| PeiCore | 64 | 64 | Build/MdeModule/DEBUG_CLANGDWARF/X64/PeiCore.efi |
| DxeCore | 4096 | 4096 | Build/MdeModule/DEBUG_CLANGDWARF/X64/DxeCore.efi |
| PeiCore | 64 | 64 | Build/MdeModule/DEBUG_GCC/IA32/PeiCore.efi |
| DxeCore | 64 | 64 | Build/MdeModule/DEBUG_GCC/IA32/DxeCore.efi |
| PeiCore | 64 | 64 | Build/MdeModule/DEBUG_GCC/X64/PeiCore.efi |
| DxeCore | 4096 | 4096 | Build/MdeModule/DEBUG_GCC/X64/DxeCore.efi |
| PeiCore | 32 | 32 | Build/MdeModule/DEBUG_CLANGPDB/IA32/PeiCore.efi |
| DxeCore | 32 | 32 | Build/MdeModule/DEBUG_CLANGPDB/IA32/DxeCore.efi |
| PeiCore | 32 | 32 | Build/MdeModule/DEBUG_CLANGPDB/X64/PeiCore.efi |
| DxeCore | 4096 | 32 | Build/MdeModule/DEBUG_CLANGPDB/X64/DxeCore.efi |

apop5:
@ardbiesheuvel @niruiyu @mdkinney @os-d @mikebeaton 

This should be ready for review again.

BaseTools/Conf/tools_def.template
```diff
@@ -26,6 +26,7 @@
 # 3.04 - Add toolchain for VS2026
 # 3.05 - Add CLANGPDB AARCH64 Support
 # 3.06 - Remove GCC48, GCC49 and GCC5
+# 3.07 - Swich alignemnt to 4k for x86 and X64 builds.
```
os-d:
```suggestion
# 3.07 - Switch alignment to 4k for x86 and X64 builds.
```

BaseTools/Conf/tools_def.template
```diff
@@ -930,9 +931,9 @@ DEFINE GCC_AARCH64_CC_COMMON           = DEF(GCC_ALL_CC_COMMON) -mlittle-endian
...
 DEFINE GCC_RISCV64_CC_XIPFLAGS         = -mstrict-align -mgeneral-regs-only
 DEFINE GCC_DLINK2_FLAGS_COMMON         = -Wl,--script=$(EDK_TOOLS_PATH)/Scripts/GccBase.lds
-DEFINE GCC_AARCH64_DLINK_COMMON        = -Wl,--emit-relocs -nostdlib -Wl,--gc-sections -u $(IMAGE_ENTRY_POINT) -Wl,-e,$(IMAGE_ENTRY_POINT),-Map,$(DEST_DIR_DEBUG)/$(BASE_NAME).map -z common-page-size=0x20
+DEFINE GCC_AARCH64_DLINK_COMMON        = -Wl,--emit-relocs -nostdlib -Wl,--gc-sections -u $(IMAGE_ENTRY_POINT) -Wl,-e,$(IMAGE_ENTRY_POINT),-Map,$(DEST_DIR_DEBUG)/$(BASE_NAME).map
 DEFINE GCC_LOONGARCH64_DLINK_COMMON    = -Wl,--emit-relocs -nostdlib -Wl,--gc-sections -u $(IMAGE_ENTRY_POINT) -Wl,-e,$(IMAGE_ENTRY_POINT),-Map,$(DEST_DIR_DEBUG)/$(BASE_NAME).map -z common-page-size=0x20
-DEFINE GCC_AARCH64_ASLDLINK_COMMON     = DEF(GCC_AARCH64_DLINK_COMMON) -Wl,--entry,ReferenceAcpiTable -u $(IMAGE_ENTRY_POINT) -Wl,--defsym=PECOFF_HEADER_SIZE=0 DEF(GCC_DLINK2_FLAGS_COMMON) -z common-page-size=0x20
+DEFINE GCC_AARCH64_ASLDLINK_COMMON     = DEF(GCC_AARCH64_DLINK_COMMON) -Wl,--entry,ReferenceAcpiTable -u $(IMAGE_ENTRY_POINT) -Wl,--defsym=PECOFF_HEADER_SIZE=0 DEF(GCC_DLINK2_FLAGS_COMMON)
```
os-d: Are you intended to change the ASLDLINK flags also?

BaseTools/Conf/tools_def.template
```diff
@@ -1202,13 +1203,15 @@ RELEASE_GCC_IA32_DLINK_FLAGS    = DEF(GCC_IA32_X64_DLINK_FLAGS) -flto -Os -Wl,-m
 *_GCC_X64_NASM_FLAGS            = -f elf64
 
   DEBUG_GCC_X64_CC_FLAGS        = DEF(GCC_X64_CC_FLAGS) -flto -DUSING_LTO
-  DEBUG_GCC_X64_DLINK_FLAGS     = DEF(GCC_X64_DLINK_FLAGS) -flto -Os
+  DEBUG_GCC_X64_DLINK_FLAGS     = DEF(GCC_X64_DLINK_FLAGS) -flto -Os -z common-page-size=0x1000
+  DEBUG_GCC_X64_DLINK_XIPFLAGS  = -z common-page-size=0x40
```
os-d: Because the XIPFLAGS are the same, you can do `*_GCC_X64_DLINK_XIPLAGS` at the end, a la https://github.com/tianocore/edk2/blob/e2b0e207f7b1009a6729f7973ed17242cb6672a4/BaseTools/Conf/tools_def.template#L735. Because we'll also need the same for NOOPT, right?

BaseTools/Conf/tools_def.template
```diff
@@ -1419,6 +1422,8 @@ NOOPT_CLANGPDB_X64_DLINK_FLAGS      = /NOLOGO /NODEFAULTLIB /IGNORE:4001 /OPT:RE
 NOOPT_CLANGPDB_X64_DLINK2_FLAGS     =
 NOOPT_CLANGPDB_X64_GENFW_FLAGS      = --keepexceptiontable
 
+*_CLANGPDB_X64_DLINK_XIPFLAGS       = /ALIGN:32
```
os-d, line 1423: Probably worth the comment on what this is, even though it is commented in the VS2022 ones.

BaseTools/Conf/tools_def.template
```diff
@@ -1570,14 +1575,17 @@ NOOPT_CLANGDWARF_X64_NASM_FLAGS       = -O0 -f elf64 -g
 DEBUG_CLANGDWARF_X64_CC_FLAGS         = DEF(CLANGDWARF_ALL_CC_FLAGS) -m64 "-DEFIAPI=__attribute__((ms_abi))" -mno-red-zone -mcmodel=small -fpie -fdirect-access-external-data -Oz -flto DEF(CLANGDWARF_X64_TARGET) -g
 DEBUG_CLANGDWARF_X64_DLINK_FLAGS      = DEF(CLANGDWARF_X64_TARGET) DEF(CLANGDWARF_IA32_X64_DLINK_FLAGS) -flto -Wl,-O3 -Wl,-melf_x86_64 -Wl,--oformat,elf64-x86-64 -Wl,-pie -mcmodel=small -Wl,--apply-dynamic-relocs
```
os-d: 4K on CLANGDWARF?

BaseTools/Conf/tools_def.template
apop5: NOOPT is running unit tests on host system. I am hesitant to change the NOOPT because it has repercussions for host based unit test execution

mdkinney: Unit tests can do their own overrides of tools_def.txt settings in UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc.

Use of NOOPT is valid for FW builds.  It is typically used to help improve debug when working on race conditions or  when the optimizing compiler makes it hard to view the intermediate values of local variables that can be optimized into registers.

apop5: I'll switch it to apply to all builds.

Based on a previous test, this will have impact on ovmf for some size increases.

BaseTools/Conf/tools_def.template
apop5: clangdwarf, if no common-page-size is set, will default to 4k alignment.

if we want to be explicit about it, I can add the common-page-size

BaseTools/Conf/tools_def.template
apop5: no, reverted back

BaseTools/Conf/tools_def.template
os-d: I think we should be explicit. This is behavior we require for security, shouldn't rely on the default potentially changing

apop5: Made it explicit 4k

BaseTools/Conf/tools_def.template
```diff
@@ -1237,15 +1241,15 @@ RELEASE_GCC_X64_DLINK_FLAGS     = DEF(GCC_X64_DLINK_FLAGS) -flto -Os
 
   DEBUG_GCC_AARCH64_CC_FLAGS    = DEF(GCC_AARCH64_CC_FLAGS) -flto
   DEBUG_GCC_AARCH64_DLINK_FLAGS = DEF(GCC_AARCH64_DLINK_FLAGS) -flto -Os -L$(WORKSPACE)/BaseTools/Bin/GccLto -llto-aarch64 -Wl,-plugin-opt=-pass-through=-llto-aarch64 -Wno-lto-type-mismatch
```
os-d: These need `-z common-page-size=0x1000`? I don't see the common flags setting it. I would set it in the common flags, personally, since GCC AARCH64 doesn't have any difference in behavior.

os-d: This PR does say it is scoped to X64, so I'd either not change the AARCH64 one now or amend your PR title/description. I do agree it should be changed.

BaseTools/Conf/tools_def.template
```diff
@@ -1236,16 +1240,16 @@ RELEASE_GCC_X64_DLINK_FLAGS     = DEF(GCC_X64_DLINK_FLAGS) -flto -Os
...
-RELEASE_GCC_AARCH64_DLINK_FLAGS = DEF(GCC_AARCH64_DLINK_FLAGS) -flto -Os -L$(WORKSPACE)/BaseTools/Bin/GccLto -llto-aarch64 -Wl,-plugin-opt=-pass-through=-llto-aarch64 -Wno-lto-type-mismatch
-RELEASE_GCC_AARCH64_DLINK_XIPFLAGS = -z common-page-size=0x20
+RELEASE_GCC_AARCH64_DLINK_FLAGS = DEF(GCC_AARCH64_DLINK_FLAGS) -flto -Os -L$(WORKSPACE)/BaseTools/Bin/GccLto -llto-aarch64 -Wl,-plugin-opt=-pass-through=-llto-aarch64 -Wno-lto-type-mismatch -z common-page-size=0x1000
 
   NOOPT_GCC_AARCH64_CC_FLAGS    = DEF(GCC_AARCH64_CC_FLAGS) -O0
-  NOOPT_GCC_AARCH64_DLINK_FLAGS = DEF(GCC_AARCH64_DLINK_FLAGS) -O0
-  NOOPT_GCC_AARCH64_DLINK_XIPFLAGS = -z common-page-size=0x20 -O0
+  NOOPT_GCC_AARCH64_DLINK_FLAGS = DEF(GCC_AARCH64_DLINK_FLAGS) -O0 -z common-page-size=0x1000
+
+# BaseTools cannot handle XIP modules with different section and file alignment
```
ardbiesheuvel: It is not a BaseTools limitation, but a fundamental property of XIP execution that the section (i.e., memory) and file alignment are the same, no? The file is essentially mapped in memory directly rather than loaded by a PE loader, and so if the file layout deviates from the memory layout, things will be fundamentally broken.

So I don't think it is helpful to frame this as a BaseTools limitation (which could potentially be fixed). If you want to clarify this, explain that the file and memory views of XIP executables are necessarily identical.

os-d: True. The BaseTools limitation is that all modules in an FV with XIP modules must have the same file alignment and section alignment, even if the modules are not XIP themselves. There actually was a PR up that fixed that, introducing a concept of XIP in an FDF to tell BaseTools which modules needed the same file alignment and section alignment, unfortunately that PR was closed and dropped by the submitter. We have it on our backlog to resurrect, but haven’t gotten to it.

That’s why all SEC and PEI modules get the “XIPFLAGS” applied to them (in build_rules.template), because at this point in time BaseTools doesn’t know which modules are XIP and which aren’t and so believes all modules in an uncompressed FV that is indicated it should be rebased should have the same file and section alignment.

I had originally asked the clarifying comment, because so much of tools_def.template is not commented, but we can also drop the comment. Agreed that if it is there, it should be accurate.

apop5: Updated to remove comments.

apop5:
rebased to master and force pushed to get clandpdb workflow to run.

mdkinney:
@apop5 Unfortunately, this change seems to have introduced a regression for EmulatorPkg X64 builds for Windows DEBUG CLANGPDB.  Not sure how this is missed in edk2 CI with CLANGPDB being added recently. 

```
INFO - 
INFO - Fd File Name:FV_RECOVERY (D:\a\edk2\edk2\Build\EmulatorX64\DEBUG_CLANGPDB\FV\FV_RECOVERY.fd)
INFO - 
INFO - Generate Region at Offset 0x0
INFO -    Region Size = 0xC80000
INFO -    Region Name = FV
INFO - 
INFO - Generating FVRECOVERY FV
INFO - ######## ['GenFv', '-a', 'D:\\a\\edk2\\edk2\\Build\\EmulatorX64\\DEBUG_CLANGPDB\\FV\\Ffs\\FVRECOVERY.inf', '-o', 'D:\\a\\edk2\\edk2\\Build\\EmulatorX64\\DEBUG_CLANGPDB\\FV\\FVRECOVERY.Fv', '-i', 'D:\\a\\edk2\\edk2\\Build\\EmulatorX64\\DEBUG_CLANGPDB\\FV\\FVRECOVERY.inf']
INFO - Return Value = 2
INFO - GenFv: ERROR 3000: Invalid
INFO -   PE image Section-Alignment and File-Alignment do not match : D:\a\edk2\edk2\Build\EmulatorX64\DEBUG_CLANGPDB\FV\Ffs\BCAF98C9-22B0-3B4F-9CBD-C8A6B4DBCEE9EmuSec\BCAF98C9-22B0-3B4F-9CBD-C8A6B4DBCEE9.ffs.
INFO - GenFv: ERROR 3000: Invalid
INFO -   Could not rebase D:\a\edk2\edk2\Build\EmulatorX64\DEBUG_CLANGPDB\FV\Ffs\BCAF98C9-22B0-3B4F-9CBD-C8A6B4DBCEE9EmuSec\BCAF98C9-22B0-3B4F-9CBD-C8A6B4DBCEE9.ffs.
```

And Windows Mingw EmulatorPkg X64 DEBUG CLANGDWARF.

```
INFO - lld: error: unknown argument: -z
INFO - clang-20: error: linker command failed with exit code 1 (use -v to see invocation)
INFO - mingw32-make: *** [GNUmakefile:383: D:\a\edk2\edk2\Build\EmulatorX64Mingw\DEBUG_CLANGDWARF\X64\MdeModulePkg\Universal\FaultTolerantWritePei\FaultTolerantWritePei\DEBUG\FaultTolerantWritePei.dll] Error 1
INFO - 
INFO - 
INFO - build.py...
INFO -  : error 7000: Failed to execute command
INFO - 	mingw32-make tbuild [D:\a\edk2\edk2\Build\EmulatorX64Mi
```

mikebeaton:
> @apop5 Unfortunately, this change seems to have introduced a regression for EmulatorPkg X64 builds for Windows DEBUG CLANGPDB. Not sure how this is missed in edk2 CI with CLANGPDB being added recently.

It was explicitly excluded, as EmulatorPkg was not building under CLANGPDB at the time I first prepared the CI PR.

I'll put up a PR to add it back.

mikebeaton:
~https://github.com/tianocore/edk2/pull/12418~ **EDIT**: https://github.com/tianocore/edk2/pull/12431

PR #11455, opened 2025-08-29
PrmPkg: PrmSample Make build options Clang/Gcc compatible.

apop5:
# Description

While attempting to compile PrmSampleAcpiParameterBufferModule and PrmSampleContextBufferModule with ClangDwarf under windows, errors occured because PRM used custom buildoption --require-defined. This is not universally supported by all clang linkers, specifically the ld.lld that is distributed with windows clang installations.

Switch to using -Wl,-u,<symbol to export> ensures that the associated symbol is in the final binary, even if its not referenced, and it is supported by both clang and gcc.

- [ ] Breaking change?

- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Local platform build

## Integration Instructions
No integration necessary.

### DIFF
--- a/PrmPkg/Include/Prm.h
+++ b/PrmPkg/Include/Prm.h
@@ -15,9 +15,9 @@
 #include <PrmContextBuffer.h>
 
 #if defined (_MSC_VER)
-#define PRM_EXPORT_API  __declspec(dllexport)
+#define PRM_EXPORT_API  volatile __declspec(dllexport)
 #elif defined (__GNUC__)
-#define PRM_EXPORT_API  __attribute__ ((visibility ("default")))
+#define PRM_EXPORT_API  __attribute__((used)) __attribute__ ((visibility ("default")))
 #else
 #define PRM_EXPORT_API
 #endif
--- a/PrmPkg/Samples/PrmSampleAcpiParameterBufferModule/PrmSampleAcpiParameterBufferModule.inf
+++ b/PrmPkg/Samples/PrmSampleAcpiParameterBufferModule/PrmSampleAcpiParameterBufferModule.inf
@@ -43,5 +43,4 @@
   MSFT:*_*_*_GENFW_FLAGS = --keepoptionalheader
 
   GCC:*_*_*_GENFW_FLAGS = --prm
-  GCC:*_*_*_DLINK_FLAGS = -Wl,--no-gc-sections -Wl,--require-defined=PrmModuleExportDescriptor -Wl,--require-defined=CheckParamBufferPrmHandler
-  GCC:*_*_X64_OBJCOPY_STRIPFLAG = --keep-symbol=PrmModuleExportDescriptor --keep-symbol=CheckParamBufferPrmHandler
+  GCC:*_*_*_DLINK_FLAGS = -Wl,--no-gc-sections
--- a/PrmPkg/Readme.md
+++ b/PrmPkg/Readme.md
@@ -73,11 +73,6 @@ on IA32 with the unique UEFI configuration required.
 * AArch64
     ``build -p PrmPkg/PrmPkg.dsc -a AARCH64 -t GCC5``
 
-   > ***Note***: Only builds with the GCC5 toolchain have been tested.
-   > ***Note***: For builds with the GCC5 toolchain, the PrmModuleExportDescriptor and any other handler entry points
-   symbols, to be listed in the PRMT, must be explicitly preserved by enumerating these in the AARCH64 linker flags.
-   The --require-defined linker flag must be used for each symbol to be preserved.
-
 ### PRM Platform GUID
 
 **IMPORTANT** PRM has a concept of a "Platform GUID" which associates a specific platform with a set of PRM modules
--- a/PrmPkg/Samples/PrmSampleContextBufferModule/PrmSampleContextBufferModule.inf
+++ b/PrmPkg/Samples/PrmSampleContextBufferModule/PrmSampleContextBufferModule.inf
@@ -46,5 +46,4 @@
   MSFT:*_*_*_GENFW_FLAGS = --keepoptionalheader
 
   GCC:*_*_*_GENFW_FLAGS = --keepoptionalheader --prm
-  GCC:*_*_*_DLINK_FLAGS = -Wl,--no-gc-sections -Wl,--require-defined=PrmModuleExportDescriptor -Wl,--require-defined=CheckStaticDataBufferPrmHandler
-  GCC:*_*_X64_OBJCOPY_STRIPFLAG = --keep-symbol=PrmModuleExportDescriptor --keep-symbol=CheckStaticDataBufferPrmHandler
+  GCC:*_*_*_DLINK_FLAGS = -Wl,--no-gc-sections


apop5:
@ardbiesheuvel @leiflindholm 

FYI as well.

PrmPkg/Samples/PrmSampleAcpiParameterBufferModule/PrmSampleAcpiParameterBufferModule.inf
```diff
@@ -43,5 +43,5 @@
   MSFT:*_*_*_GENFW_FLAGS = --keepoptionalheader
 
   GCC:*_*_*_GENFW_FLAGS = --prm
-  GCC:*_*_*_DLINK_FLAGS = -Wl,--no-gc-sections -Wl,--require-defined=PrmModuleExportDescriptor -Wl,--require-defined=CheckParamBufferPrmHandler
+  GCC:*_*_*_DLINK_FLAGS = -Wl,--no-gc-sections -Wl,-u,PrmModuleExportDescriptor -Wl,-u,CheckParamBufferPrmHandler
```
makubacki: Update the following text in PrmPkg/Readme.md:

https://github.com/tianocore/edk2/blob/2efffed93811a7c5d0e6d7d19f8ff112128ffd6a/PrmPkg/Readme.md?plain=1#L79

makubacki:
Did this only occur with LLVM lld? `--require-defined` is in the [GNU linker manual](https://www.man7.org/linux/man-pages/man1/ld.1.html). If so, can you please clarify that in the commit and PR description?

apop5:
> Did this only occur with LLVM lld? `--require-defined` is in the [GNU linker manual](https://www.man7.org/linux/man-pages/man1/ld.1.html). If so, can you please clarify that in the commit and PR description?

This was encountered with LLD 19.1.0 for windows x64.
Looking at the makefile, the linker is using ld.lld.exe, because its producing ELF binaries.

This was not encountered when building with ClangPdb, because it uses a different linker in the Clang installation path.

mdkinney:
I have verified that this issues also exists with Linux/changdwarf builds and that this proposed change allows the build to complete.

There are differences between --undefined and --require-defined.  Are the PrmPkg builds impacted by these differences.

If the goal is to guarantee that specific symbols are never optimized away, then have standard C methods of guaranteeing that been used?  For example, reference to function pointers from entry point.

mdkinney:
Perhaps adding `volatile` to `PRM_EXPORT_API` will work instead of special linker options.

```
#if defined (_MSC_VER)
#define PRM_EXPORT_API  volatile __declspec(dllexport)
#elif defined (__GNUC__)
#define PRM_EXPORT_API  volatile __attribute__ ((visibility ("default")))
#else
#define PRM_EXPORT_API  volatile
#endif
```

makubacki:
> There are differences between --undefined and --require-defined. Are the PrmPkg builds impacted by these differences.

While I did not add the GCC support, my understanding is this will achieve the same functional result as before, that the symbols are included.

> Perhaps adding volatile to PRM_EXPORT_API will work instead of special linker options.

I'd also prefer to avoid linker flags. @apop5, can you please try this? If you haven't already. I'm happy to help test anything if needed.

apop5:
--require-defined Ensures that a symbol is defined somewhere in the linking process. If it's not, the linker will error out.

-Wl,-u Forces the linker to include a symbol in the final binary, even if it's not explicitly referenced.

I agree that appending additional linker flags is not a great way to get correct functionality. I will run some tests removing the additional linker flags, and dissemble the binaries to see what the results of the suggested changes.

PrmPkg/Include/Prm.h
```diff
@@ -15,9 +15,9 @@
 #include <PrmContextBuffer.h>
 
 #if defined (_MSC_VER)
-#define PRM_EXPORT_API  __declspec(dllexport)
+#define PRM_EXPORT_API  volatile __declspec(dllexport)
```
ardbiesheuvel, line 18: Is this `volatile` actually needed?

PrmPkg/Include/Prm.h
```diff
@@ -15,9 +15,9 @@
 #include <PrmContextBuffer.h>
 
 #if defined (_MSC_VER)
-#define PRM_EXPORT_API  __declspec(dllexport)
+#define PRM_EXPORT_API  volatile __declspec(dllexport)
 #elif defined (__GNUC__)
-#define PRM_EXPORT_API  __attribute__ ((visibility ("default")))
+#define PRM_EXPORT_API  __attribute__((used)) __attribute__ ((visibility ("default")))
```
ardbiesheuvel, line 20: I don't think this is safe. This is visible only to the compiler, and the linker has no awareness of the special semantics. The only way to annotate this safely is using the linker command.

mdkinney: The feature that is required here is to guarantee a symbol is not optimized away by the compiler/linker. Applying the keyword `volatile` on a global variable is one approach to implement this feature in C code with no special compiler/linker directives.  Another option that is used in ACPI Table source files is to guarantee there is a reference to the global variable in the entry point of the module.

https://github.com/tianocore/edk2/blob/d7832b480009f8304c1f60af7c9beb6b99d1c7e4/OvmfPkg/Bhyve/AcpiTables/Hpet.aslc#L63

ardbiesheuvel: > The feature that is required here is to guarantee a symbol is not optimized away by the compiler/linker. Applying the keyword `volatile` on a global variable is one approach to implement this feature in C code with no special compiler/linker directives.

I understand the goal. I just don't think that `volatile` is appropriate (or effective) here.

Note that `volatile` is a type qualifier, and so in this case, it changes the ~~return~~ type of `PrmModuleExportDescriptor` from `PRM_MODULE_EXPORT_DESCRIPTOR_STRUCT_` to `volatile PRM_MODULE_EXPORT_DESCRIPTOR_STRUCT_`. It does not have any bearing on whether or not the symbol is preserved by an optimizing compiler or linker.

Similarly, the `__attribute__((used))` in the ELF case has no bearing on the linker behavior, given that its existence is not visibile in the ELF object file.

I understand the problem that we are trying to solve here, but as far as I can tell,  instructing the linker to retain the symbol is the only robust approach here. Also, note that the `ReferenceAcpiTable()` approach needs a command line reference as well.

mdkinney: Can you find a case where `volatile` used on a global is still optimized away by the compiler/linker?

`volatile` indicates to the compiler that something outside of the scope of the compiler/linker visibility may change the state, so the variable must be preserved.

ardbiesheuvel: > Can you find a case where `volatile` used on a global is still optimized away by the compiler/linker?
> 

The point is not whether I can produce a case. The point is that the generated code is indistinguishable so it has no effect that the assembler or linker might observe:

```
$ gcc -S -o - -xc - <<<"int i;"
	.file	"<stdin>"
	.text
	.globl	i
	.bss
	.align 4
	.type	i, @object
	.size	i, 4
i:
	.zero	4
	.ident	"GCC: (Debian 14.2.0-19+build4) 14.2.0"
	.section	.note.GNU-stack,"",@progbits
```
versus

```
$ gcc -S -o - -xc - <<<"volatile int i;"
	.file	"<stdin>"
	.text
	.globl	i
	.bss
	.align 4
	.type	i, @object
	.size	i, 4
i:
	.zero	4
	.ident	"GCC: (Debian 14.2.0-19+build4) 14.2.0"
	.section	.note.GNU-stack,"",@progbits
```

So while the `volatile` qualifier may cause the compiler itself to treat a variable differently, once it hits the object file, the linker cannot distinguish one from the other. IOW, `volatile` does not exist in the context where the linker operates.

mdkinney: GCC has a nonstandard __attribute__((used)) that guarantees that a variable or function will not be optimized away.  If I apply that to your example above, the exact same output is generated as well.  I think this tiny example with no request to link into a final executable is not sufficient to verify the behavior. Everything I find says use of `volatile` is the recommended way to achieve this goal with standard C syntax without using any non-standard compiler/linker extensions.

PrmPkg/Samples/PrmSampleAcpiParameterBufferModule/PrmSampleAcpiParameterBufferModule.inf
```diff
@@ -43,5 +43,4 @@
   MSFT:*_*_*_GENFW_FLAGS = --keepoptionalheader
 
   GCC:*_*_*_GENFW_FLAGS = --prm
-  GCC:*_*_*_DLINK_FLAGS = -Wl,--no-gc-sections -Wl,--require-defined=PrmModuleExportDescriptor -Wl,--require-defined=CheckParamBufferPrmHandler
-  GCC:*_*_X64_OBJCOPY_STRIPFLAG = --keep-symbol=PrmModuleExportDescriptor --keep-symbol=CheckParamBufferPrmHandler
+  GCC:*_*_*_DLINK_FLAGS = -Wl,--no-gc-sections
```
mdkinney, line 46: Using --no-gc-sections prevents discarding any unused sections, but this may result in must larger binaries.  Using this feature for the one global that needs to be preserved is not the best approach.  Suggest using `volatile` keyword for GCC family builds as well.

PrmPkg/Include/Prm.h
ardbiesheuvel: Please try the following:

```
$ gcc -c -o /tmp/a.o -xc - <<<"void _ModuleEntrypoint(void) {}"
$ gcc -c -o /tmp/b.o -xc - <<<"volatile int PrmModuleDescriptor = 99;"
$ gcc -o /tmp/mod.elf /tmp/a.o /tmp/b.o -Wl,-e,_ModuleEntrypoint -nostartfiles -Wl,--gc-sections
$ nm /tmp/mod.elf 
0000000000004000 D __bss_start
0000000000003f30 d _DYNAMIC
0000000000004000 D _edata
0000000000004000 D _end
0000000000002000 r __GNU_EH_FRAME_HDR
0000000000001000 T _ModuleEntrypoint
```

Here, the `volatile` on `PrmModuleDescriptor` makes no difference whatsoever. The only thing that does help is removing the `--gc-sections` linker option, but this has the side effect that all garbage collection is disabled, potentially resulting in considerable bloat.

The only way to inform the GNU ld compatible linkers that a symbol must be retained is `--require-defined`.

Also, do you have any references to the documented linker behavior wrt `volatile`? As far as I can tell, it only affects the compiler.

PrmPkg/Samples/PrmSampleAcpiParameterBufferModule/PrmSampleAcpiParameterBufferModule.inf
apop5: Yes, but removing no-gc-sections allows the linker to strip the symbols.

PrmPkg/Include/Prm.h
apop5: In testing volatile had no effect on ensuring the symbol was not stripped. Removing it.

PrmPkg/Include/Prm.h
apop5: In my testing (ubuntu, clangdwarf and gcc) attempting to use attribute(used), volatile, etc, did not prevent removal of the symbols during the linking phase.

apop5: Additionally, --require-defined is not compatiable with llvm clangdwarf compiler's ldd.

There is an alternative, -Wl,-u,<symbolname> but it is a little bit different than --require-defined, in that it will create the symbols if it does not already exist, and it will prevent the symbol from being optimized out.

PrmPkg/Samples/PrmSampleAcpiParameterBufferModule/PrmSampleAcpiParameterBufferModule.inf
mdkinney: The other option that is used for ACPI Table file builds is to guarantee that the entry point of the module makes at least one reference to the global variable that must not be optimized away.  Example from ASLC file:

https://github.com/tianocore/edk2/blob/d7832b480009f8304c1f60af7c9beb6b99d1c7e4/OvmfPkg/Bhyve/AcpiTables/Hpet.aslc#L63

Does a PRM module have an entry point where this reference can be added?  Or a library that is always required for a PRM module can make a reference to to the global in the library constructor?

mdkinney: After further discussions, and feedback from @ajfish on behavior of `used`, I have another approach to solve this issue.  The `used` attribute is a good hint to the GCC and CLANGDWARF compilers that variable need to be preserved, but can still be optimized away if there are no references.  Forcing a reference in an entry point is possible, but would require all PRM modules to add extra code to for the reference.  Another option is the add a new entry point lib for PRM modules that makes this reference in a CONSTRUCTOR.

The solution I am proposing now is similar in concept to the VS20xx/CLANGPDB use of the _MSC_EXTENSIONS to declare structures and APIs as `__declspec(dllexport)`. For GCC/CLANGDWARF, this same concept can be simulated by introducing a new section type. Only the PRM Export Descriptor Structure requires a reference because all the PRM handlers are referenced from that structure.  I am suggesting introducing a new data section called `.prmexportdescriptor` that is only applied to the PRM Export Descriptor and update the linker script to always KEEP() `.prmexportdescriptor` and merge into `.data` section. The combination of the `used` attribute and the `.prmexportdescriptor` section attribute is enough for the GCC and CLANGSDWARF tool chains to make sure the PRM Export Descriptor and PRM Handlers are preserved in the final image. Since code and data sections can not be merged, this change required extending the macros in `Prm.h` to support export attributes for data and functions independently.

```
#if defined (_MSC_EXTENSIONS)
#define PRM_EXPORT_API         __declspec(dllexport)
#define PRM_EXPORT_DESCRIPTOR  __declspec(dllexport)
#elif defined (__GNUC__) || defined (__clang__)
#define PRM_EXPORT_API
#define PRM_EXPORT_DESCRIPTOR  __attribute__((used)) __attribute__((section(".prmexportdescriptor")))
#else
#define PRM_EXPORT_API
#define PRM_EXPORT_DESCRIPTOR
#endif
```

Suggested change to `GccBase.lds`.
```
  .data ALIGN(ALIGNOF(.text)) : ALIGN(CONSTANT(COMMONPAGESIZE)) {
    *(.data .data.* .gnu.linkonce.d.*)
    *(.bss .bss.*)
    KEEP(*(.prmexportdescriptor))
  } :data
```


The next challenge was getting the DLL export table to be produced for GCC/CLANGDWARF tool chains.  GENFW has been enhanced to support a `--prm` option to build the DLL export table and add it to the generated `.efi` PE/COFF image.  This feature depends on the PRM Export Descriptor Structure being present in a symbols section of the ELF/DWARF image generated by the GCC/CLANGDWARF builds.

However, the GCC builds strip symbols from the ELF/DWARF file, but preserves them in the .DEBUG file and adds a link from the ELF/DWARF file to the .DEBUG file using an OBJCOPY operation.  CLANGDWARF does not implement that feature.  With the change above, CLANGDWARF produces DLL exports correctly, but GCC crashes in GENFW with an exception because it could not find the PRM Export Descriptor Structure symbol because it had been stripped.  If the following line is added to PRM module INF files, the symbols are not stripped from the ELF/DWARF file and GENFW can find the symbols required to generate the DLL exports.

```
GCC:*_*_*_OBJCOPY_STRIPFLAG ==
```

The final challenge was getting the ImageVersion Major.Minor value in the PE/COFF Optional Header of GCC/CLANGDWARF builds of PRM modules.  The VS20xx/CLANGPDB builds are using a linker option /VERSION:Major.Minor to set these fields. When using GCC/CLANGDWARF, the intermediate output is ELF/DWARF and there is no app/image version field in the ELF header.  GENFW performs the ELF -> PE/COFF conversion, so this is the only place in the GCC/CLANGDWARF build that the PE/COFF optional header settings can be made.  I suggest a new option in GENFW called `--image-version` to provide the version values and set them in the generated PE/COFF optional header.  Also, in order to make sure the PRM module version values are maintained in a single location, the PRM Module INF `VERSION_STRING` value can be used to specify the version value so new releases of PRM Modules only need to update the INF file [Defines] section.

With the following PRM INF statements, the version value can be set correctly in all supported tool chains and the version value maintained in a single location.

```
[Defines]

[BuildOptions]
  MSFT:*_*_*_GENFW_FLAGS = --keepoptionalheader --image-version $(VERSION_STRING)
  CLANGPDB:*_*_*_GENFW_FLAGS = --keepoptionalheader --image-version $(VERSION_STRING)
  GCC:*_*_*_GENFW_FLAGS = --prm --image-version $(VERSION_STRING)
```

The use of the VS20xx/CLANGPDB /VERSION:MM.mm option does not generate an error on overflow of the major and minor values.  These fields are `UINT16`, so setting major or minor to a value >65535 silently uses the lower 16-bits with no warning or error.  The GENFW `--image-version` option checks for overflow and breaks the build if an invalid major or minor value is used.

The final gap in adding CLANG and GCC support is the setting PRM module section alignment for OS compatibility.  PR #11574 is already in review to address this.

I will post a draft PR with all the ideas proposed here for review. With these suggested changes, I have verified `VS2019`, `VS2022`, `CLANGPDB`, `GCC`, `GCC5`, `GCCNOLTO`, and `CLANGDWARF` builds produce PRM module PE/COFF `.efi` files with correct alignment, image version values, and DLL exports.

mdkinney: Opened draft PR #11576 with these suggested changes.  Only tested with X64.

apop5: I'll close in in favor of the other PR

apop5:
Closing in favor of #11576

PR #11463, opened 2025-09-03
Edk2 usb bus dxe enhancement

MarlboroChuang:
1. 
MdeModulePkg/Bus/Usb/UsbBusDxe: Fix UsbPortReset might run into
recursive loop once the device FW can not recover by PortReset.

UsbSelectConfig will introduce the UsbConnectDriver call.
If this UsbPortReset is happened in the Usb device driver Start()
routine and the device FW can not be recovered by PortReset, the
UsbSelectConfig will introduce the recursive loop.

[Suggested solution]
Since UsbPortReset should not change the Bus Topology, the Reset flow
should only SetAddress and reconfigure the device.

2. 
MdeModulePkg/Bus/Usb/UsbBusDxe: USB Device Enumeration Compatibility for
Non-Compliant Devices

The patch enhances the USB enumeration process in EDK2 to improve
compatibility with non-standards-compliant devices that may fail during
standard enumeration sequences.
The suggested solution is based on USB specifications and references
implementations from both Linux and Windows environments.

[Suggested solution]
- Integrated a retry mechanism to sequentially execute enumeration scripts,
  inspired by the enumeration flows of Windows, Linux, and EDK2.
  This improves robustness when handling corner-case devices.
- Do sanity check while the device report the device descriptor.
- AMD XHCI might need to wait for more time while sending the CLEAR_FEATURE
  reuqest.

3. 
MdeModulePkg/Bus/Usb/UsbBusDxe: A quirk in the Interface descriptor for
InterfaceNumber

Some specific device requires a quirk in the Interface descriptor for
InterfaceNumber to work properly

[Suggested Solution]
Implement the mechanism to ensure the first InterfaceNumber not equal
to zero.

4. 
MdeModulePkg/Bus/Usb/UsbBusDxe: Manufacturer String Descriptor Caching

Certain devices require immediate follow-up commands after reading the LANGID
string to fetch Manufacturer, Product, or SerialNumber strings.

[Suggested Solution]
These strings are now cached after initial retrieval to allow
UsbIoGetStringDescriptor() to return them directly, improving efficiency
and stability.

5. 
MdeModulePkg/Bus/Usb/UsbBusDxe: TPL Handling in
UsbBusRecursivelyConnectWantedUsbIo

[Suggested Solution]
Refactored to adopt the TPL handling mechanism from UsbConnectDriver(),
enabling ConnectController() to operate correctly at TPL_CALLBACK level
for the driver connection.

6. 
MdeModulePkg/Bus/Usb/UsbBusDxe: BOS Descriptor Check for SuperSpeed Devices

Some SuperSpeed-capable devices may fall back to High-Speed mode and
cause subsequent commands to fail.

[Suggested Solution]
Check the BOS descriptor to verify SuperSpeed support and trigger a
port reset if needed to re-enumerate the device properly.

### DIFF
--- a/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbBus.c
+++ b/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbBus.c
@@ -816,13 +816,14 @@ UsbIoPortReset (
   IN EFI_USB_IO_PROTOCOL  *This
   )
 {
-  USB_INTERFACE  *UsbIf;
-  USB_INTERFACE  *HubIf;
-  USB_DEVICE     *Dev;
-  EFI_TPL        OldTpl;
-  EFI_STATUS     Status;
-  UINT8          DevAddress;
-  UINT8          Config;
+  USB_INTERFACE    *UsbIf;
+  USB_INTERFACE    *HubIf;
+  USB_DEVICE       *Dev;
+  EFI_TPL          OldTpl;
+  EFI_STATUS       Status;
+  UINT8            DevAddress;
+  UINT8            Config;
+  USB_DEVICE_DESC  *DevDesc;
 
   OldTpl = gBS->RaiseTPL (USB_BUS_TPL);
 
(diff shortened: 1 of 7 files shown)


MarlboroChuang:
@jkolakow @lgao4 , May I have your support to check the code review?
Thanks a lot.

PR #11468, opened 2025-09-05
MdePkg: Enable CompilerIntrinsicsLib for LoongArch

kilaterlee:
# Description

When arrays are lager enough, LoongArch compilers(such as GCC or Clang) will replace some libc functions to improve performace, which may cause building failures. Enabling this library can link edk2's own libc functions and fix this problem.

## How This Was Tested
1. Create a 512-member array and initialize it to 0.
2. Without this patch, the build will fail. With this patch, the build will succeed.

## Integration Instructions

N/A


MdePkg/MdeLibs.dsc.inc
```diff
@@ -68,3 +68,6 @@
   # definitions for the intrinsic functions.
   #
   NULL|MdePkg/Library/CompilerIntrinsicsLib/CompilerIntrinsicsLib.inf
+
+[LibraryClasses.LOONGARCH64]
+  NULL|MdePkg/Library/CompilerIntrinsicsLib/CompilerIntrinsicsLib.inf
```
lgao4, line 81: Can you add the comments to explain why this lib is required for LOONGARCH64?

kilaterlee: Sure.

lgao4 (approved):

PR #11474, opened 2025-09-08
MdePkg/UefiScsiLib: Resend SCSI I/O without FUA being set

annie-li:
Commit 8c654bb3ec sets FUA bit forcefully in SCSI sync I/O to avoid possible data loss. However, it causes a boot failure when Windows Server 2025 guest boots from vhost-scsi block device with "write through" configured.

Windows guest booting Error code is 0xc0000185, which indicates boot file corruption or BCD configuration error. During Windows Server 2025 booting process, the boot manager opens several Windows boot files and operates on them - such as flushing I/Os from cache to disk. However, with the FUA bit set in CDB, these SCSI commands fail with the following error.

Mode Sense Key: EFI_SCSI_SK_ILLEGAL_REQUEST
Additional Sense Code: EFI_SCSI_ASC_INVALID_FIELD

It is better to fix the possible data loss mentioned in commit 8c654bb3ec through flushing cache instead of setting FUA forcefully. However, nothing has been implemented in ScsiDiskFlushBlocks yet except for simply returning EFI_SUCCESS there.

Simply reverting commit 8c654bb3ec may cause a regression of possible data loss. This patch implements a quick fix for this booting failure by resending SCSI sync I/O without FUA bit being set.

# Description

FUA bit is set forcefully for sync SCSI command. With this set, Windows Server 2025 fails to boot from vhost-scsi with "write through" block backend. When the failure happens, the SCSI command fails with EFI_SCSI_SK_ILLEGAL_REQUEST and EFI_SCSI_ASC_INVALID_FIELD. This patch resends the SCSI IO without FUA set after these errors happen. These errors may also be triggered due to other reasons, but these errors may not happen frequently and resending SCSI IO do not cause serious effects. The approach of resending SCSI IO is better than reverting the original patch of setting FUA bit for avoiding possible data loss.

- [ ] Breaking change?
  - No
- [ ] Impacts security?
  - No
- [ ] Includes tests?
  - No

## How This Was Tested

Boot Windows 2025 from vhost-scsi with "write through" backend.

## Integration Instructions

N/A

### DIFF
--- a/MdePkg/Include/Library/UefiScsiLib.h
+++ b/MdePkg/Include/Library/UefiScsiLib.h
@@ -654,6 +654,7 @@ ScsiRead10Command (
   @param[in, out] DataLength           The length of data buffer.
   @param[in]      StartLba             The start address of LBA.
   @param[in]      SectorSize           The number of contiguous logical blocks of data that shall be transferred.
+  @param[in]      SetFUA               If TRUE, FUA bit will be set.
 
   @retval  EFI_SUCCESS                 Command is executed successfully.
   @retval  EFI_BAD_BUFFER_SIZE         The SCSI Request Packet was executed, but the entire DataBuffer could
@@ -679,7 +680,8 @@ ScsiWrite10Command (
   IN OUT VOID                  *DataBuffer   OPTIONAL,
   IN OUT UINT32                *DataLength,
   IN     UINT32                StartLba,
-  IN     UINT32                SectorSize
+  IN     UINT32                SectorSize,
+  IN     BOOLEAN               SetFUA
   );
 
 /**
@@ -774,6 +776,7 @@ ScsiRead16Command (
   @param[in, out] DataLength           The length of data buffer.
   @param[in]      StartLba             The start address of LBA.
   @param[in]      SectorSize           The number of contiguous logical blocks of data that shall be transferred.
+  @param[in]      SetFUA               If TRUE, FUA bit will be set.
 
   @retval  EFI_SUCCESS                 Command is executed successfully.
   @retval  EFI_BAD_BUFFER_SIZE         The SCSI Request Packet was executed, but the entire DataBuffer could
@@ -799,7 +802,8 @@ ScsiWrite16Command (
   IN OUT VOID                  *DataBuffer   OPTIONAL,
   IN OUT UINT32                *DataLength,
   IN     UINT64                StartLba,
-  IN     UINT32                SectorSize
+  IN     UINT32                SectorSize,
+  IN     BOOLEAN               SetFUA
   );
 
 /**
--- a/MdeModulePkg/Bus/Scsi/ScsiDiskDxe/ScsiDisk.c
+++ b/MdeModulePkg/Bus/Scsi/ScsiDiskDxe/ScsiDisk.c
@@ -2473,9 +2473,11 @@ ScsiDiskDetectMedia (
   // Get Write Cache Mode
   //
   for (Retry = 0; Retry < MaxRetry; Retry++) {
-    Status = ScsiDiskWriteCacheMode (ScsiDiskDevice);
-    if (!EFI_ERROR (Status)) {
-      break;
+    if (ScsiDiskDevice->DeviceType == EFI_SCSI_TYPE_DISK) {
+      Status = ScsiDiskWriteCacheMode (ScsiDiskDevice);
+      if (!EFI_ERROR (Status)) {
+        break;
+      }
     }
   }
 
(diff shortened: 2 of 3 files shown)


leiflindholm:
@makubacki I can't figure out what's going wrong here, can you?

makubacki:
> @makubacki I can't figure out what's going wrong here, can you?

Are you referring to [this check](https://github.com/tianocore/edk2/commit/6d571c0070161b1b96049410f8584c0955d73536) being activated? The introduction of a new return code should at least be separated from the coupled functional changes.

leiflindholm:
> > @makubacki I can't figure out what's going wrong here, can you?
> 
> Are you referring to [this check](https://github.com/tianocore/edk2/commit/6d571c0070161b1b96049410f8584c0955d73536) being activated?

I was seeing lots of errors I think was unrelated to that. Or I failed to parse the output. Either way it now looks resolved.

annie-li:
@leiflindholm @makubacki thanks for checking the failure. 
`The introduction of a new return code should at least be separated from the coupled functional changes.`
So to fix this patch check error, I suppose I need to separate the patch into two, one is for functional changes, another is for introduction of a new return code?

MdePkg/Include/Uefi/UefiBaseType.h
```diff
@@ -143,6 +143,7 @@ typedef union {
 #define EFI_COMPROMISED_DATA      RETURN_COMPROMISED_DATA
 #define EFI_IP_ADDRESS_CONFLICT   RETURN_IP_ADDRESS_CONFLICT
 #define EFI_HTTP_ERROR            RETURN_HTTP_ERROR
+#define EFI_DEVICE_NO_FUA         RETURN_DEVICE_NO_FUA
```
lgao4: Is this error defined in the latest UEFI spec?

MdePkg/Include/Library/UefiScsiLib.h
```diff
@@ -679,7 +680,8 @@ ScsiWrite10Command (
   IN OUT VOID                  *DataBuffer   OPTIONAL,
   IN OUT UINT32                *DataLength,
   IN     UINT32                StartLba,
-  IN     UINT32                SectorSize
+  IN     UINT32                SectorSize,
```
lgao4, line 683: I suggest to introduce new API for this support, don't update the existing one.

lgao4:
> @leiflindholm @makubacki thanks for checking the failure. `The introduction of a new return code should at least be separated from the coupled functional changes.` So to fix this patch check error, I suppose I need to separate the patch into two, one is for functional changes, another is for introduction of a new return code?

You need to separate the change for each package. One commit is the change in MdePkg, another commit is the change in MdeModulePkg. I don't think the change in BaseTools is required.

makubacki:
> You need to separate the change for each package. One commit is the change in MdePkg, another commit is the change in MdeModulePkg. I don't think the change in BaseTools is required.

Also, even in `MdePkg`, I suggest you separate the introduction of a new status code and any other changes into separate commits. The status code is global and independent of the other changes. If, for some reason, the other changes need to be reverted in the future, it should be possible to do that while also not having to revert the status code definition that might, at that time, be used by other code than the code in these changes.

MdePkg/Include/Uefi/UefiBaseType.h
annie-li: Nope. This may be a dumb question -  so a macro has to be defined in the spec first before it is introduced?

annie-li:
@lgao4 @makubacki Thanks for your feedback. Will improve the patch as suggested.

MdePkg/Include/Uefi/UefiBaseType.h
leiflindholm: @annie-li error codes described in UefiBaseType.h really ought to exist in https://uefi.org/specs/UEFI/2.11/Apx_D_Status_Codes.html, yes.

Also, in general, everything with an EFI_/Efi prefix is assumed to be in UEFI specification.

annie-li:
I've split the patch into two patches, however, splitting the patch doesn't help with the patchcheck error(`Bash exited with code '255'.`). It is not possible to only change the code for one package at one time since it won't pass the compiling. Anyway, there're no other checking errors, I suppose this is fine?
The new patch set avoids the new macro definition, and adds check of Write Cache Mode before resending SCSI I/O without FUA.

leiflindholm:
The actual patchcheck error, on display in the bottom of a locked filing cabinet stuck in a disused lavatory with a sign on the door saying 'Beware of the Leopard.", is:

```
 * MdePkg
 * MdeModulePkg
```

This isn't really avoidable for that change, so please add
to the commit message of the second commit.

PR #11475, opened 2025-09-08
tools_def.template: Remove /FILEALIGN from *_CLANGPDB_X64_DLINK_FLAGS

makubacki:
# Description

`FileAlignment` in the PE/COFF Specification states:

```
The alignment factor (in bytes) that is used to align the raw data of
sections in the image file. The value should be a power of 2 between
512 and 64 K, inclusive. The default is 512. If the SectionAlignment
is less than the architecture's page size, then FileAlignment must
match SectionAlignment.
```

Although in UEFI we often use sizes less than 512, that is outside the allowed spec-defined range. Today, `/FILEALIGN` is not set in X64 VS DLINK flags.

This change modifies X64 CLANGPDB linker flags in tools_def.template to:

1. Be PE/COFF spec-compliant by not explicitly specifying a size less than the spec-supported minimum.
2. Make the `/FILELALIGN` value consistent for X64 modules between the VS and CLANGPDB flags.

> Note: A smaller `/FILEALIGN` value is often used to save flash space
      particularly for uncompressed images that reside directly in
      flash. For that reason,, the CLANGPDB_IA32 values were left
      to keep `/FILEALIGN:32` to not impact platforms where that is
      more likely the case.

Platforms with 64-bit PEIMs, can explicitly set `/FILEALIGN` for their active toolchain as needed. That is a platform decision to opt out of PE/COFF Specification compatibility to achieve platform size requirements.

This is particularly important because some tools that operate on PE/COFF images, strictly follow the specification behavior. With this change, by default CLANGPDB X64 images will be compliant with those expectations.

- [x] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- Verified file alignment in CLANGPDB X64 binaries

## Integration Instructions

- Review the information in this PR description. If a smaller file alignment value is needed, set that in a platform location such as `[BuildOptions]` in a DSC file.
- This is marked as a breaking change because a larger file alignment value for X64 modules may cause more space flash to be consumed if using CLANGPDB.


nate-desimone:
Hi @makubacki - It looks like this will result in CLANGPDB images that by-default have FileAlignment != SectionAlignment. Did you test to make sure these PE/COFF images load successfully? I remember running into problems trying to use images with this property on EmulatorPkg on Windows.

PR #11478, opened 2025-09-09
UnitTestFrameworkPkg: Make includes public

trhpx:
# Description

Make private Include files public.

UnitTestBootLib.h needs to be public to allow unit tests in other Pkg to use a UnitTestBootLib implementation, and also to  allow for new implementations of this LibraryClass in other Pkg

UnitTestResultReportLib.h needs to be public to allow unit tests in other 'packages' to use a UnitTestResultReportLib implementation, or to allow for new implementations of this LibraryClass in other 'packages'.

UnitTestFrameworkTypes.h needs to be public because it is included by UnitTestResultReportLib.h

Update UnitTestFrameworkPkg.dec to make public.

Signed-off-by: Tim Hoppen <tim.hoppen@phoenixtech.com>

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Build test of UnitTestFrameworkPkg\UnitTestFrameworkPkg.dsc
No other package is able to use the affected files (this is what the PR addresses), so no other package testing should be needed.

## Integration Instructions

N/A


trhpx:
I'd like to start leveraging UnitTestFrameworkPkg for custom tests in my own package, but these library classes being private is making that (needlessly, IMHO) complicated.

@makubacki @mdkinney @spbrogan If I'm way off base in my view of this, please help to explain. Otherwise, can we look at this for approval?

mdkinney:
Making UnitTestBootLib and UnitTestResultReportLib public includes makes sense to me.  Especially with the comments in the include file headers.

Making all of UnitTestFrameworkType.h public does not make sense.  Which elements do you think are required?

trhpx:
I don't have any need for UnitTestFrameworkTypes.h, but UnitTestResultReportLib.h is including it.

I found that including UnitTestResultReportLib.h from outside of this package required UnitTestFrameworkTypes.h to also not be private.

mdkinney:
I think that include is a bug.  The API in that lib only uses the framework handle that is defined in the public UnitTestLib.h in the MdePkg.

Can you try replacing #include <UnitTestFrameworkTypes.h> with #include <Library/UnitTestLib.h>

trhpx:
I agree that it looks like a bug. I'll give that a try when I get a chance.

trhpx:
Trying that caused some build failures... there are 3 C files that currently depend on UnitTestResultReportLib.h having a #include for UnitTestFrameworkTypes.h:

- UnitTestFrameworkPkg\Library\UnitTestLib\RunTests.c
- UnitTestFrameworkPkg\Library\UnitTestLib\UnitTestLib.c
- UnitTestFrameworkPkg\Library\UnitTestResultReportLib\UnitTestResultReportLib.c

If `#include <UnitTestFrameworkTypes.h>` is added to these 3 C files, everything in _edk2_ builds fine.

But back to my original use case, one of the goals I have (not specified until now) is to use UnitTestResultReportJUnitFormatLib from Microsoft's mu_plus\XmlSupportPkg, which does use most of UnitTestFrameworkTypes.h contents:

- UNIT_TEST_MAX_STRING_LENGTH
- UNIT_TEST_FINGERPRINT_SIZE
- FAILURE_TYPE
- FAILURETYPE_NOFAILURE
- FAILURETYPE_OTHER
- FAILURETYPE_ASSERTTRUE
- FAILURETYPE_ASSERTFALSE
- FAILURETYPE_ASSERTEQUAL
- FAILURETYPE_ASSERTNOTEQUAL
- FAILURETYPE_ASSERTNOTEFIERROR
- FAILURETYPE_ASSERTSTATUSEQUAL
- FAILURETYPE_ASSERTNOTNULL
- UNIT_TEST
- UNIT_TEST_LIST_ENTRY
- UNIT_TEST_SUITE
- UNIT_TEST_SUITE_LIST_ENTRY
- UNIT_TEST_FRAMEWORK

mdkinney:
Could support for JUINT format be added to UnitTestFrameworkPkg, so it can use the private includes?

mdkinney:
If the goal is to support custom report libraries outside the UnitTestFrameworkPkg, then I think new APIs are required in the UnitTestLib.  The UnitTestLib today supports creating and registering and running unit tests and provides macros/functions to implement unit tests.  It does not provide services to collect the results from unit tests after they have been run.  The one implementation of UnitTestResultReportLib in the UnitTestFrameworkPkg has to know the internal structures to collect results.

Can you propose to API additions to the UnitTestLib that would allow all the registered unit tests to be traversed and collect the test results?  I think that would be a great improvement that would allow many different unit test results report generators.

Issue #11483, opened 2025-09-10
[Feature]: iSCSI Support can't eable at build time


r1chard-lyu:
### Feature Overview

Since f0424ec80e, the iSCSI DXE in OVMF has been disabled by PcdEntryPointOverrideDefaultValue
The user needs to enable the iSCSI DXE explicitly using the -qemu syntax '-fw_cfg "name=opt/org.tianocore/ISCSISupport,string=yes"'

However, this approach may not be very user-friendly. Contrary to what the commit message says, I believe that the iSCSI code doesn’t cause a massive slowdown unless the user has explicitly configured iSCSI boot attempts. (This is in stark contrast to HTTP boot attempts, which are always tried twice for every network interface, with ~30s timeout per attempt.)

### Solution Overview

Would it be possible to set the default of NETWORK_ISCSI_ENABLE to FALSE and remove the PcdEntryPointOverrideDefaultValue from NetworkComponents.dsc.inc, or alternatively add a build flag or PCD so that iSCSI support can be specified at build time?

### What packages are impacted?

OvmfPkg

### Urgency

Medium

### Are you going to implement the feature request?

I will implement the feature

### Do you need maintainer feedback?

Maintainer feedback requested

r1chard-lyu:
The fixed patch has been merged.

PR #11484, opened 2025-09-10
OvmfPkg: Disable iSCSI support at build time by default

r1chard-lyu:
REF: https://github.com/tianocore/edk2/issues/11483

Since f0424ec80e, the iSCSI DXE in OVMF has been disabled by PcdEntryPointOverrideDefaultValue. Currently, users must enable the iSCSI DXE explicitly using the QEMU syntax:
'-fw_cfg name=opt/org.tianocore/ISCSISupport,string=yes'.

This approach is not user-friendly because it does not allow deciding at build time whether to enable iSCSI support. Instead, it requires adding an extra command line option at runtime, which is also inconvenient when using Libvirt.

This commit changes iSCSI support to be disabled at build time by default, allowing users to decide whether to enable iSCSI support at build time.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
1. Default
build -a X64 -b DEBUG -t GCC5 -p OvmfPkg/OvmfPkgX64.dsc
2. Enable iSCSI Support
build -D NETWORK_ISCSI_ENABLE=TRUE -a X64 -b DEBUG -t GCC5 -p OvmfPkg/OvmfPkgX64.dsc

## Integration Instructions

N/A


kraxel:
Problem is that iscsi initialization needs significant amount of time, so turning this off by default speeds up boot.

Having a compile time option instead of a runtime option IMHO is worse for users because they must compile their own firmware images instead of simply using the builds shipped by the distro.

libvirt has support for setting these options btw:
```
  <sysinfo type='fwcfg'>
    <entry name='opt/org.tianocore/ISCSISupport'>yes</entry>
  </sysinfo>
```

OvmfPkg/OvmfPkgX64.dsc
```diff
@@ -50,7 +50,7 @@
   DEFINE NETWORK_IP6_ENABLE             = FALSE
   DEFINE NETWORK_HTTP_BOOT_ENABLE       = FALSE
   DEFINE NETWORK_ALLOW_HTTP_CONNECTIONS = TRUE
-  DEFINE NETWORK_ISCSI_ENABLE           = TRUE
+  DEFINE NETWORK_ISCSI_ENABLE           = FALSE
```
ardbiesheuvel, line 53: What is wrong with simply building your firmware with `-D NETWORK_ISCSI_ENABLE=FALSE` on the command line if you want to omit iSCSI from the build entirely?

r1chard-lyu:
Contrary to the above, I believe that the iSCSI code doesn’t cause a massive slowdown unless the user has explicitly configured iSCSI boot attempts (this is in stark contrast to HTTP boot attempts, which are always tried twice for every network interface, with ~30s timeout per attempt).

For distros, it is not possible to ship an OVMF version with iSCSI support enabled by default. In other words, users are forced to configure fwcfg in order to enable iSCSI support, which some may find inconvenient.

Alternatively, would it be possible to add a PCD or build flag to allow enabling iSCSI support at build time without requiring fwcfg configuration?

OvmfPkg/OvmfPkgX64.dsc
r1chard-lyu: I understand that this may seem counterintuitive, but since the current approach is to disable iSCSI support by default at runtime, this change moves that behavior to build time if there is a performance impact. In fact, I would prefer adding a PCD or a build flag to provide more flexibility in deciding whether to enable iSCSI at build time or runtime. However, considering that there are already many existing PCDs and build flags, including NETWORK_ISCSI_ENABLE, why not simply disable it by default?

r1chard-lyu: After this change, using ```-D NETWORK_ISCSI_ENABLE=TRUE``` allows building an OVMF version with iSCSI support enabled directly, without requiring the user to configure fwcfg.

kraxel:
> Contrary to the above, I believe that the iSCSI code doesn’t cause a massive slowdown unless the user has explicitly configured iSCSI boot attempts (this is in stark contrast to HTTP boot attempts, which are always tried twice for every network interface, with ~30s timeout per attempt).

iscsi writes large efi variables to flash and pflash emulation is slow.  It is not in the same order of magnitude of network boot timeouts, but long enough that you don't have to measure the firmware init time to notice the difference.

kraxel:
> Alternatively, would it be possible to add a PCD or build flag to allow enabling iSCSI support at build time without requiring fwcfg configuration?

build flag is fine with me.

r1chard-lyu:
> > Contrary to the above, I believe that the iSCSI code doesn’t cause a massive slowdown unless the user has explicitly configured iSCSI boot attempts (this is in stark contrast to HTTP boot attempts, which are always tried twice for every network interface, with ~30s timeout per attempt).
> 
> iscsi writes large efi variables to flash and pflash emulation is slow. It is not in the same order of magnitude of network boot timeouts, but long enough that you don't have to measure the firmware init time to notice the difference.

Thank you for the explanation; it was very helpful for me to understand the previous changes. So, is it currently possible for OVMF to enable iSCSI support by default at build time without requiring fwcfg configuration at runtime?

r1chard-lyu:
> > Alternatively, would it be possible to add a PCD or build flag to allow enabling iSCSI support at build time without requiring fwcfg configuration?
> 
> build flag is fine with me.

Okay, then I should be able to open another PR to provide a new build flag.

r1chard-lyu:
Closing this PR, thank you for your comments.

PR #11485, opened 2025-09-10
OvmfPkg: Add NETWORK_ISCSI_BUILD_ENABLE build flag

r1chard-lyu:
# Description
REF: https://github.com/tianocore/edk2/issues/11483

Introduce a new build-time flag NETWORK_ISCSI_BUILD_ENABLE to control whether iSCSI support is enabled at build time. This allows developers to decide at build stage if the IScsiDxe driver should be included, in addition to the existing runtime fwcfg option.

If NETWORK_ISCSI_BUILD_ENABLE is set to FALSE, IScsiDxe will still be built when NETWORK_ISCSI_ENABLE is TRUE, but the default PCD value (gUefiOvmfPkgTokenSpaceGuid.PcdEntryPointOverrideDefaultValue) will be set to "no". This ensures iSCSI remains disabled unless explicitly enabled by fwcfg.

This change provides more flexibility for both build-time and runtime configuration of iSCSI support.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
1. Default
build -a X64 -b DEBUG -t GCC5 -p OvmfPkg/OvmfPkgX64.dsc
2. Enable iSCSI Support at build time
build -D NETWORK_ISCSI_ENABLE -a X64 -b DEBUG -t GCC5 -p OvmfPkg/OvmfPkgX64.dsc

## Integration Instructions
N/A


r1chard-lyu:
Based on the comment in https://github.com/tianocore/edk2/pull/11484#issuecomment-3274220190, this patch was created.

kraxel:
Naming this 'BUILD' is confusing IMHO as this sets the default for the /runime/ switch.

r1chard-lyu:
> Naming this 'BUILD' is confusing IMHO as this sets the default for the /runime/ switch.

Thanks for the reminder. How about ```NETWORK_ISCSI_DEFAULT_ENABLE```,  ```NETWORK_ISCSI_AUTO_ENABLE```, ```NETWORK_ISCSI_BOOT_DEFAULT_ENABLE```, or do you have other suggestions?

kraxel:
Hmm.  None of them is very good, but I can't think of something better.  Naming is hard ...

I think `NETWORK_ISCSI_DEFAULT_ENABLE` is the best if the three suggestions.

r1chard-lyu:
> Hmm. None of them is very good, but I can't think of something better. Naming is hard ...
> 
> I think NETWORK_ISCSI_DEFAULT_ENABLE is the best if the three suggestions.

I’ve renamed this patch’s build flag to ```NETWORK_ISCSI_DEFAULT_ENABLE```. I also think this is the most suitable name for now. Perhaps in the future, if a better naming option comes up, we can adjust it again. Could you confirm if this is acceptable?

PR #11486, opened 2025-09-11
ShellPkg: Disable the shell when UEFI Secure Boot is on

mediouni-m:
When UEFI Secure Boot is on, we should not allow for the UEFI shell to be used. As such, disabling it in that scenario.

- [ ] Breaking change?
- [x] Impacts security?
  - The UEFI shell is known as insecure.
- [ ] Includes tests?

## How This Was Tested

Has been on AWS since a while, see https://github.com/aws/uefi/blob/5c3ac896feea3923a96944dc23e808385939c0a1/edk2-stable202211/0032-edk2-stable202211-uefi-shell-Disable-the-shell-when-UEFI-Secure-Boot-is-on.patch

## Integration Instructions

N/A

### DIFF
--- a/ShellPkg/Application/Shell/Shell.c
+++ b/ShellPkg/Application/Shell/Shell.c
@@ -358,6 +358,17 @@ UefiMain (
   EFI_HANDLE                      ConInHandle;
   EFI_SIMPLE_TEXT_INPUT_PROTOCOL  *OldConIn;
   SPLIT_LIST                      *Split;
+  UINT8                           *SecureBoot;
+
+  // If Secure Boot is enabled, do not launch the UEFI shell
+  SecureBoot = NULL;
+  GetEfiGlobalVariable2 (EFI_SECURE_BOOT_MODE_NAME, (VOID **)&SecureBoot, NULL);
+  if ((SecureBoot != NULL) && (*SecureBoot == SECURE_BOOT_MODE_ENABLE)) {
+    FreePool (SecureBoot);
+    return EFI_SECURITY_VIOLATION;
+  } else if (SecureBoot != NULL) {
+    FreePool (SecureBoot);
+  }
 
   if (PcdGet8 (PcdShellSupportLevel) > 3) {
     return (EFI_UNSUPPORTED);
--- a/ShellPkg/Application/Shell/Shell.h
+++ b/ShellPkg/Application/Shell/Shell.h
@@ -14,6 +14,8 @@
 
 #include <Guid/ShellVariableGuid.h>
 #include <Guid/ShellAliasGuid.h>
+#include <Guid/GlobalVariable.h>
+#include <Guid/ImageAuthentication.h>
 
 #include <Protocol/LoadedImage.h>
 #include <Protocol/SimpleTextOut.h>


ShellPkg/Application/Shell/Shell.c
```diff
@@ -358,6 +358,17 @@ UefiMain (
   EFI_HANDLE                      ConInHandle;
   EFI_SIMPLE_TEXT_INPUT_PROTOCOL  *OldConIn;
   SPLIT_LIST                      *Split;
+  UINT8                           *SecureBoot;
+
+  // If Secure Boot is enabled, do not launch the UEFI shell
```
samimujawar, line 363: Thank you for this patch.

Should there be a PCD that can be configured to enforce this. Otherwise, this would be a setback during development.

Also is this not a platform's choice to include/exclude the shell in production builds?

leiflindholm:
I agree 100% with the aim, but I'm not sure this is the best way to solve this upstream.

First, as @samimujawar points out, this is a real problem for development.
Second, I should be able to build and sign a standalone shell if that is something I want to do on my own machine.

What we really want to prevent is the lazy inclusion of the shell in the FV during development and then leaving it there when shipping. I ... don't know the best way of doing that.

NikolajSchlej:
I think we also need to prevent OEMs from signing a shell and using it as a part of their workflows, especially ones facing the user. 
UEFI shell should be considered a developer-only tool that should never be a part of secure boot chain, and should never be released as a part of such.

I do think the approach done in this patch is sound, and while an option is fine, the default for it should rather be "see SecureBoot enabled => loudly complain and hang", not just exit quietly.

If a developer actually knows what they are doing, it would be trivial for them to flip a define in a header file (I'm against a PCD because that will immediately be set to "make the issue go away" and copy-pasted into the future forever) saying YES_I_M_AWARE_THAT_MIXING_SECUREBOOT_WITH_A_SHELL_IS_REALLY_BAD_IDEA.

xpahos:
For dsc, there is already an option `DEFINE BUILD_SHELL             = FALSE`. Perhaps it would be sufficient to set its default value to `FALSE`?

mediouni-m:
UEFI Shim allows to load an untrusted by UEFI Secure Boot image if it's authorised by the user. 

Wonder if disabling startup.nsh and argument parsing when loading Shell.efi would be considered as the right thing to do then (when considering that threat model)

niruiyu (commented):
Why cannot Shell be launched when Secure Boot is on?

leiflindholm:
> Why cannot Shell be launched when Secure Boot is on?

It is designed to be a debug tool, not a part of a secured boot flow.

As a result, it is pretty much an unbounded attack surface. The `mm` command on its own rules out any serious CA from considering signing it. But many include it in the FV during development, bypassing Secure Boot checks, and some neglect to delete it for deployment.

kukrimate:
Why is this being considered now (besides the new media campaigns)? Canonical has reported the same problem and proposed a similar fix that was rejected years ago.

The conclusion there was to disable including the Shell in Secure Boot compatible builds of OVMF, but a similar patch to the Shell itself that would just make it exit with a security violation was rejected. Also note that my patch was somewhat better too because it took Setup mode into account as well.

JohnAZoidberg:
> It is designed to be a debug tool, not a part of a secured boot flow.
> As a result, it is pretty much an unbounded attack surface. The mm command on its own rules out any serious CA from considering signing it.

The shell can be built with those commands excluded.
Is it inconceivable to have a secure shell that is only used for simple scripts and does not allow dangerous operations?
Launching applications, evaluating their return codes or output, setting UEFI variables and launching other apps?

Is it more secure to have people learn the intricacies of doing those things in the UEFI C environment?

kukrimate:
> setting UEFI variables

you have already evaluated one of the footguns... being able to set boot services variables will bypass some security features like SkuSiPolicy and SBAT

leiflindholm:
> The shell can be built with those commands excluded.

Which ones? And who will continuously monitor it and make sure changes in UEFI itself doesn't turn previously safe operations unsafe?

But this isn't really about technical details. UEFI Shell isn't a _part_ of UEFI. Unlike open firmware, u-boot, etc, UEFI isn't designed with a command shell in mind. The UEFI Shell is a (very handy) development tool that escaped and started being used in CI, factory testing etc.

Wack0:
My thought here: disable running `startup.nsh` automatically when Secure Boot is enabled and Setup Mode is off.

This would prevent usage in a bootkit style scenario with automatically running commands, which seems to be the main issue here.

NikolajSchlej:
@Wack0, that will not be enough, as a [spec-compliant](https://uefi.org/sites/default/files/resources/UEFI_Shell_2_2.pdf) UEFI shell can take a script to run as a command-line option from either `ShellOpt` variable, or as part of `BootXXXX` variable it had been executed from, both of which as normal NV vars settable by OS-level attacker (see chapter 3.2 of the the spec linked above for details). 

We could disable scripting entirely, making the shell into interactive-only tool, but I'm still in favor of disabling the shell entirely, and explicitly telling the user that it's not compatible with SecureBoot in any way, shape or form.

Wack0:
> a [spec-compliant](https://uefi.org/sites/default/files/resources/UEFI_Shell_2_2.pdf) UEFI shell can take a script to run as a command-line option from either `ShellOpt` variable, or as part of `BootXXXX` variable it had been executed from, both of which as normal NV vars settable by OS-level attacker (see chapter 3.2 of the the spec linked above for details).

I was unaware of that part of the spec (although I have seen shells in the past implementing the latter, I was unaware it was part of the spec). In that case my thought can easily be revised to "disable any way a script or command can automatically run on shell start if Secure Boot is enabled". That's what is causing exploitability issues here; that "dangerous" commands can automatically execute without confirmation by a physically present user.

leiflindholm:
> I was unaware of that part of the spec (although I have seen shells in the past implementing the latter, I was unaware it was part of the spec). In that case my thought can easily be revised to "disable any way a script or command can automatically run on shell start if Secure Boot is enabled". That's what is causing exploitability issues here; that "dangerous" commands can automatically execute without confirmation by a physically present user.

The Shell is inherently not secure. I don't think there is value, and I absolutely see risk, in trying to retrofit a "securable" sub-configuration.

mikebeaton:
| | NOOPT/DEBUG | RELEASE |
|---|---|---|
| Secure Boot | Build and sign shell, and include in FV | Build and sign shell, but don't include in FV (can be manually loaded) |
| Not Secure Boot | Build and include shell | Build and include shell |

Is there any way we could build the shell and include in FV in DEBUG, build and sign but not include in FV in RELEASE? Then developers get the shell 'for free' in DEBUG, and can save it somewhere, navigate to it and launch it, in RELEASE.

Flickdm:
> NOOPT/DEBUG	RELEASE
> Secure Boot	Build and sign shell, and include in FV	Build and sign shell, but don't include in FV (can be manually loaded)
> Not Secure Boot	Build and include shell	Build and include shell
> Is there any way we could build the shell and include in FV in DEBUG, build and sign but not include in FV in RELEASE? Then developers get the shell 'for free' in DEBUG, and can save it somewhere, navigate to it and launch it, in RELEASE.

If the binary is in the FV it does not need to be signed (FV Apps are implicitly trusted) that being said at Microsoft, our DEBUG firmware (never released to the public and will not work on a fused platform once it leaves the factory) is allowed to include shell simply because the reduced security posture during firmware development. 

In RELEASE firmware, shell is not built nor is it included in the FV.

When I need shell in RELEASE firmware with secure boot enabled I do the following:

1. Disable secureboot
2. Enter Setup Mode
3. Add a certificate to the DB
4. Sign my shell with the following command:

```powershell
& 'signtool.exe' sign /fd sha256 /f DB.pfx /p <password> shell.efi
```

There should be no reason why in RELEASE firmware shell needs to be included. As @NikolajSchlej said Shell and Secure boot are not compatible.

I'm happy to document the exact steps at https://github.com/microsoft/secureboot_objects

mikebeaton:
> If the binary is in the FV it does not need to be signed (FV Apps are implicitly trusted) that being said at Microsoft, our DEBUG firmware (never released to the public and will not work on a fused platform once it leaves the factory) is allowed to include shell simply because the reduced security posture during firmware development.
> 
> In RELEASE firmware, shell is not built nor is it included in the FV.
> 
> When I need shell in RELEASE firmware with secure boot enabled I do the following:
> 
> 1. Disable secureboot
> 2. Enter Setup Mode
> 3. Add a certificate to the DB
> 4. Sign my shell with the following command:
> 
> ```powershell
> & 'signtool.exe' sign /fd sha256 /f DB.pfx /p <password> shell.efi
> ```
> 
> There should be no reason why in RELEASE firmware shell needs to be included. As @NikolajSchlej said Shell and Secure boot are not compatible.
> 
> I'm happy to document the exact steps at https://github.com/microsoft/secureboot_objects

Fair point that a binary in FV does not even need to be signed. All I was saying was, could the shell not be _built_, but not placed into any FV, in RELEASE build with secure boot enabled? Even better if there is no question of it being signed in this process - you are quite correct that any developer should know various ways to load the built binary (by signing it, or disabling secure boot) if they need it.

pagabuc:
It is unfortunate to see this closed by inactivity. I believe the thread provides exactly the kind of informed, multi-stakeholder analysis needed to make a security decision. If possible, this PR should be reopened and either merged (with or without modifications), or rejected with a documented rationale.

mikebeaton:
Does anyone else who's commented on this want to come out for or against my idea that [the shell might as well be _built_ in RELEASE, just not included in FV](https://github.com/tianocore/edk2/pull/11486#issuecomment-3616783708)?

Whichever the general consensus, fine with me - I don't want my suggestion to sit there unresolved and potentially blocking making any change at all.

leiflindholm:
> Does anyone else who's commented on this want to come out for or against my idea that [the shell might as well be _built_ in RELEASE, just not included in FV](https://github.com/tianocore/edk2/pull/11486#issuecomment-3616783708)?
> 
> Whichever the general consensus, fine with me - I don't want my suggestion to sit there unresolved and potentially blocking making any change at all.

I mean, that's a good idea. But the FV is platform specific, so I'm unsure what we'd do as a project? Just make it conditional in the virtual machine targets' .fdfs?

If we were to actually attempt to prevent people from shooting themselves with this particular footgun, we'd need to integrate innate knowledge of the Shell in one or more of the Gen* utilities.

ardbiesheuvel:
> Does anyone else who's commented on this want to come out for or against my idea that [the shell might as well be _built_ in RELEASE, just not included in FV](https://github.com/tianocore/edk2/pull/11486#issuecomment-3616783708)?
> 

I think it is fine to build it, but any automated signing should be avoided IMO. Note that executables delivered in a FV do not need to be signed to begin with.

I'd prefer a model where, if the UEFI shell is incorporated at all, it is done so in a way where it is not implicitly trusted or signed. When secure boot is disabled, the shell could just execute as usual. When secure boot is enabled, the user would need to enter the BDS menu to add the hash of the UEFI shell binary to `db` in order to allow it to be executed.

Also, I am not a fan of these `DEBUG` vs `RELEASE` shenanigans - secure boot should work on a `DEBUG` build as it does on a `RELEASE` build if you care to enable it.

mikebeaton:
> It is unfortunate to see this closed by inactivity. I believe the thread provides exactly the kind of informed, multi-stakeholder analysis needed to make a security decision. If possible, this PR should be reopened and either merged (with or without modifications), or rejected with a documented rationale.

@pagabuc - Having read the new responses above, IMO unfortunately the PR as is should be rejected, for reasons given by @leiflindholm and @ardbiesheuvel (and somewhat myself).

I believe it comes down to:
 - It should still be possible to build and run the shell myself, when I want to; even when secure boot is enabled (which I can do by signing with my own key, _not_ any publicly distributed one, and enrolling the key on my machine; or just by adding a hash, on my machine)
 
Assuming we agree that that should still be possible, what we are trying to avoid being done by mistake is:

 - Shipping signed (with publicly distributed keys) shells
 - Including shell in FV (where it would be unsigned, but runnable)

ShellPkg/Application/Shell/Shell.c
spbrogan: I don't think this is a necessary feature and should definitely not be viewed as a security feature.  There are numerous ways a platform can and should control the access to the shell (or any other tool that provides functionality that can be used to circumvent platform/boot security).  It is very common to include the shell in the firmware of non-production builds to make debugging and other developer focus activities easier.  I would expect this conditional to be disabled.

benjamindoron:
I'm against this too, on the basis that a developer might leave a shell in their personal firmware for development, enable Secure Boot to protect the OS, and rely on physical control of the hardware to ensure that nobody else gets into the shell. They'd argue that the presence of NSH files already indicates compromise, that they'd make a best effort attempt to protect the platform, and with that in place, an attacker probably won't be both placing the file and setting BootNext to trigger the shell (since this implies compromise already). It's delicate and implies being okay with degraded trust, but I'd see it happening

How about something that allows the firmware to communicate to an OS that it may not be trustworthy? I'd recommend a locked, volatile, runtime variable, but this probably isn't good enough. Bootkits already hook the variable services to spoof that Secure Boot is enabled. Obviously, there are the TPM PCRs (on a typical boot, the shell will not be measured), but while this provides strong guarantees, it's inflexible. Still, it might be appropriate for platforms that are carefully administered and need a high degree of trust.
- The inflexibility of PCR extend could be mitigated with a new entry in the event log that indicates this 'untrustworthy' state, but again, it ties security to the TPM. Thoughts on if this is okay?

PR #11487, opened 2025-09-11
add some igvm support

kraxel:
- **OvmfPkg/igvm: reserve two pages for igvm support in memfd**
- **OvmfPkg/igvm: add igvm regions to reset vector**
- **OvmfPkg/igvm: add igvm memory map support.**
- **OvmfPkg/igvm: add IgvmData struct header + guid**
- **OvmfPkg/igvm: handle igvm data hobs**
- **OvmfPkg/QemuKernelLoaderFsDxe: add support for igvm data blobs**

### DIFF
--- a/OvmfPkg/Library/PlatformInitLib/Igvm.c
+++ b/OvmfPkg/Library/PlatformInitLib/Igvm.c
@@ -21,6 +21,9 @@
 #define MEMORY_MAP_ENTRIES  8
 #define MEMORY_MAP_SIZE     (MEMORY_MAP_ENTRIES * sizeof(IGVM_MEMORY_MAP_ENTRY))
 
+#define VP_COUNT_OFFSET  (MEMORY_MAP_OFFSET + MEMORY_MAP_SIZE)
+#define VP_COUNT_SIZE    16
+
 #define IGVM_MM_ENTRY_TYPE_MEMORY      0x00
 #define IGVM_MM_ENTRY_TYPE_RESERVED    0x01
 #define IGVM_MM_ENTRY_TYPE_PERSISTENT  0x02
@@ -179,4 +182,29 @@ PlatformIgvmDataHobs (
       EfiBootServicesData
       );
   }
+}
+
+UINT32
+EFIAPI
+PlatformIgvmVpCount (
+  VOID
+  )
+{
+  UINT64  Address;
+  UINT32  *VpCount;
+
+  Address = FixedPcdGet64 (PcdOvmfIgvmParamBase);
+  if (Address == 0) {
+    // no parameter area
+    return 0;
+  }
+
+  VpCount = (VOID *)(UINTN)(Address + VP_COUNT_OFFSET);
+  if (!(*VpCount)) {
+    // no vp count
+    return 0;
+  }
+
+  DEBUG ((DEBUG_INFO, "%a: vp-count=%d\n", __func__, *VpCount));
+  return *VpCount;
 }
--- a/OvmfPkg/ResetVector/X64/IgvmMetadata.asm
+++ b/OvmfPkg/ResetVector/X64/IgvmMetadata.asm
@@ -12,6 +12,7 @@ ALIGN 16
 
 %define IGVM_ID_PARAM_AREA                               0x100
 %define IGVM_ID_PARAM_MEMORY_MAP                         0x101
+%define IGVM_ID_PARAM_VP_COUNT                           0x102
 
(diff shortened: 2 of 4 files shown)


ardbiesheuvel:
Where is this IGVM format documented?

kraxel:
> Where is this IGVM format documented?

specs (and rust crate):
https://docs.rs/igvm_defs/0.4.0/igvm_defs/

some tools arounds that is here:
https://gitlab.com/kraxel/virt-firmware-rs/-/commits/igvm-params

qemu support for igvm is evolving still, initial support has been added in version 10.1.

kraxel:
ping ard, anything missing to move this forward?

fyi: latest update adds one patch to support for another igvm parameter (vcpu count).

OvmfPkg/ResetVector/X64/IgvmMetadata.asm
```diff
@@ -0,0 +1,52 @@
...
+ALIGN 16
+
+%define IGVM_ID_PARAM_AREA                               0x100
+%define IGVM_ID_PARAM_MEMORY_MAP                         0x101
+
+%define IGVM_ID_HOB_AREA                                 0x200
+
+%define MEMORY_MAP_OFFSET                                    0
+%define MEMORY_MAP_ENTRIES                                   8
+%define MEMORY_MAP_SIZE              (MEMORY_MAP_ENTRIES * 24)
```
tlendacky, line 22: Does this mean that an entry is 24 bytes in size and the area can only hold a maximum of 8 entries? Some kind of comments that refer to the eventual support and what has to be kept in sync would be good to have here.

OvmfPkg/Library/PlatformInitLib/Igvm.c
```diff
@@ -0,0 +1,110 @@
...
+
+**/
+
+#include <Library/PlatformInitLib.h>
+#include <Library/DebugLib.h>
+
+/* keep in sync with OvmfPkg/ResetVector/X64/IgvmMetadata.asm */
+#define MEMORY_MAP_OFFSET   0
+#define MEMORY_MAP_ENTRIES  8
+#define MEMORY_MAP_SIZE     (MEMORY_MAP_ENTRIES * 24)
```
tlendacky: Should this use the size of the IGVM_MEMORY_MAP_ENTRY instead of hardcoding 24?

OvmfPkg/ResetVector/X64/IgvmMetadata.asm
```diff
@@ -15,6 +15,7 @@ ALIGN 16
 
 %define IGVM_ID_HOB_AREA                                 0x200
 
+; keep in sync with OvmfPkg/Library/PlatformInitLib/Igvm.c
```
tlendacky, line 19: Ah, this covers my concern in the previous patch.

OvmfPkg/Library/PlatformInitLib/Igvm.c
```diff
@@ -108,3 +115,68 @@ PlatformIgvmScanE820 (
...
+
+  Start = FixedPcdGet64 (PcdOvmfIgvmHobBase);
+  End   = Start + FixedPcdGet64 (PcdOvmfIgvmHobSize);
+  if (Start == 0) {
+    // no parameter area
+    return;
+  }
+
+  Hob.Raw = (VOID *)(UINTN)Start;
+  for ( ; ;) {
```
tlendacky: A nit, but could the "Hob.Raw = (VOID *)(UINTN)Start" be the assignment part of the for loop, the HobType check be the conditional and then the "GET_NEXT_HOB (Hob)" be the next loop value. It just seems strange seeing the for loop like this.

OvmfPkg/ResetVector/X64/IgvmMetadata.asm
kraxel: Yes, it's 24 bytes per entry + room for 8.  A typical VM configuration needs three entries.

OvmfPkg/Library/PlatformInitLib/Igvm.c
kraxel: Fixed.

OvmfPkg/Library/PlatformInitLib/Igvm.c
kraxel: Moved start + next.  Left the conditionals.  It's actually three conditions, and leaving them as-is is IMHO better than moving them all into the if conditional, or split them up and move only one.

PR #11489, opened 2025-09-11
OvmfPkg: Use TSC instead of ACPI timer

nxfs:
Avoid unnecessary VM exits and use TSC as performance counter
implementation. This mirrors the PcAtChipsetPkg implementation.

We observed VM exit overhead when the booting processor waits for APs in
TimedWaitForApFinish.

### DIFF
--- a/OvmfPkg/Library/AcpiTimerLib/AcpiTimerLib.c
+++ b/OvmfPkg/Library/AcpiTimerLib/AcpiTimerLib.c
@@ -124,8 +124,7 @@ NanoSecondDelay (
 
 **/
 UINT64
-EFIAPI
-GetPerformanceCounter (
+AcpiGetPerformanceCounter (
   VOID
   )
 {
@@ -156,8 +155,7 @@ GetPerformanceCounter (
 
 **/
 UINT64
-EFIAPI
-GetPerformanceCounterProperties (
+AcpiGetPerformanceCounterProperties (
   OUT      UINT64  *StartValue   OPTIONAL,
   OUT      UINT64  *EndValue     OPTIONAL
   )
--- a/OvmfPkg/Library/AcpiTimerLib/BaseAcpiTimerLib.c
+++ b/OvmfPkg/Library/AcpiTimerLib/BaseAcpiTimerLib.c
@@ -10,6 +10,8 @@
 #include <Library/IoLib.h>
 #include <Library/PciLib.h>
 #include <OvmfPlatforms.h>
+#include "AcpiTimerLib.h"
+#include "TscTimerLib.h"
 
 //
 // Cached ACPI Timer IO Address
@@ -107,4 +109,41 @@ InternalAcpiGetTimerTick (
   //   Return the current ACPI timer value.
   //
   return IoRead32 (mAcpiTimerIoAddr);
+}
+
+/**
+  Retrieves the current value of a 64-bit free running performance counter.
+
+  @return The current value of the free running performance counter.
+
+**/
+UINT64
+EFIAPI
+GetPerformanceCounter (
+  VOID
+  )
+{
+  return IsTscTimerFrequencyAvailable () ? TscGetPerformanceCounter () : AcpiGetPerformanceCounter ();
+}
+
+/**
+  Retrieves the 64-bit frequency in Hz and the range of performance counter
+  values.
+
+  @param  StartValue  The value the performance counter starts with when it
+                      rolls over.
+  @param  EndValue    The value that the performance counter ends with before
+                      it rolls over.
+
+  @return The frequency in Hz.
+
+**/
+UINT64
+EFIAPI
+GetPerformanceCounterProperties (
+  OUT      UINT64  *StartValue   OPTIONAL,
+  OUT      UINT64  *EndValue     OPTIONAL
+  )
+{
+  return IsTscTimerFrequencyAvailable () ? TscGetPerformanceCounterProperties (StartValue, EndValue) : AcpiGetPerformanceCounterProperties (StartValue, EndValue);
 }
--- a/OvmfPkg/Library/AcpiTimerLib/BaseAcpiTimerLib.inf
+++ b/OvmfPkg/Library/AcpiTimerLib/BaseAcpiTimerLib.inf
@@ -21,6 +21,7 @@
   AcpiTimerLib.c
   AcpiTimerLib.h
   BaseAcpiTimerLib.c
+  TscTimerLib.c
 
 [Packages]
   MdePkg/MdePkg.dec
--- a/OvmfPkg/Library/AcpiTimerLib/BaseAcpiTimerLibBhyve.c
+++ b/OvmfPkg/Library/AcpiTimerLib/BaseAcpiTimerLibBhyve.c
@@ -10,6 +10,8 @@
 
 #include <Library/IoLib.h>
 #include <OvmfPlatforms.h>
+#include "AcpiTimerLib.h"
+#include "TscTimerLib.h"
 
 /**
   Internal function to read the current tick counter of ACPI.
@@ -29,4 +31,41 @@ InternalAcpiGetTimerTick (
   // Return the current ACPI timer value.
   //
   return IoRead32 (BHYVE_ACPI_TIMER_IO_ADDR);
+}
+
+/**
+  Retrieves the current value of a 64-bit free running performance counter.
+
+  @return The current value of the free running performance counter.
+
+**/
+UINT64
+EFIAPI
+GetPerformanceCounter (
+  VOID
+  )
+{
+  return IsTscTimerFrequencyAvailable () ? TscGetPerformanceCounter () : AcpiGetPerformanceCounter ();
+}
+
+/**
+  Retrieves the 64-bit frequency in Hz and the range of performance counter
+  values.
+
+  @param  StartValue  The value the performance counter starts with when it
+                      rolls over.
+  @param  EndValue    The value that the performance counter ends with before
+                      it rolls over.
+
+  @return The frequency in Hz.
+
+**/
+UINT64
+EFIAPI
+GetPerformanceCounterProperties (
+  OUT      UINT64  *StartValue   OPTIONAL,
+  OUT      UINT64  *EndValue     OPTIONAL
+  )
+{
+  return IsTscTimerFrequencyAvailable () ? TscGetPerformanceCounterProperties (StartValue, EndValue) : AcpiGetPerformanceCounterProperties (StartValue, EndValue);
 }
--- a/OvmfPkg/Library/AcpiTimerLib/BaseAcpiTimerLibBhyve.inf
+++ b/OvmfPkg/Library/AcpiTimerLib/BaseAcpiTimerLibBhyve.inf
@@ -21,10 +21,15 @@
   AcpiTimerLib.c
   AcpiTimerLib.h
   BaseAcpiTimerLibBhyve.c
+  TscTimerLib.c
+  TscTimerLib.h
 
 [Packages]
   MdePkg/MdePkg.dec
   OvmfPkg/OvmfPkg.dec
 
 [LibraryClasses]
+  BaseLib
+  BaseMemoryLib
+  DebugLib
   IoLib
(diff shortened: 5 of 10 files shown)


OvmfPkg/Library/AcpiTimerLib/AcpiTimerLib.c
```diff
@@ -10,10 +10,13 @@
 
 #include <Library/DebugLib.h>
 #include <Library/BaseLib.h>
+#include <Library/BaseMemoryLib.h>
 #include <IndustryStandard/Acpi.h>
 
 #include "AcpiTimerLib.h"
 
+STATIC UINT64  mTscFrequency = 0;
+
```
kraxel: Using global variables here is a non-starter.  The library is used for PEI boot phase, global variables are only allowed in DXE boot phase.

OvmfPkg/Library/AcpiTimerLib/AcpiTimerLib.c
```diff
@@ -129,7 +132,105 @@ GetPerformanceCounter (
...
-  return (UINT64)InternalAcpiGetTimerTick ();
+  return AsmReadTsc ();
+}
+
+/**
+  Calculate TSC frequency.
+
+  The TSC counting frequency is determined by comparing how far it counts
+  during a 101.4 us period as determined by the ACPI timer.
+  The ACPI timer is used because it counts at a known frequency.
```
kraxel: No.  Doing timer calibration like that simply does not work reliable in virtual machines.  If the host stops the vcpu in the middle of the loop the result can be /way/ off.  This is the reason why edk2 uses a fixed frequency timer.

nxfs: Makes sense. Would it be acceptable to remove the calibration fallback and use TSC only when its frequency is advertised in CPUID?

OvmfPkg/Library/AcpiTimerLib/AcpiTimerLib.c
nxfs: Ok. This global variable can easily be removed if we eliminate the calibration path as per your other comment.

nxfs:
> Which qemu version is this btw? The just released 10.2 version should behave much better thanks to qemu commit 4ae5e2b2bf65452538511a2895d7a4e2115058a5

For L1TF reloaded some hypervisors stop running both vCPU siblings when one exits, so the booting CPU slows down the AP significantly if we use an emulated interface

OvmfPkg/Library/AcpiTimerLib/AcpiTimerLib.c
dwmw2: Hm, in that case why is it OK to have `mAcpiTimerIoAddr` in BaseAcpiTimerLib.c?

dwmw2: ... and DxeAcpiTimerLib.c too.

kraxel: There are two variants of the library, one for early boot phases (BaseRomAcpiTimerLib.inf) not using global variables and one for the later boot phases (BaseAcpiTimerLib.inf) which does cache stuff in global variables instead of calculating things each time they are needed.

dwmw2: OK, so we could use global variables in the later boot phases but not in BaseRomAcpiTimerLib. I think that's sufficient.

The main driver for this is a hypervisor which does not allow sibling CPU threads to run while one of their siblings has trapped to the hypervisor (see L1TF reloaded for an example of why a hypervisor might want to do that).

Basically, while the BSP has sent INIT/SIPI to BSPs to wake them up, bashing on the PM Timer to wait for a timeout is actively preventing them from making progress. Hence wanting to use the TSC instead, which is handled without a vmexit and *does* allow the siblings to progress.

OvmfPkg/Library/AcpiTimerLib/AcpiTimerLib.c
kraxel: I think you only get that when asking for it explicitly (via `qemu -cpu +invtsc`).
So not sure how well that'll work in practice.  But, yes, that would be better.

kraxel:
> For L1TF reloaded some hypervisors stop running both vCPU siblings when one exits, so the booting CPU slows down the AP significantly if we use an emulated interface

Where do you have that?  PEI or DXE or both?

PEI is more tricky because you can't cache the detection and I think cpuid will vmexit too.

Maybe the better approach is to change the TimedWaitForApFinish function to do the timeout checking in a more virtualization friendly way.

nxfs:
> Where do you have that? PEI or DXE or both?

It's on TimedWaitForApFinish called from WakeUpAP the first time we collect CPUs. As far as I can tell this is DXE and BaseAcpiTimerLib.c is invoked. I've done A/B testing and checked that the VM exit disappeared.

> Maybe the better approach is to change the TimedWaitForApFinish function to do the timeout checking in a more virtualization friendly way.

Yes we can also do that, add backoff to the busy wait loop.

nxfs:
New revision addresses the comment and doesn't use cache data for BaseRomAcpiTimerLib. Also reorganised the code to make it somewhat more logical with a separate file to handle Tsc.

kraxel:
The new file for Tsc support certainly is an improvement.  Hiding the acpi fallback logic there is not so nice though.  Can we move the branching to acpi / tsc code to the `GetPerformanceCounter()` function, so it is easier to follow what is happening?

nxfs:
> The new file for Tsc support certainly is an improvement. Hiding the acpi fallback logic there is not so nice though. Can we move the branching to acpi / tsc code to the GetPerformanceCounter() function, so it is easier to follow what is happening?

Done in new revision, and also fixed BaseAcpiTimerLibBhyve.c build issues picked up by CI.

nxfs:
Checking CI failures. The OVMF_IA32X64_NOOPT seems to be because I'm using native multiplication, I'll update the PR with a fix. Local build works though, so not sure exactly how my config deviates from CI. 

The other X64_MM_DEBUG failure seems like unrelated timeout? I can't repro locally, it runs fine.

nxfs:
As per discussion added timed wait backoff.
I put it on a separate PR. Let me know if preferable to keep in the same PR.
https://github.com/tianocore/edk2/pull/11501

nxfs:
I see the X64_MM_DEBUG continues to fail on the CI. I would appreciate some guidance on how to troubleshoot it as I am unable to repro locally.

I ran the following steps:

```
sudo rm -rf Build && time docker run --rm --dns=192.168.1.1 --dns=8.8.8.8 -v $(pwd):/home/edk2 -w /home/edk2 ghcr.io/tianocore/containers/fedora-41-test:4dbfa9e bash -c "dnf install -y acpica-tools && git config --global --add safe.directory '*' && stuart_setup -c OvmfPkg/PlatformCI/PlatformBuild.py TOOL_CHAIN_TAG=GCC5 TARGET=DEBUG -a X64 && stuart_update -c OvmfPkg/PlatformCI/PlatformBuild.py TOOL_CHAIN_TAG=GCC5 TARGET=DEBUG -a X64 && stuart_build -c OvmfPkg/PlatformCI/PlatformBuild.py TOOL_CHAIN_TAG=GCC5 TARGET=DEBUG -a X64 BLD_*_SECURE_BOOT_ENABLE=1 BLD_*_SMM_REQUIRE=1 BLD_*_TPM2_ENABLE=1 BLD_*_NETWORK_TLS_ENABLE=1 BLD_*_NETWORK_IP6_ENABLE=1 BLD_*_NETWORK_HTTP_BOOT_ENABLE=1 BLD_*_STANDALONE_MM_ENABLE=1 MAKE_STARTUP_NSH=TRUE QEMU_HEADLESS=TRUE && stuart_build -c OvmfPkg/PlatformCI/PlatformBuild.py TOOL_CHAIN_TAG=GCC5 TARGET=DEBUG -a X64 BLD_*_SECURE_BOOT_ENABLE=1 BLD_*_SMM_REQUIRE=1 BLD_*_TPM2_ENABLE=1 BLD_*_NETWORK_TLS_ENABLE=1 BLD_*_NETWORK_IP6_ENABLE=1 BLD_*_NETWORK_HTTP_BOOT_ENABLE=1 BLD_*_STANDALONE_MM_ENABLE=1 MAKE_STARTUP_NSH=TRUE QEMU_HEADLESS=TRUE --FlashOnly"
```

Output - SmmRelocation works (hangs on CI report https://dev.azure.com/tianocore/edk2-ci/_build/results?buildId=200048&view=logs&j=7a85138a-5e2d-569e-7f8d-012982ba61ab&t=cc467a88-110e-5c2d-78a0-8cccd3d88f4a)
```
...
INFO - SmmRelocationInit Start                                                              
INFO - SmmRelocationInit - SmmRelocationSize: 0x0000E000
INFO - SmmRelocationInit - SmmRelocationStart: 0x06001000
INFO - SmmRelocationInit - SmmStackSize: 0x00001000                      
INFO - SmmRelocationInit - SmmStacks: 0x05F97000                         
INFO - CreateSmmBaseHob - MaxCapOfProcessorsInHob: 8188
INFO - CreateSmmBaseHob - SmmBaseHobData[0]->ProcessorIndex: 0
INFO - CreateSmmBaseHob - SmmBaseHobData[0]->NumberOfProcessors: 4       
INFO - CreateSmmBaseHob - SmmBaseHobData[0]->SmBase[0]: 0x05FF9000       
INFO - CreateSmmBaseHob - SmmBaseHobData[0]->SmBase[1]: 0x05FFB000
INFO - CreateSmmBaseHob - SmmBaseHobData[0]->SmBase[2]: 0x05FFD000                          
INFO - CreateSmmBaseHob - SmmBaseHobData[0]->SmBase[3]: 0x05FFF000       
INFO - SmmRelocationInit Done
...
```

Command completes in 37s (most of it being the prep step)
```
docker run --rm --dns=192.168.1.1 --dns=8.8.8.8 -v $(pwd):/home/edk2 -w   bas  0.08s user 0.13s system 0% cpu 37.239 total
```

OvmfPkg/Library/AcpiTimerLib/AcpiTimerLib.c
nxfs: Latest rev addresses this comment

kraxel:
fallback to acpi timer not working properly?  I see earlier in the log:
```
INFO - No TSC timer frequency available
```

nxfs:
> fallback to acpi timer not working properly? I see earlier in the log:
> 
> ```
> INFO - No TSC timer frequency available
> ```

Thank you for your engagement.

I saw the log too, it's now expected behavior for a library that uses TSC.

Take OvmfPkg/Library/AcpiTimerLib/BaseAcpiTimerLib.c

call GetPerformanceCounterProperties -> 
call IsTscTimerFrequencyAvailable ->
call InternalInitializeTimerSource ->
looks for TSC freq, doesn't find it, prints debug message, returns 0 freq
-> IsTscTimerFrequencyAvailable returns false since freq is zero
-> GetPerformanceCounterProperties uses ACPI timer fallback instead

I've triggered the CI a few times and OVMF_X64_MM_DEBUG seems to consistently hang there

```
INFO - SmmRelocationInit Start
INFO - SmmRelocationInit - SmmRelocationSize: 0x0000E000
INFO - SmmRelocationInit - SmmRelocationStart: 0x06001000
INFO - SmmRelocationInit - SmmStackSize: 0x00001000
INFO - SmmRelocationInit - SmmStacks: 0x05F97000
```

I have looked at that code in this area, I don't understand how it can interact with the TSC logic that I implemented.
You mentioned we cannot use static variables in PEI, very long shot but I'm wondering if this might be accidentally violated in this scenario and cause some memory corruption..? I don't have sufficient understanding of memory handling in EDK yet to reason about whether this might be a cause.

kraxel:
> I saw the log too, it's now expected behavior for a library that uses TSC.

Yea, sure.  In CI qemu runs with the default config, which I think excludes +invtsc so the quoted message and the code trying to fallback to the acpi timer is expected.  So maybe tsc works and acpi does not?

Building with `-DSMM_REQUIRE=TRUE` will produce the SMM binary btw if you want one for local testing.

PR #11492, opened 2025-09-11
MdeModulePkg: Add PcdDelayedDispatchMaxEntries

makubacki:
# Description

The current fixed value of 8 for `DELAYED_DISPATCH_MAX_ENTRIES` is
not large enough to accommodate platform usage. This change replaces
the macro with a FixedAtBuild PCD that can be configured by
platforms.

In the case the default PCD value is too small, an error message
explaining that the PCD should be updated will be printed followed
by an assert.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- MdeModulePkg CI
- Integrate into a platform with >8 delayed dispatch entries and
  `PcdDelayedDispatchMaxEntries` set to a platform-specific value.

## Integration Instructions

- No action needed. A platform can choose to configure `PcdDelayedDispatchMaxEntries` as needed.

### DIFF
--- a/MdeModulePkg/Core/Pei/Dispatcher/Dispatcher.c
+++ b/MdeModulePkg/Core/Pei/Dispatcher/Dispatcher.c
@@ -134,9 +134,15 @@ PeiDelayedDispatchRegister (
   }
 
   // Check for available entry slots
-  ASSERT (DelayedDispatchTable->Count <= DELAYED_DISPATCH_MAX_ENTRIES);
-  if (DelayedDispatchTable->Count == DELAYED_DISPATCH_MAX_ENTRIES) {
-    DEBUG ((DEBUG_ERROR, "%a Too many entries requested\n", __func__));
+  if (DelayedDispatchTable->Count >= PcdGet32 (PcdDelayedDispatchMaxEntries)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a DelayedDispatchTable->Count = %d. PcdDelayedDispatchMaxEntries (%d) is too small.\n",
+      __func__,
+      DelayedDispatchTable->Count,
+      PcdGet32 (PcdDelayedDispatchMaxEntries)
+      ));
+    ASSERT (DelayedDispatchTable->Count < PcdGet32 (PcdDelayedDispatchMaxEntries));
     Status = EFI_OUT_OF_RESOURCES;
     goto Exit;
   }
@@ -1807,7 +1813,7 @@ PeiDispatcher (
     if (GuidHob != NULL) {
       Private->DelayedDispatchTable = (DELAYED_DISPATCH_TABLE *)(GET_GUID_HOB_DATA (GuidHob));
     } else {
-      TableSize                     = sizeof (DELAYED_DISPATCH_TABLE) + ((DELAYED_DISPATCH_MAX_ENTRIES - 1) * sizeof (DELAYED_DISPATCH_ENTRY));
+      TableSize                     = sizeof (DELAYED_DISPATCH_TABLE) + (PcdGet32 (PcdDelayedDispatchMaxEntries) * sizeof (DELAYED_DISPATCH_ENTRY));
       Private->DelayedDispatchTable = BuildGuidHob (&gEfiDelayedDispatchTableGuid, TableSize);
       if (Private->DelayedDispatchTable != NULL) {
         ZeroMem (Private->DelayedDispatchTable, TableSize);
--- a/MdeModulePkg/Include/Guid/DelayedDispatch.h
+++ b/MdeModulePkg/Include/Guid/DelayedDispatch.h
@@ -1,7 +1,7 @@
 /** @file
   Definition for structure & defines exported by Delayed Dispatch PPI
 
-  Copyright (c), Microsoft Corporation.
+  Copyright (c) Microsoft Corporation.
 
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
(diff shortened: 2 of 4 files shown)


MdeModulePkg/Include/Guid/DelayedDispatch.h
```diff
@@ -18,7 +18,7 @@
 //
 // Maximal number of Delayed Dispatch entries supported
 //
-#define DELAYED_DISPATCH_MAX_ENTRIES  8
+#define DELAYED_DISPATCH_MAX_ENTRIES  32
```
lgao4: Is this one defined in PI spec？

makubacki: [`EFI_DELAYED_DISPATCH_PPI`](https://uefi.org/specs/PI/1.9/V1_Additional_PPIs.html#efi-delayed-dispatch-ppi) is defined in the PI Spec. I'd expect that this is considered more of an implementation detail.

lgao4: So, this macro value is the implementation detail, right?

makubacki: Yes. There is a platform I'm aware of that has around 14 or so entries. This is mean to be a reasonable increase to provide some buffer to that known max case.

mdkinney: There are a couple PCDs already defined for the Delayed Dispatch feature.

```
  ## Delayed Dispatch Maximum Delay in us (microseconds)
  # Maximum delay for any particular delay request - 5 seconds
  gEfiMdeModulePkgTokenSpaceGuid.PcdDelayedDispatchMaxDelayUs|5000000|UINT32|0x3000104A

  ## Delayed Dispatch timeout in us (microseconds)
  # Maximum delay when waiting for completion (ie EndOfPei) - 10 seconds
  gEfiMdeModulePkgTokenSpaceGuid.PcdDelayedDispatchCompletionTimeoutUs|10000000|UINT32|0x3000104B
```

Would it be better to define a PCD for the mac entries with a default of 8 and can be configured to larger values if needed for specific platforms?

makubacki: The original implementation proposal had a PCD with a default value of `8` - https://edk2.groups.io/g/devel/message/109471.

I understand the goal of minimizing configuration options where possible, but this seems like a reasonable PCD. I’m happy to update the patch if there are no objections.

mdkinney:
I would prefer no #define and no PCD.  Just grow the buffer as needed, but this consumes extra PEI HOB list memory to alloc new and free old.

The PCD works because there is an ASSERT() with error message.  If the error message told what PCD value needed to be adjusted, then platform integration can determine what size is needed.

MdeModulePkg/Include/Guid/DelayedDispatch.h
makubacki: Closing as no objections in this thread and another comment indicates the PCD is acceptable with the size being reported if too small.

makubacki:
> I would prefer no #define and no PCD. Just grow the buffer as needed, but this consumes extra PEI HOB list memory to alloc new and free old.
> 
> The PCD works because there is an ASSERT() with error message. If the error message told what PCD value needed to be adjusted, then platform integration can determine what size is needed.

Added the PCD with the error message and ASSERT().

mdkinney:
Recommend PCD allow FixedAtBuiild and PatchableInModule in DEC and use PcdGetxx() instead of FixedPcdGetxx().

makubacki:
> Recommend PCD allow FixedAtBuiild and PatchableInModule in DEC and use PcdGetxx() instead of FixedPcdGetxx().

Done

MdeModulePkg/Include/Guid/DelayedDispatch.h
```diff
@@ -37,7 +32,7 @@ typedef struct {
 typedef struct {
   UINT32                    Count;
   UINT32                    DispCount;
-  DELAYED_DISPATCH_ENTRY    Entry[DELAYED_DISPATCH_MAX_ENTRIES];
+  DELAYED_DISPATCH_ENTRY    Entry[1];   // Number of entries is based on PcdDelayedDispatchMaxEntries
```
mdkinney: For structure with flexible array member, it should be [].  Not [1] and not [0].

makubacki: Updated. I also filed https://github.com/tianocore/edk2/issues/11528 so the 254 other cases in the codebase can be tracked for cleanup.

makubacki:
@lgao4, please revisit.

makubacki:
@mdkinney, please revisit as well.

MdeModulePkg/Core/Pei/Dispatcher/Dispatcher.c
```diff
@@ -1807,7 +1813,7 @@ PeiDispatcher (
     if (GuidHob != NULL) {
       Private->DelayedDispatchTable = (DELAYED_DISPATCH_TABLE *)(GET_GUID_HOB_DATA (GuidHob));
     } else {
-      TableSize                     = sizeof (DELAYED_DISPATCH_TABLE) + ((DELAYED_DISPATCH_MAX_ENTRIES - 1) * sizeof (DELAYED_DISPATCH_ENTRY));
+      TableSize                     = sizeof (DELAYED_DISPATCH_TABLE) + ((PcdGet32 (PcdDelayedDispatchMaxEntries)  - 1) * sizeof (DELAYED_DISPATCH_ENTRY));
```
mdkinney: I think there is a logic bug in the old code and the new code.

The old code uses sizeof(DELAYED_DISPATCH_TABLE) which includes the array of DELAYED_DISPATCH_MAX_ENTRIES, and then TableSize adds another DELAYED_DISPATCH_MAX_ENTRIES-1 entries.  For 8, TableSize would be set to 15 entries.

The new code subtracts 1, which for 8, would only allocate 7 entries and it should allocate 8 entries.

makubacki: @mdkinney, you're correct. Sorry I didn't revaluate that logic. The update is pushed.

mdkinney (approved):

PR #11494, opened 2025-09-11
Global: Remove ARM32 Support

os-d:
# Description

edk2 has decided to remove support for the deprecated ARM32 architecture. This commit removes all ARM32 support from edk2.

A few notes:
- Industry standard definitions were left alone, but implementations were removed
- Some potentially ARM32 only things were left in tree due to my uncertainty whether it was truly ARM32 or for one of the following reasons
  - Unclear if some ARMv7 values were scoped to ARM32 (since they were noted for compatibility in AARCH64)
  - ARM is used to refer to ARM32, AARCH64, and the company, so it can be hard to piece out which it is actually talking about
  - ARM32 only PCDs, drivers, libraries, etc may have been left if they did not specifically call out they were ARM32 only
- There may be more optimal merging of existing code, e.g. many libraries have the pattern of Common.c calls into Arm/Arch.c or AArch64/Arch.c. I removed the Arm directory and moved the AArch64/Arch.c file up to the top level, getting rid of the AArch64 directory. However, it may make sense to then combine Common.c and Arch.c. This will need to be done by the maintainers of those components if they see fit. 

- [x] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Build tests.

## Integration Instructions

ARM32 is no longer supported by edk2. ARM32 platform development can still be done on edk2-stable202508.

### DIFF
--- a/BaseTools/Conf/tools_def.template
+++ b/BaseTools/Conf/tools_def.template
@@ -1663,7 +1663,7 @@ DEFINE CLANGPDB_IA32_TARGET          = -target i686-pc-windows-msvc
 DEFINE CLANGPDB_X64_TARGET           = -target x86_64-pc-windows-msvc
 
 DEFINE CLANGPDB_WARNING_OVERRIDES    = -Wno-parentheses-equality -Wno-tautological-compare -Wno-tautological-constant-out-of-range-compare -Wno-empty-body -Wno-unused-const-variable -Wno-varargs -Wno-unknown-warning-option -Wno-unused-but-set-variable -Wno-unused-const-variable -Wno-unaligned-access -Wno-microsoft-enum-forward-reference
-DEFINE CLANGPDB_ALL_CC_FLAGS         = DEF(GCC48_ALL_CC_FLAGS) DEF(CLANGPDB_WARNING_OVERRIDES) -fno-stack-protector -funsigned-char -ftrap-function=undefined_behavior_has_been_optimized_away_by_clang -Wno-address -Wno-shift-negative-value -Wno-unknown-pragmas -Wno-incompatible-library-redeclaration -Wno-null-dereference -mno-implicit-float -mms-bitfields -mno-stack-arg-probe -fno-omit-frame-pointer -U _MSC_VER -D __GNUC__
+DEFINE CLANGPDB_ALL_CC_FLAGS         = DEF(GCC48_ALL_CC_FLAGS) DEF(CLANGPDB_WARNING_OVERRIDES) -fno-stack-protector -funsigned-char -ftrap-function=undefined_behavior_has_been_optimized_away_by_clang -Wno-address -Wno-shift-negative-value -Wno-unknown-pragmas -Wno-incompatible-library-redeclaration -Wno-null-dereference -mno-implicit-float -mms-bitfields -mno-stack-arg-probe -fno-omit-frame-pointer -U _MSC_VER -D __GNUC__=4 -D __GNUC_MINOR__=2 -D __GNUC_PATCHLEVEL__=1 -D __MINGW32__=1
 
 ###########################
 # CLANGPDB IA32 definitions
--- a/CryptoPkg/Driver/Crypto.c
+++ b/CryptoPkg/Driver/Crypto.c
@@ -5125,6 +5125,30 @@ CryptoServiceTlsSetServerName (
 }
 
 /**
+  Set the Tls security level.
+
+  This function Set the Tls security level.
+  If Tls is NULL, nothing is done.
+
+  @param[in]  Tls                Pointer to the TLS object.
+  @param[in]  Level              The Tls Security level need to set.
+
+  @retval  EFI_SUCCESS           The Tls security level was set successfully.
+  @retval  EFI_INVALID_PARAMETER The parameters are invalid.
+  @retval  EFI_UNSUPPORTED       The requested TLS set security level is not supported.
+
+**/
+EFI_STATUS
+EFIAPI
+CryptoServiceTlsSetSecurityLevel (
+  IN VOID   *Tls,
+  IN UINT8  Level
+  )
+{
+  return CALL_BASECRYPTLIB (TlsSet.Services.SecurityLevel, TlsSetSecurityLevel, (Tls, Level), EFI_UNSUPPORTED);
+}
+
+/**
   Set the signature algorithm list to used by the TLS object.
 
   This function sets the signature algorithms for use by a specified TLS object.
@@ -7137,6 +7161,7 @@ const EDKII_CRYPTO_PROTOCOL  mEdkiiCrypto = {
   CryptoServicePkcs1v2Decrypt,
   CryptoServiceRsaOaepEncrypt,
   CryptoServiceRsaOaepDecrypt,
-  /// TLS Set (continued)
+  /// TLS Set (Continued)
   CryptoServiceTlsSetServerName,
+  CryptoServiceTlsSetSecurityLevel,
 };
--- a/CryptoPkg/Include/Library/TlsLib.h
+++ b/CryptoPkg/Include/Library/TlsLib.h
@@ -645,6 +645,26 @@ TlsSetEcCurve (
   );
 
 /**
+  Set the Tls security level.
+
+  This function Set the Tls security level.
+  If Tls is NULL, nothing is done.
+
+  @param[in]  Tls                Pointer to the TLS object.
+  @param[in]  Level              Tls Security level need to set.
+
+  @retval  EFI_SUCCESS           The Tls security level was set successfully.
+  @retval  EFI_INVALID_PARAMETER The parameters are invalid.
+
+**/
+EFI_STATUS
+EFIAPI
+TlsSetSecurityLevel (
+  IN VOID   *Tls,
+  IN UINT8  Level
+  );
+
+/**
   Gets the protocol version used by the specified TLS connection.
 
   This function returns the protocol version used by the specified TLS
--- a/CryptoPkg/Include/Pcd/PcdCryptoServiceFamilyEnable.h
+++ b/CryptoPkg/Include/Pcd/PcdCryptoServiceFamilyEnable.h
@@ -328,6 +328,7 @@ typedef struct {
       UINT8    SignatureAlgoList  : 1;
       UINT8    EcCurve            : 1;
       UINT8    ServerName         : 1;
+      UINT8    SecurityLevel      : 1;
     } Services;
     UINT32    Family;
   } TlsSet;
--- a/CryptoPkg/Library/BaseCryptLibOnProtocolPpi/CryptLib.c
+++ b/CryptoPkg/Library/BaseCryptLibOnProtocolPpi/CryptLib.c
@@ -4403,6 +4403,30 @@ TlsSetEcCurve (
 }
 
 /**
+  Set the Tls security level.
+
+  This function Set the Tls security level.
+  If Tls is NULL, nothing is done.
+
+  @param[in]  Tls                Pointer to the TLS object.
+  @param[in]  Level              Tls Security level need to set.
+
+  @retval  EFI_SUCCESS           The Tls security level was set successfully.
+  @retval  EFI_INVALID_PARAMETER The parameters are invalid.
+  @retval  EFI_UNSUPPORTED       The requested TLS set security level is not supported.
+
+**/
+EFI_STATUS
+EFIAPI
+TlsSetSecurityLevel (
+  IN VOID   *Tls,
+  IN UINT8  Level
+  )
+{
+  CALL_CRYPTO_SERVICE (TlsSetSecurityLevel, (Tls, Level), EFI_UNSUPPORTED);
+}
+
+/**
   Gets the protocol version used by the specified TLS connection.
 
   This function returns the protocol version used by the specified TLS
--- a/CryptoPkg/Library/TlsLib/InternalTlsLib.h
+++ b/CryptoPkg/Library/TlsLib/InternalTlsLib.h
@@ -25,6 +25,16 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <openssl/bio.h>
 #include <openssl/err.h>
 
+///
+/// TLS Minimum security level
+#define MIN_SECURITY_LEVEL  2
+///
+
+///
+/// TLS Maximum security level
+#define MAX_SECURITY_LEVEL  5
+///
+
 typedef struct {
   //
   // Main SSL Connection which is created by a server or a client
--- a/CryptoPkg/Library/TlsLib/TlsConfig.c
+++ b/CryptoPkg/Library/TlsLib/TlsConfig.c
@@ -1192,6 +1192,46 @@ TlsSetEcCurve (
 }
 
 /**
+  Set the Tls security level.
+
+  This function Set the Tls security level.
+  If Tls is NULL, nothing is done.
+
+  @param[in]  Tls                Pointer to the TLS object.
+  @param[in]  Level              Tls Security level need to set.
+
+  @retval  EFI_SUCCESS           The Tls security level was set successfully.
+  @retval  EFI_INVALID_PARAMETER The parameters are invalid.
+  @retval  EFI_UNSUPPORTED       The requested TLS set security level is not supported.
+
+**/
+EFI_STATUS
+EFIAPI
+TlsSetSecurityLevel (
+  IN VOID   *Tls,
+  IN UINT8  Level
+  )
+{
+  TLS_CONNECTION  *TlsConn;
+
+  TlsConn = (TLS_CONNECTION *)Tls;
+
+  if ((TlsConn == NULL) || (TlsConn->Ssl == NULL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  // Check if the security level is within the valid range (2 to 5).
+  // Return EFI_INVALID_PARAMETER if the level is out of bounds.
+  if ((Level < MIN_SECURITY_LEVEL) || (Level > MAX_SECURITY_LEVEL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  SSL_set_security_level (TlsConn->Ssl, Level);
+
+  return EFI_SUCCESS;
+}
+
+/**
   Gets the protocol version used by the specified TLS connection.
 
   This function returns the protocol version used by the specified TLS
--- a/CryptoPkg/Library/TlsLib/TlsInit.c
+++ b/CryptoPkg/Library/TlsLib/TlsInit.c
@@ -188,7 +188,7 @@ TlsNew (
   //
   // This retains compatibility with previous version of OpenSSL.
   //
-  SSL_set_security_level (TlsConn->Ssl, 0);
+  SSL_set_security_level (TlsConn->Ssl, 3);
 
   //
   // Initialize the created SSL Object
--- a/CryptoPkg/Library/TlsLibNull/TlsConfigNull.c
+++ b/CryptoPkg/Library/TlsLibNull/TlsConfigNull.c
@@ -401,6 +401,31 @@ TlsSetEcCurve (
 }
 
 /**
+  Set the Tls security level.
+
+  This function Set the Tls security level.
+  If Tls is NULL, nothing is done.
+
+  @param[in]  Tls                Pointer to the TLS object.
+  @param[in]  Level              Tls Security level need to set.
+
+  @retval  EFI_SUCCESS           The Tls security level was set successfully.
+  @retval  EFI_INVALID_PARAMETER The parameters are invalid.
+  @retval  EFI_UNSUPPORTED       The requested TLS set security level is not supported.
+
+**/
+EFI_STATUS
+EFIAPI
+TlsSetSecurityLevel (
+  IN VOID   *Tls,
+  IN UINT8  Level
+  )
+{
+  ASSERT (FALSE);
+  return EFI_UNSUPPORTED;
+}
+
+/**
   Gets the protocol version used by the specified TLS connection.
 
   This function returns the protocol version used by the specified TLS
--- a/CryptoPkg/Private/Protocol/Crypto.h
+++ b/CryptoPkg/Private/Protocol/Crypto.h
@@ -21,7 +21,7 @@
 /// the EDK II Crypto Protocol is extended, this version define must be
 /// increased.
 ///
-#define EDKII_CRYPTO_VERSION  18
+#define EDKII_CRYPTO_VERSION  19
 
 ///
 /// EDK II Crypto Protocol forward declaration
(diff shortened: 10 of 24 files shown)


BaseTools/Conf/XMLSchema/DistributionPackage.xsd
```diff
@@ -330,7 +330,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
               having an error
             </xs:documentation>
             <xs:documentation xml:lang="en-us">
-              #3 Allow for inclusion of ARM and future architecture
+              #3 Allow for inclusion of  and future architecture
```
ardbiesheuvel: No idea what this file does but this change looks slightly dodgy. Care to fix this, or explain why changing it like this is appropriate?

ardbiesheuvel (changes requested):
I stopped halfway reviewing this patch.

While I highly appreciate the effort you are making here to perform this major cleanup, there is really no reason whatsoever to do all of this in a single patch, and I dread the day n months from now where bisecting a heisenbug on AArch64 leads to this change, and we'll have to dig through all of the python, tools_def and code changes to narrow it down.

So please split this into:
- removal of CI plumbing
- tools_def/BaseTools changes removing the ability to build for the ARM target
- removal of ARM support from generic modules (at least 1 patch per module)
- removal of ARM support from ArmPkg/ArmPlatformPkg (split into meaningful chunks)

If you don't have the bandwidth, we can try and find someone to share the load.

os-d:
> I stopped halfway reviewing this patch.
> 
> While I highly appreciate the effort you are making here to perform this major cleanup, there is really no reason whatsoever to do all of this in a single patch, and I dread the day n months from now where bisecting a heisenbug on AArch64 leads to this change, and we'll have to dig through all of the python, tools_def and code changes to narrow it down.
> 
> So please split this into:
> 
> * removal of CI plumbing
> * tools_def/BaseTools changes removing the ability to build for the ARM target
> * removal of ARM support from generic modules (at least 1 patch per module)
> * removal of ARM support from ArmPkg/ArmPlatformPkg (split into meaningful chunks)
> 
> If you don't have the bandwidth, we can try and find someone to share the load.

Yeah, that’s fair. I can do that. My general thought was to make this one commit to be very explicit about when ARM32 support was dropped and to make for an easy revert if needed, but since nobody cares about either of those, I agree this should be split up.

BaseTools/Conf/XMLSchema/DistributionPackage.xsd
os-d: I also have no idea what this file is. I’ll drop the change as this looks like it is documenting changes over time, not something current.

os-d:
> I stopped halfway reviewing this patch.
> 
> While I highly appreciate the effort you are making here to perform this major cleanup, there is really no reason whatsoever to do all of this in a single patch, and I dread the day n months from now where bisecting a heisenbug on AArch64 leads to this change, and we'll have to dig through all of the python, tools_def and code changes to narrow it down.
> 
> So please split this into:
> 
> * removal of CI plumbing
> * tools_def/BaseTools changes removing the ability to build for the ARM target
> * removal of ARM support from generic modules (at least 1 patch per module)
> * removal of ARM support from ArmPkg/ArmPlatformPkg (split into meaningful chunks)
> 
> If you don't have the bandwidth, we can try and find someone to share the load.

@ardbiesheuvel I have refactored this into 49 commits, hopefully more easily reviewable. I split up the larger packages by component and I attempted to order things so that if this does need to be bisected in the future, it can be done so easily, but with the tangled interdependencies between ArmPkg and the other packages this may not be true.

I did leave BaseTools as a single commit because it felt less stable to rip out build support halfway. I thought that would break more things. But I can try harder to split that commit up if desired, let me know.

BaseTools/Source/C/GenFv/GenFvInternalLib.c
```diff
@@ -3449,7 +3352,7 @@ Routine Description:
       }
 
       // machine type is ARM, set a flag so ARM reset vector processing occurs
-      if ((MachineType == IMAGE_FILE_MACHINE_ARMTHUMB_MIXED) || (MachineType == IMAGE_FILE_MACHINE_ARM64)) {
+      if (MachineType == IMAGE_FILE_MACHINE_ARM64) {
         VerboseMsg("Located ARM/AArch64 SEC/PEI core in child FV");
```
leiflindholm: Maybe drop ARM/ here?

BaseTools/Source/C/LzmaCompress/Sdk/DOC/lzma-sdk.txt
```diff
@@ -192,7 +192,7 @@ LZMA features
   - Estimated compressing speed: about 2 MB/s on 2 GHz CPU
   - Estimated decompressing speed: 
       - 20-30 MB/s on modern 2 GHz cpu
-      - 1-2 MB/s on 200 MHz simple RISC cpu: (ARM, MIPS, PowerPC)
```
leiflindholm, line 195: LzmaCompress is an imported external codebase - is it really worth stripping ARM out of it?

ShellPkg/Library/UefiShellAcpiViewCommandLib/AcpiView.c
```diff
@@ -237,9 +237,15 @@ AcpiView (
     }
   }
 
-  if (FoundAcpiTable) {
+  if (FoundAcpiTable && (EfiConfigurationTable != NULL)) {
```
leiflindholm: The existing function could be simplified by calling EfiGetSystemConfigurationTable instead of rewriting the whole functionality locally like it's 1992. Then we wouldn't need these shenanigans, or the NULLing of the pointer on 212.

After that we could get rid of the success handling that makes up the rest of the function.

UefiCpuPkg/Library/ArmMmuLib/ArmMmuBaseLib.inf
```diff
@@ -14,21 +14,12 @@
...
-[Defines.AARCH64]
   CONSTRUCTOR                    = ArmMmuBaseLibConstructor
 
-[Sources.AARCH64]
-  ArmMmuLibInternal.h
-  AArch64/ArmMmuLibCore.c
-  AArch64/ArmMmuLibReplaceEntry.S
-
-[Sources.ARM]
+[Sources]
```
leiflindholm: Meta-comment applying to a lot of changes in this PR:
Do we really want to move all [Sources.AARCH64] files out from AArch64/ directories and under a [Sources] section? They are still AArch64-specific.

To me, this adds problems and solves none:
- If someone wants to clone an existing AArch64 module to a new architecture, they will then need to manually go through and filter which files are arch-independent and which aren't.
- This applies also if AArch64 evolves enough that we at some point need a new set of bindings for an ABI-incompatible variant.

EmbeddedPkg/Drivers/DtPlatformDxe/DtPlatformDxe.inf
```diff
@@ -17,7 +17,7 @@
 #
 # The following information is for reference only and not required by the build tools.
 #
-#  VALID_ARCHITECTURES      = IA32 X64 ARM AARCH64
```
leiflindholm, line 20: Most of these are fundamentally wrong anyway. Instead of tweaking the guaranteed-to-never-be-up-to-date list, could we nuke them from everywhere that isn't explicitly architecture-specific?

ArmPlatformPkg/Sec/Sec.inf
```diff
@@ -15,21 +15,13 @@
   VERSION_STRING                 = 1.0
 
 [Sources.common]
+  ArchSec.c
```
leiflindholm: Commit message: this commit does not in fact delete any Pcds.

EmbeddedPkg/Drivers/DtPlatformDxe/DtPlatformDxe.inf
os-d: That's fine with me, but I don't have the bandwidth to go across the whole codebase doing that. I can eliminate all the ones in this PR that are not explicitly arch specific (and most of the arch specific ones in this PR don't have that section anyway, at least in ArmPkg). Or I can drop these changes entirely and defer that to a later PR that removes these out of date lists.

What do you prefer?

UefiCpuPkg/Library/ArmMmuLib/ArmMmuBaseLib.inf
os-d: I don't have a preference here. I can leave all of the Sources.AARCH64 and AArch64/ directories. We are mostly inconsistent, especially in ArmPkg, on where *.AARCH64 is used and it leads to a tangled web of what gets built. I prefer things to be much more straightforward: this is what gets built when you build this library.

That being said, when different architectures can use the same library that is the better case, so trying to future proof this by keeping things in the AARCH64 specific sources/dirs is fine with me. I'll make that change in the coming days unless there are any other preferences.

BaseTools/Source/C/LzmaCompress/Sdk/DOC/lzma-sdk.txt
os-d: No, I agree not. I was mixing what was edk2 on top of the external codebase. I'll drop these.

ArmPkg/ArmPkg.dsc
```diff
@@ -154,17 +154,13 @@
...
-
-[Components.AARCH64]
   ArmPkg/Drivers/ArmCrashDumpDxe/ArmCrashDumpDxe.inf
   ArmPkg/Drivers/ArmGicDxe/ArmGicV3Dxe.inf
   ArmPkg/Drivers/ArmPsciMpServicesDxe/ArmPsciMpServicesDxe.inf
   ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.inf
 
-[Components.AARCH64, Components.ARM]
   ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.inf
+
```
leiflindholm: This hunk ends up being moved out of
[Components.common.MM_CORE_STANALONE] (lol, typo)
[Components.common.MM_STANDALONE]

I know this is just the package .dsc, but it feels wrong.

EmbeddedPkg/Drivers/DtPlatformDxe/DtPlatformDxe.inf
leiflindholm: I'd be entirely onboard with either of those options. I'd _prefer_ eliminating the ones touched by this PR, but what I really don't want is the git history implying that they are well curated :)

ArmPkg/ArmPkg.dsc
os-d: This was intentional. `Components.common.module_type` is not valid, per DSC spec: 

image

https://tianocore-docs.github.io/edk2-DscSpecification/release-1.28/3_edk_ii_dsc_file_format/311_[components]_sections.html#311-components-sections

I'm not sure what the DSC parser actually does here, if it ignores it entirely or it says, I found Components.common, good enough, I'll build it (my guess is the latter). Hence why the original typo didn't matter.

I can make this a separate commit if you want, fixing up this error.

leiflindholm: Heh, explains why the typo didn't matter.
Well, I would have appreciated it as a separate commit from a reviewing standpoint, but given it's non-functional it doesn't really matter for the history perspective.

os-d: I'm making commit changes anyway, I'll split this out.

ShellPkg/Library/UefiShellAcpiViewCommandLib/AcpiView.c
pierregondois: As a side unrelated comment: unfortunately it also seems that most of the code of the ShellPkg is written following a similar pattern, i.e. long functions with a lot of variable that might be initialized (or not) in sub-cases of the function.
Ideally it would be nice to refactor these functions, but it would also mean a lot of modifications (which might also be error prone)

pierregondois:
Just 3 miss I think:

```
diff --git a/MdeModulePkg/MdeModulePkg.dec b/MdeModulePkg/MdeModulePkg.dec
index 0cb53f0a8a90..748b72de2eb3 100644
--- a/MdeModulePkg/MdeModulePkg.dec
+++ b/MdeModulePkg/MdeModulePkg.dec
@@ -985,7 +985,7 @@
   # @Prompt Enable process non-reset capsule image at runtime.
   gEfiMdeModulePkgTokenSpaceGuid.PcdSupportProcessCapsuleAtRuntime|FALSE|BOOLEAN|0x00010079
 
-[PcdsFeatureFlag.IA32, PcdsFeatureFlag.ARM, PcdsFeatureFlag.AARCH64, PcdsFeatureFlag.LOONGARCH64]
+[PcdsFeatureFlag.IA32, PcdsFeatureFlag.AARCH64, PcdsFeatureFlag.LOONGARCH64]
   gEfiMdeModulePkgTokenSpaceGuid.PcdPciDegradeResourceForOptionRom|FALSE|BOOLEAN|0x0001003a
 
 [PcdsFeatureFlag.IA32, PcdsFeatureFlag.X64]
diff --git a/MdePkg/Library/BaseMemoryLibOptDxe/AArch64/MemLibGuid.c b/MdePkg/Library/BaseMemoryLibOptDxe/AArch64/MemLibGuid.c
index 32d30cb7cecc..8e9d9211d515 100644
--- a/MdePkg/Library/BaseMemoryLibOptDxe/AArch64/MemLibGuid.c
+++ b/MdePkg/Library/BaseMemoryLibOptDxe/AArch64/MemLibGuid.c
@@ -1,5 +1,5 @@
 /** @file
-  Implementation of GUID functions for ARM and AARCH64
+  Implementation of GUID functions for AARCH64
 
   Copyright (c) 2006 - 2016, Intel Corporation. All rights reserved.<BR>
   Copyright (c) 2016, Linaro Ltd. All rights reserved.<BR>
diff --git a/StandaloneMmPkg/Core/StandaloneMmCore.inf b/StandaloneMmPkg/Core/StandaloneMmCore.inf
index 02d4e4616636..41231fa6d6d9 100644
--- a/StandaloneMmPkg/Core/StandaloneMmCore.inf
+++ b/StandaloneMmPkg/Core/StandaloneMmCore.inf
@@ -95,7 +95,7 @@
 #
 # This configuration fails for CLANGPDB, which does not support PIE in the GCC
 # sense. Such however is required for ARM family StandaloneMmCore
-# self-relocation, and thus the CLANGPDB toolchain is unsupported for ARM and
+# self-relocation, and thus the CLANGPDB toolchain is unsupported for
 # AARCH64 for this module.
 #
 [BuildOptions]
```

ShellPkg/Library/UefiShellAcpiViewCommandLib/AcpiView.c
leiflindholm: I mean, sometimes I feel like much of edk2 is written following a similar pattern.
If we had time, we absolutely should be refactoring it all. There is no situation in which it is better to leave code that confuses both humans and compilers.

However, we work with what limited resources we have. But _when_ we end up in a situation where a compiler requires us to fix broken code, we really should be fixing it, not playing "make problem go away" whack-a-mole, further convoluting the codebase.

os-d:
> Just 3 miss I think:
> 
> ```
> diff --git a/MdeModulePkg/MdeModulePkg.dec b/MdeModulePkg/MdeModulePkg.dec
> index 0cb53f0a8a90..748b72de2eb3 100644
> --- a/MdeModulePkg/MdeModulePkg.dec
> +++ b/MdeModulePkg/MdeModulePkg.dec
> @@ -985,7 +985,7 @@
>    # @Prompt Enable process non-reset capsule image at runtime.
>    gEfiMdeModulePkgTokenSpaceGuid.PcdSupportProcessCapsuleAtRuntime|FALSE|BOOLEAN|0x00010079
>  
> -[PcdsFeatureFlag.IA32, PcdsFeatureFlag.ARM, PcdsFeatureFlag.AARCH64, PcdsFeatureFlag.LOONGARCH64]
> +[PcdsFeatureFlag.IA32, PcdsFeatureFlag.AARCH64, PcdsFeatureFlag.LOONGARCH64]
>    gEfiMdeModulePkgTokenSpaceGuid.PcdPciDegradeResourceForOptionRom|FALSE|BOOLEAN|0x0001003a
>  
>  [PcdsFeatureFlag.IA32, PcdsFeatureFlag.X64]
> diff --git a/MdePkg/Library/BaseMemoryLibOptDxe/AArch64/MemLibGuid.c b/MdePkg/Library/BaseMemoryLibOptDxe/AArch64/MemLibGuid.c
> index 32d30cb7cecc..8e9d9211d515 100644
> --- a/MdePkg/Library/BaseMemoryLibOptDxe/AArch64/MemLibGuid.c
> +++ b/MdePkg/Library/BaseMemoryLibOptDxe/AArch64/MemLibGuid.c
> @@ -1,5 +1,5 @@
>  /** @file
> -  Implementation of GUID functions for ARM and AARCH64
> +  Implementation of GUID functions for AARCH64
>  
>    Copyright (c) 2006 - 2016, Intel Corporation. All rights reserved.<BR>
>    Copyright (c) 2016, Linaro Ltd. All rights reserved.<BR>
> diff --git a/StandaloneMmPkg/Core/StandaloneMmCore.inf b/StandaloneMmPkg/Core/StandaloneMmCore.inf
> index 02d4e4616636..41231fa6d6d9 100644
> --- a/StandaloneMmPkg/Core/StandaloneMmCore.inf
> +++ b/StandaloneMmPkg/Core/StandaloneMmCore.inf
> @@ -95,7 +95,7 @@
>  #
>  # This configuration fails for CLANGPDB, which does not support PIE in the GCC
>  # sense. Such however is required for ARM family StandaloneMmCore
> -# self-relocation, and thus the CLANGPDB toolchain is unsupported for ARM and
> +# self-relocation, and thus the CLANGPDB toolchain is unsupported for
>  # AARCH64 for this module.
>  #
>  [BuildOptions]
> ```

Thanks! I'll fix those as well.

ShellPkg/Library/UefiShellAcpiViewCommandLib/AcpiView.c
os-d: I have rewritten this to use the proper function.

UefiCpuPkg/Library/ArmMmuLib/ArmMmuBaseLib.inf
os-d: I have moved all of these back.

EmbeddedPkg/Drivers/DtPlatformDxe/DtPlatformDxe.inf
os-d: I eliminated the ones touched by this PR, except for where there was arch specific code (as evidenced by Sources.<arch> sections in the INF). Feel free to ask me to update any of these (or don't update). I made the choice to drop ARM from those but not touch the other archs (except in one case where ARM was the only one listed).

os-d:
@ardbiesheuvel @leiflindholm @pierregondois I believe I have addressed all the current feedback and I have addressed the two minor merge conflicts. Please re-review when you have a chance.

ShellPkg/Library/UefiShellAcpiViewCommandLib/AcpiView.c
```diff
@@ -223,23 +220,12 @@ AcpiView (
...
-  }
-
-  if (FoundAcpiTable) {
-    RsdpPtr = (UINT8 *)EfiConfigurationTable->VendorTable;
+  Status = EfiGetSystemConfigurationTable (
+             &gEfiAcpiTableGuid,
+             (VOID **)&RsdpPtr
+             );
 
+  if (!EFI_ERROR (Status) && (RsdpPtr != NULL)) {
```
leiflindholm: If `Status` is not an error, the pointer is guaranteed to be NULL. So the NULL-init above, and the NULL-check hera are redundant.

ShellPkg/Library/UefiShellAcpiViewCommandLib/AcpiView.c
```diff
@@ -209,7 +206,7 @@ AcpiView (
   //
   // set local variables to suppress incorrect compiler/analyzer warnings
   //
-  EfiConfigurationTable = NULL;
+  RsdpPtr              = NULL;
```
leiflindholm: More detail below on why this is not needed.
But additionally, OriginalAttribute, zeroed on the next line, is only used inside the block beginning on original line 304; where it is explicitly initialized.

Could we drop this whole compiler-whack-a-mole block? (And if needed move the OriginalAttribute definition to the block where it is used?)

ShellPkg/Library/UefiShellAcpiViewCommandLib/AcpiView.c
leiflindholm: We could also get rid of the success handling and return here, reducing the maximum conditional nesting in the function to 3 levels instead of 4.
I realise this is turning into cleanup-creep, but I think this is still justified. 

(I'm stopping short of suggesting the breaking up into helper functions this function sorely needs.)

os-d: Agreed. I didn't look at InstallConfigurationTable to check that it guarantees VendorTable is non-null (it does). I can flip the error checking here to remove the conditional nesting.

ShellPkg/Library/UefiShellAcpiViewCommandLib/AcpiView.c
os-d: Updated.

os-d:
@leiflindholm I've updated the AcpiView change and fixed the CI error (uncrustify in that file). Please re-review when you have a chance.

leiflindholm:
> @leiflindholm I've updated the AcpiView change and fixed the CI error (uncrustify in that file). Please re-review when you have a chance.

Done. @ardbiesheuvel @samimujawar @vishalo, any comments?

Javagedes:
approved for commit [.azurepipelines,.github,.pytool: Disable ARM32 at Top Level/Pipelines](https://github.com/tianocore/edk2/pull/11494/commits/75211f53810b5a30f0212e0b6ee35083ed75960a)

pierregondois (approved):
Levi's comment seems correct. Apart from that the PR looks ok to me

DynamicTablesPkg/Drivers/DynamicTableManagerDxe/DynamicTableManagerDxe.inf
```diff
@@ -18,14 +18,14 @@
 #
 # The following information is for reference only and not required by the build tools.
 #
-#  VALID_ARCHITECTURES           = ARM AARCH64
+#  VALID_ARCHITECTURES           = IA32 X64 AARCH64
```
samimujawar: Minor: There is another PR that was merged recently that drops IA32 support from DynamicTablesPkg, see https://github.com/tianocore/edk2/commit/32711df057bdaa34018e31ee9a5eb57431e88eef

os-d: I have fixed that, thanks.

os-d:
> I think it would be okay to remove 32 bit check for xferlist: -https://github.com/tianocore/edk2/blob/master/ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c#L189

Thanks, I have changed this to return EFI_INVALID_PARAMETER if the transfer list is not 64 bit.

PR #11501, opened 2025-09-15
UefiCpuPkg/MpInitLib: backoff on timed waits

nxfs:
This commit addresses an issue where busy wait loops in TimedWaitForApFinish can starve application processor siblings when hypervisors don't allow hyperthread siblings to run concurrently during hypervisor traps (e.g., L1TF reloaded mitigation).

### DIFF
--- a/UefiCpuPkg/Library/MpInitLib/MpLib.c
+++ b/UefiCpuPkg/Library/MpInitLib/MpLib.c
@@ -1643,14 +1643,30 @@ TimedWaitForApFinish (
                               TimeLimit,
                               &CpuMpData->CurrentTime
                               );
-  while (CpuMpData->FinishedCount < FinishedApLimit &&
-         !CheckTimeout (
-            &CpuMpData->CurrentTime,
-            &CpuMpData->TotalTime,
-            CpuMpData->ExpectedTime
-            ))
   {
-    CpuPause ();
+    UINTN  LoopCount;
+    UINTN  PauseCount;
+    UINTN  Index;
+
+    LoopCount  = 0;
+    PauseCount = 1;
+
+    while (CpuMpData->FinishedCount < FinishedApLimit &&
+           !CheckTimeout (
+              &CpuMpData->CurrentTime,
+              &CpuMpData->TotalTime,
+              CpuMpData->ExpectedTime
+              ))
+    {
+      for (Index = 0; Index < PauseCount; Index++) {
+        CpuPause ();
+      }
+
+      LoopCount++;
+      if ((LoopCount % 8) == 0 && PauseCount < 1024) {
+        PauseCount *= 2;
+      }
+    }
   }
 
   if (CpuMpData->FinishedCount >= FinishedApLimit) {


niruiyu:
My understanding is PAUSE instruction does not cause a VM exit. right?

nxfs:
> My understanding is PAUSE instruction does not cause a VM exit. right?

Correct, maybe the commit message is imprecise. What causes a VM exit is the timeout check in the loop performing an IO read to the ACPI PM timer.

niruiyu:
> > My understanding is PAUSE instruction does not cause a VM exit. right?
> 
> Correct, maybe the commit message is imprecise. What causes a VM exit is the timeout check in the loop performing an IO read to the ACPI PM timer.

If you use TimerLib that reads MSR and configure the MSR not to cause VM exit, will that solve your issue?

nxfs:
> > > My understanding is PAUSE instruction does not cause a VM exit. right?
> > 
> > 
> > Correct, maybe the commit message is imprecise. What causes a VM exit is the timeout check in the loop performing an IO read to the ACPI PM timer.
> 
> If you use TimerLib that reads MSR and configure the MSR not to cause VM exit, will that solve your issue?

In https://github.com/tianocore/edk2/pull/11489/ I proposed to use TSC when available to avoid VM exit, are you referring to an approach like this?

I think both approaches have merits, the backoff has the advantage of not relying on specific hypervisor pass through capabilities, which may be impossible to introduce for backwards compatibility reasons

kraxel:
> I think both approaches have merits, the backoff has the advantage of not relying on specific hypervisor pass through capabilities, which may be impossible to introduce for backwards compatibility reasons

Yes, I think it makes sense to have both.

ardbiesheuvel (approved):

nxfs:
Accidentally closed (and reopened) the PR instead of rebasing. Fixed now.

nxfs:
Only failing check seems to be transient CI failure.
```
ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host
```

UefiCpuPkg/Library/MpInitLib/MpLib.c
```diff
@@ -1621,14 +1621,30 @@ TimedWaitForApFinish (
...
+    {
+      for (Index = 0; Index < PauseCount; Index++) {
+        CpuPause ();
+      }
+
+      LoopCount++;
+      if ((LoopCount % 8) == 0) {
+        PauseCount *= 2;
+      }
+    }
```
niruiyu, line 1669: PauseCount, in an extreme case, will be a very large value.
How about limit PauseCount to a max value such as 1024?

niruiyu: @nxfs

nxfs: Ok, done @niruiyu

ardbiesheuvel:
Please don't merge back the master branch - rebase your branch instead.

nxfs:
> Please don't merge back the master branch - rebase your branch instead.

Woops, sorry. Done.

UefiCpuPkg/Library/MpInitLib/MpLib.c
```diff
@@ -1643,14 +1643,30 @@ TimedWaitForApFinish (
...
                               );
-  while (CpuMpData->FinishedCount < FinishedApLimit &&
-         !CheckTimeout (
-            &CpuMpData->CurrentTime,
-            &CpuMpData->TotalTime,
-            CpuMpData->ExpectedTime
-            ))
   {
-    CpuPause ();
+    UINTN  LoopCount;
```
ardbiesheuvel, line 1647: Please move these declarations to the start of the function, and drop the redundant enclosing {}

PR #11503, opened 2025-09-16
MdeModulePkg/DxeCapsuleLibFmp: Added PCD definition to control support for EmbeddedDriver in capsule updates.

PavithraG23:
# Description

There is no mechanism to enable or disable support for EmbeddedDrivers in capsule updates. So defined, a new PCD PcdEmbeddedDriverSupport allowing platforms to explicitly control whether capsules containing EmbeddedDrivers are supported.

- [ ] Breaking change?

- [ ] Impacts security?

- [ ] Includes tests?

## How This Was Tested

Validated System Firmware Update with a capsule which doesnot include embedded drivers. 

## Integration Instructions

  N/A


mdkinney:
Why would a platform optionally support embedded drivers in a capsule?  If a capsule contains embedded drivers then the capsule can not be processed unless those embedded drivers are loaded and executed otherwise the capsule would not include any embedded drivers.  This makes the capsule unusable.

PavithraG23:
> Why would a platform optionally support embedded drivers in a capsule? If a capsule contains embedded drivers then the capsule can not be processed unless those embedded drivers are loaded and executed otherwise the capsule would not include any embedded drivers. This makes the capsule unusable.

This check is essential for platforms like Qualcomm that do not support embedded drivers in capsules. It ensures that if a capsule containing embedded drivers is received, and the platform lacks support for executing those drivers, the capsule is explicitly rejected. Without this validation, the platform might proceed with processing an unsupported capsule, leading to undefined behavior or failure during the firmware update. This safeguard prevents such scenarios by enforcing compatibility upfront.

vishalo:
@mdkinney - This change aligns with Qualcomm’s current implementation, which rejects capsules containing embedded drivers. 

Would it be appropriate to add a DEBUG_INFO message in this path - logging the rejection reason (e.g., “Capsule contains embedded drivers, rejecting per platform policy”)?

spbrogan:
There was a discussion about this long ago (when FMP and all the supporting code was first introduced).  
My position has always been that supporting EmbeddedDrivers as implemented in edk2 is not secure and invalidates the security of the platform.  The reason being is the embedded drivers are verified against UEFI SB and that is a different authority than platform security (PM Auth).  When most capsules run they need to be executing with PM auth (write system flash, etc).  UEFI secure boot could be disabled by a user and then any capsule could contain any UEFI code and it would be dispatched.  There are possibilities that platform protections could be in place, but I don't see that in the OSS code and think it is dangerous.  

Project Mu has carried "overrides" for this library forever.  Unfortunately, the change has been squashed with another few capsule feature changes but you can see it here (Line 259 of DxeCapsuleLib.c).  https://github.com/microsoft/mu_basecore/commit/48d694149055c9e5774ab4a95707ff9f0c659a62#diff-cea14f6f2ac424721e02fc285d6da696d786acf464bb0c5c982271012347a7f2 

It is slightly different than what is proposed here so more analysis would need to be done to ensure a malformed/unexpected capsule could not introduce parsing errors or create invalid execution state.  

If a platform needs/wants to use EmbeddedDrivers then I think more work needs to be done in this code to facilitate secure evaluation of the drivers.

PavithraG23:
> It is slightly different than what is proposed here so more analysis would need to be done to ensure a malformed/unexpected capsule could not introduce parsing errors or create invalid execution state.
> 
> If a platform needs/wants to use EmbeddedDrivers then I think more work needs to be done in this code to facilitate secure evaluation of the drivers.

Thanks for the feedback. 
The proposed change adds a check for EmbeddedDriverCount within ValidateFmpCapsule(), which validates the capsule header before any further processing. ValidateFmpCapsule() is invoked by ProcessTheseCapsules() prior to calling ProcessThisCapsuleImage(). If ValidateFmpCapsule() fails, the capsule is not applied and the status for the respective caspule is set to EFI_ABORTED. https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleProcessLib.c#L579

MuChange: (Line 259 of DxeCapsuleLib.c). https://github.com/microsoft/mu_basecore/commit/48d694149055c9e5774ab4a95707ff9f0c659a62#diff-cea14f6f2ac424721e02fc285d6da696d786acf464bb0c5c982271012347a7f2
In addition, MuChange also includes the check for EmbeddedDriverCount within ValidateFmpCapsule().
In the latest patch, the check has been moved to line 248, where it uses FmpCapsuleHeader->EmbeddedDriverCount instead of the optional EmbeddedDriverCount parameter passed to ValidateFmpCapsule().
https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleLib.c#L248

MdeModulePkg/MdeModulePkg.dec
```diff
@@ -1696,6 +1697,14 @@
   # @Prompt Enable Capsule On Disk support.
   gEfiMdeModulePkgTokenSpaceGuid.PcdCapsuleOnDiskSupport|FALSE|BOOLEAN|0x0000002d
 
+  ## Indicates if the platform can support drivers within the capsule.
+  #  This PCD indicates if the platform can process Capsules that include EmbeddedDrivers.<BR>
+  #   TRUE  - Capsule with EmbeddedDriver is supported.<BR>
+  #   FALSE - Capsule with EmbeddedDriver is not supported.<BR>
+  #  If platform does not use this feature, this PCD should be set to FALSE.<BR><BR>
+  # @Prompt Enable Capsule with EmbeddedDriver support.
```
mdkinney, line 1705: Suggest making the default value FALSE, so embedded drivers are not supported by default and a platform must opt-in to support embedded drivers in capsules.  

With this default setting, this PR must be marked as a "Breaking Change" and the description of PR should add the information on how a platform can opt-in to the use of embedded drivers by setting the PCD to TRUE with an example DSC files PCD statement

MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleLib.c
```diff
@@ -246,6 +247,11 @@ ValidateFmpCapsule (
     return EFI_INVALID_PARAMETER;
   }
 
+  if (!PcdGetBool (PcdCapsuleEmbeddedDriverSupport) && (FmpCapsuleHeader->EmbeddedDriverCount != 0)) {
+    DEBUG ((DEBUG_ERROR, "FmpCapsuleHeader->EmbeddedDriverCount(0x%x) != 0. Fmp Capsule with EmbeddedDriver is not supported by this platform\n", FmpCapsuleHeader->EmbeddedDriverCount));
```
mdkinney: Do we need to print the EmbeddedDriverCount in this message?

Perhaps the following message is sufficient:

"FMP Capsule with one or more embedded drivers is not supported by this platform".

MdeModulePkg/MdeModulePkg.dec
```diff
@@ -1696,6 +1697,14 @@
   # @Prompt Enable Capsule On Disk support.
   gEfiMdeModulePkgTokenSpaceGuid.PcdCapsuleOnDiskSupport|FALSE|BOOLEAN|0x0000002d
 
+  ## Indicates if the platform can support drivers within the capsule.
+  #  This PCD indicates if the platform can process Capsules that include EmbeddedDrivers.<BR>
+  #   TRUE  - Capsule with EmbeddedDriver is supported.<BR>
+  #   FALSE - Capsule with EmbeddedDriver is not supported.<BR>
+  #  If platform does not use this feature, this PCD should be set to FALSE.<BR><BR>
```
mdkinney: Please provide a more detailed description.  States that the default value is FALSE which means embedded drivers are not supported, and that a platform that requires the use of embedded drivers must opt-in to that feature by setting this PCD to TRUE.

mdkinney:
One additional change,  The commit message should also describe that this is a breaking change and the default setting is to disable embedded driver support in capsules for the entire platform and that platforms have to opt-in to support embedded drivers.

This is important so platforms that sync with edk2 upstream can know that this is a breaking change from reviewing commit messages.

mdkinney:
@mergify rebase

mdkinney:
@PavithraG23 Please update the PR to allow rebase by other agents

@PavithraG23 needs to [authorize modification on its head branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).

PavithraG23:
> @PavithraG23 Please update the PR to allow rebase by other agents
> 
> @PavithraG23 needs to [authorize modification on its head branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).

I updated the PR to allow the rebase. Please re-trigger Mergify to update the base branch of the pull request.

mdkinney:
@mergify rebase

PR #11507, opened 2025-09-16
ArmPkg: support firmware update feature 

LeviYeoReum:
# Introduction

The Firmware Update Feature is based on
[Platform Security Firmware Update for the A-profile Specification 1.0][1] specification.
To update firmware, Firmware Update Feature uses FmpDevicePkg framework
for the firmware to be updated via capsule update framework.

Updates firmware with Capsule update framework in Arm with following steps:
  - Deliver firmware image via UpdateCapsule().
  - firmware image delivered to StandaloneMm via MmCommunication or FF-A.
  - StandaloneMm which is UpdateAgent write new image in firmware update storage
    according to PSA specification.
  - To apply updated firmware, Reset.

This is slight different from other architecture which using
coalescing update firmware with following steps:
  - Deliver the firmware image via UpdateCapsule().
  - Save the firmware image in the variable storage.
  - Warm Reset, and In PEI phase coalesce the firmware image scattered in physical memory.
  - Before EndofDxe, update the firmware by calling ProcessCapsules()
  - To apply the updated firmware, Reset.
for preventing arbitrary access to firmware storage device locked after EndofDxe phase.

It's the reason Arm doesn't supports coalescing way because
    - According to platform UEFI doesn't run in ROM but
      it loaded to memory by TF-A
    - According to platform, it can skip PEICORE (See EDK2_SKIP_PEICORE)
    - Arm doesn't need to lock the firmware storage device because
      it's completely isolated in StandaloneMm (at S-EL0).
      Therefore, operating system, uefi or any other software components running in
      normal world cannot access isolated firmware storage.

By doing so, it can remove WarmReset for unlocking device and support runtime
firmware update in the future.

This implementation is written for platforms where firmware storage's layout
(typical platform is Base FVP platform):
```
   +----------------------+
    |      GPT-HEADER      |
    +----------------------+
    |    FIP_A (bank0)     |
    +----------------------+
    |    FIP_B (bank1)     |
    +----------------------+
    |    FWU-Metadata      |
    +----------------------+
    |  Bkup-FWU-Metadata   |
    +----------------------+
```
and uses FwsGptSystemFipLib used to access above firmware storage.

# Overview
Here is an overview of Firmware Update Feature.

```
       UEFI (Normal world)            |          StandaloneMm (Secure world)
     ---------------------------------|--------------------------------------
                                      |                             +-------+
                                      |                    ---------|  Fws  |
                                      |                    |        +-------+
     +------------------+             |                    |       (Gpt parted)
     |   FmpDevicePkg   |             |  Read /Write Image |
     +------------------+             |                    |
       |                              |          +-------------------+
       |  SetTheImage() and etc       |          |  FwsPlatformLib   |
       |  progress via PsaFwuLib      |          +-------------------+
       |                              |                    |
       |                              |    Parsing Request | Access Fws via
       |                              |                    |    FwStore.c
       |                                                   |
       -> +----------------+     PSA ABI (MMC)   +-------------------+
          |   PsaFwuLib    |<------------------> |     FwuStMm       |
          +----------------+    PSA Error code   +-------------------+
```

When UEFI calls UpdateCapsule(), FmpDevicePkg->SetTheImage() is called.
Through FmpDeviceLib, FmpDevicePkg requests a firmware update to StandaloneMm
according to Firmware Store Update ABI defined in [PSA][1] spec via PsaFwuLib.
Then FwuStMm StandaloneMm driver parses requests from PsaFwuLib and access to firmware
storage via FwsPlatformLib which is platform specific library.

See together:
    - https://github.com/tianocore/edk2-platforms/pull/873

# Patch Sequence

Patch #1 adds partition helper macro for FF-A
Patch #2 - #4 adds MmiContext used for arm platform
Patch #5 - #6 adds firmware update feature related header and GUID.
Patch #7 adds FmpDeviceLib used with firmware update ABI.


leiflindholm:
The documentation link does not point to what it claims to, it points to the FVP marketing overview page.

leiflindholm:
> The documentation link does not point to what it claims to, it points to the FVP marketing overview page.

I can find DEN011 through internet search, but not when looking through the armdeveloper documentation library.

LeviYeoReum:
> > The documentation link does not point to what it claims to, it points to the FVP marketing overview page.
> I can find DEN011 through internet search, but not when looking through the armdeveloper documentation library.

Oh sorry. I'll update the reference correctly. Thanks

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
```diff
@@ -0,0 +1,522 @@
...
+      Table A3.2: Metadata version 2
+ */
+#define FWU_METADATA_V2_MAX_NUM_BANKS  4
+
+/**  Enum value defining op_type used in fwu_open.
+     Cf Platform Security Firmware Update for the A-profile specification,
+     1.0, 3.4.2.5 "fwu_open"
+ */
+typedef enum {
+  FWU_STREAM_READ,
```
samimujawar: FWU_OP_STREAM_READ/WRITE?
In fact enums members should not be in all capital letters. 
typedef enum {
  FwuOpStreamRead,
  FwuOpStreamWrite,
  FwuOpStreamMax
} FWU_OP_TYPE;


See Section 5.6.2.2 Enumerated Types at https://tianocore-docs.github.io/edk2-CCodingStandardsSpecification/release-2.20/5_source_files/56_declarations_and_types.html#table-8-efi-constants

ArmPkg/ArmPkg.ci.yaml
```diff
@@ -47,6 +47,7 @@
             "ArmPlatformPkg/ArmPlatformPkg.dec",
             "ArmPkg/ArmPkg.dec",
             "EmbeddedPkg/EmbeddedPkg.dec",
+            "FmpDevicePkg/FmpDevicePkg.dec",
```
samimujawar, line 50: Thank you for the ASCII diagram. It really helps to understand this patch series.

In the commit message the "FwuSmm.c" block text should be updated to "FwuStMm", to reflect diagram in the Overview section at https://github.com/tianocore/edk2/pull/11507

samimujawar: Similarly the text FmpPsaFwuLib should be changed to FmpDevicePsaFwuLib. Also would it be clearer to indicate that FmpDevicePsaFwuLib is an instance of the FmpDeviceLib? Would it be better to update the ASCII diagram as shown below?

```
    +-------------------------+
    |     FmpDevicePsaFwuLib  |
    |        (FmpDeviceLib)   |
    +-------------------------+

```

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
```diff
@@ -0,0 +1,1251 @@
...
+      DEBUG ((
+        DEBUG_ERROR,
+        "FmpPsaFwuLib: FwuOpen failed for Image Directory: %r\n",
+        Status
+        ));
+      return Status;
+    }
+  }
+
+  ASSERT (mImageDirectory != NULL);
```
samimujawar: I think this assert can be dropped.

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
```diff
@@ -0,0 +1,1251 @@
...
+        Status
+        ));
+      goto ErrorHandler;
+    }
+
+    CloseStatus = FwuCommit (Handle, 1, 0, &Progress, &TotalWork);
+    ASSERT (CloseStatus == EFI_SUCCESS);
+
+    mImageDirectorySize = TotalBytes;
+    mImageDirectory     = AllocateRuntimePool (mImageDirectorySize);
```
samimujawar: Can you clarify i.e. add a comment as to why Runtime Pool allocation is used here, please?

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
```diff
@@ -0,0 +1,1251 @@
...
+VOID
+EFIAPI
+FmpPsaVirtualAddressChangeEvent (
+  IN EFI_EVENT  Event,
+  IN VOID       *Context
+  )
+{
+  mOnRuntime = TRUE;
+  PsaFwuVirtualAddressChangeEvent (Event, Context);
+  gRT->ConvertPointer (0x00, (VOID **)&mImageDirectory);
```
samimujawar, line 65: Do you need to do ConvertPointer for mImageEntry? Can you check, please?

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
```diff
@@ -0,0 +1,1251 @@
...
+    IMAGE_ATTRIBUTE_AUTHENTICATION_REQUIRED
+    IMAGE_ATTRIBUTE_IN_USE
+    IMAGE_ATTRIBUTE_UEFI_IMAGE
+
+  @param[out] Supported  Attributes supported by this firmware device.
+  @param[out] Setting    Attributes settings for this firmware device.
+
+  @retval EFI_SUCCESS            The attributes supported by the firmware
+                                 device were returned.
+  @retval EFI_INVALID_PARAMETER  Supported is NULL.
```
samimujawar: @retval EFI_INVALID_PARAMETER    Parameters Settings or Supported is NULL.

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c
```diff
@@ -452,6 +452,30 @@ GetServiceType (
   return ServiceTypeMisc;
 }
 
+/**
```
samimujawar, line 455: This function and the corresponding change at line 512 must be moved to a separate patch.

ArmPkg/Drivers/StandaloneMmCpu/EventHandle.c
```diff
@@ -39,7 +39,7 @@ MmFoundationEntryRegister (
 // maintained in single global variable because StandaloneMm is UP-migratable
 // (which means it cannot run concurrently)
 //
-EFI_MM_COMMUNICATE_HEADER            *gGuidedEventContext = NULL;
+STATIC EFI_MM_COMMUNICATE_HEADER     *mGuidedEventContext = NULL;
```
samimujawar, line 42: Maybe the commit message can be reworded as below:

gGuidedEventContext is only used in EventHandle.c.
Therefore, make it STATIC, rename to mGuidedEventContext,
and remove the extern declaration.

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
```diff
@@ -0,0 +1,522 @@
...
+#define PSA_MM_FWU_COMMAND_WRITE_STREAM     20
+#define PSA_MM_FWU_COMMAND_READ_STREAM      21
+#define PSA_MM_FWU_COMMAND_COMMIT           22
+#define PSA_MM_FWU_COMMAND_ACCEPT_IMAGE     23
+#define PSA_MM_FWU_COMMAND_SELECT_PREVIOUS  24
+#define PSA_MM_FWU_COMMAND_MAX_ID           25
+
+/** Firmware Update Protocol error codes.
+    Cf Platform Security Firmware Update for the A-profile specification,
+    1.0, 3.4 "Firmware Store Update ABI"
```
samimujawar: nit: update section reference to '3.4.2.11 Return Status'

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
```diff
@@ -0,0 +1,522 @@
...
+#define PSA_MM_FWU_SUCCESS        0
+#define PSA_MM_FWU_UNKNOWN        (-1)
+#define PSA_MM_FWU_BUSY           (-2)
+#define PSA_MM_FWU_OUT_OF_BOUNDS  (-3)
+#define PSA_MM_FWU_AUTH_FAIL      (-4)
+#define PSA_MM_FWU_NO_PERMISSION  (-5)
+#define PSA_MM_FWU_DENIED         (-6)
+#define PSA_MM_FWU_RESUME         (-7)
+#define PSA_MM_FWU_NOT_AVAILABLE  (-8)
+#define PSA_MM_FWU_UA_ERROR       (-9)
```
samimujawar: FWU_UA_ERROR does not seem to be present in the spec. Can you point me to the relevant section in the spec, please?

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
```diff
@@ -0,0 +1,522 @@
...
+#define FWU_WRITE_PERMISSION         BIT0
+
+#define FWU_INVALID_HANDLE  0xffffffffU
+
+/**  Guid to open Firmware Image Directory via fwu_open
+      Cf Platform Security Firmware Update for the A-profile specification,
+      1.0 A3.5 or A3.8 Image properties in given FW bank version 1 & 2.
+ */
+#define FWU_IMAGE_UNACCEPTED  0
+#define FWU_IMAGE_ACCEPTED    1
```
samimujawar: #define FWU_IMAGE_ACCEPTED    BIT0

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
```diff
@@ -0,0 +1,522 @@
...
+ */
+#define FWU_ACCEPT_AFTER_ACTIVATION  BIT2
+#define FWU_READ_PERMISSION          BIT1
+#define FWU_WRITE_PERMISSION         BIT0
+
+#define FWU_INVALID_HANDLE  0xffffffffU
+
+/**  Guid to open Firmware Image Directory via fwu_open
+      Cf Platform Security Firmware Update for the A-profile specification,
+      1.0 A3.5 or A3.8 Image properties in given FW bank version 1 & 2.
```
samimujawar: Nit: would be good to add 'Table' before A3.5

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
```diff
@@ -0,0 +1,522 @@
...
+  /// Vendor specific staging flags.
+  UINT32    VendorFlags;
+
+  /// The number of elements in the update_guid array.
+  UINT32    PartialUpdateCount;
+
+  /// An array of image type GUIDs will be updated.
+  UINT8     UpdateGuid[];
+} PSA_MM_FWU_BEGIN_STAGING_REQ;
+
```
samimujawar, line 190: Might be worth adding a typedef for 3.4.2.3 fwu_end_staging
/** 
Documentation header
*/
typedef PSA_MM_FWU_PARAMETER_HEADER PSA_MM_FWU_END_STAGING_REQ;

samimujawar: Similarly add a typedef for cancel staging request, see  section '3.4.2.4 fwu_cancel_staging'

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
```diff
@@ -0,0 +1,522 @@
...
+  UINT32    Reserved;
+
+  /// Vendor specific staging flags.
+  UINT32    VendorFlags;
+
+  /// The number of elements in the update_guid array.
+  UINT32    PartialUpdateCount;
+
+  /// An array of image type GUIDs will be updated.
+  UINT8     UpdateGuid[];
```
samimujawar: Should this be EFI_GUID instead of UINT8 ?

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
```diff
@@ -0,0 +1,522 @@
...
+   */
+  UINT32      Accepted;
+
+  /// reserved (MBZ)
+  UINT32      Reserved;
+} PSA_MM_FWU_IMAGE_PROPERTIES;
+
+/** Metadata image entry version 1.
+    Cf Platform Security Firmware Update for the A-profile specification,
+    1.0, A.3.2.3 "Metadata version 1"
```
samimujawar: Is metadata version 1 going to be supported? If not can we not include the definitions, please?

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
```diff
@@ -0,0 +1,522 @@
...
+   Cf Platform Security Firmware Update for the A-profile specification,
+   1.0, A.3.2.3 "Metadata version 1"
+ */
+typedef struct {
+  /// Metadata Common Header.
+  PSA_MM_FWU_METADATA_COMMON_HEADER    Header;
+
+  /// Image entry information
+  PSA_MM_FWU_IMAGE_ENTRY_V1            ImageEntry[];
+} PSA_MM_FWU_METADATA_V1;
```
samimujawar: Can we drop the Metadata V1 structures as we are not going to be supporting this, please?

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
```diff
@@ -0,0 +1,1254 @@
...
+#include <Library/MemoryAllocationLib.h>
+#include <Library/UefiBootServicesTableLib.h>
+#include <Library/UefiRuntimeServicesTableLib.h>
+#include <Library/FmpDeviceLib.h>
+
+#include <Guid/SystemResourceTable.h>
+
+#include "PsaFwuLib.h"
+
+STATIC BOOLEAN                     mSupportFunction[PSA_MM_FWU_COMMAND_MAX_ID];
```
samimujawar: I think it would be better to use a bitmap to represent the supported functions, with helper function to set and test if a function is supported.

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.inf
```diff
@@ -0,0 +1,68 @@
...
+  PsaFwuLib.h
+
+[Packages]
+  MdePkg/MdePkg.dec
+  FmpDevicePkg/FmpDevicePkg.dec
+  ArmPkg/ArmPkg.dec
+
+[LibraryClasses]
+  ArmFfaLib
+  DebugLib
```
samimujawar: Alphabetic order, please

ArmPkg/Library/FmpDevicePsaFwuLib/PsaFwuLib.c
```diff
@@ -0,0 +1,1163 @@
...
+  EFI_FFA_PART_INFO_DESC  *PartInfo;
+  EFI_GUID                *ServiceGuid;
+  VOID                    *RxBuffer;
+  UINT64                  RxBufferSize;
+  UINT32                  Count;
+  UINT32                  Size;
+  UINT16                  PartId;
+
+  if (!IsFfaSupported ()) {
+    *FwuReqMode = FWU_REQ_MODE_MM_PROTOCOL;
```
samimujawar: The macro FWU_REQ_MODE_MM_PROTOCOL is misleading. In this case this should have been called FWU_REQ_MODE_SPM_MM to indicate that the transport is SPM-MM and the MM COMMUNICATE mechanism is being used.

This means there are 3 possible modes:
1. SPM-MM with MM_COMMUNICATE
2. FF-A 1.1 with Direct Req1 and MM_COMMUNICATE
3. FFA >= 1.2 with Direct Req2

Since we plan to drop the support for SPM-MM in 2026Q1, I think we should remove any  support for SPM-MM.
Since we do not support FF-A 1.1 we can further drop the MM_COMMUINICATE support.

This means that this function should check if FF-A version is >= 1.2 and set the transport to FWU_REQ_MODE_FFA_DIRECT_REQ2. For anything else it should return EFI_UNSUPPORTED.

Ref: Section 3.4.1 Transport layer, Platform Security Firmware Update for the A-profile Arm Architecture, DEN0118, EAC1, 1.0 A

ArmPkg/Library/FmpDevicePsaFwuLib/PsaFwuLib.c
```diff
@@ -0,0 +1,1163 @@
...
+  Status = ArmFfaLibGetRxTxBuffers (
+             NULL,
+             NULL,
+             &RxBuffer,
+             &RxBufferSize
+             );
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a: Failed to get Tx/Rx Buffer. Status:%r\n",
```
samimujawar: "%a: Failed to get Rx Buffer. Status:%r\n"

ArmPkg/Library/FmpDevicePsaFwuLib/PsaFwuLib.c
```diff
@@ -0,0 +1,1163 @@
...
+      return EFI_NOT_FOUND;
+    case PSA_MM_FWU_BUSY:
+      return EFI_NOT_READY;
+    case PSA_MM_FWU_OUT_OF_BOUNDS:
+      return EFI_OUT_OF_RESOURCES;
+    case PSA_MM_FWU_AUTH_FAIL:
+      return EFI_SECURITY_VIOLATION;
+    case PSA_MM_FWU_NO_PERMISSION:
+      return EFI_ACCESS_DENIED;
+    case PSA_MM_FWU_DENIED:
```
samimujawar, line 265: I think PSA_MM_FWU_NO_PERMISSION and PSA_MM_FWU_DENIED should return EFI_ACCESS_DENIED

ArmPkg/Library/FmpDevicePsaFwuLib/PsaFwuLib.c
```diff
@@ -0,0 +1,1163 @@
...
+FwuBeginStaging (
+  IN CONST EFI_GUID  *UpdateGuids,
+  IN UINT32          UpdateCount,
+  IN UINT32          VendorFlags
+  )
+{
+  EFI_STATUS                    Status;
+  PSA_MM_FWU_BEGIN_STAGING_REQ  ReqData;
+
+  if ((UpdateCount < 0) || ((UpdateCount > 0) && (UpdateGuids == NULL))) {
```
samimujawar: Should the check be (UpdateCount == 0) instead of (UpdateCount < 0), Can you recheck this condition, please?

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
```diff
@@ -0,0 +1,1254 @@
...
+    Status = FwuSelectPrevious ();
+    if (EFI_ERROR (Status)) {
+      DEBUG ((
+        DEBUG_ERROR,
+        "FmpPsaFwuLib: Failed to select previous... %r\n",
+        Status
+        ));
+    }
+
+    mUpdateDisabled = TRUE;
```
samimujawar: I think this can go at line 1198

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
```diff
@@ -0,0 +1,1254 @@
...
+
+    /**
+     * Though fwu_select_previous failed,
```
samimujawar: Not entirely sure what you mean here. Can you revise this comment, please?

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
```diff
@@ -0,0 +1,1254 @@
...
+     * firmware update feature.
+     */
+    Status = EFI_SUCCESS;
+
+    goto ErrorHandler;
+  }
+
+  if (mImageEntry->Accepted == FWU_IMAGE_UNACCEPTED) {
+    /**
+     * When UEFI boot and load this Fmp device driver,
```
samimujawar: Would the following text be clearer?
If the firmware is able to boot to this stage and load the FMP device driver, we can consider that the Boot is successful. Therefore, accept the image in the ReadyToBoot phase.

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
```diff
@@ -0,0 +1,1254 @@
...
+  UINT32                     Handle;
+  UINT32                     ReadBytes;
+  UINT32                     TotalBytes;
+  UINT32                     Progress;
+  UINT32                     TotalWork;
+  VOID                       *Buffer;
+  PSA_MM_FWU_IMG_INFO_ENTRY  *ImgEntry;
+  UINTN                      Idx;
+
+  Status = FwuOpen (&gPsaFwuImageDirectoryGuid, FWU_STREAM_READ, &Handle);
```
samimujawar: I think it would be clearer to move this FwuOpen() in the if (mImageDirectory == NULL) block. 
Similarly, the FwuOpen() at line 199 can be moved out to line 209.

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
```diff
@@ -0,0 +1,1254 @@
...
+  }
+
+  /**
+   * Refresh Image Directory.
+   */
+  Status = GetFwuImageDirectory ();
+  if (EFI_ERROR (Status) ||
+      (mImageEntry->Accepted == FWU_IMAGE_UNACCEPTED))
+  {
+    DEBUG ((DEBUG_ERROR, "FmpPsaFwuLib: Not renewed Image Directory after accept - %g\n", &gEfiCallerIdGuid));
```
samimujawar: "FmpPsaFwuLib: Image not renewed in Image Directory after calling accept - %g\n"

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
```diff
@@ -0,0 +1,1254 @@
...
+FmpDeviceGetLowestSupportedVersion (
+  IN OUT UINT32  *LowestSupportedVersion
+  )
+{
+  //
+  // Retrieve the lowest support version from a PCD
+  // NOTE: This method of using a PCD can only be used for the system firmware
+  //       FMP instance that is updated every time the system firmware is
+  //       updated.  If system firmware updates support partial updates that
+  //       would not include the system firmware FMP instance, then a PCD can
```
samimujawar, line 533: can not => cannot

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
```diff
@@ -0,0 +1,1254 @@
...
+  if (VersionString == NULL) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  //
+  // Retrieve the version string from a PCD
+  // NOTE: This method of using a PCD can only be used for the system firmware
+  //       FMP instance that is updated every time the system firmware is
+  //       updated.  If system firmware updates support partial updates that
+  //       would not include the system firmware FMP instance, then a PCD can
```
samimujawar, line 582: can not => cannot

samimujawar: same comment for line 620

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
```diff
@@ -0,0 +1,1254 @@
...
+  }
+
+  if (Progress != NULL) {
+    Progress (100);
+  }
+
+  *LastAttemptStatus = LAST_ATTEMPT_STATUS_SUCCESS;
+
+  /**
+   * Renewal CorrectBoot value on ImageDirectory.
```
samimujawar: Reload CorrectBoot value from ImageDirectory.

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
```diff
@@ -0,0 +1,1254 @@
...
+   */
+  Status = GetFwuImageDirectory ();
+
+  ASSERT (Status == EFI_SUCCESS);
+  ASSERT (!(mImageDirectory->CorrectBoot));
+
+  /**
+   * After end staging, active idx is changed.
+   * In other word, the UpdateAgent's status out of correct boot.
```
samimujawar: Can you rephrase this comment, please?

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c
LeviYeoReum: Okay. I'll make separate patch.

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
LeviYeoReum: Thanks! I'll change it!

ArmPkg/ArmPkg.ci.yaml
LeviYeoReum: > In the commit message the "FwuSmm.c" block text should be updated to "FwuStMm", to reflect diagram in the Overview section at https://github.com/tianocore/edk2/pull/11507

Sorry. the update name is "FwuSmm" not "FwuStMm" when you see the source:
  - https://github.com/tianocore/edk2-platforms/pull/873/commits/4abb32c3acd04ad6c777b662c82eba0167b26a76
  
I'll update it.

> Similarly the text FmpPsaFwuLib should be changed to FmpDevicePsaFwuLib. Also would it be clearer to indicate that FmpDevicePsaFwuLib is an instance of the FmpDeviceLib? Would it be better to update the ASCII diagram as shown below?

That's much clear. I'll update it. Thanks!

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
LeviYeoReum: Okay. I'll remove it.

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
LeviYeoReum: This is for supporting runtime capsule update.
I'll add the comment. Thanks!

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
LeviYeoReum: I think you miss the latest version which includes your point.

Thanks ;)

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
LeviYeoReum: Thanks. I'll rephrase it.

ArmPkg/Drivers/StandaloneMmCpu/EventHandle.c
LeviYeoReum: Thanks for your suggestion.
I'll reword it.

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
LeviYeoReum: Thanks. I'll  update  comments.

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
LeviYeoReum: Ah. this might be a out of specification as spec updated.
I think this would come from the "draft" specification but I notice there is no return value for this.
I'll remove this.

Thanks!

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
LeviYeoReum: Thanks. I'll apply it.

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
LeviYeoReum: Ah. I've forgotten. I'll add it.

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
LeviYeoReum: Nope When you see the `FwuCommunicate()`, PSA_MM_FWU_CMD_DATA includes the `PSA_MM_FWU_PARAMETER_HEADER` and each REQ parameters are delivers as "DATA".

That's why the "_REQ" (request parameter) is defined pure reuqest data without "PSA_MM_FW_PARAMETER_HEADER"

Since the spec said the both function doesn't have "PARAMTER", I think I'll remain it as it is.

Thanks.

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
LeviYeoReum: Nope. I'll remove the Metadata V1.

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
LeviYeoReum: Okay. I'll drop it

ArmPkg/Include/IndustryStandard/PsaMmFwUpdate.h
LeviYeoReum: Thanks. That's would be better.

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.inf
LeviYeoReum: Thanks. I'll fix it.

ArmPkg/Library/FmpDevicePsaFwuLib/PsaFwuLib.c
LeviYeoReum: Right. I'll get rid of all of support of MM_COMMUNICATION and DIRECT_REQ for firwmare update.
Thanks!

ArmPkg/Library/FmpDevicePsaFwuLib/PsaFwuLib.c
LeviYeoReum: Thanks. I'll change it.

ArmPkg/Library/FmpDevicePsaFwuLib/PsaFwuLib.c
LeviYeoReum: Okay. I'll change it. Thanks!

ArmPkg/Library/FmpDevicePsaFwuLib/PsaFwuLib.c
LeviYeoReum: UpdateCount == 0 means. update All image in the bank.
See the FmpDeviceSetImageWithStatus() usage and spec 3.4.2.2 fwu_begin_staing the case when
partial_update_count is 0.

But, UpdateCount < 0 is useless. So I'll remove that one. Thanks!

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
LeviYeoReum: That would be better. 
Thanks!

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
LeviYeoReum: Okay. I'll revise. Thanks!

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
LeviYeoReum: Thanks for your suggestion.
I'll apply this.

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
LeviYeoReum: Thanks. I'll change it

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
LeviYeoReum: Okay. I'll revise it.

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
LeviYeoReum: Thanks. I'll change to it.

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
LeviYeoReum: Thanks.

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
LeviYeoReum: Thanks.

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
LeviYeoReum: That would be better. I'll change it!

ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
LeviYeoReum: No. There is "No seek" operation for Fwu file read.
IOW Image directory size is found, It couldn't reuse the handle, but it need to open it again.
Therefore, I'll remain it as it is.

MdePkg/Include/IndustryStandard/ArmFfaSvc.h
```diff
@@ -182,13 +182,13 @@
      ARM_FFA_SET_MEM_ATTR_DATA_PERM_SHIFT))
 
 /*
- * macro used in FFA_MSG_DIRECT_REQ/REQ2
- * See FF-A spec Chapther 15.2 FFA_MSG_SEND_DIRECT_REQ and
- * 15.4 FFA_MSG_SEND_DIRECT_REQ2
+ * macro used for partition information
+ * See FF-A spec Chapther 6.1 Partition identification
```
ardbiesheuvel: 'chapter'

ArmPkg/ArmPkg.ci.yaml
```diff
@@ -84,7 +84,10 @@
 
     ## options defined .pytool/Plugin/LibraryClassCheck
     "LibraryClassCheck": {
-        "IgnoreHeaderFile": []
+        "IgnoreHeaderFile": [
+          "Include/Library/ArmMmHandlerContext.h"
```
ardbiesheuvel, line 89: Why is this needed?

LeviYeoReum: Because, `ArmMmHandlerContext.h` doesn't have Library like ArmMmHandlerContextLib.
This only has "header".

LeviYeoReum:
> Looks ok to me but I will leave the final word to @samimujawar

Thanks!

MdePkg/Include/IndustryStandard/ArmFfaSvc.h
LeviYeoReum: Thanks. I'll fix it.

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c
```diff
@@ -869,13 +869,15 @@ DelegatedEventLoop (
...
-  UINTN         CommBufferAddr;
+  EFI_STATUS              Status;
+  UINT64                  Uuid[2];
+  VOID                    *CommData;
+  EFI_GUID                ServiceGuid;
+  SERVICE_TYPE            ServiceType;
+  UINTN                   CommBufferAddr;
+  ARM_MM_HANDLER_CONTEXT  MmHandlerContext;
+  FFA_MSG_INFO            *FfaMsgInfo;
+  SPM_MM_MSG_INFO         *SpmMmInfo;
```
leiflindholm: I realise this is a late comment, but:

This function was already a bit of a problem. With this, it's pushing 150 lines, and the number of local variables is at the very least getting unseemly.

Could we at minimum break the contents of the different CommProtocol cases out into separat helpers so that anyone can plausibly at least pretend to understand the flow?

ArmPkg/Library/FmpDevicePsaFwuLib/PsaFwuLib.h
```diff
@@ -0,0 +1,273 @@
...
+ *
+ */
+VOID
+EFIAPI
+PsaFwuVirtualAddressChangeEvent (
+  IN EFI_EVENT  Event,
+  IN VOID       *Context
+  );
+
+#endif /**< End of __PSA_FWU_LIB_H__ */
```
leiflindholm: Guard macro does not have leading __, so this shouldn't either.

ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c
LeviYeoReum: Thanks. I'll separate for this.

ArmPkg/Library/FmpDevicePsaFwuLib/PsaFwuLib.h
LeviYeoReum: Ah. This is only wrong comment. Guard is `PSA_FWU_LIB_H_`.
I'll change it. Thanks!

PR #11511, opened 2025-09-17
BaseTools/tools_def.template: use -funsigned-chars for LOONGARCH64

leiflindholm:
As dissussed at some length in commit message for 3f7f28717892, ("BaseTools: use unsigned chars on ARM architectures"), edk2 pretty much assumes unsigned chars, so set this explicitly for LOONGARCH64 also.

I *hope* this isn't a breaking change, but if it is, it probably fixes as much as it breaks, and should prevent future issues.

## How This Was Tested

Limited compile testing.


MarsDoge:
According to my verification, on ARM platforms, `plain char` should be unsigned by default, but on LoongArch/x86, it should be signed by default.  So, ` -fsigned-char `is actually redundant on LoongArch.
Back to the ARM modification before, if `plain char` needs to be unified, does x86 also need to explicitly adjust from signed to unsigned ? Looking forward to your reply. @leiflindholm

leiflindholm:
@MarsDoge yes. Although, hmm, I note this is already the case for the clang profiles:
https://github.com/tianocore/edk2/blob/master/BaseTools/Conf/tools_def.template#L1936,
https://github.com/tianocore/edk2/blob/master/BaseTools/Conf/tools_def.template#L2043
but not for GCC.

Although in fact CLANGDWARF_ALL_CC_FLAGS/CLANGPDB_ALL_CC_FLAGS are used _only_ for IA32/X64 profiles, and include several extremely x86-specific command line options. Urgh, this needs bigger untangling.

I'll close this and raise a new bigger cleanup PR that will end up including the same ultimate change for LOONGARCH64.

PR #11512, opened 2025-09-17
BaseTools,EmbeddedPkg: enable EmbeddedPkg builds for LOONGARCH64/RISCV64

leiflindholm:
Several LOONGARCH64 and RISCV64 platforms make use of modules from EmbeddedPkg,
so add these as a supported architecture.

Also add the missing plumbing to permit building NOOPT target with LOONGARCH64
GCC toolchain profiles.


ardbiesheuvel:
Not sure how to fix this PR after I tried using the web UI to resolve a conflict.

bexcran (approved):

PR #11519, opened 2025-09-18
MdePkg/Library/BaseLib: Remove __chkstk() from BaseLib

mdkinney:
# Description

Remove workaround for GCC compilers that is no longer required. Keeping this implementation of __chkstk() generates duplicate symbol build failures with Windows/CLANGPDB builds host-based unit test builds.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Windows/CLANGPDB host-based unit tests build fail with a duplicate __chkstk() symbol before this change.
Builds pass after this change.

## Integration Instructions

N/A


mdkinney:
@os-d The workaround being removed here has been in place for 17+ years.  I do not think it applies anymore.  Please review to make sure removing this does not have any impact to the recently added stack check features.

os-d:
> This does not affect the current stack cookie implementation. GCC relies on `__stack_chk_fail` for reporting stack cookie errors:
> 
> https://github.com/tianocore/edk2/blob/2a8d98d0a5e8d662726b79e632658919cc8321f7/MdePkg/Library/StackCheckLib/StackCheckLibCommonGcc.c#L33
> 
> .
> Looks like this is was to support GCC targeting Windows, like mingw. Windows does not use this for stack cookies, either.

OVMF IA32X64 and ArmVirtPkg QEMU both build and run with stack cookies enabled, so we should have seen any failure in CI if there was some intersection.

lgao4 (approved):

PR #11526, opened 2025-09-18
MdeModulePkg/HiiDatabaseDxe: Avoid unexpected memory free

EricGao2015:
TempBuffer variable is only dynamically allocated when the opcode is EFI_IFR_STRING_OP and the string value is not 0, so it should be freed under the same condition. Otherwise, the variable may be unexpected freed.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions

### DIFF
--- a/MdeModulePkg/Universal/HiiDatabaseDxe/ConfigRouting.c
+++ b/MdeModulePkg/Universal/HiiDatabaseDxe/ConfigRouting.c
@@ -4298,7 +4298,7 @@ GenerateAltConfigResp (
           DefaultString = NULL;
         }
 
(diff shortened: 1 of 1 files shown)


EricGao2015:
@lgao4 @dandanbi Hi all, please help to review this patch, thanks.

kilaterlee:
@dandanbi Can you help to review this PR? THX!

kilaterlee:
@lgao4 Hi Liming, I found you have approved this patch, can I give the `push` lable if there have no objection?

PR #11527, opened 2025-09-18
ShellPkg: Code refactoring

pierregondois:
# Description

In an effort to simplify the code in the ShellPkg:
- Refactor the code of some shell commands
- Remove deprecated/unsused code
- Introduce aliases for the `ShellPrintEx()` and `ShellPrintHiiEx()` functions to remove optional parameters.

For most patches, no functional change is introduced.
 
- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Compared the output of the following commands with/without the patches:
```
memmap
memmap -sfo

dmem
dmem -verbose

drvcfg
# Compared the content of XXX
drvcfg -o XXX

connect
# output:
# Connect - Handle [17] Result Success.
# output: Connect - Handle [38] Result Success.
# Then
connect 17
connect 38
```

### DIFF
--- a/ShellPkg/Library/UefiShellLevel2CommandsLib/Load.c
+++ b/ShellPkg/Library/UefiShellLevel2CommandsLib/Load.c
@@ -125,54 +125,51 @@ LoadDriver (
     }
 
     ShellPrintHiiDefaultEx (STRING_TOKEN (STR_LOAD_NOT_IMAGE), gShellLevel2HiiHandle, FileName, Status);
+    goto Exit;
+  }
+
+  //
+  // Make sure it is a driver image
+  //
+  Status = gBS->HandleProtocol (LoadedDriverHandle, &gEfiLoadedImageProtocolGuid, (VOID *)&LoadedDriverImage);
+
+  ASSERT (LoadedDriverImage != NULL);
+
+  if (EFI_ERROR (Status) ||
+      ((LoadedDriverImage->ImageCodeType != EfiBootServicesCode) &&
+       (LoadedDriverImage->ImageCodeType != EfiRuntimeServicesCode)))
+  {
+    ShellPrintHiiDefaultEx (STRING_TOKEN (STR_LOAD_NOT_DRIVER), gShellLevel2HiiHandle, FileName);
+
+    //
+    // Exit and unload the non-driver image
+    //
+    gBS->Exit (LoadedDriverHandle, EFI_INVALID_PARAMETER, 0, NULL);
+    Status = EFI_INVALID_PARAMETER;
+    goto Exit;
+  }
+
+  //
+  // Start the image
+  //
+  Status = gBS->StartImage (LoadedDriverHandle, NULL, NULL);
+  if (EFI_ERROR (Status)) {
+    ShellPrintHiiDefaultEx (STRING_TOKEN (STR_LOAD_ERROR), gShellLevel2HiiHandle, FileName, Status);
   } else {
-    //
-    // Make sure it is a driver image
-    //
-    Status = gBS->HandleProtocol (LoadedDriverHandle, &gEfiLoadedImageProtocolGuid, (VOID *)&LoadedDriverImage);
-
-    ASSERT (LoadedDriverImage != NULL);
-
-    if (  EFI_ERROR (Status)
-       || (  (LoadedDriverImage->ImageCodeType != EfiBootServicesCode)
-          && (LoadedDriverImage->ImageCodeType != EfiRuntimeServicesCode))
-          )
-    {
-      ShellPrintHiiDefaultEx (STRING_TOKEN (STR_LOAD_NOT_DRIVER), gShellLevel2HiiHandle, FileName);
-
+    ShellPrintHiiDefaultEx (STRING_TOKEN (STR_LOAD_LOADED), gShellLevel2HiiHandle, FileName, LoadedDriverImage->ImageBase, Status);
+    if (Connect) {
       //
-      // Exit and unload the non-driver image
+      // Connect it...
       //
-      gBS->Exit (LoadedDriverHandle, EFI_INVALID_PARAMETER, 0, NULL);
-      Status = EFI_INVALID_PARAMETER;
+      Status = ConnectAllEfi ();
     }
   }
 
-  if (!EFI_ERROR (Status)) {
-    //
-    // Start the image
-    //
-    Status = gBS->StartImage (LoadedDriverHandle, NULL, NULL);
-    if (EFI_ERROR (Status)) {
-      ShellPrintHiiDefaultEx (STRING_TOKEN (STR_LOAD_ERROR), gShellLevel2HiiHandle, FileName, Status);
-    } else {
-      ShellPrintHiiDefaultEx (STRING_TOKEN (STR_LOAD_LOADED), gShellLevel2HiiHandle, FileName, LoadedDriverImage->ImageBase, Status);
-    }
-  }
-
-  if (!EFI_ERROR (Status) && Connect) {
-    //
-    // Connect it...
-    //
-    Status = ConnectAllEfi ();
-  }
-
+Exit:
   //
   // clean up memory...
   //
-  if (FilePath != NULL) {
-    FreePool (FilePath);
-  }
+  FreePool (FilePath);
 
   return (Status);
 }


leiflindholm:
Before I start nitpicking, I just really want to say: thank you for this brilliant bit of cleanup.

ShellPkg/Library/UefiShellDebug1CommandsLib/Dmem.c
```diff
@@ -504,20 +501,12 @@ DisplaySystemTable (
...
-    if (ShellStatus == SHELL_SUCCESS) {
-      ShellStatus = DisplayConformanceProfiles (ConformanceProfileTableAddress);
-    }
+    DisplayRtProperties (RtPropertiesTableAddress);
+    DisplayImageExecutionEntries (ImageExecutionTableAddress);
+    DisplayConformanceProfiles (ConformanceProfileTableAddress);
   }
 
-  return ShellStatus;
+  return SHELL_SUCCESS;
```
leiflindholm: I agree with this change, but the commit message is actually incorrect. ShellStatus _is_ used, as the return value from each previous function call.

But a) I don't necessarily see great value in that.
And b)
- DisplayRtProperties looks broken for this anyway:
It takes an address in, ignores it, looks up the config table anyway, doesn't check the return value, uses the result blindly, and then only does an error check on whether the printout succeeded...
- DisplayImageExecutionEntries and DisplayConformanceProfiles also ignore the address parameter completely, and forego any useful error handling.

So, please keep this change, but also update the commit message indicating that the change _is_ functional. Just not practically so.

I see you rework these helper functions in later cmomits, so I'll comment on their pre-existing brokenness there, I think there is more cleanup required.

ShellPkg/Library/UefiShellDebug1CommandsLib/MemMap.c
```diff
@@ -147,6 +147,203 @@ AddMemoryLength (
...
+  MEMORY_LENGTH_ENTRY     *Entry;
+
+  InitializeListHead (&MemoryList);
+
+  TotalPages = 0;
+
+  // Reset the MemoryPageCount array.
+  ZeroMem (MemoryPageCount, sizeof (MemoryPageCount));
+
+  for ( Walker = Descriptors
```
leiflindholm: Stray leading space after (? Ah, retained from original. But I think it's worth cleaning up.

ShellPkg/Library/UefiShellDebug1CommandsLib/MemMap.c
```diff
@@ -147,6 +147,203 @@ AddMemoryLength (
...
 }
 
+/** Parse Memory Descriptors
+
+@param[in]  Descriptors   Array of pointers to Memory Descriptors.
+@param[in]  Size          Size of the Descriptors array.
+@param[in]  ItemSize      Size of a Memory Descriptor.
+@param[in]  Sfo           Whether 'Standard Format Output' should be used.
+**/
+SHELL_STATUS
```
leiflindholm, line 158: static?

ShellPkg/Include/Library/ShellLib.h
```diff
@@ -968,6 +968,16 @@ ShellPrintHiiEx (
   ...
   );
 
+/** Macro aliasing the ShellPrintEx() function and providing default parameters for
+    the Col and Row position.
+**/
+#define ShellPrintDfltEx(...)  ShellPrintEx(-1, -1, __VA_ARGS__)
```
leiflindholm: I'm all for the idea, but Dflt makes for annoying CamelCase. Please expand to Default (for both).

leiflindholm (changes requested):

ShellPkg/Library/UefiShellDebug1CommandsLib/Dmem.c
leiflindholm: Ah, no, that was just the printouts. The actual logic change was preceding...

ShellPkg/Library/UefiShellDebug1CommandsLib/Dmem.c
```diff
@@ -97,47 +97,44 @@ DisplayRtProperties (
 {
   EFI_RT_PROPERTIES_TABLE  *RtPropertiesTable;
   UINT32                   RtServices;
-  SHELL_STATUS             ShellStatus;
   EFI_STATUS               Status;
 
-  ShellStatus = SHELL_SUCCESS;
-
-  if (Address != 0) {
```
leiflindholm, line 105: Yeah, so here - if you look at the original code, it never uses Address at all. And it never checks the return value ot EfiGetSystemConfigurationTable.

The pattern repeats across all three functions.
I'd suggest turning this into a functional change:
- dropping the EfiGetSystemConfigurationTable calls (which will already have been performed before calling this function with its address)
- actually using the address
- and then considering whether these need return values at all for a printout

pierregondois: Hoping I understood the comment correctly, the changes should have been made

ShellPkg/Library/UefiShellDebug1CommandsLib/Dmem.c
```diff
@@ -88,27 +88,24 @@ DisplayMmioMemory (
...
-  SHELL_STATUS             ShellStatus;
-  EFI_STATUS               Status;
+  UINT32        RtServices;
+  SHELL_STATUS  ShellStatus;
+  EFI_STATUS    Status;
 
   ShellStatus = SHELL_SUCCESS;
 
-  if (Address == 0) {
+  if (RtPropertiesTable == 0) {
```
leiflindholm: I approve of the changing the "Address" parameters to pointers, but then it also follows that the tests should be for NULL, not 0.

pierregondois: yes right indeed

ShellPkg/Library/UefiShellLevel2CommandsLib/Load.c
```diff
@@ -154,9 +154,9 @@
     //
     Status = gBS->StartImage (LoadedDriverHandle, NULL, NULL);
     if (EFI_ERROR (Status)) {
-      ShellPrintHiiEx (-1, -1, NULL, STRING_TOKEN (STR_LOAD_ERROR), gShellLevel2HiiHandle, FileName, Status);
+      ShellPrintHiiDefaultEx (STRING_TOKEN (STR_LOAD_ERROR), gShellLevel2HiiHandle, FileName, Status);
     } else {
-      ShellPrintHiiEx (-1, -1, NULL, STRING_TOKEN (STR_LOAD_LOADED), gShellLevel2HiiHandle, FileName, LoadedDriverImage->ImageBase, Status);
+      ShellPrintHiiDefaultEx (STRING_TOKEN (STR_LOAD_LOADED), gShellLevel2HiiHandle, FileName, LoadedDriverImage->ImageBase, Status);
```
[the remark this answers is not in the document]
leiflindholm: Another pre-existing issue caused by appalling code design (looking like the traditional misintepretation of MISRA). Ultimately, success-handling.

Untangle the success-handling and the error report (for the case that cannot be reached) goes away.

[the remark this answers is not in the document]
pierregondois: Yes ok, I thought it was a false positive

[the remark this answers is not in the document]
leiflindholm: Tbh, I was thinking to take it further:
https://github.com/leiflindholm/edk2/commit/b443101616de1ac30ab8f7257369f69d771d87be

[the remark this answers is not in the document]
pierregondois: yes ok, I replaced the last patch by yours if it's ok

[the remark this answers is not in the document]
leiflindholm: My bad (uncrustify), update pushed to https://github.com/leiflindholm/edk2/tree/rework/11527

pierregondois:
Just a ping in case the PR looks sensible

leiflindholm:
> Just a ping in case the PR looks sensible

Wanted to give others a chance to comment. Thanks for the ping, I've set the push label.

PR #11529, opened 2025-09-18
FmpDevicePkg/FmpDxe: Improve handling of XDR certs

antklein:
# Description
The current implementation for XDR certificate handling assumes the PublicKeyDataXdr buffer is 4-byte aligned. This is not always the case. For example, if PublicKeyDataXdr=0x02, and the length is 0x05, the correct offset should be 0x2 + align_up(0x5, 4) -> 0x2 + 0x8 = 0xA. The current logic produces align_up(0x2 + 0x5, 4) -> align_up(0x7, 4) = 0x8.

RFC XDR reference on 4-byte alignment. https://www.rfc-editor.org/rfc/rfc4506#page-16
```
   (4) Why is the XDR unit four bytes wide?

   There is a tradeoff in choosing the XDR unit size.  Choosing a small
   size, such as two, makes the encoded data small, but causes alignment
   problems for machines that aren't aligned on these boundaries.  A
   large size, such as eight, means the data will be aligned on
   virtually every machine, but causes the encoded data to grow too big.
   We chose four as a compromise.  Four is big enough to support most
   architectures efficiently, except for rare machines such as the
   eight-byte-aligned Cray*.  Four is also small enough to keep the
   encoded data restricted to a reasonable size.
```

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Verified this change properly aligns the PublicKeyDataXdr pointer to the next certificate in the list, even when the initial pointer is not 4-byte aligned.

## Integration Instructions
N/A

### DIFF
--- a/FmpDevicePkg/FmpDxe/FmpDxe.c
+++ b/FmpDevicePkg/FmpDxe/FmpDxe.c
@@ -923,8 +923,7 @@ CheckTheImageInternal (
         break;
       }
 
-      PublicKeyDataXdr += PublicKeyDataLength;
-      PublicKeyDataXdr  = (UINT8 *)ALIGN_POINTER (PublicKeyDataXdr, sizeof (UINT32));
+      PublicKeyDataXdr += ALIGN_VALUE (PublicKeyDataLength, 4);
     }
   }
 


antklein:
@lgao4, @mdkinney, @xuweiintel Please help review this change at your earliest convenience.

mdkinney:
The initial pointer value is from PcdGetPtr(), which does not guarantee any alignment.  PCD stores the data.

```
/**
  Retrieves a pointer to a value for a given PCD token.

  Retrieves the current pointer to the buffer for a PCD token number.
  Do not make any assumptions about the alignment of the pointer that
  is returned by this function call.  If the TokenNumber is invalid,
  the results are unpredictable.

  @param[in]  TokenNumber The PCD token number.

  @return The pointer to the buffer to be retrived.

**/
VOID *
EFIAPI
PeiPcdGetPtr (
  IN UINTN  TokenNumber
  );
```

The XDR buffer is generated with tools, and can not know if the PCD storage will be aligned or not, so all lengths are relative to the beginning of the buffer assuming alignment.  If this change is implemented, then I think the wrong offset would be calculated if PcdGetPtr() returns an unaligned pointer value.

antklein:
> The initial pointer value is from PcdGetPtr(), which does not guarantee any alignment. PCD stores the data.
> 
> ```
> /**
>   Retrieves a pointer to a value for a given PCD token.
> 
>   Retrieves the current pointer to the buffer for a PCD token number.
>   Do not make any assumptions about the alignment of the pointer that
>   is returned by this function call.  If the TokenNumber is invalid,
>   the results are unpredictable.
> 
>   @param[in]  TokenNumber The PCD token number.
> 
>   @return The pointer to the buffer to be retrived.
> 
> **/
> VOID *
> EFIAPI
> PeiPcdGetPtr (
>   IN UINTN  TokenNumber
>   );
> ```
> 
> The XDR buffer is generated with tools, and can not know if the PCD storage will be aligned or not, so all lengths are relative to the beginning of the buffer assuming alignment. If this change is implemented, then I think the wrong offset would be calculated if PcdGetPtr() returns an unaligned pointer value.

Right, that's what my PR is fixing. When PcdGetPtr() returns an unaligned pointer value the current pointer math does not properly address this unaligned pointer.

For example, here's the current logic on a simplified design.
```c
PublicKeyDataXdr += PublicKeyDataLength;
PublicKeyDataXdr  = (UINT8 *)ALIGN_POINTER (PublicKeyDataXdr, sizeof (UINT32));
```
Assume the pointer is set to an unaligned value 0x02 and the data length is 0x05.
```
PublicKeyDataXdr = 0x02;
PublicKeyDataLength = 0x05;

PublicKeyDataXdr = 0x02 + 0x05; // => 0x07
PublicKeyDataXdr = (UINT8 *)ALIGN_POINTER (0x07, 4); // => 0x08
```

This design is incorrect, as the expected value for the end of the certificate should be 0xA.
`PublicKeyDataXdr += ALIGN_VALUE (PublicKeyDataLength, 4);`

```
PublicKeyDataXdr = 0x02;
PublicKeyDataLength = 0x05;

PublicKeyDataXdr = 0x02 + ALIGN_VALUE (0x05, 4); // => 0x02 + 0x08 = 0x0A
```

The initial value of PublicKeyDataXdr is not being correctly accounted for in the original math, since the pointer alignment loses it as part of the alignment.

mdkinney:
Thanks for the additional explanation.  That makes sense now.  I think this issue would be observed if PcdGetPtr() returned an unaligned pointer and there is more than one public key.

antklein:
> Thanks for the additional explanation. That makes sense now. I think this issue would be observed if PcdGetPtr() returned an unaligned pointer and there is more than one public key.

Yes, exactly! That's the exact reason this issue was found when debugging on our system.

PR #11531, opened 2025-09-19
EmulatorPkg: Add support for CLANGDWARF

mdkinney:
# Description

Update EmulatorPkg to support Linux/CLANGDWARF builds

* Fix warnings from CLANGDWARF compiler
* Update build_rule.template for CLANGDWARF builds
* Update EmulatorPkg/Unix/Host to support all GCC family builds including CLANGDWARF

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Linux/CLANGDWARF build fails before this change and passes and boots after this change.

All of the following builds pass and boot to UEFI Shell.
```
build -a X64 -p EmulatorPkg/EmulatorPkg.dsc -t CLANGDWARF -b DEBUG
build -a X64 -p EmulatorPkg/EmulatorPkg.dsc -t CLANGDWARF -b RELEASE
build -a X64 -p EmulatorPkg/EmulatorPkg.dsc -t CLANGDWARF -b NOOPT
```

## Integration Instructions

N/A


os-d (approved):
Approved for EmulatorPkg. @lgao4 @bexcran can you review the BaseTools change? It looks okay to me. @mdkinney , my only question is do we think CLANG* needs more plumbing through build_rule.template? I would have thought it would have followed the GCC path (but I guess it isn't family, but the sub family thing I'm forgetting the name for).

os-d:
> Approved for EmulatorPkg. @lgao4 @bexcran can you review the BaseTools change? It looks okay to me. @mdkinney , my only question is do we think CLANG* needs more plumbing through build_rule.template? I would have thought it would have followed the GCC path (but I guess it isn't family, but the sub family thing I'm forgetting the name for).

Question rescinded, I see you put up a PR to address exactly this :)

PR #11536, opened 2025-09-19
Avoid using MTRRs with AMD SEV-SNP

RARelph:
# Description

MTRRs are not supported in guests running under AMD's SEV-SNP. These 2 commits avoid using the MTRRs in a way that is similar to the way MTRR usage is avoided with TDX. The first commit establishes the CCGuestAttr value earlier than it previously was for SEV guests so that it can be used by the second commit, which tests that value to avoid using MTRRs.

## How This Was Tested

This issue was discovered by the Coconut-SVSM project and was fixed and tested in their EDK2 fork.

## Integration Instructions

N/A


RARelph:
@kraxel - @tlendacky thought you might be able to advise how to resolve this remaining failure. My change needs to access a PCD, but the OVMF_X64_MM_DEBUG configuration is somehow using the PcdLibNull.

kraxel:
Not sure why the NULL version of the lib is used, possibly simply because nobody used dynamic PCDs before.  Maybe you simply need to change the library, simliar to commit d6f828b3849e5adcf9a04d3c16e79388db9b3614

RARelph:
@kraxel, @tlendacky - Thanks. I tried putting "PcdLib|MdePkg/Library/PeiPcdLib/PeiPcdLib.inf" in LibraryClasses.common.PEI_CORE (as you had), but the same problem occurred. Then I tried putting it under MdeModulePkg/Universal/PCD/Pei/Pcd.inf. That didn't fix the problem either. I feel like a bull in a china shop. I have no relevant experience here. I'd really appreciate some help figuring out what to change in the 1100 line OvmfPkg/OvmfPkgX64.dsc file to change just the OVMF_X64_MM_DEBUG build.

RARelph:
FWIW, the assert pops because InitializeMpServiceData() in UefiCpuPkg/PiSmmCpuDxeSmm/MpService.c calls MtrrGetAllMtrrs(). That ends up at MtrrLibIsMtrrSupported(), which does
`if (TdIsEnabled () || SnpIsEnabled ())`
TdIsEnabled() is for TDX, of course, and does a bunch of CPUID operations to check for TDX. I had initially approached creating SnpIsEnabled() the same way, but Tom advised that we're not supposed to add anything to MdePkg/Library/BaseLib/X64, which is where TdIsEnabled() is defined (in TdProbe.c). He advised we should be using PCDs, so... that's what I am trying to do.

kraxel:
Maybe create two variants of the library?  Move the checks for SNP and TDX to a new source file.  Have two variants of that source file.  One which simply says 'no TDX / SNP', and one which actually runs the checks.  The latter can be used in OVMF builds which are compatible with TDX/SNP, the former everywhere else.

The OVMF builds with SMM support (no matter whenever standalone or not) are incompatible with TDX/SNP because SMM emulation depends on the host accessing guest register state, so that should get us past the SMM CI failures.

RARelph:
@kraxel - Thanks. With @tlendacky's help, I am now able to build and test this failing configuration locally. With your suggestion, I tried an experiment removing the code that does the PCD read, rebuilt, and re-run. That gets further... but still fails. So I checked out upstream/master, rebuilt, and re-ran... and that fails the same way.

Ftw: Remaining work space size - FE0
MmInstallProtocolInterface: 3868FC3B-7E45-43A7-906C-4BA47DE1754D 7F120020
  DriverEntry (Scheduled) - 7EE2C0C1-C21A-4113-A53A-66824A95696F
MmLoadImage - 7EE2C0C1-C21A-4113-A53A-66824A95696F

ASSERT_EFI_ERROR (Status = Out of Resources)
ASSERT [StandaloneMmCore] Dispatcher.c(182): !(((RETURN_STATUS)(Status)) >= 0x8000000000000000ULL)

While I accept it's probably a problem in my local environment somehow, can you confirm my assumption that no change gets made on the master branch unless ALL the checks pass?

RARelph:
@kraxel - I'll head down the 2 libraries approach.
But @tlendacky and I have observed that if, immediately prior to the GetPcd64() that causes the abort, we put in a GetPcd32() against the other PCD (gUefiCpuPkgTokenSpaceGuid.PcdCpuNumberOfReservedVariableMtrrs) that is declared in MtrrLib.inf... that succeeds. We don't understand why GetPcd64() aborts, but GetPcd32() works. Note that it's the 3rd time through this block that the abort occurs... not the first or second.

```
  if (!mCurrentAttrRead) {  
    DEBUG((DEBUG_INFO, "Got here %08x\n", PcdGet32 (PcdCpuNumberOfReservedVariableMtrrs)));
    mCurrentAttr     = PcdGet64 (PcdConfidentialComputingGuestAttr);
    DEBUG((DEBUG_INFO, "Got there %08x\n", PcdGet32 (PcdCpuNumberOfReservedVariableMtrrs)));
    mCurrentAttrRead = TRUE;
  }

```

kraxel:
There are different kinds of PCDs.  Some are fixed, these are compile time constants and work fine.  Some are dynamic (can be changed at runtime), these are looked up via PcdLib.  Go look at the `AutoGen.h` files in your `Build/` directory of you want see how this works behind the scenes.

UefiCpuPkg/Library/MtrrLib/DetectCcNull.c
niruiyu, line 1: I cannot understand the file name.
The file name contains "Mtrr" + "Sup" + "Smm".
What's the meaning of "Sup"?
Why "Smm" but MtrrLib calls the CcMode implemented in this file always?

The CC version instance depends on a PCD (possibly dynamic PCD), is that intended? As earlier phase in bios boot dynamic PCD service is not available.

RARelph: "Sup" is short for "Support".
MtrrLib.c always calls CcMode(). But there are 2 CcMode() functions implemented. One in MtrrSupSmm.c (always returns FALSE), the other in MtrrSupCc.c (does real work.) I was advised by Gerd (see conversation above) to split it in to 2 libraries. Which MtrrSup*.c file is used is determined by the .inf file referenced in the build's .dsc file.
Yes, the Cc version depends on a dynamic PCD. Early in the implementation, I was advised that this was preferable to trying to add something to MdePkg.

OvmfPkg/AmdSev/AmdSevX64.dsc
```diff
@@ -143,7 +143,12 @@
   IoLib|MdePkg/Library/BaseIoLibIntrinsic/BaseIoLibIntrinsicSev.inf
   OemHookStatusCodeLib|MdeModulePkg/Library/OemHookStatusCodeLibNull/OemHookStatusCodeLibNull.inf
   SerialPortLib|PcAtChipsetPkg/Library/SerialIoLib/SerialIoLib.inf
+!if $(STANDALONE_MM_ENABLE) != TRUE
+  MtrrLib|UefiCpuPkg/Library/MtrrLib/MtrrLibCc.inf
```
kraxel: No need to check for STANDALONE_MM_ENABLE, the AmdSev build does not support SMM, just use MtrrLibCc.inf

OvmfPkg/Bhyve/BhyveX64.dsc
```diff
@@ -157,7 +157,12 @@
   IoLib|MdePkg/Library/BaseIoLibIntrinsic/BaseIoLibIntrinsicSev.inf
   OemHookStatusCodeLib|MdeModulePkg/Library/OemHookStatusCodeLibNull/OemHookStatusCodeLibNull.inf
   SerialPortLib|PcAtChipsetPkg/Library/SerialIoLib/SerialIoLib.inf
+!if $(STANDALONE_MM_ENABLE) != TRUE
+  MtrrLib|UefiCpuPkg/Library/MtrrLib/MtrrLibCc.inf
+!else
+  # CC is not supported with standalone MM enabled
   MtrrLib|UefiCpuPkg/Library/MtrrLib/MtrrLib.inf
+!endif
```
kraxel: No need to change this, bhyve has neither smm emulation nor confidential computing support.

OvmfPkg/CloudHv/CloudHvX64.dsc
```diff
@@ -166,7 +166,12 @@
   IoLib|MdePkg/Library/BaseIoLibIntrinsic/BaseIoLibIntrinsicSev.inf
   OemHookStatusCodeLib|MdeModulePkg/Library/OemHookStatusCodeLibNull/OemHookStatusCodeLibNull.inf
   SerialPortLib|PcAtChipsetPkg/Library/SerialIoLib/SerialIoLib.inf
+!if $(STANDALONE_MM_ENABLE) != TRUE
```
kraxel: ... same four cloud hypervisor.

OvmfPkg/IntelTdx/IntelTdxX64.dsc
```diff
@@ -147,7 +147,12 @@
   IoLib|MdePkg/Library/BaseIoLibIntrinsic/BaseIoLibIntrinsicSev.inf
   OemHookStatusCodeLib|MdeModulePkg/Library/OemHookStatusCodeLibNull/OemHookStatusCodeLibNull.inf
   SerialPortLib|PcAtChipsetPkg/Library/SerialIoLib/SerialIoLib.inf
+!if $(STANDALONE_MM_ENABLE) != TRUE
```
kraxel: Use can use the CC variant unconditionally (simliar to the AmdSev build).

OvmfPkg/Microvm/MicrovmX64.dsc
```diff
@@ -166,7 +166,12 @@
   IoLib|MdePkg/Library/BaseIoLibIntrinsic/BaseIoLibIntrinsicSev.inf
   OemHookStatusCodeLib|MdeModulePkg/Library/OemHookStatusCodeLibNull/OemHookStatusCodeLibNull.inf
   SerialPortLib|PcAtChipsetPkg/Library/SerialIoLib/SerialIoLib.inf
+!if $(STANDALONE_MM_ENABLE) != TRUE
```
kraxel: Not needed, microvm does not support SMM emulation.

OvmfPkg/OvmfXen.dsc
```diff
@@ -158,7 +158,12 @@
   IoLib|MdePkg/Library/BaseIoLibIntrinsic/BaseIoLibIntrinsicSev.inf
   OemHookStatusCodeLib|MdeModulePkg/Library/OemHookStatusCodeLibNull/OemHookStatusCodeLibNull.inf
   SerialPortLib|PcAtChipsetPkg/Library/SerialIoLib/SerialIoLib.inf
+!if $(STANDALONE_MM_ENABLE) != TRUE
```
kraxel: I think this ns not needed too.

UefiCpuPkg/Library/MtrrLib/DetectCcNull.c
kraxel: Usual naming convention is to use 'Null' in the file name for the stub version, so you could go -- for example -- with `DetectCc.c` and `DetectCcNull.c`.

RARelph: @kraxel - I struggled with the notion of whether "Smm" or "Cc" should be 'default'. But for the .c files, your suggestion is good. I'll incorporate that in to my next push.

RARelph: Thanks, I’ve made the change as suggested. Resolving this thread.

OvmfPkg/AmdSev/AmdSevX64.dsc
RARelph: Thanks, I’ve made the change as suggested. Resolving this thread.

OvmfPkg/CloudHv/CloudHvX64.dsc
RARelph: Thanks, I’ve made the change as suggested. Resolving this thread.

OvmfPkg/Microvm/MicrovmX64.dsc
RARelph: Thanks, I’ve made the change as suggested. Resolving this thread.

OvmfPkg/IntelTdx/IntelTdxX64.dsc
RARelph: Thanks, I’ve made the change as suggested. Resolving this thread.

OvmfPkg/OvmfXen.dsc
RARelph: Thanks, I’ve made the change as suggested. Resolving this thread.

OvmfPkg/Bhyve/BhyveX64.dsc
RARelph: Thanks, I’ve made the change as suggested. Resolving this thread.

UefiCpuPkg/Library/MtrrLib/DetectCcNull.c
kraxel: > @kraxel - I struggled with the notion of whether "Smm" or "Cc" should be 'default'.

There are very few builds which need the Cc variant, specifically `AndSevX64.dsc` + `IntelTdx.dsc` (unconditionally) and `OvmfPkgX64.dsc` + `OvmfPkgIa32X64.dsc` (if build without SMM support).  So the non-CC variant should be the default, especially as all physical hardware platforms need that one too.

tlendacky: Agree. I think the flow should be:

- Create new "Cc" library with DetectCc.c that performs the same thing the current library does today with both the original and Cc version of the library using DetectCc.c
- Switch the Ovmf .dsc files over to the new "Cc" library
- Create the new DetectCcNull.c and switch the original library over to use it
- Add SEV-SNP support to DetectCc.c

I think that should accomplish the goal.

RARelph:
I think all of the comments were addressed 3 weeks ago. I've just rebased the changes and re-tested locally to confirm no unexpected problems. Please review and merge.

UefiCpuPkg/Library/MtrrLib/MtrrLib.c
```diff
@@ -19,6 +19,9 @@
 #include <Library/CpuLib.h>
 #include <Library/BaseMemoryLib.h>
 #include <Library/DebugLib.h>
+#include <Library/PcdLib.h>
```
tlendacky: Shouldn't this include be in the DetectCc.c file now?

UefiCpuPkg/Library/MtrrLib/DetectCc.c
```diff
@@ -0,0 +1,61 @@
...
+  //
+  if (((RShiftU64 (CurrentAttr, 8)) & 0xff) == 1) {
+    UINT64  CurrentLevel = CurrentAttr & CCAttrTypeMask;
+
+    if (CurrentLevel == CCAttrAmdSevSnp) {
+      return TRUE;
+    }
+  }
+
+  return FALSE;
```
tlendacky: Maybe I'm missing something, but couldn't this just be written as:

return ((CurrentAttr & CCAttrTypeMask) == CCAttrAmdSevSnp) ? TRUE : FALSE;

OvmfPkg/IntelTdx/IntelTdxX64.dsc
```diff
@@ -147,7 +147,7 @@
   IoLib|MdePkg/Library/BaseIoLibIntrinsic/BaseIoLibIntrinsicSev.inf
   OemHookStatusCodeLib|MdeModulePkg/Library/OemHookStatusCodeLibNull/OemHookStatusCodeLibNull.inf
   SerialPortLib|PcAtChipsetPkg/Library/SerialIoLib/SerialIoLib.inf
-  MtrrLib|UefiCpuPkg/Library/MtrrLib/MtrrLib.inf
+  MtrrLib|UefiCpuPkg/Library/MtrrLib/MtrrLibCc.inf
```
tlendacky: This could cause a bisect issue...  The previous commit will allow a TDX guest to run with the version of the library that is now using the "Null" check before this commit is applied.

tlendacky (commented):
To maintain bisectability of the tree this will need to be done in more than 3 commits.

- Set PCD earlier
- Create DetectCc.c that extracts the current TdIsEnabled() check, modify MtrrLib.inf to use it, and create MtrrLibCc.inf that also uses it.
- Change the dsc files under the OvmfPkg directory
- Create DetectCcNull.c with the "null" implementation for use in MtrrLib.inf
- Modify DetectCc.c to add the SNP support.

UefiCpuPkg/Library/MtrrLib/MtrrLib.c
RARelph: Well, I'm confused. It was necessary when I had the SNP Pcd code in this file. So I can remove this here. Why it isn't needed in DetectCc.c... I'm not sure. But DetectCc.c builds without it and none of the CI checks complain.

UefiCpuPkg/Library/MtrrLib/DetectCc.c
RARelph: I think I was following some pattern in existing code somewhere, but I agree. I'll fix this. It simplifies further to:
return (CurrentAttr & CCAttrTypeMask) == CCAttrAmdSevSnp;

OvmfPkg/IntelTdx/IntelTdxX64.dsc
RARelph: Agreed. Fixing.

UefiCpuPkg/Library/MtrrLib/DetectCc.c
tlendacky: Yes and no.  BOOLEAN isn't the same as bool, it is typedef'd as unsigned char. That's why I did the explicit return of TRUE and FALSE. In this case, it probably doesn't matter, but I was told in the past to explicitly use TRUE and FALSE when returning a BOOLEAN. Not sure if anyone still cares about that today.

RARelph: True. I'll use TRUE and FALSE.

UefiCpuPkg/Library/MtrrLib/DetectCcNull.c
```diff
@@ -0,0 +1,31 @@
...
+
+#include <Library/BaseLib.h>
+#include "DetectCc.h"
+
+BOOLEAN
+CcMode (
+  VOID
+  )
+{
+  STATIC BOOLEAN  mCcMode     = FALSE;
```
kraxel: Shouldn't the Null variant of this just return FALSE, unconditionally?

UefiCpuPkg/Library/MtrrLib/DetectCcNull.c
```diff
@@ -11,21 +11,12 @@
...
-  STATIC BOOLEAN  mCcMode     = FALSE;
-  STATIC BOOLEAN  mCcModeRead = FALSE;
-
-  if (!mCcModeRead) {
-    mCcMode     = TdIsEnabled ();
-    mCcModeRead = TRUE;
-  }
-
-  return mCcMode;
+  return FALSE;
```
kraxel: Ah, it's just temporary, probably to not break things in the middle of the reorganization?

UefiCpuPkg/Library/MtrrLib/DetectCcNull.c
RARelph: That would break the Intel TDX build during a bisect.

UefiCpuPkg/Library/MtrrLib/DetectCcNull.c
RARelph: Exactly. Bisect rule meets no changes across code areas rule.

RARelph:
We really need someone with write access to provide an approving review.

UefiCpuPkg/Library/MtrrLib/DetectCc.c
```diff
@@ -0,0 +1,31 @@
...
+#include <Library/BaseLib.h>
+#include "DetectCc.h"
+
+BOOLEAN
+CcMode (
+  VOID
+  )
+{
+  STATIC BOOLEAN  mCcMode     = FALSE;
+  STATIC BOOLEAN  mCcModeRead = FALSE;
```
ardbiesheuvel: Is this library supposed to be usable by PEIMs? As in that case, writable globals such as these are not permitted.

RARelph: @ardbiesheuvel:
Forgive my ignorance and the dumb question... I am new to this code base. Why can't we use writable globals here? It works (admittedly, "for me") and it passes CI.

If the caching - what the 2 static variables exist to do - is removed the system fails to boot with multiple vCPUs:
ASSERT [CpuMpPei] PeiServicesTablePointer.c(42): PeiServices != ((void *) 0)

Another alternative I see is to do what Intel did for their Confidential Computing environment and add code to MdePkg/Library/BaseLib/X64, which I was advised to not do.

ardbiesheuvel: PEIMs execute in place from [emulated] NOR flash, and so stores to global variables are not generally permitted in libraries that are supposed to be usable in PEI. 

In this particular case, the store might just be dropped by the emulation, but on other platforms, this could trigger some kind of exception, so we should not rely on that.

RARelph: @ardbiesheuvel, @kraxel : Does it matter that these statics are only used for OVMF builds?

ardbiesheuvel: Not really.

This is all very complicated code. PEIMs can be shadowed (and PlatformPei likely is), and so it is dispatched twice: once executing from place in NOR flash, and again running from RAM. 

If `CcMode()` is only called the second time around, things will work as expected, but only by accident. Any refactoring occurring in a completely different part of the PEI stage code could trigger changes that would potentially result in boot errors that are very difficult to diagnose.

The usual idiom is to create a HOB or set a PCD if you want to create global state in PEI. Static variables are not an  option here.

RARelph: The first change in this set of 4 moves setting a dynamic PCD earlier in time specifically so it can be referenced by this code. However, without caching the result of the logic based on the PCD's value, we get the error I noted above. I don't know how to proceed here. I have code that works and passes CI, but depends on caching the value. If I don't cache the value, something actually breaks.

Can OVMF code ever be executed from NOR flash? I thought the "VM" in OVMF stood for "virtual machine".

ardbiesheuvel: > Can OVMF code ever be executed from NOR flash? I thought the "VM" in OVMF stood for "virtual machine".

OVMF boots from virtual NOR flash, but it actually runs PEI from RAM not flash - this means the point I made above about pre-shadow does not actually apply to OVMF, and using writable globals in PEI is fine.

*However*, that does mean this particular library implementation belongs in OvmfPkg and not UefiCpuPkg, and the variables should have a big fat comment that explains why the use of such variables is acceptable in this particular case, even though the library is used by a PEIM.

RARelph: That's a fair point. The change will still impact UefiCpuPkg/MtrrLib, since that's where the bug is (ALL Confidential Computing environments do not support MTRRs, not just Tdx), but I agree the DetectCc pieces can probably be moved to OVMF... somewhere.

mdkinney: These variables appear to cache state.  What is the perf impact of detecting the state every time it is needed?  The MTRR Library services are not used very often.  Not using global variables maximizes compatibility.

Also wondering if a new INF is really needed. It the check is light weight and configurable through PCDs, then a single INF that works for CC and all other envs appears possible.

RARelph: @mdkinney @lgao4 - If I did for SEV what was done to avoid MTRR accesses for TDX, would that be "approved"? Currently, code in MtrrLib calls TdIsEnabled(), which is defined in MdePkg/LibraryBaseLib. I would propose to put SEV detection in a file parallel to TdProbe.c and then have MtrrLib reference the new SEV detection function in addition to checking TdIsEnabled() to avoid accessing MTRRs.

RARelph: @mdkinney - Sorry, I missed replying directly to your message. Unfortunately, when I remove the caching, I get an assert when the VM is run with more than 1 vCPU:
ASSERT [CpuMpPei] PeiServicesTablePointer.c(42): PeiServices != ((void *) 0)

As for the necessity of multiple .infs, it was discovered that some .dsc files are using a Null PCD library, and since the PCD we're using is dynamic, that doesn't work. See the conversation with @kraxel in mid-to-late September above, where he recommended the new INF approach.

tlendacky: Moving the check to MdePkg would use CPUID and RDMSR to determine if SNP is active and not use the PCD, which would avoid the NULL PCD library issue and issues with APs accessing PCDs and also eliminate the need for multiple .inf files.

mdkinney: Should not use any PEI Services in the BASE MtrrLib since the MtrrLib services are called from all CPUs to align MTRR setting across all CPUs.  Using globals to avoid calling PEI Service on APs is not the right approach.  Removing the PEI Services dependency in MtrrLib is the correct aproach.

RARelph: @mdkinney - I don't speak EDK2 fluently, but I think you're saying I *should* use the MdePkg approach to have MtrrLib avoid using MTRRs for SEV guests the way TdIsEnabled() is used in MtrrLib to avoid using MTRRs for TDX guests.

kraxel: > @ardbiesheuvel, @kraxel : Does it matter that these statics are only used for OVMF builds?

It happens to kind-of work for OVMF because the PEI firmware volume is compressed and unpacked to RAM before being started.  There is no gurantee that PEI runs in RAM, ArmVirt for example runs PEI from read/only flash.  Also changing the data section of PEI code makes the PCR0 measurements of OVMF unstable.

OVMF stores any state needed in EFI_HOB_PLATFORM_INFO to avoid the use of global variables in PEI code.

RARelph: I've gone the way of using MdePkg so this issue isn't relevant any longer.

RARelph:
OK, so switching strategies to resolve this issue. Using the method used for Intel Tdx to avoid using MTRRs. Avoided this at first because it involves making changes to MdePkg, but it seems there really aren't any workable alternatives.

RARelph:
@mdkinney - Can you approve this (completely new) change? It's much simpler now that I've resigned myself to having to change MdePkg. All the checks passed, but during the successful CI run, the base branch changed... so I've rebased and that has restarted the hour+ CI.

UefiCpuPkg/Library/MtrrLib/MtrrLib.c
```diff
@@ -162,9 +162,9 @@ MtrrLibIsMtrrSupported (
   MSR_IA32_MTRRCAP_REGISTER  MtrrCap;
 
   //
-  // MTRR is not supported in TD-Guest.
+  // MTRR is not supported in TD-Guest or any SEV guest.
   //
-  if (TdIsEnabled ()) {
+  if (TdIsEnabled () || SevGuestIsEnabled ()) {
```
mdkinney, line 167: Are there any issues with running these functions on all possible x86 CPUs?  Do any additional checks need to be made in these function to verify the x86 CPU manufacturer or model before looking at next level of CPUID and MSR details?

MdePkg/Library/BaseLib/X64/SevProbe.c
```diff
@@ -0,0 +1,44 @@
...
+
+  // Check SEV support
+  AsmCpuid (CPUID_MEMORY_ENCRYPTION_INFO, &RegEax.Uint32, NULL, NULL, NULL);
+  if (RegEax.Bits.SevBit == 0) {
+    return FALSE;
+  }
+
+  // Read SEV_STATUS MSR
+  SevStatus = AsmReadMsr64 (MSR_SEV_STATUS);
+  return (SevStatus & 0x7) != 0 ? TRUE : FALSE;
```
mdkinney: Is there a structure for MSR_SEV_STATUS in the Msr include files?  Recommend adding if not present to make this code easier to maintain without hard coded constants.

MdePkg/Library/BaseLib/X64/SevProbe.c
```diff
@@ -0,0 +1,44 @@
...
+**/
+BOOLEAN
+EFIAPI
+SevGuestIsEnabled (
+  VOID
+  )
+{
+  UINT32                            MaxExtendedLeaf;
+  UINT64                            SevStatus;
+  CPUID_MEMORY_ENCRYPTION_INFO_EAX  RegEax;
```
mdkinney: Change local variable name to MemoryEncryptionInfo or MemoryEncryptionInfoEax.  If that CPUID leaf only provide info in EAX, then Eax in var name not required.

MdePkg/Library/BaseLib/X64/SevProbe.c
RARelph: There is an MSR_SEV_STATUS_REGISTER typedef that I can use, yes.

UefiCpuPkg/Library/MtrrLib/MtrrLib.c
RARelph: @tlendacky asked me to remove the AuthenticAMD check because that value is overridable by hypervisors and that the CPUID_EXTENDED_FUNCTION CPUID check to ensure the existence of the CPUID_MEMORY_ENCRYPTION_INFO CPUID value is safe across all x86 CPUs.

MdePkg/Library/BaseLib/X64/SevProbe.c
RARelph: That CPUID leaf does return values in the other registers as well. But I can change the name of the variable  to MemoryEncryptionInfoEax since we only care about the one value.

MdePkg/Library/BaseLib/X64/SevProbe.c
RARelph: Done.

MdePkg/Library/BaseLib/X64/SevProbe.c
RARelph: Done.

mdkinney:
The updates look good.

The Signed-off-by tag in the commit message are not correct and need to be updated.  Recommend updating git config with correct email address

RARelph:
Fixed the Signed-Off-By. I'm not sure how that happened, but I'll guess that's what happens when you do a rebase from the GitHub UI. Something like that.

lgao4:
@RARelph ， do you want to merge it for this stable tag 202511? I think this PR has been approved before the soft feature freeze.

mdkinney:
@niruiyu Still waiting for UefiCpuPkg maintainer approval

niruiyu:
@mdkinney I approved. May need you or @lgao4 to add push label.

PR #11537, opened 2025-09-19
ArmPkg/SemihostFs: fix crash when failed to open file

LeviYeoReum:
Since FileFcb's list entry doesn't initialised at allocation,
SemihostFs reports below crash while freeing FileFcb when
get failed to size of file in FileOpen():

  ASSERT [SemihostFs] LinkedList.c(75): List->ForwardLink != ((void *) 0)

To address this, initailise list entry properly at FCB allocation
and move RemoveListEntry() to FileClose()
since the FCB entry will be on the list when VolumeOpen() or
FileOpen() is successed.

Signed-off: Yeoreum Yun <yeoreum.yun@arm.com>

### DIFF
--- a/ArmPkg/Filesystem/SemihostFs/AArch64/SemihostFs.c
+++ b/ArmPkg/Filesystem/SemihostFs/AArch64/SemihostFs.c
@@ -95,10 +95,13 @@ AllocateFCB (
   SEMIHOST_FCB  *Fcb;
 
   Fcb = AllocateZeroPool (sizeof (SEMIHOST_FCB));
-  if (Fcb != NULL) {
-    CopyMem (&Fcb->File, &gSemihostFsFile, sizeof (gSemihostFsFile));
-    Fcb->Signature = SEMIHOST_FCB_SIGNATURE;
+  if (Fcb == NULL) {
+    return NULL;
   }
+
+  CopyMem (&Fcb->File, &gSemihostFsFile, sizeof (gSemihostFsFile));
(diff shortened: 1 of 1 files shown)


samimujawar:
Can you describe the steps to recreate the crash/assertion, please?

LeviYeoReum:
> Can you describe the steps to recreate the crash/assertion, please?

This is simple. Just run the UEFI with FVP,
then on the UEFI shell, change to FS to semihost and try "ls" 

Then you can get that error easily.

ArmPkg/Filesystem/SemihostFs/AArch64/SemihostFs.c
```diff
@@ -98,6 +98,7 @@ AllocateFCB (
   if (Fcb != NULL) {
     CopyMem (&Fcb->File, &gSemihostFsFile, sizeof (gSemihostFsFile));
     Fcb->Signature = SEMIHOST_FCB_SIGNATURE;
+    InitializeListHead (&Fcb->Link);
```
ardbiesheuvel: Please take the opportunity to fix this 'success handling' anti-pattern, and do something like

```
if (Fcb == NULL) {
  return NULL;
}

...

return Fcb;
```

LeviYeoReum: Thanks. I'll change it.

LeviYeoReum:
> Please fix the grammer in the commit log: initialised -> get initialised, 'get failed to size of file' ???, successed ???

Sorry. I'll refine commit message.

PR #11540, opened 2025-09-22
OvmfPkg/MemFD: swap memory log buffer and pei firmware volume

kraxel:
This places the memory log buffer next to initial memory for heap and
stack (TempRamBase) and includes it in the memory descriptor #3 for
SEV-SNP (SNP_SEC_MEM_{BASE,SIZE}_DESC_3).  This makes sure the memory
is pre-validated at launch time, which is needed to make memory debug
logging work in SEV-SNP mode.


ajyoung-oracle:
With this one commit/change, does Memory Debug Logging work with SEV-[EP/SNP]?

Or are more changes required (such as to mark the memory buffer pages as unencrypted)?

kraxel:
> With this one commit/change, does Memory Debug Logging work with SEV-[EP/SNP]?

It works for in-guest usage (i.e. read the log after boot via sysfs with 6.17 linux kernel).
SEV-ES works fine without this.  SEV-SNP needs this, otherwise the log buffer is not validated
and you'll get a fault when the firmware tries to write the first line to the buffer.

> Or are more changes required (such as to mark the memory buffer pages as unencrypted)?

Accessing the log from the host needs more changes.

PR #11541, opened 2025-09-22
RIMT parser for acpiview command

krzysztof-drobinski-plastelinet:
# Description

RISC-V IO Mapping Table (RIMT) is the ACPI table describing relations between IOMMU and other devices on RISC-V architecture.

See https://github.com/riscv-non-isa/riscv-acpi-rimt for details.

This PR extends acpiview command by adding RIMT parser

- [ ] Breaking change?
  - **No** 
- [ ] Impacts security?
  - **No**
- [ ] Includes tests?
  - **No**

## How This Was Tested

Tested on qemu-10.1 with RISC-V IOMMU enabled configured as SYS as well as PCIe device
```
qemu-system-riscv64 \
	-cpu rva23s64 \
	-smp 4 \
	-nographic \
	-M virt,acpi=on,aia=aplic-imsic,aia-guests=5,iommu-sys=on,pflash0=pflash0,pflash1=pflash1,  \
	-serial mon:stdio \
	-device virtio-gpu-pci \
	-device qemu-xhci \
	-device usb-kbd \
	-device virtio-rng-pci \
	-device ahci,id=ahci -device ide-hd,drive=disk,bus=ahci.0 -drive id=disk,file=./fat32.img,format=raw,if=none \
	-blockdev node-name=pflash0,driver=file,read-only=on,filename=./RISCV_VIRT_CODE.fd \
	-blockdev node-name=pflash1,driver=file,filename=.//RISCV_VIRT_VARS.fd \
	-bios ./build_opensbi/platform/generic/firmware/fw_dynamic.bin \
	-kernel ./build_linux/arch/riscv/boot/Image -append "root=/dev/sda" \
	--trace "riscv_iommu*"
```
Then in UEFI Shell

```
Shell> acpiview -s rimt


 --------------- RIMT Table --------------- 

Address  : 0x875D7000
Length   : 128
  
00000000 : 52 49 4D 54 80 00 00 00 - 01 E1 42 4F 43 48 53 20   RIMT......BOCHS 
00000010 : 42 58 50 43 20 20 20 20 - 01 00 00 00 42 58 50 43   BXPC    ....BXPC
00000020 : 01 00 00 00 02 00 00 00 - 30 00 00 00 00 00 00 00   ........0.......
00000030 : 00 01 28 00 00 00 00 00 - 52 53 43 56 30 30 30 34   ..(.....RSCV0004
00000040 : 00 00 01 03 00 00 00 00 - 00 00 00 00 00 00 00 00   ................
00000050 : 00 00 00 00 00 00 28 00 - 01 01 28 00 00 00 01 00   ......(...(.....
00000060 : 00 00 00 00 00 00 00 00 - 14 00 01 00 00 00 00 00   ................
00000070 : FF FF 00 00 00 00 00 00 - 30 00 00 00 00 00 00 00   ........0.......

Table Checksum : OK

RIMT                                 :
  Signature                          : RIMT
  Length                             : 128
  Revision                           : 1
  Checksum                           : 0xE1
  Oem ID                             : BOCHS 
  Oem Table ID                       : BXPC    
  Oem Revision                       : 0x1
  Creator ID                         : BXPC
  Creator Revision                   : 0x1
  Number of RIMT Nodes               : 2
  Offset to RIMT Node Array          : 0x30
  Reserved                           : 0x0
  RIMT Node [0]                      :
    Type                             : IOMMU
    Revision                         : 1
    Length                           : 40
    Reserved                         : 0x0
    Id                               : 0
    Hardware ID                      : RSCV0004 (0x3430303056435352)
    Base Address                     : 0x3010000
    Flags                            : 0x0
      Device Type                    : "Platform"
      Proximity Domain               : "Invalid"
      Reserved                       : 0x0

    Proximity Domain                 : 0
    PCIe Segment number              : 0x0
    PCIe B/D/F                       : 0
    Number of interrupt wires        : 0x0
    Interrupt wire array offset      : 0x28
  RIMT Node [1]                      :
    Type                             : PCIe Root Complex
    Revision                         : 1
    Length                           : 40
    Reserved                         : 0x0
    Id                               : 1
    Flags                            : 0x0
      ATS                            : "Not Supported"
      PRI                            : "Not Supported"
      Reserved                       : 0x0

    Reserved                         : 0x0
    PCIe Segment Number              : 0
    Id Mapping Array Offset          : 0x14
    Number Of Id Mappings            : 1
    ID Mapping [0]                   :
      Source ID Base                 : 0x0
      Number of IDs                  : 65535
      Destination Device ID Base     : 0
      Destination IOMMU Offset       : 0x30
      Flags                          : 0x0
        ATS                          : "Not Required"
        PRI                          : "Not Required"
        Reserved                     : 0x0


Table Statistics:
        0 Error(s)
        0 Warning(s)
```

## Integration Instructions
N/A


MdePkg/Include/IndustryStandard/Acpi66.h
```diff
@@ -826,6 +826,98 @@ typedef struct {
 
 #define EFI_ACPI_6_6_PLIC_STRUCTURE_VERSION  1
 
+///
```
leiflindholm: Commit message: Please put a reference to an RIMT specification version. And please write an actual message, not just "go look over there".

ACPI 6.6 contains a link saying go look at https://uefi.org/acpi for the suplementary specification, but I find no such reference in that location. However, as a specification maintained outside of ACPI, could we move this to its own header, like Arm's IORT in IoRemappingTable.h?

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
```diff
@@ -0,0 +1,295 @@
...
+ **/
+STATIC
+VOID
+EFIAPI
+PrintRimtNodeIommu (
+  IN EFI_ACPI_6_6_RIMT_IOMMU_NODE_STRUCTURE  *Node
+  )
+{
+  PrintRimtNodeHeader ((EFI_ACPI_6_6_RIMT_NODE_HEADER_STRUCTURE *)Node);
+  HardwareIdToString (&Node->HardwareId, HardwareIdStr);
```
leiflindholm: Could replace with call to AsciiStrnCpyS instead of custom function?

ShellPkg/Library/UefiShellAcpiViewCommandLib/UefiShellAcpiViewCommandLib.uni
```diff
@@ -92,6 +92,7 @@
 "       MPAM  - Memory System Resource Partitioning and Monitoring Table\r\n"
 "       PPTT  - Processor Properties Topology Table\r\n"
 "       RSDP  - Root System Description Pointer\r\n"
+"       RIMT  - RISC-V IO Mapping Table\r\n"
```
leiflindholm, line 94: Move one line up for sorting purposes?

MdePkg/Include/IndustryStandard/Acpi66.h
krzysztof-drobinski-plastelinet: Make sense, Will follow Arm's approach here

ShellPkg/Library/UefiShellAcpiViewCommandLib/UefiShellAcpiViewCommandLib.uni
krzysztof-drobinski-plastelinet: Ok, will do,

MdePkg/Include/IndustryStandard/Acpi66.h
krzysztof-drobinski-plastelinet: @leiflindholm 

> ACPI 6.6 contains a link saying go look at https://uefi.org/acpi for the suplementary specification, but I find no such reference in that location. However, as a specification maintained outside of ACPI, could we move this to its own header, like Arm's IORT in IoRemappingTable.h?

I have created https://github.com/riscv-non-isa/riscv-acpi-rimt/issues/43 to clarify above.
Meanwhile, I will leave only table signature defined in `Acpi66.h` and move everything else to new header `IndustryStandard/RiscVIoMappingTable.h` That approach should reflect real content of current ACPI-6.6 spec

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
```diff
@@ -0,0 +1,295 @@
...
+  AsciiPrint ("\tHardware Id                 : %a   (0x%lx)\n", HardwareIdStr, Node->HardwareId);
+  Print (L"\tBase Address                : 0x%x\n", Node->BaseAddress);
+  Print (
+    L"\tFlags                       : (0x%x) %s Device | Proximity Domain %s\n",
+    Node->Flags,
+    RimtNodeIommuTypeName[GET_BIT (0, Node->Flags)],
+    RimtNodeIommuProximityDomainTypeName[GET_BIT (1, Node->Flags)]
+    );
+  Print (L"\tProximity Domain            : %d\n", Node->ProximityDomain);
```
pierregondois: Hello Krzysztof,
Would it be possible to follow the table-based approach as the other parsers do ?
The Mcfg Parser should provide a good example:
`ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Mcfg/McfgParser.c`
and there are also wrappers to print bitfields, cf., the `ParseAcpiBitFields()` function

krzysztof-drobinski-plastelinet: Hi Pierre,
I could but the issue I see with table approach is that it respects ACPI structures from `MdePkg/Include/IndustryStandard/` only partially (only generic ACPI header from `Acpi.h` in fact)
I looked at `MdePkg/Include/IndustryStandard/IoRemappingTable.h` and `ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Iort/IortParser.c` for instance.
`IortParser` duplicates most of IORT structures in a way of offset definitions in parser tables.
There is no sync point between state of `IndustryStandard` headers and parser tables in IORT `acpiview` for example. Do you recommend rewriting to `table-based` approach?

pierregondois: There is effectively no synchronization between the MdePkg definitions and the AcpiView definitions, however using the table-based approach:
- allows to have a uniform way to display the information. Modifying `ParseAcpi()` would allow to update the way to print all tables without applying changes to each parser individually
- does some additional checking. In the RIMT parser, there doesn't seem to be any check regarding the size of the table: if the table contains an error, the parser will try to access data outside of the table.
- allows to have cross-table validation. This is not present right now, but there was an idea of keeping track of some field values of some tables in order to validate the fields of some other tables. If the parser just prints the fields, this is not possible to do anymore.

krzysztof-drobinski-plastelinet: @pierregondois Ok What about something like 

```
diff --git a/ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c b/ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
index 5cdd34f442..5410355548 100644
--- a/ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
+++ b/ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
@@ -29,9 +29,9 @@ STATIC EFI_ACPI_6_6_RIMT_STRUCTURE  RimtHdrInfo;
 **/
 STATIC CONST ACPI_PARSER  RimtParser[] = {
   PARSE_ACPI_HEADER (&RimtHdrInfo.Header),
-  { L"Number of RIMT Nodes",              4,  36, L"%d",   NULL, NULL, NULL, NULL },
-  { L"Offset to RIMT Node Array",         4,  40, L"0x%x", NULL, NULL, NULL, NULL },
-  { L"Reserved",                          4,  44, L"0x%x", NULL, NULL, NULL, NULL }
+  { L"Number of RIMT Nodes",              sizeof(RimtHdrInfo.NumberOfRimtNodes), OFFSET_OF(EFI_ACPI_6_6_RIMT_STRUCTURE, NumberOfRimtNodes), L"%d",   NULL, NULL, NULL, NULL },^M
+  { L"Offset to RIMT Node Array",         sizeof(RimtHdrInfo.OffsetToRimtNodeArray), OFFSET_OF(EFI_ACPI_6_6_RIMT_STRUCTURE, OffsetToRimtNodeArray), L"0x%x", NULL, NULL, NULL, NULL },^M
+  { L"Reserved",                          sizeof(RimtHdrInfo.Reserved), OFFSET_OF(EFI_ACPI_6_6_RIMT_STRUCTURE, Reserved), L"0x%x", NULL, NULL, NULL, NULL }^M
```
This way acpiview could validate IndustrialStandard table header for real.

leiflindholm: > This way acpiview could validate IndustrialStandard table header for real.

Yeah, that looks like something we would want to do across the board. So would be good to do start doing it at least for new additions.

pierregondois: Yes nice, it is better to avoid hard coded value. (+ tables are used for parsing)

samimujawar: @leiflindholm the original intent behind the current approach in acpiview was that the first three fields of the ACPI_PARSER—Field, Byte Length, and Byte Offset—would be taken directly from the ACPI table specification. For example, these correspond to the first three columns in the MADT table definition: https://uefi.org/specs/ACPI/6.6/05_ACPI_Software_Programming_Model.html#multiple-apic-description-table-madt
This approach helps verify that the offsets in the specification are correct, since the parser would flag an error if there were any mismatches.

I’m not sure if this is still considered useful, but it has definitely helped in the past to catch issues. 
Personally, I’d be inclined to keep this scheme if possible.

leiflindholm: @samimujawar I'm not sure I understand.
Are you saing it's beneficial to hard-code the values instead of using the core definitions because people would be unlikely to get it wrong in the same way twice when manually transcribing?

samimujawar: Thinking about it again it is a maintenance overhead as one has to review that the first 3 fields have been populated correctly. If the field definitions from the structure are used, those definitions would need to reflect the spec, and if the size and offset mismatch it will get flagged by the parser.
So we can go with the proposed approach rather that hardcoding the values.

MdePkg/Include/IndustryStandard/Acpi66.h
krzysztof-drobinski-plastelinet: Commit message updated

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: done

ShellPkg/Library/UefiShellAcpiViewCommandLib/UefiShellAcpiViewCommandLib.uni
krzysztof-drobinski-plastelinet: done

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: @pierregondois Switched to ACPI_PARSERs

MdePkg/Include/IndustryStandard/RiscVIoMappingTable.h
```diff
@@ -0,0 +1,118 @@
...
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+  @par Reference(s):
+  - RISC-V IO Mapping Table (RIMT) Specification, Version v1.0
+    (https://github.com/riscv-non-isa/riscv-acpi-rimt/releases/download/v1.0/rimt-spec.pdf)
+  - Advanced Configuration and Power Interface (ACPI) specification, Version 6.6
+    (https://uefi.org/specs/ACPI/6.6/)
+**/
+
+#ifndef __RISCV_IO_MAPPING_TABLE_H_
```
leiflindholm: Symbol names with leading underscore are reserved for toolchain use. (Please drop leading __).

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
```diff
@@ -0,0 +1,876 @@
...
+STATIC
+VOID
+EFIAPI
+RimtValidateNumberOfNodes (
+  UINT8   *Ptr,
+  UINT32  Length,
+  VOID    *Context
+  )
+{
+  UINT32  data; // "Number of RIMT Nodes" field is 4 byte long in spec
```
leiflindholm: 'Data', not 'data'.

leiflindholm: Although if this is a variable to hold specifically the number of nodes, it probably ought to be called NumNodes or something more descriptive.

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
```diff
@@ -0,0 +1,876 @@
...
+#include <Library/UefiLib.h>
+#include "AcpiParser.h"
+#include "AcpiTableParser.h"
+
+typedef enum {
+  IOMMU = 0,
+  PCIERC,
+  PLATFORM,
+  UNSUPPORTED
+} RimtNodeTypes;
```
pierregondois: Ideally this should go in the MdePkg as these values are spec-related

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
```diff
@@ -0,0 +1,876 @@
...
+  CONST CHAR16  *Format,
+  UINT8         *Ptr,
+  UINT32        Length
+  )
+{
+  CHAR8  *Str;
+  UINT8  NodeType;
+
+  NodeType = *(UINT8 *)Ptr;
+  if (NodeType == IOMMU) {
```
pierregondois: I think it would be easier to use an array of strings, cf. the `ErstActionTable` for instance

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
```diff
@@ -0,0 +1,876 @@
...
+
+  AsciiPrint (Str);
+}
+
+/**
+  Helper Macro for populating the RIMT Node header in the ACPI_PARSER array.
+  Every RIMT Node (IOMMU, PCIe Root Complex, PLatform Device) has this header
+  common
+**/
+STATIC UINT8                                    *RimtNodeType;
```
pierregondois: In general, static variables should be prefixed by `m`: `mRimtNodeType`
And also ideally be placed at the beginning of the file.
Same comment at other places

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
```diff
@@ -0,0 +1,876 @@
...
+
+
+/**
+An ACPI_PARSER array describing Parser for the RIMT node header structure.
+**/
+STATIC CONST ACPI_PARSER  RimtNodeHeaderParser[] = {
+  RIMT_PARSE_NODE_HEADER
+};
+
+#define RIMT_MINIMAL_NUMBER_OF_NODES_ALLOWED  2
```
pierregondois: Similarly, might move the definition to the MdePkg.
Please move all the possible RIMT definitions to the MdePkg

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
```diff
@@ -0,0 +1,876 @@
...
+
+/**
+  This macro generates function definition which traces bits in Flags field.
+  If no format string is specified the Format must be NULL.
+
+  @param [in] Parser    Optional format string for tracing the data.
+  @param [in] Active    String to display if interpreted bit is active.
+  @param [in] Inactive  String to display if interpreted bit is inactive.
+**/
+#define RIMT_DECLARE_FLAGS_BIT_PARSER_FUNC(Parser, Active, Inactive) \
```
pierregondois, line 255: This seems quite complex to have a mapping:
0 -> Inactive
1 -> Active

Would it be ok to just print the bit value ?

If no, would it possible instead to define an generic abstraction in:
`ShellPkg/Library/UefiShellAcpiViewCommandLib/AcpiParser.c` using a similar logic as in the `MadtParser.c::DumpValue()` so that other parsers can re-use it ?
Tagging @leiflindholm just in case as what I m suggesting is slightly questionable.

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
```diff
@@ -0,0 +1,876 @@
...
+  }
+};
+
+/**
+  This macro generates function definition for parser tracing
+  Flags field of RIMT Node.
+
+  @param [in] Parser  An ACPI_PARSER for tracing the data.
+**/
+#define RIMT_DECLARE_NODE_FLAGS_PARSER_FUNC(Parser) \
```
pierregondois, line 222: This might be a bit over-engineering. I don't know what other people think

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: @pierregondois Not sure about that. It will make reading this simple code harder (by need of jumping to array of strings to see what's behind the `Str`). Additionally, we will loose relation to `RimtNodeTypes` here in favor of array index. Those strings are not used anywhere except this routine. What's the benefit?

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: @pierregondois 

> In general, static variables should be prefixed by m: mRimtNodeType

Ok

> And also ideally be placed at the beginning of the file.

Is that approach required by EDK2? IMHO, It makes code harder to read. Declaring (if possible) variables in places where they are going to be used for the first time simplifies reading.

leiflindholm: It is a general rule, yeah. Each variant has its own benefits, we picked this one.

https://tianocore-docs.github.io/edk2-CCodingStandardsSpecification/release-2.20/5_source_files/54_code_file_structure.html#54-code-file-structure

MdePkg/Include/IndustryStandard/RiscVIoMappingTable.h
krzysztof-drobinski-plastelinet: @leiflindholm Removed leading underscores

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: @leiflindholm  Renamed

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: @pierregondois Moved to RIMT header

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: @pierregondois Variables renamed and moved to the beginning of file.

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: @pierregondois Moved

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: @pierregondois Ok, but what's the alternative? Dropping these macros wii make many functions looking almost the same. Parser API doesn't allow to pass arguments for reusability purposes.

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: @pierregondois Do you think this comment can be resolved?

krzysztof-drobinski-plastelinet:
@pierregondois @leiflindholm What's the policy for resolving comments? Shall I resolve addressed comments by  myself or should wait for you guys?

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
```diff
@@ -0,0 +1,865 @@
...
+// Local variables
+STATIC UINT8                                               *mRimtNodeType;
+STATIC UINT16                                              *mRimtNodeLength;
+STATIC EFI_ACPI_6_6_RIMT_NODE_HEADER_STRUCTURE             mRimtNodeHeader;
+STATIC EFI_ACPI_6_6_RIMT_STRUCTURE                         mRimtInfo;
+STATIC CONST UINT32                                        *mNumberOfRimtNodes;
+STATIC CONST UINT32                                        *mOffsetToRimtNodeArray;
+STATIC EFI_ACPI_6_6_RIMT_ID_MAPPING_STRUCTURE              mRimtIdMappingNode;
+STATIC CONST UINT16                                        *mNumberOfIdMappings;
+STATIC UINT8                                               *IdMappingArrayOffset;
```
leiflindholm: Needs the m prefix like the others.

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
```diff
@@ -0,0 +1,865 @@
...
+STATIC
+VOID
+EFIAPI
+RimtValidateNumberOfNodes (
+  UINT8   *Ptr,
+  UINT32  Length,
+  VOID    *Context
+  )
+{
+  UINT32  NumNodes; // "Number of RIMT Nodes" field is 4 byte long in spec
```
leiflindholm: With the rename, the comment is a bit superfluous.

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
leiflindholm: (Apologies for slow responses, a bunch of us are attending the UEFI devcon and osfc.io event this week.)

I'm not entirely opposed to simply printing the bit, although that loses some information.

But couldn't a single helper function be used for this instead of a function generation macro?
i.e.
```
RimtPrintFlagState (Ptr, "Active", "Inactive");
```

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
```diff
@@ -0,0 +1,865 @@
...
+  IN UINT8   *Ptr,
+  IN UINT32  Length,
+  IN VOID    *Context
+  )
+{
+  UINT8  NodeType = *(UINT8 *)Ptr;
+
+  if (NodeType >= UNSUPPORTED) {
+    IncrementErrorCount ();
+    Print (L"\nERROR: Urecognized RIMT node type. Type value must be less than %d but got %d", UNSUPPORTED, NodeType);
```
pierregondois: NIT: Unrecognized

MdePkg/Include/IndustryStandard/RiscVIoMappingTable.h
```diff
@@ -0,0 +1,133 @@
...
+
+#define EFI_ACPI_6_6_RIMT_STRUCTURE_VERSION  1
+
+#pragma pack()
+
+///
+/// RIMT node structure types as defined in RIMT spec, Chapter 2.1
+///
+typedef enum {
+  IOMMU = 0,
```
pierregondois: I think the enum names should be camelcased + ideally be less generic.
Maybe: `RimtNodeIommu` for instance ?

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
```diff
@@ -0,0 +1,865 @@
...
+  AsciiPrint (Buffer);
+  Print (L" (0x%lx)", *HardwareId);
+}
+
+/**
+  This macro calls generated function which traces bits in Flags field
+  defined by DECLARE_FLAG_INTERPRETER()
+**/
+
+#define RIMT_FLAGS_IOMMU_TYPE_BIT_OFFSET              0
```
pierregondois: Would it be possible to also move these definitions to the MdePkg ?
Same comment for the `RIMT_ID_MAPPING_*` definitions

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: Not sure how I could pass suggested `RimtPrintFlagState (Ptr, "Active", "Inactive")` to ACPI_PARSER

The issue is that it could be also 

`RimtPrintFlagState (Ptr, "Required", "Not Required");` 

or

`RimtPrintFlagState (Ptr, "PCIe", "Platform");` 

what means we will still get many almost identical  `_RimtNodeFlagsParser`s, like

```
RIMT_DECLARE_FLAGS_BIT_PARSER_FUNC (IommuType, "PCIe", "Platform");
RIMT_DECLARE_FLAGS_BIT_PARSER_FUNC (ProximityDomain, "Valid", "Invalid");
RIMT_DECLARE_FLAGS_BIT_PARSER_FUNC (PcieRcAtsSupport, "Supported", "Not Supported");
RIMT_DECLARE_FLAGS_BIT_PARSER_FUNC (PcieRcPriSupport, "Supported", "Not Supported");

```

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
```diff
@@ -0,0 +1,865 @@
...
+EFIAPI
+RimtParseIdMappingArray (
+  IN UINT8   *Node,
+  IN UINT16  Length,
+  IN UINT32  NumberOfIdMappings
+  )
+{
+  UINT32  Index;
+  UINT32  Offset;
+  CHAR8   Buffer[32]; // Used for AsciiName param of ParseAcpi
```
pierregondois: Effectively it should be ok to remove these comments as the functions are short enough.

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
pierregondois: Effectively, I find it more readable, but it should not really be that important

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
pierregondois: Unfortunately we need to populate a parser array with these functions, so input arguments cannot be provided.

I'm not a fan of the macro solution as having a mapping: `id -> string` to print seems to be a generic issue (cf. the Mpam parser for instance), but I assume this is better than extending the `ACPI_PARSER` for now ...

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
pierregondois: Similar comment as for the `RIMT_DECLARE_FLAGS_BIT_PARSER_FUNC()` macro: the issue seems generic enough to create a generic solution, but this would require to extend the `ACPI_PARSER`. So I it might be effectively better to do what you suggest for now

MdePkg/Include/IndustryStandard/RiscVIoMappingTable.h
krzysztof-drobinski-plastelinet: Ok, renamed

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: yes, prefix added

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: fixed

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: comment removed

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: Sure, definitions moved, naming unified

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: comment removed

krzysztof-drobinski-plastelinet:
@leiflindholm What about your comments, are you ok with the changes?

lgao4 (approved):

lgao4:

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c
krzysztof-drobinski-plastelinet: @pierregondois This was related to moving to table approach. Can we close this one?

pierregondois: Hi Krzysztof,
I think it should be ok to close the comments and merge the PR. I don't see the 'resolve' button anymore.
If the other persons approve, I assume it can be merged

krzysztof-drobinski-plastelinet: Ok, resolving

PR #11544, opened 2025-09-22
MdeModulePkg: DxeMain: Honor memory bin during memory map merging

kuqin12:
# Description

The current memory map merging logic in `CoreAddRange` only checks the memory type without considering the bin type.

However, the `CoreGetMemoryMap` routine will only report the entries that are entirely inside a specific bin type. This will confuse the consumer and report false free memory that belongs to special memory buckets.

This change fixes the issue by gating the merging logic by checking the bin type first. The merging is only allowed either if the bin type is not one of the special types, or if the bin type is the same for both entries.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This change was tested on both virtual platform and proprietary hardware platforms and ensure that the check asserts no longer trip.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/DxeMain.h
+++ b/MdeModulePkg/Core/Dxe/DxeMain.h
@@ -2804,4 +2804,22 @@ CoreInitializeHandleServices (
   VOID
   );
 
+/**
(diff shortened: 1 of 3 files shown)


kuqin12:
Closing this change as this solution will waste memory

PR #11546, opened 2025-09-23
UefiCpuPkg/CpuDxe: fix page table walk in confidential VM

luigix25:
`PageStartAddress` variable was not set correctly because the encryption bit was not considered, which broke the page walk logic.

This resulted in a Page Fault when booting a CVM when PcdNullPointerDetectionPropertyMask was set.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on SNP and TDX.

## Integration Instructions

### DIFF
--- a/UefiCpuPkg/CpuDxe/CpuPageTable.c
+++ b/UefiCpuPkg/CpuDxe/CpuPageTable.c
@@ -975,6 +975,7 @@ RefreshGcdMemoryAttributesFromPaging (
   UINT64                           Capabilities;
   UINT64                           NewAttributes;
   UINTN                            Index;
+  UINT64                           AddressEncMask;
 
   //
   // Assuming that memory space map returned is sorted already; otherwise sort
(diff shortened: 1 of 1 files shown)


luigix25:
Ping :)

luigix25:
@ardbiesheuvel could you help me with this PR?

UefiCpuPkg/CpuDxe/CpuPageTable.c
```diff
@@ -1055,10 +1056,18 @@ RefreshGcdMemoryAttributesFromPaging (
...
         }
 
+        //
+        // Get the mask for the memory encryption bit for Tdx and Sev
+        //
+        AddressEncMask = PcdGet64 (PcdPteMemoryEncryptionAddressOrMask);
+        if (AddressEncMask == 0) {
+          AddressEncMask = PcdGet64 (PcdTdxSharedBitMask);
+        }
+
```
ardbiesheuvel, line 1066: I am not a fan of adding SEV and TDX specific tweaks all over the code, so going forward, I would prefer to see some kind of library abstraction to deal with encryption bits in the page table descriptors.

For now, given that this fixes a bug, I'll accept it.

luigix25: Ok, thanks!

PR #11549, opened 2025-09-23
Add CLANGPDB and CLANGDWARF support to UnitTestFrameworkPkg

mdkinney:
# Description

* Add CLANGPDB and CLANGWARF support to UnitTestFrameworkPkgHost.dsc.inc and UnitTestFrameworkPkgCommon.dsc.inc and 
* Update GoogleTestLib for CLANG with special implementation of EXPECT_THROW_MESSAGE() and ASSERT_THROW_MESSAGE() macros to address compatibility issues between GoogleTest and ASAN on Windows.
* Update CmockaLib CC_FLAGS for CLANG
* Update SubhookLib CC_FLAGS for CLANG
* Remove redundant MSFT CC_FLAGS options from UnitTestDebugAddressLib and GoogleTestLib
* Update HostBasedUnitTestRunner to generate merged Cobertura coverage.xml output file using CLANG code coverage data    from -fprofile-instr-generate -fcoverage-mapping options.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Host based unit test builds using CLANGPDB and CLANGDWARF failed before this change and pass after this change.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
+++ b/BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
@@ -61,6 +61,10 @@ class HostBasedUnitTestRunner(IUefiBuildPlugin):
         # Set up the reporting type for Cmocka.
         shell_env.set_shell_var('CMOCKA_MESSAGE_OUTPUT', 'xml')
 
+        # Configure LLVM code coverage collection to generate unique file for
+        # each host based unit test executable.
+        shell_env.set_shell_var('LLVM_PROFILE_FILE', '%m.profraw')
+
         for arch in thebuilder.env.GetValue("TARGET_ARCH").split():
             logging.log(edk2_logging.get_subsection_level(),
                         "Testing for architecture: " + arch)
@@ -138,8 +142,12 @@ class HostBasedUnitTestRunner(IUefiBuildPlugin):
                                         failure_count += 1
 
             if thebuilder.env.GetValue("CODE_COVERAGE") != "FALSE":
-                if thebuilder.env.GetValue("TOOL_CHAIN_TAG") == "GCC5":
+                if thebuilder.env.GetValue("TOOL_CHAIN_TAG").startswith("GCC"):
                     ret = self.gen_code_coverage_gcc(thebuilder)
+                    if ret != 0:
+                        failure_count += 1
+                elif thebuilder.env.GetValue("TOOL_CHAIN_TAG").startswith("CLANG"):
+                    ret = self.gen_code_coverage_clang(thebuilder)
                     if ret != 0:
                         failure_count += 1
                 elif thebuilder.env.GetValue("TOOL_CHAIN_TAG").startswith ("VS"):
@@ -147,7 +155,7 @@ class HostBasedUnitTestRunner(IUefiBuildPlugin):
                     if ret != 0:
                         failure_count += 1
                 else:
-                    logging.info("Skipping code coverage. Currently, support GCC and MSVC compiler.")
+                    logging.info("Skipping code coverage. Currently, support GCC, CLANG, and MSVC compiler.")
 
         return failure_count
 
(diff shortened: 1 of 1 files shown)


gguo11837463 (approved):

UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
```diff
@@ -69,22 +102,40 @@
...
   #
-  # Enable Address Sanitizer for GCC for HOST_APPLICATION
+  # Set CLANGPDB HOST_APPLICATION options
   #
-  GCC:*_*_*_DLINK_FLAGS    = -fsanitize=address
+  CLANGPDB:*_*_*_DLINK_FLAGS   == /nologo /SUBSYSTEM:CONSOLE /DEBUG /out:"$(BIN_DIR)\$(MODULE_NAME_GUID).exe" /pdb:"$(BIN_DIR)\$(MODULE_NAME_GUID).pdb" /WHOLEARCHIVE
+  CLANGPDB:*_*_X64_DLINK_FLAGS  = /LIBPATH:"%VCToolsInstallDir%lib\x64" /LIBPATH:"%UniversalCRTSdkDir%lib\%UCRTVersion%\ucrt\x64" /LIBPATH:"%WindowsSdkDir%lib\%WindowsSDKLibVersion%\um\x64"
+!if $(UNIT_TESTING_ADDRESS_SANITIZER_ENABLE)
+  CLANGPDB:*_*_X64_DLINK_FLAGS  = clang_rt.profile-x86_64.lib clang_rt.asan_dynamic-x86_64.lib clang_rt.asan_dynamic_runtime_thunk-x86_64.lib
 !endif
```
makubacki, line 139: Were `IA32` libraries intentionally not set here? IA32 still seems supported in the rest of the build options.

makubacki:
nit: There is a typo in this commit message that you could clean up before completing the PR.

```
UnitTestFrameworkPkg/CmockaLib: Add CLANGPDB support

Add CLANGPDB family defines to match MSFT family defines.
Also add -Wno-deprecated-declarations
to CLANGPDB family
to ignore warnings detected wen building cmocka sources
with clang compiler.
```

"wen" -> "when"

UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
mdkinney: The libs listed in this line are only or X64 so that line is correct.

I do not know the correct settings for IA32 CLANGPDB host unit test builds yet, and stuart build of unit tests only allow X64 due to other IA32 related issues.  We can leave IA32 CLANGPDB host unit test build support as an open issue for now.

mdkinney:
> nit: There is a typo in this commit message that you could clean up before completing the PR.
> 
> ```
> UnitTestFrameworkPkg/CmockaLib: Add CLANGPDB support
> 
> Add CLANGPDB family defines to match MSFT family defines.
> Also add -Wno-deprecated-declarations
> to CLANGPDB family
> to ignore warnings detected wen building cmocka sources
> with clang compiler.
> ```
> 
> "wen" -> "when"

Updated PR with this fixed.

mdkinney:
@Javagedes Do you have any comments on the BaseTools plugin changes?

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -224,6 +232,125 @@ def gen_code_coverage_gcc(self, thebuilder):
...
+            testList = glob.glob(os.path.join(buildOutputBase, "**", "*Test*"), recursive=True)
+            testList = [f for f in testList if os.path.isfile(f) and os.path.splitext(f)[1] == ""]
+        elif GetHostInfo().os.upper() == "WINDOWS":
+            # Collect test executables with a .exe file extension
+            testList = glob.glob(os.path.join(buildOutputBase, "**","*Test*.exe"), recursive=True)
+        else:
+            raise NotImplementedError("Unsupported Operating System")
+        if not testList:
+            logging.warning("UnitTest Coverage: No test binaries found.")
+            return 1
```
Javagedes: I'm not against returning an error for no tests at this point in the plugin. That is because in `do_post_build` we already validate some tests exist. If we don't find any now, that would be different than what we already determined above and is definitely an error.

However, by doing so, we are creating a distinct different between `gen_code_coverage_clang` and `gen_code_coverage_gcc` / `gen_code_coverage_msvc` as they do not return errors when their test globs return nothing (`testCoverageList` for `gen_code_coverage_gcc` and `testList` for `gen_code_coverage_msvc`.

I do think returning an error is the correct thing to do. But this means we need to update `gen_code_coverage_gcc` and `gen_code_coverage_msvc`. That seems outside the scope of this PR, so it seems more appropriate to set this to `return 0` here and then update all three in a new PR, but I'm open to other ways of handling this situation.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -224,6 +232,125 @@ def gen_code_coverage_gcc(self, thebuilder):
...
+                    for package in packages.findall("package"):
+                        base_packages.append(package)
+
+            # Recalculate line-rate and branch-rate for the merged XML
+            lines_covered = 0
+            lines_valid = 0
+            branches_covered = 0
+            branches_valid = 0
+
+            for package in base_packages.findall("package"):
```
Javagedes, line 329: This is quite the nesting. I would try my best to reduce this as much as possible. This is something I was able to come up with. Warning: This is untested, just from eyeballing:

```python
def parse_condition_coverage(cond_cov: str) -> int:
    """
    Parse a condition-coverage string like '50% (1/2)'.
    Returns the number of covered branches.
    """
    try:
        inside_parens = cond_cov.split("(", 1)[1].split(")", 1)[0]  # "1/2"
        covered, total = map(int, inside_parens.split("/"))
        return covered
    except Exception:
        return 0

def merge_covertura_xml_files(self, xml_file_list, output_file):
    ...
    for package in base_packages.findall("package"):
        for packageclass in package.findall("classes/class"):
            lines = packageclass.find("lines")
            if lines is None:
                continue

            for line in lines.findall("line"):
                # Line coverage
                hits = line.attrib.get("hits")
                if hits and int(hits) > 0:
                    lines_covered += 1
                lines_valid += 1

                # Branch Coverage
                if line.attrib.get("branch") != "true":
                    continue
                cond_cov = line.attrib.get("condition-coverage")
                if cond_cov:
                    branches_covered += parse_condition_coverage(cond_cov)
```

Javagedes (approved):
Overall looks good. Left a few comments. Approving with the assumption you will resolve them as necessary.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
mdkinney: I agree and will update PR to return 0.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
mdkinney: Updated this this suggestion

Javagedes: This change looks good to me. The only thing you _might_ consider is changing the condition-coverage logic to just be the below, as (If I did the character count right) stop you from needing to do the 3 line wrapping call of `parse_condition_coverage)`.

It's purely cosmetic though, so feel free to ignore this suggestion.

``` python
cond_cov = line.attrib.get("condition-coverage") or "0% (0/0)"
branches_covered += parse_condition_coverage(cond_cov)
```

Javagedes: Please feel free to close this particular comment after reviewing. I will approve, as this looks good to me whether you make the change or not.

Javagedes (approved):

UnitTestFrameworkPkg/Include/Library/GoogleTestLib.h
```diff
@@ -9,9 +9,20 @@
 #ifndef GOOGLE_TEST_LIB_H_
 #define GOOGLE_TEST_LIB_H_
 
+//
+// For Windows/clang builds if _MSC_VER is not defined, then set to
+// Visual Studio 2015 value of 1900
+//
+#if defined (__clang__) && defined (_WIN32)
+  #ifndef _MSC_VER
+#define _MSC_VER  1900
```
jaykrell, line 18: Is this really a good idea?

mdkinney: It was not a good idea and was later removed by https://github.com/tianocore/edk2/pull/11914

jaykrell: Thank you, that is better. I wonder if we should go further? I haven't looked, I admit, but.. seems messy to be definining, undefining... but I guess `__clang__` + `_MSC_VER` is now a deliberate construct, meaning definitely clang, with some strong albeit flawed or non-ideal degree of Msvc compat?

I now find myself writing this:
```
if __clang__ || __GNUC__
..
elif _MSC_VER
..
else
error unknown compiler, or write portable code here
endif
```

while previously I was writing:
```
if _MSC_VER // definitely only Msvc
..
elsif __clang__ || __GNUC__
..
else
error unknown compiler, or portable code
endif
```

I'll stop doing that, and fixed a few cases.

PR #11550, opened 2025-09-24
MdeModulePkg/Ps2KeyboardDxe: use MMIO when PcdPciIoTranslation is set

MarsDoge:
On IA32/X64, legacy port I/O is available and IoRead/IoWrite works as-is. On LOONGARCH/ARM/RISC-V there is no port I/O, and platforms usually expose a translated MMIO window for legacy I/O ranges.

This change reads PcdPciIoTranslation and, when non-zero, performs MMIO accesses at (Translation + Port) instead of using IoRead/IoWrite. When the PCD is zero, behavior is unchanged and IoRead/IoWrite is used.

Implementation details:
- Add small helpers to select MmioRead*/MmioWrite* vs IoRead*/IoWrite*.
- Use FixedPcdGet64(PcdPciIoTranslation) when possible to avoid runtime PCD reads in hot paths.
- Update all keyboard status/data accessors to the new helpers.
- Wire up INF: add Pcd and LibraryClasses entries.

This keeps x86 behavior identical and enables the driver to work on LOONGARCH/ARM/RISC-V platforms that map legacy I/O through MMIO.

Cc: Chao Li <lichao@loongson.cn>
Cc: Gao Qihang <gaoqihang@loongson.cn>
Signed-off-by: Dongyan Qian <qiandongyan@loongson.cn>
Tested-by: Chen Zhang <zhangchen@loongson.cn>

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions

### DIFF
--- a/MdeModulePkg/Bus/Isa/Ps2KeyboardDxe/Ps2KbdCtrller.c
+++ b/MdeModulePkg/Bus/Isa/Ps2KeyboardDxe/Ps2KbdCtrller.c
@@ -689,7 +689,19 @@ KeyReadDataRegister (
   )
 
 {
-  return IoRead8 (ConsoleIn->DataRegisterAddress);
+  UINT32  Port;
+  UINT64  Translation;
+  UINT64  MmioAddress;
+
+  Port        = ConsoleIn->DataRegisterAddress;
+  Translation = PcdGet64 (PcdPciIoTranslation);
+  if (Translation != 0) {
+    MmioAddress = Translation + Port;
+    ASSERT (MmioAddress >= Port);
+    return MmioRead8 ((UINTN)MmioAddress);
+  }
+
+  return IoRead8 (Port);
 }
 
 /**
@@ -705,7 +717,19 @@ KeyWriteDataRegister (
   IN UINT8                    Data
   )
 {
-  IoWrite8 (ConsoleIn->DataRegisterAddress, Data);
+  UINT32  Port;
+  UINT64  Translation;
+  UINT64  MmioAddress;
+
+  Port        = ConsoleIn->DataRegisterAddress;
+  Translation = PcdGet64 (PcdPciIoTranslation);
+  if (Translation != 0) {
+    MmioAddress = Translation + Port;
+    ASSERT (MmioAddress >= Port);
+    MmioWrite8 ((UINTN)MmioAddress, Data);
+  } else {
+    IoWrite8 (Port, Data);
+  }
 }
 
 /**
@@ -721,7 +745,19 @@ KeyReadStatusRegister (
   IN KEYBOARD_CONSOLE_IN_DEV  *ConsoleIn
   )
 {
-  return IoRead8 (ConsoleIn->StatusRegisterAddress);
+  UINT32  Port;
+  UINT64  Translation;
+  UINT64  MmioAddress;
+
+  Port        = ConsoleIn->StatusRegisterAddress;
+  Translation = PcdGet64 (PcdPciIoTranslation);
+  if (Translation != 0) {
+    MmioAddress = Translation + Port;
+    ASSERT (MmioAddress >= Port);
+    return MmioRead8 ((UINTN)MmioAddress);
+  }
+
+  return IoRead8 (Port);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


mdkinney:
This approach potentially moves this handling into many drivers.

I still do not understand why the logic proposed in this PR can not be moved into an IoLib instance.

For example, the IoLib implementation in MdePkg\Library\BaseIoLibIntrinsic could add an extra INF file for the platforms that set gEfiMdePkgTokenSpaceGuid.PcdPciIoTranslation to a non zero value.  This new INF would not use IoLibNoIo.c.  Instead, it would use a new file that uses PcdPciIoTranslation in the exact same way as the change proposed in the Ps2KeyboardDxe module to map the IoRead/Writexx() services to an MmioRead/Writexx() services with a translation of PcdPciIoTranslation.  That would be a single change that would allow drivers like Ps2KeyboardDxe to work with no modifications.  The Platform DSC would select this new IoLib instance for modules like the Ps2KeyboardDxe.

MarsDoge:
As far as I know, on the ARM platform, the Ps2 driver has been independently added to the interface through the override method, and there is no IoRead/Write operation in the form of CPU-side IO. This PcdPciIoTranslation address may be an LPC address port, not the so-called IO space. So I think it is not reasonable to add base IO support to BaseIoLibIntrinsic. @mdkinney

mdkinney:
Does this mean that the following usages of this PCD are incorrect because they use this PCD for all I/O Port ranges.  Not a subset.

https://github.com/tianocore/edk2/blob/17691a2641c2715fc1196aa1b0fe85b85db6c240/UefiCpuPkg/CpuMmio2Dxe/CpuMmio2Dxe.c#L406
https://github.com/tianocore/edk2/blob/17691a2641c2715fc1196aa1b0fe85b85db6c240/UefiCpuPkg/CpuMmio2Dxe/CpuMmio2Dxe.c#L492
https://github.com/tianocore/edk2/blob/17691a2641c2715fc1196aa1b0fe85b85db6c240/ArmPkg/Drivers/ArmPciCpuIo2Dxe/ArmPciCpuIo2Dxe.c#L405
https://github.com/tianocore/edk2/blob/17691a2641c2715fc1196aa1b0fe85b85db6c240/ArmPkg/Drivers/ArmPciCpuIo2Dxe/ArmPciCpuIo2Dxe.c#L491

MarsDoge:
Yes, I understand that PCI I/O should be regarded as a subset of CPU I/O. In architectures that employ a unified addressing scheme rather than independent addressing, not only PCI devices but also other devices expose an I/O space that is accessible to the CPU through MMIO. For example: non-PCI mounted LPC device.
It’s just that most of the IOs are currently mounted under PCI, and can be accessed through a subset of PCIIO.

MarsDoge:
@ardbiesheuvel Hi, as far as I know, ARM is facing the same problem. I would like to know your thoughts on this.

ardbiesheuvel:
> Does this mean that the following usages of this PCD are incorrect because they use this PCD for all I/O Port ranges. Not a subset.
> 
> https://github.com/tianocore/edk2/blob/17691a2641c2715fc1196aa1b0fe85b85db6c240/UefiCpuPkg/CpuMmio2Dxe/CpuMmio2Dxe.c#L406
> 
> 
> https://github.com/tianocore/edk2/blob/17691a2641c2715fc1196aa1b0fe85b85db6c240/UefiCpuPkg/CpuMmio2Dxe/CpuMmio2Dxe.c#L492
> 
> 
> https://github.com/tianocore/edk2/blob/17691a2641c2715fc1196aa1b0fe85b85db6c240/ArmPkg/Drivers/ArmPciCpuIo2Dxe/ArmPciCpuIo2Dxe.c#L405
> 
> 
> https://github.com/tianocore/edk2/blob/17691a2641c2715fc1196aa1b0fe85b85db6c240/ArmPkg/Drivers/ArmPciCpuIo2Dxe/ArmPciCpuIo2Dxe.c#L491

Indeed. `PcdPciIoTranslation` is problematic because it assumes that all PCIe host bridges in the system use the same translation for port I/O, which is not very likely in practice.

As pointed out, on non-x86 systems, only PCI provides port I/O to begin with, so this is not a huge issue in practice, given how rarely port I/O is used in PCIe systems, even on legacy endpoints. Only in cases such as this one, where legacy peripherals such as PS/2 or serial ports on PCIe expansion cards are used do we actually run into these problems.

ardbiesheuvel:
> @ardbiesheuvel Hi, as far as I know, ARM is facing the same problem. I would like to know your thoughts on this.

You should be using `MdePkg/Library/DxeIoLibCpuIo2/DxeIoLibCpuIo2.inf` with this driver, and ensure that your implementation of `gEfiCpuIo2ProtocolGuid` does the right thing.

MarsDoge:
> You should be using `MdePkg/Library/DxeIoLibCpuIo2/DxeIoLibCpuIo2.inf` with this driver, and ensure that your implementation of `gEfiCpuIo2ProtocolGuid` does the right thing.

@ardbiesheuvel I totally agree with your idea, because of this, my option is to allow PS/2 to call IoLib through the `DxeIoLibCpuIo2` protocol at first, so I submitted the PR https://github.com/tianocore/edk2/pull/11414, but @mdkinney  told me that CPU I/O Protocols belongs to the PI specification, not the UEFI specification, so it cannot be used to link the UEFI PS/2 driver .

PR #11557, opened 2025-09-25
MdeModulePkg: DXE Core: Correct Usage of EFI_MEMORY_ATTRIBUTE_MASK

os-d:
# Description

edk2 commit 3bd5c994c879f78e8e3d5346dc3b627f199291aa added usage of EFI_MEMORY_ATTRIBUTE_MASK to edk2. However, it applied it incorrectly to some places that should instead use EFI_MEMORY_ACCESS_MASK. EFI_MEMORY_ACCESS_MASK contains the actual HW page table access attributes (read protect, read only, no-execute), whereas EFI_MEMORY_ATTRIBUTE_MASK contains the access attributes in addition to some virtual attributes (special purpose and cpu crypto).

The GCD has a behavior where if SetMemorySpaceAttributes() is called with only virtual attributes set, it will not call into CpuDxe to change the attributes; 0 is a valid page table attribute set (it means RWX). However, after the above change, this behavior was altered so that if EFI_MEMORY_SP or EFI_MEMORY_CPU_CRYPTO is applied, in attempt to just update these virtual attributes, the GCD will call into CpuDxe and apply RWX instead, which is not the intention of the caller.

One other place this was done incorrectly was in CoreGetMemoryMap, but that was fixed in f1567720b13a578ffa54716119f826df622babcd.

SetUefiImageMemoryAttributes() is also updated here because that logic was copied from the check the GCD has about whether to call CpuDxe or not. Now that the GCD has been corrected, this also needs to be corrected.

- [ ] Breaking change?
- [x] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Applying EFI_MEMORY_SP or EFI_MEMORY_CPU_CRYPTO only in Ovmf no longer sets that range to be RWX.

## Integration Instructions

N/A.

### DIFF
--- a/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
+++ b/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
@@ -692,7 +692,7 @@ ConverToCpuArchAttributes (
 {
   UINT64  CpuArchAttributes;
 
-  CpuArchAttributes = Attributes & EFI_MEMORY_ATTRIBUTE_MASK;
+  CpuArchAttributes = Attributes & EFI_MEMORY_ACCESS_MASK;
 
   if ((Attributes & EFI_MEMORY_UC) == EFI_MEMORY_UC) {
     CpuArchAttributes |= EFI_MEMORY_UC;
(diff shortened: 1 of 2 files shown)


os-d:
@lgao4 can we merge this?

PR #11559, opened 2025-09-26
OvmfPkg/build.sh: Remove support for IA32 architecture

Damien-Chen:
Removes the logic for building the IA32 architecture from the `build.sh` script since AMD and Intel CPU are X64 only.

## How This Was Tested
Build QEMU with `build.sh`

## Integration Instructions
N/A


kraxel:
This is the mixed mode build (ia32 sec+pei, x64 dxe), support for that will stay in edk2, and so should the ovmf support (for CI purposes).

Damien-Chen:
Yes, and I know the script is used to build OvmfPkgIa32X64.dsc as well. But is these IA32 Architecture thing really needs to exist ?

kraxel:
As long as edk2 support SEC and PEI in IA32 we should continue to build the relevant code in CI, and OvmfPkgIa32X64.dsc does exactly that ...

Hmm, all CI passes, so apparently the script is not used for the CI builds.  Guess we can update the build.sh script then, so people stop using OvmfPkgIa32X64.dsc (CI is really the only reason for this to exist, everything else should move on and use OvmfPkgX64.dsc instead).

kraxel (approved):

Damien-Chen:
Correct me if I am wrong.
The build.sh exist here as I know is a helper script to help developer build OvmfX64Pkg.dsc more conviniently.  And since modern systems only exist X64 and in the future. That is the reason why I modify the script to Drop for IA32 related since it`s no use anymore.

Issue #11364, opened 2025-07-31
[Bug]: Incorrect behavior when switching language in Front Page and Driver Health Manager


Sitplek:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

MdeModulePkg

### Which targets are impacted by this bug?

RELEASE

**Current Behavior**

The interface incorrectly updates after changing language through "Select Language":

1. On the main page **(Front Page)**: The **Continue** and **Reset** buttons immediately display in the selected language, however the **Device Manager**, **Boot Manager**, and **Boot Maintenance Manager** items remain in the previous language until reboot via Continue or Reset.

2. In **Driver Health Manager**: The form language switches correctly at first, but if you open the form with French language set and exit, on the next entry it displays in **English** again, despite the rest of the interface remaining in French.

3. On the main page **(Front Page)**: navigation and control help buttons like "=Move Highlight", "<Enter>=Select Entry" do not translate at all.

4. Dialog boxes like "ERROR", "WARNING", "INFO" do not switch to another language, and the "Yes" and "No" buttons also do not switch.

---

**Expected Behavior**

* All interface elements should immediately display in the selected language without reboot
* Driver Health Manager should preserve the selected language and not change it on its own

---

**Steps to Reproduce**

1. Open the `DeviceManagerStrings.uni` file
2. Change the `STR_EDKII_MENU_TITLE` variable:
   ```
   #language fr-FR  "Device Manager"
   ```
   to:
   ```
   #language fr-FR "Gestionnaire de Périphériques"
   ```

3. Open the `DriverHealthManagerStrings.uni` file
4. Change the `STR_FORM_TITLE` variable:
   ```
   #language fr-FR  "Driver Health Manager"
   ```
   to:
   ```
   #language fr-FR  "Gestionnaire d'Appareilsgestionnaire de santé du conducteur"
   ```

5. Open the `FormDisplayStr.uni` file
6. Change the `INFO_POPUP_STRING` variable:
   ```
   #language fr-FR  "INFO"
   ```
   to:
   ```
   #language fr-FR  "INFORMATION"
   ```

7. Open the `CustomizedDisplayLib.uni` file
   You don't need to change anything - it already has French language different from the English version

8. Save the modified files
9. Build the project
10. Run the built firmware through QEMU
11. In the "Please select boot device" menu, select "EFI Firmware Setup"
[...]

PR #11560, opened 2025-09-28
MdeModulePkg: Improve display when the language changes on Setup

EricGao2015:
# Description

This PR can partially solve https://github.com/tianocore/edk2/issues/11364.

- **MdeModulePkg/UiApp: Update front page form when the language changes**

In current code, when changing language on Setup front page, DeviceManager, BootManager and BootMaintenanceManager option will not change. This occurs because the PromptId and HelpId for these Goto options are set via the HiiSetString function, but this function will set string to all supported languages if language parameter is NULL. So, when language is changed and the string id is not changed, the three options still remain unchanged.
This patch adds update operation within the callback function of front page. Upon the language change, the PromptId and HelpId are refreshed to ensure that the corresponding option texts are correctly displayed in the newly selected language.

- **MdeModulePkg: Fix issue that French package is unexpectedly cleaned**

The DriverHealthManagerStrings.uni file defines both English and French languages, resulting in the generation of two Unicode string packages: the first for English and the second for French. When cleaning the dynamic strings from the HII package list, the pointer which should point to dynamic string package incorrectly points to the static French package, causing it to be unexpectedly cleaned.
This patch creates a default HiiPackageList that only contains static STRING_ARRAY_NAME data once the driver is initialized and directly updates this HiiPackageList in DriverHealthManagerCleanDynamicString function. This way, there is no need for concern regarding the number of language packages available.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

1. Change language from English to French, `Device Manager`, `Boot Manager ` and `Boot Maintenance Manager` options are also changed from English to French.
2. Change language from English to French, enter  `Device Manager` page, `Driver Health Manager` page, and then return to `Device Manager` page, the `Driver Health Manager` is shown as French.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Universal/DriverHealthManagerDxe/DriverHealthManagerDxe.c
+++ b/MdeModulePkg/Universal/DriverHealthManagerDxe/DriverHealthManagerDxe.c
@@ -47,6 +47,7 @@ EFI_HII_HANDLE                       mDriverHealthManagerHiiHandle;
 EFI_BOOT_MANAGER_DRIVER_HEALTH_INFO  *mDriverHealthManagerHealthInfo     = NULL;
 UINTN                                mDriverHealthManagerHealthInfoCount = 0;
 EFI_HII_DATABASE_PROTOCOL            *mDriverHealthManagerDatabase;
+EFI_HII_PACKAGE_LIST_HEADER          *mBackupHiiPackageList = NULL;
 
 extern UINT8  DriverHealthManagerVfrBin[];
 extern UINT8  DriverHealthConfigureVfrBin[];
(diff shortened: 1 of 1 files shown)


MdeModulePkg/Universal/DriverHealthManagerDxe/DriverHealthManagerDxe.c
```diff
@@ -898,6 +898,7 @@ DriverHealthManagerCleanDynamicString (
   PackageHeader = (EFI_HII_PACKAGE_HEADER *)(HiiPackageList + 1);
   CopyMem (PackageHeader, STRING_ARRAY_NAME + sizeof (UINT32), FixedStringSize);
 
+  PackageHeader         = (EFI_HII_PACKAGE_HEADER *)((UINT8 *)PackageHeader + PackageHeader->Length);
```
lgao4: This logic assumes two string packages in this driver. Once this driver adds new lanaguage support, this logic will be updated. Could you use the general logici to handle it？

EricGao2015: My idea is to create a default HiiPackageList that only contains static STRING_ARRAY_NAME data once the driver is initialized and directly update this HiiPackageList in `DriverHealthManagerCleanDynamicString` function. This way, computing pointer offset can be avoided. Can you accept this method？

EricGao2015: @lgao4  Hi Liming, I have updated the patch with the method mentioned above, please give some suggestions.

lgao4: Yes. Please try it.

EricGao2015: >> My idea is to create a default HiiPackageList that only contains static STRING_ARRAY_NAME data once the driver is initialized and directly update this HiiPackageList in DriverHealthManagerCleanDynamicString function. This way, computing pointer offset can be avoided. Can you accept this method？

> Yes. Please try it.

@lgao4 I have updated the new code with the method mentioned above, please re-review it.

MdeModulePkg/Application/UiApp/FrontPageCustomizedUiSupport.c
```diff
@@ -145,6 +146,8 @@ LanguageChangeHandler (
       FreePool (Lang);
       return EFI_DEVICE_ERROR;
     }
+
```
lgao4, line 149: For this problem, HiiSetString () should be called for each supported lanaguage in main page. The first call is to create StringId, the second call is to set different language.

PR #11561, opened 2025-09-29
BaseTools: Disable GCC relax on LoongArch

kilaterlee:
# Description

It might appear some odd problems if enable GCC relax. Disabling it can avoid these problems.

## How This Was Tested

Currently, it has not fixed reproduction method, just disable GCC relax to avoid the unknown errors.

## Integration Instructions

N/A.


kilaterlee:
@lgao4 @bexcran @gapalomi @ponchofigueroa @YuweiChen1110 Hi maintainers, please help to review this patch if you have time. :)

PR #11564, opened 2025-09-30
NetworkPkg/UefiPxeBcDxe: Add buffer check before reporting status code

MarcChen46:
# Description

When PxeBcLoadBootFile() fails, it reports a status code to notify listeners (such as telemetry/SEL logging drivers) about the error. However, the current implementation reports the status code for all error conditions, including benign case, such as where the caller passes Buffer == NULL with EFI_BUFFER_TOO_SMALL to query the required size.

This causes false positive error reports in telemetry systems, as EFI_BUFFER_TOO_SMALL with Buffer == NULL is an expected and normal operation for size queries, not an actual error condition.

The PXE-E05 error message is already correctly guarded with the condition "(Status == EFI_BUFFER_TOO_SMALL) && (Buffer != NULL)", but the status code reporting was unconditional.

Add a buffer null check before REPORT_STATUS_CODE_WITH_EXTENDED_DATA to ensure status codes are only reported for actual errors:
- Report when Status != EFI_BUFFER_TOO_SMALL (all other errors)
- Report when Status == EFI_BUFFER_TOO_SMALL AND Buffer != NULL (PXE-E05)
- Skip when Status == EFI_BUFFER_TOO_SMALL AND Buffer == NULL (size query)

This prevents spurious error logs while maintaining proper error reporting for genuine failures including PXE-E05, PXE-E07, PXE-E09, PXE-E99, and all other error conditions.

Signed-off-by: Marc Chen <marc.chen@microsoft.com>

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verified on a PXE boot pass flow and confirmed the status code will not be reported.

## Integration Instructions

N/A


MarcChen46:
Hi @SaloniKasbekar @makubacki , please review this PR and let me know if there is any feedback, thanks.

NetworkPkg/UefiPxeBcDxe/PxeBcBoot.c
```diff
@@ -1268,12 +1268,19 @@ PxeBcLoadBootFile (
     AsciiPrint ("\n  PXE-E99: Unexpected network error.\n");
   }
 
-  REPORT_STATUS_CODE_WITH_EXTENDED_DATA (
-    EFI_ERROR_CODE,
-    (EFI_STATUS_CODE_VALUE)(EFI_IO_BUS_IP_NETWORK | EFI_OEM_SPECIFIC | ((EFI_STATUS_CODE_VALUE)(Status & 0x1F))),
-    (VOID *)&(PxeBcMode->UsingIpv6),
-    sizeof (PxeBcMode->UsingIpv6)
-    );
+  //
```
leiflindholm, line 1271: Commit message: suggest rewording
"including the benign case where the" as
"including benign cases, such as where the"

NetworkPkg/UefiPxeBcDxe/PxeBcBoot.c
```diff
@@ -1268,12 +1268,19 @@ PxeBcLoadBootFile (
...
-    (EFI_STATUS_CODE_VALUE)(EFI_IO_BUS_IP_NETWORK | EFI_OEM_SPECIFIC | ((EFI_STATUS_CODE_VALUE)(Status & 0x1F))),
-    (VOID *)&(PxeBcMode->UsingIpv6),
-    sizeof (PxeBcMode->UsingIpv6)
-    );
+  //
+  // Report status code for all errors printed above.
+  // Skip reporting when Buffer is NULL and Status is EFI_BUFFER_TOO_SMALL,
+  // as this is an expected size query, not an error.
+  //
+  if ((Status != EFI_BUFFER_TOO_SMALL) || (Buffer != NULL)) {
```
leiflindholm, line 1276: This will still log an error if `Buffer` is non-NULL but too small, and I don't believe that case is any less valid. Unless I'm missing something, can we drop the NULL test?

MarcChen46: Hi @leiflindholm , Please check below positive and negative scenarios and let me know if you see concern, thanks.

Positive scenario: Below two condition will still invoke report status code.
Case1: `(Status == EFI_BUFFER_TOO_SMALL) && (Buffer != NULL)`
Case2: `(Status != EFI_BUFFER_TOO_SMALL)`

Negative scenario: Below condition will **NOT** invoke report status code
Case3: `(Status == EFI_BUFFER_TOO_SMALL) && (Buffer == NULL)`

NetworkPkg/UefiPxeBcDxe/PxeBcBoot.c
MarcChen46: Thanks, updated in the PR description field.

NetworkPkg/UefiPxeBcDxe/PxeBcBoot.c
leiflindholm: Right, so PXE-E05 is actually a defined PXE error condition?
Then I guess what I'm disagreeing with is the PXE spec rather than this commit.

NetworkPkg/UefiPxeBcDxe/PxeBcBoot.c
leiflindholm: Please also update the commit message.

MarcChen46: Thank you @leiflindholm, I have updated the commit message too.

PR #11567, opened 2025-09-30
ArmPkg,ArmPlatformPkg,SecurityPkg: rename *NullLib *LibNull

leiflindholm:
# Description

The edk2 convention is that null libraries have a name ending in LibNull.
A few libraries in ArmPkg and ArmPlatformPkg did not follow this convention,
so this PR fixes that.

Finally it also renames the just merged PeilessSecMeasureLibNull in SecurityPkg.

## How This Was Tested

Simple mechanical change: build tested only.

## Integration Instructions

If someone has explicitly referenced these NULL-variant libraries downstream, they may
need to update their build config, but I believe these are in reality only used for
package .dsc buildability.


leiflindholm:
Err, need to fix up the one in SecurityPkg as well.

LeviYeoReum (approved):

PR #11571, opened 2025-09-30
SecurityPkg: Trace messages are added.

arunsbaskaran:
Added debug trace messages on failure.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

### DIFF
--- a/BaseTools/Source/Python/FirmwareStorageFormat/FvHeader.py
+++ b/BaseTools/Source/Python/FirmwareStorageFormat/FvHeader.py
@@ -4,7 +4,6 @@
 # Copyright (c) 2021-, Intel Corporation. All rights reserved.<BR>
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 ##
-from ast import Str
 from struct import *
 from ctypes import *
 from FirmwareStorageFormat.Common import *
--- a/DynamicTablesPkg/DynamicTables.dsc.inc
+++ b/DynamicTablesPkg/DynamicTables.dsc.inc
@@ -79,9 +79,11 @@
       #   Common
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiCedtLib/AcpiCedtLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiFadtLib/AcpiFadtLib.inf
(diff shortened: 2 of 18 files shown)


SecurityPkg/Tcg/Tcg2PlatformDxe/Tcg2PlatformDxe.c
```diff
@@ -80,6 +81,10 @@ Tcg2PlatformDxeEntryPoint (
             );
 
   ASSERT (Event != NULL);
+  if (Event == NULL) {
```
luigix25, line 84: This condition (may) be always false in DEBUG according to the assert configuration.
This is not true in RELEASE. `Tcg2PlatformDxeEntryPoint` returns success regardless of `Event`.
So, IIUC you are handling an error case in this PR, but imho this is misleading and the commit message
doesn't match what you are doing.

arunsbaskaran: Missed your comment. Addressed your feedback, please check it.

jyao1: I don't understand the logic.
We have ASSERT(Event != NULL), which means Event could NOT be NULL...

If you think it could be NULL, please remove ASSERT.

arunsbaskaran: ASSERT (Event != NULL); will be helpful only in DEBUG mode.
In release mode this ASSERT is dummy.

if (Event == NULL) { -> this will be helpful in RELEASE mode

luigix25: Yeah, there is no need to have both an assert and a check if event is NULL. 

Because the asserts are removed in the release build, the if is enough.

arunsbaskaran:
Did some mistakes in git commands. So, closing it. I will open a new one.

arunsbaskaran:
New PR
https://github.com/tianocore/edk2/pull/11652

PR #11575, opened 2025-10-02
ArmPkg/Library: change transfer list's checksum calculation

LeviYeoReum:
# Description

Since the commit ae94bd1ff5b5 ("Switch to an xor for the checksum") in
firmware handoff specification [0],
the checksum calculation method is changed
from additive sum to XOR in 2023.

To align with this specification,
TF-A updates the method in last month by:
  commit e2fc189e1bd1 ("fix: use XOR for checksum") in libtl [1].

Before the above commit, ArmTransferListLib is implemented with
additive sum for checksum to align with TF-A once.

As transfer list library used by TF-A updates the
calculation method for checksum according to the specification,
let's update the checksum calculation method in ArmTransferListLib.

Link: https://github.com/FirmwareHandoff/firmware_handoff/commit/ae94bd1ff5b5d4a1024782934df8299b4c8d097b [0]
Link: https://review.trustedfirmware.org/plugins/gitiles/shared/transfer-list-library/+/1712685fdfaa0fa8887026a77568d2039a0c37f5 [1]

- [X] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

With the latest TF-A, test on booting with transfer list in FVP RevC model.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
+++ b/ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
@@ -16,6 +16,45 @@
 #include <Library/HobLib.h>
 
 /**
+  Returns the xor of all elements in a buffer in unit of UINT8.
+  During calculation, the carry bits are dropped.
+
+  This function calculates the xor value of all elements in a buffer
+  in unit of UINT8 start from Zero.
+  The result is returned as UINT8. If Length is Zero, then Zero is
+  returned.
+
+  If Buffer is NULL, then ASSERT().
+  If Length is greater than (MAX_ADDRESS - Buffer + 1), then ASSERT().
+
+  @param[in]  Buffer      The pointer to the buffer to carry out the xor operation.
+  @param[in]  Length      The size, in bytes, of Buffer.
+
+  @return CheckSum    The xor of Buffer calculated from Zero.
+
+**/
+STATIC
+UINT8
+EFIAPI
+CalculateXor8 (
+  IN      CONST UINT8  *Buffer,
+  IN      UINTN        Length
+  )
+{
+  UINT8  CheckSum;
+  UINTN  Count;
+
+  ASSERT (Buffer != NULL);
+  ASSERT (Length <= (MAX_ADDRESS - ((UINTN)Buffer) + 1));
+
+  for (CheckSum = 0, Count = 0; Count < Length; Count++) {
+    CheckSum ^= *(Buffer + Count);
+  }
+
+  return CheckSum;
+}
+
+/**
   Get the TransferList from HOB list.
 
   @param[out] TransferList  TransferList
@@ -76,7 +115,7 @@ TransferListVerifyChecksum (
     return TRUE;
   }
 
-  return (CalculateSum8 ((UINT8 *)TransferListHeader, TransferListHeader->UsedSize) == 0);
+  return (CalculateXor8 ((UINT8 *)TransferListHeader, TransferListHeader->UsedSize) == 0);
 }
 
 /**


leiflindholm:
Why? Was the original submission incorrect, or has the spec changed?

Reference [0] gives a "not found" error message.

LeviYeoReum:
> Why? Was the original submission incorrect

Yes. the original submission incorrect since the TF-A uses calculation as summation and 
to work with it, I've submiited with summation calcalation. 

So, this is a fix.

LeviYeoReum:
> > Why? Was the original submission incorrect
> Yes. the original submission incorrect since the TF-A uses calculation as summation and
> to work with it, I've submiited with summation calcalation.

Note: 
in the firmware specification, 
commit ae94bd1ff5b5 ("Switch to an xor for the checksum") changes the checksum way
from sum to xor in 2023 (https://github.com/FirmwareHandoff/firmware_handoff/commit/ae94bd1ff5b5d4a1024782934df8299b4c8d097b)

But, the TF-A doesn't reflect this change until last month.
As the it aligns according to specification in TF-A, now fix it according to specification.


> Reference [0] gives a "not found" error message.
Sorry. I'll fix this

leiflindholm:
OK, but the commit hash in a different repo unrelated to this implementation, while relevant in the PR description, isn't relevant for the commit message.
Same goes for some random repository that isn't public.

So could you update the commit message to say the relevant bits:
- Original submission was based on trustedfirmware.org transfer-list-lib
- transfer-list-lib did not follow specification
- transfer-list-lib has now been fixed, so implement the matching fix here
?

Then please drop the links from the commit message (not needed there), keep them in PR description (very useful here), but update [0] to point to https://git.trustedfirmware.org/plugins/gitiles/shared/transfer-list-library/ instead, which is publicly accessible.

leiflindholm:
> But, the TF-A doesn't reflect this change until last month.

Ah, if you could add that to commit message as well, including the commit hash for _that_, we're all good.

LeviYeoReum:
Thanks for suggestion. I'll change it!

leiflindholm (approved):

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -15,6 +15,45 @@
...
+
+  This function calculates the xor value of all elements in a buffer
+  in unit of UINT8 start from Zero.
+  The result is returned as UINT8. If Length is Zero, then Zero is
+  returned.
+
+  If Buffer is NULL, then ASSERT().
+  If Length is greater than (MAX_ADDRESS - Buffer + 1), then ASSERT().
+
+  @param  Buffer      The pointer to the buffer to carry out the xor operation.
```
vishalo: Please add [in] - to keep it consistent with rest of the file.

LeviYeoReum: okay. Thanks.

vishalo:
Is this required if we pick - https://github.com/tianocore/edk2/pull/11577 ?

LeviYeoReum:
> Is this required if we pick - #11577 ?

If that commit merged soon. this one doesn't need, But, as a bug fix and remain commit log,
This seems good to be merged.
However, I don't have any strong comment. I'll remain it to @samimujawar  or @leiflindholm .

leiflindholm:
So, TF-A has not yet stepped the version of its libtl submodule to a version that includes this fix. I don't think we can merge this until that happens.

leiflindholm:
> So, TF-A has not yet stepped the version of its libtl submodule to a version that includes this fix. I don't think we can merge this until that happens.

And indeed I don't think we can merge https://github.com/tianocore/edk2/pull/11577 either until that point.

LeviYeoReum:
> So, TF-A has not yet stepped the version of its libtl submodule to a version that includes this fix. I don't think we can merge this until that happens.

Yes. you're right. We'll need to wait the TF-A's change.

PR #11576, opened 2025-10-02
Add GCC/CLANG Support to PrmPkg

mdkinney:
# Description

## PrmPkg updates
* Set X64 4KB and AARCH64 64KB alignment for PRM modules in PrmPkg.dsc
* Remove notes from Readme that do not apply
* PrmPkg/Samples: Update INF files for GCC/CLANG
    * Remove redundant /DLL option from MSFT
    * Use VERSION_STRING from [Defines] section to set
      the PE/COFF optional header image version fields.
      The use of the MSFT linker /VERSION option is
      removed and replaced with the GenFw --image-version
      option to use same method for all tool chains and
      provide better error checking.
    * Add CLANGPDB tool chain family support that matches
      MSFT settings.
    * Add GCC family tool chain support that overrides
      OBJCOPY_STRIPFLAGS to nothing to prevent symbols
      from being stripped from PRM Modules so the
      PRM Export Descriptor Structure can be found by
      GENFW when the --prm option is used.
    * Remove all -Wl, options that are not compatible
      across all support tool chains. This also removes
      the need to list the PRM Export Descriptor Structure
      and PRM Handler APIs in the INF file and instead
      depend in GENFW with --prm option to find these
      symbols and generate the PE/COFF DLL export section.
* PrmPkg/Include: Fix GCC/CLANG PRM Module DLL Export issues
  * Use _MSC_EXTENSIONS instead of _MSC_VER when MSC specific
    extensions are required such as __declspec(dllexport).
    CLANGPDB builds do not set  _MSC_VER but CLANGPDB builds
    in Windows and Linux do support use of MSC specific
    extensions.
  * Add PRM_EXPORT_DESCRIPTOR macro that provides attributes
    required to make sure the PRM Export Descriptor Structure
    is not optimized away by the compiler/linker.
    PRM_EXPORT_API can not be used for this because the
    attributes for code and data may be different with
    different tool chains.
  * For GCC/CLANGDWARF tool chains, PRM_EXPORT_DESCRIPTOR
    sets the 'used' and 'section(.prmexportdescriptor)'
    attributes and depends on the GCC linker script
    to make sure the `.prmexportdescriptor` section is
    preserved.
  * Fix incorrect location of ; in PRM_MODULE_EXPORT()
    macro that was found by CLANG.

## BaseTools updates
* BaseTools/Scripts: KEEP .prmexportdescriptor data sections
* BaseTools/Source/C/GenFw: Add --image-version option
* BaseTools/Source/C/GenFw: Add no symbols check to --prm

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Build and evaluate generated PE/COFF images with X64 using VS019, VS2022, CLANGPDB, GCC, GCC5, GCCNOLTO, and CLANGDWARF.

## Integration Instructions

Existing PRM modules must update their DSC and INF files to use the templates from the samples in the PrmPkg to support GCC/CLANG toolchains.


BaseTools/Source/C/GenFw/Elf64Convert.c
```diff
@@ -2398,6 +2400,10 @@ WriteExport64 (
   UINT16                              Index;
   UINT8                               *Tdata = NULL;
 
+  if (mExportSymNum == 0) {
+    Error (NULL, 0, 3000, "Invalid", "%s has no export symbols.", mInImageName);
```
apop5: Could this error be made explicit? Mentioning that prm parameter was specified but there are not exports available for it?

mdkinney: I agree.  I will review all the new error messages in this PR to make sure they are useful.

mdkinney: Updated PR

BaseTools/Source/C/GenFw/Elf64Convert.c
```diff
@@ -2398,6 +2400,10 @@ WriteExport64 (
   UINT16                              Index;
   UINT8                               *Tdata = NULL;
 
+  if (mExportSymNum == 0) {
+    Error (NULL, 0, 3000, "Invalid", "--prm otion set but no export symbols were found in %s", mInImageName);
```
makubacki:
```suggestion
    Error (NULL, 0, 3000, "Invalid", "--prm option set but no export symbols were found in %s", mInImageName);
```

BaseTools/Source/C/GenFw/GenFw.c
```diff
@@ -1358,6 +1369,49 @@ Routine Description:
       continue;
     }
 
+    if (stricmp(argv[0], "--image-version") == 0) {
+      ImageVersionFlag = TRUE;
+      // Split argv[1] at '.'
+      ImageVersionDot = strchr(argv[1], '.');
```
makubacki, line 1379: If I understand correctly, `GenFw --image-version` would mean `argv[1]` is not valid. So, this should check for the existence `argv[1]` before accessing it.

BaseTools/Source/C/GenFw/Elf64Convert.c
mdkinney: I agree.  Fixed in latest push

BaseTools/Source/C/GenFw/GenFw.c
mdkinney: I agree.  Good catch.  Fixed in latest push

mdkinney:
@apop5 Please review with latest updates

@lgao4 @gapalomi Please review BaseTools changes

apop5:
ClangPdb all appears to be good. I was able to CI build and manually examine the output .efi and see CheckParamBufferPrmHandler existed.

ClandDwarf is a different matter.
When dumping the resulting efi (or dll, if you want to view symbols) CheckParambufferPrmHandler does not appear in the binary.

mdkinney:
> ClangPdb all appears to be good. I was able to CI build and manually examine the output .efi and see CheckParamBufferPrmHandler existed.
> 
> ClandDwarf is a different matter. When dumping the resulting efi (or dll, if you want to view symbols) CheckParambufferPrmHandler does not appear in the binary.

That is odd.  I checked CLANGFPDB and CLANGDWARF.  I used PE/COFF tools to show the DLL exports and the expected DLL exports were present.  The symbols that are used are not public but they are in the symbol tables parsed by GenFW.  

Please verify that the DLL export table is correct or not.  GenFw has been updated to fail with an error if the required symbols can not be found

apop5:
> > ClangPdb all appears to be good. I was able to CI build and manually examine the output .efi and see CheckParamBufferPrmHandler existed.
> > ClandDwarf is a different matter. When dumping the resulting efi (or dll, if you want to view symbols) CheckParambufferPrmHandler does not appear in the binary.
> 
> That is odd. I checked CLANGFPDB and CLANGDWARF. I used PE/COFF tools to show the DLL exports and the expected DLL exports were present. The symbols that are used are not public but they are in the symbol tables parsed by GenFW.
> 
> Please verify that the DLL export table is correct or not. GenFw has been updated to fail with an error if the required symbols can not be found

Which tools did you use exactly? I dumped using llvm-objdump installed with the llvm tools (bin\llvm-objdump)

When I attempt to use dumpbin (VS tool) on the clangdwarf compiled binary, it gives a known error (PrmSampleAcpiParameterBufferModule.dll : warning LNK4048: Invalid format file; ignored)

apop5:
> > > ClangPdb all appears to be good. I was able to CI build and manually examine the output .efi and see CheckParamBufferPrmHandler existed.
> > > ClandDwarf is a different matter. When dumping the resulting efi (or dll, if you want to view symbols) CheckParambufferPrmHandler does not appear in the binary.
> > 
> > 
> > That is odd. I checked CLANGFPDB and CLANGDWARF. I used PE/COFF tools to show the DLL exports and the expected DLL exports were present. The symbols that are used are not public but they are in the symbol tables parsed by GenFW.
> > Please verify that the DLL export table is correct or not. GenFw has been updated to fail with an error if the required symbols can not be found
> 
> Which tools did you use exactly? I dumped using llvm-objdump installed with the llvm tools (bin\llvm-objdump)
> 
> When I attempt to use dumpbin (VS tool) on the clangdwarf compiled binary, it gives a known error (PrmSampleAcpiParameterBufferModule.dll : warning LNK4048: Invalid format file; ignored)

Everything is fine.

llvm-objdump -p  for clangpdb is showing:

Export Table:
 DLL name: PrmSampleAcpiParameterBufferModule.dll
 Ordinal base: 1
 Ordinal      RVA  Name
       1   0x10d3  CheckParamBufferPrmHandler
       2   0x3010  PrmModuleExportDescriptor
       
llvm-objdump -p for clangdwarf is showing the following:
Export Table:
 DLL name: C:\Code\edk2\Build\Prm\DEBUG_CLANGDWARF\X64\PrmPkg\Samples\PrmSampleAcpiParameterBufferModule\PrmSampleAcpiParameterBufferModule\DEBUG\PrmSampleAcpiParameterBufferModule.dll
 Ordinal base: 1
 Ordinal      RVA  Name
       1        0  CheckParamBufferPrmHandler
       2   0x2020  PrmModuleExportDescriptor
       
       
Interestingly llvm-objdump -d for the clangdwarf does not show the disassembly for PrmModuleExportDescriptor, It only shows the code for ModuleEntryPoint, which is where my confusion stemmed from.

apop5 (approved):

BaseTools/Source/C/GenFw/Elf64Convert.c
```diff
@@ -2398,6 +2400,10 @@ WriteExport64 (
   UINT16                              Index;
   UINT8                               *Tdata = NULL;
 
+  if (mExportSymNum == 0) {
```
lgao4, line 2403: Seemly, mExportSymNum zero value is invalid. If so, why need to add check for mExportSymNum in the above code.

mdkinney: Because there can be multiple sections in the image with symbols.  There is an outer loop to go through all sections.  Only one will contain the PRM Export Descriptor symbol.  That section of code will access array entries beyond the end of the array if this check is not added.

lgao4: Got it.

lgao4 (approved):

PR #11577, opened 2025-10-03
Arm transfer list lib

HarishManchikanti:
**# Description**

This series of commits introduces support for the Transfer List Library (libtl) in ArmPkg/ArmTransferListLib.

The goal is to align EDK2 with the Firmware Handoff Specification by reusing the common libtl implementation instead of maintaining separate custom logic.
ArmTransferListLib now acts as a thin wrapper around libtl, exposing the same functionality via an EDK2-style API so platform code can work with Transfer Lists consistently.

**Summary of changes**

- Add libtl as a git submodule under ArmPkg/Library/ArmTransferListLib/libtl.
- Update ArmTransferListLib.inf to build with libtl sources.
- Replace previous in-tree custom Transfer List handling with calls into libtl.
- Add STATIC_ASSERTs to guarantee structure size/layout matches libtl.
- Guard register-convention macros in ArmTransferList.h to avoid conflicts with libtl headers.
- Extend ArmTransferListLib with additional wrapper APIs and helpers.

This ensures EDK2 stays aligned with the libtl reference implementation and avoids duplication.

Breaking change?
No. Existing ArmTransferListLib API continues to be provided, now backed by libtl.

Impacts security?
No direct security changes. Transfer List parsing continues to validate headers and sizes.

Includes tests?
No new unit tests. Verified manually with builds and integration tests on platform firmware.

**How This Was Tested**

- Built ArmPkg successfully for AARCH64 target.
- Validated that ArmTransferListLib links against libtl sources.
- Ran integration tests on platform firmware (Phoenix/FVP) confirming TL handoff still functions.
- Verified structure layout compatibility with libtl using STATIC_ASSERT.
- Tested with ftpm branch for any issues .

**Integration Instructions**

Consumers of ArmTransferListLib continue to include <Library/ArmTransferListLib.h> and link as before.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -35,3 +35,6 @@
 [submodule "SecurityPkg/DeviceSecurity/SpdmLib/libspdm"]
 	path = SecurityPkg/DeviceSecurity/SpdmLib/libspdm
 	url = https://github.com/DMTF/libspdm.git
+[submodule "ArmPkg/Library/ArmTransferListLib/libtl"]
+	path = ArmPkg/Library/ArmTransferListLib/libtl
+	url = https://review.trustedfirmware.org/shared/transfer-list-library.git
--- a/ArmPkg/Include/IndustryStandard/ArmTransferList.h
+++ b/ArmPkg/Include/IndustryStandard/ArmTransferList.h
@@ -30,16 +30,27 @@
  * For register convention, please see below:
  * https://github.com/FirmwareHandoff/firmware_handoff/blob/main/source/register_conventions.rst
  */
+#ifndef REGISTER_CONVENTION_VERSION_SHIFT_64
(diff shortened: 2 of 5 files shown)


leiflindholm:
As mentioned in #11575, we can't merge this until TF-A steps its version of transfer-list-library to one that includes the checksum fix. Do we have any idea when that may happen?

HarishManchikanti:
> As mentioned in #11575, we can't merge this until TF-A steps its version of transfer-list-library to one that includes the checksum fix. Do we have any idea when that may happen?

I see there’s an ongoing discussion here: https://github.com/FirmwareHandoff/firmware_handoff/issues/78 regarding how and when TF-A will adopt the libtl changes (including the checksum fix).

PR #11579, opened 2025-10-03
MdePkg: Add PCIe capability structures for DPC and DLF

jkolakow:
Add definitions for Downstream Port Containment (DPC) Extended Capability. Add definitions for Data Link Feature (DLF).
Update existing definitions for PASID and Precision Time Measurement (PTM) and move it from PciExpress40.h to PciExpress31.h as it was defined in PCIe 3.1. Mark some duplicate symbols 'obsolete' to remove in future.

# Description

- [ ] Breaking change?
  - No
- [ ] Impacts security?
  - No
- [ ] Includes tests?
  - No

## How This Was Tested

Tested in recently developed Intel server platforms.

## Integration Instructions

N/A

### DIFF
--- a/MdePkg/Include/IndustryStandard/Pci30.h
+++ b/MdePkg/Include/IndustryStandard/Pci30.h
@@ -42,7 +42,7 @@
 ///
 /// PCI Capability List IDs and records
 ///
-#define EFI_PCI_CAPABILITY_ID_PCIEXP  0x10
+#define EFI_PCI_CAPABILITY_ID_PCIEXP  0x10  // Obsolete, use PCI_EXPRESS_CAPABILITY_ID
 
 #pragma pack(1)
 
--- a/MdePkg/Include/IndustryStandard/PciExpress21.h
+++ b/MdePkg/Include/IndustryStandard/PciExpress21.h
@@ -31,8 +31,13 @@
 
 #pragma pack(1)
 ///
-/// PCI Express Capability Structure
+/// PCI Express Capability Structure version 2.
+/// Mandatory for PCI Express devices. If not present it is not PCI Express device, thus no extended config space.
 ///
+#define PCI_EXPRESS_CAPABILITY_ID         0x0010
+#define PCI_EXPRESS_CAPABILITY_VER1       0x1    // Version 1 ends at PCI_CAPABILITY_PCIEXP::RootStatus register
+#define PCI_EXPRESS_CAPABILITY_VER2       0x2    // Version 2 extends version 1 up to PCI_CAPABILITY_PCIEXP::SlotStatus2 register
+
 typedef union {
   struct {
     UINT16    Version                : 4;
@@ -396,29 +401,29 @@ typedef union {
 } PCI_REG_PCIE_SLOT_CAPABILITY2;
 
 typedef struct {
-  EFI_PCI_CAPABILITY_HDR             Hdr;
-  PCI_REG_PCIE_CAPABILITY            Capability;
-  PCI_REG_PCIE_DEVICE_CAPABILITY     DeviceCapability;
-  PCI_REG_PCIE_DEVICE_CONTROL        DeviceControl;
-  PCI_REG_PCIE_DEVICE_STATUS         DeviceStatus;
-  PCI_REG_PCIE_LINK_CAPABILITY       LinkCapability;
-  PCI_REG_PCIE_LINK_CONTROL          LinkControl;
-  PCI_REG_PCIE_LINK_STATUS           LinkStatus;
-  PCI_REG_PCIE_SLOT_CAPABILITY       SlotCapability;
-  PCI_REG_PCIE_SLOT_CONTROL          SlotControl;
-  PCI_REG_PCIE_SLOT_STATUS           SlotStatus;
-  PCI_REG_PCIE_ROOT_CONTROL          RootControl;
-  PCI_REG_PCIE_ROOT_CAPABILITY       RootCapability;
-  PCI_REG_PCIE_ROOT_STATUS           RootStatus;
-  PCI_REG_PCIE_DEVICE_CAPABILITY2    DeviceCapability2;
-  PCI_REG_PCIE_DEVICE_CONTROL2       DeviceControl2;
-  UINT16                             DeviceStatus2;
-  PCI_REG_PCIE_LINK_CAPABILITY2      LinkCapability2;
-  PCI_REG_PCIE_LINK_CONTROL2         LinkControl2;
-  PCI_REG_PCIE_LINK_STATUS2          LinkStatus2;
-  PCI_REG_PCIE_SLOT_CAPABILITY2      SlotCapability2;
-  UINT16                             SlotControl2;
-  UINT16                             SlotStatus2;
+  EFI_PCI_CAPABILITY_HDR             Hdr;                         // Offset 00 size 2
+  PCI_REG_PCIE_CAPABILITY            Capability;                  // Offset 02 size 2
+  PCI_REG_PCIE_DEVICE_CAPABILITY     DeviceCapability;            // Offset 04 size 4
+  PCI_REG_PCIE_DEVICE_CONTROL        DeviceControl;               // Offset 08 size 2
+  PCI_REG_PCIE_DEVICE_STATUS         DeviceStatus;                // Offset 0A size 2
+  PCI_REG_PCIE_LINK_CAPABILITY       LinkCapability;              // Offset 0C size 4
+  PCI_REG_PCIE_LINK_CONTROL          LinkControl;                 // Offset 10 size 2
+  PCI_REG_PCIE_LINK_STATUS           LinkStatus;                  // Offset 12 size 2
+  PCI_REG_PCIE_SLOT_CAPABILITY       SlotCapability;              // Offset 14 size 4
+  PCI_REG_PCIE_SLOT_CONTROL          SlotControl;                 // Offset 18 size 2
+  PCI_REG_PCIE_SLOT_STATUS           SlotStatus;                  // Offset 1A size 2
+  PCI_REG_PCIE_ROOT_CONTROL          RootControl;                 // Offset 1C size 2
+  PCI_REG_PCIE_ROOT_CAPABILITY       RootCapability;              // Offset 1E size 2
+  PCI_REG_PCIE_ROOT_STATUS           RootStatus;                  // Offset 20 size 4 - Ver1 ends here
+  PCI_REG_PCIE_DEVICE_CAPABILITY2    DeviceCapability2;           // Offset 24 size 4
+  PCI_REG_PCIE_DEVICE_CONTROL2       DeviceControl2;              // Offset 28 size 2
+  UINT16                             DeviceStatus2;               // Offset 2A size 2
+  PCI_REG_PCIE_LINK_CAPABILITY2      LinkCapability2;             // Offset 2C size 4
+  PCI_REG_PCIE_LINK_CONTROL2         LinkControl2;                // Offset 30 size 2
+  PCI_REG_PCIE_LINK_STATUS2          LinkStatus2;                 // Offset 32 size 2
+  PCI_REG_PCIE_SLOT_CAPABILITY2      SlotCapability2;             // Offset 34 size 4
+  UINT16                             SlotControl2;                // Offset 38 size 2
+  UINT16                             SlotStatus2;                 // Offset 3A size 2
 } PCI_CAPABILITY_PCIEXP;
 
 #define EFI_PCIE_CAPABILITY_BASE_OFFSET                           0x100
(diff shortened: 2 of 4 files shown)


MdePkg/Include/IndustryStandard/Pci30.h
```diff
@@ -42,7 +42,7 @@
 ///
 /// PCI Capability List IDs and records
 ///
-#define EFI_PCI_CAPABILITY_ID_PCIEXP  0x10
+#define EFI_PCI_CAPABILITY_ID_PCIEXP  0x10  // Obsolete, use PCI_EXPRESS_CAPABILITY_ID
```
mdkinney: Why is this obsolete?  Recommend moving this comment with a more complete description into the comment block above the define.  Refer to spec version/section where it became obsolete.

MdePkg/Include/IndustryStandard/PciExpress21.h
```diff
@@ -30,9 +30,15 @@
...
-/// PCI Express Capability Structure
-///
+
+//
+// PCI Express Capability Structure version 2.
+// Mandatory for PCI Express devices. If not present it is not PCI Express device, thus no extended config space.
+//
+#define PCI_EXPRESS_CAPABILITY_ID    0x0010
+#define PCI_EXPRESS_CAPABILITY_VER1  0x1         // Version 1 ends at PCI_CAPABILITY_PCIEXP::RootStatus register
+#define PCI_EXPRESS_CAPABILITY_VER2  0x2         // Version 2 extends version 1 up to PCI_CAPABILITY_PCIEXP::SlotStatus2 register
```
mdkinney: Move long comment lines into comment blocks above each define fir format to 80 columns

MdePkg/Include/IndustryStandard/PciExpress21.h
```diff
@@ -429,12 +435,12 @@ typedef struct {
...
 
 //
-// for SR-IOV
+// Single Root IO Virtualization (SR-IOV) Extended Capability Structure.
 //
-#define EFI_PCIE_CAPABILITY_ID_ARI    0x0E
-#define EFI_PCIE_CAPABILITY_ID_ATS    0x0F
-#define EFI_PCIE_CAPABILITY_ID_SRIOV  0x10
-#define EFI_PCIE_CAPABILITY_ID_MRIOV  0x11
+#define EFI_PCIE_CAPABILITY_ID_ARI    0x0E         // Obsolete, use PCI_EXPRESS_EXTENDED_CAPABILITY_ARI_CAPABILITY_ID
```
mdkinney: Add comment block above group of defines and explain why obsolete.,

MdePkg/Include/IndustryStandard/PciExpress31.h
```diff
@@ -67,41 +171,41 @@ typedef struct {
...
+    UINT32    EPtmCapable                     : 1;    // [3]
+    UINT32    PtmPropatagionDelayAdaptCapable : 1;    // [4]
+    UINT32    Reserved0                       : 3;    // [7:5]
+    UINT32    LocalClockGranularity           : 8;    // [15:8]
+    UINT32    Reserved1                       : 16;   // [31:16]
   } Bits;
-  UINT16    Uint16;
-} PCI_EXPRESS_EXTENDED_CAPABILITIES_PASID_CAPABILITY;
+  UINT32    Uint32;
+} PCI_EXPRESS_REG_PTM_CAPABILITY;
```
mdkinney: Is this a non-backward compatible change to change the size and type name?

MdePkg/Include/IndustryStandard/PciExpress31.h
```diff
@@ -67,41 +171,41 @@ typedef struct {
...
+    UINT32    PtmEnable            : 1;    // [0]
+    UINT32    RootSelect           : 1;    // [1]
+    UINT32    Reserved0            : 6;    // [7:2]
+    UINT32    EffectiveGranularity : 8;    // [15:8]
+    UINT32    Reserved1            : 16;   // [31:16]
   } Bits;
-  UINT16    Uint16;
-} PCI_EXPRESS_EXTENDED_CAPABILITIES_PASID_CONTROL;
+  UINT32    Uint32;
+} PCI_EXPRESS_REG_PTM_CONTROL;
```
mdkinney: Is this a non-backward compatible change to change the size and type name?

MdePkg/Include/IndustryStandard/PciExpress31.h
```diff
@@ -67,41 +171,41 @@ typedef struct {
...
 typedef struct {
-  PCI_EXPRESS_EXTENDED_CAPABILITIES_HEADER              Header;
-  PCI_EXPRESS_EXTENDED_CAPABILITIES_PASID_CAPABILITY    Capability;
-  PCI_EXPRESS_EXTENDED_CAPABILITIES_PASID_CONTROL       Control;
-} PCI_EXPRESS_EXTENDED_CAPABILITIES_PASID;
-///@}
+  PCI_EXPRESS_EXTENDED_CAPABILITIES_HEADER    Header;
+  PCI_EXPRESS_REG_PTM_CAPABILITY              Capability; // Offset 04h size 4
+  PCI_EXPRESS_REG_PTM_CONTROL                 Control;    // Offset 08h size 4
+} PCI_EXPRESS_EXTENDED_CAPABILITIES_PTM;
```
mdkinney: Is this a non-backward compatible change to change the size and type name?

MdePkg/Include/IndustryStandard/PciExpress40.h
```diff
@@ -16,47 +16,42 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 
 #pragma pack(1)
 
-/// Precision Time Management Extended Capability definitions.
-///
-/// Based on section 7.9.16 of PCI Express Base Specification 4.0.
-///@{
-#define PCI_EXPRESS_EXTENDED_CAPABILITY_PTM_ID    0x001F
```
mdkinney, line 22: Why is PMT removed?

MdePkg/Include/IndustryStandard/Pci30.h
jkolakow: This symbol use EFI_ prefix, but it is not UEFI spec defined. The ID 10h is defined in PCI Express specification. That is why I would like to discourage users from using this definition and replace it with PCI_EXPRESS_CAPABILITY_ID. Second reason is that the registers structure PCI_CAPABILITY_PCIEXP for this capability is defined in PciExpress21.h, just without the ID symbol. It is inconsistent with all other capability registers, where capability ID is defined just above capability registers structures.

MdePkg/Include/IndustryStandard/PciExpress21.h
jkolakow: My assumption was that up to 120 is fine and this format seems more readable to me, but of course can move to the comment block above.

MdePkg/Include/IndustryStandard/PciExpress21.h
jkolakow: Similar like EFI_PCI_CAPABILITY_ID_PCIEXP, these symbols use EFI_ prefix, but are not EFI spec defined, but PCI Express spec defined. They are also duplicate with the definitions that I point in the comment.

MdePkg/Include/IndustryStandard/PciExpress31.h
jkolakow: Actually this is not rename. The PASID definitions are moved above to follow the order of ID value which is usually followed by other definitions, just this one does not follow. And the PTM definitions are moved to this file from PciExpress40.h because PTM capability was added to PCI Express spec v3.1.
However, there is also a cleanup in PTM structure that I propose. I changed names like PTMRequesterCapable to PtmRequesterCapable to follow camel case. If you don't like such cleanup I can revert it.

MdePkg/Include/IndustryStandard/PciExpress31.h
jkolakow: This is not rename, PAS is moved to top of file and PTM is moved here from PciExpress40.h

MdePkg/Include/IndustryStandard/PciExpress31.h
jkolakow: This is not rename, PAS is moved to top of file and PTM is moved here from PciExpress40.h

MdePkg/Include/IndustryStandard/PciExpress40.h
jkolakow: I propose to move PTM definitions to PciExpress31.h because in the PCI Express Spec v3.1 it was added.

jkolakow: Moving PTM to original location to avoid confusion.

MdePkg/Include/IndustryStandard/PciExpress31.h
jkolakow: Moved to original location to avoid confusion.

MdePkg/Include/IndustryStandard/PciExpress40.h
jkolakow: Moved to original location to avoid confusion.

MdePkg/Include/IndustryStandard/PciExpress21.h
```diff
@@ -429,12 +437,12 @@ typedef struct {
...
+#define EFI_PCIE_CAPABILITY_ID_ATS    0x0F         // Obsolete, use PCI_EXPRESS_EXTENDED_CAPABILITY_ATS_ID
+#define EFI_PCIE_CAPABILITY_ID_SRIOV  0x10         // Obsolete, use PCI_EXPRESS_EXTENDED_CAPABILITY_SRIOV_ID
```
mdkinney: So the first 3 should never have been here because they use EFI_ prefix.  So the only reason to keep them is for source compatibility with any components that use those names.  Since it is for compat, I think the first three should be #define from EFI_* name to PCI_EXPRESS_* name with a comment block above that clearly states that these 3 are only here for compatibility and may be removed in the future and that the PCI_EXPRESS_* names should be used.

Are there any known use cases of EFI_PCIE_CAPABILITY_ID_MRIOV in source files?  Could we consider removing all together?

jkolakow: OK, will do the redefine. Regarding the MR-IOV, it is used to locate the capability in device config space, but the located offset is never used, so it seems safe to remove. I was thinking about removing in separate PR, but can also add to this PR.

MdeModulePkg/MdeModulePkg.dec
```diff
@@ -2050,12 +2050,6 @@
   # @Prompt Enable ARI support.
   gEfiMdeModulePkgTokenSpaceGuid.PcdAriSupport|TRUE|BOOLEAN|0x10000045
 
-  ## Indicates if the Multi Root I/O virtualization is supported.<BR><BR>
```
lgao4, line 2061: I suggest to update description to say it will not be used instead of remove it.

MdePkg/Include/IndustryStandard/PciExpress21.h
```diff
@@ -430,13 +438,18 @@ typedef struct {
...
-// for SR-IOV
+// Definitions EFI_PCIE_CAPABILITY_ID_ARI, EFI_PCIE_CAPABILITY_ID_ATS, EFI_PCIE_CAPABILITY_ID_SRIOV,
+// are obsolete, will be removed in future. Instead use PCI Express definitions
+// PCI_EXPRESS_EXTENDED_CAPABILITY_ARI_CAPABILITY_ID, PCI_EXPRESS_EXTENDED_CAPABILITY_ATS_ID,
+// PCI_EXPRESS_EXTENDED_CAPABILITY_SRIOV_ID.
 //
-#define EFI_PCIE_CAPABILITY_ID_ARI    0x0E
-#define EFI_PCIE_CAPABILITY_ID_ATS    0x0F
-#define EFI_PCIE_CAPABILITY_ID_SRIOV  0x10
-#define EFI_PCIE_CAPABILITY_ID_MRIOV  0x11
```
lgao4, line 437: Add comments to say EFI_PCIE_CAPABILITY_ID_MRIOV  is not used, but its defintiion is still kept.

MdePkg/Include/IndustryStandard/PciExpress40.h
```diff
@@ -16,47 +16,79 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
...
   struct {
-    UINT32    PTMRequesterCapable                  : 1;
-    UINT32    PTMResponderCapable                  : 1;
-    UINT32    PTMRootCapable                       : 1;
-    UINT32    ePTMCapable                          : 1;
-    UINT32    PTMPropagationDelayAdaptationCapable : 1;
-    UINT32    Reserved                             : 3;
-    UINT32    LocalClockGranularity                : 8;
-    UINT32    Reserved2                            : 16;
+    UINT32    PtmRequesterCapable             : 1;    // [0]
```
lgao4: I don't suggest to update the field name.

MdeModulePkg/MdeModulePkg.dec
jkolakow: OK, makes sense.

MdePkg/Include/IndustryStandard/PciExpress40.h
jkolakow: OK

PR #11582, opened 2025-10-03
IntelFsp2Pkg/GenCfgOpt.py: Fix line endings in Linux environments

zurcher:
# Description

This change specifies CRLF line endings in the output of GenCfgOpt.py to align with EDK2 coding standards.
Previously when running the tool in Linux environments (such as WSL), files were generated with LF-only line endings.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verified output no longer creates whole-file diffs in generated header files.

## Integration Instructions

N/A


zurcher:
Hi @nate-desimone and @ChaselChiu, would you mind taking a look at this?

PR #11590, opened 2025-10-06
OvmfPkg/PlatformInitLib: add sanity checks to igvm code

kraxel:
With SEV or SEV-ES being active uninitialized memory is random cipher
text instead of being zero-initialized.  So checking against zero to
figure whenever igvm parameters are present or not is not sufficient.
Apply more strict sanity checks.


OvmfPkg/Library/PlatformInitLib/Igvm.c
```diff
@@ -205,6 +222,12 @@ PlatformIgvmVpCount (
     return 0;
   }
 
+  if (*VpCount & 0xffff0000) {
+    DEBUG ((DEBUG_INFO, "%a: vp-count sanity check failed (>4096), ignoring\n", __func__));
```
tlendacky: ">65,535" now

bssrikanth:
Have tested this draft patch, it resolves the sev/sev-es guest boot issues.

Reported-by: Srikanth Aithal <sraithal@amd.com>
Tested-by: Srikanth Aithal <sraithal@amd.com>

OvmfPkg/Library/PlatformInitLib/Igvm.c
```diff
@@ -58,6 +64,13 @@ PlatformIgvmMemoryMapFind (
     return NULL;
   }
 
+  if (Map[0].EntryType > 0x0f ||
```
luigix25: what about using a constant here?

kraxel: There isn't really a constant.  Right now there are 4 or 5 page types, plus two unstable ones.  I've added some wiggle room for future new types and ended up with allowing 0x0f (aka 4 bits used) max, the remaining 12 bits in the page type field must all be zero.

PR #11592, opened 2025-10-06
ShellPkg/AcpiView: Update DumpAcpiTableToFile

leon9078:
# Description

This PR slightly modify DumpAcpiTableToFile() function inside AcpiView.c
It essentially improve 3 aspects:
* Change the dumped ACPI table file extension from ".bin" to ".aml" in order to follow official Intel (ACPICA) iASL tool. Source can be found [here](https://github.com/acpica/acpica/blob/53553c302b5fc57afe55dd3d96ce3aa861dea1df/source/compiler/asltypes.h#L315). <br/>
ACPI tables are AML generated code, so we would use the more specific "aml" file extension instead of the more generic "bin"

* Output filename Signature is now always uppercase, consistent with ACPI spec. See [here](https://uefi.org/specs/ACPI/6.6/05_ACPI_Software_Programming_Model.html#description-header-signatures-for-tables-defined-by-acpi) and [here](https://uefi.org/specs/ACPI/6.6/05_ACPI_Software_Programming_Model.html#description-header-signatures-for-tables-reserved-by-acpi)
<br/> E.g.
before: `acpiview -d -s dsdt` will create `dsdt0000.bin` file in current working directory
after: `acpiview -d -s dsdt` will create `DSDT0000.aml` file in current working directory
* Add length info of the ACPI dumped table to the Print log
<br/> E.g.
before: `Dumping ACPI table to : .\dsdt0000.bin ... `
after: `Dumping ACPI table to: .\DSDT0000.aml (68192 bytes) ... `

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Ran `acpiview -d -s dsdt` in UEFI shell and verified the newly added file with `ls`

## Integration Instructions

N/A


pierregondois:
Hello Leonardo,
About changing the file extension to `.aml`:
- The AML should be the compiled version of ASL files. All ACPI tables are not necessarily written in ASL language. For instance the GTDT table is not written in ASL, should we still dump the table with a `.aml` extension ?
- The `acpixtract` tool from the acpica-tools suite extract ACPI tables from a text file. The extracted tables have the `.dat` extension. So there is not a clear consensus on which file extension to use I believe.
- In general, if some scripts have been written to copy all the `.bin` files the `acpiview` tool created, these script will not work anymore after this patch.


About using upper case for file names, the UEFI shell should be:
- Case-preserving when creating files
- Case-insensitive when matching filenames
So it shouldn't make a difference, but the change doesn't seem to solve any existing issue.


About printing the size of files:
- I don't see any specific issue with that.

PR #11593, opened 2025-10-06
BaseTools: use double hyphens for CLANGPDB llvm-rc

xpahos:
# Description

When compiling rc files in OSX, llvm-rc treats '/Fo/PathToFile' as a positional argument. This results in the error "Exactly one input file should be provided". Using double hyphens llvm splits all data on the left side as options and data on the right side as positional arguments.

See: https://llvm.org/docs/CommandLine.html
Section: Specifying positional options with hyphens

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Run the following command on OS X(M3) and Linux(x64):

```
build -a X64 -t CLANGPDB -p OvmfPkg/OvmfPkgX64.dsc -b DEBUG
```

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Conf/build_rule.template
+++ b/BaseTools/Conf/build_rule.template
@@ -706,9 +706,13 @@
     <OutputFile.XCODE>
         $(OUTPUT_DIR)(+)$(MODULE_NAME)hii.rc
 
-    <Command.MSFT, Command.INTEL, Command.CLANGPDB>
+    <Command.MSFT, Command.INTEL>
         "$(GENFW)" -o $(OUTPUT_DIR)(+)$(MODULE_NAME)hii.rc -g $(MODULE_GUID) --hiipackage $(HII_BINARY_PACKAGES) $(GENFW_FLAGS)
         "$(RC)" /Fo${dst} $(OUTPUT_DIR)(+)$(MODULE_NAME)hii.rc
+
+    <Command.CLANGPDB>
+        "$(GENFW)" -o $(OUTPUT_DIR)(+)$(MODULE_NAME)hii.rc -g $(MODULE_GUID) --hiipackage $(HII_BINARY_PACKAGES) $(GENFW_FLAGS)
+        "$(RC)" /Fo${dst} -- $(OUTPUT_DIR)(+)$(MODULE_NAME)hii.rc
 
     <Command.GCC>
         "$(GENFW)" -o $(OUTPUT_DIR)(+)$(MODULE_NAME)hii.rc -g $(MODULE_GUID) --hiibinpackage $(HII_BINARY_PACKAGES) $(GENFW_FLAGS)


lgao4:
Do you verify the module MdeModulePkg\Universal\HiiResourcesSampleDxe? This module depends on rc comiler.

xpahos:
> Do you verify the module MdeModulePkg\Universal\HiiResourcesSampleDxe? This module depends on rc comiler.

Apologies for the long delay. I have re-submitted the pull request. I am running:

```
PATH=/opt/local/libexec/llvm-18/bin:$PATH build -p MdeModulePkg/MdeModulePkg.dsc \
      -m MdeModulePkg/Universal/HiiResourcesSampleDxe/HiiResourcesSampleDxe.inf \
      -a X64 -t CLANGPDB -b DEBUG

Build environment: macOS-15.7.1-arm64-arm-64bit-Mach-O
SOURCE_DATE_EPOCH not set - using 1762870460
Build start time: 17:14:20, Nov.11 2025

...


Processing meta-data .
Architecture(s)  = X64
Build target     = DEBUG
Toolchain        = CLANGPDB

...

- Done -
Build end time: 17:14:21, Nov.11 2025
Build total time: 00:00:02

```

```
PATH=/opt/local/libexec/llvm-19/bin:$PATH build -p MdeModulePkg/MdeModulePkg.dsc \
      -m MdeModulePkg/Universal/HiiResourcesSampleDxe/HiiResourcesSampleDxe.inf \
      -a X64 -t CLANGPDB -b DEBUG

Build environment: macOS-15.7.1-arm64-arm-64bit-Mach-O
SOURCE_DATE_EPOCH not set - using 1762870413
Build start time: 17:13:33, Nov.11 2025

...


Processing meta-data .
Architecture(s)  = X64
Build target     = DEBUG
Toolchain        = CLANGPDB

...

- Done -
Build end time: 17:13:35, Nov.11 2025
Build total time: 00:00:02
```


The binary file has been compiled:

```
file Build/MdeModule/DEBUG_CLANGPDB/X64/MdeModulePkg/Universal/HiiResourcesSampleDxe/HiiResourcesSampleDxe/OUTPUT/HiiResourcesSample.efi

Build/MdeModule/DEBUG_CLANGPDB/X64/MdeModulePkg/Universal/HiiResourcesSampleDxe/HiiResourcesSampleDxe/OUTPUT/HiiResourcesSample.efi: MS-DOS executable PE32+ executable (DLL) (EFI boot service driver) x86-64, for MS Windows
```

mikebeaton:
This change looks valid and useful to me.

I've confirmed it also allows building MdeModulePkg.dsc with the homebrew versions of llvm and lld (which do not otherwise work, with the same error as reported here).

@xpahos - I think the best way to merge your change into the latest build files is to add
```
*_CLANGPDB_*_RC_FLAGS                = --
```
at the end of the top section for CLANGPDB in `tools_def.template` rather the replicating and modifying the lines in `build_rule.template`.

xpahos:
> @xpahos - I think the best way to merge your change into the latest build files is to add
> 
> ```
> *_CLANGPDB_*_RC_FLAGS                = --
> ```
> 
> at the end of the top section for CLANGPDB in `tools_def.template` rather the replicating and modifying the lines in `build_rule.template`.

I rebased the commit and moved the change to `tools_def.template` with an additional comment. Also think this is the best option, since CLANGDWARF does not use llvm-rc.

mikebeaton:
Can I cc @mdkinney @os-d on this?

It looks sane and useful to me. I've confirmed it doesn't break the current/new CLANG CI.

(@xpahos (and others if they want to comment?) -  I'm not sure we need the new comment by the added line? It's possible to end up with overkill commenting everything, and git blame can be helpful for finding the explanation of specific lines without overloading the files themselves with comments.)

xpahos:
> Can I cc @mdkinney @os-d on this?

I think yes. Sorry, I'm not very familiar with the code review process. So far, I've only had a few pull requests where reviewers added a script from CI :)

> (@xpahos (and others if they want to comment?) - I'm not sure we need the new comment by the added line? It's possible to end up with overkill commenting everything, and git blame can be helpful for finding the explanation of specific lines without overloading the files themselves with comments.)

I added additional details to the commit message for better understanding in the future.

mdkinney:
I am not seeing any issues with CLANGPDB use of llvm-rc on Windows or Linux.  I do not see why a change to tools_def.txt is required here.

mdkinney:
llvm-rc is also used in CLANGDWARF for EmulatorPkg now:

https://github.com/tianocore/edk2/blob/a9d076d7d04e10c0ca397d4ca13a90e53a60bf94/EmulatorPkg/EmulatorPkg.dsc#L650

mikebeaton:
> I am not seeing any issues with CLANGPDB use of llvm-rc on Windows or Linux. I do not see why a change to tools_def.txt is required here.

I'm not exactly sure why this _isn't_ an issue for CLANGPDB on Windows or Linux, but the @xpahos certainly appears to be correct - from my tests and theirs - that this change is necessary and sufficient for CLANGPDB to build on macOS, using either Homebrew (my test) or MacPorts (apparently their test) distribution of LLVM - and without it, from current `master`, you get the error stated.

os-d:
I don't have an opinion on this change. I tested it on Windows and it worked, if it helps the MacOS experience I think that's fine, though that is of course a small user set.

mdkinney:
With the recent updated to build_rules.template, the /Fo option was removed from the $(RC) action. 

PR: https://github.com/tianocore/edk2/pull/11929
Commit: https://github.com/tianocore/edk2/pull/11929/commits/92625b33a3abab5951e629030befe748efbb8010

Please sync to the latest edk2/master and provide the error log from the build command without this proposed change.  If that no longer fails, then this PR can be closed.

xpahos:
> Please sync to the latest edk2/master and provide the error log from the build command without this proposed change. If that no longer fails, then this PR can be closed.

Already rebased.

In my opinion, `ArgList::AddAllArgValues` in the file `llvm/lib/Option/ArgList.cpp` behaves differently in OS X and Linux. The size of `Output` is 0 on OS X and 1 on Linux. This is the result of an empty list in the `filtered` call. It seems to be related to something in the OS X SDK. I'm not sure it will be possible to change llvm-rc in the OS X SDK.

mdkinney:
Are these the XCODE tools or the homebrew llvm tools?

I also see in the comments above that llvm 19 is being used.  I have seen issues with use of llvm 19.  Recommend moving to llvm 20.1.8.

mikebeaton:
> ... though that is of course a small user set.

True enough, although it's one of the reasons I'm here! (And us working to keep XCODE5 building (see e.g. second part of this message: https://github.com/tianocore/containers/pull/101#issuecomment-3573210480) has been one of the reasons that bringing CLANGPDB back on stream recently hasn't been even more of a pain than it would have been, since the errors that XCODE5 and the other CLANG toolchains catch - when errors aren't disabled - are not surprisingly very similar. Of course that's slightly tangential, since the current issue is about CLANGPDB on macOS, but just to show that helpful input to EDK2 is still coming from macOS users... :) )

> Are these the XCODE tools or the homebrew llvm tools?

Not Xcode clang - I believe that would only be appropriate for the XCODE5 toolchain.

When using clang, lld, llvm-rc, etc., from either MacPorts (in @xpahos' test) or Homebrew (my test) distributions, it gives these errors:

```
$ export PATH="/opt/homebrew/opt/llvm/bin:$PATH"
$ build -a X64 -t CLANGPDB -p MdeModulePkg/MdeModulePkg.dsc
....
Exactly one input file should be provided.
make: *** [/Users/mjsbeaton/OpenSource/w/edk2/Build/MdeModule/DEBUG_CLANGPDB/X64/MdeModulePkg/Application/HelloWorld/HelloWorld/OUTPUT/HelloWorldhii.lib] Error 1
...
```

The specific failing command is:
```
$ "llvm-rc"  /Users/mjsbeaton/OpenSource/w/edk2/Build/MdeModule/DEBUG_CLANGPDB/X64/MdeModulePkg/Application/HelloWorld/HelloWorld/OUTPUT/HelloWorldhii.rc
Exactly one input file should be provided.
```

or with @xpahos' change applied:

```
$ "llvm-rc" -- /Users/mjsbeaton/OpenSource/w/edk2/Build/MdeModule/DEBUG_CLANGPDB/X64/MdeModulePkg/Application/HelloWorld/HelloWorld/OUTPUT/HelloWorldhii.rc
```

which succeeds, as does the whole build.

@xpahos - This means the commit message needs updating, as the issue isn't specific to the /Fo option, but (apparently?) to the path being interpreted as a flag.

The above is just now with the absolutely current master:
```
$ git rev-parse HEAD
1588d5b9f0112a448ff76e80a872e35c1628176e
```
https://github.com/tianocore/edk2/commit/1588d5b9f0112a448ff76e80a872e35c1628176e

xpahos:
The problem is that `llvm-rc` in OS X interprets '/User/... ' as ['/U', 'ser', '...' ] when parsing arguments. 

```
(const llvm::opt::OptTable::Info) {
  Prefixes = {
    Data = 0x0000000100072648
    Length = 2
  }
  PrefixedName = {
    llvm::StringRef = (Data = "/U", Length = 2)
  }
  HelpText = 0x0000000100069599 "Undefine a symbol for the C preprocessor."
  HelpTextsForVariants = {
    __elems_ = {
      [0] = {
        first = {
          __elems_ = ([0] = 0, [1] = 0)
        }
        second = 0x0000000000000000
      }
    }
  }
  MetaVar = 0x0000000000000000
  ID = 24
  Kind = '\v'
  Param = '\0'
  Flags = 0
  Visibility = 1
  GroupID = 0
  AliasID = 0
  AliasArgs = 0x0000000000000000
  Values = 0x0000000000000000
}
```

I have updated the commit message.

mdkinney:
If the filename is put in quotes does that also resolve the issue?

xpahos:
> If the filename is put in quotes does that also resolve the issue?

```
xpahos-osx [xpahos] [/Users/xpahos/git/xpahos-edk2] PATH=/opt/local/libexec/llvm-18/bin:$PATH llvm-rc "/Users/xpahos/git/xpahos-edk2/Build/OvmfX64/DEBUG_CLANGPDB/X64/MdeModulePkg/Logo/LogoDxe/OUTPUT/LogoDxehii.rc"
Exactly one input file should be provided.
xpahos-osx [xpahos] [/Users/xpahos/git/xpahos-edk2] PATH=/opt/local/libexec/llvm-18/bin:$PATH llvm-rc '/Users/xpahos/git/xpahos-edk2/Build/OvmfX64/DEBUG_CLANGPDB/X64/MdeModulePkg/Logo/LogoDxe/OUTPUT/LogoDxehii.rc'
Exactly one input file should be provided.
```

Unfortunately, neither single nor double quotation marks help.

mikebeaton:
> The problem is that `llvm-rc` in OS X interprets '/User/... ' as ['/U', 'ser', '...' ] when parsing arguments.

Just to clarify, from some quick tests, the same issue applies in Linux if you make the required file live at `/User/...` (or even at `/usr/...`, it turns out). Which I would say makes this look less like a weird fix only required for macOS.

mdkinney:
> > The problem is that `llvm-rc` in OS X interprets '/User/... ' as ['/U', 'ser', '...' ] when parsing arguments.
> 
> Just to clarify, from some quick tests, the same issue applies in Linux if you make the required file live at `/User/...` (or even at `/usr/...`, it turns out). Which I would say makes this look less like a weird fix only required for macOS.

And with that info, I can reproduce in Windows cmd as well:

```
llvm-rc "/Users/x.a"
Exactly one input file should be provided.
```
And adding -- resolves

```
llvm-rc -- "/Users/x.a"
clang: error: no such file or directory: '/Users/x.a'
clang: error: no input files
llvm-rc: Preprocessing failed.
```

I agree with the change in this PR.  The rule that needs to be followed is that whenever llvm-rc is used as RC_PATH, then RC_FLAGS must end with `--`.

The EmulatorPkg.dsc file also uses llvm-rc, so that file needs to be updated in this PR as well.

I also suggest that the RC_FLAGS assignment to -- in tools_def.template be moved down to go just after GENFWHII_FLAGS so those two FLAGS are grouped together since they are used in the same rule.

os-d (approved):
Approved for EmulatorPkg. However, please split this into two commits, one for BaseTools and one for EmulatorPkg. That’s why PatchCheck is failing. You can run it locally to confirm your commits are matching edk2 style, after the split it would be python BaseTools/Scripts/PatchCheck.py -2

xpahos:
> Approved for EmulatorPkg. However, please split this into two commits, one for BaseTools and one for EmulatorPkg. That’s why PatchCheck is failing. You can run it locally to confirm your commits are matching edk2 style, after the split it would be python BaseTools/Scripts/PatchCheck.py -2

Sorry, I accidentally sent a commit to the main repository. I tried to test the build on Linux. I made some hacks to try to build EmulatorPkg with CLANGDWARF using the darwin target platform. There are many different problems, for example with the linker, and it seems that using CLANGDWARF on OS X is a bad idea. Overall, there is no way to test EmulatorPkg changes on OS X, and it doesn't make sense.

mdkinney:
The code changes look good.  The commit messages describe the issue as OSX specific, but I think the issue is in llvm-rc argument parsing for all OSes llvm-rc may be used.  I suggest the commit message me refined to describe it as a change for llvm-rc compatibility and it is not considered a work around anymore.

mikebeaton:
> The code changes look good. The commit messages describe the issue as OSX specific, but I think the issue is in llvm-rc argument parsing for all OSes llvm-rc may be used. I suggest the commit message me refined to describe it as a change for llvm-rc compatibility and it is not considered a work around anymore.

I agree.

I think you could however briefly mention (in the main CLANGPDB commit) that we typically hit this in macOS, and that this change allows CLANGPDB to work on macOS with typical directory names.

It might make more sense to use macOS not OS X as the OS name in the commit messages? I know that Mac OS is also a very old name for a different OS, previously on Macs.

xpahos:
> > The code changes look good. The commit messages describe the issue as OSX specific, but I think the issue is in llvm-rc argument parsing for all OSes llvm-rc may be used. I suggest the commit message me refined to describe it as a change for llvm-rc compatibility and it is not considered a work around anymore.
> 
> I agree.
> 
> I think you could however briefly mention (in the main CLANGPDB commit) that we typically hit this in macOS, and that this change allows CLANGPDB to work on macOS with typical directory names.
> 
> It might make more sense to use macOS not OS X as the OS name in the commit messages? I know that Mac OS is also a very old name for a different OS, previously on Macs.

I replaced "OS X" with a "Unix-like". Since EmulatorPkg CLANGDWARF cannot be compiled on OS X, I changed the example to a more general path in the EmulatePkg commit message.

mdkinney:
> > > The code changes look good. The commit messages describe the issue as OSX specific, but I think the issue is in llvm-rc argument parsing for all OSes llvm-rc may be used. I suggest the commit message me refined to describe it as a change for llvm-rc compatibility and it is not considered a work around anymore.
> > 
> > 
> > I agree.
> > I think you could however briefly mention (in the main CLANGPDB commit) that we typically hit this in macOS, and that this change allows CLANGPDB to work on macOS with typical directory names.
> > It might make more sense to use macOS not OS X as the OS name in the commit messages? I know that Mac OS is also a very old name for a different OS, previously on Macs.
> 
> I replaced "OS X" with a "Unix-like". Since EmulatorPkg CLANGDWARF cannot be compiled on OS X, I changed the example to a more general path in the EmulatePkg commit message.

I reproduced the issue in Windows too using the Mingw LLVM tools that have llvm-rc.exe.  So it is a generic issue for llvm-rc for all OSes and file systems.  It is any use of / that llvm-rc interprets as an option.

xpahos:
> > > > The code changes look good. The commit messages describe the issue as OSX specific, but I think the issue is in llvm-rc argument parsing for all OSes llvm-rc may be used. I suggest the commit message me refined to describe it as a change for llvm-rc compatibility and it is not considered a work around anymore.
> > > 
> > > 
> > > I agree.
> > > I think you could however briefly mention (in the main CLANGPDB commit) that we typically hit this in macOS, and that this change allows CLANGPDB to work on macOS with typical directory names.
> > > It might make more sense to use macOS not OS X as the OS name in the commit messages? I know that Mac OS is also a very old name for a different OS, previously on Macs.
> > 
> > 
> > I replaced "OS X" with a "Unix-like". Since EmulatorPkg CLANGDWARF cannot be compiled on OS X, I changed the example to a more general path in the EmulatePkg commit message.
> 
> I reproduced the issue in Windows too using the Mingw LLVM tools that have llvm-rc.exe. So it is a generic issue for llvm-rc for all OSes and file systems. It is any use of / that llvm-rc interprets as an option.

Replaced "Unix-like" with "environments where file paths begin with /"

mikebeaton:
Could we possibly still add in the CLANGPDB commit that this issue was in fact normally affecting macOS users, but not others, since paths in macOS typically start with /User (/U), which looks like a flag, even though it turns out that the issue is more general and can in principle affect any OS.

xpahos:
> Could we possibly still add in the CLANGPDB commit that this issue was in fact normally affecting macOS users, but not others, since paths in macOS typically start with /User (/U), which looks like a flag, even though it turns out that the issue is more general and can in principle affect any OS.

Added more details about macOS.

mikebeaton:
> > Could we possibly still add in the CLANGPDB commit that this issue was in fact normally affecting macOS users, but not others, since paths in macOS typically start with /User (/U), which looks like a flag, even though it turns out that the issue is more general and can in principle affect any OS.
> 
> Added more details about macOS.

Many thanks. It all looks good to me now.

mikebeaton:
I believe this should be okay for someone to push?

PR #11600, opened 2025-10-08
CryptoPkg: Fix build of MbedTlsLib

nate-desimone:
# Description

Fixes build failures caused by __USE_MINGW_ANSI_STDIO not being defined when building with CLANG.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Confirmed that this change fixes build issues with MbedTlsLib when using CLANGPDB.

## Integration Instructions

N/A

### DIFF
--- a/CryptoPkg/CryptoPkg.dsc
+++ b/CryptoPkg/CryptoPkg.dsc
@@ -327,11 +327,6 @@
       OpensslLib|CryptoPkg/Library/OpensslLib/OpensslLibFull.inf
       BaseCryptLib|CryptoPkg/Library/BaseCryptLib/BaseCryptLib.inf
       TlsLib|CryptoPkg/Library/TlsLib/TlsLib.inf
-    <BuildOptions>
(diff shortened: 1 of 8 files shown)


nate-desimone:
> Recommend never using -Wno-error=undef. Instead, add -D for defines that must be defined.
> 
> I also recommend that VS20xx enable warning C4668 so VS20xx/GGC/CLANG will all break the build if #if referenced a define that is not defined.

Done.

liyi77:
Hi, could we add this define to template file like this:
https://github.com/tianocore/edk2/pull/11386/files

This Macro should be a common Mingw define instead of Mbedtls specific as my understand.
https://sourceforge.net/p/mingw-w64/wiki2/gnu%20printf/

nate-desimone:
> Hi, could we add this define to template file like this: https://github.com/tianocore/edk2/pull/11386/files
> 
> This Macro should be a common Mingw define instead of Mbedtls specific as my understand. https://sourceforge.net/p/mingw-w64/wiki2/gnu%20printf/

Imho, I don't think we should do this in the general case. EDK II doesn't implement printf(), so there shouldn't be any need to use this macro. This is more of a hack to get MbedTls building in an environment that does not have libc.

PR #11601, opened 2025-10-08
Pvpanic

xpahos:
# Description

QEMU has a pvpanic device
(https://www.qemu.org/docs/master/specs/pvpanic.html) that provides the
ability to signal guest problems to the host and deliver QMP events,
for example, to monitoring systems. The library adds support for two
of the three possible uses of pvpanic: ISA (X64) and MMIO (AARCH64). There is also a PCI pvpanic device, but the complexity of
supporting it lies in the fact that errors can occur before PCI
enumeration. Three types of events have been implemented:

1. GUEST_PANICKED is sent before calling CpuDeadLoop(), as it is
   impossible to continue working after an error occurs.
2. In cases where it is possible to continue working, GUEST_CRASHLOADED
   is sent.
3. GUEST_SHUTDOWN is sent if the system shuts down normally.

- [X] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

**X64:**

```
build -a X64 -t GCC5 -p OvmfPkg/OvmfPkgX64.dsc -b DEBUG -DDEBUG_ON_SERIAL_PORT=TRUE -DSMM_REQUIRE=TRUE -DSECURE_BOOT_ENABLE=TRUE -DTPM2_ENABLE=TRUE -DFD_SIZE_4MB -DQEMU_PV_PANIC=TRUE
```

**Test with read-only NVRAM:**

```
/opt/qemu10/bin/qemu-system-x86_64 -S -d guest_errors,unimp -D $CWD/qemu.log \
    -blockdev driver=file,filename="$CWD/OVMF_CODE.fd",node-name=uefi-code,read-only=on \
    -blockdev driver=file,filename="$CWD/OVMF_VARS.fd",node-name=uefi-vars,read-only=on \
```

```
QemuFlashDetected => FD behaves as FLASH, write-protected
QemuFlashDetected => No
ASSERT [FvbServicesSmm] QemuFlash.c(285): !_gPcd_FixedAtBuild_PcdSmmSmramRequire
[PvPanicLibSendEvent] Sending Event: 0x1
```

**Test unsigned Secure Boot Image**:

```
FSOpen: Open '\EFI\BOOT\BOOTX64.EFI' Success
[Bds] Expand PciRoot(0x0)/Pci(0xB,0x0)/Pci(0x0,0x0) -> PciRoot(0x0)/Pci(0xB,0x0)/Pci(0x0,0x0)/HD(15,GPT,C870AAEE-B8E0-4AC1-994D-493A13132A04,0x2800,0x35000)/\EFI\BOOT\BOOTX64.EFI
BdsDxe: loading Boot0002 "UEFI Misc Device" from PciRoot(0x0)/Pci(0xB,0x0)/Pci(0x0,0x0)
[Security] 3rd party image[0] can be loaded after EndOfDxe: PciRoot(0x0)/Pci(0xB,0x0)/Pci(0x0,0x0)/HD(15,GPT,C870AAEE-B8E0-4AC1-994D-493A13132A04,0x2800,0x35000)/\EFI\BOOT\BOOTX64.EFI.
DxeImageVerificationLib: Image is signed but signature is not allowed by DB and SHA256 hash of image is not found in DB/DBX.
The image doesn't pass verification: PciRoot(0x0)/Pci(0xB,0x0)/Pci(0x0,0x0)/HD(15,GPT,C870AAEE-B8E0-4AC1-994D-493A13132A04,0x2800,0x35000)/\EFI\BOOT\BOOTX64.EFI
[PvPanicLibSendEvent] Sending Event: 0x2
```

**AARCH64**:

Using this patch https://lists.gnu.org/archive/html/qemu-devel/2025-10/msg00996.html

```
build -a AARCH64 -t GCC5 -p ArmVirtPkg/ArmVirtQemu.dsc -b DEBUG -DDEBUG_ON_SERIAL_PORT=TRUE -DSMM_REQUIRE=TRUE -DSECURE_BOOT_ENABLE=TRUE -DTPM2_ENABLE=TRUE -DFD_SIZE_4MB -DQEMU_PV_PANIC=TRUE
```

**Hardcoded CpuDeadLoop in SecMain**

```
UEFI firmware (version  built at 19:43:05 on Oct  8 2025)
add-symbol-file /home/xpahos/git/xpahos-edk2/Build/ArmVirtQemu-AArch64/DEBUG_GCC5/AARCH64/ArmPlatformPkg/Sec/Sec/DEBUG/Sec.dll 0x2000
add-symbol-file /home/xpahos/git/xpahos-edk2/Build/ArmVirtQemu-AArch64/DEBUG_GCC5/AARCH64/MdeModulePkg/Core/Pei/PeiMain/DEBUG/PeiCore.dll 0xF240
[PvPanicLibSendEvent] Sending Event: 0x1
```

**Test unsigned Secure Boot Image**:

```
[Bds] Expand PciRoot(0x0)/Pci(0x2,0x0) -> <null string>
[PvPanicLibSendEvent] Sending Event: 0x2
```

```
memory_region_ops_write cpu 0 mr 0x5aa1aeecf8e0 addr 0x9021000 value 0x2 size 1 name 'pvpanic'
Guest crash loaded
```

## Integration Instructions

Add `-DQEMU_PV_PANIC=TRUE` flag while building.

### DIFF
--- a/DynamicTablesPkg/Library/Acpi/Common/AcpiCedtLib/GoogleTest/CedtGeneratorGoogleTest.cpp
+++ b/DynamicTablesPkg/Library/Acpi/Common/AcpiCedtLib/GoogleTest/CedtGeneratorGoogleTest.cpp
@@ -78,17 +78,6 @@ using ::testing::DoAll;
 using ::testing::SetArgPointee;
 using ::testing::AtLeast;
 
-#define WRAP_ACCESSOR(accessor) \
-  [this] \
-  (IN  CONST EDKII_CONFIGURATION_MANAGER_PROTOCOL *This, \
-   IN  CONST CM_OBJECT_ID                         CmObjectId, \
-   IN  CONST CM_OBJECT_TOKEN                      Token, \
-   IN  OUT   CM_OBJ_DESCRIPTOR                    *CmObject \
-  ) \
-  { \
-    return this->accessor(This, CmObjectId, Token, CmObject); \
-  }
-
 class CedtGeneratorTest : public ::testing::Test {
 protected:
   MockConfigurationManagerProtocol MockConfigMgrProtocol;
--- a/DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/GoogleTest/HmatGeneratorGoogleTest.cpp
+++ b/DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/GoogleTest/HmatGeneratorGoogleTest.cpp
@@ -1,0 +1,719 @@
+/** @file
+  Unit tests for HMAT Generator
+
+  Copyright (c) 2026, Google LLC. All rights reserved. <BR>
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#include "Base.h"
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+#include <Library/GoogleTestLib.h>
+
+extern "C" {
+  #include <Uefi.h>
+  #include <Protocol/ConfigurationManagerProtocol.h>
+  #include <Library/FunctionMockLib.h>
+  #include <Library/BaseMemoryLib.h>
+  #include <Library/DebugLib.h>
+  #include <Library/MemoryAllocationLib.h>
+  #include <Library/TableHelperLib.h>
+  #include <Library/MetadataObjLib.h>
+  #include <AcpiTableGenerator.h>
+  #include <ConfigurationManagerObject.h>
+  #include <ConfigurationManagerHelper.h>
+  #include <StandardNameSpaceObjects.h>
+  #include <IndustryStandard/Acpi64.h>
+  #include "GoogleTest/Protocol/MockConfigurationManagerProtocol.h"
+  #include "Library/Common/MetadataHandlerLib/MetadataHandler.h"
+
+  EFI_STATUS
+  EFIAPI
+  AcpiHmatLibConstructor (
+    IN EFI_HANDLE        ImageHandle,
+    IN EFI_SYSTEM_TABLE  *SystemTable
+    );
+
+  EFI_STATUS
+  EFIAPI
+  AcpiHmatLibDestructor (
+    IN EFI_HANDLE        ImageHandle,
+    IN EFI_SYSTEM_TABLE  *SystemTable
+    );
+
+  // Global generator instance
+  static ACPI_TABLE_GENERATOR  *gHmatGenerator = NULL;
+
+  static METADATA_ROOT_HANDLE  mMetadataRoot;
+
+  // C++ wrapper functions for C linkage functions
+  EFI_STATUS
+  RegisterAcpiTableGenerator (
+    IN  CONST ACPI_TABLE_GENERATOR  *CONST  TableGenerator
+    )
+  {
+    if (TableGenerator == NULL) {
+      return EFI_INVALID_PARAMETER;
+    }
+
+    gHmatGenerator = const_cast<ACPI_TABLE_GENERATOR *>(TableGenerator);
+    return EFI_SUCCESS;
+  }
+
+  EFI_STATUS
+  DeregisterAcpiTableGenerator (
+    IN  CONST ACPI_TABLE_GENERATOR  *CONST  TableGenerator
+    )
+  {
+    if (TableGenerator == NULL) {
+      return EFI_INVALID_PARAMETER;
+    }
+
+    // Clear the stored generator
+    gHmatGenerator = NULL;
+    return EFI_SUCCESS;
+  }
+
+  METADATA_ROOT_HANDLE
+  EFIAPI
+  GetMetadataRoot (
+    VOID
+    )
+  {
+    return mMetadataRoot;
+  }
+}
+
+using namespace testing;
+using ::testing::_;
+using ::testing::NiceMock;
+using ::testing::Return;
+using ::testing::DoAll;
+using ::testing::SetArgPointee;
+using ::testing::AtLeast;
+
+class HmatGeneratorTest : public ::testing::Test {
+protected:
+  MockConfigurationManagerProtocol MockConfigMgrProtocol;
+  CM_STD_OBJ_CONFIGURATION_MANAGER_INFO CfgMgrInfo = { 0 };
+
+  void
+  ValidateMemProximityDomainAttrInfo (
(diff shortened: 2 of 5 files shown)


MdePkg/Library/PvPanicLib/PvPanicLibISA.c
```diff
@@ -0,0 +1,31 @@
...
+**/
+VOID
+EFIAPI
+PvPanicLibSendEvent (
+  IN UINT8  Event
+  )
+{
+  DEBUG ((DEBUG_INFO, "[%a] Sending Event: 0x%x\n", __func__, Event));
+
+  IoWrite8 (PcdGet32 (PcdPvPanicIoPort), Event);
```
kraxel: So you are trying to talk to the device without checking it is actually present.

kraxel (commented):
Also: why is this in MdePkg?  Shouldn't it go to OvmfPkg instead?
What happens if the firmware triggers an ASSERT(), is that signaled to the host?
Any specific reason why this is a compile time option?

xpahos:
> Also: why is this in MdePkg? Shouldn't it go to OvmfPkg instead?

I looked at the dependencies of the MdePkg and MdeModulePkg modules and did not find any dependencies on OvmfPkg. Would it be the right decision to make these modules dependent on OvmfPkg?

> What happens if the firmware triggers an ASSERT(), is that signaled to the host?

I'm not sure whether it's necessary to add a signal to ASSERT. ASSERT is supposed to work in the DEBUG build, and it would probably be a good idea to add it there too. But I have my doubts that someone would run a virtual machine with a DEBUG build and leave it unattended. I assumed that since ASSERT is available, the user already knows that there is a problem and is debugging it, for example, via the serial port.

> Any specific reason why this is a compile time option?

I conceived this as an option that should only be enabled if you understand why it is needed. Never enable it by default. First, you enable options in QEMU (for example, change the MMIO address) and only then compile UEFI with the pvpanic option to avoid a situation where pvpanic is present in UEFI but not enabled in the virtual machine.

MdePkg/Library/PvPanicLib/PvPanicLibISA.c
xpahos: I tried to come up with a way to test this, but I haven't come up with anything good yet. You could try writing 0b10000000 using an undefined bit, then try reading it. QEMU will report an error, but it won't change its behaviour. Maybe there's some other way in Sec to check if the device is available?

kraxel:
> > Also: why is this in MdePkg? Shouldn't it go to OvmfPkg instead?
> 
> I looked at the dependencies of the MdePkg and MdeModulePkg modules and did not find any dependencies on OvmfPkg. Would it be the right decision to make these modules dependent on OvmfPkg?

No, MdePkg can not depend on OvmfPkg.  But this is for virtual machines only, so why MdePkg needs a dependency on it?

> > What happens if the firmware triggers an ASSERT(), is that signaled to the host?
> 
> I'm not sure whether it's necessary to add a signal to ASSERT. ASSERT is supposed to work in the DEBUG build, and it would probably be a good idea to add it there too. But I have my doubts that someone would run a virtual machine with a DEBUG build and leave it unattended. 

OVMF typically /does/ runs with DEBUG builds.

> > Any specific reason why this is a compile time option?
> 
> I conceived this as an option that should only be enabled if you understand why it is needed. Never enable it by default. First, you enable options in QEMU (for example, change the MMIO address) and only then compile UEFI with the pvpanic option to avoid a situation where pvpanic is present in UEFI but not enabled in the virtual machine.

Off by default is fine, but it should be possible to turn it on without rebuilding OVMF, otherwise nobody will use it.  Ideally OVMF would auto-detect whenever the device is present.  Not sure this is possible without parsing the acpi tables from qemu though.

xpahos:
> > > Also: why is this in MdePkg? Shouldn't it go to OvmfPkg instead?
> > 
> > 
> > I looked at the dependencies of the MdePkg and MdeModulePkg modules and did not find any dependencies on OvmfPkg. Would it be the right decision to make these modules dependent on OvmfPkg?
> 
> No, MdePkg can not depend on OvmfPkg. But this is for virtual machines only, so why MdePkg needs a dependency on it?

In my opinion, `CpuDeadLoop` is suitable as a single location for error notifications. However, `CpuDeadLoop` is defined in `MdePkg`. I would appreciate advice on the best way to do this so that there is no dependency and this code is located in `OvmfPkg`. 

> > > What happens if the firmware triggers an ASSERT(), is that signaled to the host?
> > 
> > 
> > I'm not sure whether it's necessary to add a signal to ASSERT. ASSERT is supposed to work in the DEBUG build, and it would probably be a good idea to add it there too. But I have my doubts that someone would run a virtual machine with a DEBUG build and leave it unattended.
> 
> OVMF typically /does/ runs with DEBUG builds.

I will add functionality for ASSERT directly into the macro.

> > > Any specific reason why this is a compile time option?
> > 
> > 
> > I conceived this as an option that should only be enabled if you understand why it is needed. Never enable it by default. First, you enable options in QEMU (for example, change the MMIO address) and only then compile UEFI with the pvpanic option to avoid a situation where pvpanic is present in UEFI but not enabled in the virtual machine.
> 
> Off by default is fine, but it should be possible to turn it on without rebuilding OVMF, otherwise nobody will use it. Ideally OVMF would auto-detect whenever the device is present. Not sure this is possible without parsing the acpi tables from qemu though.

I don't know what to do with the layout before PCI enumeration. Perhaps change QEMU's behaviour and add a magic number when reading from the register, which would allow determining whether the device is present or not? `fw_cfg` is available at later stages of `SEC/PEI`. On the other hand, errors at these stages in a virtual machine are unlikely, and it may indeed be worth returning errors only in `DXE` and later.

mdkinney:
I also do no think this should go into MdePkg.

Also, CpuDeadLoop() is not a good indication of a panic condition.  The best indication of a error/panic condition is an ASSERT() or REPORT_STATUS_CODE().  The DebugLib() has a PCD to define the behavior of an ASSERT() of which only one of the defined behaviors is a dead loop.  In order for this feature to catch panic conditions for all possible platform configurations, I recommend dispatching to is through an ASSERT() or REPORT_STATUS_CODE() path.

There have also been discussions in the past about adding the concept of a PANIC() macro to MdePkg to distinguish ASSERT()s that are potentially non-fatal conditions from ASSERT()s that are fatal and that the system can not continue execution past the point of detection.

If we really want something in MdePkg, I would recommend revisiting the previous discussions.

xpahos:
bump. I need more time to try to formulate a proposal on what to do with macros.

xpahos:
bump

xpahos:
bump

PR #11603, opened 2025-10-09
MdeModulePkg: ScsiDiskDxe: Query Write Protected State

os-d:
# Description

Currently, only SCSI attached CDROMs are marked as read only SCSI devices. Write failures have been seen to occur when a read only disk is attached as a SCSI device, because the device is never queried for its write protected state and the disk is listed as writeable.

This adds a new mode sense query to determine if a SCSI disk is write protected. If so, the proper BlockIO media property is set to read only, to ensure that all layers know this is a read only disk.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested by booting a VM with a read only VHD attached. Prior to this change, it would fail, with this change the boot succeeds.

## Integration Instructions

N/A.


MdeModulePkg/Bus/Scsi/ScsiDiskDxe/ScsiDisk.c
```diff
@@ -196,6 +196,64 @@ ScsiDiskDriverBindingSupported (
...
+/**
+ * Check whether the SCSI disk is write protected.
+ *
+ * @param ScsiDiskDevice          The SCSI disk device.
+ * @param WriteProtectionEnabled  A pointer to a boolean that will be set to TRUE if the disk is write protected,
+ *                                FALSE otherwise.
+ *
+ * @return EFI_SUCCESS            The operation completed successfully.
+ * @return other                  An error occurred while executing the SCSI command.
+ */
```
makubacki, line 208:
```suggestion
/**
 Check whether the SCSI disk is write protected.

 @param[in]  ScsiDiskDevice          The SCSI disk device.
 @param[out] WriteProtectionEnabled  A pointer to a Boolean that will be set to TRUE if the disk is write protected,
                                     FALSE otherwise.

  @retval EFI_SUCCESS            The operation completed successfully.
  @retval other                  An error occurred while executing the SCSI command.
 */
```

MdeModulePkg/Bus/Scsi/ScsiDiskDxe/ScsiDisk.c
```diff
@@ -196,6 +196,64 @@ ScsiDiskDriverBindingSupported (
...
+  // Initialize SCSI REQUEST_PACKET and 6-byte Cdb
+  //
+  ZeroMem (&CommandPacket, sizeof (CommandPacket));
+  ZeroMem (Cdb, sizeof (Cdb));
+
+  // Initialize output parameter to default value
+  *WriteProtectionEnabled = FALSE;
+
+  Cdb[0] = ATA_CMD_MODE_SENSE6;
+  Cdb[1] = 0b1000;                         // Setting the bit for Disable Block Descriptor
```
makubacki: nit: Given the bit macro exists, I think it's appropriate here but feel to ignore.

```suggestion
  Cdb[1] = BIT3;                         // Setting the bit for Disable Block Descriptor
```

MdeModulePkg/Bus/Scsi/ScsiDiskDxe/ScsiDisk.c
```diff
@@ -196,6 +196,64 @@ ScsiDiskDriverBindingSupported (
...
+
+  Status = ScsiDiskDevice->ScsiIo->ExecuteScsiCommand (ScsiDiskDevice->ScsiIo, &CommandPacket, NULL);
+
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  // Mode Sense 6 Byte Command returns the Write Protection status in the 3rd byte
+  // Bit 7 of the 3rd byte indicates the Write Protection status
+  *WriteProtectionEnabled = (DataBuffer[2] & 0x80) != 0;
```
makubacki:
```suggestion
  *WriteProtectionEnabled = (DataBuffer[2] & BIT7) != 0;
```

makubacki:
@lgao4 & @mdkinney, we'd like to get this in soon if possible, to pull into a platform.

Are you able to review in the next day or so? If not, that's fine. We'll just need to take a temporary approach until this can be reviewed and merged.

makubacki:
Bumping request for review @mdkinney, @lgao4, @jkolakow

lgao4 (approved):

praveensankarn:
Hi @os-d ,
    I have some questions related to the changes made in IsWriteProtected( ). 
    1) CommandPacket.InDataBuffer     = &DataBuffer; 
    Here the DataBuffer is itself a static array, so we should assign DataBuffer alone to the CommandPacket.InDataBuffer right?
    2) Before assigning WP bit to the dereferenced pointer *WriteProtectionEnabled, we should check the appropriate TargetStatus and HostAdapterStatus right before fetching WP bit?

@lgao4, @makubacki @jkolakow Please let me your comments as well.

os-d:
@praveensankarn are you hitting an issue with this code?

For the first point, it is confusing, but `DataBuffer` and `&DataBuffer` point to the same location, just with a different type. For example, see: https://godbolt.org/z/n755n89eq.

Where

image

Produces the output:

image

For TargetStatus and HostAdapterStatus I believe you are correct, those should be checked before using the data. Do you want to submit a PR for that?

PR #11608, opened 2025-10-10
MdeModulePkg,UefiPayloadPkg: Introduce platform limit for max size to pick in Resizable BAR

jkolakow:
Resizable BAR driver selects max available BAR size to configure for use. It may happen that some PCIe device declare support for size that exceeds processor address width. Platform needs an option to define constraint for the size it can accept. This change introduce dynamic PCD called PcdPcieResizableBarMaxSize where platform can provide its limit for BAR size. Such PCD can be predefined, but also can be controlled with a configuration knob or rebalance mechanisms.

- [ ] Breaking change?
  - No
- [ ] Impacts security?
  - No
- [ ] Includes tests?
  - No

## How This Was Tested

Tested in server platform with GPU with Resizable BAR capability.

## Integration Instructions

N/A


lgao4:
If the platform doesn't configure it, does it impact current logic?

MdeModulePkg/MdeModulePkg.dec
```diff
@@ -2303,6 +2303,15 @@
...
   gEfiMdeModulePkgTokenSpaceGuid.PcdFfaLibConduitSmc|TRUE|BOOLEAN|0x10000029
 
+  ## Defines platform limit for BAR size that can be accepted in Resizable BAR Capability.
+  # 0           1 MB (2^20)
+  # 1           2 MB (2^21)
+  # ...
+  # 42          4 EB (2^62)
+  # 43          8 EB (2^63)
+  # @Prompt Max size to accept in PCIe Resizable BAR Capability register.
+  gEfiMdeModulePkgTokenSpaceGuid.PcdPcieResizableBarMaxSize|43|UINT8|0x10000030
```
lgao4, line 2327: I don't understand why 43 maps to 2^63.

jkolakow: Thank you for pointing this. I'm updating description, let me know if this is clear now, or something could be improved yet.

jkolakow:
> If the platform doesn't configure it, does it impact current logic?

No, it does not. Default is 43 (43+20 = 63 -> 8 EiB) so practically no limit.
What is modified in current logic is that any error in capability registers cause the capability is ignored. Former implementation tried to "fix it" what only leads to problems if this capability is really screwed.

MdeModulePkg/Bus/Pci/PciBusDxe/PciLib.c
```diff
@@ -1857,7 +1912,7 @@ PciProgramResizableBar (
     Entries[Index].ResizableBarControl.Bits.BarSize = (UINT32)Bit;
     DEBUG ((
       DEBUG_INFO,
-      "   Resizable Bar: Offset = 0x%x, Bar Size Capability = 0x%016lx, New Bar Size = 0x%lx\n",
+      "   Resizable BAR: offset = 0x%X, size capability mask = 0x%014lX, new BAR size = 0x%lX\n",
```
lgao4: One minor comment, why output length changed from 16 to 14?

jkolakow: There are 44 capability bits total - 28 defined originally in Capability register and 16 defined later in initially unused part of Control register. 44 bits take 11 digits. I reduced it originally by one byte, but I see I could even reduce it by 2 bytes to 12. Let me update yet.

PR #5606, opened 2024-04-30
S3 cleanup

td36:
This patch set has been reviewed-by responding pkg maintainers(Ni, Ray, Wu jiaxin, Ard Biesheuvel and Gao Liming) in previous edk2 community mail loop.

This patch set is to remove some S3 related code in CpuS3.c of smm cpu driver. It contain commits to:
1) S3 MTRRs operation:
>   MdeModulePkg: Add gEdkiiS3MtrrSettingGuid
>   OvmfPkg: Save MTRR by lockbox in CpuS3DataDxe
>   UefiCpuPkg: Add locbox lib instance in DSC
>   UefiCpuPkg: Save MTRR by lockbox in CpuS3DataDxe
>   UefiCpuPkg: LoadMtrrData for all cpu in S3Resume
>   UefiCpuPkg: Remove code to load mtrr setting

2) AP page table unavailiable issue fix:
>   UefiCpuPkg: Disable PG in IA32 ApLoopCode

3) Register table cleanup:
>   UefiCpuPkg:Set PcdCpuFeaturesInitOnS3Resume to TRUE
>   UefiCpuPkg: Remove code to set register table

4)  S3 ApHltLoopCode Operation:
>   UefiCpuPkg:Abstract some DxeMpLib code to function
>   UefiCpuPkg:Move some code in DxeMpLib to common place
>   UefiCpuPkg: Install gEdkiiEndOfS3ResumeGuid in S3Resume
>   UefiCpuPkg:Relocate AP to new safe buffer in PeiMpLib
>   UefiCpuPkg:Remove code to handle APIC setting and Interrupt
>   UefiCpuPkg:Rremove code to wakeup AP and relocate ap
>   UefiCpuPkg: Remove the duplicated mpservice locate
>   MdeModulePkg: remove MpService2Ppi field in SMM_S3_RESUME_STATE

5)  Remove code to get AcpiCpuData:
>   UefiCpuPkg: Remove GetAcpiCpuData() in CpuS3.c

PR #11610, opened 2025-10-11
UefiCpuPkg/PeiMpLib: Only allocate ACPI NVS AP loop code buffer on S3 resume

makubacki:
# Description

This commit allocated an ACPI NVS buffer in PEI: https://github.com/tianocore/edk2/commit/cdc1a88272237149003d1d1ea301d320f7b3bdf6

This is a RT visible memory type that remains fragmented against the overall `EfiACPIMemoryNVS` memory bucket allocated by the DXE Core. The buffer is reallocated by `DxeMpLib` as `EfiReservedMemoryType` on non-S3 boots.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- CI
- QEMU boot to EFI shell
- Check `EfiACPIMemoryNVS` memory regions before and after the change

## Integration Instructions

- N/A


UefiCpuPkg/Library/MpInitLib/PeiMpLib.c
```diff
@@ -810,9 +810,19 @@ AllocateApLoopCodeBuffer (
...
+  ASSERT_EFI_ERROR (Status);
+
+  if (!EFI_ERROR (Status) && (BootMode == BOOT_ON_S3_RESUME)) {
+    // Only allocate a RT visible memory type on S3 resume
+    Status = PeiServicesAllocatePages (EfiACPIMemoryNVS, Pages, Address);
+  } else {
+    Status = PeiServicesAllocatePages (EfiBootServicesData, Pages, Address);
+  }
 
-  Status = PeiServicesAllocatePages (EfiACPIMemoryNVS, Pages, Address);
```
lzeng14, line 815: any specific consideration to use EfiACPIMemoryNVS before ?
If boot mode check is needed, it may be able to be added into InitMpGlobalData() when calling PrepareApLoopCode() which looks like only needed for S3 boot path.
@niruiyu for any further comment.

makubacki: This is a separate issue from what this change is addressing:

> any specific consideration to use EfiACPIMemoryNVS before ?

I'm happy to change that here, but it is a priority to fix the issue directly addressed in this PR as it impacts hibernate stability.

makubacki:
Bump to reviewers to please review when you have a moment. We need this fixed.

@arunsbaskaran @ecuevasf @jiaxinwu @kraxel @LiuZhiguang001 @niruiyu @td36

UefiCpuPkg/Library/MpInitLib/PeiMpLib.c
makubacki: @lzeng14 & @niruiyu, when do you think we can close this thread?

niruiyu: If pre-allocated NVS type memory is large enough in the MemoryTypeInformation Bin, will the issue be resolved?

I guess it's controlled by PcdPlatformEfiAcpiNvsMemorySize if MinPlatform is used.

lzeng14: It is not in the bin per my understanding since it is allocated in PEI with memory allocation HOB.

And looking at the code flow, for PEI phase, still think the whole AllocateApLoopCodeBuffer() function and its caller has no need to be run in normal boot path at all.

niruiyu: + @td36

td36: AllocateApLoopCodeBuffer() is called by PrepareApLoopCode(). The PrepareApLoopCode() was introduced into PEI MP initialization flow in this commit https://github.com/tianocore/edk2/pull/5606/commits/3f511f420ca628ccc5fac1f9521da15ebf17125f. 

The commit reused PrepareApLoopCode() to prepare new safe buffer for AP in PEI MP initialization for S3 boot flow. Maybe it's better to only call PrepareApLoopCode() when S3 boot in PeiMpLib.c implementation.

makubacki: This will also fail this test point on Intel platforms that checks for more than one bucket per runtime type: [edk2-platforms/Platform/Intel/MinPlatformPkg/Test/Library/TestPointCheckLib
/DxeCheckMemoryMap.c - TestPointCheckUefiMemoryMapEntry()](https://github.com/tianocore/edk2-platforms/blob/8279b5c005cbc5366c7dff9b5b0920b7adb40156/Platform/Intel/MinPlatformPkg/Test/Library/TestPointCheckLib/DxeCheckMemoryMap.c#L113).

@lzeng14, I'll update the commit per your suggestion if I don't run into any issues during testing on a platform.

makubacki: Thanks all for the comments. I've made the change to only call `PrepareApLoopCode()` on S3 resume.

niruiyu (approved):

PR #11615, opened 2025-10-13
Fat pkg mde module pkg fix iso639 language bug

mdkinney:
# Description

PR #11380 that removed the Unicode Collation Protocol introduced a logic bug in the use of `GetBestLanguage()`. `Iso639Language` would always be computed to `TRUE` when it should always be passed to `GetBestLanguage()` as `FALSE`
once Unicode Collation Protocol is removed.

Update `InitializeUnicodeCollationSupportWorker()` in `FatPkg` and `MdeModulePkg` to remove the local variable `Iso639Language` and always call `GetBestLanguage()` with the `Iso639Language` parameter set to `FALSE`.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Platform boot failures observed before this change are resolved after this change.

## Integration Instructions

N/A


mdkinney:
@lgao4 @os-d Need a quick review of this change due to platform boot regression introduced by PR #11380

PR #11617, opened 2025-10-14
SecurityPkg/Tcg2ConfigPei: Add gEfiPeiMemoryDiscoveredPpiGuid dependency 

abdattar:
# Description
Introduce a dependency on gEfiPeiMemoryDiscoveredPpiGuid to ensure proper PPI sequencing and prevent failures when locating gEfiPeiReadOnlyVariable2PpiGuid.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on AMD platforms.

## Integration Instructions

N/A


jyao1:
I disagree. We purposely removed memory discovered dependency for any TPM module. That is because memory init may rely on TPM readiness in history.

Would you please describe the problem clearly? then we can think about the right solution

PR #11618, opened 2025-10-14
CryptoPkg: Update mbedtls submodule for EDKII

poloaegis83:
Update mbedtls submodule from 3.3.0 to 3.6.5 in CryptoPkg.

# Description

Update mbedtls submodule from 3.3.0 to 3.6.5,
EDK2 library and code need to updated also.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

1. Build test:
  edksetup.bat
  build -a X64 -p CryptoPkg\Test\CryptoPkgHostUnitTest.dsc -b NOOPT -t VS2019 -DCRYPTO_TEST_TYPE=MBEDTLS
  build -a X64 -p CryptoPkg\Test\CryptoPkgHostUnitTest.dsc -b NOOPT -t CLANGPDB -DCRYPTO_TEST_TYPE=MBEDTLS
  build -a X64 -p CryptoPkg\CryptoPkg.dsc -b NOOPT -t VS2019
  build -a X64 -p CryptoPkg\CryptoPkg.dsc -b NOOPT -t CLANGPDB
---------------------------------------------------------
2. Unit Test:
  There is one Unit Test fail after Mbedtls update:
  CryptoPkg.BaseCryptLib.Rsa.TestVerifyRsaGenerateKeyComponents()
  
  Analysis:
  In Mbedtls 3.6.4, Mbedtls API will check input parameter "nbits",
  this size should bigger than 1024 in new code, but in test we only given size 512.
  please check below change:
  [https://github.com/Mbed-TLS/mbedtls/pull/7864](url)

---------------------------------------------------------
3. Write Unit Test of Pkcs7GetAttachedContent()

  We have 5 test:
  1.CryptoPkg.BaseCryptLib.Pkcs7Content
      Test with signature with content data, function should return true.
  
  2.CryptoPkg.BaseCryptLib.TestPkcs7Detached
       Test with signature without content data, function should return true.
  
  3.CryptoPkg.BaseCryptLib.TestVerifyPkcs7ContentData
      Test with signature with content data, verify content data returned,
       it should same as content data in tested signature.
  
  4.CryptoPkg.BaseCryptLib.TestWithInvalidPkcs7
     Test with invalid PCK7,  function should return false.
  
  5.CryptoPkg.BaseCryptLib.TestWithInvalidParameters
     Test with Invalid Parameters as input,  function should return false.
  
  For Mbedtls:
  Only pass test 4 and 5, as it expect return value false.
  For OpenSSL:
  All Test passed.
  
  To Generate Test PCk7s Der format:
  1. openssl.exe genrsa -out key.pem 2048
  2. openssl.exe req -x509 -new -key key.pem -out cert.pem -days 365 -subj "/CN=Test PKCS7/"
  3. echo "Test PKCS7" > data.txt
  
  For Attached without authenticatedAttributes Signature:
  openssl.exe cms -sign -in data.txt -signer cert.pem -inkey key.pem -outform DER -out pkcs7_mbedtls.der -nodetach -noattr -binary
  For detach without authenticatedAttributes Signature:
  openssl.exe cms -sign -in data.txt -signer cert.pem -inkey key.pem -outform DER -out pkcs7_mbedtls.der -noattr -binary

## Integration Instructions

Below test case not supported, dropped in Unit Test for MbedTls:
TestVerifyEKUsWith3CertsInSignature()
TestVerifyEKUsWith2CertsInSignature()
TestNoEKUsInSignature()
TestVerifyRsaCertPkcs1SignVerify
TestVerifyPkcs7SignVerifyNonSelfIssued()
TestVerifyDhGenerateKey()
Pkcs1v2Decrypt
Pkcs1v2EncryptDecrypt
RsaOaepEncrypt (Interface)
RsaOaepEncrypt (NoSeed)
RsaOaepEncrypt (Seeded)
RsaOaepDecrypt
RsaOaepEncryptDecrypt
RsaOaepEncryptPkcs1v2Decrypt
Pkcs1v2EncryptRsaOaepDecrypt
RsaOaepEncrypt (MdDefaultMgf1Default)
RsaOaepDecrypt (MdDefaultMgf1Default)
RsaOaepEncryptDecrypt (MdDefaultMgf1Default)
RsaOaepEncrypt (MdSha1Bgf1Sha1
RsaOaepDecrypt (MdSha1Bgf1Sha1)
RsaOaepEncryptDecrypt (MdSha1Bgf1Sha1)
RsaOaepEncrypt (MdSha256Bgf1Sha256)
RsaOaepDecrypt (MdSha256Bgf1Sha256)
RsaOaepEncryptDecrypt (MdSha256Bgf1Sha256)
TestVerifyBn()
TestVerifyEcKey()
TestPkcs7Attached()
TestPkcs7Detached()
TestVerifyPkcs7ContentData()

Note:
For Pkcs7AttachedContentTest Mbedtls not support signatures with content data, so test case dropped.
[https://github.com/Mbed-TLS/mbedtls/commit/3dafc6c3b3a02bc19bb0fd54dbbd639d1c2ded47](url)
[https://github.com/Mbed-TLS/mbedtls/commit/6cfc4692961ba808d801ab2ffbf6e076523ccac0](url)

### DIFF
--- a/MdeModulePkg/Bus/Spi/SpiNorFlashJedecSfdp/SpiNorFlash.c
+++ b/MdeModulePkg/Bus/Spi/SpiNorFlashJedecSfdp/SpiNorFlash.c
@@ -20,6 +20,8 @@
 /**
   Fill Write Buffer with Opcode, Address, Dummy Bytes, and Data.
 
+  If WriteBytes is non-zero, WriteBuffer must be a valid buffer.
+
   @param[in]    Instance               The instance of SPI_NOR_FLASH
   @param[in]    Opcode                 Opcode for transaction
   @param[in]    DummyBytes             The dummy bytes send to SPI flash device
--- a/MdeModulePkg/Bus/Spi/SpiNorFlashJedecSfdp/SpiNorFlash.h
+++ b/MdeModulePkg/Bus/Spi/SpiNorFlashJedecSfdp/SpiNorFlash.h
@@ -16,12 +16,18 @@
 #include "SpiNorFlashJedecSfdpInternal.h"
 
 /**
-  Fill Write Buffer with Opcode, Address, Dummy Bytes, and Data
+  Fill Write Buffer with Opcode, Address, Dummy Bytes, and Data.
 
-  @param[in]    Opcode      - Opcode for transaction
-  @param[in]    Address     - SPI Offset Start Address
-  @param[in]    WriteBytes  - Number of bytes to write to SPI device
-  @param[in]    WriteBuffer - Buffer containing bytes to write to SPI device
+  If WriteBytes is non-zero, WriteBuffer must be a valid buffer.
+
+  @param[in]    Instance               The instance of SPI_NOR_FLASH
+  @param[in]    Opcode                 Opcode for transaction
+  @param[in]    DummyBytes             The dummy bytes send to SPI flash device
+  @param[in]    AddressBytesSupported  Bytes of address supported by SPI flash device
+  @param[in]    UseAddress             Send the address for SPI flash command
+  @param[in]    Address                SPI Offset Start Address
+  @param[in]    WriteBytes             Number of bytes to write to SPI device
+  @param[in]    WriteBuffer            Buffer containing bytes to write to SPI device
 
   @retval       Size of Data in Buffer
 **/
--- a/OvmfPkg/RiscVVirt/Library/PlatformSecLib/SecEntry.S
+++ b/OvmfPkg/RiscVVirt/Library/PlatformSecLib/SecEntry.S
@@ -15,7 +15,7 @@ li    s0, 0
 li    a2, FixedPcdGet32 (PcdOvmfSecPeiTempRamBase)
 li    a3, FixedPcdGet32 (PcdOvmfSecPeiTempRamSize)
 
-/* Reserve region to store handoff data
+/* Reserve region to store handoff data */
 li    a4, SEC_HANDOFF_DATA_RESERVE_SIZE
 sub   a3, a3, a4
 
--- a/UefiCpuPkg/Library/BaseRiscVMmuLib/BaseRiscVMmuLib.c
+++ b/UefiCpuPkg/Library/BaseRiscVMmuLib/BaseRiscVMmuLib.c
@@ -651,6 +651,7 @@ RiscVMmuSetSatpMode  (
   UINTN                            NumberOfDescriptors;
   UINTN                            Index;
   EFI_STATUS                       Status;
+  BOOLEAN                          InterruptState;
 
   if (SatpMode > PcdGet32 (PcdCpuRiscVMmuMaxSatpMode)) {
     return EFI_DEVICE_ERROR;
@@ -720,9 +721,7 @@ RiscVMmuSetSatpMode  (
 
   FreePool ((VOID *)MemoryMap);
 
-  if (GetInterruptState ()) {
-    DisableInterrupts ();
-  }
+  InterruptState = SaveAndDisableInterrupts ();
 
   Ppn = (UINT64)TranslationTable >> RISCV_MMU_PAGE_SHIFT;
   ASSERT (!(Ppn & ~(SATP64_PPN)));
@@ -745,9 +744,7 @@ RiscVMmuSetSatpMode  (
 
   RiscVLocalTlbFlushAll ();
 
-  if (GetInterruptState ()) {
-    EnableInterrupts ();
-  }
+  SetInterruptState (InterruptState);
 
   return Status;
 }
--- a/UefiPayloadPkg/Library/FdtParserLib/FdtParserLib.c
+++ b/UefiPayloadPkg/Library/FdtParserLib/FdtParserLib.c
@@ -765,7 +765,7 @@ ParsePciRootBridge (
 
     if (AsciiStrCmp (TempStr, "ranges") == 0) {
       DEBUG ((DEBUG_INFO, "  Found ranges Property TempLen (%08X), limit %x\n", TempLen, TempLen / sizeof (UINT32)));
-
+      // TODO:  In future we should fetch these values from fdt and avoid using these Pcds
       mPciRootBridgeInfo->RootBridge[RbIndex].AllocationAttributes = EFI_PCI_HOST_BRIDGE_COMBINE_MEM_PMEM | EFI_PCI_HOST_BRIDGE_MEM64_DECODE;
       mPciRootBridgeInfo->RootBridge[RbIndex].Supports             = ROOT_BRIDGE_SUPPORTS_DEFAULT;
       mPciRootBridgeInfo->RootBridge[RbIndex].PMemAbove4G.Base     = PcdGet64 (PcdPciReservedPMemAbove4GBBase);
@@ -774,6 +774,7 @@ ParsePciRootBridge (
       mPciRootBridgeInfo->RootBridge[RbIndex].PMem.Limit           = PcdGet32 (PcdPciReservedPMemLimit);
       mPciRootBridgeInfo->RootBridge[RbIndex].UID                  = RbIndex;
       mPciRootBridgeInfo->RootBridge[RbIndex].HID                  = EISA_PNP_ID (0x0A03);
+      mPciRootBridgeInfo->RootBridge[RbIndex].DmaAbove4G           = FALSE;
 
       Data32 = (UINT32 *)(PropertyPtr->Data);
       for (Base = 0; Base < TempLen / sizeof (UINT32); Base = Base + DWORDS_TO_NEXT_ADDR_TYPE) {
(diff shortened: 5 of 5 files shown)


liyi77:
> How This Was Tested
> edksetup.bat
> build -a X64 -p CryptoPkg\Test\CryptoPkgHostUnitTest.dsc -b NOOPT -t VS2022

The default build of test agent is openssl.
Please use:
 build -a X64 -p CryptoPkg\Test\CryptoPkgHostUnitTest.dsc -b NOOPT -t VS2022 **-DCRYPTO_TEST_TYPE=MBEDTLS**

liyi77:
Please add issue link to commit message as:
```REF: https://github.com/tianocore/edk2/issues/11605```

**Comment by poloaegis83** [comment]
> > How This Was Tested
> > edksetup.bat
> > build -a X64 -p CryptoPkg\Test\CryptoPkgHostUnitTest.dsc -b NOOPT -t VS2022
> 
> The default build of test agent is openssl. Please use: build -a X64 -p CryptoPkg\Test\CryptoPkgHostUnitTest.dsc -b NOOPT -t VS2022 **-DCRYPTO_TEST_TYPE=MBEDTLS**

thank you!
Seems a lot of build errors need to be fix

File: CryptoPkg/Library/BaseCryptLibMbedTls/Pk/CryptPkcs7VerifyBase.c [hunk 1]
```diff
@@ -73,11 +75,10 @@ Pkcs7GetAttachedContent (
   if (Ret != MBEDTLS_PKCS7_SIGNED_DATA) {
     goto _Exit;
   }
-
   //
   // Check for detached or attached content
   //
-  MbedtlsContent = &(Pkcs7.signed_data.content);
+  MbedtlsContent = &(Pkcs7Attached.ContentInfo);
```
liyi77: pkcs7 changes make me confused:
1. why init mbedtls_pkcs7 Pkcs7 with type MbedtlsPkcs7SignedData, signed data should be a part of Pkcs7 struct.
2. Pkcs7Attached is not assigned before use.

I assume you want to use the MbedtlsPkcs7SignedData type in EDK2 to parse content from the pkcs7 structure in Mbedtls - I agree with this concept, but need to implement a test case to ensure the correctness of this change.

CryptoPkg/Library/MbedTlsLib/MbedTlsLib.inf
```diff
@@ -94,16 +95,24 @@
   MemoryAllocationLib
 
 [BuildOptions]
+
+  #
+  # Define UINT64_MAX for all case
+  #
+  *_*_*_CC_FLAGS = -U UINT64_MAX -D UINT64_MAX=0xFFFFFFFFFFFFFFFFULL
```
liyi77: could move it to [CrtLibSupport.h](https://github.com/tianocore/edk2/pull/11618/files#diff-9e80fe17102d95b7c7e9f5dfd48854598fab961cffca342d0b70a523f4ca55b1)

ardbiesheuvel:
@liyi77 do we have sufficient coverage in CI to confirm that PKCS#7 authentication still works as expected after this change?

liyi77:
> @liyi77 do we have sufficient coverage in CI to confirm that PKCS#7 authentication still works as expected after this change?

Except for Pkcs7GetAttachedContent(), all other PKCS#7 functions have test cases in CI to cover. 
@poloaegis83 will add test case for it in this upgrading.

CryptoPkg/Library/BaseCryptLibMbedTls/Pk/CryptPkcs7Internal.h
```diff
@@ -18,6 +18,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include "mbedtls/asn1.h"
 #include "mbedtls/asn1write.h"
 #include "mbedtls/x509.h"
+#include "x509_internal.h"
```
liyi77, line 21: do we still need to expose this header? I think we could remove it because dont need the changes of PKCS7

CryptoPkg/Test/UnitTest/Library/BaseCryptLib/Pkcs7AttachedContentSignatures.h
```diff
@@ -0,0 +1,299 @@
+/** @file
+  Application for Pkcs7 Attached Content Validation.
+  It is header file for Pkcs7 Signatures test case.
+
+Copyright (c) 2025, Intel Corporation. All rights reserved.<BR>
+SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+CONST UINT8  AttachedContents[] = {
```
liyi77, line 22: please add cmd for test cert generation.

CryptoPkg/Test/UnitTest/Library/BaseCryptLib/Pkcs7AttachedContentTest.c
```diff
@@ -0,0 +1,290 @@
...
+  @param[in]  Framework - Unit-test framework handle.
+  @param[in]  Context   - Optional context pointer for this test.
+
+  @retval UNIT_TEST_PASSED            - The required pkcs7 detached contents return be null.
+  @retval UNIT_TEST_ERROR_TEST_FAILED - Something failed, check the debug output.
+**/
+/*static
+UNIT_TEST_STATUS
+EFIAPI
+TestPkcs7DetachedNoAttributes (
```
liyi77: remove it if dont need

liyi77:
please enable ready for review if ready.

jyao1:
There is more security fix in https://github.com/Mbed-TLS/mbedtls/releases/tag/mbedtls-3.6.5.

Can we jump to 3.6.5 directly?

CryptoPkg/Test/UnitTest/Library/BaseCryptLib/Pkcs7AttachedContentSignatures.h
poloaegis83: Added a comment for Pck7 der generation

liyi77 (approved):

poloaegis83:
> There is more security fix in https://github.com/Mbed-TLS/mbedtls/releases/tag/mbedtls-3.6.5.
> 
> Can we jump to 3.6.5 directly?

This patch update to 3.6.5, also description updated, thanks

liyi77:
> The tests below are failing:
> 
> ```
> [  FAILED  ] TestVerifyEKUsWith3CertsInSignature()
> [  FAILED  ] TestVerifyEKUsWith2CertsInSignature()
> [  FAILED  ] TestNoEKUsInSignature()
> [  FAILED  ] TestVerifyRsaGenerateKeyComponents()
> [  FAILED  ] TestVerifyPkcs7SignVerifyNonSelfIssued()
> [  FAILED  ] Pkcs1v2Decrypt
> [  FAILED  ] Pkcs1v2EncryptDecrypt
> [  FAILED  ] RsaOaepEncrypt (Interface)
> [  FAILED  ] RsaOaepEncrypt (NoSeed)
> [  FAILED  ] RsaOaepEncrypt (Seeded)
> [  FAILED  ] RsaOaepDecrypt
> [  FAILED  ] RsaOaepEncryptDecrypt
> [  FAILED  ] RsaOaepEncryptPkcs1v2Decrypt
> [  FAILED  ] Pkcs1v2EncryptRsaOaepDecrypt
> [  FAILED  ] RsaOaepEncrypt (MdDefaultMgf1Default)
> [  FAILED  ] RsaOaepDecrypt (MdDefaultMgf1Default)
> [  FAILED  ] RsaOaepEncryptDecrypt (MdDefaultMgf1Default)
> [  FAILED  ] RsaOaepEncrypt (MdSha1Bgf1Sha1
> [  FAILED  ] RsaOaepDecrypt (MdSha1Bgf1Sha1)
> [  FAILED  ] RsaOaepEncryptDecrypt (MdSha1Bgf1Sha1)
> [  FAILED  ] RsaOaepEncrypt (MdSha256Bgf1Sha256)
> [  FAILED  ] RsaOaepDecrypt (MdSha256Bgf1Sha256)
> [  FAILED  ] RsaOaepEncryptDecrypt (MdSha256Bgf1Sha256)
> [  FAILED  ] TestVerifyEcKey()
> [  FAILED  ] TestPkcs7Attached()
> [  FAILED  ] TestPkcs7Detached()
> [  FAILED  ] TestVerifyPkcs7ContentData()
> ```
> 
> Even if this was the case before, I don't think we should make drastic changes to the MbedTLS support until we address this, either by disabling those tests in the MbedTLS build if we don't care about the functionality, or by fixing the code so the test passes.

This is expected because CryptoLibMbedtls haven't implemented these features, we can only guarantee that cases that passed before the upgrade will also pass after the upgrade.
>disabling those tests in the MbedTLS build if we don't care about the functionality

@poloaegis83 Could you create a test host library for Mbedtls so we can avoid this kind of confusion?

ardbiesheuvel:
> > Even if this was the case before, I don't think we should make drastic changes to the MbedTLS support until we address this, either by disabling those tests in the MbedTLS build if we don't care about the functionality, or by fixing the code so the test passes.
> 
> This is expected because CryptoLibMbedtls haven't implemented these features, we can only guarantee that cases that passed before the upgrade will also pass after the upgrade.
> 

So which test cases need to pass to ensure that the functionality we need for secure boot, measured boot, etc is working as expected?

> > disabling those tests in the MbedTLS build if we don't care about the functionality
> 
> @poloaegis83 Could you create a test host library for Mbedtls so we can avoid this kind of confusion?

Yes, please. And we need to document which use cases are supported by MbedTLS, and which use cases are only supported when using OpenSSL.

liyi77:
> So which test cases need to pass to ensure that the functionality we need for secure boot, measured boot, etc is working as expected?

Test cases for Hash, Rsa/Pkcs#7/X509 verify is necessary for those feature, and already covered by the current test suite.
@poloaegis83 has isolated an Mbedtls test suite in CryptoPkg/Test/UnitTest/Library/BaseCryptLib/BaseCryptLibUnitTestsMbedTls.c, please check.

PR #11620, opened 2025-10-14
MdeModulePkg: ArmFfaLib: Add FFA_YIELD handling

kuqin12:
# Description

If a secure partition on AArch64 platforms would consume extended time to operate with peripherals, it might elect to yield the control back to normal world and expect the normal world to callback after specified period of time.

This change adds the FFA_YIELD handling from ArmFfaLib to support long operations from secure partitions.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested on proprietary hardware platforms and booted to Windows.

## Integration Instructions

Platforms will need to add a `TimerLib` instance for the corresponding phase.

### DIFF
--- a/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
+++ b/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
@@ -143,7 +143,7 @@ FfaArgsToEfiStatus (
      * FfaStatusToEfiStatus (FfaARgs.Arg2) returns proper EFI_STATUS.
      */
     FfaStatus = ARM_FFA_RET_NOT_SUPPORTED;
-  } else if (FfaArgs->Arg0 == ARM_FID_FFA_INTERRUPT) {
+  } else if ((FfaArgs->Arg0 == ARM_FID_FFA_INTERRUPT) || (FfaArgs->Arg0 == ARM_FID_FFA_YIELD)) {
     FfaStatus = ARM_FFA_RET_INTERRUPTED;
   } else {
     FfaStatus = ARM_FFA_RET_SUCCESS;


LeviYeoReum (approved):

kuqin12:
@sachinami would you mind reviewing the change and providing feedback? Thanks in advance!

LeviYeoReum (approved):

sachinami:
> @sachinami would you mind reviewing the change and providing feedback? Thanks in advance!

Looks good to me @kuqin12 .

kuqin12:
@samimujawar could you please also provide feedback for this PR? Thanks

kuqin12:
@samimujawar could you please provide some feedback on this change? Any input is appreciated.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -144,6 +146,20 @@ FfaArgsToEfiStatus (
...
+    /*
+    * If the FF-A ABI indicates that the call was yielded, we need to pull the
+    * timeout period from Arg2 and Arg3.
+    */
+    Timeout = LShiftU64 (FfaArgs->Arg3, 32) | FfaArgs->Arg2;
+    if (Timeout != 0) {
+      /*
+      * If the timeout is non-zero, we handle the delay here.
+      */
+      NanoSecondDelay (Timeout);
```
ardbiesheuvel: This is a function that translates a FFA status to a EFI status, so it doesn't seem like the appropriate place to insert the waiting logic.

Also, doesn't a yield imply that the call needs to be repeated after the timeout expires?

kuqin12: Yeah, on the first point, I did not seem to find a proper mapping of FFA_YIELD status in EFI_STATUS domain, which led me to hijack this interrupted return value. If you have other suggestions, I would be happy to change to that return value and redo the PR so that the caller can handle it.

On the second point, my intention with this implementation is that `FFA_RUN` after the delay should be attempted from the FFA function caller side (i.e. the TPM over FFA library that invokes FFA_DIRECT_REQ2 function). But this is also tangled with the first point... If we have a more appropriate return code, we can propagate the code up to the caller and have caller handle this yield. If they elect to wait they can do delay there, otherwise they can return `EFI_TIMEOUT`.

Please let me know if you have other suggestions. Thanks in advance.

ardbiesheuvel: Having looked at the FFA spec now, I'm not convinced we need to do anything at all with the timeout values, and we can just treat this as `ARM_FID_FFA_INTERRUPT`.

The timeout is a hint to the hypervisor that there is no need to schedule the vCPU for a while, which means that

* the hypervisor may have already applied the timeout before scheduling the vCPU (if we are indeed running under a hypervisor), and we end up waiting twice the suggested time;
* we are not doing anything meaningful during this time (`NanoSecondDelay()` just spins until the timeout expires), and so calling `FFA_RUN` prematurely has the same effect, even if it ends up doing a busy-wait loop itself.

kuqin12: Thanks for the input. It seems that you are indicating to just ignore the timeout information in this case. I was a bit uneasy to take that approach when I initially implemented this change. But I think this is actually fine. Will update shortly.

kuqin12: And this is updated.

ardbiesheuvel (approved):

PR #11623, opened 2025-10-15
UnitTestFrameworkPkg: Use 8MB stack for MSFT and CLANGPDB

mdkinney:
# Description

Linux/GCC host-based unit test application builds use a default stack size of 8MB. Windows/VS20xx and Windows/CLANGPDB host-based unit test application builds use a default stack size of 1MB. This can allow Linux unit tests to pass and Windows unit tests to fail with stack overflow if large local variables are used. ASAN increases stack usage, so this condition can occur more frequently when ASAN enabled.

Update MSFT and CLANGPDB for host-based unit tests to use a default stack size of 8MB so all tool chains use the same default stack size for host-based unit tests.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Observed unit tests that fails with stack overflow exception with VS20xx, but not GCC.  With this change, the stack overflow exception does not occur and the same unit tests results are generated by both VS20xx and GCC.

## Integration Instructions

N/A

### DIFF
--- a/UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
+++ b/UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
@@ -87,7 +87,7 @@
   #
   # Set MSFT HOST_APPLICATION options
   #
-  MSFT:*_*_*_DLINK_FLAGS            == /nologo /SUBSYSTEM:CONSOLE /DEBUG /out:"$(BIN_DIR)\$(MODULE_NAME_GUID).exe" /pdb:"$(BIN_DIR)\$(MODULE_NAME_GUID).pdb" /WHOLEARCHIVE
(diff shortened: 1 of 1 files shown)


makubacki:
If I understand correctly, this is not failing in edk2 CI or pipelines would be broken.

It might be worth calling out in the UnitTestFrameworkPkg documentation how a platform can adjust their stack size locally in their DSC BuildOptions section if they still encounter a stack overflow since this is a somewhat arbitrary default (which is fine given that it's ample and consistent across toolchains) and I don't thinks we'd want to keep increasing the default further.

mdkinney:
> If I understand correctly, this is not failing in edk2 CI or pipelines would be broken.
> 
> It might be worth calling out in the UnitTestFrameworkPkg documentation how a platform can adjust their stack size locally in their DSC BuildOptions section if they still encounter a stack overflow since this is a somewhat arbitrary default (which is fine given that it's ample and consistent across toolchains) and I don't thinks we'd want to keep increasing the default further.

I agree we do not want to increase further.  Just align all the supported tool chains to the same amount of stack space.

I also agree that if these limits are exceeded, and unit tests are failing for all tool chains, then the unit test developer has a few options
* Convert local variables to variables in unit tests class object (Google Test)
* Convert local variables to pointers and add code to allocate/free buffers (Framework and Google Test)
* Increase stack size in host-based unit test DSC (Framework and Google Test)

PR #11624, opened 2025-10-15
OvmfPkg/IoMmuDxe: Fix 1M and 2M buffer handling

osteffenrh:
# Description

The IoMmu keeps a pool of pre-allocated shared buffers in various sizes
to serve requests. Usage is tracked in a bitmap.

The bitmap masks for the 1M and 2M buffer pools are incorrect, causing
the same buffers getting handed out repeatedly, causing corrupted device
accesses.

The masks needs to be kept in sync with mReservedMemRanges below.
This sets the correct values for:

- RESERVED_MEM_BITMAP_1M_MASK = (1 << 14) = 0x4000
- RESERVED_MEM_BITMAP_2M_MASK = (1 << 15) | (1 << 16) = 0x18000

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Booting a certain iso image under AMD SEV can trigger the issue. For example a current openSUSE image will do.
The loaded GRUB image is corrupted and crashes.


osteffenrh:
cc: @kraxel

PR #11621, opened 2025-10-15
Update UefiPayloadEntry to use MemoryAllocationLib

kuqin12:
# Description

With the recent addition of MemoryAllocationLib in the ArmFfaLib for SEC phase, the UefiPayloadEntry will
stop working if it needs to use FF-A interface.

This change moves the internal MemoryAllocation logic to a separate MemoryAllocationLib after breaking the
Hob creation function into its own library class.

- [x] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [ ] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

## How This Was Tested

This was tested on proprietary ARM64 hardware and booted to Windows.

## Integration Instructions

UefiPayloadEntry will need to add the following for the SEC phase:
`MemoryAllocationLib|UefiPayloadPkg/Library/PayloadEntryMemoryAllocationLib/PayloadEntryMemoryAllocationLib.inf`

PR #11625, opened 2025-10-16
MdePkg/PeilessHobLib: Add PeilessHobLib to remove dependency on PeiCore

pingy1:
Currently, Hoblib implementation uses PEI Services to retrieve the HOB List, which is unavailable when PEI Service is not initialized. so add PeilessHobLib to remove dependency on PeiCore

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions

### DIFF
--- a/MdePkg/Library/PeilessHobLib/HobLib.c
+++ b/MdePkg/Library/PeilessHobLib/HobLib.c
@@ -1,0 +1,994 @@
+/** @file
+  Provide Hob Library functions for PEI phase.
+
+Copyright (c) 2025, Intel Corporation. All rights reserved.<BR>
+SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include <PiPei.h>
+#include <Guid/MemoryAllocationHob.h>
+#include <Library/HobLib.h>
+#include <Library/DebugLib.h>
+#include <Library/BaseMemoryLib.h>
+
+/**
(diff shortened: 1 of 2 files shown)


MdePkg/Library/PeilessHobLib/HobLib.c
```diff
@@ -0,0 +1,994 @@
...
+  @param Hob              Pointer to the new HOB.
+
+  @return  EFI_SUCCESS           Success to create HOB.
+  @retval  EFI_INVALID_PARAMETER if Hob is NULL
+  @retval  EFI_NOT_AVAILABLE_YET if HobList is still not available.
+  @retval  EFI_OUT_OF_RESOURCES  if there is no more memory to grow the Hoblist.
+
+**/
+EFI_STATUS
+PeilessCreateHob (
```
kuqin12, line 219: Not entirely related to your PR specifically.. but I think this function is a tricky yet repeated function. All sec phase components use something like this to back up the HOB library. But some hobs are also created for other library implementations, such as memory allocation library for pool allocation.

Should we define this interface separately and centralize the hob lib into one? I ran into an issue like this in UefiPayloadPkg recently and created this PR:

https://github.com/tianocore/edk2/pull/11621

But I think a common implementation would be better for the common welfare.

PR #11626, opened 2025-10-16
IntelFsp2Pkg/Tools/ConfigEditor: Fixed YAML data overwrite issue

yunhsiax:
Problem Description:
The condition if force or act_cfg['value'] == '' in _set_field_value() caused valid YAML values 
to be wrongly overwritten with binary data. 
Since force=True made the condition always true, correct YAML entries like "0xFE02C000" were replaced 
by binary values such as "0x01010101", leading to corrupted configurations and incorrect FSP settings.

Solution Implementation:
The condition was updated to if act_cfg['value'] == '', removing the force dependency. 
This ensures binary data is used only when the YAML value is empty, 
preserving valid configurations and preventing overwrites.


tedkuo1 (changes requested):
This should be a right solution for the issue.

PR #11634, opened 2025-10-21
SecurityPkg: Introduce Tpm2HelpLib

### LINKED ISSUE #12799: [Breaking Change Deprecation]: Replace Tpm2Help.c with Tpm2HelpLib
### Special Notes

Subissue of breaking change Tracking issue: https://github.com/tianocore/edk2/issues/12797

Tracks the deprecation of Tpm2Help.c API in Tpm2CommandLib.

Raymond-MS:
# Description

Decoupled Tpm2Help.c from Tpm2CommandLib. This allows for use of the helper functions without the need to include Tpm2CommandLib and Tpm2DeviceLib which is usually tied to the TPM and TPM communication.

There are cases where you may need to generate a HOB for pre-DXE measurements. These cases utilize these helper functions but do not require Tpm2CommandLib or Tpm2DeviceLib as there is no communication to the TPM needed. An example is there is no Tpm2DeviceLib for SEC.

- [x] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Ran CI locallly.

## Integration Instructions

Include Tpm2HelpLib in place of Tpm2CommandLib if necessary, otherwise, just add Tpm2HelpLib to the .inf and include the header in the .c.

The helper functions in Tpm2Help.c in Tpm2CommandLib are deprecated as of the inclusion of this library. They are wrappers in place of the Tpm2HelpLib versions. Update all calls to Tpm2Help.c functions to use Tpm2HelpLib versions instead.

### DIFF
--- a/SecurityPkg/Include/Library/Tpm2HelpLib.h
+++ b/SecurityPkg/Include/Library/Tpm2HelpLib.h
@@ -1,0 +1,186 @@
+/** @file
+  Declares reusable TPM 2.0 data helper routines that are independent
+  of TPM command/device communication libraries. Includes serialization
+  and parsing helpers for common TPM data structures.
+
+Copyright (c) Microsoft Corporation.
+Copyright (c) 2013 - 2024, Intel Corporation. All rights reserved. <BR>
+SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#pragma once
+
+#include <IndustryStandard/Tpm20.h>
+
+/**
+  Return size of digest.
+
+  @param[in] HashAlgo  Hash algorithm
+
+  @return size of digest
+**/
+UINT16
+EFIAPI
+Tpm2GetHashSizeFromAlgo (
+  IN TPMI_ALG_HASH  HashAlgo
+  );
+
+/**
+  Get hash mask from algorithm.
+
+  @param[in] HashAlgo   Hash algorithm
+
+  @return Hash mask
+**/
+UINT32
+EFIAPI
+Tpm2GetHashMaskFromAlgo (
+  IN TPMI_ALG_HASH  HashAlgo
+  );
+
+/**
+  Copy AuthSessionIn to TPM2 command buffer.
+
+  @param [in]  AuthSessionIn   Input AuthSession data
+  @param [out] AuthSessionOut  Output AuthSession data in TPM2 command buffer
+
+  @return AuthSession size
+**/
+UINT32
+EFIAPI
+Tpm2CopyAuthSessionCommand (
+  IN      TPMS_AUTH_COMMAND  *AuthSessionIn  OPTIONAL,
+  OUT     UINT8              *AuthSessionOut
+  );
+
+/**
+  Copy AuthSessionIn from TPM2 response buffer.
+
+  @param [in]  AuthSessionIn   Input AuthSession data in TPM2 response buffer
+  @param [out] AuthSessionOut  Output AuthSession data
+
+  @return AuthSession size
+**/
+UINT32
+EFIAPI
+Tpm2CopyAuthSessionResponse (
+  IN      UINT8               *AuthSessionIn,
+  OUT     TPMS_AUTH_RESPONSE  *AuthSessionOut OPTIONAL
+  );
+
+/**
+  Return if hash alg is supported in HashAlgorithmMask.
+
+  @param HashAlg            Hash algorithm to be checked.
+  @param HashAlgorithmMask  Bitfield of allowed hash algorithms.
+
+  @retval TRUE  Hash algorithm is supported.
+  @retval FALSE Hash algorithm is not supported.
+**/
+BOOLEAN
+EFIAPI
+Tpm2IsHashAlgSupportedInHashAlgorithmMask (
+  IN TPMI_ALG_HASH  HashAlg,
+  IN UINT32         HashAlgorithmMask
+  );
+
+/**
+  Copy TPML_DIGEST_VALUES into a buffer
+
+  @param[in,out] Buffer             Buffer to hold copied TPML_DIGEST_VALUES compact binary.
+  @param[in]     DigestList         TPML_DIGEST_VALUES to be copied.
+  @param[in]     HashAlgorithmMask  HASH bits corresponding to the desired digests to copy.
+
+  @return The end of buffer to hold TPML_DIGEST_VALUES.
+**/
+VOID *
+EFIAPI
+Tpm2CopyDigestListToBuffer (
+  IN OUT VOID            *Buffer,
+  IN TPML_DIGEST_VALUES  *DigestList,
+  IN UINT32              HashAlgorithmMask
+  );
+
+/**
+  Copy a buffer into a TPML_DIGEST_VALUES structure.
+
+  @param[in]     Buffer             Buffer to hold TPML_DIGEST_VALUES compact binary.
+  @param[in]     BufferSize         Size of Buffer.
+  @param[out]    DigestList         TPML_DIGEST_VALUES.
+
+  @return EFI_STATUS
+  @retval EFI_SUCCESS               Buffer was successfully copied to Digest List.
+  @retval EFI_BAD_BUFFER_SIZE       Bad buffer size passed to function.
+  @retval EFI_INVALID_PARAMETER     Invalid parameter passed to function: NULL pointer or
+                                    BufferSize bigger than TPML_DIGEST_VALUES
+**/
+EFI_STATUS
+EFIAPI
+Tpm2CopyBufferToDigestList (
+  IN CONST  VOID                *Buffer,
+  IN        UINTN               BufferSize,
+  OUT       TPML_DIGEST_VALUES  *DigestList
+  );
+
+/**
+  Get TPML_DIGEST_VALUES data size.
+
+  @param[in]     DigestList    TPML_DIGEST_VALUES data.
+
+  @return TPML_DIGEST_VALUES data size.
+**/
+UINT32
+EFIAPI
+Tpm2GetDigestListSize (
+  IN TPML_DIGEST_VALUES  *DigestList
+  );
+
+/**
+  Get TPML_DIGEST_VALUES data size from HashAlgorithmMask
+
+  @param[in]     HashAlgorithmMask    Bitfield of allowed hash algorithms
+
+  @return TPML_DIGEST_VALUES data size.
+**/
+UINT32
+EFIAPI
+Tpm2GetDigestListSizeFromHashAlgorithmMask (
+  IN UINT32  HashAlgorithmMask
+  );
+
+/**
+  This function get digest from digest list.
+
+  @param[in]  HashAlg       Digest algorithm
+  @param[in]  DigestList    Digest list
+  @param[out] Digest        Digest
+
+  @retval EFI_SUCCESS            Digest is found and returned.
+  @retval EFI_NOT_FOUND          Digest is not found.
+  @retval EFI_INVALID_PARAMETER  DigestList or Digest invalid.
+**/
+EFI_STATUS
+EFIAPI
+Tpm2GetDigestFromDigestList (
+  IN TPMI_ALG_HASH       HashAlg,
+  IN TPML_DIGEST_VALUES  *DigestList,
+  OUT VOID               *Digest
+  );
+
+/**
+  Check if all hash algorithms supported in HashAlgorithmMask are
+  present in the DigestList.
+
+  @param DigestList         Digest list
+  @param HashAlgorithmMask  Bitfield of allowed hash algorithms.
+
+  @retval TRUE  All hash algorithms present.
+  @retval FALSE Some hash algorithms not present.
+**/
+BOOLEAN
+EFIAPI
+Tpm2IsDigestListInSyncWithHashAlgorithmMask (
+  IN TPML_DIGEST_VALUES  *DigestList,
+  IN UINT32              HashAlgorithmMask
+  );
--- a/SecurityPkg/Library/PeilessSecMeasureLib/PeilessSecMeasureLib.c
+++ b/SecurityPkg/Library/PeilessSecMeasureLib/PeilessSecMeasureLib.c
@@ -17,6 +17,7 @@
 #include <Library/DebugLib.h>
 #include <Library/BaseMemoryLib.h>
 #include <Library/Tpm2CommandLib.h>
+#include <Library/Tpm2HelpLib.h>
 #include <Library/HashLib.h>
 #include <Library/HobLib.h>
 #include <Library/MemoryAllocationLib.h>
@@ -68,12 +69,12 @@ LogHashEvent (
   UINT8           *DigestBuffer;
 
   //
-  // Use GetDigestListSize (DigestList) in the GUID HOB DataLength calculation
+  // Use Tpm2GetDigestListSize (DigestList) in the GUID HOB DataLength calculation
   // to reserve enough buffer to hold TPML_DIGEST_VALUES compact binary.
   //
   HobData = BuildGuidHob (
               &gTcgEvent2EntryHobGuid,
-              sizeof (TcgPcrEvent2->PCRIndex) + sizeof (TcgPcrEvent2->EventType) + GetDigestListSize (DigestList) + sizeof (TcgPcrEvent2->EventSize) + NewEventHdr->EventSize
+              sizeof (TcgPcrEvent2->PCRIndex) + sizeof (TcgPcrEvent2->EventType) + Tpm2GetDigestListSize (DigestList) + sizeof (TcgPcrEvent2->EventSize) + NewEventHdr->EventSize
               );
   if (HobData == NULL) {
     return EFI_OUT_OF_RESOURCES;
@@ -90,7 +91,7 @@ LogHashEvent (
    *     - TPM_ALG_SHA384
    *     - TPM_ALG_SHA512
    */
-  DigestBuffer = CopyDigestListToBuffer (
+  DigestBuffer = Tpm2CopyDigestListToBuffer (
                    DigestBuffer,
                    DigestList,
                    TPM_ALG_SHA256 | TPM_ALG_SHA384 | TPM_ALG_SHA512
--- a/SecurityPkg/Library/Tpm2HelpLib/Tpm2Help.c
+++ b/SecurityPkg/Library/Tpm2HelpLib/Tpm2Help.c
@@ -1,0 +1,562 @@
+/** @file
+  Defines reusable TPM 2.0 data helper routines that are independent
+  of TPM command/device communication libraries. Includes serialization
+  and parsing helpers for common TPM data structures.
+
+Copyright (c) Microsoft Corporation.
+Copyright (c) 2013 - 2018, Intel Corporation. All rights reserved. <BR>
+SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include <IndustryStandard/UefiTcgPlatform.h>
+#include <Library/Tpm2HelpLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/BaseLib.h>
+#include <Library/DebugLib.h>
+
+typedef struct {
+  TPMI_ALG_HASH    HashAlgo;
+  UINT16           HashSize;
+  UINT32           HashMask;
+} INTERNAL_HASH_INFO;
+
+STATIC INTERNAL_HASH_INFO  mHashInfo[] = {
+  { TPM_ALG_SHA1,    SHA1_DIGEST_SIZE,    HASH_ALG_SHA1    },
+  { TPM_ALG_SHA256,  SHA256_DIGEST_SIZE,  HASH_ALG_SHA256  },
+  { TPM_ALG_SM3_256, SM3_256_DIGEST_SIZE, HASH_ALG_SM3_256 },
+  { TPM_ALG_SHA384,  SHA384_DIGEST_SIZE,  HASH_ALG_SHA384  },
+  { TPM_ALG_SHA512,  SHA512_DIGEST_SIZE,  HASH_ALG_SHA512  },
+};
+
+/**
+  Check if DigestList has an entry for HashAlg.
+
+  @param DigestList         Digest list.
+  @param HashAlg            Hash algorithm id.
+
+  @retval TRUE  Match found.
+  @retval FALSE No match found.
+**/
+STATIC
+BOOLEAN
+CheckDigestListForHashAlg (
+  IN TPML_DIGEST_VALUES  *DigestList,
+  IN TPM_ALG_ID          HashAlg
+  )
+{
+  UINT32  Index;
+
+  if (DigestList->count > HASH_COUNT) {
+    return FALSE;
+  }
+
+  for (Index = 0; Index < DigestList->count; Index++) {
+    if (DigestList->digests[Index].hashAlg == HashAlg) {
+      DEBUG ((DEBUG_INFO, "Hash alg 0x%x found in DigestList.\n", HashAlg));
+      return TRUE;
+    }
+  }
+
+  DEBUG ((DEBUG_INFO, "Hash alg 0x%x not found in DigestList.\n", HashAlg));
+  return FALSE;
+}
+
+/**
+  Return size of digest.
+
+  @param[in] HashAlgo  Hash algorithm
+
+  @return size of digest
+**/
+UINT16
+EFIAPI
+Tpm2GetHashSizeFromAlgo (
+  IN TPMI_ALG_HASH  HashAlgo
+  )
+{
+  UINTN  Index;
+
+  for (Index = 0; Index < sizeof (mHashInfo)/sizeof (mHashInfo[0]); Index++) {
+    if (mHashInfo[Index].HashAlgo == HashAlgo) {
+      return mHashInfo[Index].HashSize;
+    }
+  }
+
+  return 0;
+}
+
+/**
+  Get hash mask from algorithm.
+
+  @param[in] HashAlgo   Hash algorithm
+
+  @return Hash mask
+**/
+UINT32
+EFIAPI
+Tpm2GetHashMaskFromAlgo (
+  IN TPMI_ALG_HASH  HashAlgo
+  )
+{
+  UINTN  Index;
+
+  for (Index = 0; Index < sizeof (mHashInfo)/sizeof (mHashInfo[0]); Index++) {
+    if (mHashInfo[Index].HashAlgo == HashAlgo) {
+      return mHashInfo[Index].HashMask;
+    }
+  }
+
+  return 0;
+}
+
+/**
+  Copy AuthSessionIn to TPM2 command buffer.
+
+  @param [in]  AuthSessionIn   Input AuthSession data
+  @param [out] AuthSessionOut  Output AuthSession data in TPM2 command buffer
+
+  @return AuthSession size
+**/
+UINT32
+EFIAPI
+Tpm2CopyAuthSessionCommand (
+  IN      TPMS_AUTH_COMMAND  *AuthSessionIn  OPTIONAL,
+  OUT     UINT8              *AuthSessionOut
+  )
+{
+  UINT8  *Buffer;
+
+  if (AuthSessionOut == NULL) {
+    return 0;
+  }
+
+  Buffer = (UINT8 *)AuthSessionOut;
+
+  //
+  // Add in Auth session
+  //
+  if (AuthSessionIn != NULL) {
+    //  sessionHandle
+    WriteUnaligned32 ((UINT32 *)Buffer, SwapBytes32 (AuthSessionIn->sessionHandle));
+    Buffer += sizeof (UINT32);
+
+    // nonce
+    WriteUnaligned16 ((UINT16 *)Buffer, SwapBytes16 (AuthSessionIn->nonce.size));
+    Buffer += sizeof (UINT16);
+
+    CopyMem (Buffer, AuthSessionIn->nonce.buffer, AuthSessionIn->nonce.size);
+    Buffer += AuthSessionIn->nonce.size;
+
+    // sessionAttributes
+    *(UINT8 *)Buffer = *(UINT8 *)&AuthSessionIn->sessionAttributes;
+    Buffer++;
+
+    // hmac
+    WriteUnaligned16 ((UINT16 *)Buffer, SwapBytes16 (AuthSessionIn->hmac.size));
+    Buffer += sizeof (UINT16);
+
+    CopyMem (Buffer, AuthSessionIn->hmac.buffer, AuthSessionIn->hmac.size);
+    Buffer += AuthSessionIn->hmac.size;
+  } else {
+    //  sessionHandle
+    WriteUnaligned32 ((UINT32 *)Buffer, SwapBytes32 (TPM_RS_PW));
+    Buffer += sizeof (UINT32);
+
+    // nonce = nullNonce
+    WriteUnaligned16 ((UINT16 *)Buffer, SwapBytes16 (0));
+    Buffer += sizeof (UINT16);
+
+    // sessionAttributes = 0
+    *(UINT8 *)Buffer = 0x00;
+    Buffer++;
+
+    // hmac = nullAuth
+    WriteUnaligned16 ((UINT16 *)Buffer, SwapBytes16 (0));
+    Buffer += sizeof (UINT16);
+  }
+
+  return (UINT32)((UINTN)Buffer - (UINTN)AuthSessionOut);
+}
+
+/**
+  Copy AuthSessionIn from TPM2 response buffer.
+
+  @param [in]  AuthSessionIn   Input AuthSession data in TPM2 response buffer
+  @param [out] AuthSessionOut  Output AuthSession data
+
+  @return 0    copy failed
+          else AuthSession size
+**/
+UINT32
+EFIAPI
+Tpm2CopyAuthSessionResponse (
+  IN      UINT8               *AuthSessionIn,
+  OUT     TPMS_AUTH_RESPONSE  *AuthSessionOut OPTIONAL
+  )
+{
+  UINT8               *Buffer;
+  TPMS_AUTH_RESPONSE  LocalAuthSessionOut;
+
+  if (AuthSessionOut == NULL) {
+    AuthSessionOut = &LocalAuthSessionOut;
+  }
+
+  if (AuthSessionIn == NULL) {
+    return 0;
+  }
+
+  Buffer = (UINT8 *)AuthSessionIn;
+
+  // nonce
+  AuthSessionOut->nonce.size = SwapBytes16 (ReadUnaligned16 ((UINT16 *)Buffer));
+  Buffer                    += sizeof (UINT16);
+  if (AuthSessionOut->nonce.size > sizeof (TPMU_HA)) {
+    DEBUG ((DEBUG_ERROR, "Tpm2CopyAuthSessionResponse - nonce.size error %x\n", AuthSessionOut->nonce.size));
+    return 0;
+  }
+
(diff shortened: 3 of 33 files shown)


kraxel:
What is the use case for the new library?

Raymond-MS:
> What is the use case for the new library?

There are instances where use of these helper functions wouldn't require the inclusion of Tpm2CommandLib which is dependent on Tpm2DeviceLib. Both of which are mainly for communicating with a TPM. Generating TPM logs as HOBs would be an example.

Raymond-MS:
> > What is the use case for the new library?
> 
> There are instances where use of these helper functions wouldn't require the inclusion of Tpm2CommandLib which is dependent on Tpm2DeviceLib. Both of which are mainly for communicating with a TPM. Generating TPM logs as HOBs would be an example.

I see multiple instances of helper functions re-written in various packages: GetHashMaskFromAlgo is an example (See OvmfPkg: TdTcg2Dxe.c). With this new helper lib, we can consolidate helper functions into one location and include that library rather than have duplicate instances.

kraxel (approved):

ardbiesheuvel:
> ## Integration Instructions
> Include Tpm2HelpLib in place of Tpm2CommandLib if necessary, otherwise, just add Tpm2HelpLib to the .inf and include the header in the .c

This will break every platform .DSC that incorporates such INFs but does not specify a resolution for `Tpm2HelpLib`. Please prepare a sibling PR for `edk2-platforms` that fixes up such occurrences.

Raymond-MS:
@jyao1 When you can, please take a look at this PR. Thanks!

os-d:
@jyao1 can you please help review this? It has been open for 2 months without SecurityPkg review. Thanks!

SecurityPkg/Library/Tpm2HelpLib/Tpm2Help.c
```diff
@@ -27,6 +27,35 @@ STATIC INTERNAL_HASH_INFO  mHashInfo[] = {
...
+**/
+STATIC
+BOOLEAN
+CheckDigestListForHashAlg (
+  IN TPML_DIGEST_VALUES  *DigestList,
+  IN TPM_ALG_ID          HashAlg
+  )
+{
+  UINT32  Index;
+
```
AshrafAliS, line 49: Do we need to validate DigestList before we access it, either in this function or the caller

SecurityPkg/SecurityPkg.dec
```diff
@@ -46,6 +46,10 @@
   #
   Tpm2CommandLib|Include/Library/Tpm2CommandLib.h
 
+  ##  @libraryclass  Provides interfaces for other modules to send TPM 2.0 command.
+  #
+  Tpm2HelpLib|Include/Library/Tpm2HelpLib.h
```
spbrogan, line 51: This looks like copy and paste from above.  It is not accurate.  Especially for something named "Help" you should be really clear what the library class provides and its intended purpose.

SecurityPkg/Library/Tpm2HelpLib/Tpm2HelpLib.inf
```diff
@@ -0,0 +1,29 @@
+## @file
+#  Provides helper functions for TPM 2.0 interactions
```
spbrogan: This needs better and more complete information.

SecurityPkg/Library/Tpm2HelpLib/Tpm2Help.c
```diff
@@ -291,31 +320,93 @@ CopyDigestListToBuffer (
...
+        );
+      return EFI_SUCCESS;
+    }
+  }
+
+  return EFI_NOT_FOUND;
+}
+
+/**
+  Copy a buffer into  TPML_DIGEST_VALUES structure.
```
spbrogan: Now that you are making this a standalone libraryclass linkable by multiple modules (and even implemented by different instances) you need to harden the code and define more.  Buffer format should be define and validated.  More parameter checking is also needed.

SecurityPkg/Include/Library/Tpm2HelpLib.h
```diff
@@ -132,4 +132,55 @@ GetDigestFromDigestList (
...
+  Check if all hash algorithms supported in HashAlgorithmMask are
+  present in the DigestList.
+
+  @param DigestList         Digest list
+  @param HashAlgorithmMask  Bitfield of allowed hash algorithms.
+
+  @retval TRUE  All hash algorithms present.
+  @retval FALSE Some hash algorithms not present.
+**/
+BOOLEAN
```
spbrogan, line 181: this needs EFIAPI

SecurityPkg/Library/Tpm2HelpLib/Tpm2Help.c
Raymond-MS: I added the check into the caller of this function. As this is static to this file, we should make sure all callers to this function are providing a valid DigestList. Good callout, thanks!

Raymond-MS:
> @jyao1 When you can, please take a look at this PR. Thanks!

@jyao1 I could still use your review on this PR please! Thanks!

Raymond-MS:
> > @jyao1 When you can, please take a look at this PR. Thanks!
> 
> @jyao1 I could still use your review on this PR please! Thanks!

@jyao1 or @liyi77 This review has been up for quite some time now. I could really use a review, please! Thank you! Let me know if there is anything I can do to help move this along.

Raymond-MS:
> > > @jyao1 When you can, please take a look at this PR. Thanks!
> > 
> > 
> > @jyao1 I could still use your review on this PR please! Thanks!
> 
> @jyao1 or @liyi77 This review has been up for quite some time now. I could really use a review, please! Thank you! Let me know if there is anything I can do to help move this along.

I was hoping to get some help with reviewing this PR. It has been open since October of last year without a SecurityPkg review. Please let me know if there is anything you need from me. Thanks!

Tianocore Stewards
----------------------
F: *
M: Andrew Fish <afish@appl.com> @ajfish 
M: Ard Biesheuvel <ardb+tianocore@kernel.org> @ardbiesheuvel 
M: Leif Lindholm <leif.lindholm@oss.qualcomm.com> @leiflindholm 
M: Michael D Kinney <michael.d.kinney@intel.com> @mdkinney 
M: Michael Kubacki <mikubacki@linux.microsoft.com> @makubacki

liyi77:
> > > @jyao1 When you can, please take a look at this PR. Thanks!
> > 
> > 
> > @jyao1 I could still use your review on this PR please! Thanks!
> 
> @jyao1 or @liyi77 This review has been up for quite some time now. I could really use a review, please! Thank you! Let me know if there is anything I can do to help move this along.

I am not the reviewer of SecurityPkg, maybe @arunsbaskaran can also help on it.

Raymond-MS:
> > > > @jyao1 When you can, please take a look at this PR. Thanks!
> > > 
> > > 
> > > @jyao1 I could still use your review on this PR please! Thanks!
> > 
> > 
> > @jyao1 or @liyi77 This review has been up for quite some time now. I could really use a review, please! Thank you! Let me know if there is anything I can do to help move this along.
> 
> I was hoping to get some help with reviewing this PR. It has been open since October of last year without a SecurityPkg review. Please let me know if there is anything you need from me. Thanks!
> 
> ## Tianocore Stewards
> F: * M: Andrew Fish [afish@appl.com](mailto:afish@appl.com) @ajfish M: Ard Biesheuvel [ardb+tianocore@kernel.org](mailto:ardb+tianocore@kernel.org) @ardbiesheuvel M: Leif Lindholm [leif.lindholm@oss.qualcomm.com](mailto:leif.lindholm@oss.qualcomm.com) @leiflindholm M: Michael D Kinney [michael.d.kinney@intel.com](mailto:michael.d.kinney@intel.com) @mdkinney M: Michael Kubacki [mikubacki@linux.microsoft.com](mailto:mikubacki@linux.microsoft.com) @makubacki

Pinging again to get some support on this PR, please. Thanks!

@ajfish @ardbiesheuvel @leiflindholm @mdkinney @makubacki

leiflindholm:
> Pinging again to get some support on this PR, please. Thanks!
> 
> @ajfish @ardbiesheuvel @leiflindholm @mdkinney @makubacki

This really needs review by SecurityPkg maintainers, of which we appear to be having a shortage. I think that needs to be a topic for the next stewards meeting.

Some high-level comments:
- This was a problem with the code already, but should definitely be addressed if the intent is now to be used outside of SecurityPkg: namespacing. Function names should have some prefix, for example Tpm2.
- Why are the declarations moved around relatively to each other as part of the breakout? If you feel they need reorganising, please do that separately in a preceding commit.
- The above makes it a lot harder to spot that CopyBufferToDigestList and GetDigestListSizeFromHashAlgorithmMask disappear as part of the "move". And then for some reason reintroduced in a separate, subsequent commit. I'm all for breaking large moves up in steps to help with review, but this does the opposite as well as introduces a potential bisect breakage.

jyao1:
My comment is same as before.

1) My biggest concern is still compatibility issue, because once this is checked in, every platform need to be updated to add the new library.

2) If the problem statement is "This allows for use of the helper functions without the need to include Tpm2CommandLib and Tpm2DeviceLib which is usually tied to the TPM and TPM communication.", can we just add a Tpm2HelpLib, without changing others?

As such, there will be no compatibility problem in this case, and you have freedom to change / remove / add helper function if needed.

Also, please help me understand what is the use case that a module only need link Tpm2HelpLib? I feel the help lib API design should be based on the use case, instead of current implementation.

3) Yes, I agree the feedback "This was a problem with the code already, but should definitely be addressed if the intent is now to be used outside of SecurityPkg: namespacing. Function names should have some prefix, for example Tpm2."

It is a good idea to add `Tpm2` prefix, if we introduce a new lib.

Raymond-MS:
> My comment is same as before.
> 
> 1. My biggest concern is still compatibility issue, because once this is checked in, every platform need to be updated to add the new library.
> 2. If the problem statement is "This allows for use of the helper functions without the need to include Tpm2CommandLib and Tpm2DeviceLib which is usually tied to the TPM and TPM communication.", can we just add a Tpm2HelpLib, without changing others?
> 
> As such, there will be no compatibility problem in this case, and you have freedom to change / remove / add helper function if needed.
> 
> Also, please help me understand what is the use case that a module only need link Tpm2HelpLib? I feel the help lib API design should be based on the use case, instead of current implementation.
> 
> 3. Yes, I agree the feedback "This was a problem with the code already, but should definitely be addressed if the intent is now to be used outside of SecurityPkg: namespacing. Function names should have some prefix, for example Tpm2."
> 
> It is a good idea to add `Tpm2` prefix, if we introduce a new lib.

@jyao1 There are cases where you may need to generate a HOB for pre-DXE measurements. These cases utilize these helper functions but do not require Tpm2CommandLib or Tpm2DeviceLib as there is no communication to the TPM needed. An example is there is no Tpm2DeviceLib for SEC. I specifically ran into this problem which is where this work spawned from. I don't see the reason to have these helper functions coupled to Tpm2CommandLib when it isn't explicitly necessary. Also, there are multiple instances of these helper functions strewn throughout the repo. Having a single place where helper functions can reside reduces this duplication and reduces the chances of bugs and maintenance. Please refer to my response to @kraxel.

Raymond-MS:
> > Pinging again to get some support on this PR, please. Thanks!
> > @ajfish @ardbiesheuvel @leiflindholm @mdkinney @makubacki
> 
> This really needs review by SecurityPkg maintainers, of which we appear to be having a shortage. I think that needs to be a topic for the next stewards meeting.
> 
> Some high-level comments:
> 
> * This was a problem with the code already, but should definitely be addressed if the intent is now to be used outside of SecurityPkg: namespacing. Function names should have some prefix, for example Tpm2.
> * Why are the declarations moved around relatively to each other as part of the breakout? If you feel they need reorganising, please do that separately in a preceding commit.
> * The above makes it a lot harder to spot that CopyBufferToDigestList and GetDigestListSizeFromHashAlgorithmMask disappear as part of the "move". And then for some reason reintroduced in a separate, subsequent commit. I'm all for breaking large moves up in steps to help with review, but this does the opposite as well as introduces a potential bisect breakage.

I agree adding the prefix is a good idea. The reason for the declarations moving is due to those functions being added in a separate commit. I do not know why they were placed where they were when they were added but it made sense to add them to the end of the file as they are new functions being added to the library. My method was to take this in stages as if it were a new library, not as if I were just moving the file. The first set of functions to move were the original helpers with the additional helpers being added later.

jyao1:
@Raymond-MS , understood your use case now. Thanks to explain that to me.

I agree with you that it is a good idea to have Tpm2HelpLib, according to your problem statement "This allows for use of the helper functions without the need to include Tpm2CommandLib and Tpm2DeviceLib which is usually tied to the TPM and TPM communication."

However, I am still not convinced to change the original code which impacts all rest platform.
To resolve your problem statement above, just adding a new Tpm2HelpLib is good enough to start with. It is compatible, and has no impact to others. (Cleaning up the exiting code is definitely another **HARDER** task.)

os-d:
Edit: My apologies, I made my original comment without re-reading the PR, which I should have done. I forgot the exact semantics of it since it has been up so long.

@jyao1 adding a new library class is trivial for platform developers to do. I think that your suggested approach could work because the new lib functions already exist in Tpm2Help.c and are not spread across other files. I don't think that is a pattern we want to create though, especially if the only concern is adding a single line in the DSC.

If there is concern about how the logic is moved into the new library class, that is totally fair and we should discuss what the right method is, but if the only concern is that this makes platform developers add a single line to the DSC, then I think we should consider that acceptable, as the rest of edk2 does, for platforms to do when consuming new stable tags.

Do you have concerns about how the logic is moved? If so, let’s address those. Happy to get feedback from others on appropriateness of requiring platform devs to add a single line to their DSC.

os-d:
I’ll also note Ard already asked Raymond to do the extra step of fixing up the platforms in edk2-platforms, which he has a PR for here: https://github.com/tianocore/edk2-platforms/pull/896.

jyao1:
@os-d, I do not have concern on the code logic.
I understand your use case now, and I agree that adding Tpm2HelpLib is good solution (with Tpm2 prefix). I can approve that part.


> where helper functions can reside reduces this duplication and reduces the chances of bugs and maintenance.

I do think there is a way to avoid duplication and reduce the changes of bugs and maintenance, without removing the code from original library.

For example, we can put Tpm2Help.inf under [Tpm2Command](https://github.com/tianocore/edk2/tree/master/SecurityPkg/Library/Tpm2CommandLib).
We will have
```
https://github.com/tianocore/edk2/tree/master/SecurityPkg/Library/Tpm2CommandLib/Tpm2CommandLib.inf
https://github.com/tianocore/edk2/tree/master/SecurityPkg/Library/Tpm2CommandLib/Tpm2HelpLib.inf
https://github.com/tianocore/edk2/tree/master/SecurityPkg/Library/Tpm2CommandLib/Tpm2Help.c
```
As such, no file needs to be moved. You naturally get a new library with current function, with just adding one new file. This is a lightweight change.

My question would be: **Why we have to choose a solution which breaks the compatibility?**

What is the problem statement? We just need to resolve the problem.

Again, my big concern is about platform DSC compatibility (if it is moved or deleted). And it is subjective to say "it is a small issue" or "it is a big issue".
To address this concern, I suggest to get **Steward team's aligned response** on the compatibility impact.

leiflindholm:
> My method was to take this in stages as if it were a new library, not as if I were just moving the file. 

And my view is the method taken here makes review complicated, which could be resolved by _not_ using this method. This is Security code. Reviewability (and hence traceability) is a high priority for all PRs, but _especially_ here.

So, please rework using the approach of first reorganising functions and prototypes as they will appear in the new library, and then moving them. Or skip reorganising them.

And please move _all_ of the functions in one go, the current implementation breaks bisect by leaving a commit where CopyBufferToDigestList and GetDigestListSizeFromHashAlgorithmMask don't exist.

IsDigestListInSyncWithHashAlgorithmMask appears to be new, and hence belongs in a subsequent commit, but it is the only exported function not mentioned in that commit message.

makubacki:
I agree with Leif's points about reorganizing the commits. Bisectability should obviously not be broken, and code movement should be logical within commits.

I am not a fan of this approach:

```
https://github.com/tianocore/edk2/tree/master/SecurityPkg/Library/Tpm2CommandLib/Tpm2CommandLib.inf
https://github.com/tianocore/edk2/tree/master/SecurityPkg/Library/Tpm2CommandLib/Tpm2HelpLib.inf
https://github.com/tianocore/edk2/tree/master/SecurityPkg/Library/Tpm2CommandLib/Tpm2Help.c
```

This would be encouraging a pattern that is overall confusing and favors implementation convenience over interface scope and clarity. While more than one INF is placed in library directories in various places for various reasons, there is usually a higher degree of cohesion than here (like INFs for different phases otherwise operating on the same underlying implementation for a single module/interface). This is presented as a generic "helper library" that within its scope of responsibility may be used outside of the Tpm2CommandLib.

If the idea is that those using Tpm2CommandLib don't need to add Tpm2HelpLib to their DSC (e.g. avoiding a breaking change) because it won't depend on the library class, I think that is counter to library class design. The expectation would be that if a platform wants to substitute the functionality in Tpm2HelpLib, for whatever reason, such as making a bug fix, it should be possible to apply that change by specifying the desired library instance for the class in the platform DSC file. I don't think its implementation should be leaked outside that contract.

Raymond-MS:
> > My method was to take this in stages as if it were a new library, not as if I were just moving the file.
> 
> And my view is the method taken here makes review complicated, which could be resolved by _not_ using this method. This is Security code. Reviewability (and hence traceability) is a high priority for all PRs, but _especially_ here.
> 
> So, please rework using the approach of first reorganising functions and prototypes as they will appear in the new library, and then moving them. Or skip reorganising them.
> 
> And please move _all_ of the functions in one go, the current implementation breaks bisect by leaving a commit where CopyBufferToDigestList and GetDigestListSizeFromHashAlgorithmMask don't exist.
> 
> IsDigestListInSyncWithHashAlgorithmMask appears to be new, and hence belongs in a subsequent commit, but it is the only exported function not mentioned in that commit message.

Thanks @leiflindholm, you are correct. For some reason we had that change as a separate commit in our history which is why I added those functions as a separate commit. It looks like they were already upstreamed which is where the confusion was coming from. Apologies on my end. I should have resolved that change and now only the one function is being added.

Raymond-MS:
@leiflindholm & @jyao1 Would you be able to take another look at this PR? It hasn't received any traction since the last commit I pushed up. I should have addressed most if not all of the feedback outside of keeping the library to Tpm2CommandLib. Thanks!

jyao1:
I hold my previous opinion. Let me copy/paste below:

> My question would be: Why we have to choose a solution which breaks the compatibility?

> What is the problem statement? We just need to resolve the problem.

> Again, my big concern is about platform DSC compatibility (if it is moved or deleted). And it is subjective to say "it is a small issue" or "it is a big issue".
> To address this concern, I suggest to get Steward team's aligned response on the compatibility impact.

SecurityPkg/Library/PeilessSecMeasureLib/PeilessSecMeasureLib.c
makubacki, line 1: I don't see where you've added `Tpm2HelpLib` to PeilessSecMeasureLib.inf.

SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
makubacki, line 1: I don't see where you've added Tpm2HelpLib to Tcg2Pei.inf.

makubacki:
For the DSC compatibility decision, it ultimately comes down to whether a platform that depends on these INFs:

- ‎SecurityPkg/Library/HashLibBaseCryptoRouter/HashLibBaseCryptoRouterDxe.inf‎
- SecurityPkg/Library/PeilessSecMeasureLib/PeilessSecMeasureLib.inf
- SecurityPkg/Library/Tpm2CommandLib/Tpm2CommandLib.inf
- SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.inf
- SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.inf

Should have to add a `Tpm2HelpLib` line to their DSC file. Unfortunately, you have and are going to get differing opinions to the impact. For that particular problem, other packages have used a `.dsc.inc` file (e.g. [MdeLib.dsc.inc](https://github.com/tianocore/edk2/blob/master/MdePkg/MdeLibs.dsc.inc)).

Perhaps SecurityPkg could adopt something similar to ease DSC integration.

---

I think the more impactful, though less likely, work on a platform developer is (1) the functions being removed from the public Tpm2CommandLib API and (2) those functions being renamed in Tpm2HelpLib. Obviously, code in SecurityPkg will be updated, but any platform code using those functions will need source changes.

---

Now, we have two pieces:

1. Add Tpm2HelpLib to platform DSC (for platforms that use TPM code listed above)
2. Potential source code impact for platform users of modified Tpm2CommandLib APIs

---

If we agree the split to Tpm2HelpLib makes sense long term and the cleanup is the concern. Perhaps we could merge this PR with compatibility to existing code with a clear warning that the cleanup will occur after the 202605 stable tag and then make those changes then. Platforms would then have time to prepare and no compatibility impact to the next stable tag (202605) which will include both changes.

SecurityPkg/Library/PeilessSecMeasureLib/PeilessSecMeasureLib.c
Raymond-MS: Good catch! Not sure how this got dropped. Fixed!

SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
Raymond-MS: Good catch! Not sure how this got dropped. Fixed!

Raymond-MS:
I just wanted to leave a comment explaining my plans going forward to implement the feedback I have received. Currently the plan is to revert the change where Tpm2Help.c is removed from Tpm2CommandLib and instead just introduce the new Tpm2HelpLib with the same functions but with the prefix instead (i.e. squashing the prefix commit into the introduction commit). Afterwards, I am going to add a new commit that makes it so that the functions in Tpm2CommandLib's Tpm2Help.c are basically wrappers which call into the new Tpm2HelpLib. This way anyone who is calling the old functions are not broken and the only change they need to make is adding the Tpm2HelpLib.inf to their platform .dsc as Tpm2CommandLib will now depend on Tpm2HelpLib. At some point, we should be able to drop Tpm2Help.c from Tpm2CommandLib once most platforms have updated and moved to using Tpm2HelpLib.

If there are any questions or concerns, please feel free to let me know. 

@jyao1 I have answers to your questions below:

"My question would be: Why we have to choose a solution which breaks the compatibility?"
I don't have a great response for this other than there is a major downside to having duplicate code between Tpm2CommandLib's Tpm2Help.c and Tpm2HelpLib's Tpm2Help.c. We do not want to have to maintain both copies and if one changes the other will also need to change. Having a single place where these helper functions live is going to be much better for the overall health of the codebase and reduce the maintenance burden.

"What is the problem statement? We just need to resolve the problem."
The problem is that there are instances where you do not need/want to include Tpm2CommandLib and thus Tpm2DeviceLib when you only need access to the Tpm2Help.c functions. An example is when you need to generate pre-DXE logging HOBs which utilize these helper functions to populate digests/etc. but do not need/require any TPM communication which comes with Tpm2CommandLib and Tpm2DeviceLib. There are also instances where multiple of these helper functions are strewn throughout the codebase, I assume, for this very reason. Having a single library to hold helper functions that can be used as needed is a good approach to reduce code duplication and reduce maintenance.

Raymond-MS:
@jyao1 Please review the changes, I should have addressed your concerns in the latest commit I pushed up. If possible, I would like to get this merged. Thank you!

makubacki:
@jyao1, @Raymond-MS has put a lot of effort into accommodating feedback on this PR. If you're busy, can you please let him know approximately when you'll be able to review again?

@arunsbaskaran, since you're a SecurityPkg TCG reviewer, your feedback is appreciated as well.

Raymond-MS:
@jyao1 Pinging again in hopes that you will get a chance to review. It has now been 3 weeks since my last ping and 2 weeks since @makubacki pinged asking for reviews. Please review when you get a chance. This PR has been up for 7+ months, I would really like to get this merged and off my plate. Thank you!

makubacki:
@jyao1, can you please reply if you are going to review this again? If there is not a reply by tomorrow, I will re-review the latest and we'll move on with that and other reviews.

SecurityPkg/Library/Tpm2HelpLib/Tpm2HelpLib.inf
makubacki, line 1: I think that the first commit "SecurityPkg: Introduce Tpm2HelpLib" (https://github.com/tianocore/edk2/pull/11634/changes/cf3333c46511f93b8666aebd4069c8190e26b6ff), should just add the full implementation for `Tpm2HelpLib`.

In any case, it needs to add `‎SecurityPkg/Library/Tpm2HelpLib/Tpm2HelpLib.inf` to SecurityPkg.dsc (so its build is verified) and add the library class being introduced (`Tpm2HelpLib`) to SecurityPkg.dec.

Logically, it makes sense to proceed to package integration (the following commits as they are) of that library class in anticipation for the breaking change of adding that class to libraries used elsewhere.

Then, what is currently commit https://github.com/tianocore/edk2/pull/11634/changes/c12d4037e94c3c90e4f291595cace6e63ffce1ad could retain the changes to SecurityPkg code to use `Tpm2HelpLib`.

Other packages that might now be indirectly dependent on `Tpm2HelpLib` should have been updated to specify the `Tpm2HelpLib` library class in their DSC files at this point.

SecurityPkg/Library/Tpm2HelpLib/Tpm2Help.c
makubacki, line 1: I think you should merge https://github.com/tianocore/edk2/pull/11634/changes/f5813fed774c28f85c68b7905b24110ab862d102 into the commit that adds the overall `Tpm2HelpLib` implementation.

If it stays separate, the commit message should explain why it is separate for clarity. If it was just for review clarity based on the origin of the function, I still think it is fine to be a part of the commit with the rest of the library instance implementation.

SecurityPkg/Library/Tpm2HelpLib/Tpm2Help.c
```diff
@@ -0,0 +1,544 @@
...
+  CONST UINT8  *BufferPtr;
+
+  if ((Buffer == NULL) || (DigestList == NULL) || (BufferSize > sizeof (TPML_DIGEST_VALUES))) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  DigestList->count = SwapBytes32 (ReadUnaligned32 ((CONST UINT32 *)Buffer));
+  if (DigestList->count > HASH_COUNT) {
+    return EFI_INVALID_PARAMETER;
+  }
```
makubacki, line 377: A few functions (like this one, `Tpm2GetDigestListSize()`, and `Tpm2CopyDigestListToBuffer()`) have a `TPML_DIGEST_VALUES` parameter but out of those, this is the only one that validates `DigestList->count <= HASH_COUNT`.

You could consider moving that check to a small helper function and then consistently applying it across all external facing APIs that accept a `TPML_DIGEST_VALUES` argument.

SecurityPkg/Library/Tpm2HelpLib/Tpm2Help.c
```diff
@@ -0,0 +1,544 @@
...
+  IN UINT32              HashAlgorithmMask
+  )
+{
+  UINTN   Index;
+  UINT16  DigestSize;
+  UINT32  DigestListCount;
+  UINT32  *DigestListCountPtr;
+
+  if ((Buffer == NULL) || (DigestList == NULL)) {
+    return NULL;
```
makubacki, line 315: Can you update the function description to mention that this might return `NULL` and check callers are handling it properly?

SecurityPkg/Library/Tpm2HelpLib/Tpm2HelpLib.inf
makubacki, line 1: Files like [OvmfPkg/Tcg/TdTcg2Dxe/TdTcg2Dxe.c](https://github.com/tianocore/edk2/blob/master/OvmfPkg/Tcg/TdTcg2Dxe/TdTcg2Dxe.c) still define functions now available in `Tpm2HelpLib` like [`GetHasSizeFromAlgo()`](https://github.com/tianocore/edk2/blob/4bfc280b57971743591f923516832a3fa84f5dde/OvmfPkg/Tcg/TdTcg2Dxe/TdTcg2Dxe.c#L148).

Did you consider moving those to use `Tpm2HelpLib`?

SecurityPkg/Include/Library/Tpm2HelpLib.h
```diff
@@ -0,0 +1,187 @@
...
+UINT32
+EFIAPI
+Tpm2GetDigestListSize (
+  IN TPML_DIGEST_VALUES  *DigestList
+  );
+
+/**
+  Get TPML_DIGEST_VALUES data size from HashAlgorithmMask
+
+  @param[in]     DigestList    TPML_DIGEST_VALUES data.
```
makubacki:
```suggestion
  @param[in]     HashAlgorithmMask    Bitfield of allowed hash algorithms.
```

SecurityPkg/Library/Tpm2HelpLib/Tpm2Help.c
```diff
@@ -0,0 +1,544 @@
+/** @file
+  Implement TPM2 help.
+
+Copyright (c), Microsoft Corporation.
```
makubacki:
```suggestion
Copyright (c) Microsoft Corporation.
```

SecurityPkg/Include/Library/Tpm2HelpLib.h
```diff
@@ -0,0 +1,187 @@
+/** @file
+  Definitions for TPM2 helper functions
+
+Copyright (c), Microsoft Corporation.
```
makubacki:
```suggestion
Copyright (c) Microsoft Corporation.
```

SecurityPkg/Library/Tpm2HelpLib/Tpm2HelpLib.inf
```diff
@@ -0,0 +1,29 @@
+## @file
+#  Provides helper functions for interfacing with TPM 2.0 data such
+#  as the Digest and AuthSession. Provides a place for common TPM 2.0
+#  functionality.
+#
+# Copyright (c) 2013 - 2018, Intel Corporation. All rights reserved.<BR>
+# SPDX-License-Identifier: BSD-2-Clause-Patent
```
makubacki, line 8: Since this is a new file and to match Tpm2HelpLib.c and Tpm2HelpLib.h, I think a copyright is appropriate here.


```suggestion
# Copyright (c) 2013 - 2018, Intel Corporation. All rights reserved.<BR>
# Copyright (c) Microsoft Corporation.
# SPDX-License-Identifier: BSD-2-Clause-Patent
```

makubacki:
@Raymond-MS, please update the "Integration Instructions" in the PR description to include the deprecation steps for the functions left in `Tpm2CommandLib`.

makubacki:
Based on prior notices about the length of time this PR has been open, once the current feedback is confirmed to be addressed, I will merge the PR if there is no other outstanding feedback.

SecurityPkg/Library/Tpm2HelpLib/Tpm2Help.c
Raymond-MS: Fixed.

SecurityPkg/Include/Library/Tpm2HelpLib.h
Raymond-MS: Fixed.

SecurityPkg/Library/Tpm2HelpLib/Tpm2Help.c
Raymond-MS: Sounds good. This should be fine with the current change since we are no longer directly removing the functions from Tpm2CommandLib.

SecurityPkg/Library/Tpm2HelpLib/Tpm2HelpLib.inf
Raymond-MS: Sounds good. Will do.

Raymond-MS: At first glance, I had left these alone because they used a different structure. However, after a deeper look they are basically identical. There should be no reason why I can't move those functions to use Tpm2HelpLib. I will update. Thanks!

SecurityPkg/Library/Tpm2HelpLib/Tpm2HelpLib.inf
Raymond-MS: Fixed.

SecurityPkg/Library/Tpm2HelpLib/Tpm2Help.c
Raymond-MS: Good catch, I believe this was only done because we were copying from a buffer which we want to validate the count is correct but there is no reason we can't validate the count before using it in other functions. Added.

SecurityPkg/Include/Library/Tpm2HelpLib.h
Raymond-MS: Nice catch! Fixed.

SecurityPkg/Library/Tpm2HelpLib/Tpm2Help.c
Raymond-MS: Of course, updated.

cfernald (approved):
I've reviewed these changes in MU. I'm not seeing any delta that is worth rehashing the changes over and seems all the updates for the commit and library structure are goodness.

SecurityPkg/Include/Library/Tpm2HelpLib.h
```diff
@@ -0,0 +1,170 @@
...
+  Definitions for TPM2 helper functions
+
+Copyright (c) Microsoft Corporation.
+Copyright (c) 2013 - 2024, Intel Corporation. All rights reserved. <BR>
+SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef TPM2_HELP_LIB_H_
+#define TPM2_HELP_LIB_H_
```
makubacki: Please use `#pragma once` to match current convention.

SecurityPkg/Library/Tpm2HelpLib/Tpm2Help.c
```diff
@@ -0,0 +1,473 @@
+/** @file
+  Implement TPM2 help.
```
makubacki: nit: Since you're going to push another update, can you please replace this with a more relevant description?

SecurityPkg/Include/Library/Tpm2HelpLib.h
makubacki, line 1: Why is this commit broken out from the commit that introduces the `TpmHelp` implementation? [SecurityPkg: Additional helper function(s)](https://github.com/tianocore/edk2/commit/0425f8d72b66c389a3da94a51e2b4e720930b713)

SecurityPkg/Include/Library/Tpm2HelpLib.h
Raymond-MS: Updated.

SecurityPkg/Library/Tpm2HelpLib/Tpm2Help.c
Raymond-MS: Will do!

SecurityPkg/Include/Library/Tpm2HelpLib.h
Raymond-MS: You are right. This should be fine to squash into the first commit. Will do. Thanks!

makubacki (approved):

makubacki:
@mergifyio queue

PR #11640, opened 2025-10-21
 BaseTools/Conf: Remove OBJCOPY and SYMRENAME operations

mdkinney:
# Description

Remove unused `SYMRENAME_PATH` and `SYMRENAME_FLAGS` from `tools_def.template` that are not used by any build rules in `build_rule.template`.

Remove `OBJCOPY_PATH`, `OBJCOPY_FLAGS`, `OBJCOPY_ADDDEBUGFLAG`, and `OBJCOPY_STRIPFLAG` from `tools_def.template` and remove all `OBJCOPY` steps in `build_rule.template`.
* `OBJCOPY_FLAGS` was either set to nothing or set to string to force a build error. This implies that `OBJCOPY` steps with `OBJCOPY_FLAGS` do not generate any new files. These steps can be safely removed with no side effects
* `OBJCOPY_ADDDEBUGFLAG` is only used in `GCC` builds to generate a section in an ELF image with a link to the `.debug` DWARF symbol file for the `UNIXGCC` tool chain that directly generates PE/COFF images. The `UNIXGCC` tool chain is no longer supported, and the PE/COFF images generated with currently supported tool chains contain a debug directory entry with a link to the DWARF symbol file. The steps using `OBJCOPY_ADDDEBUGFLAG` can be safely removed with no side effects.
* `OBJCOPY_STRIPFLAG` is only used in GCC builds to strip symbols before the `GENFW` ELF->PE/COFF conversion. No other tool chain configurations perform a symbol strip operation. Stripping symbols from images induces failures in some of the `GENFW` operations to parse symbols (such as PRM use cases). Removing the `OBJCOPY_STRIPFLAG` operations from `GCC` build rules aligns `GCC` with all other tool chains on treatment of symbols.
* Add `-Wl,--no-relax` to `tools_def.template` for `GCC` tool chains for `X64`. This aligns with other CPU architecture `GCC` linker settings. This is required for `X64` because the removal of the `OBJCOPY` operations to not strip symbols introduces the use of ELF relocation types not supported by `GENFW` ELF->PE/COFF conversions. Adding `--no-relax` prevents the use of the unsupported ELF relocation types such as `R_X86_64_REX_GOTPCRELX`.

Remove `OBJCOPY_STRIPFLAG` override in `PrmPkg`. The use of `OBJCOPY_STRIPFLAG` has been removed from `build_rule.template`. The override of `OBJCOPY_STRIPFLAG` to nothing to prevent symbols required to generate DLL export section in PRM images is no longer required.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Build with `GCC` and `CLANGDWARF` before and after this change with build issues or regressions.

## Integration Instructions

N/A

### DIFF
--- a/PrmPkg/Samples/PrmSampleAcpiParameterBufferModule/PrmSampleAcpiParameterBufferModule.inf
+++ b/PrmPkg/Samples/PrmSampleAcpiParameterBufferModule/PrmSampleAcpiParameterBufferModule.inf
@@ -46,4 +46,3 @@
   CLANGPDB:*_*_*_GENFW_FLAGS = --keepoptionalheader
 
   GCC:*_*_*_GENFW_FLAGS = --prm --image-version $(VERSION_STRING)
-  GCC:*_*_*_OBJCOPY_STRIPFLAG ==
--- a/PrmPkg/Samples/PrmSampleContextBufferModule/PrmSampleContextBufferModule.inf
+++ b/PrmPkg/Samples/PrmSampleContextBufferModule/PrmSampleContextBufferModule.inf
@@ -49,4 +49,3 @@
   CLANGPDB:*_*_*_GENFW_FLAGS = --keepoptionalheader
 
   GCC:*_*_*_GENFW_FLAGS = --prm --image-version $(VERSION_STRING)
-  GCC:*_*_*_OBJCOPY_STRIPFLAG ==
--- a/PrmPkg/Samples/PrmSampleHardwareAccessModule/PrmSampleHardwareAccessModule.inf
+++ b/PrmPkg/Samples/PrmSampleHardwareAccessModule/PrmSampleHardwareAccessModule.inf
@@ -46,4 +46,3 @@
   CLANGPDB:*_*_*_GENFW_FLAGS = --keepoptionalheader --image-version $(VERSION_STRING)
 
   GCC:*_*_*_GENFW_FLAGS = --prm --image-version $(VERSION_STRING)
-  GCC:*_*_*_OBJCOPY_STRIPFLAG ==


ardbiesheuvel:
> * Add `-Wl,--no-relax` to `tools_def.template` for `GCC` and `CLANGDWARF` tool chains for `X64`. This aligns with other CPU architecture linker settings. This is required for `X64` because the removal of the `OBJCOPY` operations to not strip symbols introduces the use of ELF relocation types not supported by `GENFW` ELF->PE/COFF conversions. Adding `--no-relax` prevents the use of the unsupported ELF relocation types such as `R_X86_64_REX_GOTPCRELX`.
> 

Do both GCC and CLANGDWARF require this? The `ld.lld` and `ld.bfd` linkers behave differently when it comes to relaxable relocations: the former leaves the original relocations in the resulting binary, whereas the latter will update them to whatever describes the relaxed situation best.

Given that the codegen is worse without relaxation (i.e., taking the address of a symbol will involve a MOV from the GOT in memory rather than a RIP-relative LEA which is just an arithmetic operation, and so the resulting code is worse in terms of performance, memory footprint and the number of relocation fixups that need to be applied), I'd prefer to only enable this for `ld.lld` if possible.

mdkinney:
> > * Add `-Wl,--no-relax` to `tools_def.template` for `GCC` and `CLANGDWARF` tool chains for `X64`. This aligns with other CPU architecture linker settings. This is required for `X64` because the removal of the `OBJCOPY` operations to not strip symbols introduces the use of ELF relocation types not supported by `GENFW` ELF->PE/COFF conversions. Adding `--no-relax` prevents the use of the unsupported ELF relocation types such as `R_X86_64_REX_GOTPCRELX`.
> 
> Do both GCC and CLANGDWARF require this? The `ld.lld` and `ld.bfd` linkers behave differently when it comes to relaxable relocations: the former leaves the original relocations in the resulting binary, whereas the latter will update them to whatever describes the relaxed situation best.
> 
> Given that the codegen is worse without relaxation (i.e., taking the address of a symbol will involve a MOV from the GOT in memory rather than a RIP-relative LEA which is just an arithmetic operation, and so the resulting code is worse in terms of performance, memory footprint and the number of relocation fixups that need to be applied), I'd prefer to only enable this for `ld.lld` if possible.


@ardbiesheuvel I have not observed this issue with CLANGDWARF, but my testing with that tool chain was incomplete at this time due to other issues. I did observe this with GCC.  I can remove the CLANGDWARF changes in this PR and if issues are observed in the future with CLANGDWARF, we can discuss options.

mdkinney:
@ardbiesheuvel `-Wl,--no-relax` has been removed from from `CLANGDWARF`

lgao4:
After remove STRIPFLAG , have you compare the generated efi image before and after this change?

mdkinney:
> After remove STRIPFLAG , have you compare the generated efi image before and after this change?

This operation only impacts the .DLL file.  The ELF .DLL file is converted to a PE/COFF .EFI file with no symbols in the .EFI file.  SO stripping symbols has no impact to the final EFI file.  It only impacts the ELF/DWARF information and there is no value in stripping symbols.  It actually breaks PRM processing to extract symbols for export table generation.  The VS20xx and CLANG tools chains do not strip symbols.  So this change makes GCC the same as VS20xx and CLANG with respect to symbol stripping.

lgao4:
> > After remove STRIPFLAG , have you compare the generated efi image before and after this change?
> 
> This operation only impacts the .DLL file. The ELF .DLL file is converted to a PE/COFF .EFI file with no symbols in the .EFI file. SO stripping symbols has no impact to the final EFI file. It only impacts the ELF/DWARF information and there is no value in stripping symbols. It actually breaks PRM processing to extract symbols for export table generation. The VS20xx and CLANG tools chains do not strip symbols. So this change makes GCC the same as VS20xx and CLANG with respect to symbol stripping.

If no impact for the generated EFI image, I have no other comments for this change.

ardbiesheuvel:
> > > * Add `-Wl,--no-relax` to `tools_def.template` for `GCC` and `CLANGDWARF` tool chains for `X64`. This aligns with other CPU architecture linker settings. This is required for `X64` because the removal of the `OBJCOPY` operations to not strip symbols introduces the use of ELF relocation types not supported by `GENFW` ELF->PE/COFF conversions. Adding `--no-relax` prevents the use of the unsupported ELF relocation types such as `R_X86_64_REX_GOTPCRELX`.
> > 
> > 
> > Do both GCC and CLANGDWARF require this? The `ld.lld` and `ld.bfd` linkers behave differently when it comes to relaxable relocations: the former leaves the original relocations in the resulting binary, whereas the latter will update them to whatever describes the relaxed situation best.
> > Given that the codegen is worse without relaxation (i.e., taking the address of a symbol will involve a MOV from the GOT in memory rather than a RIP-relative LEA which is just an arithmetic operation, and so the resulting code is worse in terms of performance, memory footprint and the number of relocation fixups that need to be applied), I'd prefer to only enable this for `ld.lld` if possible.
> 
> @ardbiesheuvel I have not observed this issue with CLANGDWARF, but my testing with that tool chain was incomplete at this time due to other issues. I did observe this with GCC. I can remove the CLANGDWARF changes in this PR and if issues are observed in the future with CLANGDWARF, we can discuss options.

Interesting. As far as I am aware, it is LLVM's `lld` that retains the `R_X86_64_REX_GOTPCRELX` relocations after relaxing them, whereas GNU `ld` converts them into different ones that more accurately describe the relaxed version of the instruction.

mdkinney:
> > > > * Add `-Wl,--no-relax` to `tools_def.template` for `GCC` and `CLANGDWARF` tool chains for `X64`. This aligns with other CPU architecture linker settings. This is required for `X64` because the removal of the `OBJCOPY` operations to not strip symbols introduces the use of ELF relocation types not supported by `GENFW` ELF->PE/COFF conversions. Adding `--no-relax` prevents the use of the unsupported ELF relocation types such as `R_X86_64_REX_GOTPCRELX`.
> > > 
> > > 
> > > Do both GCC and CLANGDWARF require this? The `ld.lld` and `ld.bfd` linkers behave differently when it comes to relaxable relocations: the former leaves the original relocations in the resulting binary, whereas the latter will update them to whatever describes the relaxed situation best.
> > > Given that the codegen is worse without relaxation (i.e., taking the address of a symbol will involve a MOV from the GOT in memory rather than a RIP-relative LEA which is just an arithmetic operation, and so the resulting code is worse in terms of performance, memory footprint and the number of relocation fixups that need to be applied), I'd prefer to only enable this for `ld.lld` if possible.
> > 
> > 
> > @ardbiesheuvel I have not observed this issue with CLANGDWARF, but my testing with that tool chain was incomplete at this time due to other issues. I did observe this with GCC. I can remove the CLANGDWARF changes in this PR and if issues are observed in the future with CLANGDWARF, we can discuss options.
> 
> Interesting. As far as I am aware, it is LLVM's `lld` that retains the `R_X86_64_REX_GOTPCRELX` relocations after relaxing them, whereas GNU `ld` converts them into different ones that more accurately describe the relaxed version of the instruction.

I do not see this relocation type generated in newer versions of GCC.  I only see this relocation type in the final ELF images in older versions of GCC.  I believe it was gcc-7.

ardbiesheuvel:
> > > > > * Add `-Wl,--no-relax` to `tools_def.template` for `GCC` and `CLANGDWARF` tool chains for `X64`. This aligns with other CPU architecture linker settings. This is required for `X64` because the removal of the `OBJCOPY` operations to not strip symbols introduces the use of ELF relocation types not supported by `GENFW` ELF->PE/COFF conversions. Adding `--no-relax` prevents the use of the unsupported ELF relocation types such as `R_X86_64_REX_GOTPCRELX`.
> > > > 
> > > > 
> > > > Do both GCC and CLANGDWARF require this? The `ld.lld` and `ld.bfd` linkers behave differently when it comes to relaxable relocations: the former leaves the original relocations in the resulting binary, whereas the latter will update them to whatever describes the relaxed situation best.
> > > > Given that the codegen is worse without relaxation (i.e., taking the address of a symbol will involve a MOV from the GOT in memory rather than a RIP-relative LEA which is just an arithmetic operation, and so the resulting code is worse in terms of performance, memory footprint and the number of relocation fixups that need to be applied), I'd prefer to only enable this for `ld.lld` if possible.
> > > 
> > > 
> > > @ardbiesheuvel I have not observed this issue with CLANGDWARF, but my testing with that tool chain was incomplete at this time due to other issues. I did observe this with GCC. I can remove the CLANGDWARF changes in this PR and if issues are observed in the future with CLANGDWARF, we can discuss options.
> > 
> > 
> > Interesting. As far as I am aware, it is LLVM's `lld` that retains the `R_X86_64_REX_GOTPCRELX` relocations after relaxing them, whereas GNU `ld` converts them into different ones that more accurately describe the relaxed version of the instruction.
> 
> I do not see this relocation type generated in newer versions of GCC. I only see this relocation type in the final ELF images in older versions of GCC. I believe it was gcc-7.

I actually fixed this for `R_X86_64_REX_GOTPCRELX` relocations in `GenFw` in commit 051ef932bd19, and so I would like to understand which relocations are actually being suppressed here, given that only MOV to LEA is feasible in the position independent code model.

Do you have a reproducer for the issue you are addressing here?

mdkinney:
> > > > > > * Add `-Wl,--no-relax` to `tools_def.template` for `GCC` and `CLANGDWARF` tool chains for `X64`. This aligns with other CPU architecture linker settings. This is required for `X64` because the removal of the `OBJCOPY` operations to not strip symbols introduces the use of ELF relocation types not supported by `GENFW` ELF->PE/COFF conversions. Adding `--no-relax` prevents the use of the unsupported ELF relocation types such as `R_X86_64_REX_GOTPCRELX`.
> > > > > 
> > > > > 
> > > > > Do both GCC and CLANGDWARF require this? The `ld.lld` and `ld.bfd` linkers behave differently when it comes to relaxable relocations: the former leaves the original relocations in the resulting binary, whereas the latter will update them to whatever describes the relaxed situation best.
> > > > > Given that the codegen is worse without relaxation (i.e., taking the address of a symbol will involve a MOV from the GOT in memory rather than a RIP-relative LEA which is just an arithmetic operation, and so the resulting code is worse in terms of performance, memory footprint and the number of relocation fixups that need to be applied), I'd prefer to only enable this for `ld.lld` if possible.
> > > > 
> > > > 
> > > > @ardbiesheuvel I have not observed this issue with CLANGDWARF, but my testing with that tool chain was incomplete at this time due to other issues. I did observe this with GCC. I can remove the CLANGDWARF changes in this PR and if issues are observed in the future with CLANGDWARF, we can discuss options.
> > > 
> > > 
> > > Interesting. As far as I am aware, it is LLVM's `lld` that retains the `R_X86_64_REX_GOTPCRELX` relocations after relaxing them, whereas GNU `ld` converts them into different ones that more accurately describe the relaxed version of the instruction.
> > 
> > 
> > I do not see this relocation type generated in newer versions of GCC. I only see this relocation type in the final ELF images in older versions of GCC. I believe it was gcc-7.
> 
> I actually fixed this for `R_X86_64_REX_GOTPCRELX` relocations in `GenFw` in commit [051ef93](https://github.com/tianocore/edk2/commit/051ef932bd1911ac38a7c1c730b1d7d7b44da4b5), and so I would like to understand which relocations are actually being suppressed here, given that only MOV to LEA is feasible in the position independent code model.
> 
> Do you have a reproducer for the issue you are addressing here?

I need to update the PR description and commit message.  I cross referenced to the incorrect symbolic name for the relocation type.  The relocation type observed by GENFW that was not supported from an ELF image generated by a GCC build was type 0x82.

PR #11650, opened 2025-10-22
MdeModulePkg/AcpiTableDxe: Initialize NeedToInstallTable before use

pttuan:
# Description
Fix a warning from the compiler about the NeedToInstallTable variable being potentially uninitialized before use.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
N/A
## Integration Instructions
N/A

### DIFF
--- a/MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiTableProtocol.c
+++ b/MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiTableProtocol.c
@@ -2158,10 +2158,17 @@ InstallAcpiTableFromAcpiSiliconHob (
       //
       // According ACPI spec, if XDsdt field contains a nonzero value which can be used by the OSPM, then the Dsdt field must be ignored by the OSPM.
       //
+      NeedToInstallTable = NULL;
       if (((EFI_ACPI_3_0_FIXED_ACPI_DESCRIPTION_TABLE *)SocEntryTable)->XDsdt != 0) {
         NeedToInstallTable = (VOID *)(UINTN)((EFI_ACPI_3_0_FIXED_ACPI_DESCRIPTION_TABLE *)SocEntryTable)->XDsdt;
       } else if (((EFI_ACPI_3_0_FIXED_ACPI_DESCRIPTION_TABLE *)SocEntryTable)->Dsdt != 0) {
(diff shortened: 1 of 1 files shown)


MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiTableProtocol.c
```diff
@@ -2158,12 +2158,15 @@ InstallAcpiTableFromAcpiSiliconHob (
...
       // According ACPI spec, if XDsdt field contains a nonzero value which can be used by the OSPM, then the Dsdt field must be ignored by the OSPM.
       //
+      NeedToInstallTable = NULL;
       if (((EFI_ACPI_3_0_FIXED_ACPI_DESCRIPTION_TABLE *)SocEntryTable)->XDsdt != 0) {
         NeedToInstallTable = (VOID *)(UINTN)((EFI_ACPI_3_0_FIXED_ACPI_DESCRIPTION_TABLE *)SocEntryTable)->XDsdt;
       } else if (((EFI_ACPI_3_0_FIXED_ACPI_DESCRIPTION_TABLE *)SocEntryTable)->Dsdt != 0) {
         NeedToInstallTable = (VOID *)(UINTN)((EFI_ACPI_3_0_FIXED_ACPI_DESCRIPTION_TABLE *)SocEntryTable)->Dsdt;
       }
 
+      ASSERT (NeedToInstallTable != NULL);
```
AshrafAliS: in RELEASE BIOS, ASSERT will not hit, do we need to do the break here?

pttuan: NeedToInstallTable cannot be NULL in any cases. The ASSERT statement detects unexpected value during debug build before creating a release build, so a break is not needed here.

MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiTableProtocol.c
```diff
@@ -2158,12 +2158,15 @@ InstallAcpiTableFromAcpiSiliconHob (
...
+
```
lgao4, line 2173: I think the below code should also need to check the value of NeedToInstallTable.

PR #11654, opened 2025-10-24
MdeModulePkg: Libray cumulative codeql issues.

apop5:
# Description

Codeql on MdeModulePkg/Libraries.

The majority of changes were unchecked return statuses that resulted in using variables which may be null/uninitialized. 

There were some conditional uninitialized variables (depending on the conditionals, a variable would be uninitialized and then the value would be used)

One instance of switching to SafeIntLib to allow catching of overflow based on unvalidated input (from Hii).

this is a continuation of #11252 which went stale.

- [ ] Breaking change?
- [x] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Local CI using --codeql.
Verified by booting virtual platforms.

## Integration Instructions
No integration necessary.


MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenanceManagerCustomizedUiSupport.c
```diff
@@ -379,88 +381,98 @@ BmmListThirdPartyDrivers (
...
+          //
+          EmptyLineAfter = FALSE;
+          if (SpecialHandlerFn (String, &NewName, &EmptyLineAfter)) {
+            FreePool (String);
+            String                              = NewName;
+            DriverListPtr[Count].EmptyLineAfter = EmptyLineAfter;
+          }
+        }
+
+        DriverListPtr[Count].PromptId = HiiSetString (HiiHandle, 0, String, NULL);
```
mikebeaton: Should we be calling HiiSetString if String is NULL?

MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenanceManagerCustomizedUiSupport.c
```diff
@@ -379,88 +381,98 @@ BmmListThirdPartyDrivers (
...
+          FreePool (String);
+        }
+
+        String = HiiGetString (HiiHandles[Index], TokenHelp, NULL);
+        if (String == NULL) {
+          String = HiiGetString (HiiHandle, STRING_TOKEN (STR_MISSING_STRING), NULL);
+          ASSERT (String != NULL);
+        }
+
+        DriverListPtr[Count].HelpId = HiiSetString (HiiHandle, 0, String, NULL);
```
mikebeaton: Should we be calling HiiSetString if String is NULL?

MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenance.c
```diff
@@ -1510,7 +1513,7 @@ InitializeBmmConfig (
   // Initialize data which located in BMM main page
   //
   CallbackData->BmmFakeNvData.BootNext = NONE_BOOTNEXT_VALUE;
-  for (Index = 0; Index < BootOptionMenu.MenuNumber; Index++) {
+  for (Index = 0; (UINTN)Index < BootOptionMenu.MenuNumber; Index++) {
```
mikebeaton: What was the reasoning behind changing Index from UINT16 to UINTN in some instances, but casting it from UNINT16 to UINTN (as here) in others?

MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenanceManagerCustomizedUiSupport.c
```diff
@@ -379,88 +381,98 @@ BmmListThirdPartyDrivers (
...
+        Count++;
+        if (Count >= CurrentSize) {
+          DriverListPtr = ReallocatePool (
+                            CurrentSize * sizeof (UI_HII_DRIVER_INSTANCE),
+                            (Count + UI_HII_DRIVER_LIST_SIZE)
+                            * sizeof (UI_HII_DRIVER_INSTANCE),
+                            gHiiDriverList
+                            );
+          ASSERT (DriverListPtr != NULL);
+          if (DriverListPtr != NULL) {
```
mikebeaton: Shouldn't we break if we can't reallocate?

MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenanceManagerCustomizedUiSupport.c
```diff
@@ -379,88 +381,98 @@ BmmListThirdPartyDrivers (
...
-                        );
-      ASSERT (DriverListPtr != NULL);
-      gHiiDriverList = DriverListPtr;
-      CurrentSize   += UI_HII_DRIVER_LIST_SIZE;
-    }
-  }
-
-  FreePool (HiiHandles);
+    Index = 0;
+    while (gHiiDriverList[Index].PromptId != 0) {
```
mikebeaton: As far as I can see, given the exceptional conditions above which we're trying to handle in this PR, gHiiDriverList can be NULL here.

MdeModulePkg/Library/BootMaintenanceManagerUiLib/UpdatePage.c
```diff
@@ -484,11 +484,16 @@ UpdateConsolePage (
...
       break;
+
+    default:
+      return;
   }
 
-  ASSERT (ConsoleCheck != NULL);
+  if (ConsoleCheck == NULL) {
+    ASSERT (ConsoleCheck != NULL);
+  }
```
mikebeaton: !

mikebeaton: Maybe this is intended to have a return as well. It's not clear the original ASSERT (or an added return with it) is really needed, given the code immediately above it.

MdeModulePkg/Library/BrotliCustomDecompressLib/BrotliDecompress.c
```diff
@@ -182,12 +182,12 @@ BrGetDecodedSizeOfBuf (
...
 {
   UINT64  DecodedSize;
-  INTN    Index;
+  UINT8   Index;
 
   /* Parse header */
   DecodedSize = 0;
-  for (Index = EndOffset - 1; Index >= StartOffset; Index--) {
-    DecodedSize = LShiftU64 (DecodedSize, 8) + EncodedData[Index];
+  for (Index = EndOffset; Index > StartOffset; Index--) {
```
mikebeaton: This seems like a separate PR - this is more than just CodeQL?

MdeModulePkg/Library/SmmLockBoxLib/SmmLockBoxMmLib.c
```diff
@@ -834,6 +844,7 @@ RestoreLockBox (
   @retval RETURN_SUCCESS            the information is restored successfully.
   @retval RETURN_NOT_STARTED        it is too early to invoke this interface
   @retval RETURN_UNSUPPORTED        the service is not supported by implementaion.
+  @retval RETURN_OUT_OF_RESOURCES   noT enough resourceS to save the information.
```
mikebeaton: capitalisation typos

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -214,10 +213,12 @@ BmAdjustFvFilePath (
          (VOID **)&LoadedImage
          );
   NewDevicePath = AppendDevicePathNode (DevicePathFromHandle (LoadedImage->DeviceHandle), FvFileNode);
-  FullPath      = BmAdjustFvFilePath (NewDevicePath);
-  FreePool (NewDevicePath);
-  if (FullPath != NULL) {
-    return FullPath;
+  if (NewDevicePath != NULL) {
```
mikebeaton: Dropping through to second stage below is not obviously correct, if AppendDevicePathNode returns NULL. Since it's clearly not supposed to return NULL, we certainly are no longer correctly verifying that the file is NOT in the current firmware volume (which would be the correct reason to drop through). Maybe we should just return with NULL in this case.

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -239,7 +240,12 @@ BmAdjustFvFilePath (
     }
 
     NewDevicePath = AppendDevicePathNode (DevicePathFromHandle (FvHandles[Index]), FvFileNode);
-    FullPath      = BmAdjustFvFilePath (NewDevicePath);
+    if (NewDevicePath == NULL) {
+      ASSERT (NewDevicePath != NULL);
+      continue;
+    }
+
```
mikebeaton: I feel like the approach of just returning NULL immediately in this case might be better here, too.

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -683,48 +695,52 @@ BmExpandFileDevicePath (
   EFI_DEVICE_PATH_PROTOCOL  *NextFullPath;
   BOOLEAN                   GetNext;
 
+  HandleCount = 0;
+  Handles     = NULL;
   EfiBootManagerConnectAll ();
   Status = gBS->LocateHandleBuffer (ByProtocol, &gEfiSimpleFileSystemProtocolGuid, NULL, &HandleCount, &Handles);
-  if (EFI_ERROR (Status)) {
-    HandleCount = 0;
-    Handles     = NULL;
```
mikebeaton, line 690: Swapping these to above is not necessarily exactly the same semantics, depending on how LocateHandleBuffer fails internally.

mikebeaton: In any event, looking at the code below, I believe we could just immediately return NULL here, as that is what would have happened with the previous code in the error case anyway.

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -683,48 +695,52 @@ BmExpandFileDevicePath (
...
-        BlockIo = NULL;
-      }
+        if (((MediaType == 0) && (BlockIo != NULL) && BlockIo->Media->RemovableMedia) ||
+            ((MediaType == 1) && (BlockIo != NULL) && !BlockIo->Media->RemovableMedia) ||
+            ((MediaType == 2) && (BlockIo == NULL))
+            )
+        {
+          NextFullPath = AppendDevicePath (DevicePathFromHandle (Handles[Index]), FilePath);
+          if (NextFullPath == NULL) {
+            continue;
```
mikebeaton: I believe we should return NULL here, as this is a general error (and NULL is already used, more or less implicitly, as the general error return from this function).

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -756,41 +772,39 @@ BmExpandUriDevicePath (
   EFI_DEVICE_PATH_PROTOCOL  *RamDiskDevicePath;
   BOOLEAN                   GetNext;
 
+  HandleCount = 0;
+  Handles     = NULL;
```
mikebeaton: As above, it might be simpler to just immediately return NULL here if LocateHandleBuffer below fails. And we (at least logically, I don't know about compilers?) shouldn't need to initialise these here, in that case.

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -923,6 +937,10 @@ BmExpandPartitionDevicePath (
       // partial partition boot option. Second, check whether the instance could be connected.
       //
       Instance = GetNextDevicePathInstance (&TempNewDevicePath, &Size);
+      if (Instance == NULL) {
+        break;
```
mikebeaton: I don't know if anyone else wants to chime in, but it seems to me that again, in these general error conditions, returning NULL immediately is about as good as we can do, and better than carrying on round the loop in a state we were certainly never supposed to be in.

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -941,31 +959,33 @@ BmExpandPartitionDevicePath (
           //   2. ACPI()/PCI()/ATA()/Partition()/Partition(A2)/EFI/BootX64.EFI
           // For simplicity, only #1 is returned.
           //
-          FullPath = BmGetNextLoadOptionDevicePath (TempDevicePath, NULL);
-          FreePool (TempDevicePath);
+          if (TempDevicePath != NULL) {
```
mikebeaton: Return NULL?

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -991,45 +1012,46 @@ BmExpandPartitionDevicePath (
...
-      if (BlockIoDevicePath == NULL) {
-        continue;
-      }
+        if (BmMatchPartitionDevicePathNode (BlockIoDevicePath, (HARDDRIVE_DEVICE_PATH *)FilePath)) {
+          //
+          // Find the matched partition device path
+          //
+          TempDevicePath = AppendDevicePath (BlockIoDevicePath, NextDevicePathNode (FilePath));
+          if (TempDevicePath == NULL) {
+            continue;
```
mikebeaton: return NULL?

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -1160,7 +1185,11 @@ BmExpandMediaDevicePath (
     // Get the device path size of SimpleFileSystem handle
     //
     TempDevicePath = DevicePathFromHandle (SimpleFileSystemHandles[Index]);
-    TempSize       = GetDevicePathSize (TempDevicePath) - END_DEVICE_PATH_LENGTH;
+    if (TempDevicePath == NULL) {
+      continue;
```
mikebeaton: return NULL?

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -1248,19 +1279,18 @@ BmExpandNetworkFileSystem (
...
   Status = gBS->LocateHandleBuffer (
                   ByProtocol,
                   &gEfiBlockIoProtocolGuid,
                   NULL,
                   &HandleCount,
                   &Handles
                   );
-  if (EFI_ERROR (Status)) {
-    Handles     = NULL;
-    HandleCount = 0;
```
mikebeaton, line 1260: As before, moving these above LocateHandleBuffer is not the same semantics and depends (more than the existing code does) on how LocateHandleBuffer fails internally.

mikebeaton: Also as before, I believe:
a) We can just return NULL immediately here, as this is what the existing code effectively does anyway
b) There is no logical need to initialise these two before the call (and none after, either, if we are simply returning)

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -2304,12 +2334,20 @@ BmEnumerateBootOptions (
...
       BootOptions = ReallocatePool (
                       sizeof (EFI_BOOT_MANAGER_LOAD_OPTION) * (*BootOptionCount),
                       sizeof (EFI_BOOT_MANAGER_LOAD_OPTION) * (*BootOptionCount + 1),
                       BootOptions
                       );
-      ASSERT (BootOptions != NULL);
+      if (BootOptions == NULL) {
+        ASSERT (BootOptions != NULL);
+        FreePool (Description);
+        continue;
```
mikebeaton: Once again, a general (presumably out of resources) error, and it seems we would be better to return NULL here rather than to plough onwards?

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -2355,12 +2393,20 @@ BmEnumerateBootOptions (
...
     BootOptions = ReallocatePool (
                     sizeof (EFI_BOOT_MANAGER_LOAD_OPTION) * (*BootOptionCount),
                     sizeof (EFI_BOOT_MANAGER_LOAD_OPTION) * (*BootOptionCount + 1),
                     BootOptions
                     );
-    ASSERT (BootOptions != NULL);
+    if (BootOptions == NULL) {
+      ASSERT (BootOptions != NULL);
+      FreePool (Description);
+      continue;
```
mikebeaton: return NULL?

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -2399,12 +2445,20 @@ BmEnumerateBootOptions (
...
+
     BootOptions = ReallocatePool (
                     sizeof (EFI_BOOT_MANAGER_LOAD_OPTION) * (*BootOptionCount),
                     sizeof (EFI_BOOT_MANAGER_LOAD_OPTION) * (*BootOptionCount + 1),
                     BootOptions
                     );
-    ASSERT (BootOptions != NULL);
+    if (BootOptions == NULL) {
+      ASSERT (BootOptions != NULL);
+      FreePool (Description);
```
mikebeaton: return NULL?

MdeModulePkg/Library/UefiBootManagerLib/BmConsole.c
```diff
@@ -204,9 +206,14 @@ EfiBootManagerGetGopDevicePath (
...
-          gBS->FreePool (ReturnDevicePath);
-          gBS->FreePool (TempDevicePath);
+          if (ReturnDevicePath != NULL) {
+            GopPool = AppendDevicePathInstance (GopPool, ReturnDevicePath);
+            gBS->FreePool (ReturnDevicePath);
+          }
+
+          if (TempDevicePath != NULL) {
+            gBS->FreePool (TempDevicePath);
+          }
```
mikebeaton: There's a mistake here. With the current change you may fail to generate a new GopPool, but still free the old GopPool. I believe you want:
```
          ReturnDevicePath = EfiBootManagerGetGopDevicePath (OpenInfoBuffer[Index].ControllerHandle);
          if (ReturnDevicePath != NULL) {
            TempDevicePath   = GopPool;
            GopPool          = AppendDevicePathInstance (GopPool, ReturnDevicePath);
            if (TempDevicePath != NULL) {
              gBS->FreePool (TempDevicePath);
            }
            gBS->FreePool (ReturnDevicePath);
          }
```

MdeModulePkg/Library/UefiBootManagerLib/BmConsole.c
```diff
@@ -464,29 +474,34 @@ EfiBootManagerUpdateConsoleVariable (
       // Check if there is part of CustomizedConDevicePath in NewDevicePath, delete it.
       //
       NewDevicePath = BmDelPartMatchInstance (NewDevicePath, CustomizedConDevicePath);
-      //
-      // In the first check, the default console variable will be _ModuleEntryPoint,
-      // just append current customized device path
-      //
-      TempNewDevicePath = NewDevicePath;
-      NewDevicePath     = AppendDevicePathInstance (NewDevicePath, CustomizedConDevicePath);
+      if (NewDevicePath != NULL) {
```
mikebeaton: I don't think you want this new condition. TempNewDevicePath _should_ be NULL, if NewDevicePath is.

MdeModulePkg/Library/UefiBootManagerLib/BmHotkey.c
```diff
@@ -1014,27 +1025,31 @@ EfiBootManagerAddKeyOptionVariable (
...
+      }
     }
-  }
 
-  BmFreeKeyOptions (KeyOptions, KeyOptionCount);
+    BmFreeKeyOptions (KeyOptions, KeyOptionCount);
+  }
 
   if (Index < KeyOptionCount) {
     return EFI_ALREADY_STARTED;
```
mikebeaton, line 1040: If you move the close of the `if (KeyOptions != NULL) {` conditional down to here, then you don't need the slightly artificial looking change above of initialising the loop variable outside the `for`.

MdeModulePkg/Library/UefiBootManagerLib/BmHotkey.c
```diff
@@ -1014,27 +1025,31 @@ EfiBootManagerAddKeyOptionVariable (
     return Status;
   }
 
+  Index           = 0;
+  KeyOptionCount  = 0;
```
mikebeaton: You don't need this extra/changed initialisation, if you make the small change [mentioned in the code immediately below](https://github.com/tianocore/edk2/pull/11654/changes#r2658861866).

MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenanceManagerCustomizedUiSupport.c
apop5: Are you considering the case where STR_MISSING_STRING is not in the database for this question?

MdeModulePkg/Library/BootMaintenanceManagerUiLib/UpdatePage.c
apop5: Yes, adding a return here.

MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenanceManagerCustomizedUiSupport.c
mikebeaton: I'm just looking at the implementations of HiiSetString (in MdeModulePkg/Universal/HiiDatabaseDxe/String.c and MdeModulePkg/Library/UefiHiiLib/HiiString.c). One returns with invalid parameter if String is null, and one ASSERTS that String should not be null but carries on regardless.

mikebeaton: Digging in a bit further, the relevant implementation is the one in MdeModulePkg/Universal/HiiDatabaseDxe/String.c (from gEfiHiiStringProtocolGuid installed in MdeModulePkg/Universal/HiiDatabaseDxe/HiiDatabaseEntry.c) and I would say the behaviour is defined and will always return a StringId of zero in case String is NULL. Since we have already done AllocateZeroPool this is, at least, optional - IMO - if you agree! So my tendency would still be to avoid calling it (and relying on any possible implementation of it handling this correctly).

MdeModulePkg/Library/BootMaintenanceManagerUiLib/UpdatePage.c
mikebeaton: As mentioned above, if looking at the preceding code, it really is pretty tautologous to even have this ASSERT and return at all. (I mean, should we be checking the other two variables which are set in the case statements as well??) I suppose I don't have a super-strong preference, but if I was choosing I'd remove it (or at least update it to check the other two variables as well, but that is certainly belt and braces and something else to hold pants/trousers up!).

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
apop5: adding a goto exit to force freeing of LocateHandleBuffer

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
apop5: adding a goto exit to get allocated resources freed.

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
apop5: reusing same goto exit

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
apop5: same goto exit.

MdeModulePkg/Library/UefiBootManagerLib/BmConsole.c
apop5: good catch, changed it to your suggestion.

MdeModulePkg/Library/UefiBootManagerLib/BmConsole.c
apop5: dropped the change.

MdeModulePkg/Library/UefiBootManagerLib/BmHotkey.c
apop5: moved it down

MdeModulePkg/Library/BootMaintenanceManagerUiLib/BootMaintenance.c
apop5: These casts were incorrectly applied. They still exposed the problem that casting from a UINT16 would potentionaly never be able to exit the loop (a uint16 can never match a Uintn).

All the casts inside of loop controls have been removed. The codeql errors were addressed by changing the UINT16s to UINTNs, even though Hii menus, etc have a maximum size of UINT16.

MdeModulePkg/Library/BootMaintenanceManagerUiLib/UpdatePage.c
apop5: with the addition of the default return, the other two variables will always be a valid value.

The console check is the only variable that is being pulled out of the callback data and contains no sanity check

MdeModulePkg/Library/BrotliCustomDecompressLib/BrotliDecompress.c
apop5: Removed the changes to the for loop, and addressed the size different by increasing the Index variable.

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
apop5: This is good advice, I took it and applied to other places which would benefit as well.

MdeModulePkg/Library/DxeCorePerformanceLib/DxeCorePerformanceLib.c
```diff
@@ -1218,14 +1234,20 @@ InsertFpdtRecord (
...
+      if (EFI_ERROR (Status)) {
+        DEBUG ((DEBUG_ERROR, "Failed to get Module Info from Handle! Status = %r\n", Status));
+      }
+
+      StringPtr = NULL;
+
       if (String != NULL) {
         StringPtr = String;
-      } else {
+      } else if (ModuleName != NULL) {
```
mikebeaton: This is incorrect. ModuleName can't be null, it's a local buffer.

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -214,7 +213,11 @@ BmAdjustFvFilePath (
          (VOID **)&LoadedImage
          );
   NewDevicePath = AppendDevicePathNode (DevicePathFromHandle (LoadedImage->DeviceHandle), FvFileNode);
-  FullPath      = BmAdjustFvFilePath (NewDevicePath);
+  if (NewDevicePath == NULL) {
+    return NULL;
```
mikebeaton: Since line 246 below has an ASSERT, shouldn't this be the same?

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -239,7 +242,12 @@ BmAdjustFvFilePath (
     }
 
     NewDevicePath = AppendDevicePathNode (DevicePathFromHandle (FvHandles[Index]), FvFileNode);
-    FullPath      = BmAdjustFvFilePath (NewDevicePath);
+    if (NewDevicePath == NULL) {
+      ASSERT (NewDevicePath != NULL);
+      return NULL;
```
mikebeaton: We need to free FvHandles here.

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -923,6 +941,10 @@ BmExpandPartitionDevicePath (
       // partial partition boot option. Second, check whether the instance could be connected.
       //
       Instance = GetNextDevicePathInstance (&TempNewDevicePath, &Size);
+      if (Instance == NULL) {
+        return NULL;
```
mikebeaton: We need to free CachedDevicePath here.

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -941,6 +963,10 @@ BmExpandPartitionDevicePath (
           //   2. ACPI()/PCI()/ATA()/Partition()/Partition(A2)/EFI/BootX64.EFI
           // For simplicity, only #1 is returned.
           //
+          if (TempDevicePath == NULL) {
+            return NULL;
```
mikebeaton: And we need to free CachedDevicePath here.

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -983,6 +1009,7 @@ BmExpandPartitionDevicePath (
   // to search all devices in the system for a matched partition
   //
   BlockIoBuffer       = NULL;
+  BlockIoHandleCount  = 0;
```
mikebeaton: I am not sure we needed to add this initializer?

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -992,8 +1019,7 @@ BmExpandPartitionDevicePath (
 
     Status = gBS->LocateHandleBuffer (ByProtocol, &gEfiBlockIoProtocolGuid, NULL, &BlockIoHandleCount, &BlockIoBuffer);
     if (EFI_ERROR (Status)) {
-      BlockIoHandleCount = 0;
-      BlockIoBuffer      = NULL;
+      return NULL;
```
mikebeaton: I am not sure this is right? If there were no BLOCK_IO devices, shouldn't we carry on with the rest of the method anyway?

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -1169,13 +1206,16 @@ BmExpandMediaDevicePath (
...
         break;
       } else {
-        GetNext = (BOOLEAN)(CompareMem (NextFullPath, FullPath, GetDevicePathSize (NextFullPath)) == 0);
-        FreePool (NextFullPath);
-        NextFullPath = NULL;
+        if (NextFullPath != NULL) {
+          GetNext = (BOOLEAN)(CompareMem (NextFullPath, FullPath, GetDevicePathSize (NextFullPath)) == 0);
+          FreePool (NextFullPath);
+          NextFullPath = NULL;
+        }
```
mikebeaton: I think we should abort and return NULL from the method if NextFullPath is NULL, rather than plowing on.

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -1581,7 +1620,7 @@ BmExpandLoadFiles (
     }
 
     for (Index = 0; Index < HandleCount; Index++) {
-      if (BmMatchHttpBootDevicePath (DevicePathFromHandle (Handles[Index]), FilePath)) {
+      if ((Handles != NULL) && BmMatchHttpBootDevicePath (DevicePathFromHandle (Handles[Index]), FilePath)) {
```
mikebeaton: This is kind of a false positive as well. The added check is only needed if LocateHandleBuffer can return a NULL Handles list together with a non-zero HandleCount.

There would need to be an awful lot of defensive coding added all over the codebase, if we really need to assume this can happen.

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
mikebeaton, line 2460: I'm not clear why this gets a free pass (for returning non-NULL), when EfiBootManagerGetLoadOptions below doesn't?

mikebeaton:
@apop5 - For reviewing and re-reviewing, it would really help if this PR was split out into smaller chunks. At least maybe the 'Fix missing NULL tests' commit as a separate PR from the rest.

MdeModulePkg/Library/UefiBootManagerLib/BmBootDescription.c
```diff
@@ -571,7 +578,11 @@ BmGetUsbDescription (
 
   DescMaxSize = StrSize (Manufacturer) + StrSize (Product) + StrSize (SerialNumber);
   Description = AllocateZeroPool (DescMaxSize);
-  ASSERT (Description != NULL);
+  if (Description == NULL) {
+    ASSERT (Description != NULL);
+    return NULL;
```
mikebeaton: We still need to do some of the FreePool's below.

MdeModulePkg/Library/UefiBootManagerLib/BmBootDescription.c
mikebeaton, line 1065: Should we not return NULL early here, too?

MdeModulePkg/Library/UefiBootManagerLib/BmConsole.c
```diff
@@ -204,9 +206,19 @@ EfiBootManagerGetGopDevicePath (
           DEBUG ((DEBUG_INFO, "[Bds] Looking for GOP child deeper ... \n"));
           TempDevicePath   = GopPool;
```
mikebeaton, line 205: This needs to be removed...

MdeModulePkg/Library/UefiBootManagerLib/BmConsole.c
```diff
@@ -204,9 +206,19 @@ EfiBootManagerGetGopDevicePath (
...
+            if (TempDevicePath != NULL) {
+              gBS->FreePool (TempDevicePath);
+            }
+
+            gBS->FreePool (ReturnDevicePath);
+          }
+
+          if (TempDevicePath != NULL) {
+            gBS->FreePool (TempDevicePath);
+          }
```
mikebeaton: And this needs to be moved inside the if statement.

MdeModulePkg/Library/UefiBootManagerLib/BmConsole.c
```diff
@@ -426,6 +438,9 @@ EfiBootManagerUpdateConsoleVariable (
   EFI_DEVICE_PATH_PROTOCOL  *NewDevicePath;
   EFI_DEVICE_PATH_PROTOCOL  *TempNewDevicePath;
 
+  Status            = EFI_SUCCESS;
+  TempNewDevicePath = NULL;
```
mikebeaton: I don't believe TempNewDevicePath needs to be initialized.

MdeModulePkg/Library/UefiBootManagerLib/BmConsole.c
```diff
@@ -464,29 +479,33 @@ EfiBootManagerUpdateConsoleVariable (
...
-  //
-  Status = gRT->SetVariable (
-                  mConVarName[ConsoleType],
-                  &gEfiGlobalVariableGuid,
-                  EFI_VARIABLE_BOOTSERVICE_ACCESS | EFI_VARIABLE_RUNTIME_ACCESS
-                  | ((ConsoleType < ConInDev) ? EFI_VARIABLE_NON_VOLATILE : 0),
-                  GetDevicePathSize (NewDevicePath),
-                  NewDevicePath
-                  );
+  if (NewDevicePath != NULL) {
```
mikebeaton: It might be nicer to not initialize Status at the top, but to set it here to the appropriate value, if this condition does not pass.

And it looks like the appropriate value should probably be EFI_UNSUPPORTED, not EFI_SUCCESS?

MdeModulePkg/Library/UefiBootManagerLib/BmDriverHealth.c
```diff
@@ -540,6 +540,7 @@ BmRepairAllControllers (
   UINT32                               MaxRepairCount;
   UINT32                               RepairCount;
 
+  Count = 0;
```
mikebeaton: This does not appear to need initializing?

MdeModulePkg/Library/UefiBootManagerLib/BmDriverHealth.c
```diff
@@ -561,6 +565,10 @@ BmRepairAllControllers (
     // Deal with Repair Required
     //
     DriverHealthInfo = EfiBootManagerGetDriverHealthInfo (&Count);
+    if (DriverHealthInfo == NULL) {
+      return;
```
mikebeaton: This looks like it should have an ASSERT added.

MdeModulePkg/Library/UefiBootManagerLib/BmDriverHealth.c
```diff
@@ -622,6 +630,10 @@ BmRepairAllControllers (
   RebootRequired    = FALSE;
   ReconnectRequired = FALSE;
   DriverHealthInfo  = EfiBootManagerGetDriverHealthInfo (&Count);
+  if (DriverHealthInfo == NULL) {
+    return;
```
mikebeaton: This looks like it should have an ASSERT added.

apop5:
> @apop5 - For reviewing and re-reviewing, it would really help if this PR was split out into smaller chunks. At least maybe the 'Fix missing NULL tests' commit as a separate PR from the rest.

I whole heartedly agree. Too big of a pr makes it difficult to review, and to get all the feedback implemented correctly. After these merge, I don't plan any more large change sets, or I will break them up smaller than this.

MdeModulePkg/Library/UefiBootManagerLib/BmHotkey.c
mikebeaton: @apop5 - I can't unresolve anything, but these are not changed.

MdeModulePkg/Library/UefiBootManagerLib/BmHotkey.c
mikebeaton: @apop5 - I can't unresolve anything, but these are not changed.

MdeModulePkg/Library/UefiBootManagerLib/BmLoadOption.c
mikebeaton: These conditionals after each other don't make sense.

MdeModulePkg/Library/UefiBootManagerLib/BmLoadOption.c
```diff
@@ -455,7 +475,10 @@ EfiBootManagerSortLoadOptionVariable (
   // Create new ****Order variable
   //
   OptionOrder = AllocatePool (LoadOptionCount * sizeof (UINT16));
-  ASSERT (OptionOrder != NULL);
```
mikebeaton, line 458: We've lost this ASSERT.

MdeModulePkg/Library/UefiBootManagerLib/BmLoadOption.c
```diff
@@ -1107,7 +1134,10 @@ EfiBootManagerGetLoadOptions (
     *OptionCount = OptionOrderSize / sizeof (UINT16);
 
     Options = AllocatePool (*OptionCount * sizeof (EFI_BOOT_MANAGER_LOAD_OPTION));
-    ASSERT (Options != NULL);
+    if (Options == NULL) {
+      ASSERT (Options != NULL);
+      return NULL;
```
mikebeaton: We need to free OptionOrder.

MdeModulePkg/Library/UefiBootManagerLib/BmLoadOption.c
```diff
@@ -1130,7 +1160,12 @@ EfiBootManagerGetLoadOptions (
 
     if (OptionIndex < *OptionCount) {
       Options = ReallocatePool (*OptionCount * sizeof (EFI_BOOT_MANAGER_LOAD_OPTION), OptionIndex * sizeof (EFI_BOOT_MANAGER_LOAD_OPTION), Options);
-      ASSERT (Options != NULL);
+      if (Options == NULL) {
+        ASSERT (Options != NULL);
+        FreePool (OptionOrder);
```
mikebeaton: This is incorrect, it has just been freed.

MdeModulePkg/Library/UefiBootManagerLib/BmMisc.c
```diff
@@ -57,7 +57,10 @@ BmDelPartMatchInstance (
       }
     }
 
-    FreePool (Instance);
+    if (Instance != NULL) {
```
mikebeaton:
```
while (Instance != NULL) {
   ...
   if (Instance != NULL) {
```
?

MdeModulePkg/Library/UefiHiiLib/HiiLib.c
```diff
@@ -82,7 +89,10 @@ InternalHiiExtractGuidFromHiiHandle (
 
   if (Status == EFI_BUFFER_TOO_SMALL) {
     HiiPackageList = AllocatePool (BufferSize);
-    ASSERT (HiiPackageList != NULL);
+    if (HiiPackageList == NULL) {
+      ASSERT (HiiPackageList != NULL);
```
mikebeaton: I'm not convinced it's good practice to pair EFI_OUT_OF_RESOURCES returns with ASSERTs every time? I don't know if someone more of a gatekeeper than me can comment?

MdeModulePkg/Library/UefiHiiLib/HiiLib.c
```diff
@@ -95,7 +105,9 @@ InternalHiiExtractGuidFromHiiHandle (
   //
   // Extract GUID
   //
-  CopyGuid (Guid, &HiiPackageList->PackageListGuid);
+  if (HiiPackageList != NULL) {
```
mikebeaton: Given your above change, this can no longer be NULL.

MdeModulePkg/Library/UefiHiiLib/HiiLib.c
```diff
@@ -2210,7 +2246,10 @@ InternalHiiValidateCurrentSetting (
     // Skip header part.
     //
     StringPtr = StrStr (ConfigResp, L"PATH=");
-    ASSERT (StringPtr != NULL);
+    if (StringPtr == NULL) {
+      ASSERT (StringPtr != NULL);
+      return EFI_OUT_OF_RESOURCES;
```
mikebeaton: I believe this should be EFI_INVALID_PARAMETER.

MdeModulePkg/Library/UefiHiiLib/HiiLib.c
mikebeaton, line 2890: Maybe we should add an ASSERT here?

MdeModulePkg/Library/UefiHiiLib/HiiLib.c
```diff
@@ -2896,6 +2938,11 @@ HiiGetBrowserData (
   Size       = (StrLen (mConfigHdrTemplate) + 1) * sizeof (CHAR16);
   Size       = Size + (StrLen (ResultsData) + 1) * sizeof (CHAR16);
   ConfigResp = AllocateZeroPool (Size);
+  if (ConfigResp == NULL) {
+    FreePool (ResultsData);
+    return FALSE;
```
mikebeaton: Maybe we should add an ASSERT here?

MdeModulePkg/Library/UefiHiiLib/HiiLib.c
```diff
@@ -2976,6 +3023,10 @@ HiiSetBrowserData (
     //
     Size          = (StrLen (mConfigHdrTemplate) + 32 + 1) * sizeof (CHAR16);
     ConfigRequest = AllocateZeroPool (Size);
+    if (ConfigRequest == NULL) {
+      return FALSE;
```
mikebeaton: Maybe we should add an ASSERT here?

MdeModulePkg/Library/UefiHiiLib/HiiLib.c
```diff
@@ -2985,6 +3036,10 @@ HiiSetBrowserData (
     Size          = StrLen (mConfigHdrTemplate) * sizeof (CHAR16);
     Size          = Size + (StrLen (RequestElement) + 1) * sizeof (CHAR16);
     ConfigRequest = AllocateZeroPool (Size);
+    if (ConfigRequest == NULL) {
+      return FALSE;
```
mikebeaton: Maybe we should add an ASSERT here?

MdeModulePkg/Library/VarCheckHiiLib/VarCheckHiiGenFromFv.c
```diff
@@ -178,6 +178,8 @@ ParseFfs (
   UINT8                   NumberofMatchingVfrBin;
   UINTN                   *VfrBinBaseAddress;
 
+  VfrBinBaseAddress = NULL;
```
mikebeaton: This only needs initializing if the method it is passed to, which is meant to initialize it, doesn't do so. Do we need to allow for that?

MdeModulePkg/Library/VarCheckPcdLib/VarCheckPcdLibNullClass.c
```diff
@@ -429,6 +429,8 @@ LocateVarCheckPcdBin (
   VAR_CHECK_PCD_VARIABLE_HEADER  *VarCheckPcdBin;
   UINTN                          VarCheckPcdBinSize;
 
+  VarCheckPcdBin = NULL;
```
mikebeaton: Similar comment to preceding. This only needs initializing if the method it is passed to, which is meant to initialize it, doesn't do so. (And this instance is worse, because if that is the case, it would then be passed as NULL to AllocateRuntimeCopyPool without any further checks.)

MdeModulePkg/Library/UefiHiiLib/HiiLib.c
```diff
@@ -2146,14 +2150,20 @@ GetBlockDataInfo (
...
+        (NewBlockData->Offset <= Sum1) &&
+        (Sum2 > Sum1))
+    {
+      Sum1 = BlockData->Width;
+      if (!EFI_ERROR (SafeUint16Sub (Sum2, BlockData->Offset, &BlockData->Width))) {
+        RemoveEntryList (Link->ForwardLink);
+        FreePool (NewBlockData);
+        continue;
+      } else {
+        BlockData->Width = Sum1;
```
mikebeaton: The commit for the changes here could definitely be moved into a separate PR, not least to give someone else other than my poor aching head a chance to review it.

mikebeaton:
> I whole heartedly agree. Too big of a pr makes it difficult to review, and to get all the feedback implemented correctly. After these merge, I don't plan any more large change sets, or I will break them up smaller than this.

I think I'm rather suggesting splitting them up now, since we're in a review and re-review cycle.

At least I think into:

 - 'Fix missing NULL tests' (which is still enormous, but slightly less so if the rest is split out)
 - 'Fix Comparison overflow'
 - Everything else

apop5:
> > I whole heartedly agree. Too big of a pr makes it difficult to review, and to get all the feedback implemented correctly. After these merge, I don't plan any more large change sets, or I will break them up smaller than this.
> 
> I think I'm rather suggesting splitting them up now, since we're in a review and re-review cycle.
> 
> At least I think into:
> 
> * 'Fix missing NULL tests' (which is still enormous, but slightly less so if the rest is split out)
> * 'Fix Comparison overflow'
> * Everything else

Thanks Mike,

I will apply the current set of feedback, and then split this PR following your guidance.

Would it be okay for this PR to become "Everything Else" and just open 2 others for NUL and Comparison?

MdeModulePkg/Library/BootManagerUiLib/BootManager.c
```diff
@@ -853,6 +866,9 @@ BootManagerCallback (
   }
 
   BootOption = EfiBootManagerGetLoadOptions (&BootOptionCount, LoadOptionTypeBoot);
+  if (BootOption == NULL) {
+    return EFI_INVALID_PARAMETER;
```
os-d: nit: maybe EFI_UNSUPPORTED? That return is already made here, but needs to be added to function header.

MdeModulePkg/Library/DxeCorePerformanceLib/DxeCorePerformanceLib.c
```diff
@@ -1074,7 +1074,11 @@ InsertFpdtRecord (
   switch (PerfId) {
     case MODULE_START_ID:
     case MODULE_END_ID:
-      GetModuleInfoFromHandle ((EFI_HANDLE)CallerIdentifier, ModuleName, sizeof (ModuleName), &ModuleGuid);
+      Status = GetModuleInfoFromHandle ((EFI_HANDLE)CallerIdentifier, ModuleName, sizeof (ModuleName), &ModuleGuid);
+      if (EFI_ERROR (Status)) {
+        DEBUG ((DEBUG_ERROR, "Failed to get Module Info from Handle! Status = %r\n", Status));
+      }
```
os-d: For two of these, you set `StringPtr` to `NULL` and then others just continue through. `ModuleName` gets used after this, is it safe?

MdeModulePkg/Library/DxePrintLibPrint2Protocol/PrintLib.c
```diff
@@ -2060,7 +2060,11 @@ InternalPrintLibSPrintMarker (
     return (LengthToReturn / BytesPerOutputCharacter);
   }
 
-  ASSERT (Buffer != NULL);
+  if (Buffer == NULL) {
```
os-d: This check seems like it should be much higher up because we used `Buffer` in `InteralPrintLibFillBuffer` in a bunch of places.

apop5: There is a check at the start, SAFE_PRINT_CONSTRAINT_CHECK ((Buffer != NULL), 0);

Buffer is getting checked multiple times through the function, but it is being manipulated multiple times as well

I think I am going to drop this specific change.

MdeModulePkg/Library/BootManagerUiLib/BootManager.c
apop5: Changing to unsupported.

I was reusing the EFI_INVALID_PARAMTER, but that doesn't even appear to be a specification defined return value

https://uefi.org/specs/UEFI/2.11/35_HII_Configuration_Processing_and_Browser_Protocol.html#efi-hii-config-access-protocol-callback

MdeModulePkg/Library/DxeCorePerformanceLib/DxeCorePerformanceLib.c
apop5: ModuleName is a local buffer that was zeroed

Using this will result it in just getting a null terminated zero length string, which was matching existing functionality.

The debug message will satisfy code scanning, and will match existing functionality.

os-d: If the return status from this function really doesn't matter and doesn't change behavior, can we just include this in the codeql ignore file?

apop5: dropped all the changes in the file.

apop5:
@os-d @dandanbi @jiaxinwu @lgao4 @niruiyu @khaliid2040

Reviews would be appreciated. I would really like to get this in before the stable tag.

lgao4 (approved):

khaliid2040 (approved):

os-d (approved):

PR #11656, opened 2025-10-25
MdeModulePkg: Xhci DMA buffer Allocation fix (#1539)

eeshanl:
# Description

Changes Pool allocations to use the PciIo interface's AllocateBuffer routine

DMA buffers should be page aligned and always allocated with the appropriate PciIo interface.

On some platforms, where IoMmu is enabled it is possible that if two allocations occur within the same page, and one gets unmapped via PciIo->Unmap, it will also unmap the other allocated address.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on physical device

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c
+++ b/MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c
@@ -1396,6 +1396,7 @@ XhciDelAsyncIntTransfer (
   LIST_ENTRY              *Next;
   URB                     *Urb;
   VOID                    *UrbData;
+  UINTN                   DataLen;
   EFI_USB_DATA_DIRECTION  Direction;
   EFI_STATUS              Status;
 
@@ -1425,9 +1426,15 @@ XhciDelAsyncIntTransfer (
       // and allocates and manages its own data buffer, so free it here.
       //
       UrbData = Urb->Data;
+      DataLen = Urb->DataLen;
       XhcFreeUrb (Xhc, Urb);
       if (UrbData != NULL) {
-        FreePool (UrbData);
+        Status = Xhc->PciIo->FreeBuffer (Xhc->PciIo, EFI_SIZE_TO_PAGES (DataLen), UrbData);
+        if (EFI_ERROR (Status)) {
+          DEBUG ((DEBUG_ERROR, "%a: failed to free buffer via PciIo\n", __func__));
+          ASSERT_EFI_ERROR (Status);
+          return EFI_SUCCESS;
+        }
       }
 
       return EFI_SUCCESS;
@@ -1452,6 +1459,7 @@ XhciDelAllAsyncIntTransfers (
   LIST_ENTRY  *Next;
   URB         *Urb;
   VOID        *UrbData;
+  UINTN       DataLen;
   EFI_STATUS  Status;
 
   BASE_LIST_FOR_EACH_SAFE (Entry, Next, &Xhc->AsyncIntTransfers) {
@@ -1472,9 +1480,13 @@ XhciDelAllAsyncIntTransfers (
     // and allocates and manages its own data buffer, so free it here.
     //
     UrbData = Urb->Data;
+    DataLen = Urb->DataLen;
     XhcFreeUrb (Xhc, Urb);
     if (UrbData != NULL) {
-      FreePool (UrbData);
+      Status = Xhc->PciIo->FreeBuffer (Xhc->PciIo, EFI_SIZE_TO_PAGES (DataLen), UrbData);
+      if (EFI_ERROR (Status)) {
+        DEBUG ((DEBUG_ERROR, "%a: failed to free buffer via PciIo\n", __func__));
+      }
     }
   }
 }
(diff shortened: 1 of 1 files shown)


vishalo:
@josephoresko

makubacki:
@eeshanl, can you please update the test procedure with more details about the type of device and testing performed?

makubacki:
Also, please wait for @ardbiesheuvel to review this before merging.

ardbiesheuvel:
> DMA buffers should be page aligned and always allocated with the appropriate PciIo interface.
> 

This is inaccurate. Common buffers (those that may be accessed by the CPU and the bus master at the same time) need to be allocated with the appropriate PciIo API, and are therefore page granular, although this is not a fundamental requirement.
The `PciIo::Map()` interface permits the mapping of arbitrary memory regions for streaming DMA., i.e., where a buffer is handed back and forth between the bus master and the CPU, but only ever accessed by one at a time.

> On some platforms, where IoMmu is enabled it is possible that if two allocations occur within the same page, and one gets unmapped via PciIo->Unmap, it will also unmap the other allocated address.
> 

This is a bug in the IoMmu implementation. A mapping should not disappear because another unrelated mapping is released.

In any case, it is not necessary to use `PciIo::AllocateBuffer()` here - ordinary page based allocations should be sufficient to work around the problem. However, I don't think there is anything wrong with this driver, apart from the fact that it uses pool allocations for DMA without taking the DMA alignment (in `CpuArchProtocol::DmaBufferAlignment`) into account, although I'm not sure how a `UEFI_DRIVER` should access this information.

TL;DR using page based allocations is acceptable here, as it ensures the DMA buffer alignment, but not using `PciIo::AllocateBuffer()`.

Edit: the underlying `PciIo` implementation will take the DMA buffer alignment into account, and will revert to bounce buffering if needed. I did notice that `NonDiscoverablePciDeviceDxe` is hopelessly inefficient in this regard when using non-coherent DMA, so I'll be proposing a fix for that shortly.

eeshanl:
> > DMA buffers should be page aligned and always allocated with the appropriate PciIo interface.
> 
> This is inaccurate. Common buffers (those that may be accessed by the CPU and the bus master at the same time) need to be allocated with the appropriate PciIo API, and are therefore page granular, although this is not a fundamental requirement. The `PciIo::Map()` interface permits the mapping of arbitrary memory regions for streaming DMA., i.e., where a buffer is handed back and forth between the bus master and the CPU, but only ever accessed by one at a time.
> 
> > On some platforms, where IoMmu is enabled it is possible that if two allocations occur within the same page, and one gets unmapped via PciIo->Unmap, it will also unmap the other allocated address.
> 
> This is a bug in the IoMmu implementation. A mapping should not disappear because another unrelated mapping is released.
> 
> In any case, it is not necessary to use `PciIo::AllocateBuffer()` here - ordinary page based allocations should be sufficient to work around the problem. However, I don't think there is anything wrong with this driver, apart from the fact that it uses pool allocations for DMA without taking the DMA alignment (in `CpuArchProtocol::DmaBufferAlignment`) into account, although I'm not sure how a `UEFI_DRIVER` should access this information.
> 
> TL;DR using page based allocations is acceptable here, as it ensures the DMA buffer alignment, but not using `PciIo::AllocateBuffer()`.
> 
> Edit: the underlying `PciIo` implementation will take the DMA buffer alignment into account, and will revert to bounce buffering if needed. I did notice that `NonDiscoverablePciDeviceDxe` is hopelessly inefficient in this regard when using non-coherent DMA, so I'll be proposing a fix for that shortly.

Thank you for the feedback, I'll go ahead and change this to just use standard page allocations.
I would prefer that all DMA buffers at-least use aligned pages for the allocated memory for performance reasons when IoMmu is enabled. That way we don't have to do the double buffer re-mapping case as seen in https://github.com/microsoft/mu_silicon_intel_tiano/blob/dev/202502/IntelSiliconPkg/Feature/VTd/IntelVTdCoreDxe/BmDma.c#L189.

ardbiesheuvel:
> 
> Thank you for the feedback, I'll go ahead and change this to just use standard page allocations. I would prefer that all DMA buffers at-least use aligned pages for the allocated memory for performance reasons when IoMmu is enabled. That way we don't have to do the double buffer re-mapping case as seen in https://github.com/microsoft/mu_silicon_intel_tiano/blob/dev/202502/IntelSiliconPkg/Feature/VTd/IntelVTdCoreDxe/BmDma.c#L189.

I don't see how that is going to help. Even if you allocate at page granularity, the size of the _mapping_ comes from the caller, and this is often a much smaller value (~8 bytes for polling a USB keyboard EP, or ~1500 bytes for a network packet), and so you will always hit the remap case, even when using page allocations.

So you would have to change all IOMMU drivers to round up `*NumberOfBytes` to page granularity if you want to avoid taking this path. 

Why does a misaligned buffer require a remap to begin with?

eeshanl:
> > Thank you for the feedback, I'll go ahead and change this to just use standard page allocations. I would prefer that all DMA buffers at-least use aligned pages for the allocated memory for performance reasons when IoMmu is enabled. That way we don't have to do the double buffer re-mapping case as seen in https://github.com/microsoft/mu_silicon_intel_tiano/blob/dev/202502/IntelSiliconPkg/Feature/VTd/IntelVTdCoreDxe/BmDma.c#L189.
> 
> I don't see how that is going to help. Even if you allocate at page granularity, the size of the _mapping_ comes from the caller, and this is often a much smaller value (~8 bytes for polling a USB keyboard EP, or ~1500 bytes for a network packet), and so you will always hit the remap case, even when using page allocations.
> 
> So you would have to change all IOMMU drivers to round up `*NumberOfBytes` to page granularity if you want to avoid taking this path.
> 
> Why does a misaligned buffer require a remap to begin with?

The iommu/smmu can only do page wide mappings. Atleast with smmu, when you write the physical address to the leaf level page table, you drop the bottom 12 bits. And it maps that entire page till the end of the page boundary. Which is why as you can see in the vtd driver, it will check for address and size to be aligned and copy the contents into a new buffer so that it does not over-map a region where you could be mapping a region outside of the allocated DMA buffer. Hence why I would like all dma buffers to be page allocated. That way the iommu driver could still round up the *NumberOfBytes like you mention.

PR #11662, opened 2025-10-28
IntelFsp2Pkg: Add check if current OS support tkinter or not

hongbin123:
- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?


niruiyu (approved):

ChaselChiu:
It would be good if we still could show dailog for warning message when tkinter available.

niruiyu:
> It would be good if we still could show dailog for warning message when tkinter available.

@ChaselChiu It's not a good practice to pop up a dialog from a command-line tool.

ChaselChiu:
> > It would be good if we still could show dailog for warning message when tkinter available.
> 
> @ChaselChiu It's not a good practice to pop up a dialog from a command-line tool.

Same script is consumed by ConfigEditor.py which is a GUI tool 

IMO, the script was implemented in a bad style from the beginning.
The UI/logic code is mixed together and any function can pop up dialogs.

I do no think we need to refactor the code for now.
I agree to pop up a dialog if tkinter is available.

AshrafAliS (commented):
tkinter is being used in this file.

image

hongbin123:
> tkinter is being used in this file.
> 
> image

yes, modify code to check if tkinter existed, it should be better now.

hongbin123:
> > > It would be good if we still could show dailog for warning message when tkinter available.
> > 
> > 
> > @ChaselChiu It's not a good practice to pop up a dialog from a command-line tool.
> 
> Same script is consumed by ConfigEditor.py which is a GUI tool
> 
> IMO, the script was implemented in a bad style from the beginning. The UI/logic code is mixed together and any function can pop up dialogs.
> 
> I do no think we need to refactor the code for now. I agree to pop up a dialog if tkinter is available.

Add the tkinter checking code.

IntelFsp2Pkg/Tools/ConfigEditor/GenYamlCfg.py
```diff
@@ -639,6 +637,11 @@ def __init__(self):
         self._macro_dict = {}
         self.binseg_dict = {}
         self.initialize()
+        if self.tkinter_supported():
+            import tkinter.messagebox as messagebox
+            self.messagebox = messagebox
+            import tkinter
+            self._tk = tkinter
```
niruiyu:
```python
self._tk = self.import_tkinter

def import_tkinter(self):
  try:
    import tkinter
    import tkinter.messagebox
    return tkinter
  except ImportError:
    print(...)
    return None


if self._tk:
  self._tk.messagebox(...)
else:
  print(...)
```

niruiyu (approved):

IntelFsp2Pkg/Tools/ConfigEditor/GenYamlCfg.py
hongbin123: updated

PR #11666, opened 2025-10-28
ArmPlatformPkg: Update transfer list register usage before stack setup

PrachotanReddy:
Previously register x6 was used to temporarily hold the transfer list address before the stack was setup and the jump to C code. This however is not working for RPi3, platform code uses w6 which aliases x6 causing overwrites.
Instead we now use TPIDRRO_EL0  to temporarily hold the value.
This is inspired by PEI using TPIDR_EL0 to store the PEI services table pointer.

# Description

Use TPIDRRO_EL0 instead of x6 to temporarily hold the transfer list address before the stack is setup.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on RPi3

## Integration Instructions

N/A

### DIFF
--- a/ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
+++ b/ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
@@ -20,8 +20,8 @@ ASM_FUNC(_ModuleEntryPoint)
   // Skip TransferList init if x2 is not equal to 0
   cbnz  x2, _SkipTransferList
 
-  // Set the TransferList Base Address from register x3
-  mov   x6, x3
+  // Set the TransferList Base Address from register x3 to TPIDRRO_EL0
+  msr   TPIDRRO_EL0, x3
 
 _SkipTransferList:
   // Do early platform specific actions
@@ -93,7 +93,7 @@ _GetStackBase:
   sub   x1, x1, x2
 
   // Pass Transfer List Base Address
-  mov   x2, x6
+  mrs   x2, TPIDRRO_EL0
   // Move sec startup address into a data register
   // Ensure we're jumping to FV version of the code (not boot remapped alias)
   ldr   x4, =ASM_PFX(CEntryPoint)


vbpandya:
Could you please point me to the code in RPi3 which is using x6 ?
And is it too hard for RPi3 to use any other register than x6 ?

PrachotanReddy:
> Could you please point me to the code in RPi3 which is using x6 ?

Please see: https://github.com/tianocore/edk2-platforms/blob/a244c956e8be460ee904948bf7b3ea6a7ddad61d/Platform/RaspberryPi/Library/PlatformLib/AArch64/RaspberryPiHelper.S#L17

> And is it too hard for RPi3 to use any other register than x6 ?

Based on a discussion with Leif and Ard, using a parameter/result register (x19-x28) was recommended. We could change RPi3 to not use x6 but still need to standardize the usage of x6 so all platforms comply. But if we are trying to standardize a register it might be worth to just use a parameter/result register. See: https://edk2.groups.io/g/devel/topic/115317898#msg121608

leiflindholm:
So, it's worth clarifying a few things - in PR description and commit message.
- The x6 use is in fact w6, which aliases x6.
- The use is in ArmPlatformPeiBootAction in PlatformLib (which presumably means RPi4 is also affected?)
- The change isn't "use x20 instead of x6, because that happens to work right now". The change is "use a register that is callee-saved according to the PCS, even though we don't even have a stack here, because that's _some_ sane rule to follow".

ardbiesheuvel:
> * The change isn't "use x20 instead of x6, because that happens to work right now". The change is "use a register that is callee-saved according to the PCS, even though we don't even have a stack here, because that's _some_ sane rule to follow".

We might even consider using `TPIDRRO_EL0` - there is prior art here, as both ~~PEI and~~ PeilessSec use `TPIDR_EL0` to store the HOB list pointer. (edit: PEI uses it to store the PEI services table pointer)

PrachotanReddy:
From the docs:

> TPIDR_EL0, EL0 Read/Write Software Thread ID Register
> Provides a location where software executing at EL0 can store thread identifying information, for OS management purposes.

Based on this description, is it still valid to use it to hold information like the Transfer List Base address?
Additionally since TPIDR_EL0 is already being used, is there another register that can serve the purpose?

ardbiesheuvel:
> From the docs:
> 
> > TPIDR_EL0, EL0 Read/Write Software Thread ID Register
> > Provides a location where software executing at EL0 can store thread identifying information, for OS management purposes.
> 
> Based on this description, is it still valid to use it to hold information like the Transfer List Base address? Additionally since TPIDR_EL0 is already being used, is there another register that can serve the purpose?

`TPIDRRO_EL0` not `TPIDR_EL0`

PrachotanReddy:
I've updated the fix to use TPIDRRO_EL0 instead. But is TPIDRRO_EL0 callee-saved? Wouldn't we be in the same place if another platform decides to use TPIDRRO_EL0 to store something else?

ardbiesheuvel:
> I've updated the fix to use TPIDRRO_EL0 instead. But is TPIDRRO_EL0 callee-saved? Wouldn't we be in the same place if another platform decides to use TPIDRRO_EL0 to store something else?

The problem is that this code runs before there is a stack, and so `callee-saved` is meaningless - the callee cannot be expected to save anything.

`TPIDRRO_EL0` is much less likely to be in use already than any of the GPRs.

PR #11667, opened 2025-10-29
 MdePkg: Introduce EFI_DT_FIXUP_PROTOCOL protocol

dimorinny:
# Description

Adds the EFI_DT_FIXUP_PROTOCOL header. This protocol allows firmware to apply boot-time fixups and process memory reservations from the device tree. It is expected to be invoked by the boot manager before passing the updated device tree to the OS.

This protocol is used by:

1. Android Generic Bootloader (GBL)
2. Systemd-boot, Systemd-stub
3. Ubuntu's grub2

Original proposal:
https://github.com/U-Boot-EFI/EFI_DT_FIXUP_PROTOCOL

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Used by GBL on top of QCOM EDK2-based FW.

## Integration Instructions

N/A

### DIFF
--- a/MdePkg/Include/Protocol/DtFixup.h
+++ b/MdePkg/Include/Protocol/DtFixup.h
@@ -1,0 +1,50 @@
+/** @file
+
+  Copyright (c) 2025, Heinrich Schuchardt.
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+/*
+  DT Fix-up Protocol.
+  Lets firmware patch a flattened-device-tree and reserve memory ranges.
+
+  Related docs:
+  https://github.com/U-Boot-EFI/EFI_DT_FIXUP_PROTOCOL
+*/
+
+#ifndef DT_FIXUP_PROTOCOL_H_
+#define DT_FIXUP_PROTOCOL_H_
+
+#include <Uefi/UefiBaseType.h>
+
+//
+// {e617d64c-fe08-46da-f4dc-bbd5870c7300}
+//
+#define DT_FIXUP_PROTOCOL_GUID \
+  { 0xe617d64c, 0xfe08, 0x46da, { 0xf4, 0xdc, 0xbb, 0xd5, 0x87, 0x0c, 0x73, 0x00 } }
+
+#define DT_FIXUP_PROTOCOL_REVISION  0x00010000
+
+typedef struct _DT_FIXUP_PROTOCOL DT_FIXUP_PROTOCOL;
+
+#define DT_FIXUP_APPLY_FIXUPS    0x00000001
+#define DT_FIXUP_RESERVE_MEMORY  0x00000002
+#define DT_FIXUP_ALL             (DT_FIXUP_APPLY_FIXUPS | DT_FIXUP_RESERVE_MEMORY)
+
+typedef
+EFI_STATUS
+(EFIAPI *DT_FIXUP)(
+  IN DT_FIXUP_PROTOCOL *This,
+  IN OUT VOID          *Fdt,
+  IN OUT UINTN         *BufferSize,
+  IN UINT32            Flags
+  );
+
+struct _DT_FIXUP_PROTOCOL {
+  UINT64      Revision;
+  DT_FIXUP    Fixup;
+};
+
+#endif // DT_FIXUP_PROTOCOL_H_


ardbiesheuvel (commented):
The change looks reasonable in terms of idiom and coding style, but I'd like to understand how this is going to be used. Typically, it is the firmware that authoritatively describes the platform to the OS, not the other way around. Also, this protocol does two wildly different things in a single call
- modify a DT description in memory according to the firmware's view of the platform
- modify the firmware's view of the platform's memory map based on a DT description.

Especially the latter is problematic imo: we have plenty of APIs for doing memory allocation and reservation, and doing a batch update of the EFI memory map (which is internally derived from the GCD memory map) based on a DT blob seems like a fragile proposition to me.

If no providers or users of this protocol are going to be contributed, perhaps we don't need this definition either. But if we do, and given the fact that this protocol [sadly] appears to be a de-facto industry standard, it would belong in MdePkg rather than EmbeddedPkg

@mdkinney @lgao4

lgao4:
> 3\. Ubuntu's grub2

Do you mean Linux Grub may also consume it? If so, this protocol should be the standard definition.

dimorinny:
> Typically, it is the firmware that authoritatively describes the platform to the OS, not the other way around.

Yes, I'm agree that this protocol turns the DT logic around, but this could be useful for cases where boot app is responsible for loading / verifying the HLOS images so owns them. In this case FW is the one to adjust the DT with board-specific details before it's getting handed off to HLOS via boot app. That's exactly what we [do](https://cs.android.com/android/kernel/superproject/+/common-android-mainline:bootable/libbootloader/gbl/docs/efi_protocols.md;l=183) in [GBL](https://source.android.com/docs/core/architecture/bootloader/generic-bootloader) which is Google-provided Android boot UEFI app.

> Especially the latter is problematic imo: we have plenty of APIs for doing memory allocation and reservation, and doing a batch update of the EFI memory map (which is internally derived from the GCD memory map) based on a DT blob seems like a fragile proposition to me.

It's hard for me to speculate about this one since GBL doesn't use this feature, @xypron do you have any details to share to address this point?

> If no providers or users of this protocol are going to be contributed, perhaps we don't need this definition either.

This protocol is going to be used by EDK2-based Android ODMs and having it upstreamed (which is already done for U-boot, LK) could be extremely useful from OSS-compatibility and further standardization perspective. CC @leiflindholm

> it would belong in MdePkg rather than EmbeddedPkg
> Do you mean Linux Grub may also consume it? If so, this protocol should be the standard definition.

Ubuntu's grub2 [carries](https://git.launchpad.net/ubuntu/+source/grub2/tree/debian/patches/fdt-device-tree-fixup-protocol.patch) it as build-time patch. It's not used by upstream grub2 though. Moved it into `MdePkg`, thanks for the pointer.

xypron:
> > Typically, it is the firmware that authoritatively describes the platform to the OS, not the other way around.

This does not hold true for device trees for ARM and RISC-V hardware.

Device-trees for ARM and RISC-V hardware are developed inside the Linux kernel. They are not a stable interface. Instead the device-tree for new hardware develops over months and years. Firmware like U-Boot follows the development in Linux.

With a dated device-tree from firmware you may only get partial hardware support.

> 
> Yes, I'm agree that this protocol turns the DT logic around, but this could be useful for cases where boot app is responsible for loading / verifying the HLOS images so owns them. In this case FW is the one to adjust the DT with board-specific details before it's getting handed off to HLOS via boot app. That's exactly what we [do](https://cs.android.com/android/kernel/superproject/+/common-android-mainline:bootable/libbootloader/gbl/docs/efi_protocols.md;l=183) in [GBL](https://source.android.com/docs/core/architecture/bootloader/generic-bootloader) which is Google-provided Android boot UEFI app.
> 
> > Especially the latter is problematic imo: we have plenty of APIs for doing memory allocation and reservation, and doing a batch update of the EFI memory map (which is internally derived from the GCD memory map) based on a DT blob seems like a fragile proposition to me.
> 
> It's hard for me to speculate about this one since GBL doesn't use this feature, @xypron do you have any details to share to address this point?

Some reserved memory regions depend on the operating system and on the use case. But they may still defined via the Linux device-tree. These include

* CMA pools for continuous memory allocation
* Shared video memory

There have been cases were the list of reserved memory regions supplied by firmware was incomplete. E.g. the memory used by OpenSBI was not marked as reserved though that software is needed at runtime.

U-Boot does the following:

All regions listed in the Memory Reservation Block are added to the memory map as EFI_RESERVED_MEMORY_TYPE.

For all children of the /reserved-memory/ node, if the memory block has the no-map flag, it is added to the memory map as EFI_RESERVED_MEMORY_TYPE. Otherwise it is added to the memory map as EFI_BOOT_SERVICES_DATA.

A conflict in the memory management might arise if reserved memory has been previously allocated with AllocatePages() before the invocation of the protocol and is freed after the reservation. This could result in the memory being converted to EfiConventionalMemory or the FreePages() call failing with EFI_NOT_FOUND depending on the implementation.

> Ubuntu's grub2 [carries](https://git.launchpad.net/ubuntu/+source/grub2/tree/debian/patches/fdt-device-tree-fixup-protocol.patch) it as build-time patch. It's not used by upstream grub2 though. Moved it into `MdePkg`, thanks for the pointer.

A list of known implementations and consumers of the EFI_DT_FIXUP_PROTOCOL is provided in
https://github.com/U-Boot-EFI/EFI_DT_FIXUP_PROTOCOL/blob/master/references.rst

PR #11668, opened 2025-10-29
MdeModulePkg/Core/Dxe: Fix TPL inversion from DEBUG() message

mdkinney:
# Description

PR #11443 introduced a regression by adding a DEBUG() message when the lock for events is acquired and that lock is at TPL_HIGH_LEVEL. If DEBUG() messages are routed through Report Status Code, and the Report Status Code Protocol has not been located yet, then a call to gBS->LocateProtocol() is made and that call raises TPL to TPL_NOTIFY which causes a TPL inversion.

The event lock is used to atomically update gEventSignalQueue. There is no need for the DEBUG() message to within the event lock scope.

The fix is to scope the event lock to only the InsertHeadList() call to update gEventSignalQueue.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Observed platform that maps debug messages to report status code with recursion of DEBUG() messages from:
```
DEBUG ((DEBUG_ERROR, "FATAL ERROR - RaiseTpl with OldTpl(0x%x) > NewTpl(0x%x)\n", OldTpl, NewTpl));
```

After this change, the platform boots normally.

## Integration Instructions

N/A

### DIFF
--- a/NetworkPkg/SnpDxe/Get_status.c
+++ b/NetworkPkg/SnpDxe/Get_status.c
@@ -38,49 +38,54 @@ PxeGetStatus (
   UINT32             Index;
   UINT64             *Tmp;
 
-  Tmp             = NULL;
-  Db              = Snp->Db;
-  Snp->Cdb.OpCode = PXE_OPCODE_GET_STATUS;
+  if (Snp->Cdb == NULL) {
+    DEBUG ((DEBUG_ERROR, "%a: Snp->Cdb is NULL\n", __func__));
+    return EFI_DEVICE_ERROR;
+  }
 
-  Snp->Cdb.OpFlags = 0;
+  Tmp              = NULL;
+  Db               = Snp->Db;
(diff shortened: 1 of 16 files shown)


mdkinney:
@makubacki  @leiflindholm @ardbiesheuvel Need a quick review of this PR that fixes a regression in DXE Core introduced by PR #11443

mdkinney:
@philnoh2 Please review this fix.

ardbiesheuvel (approved):

makubacki (approved):

philnoh2:
> @philnoh2 Please review this fix.

I was out of the office yesterday. Please understand delaying my review. Firstly I confirmed again that our platform doesn't reproduce the issue even on the latest EDK2 master codebase. But the update makes sense to me. Also I think that basically moving CoreAcquireEventLock/CoreReleaseEventLock to the if condition looks better. Thank you for your evaluation and quick update.

PR #11671, opened 2025-10-29
DxeCore: Adding check for underflow/overflow when iterating through resource descriptor hobs

kuqin12:
# Description

During DXE core memory service initialization, the system would check available resource descriptor hobs against the memory top from PHIT hob.

However, it is possible that a given resource descriptor hob will not be larger than the cover the memory top, causing the Length calculation to underflow. Or the hob is not properly prepared and cause the calculation to overflow.

This change adds a check for potential underflow/overflow before performing the subtraction.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested on QEMU platform and booted to Windows desktop.

## Integration Instructions

N/A


lgao4 (approved):

lgao4:
I think this change is good.

PR #11677, opened 2025-10-30
MdeModulePkg: Add gEdkiiNonDiscoverableDeviceUniqueIdProtocolGuid for deterministic UniqueId support in NonDiscoverablePciDeviceDxe

eeshanl:
# Description

MdeModulePkg: Add gEdkiiNonDiscoverableDeviceUniqueIdProtocolGuid for deterministic UniqueId support in NonDiscoverablePciDeviceDxe

With this new protocol, a platform can register a UniqueId to be used to identify the PciIo location.
For example, if a device registers with the
NonDiscoverableDeviceRegistrationLib, they can also publish this new protocol on the same EFI_HANDLE as RegisterNonDiscoverableMmioDevice. NonDiscoverablePciDeviceDxe will consume this UniqueId if found and the device's UniqueId to this value.
If not found, it will back up to the prior method of assigning the UniqueId via a static counter.
The platform must ensure all UniqueIds assigned are unique across the firmware.

Platforms may need to define their own UniqueIds through this method because the current UniqueId that gets used in
NonDiscoverablePciDeviceDxe is a static counter, which can lead to non-deterministic value's being returned by PciIo->GetLocation(). In order to have a deterministic UniqueId returned by PciIo->GetLocation(), this protocol can be used to specify a specific UniqueId for a given Handle.
For example, if a platform needs to Locate a specific NonDiscoverable PciIo protocol, they could use the determinstic UniqueId assigned through this protocol to find the expected PciIo instance. This protocol allows us to have a 1-1 defined mapping between a platform defined UniqueId and the NonDiscoverable PciIo protocol instance.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on physical platform

## Integration Instructions

When assigning a unique ID via this protocol, the platform should ensure that the ID is unique across all instances of this protocol across the platform.
The UniqueId must be in the range of:
MAX_NON_DISCOVERABLE_PCI_DEVICE_ID/2 <= UniqueId < MAX_NON_DISCOVERABLE_PCI_DEVICE_ID

Usage example:
```
EFI_HANDLE Handle = NULL;
UINTN      Size; // NonDiscoverableDevice Size
VOID       *Base; // NonDiscoverableDevice Base

Status = RegisterNonDiscoverableMmioDevice (
            NonDiscoverableDeviceTypeAhci,
            NonDiscoverableDeviceDmaTypeCoherent,
            NULL,
            &Handle,
            1,
            Base,
            Size
          );

NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL * UniqueIdProtocol = (NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL*) AllocateZeroPool (sizeof (NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL));
if (UniqueIdProtocol == NULL) {
  return EFI_OUT_OF_RESOURCES;
}
UniqueIdProtocol->Revision = NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_REVISION;
UniqueIdProtocol->UniqueId = 0x1000; // MAX_NON_DISCOVERABLE_PCI_DEVICE_ID/2 <= UniqueId < MAX_NON_DISCOVERABLE_PCI_DEVICE_ID

Status = gBS->InstallMultipleProtocolInterfaces (&Handle, &gEdkiiNonDiscoverableDeviceUniqueIdProtocolGuid, UniqueIdProtocol, NULL);
```

### DIFF
--- a/MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
+++ b/MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
@@ -9,8 +9,7 @@
 #include "NonDiscoverablePciDeviceIo.h"
 
 #include <Protocol/DriverBinding.h>
-
-#define MAX_NON_DISCOVERABLE_PCI_DEVICE_ID  (32 * 256)
+#include <Protocol/NonDiscoverableDeviceUniqueId.h>
 
 STATIC UINTN           mUniqueIdCounter = 0;
 EFI_CPU_ARCH_PROTOCOL  *mCpu;
@@ -145,11 +144,12 @@ NonDiscoverablePciDeviceStart (
   IN EFI_DEVICE_PATH_PROTOCOL     *RemainingDevicePath
   )
 {
-  NON_DISCOVERABLE_PCI_DEVICE  *Dev;
-  EFI_STATUS                   Status;
+  NON_DISCOVERABLE_PCI_DEVICE                 *Dev;
+  EFI_STATUS                                  Status;
+  NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL  *UniqueIdProtocol;
 
-  ASSERT (mUniqueIdCounter < MAX_NON_DISCOVERABLE_PCI_DEVICE_ID);
-  if (mUniqueIdCounter >= MAX_NON_DISCOVERABLE_PCI_DEVICE_ID) {
+  ASSERT (mUniqueIdCounter < (MAX_NON_DISCOVERABLE_PCI_DEVICE_ID / 2));
+  if (mUniqueIdCounter >= (MAX_NON_DISCOVERABLE_PCI_DEVICE_ID / 2)) {
     return EFI_OUT_OF_RESOURCES;
   }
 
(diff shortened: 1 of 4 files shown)


lgao4 (approved):

eeshanl:
@ardbiesheuvel please take a look, thanks

ardbiesheuvel:
Please explain why deterministic B/D/F assignments are important, given that this PCI emulation only covers the boot phase, and the OS should not care about any of this.

Also, do we really need the complexity of this protocol? Currently, the B/D/F assignments are entirely arbitrary, and so we can simply replace the static counter with a better scheme, without the need for discovery etc.

Would it help if we used a different fixed bus ID for each supported device type (ehci, xhci, ahci, etc), and permit the platform to provide an index when registering the device, which can be used to derive device and function number?

eeshanl:
> Please explain why deterministic B/D/F assignments are important, given that this PCI emulation only covers the boot phase, and the OS should not care about any of this.
> 
> Also, do we really need the complexity of this protocol? Currently, the B/D/F assignments are entirely arbitrary, and so we can simply replace the static counter with a better scheme, without the need for discovery etc.
> 
> Would it help if we used a different fixed bus ID for each supported device type (ehci, xhci, ahci, etc), and permit the platform to provide an index when registering the device, which can be used to derive device and function number?

Hi @ardbiesheuvel , 
We have a use case where we want to identify a pre-defined/known instance of PciIo, whether that is the non-discoverable kind or not, to find/identify an instance via the GetLocation() call. As a result, we wanted to make the GetLocation() call deterministic.

We wanted to add in this opt-in protocol approach because unless the platform requires the GetLocation() B/D/F to be deterministic, they don't need to care about defining what that should be. Therefore, we added this protocol so that platforms can have an opt-in way of registering a deterministic B/D/F for GetLocation().

ardbiesheuvel:
> > Please explain why deterministic B/D/F assignments are important, given that this PCI emulation only covers the boot phase, and the OS should not care about any of this.
> > Also, do we really need the complexity of this protocol? Currently, the B/D/F assignments are entirely arbitrary, and so we can simply replace the static counter with a better scheme, without the need for discovery etc.
> > Would it help if we used a different fixed bus ID for each supported device type (ehci, xhci, ahci, etc), and permit the platform to provide an index when registering the device, which can be used to derive device and function number?
> 
> Hi @ardbiesheuvel , We have a use case where we want to identify a pre-defined/known instance of PciIo, whether that is the non-discoverable kind or not, to find/identify an instance via the GetLocation() call. As a result, we wanted to make the GetLocation() call deterministic.
> 
> We wanted to add in this opt-in protocol approach because unless the platform requires the GetLocation() B/D/F to be deterministic, they don't need to care about defining what that should be. Therefore, we added this protocol so that platforms can have an opt-in way of registering a deterministic B/D/F for GetLocation().

I'd prefer it if we can make it deterministic without complicating things by introducing this protocol.

Issue #11445, opened 2025-08-27
[Bug]: Potential Infinite Loop in DumpCapsuleStatusVariable() due to Unbounded Error Handling of EFI_ERROR


AnishCherekar:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [ ] Firmware
- [x] Tool
- [ ] Unit Test

### What packages are impacted?

MdeModulePkg

### Which targets are impacted by this bug?

DEBUG, RELEASE

### Current Behavior

In the [edk2](https://github.com/tianocore/edk2/tree/master)/[MdeModulePkg](https://github.com/tianocore/edk2/tree/master/MdeModulePkg)/[Application](https://github.com/tianocore/edk2/tree/master/MdeModulePkg/Application)/[CapsuleApp](https://github.com/tianocore/edk2/tree/master/MdeModulePkg/Application/CapsuleApp)
/CapsuleDump.c file .

Function Name : **DumpCapsuleStatusVariable** 

The function  iterates over capsule status variables using a loop indexed by . It constructs variable names like ,Capsule0000 ,Capsule0001 etc, and attempts to retrieve them using GetVariable2 .
However, there's a critical flaw in the while loop:

    **while (TRUE) {**
    UnicodeSPrint (TempVarName, 5 * sizeof (CHAR16), L"%04x", Index);

    Status = GetVariable2 (
               CapsuleVarName,
               &gEfiCapsuleReportGuid,
               (VOID **)&CapsuleResult,
               NULL
               );
    **if (Status == EFI_NOT_FOUND) {
      break;
    } else if (EFI_ERROR (Status)) {
      continue;
    }**

    ASSERT (CapsuleResult != NULL);

    //
    // display capsule process status
    //
    if (CapsuleResult->VariableTotalSize >= sizeof (EFI_CAPSULE_RESULT_VARIABLE_HEADER)) {
      Print (L"CapsuleName: %s\n", CapsuleVarName);
      Print (L"  Capsule Guid: %g\n", &CapsuleResult->CapsuleGuid);
      Print (L"  Capsule ProcessedTime: %t\n", &CapsuleResult->CapsuleProcessed);
      Print (L"  Capsule Status: %r\n", CapsuleResult->CapsuleStatus);
    }

    if (CompareGuid (&CapsuleResult->CapsuleGuid, &gEfiFmpCapsuleGuid)) {
[...]

PR #11681, opened 2025-10-31
MdeModulePkg/CapsuleApp: Fix infinite loop in DumpCapsuleStatusVariable

imsalmankhalid:
The function DumpCapsuleStatusVariable() had a critical flaw where if GetVariable2() returned any EFI_ERROR other than EFI_NOT_FOUND, the loop would continue without incrementing the Index variable. This caused the loop to repeatedly attempt to retrieve the same variable name (e.g., Capsule0000) indefinitely, leading to an infinite loop.

The fix moves the Index increment to occur immediately after checking for EFI_NOT_FOUND but before checking for other errors. This ensures:
- When EFI_NOT_FOUND is returned, the loop terminates as expected
- When any other error occurs, Index increments to move forward
- The loop always makes forward progress
- The upper bound check (Index > 0xFFFF) is maintained

Fixes: https://github.com/tianocore/edk2/issues/11445

Cc: Zhichao Gao <zhichao.gao@intel.com>

# Description

This PR fixes a potential infinite loop in the DumpCapsuleStatusVariable() function that occurs when GetVariable2() returns an error other than EFI_NOT_FOUND. The bug prevents the index from incrementing, causing the same variable name to be queried repeatedly.

- [ ] Breaking change?
- [ ] Impacts security?
- [x] Includes tests?
  - Code passes PatchCheck.py validation
  - Tested with QEMU firmware environment
  
## How This Was Tested

- **Static Analysis**: Verified commit format passes PatchCheck.py
- **Code Review**: Manual inspection of the loop control flow to ensure proper error handling
- **QEMU Testing**: 
  - Built OVMF firmware with the fix
  - Booted QEMU with OVMF: `qemu-system-x86_64 -bios OVMF.fd -m 2048`
  - Launched CapsuleApp from UEFI shell to dump capsule status variables
  - Simulated error conditions by testing with corrupted/missing capsule variables
  - Verified the function properly iterates through capsule indices without hanging
  - Confirmed loop terminates correctly when reaching EFI_NOT_FOUND or Index > 0xFFFF
- **Regression Testing**: Verified existing capsule functionality remains unchanged

## Integration Instructions

N/A - This is a straightforward bug fix with no integration requirements. The change is localized to the DumpCapsuleStatusVariable() function in CapsuleApp.


lgao4:
@imsalmankhalid , please update this PR to include only one commit.

AshrafAliS (approved):

PR #11682, opened 2025-10-31
BaseTools/GenFw: Add --no-debug to skip debug information generation and enable for Release builds

zihanqi-amzn:
Add support for skipping PE/COFF debug directory generation when converting ELF binaries to PE/COFF format. This allows reducing binary size in Release builds by omitting debug information.
    
A new --no-debug option is added to GenFw. When specified, debug space allocation is skipped in ScanSections32() and ScanSections64(), and debug directory and CodeView entries are not written to the output PE/COFF file.
    
The --no-debug flag can only be used with the -e (ELF) conversion option.
    
Debug builds remain unchanged and continue to include debug information. Release builds now automatically exclude debug information for smaller binaries.

# Description

This change adds a new command line option to GenFw that allows skipping debug information generation in PE/COFF binaries. This is useful for Release builds where debug information is not needed, reducing the final binary size.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Built GenFw with the changes and verified that the --no-debug flag correctly skips debug directory generation when specified.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Source/C/GenFw/GenFw.c
+++ b/BaseTools/Source/C/GenFw/GenFw.c
@@ -89,6 +89,7 @@ UINT32 mOutImageType = FW_DUMMY_IMAGE;
 BOOLEAN mIsConvertXip = FALSE;
 BOOLEAN mExportFlag = FALSE;
 BOOLEAN mNoNxCompat = FALSE;
+BOOLEAN mSkipDebugInfo = FALSE;
 
 STATIC
 EFI_STATUS
@@ -251,6 +252,9 @@ Returns:
                         to MINOR.\n");
   fprintf (stdout, "  --keepzeropending     Don't strip zero pending of .reloc.\n\
                         This option can be used together with -e or -t.\n\
+                        It doesn't work for other options.\n");
+  fprintf (stdout, "  --no-debug            Skip debug information generation.\n\
+                        This option can be used together with -e.\n\
                         It doesn't work for other options.\n");
   fprintf (stdout, "  -r, --replace         Overwrite the input file with the output content.\n\
                         If more input files are specified,\n\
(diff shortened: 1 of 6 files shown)


BaseTools/Source/C/GenFw/GenFw.c
```diff
@@ -252,6 +253,9 @@ Routine Description:
   fprintf (stdout, "  --keepzeropending     Don't strip zero pending of .reloc.\n\
                         This option can be used together with -e or -t.\n\
                         It doesn't work for other options.\n");
+  fprintf (stdout, "  --no-debug            Skip debug information generation.\n\
+                        This option can be used together with -e or -t.\n\
+                        It doesn't work for other options.\n");
```
lgao4, line 258: this option is only used when convert elf to efi image. It doesn't work on VS tool chain.

zihanqi-amzn: True. Will update to -e only.

mdkinney:
There are many tools that can operate on elf images to remove debug sections.  Why is this feature being added to GenFW?

zihanqi-amzn:
> There are many tools that can operate on elf images to remove debug sections. Why is this feature being added to GenFW?

@mdkinney Overhead and dependency created if first generated and then removed replying on non-EDK2 toolkit.

mdkinney:
> > There are many tools that can operate on elf images to remove debug sections. Why is this feature being added to GenFW?
> 
> @mdkinney Overhead and dependency created if first generated and then removed replying on non-EDK2 toolkit.

Not sure I follow.  Can you provide more details?

Another option is to add -g0 to CC_FLAGS to prevent generation of debug sections.

zihanqi-amzn:
>Can you provide more details?

For example, using external tools like `strip` would require an additional build step and dependency on binutils, whereas `--no-debug` keeps everything within EDK2 and avoids generating debug info that would just be stripped later. 

This is intended for release builds only. Debug builds can still retain full debug information, which aligns with typical release/debug build concepts.

>Another option is to add -g0 to CC_FLAGS to prevent generation of debug sections.

`-g0` tells the compiler to generate no debug information, but GenFw adds its own PE/COFF debug directory during ELF-to-PE/COFF conversion regardless of compiler flags, that's what `--no-debug` now skips.

PR #11314, opened 2025-07-17
MdeModulePkg: Bus,Universal,Core CodeQl Changes.

apop5:
# Description

Last of codeql changes for MdeModulePkg.

With this set of changes, and the changes in the rest of the PRs, codeql for ModeModulePkg
was reporting no failures.

To get to the final state, the introduction of a CodeQlFilter.yml file was needed. This was to
suppress errors coming from the oniguruma submodules, and for 3 false positives 
where codeql was detecting missing null tests, but null was a valid input option.

- [ ] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [ ] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

## How This Was Tested
Local CI.

## Integration Instructions
No Integration necessary.

PR #11686, opened 2025-10-31
MdeModulePkg: SetupBrowerDxe Static analysis changes.

apop5:
# Description
Fixing CodeQl Scanning results for MdeModulePkg/Universal/SetupBrowserDxe.

This is split from #11314 to try to make reviewing the changes simplier. 

- [ ] Breaking change?
- [x] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Local CI.

## Integration Instructions
No Integration necessary.

### DIFF
--- a/DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtPcieLib/SsdtPcieGenerator.c
+++ b/DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtPcieLib/SsdtPcieGenerator.c
@@ -546,7 +546,7 @@ GenerateRootPortDevices (
       goto exit_handler;
     }
 
-    if (RpInfo[Index].Sun != MAX_UINT64) {
+    if (RpInfo[Index].Sun != MAX_UINT32) {
       // ASL: Name (_SUN, <Sun value>)
       Status = AmlCodeGenNameInteger (
                  "_SUN",
--- a/MdeModulePkg/Bus/Ata/AtaAtapiPassThru/AhciMode.c
+++ b/MdeModulePkg/Bus/Ata/AtaAtapiPassThru/AhciMode.c
@@ -622,9 +622,10 @@ AhciWaitDeviceReady (
 
   //
   // According to SATA1.0a spec section 5.2, we need to wait for PxTFD.BSY and PxTFD.DRQ
-  // and PxTFD.ERR to be zero. The maximum wait time is 16s which is defined at ATA spec.
+  // and PxTFD.ERR to be zero. The maximum wait time is configurable and determined by
+  // PCD PcdSataDeviceReadyTimeout.
   //
-  PhyDetectDelay = 16 * 1000;
+  PhyDetectDelay = PcdGet8 (PcdSataDeviceReadyTimeout) * 1000;
   do {
     Offset = EFI_AHCI_PORT_START + Port * EFI_AHCI_PORT_REG_WIDTH + EFI_AHCI_PORT_SERR;
     if (AhciReadReg (PciIo, Offset) != 0) {
--- a/MdeModulePkg/Bus/Ata/AtaAtapiPassThru/AtaAtapiPassThru.inf
+++ b/MdeModulePkg/Bus/Ata/AtaAtapiPassThru/AtaAtapiPassThru.inf
@@ -67,6 +67,7 @@
 [Pcd]
   gEfiMdeModulePkgTokenSpaceGuid.PcdAtaSmartEnable          ## SOMETIMES_CONSUMES
   gEfiMdeModulePkgTokenSpaceGuid.PcdAhciCommandRetryCount   ## SOMETIMES_CONSUMES
+  gEfiMdeModulePkgTokenSpaceGuid.PcdSataDeviceReadyTimeout  ## CONSUMES
 
 # [Event]
 # EVENT_TYPE_PERIODIC_TIMER ## SOMETIMES_CONSUMES
--- a/MdeModulePkg/Core/Dxe/Dispatcher/Dispatcher.c
+++ b/MdeModulePkg/Core/Dxe/Dispatcher/Dispatcher.c
@@ -773,7 +773,10 @@ FvIsBeingProcessed (
   }
 
   KnownHandle = AllocateZeroPool (sizeof (KNOWN_HANDLE));
-  ASSERT (KnownHandle != NULL);
+  if (KnownHandle == NULL) {
+    ASSERT (KnownHandle != NULL);
+    return NULL;
+  }
 
   KnownHandle->Signature = KNOWN_HANDLE_SIGNATURE;
   KnownHandle->Handle    = FvHandle;
@@ -850,6 +853,7 @@ CoreFvToDevicePath (
   @retval EFI_ALREADY_STARTED   The driver has already been started. Only one
                                 DriverName may be active in the system at any one
                                 time.
+  @retval EFI_OUT_OF_RESOURCES  If memory could not be allocated for the DriverEntry.
 
 **/
 EFI_STATUS
@@ -867,7 +871,11 @@ CoreAddToDriverList (
   // NULL or FALSE.
   //
   DriverEntry = AllocateZeroPool (sizeof (EFI_CORE_DRIVER_ENTRY));
-  ASSERT (DriverEntry != NULL);
+  if (DriverEntry == NULL) {
+    ASSERT (DriverEntry != NULL);
+    return EFI_OUT_OF_RESOURCES;
+  }
+
   if (Type == EFI_FV_FILETYPE_FIRMWARE_VOLUME_IMAGE) {
     DriverEntry->IsFvImage = TRUE;
   }
@@ -1050,13 +1058,15 @@ CoreProcessFvImageFile (
       //
       if (gSecurity != NULL) {
         FvFileDevicePath = CoreFvToDevicePath (Fv, FvHandle, FileName);
-        Status           = gSecurity->FileAuthenticationState (
-                                        gSecurity,
-                                        AuthenticationStatus,
-                                        FvFileDevicePath
-                                        );
         if (FvFileDevicePath != NULL) {
+          Status = gSecurity->FileAuthenticationState (
+                                gSecurity,
+                                AuthenticationStatus,
+                                FvFileDevicePath
+                                );
           FreePool (FvFileDevicePath);
+        } else {
+          Status = EFI_OUT_OF_RESOURCES;
         }
 
         if (Status != EFI_SUCCESS) {
--- a/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
+++ b/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
@@ -161,7 +161,10 @@ CoreDumpGcdMemorySpaceMap (
   }
 
   Status = CoreGetMemorySpaceMap (&NumberOfDescriptors, &MemorySpaceMap);
-  ASSERT (Status == EFI_SUCCESS && MemorySpaceMap != NULL);
+  if (!((Status == EFI_SUCCESS) && (MemorySpaceMap != NULL))) {
+    ASSERT ((Status == EFI_SUCCESS) && (MemorySpaceMap != NULL));
+    return;
+  }
 
   if (InitialMap) {
     DEBUG ((DEBUG_GCD, "GCD:Initial GCD Memory Space Map\n"));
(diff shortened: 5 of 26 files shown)


MdeModulePkg/Universal/SetupBrowserDxe/Expression.c
```diff
@@ -2544,7 +2594,8 @@ EvaluateExpression (
   EXPRESSION_OPCODE         *OpCode;
   FORM_BROWSER_STATEMENT    *Question;
   FORM_BROWSER_STATEMENT    *Question2;
-  UINTN                     Index;
+  UINT16                    Index;
```
leiflindholm, line 2597: These aren't NULL-check fixes.

MdeModulePkg/Universal/SetupBrowserDxe/IfrParse.c
```diff
@@ -1551,34 +1640,43 @@ ParseOpCodes (
...
       ASSERT (CurrentExpression != NULL);
-      InsertTailList (&CurrentExpression->OpCodeListHead, &ExpressionOpCode->Link);
-      if (Operand == EFI_IFR_MAP_OP) {
-        //
-        // Store current Map Expression List.
-        //
-        if (MapExpressionList != NULL) {
-          PushMapExpressionList (MapExpressionList);
-        }
+      if (CurrentExpression != NULL) {
```
leiflindholm: Please go through all of these success handling blocks messing up indentation and change them to early return/continue/break/goto. The ones that affect only one or two subsequent lines can stay as is, but anything that ends up indenting existing if-statements or other longer blocks need rewriting.

MdeModulePkg/Universal/SetupBrowserDxe/Expression.c
apop5: Fixed, moved to previous commit.

MdeModulePkg/Universal/SetupBrowserDxe/IfrParse.c
apop5: Updated.

MdeModulePkg/Universal/SetupBrowserDxe/Expression.c
```diff
@@ -3012,8 +3013,8 @@ EvaluateExpression (
           //
           Value->Value.string = NewString (gEmptyString, FormSet->HiiHandle);
         } else {
-          Index               = (UINT16)Value->Value.u64;
-          Value->Value.string = Index;
+          Index2              = (UINTN)Value->Value.u64;
```
leiflindholm: What fresh hell is this? `Index` isn't even used as an index - just a random temporary variable. Gaah!

Is there any reason why the (utterly deranged) recasting of part of a value into a different view of its original location needs a temporary variable in the first place, or can we just drop the use of `Index` here?

MdeModulePkg/Universal/SetupBrowserDxe/Expression.c
```diff
@@ -2711,13 +2712,13 @@ EvaluateExpression (
                     Value->Type = OpCode->ValueType;
                     TempBuffer  = (UINT8 *)&Value->Value;
                     ZeroMem (TempStr, sizeof (TempStr));
-                    for (Index = 0; Index < TempLength; Index++) {
-                      TempStr[0] = StrPtr[TempLength - Index - 1];
+                    for (Index2 = 0; Index2 < TempLength; Index2++) {
```
leiflindholm, line 2766: Apart from the location below, which I have comments on there, this is the only place Index2 is used. Can you make it a local definition for this block (under if (OpCode->ValueWidth...))?

MdeModulePkg/Universal/SetupBrowserDxe/IfrParse.c
```diff
@@ -2735,13 +3076,15 @@ ParseOpCodes (
...
 
     if (IsStatementOpCode (Operand)) {
-      CurrentStatement->ParentStatement = ParentStatement;
-      if (Scope != 0) {
-        //
-        // Scope != 0, other statements or options may nest in this statement.
-        // Update the ParentStatement info.
-        //
-        ParentStatement = CurrentStatement;
+      if (CurrentStatement != NULL) {
```
leiflindholm: Could you fold this test into the `if (IsStatementOpCode (Operand))` one?

MdeModulePkg/Universal/SetupBrowserDxe/Setup.c
```diff
@@ -6598,18 +6762,22 @@ ExecuteAction (
...
+
     FormSet = gCurrentSelection->FormSet;
     Form    = gCurrentSelection->Form;
+  } else {
+    FormSet = NULL;
+    Form    = NULL;
   }
 
+  Status  = EFI_SUCCESS;
+  FormSet = gCurrentSelection->FormSet;
```
leiflindholm: We don't want to override this (and next line) which we just set, right?

MdeModulePkg/Universal/SetupBrowserDxe/Expression.c
apop5: yes, Done.

MdeModulePkg/Universal/SetupBrowserDxe/Setup.c
apop5: good catch, that was just wrong.

MdeModulePkg/Universal/SetupBrowserDxe/IfrParse.c
```diff
@@ -413,7 +451,10 @@ CreateStorage (
   }
 
   if (StorageType != EFI_HII_VARSTORE_NAME_VALUE) {
-    ASSERT (StorageName != NULL);
+    if (StorageName == NULL) {
+      ASSERT (StorageName != NULL);
+      goto ErrorExit;
```
ardbiesheuvel, line 456: `Storage` and `BrowserStorage` are uninitialized here

MdeModulePkg/Universal/SetupBrowserDxe/IfrParse.c
```diff
@@ -423,14 +464,23 @@ CreateStorage (
   }
 
   Storage = AllocateZeroPool (sizeof (FORMSET_STORAGE));
-  ASSERT (Storage != NULL);
+
+  if (Storage == NULL) {
+    ASSERT (Storage != NULL);
+    goto ErrorExit;
```
ardbiesheuvel, line 470: `BrowserStorage` is uninitialized here.

MdeModulePkg/Universal/SetupBrowserDxe/IfrParse.c
```diff
@@ -449,9 +499,34 @@ CreateStorage (
...
+  if (UnicodeString != NULL) {
+    FreePool (UnicodeString);
+  }
+
+  if (BrowserStorage != NULL) {
+    FreePool (BrowserStorage);
+  }
+
+  if (Storage != NULL) {
+    if (Storage->ConfigRequest != NULL) {
```
ardbiesheuvel, line 522: ... and they are referenced here.

MdeModulePkg/Universal/SetupBrowserDxe/IfrParse.c
apop5: @ardbiesheuvel 
Thank you for these reports. Are you determine this though manual examination or through a compiler warning? if it's a compiler, can you share the details? I would like to reproduce this so I can apply it to the other static analysis as well.

Are you okay with another pr to fix these instances?

ardbiesheuvel: > @ardbiesheuvel Thank you for these reports. Are you determine this though manual examination or through a compiler warning? if it's a compiler, can you share the details? I would like to reproduce this so I can apply it to the other static analysis as well.
> 

Your changes broke the CLANGDWARF build.

> Are you okay with another pr to fix these instances?

Yes, please.

apop5: Do you think adding clang to the CI would be a good discussion? If it is able to find additional code errors, it might be worth a discussion...

ardbiesheuvel: Definitely.

MdeModulePkg/Universal/SetupBrowserDxe/Presentation.c
```diff
@@ -1606,8 +1642,7 @@ ProcessUserInput (
   if (UserInput->Action != 0) {
     Status                       = ProcessAction (UserInput->Action, UserInput->DefaultId);
     gCurrentSelection->Statement = NULL;
-  } else {
-    ASSERT (Statement != NULL);
```
mikebeaton, line 1610: We've lost an ASSERT here. We can keep it by doing:
```
} else if (Statement == NULL {
  ASSERT (Statement != NULL);
} else {
```

MdeModulePkg/Universal/SetupBrowserDxe/Expression.c
```diff
@@ -2257,6 +2293,9 @@ CompareHiiValue (
     Buf1Len = GetLengthForValue (Value1);
     Buf2    = GetBufferForValue (Value2);
     Buf2Len = GetLengthForValue (Value2);
+    if ((Buf1 == NULL) || (Buf2 == NULL)) {
+      return EFI_INVALID_PARAMETER;
+    }
```
mikebeaton, line 2298: It would be better to add a separate error return for each of these two conditions separately in the lines just above, as soon as they happen if they do.

MdeModulePkg/Universal/SetupBrowserDxe/Expression.c
```diff
@@ -2490,18 +2534,24 @@ GetQuestionValueFromForm (
...
 
   //
   // Get the question value.
   //
-  Status = GetQuestionValue (FormSet, Form, Question, GetSetValueWithEditBuffer);
-  if (EFI_ERROR (Status)) {
-    GetTheVal = FALSE;
-    goto Done;
-  }
+  if (Question != NULL) {
```
mikebeaton, line 2546: IMO a better style of fix (given the code style in use) is not to guard the code below with an `if`, but rather to follow the pattern above (which, on following it, shows up a very well hidden mistake here too, of not setting `GetTheVal = FALSE` when `Question == NULL`, in your current version of this fix, as it seems you should):

```
  if (Question == NULL ) {
    ASSERT (Question != NULL);
    GetTheVal = FALSE;
    goto Done;
  }
```

MdeModulePkg/Universal/SetupBrowserDxe/Expression.c
```diff
@@ -2708,16 +2758,18 @@ EvaluateExpression (
                   ASSERT (StrPtr != NULL);
                   TempLength = StrLen (StrPtr);
                   if (OpCode->ValueWidth >= ((TempLength + 1) / 2)) {
+                    UINTN  Index2;
```
mikebeaton, line 2761: Why have we changed to `Index2` at all? `Index` is a general purpose local index used at several places throughout this method, and I cannot see that any usage of it that interferes with this one, so it makes little sense to change only this one.

ardbiesheuvel: This PR is closed. If you feel there are worthwhile improvements to be made to the current main branch, please raise a separate PR.

MdeModulePkg/Universal/SetupBrowserDxe/Presentation.c
apop5: adding a return EFI_NOT_FOUND after the recommended assert, for the reason:

ASSERT by itself is not enough to cover cases. if DebugAssertEnabled returns false (`PcdDebugPropertyMask has bit DEBUG_PROPERTY_DEBUG_ASSERT_ENABLED cleared) , code execution will proceed pass the assert

MdeModulePkg/Universal/SetupBrowserDxe/Expression.c
apop5: Index is of type UINT16

Index2 is of type UINTN

TempLength is of type UINTN

Previously, the UINT16 was being compared against the UINTN, which resulted in a codeql Comparison of wider width (https://codeql.github.com/codeql-query-help/cpp/cpp-comparison-with-wider-type)

If the returned TempLength (UINTN) was large enough, the UINT16 would never be able to reach a high enough value to exit the for loop.

Based on PR feedback, it was turned into a Index2 with a local declaration instead of attempting to reuse Index.

MdeModulePkg/Universal/SetupBrowserDxe/Expression.c
mikebeaton: Resolved (I cannot mark it resolved for some reason).

MdeModulePkg/Universal/SetupBrowserDxe/Expression.c
mikebeaton: Resolved (I cannot resolve thread for some reason).

MdeModulePkg/Universal/SetupBrowserDxe/Presentation.c
mikebeaton: Resolved ('Resolve comment' is failing with 'Failed to resolve thread' for me for some reason).

PR #11687, opened 2025-10-31
MdeModulePkg: DisplayEngineDxe Static analysis changes.

apop5:
# Description
Fixing CodeQl Scanning results for MdeModulePkg/Universal/DisplayEngineDxe.

This is split from #11314 to try to make reviewing the changes simpler. 

- [ ] Breaking change?
- [x] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Local CI.

## Integration Instructions
No Integration necessary.

### DIFF
--- a/DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtPcieLib/SsdtPcieGenerator.c
+++ b/DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtPcieLib/SsdtPcieGenerator.c
@@ -546,7 +546,7 @@ GenerateRootPortDevices (
       goto exit_handler;
     }
 
-    if (RpInfo[Index].Sun != MAX_UINT64) {
+    if (RpInfo[Index].Sun != MAX_UINT32) {
       // ASL: Name (_SUN, <Sun value>)
       Status = AmlCodeGenNameInteger (
                  "_SUN",
--- a/MdeModulePkg/Bus/Ata/AtaAtapiPassThru/AhciMode.c
+++ b/MdeModulePkg/Bus/Ata/AtaAtapiPassThru/AhciMode.c
@@ -622,9 +622,10 @@ AhciWaitDeviceReady (
 
   //
   // According to SATA1.0a spec section 5.2, we need to wait for PxTFD.BSY and PxTFD.DRQ
-  // and PxTFD.ERR to be zero. The maximum wait time is 16s which is defined at ATA spec.
+  // and PxTFD.ERR to be zero. The maximum wait time is configurable and determined by
+  // PCD PcdSataDeviceReadyTimeout.
   //
-  PhyDetectDelay = 16 * 1000;
+  PhyDetectDelay = PcdGet8 (PcdSataDeviceReadyTimeout) * 1000;
   do {
     Offset = EFI_AHCI_PORT_START + Port * EFI_AHCI_PORT_REG_WIDTH + EFI_AHCI_PORT_SERR;
     if (AhciReadReg (PciIo, Offset) != 0) {
--- a/MdeModulePkg/Bus/Ata/AtaAtapiPassThru/AtaAtapiPassThru.inf
+++ b/MdeModulePkg/Bus/Ata/AtaAtapiPassThru/AtaAtapiPassThru.inf
@@ -67,6 +67,7 @@
 [Pcd]
   gEfiMdeModulePkgTokenSpaceGuid.PcdAtaSmartEnable          ## SOMETIMES_CONSUMES
   gEfiMdeModulePkgTokenSpaceGuid.PcdAhciCommandRetryCount   ## SOMETIMES_CONSUMES
+  gEfiMdeModulePkgTokenSpaceGuid.PcdSataDeviceReadyTimeout  ## CONSUMES
 
 # [Event]
 # EVENT_TYPE_PERIODIC_TIMER ## SOMETIMES_CONSUMES
--- a/MdeModulePkg/Core/Dxe/Dispatcher/Dispatcher.c
+++ b/MdeModulePkg/Core/Dxe/Dispatcher/Dispatcher.c
@@ -773,7 +773,10 @@ FvIsBeingProcessed (
   }
 
   KnownHandle = AllocateZeroPool (sizeof (KNOWN_HANDLE));
-  ASSERT (KnownHandle != NULL);
+  if (KnownHandle == NULL) {
+    ASSERT (KnownHandle != NULL);
+    return NULL;
+  }
 
   KnownHandle->Signature = KNOWN_HANDLE_SIGNATURE;
   KnownHandle->Handle    = FvHandle;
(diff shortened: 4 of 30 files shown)


MdeModulePkg/Universal/DisplayEngineDxe/FormDisplay.c
```diff
@@ -1178,7 +1181,7 @@ ProcessStringForDateTime (
   //
   // Enable to suppress field in the opcode base on the flag.
   //
-  if (Statement->OpCode->OpCode == EFI_IFR_DATE_OP) {
+  if ((Statement->OpCode->OpCode == EFI_IFR_DATE_OP)) {
```
leiflindholm: Unrelated and wrong.

MdeModulePkg/Universal/DisplayEngineDxe/FormDisplay.c
```diff
@@ -1203,7 +1206,7 @@ ProcessStringForDateTime (
       //
       SetUnicodeMem (&OptionString[0], 4, L' ');
     }
-  } else if (Statement->OpCode->OpCode == EFI_IFR_TIME_OP) {
+  } else if ((Statement->OpCode->OpCode == EFI_IFR_TIME_OP)) {
```
leiflindholm: Unrelated and wrong.

MdeModulePkg/Universal/DisplayEngineDxe/FormDisplay.c
```diff
@@ -1558,7 +1561,9 @@ FindTopOfScreenMenu (
     }
   } else {
     TopOfScreen = Link;
-    *SkipValue  = PreviousMenuOption->Skip - Rows;
+    if (PreviousMenuOption != NULL) {
```
leiflindholm, line 1564: Is it safe to just ignore this?

MdeModulePkg/Universal/DisplayEngineDxe/FormDisplay.c
```diff
@@ -2917,7 +2922,10 @@ UiDisplayMenu (
         if (SkipHighLight) {
           SkipHighLight = FALSE;
           MenuOption    = SavedMenuOption;
-          RefreshKeyHelp (gFormData, SavedMenuOption->ThisTag, FALSE);
+          if (SavedMenuOption != NULL) {
```
leiflindholm, line 2925: Is it safe to just ignore this?

MdeModulePkg/Universal/DisplayEngineDxe/FormDisplay.c
```diff
@@ -3001,26 +3009,28 @@ UiDisplayMenu (
...
-            } else {
-              if (((Statement->OpCode->OpCode == EFI_IFR_DATE_OP) && (StatementValue->Value.date.Month == 0xff)) || ((Statement->OpCode->OpCode == EFI_IFR_TIME_OP) && (StatementValue->Value.time.Hour == 0xff))) {
-                StringRightPtr = GetToken (HelpInfo, gFormData->HiiHandle);
-                StringErrorPtr = GetToken (STRING_TOKEN (GET_TIME_FAIL), gHiiHandle);
-                StringPtr      = AllocateZeroPool ((StrLen (StringRightPtr) + StrLen (StringErrorPtr)+ 1) * sizeof (CHAR16));
-                StrCpyS (StringPtr, StrLen (StringRightPtr) + StrLen (StringErrorPtr) + 1, StringRightPtr);
-                StrCatS (StringPtr, StrLen (StringRightPtr) + StrLen (StringErrorPtr) + 1, StringErrorPtr);
-                FreePool (StringRightPtr);
-                FreePool (StringErrorPtr);
+            if (MenuOption != NULL) {
```
leiflindholm: Return/break/continue instead of doing success handling.

MdeModulePkg/Universal/DisplayEngineDxe/FormDisplay.c
apop5: it should be. The only scenario where this comes into play is when the FormData is very broken. Broken enough that there will be lot of other problems other then returning a bad TopOfScreen integer.

MdeModulePkg/Universal/DisplayEngineDxe/FormDisplay.c
apop5: Yes, absolutly.  If we ever get into this bad state, this check will just prevent updating the Key menu (the bottom screen which tells you what key presses the setup menu will use), and what was previously displayed will stay there.

MdeModulePkg/Universal/DisplayEngineDxe/FormDisplay.c
leiflindholm: Right, this is also the final statement in the function, so no point to early return.

MdeModulePkg/Universal/DisplayEngineDxe/FormDisplay.c
```diff
@@ -3257,7 +3269,11 @@ UiDisplayMenu (
             // If the screen has no menu items, and the user didn't select UiReset
             // ignore the selection and go back to reading keys.
             //
-            ASSERT (MenuOption != NULL);
+            if (MenuOption != NULL ) {
```
pierregondois, line 3272: Hello Aaron,
I think these conditions should be inverted, this prevents from selecting any entry in the UiApp menu on my platform

apop5: Thanks @pierregondois 

I'll switch it to :
```
            if (MenuOption == NULL ) {
              ASSERT (MenuOption != NULL);
              break;
            }
```

apop5: #11803 11803

mikebeaton:
Relatively urgent I believe - https://github.com/tianocore/edk2/issues/11814

mikebeaton:
cc @pierregondois @apop5 - If one actually reads through the if/ASSERT pairs in https://github.com/tianocore/edk2/commit/901c4e32c65ded0802f641687ac1601b2c9c14b3 , it is clear the same momentary lapse of attention has occurred in multiple of them (i.e. the if condition and the assert condition are the same; in many cases it is the if condition is that is wrong compared to the clear intention, so the code is returning or breaking when it shouldn't, not asserting, and going somewhat haywire).

This obviously need fixing.

If I can take the opportunity to make the point, I was [discussing](https://github.com/tianocore/edk2/pull/11785#discussion_r2558754121) with @ardbiesheuvel whether ASSERT(FALSE) inside an if is a good pattern. I said it was, because it avoided the brainstrain and maintenance effort, for the current and any future programmers, of having to maintain two inverted expressions of the same condition. I don't suppose such decisions change quickly, if at all, in a given codebase (though I'm not at all sure EDK2 exhaustively avoids ASSERT (FALSE) inside if) but I feel it is reasonable to point out (or at least to claim) that this situation strengthens my case. I rather strongly suspect that if the coder had just had to get one condition right, and ASSERT(FALSE) inside it, these errors wouldn't have crept in.

PR #11689, opened 2025-10-31
MdeModulePkg: Universal folder Static analysis changes.

apop5:
# Description
Fixing CodeQl Scanning results for MdeModulePkg/Universal. (minus HiiDatabase, SetupBrowser and DisplayEngineDxe)

This is split from #11314 to try to make reviewing the changes simpler. 

- [ ] Breaking change?
- [x] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Local CI.

## Integration Instructions
No Integration necessary.

### DIFF
--- a/DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtPcieLib/SsdtPcieGenerator.c
+++ b/DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtPcieLib/SsdtPcieGenerator.c
@@ -546,7 +546,7 @@ GenerateRootPortDevices (
       goto exit_handler;
     }
 
-    if (RpInfo[Index].Sun != MAX_UINT64) {
+    if (RpInfo[Index].Sun != MAX_UINT32) {
       // ASL: Name (_SUN, <Sun value>)
       Status = AmlCodeGenNameInteger (
                  "_SUN",
--- a/MdeModulePkg/Bus/Ata/AtaAtapiPassThru/AhciMode.c
+++ b/MdeModulePkg/Bus/Ata/AtaAtapiPassThru/AhciMode.c
@@ -622,9 +622,10 @@ AhciWaitDeviceReady (
 
   //
   // According to SATA1.0a spec section 5.2, we need to wait for PxTFD.BSY and PxTFD.DRQ
-  // and PxTFD.ERR to be zero. The maximum wait time is 16s which is defined at ATA spec.
+  // and PxTFD.ERR to be zero. The maximum wait time is configurable and determined by
+  // PCD PcdSataDeviceReadyTimeout.
   //
-  PhyDetectDelay = 16 * 1000;
+  PhyDetectDelay = PcdGet8 (PcdSataDeviceReadyTimeout) * 1000;
   do {
     Offset = EFI_AHCI_PORT_START + Port * EFI_AHCI_PORT_REG_WIDTH + EFI_AHCI_PORT_SERR;
     if (AhciReadReg (PciIo, Offset) != 0) {
--- a/MdeModulePkg/Bus/Ata/AtaAtapiPassThru/AtaAtapiPassThru.inf
+++ b/MdeModulePkg/Bus/Ata/AtaAtapiPassThru/AtaAtapiPassThru.inf
@@ -67,6 +67,7 @@
 [Pcd]
   gEfiMdeModulePkgTokenSpaceGuid.PcdAtaSmartEnable          ## SOMETIMES_CONSUMES
   gEfiMdeModulePkgTokenSpaceGuid.PcdAhciCommandRetryCount   ## SOMETIMES_CONSUMES
+  gEfiMdeModulePkgTokenSpaceGuid.PcdSataDeviceReadyTimeout  ## CONSUMES
 
 # [Event]
 # EVENT_TYPE_PERIODIC_TIMER ## SOMETIMES_CONSUMES
--- a/MdeModulePkg/Core/Dxe/Dispatcher/Dispatcher.c
+++ b/MdeModulePkg/Core/Dxe/Dispatcher/Dispatcher.c
@@ -773,7 +773,10 @@ FvIsBeingProcessed (
   }
 
   KnownHandle = AllocateZeroPool (sizeof (KNOWN_HANDLE));
-  ASSERT (KnownHandle != NULL);
+  if (KnownHandle == NULL) {
+    ASSERT (KnownHandle != NULL);
+    return NULL;
+  }
 
   KnownHandle->Signature = KNOWN_HANDLE_SIGNATURE;
   KnownHandle->Handle    = FvHandle;
@@ -850,6 +853,7 @@ CoreFvToDevicePath (
   @retval EFI_ALREADY_STARTED   The driver has already been started. Only one
                                 DriverName may be active in the system at any one
                                 time.
+  @retval EFI_OUT_OF_RESOURCES  If memory could not be allocated for the DriverEntry.
 
 **/
 EFI_STATUS
@@ -867,7 +871,11 @@ CoreAddToDriverList (
   // NULL or FALSE.
   //
   DriverEntry = AllocateZeroPool (sizeof (EFI_CORE_DRIVER_ENTRY));
-  ASSERT (DriverEntry != NULL);
+  if (DriverEntry == NULL) {
+    ASSERT (DriverEntry != NULL);
+    return EFI_OUT_OF_RESOURCES;
+  }
+
   if (Type == EFI_FV_FILETYPE_FIRMWARE_VOLUME_IMAGE) {
     DriverEntry->IsFvImage = TRUE;
   }
@@ -1050,13 +1058,15 @@ CoreProcessFvImageFile (
       //
       if (gSecurity != NULL) {
         FvFileDevicePath = CoreFvToDevicePath (Fv, FvHandle, FileName);
-        Status           = gSecurity->FileAuthenticationState (
-                                        gSecurity,
-                                        AuthenticationStatus,
-                                        FvFileDevicePath
-                                        );
         if (FvFileDevicePath != NULL) {
+          Status = gSecurity->FileAuthenticationState (
+                                gSecurity,
+                                AuthenticationStatus,
+                                FvFileDevicePath
+                                );
           FreePool (FvFileDevicePath);
+        } else {
+          Status = EFI_OUT_OF_RESOURCES;
         }
 
         if (Status != EFI_SUCCESS) {
--- a/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
+++ b/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
@@ -161,7 +161,10 @@ CoreDumpGcdMemorySpaceMap (
   }
 
   Status = CoreGetMemorySpaceMap (&NumberOfDescriptors, &MemorySpaceMap);
-  ASSERT (Status == EFI_SUCCESS && MemorySpaceMap != NULL);
+  if (!((Status == EFI_SUCCESS) && (MemorySpaceMap != NULL))) {
+    ASSERT ((Status == EFI_SUCCESS) && (MemorySpaceMap != NULL));
+    return;
+  }
 
   if (InitialMap) {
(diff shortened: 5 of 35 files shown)


MdeModulePkg/Universal/Variable/Pei/Variable.c
```diff
@@ -894,8 +895,8 @@ FindVariableEx (
       ASSERT (Index < sizeof (IndexTable->Index) / sizeof (IndexTable->Index[0]));
       Offset  += IndexTable->Index[Index];
       MaxIndex = (VARIABLE_HEADER *)((UINT8 *)IndexTable->StartPtr + Offset);
-      GetVariableHeader (StoreInfo, MaxIndex, &VariableHeader);
-      if (CompareWithValidVariable (StoreInfo, MaxIndex, VariableHeader, VariableName, VendorGuid, PtrTrack) == EFI_SUCCESS) {
+      Valid    = GetVariableHeader (StoreInfo, MaxIndex, &VariableHeader);
```
leiflindholm, line 898: The tool is failing to flag other invocations (1084, 1175) ignoring the return value from GetVariableHeader further down in the file.

Possibly because in those locations VariableHeader is first needlessly initialized to NULL, meaning the compiler can happily UB its heart out. (In this function subsequent code actually relies on it.)

MdeModulePkg/Universal/BdsDxe/BdsEntry.c
```diff
@@ -1100,6 +1100,7 @@ BdsEntry (
       //
       // Retry to boot if any of the boot succeeds
       //
+      BootSuccess = FALSE;
```
leiflindholm: I think this change is correct, but I think it impacts needlessly convoluted code.
First of all, BootSuccess is initialized to false near the top of the function, about 400 lines up, but this is the first block actually using it.

Secondly, this appears to me to be an actual bugfix, since previously we would never exit the loop if LoadOptions becomes NULL on the next line.

Thirdly, I don't see any way the test of `if (!BootSuccess)` on original line 1115 could ever _not_ pass before. The question then becomes whether the fix here leads to desired outcome when that changes or whether _this_ BootSuccess and _that_ BootSuccess are actually unrelated items sloppily reusing a variable.

MdeModulePkg/Universal/BdsDxe/BdsEntry.c
```diff
@@ -833,23 +833,25 @@ BdsEntry (
...
     LoadOptions = EfiBootManagerGetLoadOptions (&LoadOptionCount, LoadOptionTypePlatformRecovery);
-    if (EfiBootManagerFindLoadOption (&PlatformDefaultBootOption, LoadOptions, LoadOptionCount) == -1) {
-      for (Index = 0; Index < LoadOptionCount; Index++) {
-        //
-        // The PlatformRecovery#### options are sorted by OptionNumber.
-        // Find the the smallest unused number as the new OptionNumber.
-        //
-        if (LoadOptions[Index].OptionNumber != Index) {
-          break;
+    if (LoadOptions != NULL) {
```
leiflindholm: Can we return/break/continue/goto instead?

MdeModulePkg/Universal/DriverHealthManagerDxe/DriverHealthManagerDxe.c
```diff
@@ -891,28 +899,29 @@ DriverHealthManagerCleanDynamicString (
   BufferSize      = sizeof (EFI_HII_PACKAGE_LIST_HEADER) + FixedStringSize + sizeof (EFI_HII_PACKAGE_HEADER);
   HiiPackageList  = AllocatePool (BufferSize);
   ASSERT (HiiPackageList != NULL);
+  if (HiiPackageList != NULL ) {
```
leiflindholm: Can we return/break/continue/goto instead?

MdeModulePkg/Universal/EbcDxe/EbcDebugger/EdbCmdSymbol.c
```diff
@@ -703,6 +703,8 @@ DebuggerLoadCode (
   CHAR16      *FileName;
   CHAR16      *MapFileName;
 
+  Buffer = NULL;
```
leiflindholm: Never dereferenced before it's initialized.

MdeModulePkg/Universal/EbcDxe/EbcDebugger/EdbCmdSymbol.c
```diff
@@ -800,6 +802,7 @@ DebuggerUnloadCode (
   EFI_STATUS  Status;
   VOID        *BufferPtr;
 
+  BufferPtr = NULL;
```
leiflindholm: Never referenced before it's initialized.

MdeModulePkg/Universal/FaultTolerantWritePei/FaultTolerantWritePei.c
```diff
@@ -261,7 +261,7 @@ PeimFaultTolerantWriteInitialize (
 
     if (!EFI_ERROR (Status)) {
       ASSERT (FtwLastWriteRecord != NULL);
-      if ((FtwLastWriteRecord->SpareComplete == FTW_VALID_STATE) && (FtwLastWriteRecord->DestinationComplete != FTW_VALID_STATE)) {
+      if ((FtwLastWriteRecord != NULL) && (FtwLastWriteRecord->SpareComplete == FTW_VALID_STATE) && (FtwLastWriteRecord->DestinationComplete != FTW_VALID_STATE)) {
```
leiflindholm: That ASSERT is checking whether FtwGetLastWriteRecord was called and if it did whether it returned success without updating FtwLastWriteRecord with valid information. That's not what ASSERTs are for.

Better to drop the ASSERT and add a goto End (326) or something. Yes, we end up with a bigger diff due to indentation change, but the end result will make more sense.

MdeModulePkg/Universal/PlatformDriOverrideDxe/PlatDriOverrideDxe.c
```diff
@@ -1341,7 +1350,10 @@ PlatOverMngrRouteConfig (
   @param  Value          A pointer to the data being sent to the original exporting driver.
   @param  ActionRequest  On return, points to the action requested by the callback function.
 
-  @retval EFI_SUCCESS    Always returned.
+  @retval EFI_SUCCESS           Changes were completed successfully.
+  @retval EFI_UNSUPPORTED       An action is being passed that is not supported.
+  @retval EFI_INVALID_PARMETER  Input parameters are invalid.
```
leiflindholm: Fixing the incorrect existing documentation should be a separate preceding commit so reviewer doesn't need to manually hunt for which (if any) possible return values were actually added.

MdeModulePkg/Universal/PlatformDriOverrideDxe/PlatDriOverrideLib.c
```diff
@@ -1717,6 +1724,7 @@ CheckMapping (
                                    recorded into the mapping database.
   @retval EFI_SUCCESS              The Controller and DriverImage are inserted into
                                    the mapping database successfully.
+  @retval EFI_OUT_OF_RESOURCES     Could not allocate a required resource.
```
leiflindholm, line 1726: Should also add EFI_NOT_FOUND. (And since it was already an undocumented existing return value, that could happen as part of the separate "correct documentation" commit I already requested.)

MdeModulePkg/Universal/BdsDxe/BdsEntry.c
apop5: I think returning/breaking is dangerous. This is BdsEntry, and if it just bail out, it will go back to dxe and assert.

This particular section is related to PlatformRecovery, and if there are no platform recovery boot options, then it should be able to continue on to boot.  With the current changes, it will check the returned LoadOptions and just proceed with the boot process if they are null.

leiflindholm: Fair. Could you rewrite it as 
'if ((LoadOptions != NULL) && EfiBootManagerFindLoadOption (...`
to avoid the indent changes?

MdeModulePkg/Universal/Variable/Pei/Variable.c
apop5: You are correct.

Fixed up those other locations with the same pattern to return if its not a valid header.

MdeModulePkg/Universal/EbcDxe/EbcDebugger/EdbCmdBreakpoint.c
```diff
@@ -324,6 +324,8 @@ DebuggerBreakpointSet (
   UINTN       Address;
   EFI_STATUS  Status;
 
+  Address = 0;
```
leiflindholm: Redundant. Address is always initialized before it's referenced.

apop5: Removed.

MdeModulePkg/Universal/BdsDxe/BdsEntry.c
leiflindholm: Maybe, given its potential actual bugfixness etc, break this out into its own PR?

I'm happy with the rest of the PR, but I think this one needs more review/discussion.

apop5: Moved this into #11724

leiflindholm (approved):

MdeModulePkg/Universal/PlatformDriOverrideDxe/PlatDriOverrideDxe.c
```diff
@@ -1224,9 +1225,18 @@ PlatOverMngrExtractConfig (
...
-    ConfigRequest    = AllocateZeroPool (Size);
-    ASSERT (ConfigRequest != NULL);
+    if (ConfigRequestHdr == NULL) {
+      return EFI_OUT_OF_RESOURCES;
+    }
+
+    Size          = (StrLen (ConfigRequestHdr) + 32 + 1) * sizeof (CHAR16);
+    ConfigRequest = AllocateZeroPool (Size);
+    if (ConfigRequest == NULL) {
+      ASSERT (ConfigRequest != NULL);
```
mikebeaton, line 1235: I don't think we need the ASSERT here any more, with the newly added return: thereby losing an ASSERT from the original code, but ending up following the standard pattern in EDK2 code for handling OUT_OF_RESOURCES errors.

MdeModulePkg/Universal/PlatformDriOverrideDxe/PlatDriOverrideLib.c
```diff
@@ -1919,8 +1938,9 @@ DeleteDriverImage (
     OverrideItemListIndex = GetNextNode (MappingDataBase, OverrideItemListIndex);
   }
 
-  ASSERT (Found);
-  ASSERT (OverrideItem->DriverInfoNum != 0);
```
mikebeaton, line 1923: Did we intend to lose the second ASSERT here? I believe not?

Also, in general, I believe we should keep ASSERTs as you have elsewhere, thus the patch here should be:
```
if (!Found || (OverrideItem->DriverInfoNum == 0)) {
  ASSERT (Found);
  ASSERT (OverrideItem->DriverInfoNum != 0);
  goto Exit;
}
```

apop5: I'll add the assert for found, but adding the OverrideItem check introduces another potential null pointer dereference
when the while loop skips anything, i.e. `while (!IsNull (MappingDataBase, OverrideItemListIndex))` returns TRUE on the first pass through, OverrideItem will still be NULL.

MdeModulePkg/Universal/PlatformDriOverrideDxe/PlatDriOverrideDxe.c
mikebeaton: Resolved. (Resolve comment will not work for me, for whatever reason (non-write access?).)

MdeModulePkg/Universal/PlatformDriOverrideDxe/PlatDriOverrideLib.c
mikebeaton: Resolved (I cannot resolve thread).

Issue #11635, opened 2025-10-21
Build fails with nasm 3.01 on x86_64


bbhtt:
Version: `edk2-stable202508-0-gd46aa46c8361194521391aa581593e556c707c6e`

```
        /buildstream-build/Build/Ovmf3264/RELEASE_GCC5/X64/UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHandlerLib/OUTPUT/X64/ExceptionHandlerAsm.iii:60: error: instruction not valid with 32-bit operand size
        /buildstream-build/Build/Ovmf3264/RELEASE_GCC5/X64/UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHandlerLib/OUTPUT/X64/ExceptionHandlerAsm.iii:60: error: instruction not valid with 32-bit operand size
        /buildstream-build/Build/Ovmf3264/RELEASE_GCC5/X64/UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHandlerLib/OUTPUT/X64/ExceptionHandlerAsm.iii:60: error: instruction not valid with 32-bit operand size
        /buildstream-build/Build/Ovmf3264/RELEASE_GCC5/X64/UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHandlerLib/OUTPUT/X64/ExceptionHandlerAsm.iii:60: error: instruction not valid with 32-bit operand size
        /buildstream-build/Build/Ovmf3264/RELEASE_GCC5/X64/UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHandlerLib/OUTPUT/X64/ExceptionHandlerAsm.iii:60: error: instruction not valid with 32-bit operand size
        /buildstream-build/Build/Ovmf3264/RELEASE_GCC5/X64/UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHandlerLib/OUTPUT/X64/ExceptionHandlerAsm.iii:60: error: instruction not valid with 32-bit operand size
        /buildstream-build/Build/Ovmf3264/RELEASE_GCC5/X64/UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHandlerLib/OUTPUT/X64/ExceptionHandlerAsm.iii:60: error: instruction not valid with 32-bit operand size
[...]

PR #11691, opened 2025-11-03
UefiCpuPkg/CpuExceptionHandlerLib: fix push instructions

kraxel:
Nasm 3.0 complains about 'dword' being invalid.  The comment talks about
a '8-byte value' so 'qword' should be correct here.

Fixes: https://github.com/tianocore/edk2/issues/11635


mikebeaton:
I'm not sure nasm is correctly enforcing anything here.

The 32-bit immediate push instruction is valid in 64-bit mode, and should sign extend as the comment says.

I believe we should await a nasm fix?

REFS:

- https://github.com/netwide-assembler/nasm/issues/166
- https://github.com/netwide-assembler/nasm/issues/130
- https://github.com/tianocore/edk2/issues/11635#issuecomment-3487302013

niruiyu (approved):
indeed a correct fix. thanks for catching it and fixing it.

UefiCpuPkg/Library/CpuExceptionHandlerLib/X64/ExceptionHandlerAsm.nasm
```diff
@@ -57,7 +57,7 @@ ALIGN   8
 AsmIdtVectorBegin:
 %assign Vector 0
 %rep  256
-    push    strict dword %[Vector] ; This instruction pushes sign-extended 8-byte value on stack
+    push    strict qword %[Vector] ; This instruction pushes sign-extended 8-byte value on stack
```
lzeng14: What is the original special purpose of the comment "sign-extended" ?

mikebeaton: I am only speculating, but was it just to clarify the even though the ~operand~ immediate data in the emitted instruction is a dword, it doesn't push a dword onto the stack†, but rather sign extends the dword value and pushes a qword.

† This is not as impossible as it at first sounds, since `push strict word` (`push small` as IDA disassembles it) in 64 bit mode does in fact push just a word onto the stack. (Have confirmed this with a quick test at https://app.x64.halb.it/ .)

cf My comment [here](https://github.com/netwide-assembler/nasm/issues/166#issuecomment-3488617465), which is why I am currently still not clear that it is us, rather than nasm, who should be changing.

mikebeaton:
Apologies for my running commentary, but even though I'm [starting to understand what is going on](https://github.com/netwide-assembler/nasm/issues/166#issuecomment-3488617465), I'm rather confused about what the nasm semantics really are, here.

The key thing is that the code in question needs to be able to ensure that each push takes the same number of bytes, even if some could be encoded in fewer*: we need a way to control that. So are the nasm size keywords really to control the true operand size (the amount of data pushed, in this case), or to control the immediate size (the amount of data used in the emitted instruction to encode the operand)? (The latter being what we need to control.) The answer, as far as the recent change in nasm goes, seems to be 'the true operand size'. But we still need to control the immediate size! And the true status in nasm still seems to be 'a bit of both'. At which point it might be valid to ask, why the change?

*0-127 would fit in a sign-extended byte; 128-255 can only fit in a sign-extended word, or larger, but there is no such word immediate length instruction (the closest such instruction actually pushes only a word to the stack, which is _not_ what we want!); so we need to ensure that every push uses a dword immediate value... which the old code did, in old nasm. And the new code does, in both old and new nasm. But the logic of the change in nasm still remains obscure, to me.

(Per [Peter Anvin's comment](https://github.com/netwide-assembler/nasm/issues/166#issuecomment-3487459065), and also tested at https://app.x64.halb.it/ , `push strict` with no size specifier also works as a fix - which makes more sense once you look at https://www.nasm.us/xdoc/3.01/nasmdoc.pdf §3.7 - but I believe it is much less clear what `push strict` or `push strict qword` are actually doing here, rather than the previous code `push strict dword` - which I continue to propose probably should not have been made invalid by nasm.)

mikebeaton:
I'll roll back (more than) a little, and say that it's pretty clear this fix is never going to become wrong, regardless of what nasm do or don't eventually change. (For example, `ndisasm` (v3 or v2) disassembles the resulting instruction `68 00 00 00 00` (the instruction which the code here is and was meant to generate) as `push qword 0x0`, and `strict` just means don't optimize by encoding in fewer bytes, even if possible.)

@kraxel - Should the comment just be removed, if this fix is applied? c.f. the 32 bit version: https://github.com/tianocore/edk2/blob/1c74842bd07a48070d1cf3458f0b6b377e3ec8b9/UefiCpuPkg/Library/CpuExceptionHandlerLib/Ia32/ExceptionHandlerAsm.nasm#L37

AshrafAliS (approved):

mikebeaton:
This is good to push.

PR #11697, opened 2025-11-04
MdeModulePkg: CoreDxe: Handle multiple MemoryAllocationModules

kuqin12:
# Description

The current implementation from Dxe/Image/Image.c does not handle the configuration where there might be multiple MemoryAllocationModules.

Given that the `ModuleName` is included in the hob data and used for targetting the consumer, DXE core should specify the GUID when looking up for its own MemoryAllocationModule.

This change adds a check to ensure the located hob is targetting DXE core.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This change was tested locally and booted to Windows Desktop on QEMU Q35.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Image/Image.c
+++ b/MdeModulePkg/Core/Dxe/Image/Image.c
@@ -200,7 +200,9 @@ CoreInitializeImageServices (
       //
       // Find Dxe Core HOB
       //
-      break;
+      if (CompareGuid (&DxeCoreHob.MemoryAllocationModule->ModuleName, &gEfiCallerIdGuid)) {
+        break;
+      }
     }
 
     DxeCoreHob.Raw = GET_NEXT_HOB (DxeCoreHob);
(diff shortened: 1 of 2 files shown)


os-d:
@kuqin12 it looks like that change is also needed here: https://github.com/tianocore/edk2/blob/05b677c9de2cb3c4486736eb795c12e8008c3797/MdeModulePkg/Core/Dxe/Mem/MemoryProfileRecord.c#L548

kuqin12:
> @kuqin12 it looks like that change is also needed here:
> 
Thanks. Nice call. This is updated.

kuqin12:
@lgao4 Could you please provide feedback on this change and let me know whether we can make this through soft freeze? The DxeCore could run into asserts if there are multiple allocation module hobs produced.

Please let me know your thoughts. Thanks in advance.

os-d (approved):

lgao4 (approved):

lgao4:
This change is good.

PR #11703, opened 2025-11-04
treewide: fix various typos

phip1611:
# Description

This PR is a series of multiple commits to fix various
typos in the code, originating mostly from copy&paste over the years.

Most of them only affect documentation and not code. I discovered them using [typos](https://github.com/crate-ci/typos). There are a lot of typos but I addressed some of the most low-hanging fruits.

It is my first contribution to EDK2 and I hope to match your expectations. I'm happy to adjust the PR if necessary.

- [ ] Breaking change? NO
- [ ] Impacts security? NO
- [ ] Includes tests? NO

## How This Was Tested

-

## Integration Instructions

-
This commit is the first in a series of multiple commits to fix various
typos in the code, originating mostly from copy&paste over the years.

Most of them only affect documentation and not code.

Signed-off-by: Philipp Schuster <philipp.schuster@cyberus-technology.de>

### DIFF
--- a/UnitTestFrameworkPkg/Test/UnitTest/Sample/SampleUnitTest/SampleUnitTest.c
+++ b/UnitTestFrameworkPkg/Test/UnitTest/Sample/SampleUnitTest/SampleUnitTest.c
@@ -1,5 +1,5 @@
 /** @file
-  This is a sample to demostrate the usage of the Unit Test Library that
+  This is a sample to demonstrate the usage of the Unit Test Library that
   supports the PEI, DXE, SMM, UEFI SHell, and host execution environments.
 
   Copyright (c) Microsoft Corporation.<BR>


ardbiesheuvel (approved):

leiflindholm:
> Approved for EmulatorPkg. I’ll note that I’m in favor of fixing spelling errors, but it is much more effective to prevent them. Right now a number of packages have cspell in audit mode, so it doesn’t fail CI. If we are going to take the time to fix spelling errors, I’d prefer we fix all of them in a package and remove audit mode, so that we catch future errors in PR checks, before they are checked in.
> 
> There is a separate question of do we have too permissive of cspell dictionaries (I think yes, cpp is particularly bad, it had lots of wildcards in there that let misspelled words in if you have some substring).
> 
> You don’t need to fix this in your PR, but just noting thay is a more impactful change.

If you feel that strongly about it, you could substantially lower the barrier for anyone doing what you're asking by submitting fixes to the various misspellings in the tree that would break if cspell was enabled.

Regardless, a PR is not the place to discuss this. I'm reliably informed by your colleagues that https://github.com/tianocore/edk2/discussions is "A better, more accessible/traceable/reliable way to discuss community topics". Perhaps move the conversation there, or an issue?

ardbiesheuvel:
> Approved for EmulatorPkg. I’ll note that I’m in favor of fixing spelling errors, but it is much more effective to prevent them. Right now a number of packages have cspell in audit mode, so it doesn’t fail CI. If we are going to take the time to fix spelling errors, I’d prefer we fix all of them in a package and remove audit mode, so that we catch future errors in PR checks, before they are checked in.
> 

While I agree with this in principle, the reality is that for packages such as ArmPkg, the list with exceptions becomes so long, and so tedious to maintain, that I'd rather keep it in audit mode. Once we manage to train SpellCheck on the ARM ARM, we could revisit this, but otherwise I don't see the point.

os-d:
> If you feel that strongly about it, you could substantially lower the barrier for anyone doing what you're asking by submitting fixes to the various misspellings in the tree that would break if cspell was enabled.
> 
> Regardless, a PR is not the place to discuss this. I'm reliably informed by your colleagues that https://github.com/tianocore/edk2/discussions is "A better, more accessible/traceable/reliable way to discuss community topics". Perhaps move the conversation there, or an issue?

I don't feel that strongly about it :)

I was mostly noting to @phip1611, who noted they are a first time contributor that these fixes are appreciated and if they want to take it further, it can be, to more effect. I don't think I made that clear that I wasn't asking for changes to this PR, but saying that we can play whack-a-mole with spelling errors or we can fix it. My assumption was that the contributor, having made a first, simple spelling fix change, may be looking for more items to take up.

If you want, I can file an issue or start a discussion, but I don't have the bandwidth to take up cspell violations at the moment.

makubacki:
We've brought up enforce cspell in the past and ultimately what I think has been missing is (1) a structured, well communicated plan and (2) bandwidth to execute the plan. A GitHub discussion/issue would be a great place to devise the plan and get feedback. I also think the plan needs to include considerations like a common project dictionary ("edk2" dictionary), so the package's only need to maintain a delta of truly unique words in their specific package which should be relatively small.

phip1611:
FYI: typos appears to be the new de facto standard replacing cspell.

https://github.com/crate-ci/typos

It comes with a large, community-maintained replace-list of common typos receiving fresh updates every couple of weeks. 

Would nevertheless be a pain to properly set up for the whole repo. You'd need a `.typos.toml` listing a ton of exceptions (individual ones and whole files/directories).

liyi77 (approved):

gguo11837463 (approved):
Approve from UefiPayloadPkg

PR #11706, opened 2025-11-04
MdeModulePkg: Universal/Disk Static analysis changes.

apop5:
# Description
Fixing CodeQl Scanning results for MdeModulePkg/Universal/Disk.

- [ ] Breaking change?
- [x] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Local CI.

## Integration Instructions
No Integration necessary.


apop5:
not stale

mikebeaton (approved):
Have checked through these changes and they all look reasonable to me.

apop5:
> Have checked through these changes and they all look reasonable to me.

Thanks mike!

MdeModulePkg/Universal/Disk/CdExpressPei/PeiCdExpress.c
mikebeaton, line 193: I don't like the fact that we've had to change from the easily readable `if (BlockIo2) ... else` to the pair of conditions in each case below checking whether `BlockIoPpi`, `BlockIo2Ppi` are initialised. The actual issue is that the analyser can't know whether these two vars were set non-NULL by the two calls here.

Can we add a check for non-null, with an ASSERT and break if the check fails, after each of these two calls, and return to the more readable `if (BlockIo2) ... else` below.

MdeModulePkg/Universal/Disk/DiskIoDxe/DiskIo.c
```diff
@@ -997,10 +997,13 @@ DiskIo2ReadWriteDisk (
...
     EfiAcquireLock (&Instance->TaskQueueLock);
-    RemoveEntryList (&Task->Link);
+    if (Task != NULL) {
+      RemoveEntryList (&Task->Link);
+    }
+
     EfiReleaseLock (&Instance->TaskQueueLock);
 
-    if (!EFI_ERROR (Status) && (Task->Token != NULL)) {
+    if (!EFI_ERROR (Status) && (Task != NULL) && (Task->Token != NULL)) {
```
mikebeaton: It's not logically possible for `Task` to be NULL here, because of the `!Blocking` check, and the earlier `Blocking` logic above.

I do not like having to introduce extra code, which is less readable and harder to maintain, because an analyser can't understand the logic of the code - if that is what has happened here.

MdeModulePkg/Universal/Disk/PartitionDxe/Mbr.c
```diff
@@ -127,7 +127,7 @@ PartitionInstallMbrChildHandles (
 {
   EFI_STATUS                   Status;
   MASTER_BOOT_RECORD           *Mbr;
-  UINT32                       ExtMbrStartingLba;
+  UINT64                       ExtMbrStartingLba;
```
mikebeaton: It looks like this logically only ever contains a UINT32 value.

Can we change to adding casts (only) where necessary below, to avoid warnings when using this value, rather than changing the type here.

MdeModulePkg/Universal/Disk/RamDiskDxe/RamDiskProtocol.c
```diff
@@ -234,7 +238,7 @@ RamDiskPublishNfit (
     }
   }
 
-  if (!EFI_ERROR (Status)) {
+  if (!EFI_ERROR (Status) && (TableHeader != NULL)) {
```
mikebeaton: This one doesn't make sense. TableHeader can only be non-NULL here in the case were it has already been de-referenced, without checking whether it was non-NULL, in line 233 just above. I don't quite understand how the static analyser you are using is coming up with these recommendations....

I think you could reasonably have `if (!EFI_ERROR (Status) && (TableHeader != NULL))` in line 230 above, instead of here.

mikebeaton (changes requested):

mikebeaton:
> Have checked through these changes and they all look reasonable to me.

@apop5 - Apologies, I'm not sure where they're suddenly not looking so reasonable to me now, but please consider the revised comments. Thanks.

apop5:
> > Have checked through these changes and they all look reasonable to me.
> 
> @apop5 - Apologies, I'm not sure where they're suddenly not looking so reasonable to me now, but please consider the revised comments. Thanks.

No worries, I will look through the feedback and respond/fix/discuss

MdeModulePkg/Universal/Disk/CdExpressPei/PeiCdExpress.c
apop5: I agree with your acessment.

I've reverted these changes. We can deal with the false positives with a suppression file later.

MdeModulePkg/Universal/Disk/PartitionDxe/Mbr.c
apop5: I've reverted it. After reverting it, the error wasn't reported again.

MdeModulePkg/Universal/Disk/RamDiskDxe/RamDiskProtocol.c
apop5: I've reverted it as well.

MdeModulePkg/Universal/Disk/DiskIoDxe/DiskIo.c
apop5: I've reverted this as well. Will revisit it later.

MdeModulePkg/Universal/Disk/RamDiskDxe/RamDiskProtocol.c
mikebeaton: This doesn't appear to have been successfully reverted.

mikebeaton: I don't seem to be able to unresolve the other comment at this point, but have added a comment in there.

apop5: Now its reverted.

MdeModulePkg/Universal/Disk/RamDiskDxe/RamDiskImpl.c
```diff
@@ -391,7 +391,7 @@ HiiCreateRamDisk (
                   &BufferSize,
                   (VOID *)(UINTN)StartingAddr
                   );
-    if (BufferSize != FileInformation->FileSize) {
+    if ((FileInformation != NULL) && (BufferSize != FileInformation->FileSize)) {
```
mikebeaton: Sorry, another similar one. FileInformation also cannot be NULL here, given the actual logic of the code above.

apop5: reverted.

mikebeaton (approved):

apop5:
not stale, waiting for review.

apop5:
@lgao4 

If you could take a look at this?

lgao4 (approved):

PR #11707, opened 2025-11-04
Handle when MM communication buffer region is added to GCD before MMCommunicationDxe

kuqin12:
# Description

The MM communication buffer memory is being added in MM communication DXE driver through gDS calls:
https://github.com/tianocore/edk2/blob/05b677c9de2cb3c4486736eb795c12e8008c3797/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c#L758

This means that the region should not be reported as available through resource descriptor hob.

This change started with moving all PCD get calls to local variables in MemoryInitPeiLib.c and then handles the incoming MM comm buffer:

1. If it is not added, add and set attributes;
2. If it is already added, assume it is already allocated (otherwise it will cause other memory corruption issues), first free the space, then remove the space and add it like before. This is to ensure the attributes are set properly.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested on QEMU SBSA platform and booted to UEFI shell as well as Windows desktop.

## Integration Instructions

N/A


ArmPlatformPkg/MemoryInitPei/MemoryInitPeiLib.c
```diff
@@ -114,12 +116,48 @@ MemoryPeim (
...
+        );
+    } else {
+      // The mm communication buffer is out of the system memory range
+      BuildResourceDescriptorHob (
+        EFI_RESOURCE_SYSTEM_MEMORY,
+        ResourceAttributes,
+        SystemMemoryBase,
+        PcdGet64 (PcdSystemMemorySize)
+        );
+    }
```
ardbiesheuvel: I don't like this at all tbh. Punching holes into a reported resource like this is kind of feasible for the first one, but once there are more regions to consider, it quickly gets out of hand.

Couldn't we generate a memory reservation HOB at this stage, and make the MmCommunicate DXE deal with the possibility that the memory region already exists in the GCD?

(In any case, please don't disregard the first patch even if we decide to take another route with this)

kuqin12: Thanks for the feedback! I understand your concern and agree that the current change only addresses a single region. One possible improvement could be to:

1. Coalesce all region start and end points
2. Sort them
3. Process them in a “scanning” fashion

This approach would allow us to handle multiple regions more flexibly.

Memory allocation hob is also feasible. However, since there’s no explicit contract, we’d still need to verify whether a region (in this case, the MM comm buffer) overlaps with system memory, and reserve memory holes accordingly. As I wonder some platform logic might already account for this.

That said, I agree your proposed approach is simpler. :) Please let me know if you have a strong preference; otherwise, I’ll proceed with updates in the next push.

ardbiesheuvel: > Thanks for the feedback! I understand your concern and agree that the current change only addresses a single region. One possible improvement could be to:
> 
> 1. Coalesce all region start and end points
> 2. Sort them
> 3. Process them in a “scanning” fashion
> 
> This approach would allow us to handle multiple regions more flexibly.
> 

Sure, that would be better. But I don't think such complicated logic belongs here to begin with.

> Memory allocation hob is also feasible. However, since there’s no explicit contract, we’d still need to verify whether a region (in this case, the MM comm buffer) overlaps with system memory, and reserve memory holes accordingly. As I wonder some platform logic might already account for this.
> 

Conceptually, what we need to do is
* describe the region as system memory
* reserve it for use by the MmCommunicate DXE

Omitting it from system memory entirely is dubious, because we lose the 'RAM' annotation too, which is important on ARM systems.

If we cannot do the above by creating resource descriptor HOBs, perhaps we should just allocate the pages as soon as permanent memory becomes available in PEI.

kuqin12: Sorry for the delay.

I explored several approaches to make the MM communication buffer a reserved region. However, on a PEI-enabled platform, the memory manager isn’t advanced enough to allocate at a specific address, and any memory allocation HOB after SEC handoff won’t be honored by the PEI core. Therefore, the carve-out for the MM communication buffer must happen in the SEC phase for the reservation to take effect.

I could update the PR to add a memory allocation HOB here or in ArmPlatformLib and have the MM communication DXE handle it more appropriately. That said, this approach would likely only work for PeiLess ARM64 platforms.

I’m open to making our platform PeiLess, but I’d like to understand if there’s a common pattern we should follow for ARM64 platforms—which often have multiple carve-out regions—to ensure PEI-enabled architectures can handle such reservations correctly.

kuqin12: I updated the PR and replaced the hole-poking code with GCD memory handling from the MM communication DXE side.

Happy New Year!

ArmPlatformPkg/MemoryInitPei/MemoryInitPeiLib.c
```diff
@@ -88,15 +89,21 @@ MemoryPeim (
                         EFI_RESOURCE_ATTRIBUTE_TESTED
                         );
 
+  //
+  // Use local variable to avoid multiple PcdGet64 calls
+  //
```
mikebeaton, line 94: Shouldn't this commit's changes be a separate PR? (IMO - happy to be confirmed or overridden by those who know better.)

Is there much to be gained by avoiding multiple PcdGet64 'calls', since they're just a macro expansion? E.g. the old code is no less efficient (given how the macros actually work) and makes it more explicit where the values are used. (Again, IMO...)

kuqin12: This is kept because of the comment from Ard earlier (https://github.com/tianocore/edk2/pull/11707#discussion_r2526506454).

I agree that the first commit becomes less relevant now that we’ve shifted from using resource descriptions in PEI/SEC to relying on GCD processing in DXE. That said, the change still has value: it ensures the PCD is not read multiple times, making it easier to update the source in one place and enabling potential preprocessing or other future enhancements.

ardbiesheuvel (approved):

PR #11708, opened 2025-11-04
OvmfPkg/RiscVVirt/PlatformSecLib: Actually reserve handoff region

benjamindoron:
# Description

An unterminated comment meant that the handoff region wasn't reserved. This will result in stack corruption.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This is a port of https://github.com/tianocore/edk2-platforms/pull/898, which was tested by booting the QEMU virt board. As OVMF uses similar logic, it's assumed this is correct and necessary here too.

## Integration Instructions

N/A

### DIFF
--- a/OvmfPkg/RiscVVirt/Library/PlatformSecLib/SecEntry.S
+++ b/OvmfPkg/RiscVVirt/Library/PlatformSecLib/SecEntry.S
@@ -15,7 +15,7 @@ li    s0, 0
 li    a2, FixedPcdGet32 (PcdOvmfSecPeiTempRamBase)
 li    a3, FixedPcdGet32 (PcdOvmfSecPeiTempRamSize)
 
-/* Reserve region to store handoff data
+/* Reserve region to store handoff data */
 li    a4, SEC_HANDOFF_DATA_RESERVE_SIZE
 sub   a3, a3, a4
 


vlsunil (approved):

vlsunil:
@benjamindoron : Looks like some permission is required for your branch. Could you please take a look?

benjamindoron:
> @benjamindoron : Looks like some permission is required for your branch. Could you please take a look?

I don't know what the problem was, but it looks like it's fixed now

ardbiesheuvel:
> > @benjamindoron : Looks like some permission is required for your branch. Could you please take a look?
> 
> I don't know what the problem was, but it looks like it's fixed now

I committed it by hand. Normally, we use mergify which expects to be able to make changes to your branch, so for future PRs, please keep this in mind.

benjamindoron:
> > > @benjamindoron : Looks like some permission is required for your branch. Could you please take a look?
> > 
> > 
> > I don't know what the problem was, but it looks like it's fixed now
> 
> I committed it by hand. Normally, we use mergify which expects to be able to make changes to your branch, so for future PRs, please keep this in mind.

Oh, I see. Sorry for the trouble, and thanks for the help. I'll double check that the box is ticked next time

PR #11711, opened 2025-11-05
EmulatorPkg/Application/RedfishPlatformConfig: Add EndPointer parameter to StrDecimalToUintnS for potential use

SharingYang:
Now transfer null parameter to StrDecimalToUintnS as fllows:
ReturnStatus = StrDecimalToUintnS (Argv[6], NULL, &RedfishServiceIpPort);
It's not convenient to debug when have problems, because there are no return value.
So, Add EndPointer parameter to StrDecimalToUintnS for potential use.


SharingYang:
Close the pr because it's not needed.

PR #11712, opened 2025-11-05
Do some improvement for CreateRscHandlerCallbackPacket

SharingYang:
Do some improvement for CreateRscHandlerCallbackPacket:
1.Return if NumberOfEntries is NULL to avoid null pointer access, because ASSERT don't block code execution   with release version.
2.Zero Hob data when create Hob successfully. Ther previous code just zero the first UINTN.

### DIFF
--- a/MdeModulePkg/Universal/ReportStatusCodeRouter/Pei/ReportStatusCodeRouterPei.c
+++ b/MdeModulePkg/Universal/ReportStatusCodeRouter/Pei/ReportStatusCodeRouterPei.c
@@ -57,6 +57,9 @@ CreateRscHandlerCallbackPacket (
                       sizeof (EFI_PEI_RSC_HANDLER_CALLBACK) * 64 + sizeof (UINTN)
                       );
   ASSERT (NumberOfEntries != NULL);
+  if (NumberOfEntries == NULL) {
+    return NumberOfEntries;
+  }
 
   *NumberOfEntries = 0;
 
@@ -135,6 +138,9 @@ Register (
 
   if (FreePacket == NULL) {
     FreePacket = CreateRscHandlerCallbackPacket ();
+    if (FreePacket == NULL) {
+      return EFI_OUT_OF_RESOURCES;
+    }
   }
 
   CallbackEntry                 = (EFI_PEI_RSC_HANDLER_CALLBACK *)(FreePacket + 1);


panyongg:
@lgao4 @mdkinney , Could you please help to review this code changes? And plesae help to approve it if no concerns. Thanks.

MdeModulePkg/Universal/ReportStatusCodeRouter/Pei/ReportStatusCodeRouterPei.c
```diff
@@ -57,8 +59,16 @@ CreateRscHandlerCallbackPacket (
...
+  if (NumberOfEntries == NULL) {
+    return NumberOfEntries;
+  }
 
+  //
+  // Clear Hob Data.
+  //
   *NumberOfEntries = 0;
+  CallbackEntry    = (EFI_PEI_RSC_HANDLER_CALLBACK *)(NumberOfEntries + 1);
+  SetMem (CallbackEntry, 64 * sizeof (EFI_PEI_RSC_HANDLER_CALLBACK), 0);
```
niruiyu: What bug will appear if no ZeroMem()?

SharingYang: 1.No Bug was encountered.
2.The original code set the CallbackEntry[Index] as NULL using the following code  to prevent gabage data to lead misjudgement lik **if (CallbackEntry[Index] == NULL)**:
  for (Index = 0; Index < 64; Index++) {
    CallbackEntry[Index] = NULL;
  }
3.We then found bugs "Solve the link warning: unresolved external symbol memset" and changed it to using edk2 memory lib.

image

niruiyu: But I cannot find the for-loop you mentioned in the original code.

SharingYang: 1.It's in the git logs for ExReportStatusCodeRouterPei in Intel\ServerPlatformPkg\Universal\PeiExStatusCodeRouter.
image
image
2. The handler dispatcher in PEI statuscode router driver check if handler is NULL, if it is not NULL, then handler can be called.
Assign the initial value for  CallbackEntry as NULL to ensure safety.
image

niruiyu: ReportDispatcher() runs the for-loop for *NumberOfEntries times.
So if the first UINTN is assigned to 0 properly, there should be no functionality issue.

If there is no bug, why do we change the code?

SharingYang: Hi Ray,
I checked the code and did a test pr to do coverity check, It shows the code is really redundant.
I will remove the code if you agree.

MdeModulePkg/Universal/ReportStatusCodeRouter/Pei/ReportStatusCodeRouterPei.c
niruiyu, line 146: FreePacket may be NULL. Do we need to add a check? otherwise static-code checker may still complain.

SharingYang: Thanks, Fixed it. Please help review.

lgao4:
@os-d , does codeql detect this issue?

os-d:
@niruiyu @lgao4 Coverity is not run for edk2, but CodeQL is and has a similar purpose. Potential null pointer dereference is often caught by CodeQL. However, edk2 has CodeQL configured to ignore all existing issues and only flag new issues. You can see there are 304 flagged issues in the current codebase: https://github.com/tianocore/edk2/security/code-scanning. So, I suspect that this was fixed at all because when making changes to this module caused CodeQL to flag an existing issue because new changes were made in this area.

Static analysis tools are not completely perfect and/or deterministic, sometimes they miss errors until code around them changes.

lgao4:
if no more feedback, I will merge it this week.

PR #11714, opened 2025-11-05
BaseTools: Scripts: enhance UEFI debug python scripts.

grandpaul:
# Description
This PR enhance the UEFI debug python scripts a little bit. Firstly we wrap the top-level code to let people
able to import that python script as a library. And we fix the gST searching logic to make it fail more gracefully.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

I manually test it with qemu + gdb.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Scripts/efi_gdb.py
+++ b/BaseTools/Scripts/efi_gdb.py
@@ -572,14 +572,16 @@ class EfiTablesCmd (gdb.Command):
 
         gST = gdb.lookup_global_symbol('gST')
         if gST is None:
-            print('Error: This command requires symbols for gST to be loaded')
-            return
+            table = EfiConfigurationTable(self.file)
+        else:
+            table = EfiConfigurationTable(
+                self.file, int(gST.value(gdb.selected_frame())))
 
-        table = EfiConfigurationTable(
-            self.file, int(gST.value(gdb.selected_frame())))
         if table:
             print(table, '\n')
-
+        elif gST is None:
+            print('Error: This command requires symbols for gST to be loaded')
+            return
 
 class EfiSymbolsCmd (gdb.Command):
     """Load Symbols for EFI. Type 'efi symbols -h' for more info."""
--- a/BaseTools/Scripts/efi_gdb.py
+++ b/BaseTools/Scripts/efi_gdb.py
@@ -876,43 +876,46 @@ class LoadEmulatorEfiSymbols(gdb.Breakpoint):
         # keep running
         return False
 
+def main():
+    # Get python backtraces to debug errors in this script
+    gdb.execute("set python print-stack full")
 
-# Get python backtraces to debug errors in this script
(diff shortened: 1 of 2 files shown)


mikebeaton:
Please clarify why you need this change, i.e. use-case for modified code that can't be done with original.

grandpaul:
> Please clarify why you need this change, i.e. use-case for modified code that can't be done with original.
@mikebeaton 

Sorry for late reply. We made some extra commands. Not sure if it is ok to upstream these. 
We implement 3 extra commands for gdb. So we re-use the code in edk2. Code is here. https://paste.debian.net/1412261/
1. efi showvendor. To show who provides the EFI debug functions. Could be edk2, u-boot or LittleKernel.
2. efi loadsymbols: Calculate the address and log debug symbols. We are using this command to debug AOSP GBL (Generic boot loader) through hardware debugger.
3. efi list: List the apps address and path.  

Without the modification, when we do "from efi_gdb import EfiCmd, GdbFileObject, EfiTablesCmd" we will get an error.

Or do you think I can also upstream these features too?

Thanks.

mikebeaton:
The explanation makes sense.

Some of your commits are missing Signed-off-by. You don't need merge commits in your PR. Go back to how it was without merge commits, rebase it onto master and force push it to your PR branch.

grandpaul:
> Some of your commits are missing Signed-off-by. You don't need merge commits in your PR. Go back to how it was without merge commits, rebase it onto master and force push it to your PR branch.

Thanks.
Rebased and pushed. I've checked the commits messages, it seems all of them has the Signed-off-by.

mikebeaton:
I don't use these scripts, because I work on a project which has its own UEFI debug scripts. I would have to get up to speed on running these ones to be able to give more definitive feedback. Hence the silence so far, after your previous changes - for which, thanks.

BaseTools/Scripts/efi_gdb.py
```diff
@@ -572,14 +572,16 @@ def invoke(self, arg, from_tty):
...
-            print('Error: This command requires symbols for gST to be loaded')
-            return
+            table = EfiConfigurationTable(self.file)
+        else:
+            table = EfiConfigurationTable(
+                self.file, int(gST.value(gdb.selected_frame())))
 
-        table = EfiConfigurationTable(
-            self.file, int(gST.value(gdb.selected_frame())))
         if table:
```
mikebeaton, line 580: I am not sure the original logic was right here. I think `if table:` will always be true. (`table` does not implement `__len__` or `__bool__`, and does not inherit from `dict`.)

grandpaul: Sorry for late reply. I'm in holidays for the new year. 
Yeah, you are right. I've implemented the __bool__ function to make it work. Thanks.

PR #6087, opened 2024-08-21
OvmfPkg, ArmVirtPkg: Fix unable to build with -D NETWORK_ENABLE=0

mikebeaton:
# Description

This PR fixes https://bugzilla.tianocore.org/show_bug.cgi?id=4829 , namely the issue that Intel OvmfPkg platforms can no longer be built with -D NETWORK_ENABLE=0 after 7f17a155640a2a9e1f7b0f3522628ee2c6f62624 . In addition it fixes the same issue in non-Intel OvmfPkg platforms and ArmVirtPkg platforms, even though the issue in those other platforms is masked until https://github.com/tianocore/edk2/pull/6092 is applied.

- [ ] Breaking change? NO
- [ ] Impacts security? NO
- [ ] Includes tests? NO
  - Perhaps ideally this would update the CI to build all these platforms also with -D NETWORK_ENABLE=0, but I have not (at least so far) done this. (Note that everything built in CI with -D NETWORK_ENABLE=1 (default) (which is some but I am not sure if all of the platforms I have changed) still passes.)

## How This Was Tested

Confirm that after these changes Intel OvmfPkg variants build and run without network support, when compiled with -D NETWORK_ENABLE=0. Confirm that when paired with https://github.com/tianocore/edk2/pull/6092 the other affected packages build correctly with -D NETWORK_ENABLE=0. Confirm that everything still builds and runs successfully in CI.

## Integration Instructions

Paired PR https://github.com/tianocore/edk2/pull/6092 addresses the issue discovered while preparing this, namely that other OvmfPkg platforms and ArmVirtPkg platforms are only not showing this same issue because it is masked in those cases by the fact that those packages are still incorrectly including some NetworkPkg components when they should be disabled.

PR #11719, opened 2025-11-05
OvmfPkg,ArmVirtPkg,NetworkPkg: Fix build with -D NETWORK_ENABLE=0

mikebeaton:
# Description

Previous PR https://github.com/tianocore/edk2/pull/6087 restored the ability to build OvmfPkg with -D NETWORK_ENABLE=0.

Unfortunately b3b3cfab7eb52acd77558a9727196e30d05d1f2a has broken it again, since the .dsc references to OvmfPkg/VirtioNetDxe/VirtioNet.inf were moved from outside !if $(NETWORK_ENABLE) == TRUE to inside it, while the .fdf references remained outside.

In discussion below it was decided to move the .fdf references inside the conditional, i.e. not to include VirtioNetDxe unless the rest of the network stack is being built.

Removal of VirtioNetDxe driver on -D NETWORK_ENABLE=0 has only been applied to those packages which are already using OvmfPkg/Dsc/Includes/NetworkComponents.dsc.inc.

f9408b7cc160030c3bb9627c8086a792cc2d046c introduces new PCDs which also need to be moved inside a NETWORK_ENABLE test in order to allow building with -D NETWORK_ENABLE=0, which is also done here.

- [ ] Breaking change?
  - NO
- [ ] Impacts security?
  - NO
- [ ] Includes tests?
  - NO

## How This Was Tested

 - Test building OvmfPkgX64, OvmfPkgIa32X64, ArmVirtQemu, ArmVirtQemuKernel, CloudHvX64 w/ & w/out NETWORK_ENABLE=0, and confirm VirtioNetDxe isn't/is built.
 - Test running OvmfPkgX64, OvmfPkgIa32X64, ArmVirtQemu w/ & w/out NETWORK_ENABLE=0, and confirm running normally and VirtioNetDxe.efi isn't/is loaded.

## Integration Instructions

N/A

### DIFF
--- a/ArmVirtPkg/ArmVirtQemu.dsc
+++ b/ArmVirtPkg/ArmVirtQemu.dsc
@@ -486,7 +486,6 @@
   OvmfPkg/Fdt/HighMemDxe/HighMemDxe.inf
   OvmfPkg/VirtioBlkDxe/VirtioBlk.inf
   OvmfPkg/VirtioScsiDxe/VirtioScsi.inf
-  OvmfPkg/VirtioNetDxe/VirtioNet.inf
   OvmfPkg/VirtioRngDxe/VirtioRng.inf
   OvmfPkg/VirtioSerialDxe/VirtioSerial.inf {
     <PcdsFixedAtBuild>
--- a/ArmVirtPkg/ArmVirtQemuFvMain.fdf.inc
+++ b/ArmVirtPkg/ArmVirtQemuFvMain.fdf.inc
@@ -101,7 +101,9 @@ READ_LOCK_STATUS   = TRUE
   # Platform Driver
   #
   INF OvmfPkg/VirtioBlkDxe/VirtioBlk.inf
+!if $(NETWORK_ENABLE) == TRUE
   INF OvmfPkg/VirtioNetDxe/VirtioNet.inf
+!endif
   INF OvmfPkg/VirtioScsiDxe/VirtioScsi.inf
   INF OvmfPkg/VirtioRngDxe/VirtioRng.inf
   INF OvmfPkg/VirtioSerialDxe/VirtioSerial.inf
--- a/ArmVirtPkg/ArmVirtQemuKernel.dsc
+++ b/ArmVirtPkg/ArmVirtQemuKernel.dsc
@@ -379,7 +379,6 @@
   OvmfPkg/Fdt/HighMemDxe/HighMemDxe.inf
   OvmfPkg/VirtioBlkDxe/VirtioBlk.inf
   OvmfPkg/VirtioScsiDxe/VirtioScsi.inf
-  OvmfPkg/VirtioNetDxe/VirtioNet.inf
   OvmfPkg/VirtioRngDxe/VirtioRng.inf
   OvmfPkg/VirtioSerialDxe/VirtioSerial.inf
 
--- a/NetworkPkg/NetworkDynamicPcds.dsc.inc
+++ b/NetworkPkg/NetworkDynamicPcds.dsc.inc
@@ -11,15 +11,18 @@
 #
 ##
 
-#
-# IPv4 and IPv6 PXE Boot support.
-#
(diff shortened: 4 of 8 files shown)


kraxel:
> Somewhat non-obviously, VirtioNet has always been intended to be included even if NETWORK_ENABLE was false (see for instance the comment introduced at https://github.com/tianocore/edk2/blob/7a1f59476d4/OvmfPkg/README#L159-L160 which is still present).

I don't think this is what the comment refers to.  non-virtio qemu NICs get a EFI (from ipxe) driver via pci option rom, so they will only be loaded in case the device is actually present.  the virtio-net driver is always there because it comes with the firmware image instead.

IMHO excluding the drivers with NETWORK_ENABLE=FALSE is fine, I'd suggest to fix the fdf instead.

mikebeaton:
Hi!

Right, I see what you're saying the comment refers to, so I should change that in the commit message.

But still, up to now, it seems VirtioNetDxe has intentionally rather than mistakenly been included whether or not the network stack is included. I'm not quite sure historically why? Perhaps this is the main point to be sure about before moving it inside?

I did also mean to mention this in the commit message, but  since https://github.com/tianocore/edk2/commit/b9cb18206a301daef2a843c024da578eaa3f4a91 we have an option to control whether VirtioNetDxe starts anyway, so given the new PCD, possibly forcing VirtioNetDxe to be included or not with the rest of the network stack unnecessarily removes some fine-grained control?

I had a situation recently where I was testing the network stack distributed as separate drivers which can be loaded by our bootloader, and I actually wanted VirtioNetDxe but not the rest of the network stack. :-) (Although, to argue against myself, if I'm manually loading certain drivers in a bootloader anyway, I could always have loaded VirtioNetDxe if it wasn't there.)

kraxel:
Well, if there is an actual use case for virtio-net driver without the rest of the network stack I'm fine with it too.

mikebeaton:
@kraxel - I will look into it some more and get back in a day or so. On reflection I'm tending towards agreeing with you, unless anyone can come back with a good reason as to why it wasn't like this in the first place.

mikebeaton:
@kraxel - Hi, after discussion with other members of the project I work on, and in agreement with your initial instincts, we have decided that it is better to move the .fdf references to VirtioNetDxe inside NETWORK_ENABLE tests, i.e. to only include VirtioNetDxe.efi when at least some of the rest of the network stack is being built. So I have updated the PR to do it that way.

In the probably fairly unusual situation which I mentioned above, that this driver is needed because other drivers which need it (e.g. the network stack distributed as separate drivers) are being loaded post-end-of-DXE (e.g. by a bootloader) then it actually makes more sense for VirtioNetDxe to be loaded as a stand-alone driver at that time too.

ardbiesheuvel (approved):
@lgao4 @leiflindholm @mdkinney  @makubacki

This is a build fix that we should consider during the soft freeze IMHO

lgao4:
> @lgao4 @leiflindholm @mdkinney @makubacki
> 
> This is a build fix that we should consider during the soft freeze IMHO

I agree to merge it.

mikebeaton:
Not merged during soft freeze?

ardbiesheuvel:
> Not merged during soft freeze?

No, it appears this one slipped through the cracks - sorry about that.

mikebeaton:
How does push label work? I gets pushed shortly, as soon as someone adds it, but not during soft freeze? Or even during soft freeze, if someone adds it then?

ardbiesheuvel:
During the freeze, everything is handled manually by Liming (@lgao4)

mikebeaton:
I don't think we care too much in this case, as we have several other pending changes anyway. But just for future reference, had I wanted to chase this, once Liming had agreed to merge, as above, he or yourself could have added a push label and it would have gone in then?

leiflindholm:
> I don't think we care too much in this case, as we have several other pending changes anyway. But just for future reference, had I wanted to chase this, once Liming had agreed to merge, as above, he or yourself could have added a push label and it would have gone in then?

During code freeze, only Liming can add the push label, triggering the merge.

Once the default branch is unlocked (manually) after the release, anyone in the maintainers group can add the "push" label.

PR #11720, opened 2025-11-06
Fixes for gcc 15 / C23

osteffenrh:
# Description

Small fixes to allow building with GCC15 which defaults to C23 standard,
which has stricter warnings on type casts.

This is needed to use the new Fedora 43 images in the CI (for the Linux) builds.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Running builds locally.

### DIFF
--- a/MdeModulePkg/Universal/DebugSupportDxe/Ia32/PlDebugSupport.c
+++ b/MdeModulePkg/Universal/DebugSupportDxe/Ia32/PlDebugSupport.c
@@ -192,7 +192,7 @@ RegisterPeriodicCallback (
   IN EFI_PERIODIC_CALLBACK       PeriodicCallback
   )
 {
-  return ManageIdtEntryTable (PeriodicCallback, SYSTEM_TIMER_VECTOR);
+  return ManageIdtEntryTable ((CALLBACK_FUNC)PeriodicCallback, SYSTEM_TIMER_VECTOR);
 }
 
 /**
@@ -221,7 +221,7 @@ RegisterExceptionCallback (
   IN EFI_EXCEPTION_TYPE          ExceptionType
   )
 {
-  return ManageIdtEntryTable (ExceptionCallback, ExceptionType);
+  return ManageIdtEntryTable ((CALLBACK_FUNC)ExceptionCallback, ExceptionType);
 }
 
 /**


SecurityPkg/DeviceSecurity/SpdmSecurityLib/SpdmConnectionInit.c
```diff
@@ -222,8 +222,8 @@ CreateSpdmDeviceContext (
     SpdmDeviceInfo->MaxSpdmMsgSize,
     SpdmDeviceInfo->TransportHeaderSize,
     SpdmDeviceInfo->TransportTailSize,
-    SpdmDeviceInfo->TransportEncodeMessage,
-    SpdmDeviceInfo->TransportDecodeMessage
+    (libspdm_transport_encode_message_func)SpdmDeviceInfo->TransportEncodeMessage,
```
luigix25, line 225: What about this?

```
diff --git a/SecurityPkg/Include/Library/SpdmSecurityLib.h b/SecurityPkg/Include/Library/SpdmSecurityLib.h
index 96a7841381..36975f000b 100644
--- a/SecurityPkg/Include/Library/SpdmSecurityLib.h
+++ b/SecurityPkg/Include/Library/SpdmSecurityLib.h
@@ -129,9 +129,9 @@ typedef
   SPDM_RETURN
 (*SPDM_TRANSPORT_ENCODE_MESSAGE_FUNC)(
   IN VOID              *SpdmContext,
-  IN OUT CONST UINT32  *SessionId,
-  IN BOOLEAN           IsAppMessage,
-  IN BOOLEAN           IsRequester,
+  IN OUT CONST UINT32  SessionId,^M
+  IN bool              IsAppMessage,^M
+  IN bool              IsRequester,^M
   IN UINTN             MessageSize,
   IN OUT VOID          *Message,
   IN OUT UINTN         *TransportMessageSize,
```

In this way we don't need to add a cast here.

osteffenrh: I think we should not start mixing boolean types in public EDK2 header files.

A proper long-term fix could be to modernize EDK2s BOOLEAN to use bool (supported since C99), but that is unlikely to happen quickly, I guess.

luigix25: > I think we should not start mixing boolean types in public EDK2 header files.
> 
> A proper long-term fix could be to modernize EDK2s BOOLEAN to use bool (supported since C99), but that is unlikely to happen quickly, I guess.

It would be nice, but yes, It won't happen quickly. 

So yes, a cast is therefore needed in that case.

jyao1 (approved):

osteffenrh:
Included in https://github.com/tianocore/edk2/pull/11758. Closing.

PR #11721, opened 2025-11-06
Global: fix boot failure with LPA2

LeviYeoReum:
# Description

When FEAT_LPA2 is enabled:
  1. The DS bit is located at bit 32, not bit 59, when HCR_EL2.E2H = 0.
  2. The root level of the page table should be -1 when using a 52-bit address with a 4KB TxSZ and LPA2 feature is enabled.

However, the current implementation does not handle these conditions correctly, 
causing EFI to fail during boot when CPU is EL 0&1 or EL 0&2.

Fix this boot failure.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

When run FVP_BASE_RevC model, add below parameter for LPA2 with 4KB, PA 52bits feature:
  -C cluster0.has_52bit_address_with_4k=2
  -C cluster1.has_52bit_address_with_4k=2
  -C cluster0.PA_SIZE=52
  -C cluster1.PA_SIZE=52

and 
  -C cluster0.has_arm_v9-6=1 for E2H0 enabled (HCR_EL2.E2H alaways set as 1).

## Integration Instructions

NP

### DIFF
--- a/ArmPkg/Drivers/CpuDxe/AArch64/Mmu.c
+++ b/ArmPkg/Drivers/CpuDxe/AArch64/Mmu.c
@@ -48,7 +48,8 @@ GetRootTranslationTableInfo (
   OUT UINTN  *RootTableEntryCount
   )
 {
-  *RootTableLevel      = (INTN)(T0SZ - MIN_T0SZ) / BITS_PER_LEVEL;
+  *RootTableLevel = (T0SZ < MIN_T0SZ) ? -1 : (INTN)(T0SZ - MIN_T0SZ) / BITS_PER_LEVEL;
+  ASSERT (*RootTableLevel >= 0 || ArmLpa2Enabled ());
   *RootTableEntryCount = TT_ENTRY_COUNT >> (INTN)(T0SZ - MIN_T0SZ) % BITS_PER_LEVEL;
 }
 
@@ -241,7 +242,7 @@ GetNextEntryAttribute (
 
       // Increase the level number and scan the sub-level table
       GetNextEntryAttribute (
-        (UINT64 *)GetOutputAddress (Entry, ArmGetTCR () & TCR_DS),
+        (UINT64 *)GetOutputAddress (Entry, ArmLpa2Enabled ()),
         TT_ENTRY_COUNT,
         TableLevel + 1,
         (BaseAddress + (Index * TT_ADDRESS_AT_LEVEL (TableLevel))),
@@ -329,7 +330,7 @@ SyncCacheConfig (
   GetRootTranslationTableInfo (T0SZ, &TableLevel, &TableCount);
 
   // First Attribute of the Page Tables
-  PageAttribute = GetFirstPageAttribute (FirstLevelTableAddress, TableLevel, ArmGetTCR () & TCR_DS);
+  PageAttribute = GetFirstPageAttribute (FirstLevelTableAddress, TableLevel, ArmLpa2Enabled ());
 
   // We scan from the start of the memory map (ie: at the address 0x0)
   BaseAddressGcdRegion = 0x0;
@@ -458,7 +459,7 @@ GetMemoryRegionRec (
   EntryType  = *BlockEntry & TT_TYPE_MASK;
 
   if ((TableLevel < 3) && (EntryType == TT_TYPE_TABLE_ENTRY)) {
-    NextTranslationTable = (UINT64 *)GetOutputAddress (*BlockEntry, ArmGetTCR () & TCR_DS);
+    NextTranslationTable = (UINT64 *)GetOutputAddress (*BlockEntry, ArmLpa2Enabled ());
 
     // The entry is a page table, so we go to the next level
     Status = GetMemoryRegionRec (
--- a/UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
+++ b/UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
@@ -142,7 +142,12 @@ GetRootTableLevel (
   IN  UINTN  T0SZ
   )
 {
-  return (INTN)(T0SZ - MIN_T0SZ) / BITS_PER_LEVEL;
+  INTN  RootTableLevel;
+
+  RootTableLevel =  (T0SZ < MIN_T0SZ) ? -1 : (INTN)(T0SZ - MIN_T0SZ) / BITS_PER_LEVEL;
+  ASSERT (RootTableLevel >= 0 || ArmLpa2Enabled ());
+
+  return RootTableLevel;
 }
 
 STATIC
@@ -306,7 +311,7 @@ UpdateRegionMappingRecursive (
     // the MMU in order to update page table entries safely, so prefer page
     // mappings in that particular case.
     //
-    if ((Level == 0) || (((RegionStart | BlockEnd) & BlockMask) != 0) ||
+    if ((Level <= 0) || (((RegionStart | BlockEnd) & BlockMask) != 0) ||
         ((Level < 3) && (((UINT64)PageTable & ~BlockMask) == RegionStart)) ||
         IsTableEntry (*Entry, Level))
     {
@@ -616,7 +621,7 @@ ArmSetMemoryAttributes (
            PageAttributeMask,
            ArmGetTTBR0BaseAddress (),
            TRUE,
-           ArmGetTCR () & TCR_DS
+           ArmLpa2Enabled ()
            );
 }
 
@@ -680,7 +685,7 @@ ArmConfigureMmu (
     } else if (MaxAddress < SIZE_256TB) {
       TCR |= TCR_PS_256TB;
     } else if ((MaxAddress < SIZE_4PB) && ArmHas52BitTgran4 ()) {
-      TCR |= TCR_PS_4PB | TCR_DS;
+      TCR |= TCR_PS_4PB | TCR_DS_NVHE;
     } else {
       DEBUG ((
         DEBUG_ERROR,
@@ -764,7 +769,7 @@ ArmConfigureMmu (
   ZeroMem (TranslationTable, RootTableEntryCount * sizeof (UINT64));
 
   while (MemoryTable->Length != 0) {
-    Status = FillTranslationTable (TranslationTable, MemoryTable, TCR & TCR_DS);
+    Status = FillTranslationTable (TranslationTable, MemoryTable, ArmLpa2Enabled ());
     if (EFI_ERROR (Status)) {
       goto FreeTranslationTable;
     }


ArmPkg/Drivers/CpuDxe/AArch64/Mmu.c
```diff
@@ -48,7 +48,8 @@ GetRootTranslationTableInfo (
   OUT UINTN  *RootTableEntryCount
   )
 {
-  *RootTableLevel      = (INTN)(T0SZ - MIN_T0SZ) / BITS_PER_LEVEL;
```
leiflindholm, line 51: Commit message:
Both existing comments and what I can find in the Arm ARM use `EL1&0` and `EL2&0`.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -736,6 +737,10 @@ ArmConfigureMmu (
   // Set TCR
   ArmSetTCR (TCR);
 
+  if ((MaxAddress >= SIZE_256TB) && !ArmLpa2Enabled ()) {
```
leiflindholm: Fold this in at original lines 683/711 instead. Probably want to upgrade it to DEBUG_ERROR as well. We should probably look at untangling the whole 665-721 block (or just breaking it out into a helper), but doesn't have to be part of this PR.

This does feel completely unrelated to the overall PR and commit description though. It should probably be broken out as a separate commit.

MdePkg/Include/AArch64/AArch64Mmu.h
```diff
@@ -129,7 +129,8 @@
 #define TTBR_ASID_MASK   (0xFF << TTBR_ASID_FIELD)
 #define TTBR_BADDR_MASK  (0xFFFFFFFFFFFF )                     // The width of this field depends on the values in TxSZ. Addr occupies bottom 48bits
 
-#define TCR_DS  (1UL << 59)
+#define TCR_DS       (1UL << 59)
```
leiflindholm, line 132: Please break these additions, as well as the added helper function, out into its own commit.

leiflindholm (changes requested):
Does this fix a breakage introduced since August stable tag, and hence should be included in November tag currently underway?

ArmPkg/Drivers/CpuDxe/AArch64/Mmu.c
LeviYeoReum: Oh Thank. I'll change as "and" not "or".

leiflindholm: Well, I mainly meant `EL1&0` as opposed to `EL 0&1` form.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
LeviYeoReum: Hmm. I'm not sure it could be fold on 683/711 since 
This is situation though TCR should set the "DS" bit but it isn't set.
I observed this case when model is run without -C has_arm_v8-1=1 or above option.

I agree upgrade this with DEBUG_ERROR. but folding mean it should move all ArmSetTCR() in each line
and it seems to add code duplication.

Am I missing something?

ArmPkg/Drivers/CpuDxe/AArch64/Mmu.c
LeviYeoReum: Oh. I see. let me modify them.

LeviYeoReum:
> Does this fix a breakage introduced since August stable tag, 

Yes.

> and hence should be included in November tag currently underway?

Personally It would be good to be added but since many user seems doesn't use FEAT_LPA2,
I thiink it would be merged later. 

@samimujawar, What do you think?

MdePkg/Include/AArch64/AArch64Mmu.h
LeviYeoReum: Okay. I'll separate. Thanks!

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
leiflindholm: We have a duplicated (eew) set of valid size checks for either outcome of the TranslationRegimeIsDual () test.
With your fix above, we correctly set DS_NVHE or DS dependent on state of HCR_EL2.E2H.
What are you testing for here that isn't already being tested for above?

We don't need to read the value we wrote into TCR back in order to know if we have a valid configuration or not. If we do, we're missing other checks earlier.

LeviYeoReum: > What are you testing for here that isn't already being tested for above?

I believe former test with the model running without `-C has_arm_v8-1=1 or above option` used widely
-- in this case the model runs with armv8.0 extension where TCR_EL2's high 32bits doesn't implemented.
So, top 32 bits of TCR_EL2 are ignored.

IOW, though "TCR variable" has DS bit on 32nd bit, but it doesn't set.

The reason why former version works when DS bit is cleared, page table walks start with level 0.
(and the former version only set the page level start from 0).

In contrast, when hash_arm_v8-1=1 or E2H0 is enabled (E2H bit always set)
TCR_EL2 DS bits set properly (and I believe these cases aren't tested at that time).
Since this case runs page table walks with level -1, former patch doesn't work.

> We don't need to read the value we wrote into TCR back in order to know if we have a valid configuration or not. If we do, we're missing other checks earlier.

Agree, but according to model option, as I said above, something doesn't set properly.
(that's why I've added some WARN in the patch).
But I'm not sure whether make it ASSSERT(0) or let the run with WARN message.
(otherwise, it's "REALLY" hard to catch....).

leiflindholm: > Agree, but according to model option, as I said above, something doesn't set properly.

Something doesn't set properly because we haven't properly identified the current state in software, which we could.

If we properly identify that, we don't need to verify whether the write "takes".

The place to properly identify that feels to me like it should be where we're already performing one of the two tests that _this_ patch performs in this location. I.e. original lines 683/711.

What I want is to eliminate the possibility of reaching original line 722 with an invalid configuration. Not check at a later point whether we did.

LeviYeoReum: > Something doesn't set properly because we haven't properly identified the current state in software, which we could.

Well, I don't think so. since the "DS" bit in `TCR_EL2` is added in `FEAT_LPA2` and 
it already checked properly by `ArmHas52BitTgran4()`.

But I think this is my mistake.
Since if model doesn't has a option `has_armv8_6=1` or above, the TCR_EL2's DS bit is RES0
So, although try to set the "TCR_DS", remain as "RES0" is right behaviour.
In this case, the ArmLpa2Enabled() is always properly handle this case, this pointless check can be removed.

Thanks!

ArmPkg/Drivers/CpuDxe/AArch64/Mmu.c
```diff
@@ -48,7 +48,8 @@ GetRootTranslationTableInfo (
   OUT UINTN  *RootTableEntryCount
   )
 {
-  *RootTableLevel      = (INTN)(T0SZ - MIN_T0SZ) / BITS_PER_LEVEL;
+  *RootTableLevel = ((T0SZ < MIN_T0SZ) && ArmLpa2Enabled ()) ?
+                    -1 : (INTN)(T0SZ - MIN_T0SZ) / BITS_PER_LEVEL;
```
ardbiesheuvel: This should rely only on `T0SZ` - the fact that LPA2 needs to be enabled in order for level `-1` to be permitted is a separate matter, and returning `0` instead does not really help.

So better do something like

```
*RootTableLevel = (T0SZ < MIN_T0SZ) ? -1 : (INTN)(T0SZ - MIN_T0SZ) / BITS_PER_LEVEL;
ASSERT (*RootTableLevel >= 0 || ArmLpa2Enabled ());
```

LeviYeoReum: Hi Ard,
There is on corner case.
When LPA2 feature enable, but DS but is not set (!ArmLpa2Enabled()), the page table level started from 0.

Your ASSERT() is rational but, the change of root level seems break above case.

ardbiesheuvel: > Hi Ard, There is on corner case. When LPA2 feature enable, but DS but is not set (!ArmLpa2Enabled()), the page table level started from 0.
> 
> Your ASSERT() is rational but, the change of root level seems break above case.

That case will be broken anyway. Setting `TCR.T0SZ` to a value < 16 when `DS` unset is going to crash the system anyway, so whether the rootlevel is reported as `0` or `1` makes no difference whatsoever.

LeviYeoReum: > That case will be broken anyway. Setting TCR.T0SZ to a value < 16 when DS unset is going to crash the system anyway, so whether the rootlevel is reported as 0 or 1 makes no difference whatsoever.

Nope. When you check in the AARCH64.FullTranslate(), when TCR.T0SZ sets 12 but if DS is 0, 
page translation works as with 4 level page table works (since effective T0SZ becomse "16").

As the example, when model runs with FEAT_LPA2 option without has.armv8_6, it works with 4 level page tables.
even, This is the way current code works.

ardbiesheuvel: Even if the architecture is tolerant to misprogramming errors, that does not mean we should do the same. So at the very least, I want the DEBUG() build to produce an ASSERT() here when using a `T0SZ` value that is not supported when `DS == 0`

LeviYeoReum: > Even if the architecture is tolerant to misprogramming errors, that does not mean we should do the same. So at the very least, I want the DEBUG() build to produce an ASSERT() here when using a T0SZ value that is not supported when DS == 0

Okay. and runs FEAT_LPA2 without armv8_6 extension seems broken setup for model itself too.
I'll change as your suggestion.

Thanks.

leiflindholm:
@LeviYeoReum could you add a `Fixes: 90771630bfe` to commit message of 2/2?

UefiCpuPkg/Include/Library/ArmMmuLib.h
```diff
@@ -71,4 +71,9 @@ ArmSetMemoryAttributes (
   IN UINT64                AttributeMask
   );
 
+BOOLEAN
```
lgao4, line 82: Function header is requried to be added.

LeviYeoReum: Thanks. I'll add the comments.

LeviYeoReum:
> @LeviYeoReum could you add a `Fixes: 90771630bfe` to commit message of 2/2?

Thanks. :) I've added

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -811,6 +816,20 @@ ArmConfigureMmu (
   return Status;
 }
 
+BOOLEAN
+ArmLpa2Enabled (
```
lgao4, line 827: Please also copy the header info here.

LeviYeoReum: Thanks. Done

PR #11722, opened 2025-11-06
MdeModulePkg/NonDiscoverablePciDeviceDxe: Avoid uncached bounce buffers

ardbiesheuvel:
The current handling of non-coherent DMA is very inefficient:

- outbound DMA uses bounce buffers unnecessarily, considering that a bus master read cannot break coherency, and can therefore access the host buffer directly without the risk of data corruption;

- bounce buffers are allocated from uncached memory, resulting in slow CopyMem() performance, and unnecessary overhead in allocating, remapping and freeing the pages, which also involves EFI memory map changes each time.

So refactor the logic, and tweak it so that:

- Bounce buffering is only performed when the host buffer is out of DMA range for the device, or when performing misaligned inbound DMA on a host buffer that is mapped cacheable;

- Cache maintenance is only performed when the buffer presented to the device is mapped cacheable.

- Bounce buffers are pool allocated, which is much more efficient and does not involve EFI memory map changes in the common case.

### DIFF
--- a/MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceIo.c
+++ b/MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceIo.c
@@ -983,6 +983,7 @@ NonCoherentPciIoFreeBuffer (
   // Find the uncached allocation list entry associated
   // with this allocation
   //
+  EfiAcquireLock (&Dev->UncachedAllocationListLock);
   for (Entry = Dev->UncachedAllocationList.ForwardLink;
        Entry != &Dev->UncachedAllocationList;
        Entry = Entry->ForwardLink)
@@ -1010,7 +1011,8 @@ NonCoherentPciIoFreeBuffer (
 
   if (!Found) {
     ASSERT_EFI_ERROR (EFI_NOT_FOUND);
-    return EFI_NOT_FOUND;
+    Status = EFI_NOT_FOUND;
+    goto ReleaseLock;
   }
 
   EndPages = Alloc->NumPages - (Pages + StartPages);
@@ -1018,7 +1020,8 @@ NonCoherentPciIoFreeBuffer (
   if (StartPages != 0) {
     AllocHead = AllocatePool (sizeof *AllocHead);
     if (AllocHead == NULL) {
-      return EFI_OUT_OF_RESOURCES;
+      Status = EFI_OUT_OF_RESOURCES;
+      goto ReleaseLock;
     }
 
     AllocHead->HostAddress = Alloc->HostAddress;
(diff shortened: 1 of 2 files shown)


ardbiesheuvel:
@eeshanl

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceIo.c
```diff
@@ -1288,113 +1289,130 @@ NonCoherentPciIoMap (
...
-        break;
-
-      default:
-        ASSERT (FALSE);
-    }
+  //
+  // Check whether the host address refers to an uncached mapping. This is
+  // required for common buffer mappings.
+  //
+  Status = gDS->GetMemorySpaceDescriptor (
```
os-d: I know this was doing this before, but it is (unfortunately) unreliable to rely on the GCD for memory attributes. Consumers often bypass the GCD with the CpuArch protocol (or memory attributes protocol). E.g. there is a smattering of usage across edk2-platforms that is setting cache attributes directly via the CpuArch protocol: https://github.com/search?q=repo%3Atianocore%2Fedk2-platforms%20setmemoryattributes&type=code.

To be more reliable, this would need to use the memory attribute protocol to get the actually set attributes.

That being said, it can be said to be a programming error to not have the GCD updated, so up to you on how you want to handle it.

ardbiesheuvel: Thanks for taking a look.

> I know this was doing this before, but it is (unfortunately) unreliable to rely on the GCD for memory attributes. Consumers often bypass the GCD with the CpuArch protocol (or memory attributes protocol). E.g. there is a smattering of usage across edk2-platforms that is setting cache attributes directly via the CpuArch protocol: https://github.com/search?q=repo%3Atianocore%2Fedk2-platforms%20setmemoryattributes&type=code.
> 

Agree with this in principle. However, in this particular case, the question being answered is essentially whether or not the buffer in question was allocated using `NonCoherentPciIoAllocateBuffer ()`, and so the GCD should be in sync. Note that the whole point of this driver is being able to use generic ?HCI drivers with host controllers that are not PCI based, and these drivers use the `PciIo` abstraction without reasoning about coherency at all.

os-d: I think that's reasonable to say consumers need to rely on the `PciIo` abstraction

os-d: Github won't let me resolve the comment :)

PR #11724, opened 2025-11-07
MdeModulePkg: Bds Fix potential infinite loop.

apop5:
Bds will get the List of Boot Options and attempt to boot.
If there were no valid load options, the system could
infinitely be stuck in the attempt boot loop.

Add NULL checks to prevent a protentional infinite
boot retry loop.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
CI

## Integration Instructions
No integration necessary.


MdeModulePkg/Universal/BdsDxe/BdsEntry.c
```diff
@@ -1094,6 +1094,7 @@ BdsEntry (
       //
       // Retry to boot if any of the boot succeeds
       //
+      BootSuccess = FALSE;
```
leiflindholm, line 1100: Oh, err, I don't think this change makes sense without the `if LoadOptions != NULL)` check below. So please include that too.

apop5: Yes, I'll do that.

The commit messages need fixed up anyway.

ardbiesheuvel: This breaks the build on Clang, as it thinks the `if (!BootSuccess)` test on line 1113 is reachable without `BootSuccess` ever having been initialized, and as far as I can tell, this is accurate.

leiflindholm: I think you're right.
Must have got confused by the 72 layers of indentation when reviewing.
I guess the quick fix is to reinstate the `BootSuccess = FALSE;` 300 lines up.

Alternatively, refactoring this function into something readable using several helper functions.

apop5: Two PRs are up for these build breaks:

https://github.com/tianocore/edk2/pull/11788

https://github.com/tianocore/edk2/pull/11803

Issue #11725, opened 2025-11-08
OvmfPkg `build.sh` has unnecessarily removed support for still-supported OvmfPkgIa32X64.dsc


mikebeaton:
https://github.com/tianocore/edk2/pull/11559 should not have been accepted as is.

The OvmfPkg `build.sh` helper script previously allowed building OvmfPkgIa32.dsc, OvmfPkgIa32X64.dsc, OvmfPkgX64.dsc.

OvmfPkgIa32.dsc has now been removed, but OvmfPkgIa32X64.dsc has not, and I believe there is no good reason to remove `build.sh` support for it while it is still there.

mikebeaton:
@kraxel - I would humbly disagree with:

> Hmm, all CI passes, so apparently the script is not used for the CI builds. Guess we can update the build.sh script then, so people stop using OvmfPkgIa32X64.dsc (CI is really the only reason for this to exist, everything else should move on and use OvmfPkgX64.dsc instead).

As per this commit (reverting another change) https://github.com/tianocore/edk2/commit/4d196352f35ac516b477e568265b4e537b0283d8, `build.sh` remains a convenient tool to build then run the different Ovmf packages.

i.e. `./build.sh [build options]` or `./build.sh -a X64 [build options]` to build OvmfPkgX64 (we can also specify build target -b and toolchain -t); followed by `./build.sh [-b ... -t ...] qemu [qemu options]` to run the matching build.

Similarly `./build.sh -a IA32 -a X64 [build options]` to build OvmfPkgIa32X64 (we can also specify build target -b and toolchain -t); followed by `./build.sh -a IA32 -a X64 [-b ... -t ...] qemu [qemu options]` to run it.

I can't see any good reason to remove support for this.

Issue #9413, opened 2023-08-20
Recent OVMF build.sh change breaks useful functionality (Bugzilla Bug 4528)


tianocore-issues:
This issue was created automatically with [bugzilla2github](https://github.com/makubacki/bugzilla2github/tree/bz2gh_tianocore)
# Bugzilla Bug 4528
Date: 2023-08-20T07:34:57+00:00
From: Mike Beaton &lt;<mjsbeaton>&gt;
To:   unassigned &lt;<unassigned>&gt;
CC:   @lgao4, @lersek, @bexcran

Last updated: 2023-10-26T16:10:26+00:00

tianocore-issues:
## Comment 21838
Date: 2023-08-20 07:34:57 +0000
From: Mike Beaton &lt;<mjsbeaton>&gt;

- Industry Specification: ---
- Release Observed: ['EDK II Master', 'edk2-stable202305']
- Releases to Fix: EDK II Master
- Target OS: ---
- Bugzilla Assignee(s): unassigned &lt;<unassigned>&gt;

Prior to  https://github.com/tianocore/edk2/commit/173a7a7daaad560cd69e1000faca1d2b91774c46  :

`./build.sh -a X64 -b RELEASE` (or whichever arch and build target are required) built OVMF

then:

`./build.sh -a X64 -b RELEASE qemu {my-qemu-flags}` ran OVMF, using the correct qemu executable and built binary for the arch and target (failing if this had not yet been built, e.g. by the previous command).

The recent commit makes it so that every use of `./build.sh ... qemu ...` rebuilds, before running qemu with the built binary.

This is a problem because:

a) It wastes some unnecessary time
b) It trashes (i.e. rebuilds) the NVRAM contents of the virtual machine, each time it is re-run in this way

I believe the previous behaviour was almost certainly intentional, and useful as it was.

FWIW, it is also documented as it was in multiple places on the web ... including in the README for OvmfPkg itself: https://github.com/tianocore/edk2/tree/master/OvmfPkg . I would _strongly_ vote for restoring the previous behaviour, rather than updating the README.

tianocore-issues:
## Comment 22011
Date: 2023-10-18 23:10:16 +0000
From: @lgao4

Add the change owner rebecca@bsdio.co for the comments.

tianocore-issues:
## Comment 22064
Date: 2023-10-26 16:10:26 +0000
From: @lersek

[...]

PR #11726, opened 2025-11-08
OvmfPkg: Prevent ./build.sh from building X64 when legacy IA32 is specified

mikebeaton:
# Description

1fb88ffe284782cc79e306306b8d19829b6248b7 removed OVMF IA32, but unfortunately left the legacy flag `-a IA32` which previously triggered that build in `OvmfPkg/build.sh` as building X64 instead (even though `-a IA32 -a X64` to build IA32X64 still worked correctly).

This was followed by 2a6708a786eb0f0e9b979cd93c5dc12c42e337a1, which removed `OvmfPkg/build.sh` support for the `-a IA32` flag altogether, fixing the problem introduced by 1fb88ffe284782cc79e306306b8d19829b6248b7, but also removing support for IA32X64.

As per https://github.com/tianocore/edk2/issues/11725, there should be no need to remove IA32X64 support from `build.sh`. This first commit here reverts 2a6708a786eb0f0e9b979cd93c5dc12c42e337a1 to restore support for building IA32X64. The second commit fixes the mentioned issue with 1fb88ffe284782cc79e306306b8d19829b6248b7.

- [ ] Breaking change?
  - NO.
- [ ] Impacts security?
  - NO.
- [ ] Includes tests?
  - NO.

## How This Was Tested

Confirm builds supported packages, and refuses to build no longer supported package.

## Integration Instructions

N/A

### DIFF
--- a/OvmfPkg/build.sh
+++ b/OvmfPkg/build.sh
@@ -153,6 +153,10 @@ if [[ "$ARCH_IA32" == "yes" && "$ARCH_X64" == "yes" ]]; then
   BUILD_OPTIONS="$BUILD_OPTIONS -a IA32 -a X64"
   PLATFORM_BUILD_DIR=Ovmf3264
   BUILD_ROOT_ARCH=X64
+elif [[ "$ARCH_IA32" == "yes" && "$ARCH_X64" == "no" ]]; then
+    echo "Unsupported architecture '-a IA32'"
+    echo "Only '-a IA32 -a X64' or '-a X64' are supported"
+    exit 1
 else
   PROCESSOR=X64
   Processor=X64


kraxel:
There also is no reason to keep IA32X64 support in `build.sh`.  The only reason we still have this mixed mode build is being able to run CI builds (which no not need build.sh), so IA32 PEI support does not regress unnoticed.  For running virtual machines it has zero relevance.

mikebeaton:
It doesn't have zero relevance.

As I mentioned in [the related Issue to this PR](https://github.com/tianocore/edk2/issues/11725), in an earlier case I [requested a reversion of an earlier change to this file](https://github.com/tianocore/edk2/pull/4791) as it affected a usage I make of it, and which I believe it was originally intended to have.

In that previous case my argument that this usage should continue to be supported was accepted.

Specifically, I can build supported OVMF packages with commands such as:
```
$./build.sh -b NOOPT -D DEBUG_ON_SERIAL_PORT
$./build.sh -a X64 -t GCC -D TMP2_ENABLE=0
$./build.sh -a IA32 -a X64 -D NETWORK_ENABLE=0
```
and then I can run (and re-run, without rebuilding) the matching builds with:
```
$./build.sh -b NOOPT qemu -gdb tcp::8864 -serial stdio [additional qemu opts]
$./build.sh -a X64 -t GCC qemu [additional qemu opts]
$./build.sh -a IA32 -a X64 qemu [additional qemu opts]
```

I don't see why this feature, which still works, needs to be removed.

Link to (originally) bugzilla bug report for the earlier issue: https://github.com/tianocore/edk2/issues/9413

mikebeaton:
I should perhaps clarify that I work on a project which continues to support very old machines, and supports running legacy OSes in VMs. We maintain a working fork of DuetPkg, for instance(!). But we rebase regularly off EDK 2, and we (at least I) do still use the feature mentioned here - as we still want to run VMs in mixed mode. (Actually we still want to run VMs in 32-bit mode too - but that means we're going to need to maintain our own fork of OvmfPkgIa32, moving forwards.)

So again, however niche, I can't see the need to _remove_ this still working feature, especially as we are offering a fix for the only problem that I am aware of with how it has currently ended up.

kraxel:
I do not want give the impression that running the OvmfPkgIa32X64 port makes any sense, because it doesn't.

S3 not being fully supported with 64-bit PEI used to be a problem.  That was a good reason to prefer  OvmfPkgIa32X64 over OvmfPkfX64, but that has been finally fixed more two years ago (see commit 098c55702318fd907de6fad7b43b5e9a6ad9ff7f).

Not sure why you insist on continuing to use the mixed mode build.  If you /really/ want you still can, you only have to run the BaseTools build script directly.

mikebeaton:
> I do not want give the impression that running the OvmfPkgIa32X64 port makes any sense, because it doesn't.

I genuinely don't understand why it doesn't, when it's still there and it still runs?

I mean, surely this is a _developer_ script, more than an end user script? If a developer still wants to look at, examine and run OvmfPkgIa32X64, the way they used to be able to, shouldn't they still be able to? Is this the right place to make changes that are (valid) recommendations of what users of the package(s) should do, if they only want to run stuff within in? (I also can't image that any user wanting to run 64 bit efi or OS code within OVMF is going to think, "hmm, I wonder whether `-a IA32 -a X64` might be the best way to get what I want"!) It's also quite unobvious what the right way is to build OvmfPkgIa32X64 (while it's still there), without this being implicitly documented in this script.

ardbiesheuvel:
> > I do not want give the impression that running the OvmfPkgIa32X64 port makes any sense, because it doesn't.
> 
> I genuinely don't understand why it doesn't, when it's still there and it still runs?
> 

Because all platforms capable of running the IA32X64 build can run the X64 build as well. At this point, IA32X64 is only retained for CI coverage of 32-bit PEI, which still might be in use in downstream platforms, given that none of the original reasons for having this variant of OVMF still apply. 

32-bit support is going to be removed in is entirety after the next (not the upcoming) stable tag.

mikebeaton:
Then wouldn't that be the time to remove it from this script? 🤷‍♂️

mikebeaton:
> 32-bit support is going to be removed in is entirety after the next (not the upcoming) stable tag.

Still, I guess I should take approval of https://github.com/tianocore/edk2/commit/2a6708a786eb0f0e9b979cd93c5dc12c42e337a1 and declining this PR as advance notice of the above. :)

mikebeaton:
> 32-bit support is going to be removed in is entirety after the next (not the upcoming) stable tag.

@kraxel - I couldn't find any details for the above in https://edk2.groups.io/g/devel/topic/rfc_remove_ovmf_ia32_and/114152215 - at least, not unless 'after' just means 'at some point after, possibly several more stable tags', in which case yes.

ardbiesheuvel:
> > 32-bit support is going to be removed in is entirety after the next (not the upcoming) stable tag.
> 
> @kraxel - I couldn't find any details for the above in https://edk2.groups.io/g/devel/topic/rfc_remove_ovmf_ia32_and/114152215 - at least, not unless 'after' just means 'at some point after, possibly several more stable tags', in which case yes.

This was decided by the Tianocore stewards at the UEFI Plugfest last month. The upcoming stable tag release notes will contain a mention that all 32-bit support is considered obsolete, and will be removed right after the next stable tag (202602) is released (unless there is [motivated] pushback from the community)

@lgao4 @mdkinney @leiflindholm @makubacki @ajfish

PR #11729, opened 2025-11-08
MdeModulePkg/BrotliCustomDecompressLib: Correct BrotliDecompress

mikebeaton:
# Description

This PR fixes a relatively minor but genuine code bug identified by XCODE5 build.

- [ ] Breaking change?
  - NO.
- [ ] Impacts security?
  - NO.
- [ ] Includes tests?
  - NO.

## How This Was Tested

Confirm valid XCODE5 build after changes. Confirm no break in EDK 2 CI.

## Integration Instructions

N/A


leiflindholm:
Presumably this bug was hidden from other toolchains by the otherwise pointless initialization to 0 on line 279.

However, `BrotliUefiDecompress ()` is the only caller of the library-internal `BrotliDecompress ()`, and neither of them make any use of the value stored.
Is this just some weird copy-paste error via `BrotliUefiDecompressGetInfo ()`?
It seems the variable, and the argument, could be dropped outright.

mikebeaton:
@leiflindholm - Looking at the code, IMO the intention may have been to provide something which could be called elsewhere. However it is not declared in any header file and never is called elsewhere. So I've updated the PR to remove the broken parameter completely, and marked the method STATIC and renamed it as ...Internal.

mikebeaton:
@ardbiesheuvel - Not sure whether to ping yourself or Liming better? This one should be okay to push.

mikebeaton:
Same as others, don't think I can change it?

mikebeaton:
@mergifyio requeue

PR #11730, opened 2025-11-08
SecurityPkg: Update libspdm to 3.7.0

mikebeaton:
# Description

https://github.com/DMTF/libspdm/commit/0f6c6a3e800f487242b973b5858c534186a7db05 fixes incorrectly typed return values in `libspdm_hmac_new`.

Without this commit XCODE5 refuses to build SecurityPkg, with multiple errors corresponding to the lines fixed above, each of the form:

libspdm_crypt_hmac.c:31:16: error: expression which evaluates to zero treated as a null pointer constant of type 'void *'

I have updated libspdm to the minimum whole build number which includes the required commit, though higher releases than 3.7.0, up to 3.8.1, are available.

- [ ] Breaking change?
  - NO.
- [ ] Impacts security?
  - NO.
- [ ] Includes tests?
  - NO.

## How This Was Tested

Confirm successful build of SecurityPkg with XCODE5 after change. Confirm EDK 2 CI is not broken.

## Integration Instructions

N/A


mikebeaton:
Note: The same issue affects CLANGDWARF and CLANGPDB with [slightly more aggressive warnings enabled](https://github.com/tianocore/edk2/pull/11733/commits), which it would be realistic to do after this and the other small recent cleanups which I've submitted are merged. (Although EDK 2 CI doesn't currently run CLANGDWARF or CLANGPDB CI, Acidanthera AUDK CI does, so I can confirm that with the upgraded warnings these toolchains succeed or fail on basically the same files as XCODE5.)

@leiflindholm - cf https://github.com/tianocore/edk2/pull/11729#issuecomment-3506445390

mikebeaton:
@ardbiesheuvel - Also here, not sure whether to ping yourself or Liming better? This one should be okay to push too.

ardbiesheuvel:
@makubacki @mdkinney @leiflindholm this seems to be a glitch with how we handle the permissions during the freeze: Jiewen's approval is no longer considered sufficient, even though he is an owner of SecurityPkg.

makubacki:
> @makubacki @mdkinney @leiflindholm this seems to be a glitch with how we handle the permissions during the freeze: Jiewen's approval is no longer considered sufficient, even though he is an owner of SecurityPkg.

With the 2511 stable tag completed, I've restored write permission to maintainers. Jiewen's approval meets the requirement now.

ardbiesheuvel:
Does that apply to mergify too? I guess that is why all the merges were failing?

mikebeaton:
@jyao1 - Hi! Would it be okay to give this a push label?

makubacki:
> Does that apply to mergify too? I guess that is why all the merges were failing?

If you're referencing the mergify issue I think you are, that was because commit https://github.com/tianocore/edk2/commit/96158b339a1c8937d065bdfdfeeaf168f1ccca4c modified a file in the `.github/workflows` directory. Because that directory has code that is directly executed on a build agent, GitHub assigns it special permissions. Mergify does not have workflow write permission, so when it rebases that change onto older branches, it was blocked. We could consider giving it that permission, but this is relatively rare and there would be some security considerations to work through.

PR #11740, opened 2025-11-10
Prevent VariablePei Memory Bucket Fragmentation

### LINKED ISSUE #11739: [Bug]: VariablePei allocates runtime memory fragmenting memory buckets

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

MdeModulePkg

DEBUG, RELEASE

### Current Behavior

Commit d8f513de3e3ef228af7e6facf0ad3e35c3224032 updated [MdeModulePkg/Universal/VariablePei/Variable.c](https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Universal/Variable/Pei/Variable.c) to allocate the UEFI variable runtime cache into runtime pages. This runtime allocation does not get consolidated into the runtime memory buckets that are allocated by the DXE Core and stay fragmented throughout boot which impacts hibernate (S4) stability.

### Expected Behavior

All runtime allocations for a given runtime memory type should be consolidated in a single memory bucket for that memory type.

### Steps To Reproduce

Inspect the runtime memory allocations at End of DXE or Exit Boot Services to see that the UEFI runtime cache range remains separated from the DXE-allocated memory bucket.

### Build Environment

```markdown
N/A
```

### Version Information

```text
Commit: d8f513d+
```

### Urgency

Medium

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

makubacki:
# Description

Fixes #11739 

---

**MdeModulePkg/Variable: Move RT cache buffer allocation to DXE**

Commit `d8f513de3e3ef228af7e6facf0ad3e35c3224032` added runtime
memory allocation to VariablePei. That will cause runtime memory
bucket fragmentation affecting hibernate stability as these runtime
allocations are separate from the main runtime memory type
bucket allocation made by the DXE Core.

To preserve the existing functionality but allow the runtime buffer
to come from the DXE bucket, the allocations in PEI are changed to
boot services data allocations and the contents are moved to the
runtime memory buffers (and those pages unblocked to MM) in the
DXE variable driver after the bucket has been allocated by the DXE
core.

---

**MdeModulePkg/VariableSmmRuntimeDxe: Disable var RT cache on alloc failure**

If a RT data buffer fails to be allocated or unblocked, clean up the
RT cache state by (1) freeing any allocated buffers and (2) resetting
the buffer addresses back to zero to indicate the RT cache is not
available.

---

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- Boot with the UEFI variable runtime cache enabled on QEMU and a physical X64 platform

## Integration Instructions

- M/A

### DIFF
--- a/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
+++ b/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
@@ -1622,6 +1622,21 @@ InitVariableStoreHeader (
 }
 
 /**
+  Invalidates all runtime cache pointers to prevent access to boot services memory.
+
+**/
+VOID
+InvalidateAllRuntimeCaches (
+  VOID
+  )
+{
+  mVariableRtCacheInfo.CacheInfoFlagBuffer        = 0;
+  mVariableRtCacheInfo.RuntimeHobCacheBuffer      = 0;
+  mVariableRtCacheInfo.RuntimeNvCacheBuffer       = 0;
+  mVariableRtCacheInfo.RuntimeVolatileCacheBuffer = 0;
+}
+
+/**
   Initialize the runtime variable cache related content.
 
   @param[in]  RuntimeCacheInfoGuidHob     Pointer to the gEdkiiVariableRuntimeCacheInfoHobGuid HOB.
@@ -1644,8 +1659,14 @@ InitVariableCache (
   UINTN                        AllocatedVolatileCacheSize;
   VARIABLE_RUNTIME_CACHE_INFO  *VariableRuntimeCacheInfo;
   VOID                         *RuntimeBuffer;
-  EFI_PHYSICAL_ADDRESS         RuntimeBufferPhysicalAddress;
   UINTN                        Pages;
+
+  // Storage for temporary allocations - only committed to mVariableRtCacheInfo if all succeed
+  EFI_PHYSICAL_ADDRESS  TempCacheInfoBuffer             = 0;
+  EFI_PHYSICAL_ADDRESS  TempHobCacheBuffer              = 0;
+  EFI_PHYSICAL_ADDRESS  TempNvCacheBuffer               = 0;
+  EFI_PHYSICAL_ADDRESS  TempVolatileCacheBuffer         = 0;
+  BOOLEAN               AllRtBufferAllocationsSucceeded = TRUE;
 
   //
   // Get needed runtime cache buffer size and check if auth variables are to be used from SMM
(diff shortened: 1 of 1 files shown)


makubacki:
@lgao4 & @td36, I posted this PR to raise the issue and show a possible approach to address the problem.

Can you please clarify the requirements that are needed for the runtime cache allocation in the variable PEIM? If the runtime type memory allocation is required (cannot be done in DXE as proposed here), can you describe that use case? Then, we can determine what alternatives may be viable that support it and can prevent memory bucket fragmentation.

makubacki:
@lgao4 & @td36 , please provide feedback on the change. I'd just like you to check if this direction is acceptable to you or if you would like to propose an alternative.

lgao4:
> @lgao4 & @td36 , please provide feedback on the change. I'd just like you to check if this direction is acceptable to you or if you would like to propose an alternative.

I don't think it is necessary to allocate runtime memory for variable in pei phase. I agree your proposal to migrate them in DXE phase from  boot memory to runtime memory. 

@niruiyu , have you any suggestion for this PR?

niruiyu:
> @lgao4 & @td36, I posted this PR to raise the issue and show a possible approach to address the problem.
> 
> Can you please clarify the requirements that are needed for the runtime cache allocation in the variable PEIM? If the runtime type memory allocation is required (cannot be done in DXE as proposed here), can you describe that use case? Then, we can determine what alternatives may be viable that support it and can prevent memory bucket fragmentation.

@makubacki The edk2 standalone MM does not accept Unblock request after it has been launched. A unblock request in PEI is translated to a GUIDed HOB and passed to standalone MM env for paging setup. A unblock request in DXE is ignored (actually there is no DXE version of Unblock implementation in edk2). The purpose of not accepting Unblock in DXE is for better security.

niruiyu:
> > @lgao4 & @td36 , please provide feedback on the change. I'd just like you to check if this direction is acceptable to you or if you would like to propose an alternative.
> 
> I don't think it is necessary to allocate runtime memory for variable in pei phase. I agree your proposal to migrate them in DXE phase from boot memory to runtime memory.
> 
> @niruiyu , have you any suggestion for this PR?

As I mentioned above, migration to DXE does not work according to current standalone MM design.

makubacki:
@niruiyu, I understand the edk2 Standalone MM design and it was the main reason for keeping this in draft to discuss a plan.

The requirement is the same as it has been since the concept of memory buckets was introduced and that is to keep runtime memory allocations stable across S4. Allocating runtime in memory in PEI prior to the DXE core establishing the buckets used for runtime memory allocations leads to fragmentation. What is the edk2 design answer for this?

makubacki:
I'm letting this stay stale. It is superseded by PEI memory bucket support in https://github.com/tianocore/edk2/pull/12086.

PR #11743, opened 2025-11-11
DynamicTablesPkg: Fix always true condition

vbpandya:
In "if (RpInfo[Index].Sun != MAX_UINT64)", "RpInfo[Index].Sun" is UINT32 type.
Which means it is never going to be equal to MAX_UINT64, so use "RpInfo[Index].Sun != MAX_UINT32" instead.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This issue causes an error while building with clang, with this commit there is no error.

## Integration Instructions

N/A


pierregondois:
Hello Varshit,
The changes seems correct, but would it be possible to remove the 'Merge' commit ?
I.e. just rebasing on the latest master

vbpandya:
> Hello Varshit, The changes seems correct, but would it be possible to remove the 'Merge' commit ? I.e. just rebasing on the latest master

Apologies, didn't mean to do that, I guess I don't understand the rebase from GUI yet :D

PR #11746, opened 2025-11-12
BaseTools: Fix check and error message by moving the use of the variable

### LINKED ISSUE #11744: [Bug]: <title>toolsetup.bat is trying to use delayed expansion incorrectly.

### Bug Type

- [ ] Firmware
- [x] Tool
- [ ] Unit Test

BaseTools

_No response_

### Current Behavior

I downloaded a fresh copy of the EDK2 repo and the submodules, then I tried to run edksetup.bat and I get the following error: 

Image

After looking into edksetup.bat, I realized the error was coming from BaseTools\toolsetup.bat. 

I narrowed the error down to this massive nested if statement. 
` if not defined EDK_TOOLS_BIN (
    if defined BASETOOLS_MINGW_BUILD (
      if /I "%MINGW_ARCH%"=="X64" (
        set EDK_TOOLS_BIN=%EDK_TOOLS_PATH%\Bin\Win64
      ) else (
        if /I "%MINGW_ARCH%"=="AARCH64" (
          set EDK_TOOLS_BIN=%EDK_TOOLS_PATH%\Bin\Win64
        ) else (
          set EDK_TOOLS_BIN=%EDK_TOOLS_PATH%\Bin\Win32
        )
      )
      set MINGW_ARCH=
    ) else (
      set EDK_TOOLS_BIN=%EDK_TOOLS_PATH%\Bin\Win32
    )
    if not defined FORCE_REBUILD (
      if not defined REBUILD (
        if not exist "%EDK_TOOLS_BIN%" (
          echo.
          echo !!! ERROR !!! Cannot find BaseTools Bin Win32!!!
          echo Please check the directory %EDK_TOOLS_BIN%
          echo Or configure EDK_TOOLS_BIN env to point to Bin directory.
          echo.
        )
      )
    )
  )`

This was added on 6/6/2025 as part of a larger improvement. 
The problem is that when the variable "EDK_TOOLS_BIN" is set within the if statement, "delayed expansion" is needed in order to get the updated valued of the variable from within the same statement. 

A simpler fix is to split the if statement into smaller pieces. 
Maybe flip the top if statement into something like this:
`if not defined EDK_TOOLS_BIN GOTO :EDKToolsDefined`
and then add the label ":EDKToolsDefined" below the current if statement. 

### Expected Behavior

I expected edksetup.bat to run successfully after pulling a new clone of the EDK2 repo.

### Steps To Reproduce

I pulled a fresh copy of the EDK2 repo, and ran edksetup.bat on a Win11 system that was not yet fully configured to build EDK2 projects.
This gave me an unexpected error as well as some expected errors. 

### Build Environment

```markdown
- OS(s): Win 11
- Tool Chain(s): cmd prompt + edksetup.bat
```

### Version Information

```text
Commit: SHA-1: c0ef2b017855177903d87937ac4cdaec5c7851db
```

### Urgency

Low

I will fix it

Maintainer feedback requested

### Anything else?

_No response_

FuntimeMS:
# Description
ToolSetup.bat had a bug where the check for %EDK_TOOLS_BIN% was happening inside the same code block where it was assigned, so the check was failing. 
I moved the check outside of the code block (if statement) where it is assigned (in 2 locations) and now it works as expected. 
fixes #11744

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
I verified that the error is now reported correctly, with the missing path being displayed on the screen.

## Integration Instructions
N/A

### DIFF
--- a/BaseTools/toolsetup.bat
+++ b/BaseTools/toolsetup.bat
@@ -243,15 +243,15 @@ if /I "%1"=="/?" goto Usage
     ) else (
       set EDK_TOOLS_BIN=%EDK_TOOLS_PATH%\Bin\Win32
     )
-    if not defined FORCE_REBUILD (
-      if not defined REBUILD (
-        if not exist "%EDK_TOOLS_BIN%" (
-          echo.
-          echo !!! ERROR !!! Cannot find BaseTools Bin Win32!!!
-          echo Please check the directory %EDK_TOOLS_BIN%
-          echo Or configure EDK_TOOLS_BIN env to point to Bin directory.
-          echo.
-        )
+  )
+  if not defined FORCE_REBUILD (
+    if not defined REBUILD (
+      if not exist "%EDK_TOOLS_BIN%" (
+        echo.
+        echo !!! ERROR !!! Cannot find BaseTools Bin Win32!!!
+        echo Please check the directory %EDK_TOOLS_BIN%
+        echo Or configure EDK_TOOLS_BIN env to point to Bin directory.
+        echo.
       )
     )
   )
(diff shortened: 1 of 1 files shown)


gapalomi:
CI its failing with this errors. Please correct commit message format

```
 * Subject line not in "package,package: description" format!
 * Missing Signed-off-by! (Note: this must be added by the code contributor!)
 * First line of commit message (subject line) is too long (141 > 75).
```

FuntimeMS:
Okay, I have added a new commit with the updated description.

FuntimeMS:
Apparently, I didn't do something right. Is it still complaining about the original commit message? 
Was I supposed to modify the existing commit message instead of creating a new commit for the updated message?

FuntimeMS:
Okay, I found the error messages, modified the commit message and repushed it. 
That fixed the latest commit, but I'm still getting the error for the original commit, I guess I'll see if I can re-write history and fix the original commit.

PR #11748, opened 2025-11-12
BaseTools,OvmfPkg/Bhyve: Line ending fixes

mikebeaton:
# Description

Minor line ending fixes.

These were discovered as part of cleaning up line endings in our EDK 2 fork, a few of which had drifted. Running the resultant scripts in EDK 2 confirmed these three minor anomalies are from upstream.

- [ ] Breaking change?
  - NO.
- [ ] Impacts security?
  - NO.
- [ ] Includes tests?
  - NO.

## How This Was Tested

N/A (except to confirm that @bexcran agrees .sh change is correct)

## Integration Instructions

N/A


mikebeaton:
@bexcran - Can you confirm I'm not missing anything stupid as to why Bhyve .sh line ending change would be wrong?

bexcran:
> @bexcran - Can you confirm I'm not missing anything stupid as to why Bhyve .sh line ending change would be wrong?

I can't tell from the diff what the original or new line endings were/are, but as long as they end up on people's computers as UNIX (LF), the script will run fine.

PR #11750, opened 2025-11-12
SecurityPkg: Fix Tcg2SubmitCommand in TPM field upgrade scenario

liqiqiii:
# Description

Fix Tcg2SubmitCommand in TPM field upgrade scenario.
There's a bug that after the device was forced shut down during the tpm firmware update process, 
then the TPM will stay in field upgrade mode and refuse to work properly. This will block TPM startup and cause the device to bootloop.
Now in the TCG_DXE_DATA struct we don’t have a flag to indicate the current TPM mode, and we will simply treat the upper
Scenario with this DEVICE_ERROR and set the TPMPresentFlag to FALSE.
Later in Tcg2SubmitCommand we will just return DEVICE_ERROR and skip the actual TPM recovery.
We should have an extra flag that check for the TPM response code and if it's TPM_RC_UPGRADE, we knew the device is in field upgrade mode and should continue the workflow.
After I commented out the TpmPresentFlag check, the device will be able to continue with the TPM recovery using the data stashed in persisted blob and finish the TPM capsule update successfully, and I can boot to the OS with TPM working.
We should only return EFI_DEVICE_ERROR when both TPMPresentFlag and TpmUpdateFlag are false.
The field upgrade is part of TCG spec, and the capsule update/recovery is part of UEFI spec.
Make this PR to bring in the fix for this corner case.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested with a device running dTPM, force shutdown the device during TPM capsule update, and in the reboot it can go into tpm recovery flow and update the TPM firmware correctly. Checked in Device Manager, version has been updated.


jyao1:
That is good problem statement.

Question: If the TPM failed due to Field Upgrade and cause PCR missing event log entry, do we want to continue boot OS directly without reset?

liqiqiii:
> That is good problem statement.
> 
> Question: If the TPM failed due to Field Upgrade and cause PCR missing event log entry, do we want to continue boot OS directly without reset?

Hi Jiewen, this is a good question. I believe when the TPM failed the field upgrade, we won't be able to boot to the OS, as the TPM will stay in field upgrade mode, which has very limited functionalities in that case. The change here just makes sure that when the TPM is in this mode we can still talk to TPM and try to do TPM recovery the next boot. 
During the early investigation of this bug, I tried to simply comment out some code and try to boot to the OS, but it will never get there, because it's fatal bug. My change won't allow the device to boot because TPM is not startup, it just allows the communication between TPM and UEFI.

liqiqiii:
@jyao1 @arunsbaskaran I have updated the PR and all tests are passing, could you take a look?

liqiqiii:
@jyao1 @arunsbaskaran A kindly reminder for this PR. Any thoughts on this?

jyao1:
OK. Let's align the design and expectation first.

We have 3 states for TPM response. See the following table.

| TPM Return Status | Support TPM in BIOS | Support TPM in OS | TPMPresentFlag | TpmUpdateFlag |
|---------------------|-------------------------|----------------------|-------------------|-------------------|
| SUCCESS               | YES                             | YES                         | TRUE                   | FALSE                 |
| FIELD_UPGRADE   | YES                             | NO                          | FALSE                  | TRUE                  |
| Other FAIL             | NO                             | NO                          | FALSE                 | FALSE                  |

I prefer TPMPresentFlag=FALSE in FIELD_UPGRADE, because the TPM PCR cannot reflect the BIOS. The chain of trust is broken.
We just give chance to finish upgrade in BIOS phase.

Thought?

If you agree, I prefer to include this table in the code as comment.

liqiqiii:
> OK. Let's align the design and expectation first.
> 
> We have 3 states for TPM response. See the following table.
> 
> TPM Return Status	Support TPM in BIOS	Support TPM in OS	TPMPresentFlag	TpmUpdateFlag
> SUCCESS	YES	YES	TRUE	FALSE
> FIELD_UPGRADE	YES	NO	FALSE	TRUE
> Other FAIL	NO	NO	FALSE	FALSE
> I prefer TPMPresentFlag=FALSE in FIELD_UPGRADE, because the TPM PCR cannot reflect the BIOS. The chain of trust is broken. We just give chance to finish upgrade in BIOS phase.
> 
> Thought?
> 
> If you agree, I prefer to include this table in the code as comment.

Hi @jyao1, thanks for the insights. I am completely with you; the table is a good way to think about it in a more comprehensive way. I have made changes based on this, leave the PresentFlag as false and add an additional check to make sure it will only return EFI_DEVICE_ERROR when TpmUpdateFlag is also false. Leave the PresentFlag as false won't hurt because that's how I tested originally and it works. We don't want Tpm to be functional in this specific tpm recovery boot and it will be good the next boot, which is expected. Thank you!

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -81,6 +81,7 @@ typedef struct {
 
 typedef struct _TCG_DXE_DATA {
```
jyao1, line 90: Would you please the following table to the code as comment? It will help people who review the code later.

```

+-------------------+---------------------+-------------------+----------------+---------------+
| TPM Return Status | Support TPM in BIOS | Support TPM in OS | TPMPresentFlag | TpmUpdateFlag |
|-------------------|---------------------|-------------------|----------------|---------------|
| SUCCESS           | YES                 | YES               | TRUE           | FALSE         |
| FIELD_UPGRADE     | YES                 | NO                | FALSE          | TRUE          |
| Other FAIL        | NO                  | NO                | FALSE          | FALSE         |
+-------------------+---------------------+-------------------+----------------+---------------+
```

liqiqiii: Hi @jyao1, I have added this. Thanks for the reminder.

PR #11757, opened 2025-11-14
Add support for complete out of tree build

gonzoua:
# Description

This pull requests adds support for the complet out-of-tree builds. Although the out-of-tree builds were supported by the `build` utility for some time, the `BaseTools` part still created files in-tree. This patchset addresses this blocker by:
  - Moving the `BaseTools` interim files and binaries to `$WORKSPACE/BaseTools/Build` directory
  - Making `BaseTool/Tests` use system's tmp directory for its interim files

Having full out-of-tree builds is convenient because it allows integrating EDK2 builds in a system that enforces read-only access to the codebase to prevent cross-target build contamination.

- [x] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

### Normal build env test
```
host$ docker run --rm -ti -v "${HOME}":"${HOME}" \
    -e EDK2_DOCKER_USER_HOME="${HOME}"  \
    -v $(pwd):/host -w /host \
    --name edk2-ubuntu-dev \
    ghcr.io/tianocore/containers/ubuntu-22-dev:latest /bin/bash
edk2-ubuntu-dev$ source edksetup.sh
edk2-ubuntu-dev$ make -j -C BaseTools
edk2-ubuntu-dev$ export PACKAGES_PATH=$(pwd)
edk2-ubuntu-dev$ build -t GCC5 -a AARCH64 -p ArmVirtPkg/ArmVirtQemu.dsc
```

### Read-only codebase builds test
```
host$ docker run --rm -ti -v "${HOME}":"${HOME}" \
    -e EDK2_DOCKER_USER_HOME="${HOME}" \
    -v $(pwd):/host:ro -w /host \
    --name edk2-ubuntu-dev \
    ghcr.io/tianocore/containers/ubuntu-22-dev:latest /bin/bash
edk2-ubuntu-dev$ export WORKSPACE=/tmp/workspace
edk2-ubuntu-dev$ mkdir -p $WORKSPACE/Conf
edk2-ubuntu-dev$ export EDK_TOOLS_PATH=$(pwd)/BaseTools
edk2-ubuntu-dev$ source edksetup.sh
edk2-ubuntu-dev$ make -j -C BaseTools
edk2-ubuntu-dev$ export PACKAGES_PATH=$(pwd)
edk2-ubuntu-dev$ build -t GCC5 -a AARCH64 -p ArmVirtPkg/ArmVirtQemu.dsc
```

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
+++ b/ArmPkg/Library/FmpDevicePsaFwuLib/FmpDevicePsaFwuLib.c
@@ -458,7 +458,7 @@ FmpDeviceGetImageTypeIdGuidPtr (
 
   @retval EFI_SUCCESS            The attributes supported by the firmware
                                  device were returned.
-  @retval EFI_INVALID_PARAMETER  One of the parameters — Settings or Supported — is NULL.
+  @retval EFI_INVALID_PARAMETER  One of the parameters - Settings or Supported - is NULL.
 
 **/
 EFI_STATUS
--- a/ArmPlatformPkg/ArmPlatformPkg.ci.yaml
+++ b/ArmPlatformPkg/ArmPlatformPkg.ci.yaml
@@ -93,9 +93,42 @@
         "IgnoreFiles": [],           # use gitignore syntax to ignore errors
                                      # in matching files
         "ExtendWords": [
+            "afsel",
+            "ctsen",
+            "dataen",
+            "daten",
+            "deassert",
+            "hbackporch",
             "hdlcd",
+            "hsync",
             "icdsgir",
-            "primecells"
+            "inten",
+            "jedec",
+            "pcell",
+            "primecells",
+            "pxclk",
+            "resen",
+            "rtsen",
+            "tpidrurw",
+            "uartcr",
+            "uartdmacr",
+            "uartdr",
+            "uartecr",
+            "uartfbrd",
+            "uartfr",
+            "uartibrd",
+            "uarticr",
+            "uartifls",
+            "uartilpr",
+            "uartimsc",
+            "uartris",
+            "vbackporch",
+            "vsync",
+            "vsyncwidth",
+            "wdogmis",
+            "wdogris",
+            "wsvga",
+            "wsxga"
            ],           # words to extend to the dictionary for this package
         "IgnoreStandardPaths": [    # Standard Plugin defined paths that
             "*.asm", "*.s"          # should be ignore
--- a/BaseTools/Source/C/Common/StringFuncs.c
+++ b/BaseTools/Source/C/Common/StringFuncs.c
@@ -110,7 +110,6 @@ SplitStringByWhitespace (
   CHAR8       *EndOfSubString;
   CHAR8       *EndOfString;
   STRING_LIST *Output;
-  UINTN       Item;
 
   String = CloneString (String);
   if (String == NULL) {
@@ -120,7 +119,7 @@ SplitStringByWhitespace (
 
   Output = NewStringList ();
 
-  for (Pos = String, Item = 0; Pos < EndOfString; Item++) {
+  for (Pos = String; Pos < EndOfString;) {
     while (isspace ((int)*Pos)) {
       Pos++;
     }
--- a/BaseTools/Source/C/EfiRom/EfiRom.c
+++ b/BaseTools/Source/C/EfiRom/EfiRom.c
@@ -44,7 +44,6 @@ Returns:
   FILE_LIST *FList;
   UINT32    TotalSize;
   UINT32    Size;
-  CHAR8     *Ptr0;
 
   SetUtilityName(UTILITY_NAME);
 
@@ -75,7 +74,7 @@ Returns:
   //
   if (mOptions.DumpOption == 1) {
     if (mOptions.FileList != NULL) {
-      if ((Ptr0 = strstr ((CONST CHAR8 *) mOptions.FileList->FileName, DEFAULT_OUTPUT_EXTENSION)) != NULL) {
+      if (strstr ((CONST CHAR8 *) mOptions.FileList->FileName, DEFAULT_OUTPUT_EXTENSION) != NULL) {
         DumpImage (mOptions.FileList);
         goto BailOut;
       } else {
--- a/EmbeddedPkg/GdbStub/GdbStub.c
+++ b/EmbeddedPkg/GdbStub/GdbStub.c
@@ -1149,8 +1149,8 @@ GdbExceptionHandler (
           AsciiSPrint (gOutBuffer, MAX_BUF_SIZE, "qXfer:libraries:read+;PacketSize=%d", MAX_BUF_SIZE);
           SendPacket (gOutBuffer);
         } else if (AsciiStrnCmp (gInBuffer, "qXfer:libraries:read::", 22) == 0) {
-          // ‘qXfer:libraries:read::offset,length
-          // gInBuffer[22] is offset string, ++Ptr is length string’
+          // 'qXfer:libraries:read::offset,length
+          // gInBuffer[22] is offset string, ++Ptr is length string'
           for (Ptr = &gInBuffer[22]; *Ptr != ','; Ptr++) {
           }
 
--- a/EmbeddedPkg/GdbStub/GdbStubInternal.h
+++ b/EmbeddedPkg/GdbStub/GdbStubInternal.h
@@ -320,7 +320,7 @@ SendNotSupported (
   VOID
   );
 
-/** ‘p n’
+/** 'p n'
  Reads the n-th register's value into an output buffer and sends it as a packet
  @param     SystemContext       Register content at time of the exception
  @param     InBuffer            This is the input buffer received from gdb server
@@ -331,7 +331,7 @@ ReadNthRegister (
   IN    CHAR8               *InBuffer
   );
 
-/** ‘g’
+/** 'g'
  Reads the general registers into an output buffer  and sends it as a packet
  @param     SystemContext           Register content at time of the exception
  **/
@@ -341,7 +341,7 @@ ReadGeneralRegisters (
   IN    EFI_SYSTEM_CONTEXT  SystemContext
   );
 
-/** ‘P n...=r...’
+/** 'P n...=r...'
  Writes the new value of n-th register received into the input buffer to the n-th register
  @param     SystemContext       Register content at time of the exception
  @param     InBuffer            This is the input buffer received from gdb server
@@ -353,7 +353,7 @@ WriteNthRegister (
   IN    CHAR8               *InBuffer
   );
 
-/** ‘G XX...’
+/** 'G XX...'
  Writes the new values received into the input buffer to the general registers
  @param     SystemContext               Register content at time of the exception
  @param     InBuffer                    Pointer to the input buffer received from gdb server
@@ -366,7 +366,7 @@ WriteGeneralRegisters (
   IN    CHAR8               *InBuffer
   );
 
-/** ‘m addr,length ’
+/** 'm addr,length '
  Find the Length of the area to read and the start address. Finally, pass them to
  another function, TransferFromMemToOutBufAndSend, that will read from that memory space and
  send it as a packet.
@@ -379,7 +379,7 @@ ReadFromMemory (
   IN  CHAR8  *PacketData
   );
 
-/** ‘M addr,length :XX...’
+/** 'M addr,length :XX...'
  Find the Length of the area in bytes to write and the start address. Finally, pass them to
  another function, TransferFromInBufToMem, that will write to that memory space the info in
  the input buffer.
@@ -392,7 +392,7 @@ WriteToMemory (
   IN CHAR8  *PacketData
   );
 
-/** ‘c [addr ]’
+/** 'c [addr ]'
  Continue. addr is Address to resume. If addr is omitted, resume at current
  Address.
 
@@ -407,7 +407,7 @@ ContinueAtAddress (
   IN  CHAR8               *PacketData
   );
 
-/** ‘s [addr ]’
+/** 's [addr ]'
  Single step. addr is the Address at which to resume. If addr is omitted, resume
  at same Address.
 
@@ -442,10 +442,10 @@ RemoveSingleStep (
   );
 
 /**
-  ‘Z1, [addr], [length]’
-  ‘Z2, [addr], [length]’
-  ‘Z3, [addr], [length]’
-  ‘Z4, [addr], [length]’
+  'Z1, [addr], [length]'
+  'Z2, [addr], [length]'
+  'Z3, [addr], [length]'
+  'Z4, [addr], [length]'
 
   Insert hardware breakpoint/watchpoint at address addr of size length
 
@@ -461,10 +461,10 @@ InsertBreakPoint (
   );
 
 /**
-  ‘z1, [addr], [length]’
-  ‘z2, [addr], [length]’
-  ‘z3, [addr], [length]’
-  ‘z4, [addr], [length]’
+  'z1, [addr], [length]'
+  'z2, [addr], [length]'
+  'z3, [addr], [length]'
+  'z4, [addr], [length]'
 
   Remove hardware breakpoint/watchpoint at address addr of size length
 
--- a/EmbeddedPkg/GdbStub/Ia32/Processor.c
+++ b/EmbeddedPkg/GdbStub/Ia32/Processor.c
@@ -173,7 +173,7 @@ BasicReadRegister (
   return OutBufPtr;
 }
 
-/** ‘p n’
+/** 'p n'
  Reads the n-th register's value into an output buffer and sends it as a packet
 
  @param   SystemContext   Register content at time of the exception
@@ -204,7 +204,7 @@ ReadNthRegister (
   SendPacket (OutBuffer);
 }
 
-/** ‘g’
+/** 'g'
  Reads the general registers into an output buffer  and sends it as a packet
 
  @param   SystemContext     Register content at time of the exception
@@ -274,7 +274,7 @@ BasicWriteRegister (
   return InBufPtr;
 }
 
-/** ‘P n...=r...’
+/** 'P n...=r...'
  Writes the new value of n-th register received into the input buffer to the n-th register
 
  @param   SystemContext   Register content at time of the exception
@@ -313,7 +313,7 @@ WriteNthRegister (
   SendSuccess ();
 }
 
-/** ‘G XX...’
+/** 'G XX...'
  Writes the new values received into the input buffer to the general registers
 
  @param   SystemContext       Register content at time of the exception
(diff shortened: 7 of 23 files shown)


niruiyu:
If I understand correctly, you are fixing the BaseTools binary build issue which generates the obj/exe files under BaseTools directory. I like it very much.

Can you explain a bit more about where the obj/exe files will be with your patch?

gonzoua:
The executables  are placed in `$WORKSPACE/BaseTools/bin` and the obj files are placed in the mirrored dir structure in $WORKSPACE, for instance obj files for `BaseTools/Source/C/LzmaCompress/Sdk/C/LzmaDec.c` is `$WORKSPACE/BaseTools/Source/C/LzmaCompress/Sdk/C/LzmaDec.o`

BaseTools/BinWrappers/PosixLike/BrotliCompress
```diff
@@ -36,7 +36,13 @@ do
 done
 
 
-if [ -n "$EDK_TOOLS_PATH" ] && [ -e "$EDK_TOOLS_PATH/Source/C/bin" ]
+if [ -n "$WORKSPACE" ] && [ -e "$WORKSPACE/BaseTools/Source/C/bin" ]
```
leiflindholm: I think this would break some existing users (like myself) who do the "mostly out of tree build" already supported. WORKSPACE is not tied to being the edk2 repository, which is why EDK_TOOLS_PATH exists in the first place.
Or am I missing something?

BaseTools/Source/C/GNUmakefile
```diff
@@ -22,7 +28,6 @@ all: makerootdir subdirs
 	@echo Finished building BaseTools C Tools with HOST_ARCH=$(HOST_ARCH)
 
 LIBRARIES = Common
-VFRAUTOGEN = VfrCompile/VfrLexer.h
```
leiflindholm, line 25: This does not look obviously related to the described change?
If it is general cleanup to make the end result tidier, could it be broken out into a separate commit?

BaseTools/BinWrappers/PosixLike/BrotliCompress
ardbiesheuvel: This is how I use the WORKSPACE feature as well, and it appears to work fine. That path is where the BaseTools build will put the resulting binaries.

leiflindholm: Ah, got it, thanks!

BaseTools/Source/C/GNUmakefile
gonzoua: It's general cleanup, added as a separate commit.

niruiyu:
> The executables are placed in `$WORKSPACE/BaseTools/bin` and the obj files are placed in the mirrored dir structure in $WORKSPACE, for instance obj files for `BaseTools/Source/C/LzmaCompress/Sdk/C/LzmaDec.c` is `$WORKSPACE/BaseTools/Source/C/LzmaCompress/Sdk/C/LzmaDec.o`

I thought usually $WORKSPACE points to the project root directory. Then your approach that places the intermediate files in $WORKSPACE might not work. Why not place these intermediate files in Build directory?

gonzoua:
> > The executables are placed in `$WORKSPACE/BaseTools/bin` and the obj files are placed in the mirrored dir structure in $WORKSPACE, for instance obj files for `BaseTools/Source/C/LzmaCompress/Sdk/C/LzmaDec.c` is `$WORKSPACE/BaseTools/Source/C/LzmaCompress/Sdk/C/LzmaDec.o`
> 
> I thought usually $WORKSPACE points to the project root directory. Then your approach that places the intermediate files in $WORKSPACE might not work. Why not place these intermediate files in Build directory?

Could you please elaborate? My understanding is WORKSPACE is where the main (non-BaseTools) build places its files, it can be `/tmp/workspace` for out-of-tree build or it can point to the edk2 root dir. If WORKSPACE points to the project root, it by definition should be writable, and the behavior I described is the same as the one that currently exists.

niruiyu:
> > > The executables are placed in `$WORKSPACE/BaseTools/bin` and the obj files are placed in the mirrored dir structure in $WORKSPACE, for instance obj files for `BaseTools/Source/C/LzmaCompress/Sdk/C/LzmaDec.c` is `$WORKSPACE/BaseTools/Source/C/LzmaCompress/Sdk/C/LzmaDec.o`
> > 
> > 
> > I thought usually $WORKSPACE points to the project root directory. Then your approach that places the intermediate files in $WORKSPACE might not work. Why not place these intermediate files in Build directory?
> 
> Could you please elaborate? My understanding is WORKSPACE is where the main (non-BaseTools) build places its files, it can be `/tmp/workspace` for out-of-tree build or it can point to the edk2 root dir. If WORKSPACE points to the project root, it by definition should be writable, and the behavior I described is the same as the one that currently exists.

You can open edksetup.bat. The logic there assumes the WORKSPACE points to the parent folder of "BaseTools". It's the root of the source code, not the temp directory that holds output files.

gonzoua:
> > > > The executables are placed in `$WORKSPACE/BaseTools/bin` and the obj files are placed in the mirrored dir structure in $WORKSPACE, for instance obj files for `BaseTools/Source/C/LzmaCompress/Sdk/C/LzmaDec.c` is `$WORKSPACE/BaseTools/Source/C/LzmaCompress/Sdk/C/LzmaDec.o`
> > > 
> > > 
> > > I thought usually $WORKSPACE points to the project root directory. Then your approach that places the intermediate files in $WORKSPACE might not work. Why not place these intermediate files in Build directory?
> > 
> > 
> > Could you please elaborate? My understanding is WORKSPACE is where the main (non-BaseTools) build places its files, it can be `/tmp/workspace` for out-of-tree build or it can point to the edk2 root dir. If WORKSPACE points to the project root, it by definition should be writable, and the behavior I described is the same as the one that currently exists.
> 
> You can open edksetup.bat. The logic there assumes the WORKSPACE points to the parent folder of "BaseTools". It's the root of the source code, not the temp directory that holds output files.

Right, these are default values if neither WORKSPACE nor EDK_TOOLS_PATH are defined. But it's not a strict requirement, as far as I understand. They can be specified them to different locations. For instance, section [4.1.1 Development Environments](https://tianocore-docs.github.io/edk2-BuildSpecification/release-1.28/4_edk_ii_build_process_overview/41_edk_ii_build_system.html) sets WORKSPACE to one level up from the edk2 root and uses PACKAGES_PATH to set search path.

Section "Read-only codebase builds test" in the PR's description also demonstrates that this approach works.

ardbiesheuvel:
I personally always use a workspace that is separate from the edk2/ directory, by setting `WORKSPACE` to a different directory.

I guess using `$WORKSPACE/Build` rather than `$WORKSPACE` for the BaseTools would make this feature useful even in the case where WORKSPACE points to the main repo, so perhaps this is something to consider. (I.e., you can clean the working area by deleting everything under Build/)

gonzoua:
Reworked to put everything in `$WORKSPACE/Build/BaseTools`. Implementation-wise it's not a big difference. My rationale for keeping it pre-change way for WORKSPACE==<edk2 root> case was to keep it backward compatible compatible with potential external build workflows. I also added a paragraph for the new behavior to the commit message.

mikebeaton (changes requested):
Hi! I am used to being able to `rm -rf Build` without it removing already compiled BaseTools - could they go in peer directory of Build rather than subdirectory?

Another point, run within edk2 dir after the above steps, does this matter?:
```
$ git clean -xdn
Would remove BaseTools/Source/Python/AutoGen/__pycache__/
Would remove BaseTools/Source/Python/Common/__pycache__/
Would remove BaseTools/Source/Python/CommonDataClass/__pycache__/
Would remove BaseTools/Source/Python/GenFds/__pycache__/
Would remove BaseTools/Source/Python/GenPatchPcdTable/__pycache__/
Would remove BaseTools/Source/Python/PatchPcdValue/__pycache__/
Would remove BaseTools/Source/Python/Workspace/__pycache__/
Would remove BaseTools/Source/Python/__pycache__/
Would remove BaseTools/Source/Python/build/__pycache__/
Would remove BaseTools/Tests/__pycache__/
```

gonzoua:
> Hi! I am used to being able to `rm -rf Build` without it removing already compiled BaseTools - could they go in peer directory of Build rather than subdirectory?

The Build directory already contains multiple peer directories, for instance: `Build/ArmVirtQemu-AArch64` and `Build/BaseTools` so you can delete  a target build without deleting the BaseTools, it's just one level deeper.  Having a single entry point for nuking everything, IMO, is more preferable to multiple top-level peer directories. 

> Another point, run within edk2 dir after the above steps, does this matter?:
> 
> ```
> $ git clean -xdn
> Would remove BaseTools/Source/Python/AutoGen/__pycache__/
> Would remove BaseTools/Source/Python/Common/__pycache__/
> Would remove BaseTools/Source/Python/CommonDataClass/__pycache__/
> Would remove BaseTools/Source/Python/GenFds/__pycache__/
> Would remove BaseTools/Source/Python/GenPatchPcdTable/__pycache__/
> Would remove BaseTools/Source/Python/PatchPcdValue/__pycache__/
> Would remove BaseTools/Source/Python/Workspace/__pycache__/
> Would remove BaseTools/Source/Python/__pycache__/
> Would remove BaseTools/Source/Python/build/__pycache__/
> Would remove BaseTools/Tests/__pycache__/
> ```

No, this is not critical. If your src base is mounted read-only the pycache bytecode is just not created. This behaviour also can be switched off by setting the `PYTHONDONTWRITEBYTECODE` env variable to `1`. Alternatively it also can be moved to the Build/BaseTool dir by setting `PYTHONPYCACHEPREFIX` in edksetup.sh, but it can be done as a further impovement if neccessary.  I don't know if this caching provides any significant speed-up.

mikebeaton:
> Having a single entry point for nuking everything, IMO, is more preferable to multiple top-level peer directories.

Thank you for the clarification on the pycache point.

On the above point, I am afraid I still disagree. I do not work on BaseTools as often as I work on packages, but when I do I know I have `make -C BaseTools clean` and `make -C BaseTools/Source/C clean` available to remove any possible caching effects.

But I often build various different packages, sometimes very many of them as I'm quite often working on toolchain changes, and when I'm doing that I often want to say 'just reset the state of everything I've built', without having to remember exactly what I have and haven't built, and without having to rebuild BaseTools.

gonzoua:
Moved BaseTools interim files and binaries out of `Build/` directory to `BaseTools/Build` directory. Now they can be removed separately.

mikebeaton:
> Moved BaseTools interim files and binaries out of `Build/` directory to `BaseTools/Build` directory. Now they can be removed separately.

Thank you.

It occurred to me, could `$WORKSPACE/Conf` helpfully be moved to `$WORKSPACE/BaseTools/Conf`? Fewer subdirectories, related files together, and we can then `rm -rf $WORKSPACE/BaseTools` to clear down BaseTools plus conf, which would be a new convenient feature, while still being able to clean them both separately.

Also, could it be convenient if `edksetup.sh` created the `Conf` directory?

These are both just thoughts/suggestions, if they strike you as sensible too - not at all blockers or definite requests for changes, in this case.

ardbiesheuvel:
> > Moved BaseTools interim files and binaries out of `Build/` directory to `BaseTools/Build` directory. Now they can be removed separately.
> 
> Thank you.
> 
> It occurred to me, could `$WORKSPACE/Conf` helpfully be moved to `$WORKSPACE/BaseTools/Conf`? Fewer subdirectories, related files together, and we can then `rm -rf $WORKSPACE/BaseTools` to clear down BaseTools plus conf, which would be a new convenient feature, while still being able to clean them both separately.
> 
> Also, could it be convenient if `edksetup.sh` created the `Conf` directory?
> 
> These are both just thoughts/suggestions, if they strike you as sensible too - not at all blockers or definite requests for changes, in this case.

Could we please revisit this in the context of a different PR? Personally, I never clean Conf/ at all, because tools_def.txt and build_rule.txt are symlinked to the templates under BaseTools/

mikebeaton (approved):

mikebeaton:
> > > Moved BaseTools interim files and binaries out of `Build/` directory to `BaseTools/Build` directory. Now they can be removed separately.
> > 
> > 
> > Thank you.
> > It occurred to me, could `$WORKSPACE/Conf` helpfully be moved to `$WORKSPACE/BaseTools/Conf`? Fewer subdirectories, related files together, and we can then `rm -rf $WORKSPACE/BaseTools` to clear down BaseTools plus conf, which would be a new convenient feature, while still being able to clean them both separately.
> > Also, could it be convenient if `edksetup.sh` created the `Conf` directory?
> > These are both just thoughts/suggestions, if they strike you as sensible too - not at all blockers or definite requests for changes, in this case.
> 
> Could we please revisit this in the context of a different PR? Personally, I never clean Conf/ at all, because tools_def.txt and build_rule.txt are symlinked to the templates under BaseTools/

Yes, of course, we can. This change approved as is could go in this release, I think? (Don't want it to have ground to a halt because of my additional suggestion!)

ardbiesheuvel:
Thanks for respinning this. I'd like to get this in for the next stable tag (@lgao4)

Some minor comments remaining, including a typo in the commit log of patch 3 (s/mart/part) [Annoyingly, github does not let me respond to that)

BaseTools/Tests/GNUmakefile
ardbiesheuvel: These are redundant now, right?

ardbiesheuvel: Fixed this up and pushed a new version

gonzoua:
Yes, that was redundant logic re-introduced by bad conflict resolution during rebase. Thank you for fixing it Ard, appreciate your help.

lgao4:
> Thanks for respinning this. I'd like to get this in for the next stable tag (@lgao4)
> 
> Some minor comments remaining, including a typo in the commit log of patch 3 (s/mart/part) [Annoyingly, github does not let me respond to that)

Do you mean 202602 stable tag? I see this change has passed code review before code freeze. It can be merged.

lgao4:
@gonzoua , thanks for your work to support this feature. Current changes only support Linux development environment. I think Windows BaseTools Makefile can also be updated for it. This can be another separate task.

lgao4 (approved):

mdkinney:
This PR appears to have introduced regression for Mingw CLANG environments.  I am investigating.  Must fix for the stable tag.

gonzoua:
Ugh, sorry for the caused breakage.

mdkinney:
@gonzoua There is a FIT tool in edk2-platforms.  I am seeing some issues building that tool after these changes.  Can you offer suggestions on how to update the makefiles in https://github.com/tianocore/edk2-platforms/tree/master/Silicon/Intel/Tools

mdkinney:
@gonzoua I think edk2-platforms make need a change like the following for FitGen tool to add BUILDROOT.

```
diff --git a/Silicon/Intel/Tools/GNUmakefile b/Silicon/Intel/Tools/GNUmakefile
index 6910b6ad0..659454fc8 100644
--- a/Silicon/Intel/Tools/GNUmakefile
+++ b/Silicon/Intel/Tools/GNUmakefile
@@ -7,13 +7,14 @@
##

MAKEROOT = $(EDK_TOOLS_PATH)/Source/C
+BUILDROOT=$(WORKSPACE)/BaseTools/Build/Source/C^M

APPLICATIONS = \
   FitGen \

SUBDIRS := $(APPLICATIONS)

-$(APPLICATIONS): $(MAKEROOT)/bin
+$(APPLICATIONS): $(BUILDROOT)/bin^M

.PHONY: subdirs $(SUBDIRS)
subdirs: $(SUBDIRS)
```

gonzoua:
I think more complete patch would be something like this: https://github.com/gonzoua/edk2-platforms/tree/fix-fitgen-build

It fixes both in-tree and out-of-tree builds, but as it is used in build_bios.py, the script may also need to be fixed/tested. Do you have the end-to-end repro scenario?

mdkinney:
> I think more complete patch would be something like this: https://github.com/gonzoua/edk2-platforms/tree/fix-fitgen-build
> 
> It fixes both in-tree and out-of-tree builds, but as it is used in build_bios.py, the script may also need to be fixed/tested. Do you have the end-to-end repro scenario?

This looks like the right fix for edk2-platforms.  Can you open a PR for edk2-platforms?

gonzoua:
https://github.com/tianocore/edk2-platforms/pull/941

mdkinney:
After reviewing side effects from this change on downstream platforms, I think the fact that this change is a breaking change is a big issue.  I think there is a way to make an additional change here to convert this to a compatible change with an option to opt into out-of-tree builds and make the feature more flexible.

This current implementation assumes the tool output files will go into `WORKSPACE/BaseTools/Build` if `WORKSPACE` is set.  It will go into the old location if WORKSPACE is not set. Many downstream config set WORKSPACE before building tools which means they are forced into the new location.

In order to preserve current behavior and opt-in to the out-of-tree feature when a platform is ready, I recommend a new environment variable be used to specify the out-of-tree location.  This would allow tools to be put in directory inside `Build` or in a unique tool output directory.

My suggestion is use a new env var called `EDK_TOOLS_BUILD_PATH`.  If it is not defined, then `BUILDROOT` is set to `MAKEROOT` for compatibility.  If `EDK_TOOLS_BUILD_PATH` is defined, then `BUILDROOT` is set to `EDK_TOOLS_BUILD_PATH/Source/C`.

mdkinney:
Opened PR https://github.com/tianocore/edk2/pull/12157 to make this feature opt-in and make output directory configurable.

gonzoua:
I addresses the proposed fix approach in PR #12157 and would like to follow up here on the side effects. Do you have more examples of regressions? The BaseTools, as far as I understood, was supposed to be a hermetic build and platform code would use it through wrappers. 

AFAIK there is no "contract" between edk2 and edk2-platforms repos on how to hook platform-specific base tool into the  build, so tools like FitGen  either need to be a standalone, hermetic build or edk2 needs an official way to add a platform tool into a Base tools build. Using app.makefile directly is not the right thing to do and in a long term need to be fixed, IMO.

mdkinney:
> I addresses the proposed fix approach in PR #12157 and would like to follow up here on the side effects. Do you have more examples of regressions? The BaseTools, as far as I understood, was supposed to be a hermetic build and platform code would use it through wrappers.
> 
> AFAIK there is no "contract" between edk2 and edk2-platforms repos on how to hook platform-specific base tool into the build, so tools like FitGen either need to be a standalone, hermetic build or edk2 needs an official way to add a platform tool into a Base tools build. Using app.makefile directly is not the right thing to do and in a long term need to be fixed, IMO.

The scenario that breaks is a platform build that depends on BaseTools from edk2 and FitGen tool from edk2-platforms.  Those both need to be built before the platform can be built.

I agree that the root cause of build failure of the edk2-platforms build of FitGen is due to FitGen depending on makefile content from the edk2 repo and that is an issue that needs to be fixed. I also do not like that FitGen generates output in the same directory as BaseTools because there could be collisions if that pattern is continued.

I think a reasonable fix to edk2-platforms FitGen is to add a copy of the dependent makefiles to the edk2-platforms repo and for edk2-platforms to have its own tools output directory.

Unfortunately, it is hard to know how many custom tools have follow the FitGen example.

PR #11761, opened 2025-11-16
BaseTools: Remove no longer needed warning suppression

mikebeaton:
# Description

Since https://github.com/tianocore/edk2/commit/ae83c6b7fd83a5906e016a32027c1bcd792a624e `-Wno-unused-but-set-variable`, `-Wno-unused-const-variable` and `-Wno-unused-variable` warning suppression is no longer needed in any builds, and the warnings can be re-enabled to catch real errors.

On pushing this to CI we find two such minors errors introduced in https://github.com/tianocore/edk2/commit/1634dd5b9396702556aa784cfd47c3408d72006f and https://github.com/tianocore/edk2/commit/9c651ef83a74d83219075316aa5d1c189ed608c8, so we fix these in additional commits.

- [ ] Breaking change?
  - NO.
- [ ] Impacts security?
  - NO.
- [ ] Includes tests?
  - NO.

## How This Was Tested

EDK 2 CI (GCC only)
Acidanthera [AUDK CI](https://github.com/acidanthera/audk/commit/2950fc4d6a6c47608a69893b3534a5306524110c) (also tests CLANGDWARF and CLANGPDB)

## Integration Instructions

N/A

### DIFF
--- a/StandaloneMmPkg/Core/StandaloneMmCore.c
+++ b/StandaloneMmPkg/Core/StandaloneMmCore.c
@@ -265,7 +265,9 @@ MmReadyToLockHandler (
   // Display any drivers that were not dispatched because dependency expression
   // evaluated to false if this is a debug build
   //
-  // MmDisplayDiscoveredNotDispatched ();
+  DEBUG_CODE_BEGIN ();
+  MmDisplayDiscoveredNotDispatched ();
+  DEBUG_CODE_END ();
 
   return Status;
 }
@@ -455,7 +457,7 @@ MmCorePrepareCommunicationBuffer (
   mMmCommunicationBuffer = (MM_COMM_BUFFER *)GET_GUID_HOB_DATA (GuidHob);
   DEBUG ((
     DEBUG_INFO,
-    "MM Communication Buffer is at %x, number of pages is %x\n",
+    "MM Communication Buffer is at 0x%x, number of pages is %d\n",
     mMmCommunicationBuffer->PhysicalStart,
     mMmCommunicationBuffer->NumberOfPages
     ));


BaseTools/Conf/tools_def.template
```diff
@@ -983,7 +983,7 @@ DEFINE GCC5_LOONGARCH64_PP_FLAGS           = -mabi=lp64d -march=loongarch64 DEF(
 *_GCC48_IA32_NASM_FLAGS           = -f elf32
 
   DEBUG_GCC48_IA32_CC_FLAGS       = DEF(GCC48_IA32_CC_FLAGS)
-RELEASE_GCC48_IA32_CC_FLAGS       = DEF(GCC48_IA32_CC_FLAGS) -Wno-unused-but-set-variable
+RELEASE_GCC48_IA32_CC_FLAGS       = DEF(GCC48_IA32_CC_FLAGS)
```
ardbiesheuvel: Often, hacks such as these exist because of bugs in the compiler, not bugs in our code.

IOW, dropping this without checking whether or not GCC 4.8 still works is not a great idea. Bonus points for finding out that GCC 4.8 doesn't even build to being with, in which case we can consider dropping GCC 4.8 support altogether. Or did you do a test build with GCC 4.8 already?

mikebeaton: No, will test.

leiflindholm: These days, we tend to see them because our codebase is crammed full of several-hundred-line functions with 30+ live variables throughout with overlapping meanings across.

The last GCC 4.8 point release was made over 10 years ago.
And we do have this note, from May 2023:
https://github.com/tianocore/edk2/blob/master/BaseTools/Conf/tools_def.template#L22

In fact, I was surprised to notice recently we still had GCC 4.8 support in at all. I had mistakenly thought we threw that out when we decided to reinterpret GCC5 as LTO with GCC49 as NOLTO.

Which is a long-winded way of saying:
a) I think it's past time to yeet 4.8 (and maybe 4.9, 5.0)
b) I would quite like to see what happens if we merge this

As always, we can revert (for now) if it causes significant impact _and_ those affected are willing to start working towards removing their dependencies on outdated tools.

mikebeaton: So on Ubuntu 14.04.1, GCC 4.8.4, Python 3.5.2, it is struggling to make BaseTools, before I ever get to trying the toolchain.

Here's what I'm seeing - let me know if I'm missing something and there's a quick way to get past this!

----

Initially we get:

```
In file included from BrotliCompress.c:21:0:
./brotli/c/common/constants.h:196:1: error: ‘model’ attribute directive ignored [-Werror=attributes]
 BrotliPrefixCodeRange _kBrotliPrefixCodeRanges[BROTLI_NUM_BLOCK_LEN_SYMBOLS];
 ^
cc1: error: unrecognized command line option "-Wno-restrict" [-Werror]
cc1: error: unrecognized command line option "-Wno-stringop-truncation" [-Werror]
```

If I just remove those options from `BaseTools/Source/C/Makefiles/header.makefile` then I get to:

```
In file included from BrotliCompress.c:21:0:
./brotli/c/common/constants.h:196:1: error: ‘model’ attribute directive ignored [-Werror=attributes]
 BrotliPrefixCodeRange _kBrotliPrefixCodeRanges[BROTLI_NUM_BLOCK_LEN_SYMBOLS];
 ^
```

If I then add `-Wno-error=attributes`, we get to:

```
.
.
.
In file included from brotli/c/enc/encoder_dict.h:15:0,
                 from brotli/c/enc/params.h:13,
                 from brotli/c/enc/command.h:15,
                 from brotli/c/enc/backward_references_hq.h:14,
                 from brotli/c/enc/encode.c:16:
brotli/c/enc/static_dict_lut.h:40:5: warning: ‘model’ attribute directive ignored [-Wattributes]
     kStaticDictionaryBuckets[BROTLI_ENC_STATIC_DICT_LUT_NUM_BUCKETS];
     ^
brotli/c/enc/static_dict_lut.h:42:5: warning: ‘model’ attribute directive ignored [-Wattributes]
     kStaticDictionaryWords[BROTLI_ENC_STATIC_DICT_LUT_NUM_ITEMS];
     ^
In file included from brotli/c/enc/backward_references_hq.h:14:0,
                 from brotli/c/enc/encode.c:16:
brotli/c/enc/command.h:23:1: warning: ‘model’ attribute directive ignored [-Wattributes]
 uint32_t kBrotliInsBase[BROTLI_NUM_INS_COPY_CODES];
 ^
brotli/c/enc/command.h:25:1: warning: ‘model’ attribute directive ignored [-Wattributes]
 uint32_t kBrotliInsExtra[BROTLI_NUM_INS_COPY_CODES];
 ^
brotli/c/enc/command.h:27:1: warning: ‘model’ attribute directive ignored [-Wattributes]
 uint32_t kBrotliCopyBase[BROTLI_NUM_INS_COPY_CODES];
 ^
brotli/c/enc/command.h:29:1: warning: ‘model’ attribute directive ignored [-Wattributes]
 uint32_t kBrotliCopyExtra[BROTLI_NUM_INS_COPY_CODES];
 ^
In file included from brotli/c/enc/brotli_bit_stream.h:22:0,
                 from brotli/c/enc/encode.c:19:
brotli/c/enc/entropy_encode.h:78:1: warning: ‘model’ attribute directive ignored [-Wattributes]
 BROTLI_INTERNAL extern BROTLI_MODEL("small") const size_t kBrotliShellGaps[6];
 ^
In file included from brotli/c/enc/encode.c:24:0:
brotli/c/enc/dictionary_hash.h:36:5: warning: ‘model’ attribute directive ignored [-Wattributes]
     kStaticDictionaryHashWords[BROTLI_ENC_NUM_HASH_BUCKETS];
     ^
brotli/c/enc/dictionary_hash.h:38:5: warning: ‘model’ attribute directive ignored [-Wattributes]
     kStaticDictionaryHashLengths[BROTLI_ENC_NUM_HASH_BUCKETS];
     ^
brotli/c/enc/encode.c: In function ‘InitCommandPrefixCodes’:
brotli/c/enc/encode.c:203:3: warning: ‘model’ attribute directive ignored [-Wattributes]
   static const BROTLI_MODEL("small") uint8_t kDefaultCommandDepths[128] = {
   ^
brotli/c/enc/encode.c:213:3: warning: ‘model’ attribute directive ignored [-Wattributes]
   static const BROTLI_MODEL("small") uint16_t kDefaultCommandBits[128] = {
   ^
brotli/c/enc/encode.c:227:3: warning: ‘model’ attribute directive ignored [-Wattributes]
   static const BROTLI_MODEL("small") uint8_t kDefaultCommandCode[] = {
   ^
brotli/c/enc/encode.c: In function ‘EstimateEntropy’:
brotli/c/enc/encode.c:248:3: error: ‘for’ loop initial declarations are only allowed in C99 mode
   for (size_t i = 0; i < size; ++i) {
   ^
brotli/c/enc/encode.c:248:3: note: use option -std=c99 or -std=gnu99 to compile your code
brotli/c/enc/encode.c: In function ‘ChooseContextMap’:
brotli/c/enc/encode.c:269:3: warning: ‘model’ attribute directive ignored [-Wattributes]
   uint32_t kStaticContextMapContinuation[64] = {
   ^
brotli/c/enc/encode.c:276:3: warning: ‘model’ attribute directive ignored [-Wattributes]
   uint32_t kStaticContextMapSimpleUTF8[64] = {
   ^
brotli/c/enc/encode.c: In function ‘ShouldUseComplexStaticContextMap’:
brotli/c/enc/encode.c:333:3: warning: ‘model’ attribute directive ignored [-Wattributes]
   uint32_t kStaticContextMapComplexUTF8[64] = {
   ^
make[2]: *** [brotli/c/enc/encode.o] Error 1
make[2]: Leaving directory `/home/mjsbeaton/OpenSource/edk2/BaseTools/Source/C/BrotliCompress'
make[1]: *** [BrotliCompress] Error 2
make[1]: Leaving directory `/home/mjsbeaton/OpenSource/edk2/BaseTools/Source/C'
make: *** [Source/C] Error 2
```

with several pages more similar above what I've pasted.

ardbiesheuvel:
Thanks for doing the due diligence. Let's drop GCC 4.8 support rather than waste any more time on this.

mikebeaton:
> Thanks for doing the due diligence. Let's drop GCC 4.8 support rather than waste any more time on this.

Would that be and drop 4.9 as well? (I don't have 'skin in the game', just wondering!)

mikebeaton:
Hmmm... for GCC5, the newest version of Ubuntu which comes with GCC 5 by default is 16.04, but that comes with Python 3.5 by default and no easy access to Python 3.6, so we can build BaseTools, but when we come to build a package we get:

```
$ build -a X64 -t GCC5 -p MdePkg/MdePkg.dsc
Traceback (most recent call last):
  File "/home/mjsbeaton/OpenSource/edk2/BaseTools/BinWrappers/PosixLike/../../Source/Python/build/build.py", line 33, in <module>
    import secrets
ImportError: No module named 'secrets'
```

Apparently 'secrets' is a built-in thing, in Python 3.6+.

Presumably we (I!) can work round this one by going to a newer Linux and installing gcc 5.

ardbiesheuvel:
Again, thanks for doing the due diligence.

A while ago, we cloned GCC49 to GCCNOLTO and GCC5 to GCC, in order to decouple our GCC support from the major/minor GCC versions.

Perhaps by the time the upcoming stable tag lands, it's time to
a) get rid of GCC48 and GCC49
b) decide which version of GCC we still want to support, as I suspect that there is little point to supporting anything older than 8 at this time
c) get rid of GCC5, or keep it out of nostalgia, but document that only GCC 8 and older are still supported.

mikebeaton:
FWIW with gcc 7 these changes don't break sample test builds X46 RELEASE of MdePkg, MdeModulePkg, SecurityPkg, UefiCpuPkg with **GCC5**.

The changes in this PR also don't break **GCC48** or **GCC49** X46 RELEASE MdePkg with gcc 7. MdeModulePkg, SecurityPkg and UefiCpuPkg fail with GCC48 and GCC49 on the same test setup, but not for reasons to do with this PR.

Worth noting we are now reliant on **nasm 2.15.01**+ (due to use of shadow stack instructions https://github.com/netwide-assembler/nasm/blob/master/doc/changes.src#L479), which is also not found naturally with older gcc (e.g. I compiled 3.01 from source).

mikebeaton:
> Perhaps by the time the upcoming stable tag lands, it's time to a) get rid of GCC48 and GCC49 b) decide which version of GCC we still want to support, as I suspect that there is little point to supporting anything older than 8 at this time c) get rid of GCC5, or keep it out of nostalgia, but document that only GCC 8 and older are still supported.

One more quick point: it may be worth noting that there is a Ubuntu LTS with gcc 7 as the default (18.04), but not one with 8 (20.04 skips to 9). Since that is one (of several) ways of quickly accessing a test environment, that might be some degree of a reason to prefer 7 rather than 8 as a minimal supported gcc version, depending on what the other reasons are.

mikebeaton:
After discussion and due diligence as above, is this one agreed to merge as is now?

mikebeaton:
I'll have a poke around in edk2-platforms, and see what changes this would require, there.

mikebeaton:
Rebased, and fixed very minor conflict in two places where changed lines here were right next to lines added for LOONGARCH64 in c33aa63b7dcf7f7377f92becef1a792fe2abecd2. Rebase has also dropped one of the two previously required preceding commits, which has now been merged as 6e5835d76ae27859b761003b6d57ab4ec190428f.

SecurityPkg/Library/HashLibTpm2/HashLibTpm2PeilessSecLib.c
```diff
@@ -585,7 +584,6 @@ HasLibTpm2PeilessSecLibConstructor (
     goto DisableHandler;
   }
 
-  TcgPcrEvent             = (TCG_PCR_EVENT *)EventLog;
   TcgEfiSpecIdEventStruct = (TCG_EfiSpecIDEventStruct *)
                             (EventLog + OFFSET_OF (TCG_PCR_EVENT, Event));
```
ardbiesheuvel: This should be

```
TcgEfiSpecIdEventStruct = (TCG_EfiSpecIDEventStruct *)TcgPcrEvent->Event;
```

which also fixes the unused warning.

SecurityPkg/Library/HashLibTpm2/HashLibTpm2PeilessSecLib.c
```diff
@@ -586,9 +586,7 @@ HasLibTpm2PeilessSecLibConstructor (
   }
 
   TcgPcrEvent             = (TCG_PCR_EVENT *)EventLog;
-  TcgEfiSpecIdEventStruct = (TCG_EfiSpecIDEventStruct *)
-                            (EventLog + OFFSET_OF (TCG_PCR_EVENT, Event));
-
+  TcgEfiSpecIdEventStruct = (TCG_EfiSpecIDEventStruct *)TcgPcrEvent->Event;
```
ardbiesheuvel, line 589: Yeah better, thanks.

But please drop the URL from the commit log, as it will go stale very quickly, and no need for the stream-of-consciousness.

mikebeaton:
I've successfully built multiple platforms in edk2-platforms with GCC, with this change, with [almost](https://github.com/tianocore/edk2-platforms/pull/912) no problems. Including Arm, Intel, etc. As usual, I think, not all packages in there build properly, but not for reasons related to this change.

TL/DR of my comments above, this doesn't even break GCC48/GCC49 with old (but not as old as 4.8 and 4.9) gcc, so possibly finally removing these doesn't need to be a blocker to this PR? Or were you hoping that would be done too, in this PR?

> decide which version of GCC we still want to support, as I suspect that there is little point to supporting anything older than 8 at this time

I also suggested above that 7 might be a better choice (due to what different Ubuntu LTS versions support).

mikebeaton:
Can I cc @mdkinney @os-d @makubacki - Would this change be good to include in the current round of tooling updates? (I think so.)

**Notes**:
 - I have tested that this doesn't break GCC48 (with a more recent, but still somewhat old gcc). That said, I could remove the changes for GCC48, GCC49 and just leave for everything else (GCC, GCC5, XCODE5 (tested), CLANGPDB, CLANGDWARF) if preferred?
 - Also now tested that it doesn't break a fairly large representative sample of edk2-platforms projects, in a sample of toolchains. There's a mountain of stuff there to test completely, but it is certainly not breaking everything left right and centre, or anywhere close.

bexcran:
Oh! Yes I did sign up to remove VS2015. I'll try and get back to that soon.

mikebeaton:
> I'm all for removing these and in general auditing what we have set in tools def. I also agree with Ard and Leif that GCC48, GCC49, and GCC5 can all be removed. Same goes with VS2015 (I believe @bexcran signed up to do that, IIRC), and VS2017. Whether an RFC is needed or not I'll defer to the stewards.

I think GCC5 cannot be removed yet.

I there any chance that this can be accepted separately from removing GCC48, GCC49 or should I be looking at that too? :)

bexcran:
GCC5 has been marked as deprecated for a while though, and it should be simple for people to move to 'GCC' instead.

mikebeaton:
> c) get rid of GCC5, or keep it out of nostalgia, but document that only GCC 8 and older are still supported.

I am - perhaps incorrectly - reading the above from @ardbiesheuvel as really meaning "keep it due to the fact that quite a few people are probably still using it right now, even though they shouldn't be". I assume the quite literal reading is unlikely to be what is really meant (?!), but perhaps my proposed reinterpretation (which is, not coincidentally, what I think) is wrong.

os-d:
IMO, it is simple to mark it as a breaking change and have the integration instruction of move `GCC5` to `GCC`. Everything else still works.

> I there any chance that this can be accepted separately from removing GCC48, GCC49 or should I be looking at that too? :)

I'm fine to get this PR in before the removal, but I'm also not a BaseTools maintainer nor a steward, so I will defer to those folks :)

vishalo:
The soft freeze for the upcoming stable tag is scheduled for 2026‑02‑02. Would it be possible to defer the warning‑suppression‑disabling change until after that stable tag release? Deferring the change would provide downstream teams with adequate time to update and clean up their code.

leiflindholm:
@vishalo that feels like excessive caution for a trivially locally-revertable change.

vishalo:
Thanks @leiflindholm . I thought about this further - we can move it to downstream dsc file for now, instead of reverting. +1 to merge.

mikebeaton:
To avoid any duplicate work, I'll mention here that I'm having a go at stripping out GCC48, GCC49 and GCC5, to try to put up a PR in time to be considered for inclusion before the soft freeze, i.e. by early next week.

mikebeaton:
@ardbiesheuvel @leiflindholm - Hi! Is this one looking okay to merge? I've rebased it (thereby becoming slightly less controversial, as we don't have to decide whether or not to change GCC48, GCC49 or not, as they are no longer there). And I've rechecked it passes all CI in [GCC](https://github.com/tianocore/edk2/pull/11761/commits), [CLANGPDB](https://github.com/tianocore/edk2/pull/11906/commits), [CLANGDWARF](https://github.com/tianocore/edk2/pull/11905/commits).

PR #11769, opened 2025-11-18
MdeModulePkg: StatusCodeHandler Stmm remove assert.

apop5:
# Description

In StandaloneMM mode, IsStatusCodeUsingSerialPort is expecting to find gMmStatusCodeUseSerialHobGuid, and will assert if it is not found.

Change the logic so that if the Guided Hob is not found, to let the function return FALSE and progress without serial messages.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
A platform which did not provide a gMmStatusCodeUseSerialHobGuid would assert.
After changing, the platform proceeds to boot.

## Integration Instructions
No integration necessary.


apop5:
@niruiyu @lgao4 

Would it be possible to review this minor change?

lgao4 (approved):

PR #11775, opened 2025-11-19
Fix retry logic in UFS device detection APIs

NishanthSanjeevi:
# Description
Fix the logic in the UFS device detection APIs

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Boot tested

## Integration Instructions
N/A


mikebeaton:
@NishanthSanjeevi - Hi

The commits which introduce the code you are changing are
 - https://github.com/tianocore/edk2/commit/13d9e8ec98e
 - https://github.com/tianocore/edk2/commit/f426d8744f3
 
I does seem that both these commits break the previous retry logic, but also that a correct fix would need to take account of the code before and after those commits. E.g. before the change in both cases, anything other than EFI_NOT_FOUND was an immediate error, it looks to me as if that logic should be restored if you are restoring the retries. Just from reading the before and after code, it also looks to me as if the UfsWaitMemSet command, once it's reached, doesn't need to retry, it can probably be allowed to succeed or fail immediately with no retries.

Is restoring the retry logic here based on specific hardware which works when the retry is restored, but not otherwise?

PR #11776, opened 2025-11-19
MdeModulePkg/PciHostBridgeDxe: Ignore unsupported I/O BARs.

michaelescue:
# Description
Bootup fails when an endpoint requests I/O space BARs and the Root Bridge does not support I/O space causing a resource conflict during bus enumeration. While I/O space is traditionally supported in x86_64, platforms using an Arm architecture may or may not support I/O space.

This change adds logic allowing the platform to detect I/O space support for each Root Bridge and prevent submission of the required I/O resources for allocation when unsupported. Endpoints will not receive I/O resources, but they will continue to receive Memory BAR resources.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- Boot on ARM64 platform with a mix of PCIe endpoints that request and do not request I/O Space BARs.

## Integration Instructions

- N/A

### DIFF
--- a/MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciHostBridge.c
+++ b/MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciHostBridge.c
@@ -1569,6 +1569,14 @@ SubmitResources (
         RootBridge->ResAllocNode[Type].Length    = Descriptor->AddrLen;
         RootBridge->ResAllocNode[Type].Alignment = Descriptor->AddrRangeMax;
         RootBridge->ResAllocNode[Type].Status    = ResSubmitted;
+        //
+        // If the PCI root bridge does not support IO space, do not submit requested resources
+        // for allocation.
+        //
+        if ((Type == TypeIo) && (RootBridge->Io.Base > RootBridge->Io.Limit)) {
+          DEBUG ((DEBUG_INFO, " I/O: Resource not submitted. Unsupported aperture.\n"));
+          RootBridge->ResAllocNode[Type].Status = ResNone;
+        }
       }
 
       RootBridge->ResourceSubmitted = TRUE;


ardbiesheuvel (approved):
This looks like a reasonable workaround to me - it is a bit of a hack, but it gets rid of the system-wide PCD, and takes the properties of the individual root bridges into account.

The only thing missing here is a DEBUG() that explains that I/O resources are being disregarded because the I/O aperture is missing entirely.

MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciHostBridge.c
```diff
@@ -1568,7 +1568,15 @@ SubmitResources (
 
         RootBridge->ResAllocNode[Type].Length    = Descriptor->AddrLen;
         RootBridge->ResAllocNode[Type].Alignment = Descriptor->AddrRangeMax;
-        RootBridge->ResAllocNode[Type].Status    = ResSubmitted;
+        //
+        // If the PCI root bridge does not suport IO Space, do not submit requested resources
+        // for allocation.
+        //
+        if ((Type == TypeIo) && (RootBridge->Io.Base > RootBridge->Io.Limit)) {
```
mdkinney, line 1576: Could make this a more generic check for all 5 if the resource types considered(Mem32, PMem32, Mem64, PMem64,  Io). 

Could also keep the line unconditionally sets Status to ResSubmitted, and then set to ResNone if the Root Bridge does not support the requested Type.

ardbiesheuvel: > Could make this a more generic check for all 5 if the resource types considered(Mem32, PMem32, Mem64, PMem64, Io).
> 

In terms of logic, this might seem reasonable. But in practice, I don't think this makes a lot of sense. 64-bit MMIO BARs will already be allocated from 32-bit resource windows if needed, and similarly, prefetchable BARs will be allocated from non-prefetchable windows if there are insufficient resources. So we'll only hit this condition if there is not even a 32-bit non-prefetchable MMIO resource window, and all MMIO BARs will remain unallocated.

I/O resources are different because
a) only legacy PCIe endpoints are permitted to have I/O BARs in the first place;
b) such legacy endpoints are usually VGA compatible GPUs or IDE compatible SATA controllers, where the I/O BARs are never used when driving the device natively - only x86 systems will ever use the VGA or IDE legacy interfaces.

mdkinney: Good points.

I was thinking of the hypothetical case where MMIO resources below 4GB were not available from a root bridge, in which case Mem32 and Pmem32 would not be available, but Mem64 and PMem64 would be available.

I am ok with I/O only case now.

josephoresko:
Looks good to me.

michaelescue:
@ardbiesheuvel, @mdkinney, 
Thanks for the feedback! 

>The only thing missing here is a DEBUG() that explains that I/O resources are being disregarded because the I/O aperture is missing entirely.

I've added the debug print statement as requested.

>Could also keep the line unconditionally sets Status to ResSubmitted, and then set to ResNone if the Root Bridge does not support the requested Type.

I agree with this and have made the necessary change.

michaelescue:
Corrected spelling errors and capitalization.

PR #11777, opened 2025-11-21
VfrCompile bugfixes

mikebeaton:
# Description

Glibc 2.39 (in Ubuntu 24.04) runtime hardening revealed out of bounds memory accesses in VfrCompile, leading to crashes when compiling MdeModulePkg under that version.

Additional investigation with clang santizers revealed four separate bugs affecting VfrCompile which are fixed in these commits. A further similar bug had already been fixed recently within Acidanthera and is also included here.

See individual commit messages for details.

- [ ] Breaking change?
  - NO.
- [ ] Impacts security?
  - NO.
- [ ] Includes tests?
  - NO.

## How This Was Tested

Runtime sanitizers, single step debugging. Verified that VfrCompile `.lst` dump files generated during build are the same before and after changes, except for wanted alignment changes caused by fix by ee87218cdf8689cae07451bae6deb4a437fdbfe4.

## Integration Instructions

N/A


mikebeaton:
(Force pushed to streamline - but not logically change - fixed bitfield calculation.)

BaseTools/Source/C/Include/Common/BaseTypes.h
```diff
@@ -151,7 +151,13 @@ typedef struct {
 //
 // Macro that returns the byte offset of a field in a data structure.
 //
-#define OFFSET_OF(TYPE, Field) ((UINTN) &(((TYPE *)0)->Field))
+#if (defined (__GNUC__) && __GNUC__ >= 4) || defined (__clang__)
```
leiflindholm: GCC 3.4.6, the latest patch version of a pre-GCC4, was released nearly 20 years ago.
We could probably drop that >= 4.

BaseTools/Source/C/VfrCompile/VfrUtilityLib.cpp
```diff
@@ -1241,6 +1241,7 @@ CVfrVarDataTypeDB::DataTypeAddField (
   pNewField->mFieldType    = pFieldType;
   pNewField->mArrayNum     = ArrayNum;
   pNewField->mIsBitField   = FALSE;
+  pNewField->mBitWidth     = 0;
```
leiflindholm: Why is zero a safe value?
Comments on line 1178 above appears to suggest it has special meaning.

BaseTools/Source/C/Include/Common/UefiInternalFormRepresentation.h
```diff
@@ -777,6 +777,7 @@ typedef struct _EFI_IFR_FORM_SET {
   EFI_STRING_ID            FormSetTitle;
   EFI_STRING_ID            Help;
   UINT8                    Flags;
+  UINT8                    Align;
```
leiflindholm: If there are alignment requirements, why is the struct inside a `#pragma pack(1)` block?

MdePkg/Include/Uefi/UefiInternalFormRepresentation.h
```diff
@@ -879,6 +879,7 @@ typedef struct _EFI_IFR_FORM_SET {
   EFI_STRING_ID        FormSetTitle;
   EFI_STRING_ID        Help;
   UINT8                Flags;
+  UINT8                Align;
```
leiflindholm: If there are alignment requirements, why is the struct inside a `#pragma pack(1)` block?

BaseTools/Source/C/VfrCompile/VfrSyntax.g
```diff
@@ -895,16 +895,16 @@ vfrExtensionData[UINT8 *DataBuff, UINT32 Size, CHAR8 *TypeName, UINT32 TypeSize,
...
+                Offset = (FieldOffset - PreBits) / 8;
                 Mask <<= PreBits;
               }
               switch (FieldType) {
               case EFI_IFR_TYPE_NUM_SIZE_8:
                  Data_U8 = _STOU8(RD->getText(), RD->getLine());
                  if (BitField) {
                    //
-                   // Set the value to the bit fileds.
+                   // Set the value to the bit fields.
```
leiflindholm, line 907: This change is unrelated and non-functional. Please put it in a separate commit.

leiflindholm: In fact, it could go in last commit: "Minor code cleanup"

BaseTools/Source/C/VfrCompile/VfrError.cpp
```diff
@@ -35,7 +35,7 @@ static SVFR_ERROR_HANDLE VFR_ERROR_HANDLE_TABLE [] = {
   { VFR_RETURN_GET_NVVARSTORE_ERROR, ": get name value varstore error" },
   { VFR_RETURN_QVAR_REUSE, ": variable reused by more than one question" },
   { VFR_RETURN_FLAGS_UNSUPPORTED, ": flags unsupported" },
-  { VFR_RETURN_ERROR_ARRARY_NUM, ": array number error, the valid value is in (0 ~ MAX_INDEX-1) for UEFI vfr and in (1 ~ MAX_INDEX) for Framework Vfr" },
+  { VFR_RETURN_ERROR_ARRAY_NUM, ": array number error, the valid value is in (0 ~ MAX_INDEX-1) for UEFI vfr and in (1 ~ MAX_INDEX) for Framework Vfr" },
```
leiflindholm, line 38: Commit-message: please call out that this is a non-functional change.

BaseTools/Source/C/Include/Common/UefiInternalFormRepresentation.h
mikebeaton: Search me. But there are, and it is.

mikebeaton: I suspect some programmers - possibly including myself in the past - use `pack(1)` to mean 'I want to manually control alignment, here'.

BaseTools/Source/C/VfrCompile/VfrUtilityLib.cpp
mikebeaton: The comment above only applies when adding a bitfield field definition. It refers to cases like the following:

```
typedef struct {
  UINT16    NestByteField;
  UINT8                     : 1; // unamed field can be used for padding
  UINT8     NestBitCheckbox : 1;
  UINT8     NestBitOneof    : 2;
  UINT8                     : 0; // Special width 0 can be used to force alignment at the next word boundary
  UINT8     NestBitNumeric  : 4;
} MY_BITS_DATA;
```

The PR change is setting `mBitWidth` of a non-bitfield field definition to zero (as opposed to just leaving it uninitialised); e.g. the field definition that will be generated for `BitsData` here:

```
typedef struct {
  MY_BITS_DATA    BitsData;
  UINT32          EfiBitGrayoutTest : 5;
  UINT32          EfiBitNumeric     : 4;
  UINT32          EfiBitOneof       : 10;
  UINT32          EfiBitCheckbox    : 1;
  UINT32                            : 0; // Width 0 used to force alignment.
} MY_EFI_BITS_VARSTORE_DATA;
```

`BitsData` (e.g.) itself isn't a bitfield and in general its `mBitWidth` will therefore not be used, which presumably is why no one thought to initialise if (i.e. `mBitWidth` specifies the bit width _of that field, iff it's a bitfield_).

But it is in fact accessed (and needs be zero, for the code to work as expected), for fields _containing_ bitfields, in method `CVfrVarDataTypeDB::GetDataFieldInfo`, when traversing down through a nested type, specified by `VarStr`, such as: `"MY_EFI_BITS_VARSTORE_DATA.BitsData.NestBitCheckbox"`.

BaseTools/Source/C/Include/Common/UefiInternalFormRepresentation.h
mikebeaton: The misaligned access occurs on this line: https://github.com/tianocore/edk2/blob/9ff73832ea0440b86fa0dfa42158fce1ef3b57d2/BaseTools/Source/C/VfrCompile/VfrFormPkg.h#L684

MdePkg/Include/Uefi/UefiInternalFormRepresentation.h
mikebeaton: I'll switch to answering on just one of these.

BaseTools/Source/C/Include/Common/UefiInternalFormRepresentation.h
mikebeaton: It is worth noting the (IMO clearly wrong) alignment in the `.lst` dump files which VfrCompile was generating (always generates, during a normal build), vs. the sane alignment after this fix:

Screenshot 2025-11-21 at 12 38 28

Screenshot 2025-11-21 at 12 38 40

<br>Screenshot 2025-11-21 at 12 52 09

<br>All the `.lst` files generated (e.g.) during the build of MdeModulePkg benefit from the same alignment fix, which is caused by exactly this change:

Screenshot 2025-11-21 at 12 38 49

mhaeuser: > If there are alignment requirements, why is the struct inside a `#pragma pack(1)` block?

Multiple places in EDK II, including this, ignore the semantics of packed data. Packed data is only safe to access through its packed declaration and, to my knowledge, not really by any other means. In particular, as you surely know, it is *not* safe to create and dereference pointers to it: https://github.com/llvm/llvm-project/commit/722a4db1985ca9a8b982074d658dfee9c4624d53

I feel like EDK II suffers from a "packing abuse" where every single thing that is considered to need "interoperability" is packed, for some reason. Some seem to believe that's the best way to deal with different alignment requirements across ABIs and that throwing a pragma solves the issues. Well, as we see here, not really. Rather, definitions like PE, ELF, and Mach-O exploiting natural alignment (which is a given from the UEFI spec) are the safest solution.

Further notes:
- At least some of the structs in the block, this in particular, seem to *not* be packed by the UEFI specification. So, this change takes one step towards aligning it with the spec (unless I missed something).
- Adding internal alignment variables to packed structs is something done in other places in EDK II, but it is very awkward. While internal alignment is "fixed", external alignment remains the same and the compiler can place things at arbitrary locations. And direct accesses are still "slow" despite the stronger-than-really-packed guarantees. It feels like there are nothing but drawbacks to this paradigm.
- We have `STATIC_ASSERT` and `_Alignof`, so we can easily ensure correctness at compile time for un-packed struct declarations.

mikebeaton: @leiflindholm - pace my own and others' comments above, I am coming round to your way of thinking and am going to attempt a more conservative (and, I believe you may be right, more correct) fix here.

BaseTools/Source/C/VfrCompile/VfrFormPkg.h
```diff
@@ -681,7 +681,7 @@ class CIfrFormSet : public CIfrObj, public CIfrOpHeader {
   }
 
   VOID SetClassGuid (IN EFI_GUID *Guid) {
-    memmove (&(mClassGuid[mFormSet->Flags++]), Guid, sizeof (EFI_GUID));
+    memmove ((VOID *) &(mClassGuid[mFormSet->Flags++]), Guid, sizeof (EFI_GUID));
```
ardbiesheuvel: This is bizarre. What kind of diagnostic are you getting?

BaseTools/Source/C/Include/Common/BaseTypes.h
mikebeaton: A quick grep finds the following lines (three irrelevant results removed manually when pasting):
```
$ grep -r "__GNUC__.*[0-9]" | grep -Ev "/(googletest|libspdm|openssl|mbedtls|brotli)/"
MdePkg/Include/Base.h:#if (defined (__GNUC__) && __GNUC__ >= 4) || defined (__clang__)
RedfishPkg/Include/Library/RedfishCrtLib.h:#if defined (__GNUC__) && (__GNUC__ >= 2)
CryptoPkg/Library/Include/CrtLibSupport.h:#if defined (__GNUC__) && (__GNUC__ >= 2)
MdeModulePkg/Library/LzmaCustomDecompressLib/Sdk/C/CpuArch.h:                                       (defined (__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 3))) \
MdeModulePkg/Library/LzmaCustomDecompressLib/Sdk/C/7zTypes.h:#if defined (__GNUC__) && (__GNUC__ >= 4)
BaseTools/Source/C/Include/Common/BaseTypes.h:#if (defined (__GNUC__) && __GNUC__ >= 4) || defined (__clang__)
BaseTools/Source/C/LzmaCompress/Sdk/C/CpuArch.h:       (defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 3))) \
BaseTools/Source/C/LzmaCompress/Sdk/C/7zTypes.h:#if defined (__GNUC__) && (__GNUC__ >= 4)
```
So maybe this is more one to change in a separate commit? Or I could just not include the test here? Though it was specifically meant to match what is in MdePkg/Include/Base.h.

BaseTools/Source/C/VfrCompile/VfrUtilityLib.cpp
mikebeaton: There is probably a more correct fix, making this never accessed for non-bitfields, even if they are bitfield parents, though it's a bit fiddly. (This is certainly what the code was de facto doing, but explicitly setting to 0 something which needed to be 0, for how it currently works to not rely on uninitialized memory being zero.)

makubacki:
> ⚠ **WARNING: Cannot add some reviewers**: A user specified as a reviewer for this PR is not a collaborator of the repository. Please add them as a collaborator to the repository so they can be requested in the future.
> 
> Non-collaborators requested:
> 
> * @mikebeaton
> 
> Attn Admins:
> 
> * @jljusten
> * @ardbiesheuvel
> * @lgao4
> * @mdkinney
> * @makubacki
> * @jkmathews
> 
> **Admin Instructions:**
> 
> * Add the non-collaborators as collaborators to the appropriate team(s) listed in [teams](https://github.com/orgs/tianocore/teams)
> * If they are no longer needed as reviewers, remove them from [`Maintainers.txt`](https://github.com/tianocore/edk2/blob/HEAD/Maintainers.txt)

I've invited @mikebeaton to the "EDK II Reviewers" team.

mikebeaton:
> I've invited @mikebeaton to the "EDK II Reviewers" team.

Accepted - tyvm

BaseTools/Source/C/Include/Common/BaseTypes.h
leiflindholm: Yeah, I get the wanting to match, it just made me twitch a bit when I saw it.

But it looks problematic that several of these don't check for __clang__.

Actually, is this how we ended up with the weirdness of 12797dd337? Because clang doesn't set the GCC macros when called with a windows target?

mikebeaton: Removed in this commit.

BaseTools/Source/C/VfrCompile/VfrUtilityLib.cpp
mikebeaton: Made the more correct fix. Ty.

mikebeaton:
@leiflindholm - All your points addressed (I hope). Thank you for helpful feedback.

mikebeaton:
@leiflindholm - This is still tagged as pending requested changes from yourself, but I don't think there are any remaining? 🙏

leiflindholm:
> @leiflindholm - This is still tagged as pending requested changes from yourself, but I don't think there are any remaining? 🙏

Sorry, I went down a rabbit hole around the packed EFI_GUID thing, and got distracted.
I think this is a mess edk2 needs to detangle at some point, but this is not it.
(Grub handles this by having `grub_packed_guid_t` as well as `grub_guid_t` as an example.)

BaseTools/Source/C/VfrCompile/VfrFormPkg.h
leiflindholm: Could you add a comment here, along the lines of
// Cast destination to `VOID *` to deal with abuse of packed (1) in
// UefiInternalFormRepresentation.h without a full rewrite.
?

leiflindholm: Final comment, honest.

mikebeaton: Is this correct?

Obviously in the very first instance, the (VOID *) is there to quiet the sanitizers, for something they ought not to worry about (to allow VfrCompile to be run with sanitizers and - hopefully - to confirm there are no other catchable bugs).

But beyond that, the issue is copying _to_ an unaligned address. EFI_GUID *mClassGuid is an unaligned address - and it is meant to be, regardless of whether the packing is fixed or not. The serial VFR format does not require this GUID/these GUIDs to occur aligned on a multiple of 4, or anything else, bytes in the stream, and they don't.

That is the issue as I understand it. I'm not even sure the pack(1) _is_ being abused in this place where it occurs, after all the discussion, and given the nature of this serial format.

I hesitate to post this, but my understanding has been shaped by this summary of what the web knows about this(!):

image

mikebeaton: Ofc assuming you agree with all that and still want the comment, I'll just add it!

leiflindholm: Please post references or reasoning, not slop.

leiflindholm: In fact, your reasoning is fine, regardless of how you got to it. The slop is just aggravating noise.

ardbiesheuvel: What I don't understand is why the alignment matters at all when invoking `memmove()`, which takes void pointers with any alignment.

leiflindholm: OK, so if the destination is supposed to be packed, then it would seem to me that the destination should be packed, not cast at the point where the compiler complains.

I.e., the class definition should be inside a `#pragma packed (1)` block.

mikebeaton: Okay, so not yet approved, I'll try this and see if I can make a simple change which quiets the sanitizers and still produces the same output.

FWIW, the _compiler_ doesn't complain, the compiled-in sanitizer code complains, at runtime, in this case. But more or less same difference.

mikebeaton: Wait, this won't work. The class is:

```
  EFI_IFR_FORM_SET *mFormSet;
  EFI_GUID *mClassGuid;
```

i.e. pointers.

The correct fix is almost certainly to make an unaligned GUID type, as you say. But then my reservation about that is that there would (pretty clearly?) be multiple places in VfrCompile (at least, and possibly in Common as well? not sure?) where EFI_GUID should be replaced with UNALIGNED_EFI_GUID, and it would take _ages_ to correctly identify them all (and I probably wouldn't correctly identify them all, to be fair, however long I spent - I'm sure I'd end up with false positives and false negatives).

So two options are:

 - Create an UNALIGNED_EFI_GUID type, but _just_ apply it here.
 - Don't create an UNALIGNED_EFI_GUID, but do add a comment, saying that the correct fix would be to create an UNALIGNED_EFI_GUID, and then apply it correctly throughout VfrCompile.
 
 How does that reasoning stack up?

mikebeaton: I could certainly actually test that option 1 _works_; before doing option 2 in the PR; if that would be sufficient?

**EDIT**: Actually, maybe option 1 is at least a first step. Let me try it first, and report back.

ardbiesheuvel: Could you remind me what we are trying to fix here, and what the actual diagnostic message is?

mikebeaton:
```
VfrFormPkg.h:684:14: runtime error: store to misaligned address 0xe20000000017 for type 'EFI_GUID *', which requires 4 byte alignment
0xe20000000017: note: pointer points here
 00 00 00 01 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00
             ^ 
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior VfrFormPkg.h:684:14
```
https://github.com/tianocore/edk2/blob/master/BaseTools/Source/C/VfrCompile/VfrFormPkg.h#L684

`EFI_GUID *mClassGuid;`, defined not far above in that file, contains a misaligned pointer to an EFI_GIUD - and it is meant to, that is not an error.

We are generating a serial format which is going to be read byte-by-byte, and is not supposed to be aligned on any particular byte from the start of the stream - at least as I understand it. But we are generating it (or just a subset of it at a time? not sure?) in memory first, and doing it in a way where the semantics of our code say we do care.

In C++, having a pointer to an item which requires a certain alignment, but where the pointer does not have that alignment is UB, by definition, as I understand it. (Regardless of whether you only use unaligned operations on it or not.) Therefore the sanitizer is correct, and the code is wrong - even though its actual behaviour is not wrong.

You can silence the sanitizer with `(VOID *)`, but (I now understand) this _doesn't_ technically remove the UB violation - I guess the sanitizer is being pragmatic in that respect. Or you can fix the code so the semantics of what it says you are doing match the semantics of what you are doing.

ardbiesheuvel: > ```
> VfrFormPkg.h:684:14: runtime error: store to misaligned address 0xe20000000017 for type 'EFI_GUID *', which requires 4 byte alignment
> 0xe20000000017: note: pointer points here
>  00 00 00 01 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00
>              ^ 
> SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior VfrFormPkg.h:684:14
> ```
> 
> https://github.com/tianocore/edk2/blob/master/BaseTools/Source/C/VfrCompile/VfrFormPkg.h#L684
> 
> `EFI_GUID *mClassGuid;`, defined not far above in that file, contains a misaligned pointer to an EFI_GIUD - and it is meant to, that is not an error.
> 
> We are generating a serial format which is going to be read byte-by-byte, and is not supposed to be aligned on any particular byte from the start of the stream - at least as I understand it. But we are generating it (or just a subset of it at a time? not sure?) in memory first, and doing it in a way where the semantics of our code say we do care.
> 
> In C++, having a pointer to an item which requires a certain alignment, but where the pointer does not have that alignment is UB, by definition, as I understand it. (Regardless of whether you only use unaligned operations on it or not.) Therefore the sanitizer is correct, and the code is wrong - even though its actual behaviour is not wrong.
> 
> You can silence the sanitizer with `(VOID *)`, but (I now understand) this _doesn't_ technically remove the UB violation - I guess the sanitizer is being pragmatic in that respect. Or you can fix the code so the semantics of what it says you are doing match the semantics of what you are doing.

It looks like the compiler is doing idiom recognition, i.e., it sees a `memmove()` of a fixed size, and replaces the function call with actual loads and stores. Can you try adding `-fno-builtin-memmove` to the compiler command line to see if that silences the diagnostic?

mikebeaton: I can. But I think @leiflindholm and I would now be in agreement that what our code currently does _is_ undefined behaviour, whatever we do to silence the warning. And I can already silence the warning.

From C23 (ISO/IEC 9899:2024) (or rather [N3220 working draft thereof](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3220.pdf)) I found:

6.2.8 Alignment of objects
§1 Complete object types have alignment requirements which place restrictions on the addresses at
which objects of that type may be allocated. 

6.2.5 Types
…
§25
…
 — A pointer type may be derived from a function type or an object type, called the referenced type. A
pointer type describes an object whose value provides a reference to an entity of the referenced
type. A pointer type derived from the referenced type T is sometimes called "pointer to T".
The construction of a pointer type from a referenced type is called "pointer type derivation".
…

Which I believe is enough to see the logic: you can't _have_ an unaligned pointer to an object of an aligned type. This is _already_ undefined behaviour. At which point, we either find a (or more than one) way to silence it, or we fix it.

mikebeaton: Yes, I tried it and `-fno-builtin-memmove` silences the warning. But I would currently still feel that this is a red herring as regards the correct solution. (You and I have found different ways to stop the clang sanitizers (I'm using clang, to get better sanitizers; for all I know this can be found with gcc too, I did not check this) from warning about this, but neither of these fix the fact that this _is_ UB - on its face, as it were.)

mikebeaton:
I have redone the commit in the series to fix the particular, tricky bug under discussion. I no longer claim in the commit message that it is an 'incorrect' warning. I have fixed it by introducing an UNALIGNED_EFI_GUID type. Confirmed fixed, and still producing the same, correct, output as originally.

BaseTools/Source/C/VfrCompile/VfrFormPkg.h
ardbiesheuvel: It is only UB if you access a misaligned object through a pointer to the object's type. This also explains why it is a runtime error rather than a compile-time error: `memmove()` takes pointers to void, and so the `(VOID *)` cast is implicit, and that is why it is not caught by the build time sanitizer.

However, the `memmove()` gets lowered to a struct assignment of type `EFI_GUID`, and the runtime instrumentation catches the misalignment.

Btw I don't understand why `memmove()` is being used here in the first place - `memcpy()` would seem more appropriate to me, and given the semantics wrt overlaps, I can see why the compiler would be more eager to replace a `memmove()` of a compile-time constant size with loads followed by stores, as it guarantees the correct semantics.

mhaeuser: @ardbiesheuvel No, creating unaligned pointers is UB on its own. This is also the guarantee that allows the compiler to make this substitution in the first place, as otherwise the compiler would *introduce* UB into this example. In fact, pointers of different types don’t even need to have the same size or representation. Though to be fair, in the EDK II context, such compiler optimizations are the only realistic risk from this UB, so I think the flag is fine pragmatically.

I also think even memcpy would be substituted at least sometimes, as this is basically two 64-bit assignments. Not suitable for a general pattern, albeit sure, the better choice compared to memmove.

In fact, as we are in userland and this is a reasonably small type, a direct assignment might have been just as fine (and would have avoided any chance of UB). The compiler would optimize it to memcpy if it sees fit.

mikebeaton: @ardbiesheuvel - Not to add too much noise, but I hadn't fully grasped that your `-fno-builtin-memmove` wasn't just ('somehow') silencing the warning, but was actually changing the code to avoid the unaligned accesses that we (probably?) care about more than the technical UB of even storing an unaligned value in a pointer to an aligned type (if I understand rightly).

I guess my original (VOID *) cast was doing the same - but just in one place, whereas the compiler flag would affect any other places where this happens (and ofc minorly things slow down by avoiding genuine, valid optimisations, too).

I think an advantage of fixing the unaligned access by explicitly changing our code so that there is no (even theoretical) undefined behaviour, is that it might provide a pointer (no pun intended) to a better pattern for dealing with this stuff in the first place, in the future.

mikebeaton: I have updated the comment of my new commit (the one with EFI_GUID in the title) to reflect all this.

leiflindholm: > I also think even memcpy would be substituted at least sometimes, as this is basically two 64-bit assignments. Not suitable for a general pattern, albeit sure, the better choice compared to memmove.

@mhaeuser yeah, the problem is the x86-folks consider it rational behaviour to carefully craft their C code to avoid all situations where the compiler might generate a memcpy call, so they don't need to depend on a compilerintrinsicslib.

BaseTools/Source/C/VfrCompile/VfrFormPkg.h
```diff
@@ -655,8 +655,8 @@ static BOOLEAN            gIsStringOp = FALSE;
  */
 class CIfrFormSet : public CIfrObj, public CIfrOpHeader {
 private:
-  EFI_IFR_FORM_SET *mFormSet;
-  EFI_GUID *mClassGuid;
+  EFI_IFR_FORM_SET    *mFormSet;
+  UNALIGNED_EFI_GUID  *mClassGuid;
```
ardbiesheuvel: I don't think we need another type here: `mClassGuid` is never interpreted as a GUID, it is only used to keep track of an array of 16-byte objects in memory.

So this should be equivalent, I think:

```diff
--- a/BaseTools/Source/C/VfrCompile/VfrFormPkg.h
+++ b/BaseTools/Source/C/VfrCompile/VfrFormPkg.h
@@ -656,7 +656,7 @@ static BOOLEAN            gIsStringOp = FALSE;
 class CIfrFormSet : public CIfrObj, public CIfrOpHeader {
 private:
   EFI_IFR_FORM_SET *mFormSet;
-  EFI_GUID *mClassGuid;
+  UINT8 (*mClassGuid)[sizeof(EFI_GUID)];
 
 public:
   CIfrFormSet (UINT8 Size) : CIfrObj (EFI_IFR_FORM_SET_OP, (CHAR8 **)NULL, Size),
@@ -665,7 +665,7 @@ class CIfrFormSet : public CIfrObj, public CIfrOpHeader {
     mFormSet->FormSetTitle = EFI_STRING_ID_INVALID;
     mFormSet->Flags        = 0;
     memset (&mFormSet->Guid, 0, sizeof (EFI_GUID));
-    mClassGuid = (EFI_GUID *) (mFormSet + 1);
+    mClassGuid = (UINT8 (*)[sizeof(EFI_GUID)]) (mFormSet + 1);
   }
 
   VOID SetGuid (IN EFI_GUID *Guid) {
```

BaseTools/Source/C/VfrCompile/VfrFormPkg.h
mhaeuser: @leiflindholm Feel you… Yet we would get better optimizations and better “UB feedback”. :) memcpy is not even intrinsic, so the usual “they are unstable” (when they really aren’t) point doesn’t even apply. Oh well. I think both the flag and the packed type are nice options, the latter of course preferable if the overhead of the change is low enough.

Circling back to the spec, would be nice if someone could check what I missed regarding packing there, because these structs do not feature the pragma there (yet this clearly is the intention). Does it need a fix?

BaseTools/Source/C/VfrCompile/VfrFormPkg.h
mikebeaton: That feels wrong to me. We are precisely storing an unaligned GUID here. Why not just say that we are? I feel reassured by @leiflindholm's presumably fully legitimate claim that other codebases have felt the need to do the same in the same circumstances.

mikebeaton: Plus also this (the actually affected line) will become wrong without changes, no?

```
memmove (&(mClassGuid[mFormSet->Flags++]), Guid, sizeof (EFI_GUID));
```

ardbiesheuvel: > Plus also this (the actually affected line) will become wrong without changes, no?
> 
> ```
> memmove (&(mClassGuid[mFormSet->Flags++]), Guid, sizeof (EFI_GUID));
> ```

Why?

mikebeaton: Oh right, no, sorry. It's certainly not as simple to read, but fair enough.

ardbiesheuvel: > That feels wrong to me. We are precisely storing an unaligned GUID here. Why not just say that we are? I feel reassured by @leiflindholm's presumably fully legitimate claim that other codebases have felt the need to do the same in the same circumstances.

It is not a GUID: the program never interprets it as a GUID, until the `memmove()` gets turned into a struct assignment.

`mClassGuid` basically only keeps track of a serialization buffer. So either we type it as a pointer to an array of 16 byte objects, as I suggested, or we just use a `UINT8*`, and increment it by `sizeof(EFI_GUID)` after the copy.

BaseTools/Source/C/VfrCompile/VfrFormPkg.h
ardbiesheuvel: > @ardbiesheuvel No, creating unaligned pointers is UB on its own. This is also the guarantee that allows the compiler to make this substitution in the first place, as otherwise the compiler would _introduce_ UB into this example.

You are correct, of course.

BaseTools/Source/C/VfrCompile/VfrFormPkg.h
mikebeaton: Well, I'm happy to make the change. It fixes it - i.e. removes the unaligned access and removes even theoretical undefined behaviour. I don't prefer it, I think it is less expressive of what the code is doing, but I already said that.

Can we just wait for @leiflindholm to either confirm or deny that he thinks that this is preferable too?

leiflindholm: (Apologies, my gmail appears to be having caching issues today, notifications lag by hours.)

I'm OK with either. My preference is for the expressiveness of a dedicated type, but that conversation can be had elsewhere and later.

ardbiesheuvel: Note that redefining the struct entirely is conceptually flaky, because the packing affects both padding and alignment. EFI_GUID does not have any padding, so it makes no difference in practice. But iff we go with a new type, it should be something like

```
#pragma pack(1)
typedef struct {
    EFI_GUID Guid;
} PACKED_EFI_GUID;
```

mikebeaton: > I'm OK with either. My preference is for the expressiveness of a dedicated type, but that conversation can be had elsewhere and later.

Without trying to initiate said discussion, nor demand that it should be had at any particular time, or at all, I was already wondering where such discussions are had. This https://github.com/tianocore/edk2/discussions ? (It doesn't seem to be used very much?) edk2-devel email list? edk2-discuss mailing list (it seems to be hardly used at all). Other more private and insider channels (which I don't doubt are sometimes the appropriate venue for more speculative or pre-release discussions)? Genuinely just wondering!

mikebeaton:
I've implemented the third commit with:
```
#pragma pack(1)
typedef struct {
    EFI_GUID Guid;
} PACKED_EFI_GUID;
#pragma pack()
```

mikebeaton:
@leiflindholm - sorry - ping - approval for this PR appears to still be hanging on your request for changes. I think I gather correctly that this version would be acceptable, at least, to @ardbiesheuvel - does it look okay to you now too?

leiflindholm:
> @leiflindholm - sorry - ping - approval for this PR appears to still be hanging on your request for changes. I think I gather correctly that this version would be acceptable, at least, to @ardbiesheuvel - does it look okay to you now too?

I got the impression @ardbiesheuvel didn't like the packed EFI_GUID approach.
I'm happy with this version, but would prefer consensus.

mikebeaton:
@leiflindholm @ardbiesheuvel - I got the impression [this](https://github.com/tianocore/edk2/pull/11777#discussion_r2565493312) was a concession to that approach and would be acceptable? I also moved it out of the header file which is concerned with things which are in the spec (where I had the previous less preferred version of the struct), as such stuff isn't in the spec.

mdkinney:
@mergify rebase

PR #11778, opened 2025-11-21
UefiPayloadPkg: Setup temporary GDT for x64 UniversalPayload

AjanZhong:
Separate Exception Stacks has been enabled by default in commit [1], in this case, TSS will be appended to original GDT.

As secondary bootloader, UniversalPayload solution has different boot flow from original EDK2. When DxeMain tries to append TSS after original GDT, the original GDT is empty. It leads to system reboot when DxeMain tries to install new GDT which is appended with TSS.

To fix this issue, set up temporary GDT with 64-bit code and data descriptors.

[1] https://github.com/tianocore/edk2/commit/cec2c6

# Description

Separate Exception Stacks has been enabled, however, in original implementation, GDT has not been enabled yet for UniversalPayload, it leads to new GDT with TSS appended is illegal format for GDT.
Add temporary GDT for UniversalPayload x64 solution to ensure separated exception stack can be enabled successfully.

- [ ] Breaking change?
  - N/A
- [ ] Impacts security?
  - N/A
- [ ] Includes tests?
  - N/A

## How This Was Tested

Validated on UniversalPayload x64 solution on top of QEMU

## Integration Instructions

N/A

### DIFF
--- a/UefiPayloadPkg/Library/FdtParserLib/FdtParserLib.c
+++ b/UefiPayloadPkg/Library/FdtParserLib/FdtParserLib.c
@@ -841,21 +841,9 @@ ParsePciRootBridge (
       UINT32  TripletCells = ChildAddrCells + ParentAddrCells + SizeCells;
 
       for (Base = 0; Base < TempLen / sizeof (UINT32); Base = Base + TripletCells) {
-        UINT64  ChildBusAddress  = 0;
         UINT64  ParentBusAddress = 0;
         UINT64  DmaRangeSize     = 0;
-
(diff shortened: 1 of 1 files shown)


AjanZhong:
> is this needed for non-fit boot path as well?

It should be. However, we have no validation environment for non-fit solution. Need members who are working on non-fit solution to double check whether non-fit solution has same issue.

AjanZhong:
@gguo11837463 @gdong1 @ChaselChiu Hi, would you mind help review this patch when you are available. Thanks.

PR #11785, opened 2025-11-23
ArmPkg: Fixes to MiscChassisManufacturerFunction

mikebeaton:
# Description

Initial error in code for copying ContainedElements found while working on
 enabling CLANGDWARF CI, which reports:

ERROR - Compiler #error from /__w/1/s/ArmPkg/Universal/Smbios/SmbiosMiscDx
e/Type03/MiscChassisManufacturerFunction.c variable 'ContainedElements' is
uninitialized when passed as a const pointer argument here [-Werror,-Wunin
itialized-const-pointer]

The code should copy the optional ContainedElements from InputData.
 Note that as per SMBIOS Specification 3.8.0, an additional SKU Number
 byte comes immediately after ContainedElements (this was handled but
 less explicit in the code before).

Additionally:

 - Correct size calculations in original code, since the base size for
calculations should not include ContainedElements.
 - Add code to ASSERT and return EFI_OUT_OF_RESOURCES if Hdr.Length would
overflow.
 - Ensure that debug error message is produced even after early exits.
 - Change from AllocateZeroPool to AllocatePool since all bytes are
initialized, after changes in this PR.
 - Switch to using the exactly correct type for SkuNumberField.

Fixes: https://github.com/tianocore/edk2/commit/bfc0fae459543442f3f17e0de655a72aba5b0920

- [ ] Breaking change?
  - NO.
- [ ] Impacts security?
  - NO.
- [ ] Includes tests?
  - NO.

## How This Was Tested

Tested for fixing Build_CLANGDWARF TARGET_ARM_ARMPLATFORM only.

## Integration Instructions

N/A

### DIFF
--- a/UefiCpuPkg/ResetVector/Vtf0/Ia32/SearchForSecEntry.asm
+++ b/UefiCpuPkg/ResetVector/Vtf0/Ia32/SearchForSecEntry.asm
@@ -21,7 +21,7 @@ BITS    32
 Flat32SearchForSecEntryPoint:
 
     ;
-    ; Initialize EBP and ESI to 0
+    ; Initialize EBX and ESI to 0
     ;
     xor     ebx, ebx
     mov     esi, ebx
--- a/UefiCpuPkg/ResetVector/Vtf0/ReadMe.txt
+++ b/UefiCpuPkg/ResetVector/Vtf0/ReadMe.txt
@@ -1,10 +1,10 @@
 
 === HOW TO USE VTF0 ===
 Add this line to your DSC [Components.IA32] or [Components.X64] section:
-  UefiCpuPkg/ResetVector/Vtf0/ResetVector.inf
+  UefiCpuPkg/ResetVector/Vtf0/Vtf0.inf
 
 Add this line to your FDF FV section:
-  INF  RuleOverride=RESET_VECTOR UefiCpuPkg/ResetVector/Vtf0/ResetVector.inf
+  INF  RuleOverride=RESET_VECTOR UefiCpuPkg/ResetVector/Vtf0/Vtf0.inf
 
 In your FDF FFS file rules sections add:
   [Rule.Common.SEC.RESET_VECTOR]


mikebeaton:
cc @bexcran

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
```diff
@@ -143,7 +143,7 @@ SMBIOS_MISC_TABLE_FUNCTION (MiscChassisManufacturer) {
 
   // ContainedElements
   ASSERT (ContainedElementCount < 2);
-  (VOID)CopyMem (SmbiosRecord + 1, &ContainedElements, ExtendLength);
+  (VOID)CopyMem (&ContainedElements, SmbiosRecord + 1, ExtendLength);
```
ardbiesheuvel: This doesn't make sense: `SmbiosRecord + 1` points to a zeroed region in the pool allocation.

Shouldn't this be copying the additional ContainedElements from InputData to `SmBiosRecord + 1` if the count > 1 ? (`SMBIOS_TABLE_TYPE3` contains a single element but it is intended to be a flexible array)

mikebeaton: @ardbiesheuvel - Can you review the revised version? The calculations are not all exactly the same, but they appear to be right to me.

mikebeaton: I actually also have no idea why we would have `ASSERT (InputData->ContainedElementCount < 2);` (modifed version, but same logic as before, I believe).

@bexcran I believe your commit initially, do you have a moment to comment?

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
```diff
@@ -135,25 +132,19 @@ SMBIOS_MISC_TABLE_FUNCTION (MiscChassisManufacturer) {
...
+  // Copy base record plus ContainedElements.
+  (VOID)CopyMem (SmbiosRecord, InputData, BaseSize + ExtendSize);
 
-  SmbiosRecord->Hdr.Length = sizeof (SMBIOS_TABLE_TYPE3) + ExtendLength + 1;
+  SmbiosRecord->Hdr.Length = BaseSize + ExtendSize;
 
   SmbiosRecord->Type = OemGetChassisType ();
 
-  // ContainedElements
-  ASSERT (ContainedElementCount < 2);
```
mikebeaton, line 145: I'm not sure why we would have needed this ASSERT?

@bexcran I believe your commit initially, do you have any time to comment?

MdePkg/Include/IndustryStandard/SmBios.h
```diff
@@ -532,7 +532,9 @@ typedef struct {
   UINT8                  ContainedElementCount;
   UINT8                  ContainedElementRecordLength;
   //
-  // Can have 0 to (ContainedElementCount * ContainedElementRecordLength) contained elements
+  // Occupies strictly (ContainedElementCount * ContainedElementRecordLength) bytes.
+  // ContainedElementRecordLength is not required to be sizeof(CONTAINED_ELEMENT).
+  // Hdr.Length includes ContainedElements and any additional fields SKUNumber etc. below.
```
ardbiesheuvel: Please drop this bit. If you really think it is needed, move the change to a separate patch.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
```diff
@@ -98,71 +93,72 @@ SMBIOS_MISC_TABLE_FUNCTION (MiscChassisManufacturer) {
...
+  AssertTagStrSize = StrSize (AssertTag);
 
   TokenToGet       = STRING_TOKEN (STR_MISC_CHASSIS_SKU_NUMBER);
   ChassisSkuNumber = HiiGetPackageString (&gEfiCallerIdGuid, TokenToGet, NULL);
-  ChaNumStrLen     = StrLen (ChassisSkuNumber);
+  ChaNumStrSize    = StrSize (ChassisSkuNumber);
 
-  ContainedElementCount = InputData->ContainedElementCount;
-  ExtendLength          = ContainedElementCount * sizeof (CONTAINED_ELEMENT);
+  BaseSize   = sizeof (SMBIOS_TABLE_TYPE3) - sizeof (CONTAINED_ELEMENT);
```
ardbiesheuvel: This is risky. It might work in this particular case, but in general, you should take padding into account, which might precede the `ContaintedElements[1]` member. Better to use `OFFSET_OF(SMBIOS_TABLE_TYPE3, ContainedElemenst)` here - this is also guaranteed to work as expected regardless of whether `ContainedElements` is a fake flex array `[1]` or a true one `[]`.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
```diff
@@ -98,71 +93,72 @@ SMBIOS_MISC_TABLE_FUNCTION (MiscChassisManufacturer) {
...
+  ExtendSize = (UINTN)InputData->ContainedElementCount * (UINTN)InputData->ContainedElementRecordLength;
+
+  ASSERT (BaseSize == 0x15);
```
ardbiesheuvel: Doesn't the value `0x15` include the single `ContainedElements` array element?

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
```diff
@@ -98,71 +93,72 @@ SMBIOS_MISC_TABLE_FUNCTION (MiscChassisManufacturer) {
 
   TokenToGet   = STRING_TOKEN (STR_MISC_CHASSIS_MANUFACTURER);
   Manufacturer = HiiGetPackageString (&gEfiCallerIdGuid, TokenToGet, NULL);
-  ManuStrLen   = StrLen (Manufacturer);
+  ManuStrSize  = StrSize (Manufacturer);
```
ardbiesheuvel: Why are these changes needed? The SMBIOS record is populated with ASCII versions of these strings, no? So the length of the string (i.e., the number of CHAR16 characters) +1 equals the number of bytes needed to accommodate the ASCII version.

mikebeaton: Oh, I was trying to avoid multiple +1`s, but I've got it wrong, haven't I, as I'll now use using the original CHAR16 string size, not the new ASCII string size. Doh. Sorry.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
mikebeaton: Despite being an odd number, no, it doesn't:
https://github.com/tianocore/edk2/blob/master/MdePkg/Include/IndustryStandard/SmBios.h#L545
https://github.com/tianocore/edk2/blob/master/ShellPkg/Library/UefiShellDebug1CommandsLib/SmbiosView/PrintInfo.c#L453-L455
Screenshot 2025-11-24 at 11 01 40

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
mikebeaton: Updated, still converting to using size not length, but hopefully getting it right this time.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
```diff
@@ -98,71 +93,72 @@ SMBIOS_MISC_TABLE_FUNCTION (MiscChassisManufacturer) {
...
+  TokenToGet       = STRING_TOKEN (STR_MISC_CHASSIS_ASSET_TAG);
+  AssertTag        = HiiGetPackageString (&gEfiCallerIdGuid, TokenToGet, NULL);
 
   TokenToGet       = STRING_TOKEN (STR_MISC_CHASSIS_SKU_NUMBER);
   ChassisSkuNumber = HiiGetPackageString (&gEfiCallerIdGuid, TokenToGet, NULL);
-  ChaNumStrLen     = StrLen (ChassisSkuNumber);
 
-  ContainedElementCount = InputData->ContainedElementCount;
-  ExtendLength          = ContainedElementCount * sizeof (CONTAINED_ELEMENT);
+  BaseSize   = OFFSET_OF(SMBIOS_TABLE_TYPE3, ContainedElemenst);
```
ardbiesheuvel: This doesn't even build. Please take into consideration that every time you push a new version, everyone who has interacted with the PR will be notified by email, so please test your code locally before pushing it.

mikebeaton: Right, was not aware. Thank you.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
ardbiesheuvel: Ah ok - I missed the `[4]` on the `OemDefined` field, and the additional 3 bytes are exactly the size of the `ContainedElements[1]`.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
mikebeaton: Can revert to storing StrLen and +1-ing three times per string, or leave as is now, as preferred. (Presumably the compiler may well do the same anyway, but I also find the revised version more readable.) Or could move just this part of the change to separate, second commit, if (really) preferred. (Even though I'm really only working on this, and similar, because they are blockers to proof-of-concept for CLANGDWARF and CLANGPDB CI - but c'est la vie, I guess! And it would kind of defeat the purpose of doing that, to be fair, if I - or someone - didn't fix the errors it finds!)

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
```diff
@@ -98,71 +93,72 @@ SMBIOS_MISC_TABLE_FUNCTION (MiscChassisManufacturer) {
...
+  AssertTagStrSize = StrLen (AssertTag) + 1;
 
   TokenToGet       = STRING_TOKEN (STR_MISC_CHASSIS_SKU_NUMBER);
   ChassisSkuNumber = HiiGetPackageString (&gEfiCallerIdGuid, TokenToGet, NULL);
-  ChaNumStrLen     = StrLen (ChassisSkuNumber);
+  ChaNumStrSize    = StrLen (ChassisSkuNumber) + 1;
 
-  ContainedElementCount = InputData->ContainedElementCount;
-  ExtendLength          = ContainedElementCount * sizeof (CONTAINED_ELEMENT);
+  STATIC_ASSERT (OFFSET_OF(SMBIOS_TABLE_TYPE3, ContainedElements) == 0x15, "Base size of SMBIOS_TABLE_TYPE3 does not meet SMBIOS specification");
```
mikebeaton: Switched to STATIC_ASSERT here.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
bexcran: @mikebeaton Sorry, I don't remember this code at all. I don't know why we'd have needed the ASSERT.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
```diff
@@ -98,71 +93,72 @@ SMBIOS_MISC_TABLE_FUNCTION (MiscChassisManufacturer) {
...
-                 ExtendLength    + 1 +
-                 ManuStrLen      + 1 +
-                 VerStrLen       + 1 +
-                 SerialNumStrLen + 1 +
-                 AssertTagStrLen + 1 +
-                 ChaNumStrLen    + 1 + 1;
-  SmbiosRecord = AllocateZeroPool (RecordLength);
+  RecordLength = BaseSize                         +
+                 ExtendSize                       +
+                 sizeof (SMBIOS_TABLE_STRING)     + ///< SKUNumber
```
bexcran: This seems an odd place for a Doxygen comment. Should it just be a normal `//` comment instead?

mikebeaton: I thought that `///<` was just a standard in the codebase for a single line comment that very briefly clarifies something about the line, but doesn't merit the full three line (minimum) official comment format?

I guess I'm (also) saying that I don't think it is pulled out by Doxygen, since AFAICT that wouldn't make sense here or in the many other places where it's used (I think?!) as I say.

mikebeaton: Right - I guess you're right, and I misunderstood when to use this. Will change. Thank you.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
mikebeaton: Think I should resolve comment and leave it removed, if neither @ardbiesheuvel nor yourself see a need to keep it.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
```diff
@@ -98,71 +94,79 @@ SMBIOS_MISC_TABLE_FUNCTION (MiscChassisManufacturer) {
...
-  ContainedElementCount = InputData->ContainedElementCount;
-  ExtendLength          = ContainedElementCount * sizeof (CONTAINED_ELEMENT);
+  //
+  // Length of SMBIOS struct includes ContainedElements and SKUNumber.
+  //
+  HdrLength = BaseSize + ExtendSize + sizeof (SMBIOS_TABLE_STRING);
+  if (HdrLength > MAX_UINT8) {
+    ASSERT (FALSE);
+    return EFI_OUT_OF_RESOURCES;
+  }
```
mikebeaton: Avoid duplication of this calculation, which also allows ASSERT and error return if it should overflow.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
```diff
@@ -208,5 +204,15 @@ SMBIOS_MISC_TABLE_FUNCTION (MiscChassisManufacturer) {
...
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "[%a]:[%dL] Smbios Type03 Table Log Failed! %r \n",
+      __func__,
+      DEBUG_LINE_NUMBER,
+      Status
+      ));
+  }
+
```
mikebeaton: Debug error message was not triggered on early exits (there already was one, now there are two).

ardbiesheuvel:
Sorry for being difficult, but any patch titled 'various fixes' needs to be split up. This is messy code, and so each change needs to be obviously correct and reviewable by others (or our future selves) `n` years from now without having to page in lots of history.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
```diff
@@ -98,71 +94,82 @@ SMBIOS_MISC_TABLE_FUNCTION (MiscChassisManufacturer) {
...
+  OptionalStrPos += VerStrSize;
+  UnicodeStrToAsciiStrS (SerialNumber, OptionalStrPos, SerialNumStrSize);
+  OptionalStrPos += SerialNumStrSize;
+  UnicodeStrToAsciiStrS (AssertTag, OptionalStrPos, AssertTagStrSize);
+  OptionalStrPos += AssertTagStrSize;
+  UnicodeStrToAsciiStrS (ChassisSkuNumber, OptionalStrPos, ChaNumStrSize);
+  OptionalStrPos += ChaNumStrSize;
+  *OptionalStrPos++ = '\0';
+
+  ASSERT ((UINT8 *)OptionalStrPos - (UINT8 *)SmbiosRecord == RecordLength);
```
mikebeaton: Debug ASSERT - double check no bugs introduced here or in future edits.

mikebeaton:
> Sorry for being difficult, but any patch titled 'various fixes' needs to be split up. This is messy code, and so each change needs to be obviously correct and reviewable by others (or our future selves) `n` years from now without having to page in lots of history.

@ardbiesheuvel - I have separated out the commits, and reduced variable renaming to the bare minimum, to try to make it clear what each change is for. Each individual commit builds and passes uncrustify locally.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
```diff
@@ -116,14 +113,25 @@ SMBIOS_MISC_TABLE_FUNCTION (MiscChassisManufacturer) {
...
+
+  BaseSize     = OFFSET_OF (SMBIOS_TABLE_TYPE3, ContainedElements);
+  ExtendLength = (UINTN)InputData->ContainedElementCount * (UINTN)InputData->ContainedElementRecordLength;
+
+  //
+  // Length of SMBIOS struct includes ContainedElements and SKUNumber.
+  //
+  HdrLength = BaseSize + ExtendLength + sizeof (SMBIOS_TABLE_STRING);
+  if (HdrLength > MAX_UINT8) {
+    ASSERT (FALSE);
```
ardbiesheuvel: Please don't use `ASSERT (FALSE)`, as it produces useless diagnostic messages at runtime. In this case, better to use `ASSERT (HdrLength <= MAX_UINT8)`

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
```diff
@@ -155,7 +155,7 @@ SMBIOS_MISC_TABLE_FUNCTION (MiscChassisManufacturer) {
   // The string numbers in the fixed position portion of the record are populated in the input data.
   *SkuNumberField = 5;
 
-  StrStart = (CHAR8 *)((UINT8 *)SmbiosRecord + HdrLength);
+  StrStart = (CHAR8 *)((UINT8 *)SkuNumberField + sizeof (SMBIOS_TABLE_STRING));
```
ardbiesheuvel, line 158: What is the point of this change? I take it is supposed to produce the same value?

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
```diff
@@ -39,11 +39,11 @@ SMBIOS_MISC_TABLE_FUNCTION (MiscChassisManufacturer) {
   CHAR8               *StrStart;
   UINT8               *SkuNumberField;
   UINTN               RecordLength;
-  UINTN               ManuStrLen;
-  UINTN               VerStrLen;
-  UINTN               AssertTagStrLen;
-  UINTN               SerialNumStrLen;
-  UINTN               ChaNumStrLen;
+  UINTN               ManuStrSize;
```
ardbiesheuvel: Please keep the original names, even if you are moving the `+ 1` around. Length vs size has a very different meaning in the context of `CHAR16` strings, so I'd prefer to avoid using 'size' here altogether.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
mikebeaton: We've literally had the same discussion in our team, and reached the opposite conclusion: ASSERT with the explicit condition requires a future coder to correctly update in both places, and ASSERT (FALSE) isn't a useless message, it comes with a line number and when you go to the line number you see what failed.

FWIW, when the `if` condition is complex, it makes the issue even more pointed. (I realise it's still possible to reach the opposite conclusion, I just wanted to make the case for what I've done.)

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
```diff
@@ -156,16 +155,18 @@ SMBIOS_MISC_TABLE_FUNCTION (MiscChassisManufacturer) {
   // The string numbers in the fixed position portion of the record are populated in the input data.
   *SkuNumberField = 5;
 
-  OptionalStrStart = (CHAR8 *)((UINT8 *)SmbiosRecord + HdrLength);
-  UnicodeStrToAsciiStrS (Manufacturer, OptionalStrStart, ManuStrLen + 1);
-  StrStart = OptionalStrStart + ManuStrLen + 1;
+  StrStart = (CHAR8 *)((UINT8 *)SmbiosRecord + HdrLength);
```
ardbiesheuvel: I realize this was in the original code, but I don't think we need two casts here: `(CHAR8 *)SmbiosRecord + HdrLength` should be sufficient.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
ardbiesheuvel: I disagree. Please update it.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
mikebeaton: I was trying to clarify this in the commit message for this change (but failed). In this version of the code, we start with a pointer at the beginning of the post-'header' memory, and end up with a pointer at the end of the entire SMBIOS struct (variable parts and all) plus appended strings - hopefully - and we've only ever incrementally updated each pointer from the value of the pointer before. So then it's worthwhile to assert that we got to the end correctly. Less so if we restart with a jump relative to the beginning, in a step during this process.

At least, that was my logic for the change. If you buy it, I'm happy to update the commit message. If you don't, I'd say drop the commit (with the new ASSERT) that it's made in.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
mikebeaton: Well, that was meant to be precisely my point. We are now storing the size of the string, not the length, in fact I find it very confusing to read and edit code with string variables called 'Len' that hold size, or called 'Size' that hold length. It makes you expect (and edit in, if you are changing things) precisely the wrong '+1's in the code.

I actually think it's a very good 'standard', for that reason to always ensure the string variables that hold string sizes (+1) are _always_ called 'Size', and ones that hold Length (+0) are always called 'Len'.

I know I made an ASCII vs. CHAR16 mistake earlier, myself. But I say that's a different thing; yes, CHAR16 sizes are NOT (+1) from CHAR16 lengths, but that is an endemic problem of EDK2 code which a coder just has to watch out for (and which I failed to, in an earlier draft :-( ), separate from this issue. It is almost always the case that when we want to hold 'Size', it is precisely because we care about the memory the string occupies, and when we want to hold 'Length' it is because we care about the char count. Certainly often enough that the convention or standard I'm proposing works.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
mikebeaton: Not. It's not sufficient, unfortunately. On CLANG compilers it's required to avoid a warning (which we want to keep enabled, as it also catches other far more 'real' errors).

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
mikebeaton: In fact, what I've picked up while working on EDK2 code would appear to be exactly the convention in the EDK2 codebase!

A grep for 'Len *= *AsciiStrSize' returns 3 examples. A grep for 'Size *= *AsciiStrSize' returns 81.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
ardbiesheuvel: OK

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
mikebeaton: I think the best bet here is to change it back to Len and keep the `+1`s. I'll do that.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
mikebeaton: For now, I'll update the comment to clarify my reasoning.

mikebeaton:
@ardbiesheuvel - I have updated the first ASSERT; abandoned the (probably ill advised, if you look at it correctly - as you did), Len->Size change; and - for now? - updated the commit message where the second ASSERT (plus associated pointer arithmetic change) is introduced, to try to explain clearly enough what I am doing there, and why I (alone?!) think it might be worth it. With thanks for your time so far, on this.

Again, all individual commits build and pass uncrustify locally.

ardbiesheuvel:
Please drop the last patch, the rest looks fine to me now. Thanks.

mikebeaton:
Thank you for approving the other changes.

But if I was choosing the commit I thought added least value, it would have been the second-last - additional ASSERT plus minor pointer arithmetic changes - not the last!

For the last one, SKUNumber definitely is of type SMBIOS_TABLE_STRING. It's only not in the struct with this type because, in this case, the SMBIOS standard defines a weird struct where some struct members come after a variable size area. SMBIOS_TABLE_TYPE1 actually has a SKUNumber field of SMBIOS_TABLE_STRING in the normal part of the struct. And the comments referring to SKUNumber in the SMBIOS_TABLE_TYPE3 definition, as well as the usage of that piece of memory, make it absolutely clear that that field _is_ of type SMBIOS_TABLE_STRING.

My point being, if we had UNIT8 * used to address the SKUNumber field of SMBIOS_TABLE_TYPE1, then I guess you'd say it was a valid change to correct it to SMBIOS_TABLE_STRING *? But, and if so, this is _exactly_ analogous, and the pointer used to reference this SKUNumber field _is_ the wrong type - even though it certainly doesn't functionally change the code to use the wrong type. (But in general, in EDK2 code, we don't consider types, and the types they are defined as, interchangeable, surely?)

I certainly don't insist on using sizeof (*SkuNumberField) further down (not that I'm in a position to insist on anything!), but surely the actual pointer type change is correct?

ardbiesheuvel:
> Thank you for approving the other changes.
> 
> But if I was choosing the commit I thought added least value, it would have been the second-last - additional ASSERT plus minor pointer arithmetic changes - not the last!
> 

I was on the fence about the 3rd patch too, but that one doesn't change as many lines.

mikebeaton:
> I was on the fence about the 3rd patch too, but that one doesn't change as many lines.

Well apart from whitespace, which is unavoidable because of uncrustify, if I undo the non-needed change from `sizeof (SMBIOS_TABLE_STRING)` to `sizeof (*SkuNumberField)`, then the fourth commit - just correcting the pointer type, as it becomes (and it _is_ correcting!) - changes precisely two lines...

mikebeaton:
The fact that there are five strings - precisely the ones handled in this function - and that this is the SMBIOS_TABLE_STRING value for the fifth of them (the other four are in the fixed position part of the struct), also becomes much clearer with this change; rather than having to hunt back into structs and definitions to understand it.

ardbiesheuvel:
Fair enough.

How did you test these changes?

mikebeaton:
So, at the moment, compiling only, plus peer review. (Noting that the previous version _must_ have been broken already.) I have not yet single stepped through the new code, nor checked where this output would be consumed and watched something consuming it. Perhaps the best bet (simpler, sufficient; though I can certainly do the rest, if needed) is to run the ShellPkg dumper command that I linked to somewhere above, and check that the dump is sane. (And maybe check what on earth it said before.) I'll do that, and report back.

mikebeaton:
@ardbiesheuvel - Seems to be working correctly, and I can now see how the previous version was only just working, given the input it actually got. Hampered by the fact that something seems to have broken keyboard input - or at least response to Enter key - in all Arm platform menus. Looking for that atmo.

@leiflindholm - If I may, do you have any more info on the origin of the file under discussion here - as in, I notice you are closely associated with https://github.com/tianocore/edk2/pull/1413 where this commit https://github.com/tianocore/edk2/commit/bfc0fae459543442f3f17e0de655a72aba5b0920 I believe was merged from. I was specifically wondering whether there is any further back history for the (quite a few) new files added in that merge?

leiflindholm:
@mikebeaton hah, I didn't realise we'd tried to reverse-generate a PR history from before migrating to the PR workflow. Sadly, it's misleading: the contribution was by @bexcran:
https://edk2.groups.io/g/devel/message/68396

This implementation, as mentioned in the cover letter, started out from the driver developed by Hisilicon for their server SoCs.

leiflindholm:
> @mikebeaton hah, I didn't realise we'd tried to reverse-generate a PR history from before migrating to the PR workflow. Sadly, it's misleading: the contribution was by @bexcran: https://edk2.groups.io/g/devel/message/68396
> 
> This implementation, as mentioned in the cover letter, started out from the driver developed by Hisilicon for their server SoCs.

Ah, https://edk2.groups.io/g/devel/message/68386 was the actual cover letter.

mikebeaton:
Ard - I discovered your commit with super secret squirrel s key - which is extrememly useful - so I can access the shell even though selecting menu commands in ARM menus is currently broken. This diff is from the results of `smbiosview` run in shell in SbsaQemu:

Screenshot 2025-11-27 at 21 40 27

Note that the 25 before the fix should be, and now is, 22: 0x15 + 1 as we have been discussing.

ArmPkg/Universal/Smbios/SmbiosMiscDxe/Type03/MiscChassisManufacturerFunction.c
mhaeuser: For the current state of affairs, I vaguely agree the message turns out a bit nicer with the condition. However, it’s obvious this is prone to abuse with incomplete updates to the duplicated condition and also ambiguity to what is the right condition (like, is the “leaf” check always useful without the context of what greater branch you’re currently in?).

The obvious issue, as has been discussed plenty of times, is the ASSERT abuse. There is no sane way you end up with both an ASSERT and a runtime check spelled out in sane code - something is a precondition/postcondition/invariant/… or it is not. A higher-level macro *can* runtime-check for defensive hardening based on the project configuration, but not like this.

What ASSERT(FALSE) replicates here is our DEBUG_RAISE(), which basically is an optional breakpoint when a condition is very odd to not hold. The surrounding context is pretty much always relevant (as it is here, I can’t say that “some random header size exceeds UINT8” is particularly helpful on its own).

The reason I open this can of worms again is that in the past, we gave extensive reasons for why using ASSERT in such contexts is bad for self-documentation, error handling, and automated testing or verification. Now, what we never discussed before was debugging experience.

We are very much fine with ASSERT(FALSE)-like behaviour and simply retrieving the line number to the RAISE, which will be sufficiently clear on what the issue is, either self-documented or with a comment. If you prefer on-screen feedback at the time of the fault, the obvious answer to me is allowing to augment ASSERTs and such with a fault message (like Rust’s expect()). This allows to briefly explain the issue, which may provide the context surrounding the failed condition and is easily understood. The kind of thing we currently put as a comment at the fault site. Of course binary size is a concern (hence we didn’t introduce it ourselves), but near-useless prints are also a waste.

Point in case, this was a bug caused by refactoring success to error handling (ASSERT(FALSE), DEBUG_RAISE(), … would have caught this): https://github.com/tianocore/edk2/pull/11687#discussion_r2499928436 -> https://github.com/tianocore/edk2/pull/11687#discussion_r2565388547

PR #11796, opened 2025-11-24
OvmfPkg: Remove duplicated AmdSevDxe driver

Damien-Chen:
This commit remove duplicated AmdSevDxe driver
since it is declared in APRIORI DXE at the begging of [FV.DXEFV]

# Description

This commit remove duplicated AmdSevDxe driver
since it is declared in APRIORI DXE at the begging of [FV.DXEFV]

## How This Was Tested

Build OvmfPkgX64.fdf 

## Integration Instructions

N/A

### DIFF
--- a/OvmfPkg/AmdSev/AmdSevX64.fdf
+++ b/OvmfPkg/AmdSev/AmdSevX64.fdf
@@ -231,7 +231,6 @@ INF  OvmfPkg/QemuVideoDxe/QemuVideoDxe.inf
 INF  OvmfPkg/QemuRamfbDxe/QemuRamfbDxe.inf
 INF  OvmfPkg/VirtioGpuDxe/VirtioGpu.inf
 INF  OvmfPkg/PlatformDxe/Platform.inf
-INF  OvmfPkg/AmdSevDxe/AmdSevDxe.inf
 INF  OvmfPkg/IoMmuDxe/IoMmuDxe.inf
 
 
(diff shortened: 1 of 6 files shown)


kraxel:
Looks like it is listed twice in /all/ *.fdf files.

Damien-Chen:
It appears twice in all .fdf under OvmfPkg. Since this PR is also cover OvmfPkg folder, I think I can help to modify all .fdf

kraxel (approved):

mikebeaton:
@kraxel - It looks like _all_ the APRIORI DXE and APRIORI PEI files are listed twice, everywhere in the entire edk2 repo (mainly under OvmfPkg, plus one .fdf under UefiPayloadPkg, and one under EmulatorPkg).

To the extent that one might say that this wasn't a mistake, but intended. Something like 'these are (all) the drivers', and then 'these are the selection of all drivers that have to be loaded first'.

So are we sure we are correct in changing this? I _believe_ it would compile and work as expected, if they were all changed. But do we want to?

Damien-Chen:
@ardbiesheuvel Could you describe more detail about this? 
It would be better if we could have those information being documented, telling them compiler do not actually build those .INF defined in APRIORI block, just ensure the order of loading and dispatching

Damien-Chen:
@ardbiesheuvel 
Not sure the impact on QEMU, just verify again that the driver declare in APRIORI will be compiled as well. 
So what is the actual impact for QEMU if we made such change ?

ardbiesheuvel:
> @ardbiesheuvel Not sure the impact on QEMU, just verify again that the driver declare in APRIORI will be compiled as well. So what is the actual impact for QEMU if we made such change ?

It is not the FDF that defines what gets compiled. It defines what gets added to the image.

Damien-Chen:
@ardbiesheuvel 
I think I understand what you means, I investigate DXEFV.FV
Those driver defined in APRIORI will not be added to image, it only define ordering.
Like you said it need to defined outside that block as well.

I will close this PR. Sorry for making such dumb error

ardbiesheuvel:
> I will close this PR. Sorry for making such dumb error

No worries.

PR #11798, opened 2025-11-24
[RFC] SecurityPkg: Tpm2DeviceLibFfa: Use partition properties to select request type

mariam-elshakfy:
# Description

Use FF-A partition properties to select between legacy direct
request and direct request2 for TPM communication

TPM implementations based on FF-A 1.1 do not support direct
request2 interface introduced in FF-A 1.2

For example, fTPM SP in trusted-services complies with the TPM
Service Command Response Buffer Interface over FF-A (den0138)
which allows using the legacy direct request.

This change leverages FF-A partition properties to dynamically
decide whether to use the legacy DIRECT_REQ or DIRECT_REQ2 when
communicating with the TPM secure partition.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested on a custom FVP build using fTPM SP from trusted-services

## Integration Instructions
N/A

### DIFF
--- a/SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
+++ b/SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
@@ -27,6 +27,7 @@
 #include "Tpm2DeviceLibFfa.h"
 
 UINT16  mFfaTpm2PartitionId = TPM2_FFA_PARTITION_ID_INVALID;
+static BOOLEAN mFfaTpm2UseDirectReq2 = FALSE;
 
 /**
   Check the return status from the FF-A call and returns EFI_STATUS
@@ -158,6 +159,10 @@ FfaTpm2GetServicePartitionId (
        * Tpm partition id never be TPM2_FFA_PARTITION_ID_INVALID.
        */
       Status = EFI_DEVICE_ERROR;
+    } else {
+      if (TpmPartInfo->PartitionProps & FFA_PART_PROP_RECV_DIRECT_REQ2) {
+        mFfaTpm2UseDirectReq2 = TRUE;
+      }
     }
   }
 
@@ -185,7 +190,7 @@ Tpm2GetInterfaceVersion (
   )
 {
   EFI_STATUS       Status;
-  DIRECT_MSG_ARGS  FfaDirectReq2Args;
+  DIRECT_MSG_ARGS  FfaDirectReqArgs;
 
   if (Version == NULL) {
     Status = EFI_INVALID_PARAMETER;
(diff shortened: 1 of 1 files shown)


spbrogan:
Do you see a scenario where the platform UEFI needs dynamic support (meaning platform firmware doesn't know at build time what type of TPM SP it is communicating with)?  Do you expect a platform to change its secure partition communication protocol independently of the UEFI firmware?  

I would prefer to see the legacy FFA support not added to the same library instance.  I expect that the majority of new systems will not be using legacy and this just adds complexity and burdens the platforms with more code and bloat.  We have many examples in Edk2 where one platforms need for legacy support burdens the entire community for far longer than the viability of the platform. 

A tpm2deviceliblegacyffa could just as easily be added and get nearly the same result (unless dynamic support is desired)?  We have also had some brief discussions about potentially moving FFA communication into a service component and then having the TPM2DeviceLib interact with the abstracted component.  This could handle some of the FFA semantics in a common component (like yield, notification, interrupts, multiple requests) more easily and would provide yet another opportunity for abstraction.

mariam-elshakfy:
> Do you see a scenario where the platform UEFI needs dynamic support (meaning platform firmware doesn't know at build time what type of TPM SP it is communicating with)? Do you expect a platform to change its secure partition communication protocol independently of the UEFI firmware?

> I would prefer to see the legacy FFA support not added to the same library instance. I expect that the majority of new systems will not be using legacy and this just adds complexity and burdens the platforms with more code and bloat. We have many examples in Edk2 where one platforms need for legacy support burdens the entire community for far longer than the viability of the platform.
> 
> A tpm2deviceliblegacyffa could just as easily be added and get nearly the same result (unless dynamic support is desired)? 

Thanks. At least in our case, it's known at build time. So, yes, a new library would work. However, wouldn't this cause a maintenance headache? The new/legacy library will be an exact duplicate of the current library (minus the direct request part). So, what happens with future bugfixes in this case?

Also, we did it in the same library since TPM Service CRB Interface Over FF-A specs (DEN0138) allows the use of both request types. So, it will be up to the TPM service implementation. Quoting from [3.1 Discovery of the TPM service]:

```
Under FF-A v1.2 or later, the message addressing mode required by the TPM service must be determined at the
time of service discovery; it is a TPM service implementation choice whether the service is addressed messages by
only the FF-A partition ID through FFA_MSG_SEND_DIRECT_REQ, or by both the FF-A partition ID and the
service UUID through FFA_MSG_SEND_DIRECT_REQ2.
```

> We have also had some brief discussions about potentially moving FFA communication into a service component and then having the TPM2DeviceLib interact with the abstracted component. This could handle some of the FFA semantics in a common component (like yield, notification, interrupts, multiple requests) more easily and would provide yet another opportunity for abstraction.

I think if we could plug in a library that handles the FF-A request, this could help.

kuqin12:
The FFA library has the basic requirement of FF-A spec 1.2: https://github.com/tianocore/edk2/blob/ad5030d73171d9aae30ee74c4b2cd41686b93bcd/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c#L769

Is this tested and passed FF-A 1.2? If so, direct_req2 should be support, no?

PR #11799, opened 2025-11-24
MdeModulePkg/Pei: Update image section alignment cast

makubacki:
# Description

MSVC 14.44.35207 reports the following warning:

```
MdeModulePkg\Core\Pei\Image\Image.c(405): '~': zero extending
  'UINT32' to 'PHYSICAL_ADDRESS' of greater size.
```

This change casts the section alignment to `EFI_PHYSICAL_ADDRESS` to ensure consistent alignment mask generation.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- CI
- Build with MSVC 14.44.35207

## Integration Instructions

- N/A

### DIFF
--- a/MdeModulePkg/Core/Pei/Image/Image.c
+++ b/MdeModulePkg/Core/Pei/Image/Image.c
@@ -402,7 +402,7 @@ LoadAndRelocatePeCoffImage (
       if (ImageContext.SectionAlignment > EFI_PAGE_SIZE) {
         ImageContext.ImageAddress =
           (ImageContext.ImageAddress + ImageContext.SectionAlignment - 1) &
-          ~((UINTN)ImageContext.SectionAlignment - 1);
+          ~((EFI_PHYSICAL_ADDRESS)ImageContext.SectionAlignment - 1);
       }
 
       //


makubacki:
@lgao4, can you please review this when you get a chance?

MdeModulePkg/Core/Pei/Image/Image.c
```diff
@@ -402,7 +402,7 @@ LoadAndRelocatePeCoffImage (
       if (ImageContext.SectionAlignment > EFI_PAGE_SIZE) {
         ImageContext.ImageAddress =
           (ImageContext.ImageAddress + ImageContext.SectionAlignment - 1) &
-          ~((UINTN)ImageContext.SectionAlignment - 1);
+          ~((EFI_PHYSICAL_ADDRESS)ImageContext.SectionAlignment - 1);
```
mikebeaton, line 405: `ImageContext.ImageAddress` which you need to match size to here is actually `PHYSICAL_ADDRESS`, not `EFI_PHYSICAL_ADDRESS`, so would `(PHYSICAL_ADDRESS)` not be a better cast?

lgao4: In fact, they are the same type. PEI module can use EFI_PHYSICAL_ADDRESS.

mikebeaton: Just literally to match what it says if you navigate to ImageAddress - but also, fine.

I believe this is blocking merging PRs, as all builds are hitting it after (presumably?) a compiler upgrade in the Windows image.

mikebeaton:
This now overlaps with https://github.com/tianocore/edk2/pull/11836 - which fixes one other instance and various other somewhat related issues, so I believe would now be the choice to merge.

makubacki:
> This now overlaps with #11836 - which fixes one other instance and various other somewhat related issues, so I believe would now be the choice to merge.

Yeah, I'll close this one.

Issue #11800, opened 2025-11-25
[Bug]: non-x86 Option Rom never be reached if multiple Option Rom images are found on non-x86 platform.


EricGao2015:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

MdeModulePkg

### Which targets are impacted by this bug?

NOOPT

### Current Behavior

X86 rom image is loaded.
If the option rom in a PCIe card have multiple architecture(x86, aarch64, riscv64, loongarch64) images, the native image will never be reached on non-x86 platforms if the [emulator](https://github.com/andreiw/MultiArchUefiPkg) is introduced. 

### Expected Behavior

The native loongarch64 rom image is loaded.
If the option rom in a PCIe card have multiple architecture(x86, aarch64, riscv64,loongarch64) images, the native image can be loaded on non-x86 platforms if the [emulator](https://github.com/andreiw/MultiArchUefiPkg) is introduced. 

### Steps To Reproduce

1. Use [MultiArchUefiPkg](https://github.com/andreiw/MultiArchUefiPkg) as emulator.
2. Insert a display card on loongarch architecture motherboard. There are x86, aarch64 and loongarch64 three kinds of image in order at the option rom.
3. Determine which architecture rom image is loaded, x86 or loongarch64?

### Build Environment

```markdown
- OS(s): Centos Linux 8
- Tool Chain(s):https://github.com/loongson/build-tools/releases/download/2024.11.01/x86_64-cross-tools-loongarch64-binutils_2.43.1-gcc_14.2.0-glibc_2.40.tar.xz
```

### Version Information

```text
edk2-stable202511
```

### Urgency

Medium

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

No maintainer feedback needed

mikebeaton:
'Which targets are impacted by this bug?' answer should be 'RELEASE, NOOPT, DEBUG', I believe.

EricGao2015:
> 'Which targets are impacted by this bug?' answer should be 'RELEASE, NOOPT, DEBUG', I believe.

Modified.

EricGao2015:
PR https://github.com/tianocore/edk2/pull/11807 has been merged, so close this issue now.

PR #11807, opened 2025-11-26
MdeModulePkg: Load the native Option ROM image if the native one exists

EricGao2015:
# Description

Issue: https://github.com/tianocore/edk2/issues/11800

If there are multiple Option ROM images existed, current code will load the first image that the emulator supports. Usually, x86 ROM image is the first one and non-x86 ROM image is behind x86 ROM image. When the emulator is introduced, x86 ROM image instead of native image is loaded on non-x86 platforms.
This patch introduces a mechanism to prioritize the loading of native images. Firstly, search the native image. If the native one is found, just finish the process since the goal is reached. Otherwise, search the foriegn images that the platform may support.

## How This Was Tested

1. Use [MultiArchUefiPkg](https://github.com/andreiw/MultiArchUefiPkg) as emulator.
2. Insert a display card on loongarch architecture motherboard. There are x86, aarch64 and loongarch64 three kinds of image in order at the option rom.
3.  Boot the motherboard and determine which architecture rom image is loaded, x86 or loongarch64? The result is that 
the loonagrch64 architecture image instead of x86 architecture image is loaded successfully.

## Integration Instructions

N/A


EricGao2015:
@lgao4  @jkolakow  Hi reviewers, please review this patch at your earliest convenience.

MdeModulePkg/Bus/Pci/PciBusDxe/PciOptionRomSupport.c
```diff
@@ -638,6 +638,65 @@ RomDecode (
...
+  Scan the whole Option Rom and check if the native image exists.
+
+  @param PciDevice       Pci device instance.
+
+  @retval TRUE           The native Option Rom image exists.
+  @retval FALSE          The native Option Rom image does not exist.
+
+**/
+BOOLEAN
+FindNativeOpRom (
```
ardbiesheuvel: Instead of duplicating a lot of the logic in `ProcesOpRomImage()`, could you perhaps add a `BOOL NativeOnly` argument to it, and call it with `TRUE` first, and with `FALSE` only if the first call does not return anything?

kilaterlee: Let me think about it.
Do you mean drop this function, move the logice into `rocessorOpRomImage`, and then add an argument called `NativeOnly` to `ProcessorOpRomImage`, then call it with `TRUE` first, if not found return `EFI_NOT_FOUND`, and call it with `FALSE` again? If so, we will add some logic to the caller.

ardbiesheuvel: Yes, something like the following:

```diff
--- a/MdeModulePkg/Bus/Pci/PciBusDxe/PciDeviceSupport.c
+++ b/MdeModulePkg/Bus/Pci/PciBusDxe/PciDeviceSupport.c
@@ -329,7 +329,10 @@ RegisterPciDevice (
     // or loaded from device in the previous round of bus enumeration
     //
     if (HasEfiImage) {
-      ProcessOpRomImage (PciIoDevice);
+      Status = ProcessOpRomImage (PciIoDevice, TRUE);
+      if (Status == EFI_NOT_FOUND) {
+        ProcessOpRomImage (PciIoDevice, FALSE);
+      }
     }
   }
 ```

Comment (kilaterlee) [reply MdeModulePkg/Bus/Pci/PciBusDxe/PciOptionRomSupport.c]: OK, I have already informed @EricGao2015 about adjustment method yesterday, however the new argument order might be different from your version, I believe he will submit the new version soon. :)

File: MdeModulePkg/Bus/Pci/PciBusDxe/PciOptionRomSupport.c [hunk 2]
```diff
@@ -681,6 +685,17 @@ ProcessOpRomImage (
...
+    Status = gBS->LocateProtocol (
+                    &gEdkiiPeCoffImageEmulatorProtocolGuid,
+                    NULL,
+                    (VOID **)&PeCoffEmuProtocol
+                    );
+    if (EFI_ERROR (Status) || (PeCoffEmuProtocol == NULL)) {
+      return RetStatus;
+    }
+  }
+
```
ardbiesheuvel: Can we just drop this check? It changes the current behavior unnecessarily for platforms where the only available option ROM is non-native.

EricGao2015: Hi Ard, thanks for your suggestions. I have updated the patch, please review it again.

PR #11808, opened 2025-11-26
ShellPkg/AcpiView: Fix unused variable warnings in RimtParser

ardbiesheuvel:
Clang complains about unused variables:

ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c:21:60: error: variable 'mRimtNodeHeader' is not needed and will not be emitted [-Werror,-Wunneeded-internal-declaration]
   21 | STATIC EFI_ACPI_6_6_RIMT_NODE_HEADER_STRUCTURE             mRimtNodeHeader;
      |                                                            ^~~~~~~~~~~~~~~
ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c:25:60: error: variable 'mRimtIdMappingNode' is not needed and will not be emitted [-Werror,-Wunneeded-internal-declaration]
   25 | STATIC EFI_ACPI_6_6_RIMT_ID_MAPPING_STRUCTURE              mRimtIdMappingNode;
      |                                                            ^~~~~~~~~~~~~~~~~~
   28 | STATIC EFI_ACPI_6_6_RIMT_PCIE_ROOT_COMPLEX_NODE_STRUCTURE  mRimtPcieRootComplexNode;
      |                                                            ^~~~~~~~~~~~~~~~~~~~~~~~
ShellPkg/Library/UefiShellAcpiViewCommandLib/Parsers/Rimt/RimtParser.c:29:60: error: variable 'mRimtIommuNode' is not needed and will not be emitted [-Werror,-Wunneeded-internal-declaration]
   29 | STATIC EFI_ACPI_6_6_RIMT_IOMMU_NODE_STRUCTURE              mRimtIommuNode;
      |                                                            ^~~~~~~~~~~~~~
4 errors generated.

This is because these variables are only used to take the size of their fields using the sizeof() operator. ~So declare them as 'extern' instead, as no definition is actually needed for this use case.~

### DIFF
--- a/BaseTools/Source/Python/build/build.py
+++ b/BaseTools/Source/Python/build/build.py
@@ -24,6 +24,7 @@ import time
 import platform
 import traceback
 import multiprocessing
+import errno
 from threading import Thread,Event,BoundedSemaphore
 import threading
 from linecache import getlines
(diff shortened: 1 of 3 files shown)


ardbiesheuvel:
cc @krzysztof-drobinski-plastelinet

pierregondois:
Hi Ard,
I think `mRimtInfo` should not be changed to extern.
It seems this is what trigger the CI fails

ardbiesheuvel:
> Hi Ard, I think `mRimtInfo` should not be changed to extern. It seems this is what trigger the CI fails

~Indeed - VS2022 doesn't seem to like it.~ Hmm, I see what you mean now - I don't have access to the failed CI logs anymore, so perhaps I could have fixed this more easily.

Instead, I've replaced the `sizeof()` with a macro that uses the type directly.

pierregondois (approved):

PR #11810, opened 2025-11-26
BaseTools: AutoGen: Optimize tuple creation using tuple() for PcdDbBu…

jeremy-compostella:
…ffer

Replace manual loop-based tuple construction with the built-in tuple() function when converting PcdDbBuffer to a tuple. This change significantly improves performance—approximately three times faster—resulting in substantial build time savings in large environments.

Previously, the code iterated over each byte in PcdDbBuffer, unpacking and appending it to a tuple. The new approach leverages tuple(PcdDbBuffer) to achieve the same result more efficiently. The generated tuple remains identical to the original implementation.

TEST=The generated tuple is the same than with to original code

# Description

Replace manual loop-based tuple construction with the built-in tuple() function when converting PcdDbBuffer to a tuple. This change significantly improves performance—approximately three times faster—resulting in substantial build time savings in large environments.

Previously, the code iterated over each byte in PcdDbBuffer, unpacking and appending it to a tuple. The new approach leverages tuple(PcdDbBuffer) to achieve the same result more efficiently. The generated tuple remains identical to the original implementation.

## How This Was Tested

The generated tuple is the same than with to original code

## Integration Instructions

None


mikebeaton (approved):

mikebeaton:
Tested. Works (e.g. builds, runs Ovmf). Values are indeed identical. Lgtm.

gapalomi (approved):

PR #11813, opened 2025-11-27
BaseTools: Build: Balance thread concurrency with file descriptor limits

AydenMeng:
# Description

Problem

When the system's file descriptor limit is low (e.g., ulimit -n 1024) and the build is configured to use a high number of threads (e.g., 512), the build process can exhaust available file descriptors. This occurs because each multiprocessing worker typically consumes multiple file descriptors (for pipes, queues, semaphores, etc.).

As a result:

 - Some threads fail to create necessary IPC resources.
 - The build hangs or stalls indefinitely.
 - No clear error message is shown to the user, making debugging extremely difficult.

This issue can be confirmed via strace, which reveals failures like:
```
pipe2(0x7fff56711148, O_CLOEXEC) = -1 EMFILE (Too many open files)
```

Solution

This PR introduces two complementary fixes:

 - Commit 31d5b3:

    Explicitly catches OSError exceptions caused by file descriptor exhaustion during worker initialization or execution, logs a clear diagnostic message, and terminates the build gracefully—instead of hanging silently.

 - Commit 456393:

    Dynamically computes a safe upper bound for the number of concurrent build threads based on the current file descriptor limit. The calculation assumes ~4 FDs per worker (3 for multiprocessing pipes + 1 for POSIX named semaphores) and caps the thread count accordingly:

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

To reproduce the issue, run the following commands before starting the build (replace [THREAD_COUNT] with the actual number of threads used by your build, e.g., 16):
```
ulimit -Sn [THREAD_COUNT]      # e.g., ulimit -Sn 16
ulimit -Sn [THREAD_COUNT * 2]  # e.g., ulimit -Sn 32
ulimit -Sn [THREAD_COUNT * 4]  # e.g., ulimit -Sn 64
```

With the first two settings ([THREAD_COUNT] and [THREAD_COUNT * 2]), the build will likely fail or hang indefinitely due to file descriptor exhaustion (e.g., EMFILE: Too many open files), often without a clear error message.

The third setting ([THREAD_COUNT * 4]) usually provides enough file descriptors for the build to succeed.

After Applying the Patch

Under low FD limits (e.g., ulimit -Sn 16), the build will either:
 - Fail fast with a clear error
 -  Automatically reduce concurrency and complete successfully.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Source/Python/build/build.py
+++ b/BaseTools/Source/Python/build/build.py
@@ -838,13 +838,22 @@ class Build():
         try:
             if SkipAutoGen:
                 return True,0
+            if sys.platform == "win32":
+                SafeThreadNumber = self.ThreadNumber
+            else:
+                import resource
+                soft = resource.getrlimit(resource.RLIMIT_NOFILE)[0]
+                SafeThreadNumber = min(self.ThreadNumber, soft // 3)
+                if SafeThreadNumber < self.ThreadNumber:
+                    EdkLogger.verbose("AutoGen workers limited to %d to avoid file descriptor exhaustion" % SafeThreadNumber)
             feedback_q = mp.Queue()
             error_event = mp.Event()
             FfsCmd = DataPipe.Get("FfsCommand")
             if FfsCmd is None:
                 FfsCmd = {}
             GlobalData.FfsCmd = FfsCmd
-            auto_workers = [AutoGenWorkerInProcess(mqueue,DataPipe.dump_file,feedback_q,GlobalData.file_lock,cqueue,self.log_q,error_event) for _ in range(self.ThreadNumber)]
+
+            auto_workers = [AutoGenWorkerInProcess(mqueue,DataPipe.dump_file,feedback_q,GlobalData.file_lock,cqueue,self.log_q,error_event) for _ in range(SafeThreadNumber)]
             self.AutoGenMgr = AutoGenManager(auto_workers,feedback_q,error_event)
             self.AutoGenMgr.start()
             for w in auto_workers:


BaseTools/Source/Python/build/build.py
```diff
@@ -838,13 +839,15 @@ def StartAutoGen(self,mqueue, DataPipe,SkipAutoGen,PcdMaList,cqueue):
...
+            soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
             feedback_q = mp.Queue()
             error_event = mp.Event()
             FfsCmd = DataPipe.Get("FfsCommand")
             if FfsCmd is None:
                 FfsCmd = {}
             GlobalData.FfsCmd = FfsCmd
-            auto_workers = [AutoGenWorkerInProcess(mqueue,DataPipe.dump_file,feedback_q,GlobalData.file_lock,cqueue,self.log_q,error_event) for _ in range(self.ThreadNumber)]
+
+            auto_workers = [AutoGenWorkerInProcess(mqueue,DataPipe.dump_file,feedback_q,GlobalData.file_lock,cqueue,self.log_q,error_event) for _ in range(min(self.ThreadNumber, soft // 3))]
```
mikebeaton: Why 3 when you mention 4 in the commit message?

It would be helpful to understand if the divisor value was arrived at mainly/partly experimentally (e.g. in practice this value works, this value still exhausts file descriptors)?

AydenMeng: Oh yes, thank you for pointing that out—it was indeed a typo in the code.

To investigate file descriptor (FD) usage during the build process, I used the following command:

```
strace -o log -e trace=file,desc build -b RELEASE -t GCC5 -a LOONGARCH64 -p OvmfPkg/LoongArchVirt/LoongArchVirtQemu.dsc -n [ThreadCount=200/300/400]
oc=$(cat log | grep -P "open" | wc -l)
pc=$(cat log | grep -P "pipe" | sort | uniq | wc -l)
python3 -c "print(($oc + $pc * 2) // [ThreadCount])"
```

The result consistently came out to approximately **4**, which aligns well with our theoretical expectation: 3 FDs for multiprocessing pipes + 1 FD for the POSIX named semaphore per thread.

However, when I monitor the actual real-time FD consumption using the following script:

```
#!/bin/bash
while true; do
    pid=$(ps -aux | grep $USER | grep build.py | awk '{print $2}')
    ls /proc/$pid/fd | wc -l
done
```
I observe that the peak live FD count is roughly a little **over twice** the number of threads. I believe this discrepancy arises because many file descriptors are allocated and released very quickly, making them invisible in aggregate static analysis.

Nevertheless, **I think using 4 as the divisor remains the safest choice**—especially on lower-performance machines where resource pressure could more easily trigger FD exhaustion.

mikebeaton: I would be tempted to say that 3 is the better choice, backed up by exactly that evidence added in the commit message.

BaseTools/Source/Python/build/build.py
```diff
@@ -837,13 +838,20 @@ def StartAutoGen(self,mqueue, DataPipe,SkipAutoGen,PcdMaList,cqueue):
         try:
             if SkipAutoGen:
                 return True,0
+            if sys.platform == "win32":
+                SafeThreadNumber = self.ThreadNumber
+            else:
+                import resource
+                soft = resource.getrlimit(resource.RLIMIT_NOFILE)[0]
+                SafeThreadNumber = min(self.ThreadNumber, soft // 4)
```
mikebeaton: As well as my recommendation to stick to 3 (and add note on empirical evidence to commit message), it would be good to include an EdkLogger message, actually I would suggest at .info (always show) level, since I believe this only occurs once per package build.

If a user has specified a high thread count it would be good for this to be clearly surfaced in the build log.

Something like "Thread count limited to %d to avoid file descriptor exhaustion", ofc only to show when actually limited to less than the specified (or default, but that won't happen) value.

mikebeaton: Apologies, that should be "Workers limited to...", as I see the specified thread count value is used in other locations not affected by this change.

BaseTools/Source/Python/build/build.py
```diff
@@ -837,13 +838,22 @@ def StartAutoGen(self,mqueue, DataPipe,SkipAutoGen,PcdMaList,cqueue):
         try:
             if SkipAutoGen:
                 return True,0
+            if sys.platform == "win32":
+                SafeThreadNumber = self.ThreadNumber
+            else:
+                import resource
+                soft = resource.getrlimit(resource.RLIMIT_NOFILE)[0]
+                SafeThreadNumber = min(self.ThreadNumber, soft // 3)
```
mikebeaton, line 846: Apologies for multiple comments, since you're clearly responding fast. Since thread number _is_ used in multiple places, should it in fact be limited for all of these? AFAICT, this would involve modifying the ThreadNum() method.

ThreadNum() is called twice, and its result self.ThreadNumber is used in three places, of which this is one. (I wonder if there is way to avoid the two separate calls to ThreadNum(), which are once for self.ThreadNumber and once for LogQMaxSize?)

AFAICT the 'safer' option would be to modify the thread number everywhere, for consistency - as it could not have differed between the above places before this change - but I am not certain of the right choice here, if yourself of anyone else would like to offer an opinion?

mikebeaton: I should note that, if modifying it in all places, this allows the possibility of adding a new `-n max` syntax. Which seems kind of useful to have, given your change.

If adding that, it should be a separate commit, after the modified commit that limits the thread number in just one place so as to affect all uses of it.

AydenMeng: Thank you for your patience and valuable feedback.

Regarding whether ThreadNum() should be modified, I believe it is unnecessary. The issue we encountered stems solely from file descriptor exhaustion during dependency parsing, and the proposed fix appropriately limits the thread count based on available file descriptors.

In contrast, other usages of ThreadNum()—such as during actual compilation or log queue creation—do not significantly contribute to file descriptor consumption. Therefore, adjusting ThreadNum() globally would be unwarranted, as it could unnecessarily restrict parallelism in stages that are not FD-bound.

Hence, I recommend keeping ThreadNum() unchanged.

As a side note, while working on this patch, I looked into potential improvements and found that the ideal long-term solution would be to refactor the entire concurrency logic to use a **thread pool**. This approach would likely be safer, faster, and more resource-efficient. However, implementing such a change is currently beyond the scope of my expertise. I hope that either I or someone else in the community can tackle this enhancement in the future.

mikebeaton: > other usages of ThreadNum()—such as during actual compilation or log queue creation—do not significantly contribute to file descriptor consumption. Therefore, adjusting ThreadNum() globally would be unwarranted, as it could unnecessarily restrict parallelism in stages that are not FD-bound

Thank you for your responses.

In that case, please can you add the above quoted comment to your commit message. Can you also change the newly added log message to say "AutoGen workers limited to...", as the effect is more specific than I had realised.

mikebeaton:
I think two final ones from me: can you change the second commit subject line to 'BaseTools: Cap AutoGen thread count to avoid file descriptor exhaustion'; and can you change the newly added `.info` message to `.verbose`, after all - now that I've understood that it's not affecting the 'main' purpose of -n, I don't think we need to surface it at .info level.

BaseTools/Source/Python/build/build.py
mikebeaton: FYI resolved - but I can't mark these as resolved, I think because it's still in draft.

BaseTools/Source/Python/build/build.py
mikebeaton: Also resolved.

mikebeaton: @ardbiesheuvel - qq if I may - not sure where else to ask it but here - is it because I'm a (new) reviewer without write access, not a maintainer, that I can't Resolve comment threads which I raised? (In some views I don't see a Resolve button at all when I'd normally expect too; in others I do, but if I click it, I get a red 'Failed to resolve thread' popup.) i.e. Is that all normal and expected?

BaseTools/Source/Python/build/build.py
mikebeaton: Resolved.

BaseTools/Source/Python/build/build.py
ardbiesheuvel: Conversations sometimes just disappear into the void, and we cannot even resolve them to unblock the CI/mergify. So things rarely work how **I**'d expect, but I couldn't tell you whether or not this is how GitHub intended it to.

ardbiesheuvel (approved):

AydenMeng:
@lgao4 @bexcran Hi, would you mind help review this patch when you are available. Thanks.

lgao4:
I agree this change.

PR #11816, opened 2025-11-28
BaseTools/Plugin/HostBasedUnitTestRunner: Merge clang coverage files

cosmolai:
# Description

For clang build it doesn't generate coverage file in Build folder. This change is to generate coverage file for all modules.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Use stuart commands to run unit test and generate coverage file.
example:
cd workspace
```
python3 -m venv venv313
. ./venv313/bin/activate
pip install --upgrade -r ./Edk2/pip-requirements.txt
python3 ./Edk2/BaseTools/Edk2ToolsBuild.py -t GCC5
stuart_setup -c ./X/.pytool/CISettings.py -a X64 TOOL_CHAIN_TAG=CLANGDWARF
stuart_update -c ./X/.pytool/CISettings.py -a X64 TOOL_CHAIN_TAG=CLANGDWARF
stuart_ci_build -c ./X/.pytool/CISettings.py -a X64 -t NOOPT TOOL_CHAIN_TAG=CLANGDWARF
deactivate
```
check if Build/coverage.xml is generated.

## Integration Instructions

NA

### DIFF
--- a/BaseTools/Source/C/BrotliCompress/BrotliCompress.c
+++ b/BaseTools/Source/C/BrotliCompress/BrotliCompress.c
@@ -407,8 +407,6 @@ int main(int argc, char** argv) {
   FILE *OutputHandle;
   int Quality;
   int Gap;
-  int OutputFileLength;
-  int InputFileLength;
   int Ret;
   size_t InputFileSize;
   uint8_t *Buffer;
@@ -440,87 +438,71 @@ int main(int argc, char** argv) {
     Version();
     return 0;
   }
-  while (argc > 1) {
-    if (strcmp(argv[1], "-e") == 0 || strcmp(argv[1], "--compress") == 0 ) {
+  int i = 1;
+  while (i < argc) {
+    const char* arg = argv[i];
+    if (strcmp(arg, "-e") == 0 || strcmp(arg, "--compress") == 0) {
       CompressBool = BROTLI_TRUE;
       if (DecompressBool) {
-        printf("Can't use -e/--compress with -d/--decompess on the same time\n");
+        fprintf(stderr, "Can't use -e/--compress with -d/--decompress at the same time\n");
         return 1;
       }
-      argc--;
-      argv++;
+      i++;
       continue;
     }
-    if (strcmp(argv[1], "-d") == 0 || strcmp(argv[1], "--decompress") == 0 ) {
+    if (strcmp(arg, "-d") == 0 || strcmp(arg, "--decompress") == 0) {
       DecompressBool = BROTLI_TRUE;
       if (CompressBool) {
-        printf("Can't use -e/--compress with -d/--decompess on the same time\n");
+        fprintf(stderr, "Can't use -e/--compress with -d/--decompress at the same time\n");
         return 1;
       }
-      argc--;
-      argv++;
+      i++;
       continue;
     }
-    if (strcmp(argv[1], "-o") == 0 || strncmp(argv[1], "--output", 8) == 0) {
-      if (strcmp(argv[1], "-o") == 0) {
-        OutputFileLength = strlen(argv[2]);
-        if (OutputFileLength > _MAX_PATH) {
-          printf ("The file path %s is too long\n", argv[2]);
-          return 1;
-        }
-        OutputFile = argv[2];
-        if (OutputFile == NULL) {
-          fprintf(stderr, "Input file can't be null\n");
-          return 1;
-        }
-        argc--;
-        argv++;
-      } else {
-        OutputFileLength = strlen(argv[1] - 9);
-        OutputFile = (char *)argv[1] + 9;
-      }
-      argc--;
-      argv++;
-      continue;
-    }
-    if (strcmp(argv[1], "-q") == 0 || strncmp(argv[1], "--quality", 9) == 0) {
-      if (strcmp(argv[1], "-q") == 0) {
-        Quality = strtol(argv[2], NULL, 16);
-        argc--;
-        argv++;
-      } else {
-        Quality = strtol((char *)argv[1] + 10, NULL, 16);
-      }
-      argc--;
-      argv++;
-      continue;
-    }
-    if (strcmp(argv[1], "-g") == 0 || strncmp(argv[1], "--gap", 5) == 0) {
-      if (strcmp(argv[1], "-g") == 0) {
-        Gap = strtol(argv[2], NULL, 16);
-        argc--;
-        argv++;
-      } else {
-        Gap = strtol((char *)argv[1] + 6, NULL, 16);
-      }
-      argc--;
-      argv++;
-      continue;
-    }
-    if (argc > 1) {
-      InputFileLength = strlen(argv[1]);
-      if (InputFileLength > _MAX_PATH - 1) {
-        printf ("The file path %s is too long\n", argv[2]);
+    if (strcmp(arg, "-o") == 0) {
+      if (i + 1 >= argc) {
+        fprintf(stderr, "-o requires an argument\n");
         return 1;
       }
-      InputFile = argv[1];
-      if (InputFile == NULL) {
-       printf("Input file can't be null\n");
-       return 1;
-      }
-      argc--;
-      argv++;
+      OutputFile = argv[i + 1];
+      i += 2;
+      continue;
     }
+    if (strncmp(arg, "--output=", 9) == 0) {
+      OutputFile = (char*)arg + 9;
+      i++;
+      continue;
+    }
+    if (strcmp(arg, "-q") == 0) {
+      if (i + 1 >= argc) {
+        fprintf(stderr, "-q requires an argument\n");
+        return 1;
+      }
+      Quality = strtol(argv[i + 1], NULL, 0);
+      i += 2;
+      continue;
+    }
+    if (strncmp(arg, "--quality=", 10) == 0) {
+      Quality = strtol(arg + 10, NULL, 0);
+      i++;
+      continue;
+    }
+    if (strcmp(arg, "-g") == 0) {
+      if (i + 1 >= argc) {
+        fprintf(stderr, "-g requires an argument\n");
+        return 1;
+      }
+      Gap = strtol(argv[i + 1], NULL, 0);
+      i += 2;
+      continue;
+    }
+    if (strncmp(arg, "--gap=", 6) == 0) {
+      Gap = strtol(arg + 6, NULL, 0);
+      i++;
+      continue;
+    }
+    InputFile = argv[i];
+    i++;
   }
 
   Buffer = (uint8_t*)malloc(kFileBufferSize * 2);
@@ -543,14 +525,12 @@ int main(int argc, char** argv) {
     //
     // Decompress file for get Outputfile size
     //
-    strcpy (OutputTmpFile, OutputFile);
-    if (strlen(InputFile) + strlen(".tmp") < _MAX_PATH) {
-      strcat(OutputTmpFile, ".tmp");
-    } else {
+    if (strlen(OutputFile) + strlen(".tmp") >= _MAX_PATH) {
       printf ("Output file path is too long[%s]\n", OutputFile);
       Ret = BROTLI_FALSE;
       goto Finish;
     }
+    snprintf(OutputTmpFile, _MAX_PATH, "%s.tmp", OutputFile);
     memset(Buffer, 0, kFileBufferSize*2);
     Ret = DecompressFile(OutputFile, InputBuffer, OutputTmpFile, OutputBuffer, Quality, Gap);
     if (!Ret) {
--- a/BaseTools/Source/C/Common/EfiUtilityMsgs.c
+++ b/BaseTools/Source/C/Common/EfiUtilityMsgs.c
@@ -403,12 +403,12 @@ PrintMessage (
                        );
     }
     if (Cptr != NULL) {
-      strcpy (Line, ": ");
-      strncat (Line, Cptr, MAX_LINE_LEN - strlen (Line) - 1);
       if (LineNumber != 0) {
-        sprintf (Line2, "(%u)", (unsigned) LineNumber);
-        strncat (Line, Line2, MAX_LINE_LEN - strlen (Line) - 1);
+        snprintf(Line, MAX_LINE_LEN, ": %s(%u)", Cptr, (unsigned) LineNumber);
+      } else {
+        snprintf(Line, MAX_LINE_LEN, ": %s", Cptr);
       }
+
     }
   } else {
     //
@@ -421,7 +421,8 @@ PrintMessage (
       strncpy (Line, Cptr, MAX_LINE_LEN - 1);
       Line[MAX_LINE_LEN - 1] = 0;
       if (LineNumber != 0) {
-        sprintf (Line2, "(%u)", (unsigned) LineNumber);
+        snprintf (Line2, MAX_LINE_LEN, "(%u)", (unsigned) LineNumber);
+
         strncat (Line, Line2, MAX_LINE_LEN - strlen (Line) - 1);
       }
     } else {
@@ -448,7 +449,8 @@ PrintMessage (
   strncat (Line, ": ", MAX_LINE_LEN - strlen (Line) - 1);
   strncat (Line, Type, MAX_LINE_LEN - strlen (Line) - 1);
   if (MessageCode != 0) {
-    sprintf (Line2, " %04u", (unsigned) MessageCode);
+    snprintf (Line2, MAX_LINE_LEN, " %04u", (unsigned) MessageCode);
+
     strncat (Line, Line2, MAX_LINE_LEN - strlen (Line) - 1);
   }
   fprintf (stdout, "%s", Line);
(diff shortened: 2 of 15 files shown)


BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -272,13 +272,25 @@ def gen_code_coverage_clang(self, thebuilder):
...
-        # Merge all XML files
+        # Merge package coverage XML files
         ret = self.merge_cobertura_xml_files([f"{testFile}.coverage.xml" for testFile in testList], mergedCoverageXml)
+        if ret != 0:
+            logging.error("UnitTest Coverage: Failed to merge package coverage XML files.")
+            return 1
+
+        workspace = thebuilder.env.GetValue("WORKSPACE")
+        allCoverageXml = os.path.join(workspace, 'Build', 'coverage.xml')
+        testCoverageList = glob.glob(os.path.join(workspace, 'Build', "**", "*.coverage.xml"), recursive=True)
```
mdkinney: The gcc coverage above removes Build/coverage.xml if it exists before generated the merged coverage file.  This logic will overwrite the file if it already exists.  Probably safe to follow the gcc approach in case there is an error in merge_cobertura_xml_files() and the previous merged coverage file is still present.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -272,13 +272,25 @@ def gen_code_coverage_clang(self, thebuilder):
...
+
+        # Merge all coverage XML files
+        ret = self.merge_cobertura_xml_files(testCoverageList, allCoverageXml)
```
mdkinney: Using this function adds a dependency on OpenCppCoverage which is only available in Windows.  This will not work for Linux/clang environments.  Recommend using a different approach to merge.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
cosmolai: updated.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
cosmolai: use llvm-cov to generate lcov for all executables and convert lcov to xml.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -252,30 +256,83 @@ def gen_code_coverage_clang(self, thebuilder):
...
-                logging.error("UnitTest Coverage: Failed to generate coverage data for " + testFile)
+            if self.clang_gen_lcov_xml(mergedProfData, [testFile], f"{testFile}.coverage.lcov", f"{testFile}.coverage.xml") != 0:
                 return 1
-            coveragexmlFile = f"{testFile}.coverage.xml"
-            ret = RunCmd("lcov_cobertura",f"{lcovFile} -o {coveragexmlFile}")
-            if ret != 0:
-                logging.error("UnitTest Coverage: Failed generate filtered coverage XML.")
-                return 1
-        # Merge all XML files
-        ret = self.merge_cobertura_xml_files([f"{testFile}.coverage.xml" for testFile in testList], mergedCoverageXml)
```
mdkinney, line 276: With this line removal, then function merge_cobertura_xml_files() below can also be removed.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -236,13 +236,17 @@ def gen_code_coverage_clang(self, thebuilder):
         logging.info("Generating UnitTest code coverage")
 
         buildOutputBase = thebuilder.env.GetValue("BUILD_OUTPUT_BASE")
+        workspace = thebuilder.env.GetValue("WORKSPACE")
         if GetHostInfo().os.upper() == "LINUX":
             # Collect test executables with no file extension
             testList = glob.glob(os.path.join(buildOutputBase, "**", "*Test*"), recursive=True)
             testList = [f for f in testList if os.path.isfile(f) and os.path.splitext(f)[1] == ""]
+            allTestList = glob.glob(os.path.join(workspace, "**", "*Test*"), recursive=True)
+            allTestList = [f for f in allTestList if os.path.isfile(f) and os.path.splitext(f)[1] == ""]
```
mdkinney, line 246: Searching entire workspace for unit test executables is different behavior than gcc or msvc coverage collection and may pick up test executables from unrelated builds and seems out of scope for the stuart command invocation.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -252,30 +256,83 @@ def gen_code_coverage_clang(self, thebuilder):
...
+            return 1
+        if not output_lcov:
+            logging.error("clang_gen_lcov_xml: No output lcov file provided.")
+            return 1
+        if not output_xml:
+            logging.error("clang_gen_lcov_xml: No output xml file provided.")
+            return 1
+
+        testfiles_str = ' -object '.join(testfiles)
+        ret = RunCmd("llvm-cov", f"export -format=lcov --instr-profile={profdata} {testfiles_str} > {output_lcov}")
```
mdkinney: This can produce a very long command line that may break in some OSes.  Need to have way to do partial merges to limit command line length and combine partial merges into full merge

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
cosmolai: Update to search only workspace/Build.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
cosmolai: remove merge_cobertura_xml_files.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
cosmolai: removed

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
cosmolai: Find a way for this. llvm-cov command supports response file. Use response file for file list input. This works for both Windows and Linux.

mdkinney:
This change looks good to me.  And it removes merge_cobertura_xml_files() that I believe did not work as expected.

How was this change tested?  We want to have high confidence that that the xml and lcov coverage files are correct.  Can you compare the CLANG results to the GCC requests for code coverage of unit tests in the edk2 repo?

cosmolai:
> This change looks good to me. And it removes merge_cobertura_xml_files() that I believe did not work as expected.
> 
> How was this change tested? We want to have high confidence that that the xml and lcov coverage files are correct. Can you compare the CLANG results to the GCC requests for code coverage of unit tests in the edk2 repo?

it's not possible to compare the xml result between clang and gcc, due to too much different in the ordering in the xml files and the way they count the coverage. 
For example, gcc doesn't count the } line but clang counts the } line by default. This result in the line coverage rate in clang might be lower than it in gcc. 
However I do review couple modules and the result looks good in both.

cosmolai:
@gapalomi @makubacki
Please help to review this.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -251,117 +256,91 @@ def gen_code_coverage_clang(self, thebuilder):
...
+            logging.error("clang_gen_profdata: No profraw list file provided.")
+            return 1
+        if not output_profdata:
+            logging.error("clang_gen_profdata: No output profdata file provided.")
+            return 1
 
-        Args:
-            xml_file_list (list): List of Cobertura XML file paths to merge.
-            output_file (str): Path to the merged output XML file.
+        with open(output_profrawlistfile, "w") as f:
```
makubacki, line 304: Should this have a check for no `.profraw` files being present and give an explicit error message and return `1`?

Like:

```python
    profraw_files = glob.glob(os.path.join(basepath, "**", "*.profraw"), recursive=True)
    if not profraw_files:
       ...
```

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -251,117 +256,91 @@ def gen_code_coverage_clang(self, thebuilder):
...
-            base_root.set("branches-valid", str(branches_valid))
-            base_root.set("branch-rate", str(safe_rate(branches_covered, branches_valid)))
-
-            # Write merged XML
-            base_tree.write(output_file, encoding="utf-8", xml_declaration=True)
-            logging.info(f"Merged Cobertura XML written to {output_file}")
-            return 0
-        except Exception as e:
-            logging.error(f"Failed to merge Cobertura XML files: {e}")
+        resp_file = output_lcov + ".rsp"
```
makubacki, line 333: Can `resp_file` be deleted after it is used in this function?

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -251,117 +256,91 @@ def gen_code_coverage_clang(self, thebuilder):
...
+        with open(resp_file, "w") as f:
+            f.write('\n-object\n'.join(testfiles))
```
makubacki, line 335: Should this be:

```python
    for testfile in testfiles:
        f.write(f"-object\n{testfile}\n")
```

To include `-object` before the first file.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
makubacki: It seems like that would more cleanly handle that case.

cosmolai: That would force user to always does clean and build.
We can do delete first if it's present.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
cosmolai: It's a temp file but it's good to be there for debug purpose.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
cosmolai: The interesting part is that for llvm-cov it cannot have -object for first file. That is the reason it adds -object from 2nd file.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
cosmolai: updated.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
makubacki: That's strange. Then we can resolve this.

cosmolai:
@spbrogan @javagedes
Please help to review and push this. Thanks.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -236,13 +236,18 @@ def gen_code_coverage_clang(self, thebuilder):
         logging.info("Generating UnitTest code coverage")
 
         buildOutputBase = thebuilder.env.GetValue("BUILD_OUTPUT_BASE")
+        workspace = thebuilder.env.GetValue("WORKSPACE")
+
         if GetHostInfo().os.upper() == "LINUX":
             # Collect test executables with no file extension
             testList = glob.glob(os.path.join(buildOutputBase, "**", "*Test*"), recursive=True)
             testList = [f for f in testList if os.path.isfile(f) and os.path.splitext(f)[1] == ""]
+            allTestList = glob.glob(os.path.join(workspace, "Build", "**", "*Test*"), recursive=True)
```
Javagedes, line 245: nit: please only do this globbing in the `gen_code_coverage_clang` function. This takes time, and it is only optionally used it if `CODE_COVERAGE` environment variable is set by being passed to gen_code_coverage.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -236,13 +236,18 @@ def gen_code_coverage_clang(self, thebuilder):
...
         if GetHostInfo().os.upper() == "LINUX":
             # Collect test executables with no file extension
             testList = glob.glob(os.path.join(buildOutputBase, "**", "*Test*"), recursive=True)
             testList = [f for f in testList if os.path.isfile(f) and os.path.splitext(f)[1] == ""]
+            allTestList = glob.glob(os.path.join(workspace, "Build", "**", "*Test*"), recursive=True)
+            allTestList = [f for f in allTestList if os.path.isfile(f) and os.path.splitext(f)[1] == ""]
         elif GetHostInfo().os.upper() == "WINDOWS":
             # Collect test executables with a .exe file extension
             testList = glob.glob(os.path.join(buildOutputBase, "**","*Test*.exe"), recursive=True)
+            allTestList = glob.glob(os.path.join(workspace, "Build", "**","*Test*.exe"), recursive=True)
```
Javagedes, line 250: same here.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -251,117 +256,99 @@ def gen_code_coverage_clang(self, thebuilder):
...
+        for file in [output_profrawlistfile, output_profdata]:
+            if os.path.isfile(file):
+                os.remove(file)
 
-        Args:
-            xml_file_list (list): List of Cobertura XML file paths to merge.
-            output_file (str): Path to the merged output XML file.
+        with open(output_profrawlistfile, "w") as f:
+            f.write("\n".join(glob.glob(os.path.join(basepath, "**", "*.profraw"), recursive=True)))
+        ret = RunCmd("llvm-profdata", f"merge -sparse --input-files {output_profrawlistfile} --output {output_profdata}")
```
Javagedes, line 306: nit: I know this is existing code here that you relocated, but it would be better if we checked the output (or return value) to see if the command failed or if the command did not exist at all, and log it. Just the give the user a little bit more help.

-----

\<update\>:

Unless (and this is probably true now that I think about it), by building with CLANG, these other tools like llvm-profdata and llvm-cov are guarunteed to exist? If so, then this can be closed :)

cosmolai: It's part of clang tool-chain.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
cosmolai: This is already within function gen_code_coverage_clang and only launched when CODE_COVERAGE != FALSE.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
cosmolai: This is already within function gen_code_coverage_clang and only launched when CODE_COVERAGE != FALSE.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
Javagedes: Ah apologies. I see that now. I only looked at the consumer function, not the function this resided in. Thanks for the clarification.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
Javagedes: Thanks for validating. Closing.

PR #11817, opened 2025-11-28
ShellPkg: UefiShellLevel2CommandsLib: Extract a MainCmdXXX() function and refactor code

pierregondois:
# Description

This patch aims to help breaking down the long function present in the ShellPkg and reduce complexity/nested code and conditions. No functional change should be induced by this patches.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Briefly ran the shell commands.

### DIFF
--- a/ShellPkg/Library/UefiShellLevel2CommandsLib/Rm.c
+++ b/ShellPkg/Library/UefiShellLevel2CommandsLib/Rm.c
@@ -321,41 +321,44 @@ MainCmdRm (
     }
   }
 
-  if (ShellStatus == SHELL_SUCCESS) {
-    //
-    // loop through the list and make sure we are not aborting...
-    //
-    for ( Node = (EFI_SHELL_FILE_INFO *)GetFirstNode (&FileList->Link)
-          ; !IsNull (&FileList->Link, &Node->Link) && !ShellGetExecutionBreakFlag ()
-          ; Node = (EFI_SHELL_FILE_INFO *)GetNextNode (&FileList->Link, &Node->Link)
-          )
-    {
-      //
-      // skip the directory traversing stuff...
-      //
-      if ((StrCmp (Node->FileName, L".") == 0) || (StrCmp (Node->FileName, L"..") == 0)) {
-        continue;
-      }
-
-      //
-      // do the deleting of nodes
-      //
-      if (EFI_ERROR (Node->Status)) {
-        ShellPrintHiiDefaultEx (STRING_TOKEN (STR_RM_LOG_DELETE_ERR2), gShellLevel2HiiHandle, Node->Status);
-        ShellStatus = SHELL_ACCESS_DENIED;
-        break;
-      }
-
-      if (!IsValidDeleteTarget (FileList, Node, Package)) {
-        ShellPrintHiiDefaultEx (STRING_TOKEN (STR_RM_LOG_DELETE_ERR3), gShellLevel2HiiHandle, Node->FullName);
-        ShellStatus = SHELL_INVALID_PARAMETER;
-        break;
-      }
-
-      ShellStatus = CascadeDelete (Node, ShellCommandLineGetFlag (Package, L"-q"));
-    }
+  if ((ShellStatus != SHELL_SUCCESS) || (FileList == NULL)) {
+    goto Exit;
   }
 
+  //
+  // loop through the list and make sure we are not aborting...
+  //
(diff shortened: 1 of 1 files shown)


ShellPkg/Library/UefiShellLevel2CommandsLib/Ls.c
```diff
@@ -870,7 +870,7 @@ MainCmdLs (
       TheTime.TimeZone = EFI_UNSPECIFIED_TIMEZONE;
     }
 
-    if (ShellStatus == SHELL_SUCCESS) {
+    if ((ShellStatus == SHELL_SUCCESS) && (FullPath != NULL)) {
```
leiflindholm: If we additionally stop the success handling of lines 793 and 819, can we drop the test on this line completely?

pierregondois: I am not sure I found the conditions you were referring to at line 819.
I added a patch `ShellPkg/UefiShellLevel2/Ls: Flatten MainCmdLs()` to flatten the function while removing this condition, but CodeQl complained, so I put back a check against `FullPath`

ShellPkg/Library/UefiShellLevel2CommandsLib/Ls.c
```diff
@@ -733,6 +728,209 @@
...
+
+  if (ShellStatus != SHELL_SUCCESS) {
+    goto Exit;
+  }
+
+  ShellStatus = PrintLsOutput (
+                  ShellCommandLineGetFlag (Package, L"-r"),
+                  RequiredAttributes,
+                  ShellCommandLineGetFlag (Package, L"-sfo"),
+                  FullPath,
```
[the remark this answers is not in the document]
leiflindholm: Guess not :)

However, maybe we could do an error return/goto if `FullPath == NULL` instead?

ShellPkg/Library/UefiShellLevel2CommandsLib/Ls.c
leiflindholm: The test on 819 detects an error, sets `ShellStatus = SHELL_NOT_FOUND`, but continues with the next (success) block in the `else` case. It could do a return/goto instead.

ShellPkg/Library/UefiShellLevel2CommandsLib/Ls.c
```diff
@@ -790,112 +790,119 @@ MainCmdLs (
...
+    //
+    // Copy to the 2 strings for starting path and file search string
+    //
+    ASSERT (SearchString == NULL);
+    ASSERT (FullPath == NULL);
+    StrnCatGrow (&SearchString, NULL, L"*", 0);
+    StrnCatGrow (&FullPath, NULL, CurDir, 0);
+    Size = FullPath != NULL ? StrSize (FullPath) : 0;
+    StrnCatGrow (&FullPath, &Size, L"\\", 0);
+    goto Exit;
```
leiflindholm: This is incorrect, I believe.
As i read the code, this is the bit that generates where to start from if ls is called without arguments. So we don't want to bail here.
But we do want to bail a few lines up if `(CurDir == NULL)`.

pierregondois: Yes right, I m fixing it, sorry for this

leiflindholm: No problem, this is excellent cleanup.

ShellPkg/Library/UefiShellLevel2CommandsLib/Ls.c
```diff
@@ -820,50 +820,50 @@ MainCmdLs (
       //
       // If we got something and it doesnt have a fully qualified path, then we needed to have a CWD.
       //
-      ShellStatus = SHELL_NOT_FOUND;
```
leiflindholm: OK, so, with this major simplification of the code, the check on line 876 is now _definitely_ completely redundant. i.e. Nothing touches ShellStatus between 793 and 876.

So can you drop that test please?

Final comment, promise :)

pierregondois: Ok yes, thanks for the thorough review

ardbiesheuvel:
@mergifyio requeue

Issue #11747, opened 2025-11-12
[Bug]: QemuKernelLoaderFsDxe masks RiscVVirtQemu platform bug when fw_cfg is available


zevorn:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

OvmfPkg

### Current Behavior

The QemuKernelLoaderFsDxe module has a hard dependency on the fw_cfg device, causing it to return EFI_NOT_FOUND when fw_cfg is not available. On QEMU rvsp-ref platform (which lacks fw_cfg device), this leads to incomplete UEFI initialization that subsequently causes kernel panic during Runtime Services execution.

The issue manifests as a kernel panic in openEuler RISC-V 25.09 when running on QEMU rvsp-ref platform with EDK2 202508, specifically during UEFI Runtime Services calls that attempt to access Hand-Off Block (HOB) data structures with virtual addressing enabled.

### Expected Behavior

```
[ 0.019650] Unable to handle kernel paging request at virtual address 0x000000027fe3b018 
[ 0.020566] Oops [#1] [ 0.020832] Modules linked in: 
[ 0.021159] CPU: 0 PID: 1 Comm: swapper/0 Not tainted 6.6.0+ #1 
[ 0.021159] Hardware name: riscv-virtio,qemu (DT) 
[ 0.021159] epc : 0x000000002003ee7a ra : 0x000000002003ee6e sp : 0xffffffc800c03d10 
[ 0.023171] Call Trace: 
[ 0.023171] [<000000002003ee7a>] GetNextHob+0x1a/0x2c 
[ 0.023171] [<000000002003ef8c>] GetFirstGuidHob+0x20/0x30 
[ 0.023171] [<0000000020050a9e>] GetPerformanceCounterProperties+0x42/0x8c
...
```

The panic occurs when the kernel attempts to call UEFI Runtime Services, specifically GetTime(), which internally calls GetPerformanceCounterProperties(). This function tries to access HOB (Hand-Off Block) data at physical address 0x27FE3B018, but the address is treated as virtual in the kernel's address space, causing a page fault.

#### 1. fw_cfg Dependency in QemuKernelLoaderFsDxe

The problematic code in QemuKernelLoaderFsDxeEntrypoint:

```c
EFI_STATUS
EFIAPI
QemuKernelLoaderFsDxeEntrypoint (
[...]

PR #11819, opened 2025-11-29
UefiCpuPkg/BaseRiscV64CpuTimerLib: Add constructor to initialize mTimeBase

zevorn:
# Description

Add a constructor function that calls GetPerformanceCounterProperties(NULL, NULL) to ensure mTimeBase is initialized during DXE phase, before virtual memory is enabled.

This fixes crashes on RISC-V platforms without fw_cfg device support (e.g., QEMU rvsp-ref) where Runtime Services may call GetPerformanceCounterProperties() before mTimeBase has been initialized. When virtual memory is then enabled, the uninitialized mTimeBase value causes HOB access to fail with a page fault.

Fixes: https://github.com/tianocore/edk2/issues/11747

## How This Was Tested

To boot a kernel that has enabled the EFI timer service for time retrieval using EDKII on the QEMU rvsp-ref Machine (or other boards without fw_cfg support).

Since the rvsp-ref board has not yet been merged into the QEMU mainline, you can test it through my personal repo:

https://github.com/zevorn/qemu/tree/rvsp-ref-v4

## Integration Instructions

N/A

### DIFF
--- a/UefiCpuPkg/Library/BaseRiscV64CpuTimerLib/BaseRiscV64CpuTimerLib.inf
+++ b/UefiCpuPkg/Library/BaseRiscV64CpuTimerLib/BaseRiscV64CpuTimerLib.inf
@@ -16,6 +16,7 @@
   VERSION_STRING                 = 1.0
   LIBRARY_CLASS                  = TimerLib
   MODULE_UNI_FILE                = BaseRisV64CpuTimerLib.uni
+  CONSTRUCTOR                    = BaseRiscV64CpuTimerLibConstructor
 
 [Sources]
   CpuTimerLib.c
--- a/UefiCpuPkg/Library/BaseRiscV64CpuTimerLib/CpuTimerLib.c
+++ b/UefiCpuPkg/Library/BaseRiscV64CpuTimerLib/CpuTimerLib.c
@@ -21,8 +21,6 @@
 
 STATIC UINT64  mTimeBase;
 
-#define GET_TIME_BASE()  (mTimeBase ?: GetPerformanceCounterProperties(NULL, NULL))
-
 /**
   Stalls the CPU for at least the given number of ticks.
 
@@ -67,7 +65,7 @@ MicroSecondDelay (
     DivU64x32 (
       MultU64x32 (
         MicroSeconds,
-        GET_TIME_BASE ()
+        mTimeBase
         ),
       1000000u
       )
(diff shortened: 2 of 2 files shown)


xfan1024:
In my view, you might consider moving `mTimeBase`  initialization from `GetPerformanceCounterProperties` into `Constructor`.
Once the `Constructor` has been called, we don't need to check `mTimeBase` is zero everywhere, which can make the code logic simpler and clearer.

In this way: 

1. You can use `mTimeBase` instead of `GET_TIME_BASE()` and remove this macro.
2. In `GetPerformanceCounterProperties`, you can return `mTimeBase` directly without checking it was zero.

zevorn:
> In my view, you might consider moving `mTimeBase` initialization from `GetPerformanceCounterProperties` into `Constructor`. Once the `Constructor` has been called, we don't need to check `mTimeBase` is zero everywhere, which can make the code logic simpler and clearer.
> 
> In this way:
> 
> 1. You can use `mTimeBase` instead of `GET_TIME_BASE()` and remove this macro.
> 2. In `GetPerformanceCounterProperties`, you can return `mTimeBase` directly without checking it was zero.

Since I'm not very familiar with EDKII, to be on the safe side, I chose to make minimal modifications. Does anyone else have better suggestions?

vlsunil:
> > In my view, you might consider moving `mTimeBase` initialization from `GetPerformanceCounterProperties` into `Constructor`. Once the `Constructor` has been called, we don't need to check `mTimeBase` is zero everywhere, which can make the code logic simpler and clearer.
> > In this way:
> > 
> > 1. You can use `mTimeBase` instead of `GET_TIME_BASE()` and remove this macro.
> > 2. In `GetPerformanceCounterProperties`, you can return `mTimeBase` directly without checking it was zero.
> 
> Since I'm not very familiar with EDKII, to be on the safe side, I chose to make minimal modifications. Does anyone else have better suggestions?

@xfan1024 's suggestion to use mTimeBase instead of GET_TIME_BASE() LGTM.

zevorn:
> > > In my view, you might consider moving `mTimeBase` initialization from `GetPerformanceCounterProperties` into `Constructor`. Once the `Constructor` has been called, we don't need to check `mTimeBase` is zero everywhere, which can make the code logic simpler and clearer.
> > > In this way:
> > > 
> > > 1. You can use `mTimeBase` instead of `GET_TIME_BASE()` and remove this macro.
> > > 2. In `GetPerformanceCounterProperties`, you can return `mTimeBase` directly without checking it was zero.
> > 
> > 
> > Since I'm not very familiar with EDKII, to be on the safe side, I chose to make minimal modifications. Does anyone else have better suggestions?
> 
> @xfan1024 's suggestion to use mTimeBase instead of GET_TIME_BASE() LGTM.

I have updated the PR, please review again, thanks :)

PR #11820, opened 2025-11-29
SecurityPkg/HashLibTpm2: Fix unused variable usage

mikebeaton:
# Description

There is an unused variable in SecurityPkg/Library/HashLibTpm2/HashLibTpm2PeilessSecLib.c which is spotted by CLANG builds.

The correct fix here is not to just remove the unused assignment, but rather to fix the line following it, which should have been using the unused value in the way that it now does.

Fixes: https://github.com/tianocore/edk2/commit/9c651ef83a74d83219075316aa5d1c189ed608c8

- [ ] Breaking change?
  - NO.
- [ ] Impacts security?
  - NO.
- [ ] Includes tests?
  - NO.

## How This Was Tested

CI. Peer advice. Has not been run.

## Integration Instructions

N/A

### DIFF
--- a/SecurityPkg/Library/HashLibTpm2/HashLibTpm2PeilessSecLib.c
+++ b/SecurityPkg/Library/HashLibTpm2/HashLibTpm2PeilessSecLib.c
@@ -586,9 +586,7 @@ HasLibTpm2PeilessSecLibConstructor (
   }
 
   TcgPcrEvent             = (TCG_PCR_EVENT *)EventLog;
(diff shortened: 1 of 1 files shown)


mikebeaton:
@ardbiesheuvel It seemed sensible to [split this off](https://github.com/tianocore/edk2/pull/11761). I believe this is likely to get merged faster than the build flags change, and anyway it stands alone if it's not just removing a harmless unused variable as it was in my previous version. Can you approve (incl. your co-authorship)? :-)

PR #11821, opened 2025-11-30
BaseTools/PatchCheck.py: Disallow commit subject lines that end with period

mikebeaton:
# Description

Disallow commit subject lines that end with a period.

The vast majority of commit subject lines in the repo don't end with a period, though a very few do, and I understood it to be a standard - in which case it seems worth adding it to the patch check?

- [ ] Breaking change?
  - NO.
- [ ] Impacts security?
  - NO.
- [ ] Includes tests?
  - NO.

## How This Was Tested

Initial version of this PR had subject line that ended with period to test, and triggered newly added error.

## Integration Instructions

N/A


mikebeaton:
Wondered if @ardbiesheuvel @leiflindholm as very active contributors and reviewers had an opinion on this.

gapalomi:
I think it's good to have more uniform commits subjects. And period doesn't provide more information to the subject. 

**I'm ok with this change.** 👍

ardbiesheuvel:
@leiflindholm might have an opinion on this, I really don't.

mikebeaton:
> @leiflindholm might have an opinion on this, I really don't.

Sure - I am getting used to terse style enough to understand that no reply would have meant that too. So probably already does, for leiflindholm.

leiflindholm:
> @leiflindholm might have an opinion on this, I really don't.

Same, really. I'm not opposed.

mikebeaton:
@gapalomi - Is this okay to add push label?

mikebeaton:
Anyone okay to add push label, if this is good to go?

mikebeaton:
I think everyone agreed this is correct as far as it goes. Can anyone add the push label?

PR #11828, opened 2025-12-02
MdeModulePkg/UsbBusDxe: BOS Descriptor Check for SuperSpeed Devices

MarlboroChuang:
Some SuperSpeed-capable devices may fall back to High-Speed mode and
cause subsequent commands to fail.

[Suggested Solution]
Check the BOS descriptor to verify SuperSpeed support and trigger a
port reset if needed to re-enumerate the device properly.

### DIFF
--- a/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbDesc.c
+++ b/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbDesc.c
@@ -114,6 +114,10 @@ UsbFreeDevDesc (
     FreePool (DevDesc->Configs);
   }
 
+  if (DevDesc->BosDesc != NULL) {
+    FreePool (DevDesc->BosDesc);
+  }
+
   FreePool (DevDesc);
 }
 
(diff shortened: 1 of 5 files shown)


MdeModulePkg/Bus/Usb/UsbBusDxe/UsbDesc.c
```diff
@@ -319,6 +323,46 @@ UsbParseInterfaceDesc (
...
+BOOLEAN
+UsbIsSSDevice (
+  IN UINT8  *DescBuf,
+  IN UINTN  Len
+  )
+{
+  UINT8             *NextCap;
+  UINT8             Index;
+  UINT8             NumCapDesc;
+  USB_BOS_DESC      *BOSDesc;
```
jkolakow: Per camel case it should rather be BosDesc

MdeModulePkg/Bus/Usb/UsbBusDxe/UsbDesc.h
```diff
@@ -69,8 +69,34 @@ typedef struct {
 typedef struct {
   EFI_USB_DEVICE_DESCRIPTOR    Desc;
   USB_CONFIG_DESC              **Configs;
+  UINT8                        *BOSDesc;
```
jkolakow: Could we call it BosDesc to follow camel case?
Also if I understand correct this pointer is a pointer to USB_BOS_DESC defined below. Why don't we declare it "USB_BOS_DESC  *BosDesc;" ?  The definitions in lines 75..96 could be moved to line 32, isn't it ? What is the reason of hiding the true type under UINT8* ?

MdeModulePkg/Bus/Usb/UsbBusDxe/UsbDesc.c
```diff
@@ -319,6 +323,46 @@ UsbParseInterfaceDesc (
...
+  @param  DescBuf               The buffer of raw descriptor.
+  @param  Len                   The length of the raw descriptor buffer.
+
+  @return TRUE                  The device is a SS device
+          FALSE                 The device is not a SS device
+
+**/
+BOOLEAN
+UsbIsSSDevice (
+  IN UINT8  *DescBuf,
```
jkolakow: This function gets USB_BOS_DESC pointer as parameter, isn't it ? Why don't we just declare the parameter as USB_BOS_DESC ?

MdeModulePkg/Bus/Usb/UsbBusDxe/UsbDesc.c
MarlboroChuang: Thanks for reviewing.
resolved in current commit.

MdeModulePkg/Bus/Usb/UsbBusDxe/UsbDesc.h
MarlboroChuang: Would like to indicate that is a continous buffer.
However, I resolved it in current commit.
As per current EDK2 implementation, I moved the descriptor definition to Usb.h since that is follow USB spec.

MdeModulePkg/Bus/Usb/UsbBusDxe/UsbDesc.c
MarlboroChuang: Thanks for reviewing.
resolved it in current commit

PR #11830, opened 2025-12-02
MdeModulePkg/UsbBusDxe: UsbBusRecursivelyConnectWantedUsbIo TPL Handling

MarlboroChuang:
MdeModulePkg/UsbBusDxe: Raise TPL to CALLBACK for wanted UsbIo connect

Raise TPL to TPL_CALLBACK around the ConnectController() call in
UsbBusRecursivelyConnectWantedUsbIo() and restore the original TPL
afterward.

UsbBusRecursivelyConnectWantedUsbIo() is normally reached through the
DriverBinding Start routine invoked by gBS->ConnectController(), which
runs at TPL_APPLICATION. While the wanted UsbIo handle is being
connected, the USB device can be removed. The device removal polling
event in UsbBus runs at TPL_CALLBACK. If the connect runs at
TPL_APPLICATION, that event can preempt the connection and cause
commands to be sent to a stale or non-existent USB device address,
producing a lot of timeout transfers.

This differs from UsbConnectDriver(), which is called from the
enumeration polling event (XHCI async event) and is usually already
at TPL_CALLBACK or TPL_NOTIFY. That function uses RestoreTPL()/
RaiseTPL() to ensure ConnectController() runs at TPL_CALLBACK.

### DIFF
--- a/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbUtility.c
+++ b/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbUtility.c
@@ -1042,6 +1042,7 @@ UsbBusRecursivelyConnectWantedUsbIo (
   UINTN                     UsbIoHandleCount;
   EFI_HANDLE                *UsbIoBuffer;
   EFI_DEVICE_PATH_PROTOCOL  *UsbIoDevicePath;
+  EFI_TPL                   Tpl;
 
   if (UsbBusId == NULL) {
     return EFI_INVALID_PARAMETER;
@@ -1100,9 +1101,17 @@ UsbBusRecursivelyConnectWantedUsbIo (
         // Recursively connect the wanted Usb Io handle
         //
         DEBUG ((DEBUG_INFO, "UsbBusRecursivelyConnectWantedUsbIo: TPL before connect is %d\n", (UINT32)UsbGetCurrentTpl ()));
+        Tpl = UsbGetCurrentTpl ();
+        if (Tpl < TPL_CALLBACK) {
+          Tpl = gBS->RaiseTPL (TPL_CALLBACK);
+        }
+
         Status           = gBS->ConnectController (UsbIf->Handle, NULL, NULL, TRUE);
         UsbIf->IsManaged = (BOOLEAN) !EFI_ERROR (Status);
         DEBUG ((DEBUG_INFO, "UsbBusRecursivelyConnectWantedUsbIo: TPL after connect is %d\n", (UINT32)UsbGetCurrentTpl ()));
+        if (Tpl < TPL_CALLBACK) {
+          gBS->RestoreTPL (Tpl);
+        }
       }
     }
   }


jkolakow (approved):

mikebeaton:
@MarlboroChuang - You have not adopted the pattern from UsbConnectDriver (which lowers the TPL if needed, so as to never call ConnectController at a TPL above TPL_CALLBACK). Instead you've put in a different pattern, to raise the TPL if needed, so as to never call ConnectController at a TPL _below_ TPL_CALLBACK. This may be what you intended anyway, but it makes the commit message misleading, and also in need of more information as to why you believe this is necessary.

MarlboroChuang:
> @MarlboroChuang - You have not adopted the pattern from UsbConnectDriver (which lowers the TPL if needed, so as to never call ConnectController at a TPL above TPL_CALLBACK). Instead you've put in a different pattern, to raise the TPL if needed, so as to never call ConnectController at a TPL _below_ TPL_CALLBACK. This may be what you intended anyway, but it makes the commit message misleading, and also in need of more information as to why you believe this is necessary.

Thanks for comments. I think you are right. Will make a bit of changes since I believe that we should follow the practices from UEFI spec that the RaiseTPL / RestoreTPL should be paired and sequential.

mikebeaton:
> > @MarlboroChuang - You have not adopted the pattern from UsbConnectDriver (which lowers the TPL if needed, so as to never call ConnectController at a TPL above TPL_CALLBACK). Instead you've put in a different pattern, to raise the TPL if needed, so as to never call ConnectController at a TPL _below_ TPL_CALLBACK. This may be what you intended anyway, but it makes the commit message misleading, and also in need of more information as to why you believe this is necessary.
> 
> Thanks for comments. I think you are right. Will make a bit of changes since I believe that we should follow the practices from UEFI spec that the RaiseTPL / RestoreTPL should be paired and sequential.

But do you have any reason this think this code is ever called in a way that needs that other pattern (i.e. is ever called above TPL_CALLBACK)? If not, why the change, or if so can you clarify what issue you are seeing - thanks.

MarlboroChuang:
> > > @MarlboroChuang - You have not adopted the pattern from UsbConnectDriver (which lowers the TPL if needed, so as to never call ConnectController at a TPL above TPL_CALLBACK). Instead you've put in a different pattern, to raise the TPL if needed, so as to never call ConnectController at a TPL _below_ TPL_CALLBACK. This may be what you intended anyway, but it makes the commit message misleading, and also in need of more information as to why you believe this is necessary.
> > 
> > 
> > Thanks for comments. I think you are right. Will make a bit of changes since I believe that we should follow the practices from UEFI spec that the RaiseTPL / RestoreTPL should be paired and sequential.
> 
> But do you have any reason this think this code is ever called in a way that needs that other pattern (i.e. is ever called above TPL_CALLBACK)? If not, why the change, or if so can you clarify what issue you are seeing - thanks.

HI @mikebeaton ,
I found the previous comment from me is not correct.
It seems that my commit message is wrong but the implementation should be fine.
- With UsbBusRecursivelyConnectWantedUsbIo(), it would be normally called from gBS->ConnectController() that should be from TPL_APPLICATION level. So we need to raise it to TPL_CALLBACK. The problem here is the device could be removed while it is connecting. The device remove polling event in UsbBus is TPL_CALLBACK level. So if we don't set the TPL_CALLBACK level, it might encounter the command sent to wrong non-existing address and then cause a lot of timeout transfer.

- For UsbConnectDriver(), it will be called from polling event while device is getting enumerated and it was usually to be put as CALLBACK or NOTIFY level(From XHCI ASYNC event). So that's why it would need to restore to CALLBACK.

Does it make sense?
Thanks a lot.

MarlboroChuang:
Hi @lgao4 ,
Could you please help to review it?
Thanks a lot.

lgao4:
> Hi @lgao4 , Could you please help to review it? Thanks a lot.

Can you describe the reproduce step for this issue?

MarlboroChuang:
> > Hi @lgao4 , Could you please help to review it? Thanks a lot.
> 
> Can you describe the reproduce step for this issue?

Thanks for the reply.
Our case is that we will encounter system stalled at POST while we unplug the USB device in middle of enumeration.
Usb Keyboard Mouse will be connected while system is enummerating consolein and USB bus driver is started but without connecting all the downstream devices.
Later, the ConnectSequence will connect the USB controller again and it will run into UsbBusRecursivelyConnectWantedUsbIo() routine.
If we remove the USB device like USB storage that would be enumerated by the ConnectController this time, then it have the fail rate to encounter the device context freed prior to the Usb Device  enumeration finished.
That would introduce the stall issue.

lgao4:
> > > Hi @lgao4 , Could you please help to review it? Thanks a lot.
> > 
> > 
> > Can you describe the reproduce step for this issue?
> 
> Thanks for the reply. Our case is that we will encounter system stalled at POST while we unplug the USB device in middle of enumeration. Usb Keyboard Mouse will be connected while system is enummerating consolein and USB bus driver is started but without connecting all the downstream devices. Later, the ConnectSequence will connect the USB controller again and it will run into UsbBusRecursivelyConnectWantedUsbIo() routine. If we remove the USB device like USB storage that would be enumerated by the ConnectController this time, then it have the fail rate to encounter the device context freed prior to the Usb Device enumeration finished. That would introduce the stall issue.

Seemly, this is a corner case.

MarlboroChuang:
> > > > Hi @lgao4 , Could you please help to review it? Thanks a lot.
> > > 
> > > 
> > > Can you describe the reproduce step for this issue?
> > 
> > 
> > Thanks for the reply. Our case is that we will encounter system stalled at POST while we unplug the USB device in middle of enumeration. Usb Keyboard Mouse will be connected while system is enummerating consolein and USB bus driver is started but without connecting all the downstream devices. Later, the ConnectSequence will connect the USB controller again and it will run into UsbBusRecursivelyConnectWantedUsbIo() routine. If we remove the USB device like USB storage that would be enumerated by the ConnectController this time, then it have the fail rate to encounter the device context freed prior to the Usb Device enumeration finished. That would introduce the stall issue.
> 
> Seemly, this is a corner case.

Our testing frequently involves unplug/replug scenarios. In addition, some devices may reset themselves unexpectedly. We've actually encountered this kind of behavior quite a few times in the field. From the code logic perspective, this change aligns with the existing UsbConnectController behavior by performing the device driver connection at the TPL_CALLBACK level. Do you think that approach is reasonable?

lgao4:
> > > > > Hi @lgao4 , Could you please help to review it? Thanks a lot.
> > > > 
> > > > 
> > > > Can you describe the reproduce step for this issue?
> > > 
> > > 
> > > Thanks for the reply. Our case is that we will encounter system stalled at POST while we unplug the USB device in middle of enumeration. Usb Keyboard Mouse will be connected while system is enummerating consolein and USB bus driver is started but without connecting all the downstream devices. Later, the ConnectSequence will connect the USB controller again and it will run into UsbBusRecursivelyConnectWantedUsbIo() routine. If we remove the USB device like USB storage that would be enumerated by the ConnectController this time, then it have the fail rate to encounter the device context freed prior to the Usb Device enumeration finished. That would introduce the stall issue.
> > 
> > 
> > Seemly, this is a corner case.
> 
> Our testing frequently involves unplug/replug scenarios. In addition, some devices may reset themselves unexpectedly. We've actually encountered this kind of behavior quite a few times in the field. From the code logic perspective, this change aligns with the existing UsbConnectController behavior by performing the device driver connection at the TPL_CALLBACK level. Do you think that approach is reasonable?

I agree this fix.

lgao4 (approved):

MarlboroChuang:
> > > > > > Hi @lgao4 , Could you please help to review it? Thanks a lot.
> > > > > 
> > > > > 
> > > > > Can you describe the reproduce step for this issue?
> > > > 
> > > > 
> > > > Thanks for the reply. Our case is that we will encounter system stalled at POST while we unplug the USB device in middle of enumeration. Usb Keyboard Mouse will be connected while system is enummerating consolein and USB bus driver is started but without connecting all the downstream devices. Later, the ConnectSequence will connect the USB controller again and it will run into UsbBusRecursivelyConnectWantedUsbIo() routine. If we remove the USB device like USB storage that would be enumerated by the ConnectController this time, then it have the fail rate to encounter the device context freed prior to the Usb Device enumeration finished. That would introduce the stall issue.
> > > 
> > > 
> > > Seemly, this is a corner case.
> > 
> > 
> > Our testing frequently involves unplug/replug scenarios. In addition, some devices may reset themselves unexpectedly. We've actually encountered this kind of behavior quite a few times in the field. From the code logic perspective, this change aligns with the existing UsbConnectController behavior by performing the device driver connection at the TPL_CALLBACK level. Do you think that approach is reasonable?
> 
> I agree this fix.

Thanks @lgao4 .
May I know if this PR can be merged?
Regards,
Marlboro.

lgao4:
> > > > > > > Hi @lgao4 , Could you please help to review it? Thanks a lot.
> > > > > > 
> > > > > > 
> > > > > > Can you describe the reproduce step for this issue?
> > > > > 
> > > > > 
> > > > > Thanks for the reply. Our case is that we will encounter system stalled at POST while we unplug the USB device in middle of enumeration. Usb Keyboard Mouse will be connected while system is enummerating consolein and USB bus driver is started but without connecting all the downstream devices. Later, the ConnectSequence will connect the USB controller again and it will run into UsbBusRecursivelyConnectWantedUsbIo() routine. If we remove the USB device like USB storage that would be enumerated by the ConnectController this time, then it have the fail rate to encounter the device context freed prior to the Usb Device enumeration finished. That would introduce the stall issue.
> > > > 
> > > > 
> > > > Seemly, this is a corner case.
> > > 
> > > 
> > > Our testing frequently involves unplug/replug scenarios. In addition, some devices may reset themselves unexpectedly. We've actually encountered this kind of behavior quite a few times in the field. From the code logic perspective, this change aligns with the existing UsbConnectController behavior by performing the device driver connection at the TPL_CALLBACK level. Do you think that approach is reasonable?
> > 
> > 
> > I agree this fix.
> 
> Thanks @lgao4 . May I know if this PR can be merged? Regards, Marlboro.

It will be merged after this stable tag is created on this Friday.

PR #11831, opened 2025-12-02
MdeModulePkg/UsbBusDxe: Manufacturer String Descriptor Caching

MarlboroChuang:
MdeModulePkg/UsbBusDxe: Manufacturer String Descriptor Caching

Certain devices require immediate follow-up commands after reading the LANGID
string to fetch Manufacturer, Product, or SerialNumber strings.

[Suggested Solution]
These strings are now cached after initial retrieval to allow
UsbIoGetStringDescriptor() to return them directly, improving efficiency
and stability.

### DIFF
--- a/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbDesc.h
+++ b/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbDesc.h
@@ -69,6 +69,9 @@ typedef struct {
 typedef struct {
   EFI_USB_DEVICE_DESCRIPTOR    Desc;
   USB_CONFIG_DESC              **Configs;
+  UINT8                        *StrDescManufacturerUS;
+  UINT8                        *StrDescProductUS;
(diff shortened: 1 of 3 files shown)


MdeModulePkg/Bus/Usb/UsbBusDxe/UsbDesc.h
```diff
@@ -69,6 +69,10 @@ typedef struct {
 typedef struct {
   EFI_USB_DEVICE_DESCRIPTOR    Desc;
   USB_CONFIG_DESC              **Configs;
+  UINT8                        *BOSDesc;
```
jkolakow: Could we call it BosDesc to follow camel case ?

MarlboroChuang: This is not necessary for current PR.
Removed it.

PR #11834, opened 2025-12-02
MdeModulePkg/UsbBusDxe: Improve USB enumerating process

MarlboroChuang:
MdeModulePkg/UsbBusDxe: Improve USB enumerating process

The patch enhances the USB enumeration process in EDK2 to improve
compatibility with non-standards-compliant devices that may fail during
standard enumeration sequences.
The suggested solution is based on USB specifications and references
implementations from both Linux and Windows environments.

[Suggested solution]
- Integrated a retry mechanism to sequentially execute enumeration scripts,
  inspired by the enumeration flows of Windows, Linux, and EDK2.
  This improves robustness when handling corner-case devices.
- Do sanity check while the device report the device descriptor.
- AMD XHCI might need to wait for more time while sending the CLEAR_FEATURE
  reuqest.

### DIFF
--- a/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbEnumer.c
+++ b/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbEnumer.c
@@ -233,6 +233,7 @@ UsbCreateDevice (
   Device->ParentIf   = ParentIf;
   Device->ParentPort = ParentPort;
   Device->Tier       = (UINT8)(ParentIf->Device->Tier + 1);
+  Device->EnumScript = 0;
   return Device;
 }
 
@@ -673,6 +674,7 @@ UsbEnumerateNewDev (
   UINTN                Address;
   UINT8                Config;
   EFI_STATUS           Status;
+  UINT8                RetryCount;
 
   Parent  = HubIf->Device;
   Bus     = Parent->Bus;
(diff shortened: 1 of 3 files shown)


MdeModulePkg/Bus/Usb/UsbBusDxe/UsbEnumer.c
```diff
@@ -717,9 +722,12 @@ UsbEnumerateNewDev (
     goto ON_ERROR;
   }
 
+  Child->EnumScript = RetryCount;
```
jkolakow, line 741: If I understand correct at each retry you want to take different approach/script. Could you add few words explanation why start at UsbEnumScriptWin and what happens at UsbEnumScriptRsrv ?

MarlboroChuang: add the table to explain more for the script

jkolakow: Great, the comment explains. Just for the case '2', did you mean "Length" ?

MarlboroChuang: Fixed it.

Issue #11790, opened 2025-11-24
[Bug]: CryptoPkg cannot be built with VS2022


parfant:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

CryptoPkg

### Which targets are impacted by this bug?

RELEASE

### Current Behavior

CryptoPkg cannot be built with VS2022:
`SECTION - Building CryptoPkg Package
PROGRESS - --Running CryptoPkg: Compiler Plugin RELEASE --
PROGRESS - Start time: 2025-11-24 08:56:24.234171
PROGRESS - Setting up the Environment
PROGRESS - Running Pre Build
PROGRESS - Checking DEBUG Macros
PROGRESS - Running Build RELEASE
ERROR - Compiler #2220 from C:\edk2\CryptoPkg\Library\OpensslLib\openssl\crypto\bn\bn_gcd.c(659): the following warning is treated as an error
WARNING - Compiler #4319 from C:\edk2\CryptoPkg\Library\OpensslLib\openssl\crypto\bn\bn_gcd.c(659): '~': zero extending 'unsigned int' to 'unsigned __int64' of greater size
WARNING - Compiler #4319 from C:\edk2\CryptoPkg\Library\OpensslLib\openssl\crypto\bn\bn_gcd.c(671): '~': zero extending 'unsigned int' to 'unsigned __int64' of greater size
ERROR - Compiler #1077 from NMAKE : fatal '"C:\BuildTools\VC\Tools\MSVC\14.44.35207\bin\Hostx86\x64\cl.exe" /FoC:\edk2\Build\CryptoPkg\All\RELEASE_VS2022\X64\CryptoPkg\Library\OpensslLib\OpensslLibCrypto\OUTPUT\openssl\crypto\bn\ /showIncludes @C:\edk2\Build\CryptoPkg\All\RELEASE_VS2022\X64\CryptoPkg\Library\OpensslLib\OpensslLibCrypto\OUTPUT\cc_resp.txt @C:\edk2\Build\CryptoPkg\All\RELEASE_VS2022\X64\CryptoPkg\Library\OpensslLib\OpensslLibCrypto\OUTPUT\cc_resp_612.txt' : return code '0x2'
ERROR - Compiler #7000 from : Failed to execute command
ERROR - Compiler #7000 from : Failed to execute command
ERROR - Compiler #7000 from : Failed to execute command
[...]

PR #11836, opened 2025-12-03
CryptoPkg: Suppress VS2022 warning #4319 when building OpenSSL

ardbiesheuvel:
CI builds have started to fail with

ERROR - Compiler #2220 from D:\a\1\s\CryptoPkg\Library\OpensslLib\openssl\crypto\bn\bn_gcd.c(659): the following warning is treated as an error
WARNING - Compiler #4319 from D:\a\1\s\CryptoPkg\Library\OpensslLib\openssl\crypto\bn\bn_gcd.c(659): '~': zero extending 'unsigned int' to 'unsigned __int64' of greater size
WARNING - Compiler #4319 from D:\a\1\s\CryptoPkg\Library\OpensslLib\openssl\crypto\bn\bn_gcd.c(671): '~': zero extending 'unsigned int' to 'unsigned __int64' of greater size

which was not flagged before. Suppress the warning in CryptoPkg so the builds will succeed again. If this is a real issue, it should be reported to and fixed in the upstream project.

### DIFF
--- a/OvmfPkg/PlatformCI/.azurepipelines/Ubuntu-GCC.yml
+++ b/OvmfPkg/PlatformCI/.azurepipelines/Ubuntu-GCC.yml
@@ -51,53 +51,6 @@ jobs:
             Run.Flags: $(run_flags)
             Run: $(should_run)
 
-          OVMF_IA32X64_DEBUG:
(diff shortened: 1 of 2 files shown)


khaliid2040:
@ardbiesheuvel I think you should also close this issue #11790

PR #11838, opened 2025-12-04
OvmfPkg/MemDebugLogLib: use AcquireSpinLockOrFail

kraxel:
Drop log lines if we can't get the spin lock.  Not nice, but better than
risking a deadlock.

Some background:  Most of edk2 runs single-threaded on the BSP, so if
something holds the lock it is rather unlikely that waiting is going to
help.  Specifically I think a deadlock can happen if (a) a timer
interrupt arrives while the lock is held, and (b) some higher-TPL timer
handler tries to print something to the debug log.


ajyoung-oracle:
Hi Gerd/Ard. I just noticed the commit for this PR and wanted to comment that I agree with this change. 

And in fact I recent made a VERY similar (out-of-tree) change to our (Oracle) OVMF code where I added a new AcquireSpinLockOrFailTimeout() call used by the MemDebugLogLib code instead AcquireSpinLock(). 
AcquireSpinLockOrFailTimeout() is basically the same as AcquireSpinLock() but instead of ASSERTing on timeout, just return FALSE (i.e. lock acquire failed). Very similar idea - to avoid any potential dead-locking failures resulting in an ASSERT/boot failure.

Out of curiosity, was your change to fix a purely theoretical problem? Or was a MemDebugLogLib lock deadlock/ASSERT actually encountered/reported by someone?

Thanks!

@kraxel ^^

kraxel:
Fedora QA has seen sporadic boot failures with spinlock assert in the logs.

PR #11839, opened 2025-12-04
Use common String utilities

pierregondois:
# Description

The edk2 repository contains multiple local implementations of string utilities:
- checking a string/char uppercase/lowercase/alphanumerical/hexadecimal compliance
- replacing uppercase string/char to lowercase (or conversely)

Add string/char utilities to BaseLib and remove some of the local implementations.
Not all implementations have been replaced, but this is a first step.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

### DIFF
--- a/ShellPkg/Application/Shell/ShellProtocol.c
+++ b/ShellPkg/Application/Shell/ShellProtocol.c
@@ -3473,30 +3473,6 @@ InternalEfiShellGetListAlias (
 }
 
 /**
-  Convert a null-terminated unicode string, in-place, to all lowercase.
-  Then return it.
-
-  @param  Str    The null-terminated string to be converted to all lowercase.
-
-  @return        The null-terminated string converted into all lowercase.
-**/
-CHAR16 *
-ToLower (
-  CHAR16  *Str
-  )
-{
-  UINTN  Index;
-
-  for (Index = 0; Str[Index] != L'\0'; Index++) {
-    if ((Str[Index] >= L'A') && (Str[Index] <= L'Z')) {
-      Str[Index] -= (CHAR16)(L'A' - L'a');
-    }
-  }
-
-  return Str;
-}
-
-/**
   This function returns the command associated with a alias or a list of all
   alias'.
 
@@ -3535,7 +3511,7 @@ EfiShellGetAlias (
       return NULL;
     }
 
-    ToLower (AliasLower);
+    StrnCharToLower (AliasLower, StrLen (AliasLower));
 
     if (Volatile == NULL) {
       GetVariable2 (AliasLower, &gShellAliasGuid, (VOID **)&AliasVal, NULL);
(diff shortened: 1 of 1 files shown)


MdePkg/Library/BaseLib/String.c
```diff
@@ -1682,3 +1682,366 @@ BcdToDecimal8 (
...
+
+  @retval TRUE            Unicode Char is a lowercase letter.
+  @retval FALSE           Otherwise.
+**/
+BOOLEAN
+CharIsLowerAlpha (
+  IN CHAR16  Char
+  )
+{
+  return (Char >= L'a') && (Char <= L'z');
```
leiflindholm, line 1922: While the higher-level idea of introducing these primitives is a good one, this implementation is spectacularly incorrect.

Unicode exists for a reason. There are no shortcuts of "map to ascii" that does not produce incorrect behaviour.

pierregondois: Hello Leif,
Just to be sure I understand correctly, the issue is that the Unicode functions are only covering the ascii characters and not the full range of Unicode characters ? Like 'é' should be detected as a lowercase characters ?

The patchset should not modify the behaviour of the pre-existing code, so I'm not sure I understand how wrong this is. If the issue is that all the Unicode functions could just be replaced by:
```
CharIsLowerAlpha(X) -> (UINT16)AsciiCharIsLowerAlpha((UINT8)X)
```
isn't there still some utility to the patchset ?

leiflindholm: I agree there is no change in behaviour at the point of (a theoretical) merge.

But there's a difference between individual modules incorporating ugly hacks for practical (if short-sighted) reasons, and adding primitives to BaseLib that from the outside look like useful unicode helpers but in reality discard any of the value of unicode.

pierregondois: Hello Leif,
Getting back to the thread after the holidays:
I'm ok to close the PR, but would it be possible to detail what you mean by `discard any of the value of unicode` ? I still don't understand what is wrong

kraxel: I think what you did clearly makes sense for the `*Num` and `*HexNum` functions.

But the answer to the question whenever a character is upper case or lower case is /way/ more complex for unicode characters, and looking at the ASCII range is rather incomplete as this would not return the correct answer for characters such as Ö and ö.

pierregondois: ok right understood :+1:

leiflindholm: Apologies, back from holidays but somehow managed to get myself locked out of github so have been unable to respond before now. But yes, what @kraxel said.

However... as he also says, the numerical ones could be worth keeping.
And the ones that operate directly on Ascii strings too. I confess I missed those since the very first commit made me see red.

And certainly it would be worth raising an issue on any remaining existing code that lazily discards unicodeness of strings. They should be rewritten to use AsciiChar or fixed to use unicode properly.

PR #11843, opened 2025-12-04
ArmPlatformPkg/PL011SerialPortLib: Add hardware flow control support

xxie24:
# Description

This change introduces hardware flow control capabilities for the PL011 serial port Library.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verified UART output in console. 

## Integration Instructions

N/A

### DIFF
--- a/ArmPlatformPkg/Library/PL011SerialPortLib/PL011SerialPortLib.c
+++ b/ArmPlatformPkg/Library/PL011SerialPortLib/PL011SerialPortLib.c
@@ -57,6 +57,16 @@ SerialPortInitialize (
     return Status;
   }
 
+  if (PcdGetBool (PcdSerialUseHardwareFlowControl)) {
+    Status = PL011UartSetControl (
+               (UINTN)PcdGet64 (PcdSerialRegisterBase),
+               EFI_SERIAL_HARDWARE_FLOW_CONTROL_ENABLE
+               );
+    if (EFI_ERROR (Status)) {
+      return Status;
+    }
+  }
+
   while (SerialPortPoll ()) {
     SerialPortRead (&Scratch, sizeof (Scratch));
   }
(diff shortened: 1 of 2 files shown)


leiflindholm:
Not that it's really fundamental to the change, but I note that the MdeModulePkg .dec/.uni use 16550-specific language for this Pcd. It would be nice to get those updated if more drivers use it.

ArmPlatformPkg/Library/PL011SerialPortLib/PL011SerialPortLib.c
```diff
@@ -57,6 +57,16 @@ SerialPortInitialize (
...
+  if (PcdGetBool (PcdSerialUseHardwareFlowControl)) {
+    Status = PL011UartSetControl (
+               (UINTN)PcdGet64 (PcdSerialRegisterBase),
+               EFI_SERIAL_HARDWARE_FLOW_CONTROL_ENABLE
+               );
+    if (EFI_ERROR (Status)) {
+      return Status;
+    }
+  }
+
```
ardbiesheuvel, line 69: What does this achieve by itself? Is this sufficient to get working h/w flow control on a PL011?

The NS16550 libraries appear to include additional handling for the RTS bit, but if we don't need that, great! But I'd like to understand whether that is the case here.

xxie24: This will set PL011's UARTCR's CTSEn and RTSEn bits which are required for enabling hw flow based on TX/RX FIFO watermark level: 
 * [CTSEn/RTSEn bits](https://developer.arm.com/documentation/ddi0183/g/programmers-model/register-descriptions/control-register--uartcr?lang=en)
 * [FIFO based flow control](https://developer.arm.com/documentation/ddi0183/g/functional-overview/uart-hardware-flow-control?lang=en)

16550 UART driver as a older hw controller did manual control of RTS pin, i.e. it only assert RTS only if driver is waiting for data.  While some 16550 controllers support automatically HW flow control,  I suspect the driver implement manual RTS/CTS pin control due to hw compatibility.

ardbiesheuvel (approved):

PR #11848, opened 2025-12-05
Fix stack overflow issue due to nested interrupts

gumingyun:
This is a heavily simplified version of the fix in the OvmfPkg
NestedInterruptTplLib.  Putting it in DXE core allows CoreRestoreTpl()
to lower the TPL while keeping interrupts disabled, removing the need
for either DisableInterruptsOnIret() or the complex deferred execution
mechanism.

Instead, CoreRaiseTpl() uses the current state of the interrupt flag to
second guess whether it's being called from an interrupt handler; when
restoring the outer TPL at the end of the handler, interrupts remain
disabled until IRET.  This eliminates the possibility that a nested
invocation of the interrupt handler has the same TPL as the outer one.


ardbiesheuvel:
What happened to the other PR?

ardbiesheuvel:
Please don't repost your changes in a new PR while leaving the old one open, and with unaddressed review comments.

ardbiesheuvel:
#11276

panyongg:
> Please don't repost your changes in a new PR while leaving the old one open, and with unaddressed review comments.

thanks @ardbiesheuvel. As i mentioned in https://github.com/tianocore/edk2/pull/11276, the old branch for PR11276 is OUT OF CONTROL, the owner does not work on it, and others cannot access the branch. That's why we submitte this new PR with same code changes. Agree to have new PR? thanks.

PR #11809, opened 2025-11-26
MdePkg/BaseLib: RISC-V: Add Unaligned support

wxjstz:
# Description

RISCV64 requires unaligned access, a requirement that general-purpose implementations cannot meet. RISCV64 needs to handle byte order issues during unaligned access. This patch resolves these issues.

- [ ] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [ ] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

## How This Was Tested

Build SiFive U540 and run it on qemu

## Integration Instructions

<N/A>

Issue #10125, opened 2021-08-09
[MdePkg/BaseLib] Unaligned APIs cannot be called safely (Bugzilla Bug 3542)


tianocore-issues:
This issue was created automatically with [bugzilla2github](https://github.com/makubacki/bugzilla2github/tree/bz2gh_tianocore)
# Bugzilla Bug 3542
Date: 2021-08-09T16:55:31+00:00
From: Marvin Häuser &lt;<mhaeuser>&gt;
To:   unassigned &lt;<unassigned>&gt;
CC:   @lgao4

Last updated: 2021-08-24T22:14:19+00:00

tianocore-issues:
## Comment 16841
Date: 2021-08-09 16:55:31 +0000
From: Marvin Häuser &lt;<mhaeuser>&gt;

- Industry Specification: ---
- Release Observed: EDK II Master
- Releases to Fix: EDK II Master
- Target OS: ---
- Bugzilla Assignee(s): unassigned &lt;<unassigned>&gt;

C prohibits not only dereferencing but also casting to unaligned
pointers [1]. GCC and Clang have an optional warning to check for such potential mistakes [2]. As the target pointers of the current BaseLib unaligned APIs already have the type of what is to be accessed unaligned, they cannot be called
safely.

The most compatible fix is to change the type of the target pointers to VOID *, as any pointer implicitly casts to VOID *, and VOID * implicitly casts to any pointer. As the unaligned APIs cannot make assumptions about the alignment of the target pointers, this cannot cause hidden increasing of the required alignment. Build failures of any consuming code are not expected due to the implicit casting rules of VOID *.

[1] &quot;A pointer to an object type may be converted to a pointer to a different object type. If the resulting pointer is not correctly aligned for the referenced type, the behavior is undefined.&quot;
C11, 6.3.2.3, 7.

[2] e.g. &quot;Cast from &apos;UINT8 *&apos; (aka &apos;unsigned char *&apos;) to &apos;UINT32 *&apos; (aka &apos;unsigned int *&apos;) increases required alignment from 1 to 4&quot;
can be enabled with &quot;-Wcast-align&quot;

tianocore-issues:
## Comment 17055
Date: 2021-08-24 22:14:19 +0000
From: @lgao4

Please update the comments from the edk2 mail list.

leiflindholm:
https://github.com/tianocore/edk2/pull/11853

PR #11853, opened 2025-12-05
Mdepkg baselib remove ub

ardbiesheuvel:
# Description

As Marvin reported in #10125 a while ago, the ReadUnaliged/WriteUnaligned APIs in BaseLib are defined in a way that relies on undefined behavior in C. Reading a quantity that appears misaligned in memory cannot be done via a typed pointer, as typed pointers can only point to quantities that appear aligned.

This is essentially why we have needed a special implementation for AArch64 (and previously ARM), and the same problem has come up with RISC-V as well (#11809).

Fix the API, as well as the generic implementation, to not rely on undefined behavior, and instead, communicate to the compiler in a portable manner that the untyped pointer argument points to quantity of a certain type that may appear misaligned in memory, by using `#pragma pack(1)` and a union type encapsulating all the variants.

On x86, the [codegen](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:2,endLineNumber:15,positionColumn:2,positionLineNumber:15,selectionStartColumn:2,selectionStartLineNumber:15,startColumn:2,startLineNumber:15),source:'%23pragma+pack(1)%0A%0Aunsigned+short+readunaligned(void+*p)%0A%7B%0A++++union+%7B%0A++++++++unsigned+short+i%3B%0A++++%7D+*i++%3D+p%3B%0A%0A++++return+i-%3Ei%3B%0A%7D%0A%0Aunsigned+short+readunaligned_broken(unsigned+short+*p)%0A%7B%0A++++return+*p%3B%0A%7D'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:38.59909797158007,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:vc_v19_latest_x86,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'1',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'+-O2',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86+msvc+v19.latest+(Editor+%231)',t:'0')),header:(),k:31.190505161854627,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:cclang2110,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:2,lang:___c,libs:!(),options:'-O2',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+clang+21.1.0+(Editor+%231)',t:'0')),k:30.210396866565294,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) does not change at all, which is expected given that the underlying hardware always tolerates misalignment.

On AArch64, the [codegen](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1AB9U8lJL6yAngGVG6AMKpaAVxYMQAJlIOAMngMmABy7gBGmMQSAKykAA6oCoS2DM5uHt7xickCAUGhLBFRXLGWmNYpQgRMxARp7p4%2BZRUCVTUEeSHhkTEW1bX1GU39HYFdhT0lAJQWqK7EyOwcAKReAMxxxEzALEwA1HFMyADWEFxTyxoAgpdXrgxJwEHoewoIJAR7xJhM6PdieCemHQEAAbqg8C8AFRxC7XZYAdgAQrc9mi9vcUntESjruj8RiHoDnq93rU9nhlmtcVd8YiACJ7KF4NFUxlxKk01Ho74EeYMCkAWip2Ep1NuDIl13ujxJbw%2BXx%2Bf2YtGJwOMYWIqGOjAgMrVL3l5JhcJuyO5aN5/KZHPF8IR9I4M1onGivE8HC0pFQnEcrzmC0w2PWPFIBE0TpmxxAawAHAA6aIATiTGiTCPTJVjGg0CP0nEk7oj3s4vAUIA0YYjMzgsCQaBYcTokXIlAbTfoURqLFBscFyGQBiMXi4XErNFoBEi5YgYWLYUCNQAnpxQwvmMQlwB5MLacrh7i8BtsQRbhi0Fee3hYMKuYCOMS0cuH0hYXZGcRX194b4VUGYZ8vUwVRylcKdV14QIpxdF9VU1ZdnCwYsCGIPAWAg0h/2IMJEkwelMHfYBVSMas%2BAMYAFAANTwTAAHctziRgMP4QQRDEdgpBkQRFBUdQv10Lh9EMEwzH0PAwnLSAZlQOIbAEZ89kFLcvF4VAsNQrBJIgGZmjkzwIAcQZPEE/wxgKIo9ASJI9KMyzsj0zpzMmCxMCsPS2gGFwGj0XTKhGRzumKPp2lswSFH8szAokHSA0WaL8w4N1SA9L0fQ4PZu17QV9kHYS9hHeMuHjDQ9ggXBCBIYM1nOXgDy0KYZgQJUem00ho2qwrU2iDRoi4AA2LhY2iSQ%2Bq8PMYMLUh0LHSsUtU0sLArKsrxrGBEBQVBG2bMgKAgdttpATK%2BwHIdgBHMc%2BDoKdiBnOcv3XZcMIezcdz3awMOPRgCDPC9ixvO8H1oJ8MLfYTPy9fBfxsf9AN4YDQPAl8oNc4s4K2TdEKWL0ULQjCsJwpR8MI4jQBWsjtiomj6MYj1QxY4RRHETj6Z4tRi10HxTpAUxjHMOCtOk2SUmfJSvEUlhwtQ6xstVJ5VPUyEAPgHTXP3FJ7AYJwvIyEzNYCiYgqsnJUm14ysmslJ9YssLVZaBgPLqU2fNt9yIvyKKwpGULgtqK3Jhi%2BY4vOBKkrmkt0qO7K9lyox8qKoqSrK/AiGIKqauW%2BqoxjDR4xHDQ1iTPrYz6vrolLrwk2kCaiy/NKyyWurIwSlTkuLOuM6brCkjsSQgA) depends on whether or not strict alignment is required, which is indicated by passing the `-mstrict-align` argument. This is needed, e.g., in SEC or PEI, where code execution may occur with the MMU and caches disabled, requiring strict alignment.

On RISC-V, which never tolerates misalignment, the [codegen](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1AB9U8lJL6yAngGVG6AMKpaAVxYMQAJlIOAMngMmABy7gBGmMQSAKykAA6oCoS2DM5uHt7xickCAUGhLBFRXLGWmNYpQgRMxARp7p4%2BZRUCVTUEeSHhkTEW1bX1GU39HYFdhT0lAJQWqK7EyOwcAKReAMxxxEzALEwA1HFMyADWEFxTyxoAgpdXrgxJwEHoewoIJAR7xJhM6PdieCemHQEAAbqg8C8AFRxC7XZYAdgAQrc9mi9vcUntESjruj8RiHoDnq93rU9nhlmtcVd8YiACJ7KF4NFUxlxKk01Ho74EeYMCkAWip2Ep1NuDIl13ujxJbw%2BXx%2Bf2YtGJwOMYWIqGOjAgMrVL3l5JhcJuyO5aN5/KZHPF8IR9I4M1onGivE8HC0pFQnEcrzmC0w2PWPFIBE0TpmxxAawAHAA6aIATiTGiTCPTJVjGg0CP0nEk7oj3s4vAUIA0YYjMzgsCQaBYcTokXIlAbTfoUWIoIAbJJBchgMhkCUvJWaLQCJFyxAwsWwoEagBPTihhfMYhLgDyYW05XD3F4DbYgi3DFoK89vCwYVcwEcYlo5cPpCwuyM4ivr7w3wqoMwz5epgqjlK4U6rrwgRTi6L6qpqy7OFgxYEMQeAsBBpD/sQYSJJg9KYO%2BwCqkY1Z8AYwAKAAangmAAO5bnEjAYfwggiGI7BSDIgiKCo6hfroXD6IYJhmPoeBhOWkAzKgcQ2AIz57IKW5eLwqBYahWCSRAMzNHJngQA4gyeIJ/hjAURR6AkSR6UZlnZHpnTmZMFiYFYeltAMLgNHoumVCMjndMUfTtLZgkKP5ZmBRIOkBos0X5hwbqkB6Xo%2BhwewAEoAJJCI4gqUXsEB9oKC4EAoUx7EOyB7CU8ZePGGiFbghAkMGaznLwB5aFMMwIEqPTaaQ0btfGXCptEGjRFwPZcLG0SSD2Xh5jBhakOhXA5slxZpWWFZVleNYwIgKCoI2zZkBQEDtudIDdsVw4GEYXhcBtfB0FOxAznOX7rsuGG/ZuO57tYGHHowBBnhexY3neD60E%2BGFvsJn5evgv42P%2BgG8MBoHgS%2BUGucWcFbJuiFLF6KFoRhWE4Uo%2BGEcRoAHWR2xUTR9GMR6oYscIojiJxPM8Woxa6D4j0icY5hwVp0mySkz5KSp3rqZCAHwDprn7ik9gME4XkZCZusBRMQVWTkqT68ZWTWSkxsWWFmstAwHl1JbPmO%2B5EX5FFYUjKFwW1HbkwxfMcXnAlSUpapnCZTleUFXdkhDnsyDi3sz2jQ1TX4EQxBtR1%2B3dVGMYaHVG1rEmPaxj2PbRLXXhJtIK1Fl%2BO0WHtXWRglStRyWHCddWMxYUkdiSEAA%3D) is as expected on both GCC and Clang.


wxjstz:
```c
#if defined (_MSC_EXTENSIONS)
#define MISALIGNED  __unaligned
#else
#define MISALIGNED  __attribute__((__aligned__(1)))
#endif
```
In my tests, it does not enable the compiler to automatically generate instructions that avoid triggering exceptions when it detects unaligned addresses.

MdePkg/Library/BaseLib/Unaligned.c
```diff
@@ -101,13 +123,20 @@ ReadUnaligned24 (
...
   ASSERT (Buffer != NULL);
 
-  *Buffer = BitFieldWrite32 (*Buffer, 0, 23, Value);
+  Misaligned = Buffer;
+
+  (*Misaligned)[0] = (UINT8)Value;
+  (*Misaligned)[1] = (UINT8)(Value >> 8);
+  (*Misaligned)[2] = (UINT8)(Value >> 16);
+
   return Value;
```
wxjstz: return Value & 0xffffff;

ardbiesheuvel:
> ```c
> #if defined (_MSC_EXTENSIONS)
> #define MISALIGNED  __unaligned
> #else
> #define MISALIGNED  __attribute__((__aligned__(1)))
> #endif
> ```
> 
> In my tests, it does not enable the compiler to automatically generate instructions that avoid triggering exceptions when it detects unaligned addresses.

Thanks for testing - turns out I should be using packed structs instead,

I have updated the branch, could you please give it another try? Thanks

ardbiesheuvel:
See [here](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1AB9U8lJL6yAngGVG6AMKpaAVxYMQAJlIOAMngMmABy7gBGmMTeGqQADqgKhLYMzm4e3nEJSQIBQaEsEVFeMZaY1slCBEzEBKnunj6l5QKV1QS5IeGR0RZVNXXpjX3tgZ0F3cUAlBaorsTI7BwApBoAgkteAMx4VADUWDRB6LsQxgCyQo7G2AAaACrYwUIAkgDyT5Mr61sHo7tnzyEqz8zwA4sFsAARXa7UyuZi0PDAI5fDabTC0JSon6YQ6Yf6A4FgiHQ2HGJgEAjEPBhVwETCmKCmWJMZAAa0w6FMk0%2BazRDh2qLW8MSyM5u1oAmAEqlu2ImCY6HhYiRRwgADdUHhjgAqWK89YAdgAQl8YTCFFTXNZdksTWbzY6RarxZKjDL3XglptTWtHXboQCgSDwVDdjq8DDvdDYt7fes/eb5QQ5gxdngALTe7Ben2ow2QoWrZ1i45u6XluUKpUIl1csLEVAchgQEtHD0V2V6g12%2BOO5Op8OxvN8gscaa0TgAVl4ng4WlIqE4jl2Clm83xaJ4pAImnH0zZIE2AE4AHSGgBskin18NhoAHBouFP7z5JxxJLO94vOLwFCAYl3edx1IOBYCQNAWFiOhInIShIOg%2BgomQZBaGMIIAHdyWIFhNi8YxMFUDDRj4Oh6WIf8IDCb8wkCaoAE9OG3WjmGIejXjCbQyiA7dILYQRXgYWhGOA0gsFpYBHDETEmN4LAWEMYBxFE/B5XKdVMH/UTCLKOlFm3QJ6XfbdEQbBjnCwb8qTwFhZNIDTiDCBJMEhTAFKMREjD3aYqAMYAFAANTwTAMNeWJGDs/hBBEMR2CkGRBEUFR1FE3QuH0RSQFMYxzFM/9IGmVBYhsAQtN2DNXi8cqAHVpN4VAHOpLB8ogaYmhKzwIAcAZPHS/xRnyQo9HiRIOp64asg6jpBu6dL2oqYZxrmjFuIWtppq6KI5sWlx6j0S11oGzaJDa9cFhO/Rpy/USlw4XZVgAJTOXYWAUdVkF2dUuDPKQTkIjNqueCFJhOXBCBIW0ti4SZeCArQeVIBBq26VrSEPTYuHPTY8MNKQvGfb6vDvS6P14WyuA0GI5wXW6/wAndvNAmBEBQVAoJgsgKAgBCOZAaoWHVe8MxQgwjHxinSNocjKOo0SWIYuz5bYjiuOsOy%2BMYAhBOE79xNcSTpK07d5MU5SF1U1aNK0hcdOQPS7MMjFv1M4hzIwRYF2s2zuF4BynKUVz3KUwJQGAny/MC4LQvCudtyi4RRHEeL46StRv10N9Muy3KaRawriuSLSKqqjNXus6wMxVZF6sa7VNPgNqVuaTrut29I%2BoYdANvGLbMlG5Ilr77IGG7oblqsDrWn6NveosJvJ%2BGUfZt6NpB4Omol62065nO6GSZnUhqfqzhdn5wXK92VDFN2fHT0xjRQfwIhiEhjGYYZsOD28e9TynDR8dfI%2BDQkgrz3mPNId8n5D7flphYemcN9wky8NdGmv4P7w2mA5RIdhJBAA) for a demonstration of the resulting codegen on VS and GCC/Clang

MdePkg/Library/BaseLib/Unaligned.c
```diff
@@ -8,6 +8,14 @@
 
 #include "BaseLibInternals.h"
 
+#ifdef MDE_CPU_IA32
```
leiflindholm: Is this actually needed?

ardbiesheuvel: Yes, as MSVC does not support `__unaligned` for that architecture.

But I'll drop this in any case, and rely on `#pragma pack(1)` like we do elsewhere.

leiflindholm: Lolol.
Yeah, it would be nice not to introduce new Ia32-specific code.

MdePkg/Library/BaseLib/Unaligned.c
```diff
@@ -24,12 +34,16 @@
...
   )
 {
+  CONST MISALIGNED  *Misaligned;
+
   ASSERT (Buffer != NULL);
 
-  return *Buffer;
+  Misaligned = Buffer;
+
+  return Misaligned->Val16;
```
wxjstz: Based on my testing, it generates `ldrh    w0, [x0]` on arm64 and `lhu    a0,0(a0)` on RISC-V. This does not prevent unaligned access exceptions.

ardbiesheuvel: It works fine [here](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:44,endLineNumber:13,positionColumn:44,positionLineNumber:13,selectionStartColumn:44,selectionStartLineNumber:13,startColumn:44,startLineNumber:13),source:'%0A%23pragma+pack(1)%0A%0Aunsigned+long+readunaligned(void+*p)%0A%7B%0A++++union+%7B%0A++++++++unsigned+long+i%3B%0A++++%7D+*i++%3D+p%3B%0A%0A++++return+i-%3Ei%3B%0A%7D%0A%0Aunsigned+long+readunaligned_broken(unsigned+long+*p)%0A%7B%0A++++return+*p%3B%0A%7D'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:39.76454577801582,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:rv64-cgcc1520,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'1',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'+-O2+-Wall',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+RISC-V+(64-bits)+gcc+15.2.0+(Editor+%231)',t:'0')),header:(),k:31.733271421519277,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:armv8-cclang2110,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:2,lang:___c,libs:!(),options:'-O2+-mstrict-align',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+armv8-a+clang+21.1.0+(Editor+%231)',t:'0')),k:28.502182800464894,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

and [here](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1AB9U8lJL6yAngGVG6AMKpaAVxYMQkgJykHAGTwGTAA5dwAjTGIJAGZSAAdUBUJbBmc3Dy9fROSbAUDgsJZI6K44y0xrVKECJmICdPdPHwtMKzyGGrqCAtCIqNiLWvrGzJaFYZ6gvuKBsoBKC1RXYmR2DgBSDQBBDYAmGPjiJmAWJgBqeKZkAGsILnmt3Z3XBmTgYPRzhQQSAnPiJgmOhXmI8B9MOgIAA3VB4L4AKnijx2GwA7AAhJ7nHHnV6pc7orE7XGkvFvcGfb6/ernPAbGLE7ak9EAEXOCLwOIZ7PiDKZ2NxgIIKwYdIAtAzsPTGU82XKXhSIV8fn8AUCQcxaJTIcZwsRUDdGBBXu8qaraUiUbtMYKccLRRy%2BbLUWjWRxFrROABWXieDhaUioTiOb7LVaYQkHHikAiaD2LG4gGLeAB0aIAbJJvdm0WiABwaLje/N7fScSR%2B%2BNBzi8BQgDSx%2BOLOCwJBoFjxOhRciUDtd%2BjRYjQrPi5DAZDIYt7Rs0WgEKL1iDhavhIJ1ACenBja%2BYxA3AHlwtpKnHuLwO2xBAeGLQtwHeFhwq5gI4xLR6%2BfSFgzkZxA/vzwQEqmhTBP0DTBVEqVwF23XgggXL0v21fVN2cLBqwIYg8BYODSFA4hwiSTBWUwX9gG1Ixmz4AxgAUAA1PBMAAdwPeJGDw/hBBEMR2CkGRBEUFR1AA3QuH0QwTDMfQ8HCetIEWVB4g6T9znFA89jUgB1d9eFQAjsKweSIEWCoqjsCAHFGTxxICaYihKPQchSARrKcpIXIYXoHLmVp2mqSY3PEsyOi6epvP6Uohm6ILovC%2BzIokUzwzWJLyw4X1SH9QNgw4c4ACUAEkhEccV6POCBRzXAgFHmc4J2Qc5i1TPZUw0CrcEIEgoxiB5eDPLR5kWBANQGEzSCTXr0xiGI9jRKQ9mLLhvDmtF0srUhcK4DRG2yvTawsBsmwfFsYEQFBUE7bsyAoCB%2B2ukA6hYaF8zHZADCMRbtr4OgF2IJcVwA3dNzw4H9yPE9rDwy9GAIG872rJ8XzfWgPzwn9JP/QN8GAmxQPA3hIOg2CvwQtpqxQ4593Q9ZAywnC8IIoilFI8jKNAE6aJOBimNY9j/RjLjhFEcR%2BKFoS1GrXQyw%2BqTjHMFDjMU5TUk/dTNPFFgJmw6xxTBD49IM%2BEwPgUy2lPVJ7AYJwXCaPQ7MKRLxOcjpYpd1IItmKKQoCmLbcyYLzfMzpJk9xzgsC/2bLiqZHa9tKFBSvjPR9KsANy84npevXzneyTzkW1MuDajr8CIYger647BsTEA9nzVNvQ0RbS0LDRJCzfNvGkJCNr2msODrI6BoTdK9jTnKDpHob8MXS3JCAA%3D)

mdkinney:
Why is `MISALIGNED` added when the coding standard already defines the keyword `UNALIGNED`

https://tianocore-docs.github.io/edk2-CCodingStandardsSpecification/draft/5_source_files/56_declarations_and_types.html#table-6-efi-data-types-slightly-modified-from-uefi-231

Also, there was a use case for read/write 24-bit values.  That is why these were added in the first place.  Are they doing any hard to leave?  If we have byte-packed structures with 3-byte values, this API is useful.

MdePkg/Library/BaseLib/Unaligned.c
mdkinney: Could this be simplified to the following and remove the additional of a local variable?

return (CONST MISALIGNED *)Buffer->Val16;

mdkinney:
> Why is `MISALIGNED` added when the coding standard already defines the keyword `UNALIGNED`
> 
> https://tianocore-docs.github.io/edk2-CCodingStandardsSpecification/draft/5_source_files/56_declarations_and_types.html#table-6-efi-data-types-slightly-modified-from-uefi-231
> 
> Also, there was a use case for read/write 24-bit values. That is why these were added in the first place. Are they doing any hard to leave? If we have byte-packed structures with 3-byte values, this API is useful.

Please ignore `MISALIGNED` comment.

I see that the concept of declaring a parameter as unaligned is compiler specific and really only applies to Itanium.  Using a packed structure does seem like the best ANSI C method to get the compiler to generate the correct instructions for all archs.

MdePkg/Library/BaseLib/Unaligned.c
ardbiesheuvel: We could do `return ((CONST MISALIGNED *)Buffer)->Val16;` (mind the parens) - I'll update the PR.

ardbiesheuvel:
> There was a use case for read/write 24-bit values. That is why these were added in the first place. Are they doing any harm to leave? If we have byte-packed structures with 3-byte values, this API is useful.

There are 2 problems with this API:
- It is unclear how to distinguish an aligned UINT24 from a misaligned UINT24. The C spec does not provide any guidance here AFAIK, and there are no underlying hardware constraints that could inform this. I suppose a UINT24 appearing at address `0x0` or `0x1000` would be considered aligned. But what about an UINT24 appearing at address `0x1001` or `0x1002`?
- The current implementation is flawed, as it reads and writes 4 bytes at a time. This means that accessing a UINT24 appearing at the end of a page would trigger a page fault if the next page is not mapped (or mapped read-only).

Given that the only user of this API (removed by this PR) didn't actually make good use of it, I don't think we should retain it, given the broken implementation and the ill specified concept of (mis)alignment. If there is a use case for an API that can read or write 3-byte arrays, we can always add that back, but we should avoid reasoning about it in terms of alignment.

mdkinney:
> > There was a use case for read/write 24-bit values. That is why these were added in the first place. Are they doing any harm to leave? If we have byte-packed structures with 3-byte values, this API is useful.
> 
> There are 2 problems with this API:
> 
> * It is unclear how to distinguish an aligned UINT24 from a misaligned UINT24. The C spec does not provide any guidance here AFAIK, and there are no underlying hardware constraints that could inform this. I suppose a UINT24 appearing at address `0x0` or `0x1000` would be considered aligned. But what about an UINT24 appearing at address `0x1001` or `0x1002`?
> * The current implementation is flawed, as it reads and writes 4 bytes at a time. This means that accessing a UINT24 appearing at the end of a page would trigger a page fault if the next page is not mapped (or mapped read-only).
> 
> Given that the only user of this API (removed by this PR) didn't actually make good use of it, I don't think we should retain it, given the broken implementation and the ill specified concept of (mis)alignment. If there is a use case for an API that can read or write 3-byte arrays, we can always add that back, but we should avoid reasoning about it in terms of alignment.

I suggest we fix the implementation.  I agree the implementation should never access contents outside the 24-bit (3-byte) window.  The current use of 32-bit APIs is not safe when the 32-bit access would cross a page boundary.  Using byte access for this API would be the safest implementation.  Depending on the alignment of the input, could always be broken up into a 8-bit and a 16-bit access.  

These APIs have been in MdePkg for a very long time and could be used in downstream.  As long as they are implementable, I prefer to not introduce a non-backwards compatible change.

ardbiesheuvel:
> I suggest we fix the implementation. I agree the implementation should never access contents outside the 24-bit (3-byte) window. The current use of 32-bit APIs is not safe when the 32-bit access would cross a page boundary. Using byte access for this API would be the safest implementation. Depending on the alignment of the input, could always be broken up into a 8-bit and a 16-bit access.
> 

Fair enough. But I will retain the patch to UefiLib that removes the `ReadUnaligned24()` calls, as it performs a cast from CHAR8* to UINT32* in the call to this API, and this is UB as well.

mdkinney:
With ReadUnaligned24() input parameter change from CONST UINT32 * to CONST VOID * to match all the other Read/WriteUnalignedxx() changes, I think UefiLib can be updated from:

```
  Name1 = ReadUnaligned24 ((CONST UINT32 *)Language1);
  Name2 = ReadUnaligned24 ((CONST UINT32 *)Language2);
```
to:

```
  Name1 = ReadUnaligned24 (Language1);
  Name2 = ReadUnaligned24 (Language2);
```


I will point out that I believe we recently did changes to drop use of ISO639-2 and consolidate to RFC4646 to follow UEFI Spec.  This UefiLib function and others are only for support of ISO639-2 3-character language codes.  If there is no need for these anymore, then I think a more general ISO639-2 clean up could be done as a independent PR to delete these entire functions.

ardbiesheuvel:
> With ReadUnaligned24() input parameter change from CONST UINT32 * to CONST VOID * to match all the other Read/WriteUnalignedxx() changes, I think UefiLib can be updated from:
> 
> ```
>   Name1 = ReadUnaligned24 ((CONST UINT32 *)Language1);
>   Name2 = ReadUnaligned24 ((CONST UINT32 *)Language2);
> ```
> 
> to:
> 
> ```
>   Name1 = ReadUnaligned24 (Language1);
>   Name2 = ReadUnaligned24 (Language2);
> ```
> 

The function in question compares two 3-bytes arrays that are both located in memory. What would be the point of retaining code that reinterprets those arrays as 24-bit unsigned integers, and then moves them into 32-bit variables in order to perform the comparison using integer comparison? Let alone the fact that we still haven't established what it means for a 24-bit integer to be misaligned. This should just use `CompareMem ()`.

IMO, this API should not exist at all, and we should definitely stop using it to load char arrays from memory.

MdePkg/Library/BaseLib/Unaligned.c
```diff
@@ -101,13 +111,14 @@ ReadUnaligned24 (
...
+  OUT VOID    *Buffer,
   IN  UINT32  Value
   )
 {
   ASSERT (Buffer != NULL);
 
-  *Buffer = BitFieldWrite32 (*Buffer, 0, 23, Value);
+  WriteUnaligned16 (Buffer, (UINT16)Value);
+  ((UINT8 *)Buffer)[2] = (UINT8)(Value >> 16);
   return Value;
```
wxjstz: return Value & 0xffffff;

PR #11857, opened 2025-12-07
Global: Update CI from GCC5 to GCC

mikebeaton:
# Description

Update CI from GCC5 to GCC.

I am not sure if there was any good reason why the apparently (from their top level names in the CI) GCC CI steps were really running in GCC5, which I understand is deprecated? In case there wasn't, or in case there was but it no longer applies, I am suggesting this change.

As can be seen, all the CI build and run steps here have succeeded.

In case it is relevant, we have also been running CI of a regularly rebased fork of EDK2 in GCC (plus other toolchains) for some time with no problems.

- [ ] Breaking change?
  - NO.
- [ ] Impacts security?
  - NO.
- [ ] Includes tests?
  - NO.

## How This Was Tested

CI.

(Plus manual testing of OvmfPkg build.sh change from commit 2, although this overlaps with what CI already tests in commit 1.)

## Integration Instructions

N/A


mikebeaton:
@os-d @makubacki - Is this of interest?

ardbiesheuvel:
> @os-d @makubacki - Is this of interest?

Yes - the reason GCC (and GCCNOLTO) were created was to be able to phase out GCC49 and GCC5, as those compiler versions are ancient now, and changing the toolchain name every time the version is bumped is not worth the hassle (CLANGDWARF and CLANGPDB have no need for it either, although we may periodically have to update the minimum supported version associated with each)

mikebeaton:
This may be quite uncontroversial in the end, then.

@leiflindholm @ardbiesheuvel - How does this look for approval from the ArmPlatformPkg/ArmVirtPkg side of things?

BaseTools/Edk2ToolsBuild.py
makubacki, line 1: Was it intentional to leave the "GCC5" environment variables like `GCC5_AARCH64_PREFIX` and `GCC5_AARCH64_INSTALL` used later in this file versus `GCC_AARCH64_PREFIX` and `GCC_AARCH64_INSTALL`?

https://github.com/tianocore/edk2/blob/00e404320727af83a40add192646d27cf2d9d8d9/BaseTools/Edk2ToolsBuild.py#L253-L266

makubacki: It looks like the container has [both set](https://github.com/tianocore/containers/blob/b5628216f41680893021676cfada2bda02a18428/Fedora-43/Dockerfile#L83-L91).

At the minimum, I think `GCC_AARCH64_PREFIX` should at least be considered in this file in addition to `GCC5_AARCH64_PREFIX`.

leiflindholm: I'd say we'd want to change to GCC only.

mikebeaton: Okay, I've changed to accessing the GCC variables and saying in the comment that we support GCC (not GCC5). I've moved the changes to Edk2ToolsBuild.py into the first commit now.

PR #11866, opened 2025-12-08
OvmfPkg/PlatformInitLib: reserve igvm parameter area

kraxel:
# Description

Reserve igvm parameter area to make sure the OS does not use it.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

N/A

## Integration Instructions

N/A


bssrikanth:
Since the issue addressed by this PR is also present in the latest edk2-stable202511 tag, could we get the latest stable tag updated with this fix (e.g., by creating an edk2-stable202511.1 tag)?

PR #11870, opened 2025-12-10
Add ExStatusCodeHandler modulue

SharingYang:
# Description
Add ExStatusCodeHandler modulue to as an extentions StatusCodeHandler to support the following features:
1.Support print with multiple threads.
2.Support PcdMigrateTemporaryRamFirmwareVolumes.
3.Control Serial Status Code by Debug Print Error Level.

PR #11871, opened 2025-12-10
MdeModulePkg/Universal/StatusCodeHandler:Extend StatusCodeHandlerPei

SharingYang:
# Description
Extend StatusCodeHandlerPei to support the following features:
1.Support print with multiple threads.
2.Support PcdMigrateTemporaryRamFirmwareVolumes.
3.Control Serial Status Code by Debug Print Error Level.

### DIFF
--- a/EmulatorPkg/EmulatorPkg.dsc
+++ b/EmulatorPkg/EmulatorPkg.dsc
@@ -131,6 +131,7 @@
   IntrinsicLib|CryptoPkg/Library/IntrinsicLib/IntrinsicLib.inf
   OpensslLib|CryptoPkg/Library/OpensslLib/OpensslLibCrypto.inf
   BaseCryptLib|CryptoPkg/Library/BaseCryptLib/BaseCryptLib.inf
+  HardwareSemaphoreLib|MdeModulePkg/Library/HardwareSemaphoreLibNull/HardwareSemaphoreLibNull.inf
 
 !if $(SECURE_BOOT_ENABLE) == TRUE
   PlatformSecureLib|SecurityPkg/Library/PlatformSecureLibNull/PlatformSecureLibNull.inf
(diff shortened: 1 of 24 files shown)


os-d:
Edk2 style is to have at minimum one commit per package changed (e.g. one commit for MdeModulePkg, one for EmulatorPkg, etc.). That will make PatchCheck pass. Because the changeset is large, I would consider multiple commits for MdeModulePkg, breaking up library inclusion, etc.

This change seems rather large with having to add two new library classes to add a small amount of information to the status code handler, which is in general not incredibly useful (there are some scenarios to be sure, but far and away regular debug prints over serial are more useful). I suspect it may be possible to accomplish what you want to do with a smaller set of changes. I will defer to the MdeModulePkg maintainers on whether this change is warranted and I’ll review for EmulatorPkg after that.

SharingYang:
> Edk2 style is to have at minimum one commit per package changed (e.g. one commit for MdeModulePkg, one for EmulatorPkg, etc.). That will make PatchCheck pass. Because the changeset is large, I would consider multiple commits for MdeModulePkg, breaking up library inclusion, etc.
> 
> This change seems rather large with having to add two new library classes to add a small amount of information to the status code handler, which is in general not incredibly useful (there are some scenarios to be sure, but far and away regular debug prints over serial are more useful). I suspect it may be possible to accomplish what you want to do with a smaller set of changes. I will defer to the MdeModulePkg maintainers on whether this change is warranted and I’ll review for EmulatorPkg after that.



> Edk2 style is to have at minimum one commit per package changed (e.g. one commit for MdeModulePkg, one for EmulatorPkg, etc.). That will make PatchCheck pass. Because the changeset is large, I would consider multiple commits for MdeModulePkg, breaking up library inclusion, etc.
> 
> This change seems rather large with having to add two new library classes to add a small amount of information to the status code handler, which is in general not incredibly useful (there are some scenarios to be sure, but far and away regular debug prints over serial are more useful). I suspect it may be possible to accomplish what you want to do with a smaller set of changes. I will defer to the MdeModulePkg maintainers on whether this change is warranted and I’ll review for EmulatorPkg after that.

Thanks, Will follow your advices if we decide to proceed with this plan。
We are evaluating the altenate solution as https://github.com/tianocore/edk2/pull/11870

PR #11873, opened 2025-12-10
OvmfPkg/X86QemuLoadImageLib: flip default for EnableLegacyLoader to false

kraxel:
# Description

What happened since commit 1549bf11cc94 ("OvmfPkg/X86QemuLoadImageLib:
make legacy loader configurable.") ?

First, qemu 10.0 has been released, which brings support for the -shim
command line option so direct kernel boot with secure boot works.

Second, support has been added to libvirt (version v11.2.0 and newer).

Third, we got a bunch of linux distro releases.  Latest debian, ubuntu
and fedora releases all have new enough edk2+qemu+libvirt packages to
support direct kernel boot with shim.efi loading and proper secure boot
verification.

Lastly, the edk2 security advisory GHSA-6pp6-cm5h-86g5 and CVE-2025-2296
have been published.

Time for the next step in tightening the screws:  Flip the default for
the EnableLegacyLoader config option from true to false.

The documentation for the config option is here:
https://github.com/tianocore/edk2/blob/master/OvmfPkg/RUNTIME_CONFIG.md#user-content-security-optorgtianocoreenablelegacyloader

Upcoming final step, in a year or two: remove the legacy loader from the
code base (drop X86QemuLoadImageLib, migrade all users to use
GenericQemuLoadImageLib instead).

- [x] Breaking change?
- [x] Impacts security?
- [ ] Includes tests?

## How This Was Tested

N/A

## Integration Instructions

* see https://github.com/tianocore/edk2/blob/master/OvmfPkg/RUNTIME_CONFIG.md#user-content-security-optorgtianocoreenablelegacyloader


mikebeaton:
Hi - Would we not want to update exactly the linked documentation (as regards default value, enable, disable, etc.) in the same commit?

kraxel:
> Hi - Would we not want to update exactly the linked documentation (as regards default value, enable, disable, etc.) in the same commit?

Patch updated.

ardbiesheuvel (approved):
Happy to see that this is being phased out.

Would it be possible for OVMF to only permit the legacy loader if secure boot is disabled? Or is is already doing that? Or do we need to keep that functionality for the older installers.

kraxel:
> Happy to see that this is being phased out.
> 
> Would it be possible for OVMF to only permit the legacy loader if secure boot is disabled? Or is is already doing that? Or do we need to keep that functionality for the older installers.

Ironically the legacy loader is /more/ important for the secure boot case.  qemu versions older than 10.0 provide the (x86) linux kernel in two separate chunks, the setup header + the rest of the kernel, with the setup header patched.  The patching breaks the secure boot signature, with the effect that the efi stub boot path works fine with secure boot turned off but fails with secure boot turned on, even if you add the distro signing certificate to 'db' so you do not depend on shim loading for kernel signature checking.

PR #11876, opened 2025-12-10
OvmfPkg/QemuFwCfgLib: do not log cache lookups

kraxel:
# Description

Logging each and every successful cache lookup is not very useful and
floods the log.

The much better approach is to log everything which gets added to the
cache, so the logfile can tell you what is in the cache with /much/ less
log lines.  The cache initialization code already does that.  So drop
the lookup message without any replacement.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Boot OVMF in qemu, check debug log.

## Integration Instructions

N/A


mikebeaton:
Sorry to ask a probably stupid question, but this _is_ the verbose log - are we sure it needs removing even in that case?

PR #11861, opened 2025-12-08
Global: Demonstrate/Test CLANGDWARF CI

mikebeaton:
# Description

THIS IS A TEST PR - DO NOT MERGE

- [ ] Breaking change?
  - NO.
- [ ] Impacts security?
  - NO.
- [ ] Includes tests?
  - NO.

## How This Was Tested

CI

## Integration Instructions

N/A

PR #11877, opened 2025-12-10
RedfishPkg: Fix CLANGPDB _fltused build error

mikebeaton:
# Description

Without this fix CLANGPDB X64 and IA32 builds of RedfishPkg fail with:

`lld-link: error: undefined symbol: _fltused`

Fixes: https://github.com/tianocore/edk2/issues/11881

- [ ] Breaking change?
  - NO.
- [ ] Impacts security?
  - NO.
- [ ] Includes tests?
  - NO.

## How This Was Tested

CI - incl. confirming that all other toolchains do still build with this change (confirmed for CLANGDWARF as well as other toolchains currently in CI)

## Integration Instructions

N/A

### DIFF
--- a/ArmVirtPkg/ArmVirt.dsc.inc
+++ b/ArmVirtPkg/ArmVirt.dsc.inc
@@ -17,12 +17,8 @@
 
   DEFINE DEBUG_PRINT_ERROR_LEVEL = 0x8000004F
 
-# Dynamic stack cookies pushes the FD size up slightly over 2MB
-!if $(TARGET) != NOOPT
-  DEFINE FD_SIZE_IN_MB    = 2
-!else
-  DEFINE FD_SIZE_IN_MB    = 3
-!endif
+# Various builds now exceed 2MB so choose 3MB as the default.
+DEFINE FD_SIZE_IN_MB    = 3
 
 !if $(FD_SIZE_IN_MB) == 2
   DEFINE FD_SIZE          = 0x200000
--- a/UefiPayloadPkg/Library/FdtParserLib/FdtParserLib.c
+++ b/UefiPayloadPkg/Library/FdtParserLib/FdtParserLib.c
@@ -841,21 +841,9 @@ ParsePciRootBridge (
       UINT32  TripletCells = ChildAddrCells + ParentAddrCells + SizeCells;
 
       for (Base = 0; Base < TempLen / sizeof (UINT32); Base = Base + TripletCells) {
-        UINT64  ChildBusAddress  = 0;
         UINT64  ParentBusAddress = 0;
         UINT64  DmaRangeSize     = 0;
-
-        if (ChildAddrCells == 3) {
-          ChildBusAddress = (UINT64)Fdt32ToCpu (*(Data32 + Base + 1)) << 32 |
-                            (UINT64)Fdt32ToCpu (*(Data32 + Base + 2));
-        } else if (ChildAddrCells == 2) {
-          ChildBusAddress = (UINT64)Fdt32ToCpu (*(Data32 + Base)) << 32 |
-                            (UINT64)Fdt32ToCpu (*(Data32 + Base + 1));
-        } else if (ChildAddrCells == 1) {
-          ChildBusAddress = Fdt32ToCpu (*(Data32 + Base));
-        }
-
-        UINT32  ParentBase = Base + ChildAddrCells;
+        UINT32  ParentBase       = Base + ChildAddrCells;
         if (ParentAddrCells == 2) {
           ParentBusAddress = (UINT64)Fdt32ToCpu (*(Data32 + ParentBase)) << 32 |
                              (UINT64)Fdt32ToCpu (*(Data32 + ParentBase + 1));
--- a/UefiPayloadPkg/UefiPayloadEntry/X64/DxeLoadFuncFit.c
+++ b/UefiPayloadPkg/UefiPayloadEntry/X64/DxeLoadFuncFit.c
@@ -21,6 +21,29 @@
 
 extern VOID  *mHobList;
 
+//
+// Global Descriptor Table (GDT) - 8-byte aligned
+// GDT with 3 entries: NULL, 64-bit Code, 64-bit Data
+//
+GLOBAL_REMOVE_IF_UNREFERENCED IA32_GDT  gGdtEntries[3] = {
+  // NULL descriptor (entry 0)
+  { .Uint64 = 0                     },
+  // .short 0xffff           /* limit 15:00 */
+  // .short 0x0000           /* base 15:00 */
+  // .byte  0x00             /* base 23:16 */
+  // .byte  0b10011010       /* P(1) DPL(00) S(1) 1 C(0) R(1) A(0) */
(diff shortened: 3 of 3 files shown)


mikebeaton:
Original `_fltused` was added (apparently only for Windows; which does also still build with this revised version - I found some other fixes where it didn't) in https://github.com/tianocore/edk2/commit/739a506b18c4f694b8d5d2f3424a329c45d737ba - cc @leiflindholm

leiflindholm:
If we're using the macro in more than one package, it feels like we should move it to MdePkg, probably Base.h, instead of duplicating it.

Additionally, is there some corresponding VS option for this? If so, we could maybe also get rid of the clunky "add a fake accessor function" pattern in .aslc files?

RedfishPkg/RedfishCredentialDxe/RedfishCredentialDxe.c
```diff
@@ -13,6 +13,8 @@
 
 #define REDFISH_VERSION_DEFAULT_STRING  L"v1"
 
+extern int  _fltused;
```
mdkinney: I do not think this is required.  No code actually references this global.  It is intrinsic from compiler

mdkinney:
> If we're using the macro in more than one package, it feels like we should move it to MdePkg, probably Base.h, instead of duplicating it.
> 
> Additionally, is there some corresponding VS option for this? If so, we could maybe also get rid of the clunky "add a fake accessor function" pattern in .aslc files?

Good idea to move this to Base.h.  But this is really part of the intrinsic lib that duplicated content in multiple submodule libs.  Should really address that as a common change instead of one piece as a time. 

I am not aware of any VS option for this.

RedfishPkg/RedfishCredentialDxe/RedfishCredentialDxe.c
mikebeaton: I had convinced myself this was needed - but on rechecking it isn't. Not enough sleep, I clearly should have triple-checked. :-( Will remove.

mikebeaton:
Moved to Base.h.

Can I be excused the task of unifying IntrinsicLib across all packages for now? ~Note that the issue I have now opened would remain open after this fix: https://github.com/tianocore/edk2/issues/11881#issuecomment-3640427647~ [See below.]

My specific aim here is to get everything which can be built in [CLANGPDB](https://github.com/tianocore/edk2/pull/11860/commits) or [CLANGDWARF](https://github.com/tianocore/edk2/pull/11861/commits) CI building. (CI never builds NOOPT of any packages - except for using NOOPT to build test suites, I believe.)

mikebeaton:
> Can I be excused the task of unifying IntrinsicLib across all packages for now?

Btw https://github.com/tianocore/edk2/pull/6169 :-)

> Note that the issue I have now opened would remain open after this fix: [#11881 (comment)](https://github.com/tianocore/edk2/issues/11881#issuecomment-3640427647)

This quoted comment is now not quite correct: I have split the above into two issues, so that this PR can potentially fix and close what I originally found https://github.com/tianocore/edk2/issues/11881, leaving behind the further problem I have now found in https://github.com/tianocore/edk2/issues/11882, as an open issue which (pending further investigation) does indeed seem to be something which merits revisiting IntrinsicLib.

mdkinney:
I do not prefer updating Base.h for this issue.  I am ok with the duplication at this time.  Before moving to Base.h, I want to make sure we define appropriate names for these types of macros at that scope and that can be handled in a different issue/pr for the intrinsics.

mdkinney:
The original version of the PR that added GLOBAL_USED to RedfishPkg is the only change needed for this PR.  All other topics can be deferred.  I do not think the extra extern statements in the original PR are required.  Please test again without those.

mikebeaton:
> The original version of the PR that added GLOBAL_USED to RedfishPkg is the only change needed for this PR. All other topics can be deferred. I do not think the extra extern statements in the original PR are required. Please test again without those.

You are right. I am double-checking that a revised fix actually works in my CLANGPDB CI draft PR before pushing again here.

RedfishPkg/PrivateLibrary/RedfishCrtLib/RedfishCrtLib.c
```diff
@@ -18,8 +18,13 @@
 int   errno            = 0;
 char  errnum_message[] = "We don't support to map errnum to the error message on edk2 Redfish\n";
 
-// This is required to keep VC++ happy if you use floating-point
-int  _fltused = 1;
+// This is required to keep MSVC ABI toolchains happy if you use floating-point.
```
mikebeaton: Is this revised comment better (for the code as it is now)? And better than just removing it?

Just thought I'd flag up to get another pair of eyes specifically on this suggested changed wording.

mikebeaton: I think MSVC-mode should be fine(?!). Resolving.

mdkinney (approved):

mikebeaton:
@mdkinney - Many thanks. Is there any chance this could have push label added?

I'm hoping to work my draft/demo CLANG CI PRs into one combined live PR in the next couple of days, and this present PR is currently (pending any newly introduced clang-only problems while there's no clang CI...) the only remaining precursor PR that it would be nice to have in master before that, to keep any such CI PR clean and only requiring discussion of what's really in it.

mdkinney:
I am not RedFish maintainer.

@AbnerChang Can you please review and set push label

mdkinney:
> @mdkinney - Many thanks. Is there any chance this could have push label added?
> 
> I'm hoping to work my draft/demo CLANG CI PRs into one combined live PR in the next couple of days, and this present PR is currently (pending any newly introduced clang-only problems while there's no clang CI...) the only remaining precursor PR that it would be nice to have in master before that, to keep any such CI PR clean and only requiring discussion of what's really in it.

I also have large number of CLANG changes queued up and a POC of staged CI using GitHub actions for CLANG CI coverage.  My focus is on IA32/X64 and it looks like you are adding AARCH64.  We should combine out efforts.

changab:
> I am not RedFish maintainer.
> 
> @AbnerChang Can you please review and set push label

Done!

mikebeaton:
> I also have large number of CLANG changes queued up and a POC of staged CI using GitHub actions for CLANG CI coverage. My focus is on IA32/X64 and it looks like you are adding AARCH64. We should combine out efforts.

@mdkinney - I couldn't quite bear _not_ to [put my version up](https://github.com/tianocore/edk2/pull/11886), when I've been looking at it on and off for a year, and working on it more intensively this last month - and was literally hours from finishing it when I found out the above. :-/

It seems like the two approaches cover basically the same ground, only I:
 - 'only' fixed what was needed to get everything to build which should currently build
 - worked in Azure Pipelines

Whereas it sounds as though you:
 - have multiple other clang related fixes and improvements too 
 - worked in GitHub actions

I'm not too sure what I could contribute to your work (though if I can, I'm happy to), other than the fixes for existing things that don't build in clang, which I've been contributing back as I go. The staged-ness of my PR isn't mine, it's just what stuart (plus existing pipelines yaml, which I just copied and lightly edited where required) does already.

PR #11879, opened 2025-12-11
ArmVirtPkg: Increase firmware size

mikebeaton:
# Description

Although almost all tool chain plus package combinations currently stay under the 2MB firmware size, except for NOOPT builds, ArmVirtQemu DEBUG built with CLANGDWARF now sneaks over.

Noting that images will be padded to 64MB for before use anyway, we now choose 3MB as the default for all. But keep the 2MB vs. 3MB code which checks FD_SIZE_IN_MB, in this and other files, available for reference the next time a size change is needed.

- [ ] Breaking change?
  - NO.
- [ ] Impacts security?
  - NO.
- [ ] Includes tests?
  - NO.

## How This Was Tested

CI

## Integration Instructions

N/A

### DIFF
--- a/ArmVirtPkg/ArmVirt.dsc.inc
+++ b/ArmVirtPkg/ArmVirt.dsc.inc
@@ -17,12 +17,8 @@
 
   DEFINE DEBUG_PRINT_ERROR_LEVEL = 0x8000004F
 
-# Dynamic stack cookies pushes the FD size up slightly over 2MB
-!if $(TARGET) != NOOPT
-  DEFINE FD_SIZE_IN_MB    = 2
-!else
-  DEFINE FD_SIZE_IN_MB    = 3
-!endif
+# Various builds now exceed 2MB so choose 3MB as the default.
+DEFINE FD_SIZE_IN_MB    = 3
 
 !if $(FD_SIZE_IN_MB) == 2
   DEFINE FD_SIZE          = 0x200000


ardbiesheuvel:
No objections from me, given that `-bios` accepts images of any size, and `-pflash` requires padding to 64 MiB anyway. But perhaps @kraxel has more insights so I will let him chime in as well.

kraxel:
> But perhaps @kraxel has more insights so I will let him chime in as well.

No objections to the size increase.  I'm wondering though whenever it makes sense to have all those conditions.  The NOOPT builds are known to be significantly larger, so that makes sense.  But if normal builds start to cross the line too I'd just increase the size for everybody.

mikebeaton:
> No objections to the size increase. I'm wondering though whenever it makes sense to have all those conditions. The NOOPT builds are known to be significantly larger, so that makes sense. But if normal builds start to cross the line too I'd just increase the size for everybody.

It depends what counts as 'normal', I guess all non-NOOPT, by this line of reasoning! I was indeed wondering about the excess of conditions too.

 - All DEBUG CLANGDWARF would be somewhat lesser.
 - All DEBUG even more so (so then we'd keep the lower size for RELEASE, and apply the larger for everything else). I'm guessing this might be what you'd agree to go for?

I think upping it for RELEASE is a bigger step, and possibly not needed at this stage?

kraxel:
> It depends what counts as 'normal', I guess all non-NOOPT, by this line of reasoning!

Yes.

> I think upping it for RELEASE is a bigger step,

Not really.
 * On x86 the pflash size equals firmware image size, so changing the size does cause compatibility problems because the guest address space layout changes.  Thats why we have those size config options there.
 * On arm pflash size is fixed at 64M and the images are padded (as already mentioned by @ardbiesheuvel).  Thanks to that changing the size of the firmware images isn't a problem at all.

mikebeaton:
> > It depends what counts as 'normal', I guess all non-NOOPT, by this line of reasoning!
> 
> Yes.
> 
> > I think upping it for RELEASE is a bigger step,
> 
> Not really.
> 
> * On x86 the pflash size equals firmware image size, so changing the size does cause compatibility problems because the guest address space layout changes.  Thats why we have those size config options there.
> * On arm pflash size is fixed at 64M and the images are padded (as already mentioned by @ardbiesheuvel).  Thanks to that changing the size of the firmware images isn't a problem at all.

On looking at this further, all the following files have code for 2MB and 3MB size (i.e. conditional on the value of `FD_SIZE_IN_MB`):
```
ArmVirtPkg/ArmVirt.dsc.inc
ArmVirtPkg/ArmVirtCloudHv.fdf
ArmVirtPkg/ArmVirtQemu.fdf
ArmVirtPkg/ArmVirtQemuKernel.fdf
ArmVirtPkg/ArmVirtXen.fdf
```
if we're never using one of the two versions, should all the code for the 2MB size just go? (Again, that seems like a bigger step ... but the code would AFIACT become completely unused.)

kraxel:
> if we're never using one of the two versions, should all the code for the 2MB size just go?

I think I'd tend to leave that code in, because it helps finding all the places you have to touch if the firmware images size changes.  There will be such events sooner or later, the next one probably being the NOOPT builds becoming larger than 3M.

mikebeaton:
@kraxel - Is there any merit at all in the observation that we have got more chance of avoiding bit rot in the code for two firmware sizes if we keep both sizes active (perhaps in this and in future updates)? Suggesting - to me - a reason still to consider switching to using 2MB for RELEASE for now, and 3MB for everything else.

mikebeaton:
The red X is not a code error in the revised commit, it is a timeout in a CI step. (In case anyone could trigger a re-run? :-) )

ardbiesheuvel (approved):

PR #11880, opened 2025-12-11
DynamicTablesPkg: Add support for MSCT table generation

abdattar:
# Description
    DynamicTablesPkg: Add ACPI MSCT table generator

    Introduces an ACPI MSCT (Maximum System Characteristics Table)
    generator library that retrieves the following information from the
    configuration manager:
     - Maximum physical address
     - Processor proximity information
     - Memory proximity information
     - Clock proximity information

    The generator calculates:
     - Maximum memory length per memory domain
     - Maximum contiguous processor domain
     - Maximum memory and processor domain combination

Add MSCT table ID to the ESTD_ACPI_TABLE_ID enumeration.
Add MSCT maximum physical address configuration object and
    update parser to handle the new object type.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on AMD platform

## Integration Instructions

N/A

### DIFF
--- a/DynamicTablesPkg/DynamicTables.dsc.inc
+++ b/DynamicTablesPkg/DynamicTables.dsc.inc
@@ -2,7 +2,7 @@
 #  Dsc include file for Dynamic Tables Framework.
 #
 #  Copyright (c) 2017 - 2026, Arm Limited. All rights reserved.<BR>
-#  Copyright (C) 2024 - 2025 Advanced Micro Devices, Inc. All rights reserved.<BR>
+#  Copyright (C) 2024 - 2026 Advanced Micro Devices, Inc. All rights reserved.<BR>
 #  Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.<BR>
 #
 #  SPDX-License-Identifier: BSD-2-Clause-Patent
@@ -41,6 +41,7 @@
   DynamicTablesPkg/Library/Acpi/Common/AcpiFadtLib/AcpiFadtLib.inf
   DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/AcpiHmatLib.inf
   DynamicTablesPkg/Library/Acpi/Common/AcpiMcfgLib/AcpiMcfgLib.inf
+  DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/AcpiMsctLib.inf
   DynamicTablesPkg/Library/Acpi/Common/AcpiPcctLib/AcpiPcctLib.inf
   DynamicTablesPkg/Library/Acpi/Common/AcpiPpttLib/AcpiPpttLib.inf
   DynamicTablesPkg/Library/Acpi/Common/AcpiRawLib/AcpiRawLib.inf
@@ -105,6 +106,7 @@
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiCedtLib/AcpiCedtLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiFadtLib/AcpiFadtLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiMcfgLib/AcpiMcfgLib.inf
+      NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/AcpiMsctLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiSlitLib/AcpiSlitLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiSpcrLib/AcpiSpcrLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiSpmiLib/AcpiSpmiLib.inf
@@ -148,6 +150,7 @@
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiFadtLib/AcpiFadtLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/AcpiHmatLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiMcfgLib/AcpiMcfgLib.inf
+      NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/AcpiMsctLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiPcctLib/AcpiPcctLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiPpttLib/AcpiPpttLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiRawLib/AcpiRawLib.inf
--- a/DynamicTablesPkg/DynamicTables.dsc.inc
+++ b/DynamicTablesPkg/DynamicTables.dsc.inc
@@ -2,7 +2,7 @@
 #  Dsc include file for Dynamic Tables Framework.
 #
 #  Copyright (c) 2017 - 2026, Arm Limited. All rights reserved.<BR>
-#  Copyright (C) 2024 - 2025 Advanced Micro Devices, Inc. All rights reserved.<BR>
+#  Copyright (C) 2024 - 2026 Advanced Micro Devices, Inc. All rights reserved.<BR>
 #  Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.<BR>
 #
 #  SPDX-License-Identifier: BSD-2-Clause-Patent
@@ -41,6 +41,7 @@
   DynamicTablesPkg/Library/Acpi/Common/AcpiFadtLib/AcpiFadtLib.inf
   DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/AcpiHmatLib.inf
   DynamicTablesPkg/Library/Acpi/Common/AcpiMcfgLib/AcpiMcfgLib.inf
+  DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/AcpiMsctLib.inf
   DynamicTablesPkg/Library/Acpi/Common/AcpiPcctLib/AcpiPcctLib.inf
   DynamicTablesPkg/Library/Acpi/Common/AcpiPpttLib/AcpiPpttLib.inf
   DynamicTablesPkg/Library/Acpi/Common/AcpiRawLib/AcpiRawLib.inf
@@ -105,6 +106,7 @@
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiCedtLib/AcpiCedtLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiFadtLib/AcpiFadtLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiMcfgLib/AcpiMcfgLib.inf
+      NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/AcpiMsctLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiSlitLib/AcpiSlitLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiSpcrLib/AcpiSpcrLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiSpmiLib/AcpiSpmiLib.inf
@@ -149,6 +151,7 @@
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiFadtLib/AcpiFadtLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/AcpiHmatLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiMcfgLib/AcpiMcfgLib.inf
+      NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/AcpiMsctLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiPcctLib/AcpiPcctLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiPpttLib/AcpiPpttLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiRawLib/AcpiRawLib.inf
--- a/DynamicTablesPkg/Include/AcpiTableGenerator.h
+++ b/DynamicTablesPkg/Include/AcpiTableGenerator.h
@@ -91,6 +91,9 @@ The Dynamic Tables Framework implements the following ACPI table generators:
             Configuration Manager and builds the HEST table.
   - EINJ  : The EINJ generator collates the required platform information from
             the Configuration Manager and builds the EINJ table.
+  - MSCT  : The MSCT (Maximum System Characteristics Table) generator collates
+            the Maximum System Characteristics information from the
+            Configuration Manager and builds the MSCT table.
 */
 
 /** The ACPI_TABLE_GENERATOR_ID type describes ACPI table generator ID.
@@ -133,6 +136,7 @@ typedef enum StdAcpiTableId {
   EStdAcpiTableIdHest,                          ///< Hest Generator
   EStdAcpiTableIdEinj,                          ///< EINJ Generator
   EStdAcpiTableIdHmat,                          ///< HMAT Generator
+  EStdAcpiTableIdMsct,                          ///< MSCT Generator
   EStdAcpiTableIdMax
 } ESTD_ACPI_TABLE_ID;
 
--- a/DynamicTablesPkg/Include/AcpiTableGenerator.h
+++ b/DynamicTablesPkg/Include/AcpiTableGenerator.h
@@ -91,6 +91,9 @@ The Dynamic Tables Framework implements the following ACPI table generators:
             Configuration Manager and builds the HEST table.
   - EINJ  : The EINJ generator collates the required platform information from
             the Configuration Manager and builds the EINJ table.
+  - MSCT  : The MSCT (Maximum System Characteristics Table) generator collates
+            the Maximum System Characteristics information from the
+            Configuration Manager and builds the MSCT table.
 */
 
 /** The ACPI_TABLE_GENERATOR_ID type describes ACPI table generator ID.
@@ -134,6 +137,7 @@ typedef enum StdAcpiTableId {
   EStdAcpiTableIdEinj,                          ///< EINJ Generator
   EStdAcpiTableIdHmat,                          ///< HMAT Generator
   EStdAcpiTableIdApmt,                          ///< APMT Generator
+  EStdAcpiTableIdMsct,                          ///< MSCT Generator
   EStdAcpiTableIdMax
 } ESTD_ACPI_TABLE_ID;
 
--- a/DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/Arm/ArmMsctGenerator.c
+++ b/DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/Arm/ArmMsctGenerator.c
@@ -1,0 +1,249 @@
+/** @file
+  ARM specific ACPI Maximum System Characteristics Table (MSCT) Generator
+
+  Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#include <Library/DebugLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/MemoryAllocationLib.h>
+
+// Module specific include files.
+#include <AcpiTableGenerator.h>
+#include <ConfigurationManagerObject.h>
+#include <ConfigurationManagerHelper.h>
+#include <Library/TableHelperLib.h>
+#include <Protocol/ConfigurationManagerProtocol.h>
+#include <Library/CmObjHelperLib.h>
+
+#include "MsctGenerator.h"
+
+/**
+  ARM standard MSCT Generator
+
+  Requirements:
+    The following Configuration Manager Object(s) are used by this Generator:
+    - EArmObjGicCInfo (OPTIONAL)
+*/
+
+/** This macro expands to a function that retrieves the GIC
+    CPU interface Information from the Configuration Manager.
+*/
+GET_OBJECT_LIST (
+  EObjNameSpaceArm,
+  EArmObjGicCInfo,
+  CM_ARM_GICC_INFO
+  );
+
+/** Get architecture processor domain information.
+
+  @param [in]      CfgMgrProtocol       Pointer to the Configuration Manager
+                                        Protocol.
+  @param [out]     ProcDomainInfo       Pointer to the processor domain information.
+  @param [out]     ProcDomainInfoCount  Pointer to the count of processor domain
+                                        information structures.
+
+  @retval EFI_SUCCESS           Operation completed successfully.
+  @retval EFI_INVALID_PARAMETER A parameter is invalid.
+  @retval EFI_OUT_OF_RESOURCES  Memory allocation failed.
+  @retval Other                 Other EFI_STATUS error from called functions.
+**/
+EFI_STATUS
+EFIAPI
+GetArchProcessorDomainInfo (
+  IN  CONST EDKII_CONFIGURATION_MANAGER_PROTOCOL  *CONST  CfgMgrProtocol,
+  OUT       UINT32                                        **ProcDomainInfo,
+  OUT       UINT32                                        *ProcDomainInfoCount
+  )
+{
+  CM_ARM_GICC_INFO  *CmProcDomain;
+  EFI_STATUS        Status;
+  UINT32            Index;
+  UINT32            ProcDomainCount;
+  UINT32            *ProcDomain;
+
+  if ((CfgMgrProtocol == NULL) ||
+      (ProcDomainInfo == NULL) ||
+      (ProcDomainInfoCount == NULL))
+  {
+    ASSERT_EFI_ERROR (EFI_INVALID_PARAMETER);
+    return EFI_INVALID_PARAMETER;
+  }
+
+  CmProcDomain    = NULL;
+  ProcDomain      = NULL;
+  ProcDomainCount = 0;
+
+  Status = GetEArmObjGicCInfo (
+             CfgMgrProtocol,
+             CM_NULL_TOKEN,
+             &CmProcDomain,
+             &ProcDomainCount
+             );
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "ERROR: MSCT: Failed to get GICC Info. Status = %r\n",
+      Status
+      ));
+    return Status;
+  }
+
+  ProcDomain = AllocateZeroPool (
+                 sizeof (UINT32) *
+                 ProcDomainCount
+                 );
+  if (ProcDomain == NULL) {
+    Status = EFI_OUT_OF_RESOURCES;
+    ASSERT_EFI_ERROR (Status);
+    return Status;
+  }
+
+  for (Index = 0; Index < ProcDomainCount; Index++) {
+    Status = GetProximityDomainId (
+               CfgMgrProtocol,
+               CmProcDomain[Index].ProximityDomain,
+               CmProcDomain[Index].ProximityDomainToken,
+               &ProcDomain[Index]
+               );
+    if (EFI_ERROR (Status)) {
+      DEBUG ((
+        DEBUG_ERROR,
+        "MSCT: Failed to get Proximity Domain ID for index %d. Status = %r\n",
+        Index,
+        Status
+        ));
+      ASSERT_EFI_ERROR (Status);
+      goto error_handler;
+    }
+  }
+
+  *ProcDomainInfo      = ProcDomain;
+  *ProcDomainInfoCount = ProcDomainCount;
+  return EFI_SUCCESS;
+
+error_handler:
+
+  if (ProcDomain != NULL) {
+    FreePool (ProcDomain);
+  }
+
+  return Status;
+}
+
+/** Get the count of unique architecture clock domains.
+
+  Enumerates all GIC CPU interfaces and returns the number of unique clock
+  proximity domain IDs in the system.
+
+  @param [in]      CfgMgrProtocol    Pointer to the Configuration Manager
+                                     Protocol.
+  @param [out]     ClockDomainCount  On success, the cardinal count of unique
+                                     clock domains.
+
+  @retval EFI_SUCCESS           Operation completed successfully.
+  @retval EFI_INVALID_PARAMETER A parameter is invalid.
+  @retval EFI_OUT_OF_RESOURCES  Memory allocation failed.
+  @retval Other                 Other EFI_STATUS error from called functions.
+**/
+EFI_STATUS
+EFIAPI
+GetArchClockDomainInfo (
+  IN  CONST EDKII_CONFIGURATION_MANAGER_PROTOCOL  *CONST  CfgMgrProtocol,
+  OUT       UINT32                                        *ClockDomainCount
+  )
+{
+  EFI_STATUS        Status;
+  UINT32            ProximityDomain;
+  UINT32            CmCount;
+  UINT32            Index;
+  UINT32            UniqueIndex;
+  CM_ARM_GICC_INFO  *CmGiccInfo;
+  UINT32            *ClockDomains;
+  UINT32            UniqueCount;
+  BOOLEAN           IsUnique;
+
+  if ((CfgMgrProtocol == NULL) ||
+      (ClockDomainCount == NULL))
+  {
+    ASSERT_EFI_ERROR (EFI_INVALID_PARAMETER);
+    return EFI_INVALID_PARAMETER;
+  }
+
+  CmGiccInfo   = NULL;
+  ClockDomains = NULL;
+  UniqueCount  = 0;
+  CmCount      = 0;
+
+  Status = GetEArmObjGicCInfo (
+             CfgMgrProtocol,
+             CM_NULL_TOKEN,
+             &CmGiccInfo,
+             &CmCount
+             );
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "ERROR: MSCT: Failed to get GICC Info. Status = %r\n",
+      Status
+      ));
+    return Status;
+  }
+
+  if ((CmGiccInfo == NULL) || (CmCount == 0)) {
+    *ClockDomainCount = 0;
+    return EFI_SUCCESS;
+  }
+
+  // Allocate worst-case storage: every CPU could be in its own clock domain.
+  ClockDomains = AllocateZeroPool (sizeof (UINT32) * CmCount);
+  if (ClockDomains == NULL) {
+    Status = EFI_OUT_OF_RESOURCES;
+    ASSERT_EFI_ERROR (Status);
+    return Status;
+  }
+
+  for (Index = 0; Index < CmCount; Index++) {
+    Status = GetProximityDomainId (
+               CfgMgrProtocol,
+               CmGiccInfo[Index].ClockDomain,
+               CmGiccInfo[Index].ClockDomainToken,
+               &ProximityDomain
+               );
+    if (EFI_ERROR (Status)) {
+      DEBUG ((
+        DEBUG_ERROR,
+        "ERROR: MSCT: Failed to get Clock Domain ID for index %d. Status = %r\n",
+        Index,
+        Status
+        ));
+      ASSERT_EFI_ERROR (Status);
+      goto error_handler;
+    }
+
+    // Record the proximity domain ID only if not already seen.
+    IsUnique = TRUE;
+    for (UniqueIndex = 0; UniqueIndex < UniqueCount; UniqueIndex++) {
+      if (ClockDomains[UniqueIndex] == ProximityDomain) {
+        IsUnique = FALSE;
+        break;
+      }
+    }
+
+    if (IsUnique) {
+      ClockDomains[UniqueCount++] = ProximityDomain;
+    }
+  }
+
+  *ClockDomainCount = UniqueCount;
+  Status            = EFI_SUCCESS;
+
+error_handler:
+  if (ClockDomains != NULL) {
+    FreePool (ClockDomains);
+  }
+
+  return Status;
+}
--- a/DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
+++ b/DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
@@ -1,0 +1,1081 @@
+/** @file
+  ACPI Maximum System Characteristics Table (MSCT) Generator
+
+  Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#include <Library/DebugLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/MemoryAllocationLib.h>
+#include <Protocol/AcpiTable.h>
+#include <Library/SortLib.h>
+
+// Module specific include files.
+#include <AcpiTableGenerator.h>
+#include <ConfigurationManagerObject.h>
+#include <ConfigurationManagerHelper.h>
+#include <Library/TableHelperLib.h>
+#include <Protocol/ConfigurationManagerProtocol.h>
+#include <Library/CmObjHelperLib.h>
+
+#include "MsctGenerator.h"
+
+/** Add Id to the Seen[] array if not already present, updating UniqueCount.
+
+  Relies on the enclosing scope providing: Seen[], UniqueCount, SeenIndex,
+  and IsUnique.
+*/
+#define ADD_IF_UNIQUE(Id)                                      \
+  do {                                                         \
+    IsUnique = TRUE;                                           \
+    for (SeenIndex = 0; SeenIndex < UniqueCount; SeenIndex++) {\
+      if (Seen[SeenIndex] == (Id)) {                           \
+        IsUnique = FALSE;                                      \
+        break;                                                 \
+      }                                                        \
+    }                                                          \
+    if (IsUnique) {                                            \
+      Seen[UniqueCount++] = (Id);                              \
+    }                                                          \
+  } while (FALSE)
+
+/**
+  Standard MSCT Generator
+
+  Requirements:
+    The following Configuration Manager Object(s) are used by this Generator:
+    - EArchCommonObjMsctMaxPhysicalAddrInfo (REQUIRED)
+    - EArchCommonObjMemoryAffinityInfo (OPTIONAL)
+    - EArmObjGicCInfo (OPTIONAL)
+    - EX64ObjLocalApicX2ApicAffinityInfo (OPTIONAL)
+*/
+
+/**
+  This macro expands to a function that retrieves the maximum physical address
+  information from the Configuration Manager.
+*/
+GET_OBJECT_LIST (
+  EObjNameSpaceArchCommon,
+  EArchCommonObjMsctMaxPhysicalAddrInfo,
+  CM_ARCH_COMMON_MSCT_MAX_PHYSICAL_ADDR_INFO
+  );
+
+/**
+  This macro expands to a function that retrieves the Memory Affinity
+  information from the Configuration Manager.
+*/
+GET_OBJECT_LIST (
+  EObjNameSpaceArchCommon,
+  EArchCommonObjMemoryAffinityInfo,
+  CM_ARCH_COMMON_MEMORY_AFFINITY_INFO
+  );
+
+/** Comparison function for sorting Maximum Proximity Domain.
+
+  @param [in]  Left    Pointer to the left Proximity Domain structure.
+  @param [in]  Right   Pointer to the right Proximity Domain structure.
+
+  @retval -1 If Left proximity domain < Right proximity domain
+  @retval  0 If Left proximity domain == Right proximity domain
+  @retval  1 If Left proximity domain > Right proximity domain
+**/
+INTN
+EFIAPI
+SortByProximityDomainRange (
+  IN  CONST VOID  *Left,
+  IN  CONST VOID  *Right
+  )
+{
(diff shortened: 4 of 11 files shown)


DynamicTablesPkg/DynamicTables.dsc.inc
```diff
@@ -80,6 +81,8 @@
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiCedtLib/AcpiCedtLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiFadtLib/AcpiFadtLib.inf
       NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiMcfgLib/AcpiMcfgLib.inf
+      NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/AcpiMsctLib.inf
+      NULL|DynamicTablesPkg/Library/Acpi/Common/AcpiSratLib/AcpiSratLib.inf
```
pierregondois: I think there are 2 Srat generators with this line

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/Arm/ArmMsctGenerator.c
```diff
@@ -0,0 +1,214 @@
...
+  EFI_STATUS        Status;
+  UINT32            Index;
+  UINT32            ProcDomainCount;
+  UINT32            *ProcDomain;
+
+  if ((CfgMgrProtocol == NULL) ||
+      (ProcDomainInfo == NULL) ||
+      (ProcDomainInfoCount == NULL))
+  {
+    return EFI_INVALID_PARAMETER;
```
pierregondois, line 72: NIT: might help to have an
```
ASSERT (FALSE);
```
in these conditions

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/Arm/ArmMsctGenerator.c
```diff
@@ -0,0 +1,214 @@
...
+      goto error_handler;
+    }
+  }
+
+  *ProcDomainInfo      = ProcDomain;
+  *ProcDomainInfoCount = ProcDomainCount;
+  return EFI_SUCCESS;
+
+error_handler:
+  if (CmProcDomain != NULL) {
```
pierregondois: Is it necessary to free this pointer ? I thought it was static memory.
Same comment at other places

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/Arm/ArmMsctGenerator.c
```diff
@@ -0,0 +1,214 @@
...
+    return Status;
+  }
+
+  ProcDomain = AllocateZeroPool (
+                 sizeof (UINT32) *
+                 ProcDomainCount
+                 );
+  if (ProcDomain == NULL) {
+    Status = EFI_OUT_OF_RESOURCES;
+    goto error_handler;
```
pierregondois: If it is not necessary to free `CmProcDomain`, we could return from here.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/X64/X64MsctGenerator.c
```diff
@@ -0,0 +1,216 @@
...
+      goto error_handler;
+    }
+  }
+
+  *ProcDomainInfo      = ProcDomain;
+  *ProcDomainInfoCount = ProcDomainCount;
+  return EFI_SUCCESS;
+
+error_handler:
+  if (CmProcDomain != NULL) {
```
pierregondois: Same comments as for the Arm specific implementation I think

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/Arm/ArmMsctGenerator.c
```diff
@@ -0,0 +1,214 @@
...
+      DEBUG ((
+        DEBUG_ERROR,
+        "ERROR: MSCT: Failed to get Clock Domain ID for index %d. Status = %r\n",
+        Index,
+        Status
+        ));
+      return Status;
+    }
+
+    if (ProximityDomain > MaxClockProximityDomain) {
```
pierregondois: I m not sure I read the spec correctly, but shouldn't the `Maximum Number of Clock Domains` field be the maximum number of distinct values of proximity domains instead of the maximum value a proximity domain can take ?

```
Indicates the maximum number of Clock
Domains ever possible in the system. The
number reported in this field is (maximum
domains - 1).
```

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,823 @@
...
+  IN  CONST VOID  *Right
+  )
+{
+  CONST EFI_ACPI_6_5_MAXIMUM_PROXIMITY_DOMAIN_INFORMATION_STRUCTURE  *LeftInfo;
+  CONST EFI_ACPI_6_5_MAXIMUM_PROXIMITY_DOMAIN_INFORMATION_STRUCTURE  *RightInfo;
+
+  LeftInfo  = (CONST EFI_ACPI_6_5_MAXIMUM_PROXIMITY_DOMAIN_INFORMATION_STRUCTURE *)Left;
+  RightInfo = (CONST EFI_ACPI_6_5_MAXIMUM_PROXIMITY_DOMAIN_INFORMATION_STRUCTURE *)Right;
+
+  if (LeftInfo->ProximityDomainRangeLow < RightInfo->ProximityDomainRangeLow) {
```
pierregondois, line 109: It seems we are comparing ranges. If the ranges are interleaved, I think we should ASSERT.
Maybe this should cover these cases.

```
  ASSERT (LeftInfo->ProximityDomainRangeLow <= LeftInfo->ProximityDomainRangeHigh);
  ASSERT (RightInfo->ProximityDomainRangeLow <= RightInfo->ProximityDomainRangeHigh);
  
  // If the interval are disjoint
  if (LeftInfo->ProximityDomainRangeHigh < RightInfo->ProximityDomainRangeLow) {
    return -1;
  } else if (LeftInfo->ProximityDomainRangeLow < RightInfo->ProximityDomainRangeHigh) {
    return 1;
  } else if ((LeftInfo->ProximityDomainRangeLow == RightInfo->ProximityDomainRangeLow) &&
             (LeftInfo->ProximityDomainRangeHigh == RightInfo->ProximityDomainRangeHigh)) {
    return 0;
  }

  // This is for the interleaved case.
  // Unfortunately we cannot return an error code.
  ASSERT (FALSE);
  return 0;
```

If we don't allow to describe ranges, i.e. only one value per descriptor, the code could also be simplified.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,823 @@
...
+  UINT32      Index;
+
+  EFI_ACPI_6_5_MAXIMUM_PROXIMITY_DOMAIN_INFORMATION_STRUCTURE  *ProcDomain;
+  EFI_ACPI_6_5_MAXIMUM_PROXIMITY_DOMAIN_INFORMATION_STRUCTURE  *ProcDomainRaw;
+
+  if ((CfgMgrProtocol == NULL) ||
+      (ProcDomainInfo == NULL) ||
+      (ProcDomainInfoCount == NULL))
+  {
+    return EFI_INVALID_PARAMETER;
```
pierregondois, line 160: NIT: Maybe have 
```
ASSERT (FALSE);
```
just in case. Same comment at similar places

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,823 @@
...
+  ProcDomainArch      = NULL;
+  ProcDomainArchCount = 0;
+
+  Status = GetArchProcessorDomainInfo (
+             CfgMgrProtocol,
+             &ProcDomainArch,
+             &ProcDomainArchCount
+             );
+  if (EFI_ERROR (Status)) {
+    goto error_handler;
```
pierregondois: Similar comment as for the Arm specific implementation, but IIRC the objects returned by the configuration manager don't need to be freed.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,823 @@
...
+    ProcDomainRaw[ProximityDomainId].ProximityDomainRangeHigh = ProximityDomainId;
+    ProcDomainRaw[ProximityDomainId].MaximumProcessorCapacity++;
+
+    MaxProcDomain = MAX (MaxProcDomain, ProximityDomainId);
+  }
+
+  /// Adjust zero based proximity domain
+  MaxProcDomain++;
+  ProcDomain = AllocateZeroPool (
+                 MaxProcDomain *
```
pierregondois: IIUC `MaxProcDomain` is the maximal value a proximity domain can take. The proximity domain IDs can be any number in the UINT32 range. I think this would be better to use the maximum number of different proximity domain IDs, which should be lower.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,823 @@
...
+             &MemAffInfo,
+             &MemAffCount
+             );
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "ERROR: MSCT: Failed to get Memory Affinity Info. Status = %r\n",
+      Status
+      ));
+    goto return_handler;
```
pierregondois: Same comment as above I think

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,823 @@
...
+    goto error_handler;
+  }
+
+  /// Get the maximum memory capacity per proximity domain.
+  for (MemIndex = 0; MemIndex <= MemDomainInfoCount; MemIndex++) {
+    if (MemDomainInfo[MemIndex].ProximityDomainRangeLow == MAX_UINT32) {
+      continue;
+    }
+
+    MaxProximityDomain = MAX (
```
pierregondois: This value doesn't seem to be used

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,823 @@
...
+                           MaxProximityDomain,
+                           MemDomainInfo[MemIndex].ProximityDomainRangeHigh
+                           );
+
+    if (MemDomainInfo[MemIndex].ProximityDomainRangeLow ==
+        MemDomainInfo[MemIndex + 1].ProximityDomainRangeLow)
```
pierregondois: I think it is not guaranteed that all the memory domain information with the same proximity domain ID will follow each other. It might be necessary to order them first. Otherwise we might have this list of proximity domain IDs and this loop won't be able to group them:
0 - 1 - 0 - 2 - 0
Here the proximity domains with the ID=0 won't be grouped I think

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,823 @@
...
+      (UINT8 *)DomainInfo,
+      sizeof (EFI_ACPI_6_5_MAXIMUM_PROXIMITY_DOMAIN_INFORMATION_STRUCTURE) * DomainInfoIndex
+      );
+    MsctTable->OffsetProxDomInfo = sizeof (EFI_ACPI_6_5_MAXIMUM_SYSTEM_CHARACTERISTICS_TABLE_HEADER);
+  }
+
+  MsctTable->Header.Length = TableSize;
+
+  ///
+  /// As per the specification maximum number of proximity domain
```
pierregondois, line 830: Same question as for the maximum number of clock domains. Shouldn't this be the maximum cardinal and not the maximum ID ?

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,823 @@
...
+          );
+        DomainInfo[DomainInfoIndex].MaximumMemoryCapacity = MemDomainInfo[MemIndex].MaximumMemoryCapacity;
+        DomainInfoIndex++;
+
+        /// Invalidate the records
+        ProcDomainInfo[ProcIndex].ProximityDomainRangeLow  = MAX_UINT32;
+        ProcDomainInfo[ProcIndex].ProximityDomainRangeHigh = MAX_UINT32;
+        MemDomainInfo[MemIndex].ProximityDomainRangeLow    = MAX_UINT32;
+        MemDomainInfo[MemIndex].ProximityDomainRangeHigh   = MAX_UINT32;
+      }
```
pierregondois, line 731: IIUC the memory domain have already be grouped, so we could break out of the loop here.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,823 @@
...
+    for (MemIndex = 0; MemIndex < MemDomainInfoCount; MemIndex++) {
+      if (MemDomainInfo[MemIndex].ProximityDomainRangeLow == MAX_UINT32) {
+        continue;
+      }
+
+      if (ProcDomainInfo[ProcIndex].ProximityDomainRangeLow ==
+          MemDomainInfo[MemIndex].ProximityDomainRangeLow)
+      {
+        CopyMem (
+          &DomainInfo[DomainInfoIndex],
```
pierregondois: Isn't it possible to group the processor and memory information into one of the existing array, i.e. either `ProcDomainInfo` or `MemDomainInfo` ? That way we don't need to allocate the extra `DomainInfo` array.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,823 @@
...
+          &ProcDomainInfo[ProcIndex],
+          sizeof (EFI_ACPI_6_5_MAXIMUM_PROXIMITY_DOMAIN_INFORMATION_STRUCTURE)
+          );
+        DomainInfo[DomainInfoIndex].MaximumMemoryCapacity = MemDomainInfo[MemIndex].MaximumMemoryCapacity;
```
pierregondois, line 722: Isn't the CopyMem already copying the value ?

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,823 @@
...
+        /// Invalidate the records
+        ProcDomainInfo[ProcIndex].ProximityDomainRangeLow  = MAX_UINT32;
+        ProcDomainInfo[ProcIndex].ProximityDomainRangeHigh = MAX_UINT32;
+        MemDomainInfo[MemIndex].ProximityDomainRangeLow    = MAX_UINT32;
+        MemDomainInfo[MemIndex].ProximityDomainRangeHigh   = MAX_UINT32;
+      }
+    }
+  }
+
+  /// Club the symmetric continuous processor proximity domains
```
pierregondois: NIT: This might be more difficult, but ideally it should be possible to also group the structures which have the same `MaximumProcessorCapacity` and `MaximumMemoryCapacity`.
It might not be worth it, just noting it

DynamicTablesPkg/DynamicTables.dsc.inc
abdattar: Done.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/Arm/ArmMsctGenerator.c
abdattar: Done.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/Arm/ArmMsctGenerator.c
abdattar: Done.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/Arm/ArmMsctGenerator.c
abdattar: Done.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/X64/X64MsctGenerator.c
abdattar: Done.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: Done.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: Done.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: Done.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: Done. removed it.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: CopyMem copies the processor record, in this line we are copying the memory information as part of merging both the records.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/Arm/ArmMsctGenerator.c
abdattar: as per acpi spec 6.5 section 5.2.19, it says maximum number of domain ever possible.
image

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: I think we are doing the same.
MaxProcDomain contains the maximum number of different proximity domain ID.
at line 173 (or just two lines above this comment)
    MaxProcDomain = MAX (MaxProcDomain, ProximityDomainId);

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: Done. sorted the memory proximity domain.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: its maximum number of domain possible.
image

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: Done.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: Correcting the comment.
We already grouped both processor and memory proximity domain.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: I think we should keep the quick sort wrapper simple.
interleaving is handled while grouping the memory domain ID.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
pierregondois: I m thinking about the following case:
```
CPUX: ProximityDomain
CPU0: 0
CPU1: 10
CPU2: 2
```
Indeed, there is no reason Proximity domain should follow a specific pattern.
In such case:
```
ProcDomainArchCount == 3
ProcDomainRaw == Array of 3 elements
MaxProcDomain == 10
ProcDomain == Array of 10 elements
```

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
pierregondois: It seems you are correct, but I fail to understand why the specs does that.
IIRC there is no real constraint on the Proximity Domain values, so the processors could have:
```
CPU0 -> ProximityDomain == 0x10000
CPU1 -> ProximityDomain == 0x10001
```

This would mean that the `Proximity Domain Information Structure[Maximum Number of Proximity Domains]` field of the `Maximum System Characteristics Table (MSCT)` table will be huge.

Also there is no real interest in grouping these structure if the reserved memory still has to account for `[Maximum Number of Proximity Domains]` cells. If one tries to access the cell number 0xFFFF and structure have been grouped, then empty memory will be accesses IIUC.

abdattar: There are two different comment, let me explain for CPU proximity domain.
same logic is applied for Memory 
1) regarding huge MSCT and grouping of CPU proximity domain.
   The size of MSCT will depends on number of CPU proximity domain plus the number of processor in each domain.
   1.1) best case scenario
             e.g there are 'n' proximity CPU domain and each contains single processor
             CPU0 -> ProximityDomain == 0x10000
             CPU1 -> ProximityDomain == 0x10001
             CPUn -> ProximityDomain == 0x1000n
            Since each proximity contains only "single" processor and number of proximity are sequential 0x1000, 0x10001, 0x1000n
            Final MSCT will be having single entry
              start 0x10000 end 0x1000n, and maximum proximity domain ever possible is  0x1000n.
      1.2) worst case scenario
          e.g there are 'n' proximity CPU domain and each contains different number of processor
          CPU0->  ProximityDomain == 0x10000
          CPU1-> ProximityDomain == 0x10001
          CPU2 -> ProximityDomain == 0x10001 (same proximity domain as CPU1)
          CPU3 -> ProximityDomain == 0x10002
          CPU4 -> ProximityDomain == 0x10003
          CPU5 -> ProximityDomain == 0x10003 (same proximity domain as CPU4)
          .....
          In above worst case scenario example 
          proximity domain  0x10000 has single processor where as 0x10001 has two processor, hence it wont be grouped.
          ProximityDomain 0x10000 and 0x10002 are having single procesor but they are not sequential hence will be grouped.
          so the final outcome in MSCT(with above example), will have three records
          1)  CPU0->  ProximityDomain == 0x10000
          2) 
               CPU1-> ProximityDomain == 0x10001
               CPU2 -> ProximityDomain == 0x10001 (same proximity domain as CPU1)
           3) CPU3 -> ProximityDomain == 0x10002
           4) 
                     CPU4 -> ProximityDomain == 0x10003
          CPU5 -> ProximityDomain == 0x10003 (same proximity domain as CPU4

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/X64/X64MsctGenerator.c
pierregondois: I think it's still there actually,
normally the objects provided by the configuration manager don t need to be free

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
pierregondois: Hello Abdul,
I think the way CPUs are grouped is a second topic. What I misunderstand is the number of `Proximity Domain Information Structure` that need to be allocated and how they are indexed.

In your above example, in the best case scenario:
```
CPU0 -> ProximityDomain == 0x10000
CPU1 -> ProximityDomain == 0x10001
CPUn -> ProximityDomain == 0x1000n
```
How many `Proximity Domain Information Structure` do we need to allocated ?
- If this is `0x1000n`, then we are following the ACPI AFAIC. This also seems to be what you implemented. However ProximityDomain are UINT32 values, so theoretically, we could have a ProximityDomain=`0xFFFFFFFF`, which makes ~2**32. One cell will be at least one byte long, so we reserve ~4Gb of memory for the ACPI table. It seems a bit strange.
- If this is `n`, then we save a lot of memory. However this is not what the spec describes.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/X64/X64MsctGenerator.c
abdattar: Done. removed it

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: Done. optimize the  code.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: Done. optimize the code to allocate only for proximity domain.

pierregondois: @jmarinho 

Would it be possible to have another opinion on this question ? The spec. is a bit misleading as:
- if  `Maximum Number of Proximity Domains` is a maximal value, then the array can take a huge amount of memory. Cf. above, Proximity domain values are UINT32, which might take up to ~2**32 bytes so ~=4GB
- if `Maximum Number of Proximity Domains` is a cardinal, then the spec states `The number reported in this field is (maximum domains - 1)`, which doesn't really make sense if the reported value is a cardinal. Indeed, if there are 2 possible `Proximity domain` values, then the `Proximity Domain Information Structure` array will only have one element.

jmarinho: Hi,

my 2 cents (and looking at it from the ground up):
- The MSCT field "Maximum Number of Proximity Domains" is the num entries in the array "Proximity Domain Information Structure".
- that array has an entry (at most) per proximity domains in the system.

So to me it seems "Maximum Number of Proximity Domains" = count of proximity domains.

The spec may incorrectly suggest proximity domain ID. 

Looping others: @samerhaj, @rgough-ampere

abdattar: @pierregondois,
I have addressed the review comments and optimize the code.
I think the "Maximum Number of Proximity Domains" is cardinal value.
Thanks
AbduL

abdattar:
Hi @pierregondois , I've rebased my changes and addressed your review comments. Could you please take another look?

pierregondois:
Hello Abdul,
Sorry for the delay, I ll review the PR next week

pierregondois:
Hello Abdul,
I m sorry but I still have difficulties to understand what is meant in the spec.
```
Maximum Number of Proximity Domains:

Indicates the maximum number of Proxim-
ity Domains ever possible in the system.
The number reported in this field is (max-
imum domains - 1). For example if there
are 0x10000 possible domains in the system,
this field would report 0xFFFF.
```

If there are 0 (resp. 1) Proximity Domains, what should be the reported value ?
- 0 -> (0 - 1) = -1 ?
- 1 -> (1 - 1) = 0 ?
Plus I am still not sure how to interpret the `Maximum Number`, whether this is supposed to be a maximal value or a maximal cardinal.

I will try to find out more information, but unfortunately I think it is better to hold the PR until the spec. is clarified (or I manage to understand its meaning/purpose).

jmarinho:
> Hello Abdul, I m sorry but I still have difficulties to understand what is meant in the spec.
> 
> ```
> Maximum Number of Proximity Domains:
> 
> Indicates the maximum number of Proxim-
> ity Domains ever possible in the system.
> The number reported in this field is (max-
> imum domains - 1). For example if there
> are 0x10000 possible domains in the system,
> this field would report 0xFFFF.
> ```
> 
> If there are 0 (resp. 1) Proximity Domains, what should be the reported value ?
> 
> * 0 -> (0 - 1) = -1 ?
> * 1 -> (1 - 1) = 0 ?
>   Plus I am still not sure how to interpret the `Maximum Number`, whether this is supposed to be a maximal value or a maximal cardinal.
> 
> I will try to find out more information, but unfortunately I think it is better to hold the PR until the spec. is clarified (or I manage to understand its meaning/purpose).

Hi Both,

I can inform that the MSCT table definition is being revisited.
We'll report the corrections once the process is finished (@rgough-ampere).

jmarinho:
@pierregondois 
@abdattar 

the code-first ACPI spec change clarifying the MSCT definition was accepted.

The accepted table is in this ACPI spec PR: https://github.com/UEFI/ACPI_spec/pull/183/changes

The rendered new table is:
image

abdattar:
Rebase and resolve the conflicts.
@pierregondois  please review again, can we target this patch for edk2-202608 stable release?

pierregondois:
> Rebase and resolve the conflicts. @pierregondois please review again, can we target this patch for edk2-202608 stable release?

Hello Abdul,
I ll review the patch this week

pierregondois:
@abdattar I re-read https://github.com/tianocore/edk2/issues/12453 and https://github.com/tianocore/edk2/pull/11880#discussion_r2811943881 and FWIU:
- Maximum Number of Proximity Domains
- Maximum Number of Clock Domains

are describing a maximal value for the cardinal, not a higher bound value. I'm not sure this is what the patch does, cf:
```
      MaxProcDomain = MAX (MaxProcDomain, ProcDomainArch[Index]);
```
and
```
    if (ProximityDomain > MaxClockProximityDomain) {
      MaxClockProximityDomain = ProximityDomain;
    }
```

@jmarinho would it be possible to confirm the interpretation is correct ?

jmarinho:
> @abdattar I re-read #12453 and [#11880 (comment)](https://github.com/tianocore/edk2/pull/11880#discussion_r2811943881) and FWIU:
> 
> * Maximum Number of Proximity Domains
> * Maximum Number of Clock Domains
> 
> are describing a maximal value for the cardinal, not a higher bound value. I'm not sure this is what the patch does, cf:
> 
> ```
>       MaxProcDomain = MAX (MaxProcDomain, ProcDomainArch[Index]);
> ```
> 
> and
> 
> ```
>     if (ProximityDomain > MaxClockProximityDomain) {
>       MaxClockProximityDomain = ProximityDomain;
>     }
> ```
> 
> @jmarinho would it be possible to confirm the interpretation is correct ?

Guys, @pierregondois @abdattar , please see the clarified spec wording:
image

fields are the cardinal, or actual count of domains in the system!

abdattar:
> > @abdattar I re-read #12453 and [#11880 (comment)](https://github.com/tianocore/edk2/pull/11880#discussion_r2811943881) and FWIU:
> > 
> > * Maximum Number of Proximity Domains
> > * Maximum Number of Clock Domains
> > 
> > are describing a maximal value for the cardinal, not a higher bound value. I'm not sure this is what the patch does, cf:
> > ```
> >       MaxProcDomain = MAX (MaxProcDomain, ProcDomainArch[Index]);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and
> > ```
> >     if (ProximityDomain > MaxClockProximityDomain) {
> >       MaxClockProximityDomain = ProximityDomain;
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @jmarinho would it be possible to confirm the interpretation is correct ?
> 
> Guys, @pierregondois @abdattar , please see the clarified spec wording: image
> 
> fields are the cardinal, or actual count of domains in the system!

Hi @pierregondois  and @jmarinho , I had updated the patch to honor the cardinal value.
Could you please review it.

DynamicTablesPkg/DynamicTables.dsc.inc
```diff
@@ -40,6 +40,7 @@
   DynamicTablesPkg/Library/Acpi/Common/AcpiFadtLib/AcpiFadtLib.inf
   DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/AcpiHmatLib.inf
   DynamicTablesPkg/Library/Acpi/Common/AcpiMcfgLib/AcpiMcfgLib.inf
+  DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/AcpiMsctLib.inf
```
pierregondois, line 44: The generator is supposed to have support for `RISCV64/LOONGARCH64` aswell,
so we might need a `Default` (or better name) implementation of the arch-specific functions for these archs,
even if the implementation only returns `EFI_UNSUPPORTED`

DynamicTablesPkg/Include/AcpiTableGenerator.h
```diff
@@ -91,6 +91,9 @@ The Dynamic Tables Framework implements the following ACPI table generators:
             Configuration Manager and builds the HEST table.
   - EINJ  : The EINJ generator collates the required platform information from
             the Configuration Manager and builds the EINJ table.
+  - MSCT  : The MSCT (Maximum System Characteristics Table) generator collates
```
pierregondois, line 94: `DynamicTablesPkg: Add MSCT standard table identifier` could be merged inside `DynamicTablesPkg: Add ACPI MSCT table generator`

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,1068 @@
...
+    return EFI_OUT_OF_RESOURCES;
+  }
+
+  // The input array is sorted by ProximityDomainRangeLow; scan linearly and
+  // record a new entry only when the ID differs from the previous one.
+  UniqueCount  = 0;
+  UniqueIds[0] = MemDomainInfo[0].ProximityDomainRangeLow;
+  UniqueCount  = 1;
+  for (Index = 1; Index < MemDomainInfoCount; Index++) {
+    if (MemDomainInfo[Index].ProximityDomainRangeLow !=
```
pierregondois, line 430: If multiple memory descriptors have the same ProcDomainId,
shouldn't we increment the MaxMemory of this ProcDomainId ?

This somehow looks like `ProcDomain[ProcDomainIndex].MaximumProcessorCapacity++;`,
i.e. for both memory and CPU capacity:
1. we fetch the CmObj
2. for each CmObj, the ProxDomId is fetched
3. an array of `EFI_ACPI_6_5_MAXIMUM_PROXIMITY_DOMAIN_INFORMATION_STRUCTURE` is created, which contains a mapping of `ProxDomId` -> `MemSize`|`CpuCapa`, entries with the same `ProxDomId` are allowed in this first step
4. this array is sorted by `ProxDomId` and entries with the same `ProxDomId` are merged (resulting in `MemSize`|`CpuCapa being incremented)

Then finally the 2 arrays (memory | CpuCapacity) are merged (and that ranges of ProxDomId are allowed).
Would it make sense to have generic callbacks handling memory/CpuCapacity the same way ?

I.e.
- for 1., 2. and 3. a generic interface that would return an array of (ProxDomId, Value), with Value being CpuCount or the memory size (it could be stored in a `EFI_ACPI_6_5_MAXIMUM_PROXIMITY_DOMAIN_INFORMATION_STRUCTURE` if it is easier)
- for 4., a common function should be possible

Please let me know if I you think this is not feasible

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,1068 @@
...
+  if (MaxProximityDomain > 0) {
+    MsctTable->MaximumNumberOfProximityDomains = MaxProximityDomain - 1;
+  } else {
+    MsctTable->MaximumNumberOfProximityDomains = 0;
+  }
+
+  if (ClockDomainIdsCount > 0) {
+    MsctTable->MaximumNumberOfClockDomains = ClockDomainIdsCount - 1;
+  } else {
+    MsctTable->MaximumNumberOfClockDomains = 0;
```
pierregondois, line 843: Maybe ASSERT () in such case.

This is would be a check stricter than the ACPI spec, but we might want to avoid the case where no clock proximity domain was defined for a platform and a MSCT table is emitted.

Also please add a comment explaining the above reason.

abdattar:
Hi @pierregondois , can we please target this PR for current stable release?

pierregondois (commented):
@abdattar sorry I thought I published these comments last week but they were still in draft

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: Done. added ASSERT at line 602.

DynamicTablesPkg/Include/AcpiTableGenerator.h
abdattar: Done.

DynamicTablesPkg/DynamicTables.dsc.inc
abdattar: Done.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: > If multiple memory descriptors have the same ProcDomainId, shouldn't we increment the MaxMemory of this ProcDomainId ?
Yes, we do increment the MaxMemory at line 682.
      // Maximum memory capacity per domain.
      MemDomainInfo[MemIndex].MaximumMemoryCapacity +=
        MemDomainInfo[MemNextIndex].MaximumMemoryCapacity;

> 
> This somehow looks like `ProcDomain[ProcDomainIndex].MaximumProcessorCapacity++;`, i.e. for both memory and CPU capacity:
> 
> 1. we fetch the CmObj
> 2. for each CmObj, the ProxDomId is fetched
> 3. an array of `EFI_ACPI_6_5_MAXIMUM_PROXIMITY_DOMAIN_INFORMATION_STRUCTURE` is created, which contains a mapping of `ProxDomId` -> `MemSize`|`CpuCapa`, entries with the same `ProxDomId` are allowed in this first step
> 4. this array is sorted by `ProxDomId` and entries with the same `ProxDomId` are merged (resulting in `MemSize`|`CpuCapa being incremented)
> 
> Then finally the 2 arrays (memory | CpuCapacity) are merged (and that ranges of ProxDomId are allowed). Would it make sense to have generic callbacks handling memory/CpuCapacity the same way ?
> 
> I.e.
> 
> * for 1., 2. and 3. a generic interface that would return an array of (ProxDomId, Value), with Value being CpuCount or the memory size (it could be stored in a `EFI_ACPI_6_5_MAXIMUM_PROXIMITY_DOMAIN_INFORMATION_STRUCTURE` if it is easier)
> * for 4., a common function should be possible
> 
> Please let me know if I you think this is not feasible

I think current implementation is simple and readable, I feel making it generic makes more complex.
plus 
1. CmObj is different for processor and memory.
2. populating EFI_ACPI_6_5_MAXIMUM_PROXIMITY_DOMAIN_INFORMATION_STRUCTURE is different for processor  and memory.

abdattar:
Hi @pierregondois , i had addressed the review comments, please take a look.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,1081 @@
...
+#include <Library/CmObjHelperLib.h>
+
+#include "MsctGenerator.h"
+
+/** Add Id to the Seen[] array if not already present, updating UniqueCount.
+
+  Relies on the enclosing scope providing: Seen[], UniqueCount, SeenIndex,
+  and IsUnique.
+*/
+#define ADD_IF_UNIQUE(Id)                                      \
```
pierregondois: Is it possible to create a function instead ?

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,1081 @@
...
+  // Allocate worst-case: all entries could have distinct proximity domain IDs.
+  UniqueIds = AllocateZeroPool (sizeof (UINT32) * MemDomainInfoCount);
+  if (UniqueIds == NULL) {
+    ASSERT_EFI_ERROR (EFI_OUT_OF_RESOURCES);
+    return EFI_OUT_OF_RESOURCES;
+  }
+
+  // The input array is sorted by ProximityDomainRangeLow; scan linearly and
+  // record a new entry only when the ID differs from the previous one.
+  UniqueCount  = 0;
```
pierregondois, line 426: Can remove this one `UniqueCount  = 0;` I think

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,1081 @@
...
+      // Maximum memory capacity per domain.
+      MemDomainInfo[MemIndex].MaximumMemoryCapacity +=
+        MemDomainInfo[MemNextIndex].MaximumMemoryCapacity;
+      // Invalidate the folded record.
+      MemDomainInfo[MemNextIndex].ProximityDomainRangeLow  = MAX_UINT32;
+      MemDomainInfo[MemNextIndex].ProximityDomainRangeHigh = MAX_UINT32;
+    }
+  }
+
+  /// Perform union of MemDomainInfo and ProcDomainInfo for same proximity domain
```
pierregondois: Is it possible to add a bit more details ? Maybe:
```
  /// Perform union of MemDomainInfo and ProcDomainInfo for same proximity domain, letting:
  /// - ProcDomainInfo[] with Proximity domains describing Processors with no memory associated
  /// - MemDomainInfo[] with Proximity domains describing memory with no processor associated
```

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,1081 @@
...
+        ProcDomainInfo[ProcIndex - 1].ProximityDomainRangeHigh =
+          ProcDomainInfo[ProcIndex].ProximityDomainRangeHigh;
+
+        ProcDomainInfo[ProcIndex].ProximityDomainRangeLow  = MAX_UINT32;
+        ProcDomainInfo[ProcIndex].ProximityDomainRangeHigh = MAX_UINT32;
+      }
+    }
+  }
+
+  /// Copy the processor proximity record
```
pierregondois: ```Copy the processor proximity record (i.e. the processors with no memory associated)```

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
```diff
@@ -0,0 +1,1081 @@
...
+
+    CopyMem (
+      &DomainInfo[DomainInfoIndex],
+      &ProcDomainInfo[ProcIndex],
+      sizeof (EFI_ACPI_6_5_MAXIMUM_PROXIMITY_DOMAIN_INFORMATION_STRUCTURE)
+      );
+    DomainInfoIndex++;
+  }
+
+  /// Copy remaining memory record
```
pierregondois: ```Copy remaining memory record (i.e. the memory with no processor associated)```

pierregondois (commented):
Hello Abdul,
some last comments I think

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: Done.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: Done.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: Done.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: Done.

DynamicTablesPkg/Library/Acpi/Common/AcpiMsctLib/MsctGenerator.c
abdattar: Done.

abdattar:
Hi @pierregondois, Gentle reminder to merge the patch if there are no further review comments.

pierregondois:
> Hi @pierregondois, Gentle reminder to merge the patch if there are no further review comments.

Hi Abdul,
the CI had some issues this week. Would it be possible to rebase the PR ?

abdattar:
Hi @pierregondois , i have rebase the patch, could you please push the changes.

pierregondois (approved):

PR #11883, opened 2025-12-11
MdeModulePkg: Fix boot failure intrioduced by #11807

khaliid2040:
# Description

The changes introduced by #11807 resulted boot hang on platforms where one OpRom image is present. Due two calls to ProcessOpRomImage() which in turns two calls to AddDriver() respectively. This behaiviour resulted the same image device path to be appended twice in PCI_DRIVER_OVERRIDE_LIST. Which resulted breakage in those environment. 

This also has side effect on iteration done by EFI_BUS_SPECIFIC_DRIVER_OVERRIDE_PROTOCOL.GetDriver when getting the image handle via device path, which resulted the same image handle to be retrieved and stored twice.

Solution is, making sure only once is appended to the list and skipping it, if already appended. 

I am welcome to more solutions that could fix that bug.

- [ ] Breaking change?
  - NO
- [ ] Impacts security?
  - NO
- [ ] Includes tests?
  - NO

## How This Was Tested
On OVMF X64 build target DEBUG, run under qemu.
## Integration Instructions
N/A

### DIFF
--- a/MdeModulePkg/Bus/Pci/PciBusDxe/PciOptionRomSupport.c
+++ b/MdeModulePkg/Bus/Pci/PciBusDxe/PciOptionRomSupport.c
@@ -706,7 +706,7 @@ ProcessOpRomImage (
     // Skip the EFI PCI Option ROM image if it's machine type is not supported
     // and the native image is only considered.
     //
-    if (!EFI_IMAGE_MACHINE_TYPE_SUPPORTED (EfiRomHeader->EfiMachineType) && (NativeOnly == TRUE)) {
+    if (EFI_IMAGE_MACHINE_TYPE_SUPPORTED (EfiRomHeader->EfiMachineType) ^ (NativeOnly == TRUE)) {
       goto NextImage;
     }
 


MdeModulePkg/Bus/Pci/PciBusDxe/PciOptionRomSupport.c
```diff
@@ -750,6 +750,12 @@ ProcessOpRomImage (
       // PciOverride.GetDriver() will try to look for the Image Handle using the device path later.
       //
       AddDriver (PciDevice, NULL, PciOptionRomImageDevicePath);
+      // If reason of failure is due security violaitions, then abort image loading.
+      if (Status == EFI_ACCESS_DENIED) {
```
leiflindholm: I don't think we want to do this when secure boot _is_ enabled.

khaliid2040: I was wrong. So, i updated everything including the PR description and commit. LoadImage() failing is not neccessary even failed due secure boot.

khaliid2040:
cc: @mikebeaton, @EricGao2015, @ardbiesheuvel, @kilaterlee.

the CI workflow responsible for adding reviewers didn't add you, so please review this patch when you have time.

You were involved in changes introduced by #11807 which caused boot failure on environment with one OpRom image

MdeModulePkg/Bus/Pci/PciBusDxe/PciOptionRomSupport.c
```diff
@@ -641,12 +641,12 @@ RomDecode (
 /**
   Load and start the Option Rom image.
```
ardbiesheuvel, line 643: Please fix the spelling in the commit title, and please don't put PR references in there. If you need to refer to another patch, use the commit hash.

MdeModulePkg/Bus/Pci/PciBusDxe/PciOptionRomSupport.c
```diff
@@ -750,6 +750,12 @@ ProcessOpRomImage (
       // PciOverride.GetDriver() will try to look for the Image Handle using the device path later.
       //
       AddDriver (PciDevice, NULL, PciOptionRomImageDevicePath);
+      // If reason of failure is due security violaitions, then abort image loading.
+      if (Status == EFI_ACCESS_DENIED) {
+        RetStatus = Status;
+        FreePool (PciOptionRomImageDevicePath);
+        break;
+      }
```
ardbiesheuvel: I'd like to understand better how this logic change resulted in this issue. The intent was only for the prioritization to change, but if I understand this correctly, we may end up prioritizing an image that lacks a signature over one that does?

khaliid2040: Actually the native one is also not signed. The issue was present before even the change is introduced. LoadImage() was silently failing with secure boot all along, but since previous code the return value of the function didn't get checked, so it was skipping loading the image even native one if it is not signed.

When the change is introduced it exposed that bug and when both native and foreign ones failed it resulted boot failure. So i thought it is good idea to skip loading unsigned image at all which made boot successful.

ardbiesheuvel: With the patch in question applied, `RegisterPciDevice()` calls `ProcessOpRomImage()` a second time if the first call returns `EFI_NOT_FOUND`, but it does not propagate the return value of either call.

So it is not failure of the call itself that is causing the issue, but some other side effect that is causing problems later. Without any logs or other justification how your change address that situation, there is no way we can decide whether your fix is correct or not.

Can you provide a reproducer?

khaliid2040: I was wrong. So, i updated everything including the PR description and commit. LoadImage() failing is not neccessary even failed due secure boot.

ardbiesheuvel: OK, I think I understand now what is going on: `EFI_NOT_FOUND` does not mean no image was loaded. So the first time, the image is loaded and started. The second time time, the same is loaded and started again, because that time, we permit both native and non-native images.

Can you check if the below fixes the issue?


```diff
--- a/MdeModulePkg/Bus/Pci/PciBusDxe/PciOptionRomSupport.c
+++ b/MdeModulePkg/Bus/Pci/PciBusDxe/PciOptionRomSupport.c
@@ -706,7 +706,7 @@ ProcessOpRomImage (
     // Skip the EFI PCI Option ROM image if it's machine type is not supported
     // and the native image is only considered.
     //
-    if (!EFI_IMAGE_MACHINE_TYPE_SUPPORTED (EfiRomHeader->EfiMachineType) && (NativeOnly == TRUE)) {
+    if (EFI_IMAGE_MACHINE_TYPE_SUPPORTED (EfiRomHeader->EfiMachineType) ^ (NativeOnly == TRUE)) {
       goto NextImage;
     }
 
```

khaliid2040: Thanks, it fixed. I submitted new patch with your changes. If you satisfied with the patch, please i would like to ask you to merge as fast as possible so issue is fixed.

ardbiesheuvel (approved):

PR #11884, opened 2025-12-12
UefiCpuPkg/CpuExceptionHandlerLib:  Add LOONGARCH_IOCSR_MBUF3 RereadCount

MarsDoge:
# Description

If the multi-core IPI trapping into an SMP_CALL_FUNCTION interrupt is used, then the Context will be allowed to be null.

## How This Was Tested

Tested on LoongArch-3C6000 platform

## Integration Instructions

N/A

### DIFF
--- a/UefiCpuPkg/Library/CpuExceptionHandlerLib/LoongArch/LoongArch64/ArchExceptionHandler.c
+++ b/UefiCpuPkg/Library/CpuExceptionHandlerLib/LoongArch/LoongArch64/ArchExceptionHandler.c
@@ -244,6 +244,7 @@ IpiInterruptHandler (
   UINTN  ResumeVector = 0;
   UINTN  Parameter    = 0;
   UINTN  IpiStatus    = 0;
+  UINT8  RereadCount  = 10;
 
   IpiStatus = IoCsrRead32 (LOONGARCH_IOCSR_IPI_STATUS);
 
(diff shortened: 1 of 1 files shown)


UefiCpuPkg/Library/CpuExceptionHandlerLib/LoongArch/LoongArch64/ArchExceptionHandler.c
```diff
@@ -272,13 +272,6 @@ IpiInterruptHandler (
       if ((IpiStatus & SMP_BOOT_CPU) != 0) {
         SystemContext.SystemContextLoongArch64->PRMD &= ~((UINT64)BIT2); // Clean PIE
       } else if ((IpiStatus & SMP_CALL_FUNCTION) != 0) {
-        //
-        // Confirm that the mail box message has arrived.
-        //
-        do {
-          Parameter = IoCsrRead32 (LOONGARCH_IOCSR_MBUF3);
```
kilaterlee, line 279: I suggest it read the parameters several times, maybe 10 times, and if it still doesn't receive them, then it should exit.

kilaterlee (changes requested):

UefiCpuPkg/Library/CpuExceptionHandlerLib/LoongArch/LoongArch64/ArchExceptionHandler.c
MarsDoge: Very good, I will add the reread count to solve this.

PR #11886, opened 2025-12-12
Add CLANGDWARF and CLANGPDB CI

mikebeaton:
# Description

This sequence of commits adds CLANGDWARF and CLANGPDB CI using the `stuart_ci_build` infrastructure already used for GCC CI.

The approach taken here is to disable unit tests and NO-TARGET checks (i.e. uncrustify, cspell, etc.) except in GCC and VS2022, but to run all the same build-and-run tests for CLANGDWARF and CLANGPDB as are currently run for GCC (excepting architectures which the CLANG toolchains do not currently support).

The proposed advantages of running CLANGDWARF and CLANGPDB CI are:

 - It catches real bugs (see for example the second part of [this comment](https://github.com/tianocore/containers/pull/101#issuecomment-3573210480)) which GCC CI alone does not catch
 - It keeps these two supported tool chains working properly

- [ ] Breaking change?
  - NO.
- [ ] Impacts security?
  - NO.
- [ ] Includes tests?
  - NO.

## How This Was Tested

See https://github.com/tianocore/edk2/pull/11861/commits and https://github.com/tianocore/edk2/pull/11860/commits which demonstrate that this PR works, by 'taking over' (for demonstration purposes only) the Ubuntu-GCC.yml files. Note that all builds marked as 'Ubuntu-GCC' in those two draft PRs are CLANGDWARF and CLANGPDB builds respectively, as can be confirmed by clicking through to further details.

## Integration Instructions

In order for the CLANGDWARF and CLANGPDB builds in this PR to execute, the following files would need a pipeline added in https://dev.azure.com/tianocore/edk2-ci:

```
.azurepipelines/Ubuntu-CLANGDWARF.yml
.azurepipelines/Ubuntu-CLANGPDB.yml
OvmfPkg/PlatformCI/.azurepipelines/Ubuntu-CLANGDWARF.yml
OvmfPkg/PlatformCI/.azurepipelines/Ubuntu-CLANGPDB.yml
ArmVirtPkg/PlatformCI/.azurepipelines/Ubuntu-CLANGDWARF.yml
ArmVirtPkg/PlatformCI/.azurepipelines/Ubuntu-CLANGPDB.yml
EmulatorPkg/PlatformCI/.azurepipelines/Ubuntu-CLANGDWARF.yml
```

### DIFF
--- a/BaseTools/Source/C/GenFfs/GenFfs.c
+++ b/BaseTools/Source/C/GenFfs/GenFfs.c
@@ -398,7 +398,7 @@ Returns:
     // Based on section type to adjust offset? Todo
     //
     if ((InputFileAlign [Index] != 0) && (((Size + HeaderSize + TeOffset) % InputFileAlign [Index]) != 0)) {
-      Offset = (Size + sizeof (EFI_COMMON_SECTION_HEADER) + HeaderSize + TeOffset + InputFileAlign [Index] - 1) & ~(InputFileAlign [Index] - 1);
+      Offset = (Size + (UINT32)sizeof (EFI_COMMON_SECTION_HEADER) + HeaderSize + TeOffset + InputFileAlign [Index] - 1) & ~(InputFileAlign [Index] - 1);
       Offset = Offset - Size - HeaderSize - TeOffset;
 
       if (FileBuffer != NULL && ((Size + Offset) < *BufferLength)) {
--- a/BaseTools/Source/C/GenFv/GenFvInternalLib.c
+++ b/BaseTools/Source/C/GenFv/GenFvInternalLib.c
@@ -3176,7 +3176,7 @@ Returns:
         //
         // Only EFI_FFS_FILE_HEADER is needed for a pad section.
         //
-        CurrentOffset = (CurrentOffset + FfsHeaderSize + sizeof(EFI_FFS_FILE_HEADER) + FfsAlignment - 1) & ~(FfsAlignment - 1);
+        CurrentOffset = (CurrentOffset + FfsHeaderSize + sizeof(EFI_FFS_FILE_HEADER) + FfsAlignment - 1) & ~(UINTN)(FfsAlignment - 1);
         CurrentOffset -= FfsHeaderSize;
       }
     }
--- a/BaseTools/Source/C/GenSec/GenSec.c
+++ b/BaseTools/Source/C/GenSec/GenSec.c
@@ -538,7 +538,7 @@ Returns:
       // make sure section data meet its alignment requirement by adding one raw pad section.
       //
       if ((InputFileAlign [Index] != 0) && (((Size + HeaderSize + TeOffset) % InputFileAlign [Index]) != 0)) {
-        Offset = (Size + sizeof (EFI_COMMON_SECTION_HEADER) + HeaderSize + TeOffset + InputFileAlign [Index] - 1) & ~(InputFileAlign [Index] - 1);
+        Offset = (Size + (UINT32)sizeof (EFI_COMMON_SECTION_HEADER) + HeaderSize + TeOffset + InputFileAlign [Index] - 1) & ~(InputFileAlign [Index] - 1);
         Offset = Offset - Size - HeaderSize - TeOffset;
 
         if (FileBuffer != NULL && ((Size + Offset) < *BufferLength)) {
--- a/DynamicTablesPkg/Library/Common/AmlLib/CodeGen/AmlCodeGen.c
+++ b/DynamicTablesPkg/Library/Common/AmlLib/CodeGen/AmlCodeGen.c
@@ -4028,6 +4028,7 @@ AmlCodeGenInvokeMethod (
   }
 
   MethodInvocationNode = NULL;
+  ArgCountNode         = NULL;
 
   NodeStream = AllocateZeroPool (sizeof (AML_NODE_HANDLE) * (NumArgs + 1));
   if (NodeStream == NULL) {
@@ -4167,13 +4168,12 @@ AmlCodeGenInvokeMethod (
     goto exit_handler;
   }
 
-  ArgCountNode = NULL;
-  Status       = AmlCreateDataNode (
-                   EAmlNodeDataTypeUInt,
-                   &NumArgs,
-                   sizeof (UINT8),
-                   &ArgCountNode
-                   );
+  Status = AmlCreateDataNode (
+             EAmlNodeDataTypeUInt,
+             &NumArgs,
+             sizeof (UINT8),
+             &ArgCountNode
+             );
   if (EFI_ERROR (Status)) {
     ASSERT_EFI_ERROR (Status);
     goto exit_handler;
--- a/MdeModulePkg/Bus/Pci/PciBusDxe/PciBusDxe.inf
+++ b/MdeModulePkg/Bus/Pci/PciBusDxe/PciBusDxe.inf
@@ -2,6 +2,7 @@
 #  The PCI bus driver will probe all PCI devices and allocate MMIO and IO space for these devices.
 #  Please use PCD feature flag PcdPciBusHotplugDeviceSupport to enable hot plug supporting.
 #
+#  Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.<BR>
 #  Copyright (c) 2006 - 2021, Intel Corporation. All rights reserved.<BR>
 #
 #  SPDX-License-Identifier: BSD-2-Clause-Patent
@@ -97,6 +98,8 @@
   gEfiMdeModulePkgTokenSpaceGuid.PcdPciDisableBusEnumeration    ## SOMETIMES_CONSUMES
   gEfiMdeModulePkgTokenSpaceGuid.PcdPcieResizableBarSupport     ## CONSUMES
   gEfiMdeModulePkgTokenSpaceGuid.PcdPcieResizableBarMaxSize     ## CONSUMES
+  gEfiMdeModulePkgTokenSpaceGuid.PcdPciCrsRetryIntervalUs       ## CONSUMES
+  gEfiMdeModulePkgTokenSpaceGuid.PcdPciCrsTimeoutSeconds        ## CONSUMES
 
 [UserExtensions.TianoCore."ExtraFiles"]
   PciBusDxeExtra.uni
--- a/MdeModulePkg/Bus/Pci/PciBusDxe/PciEnumeratorSupport.c
+++ b/MdeModulePkg/Bus/Pci/PciBusDxe/PciEnumeratorSupport.c
@@ -1,6 +1,7 @@
 /** @file
   PCI emumeration support functions implementation for PCI Bus module.
 
+Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.<BR>
 Copyright (c) 2006 - 2021, Intel Corporation. All rights reserved.<BR>
 (C) Copyright 2015 Hewlett Packard Enterprise Development LP<BR>
 Copyright (C) 2023 Advanced Micro Devices, Inc. All rights reserved.<BR>
@@ -17,6 +18,18 @@ extern EDKII_DEVICE_SECURITY_PROTOCOL  *mDeviceSecurityProtocol;
 #define EVEN_ALIGN   0xFFFFFFFFFFFFFFFEULL
 #define SQUAD_ALIGN  0xFFFFFFFFFFFFFFFDULL
 #define DQUAD_ALIGN  0xFFFFFFFFFFFFFFFCULL
+
+//
+// Microseconds per second/millisecond for time conversions
+//
+#define MICROSECONDS_PER_SECOND       1000000
+#define MICROSECONDS_PER_MILLISECOND  1000
+
+//
+// Special PCI Vendor ID values per PCIe Base Specification
+//
+#define PCI_VENDOR_ID_NONE  0xFFFF  // No device present
+#define PCI_VENDOR_ID_CRS   0x0001  // Configuration Request Retry Status
 
 /**
   This routine is used to check whether the pci device is present.
@@ -42,6 +55,27 @@ PciDevicePresent (
 {
   UINT64      Address;
   EFI_STATUS  Status;
+  UINTN       CrsRetryCount;
+  UINTN       CrsRetryIntervalUs;
+  UINTN       CrsTimeoutSeconds;
+  UINTN       CrsMaxRetries;
+
+  //
+  // Get CRS retry parameters from PCDs
+  //
+  CrsRetryIntervalUs = PcdGet32 (PcdPciCrsRetryIntervalUs);
+  CrsTimeoutSeconds  = PcdGet32 (PcdPciCrsTimeoutSeconds);
+
+  //
+  // Calculate max retries. CRS retry is disabled if:
+  // - PcdPciCrsTimeoutSeconds is 0 (default)
+  // - PcdPciCrsRetryIntervalUs is 0 (avoid division by zero)
+  //
+  if ((CrsTimeoutSeconds == 0) || (CrsRetryIntervalUs == 0)) {
+    CrsMaxRetries = 0;
+  } else {
+    CrsMaxRetries = (CrsTimeoutSeconds * MICROSECONDS_PER_SECOND) / CrsRetryIntervalUs;
+  }
 
   //
   // Create PCI address map in terms of Bus, Device and Func
@@ -59,17 +93,97 @@ PciDevicePresent (
                                   Pci
                                   );
 
+  if (EFI_ERROR (Status)) {
+    return EFI_NOT_FOUND;
+  }
+
   //
-  // The host bridge may be programmed to accept Configuration Retry Status (CRS).  If the PCI device
-  // is slow, and CRS is enabled, the VendorId may read as 0x0001 when not ready.
-  // This behavior is defined in PCI spec that VendorId is 0x0001.
-  // PCI EXPRESS BASE SPECIFICATION, REV. 3.1 section 2.3.1.
-  // Skip the device, as all the other data read will be invalid.
+  // Check for valid device (not absent and not CRS)
   //
-  if (!EFI_ERROR (Status)) {
-    if (((Pci->Hdr).VendorId != 0xffff) && ((Pci->Hdr).VendorId != 0x0001)) {
+  if (((Pci->Hdr).VendorId != PCI_VENDOR_ID_NONE) && ((Pci->Hdr).VendorId != PCI_VENDOR_ID_CRS)) {
+    //
+    // Valid device found - read the entire config header
+    //
+    Status = PciRootBridgeIo->Pci.Read (
+                                    PciRootBridgeIo,
+                                    EfiPciWidthUint32,
+                                    Address,
+                                    sizeof (PCI_TYPE00) / sizeof (UINT32),
+                                    Pci
+                                    );
+
+    return EFI_SUCCESS;
+  }
+
+  //
+  // Check for no device present
+  //
+  if ((Pci->Hdr).VendorId == PCI_VENDOR_ID_NONE) {
+    return EFI_NOT_FOUND;
+  }
+
+  //
+  // CRS detected (Vendor ID = PCI_VENDOR_ID_CRS)
+  // The host bridge may be programmed to accept Configuration Retry Status (CRS).
+  // If the PCI device is slow, and CRS is enabled, the VendorId reads as PCI_VENDOR_ID_CRS.
+  // This behavior is defined in PCI EXPRESS BASE SPECIFICATION, REV. 3.1 section 2.3.1.
+  //
+
+  //
+  // If CRS retry is disabled, skip the device
+  //
+  if (CrsMaxRetries == 0) {
+    DEBUG ((
+      DEBUG_WARN,
+      "PCI %02x:%02x.%x returned CRS but retry is disabled (PcdPciCrsTimeoutSeconds=%d, PcdPciCrsRetryIntervalUs=%d)\n",
+      Bus,
+      Device,
+      Func,
+      CrsTimeoutSeconds,
+      CrsRetryIntervalUs
+      ));
+    return EFI_NOT_FOUND;
+  }
+
+  //
+  // CRS retry is enabled - wait for device to become ready
+  //
+  DEBUG ((
+    DEBUG_INFO,
+    "PCI %02x:%02x.%x returned CRS, waiting for device ready (timeout %d s)...\n",
+    Bus,
+    Device,
+    Func,
+    CrsTimeoutSeconds
+    ));
+
+  for (CrsRetryCount = 1; CrsRetryCount <= CrsMaxRetries; CrsRetryCount++) {
+    //
+    // Stall before retry
+    //
+    gBS->Stall (CrsRetryIntervalUs);
+
+    //
+    // Re-read the Vendor ID register
+    //
+    Status = PciRootBridgeIo->Pci.Read (
+                                    PciRootBridgeIo,
+                                    EfiPciWidthUint32,
+                                    Address,
+                                    1,
+                                    Pci
+                                    );
+
+    if (EFI_ERROR (Status)) {
+      return EFI_NOT_FOUND;
+    }
+
+    //
+    // Check if device is now ready (no longer returning CRS)
+    //
+    if ((Pci->Hdr).VendorId != PCI_VENDOR_ID_CRS) {
       //
-      // Read the entire config header for the device
+      // Valid device found - read the entire config header
       //
       Status = PciRootBridgeIo->Pci.Read (
                                       PciRootBridgeIo,
@@ -79,11 +193,31 @@ PciDevicePresent (
                                       Pci
                                       );
 
+      DEBUG ((
+        DEBUG_INFO,
+        "PCI %02x:%02x.%x ready after %d CRS retries (%d ms)\n",
+        Bus,
+        Device,
+        Func,
+        CrsRetryCount,
+        (CrsRetryCount * CrsRetryIntervalUs) / MICROSECONDS_PER_MILLISECOND
+        ));
+
       return EFI_SUCCESS;
-    } else if ((Pci->Hdr).VendorId == 0x0001) {
-      DEBUG ((DEBUG_WARN, "CRS response detected.  Devices that return a CRS response during enumeration are currently ignored\n"));
     }
   }
+
+  //
+  // CRS timeout - device never became ready
+  //
+  DEBUG ((
+    DEBUG_ERROR,
+    "PCI %02x:%02x.%x CRS timeout after %d seconds - device not ready\n",
+    Bus,
+    Device,
+    Func,
+    CrsTimeoutSeconds
+    ));
 
   return EFI_NOT_FOUND;
 }
--- a/MdeModulePkg/Core/Dxe/Dispatcher/Dispatcher.c
+++ b/MdeModulePkg/Core/Dxe/Dispatcher/Dispatcher.c
@@ -414,6 +414,7 @@ CoreDispatcher (
     //
     // If the dispatcher is running don't let it be restarted.
     //
+    PERF_FUNCTION_END ();
     return EFI_ALREADY_STARTED;
   }
 
@@ -428,6 +429,7 @@ CoreDispatcher (
              &DxeDispatchEvent
              );
   if (EFI_ERROR (Status)) {
+    PERF_FUNCTION_END ();
     return Status;
   }
 
--- a/MdeModulePkg/Library/DxeCorePerformanceLib/DxeCorePerformanceLib.c
+++ b/MdeModulePkg/Library/DxeCorePerformanceLib/DxeCorePerformanceLib.c
@@ -1540,7 +1540,7 @@ DxeCorePerformanceLibConstructor (
   //
   Status = gBS->CreateEventEx (
                   EVT_NOTIFY_SIGNAL,
-                  TPL_NOTIFY,
+                  TPL_CALLBACK,
                   ReportFpdtRecordBuffer,
                   NULL,
                   &gEfiEndOfDxeEventGroupGuid,
--- a/MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
+++ b/MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
@@ -2072,13 +2072,14 @@ EfiBootManagerBoot (
                    );
         ASSERT_EFI_ERROR (Status);
         );
-
       mBmLegacyBoot (BootOption);
     } else {
       BootOption->Status = EFI_UNSUPPORTED;
     }
 
     PERF_END_EX (gImageHandle, "BdsAttempt", NULL, 0, (UINT32)OptionNumber);
+    PERF_CROSSMODULE_END ("BDS");   // BEGIN is in MdeModulePkg\Universal\BdsDxe\BdsEntry.c
+    PERF_CROSSMODULE_BEGIN ("BDS"); // Keep logging BDS in case of reentry
     return;
   }
 
@@ -2109,6 +2110,8 @@ EfiBootManagerBoot (
   PERF_CODE (
     BmEndOfBdsPerfCode (NULL, NULL);
     );
+  PERF_CROSSMODULE_END ("BDS");     // BEGIN is in MdeModulePkg\Universal\BdsDxe\BdsEntry.c
+  PERF_CROSSMODULE_BEGIN ("BDS");   // Keep logging BDS in case of reentry
 
   REPORT_STATUS_CODE (EFI_PROGRESS_CODE, PcdGet32 (PcdProgressCodeOsLoaderStart));
 
--- a/MdeModulePkg/Library/UefiBootManagerLib/InternalBm.h
+++ b/MdeModulePkg/Library/UefiBootManagerLib/InternalBm.h
@@ -69,7 +69,6 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Library/DxeServicesLib.h>
 #include <Library/ReportStatusCodeLib.h>
 #include <Library/CapsuleLib.h>
-#include <Library/PerformanceLib.h>
 #include <Library/HiiLib.h>
 #include <Library/VariablePolicyHelperLib.h>
 
--- a/MdeModulePkg/Library/UefiBootManagerLib/UefiBootManagerLib.inf
+++ b/MdeModulePkg/Library/UefiBootManagerLib/UefiBootManagerLib.inf
@@ -59,7 +59,6 @@
   MemoryAllocationLib
   DxeServicesLib
   ReportStatusCodeLib
-  PerformanceLib
   HiiLib
   SortLib
   VariablePolicyHelperLib
--- a/MdeModulePkg/MdeModulePkg.dec
+++ b/MdeModulePkg/MdeModulePkg.dec
@@ -3,7 +3,7 @@
 # It also provides the definitions(including PPIs/PROTOCOLs/GUIDs and library classes)
 # and libraries instances, which are used for those modules.
 #
-# Copyright (c) 2019, NVIDIA CORPORATION. All rights reserved.
+# Copyright (c) 2019 - 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.<BR>
 # Copyright (c) 2007 - 2024, Intel Corporation. All rights reserved.<BR>
 # Copyright (c) 2016, Linaro Ltd. All rights reserved.<BR>
 # (C) Copyright 2016 - 2019 Hewlett Packard Enterprise Development LP<BR>
@@ -2324,6 +2324,20 @@
   # 43          2^(43+20) = 8 EB
   # @Prompt Max size to accept in PCIe Resizable BAR Capability register.
   gEfiMdeModulePkgTokenSpaceGuid.PcdPcieResizableBarMaxSize|43|UINT8|0x10000030
+
+  ## PCIe Configuration Request Retry Status (CRS) retry interval in microseconds.
+  #  When a PCIe device returns CRS during enumeration, the PCI bus driver will wait
+  #  this interval before retrying. Default is 10000us (10ms) to match Linux kernel.
+  #  If set to 0, CRS retry is disabled regardless of PcdPciCrsTimeoutSeconds.
+  # @Prompt PCI CRS retry interval in microseconds.
+  gEfiMdeModulePkgTokenSpaceGuid.PcdPciCrsRetryIntervalUs|10000|UINT32|0x1000002A
+
+  ## PCIe Configuration Request Retry Status (CRS) timeout in seconds.
+  #  Maximum time to wait for a device returning CRS to become ready.
+  #  When set to 0 (default), devices returning CRS are skipped during enumeration.
+  #  Platforms requiring CRS retry support should set a non-zero value.
+  # @Prompt PCI CRS timeout in seconds.
+  gEfiMdeModulePkgTokenSpaceGuid.PcdPciCrsTimeoutSeconds|0|UINT32|0x1000002B
 
 [PcdsPatchableInModule]
   ## Specify memory size with page number for PEI code when
--- a/MdeModulePkg/Universal/Acpi/FirmwarePerformanceDataTableDxe/FirmwarePerformanceDxe.c
+++ b/MdeModulePkg/Universal/Acpi/FirmwarePerformanceDataTableDxe/FirmwarePerformanceDxe.c
@@ -686,7 +686,7 @@ FirmwarePerformanceDxeEntryPoint (
   //
   Status = gBS->CreateEventEx (
                   EVT_NOTIFY_SIGNAL,
-                  TPL_NOTIFY,
+                  TPL_CALLBACK,
                   FpdtEndOfDxeEventNotify,
                   NULL,
                   &gEfiEndOfDxeEventGroupGuid,
--- a/OvmfPkg/IntelTdx/README.md
+++ b/OvmfPkg/IntelTdx/README.md
@@ -55,25 +55,40 @@ There are 2 configurations for TDVF.
      loaded.
 
 Build
-------
-- Build the TDVF (Config-A) target:
-```
-cd /path/to/edk2
-source edksetup.sh
+-----
+> For the build environment, consider reusing the existing EDKII infrastructure for containerized builds.
+> See [tianocore/containers](https://github.com/tianocore/containers) for more details.
 
-## CC_MEASUREMENT disabled
-build -p OvmfPkg/OvmfPkgX64.dsc -a X64 -t GCC -D CC_MEASUREMENT_ENABLE=FALSE -b RELEASE
+1. **(Optional)** Pull and run the tianocore/containers Ubuntu 22 Docker image:
+   ```bash
+   docker run -it \
+     -v "${HOME}":"${HOME}" \
+     -e EDK2_DOCKER_USER_HOME="${HOME}" \
+     ghcr.io/tianocore/containers/ubuntu-22-dev:latest /bin/bash
+   ```
 
-## CC_MEASUREMENT enabled
-build -p OvmfPkg/OvmfPkgX64.dsc -a X64 -t GCC -D CC_MEASUREMENT_ENABLE=TRUE -b RELEASE
-```
+2. Prepare the environment:
+   ```bash
+   cd /path/to/edk2
+   . ./edksetup.sh
+   make -C BaseTools
+   ```
 
-- Build the TDVF (Config-B) target:
-```
-cd /path/to/edk2
-source edksetup.sh
-build -p OvmfPkg/IntelTdx/IntelTdxX64.dsc -a X64 -t GCC -b RELEASE
-```
+3. Build the TDVF target:
+
+   **Config-A** (OvmfPkgX64):
+   ```bash
+   # CC_MEASUREMENT disabled
+   build -p OvmfPkg/OvmfPkgX64.dsc -a X64 -t GCC -D CC_MEASUREMENT_ENABLE=FALSE -b RELEASE
+
+   # CC_MEASUREMENT enabled
+   build -p OvmfPkg/OvmfPkgX64.dsc -a X64 -t GCC -D CC_MEASUREMENT_ENABLE=TRUE -b RELEASE
+   ```
+
+   **Config-B** (IntelTdxX64):
+   ```bash
+   build -p OvmfPkg/IntelTdx/IntelTdxX64.dsc -a X64 -t GCC -b RELEASE
+   ```
 
 Usage
 -----
@@ -81,7 +96,7 @@ Usage
 Assuming TDX-QEMU/TDX-KVM are already built, one can start a TD virtual
 machine as [launching-a-tdx-vm](https://gitlab.com/qemu/qemu/-/blob/master/docs/system/i386/tdx.rst):
 
-```
+```bash
 qemu_system_x86 \
    -accel kvm \
    -cpu host \
--- a/OvmfPkg/OvmfPkgX64.dsc
+++ b/OvmfPkg/OvmfPkgX64.dsc
@@ -35,6 +35,7 @@
   DEFINE STANDALONE_MM_ENABLE    = FALSE
   DEFINE SOURCE_DEBUG_ENABLE     = FALSE
   DEFINE CC_MEASUREMENT_ENABLE   = TRUE
+  DEFINE DEBUG_TO_MEM            = FALSE
 
 !include OvmfPkg/Include/Dsc/OvmfTpmDefines.dsc.inc
 
--- a/SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
+++ b/SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
@@ -46,8 +46,6 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Library/ReportStatusCodeLib.h>
 #include <Library/Tcg2PhysicalPresenceLib.h>
 
-#define PERF_ID_TCG2_DXE  0x3120
-
 typedef struct {
   CHAR16      *VariableName;
   EFI_GUID    *VendorGuid;
@@ -2502,7 +2500,7 @@ OnReadyToBoot (
   EFI_STATUS    Status;
   TPM_PCRINDEX  PcrIndex;
 
-  PERF_START_EX (mImageHandle, "EventRec", "Tcg2Dxe", 0, PERF_ID_TCG2_DXE);
+  PERF_FUNCTION_BEGIN ();
   if (mBootAttempts == 0) {
     //
     // Measure handoff tables.
@@ -2583,7 +2581,7 @@ OnReadyToBoot (
   // Increase boot attempt counter.
   //
   mBootAttempts++;
-  PERF_END_EX (mImageHandle, "EventRec", "Tcg2Dxe", 0, PERF_ID_TCG2_DXE + 1);
+  PERF_FUNCTION_END ();
 }
 
 /**
--- a/SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
+++ b/SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
@@ -41,8 +41,6 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Library/ResetSystemLib.h>
 #include <Library/PrintLib.h>
 
-#define PERF_ID_TCG2_PEI  0x3080
-
 typedef struct {
   EFI_GUID                     *EventGuid;
   EFI_TCG2_EVENT_LOG_FORMAT    LogFormat;
@@ -839,7 +837,7 @@ MeasureMainBios (
   EFI_FV_INFO                  VolumeInfo;
   EFI_PEI_FIRMWARE_VOLUME_PPI  *FvPpi;
 
-  PERF_START_EX (mFileHandle, "EventRec", "Tcg2Pei", 0, PERF_ID_TCG2_PEI);
+  PERF_FUNCTION_BEGIN ();
 
   //
   // Only measure BFV at the very beginning. Other parts of Static Core Root of
@@ -870,7 +868,7 @@ MeasureMainBios (
 
   Status = MeasureFvImage ((EFI_PHYSICAL_ADDRESS)(UINTN)VolumeInfo.FvStart, VolumeInfo.FvSize);
 
-  PERF_END_EX (mFileHandle, "EventRec", "Tcg2Pei", 0, PERF_ID_TCG2_PEI + 1);
+  PERF_FUNCTION_END ();
 
   return Status;
 }


mikebeaton:
cc @osteffenrh (since I think this was possibly [of interest to yourself](https://github.com/tianocore/containers/pull/101#issuecomment-3573802318), too?)

mikebeaton:
@mdkinney @os-d @makubacki - As for my previous, considerably smaller, [CI offering](https://github.com/tianocore/edk2/pull/11857) - is this of interest? :-)

@mdkinney - I'm not sure how best to address the fact that at a small number of hours before independently finishing this, [it transpired](https://github.com/tianocore/edk2/pull/11877#issuecomment-3647474704) that this may well overlap with work you (singular, and/or plural) are doing in the background. Since this is finished, working AFAIK, and since I've been working on it on and off for a year, and intensively for the last month (incl. remaining fixes elsewhere for CLANG builds to work), I thought I'd put it up for review anyway.

The main reasoning for offering this is that we (Acidanthera) and I would really like to see CLANG CI incorporated - not least so we no longer have to fix and PR back to upstream the several CLANG build bugs which have always been unintentionally introduced every time we rebase. (Beyond that of course, and as I think everyone is now aware, these often correspond to real bugs which could have been caught automatically before being merged.) Therefore, in the end, we are most keen to see CLANG CI (with as full a coverage as reasonably possible, of course) rather than tied to any particular implementation of it. (But then again we, or certainly I(!), would be perfectly happpy for this implementation to be reviewed and merged, if it's considered good enough!)

os-d:
> @mdkinney @os-d @makubacki - As for my previous, considerably smaller, [CI offering](https://github.com/tianocore/edk2/pull/11857) - is this of interest? :-)

Yes, thank you for the contribution! My main comment is the same as yours, that I know @mdkinney is working on a very similar effort to bring clang CI (and many associated fixes/updates), so we should make sure these two efforts are aligned.

I would also be interested in seeing clangpdb builds from Windows (which I believe Mike Kinney has), particularly for ARM64, since MSVC ARM64 is not heavily used and has some limitations.

For my understanding, what is the value of using CLANGDWARF over CLANGPDB? They are just a single flag different, AFAIK, i.e. we set them up here to use different syntax, etc. Trying to understand for the additional maintenance burden of running CLANGDWARF in CI. From my limited perspective, it seems like it would always be ideal for clang to directly produce PE/COFF images instead of having the spaghetti of our BaseTools ELF -> PE/COFF conversion code. Both can be built from Linux, is the CLANGDWARF desire just for supporting existing tools that rely on dwarf symbols, etc. instead of moving to tools that support PDBs, etc.? @ardbiesheuvel @leiflindholm I believe you are both CLANGDWARF users, do you mind enlightening me?

mikebeaton:
@os-d - Thanks!

Worth noting that CLANGDWARF can currently build _everything_ using clang, whereas CLANGPDB cannot (currently?) build AARCH64, RISCV64, LOONGARCH64.

Also wanted to note that apparently @mdkinney said his work on this was using GitHub actions, whereas I assumed that would almost certainly be rejected by yourselves since you currently use `stuart` with Azure Pipelines. Happy to take a steer on what's going on there.

os-d:
> Worth noting that CLANGDWARF can currently build _everything_ using clang, whereas CLANGPDB cannot (currently?) build AARCH64, RISCV64, LOONGARCH64.
> 

CLANGPDB can build AARCH64 (I don't know about RISCV64 and LOONGARCH64), but edk2 currently uses an elf directive in asm files (I always forget which, have to try it out again) which CLANGPDB can't use. We've talked for a while about addressing that by having a macro that inserts the directive when building for elf and doesn't for pdb. We've been using CLANGPDB for AARCH64 successfully for a couple years now, with a very hacky workaround for the assembly issue that we have for MSVC ARM64 support and are looking to drop :)

Javagedes:
> @os-d - Thanks!
> 
> Worth noting that CLANGDWARF can currently build _everything_ using clang, whereas CLANGPDB cannot (currently?) build AARCH64, RISCV64, LOONGARCH64.
> 
> Also wanted to note that apparently @mdkinney said his work on this was using GitHub actions, whereas I assumed that would almost certainly be rejected by yourselves since you currently use `stuart` with Azure Pipelines. Happy to take a steer on what's going on there.

We are actively looking to shift a subset of pipelines to github workflows to help reduce the queues we have due to the large amount of CI that runs per PR / merge. You can see we already have one simple reusable workflow to build platforms using stuart:

https://github.com/tianocore/edk2/blob/master/.github/workflows/BuildPlatform.yml

mhaeuser:
@os-d I think the magic is right in the names, PDB vs DWARF. :) From what I recall, people were not happy with PDB support outside the Windows/MSVC ecosystem and thus this approach was taken.

Regarding the AARCH64 issue, I recall there was some issue with StandaloneMmCore assuming ELF PIE, which is different from how MSVC does PIE. Was that addressed already in some form?

os-d:
> @os-d I think the magic is right in the names, PDB vs DWARF. :) From what I recall, people were not happy with PDB support outside the Windows/MSVC ecosystem and thus this approach was taken.
> 

Sure, my point being that the BaseTools conversion of ELF -> PE is the worst of all worlds :). I did think that clang/llvm provided a lot of Linux tools here, too, e.g. llvm-objdump, etc. that had good support for PE.

> Regarding the AARCH64 issue, I recall there was some issue with StandaloneMmCore assuming ELF PIE, which is different from how MSVC does PIE. Was that addressed already in some form?

This one doesn't ring a bell. We've been using CLANGPDB successfully with an AARCH64 platform that uses Standalone MM, so something must have been fixed :)

mdkinney:
I will write up a complete Wiki on TianoCore clang topics to help clarify these topics.

One aspect that is important here is that there are 2 linkers.  CLANGPDB uses lld-link that can directly generate PE/COFF images.  CLANGDWARF uses ld that generates Native OS images.  When clang support was first added to TianoCore, only the lld-link linker could produce PDB symbol files.  Hence the name CLANGPDB.  Since then, the clang tools have been enhanced so both DWARF and PDB symbols can be produced in a single build with a .pdb file generated and .debug section are produced. There may be advantages to producing both types of symbols.

In general, CLANGPDB is the best choice for Firmware builds to directly produce UEFI/PI conformant images. Not all CPU archs in tools_def support CLANGPDB, so some Firmware builds use CLANGDWARF with the ELF->PE/COFF conversion in GenFw.

CLANGDWARF is the best choice when a native OS application needs to be generated such as host-based unit tests from the UnitTestFrameworkPkg and the EmulatorPkg.

mhaeuser:
> Sure, my point being that the BaseTools conversion of ELF -> PE is the worst of all worlds :)

GenFw is a bit nasty, but in general, I can’t really agree. I worked *a lot* with the image stacks (and authored a lot of Acidanthera tooling for ELF and Mach-O conversion to PE) and the MSVC world gave me the worst headaches by far.

> I did think that clang/llvm provided a lot of Linux tools here, too, e.g. llvm-objdump, etc. that had good support for PE.

Last time I used them, there were issues both with PDB generation and consumption. It’s been too long to remember the details, but both CLANGDWARF and XCODE5 were easier to work with regarding tooling and RE (I think IDA was one of the pain points with CLANGPDB). Might all be fixed by now, however. PE itself was quite fine IIRC.

> This one doesn't ring a bell. We've been using CLANGPDB successfully with an AARCH64 platform that uses Standalone MM, so something must have been fixed :)

This: https://github.com/tianocore/edk2/commit/210869834639c1150b1f016e2022472a55cdd884

mikebeaton:
Removing my previous comment and engaging brain before writing, CLANGPDB definitely _does_ build StandaloneMmPkg.dsc - even though the commit and comment which Marvin highlighted are still current in the codebase - it's in my demo PR https://github.com/tianocore/edk2/pull/11860 (TARGET_OTHER).

mikebeaton:
> Removing my previous comment and engaging brain before writing, CLANGPDB definitely _does_ build StandaloneMmPkg.dsc - even though the commit and comment which Marvin highlighted are still current in the codebase - it's in my demo PR #11860 (TARGET_OTHER).

Sorry, no - that's X64 only. :-/

mikebeaton:
I've been avoiding mentioning the elephant in the room, under the assumption that I must be missing something obvious - but the elephant in the room is that (_unless_ I'm missing something obvious), CLANGPDB doesn't build AARCH64 (or RISCV64, or LOONGARCH64) at all, failing early with:

```
build.py...
 : error F000: CC_PATH doesn't exist in CLANGPDB ToolChain and AARCH64 Arch.
	[/home/vsts/w/edk2/MdePkg/Library/StandaloneMmServicesTableLib/StandaloneMmServicesTableLib.inf]
```

os-d:
> I've been avoiding mentioning the elephant in the room, under the assumption that I must be missing something obvious - but the elephant in the room is that (_unless_ I'm missing something obvious), CLANGPDB doesn't build AARCH64 (or RISCV64, or LOONGARCH64) at all, failing early with:
> 
> ```
> build.py...
>  : error F000: CC_PATH doesn't exist in CLANGPDB ToolChain and AARCH64 Arch.
> 	[/home/vsts/w/edk2/MdePkg/Library/StandaloneMmServicesTableLib/StandaloneMmServicesTableLib.inf]
> ```

Ah, we have a commit that hasn't been upstreamed that I thought was (it may have been sent out but not reviewed, not sure). I can submit this: https://github.com/microsoft/mu_basecore/commit/af03dfe9e75e3690df94ecb12d28739c8c7b6b02.

os-d:
> This: [2108698](https://github.com/tianocore/edk2/commit/210869834639c1150b1f016e2022472a55cdd884)

I do agree this should have been breaking us and tbh I'm not sure why it wasn't complaining about getting the option it doesn't know about. I'll try to take a look again when I have some time.

os-d:
> > This: [2108698](https://github.com/tianocore/edk2/commit/210869834639c1150b1f016e2022472a55cdd884)
> 
> I do agree this should have been breaking us and tbh I'm not sure why it wasn't complaining about getting the option it doesn't know about. I'll try to take a look again when I have some time.

Okay, rolling back to when we were building our QEMU SBSA platform in CI (had to disable because TF-A dropped Windows support for clang), it looks like clang happily takes -fpie even under the CLANDPDB flavor, so that doesn't explicitly fail it.

As for actually needing a PIE St MM Core on ARM64, do we actually? All the ARM64 platforms I am aware of use a fixed address for StMM Core because it is required for TF-A/Hafnium to load it, any way. PIE would work as well, but still has to be loaded at that fixed address.

mhaeuser:
> As for actually needing a PIE St MM Core on ARM64, do we actually? All the ARM64 platforms I am aware of use a fixed address for StMM Core because it is required for TF-A/Hafnium to load it, any way. PIE would work as well, but still has to be loaded at that fixed address.

Now that’s interesting. I asked the same question years ago, because self-relocation is a major pain in every regard and I wanted to avoid implementing it in my PE loader replacement, but I was told indeed for some platforms the address is not known at compile-time. I vaguely remember it might have been @ardbiesheuvel I had that conversation with - and it would make sense, as he introduced the relocation code: https://github.com/tianocore/edk2/commit/493f2c69311806a214de2b3db7fdf7bc9162ea81

If the code ever triggered without ELF-like PIE support, it may fault when writing to .text if it was made read-only by the permissions code prior to relocation.

os-d:
> If the code ever triggered without ELF-like PIE support, it may fault when writing to .text if it was made read-only by the permissions code prior to relocation.

The ARM64 StMM entry point lib manages the section permissions to avoid this: https://github.com/tianocore/edk2/blob/786de8636a28affa76de24cea3ed48da0393841b/ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/SetPermissions.c#L44.

ardbiesheuvel:
First of all, I fully agree that using a native PE/COFF toolchain is always preferred, but the debugging story on Linux is indeed a bit problematic, and there are some other issues that we need to solve. 

> > > This: [2108698](https://github.com/tianocore/edk2/commit/210869834639c1150b1f016e2022472a55cdd884)
> > 
> > 
> > I do agree this should have been breaking us and tbh I'm not sure why it wasn't complaining about getting the option it doesn't know about. I'll try to take a look again when I have some time.
> 
> Okay, rolling back to when we were building our QEMU SBSA platform in CI (had to disable because TF-A dropped Windows support for clang), it looks like clang happily takes -fpie even under the CLANDPDB flavor, so that doesn't explicitly fail it.
> 
> As for actually needing a PIE St MM Core on ARM64, do we actually? All the ARM64 platforms I am aware of use a fixed address for StMM Core because it is required for TF-A/Hafnium to load it, any way. PIE would work as well, but still has to be loaded at that fixed address.

There are platforms that incorporate standalone MM as a pseudo-TA (trusted application) in OP-TEE, and this requires runtime relocation. This is currently based on ELF, which is why we need to build with PIE. This is platform specific, though, and the self-relocation code only takes effect when the build time and runtime load addresses differ. So this is not a blocker for the package, only for the platforms in question, and this is acceptable imo.

Another dependency we have on GenFw is the patching of ADRP instructions into ADR in builds that use the GCC 'small' code model (as opposed to the 'tiny' one.) ADRP instructions return relative references rounded down to 4k, followed by an ADD/LDR instruction supplying the lower 12 bits, and this implies a minimum section alignment of 4k, which results in a lot of bloat for XIP binaries. ADR instructions have no such requirement, but they have a range of only -/+ 1 MiB, and so they cannot be used everywhere. (Also, at the time, the 'tiny' code model was only supported by GCC, not by Clang, but this appears to have changed in the meantime.)

We may be able to rely on more recent linkers performing this relaxation directly, but we'd have to check whether this is reflected correctly in the section alignment. I.e., using `lld-link` and setting a section alignment < 4k should either fail, or replace all ADRP/ADD pairs with ADR.

mhaeuser:
> The ARM64 StMM entry point lib manages the section permissions to avoid this

No, text is R-X, thus relocation would fault. Just a tiny correction to my previous comment, it would fault either way, as it starts out as R-X and the linked function won’t even touch it.

ELF avoids this, because PIE doesn’t allow relocs in .text, instead indirecting them via GOT (which still needs RELRO :( ).

os-d:
> > The ARM64 StMM entry point lib manages the section permissions to avoid this
> 
> No, text is R-X, thus relocation would fault. Just a tiny correction to my previous comment, it would fault either way, as it starts out as R-X and the linked function won’t even touch it.
> 
> ELF avoids this, because PIE doesn’t allow relocs in .text, instead indirecting them via GOT (which still needs RELRO :( ).

Oh, I see what you are saying. We have had issues when we tried the reloc path as opposed to executing from a fixed address without relocs.

mikebeaton:
> We are actively looking to shift a subset of pipelines to github workflows to help reduce the queues we have due to the large amount of CI that runs per PR / merge. You can see we already have one simple reusable workflow to build platforms using stuart:
> 
> https://github.com/tianocore/edk2/blob/master/.github/workflows/BuildPlatform.yml

@Javagedes - Thanks!

If I can ask (also @mdkinney, if I may), what's the more detailed plan? To move all CI eventually, or just some? Also, to still use stuart presumably, even in GitHub actions?

Also, what's the overall rationale? We (Acidanthera) already use GitHub actions to build and test run firmware on GCC, CLANGPDB, CLANGDWARF, VS2022, XCODE5 (with traditional EDK2 builds, marshalled by much lighter weight scripts than stuart) - I haven't noticed this process being obviously slower or faster for the amount of work done than Azure Pipelines however. Both sometimes hit queueing issues and sometimes don't, and I believe that when all's said and done, Microsoft own and run the infrastructure for both, too.

So it's not clear to me that/how/why _just_ speed/queueing issues amount to a good reason to change horses? Once again, more than happy to take a steer on the reasoning here, if that's available!

ardbiesheuvel:
> > I've been avoiding mentioning the elephant in the room, under the assumption that I must be missing something obvious - but the elephant in the room is that (_unless_ I'm missing something obvious), CLANGPDB doesn't build AARCH64 (or RISCV64, or LOONGARCH64) at all, failing early with:
> > ```
> > build.py...
> >  : error F000: CC_PATH doesn't exist in CLANGPDB ToolChain and AARCH64 Arch.
> > 	[/home/vsts/w/edk2/MdePkg/Library/StandaloneMmServicesTableLib/StandaloneMmServicesTableLib.inf]
> > ```
> 
> Ah, we have a commit that hasn't been upstreamed that I thought was (it may have been sent out but not reviewed, not sure). I can submit this: [microsoft/mu_basecore@af03dfe](https://github.com/microsoft/mu_basecore/commit/af03dfe9e75e3690df94ecb12d28739c8c7b6b02).

It would be great if we could get these changes upstream, yes.

However, I did manage to generate a broken binary with Clang PDB: building the PEI core with this CLANGPDB setup results in an image with 32 byte section alignment, i.e.

```
File: /home/ardb/build/edk2-workspace/Build/MdeModule/DEBUG_CLANGPDB/AARCH64/MdeModulePkg/Core/Pei/PeiMain/OUTPUT/PeiCore.efi
Format: COFF-ARM64
Arch: aarch64
AddressSize: 64bit
ImageFileHeader {
  Machine: IMAGE_FILE_MACHINE_ARM64 (0xAA64)
  SectionCount: 4
  TimeDateStamp: 1970-01-01 00:00:00 (0x0)
  PointerToSymbolTable: 0x0
  SymbolCount: 0
  StringTableSize: 0
  OptionalHeaderSize: 240
  Characteristics [ (0x2022)
    IMAGE_FILE_DLL (0x2000)
    IMAGE_FILE_EXECUTABLE_IMAGE (0x2)
    IMAGE_FILE_LARGE_ADDRESS_AWARE (0x20)
  ]
}
ImageOptionalHeader {
  Magic: 0x20B
  MajorLinkerVersion: 14
  MinorLinkerVersion: 0
  SizeOfCode: 28704
  SizeOfInitializedData: 1600
  SizeOfUninitializedData: 0
  AddressOfEntryPoint: 0x7210
  BaseOfCode: 0x220
  ImageBase: 0x0
  SectionAlignment: 32
  FileAlignment: 32
  ...
```

However, the produced image does contain numerous ADRP/ADD pairs, and so loading this PEI core at an alignment < 4k is guaranteed to produce a crash. For example, we have this at the start of the image

```
0000000000000220 <.text>:
...
     290: f0000028      adrp    x8, 0x7000 <.text+0x6de0>
     294: aa0003e1      mov     x1, x0
     298: 910023e2      add     x2, sp, #0x8
     29c: f943dd08      ldr     x8, [x8, #0x7b8]
```

The instructions at 0x290 and 0x29c combined load the 64-bit quantity at address <base address + 0x77b8> into register x8. However, the base address in this equation is truncated, and so this only works if the base address is 4k aligned. I.e., when loading this image at 32 bytes past a 4k aligned boundary, the load will be off by 32 bytes.

This is why we currently rely on GenFw, as the ELF tooling has similar issues. This needs to be fixed in the CLANGPDB toolchain before we can switch to it, unless we disable the small code model or use 4k section alignment as a global minimum.

os-d:
> It would be great if we could get these changes upstream, yes.
> 

I'll get a PR up today.

> However, the produced image does contain numerous ADRP/ADD pairs, and so loading this PEI core at an alignment < 4k is guaranteed to produce a crash. For example, we have this at the start of the image
> 
> ```
> 0000000000000220 <.text>:
> ...
>      290: f0000028      adrp    x8, 0x7000 <.text+0x6de0>
>      294: aa0003e1      mov     x1, x0
>      298: 910023e2      add     x2, sp, #0x8
>      29c: f943dd08      ldr     x8, [x8, #0x7b8]
> ```
> 
> The instructions at 0x290 and 0x29c combined load the 64-bit quantity at address <base address + 0x77b8> into register x8. However, the base address in this equation is truncated, and so this only works if the base address is 4k aligned. I.e., when loading this image at 32 bytes past a 4k aligned boundary, the load will be off by 32 bytes.
> 

I'm trying to understand why we never hit an issue using CLANGPDB on QEMU SBSA, which uses PEI on (with 32 byte section alignment). We should have often seen this issue because each PEIM (and PEI Core) would have had to been aligned on 4KB boundaries to succeed, which would have been rare.

Forgive my obtuseness, but why does ADRP require 4KB alignment? You said that the above loads <base address + 0x77b8> and that you need the base address to be 4KB aligned, since it gets truncated, but doesn't ADRP do a PC relative add?

image

image

So the above would be <.text + 0x290 (e.g. PC for the ADRP)> loaded into x8 and then the ldr adds 0x7b8 to this. So I would anticipate that the address it is trying to calculate is 0x77b8 from that ADRP instruction. Is that not the case?

Is there a fatal flaw in my interpretation? I would expect nothing less :)

> This is why we currently rely on GenFw, as the ELF tooling has similar issues. This needs to be fixed in the CLANGPDB toolchain before we can switch to it, unless we disable the small code model or use 4k section alignment as a global minimum.

Assuming there is a flaw in my understanding above and we do need to work around this, I think that CLANGPDB AARCH64 can be put up with a 4k section alignment. Many (if not most) ARM64 platforms don't use PEI, we already expect DXE and StMM to be 4k aligned for memory protections to be applied, so this would just affect SEC for many platforms, which is fairly small.

@ardbiesheuvel would you be willing to file an issue on clang to fix this? It will take awhile of course, but would be good to have tracked and some day fixed.

I'm not well versed enough in the tradeoffs between the tiny code and small code models to weigh in there. Should we explore using tiny? It seems like that would have issues for symbols that are more than 1MB apart? Or just be slower to access those?

os-d:
> > It would be great if we could get these changes upstream, yes.
> 
> I'll get a PR up today.
> 
> > However, the produced image does contain numerous ADRP/ADD pairs, and so loading this PEI core at an alignment < 4k is guaranteed to produce a crash. For example, we have this at the start of the image
> > ```
> > 0000000000000220 <.text>:
> > ...
> >      290: f0000028      adrp    x8, 0x7000 <.text+0x6de0>
> >      294: aa0003e1      mov     x1, x0
> >      298: 910023e2      add     x2, sp, #0x8
> >      29c: f943dd08      ldr     x8, [x8, #0x7b8]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The instructions at 0x290 and 0x29c combined load the 64-bit quantity at address <base address + 0x77b8> into register x8. However, the base address in this equation is truncated, and so this only works if the base address is 4k aligned. I.e., when loading this image at 32 bytes past a 4k aligned boundary, the load will be off by 32 bytes.
> 
> I'm trying to understand why we never hit an issue using CLANGPDB on QEMU SBSA, which uses PEI on (with 32 byte section alignment). We should have often seen this issue because each PEIM (and PEI Core) would have had to been aligned on 4KB boundaries to succeed, which would have been rare.
> 
> Forgive my obtuseness, but why does ADRP require 4KB alignment? You said that the above loads <base address + 0x77b8> and that you need the base address to be 4KB aligned, since it gets truncated, but doesn't ADRP do a PC relative add?
> 
> image image
> So the above would be <.text + 0x290 (e.g. PC for the ADRP)> loaded into x8 and then the ldr adds 0x7b8 to this. So I would anticipate that the address it is trying to calculate is 0x77b8 from that ADRP instruction. Is that not the case?
> 
> Is there a fatal flaw in my interpretation? I would expect nothing less :)
> 

Ah, oops, I do my fatal flaw here :), it isn't <.text + 0x290> because the 0x290 gets shifted out, so it is just .text + 0x77b8, which is what you are saying. And then in that case, the toolchain doesn't know the distance `y` below, but does know `x`

```text
   |>   page_start
   |
 y |    .text_start
   |
   |>  pc @ ADRP
   |
 x |
   |
   |>  Destination
```

Oh, and you know what, we did hit this once, on an ELF image and did have to take your GenFW fix, I am now dimly recalling. I am surprised that the ARM ARM doesn't call out ADRP as requiring the image to be page aligned and that clang lets this be generated.

Let me know on what you think on requiring 4KB section alignment vs the tiny code model and I'll put up the PR. And again, if you want to file a clang issue for this, that would be helpful (if you think there is a reasonable answer of what they should do, whether basing ADRP usage on multiple of 4KB section alignment).

ardbiesheuvel:
> > > It would be great if we could get these changes upstream, yes.
> > 
> > 
> > I'll get a PR up today.
> > > However, the produced image does contain numerous ADRP/ADD pairs, and so loading this PEI core at an alignment < 4k is guaranteed to produce a crash. For example, we have this at the start of the image
> > > ```
> > > 0000000000000220 <.text>:
> > > ...
> > >      290: f0000028      adrp    x8, 0x7000 <.text+0x6de0>
> > >      294: aa0003e1      mov     x1, x0
> > >      298: 910023e2      add     x2, sp, #0x8
> > >      29c: f943dd08      ldr     x8, [x8, #0x7b8]
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > The instructions at 0x290 and 0x29c combined load the 64-bit quantity at address <base address + 0x77b8> into register x8. However, the base address in this equation is truncated, and so this only works if the base address is 4k aligned. I.e., when loading this image at 32 bytes past a 4k aligned boundary, the load will be off by 32 bytes.
> > 
> > 
> > I'm trying to understand why we never hit an issue using CLANGPDB on QEMU SBSA, which uses PEI on (with 32 byte section alignment). We should have often seen this issue because each PEIM (and PEI Core) would have had to been aligned on 4KB boundaries to succeed, which would have been rare.
> > Forgive my obtuseness, but why does ADRP require 4KB alignment? You said that the above loads <base address + 0x77b8> and that you need the base address to be 4KB aligned, since it gets truncated, but doesn't ADRP do a PC relative add?
> > image image
> > So the above would be <.text + 0x290 (e.g. PC for the ADRP)> loaded into x8 and then the ldr adds 0x7b8 to this. So I would anticipate that the address it is trying to calculate is 0x77b8 from that ADRP instruction. Is that not the case?
> > Is there a fatal flaw in my interpretation? I would expect nothing less :)
> 
> Ah, oops, I do my fatal flaw here :), it isn't <.text + 0x290> because the 0x290 gets shifted out, so it is just .text + 0x77b8, which is what you are saying. And then in that case, the toolchain doesn't know the distance `y` below, but does know `x`
> 

Indeed. The key insight here is that ADRP always produces a value that has bits [11:0] cleared, and these are filled out by the subsequent ADD (or LDR). This means that the variable to which the ADRP/ADD combo refers *must* live at an address which equals those lower 12 bits modulo 4k. IOW, ADRP/ADD does not require alignment per se, but it requires that the placement modulo 4k does not change.

This is what makes the ELF handling tricky, given that ELF sections (i.e., the ones serving as input to the linker) are being used to construct PE/COFF view of the image, and those sections don't have the same requirement, given that the ADRP/ADD opcode fields have not been populated yet. Only when the loadable segments are created is there a need to uphold this requirement, because that is when the ADD immediate field gets set.

So I'd imagine there is a similar concern for native PE/COFF: it should really be the linker that, as it populates those ADD opcode immediate fields, checks the requested alignment for the executable, and/or sets it to 4k if no specific value was requested (and the default is < 4k).

Another concern we have is that, apparently, the FDF rules system is broken, given that its `Align=` directives can reduce the alignment of an executable, and it should only be able to increase the minimum. This means that currently, the only sane value for `Align` = `auto`.

> Let me know on what you think on requiring 4KB section alignment vs the tiny code model and I'll put up the PR. And again, if you want to file a clang issue for this, that would be helpful (if you think there is a reasonable answer of what they should do, whether basing ADRP usage on multiple of 4KB section alignment).

The tiny code model is not usable IIRC, because the Shell ends up being substantially larger than 1 MiB. Given that the Shell depends on `BASE` libraries, those need to be built with the small model as well. And given that SEC/PEI phase modules also rely on `BASE` libraries, we end up with conflicting requirements.

So the only workable solution today is to require 4k alignment everywhere when using CLANGPDB to build AArch64.

I'll log an issue with the LLVM/lld project.

os-d:
> Another concern we have is that, apparently, the FDF rules system is broken, given that its `Align=` directives can reduce the alignment of an executable, and it should only be able to increase the minimum. This means that currently, the only sane value for `Align` = `auto`.
> 

This only affects XIP, though, right? I.e. the FDF Align statement affects the alignment in flash, not the in memory alignment, right?

> > Let me know on what you think on requiring 4KB section alignment vs the tiny code model and I'll put up the PR. And again, if you want to file a clang issue for this, that would be helpful (if you think there is a reasonable answer of what they should do, whether basing ADRP usage on multiple of 4KB section alignment).
> 
> The tiny code model is not usable IIRC, because the Shell ends up being substantially larger than 1 MiB. Given that the Shell depends on `BASE` libraries, those need to be built with the small model as well. And given that SEC/PEI phase modules also rely on `BASE` libraries, we end up with conflicting requirements.
> 
> So the only workable solution today is to require 4k alignment everywhere when using CLANGPDB to build AArch64.
> 

Okay, I will put up a PR today with the CLANGPDB AArch64 support using 4KB alignment. Thanks for the deep dive on this, very appreciated!

> I'll log an issue with the LLVM/lld project.

Great, thank you, I'll link to that in the PR that adds the 4KB alignment everywhere, for reference.

ardbiesheuvel:
> > Another concern we have is that, apparently, the FDF rules system is broken, given that its `Align=` directives can reduce the alignment of an executable, and it should only be able to increase the minimum. This means that currently, the only sane value for `Align` = `auto`.
> 
> This only affects XIP, though, right? I.e. the FDF Align statement affects the alignment in flash, not the in memory alignment, right?
> 

Indeed. But this is where it matters most.

os-d:
> Okay, I will put up a PR today with the CLANGPDB AArch64 support using 4KB alignment. 

Here's the PR: https://github.com/tianocore/edk2/pull/11902. I took a bit of extra time to actually fix up all the build errors in edk2 and make sure that ArmVirtQemu can boot.

mikebeaton:
I have modified the commit series here so that it first of all explicitly clones GCC CI yaml to CLANGDWARF and CLANGPDB, then makes separate commits for the various changes (all removals of unsupported jobs) needed to run CI with these tool chains, compared to what GCC is able to run.

I changed this because the difference between GCC and CLANGDWARF/CLANGPDB CI is directly visible in the live demo PRs https://github.com/tianocore/edk2/pull/11860/commits and https://github.com/tianocore/edk2/pull/11861/commits, and I think is very useful to make clear exactly what is being done (now, and in future if these commits are merged), but was invisible/unclear in this sequence of commits prior to this change.

mikebeaton:
I have updated this PR with the minor nit improvement [suggested by](https://github.com/tianocore/edk2/pull/11905#discussion_r2635631806) @apop5.

I have confirmed that this proposed CLANG CI [all completes successfully when combined](https://github.com/tianocore/edk2/pull/11929#issuecomment-3705472086) with @mdkinney and @os-d's latest changes.

These two demo PRs show the extra steps that would be added by this CI for:
 - CLANGDWARF https://github.com/tianocore/edk2/pull/11861/commits
 - CLANGPDB https://github.com/tianocore/edk2/pull/11860/commits

_Note 1_: In these demo PRs I take over the GCC CI files to demonstrate how the CLANG CI would work (there is no other way to do so, given Azure Pipelines' master control over what runs), and I have also disabled the VS2022 and CodeQL jobs, since that makes it clearer what the new jobs are.

_Note 2_: The CLANGPDB demo above includes @os-d's [AARCH64 CLANGPDB support](https://github.com/tianocore/edk2/pull/11902/commits), since that enables some additional jobs which would have been included based on the logic I have taken: build everything non-test which GCC builds and which these toolchains can build.

_Note 3_: When viewing the above, there seems to be a bug in the GitHub UI whereby you have to manually navigate to the last commit of a PR with many commits in order to see its CI results (you can't just scroll down to it, 'below the fold' as they say, in the pages above). I have made a GitHub bug report.

I understand from @mdkinney (personal communication) that he sees the work I've done here as complementary to the larger scale project which he is working on, and would be happy to support adding CLANG coverage to CI in Azure Pipelines for now, as the fastest path to adding some coverage to prevent CLANG toolchain regressions.

The biggest issue as I would see it (Mike does not disagree) would be whether the load on Azure Pipelines would be acceptable if these jobs were added, given that - as I understand it - there are some concerns about the load now, with only the GCC jobs.

I'm not sure if anyone (from Microsoft?) with access to https://dev.azure.com/tianocore/edk2-ci/_build?view=folders&treeState=XENJ would be willing and able to add CI pipelines for the CLANGDWARF and CLANGPDB files which are added by this PR, to get a full view of what happens when this is all run at once?

makubacki:
> I'm not sure if anyone (from Microsoft?) with access to https://dev.azure.com/tianocore/edk2-ci/_build?view=folders&treeState=XENJ would be willing and able to add CI pipelines for the CLANGDWARF and CLANGPDB files which are added by this PR, to get a full view of what happens when this is all run at once?

@mikebeaton, I created pipelines for these files in [edk2-ci/Pipelines/Test](https://dev.azure.com/tianocore/edk2-ci/_build?definitionScope=%5CTest).

- [ArmVirtPkg - Ubuntu - CLANGDWARF](https://dev.azure.com/tianocore/edk2-ci/_build?definitionId=84)
- [EmulatorPkg - Ubuntu - CLANGDWARF](https://dev.azure.com/tianocore/edk2-ci/_build?definitionId=85)
- [OvmfPkg - Ubuntu - CLANGDWARF](https://dev.azure.com/tianocore/edk2-ci/_build?definitionId=86)
- [OvmfPkg - Ubuntu - CLANGPDB](https://dev.azure.com/tianocore/edk2-ci/_build?definitionId=87)
- [Ubuntu - CLANGDWARF](https://dev.azure.com/tianocore/edk2-ci/_build?definitionId=82)
- [Ubuntu - CLANGDWARF](https://dev.azure.com/tianocore/edk2-ci/_build?definitionId=83)

---

I cannot set up a connection to your edk2 fork so I pushed your current PR branch to a Tianocore owned repo ([edk2-staging](https://github.com/tianocore/edk2-staging)) in the [clang-ci](https://github.com/tianocore/edk2-staging/tree/clang-ci) branch.

The pipelines are set up to run against that branch right now.

mikebeaton:
> > I'm not sure if anyone (from Microsoft?) with access to https://dev.azure.com/tianocore/edk2-ci/_build?view=folders&treeState=XENJ would be willing and able to add CI pipelines for the CLANGDWARF and CLANGPDB files which are added by this PR, to get a full view of what happens when this is all run at once?
> 
> @mikebeaton, I created pipelines for these files in [edk2-ci/Pipelines/Test](https://dev.azure.com/tianocore/edk2-ci/_build?definitionScope=%5CTest).
> 
> * [ArmVirtPkg - Ubuntu - CLANGDWARF](https://dev.azure.com/tianocore/edk2-ci/_build?definitionId=84)
> * [EmulatorPkg - Ubuntu - CLANGDWARF](https://dev.azure.com/tianocore/edk2-ci/_build?definitionId=85)
> * [OvmfPkg - Ubuntu - CLANGDWARF](https://dev.azure.com/tianocore/edk2-ci/_build?definitionId=86)
> * [OvmfPkg - Ubuntu - CLANGPDB](https://dev.azure.com/tianocore/edk2-ci/_build?definitionId=87)
> * [Ubuntu - CLANGDWARF](https://dev.azure.com/tianocore/edk2-ci/_build?definitionId=82)
> * [Ubuntu - CLANGDWARF](https://dev.azure.com/tianocore/edk2-ci/_build?definitionId=83)
> 
> I cannot set up a connection to your edk2 fork so I pushed your current PR branch to a Tianocore owned repo ([edk2-staging](https://github.com/tianocore/edk2-staging)) in the [clang-ci](https://github.com/tianocore/edk2-staging/tree/clang-ci) branch.
> 
> The pipelines are set up to run against that branch right now.

@makubacki - That's great - many thanks! 

Looking at [the branch you kindly set up](https://github.com/tianocore/edk2-staging/tree/clang-ci), it does indeed run all the jobs added by this PR. I was wondering if there was any way to get a demo of _everything_ - the added jobs plus everything already present in normal (PR) CI - to give a more immediate sense of what the full result would be if this was merged (total time to run (which we can't quite be sure of without this) and total steps (which of course we can just work out anyway)).

makubacki:
> I was wondering if there was any way to get a demo of _everything_ - the added jobs plus everything already present in normal (PR) CI - to give a more immediate sense of what the full result would be if this was merged (total time to run (which we can't quite be sure of without this) and total steps (which of course we can just work out anyway)).

The existing non-Clang pipelines are set up to run against the YAML files in the tianocore/edk2 repo. I think you could achieve the same effect by pushing a change to the `clang-ci` branch in edk2-staging to trigger the Clang pipelines and open a draft PR in the edk2 repo to trigger the other pipelines. Ultimately though, it will depend on other jobs also queued in the pipelines.

mikebeaton:
> > I was wondering if there was any way to get a demo of _everything_ - the added jobs plus everything already present in normal (PR) CI - to give a more immediate sense of what the full result would be if this was merged (total time to run (which we can't quite be sure of without this) and total steps (which of course we can just work out anyway)).
> 
> The existing non-Clang pipelines are set up to run against the YAML files in the tianocore/edk2 repo. I think you could achieve the same effect by pushing a change to the `clang-ci` branch in edk2-staging to trigger the Clang pipelines and open a draft PR in the edk2 repo to trigger the other pipelines. Ultimately though, it will depend on other jobs also queued in the pipelines.

Hi Mike - Thanks for taking the time to look at this. I was really hoping to get one place to look at, with (something like) 170 jobs completed, and a certain time taken for each of them, and certain time total.

In terms of building it piece by piece, I can already take over the GCC yaml files in draft PRs to confirm how the CLANG builds would work (and indeed had in fact already [done](https://github.com/tianocore/edk2/pull/11860/commits) [so](https://github.com/tianocore/edk2/pull/11861/commits)).

The point (such as it is) being that I could then start the push of saying "here's the _full_ thing (with all 170 odd CI jobs), this is what it would look like if it was accepted; now, what is the chance of this being accepted, and what can I do to increase that chance?"

But you are correct that the CI of this very PR, plus either the CI in my 'take-over' PRs, or the CI in the branch you kindly set up, do already show that it all builds, and do allow someone to take the time to work out what all the jobs are that would build (easiest with your branch), and to _try_ to estimate what the total build time would be - though I appreciate that always depends on what other jobs are running. But it doesn't give the full (visceral?!) "this is what the full CI would be (here, in this one place), if this was accepted"...

os-d:
@mikebeaton my CLANGPDB AARCH64 PR merged, can we include that in this PR officially now?

makubacki:
@mikebeaton I think we have enough data for the build impact with the new pipelines. I'm going to introduce some additional pipeline changes soon to reduce utilization in other places to free up some resources for the Clang pipelines. Unless someone pushes back, I think you can proceed and we can adjust later if absolutely necessary.

mikebeaton:
> @mikebeaton I think we have enough data for the build impact with the new pipelines. I'm going to introduce some additional pipeline changes soon to reduce utilization in other places to free up some resources for the Clang pipelines. Unless someone pushes back, I think you can proceed and we can adjust later if absolutely necessary.

Okay - that's great! I wasn't sure if we were stuck and needed something more obvious on which to base a yay or nay discussion - but if not then of course it's fine. Thank you.

> @mikebeaton my CLANGPDB AARCH64 PR merged, can we include that in this PR officially now?

Of course. I will update shortly.

ardbiesheuvel:
I'd like to prioritize CLANGPDB over CLANGDWARF for AArch64, given that the former has all the advantages of the latter (better diagnostics, no need to install a cross compiler) while also doing away with `GenFw` entirely. The sooner we identify any potential issues wrt skipping the ELF to PE/COFF conversion, the better.

os-d:
> I'd like to prioritize CLANGPDB over CLANGDWARF for AArch64, given that the former has all the advantages of the latter (better diagnostics, no need to install a cross compiler) while also doing away with `GenFw` entirely. The sooner we identify any potential issues wrt skipping the ELF to PE/COFF conversion, the better.

Fully agreed. I would tend to say the same for x86, personally, though would defer to @mdkinney if there is value in running CLANGDWARF for x86 (outside of the host test on Linux case/EmulatorPkg on Linux). Just for CI times, since we have a lot to run. There is an overall workstream to make this better, but I do think part of that conversation should be which toolchains do we run. I believe that CLANGDWARF should either have the same issues CLANGPDB has (if clang related) or GCC has (if ELF target related), in the majority of cases.

That being said, I think @mikebeaton and @makubacki have been looking at the overall CI times, so if we don’t think CLANGDWARF is much of an impact, then that’s great.

To be clear, my concern is really on PR gate CI time, we’ve talked for awhile on nightly CI (which maybe is in the works now) and running CLANGDWARF there would be a good idea. This is my biased opinion of course, but I do feel like CLANGPDB solves the problem of being cross OS and producing PE targets. The problem with nightly CI being of course that unless you have someone who cares very very much, it will get left broken. Perhaps we would need maintainers of the nightly CI. Honestly, I would sign up to help with that if we could offload some of the PR gate CI to nightly. But we don’t need to have this conversation here, just contextualizing.

makubacki:
> That being said, I think @mikebeaton and @makubacki have been looking at the overall CI times, so if we don’t think CLANGDWARF is much of an impact, then that’s great.

The Clang sample pipelines runs are here: https://dev.azure.com/tianocore/edk2-ci/_build?definitionScope=%5CTest

CLANGDWARF has roughly equivalent build time to CLANGPDB so from an efficiency standpoint supporting those pipelines can be considered as doubling the build time.

---

We support a lot of toolchains, build environments, and workflows. This leads to bloated build times but also confusion about what is supported where and how to use it all. While this flexibility is sometimes useful because of the diversity of vendors using the project in some cases it's also detrimental allowing downstream consumers to wander down a path that might not be the best experience for them because of nuanced issues we're aware of but either don't have time to document and fix whenever another path would work fine for their use case. It also makes it more difficult to deliver a good experience because of the fragmentation splitting investments. This is like the VS 2026 AARCH64 discussion where someone could use it, but we don't recommend it based on our own experience.

Perhaps we use the more formalized TianoCore RFC process to allow focused discussions on toolchain support moving forward so the entire decision-making process is clearer, and stakeholders are more clearly identified.

---

> To be clear, my concern is really on PR gate CI time, we’ve talked for a while on nightly CI (which maybe is in the works now) and running CLANGDWARF there would be a good idea.

I am going to post a PR soon that will move today's post-merge checks to nightly. This is part of a series of upcoming changes to improve CI time including Clang GitHub workflows with stages. Regardless, less builds in the pipelines always helps.

mikebeaton:
> @mikebeaton my CLANGPDB AARCH64 PR merged, can we include that in this PR officially now?

@os-d:

I have updated the PR to officially include this. It's in this PR, and should also be building [here](https://github.com/tianocore/edk2/pull/11860/commits) in 'takover GCC' mode.

> ... you could achieve the same effect by pushing a change to the clang-ci branch in edk2-staging to trigger the Clang pipelines

@makubacki:

I believe I don't have permissions to push to edk2-staging, I've tried to [make a PR](https://github.com/tianocore/edk2-staging/pull/519) but I'm not sure that's much help - I think edk2-staging clang-ci branch just needs force pushing to match the current tip of this PR, but I can't do it.

Please can you add another pipeline, for the additional file that is now present after including CLANGPDB AARCH64: `ArmVirtPkg/PlatformCI/.azurepipelines/Ubuntu-CLANGPDB.yml`

makubacki:
> I believe I don't have permissions to push to edk2-staging, I've tried to https://github.com/tianocore/edk2-staging/pull/519 but I'm not sure that's much help - I think edk2-staging clang-ci branch just needs force pushing to match the current tip of this PR, but I can't do it.

@mikebeaton, I've added you to `edk-ii-staging-maintainers` which gives you write access to edk2-staging.

> Please can you add another pipeline, for the additional file that is now present after including CLANGPDB AARCH64: ArmVirtPkg/PlatformCI/.azurepipelines/Ubuntu-CLANGPDB.yml

Yes. I will be able to do that by tomorrow. I'll reply when it's done.

mikebeaton:
> > I believe I don't have permissions to push to edk2-staging, I've tried to [tianocore/edk2-staging#519](https://github.com/tianocore/edk2-staging/pull/519) but I'm not sure that's much help - I think edk2-staging clang-ci branch just needs force pushing to match the current tip of this PR, but I can't do it.
> 
> @mikebeaton, I've added you to `edk-ii-staging-maintainers` which gives you write access to edk2-staging.

@makubacki, Thank you - I've pushed the latest version of this PR to `clang-ci` on `edk2-staging` https://github.com/tianocore/edk2-staging/commits/clang-ci/, but it doesn't seem to want to re-run any pipelines.

It's sticking with the failed CI result which the same commit generated 3 days ago when it first appeared as a PR in edk-staging.

I'm not sure whether it would be running now, if it didn't already have a cached result from the PR? Nor whether it matters that the error from the PR CI https://github.com/tianocore/edk2-staging/pulls?q=is%3Apr+author%3Amikebeaton+is%3Aclosed https://github.com/tianocore/edk2-staging/runs/60716111576 was "The current Mergify configuration is invalid".

makubacki:
> @makubacki, Thank you - I've pushed the latest version of this PR to `clang-ci` on `edk2-staging` https://github.com/tianocore/edk2-staging/commits/clang-ci/, but it doesn't seem to want to re-run any pipelines.

Have you tried updating the trigger branches to include `clang-ci`?

```yaml
trigger:
- master
- stable/*
- clang-ci
```

> I'm not sure whether it would be running now, if it didn't already have a cached result from the PR?

It should run again based on your triggers.

> I will be able to do that by tomorrow. I'll reply when it's done.

The ArmVirtPkg Ubuntu CLANPDB pipeline is here: [ArmVirtPkg - Ubuntu - CLANGPDB](https://dev.azure.com/tianocore/edk2-ci/_build?definitionId=88)

mikebeaton:
Updated to account for ccbee92b65a35a2ce8aa4c6ec4aa6d1f09000cf7.

os-d:
@mikebeaton what is the status of this PR? Now that we are past the hard freeze are we able to merge it? I'm very keen to get CLANGPDB CI running to avoid regressions for all the fixes we've been putting in. Thanks again for doing the work to get it going!

It is still my opinion that I would prefer only adding CLANGPDB CI to PR gates and CLANGDWARF could be a nightly run, because I believe CLANGPDB is the better use case for platform FW. I defer to the stewards on what they deem is best for edk2 as a whole.

mikebeaton:
I believe it is ready to merge. I was going to go one step at a time, and look at getting [BaseTools: Remove previously deprecated GCC48, GCC49 and GCC5](https://github.com/tianocore/edk2/pull/12055) merged, then come back to [BaseTools: Remove no longer needed warning suppression](https://github.com/tianocore/edk2/pull/11761), then come back to this. But there's no requirement for any particular order.

> It is still my opinion that I would prefer only adding CLANGPDB CI to PR gates and CLANGDWARF could be a nightly run, because I believe CLANGPDB is the better use case for platform FW. I defer to the stewards on what they deem is best for edk2 as a whole.

I can look at making those changes if preferred.

os-d:
> I believe it is ready to merge. I was going to go one step at a time, and look at getting [BaseTools: Remove previously deprecated GCC48, GCC49 and GCC5](https://github.com/tianocore/edk2/pull/12055) merged, then come back to [BaseTools: Remove no longer needed warning suppression](https://github.com/tianocore/edk2/pull/11761), then come back to this. But there's no requirement for any particular order.
> 
> > It is still my opinion that I would prefer only adding CLANGPDB CI to PR gates and CLANGDWARF could be a nightly run, because I believe CLANGPDB is the better use case for platform FW. I defer to the stewards on what they deem is best for edk2 as a whole.
> 
> I can look at making those changes if preferred.

@mikebeaton based on the feedback here from Ard (his comment) and Michael (his thumbs up) and myself and that no one else has asked for CLANGDWARF CI, I suggest starting with CLANGDWARF as nightly. If the community shows more interest in the toolchain, it is easy to promote it to a PR gate later. But we have a current problem with PR gate times and I am concerned about adding too many things at once, particularly when the community interest does not seem high.

mikebeaton:
@os-d @makubacki cc @mdkinney - Have I correctly implemented what you want for 'do not add CLANGDWARF to PR gates' in commit 6 4da674ed780333aeffe653cbbda70b0e5c19848e?

What I've done here at the moment is NOT the same (as regards the `pr:` lines) as what was in https://github.com/tianocore/edk2/pull/12004/changes - or to put it another way, I'm not quite sure when this is and isn't supposed to run!

Are the cron schedules (introduced in https://github.com/tianocore/edk2/pull/12004) cloned from GCC into CLANGDWARF commit 5 8107881eef6e97fccd12f31ddd8b3ffde01b421e and GLANGPDB commit 8 7d725944ffa651d54420417adcaf100236cc2e0c suitable?

mikebeaton:
~https://github.com/tianocore/edk2/pull/12255 should be merged before this PR.~

EDIT: Has been merged.

os-d:
Thanks @mikebeaton from my point of view this seems reasonable, but I'll also be the first to admit I'm not a pipeline expert :). @makubacki @Javagedes can you please help review and ensure the pipeline logic is reasonable?

.azurepipelines/templates/platform-build-run-steps.yml
```diff
@@ -101,7 +101,10 @@ steps:
 #   are all resolved.
 - template: basetools-build-steps.yml
   parameters:
-    tool_chain_tag: ${{ parameters.tool_chain_tag }}
+    ${{ if eq(parameters.tool_chain_tag, 'VS2022') }}:
```
Javagedes: nit: I would probably do a 'startsWith' 'VS' rather then this. I can see this being easily missed if we update to VS2026 or later.

.azurepipelines/templates/pr-gate-steps.yml
```diff
@@ -46,6 +46,16 @@ steps:
     echo "##vso[task.setvariable variable=pkgs_to_build]${{ parameters.build_pkgs }}"
     echo "##vso[task.setvariable variable=pkg_count]${{ 1 }}"
 
+# Strip test targets for CI toolchains other than GCC and VS2022, as these
+# cannot run them and would be duplicating work during CI if they could.
+# Note: Within PR gate CI, NOOPT is only used for building tests.
+- ${{ if in(parameters.tool_chain_tag, 'GCC', 'VS2022') }}:
```
Javagedes: I don't like this. users of pr-gate-steps.yml should properly set the `build_targets` parameter. Us magically changing the build targets with no indication of doing so is confusing. It will also mess up any type of CI debugging because we will debug print the input as specific targets and then do something else.

.azurepipelines/templates/pr-gate-steps.yml
```diff
@@ -70,28 +80,31 @@ steps:
...
 
 # build basetools
 #   do this after setup and update so that code base dependencies
 #   are all resolved.
 - template: basetools-build-steps.yml
   parameters:
-    tool_chain_tag: ${{ parameters.tool_chain_tag }}
+    ${{ if eq(parameters.tool_chain_tag, 'VS2022') }}:
+      tool_chain_tag: ${{ parameters.tool_chain_tag }}
+    ${{ else }}:
```
Javagedes: Build tools should be able to be built with whatever toolchain the platform is being built with. If they cannot, that is an issue that needs to be resolved.

Telling consumers "If you want to build your platform with CLANG, you need to also have all the GCC toolchain things set up because you need to build the build tools with GCC" is not good.

.azurepipelines/templates/pr-gate-steps.yml
```diff
@@ -21,7 +21,7 @@ steps:
     echo "##vso[task.prependpath]${HOME}/.local/bin"
     echo "new PATH=${PATH}"
   displayName: Set PATH
-  condition: eq('${{ parameters.tool_chain_tag }}', 'GCC')
+  condition: in('${{ parameters.tool_chain_tag }}', 'GCC', 'CLANGDWARF', 'CLANGPDB')
```
Javagedes: CLANGPDB is supported on windows, even if you are not doing it right now. It makes more sense in my eyes to check what runner OS you are on, and if it is linux-esque, then do this append step. This worked before because GCC was never run on windows.

.azurepipelines/Ubuntu-CLANGDWARF.yml
```diff
@@ -0,0 +1,32 @@
...
+pr: none
+
+variables:
+  - template: templates/defaults.yml
+
+jobs:
+- template: templates/pr-gate-build-job.yml
+  parameters:
+    tool_chain_tag: CLANGDWARF
+    vm_image: 'ubuntu-24.04'
```
Javagedes: in my opinion we should move the vm_image to defaults.yml also so we only need to change it in one location. Not 100% applicable to this PR so I understand if you want to do it outside of this PR before we merge this.

.azurepipelines/Ubuntu-CLANGPDB.yml
```diff
@@ -0,0 +1,34 @@
+## @file
+# Azure Pipeline build file for a build using Ubuntu and CLANGPDB
+#
+# Copyright (c) Microsoft Corporation.
```
Javagedes, line 4: Are all of these copyrights correct? This is a new file and not touched by everyone here.

There is also a chance I misunderstand the copyright process, and this is correct, however.

.azurepipelines/templates/pr-gate-steps.yml
mikebeaton: This one I would agree with, but would try to argue it's a somewhat larger separate issue.

There isn't any simple way to build BaseTools except with gcc (though you can in fact successfully set flags to build with clang (noting that toolchains per se don't apply here, just compilers)), so as per the relevant commit comment https://github.com/tianocore/edk2/pull/11886/changes/04e0a87bb7a7a08cb47ab3035da1fcf0627201ed, my understanding is that almost everyone not on Windows is in fact building BaseTools with gcc, whichever toolchain they are using after that. I may be wrong?

So I was supposing or hoping that it would at least be okay within the CI pipeline to force BaseTools to build with gcc, at least temporarily while stuart does not support anything else.

Could this requested change to stuart be handled separately afterwards?

EDIT: I'll have a look at what would be involved to make this change too.

.azurepipelines/Ubuntu-CLANGPDB.yml
mikebeaton: Thank you very much for the detailed review.

On this issue, this is a clone of another file - hence I was assuming the existing copyrights should stay (much as they would if I cloned the file into a separate project, which the license allowed), but perhaps I don't understand the process. I don't know if yourself or anyone else can confirm either way?

And arguably I should have added myself -  I don't have any larger organisation that I can or should be assigning the copyright to - though I wouldn't necessarily presume or particularly need to do so unless it's required here?

Javagedes: hmm, that makes sense. I don't think you have to add yourself if you don't want to, but I do know you can add yourself / name and not a larger entity. There are various files throughout EDK II that are copyrighted by someone specific.

I will let others (maybe @makubacki or @mdkinney) answer if we should or should not keep the copyrights from the copied file, as I'm not familiar with the legalities or anything :)

makubacki: I am not a lawyer, but it is standard practice to preserve the copyright notices of authors when deriving from their work.

Javagedes: sounds good. We should keep the copyrights then. I will close this now. It's completely up to you on whether you wish to add yourself or not.

.azurepipelines/templates/pr-gate-steps.yml
Javagedes: Fair point. Just skimming, it looks like there is some custom logic for clang on linux builds, so I'm surprised it is not working right now on your linux pipelines.

Regardless (outside of this PR or not), we (or I) should get base tools building with CLANG for linux and windows, which will need a bit of an overhaul on that python script.

.azurepipelines/Ubuntu-CLANGDWARF.yml
mikebeaton: https://github.com/tianocore/edk2/pull/12284

.azurepipelines/templates/pr-gate-steps.yml
mikebeaton: https://github.com/tianocore/edk2/pull/12286

.azurepipelines/templates/pr-gate-steps.yml
mikebeaton: This is addressed in revised commit no. 3 99bf32dc37652af3d1ca422a608f5d763f2753a1, I have moved (and slightly improved) the code for stripping unwanted targets from `pr-gate-steps.yml` to its (only) caller, `pr-gate-build-job.yml`.

.azurepipelines/templates/pr-gate-steps.yml
mikebeaton: Actually ignore my edit in my previous comment above, can I go back to "Could this requested change to stuart be handled separately afterwards?"?

.azurepipelines/templates/platform-build-run-steps.yml
mikebeaton: Resolved in new commits.

mikebeaton:
#12292 should also (!) be merged before this PR

.azurepipelines/templates/pr-gate-steps.yml
Javagedes: It cannot really be handled by stuart itself. The logic to support build tools being built via CLANG will need to happen in [Edk2ToolsBuild.py](https://github.com/tianocore/edk2/blob/master/BaseTools/Edk2ToolsBuild.py) and probably the bat / bash scripts for people that like to build the tools outside of stuart.

I'm fine with it being handled separately afterwords though.

mdkinney: > It cannot really be handled by stuart itself. The logic to support build tools being built via CLANG will need to happen in [Edk2ToolsBuild.py](https://github.com/tianocore/edk2/blob/master/BaseTools/Edk2ToolsBuild.py) and probably the bat / bash scripts for people that like to build the tools outside of stuart.
> 
> I'm fine with it being handled separately afterwords though.

I already added that support to Edk2ToolBuild.py and am using it from Linux/clang and Mingw/clang environments.

https://github.com/tianocore/edk2/pull/11930

mikebeaton: @mdkinney - Apologies, I hadn't realised that had gone in - I can probably get away with blaming the holiday season timing!

I've removed the commit here which was forcing all BaseTools builds to use the GCC toolchain, and it all [works](https://github.com/tianocore/edk2/pull/11860/commits) completely [fine](https://github.com/tianocore/edk2/pull/11861/commits) now without it.

Many thanks.

mikebeaton:
If:

 - @makubacki - you are happy with the trigger definitions at the top of the added .yml files; 
 - @Javagedes - you are happy that I (and @mdkinney when I wasn't looking) have addressed your outstanding concerns; 
 - ~and once #12292 is merged~ (DONE);

then I think this might be ready to merge?

As I think everyone is aware, but as a reminder, @makubacki would need to kindly add live pipelines for the 7 added .yml files, before they will do anything when merged.

mikebeaton:
> If:
> 
> * @makubacki - you are happy with the trigger definitions at the top of the added .yml files;
> * @Javagedes - you are happy that I (and @mdkinney when I wasn't looking) have addressed your outstanding concerns;
> * ~and once [DynamicTablesPkg/AmlLib: Fix refactor AmlCodeGenInvokeMethod API #12292](https://github.com/tianocore/edk2/pull/12292) is merged~ (DONE);
> 
> then I think this might be ready to merge?
> 
> As I think everyone is aware, but as a reminder, @makubacki would need to kindly add live pipelines for the 7 added .yml files, before they will do anything when merged.

@mdkinney - I should have asked previously, it would be great to have final confirmation from yourself as to whether this all looks good, and fits in properly (at least as a stop gap?) with what you were already planning?

mdkinney:
> > If:
> > 
> > * @makubacki - you are happy with the trigger definitions at the top of the added .yml files;
> > * @Javagedes - you are happy that I (and @mdkinney when I wasn't looking) have addressed your outstanding concerns;
> > * ~and once [DynamicTablesPkg/AmlLib: Fix refactor AmlCodeGenInvokeMethod API #12292](https://github.com/tianocore/edk2/pull/12292) is merged~ (DONE);
> > 
> > then I think this might be ready to merge?
> > As I think everyone is aware, but as a reminder, @makubacki would need to kindly add live pipelines for the 7 added .yml files, before they will do anything when merged.
> 
> @mdkinney - I should have asked previously, it would be great to have final confirmation from yourself as to whether this all looks good, and fits in properly (at least as a stop gap?) with what you were already planning?

LGTM

makubacki (approved):

makubacki:
> If:
> 
> * @makubacki - you are happy with the trigger definitions at the top of the added .yml files;
> * @Javagedes - you are happy that I (and @mdkinney when I wasn't looking) have addressed your outstanding concerns;
> * ~and once [DynamicTablesPkg/AmlLib: Fix refactor AmlCodeGenInvokeMethod API #12292](https://github.com/tianocore/edk2/pull/12292) is merged~ (DONE);
> 
> then I think this might be ready to merge?
> 
> As I think everyone is aware, but as a reminder, @makubacki would need to kindly add live pipelines for the 7 added .yml files, before they will do anything when merged.

Thank you for all the work to keep this submission going. There will obviously be some delay due to time zones and working hours, but I will keep an eye out to add the pipelines needed.

mikebeaton:
@makubacki has kindly said he will be on hand to add the Azure Pipelines once the .yml files are in place.

Are any of the following people who I believe would have access to, and have commented above, willing and able to pull the trigger and add the 'push' label? (And/or, who else should I be checking with for final approval?) Many thanks!

cc @mdkinney @makubacki @os-d @ardbiesheuvel @Javagedes

makubacki:
@mikebeaton, I have enabled the pipelines on edk2/master. They are now located in the [\\CI](https://dev.azure.com/tianocore/edk2-ci/_build?definitionScope=%5CCI) pipeline folder.

mikebeaton:
> @mikebeaton, I have enabled the pipelines on edk2/master. They are now located in the [\CI](https://dev.azure.com/tianocore/edk2-ci/_build?definitionScope=%5CCI) pipeline folder.

Hi! Many thanks!

I'm just running CI on a test PR here: https://github.com/tianocore/edk2/pull/12304/commits

mikebeaton:
Ah - spotted something - assuming it's all passing okay, presumably it would be good to set the three CLANGPDB pipelines to required - I'm just going by this output on the pull request page:

image

mikebeaton:
Two more:

Firstly, I suspect that this may be the same issue as the above, but multiple build stages are labelled as "PR" for VS and GCC, but are not for CLANGPDB. This doesn't seem to be to do with the naming of the new pipes - I'm not sure if it indicates anything that matters?

image

image

Secondly, in the steps run [here](https://github.com/tianocore/edk2/pull/11860/commits) (expand green tick) where I was demoing CLANGPDB, TARGET_FMP_FAT_TEST only shows up in `Ubuntu GCC` (which would now have it's real name of Ubuntu CLANGPDB), but [here](https://github.com/tianocore/edk2/pull/12304/commits) as well as showing up in the expected places (GCC, CLANGPDB, VS), it shows up in OvmfPkg as well.

image

image

Is it possible one of the pipelines is pointing at the wrong .yml?

Javagedes:
`ArmVirtPkg - Ubuntu - CLANGPDB` -> `ArmVirtPkg/PlatformCI/.azurepipelines/Ubuntu-CLANGPDB.yml`
`OvmfPkg - Ubuntu - CLANGPDB` -> `.azurepipelines/Ubuntu-CLANGPDB.yml`
`Ubuntu - CLANGPDB` -> `.azurepipelines/Ubuntu-CLANGPDB.yml`

So yes it looks like `OvmfPkg` is pointing at the wrong workflow file. I don't have permissions to update, but @makubacki should be able to update easily.

Javagedes:
As for the other parts of your question. Marking a job as required is a manual step in github which can be done. 

As for the naming of the pipeline, pipelines are different the github actions. Github actions are automatically detected, associated, and then named based off the name specified in the workflow. For azure pipelines, you must manually create a pipeline, give it a name, and then associate the pipeline with a yaml file. The pipeline missing "CI" in the name is just an artifact of how the pipeline was created in azure devops before linking to the actual yaml file.

mikebeaton:
> As for the naming of the pipeline, pipelines are different the github actions. Github actions are automatically detected, associated, and then named based off the name specified in the workflow. For azure pipelines, you must manually create a pipeline, give it a name, and then associate the pipeline with a yaml file. The pipeline missing "CI" in the name is just an artifact of how the pipeline was created in azure devops before linking to the actual yaml file.

Many thanks for your answers. I saw the slight difference in naming patterns, but I don't think I'm referring to that. See my screenshots above, in that question I'm asking about it missing text "PR", which doesn't seem to be coming from the pipeline names?

Javagedes:
image

Are you referring to this "PR" or something else? If it is this "PR", then that is just part of the name that was manually given inside azure devops when creating the pipeline and linking the yaml file. If you mean something else then just let me know because I am misunderstanding what you are referring to :)

mikebeaton:
Ah, got it, there are pipelines under [PR](https://dev.azure.com/tianocore/edk2-ci/_build?definitionScope=%5CPR) with "PR" in the name and pipelines under [CI](https://dev.azure.com/tianocore/edk2-ci/_build?definitionScope=%5CCI) with "CI" in the name.

I was wrongly assuming everything relevant was coming from the CI directory which Michael linked to above, especially as the new pipelines are only in CI (as I now see), but I guess that's a change of approach as to how the older vs. newer ones are set up.

makubacki:
@mikebeaton 

- I updated the OVMF pipeline to point to `OvmfPkg/PlatformCI/.azurepipelines/Ubuntu-CLANGPDB.yml`

Everything else is intentional and expected.

1. **Naming**: Your pipeline files have CI and PR triggers. Therefore, one name is given that works for both.
2. **Requiring the pipeline**: This is done in the GitHub branch rules. I want to confirm the pipelines work as expected for a period of time before marking them required. Also, GitHub does not even recognize status checks to be added until they have been run.

---

Your confusion about naming is due to _a way_ to set up pipelines. In that approach, you create two pipeline definitions that refer to the same file. One definition is the "PR" definition. One definition is the "CI" definition. In a case like yours, where files define both triggers (which is correct/fine), you would "override" either the CI and PR trigger in each pipeline definition to disable the other trigger. For example, the PR definition would override the CI trigger to be disabled.

I don't like doing that. There is not much value (other than naming precision) and it just makes more copies of definitions which adds more maintenance. It also allows PR and CI results for the pipeline to be separately maintained but we don't really care about that either. Hence why I put a single definition in the "CI" folder that works on both of the triggers as defined in the files.

---

I will mark the pipelines as required soon today.

mikebeaton:
> @mikebeaton
> 
> * I updated the OVMF pipeline to point to `OvmfPkg/PlatformCI/.azurepipelines/Ubuntu-CLANGPDB.yml`

Thank you for the fix.
 
> Everything else is intentional and expected.
> 
> 1. **Naming**: Your pipeline files have CI and PR triggers. Therefore, one name is given that works for both.
> 2. **Requiring the pipeline**: This is done in the GitHub branch rules. I want to confirm the pipelines work as expected for a period of time before marking them required. Also, GitHub does not even recognize status checks to be added until they have been run.
> 
> Your confusion about naming is due to _a way_ to set up pipelines. In that approach, you create two pipeline definitions that refer to the same file. One definition is the "PR" definition. One definition is the "CI" definition. In a case like yours, where files define both triggers (which is correct/fine), you would "override" either the CI and PR trigger in each pipeline definition to disable the other trigger. For example, the PR definition would override the CI trigger to be disabled.
> 
> I don't like doing that. There is not much value (other than naming precision) and it just makes more copies of definitions which adds more maintenance. Hence why I put a single definition in the "CI" folder that works on both of the triggers as defined in the files.

With only public access to https://dev.azure.com/tianocore/edk2-ci I had assumed I couldn't see much about how the pipelines are set up, but in fact the pipeline definition JSON is accessible, so I had already realised that you had both triggers in the new files and it all makes sense!
 
> I will mark the pipelines as required soon today.

Thanks again.

makubacki:
> Thanks again.

No problem. All that said, I'm definitely happy to create separate definitions if anyone prefers that for consistency with existing pipelines. It just tends to complicate and lead to its own consistency problems in my experience.

mikebeaton:
> > Thanks again.
> 
> No problem. All that said, I'm definitely happy to create separate definitions if anyone prefers that for consistency with existing pipelines. It just tends to complicate and lead to its own consistency problems in my experience.

It could be nice to go the other way, and merge the PR folder triggers into the CI folder files?

makubacki:
> It could be nice to go the other way, and merge the PR folder triggers into the CI folder files?

I was thinking about that too. I don't think it's urgent enough to jump on right away, but I submitted an issue to track it - https://github.com/tianocore/edk2/issues/12305.

mikebeaton:
> I will mark the pipelines as required soon today.

@makubacki - No particular hurry, I believe, but is it okay to mark the new pipelines as required? Actually, I was wondering, what difference will that make? Since I noticed that, even if not 'Required', failures [show up as](https://github.com/tianocore/edk2/pull/12313) 'Some checks were not successful' - which (presumably?) blocks merging?

cc @Javagedes, @os-d

makubacki:
> Actually, I was wondering, what difference will that make? Since I noticed that, even if not 'Required', failures [show up as](https://github.com/tianocore/edk2/pull/12313) 'Some checks were not successful' - which (presumably?) blocks merging?

I added the CLANGPDB status checks are required. From a pure [GitHub perspective](https://docs.github.com/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks), checks that fail and are not required do not block merging. Required checks do. We often use this to evaluate new status checks before marking them required.

mikebeaton:
> I added the CLANGPDB status checks are required. From a pure [GitHub perspective](https://docs.github.com/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks), checks that fail and are not required do not block merging. Required checks do. We often use this to evaluate new status checks before marking them required.

Thank you.

Issue #11878, opened 2025-12-10
[Bug]: CLANGDWARF cannot build RELEASE AARCH64 StandaloneMmPkg


mikebeaton:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

StandaloneMmPkg

### Which targets are impacted by this bug?

RELEASE

### Current Behavior

CLANGDWARF cannot currently build AARCH64 RELEASE StandaloneMmPkg, failing with:

/__w/1/s/Build/StandaloneMm/RELEASE_CLANGDWARF/AARCH64/StandaloneMmPkg/
Core/StandaloneMmCore/DEBUG/StandaloneMmCore.dll: Bad definition for
symbol '<unknown>'@0 or unsupported symbol type.  For example, absolute
and undefined symbols are not supported.

### Expected Behavior

That CLANGDWARF can build this package.

### Steps To Reproduce

build -p StandaloneMmPkg/StandaloneMmPkg.dsc -b RELEASE -t CLANGDWARF -a AARCH64

### Build Environment

```markdown
- OS(s): fedora-43 CI image in container
- Tool Chain(s): CLANGDWARF
```

### Version Information

```text
Earlier versions not tested. edk2/master.
```

### Urgency

Medium

### Are you going to fix this?

Someone else needs to fix it

### Do you need maintainer feedback?

Maintainer feedback requested

### Anything else?

I have not had a chance to investigate further, and do not currently have a fix available.

mikebeaton:
`llvm-readelf -rW /home/vsts/w/edk2/Build/StandaloneMm/RELEASE_CLANGDWARF/AARCH64/StandaloneMmPkg/Core/StandaloneMmCore/DEBUG/StandaloneMmCore.dll` shows the `.rela.data` section full of zero lines in the problem build:

```
Relocation section '.rela.data' at offset 0x6b448 contains 52 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
0000000000000000  0000000000000000 R_AARCH64_NONE                    0
0000000000000000  0000000000000000 R_AARCH64_NONE                    0
0000000000000000  0000000000000000 R_AARCH64_NONE                    0
...
```
This seems to be normal for debug sections in RELEASE builds, e.g.:
```
Relocation section '.rela.debug_line' at offset 0x60cb8 contains 647 entries:
[...]

PR #11890, opened 2025-12-15
Basetools ignore elf none relocations

ardbiesheuvel:
The ELF psABIs for all architectures stipulate that R_*_NONE relocations
require no action at relocation time, and merely exist to declare a
dependency on a symbol in a way that cannot be conveyed by the code
itself (i.e., using an actual symbol reference). Given that EFI PE/COFF
images are always fully linked binaries, such a dependency cannot be
translated, and there are no known reasons why this would be necessary.

So instead of ignoring such relocations specifically on x86_64 only,
ignore them on all architectures when converting ELF binaries to
PE/COFF.


mikebeaton:
Pushed these into [CLANGPDB](https://github.com/tianocore/edk2/pull/11860/commits) and [CLANGDWARF](https://github.com/tianocore/edk2/pull/11861/commits) CI. :)

mikebeaton (approved):

mikebeaton:
I can confirm that this fixes the [build issue](https://github.com/tianocore/edk2/issues/11878) - tyvm - and doesn't affect any other builds on any [other](https://github.com/tianocore/edk2/pull/11860/commits) (CLANGPDB) [supported](https://github.com/tianocore/edk2/pull/11861/commits) (CLANGDWARF) [toolchains](https://github.com/tianocore/edk2/pull/11890/commits) (GCC).

mikebeaton:
Probably self-explanatory, but I think patch check is complaining because it wants the first commit to be tagged `BaseTools,UefiPayloadPkg:` not `Global:`.

lgao4:
I agree the code change is good. But, the commit log doesn't pass the patch checker.

PR #11891, opened 2025-12-15
﻿MdeModulePkg/PciBusDxe: Fix UINT8 overflow in PciAllocateBusNumber

arkadiusz-filipowski-intel:
# Description

When StartBusNumber is 0xFF NextNumber would overflow and wrap back to 0. This will be bypass check NextNumber > MaxNumberInRange as a result PCI enumeration will continue instead of breaking due to lack of resources.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

PCI hierarchy causing this issue was enumerated again with this fix and enumeration reported lack of resources as expected.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Bus/Pci/PciBusDxe/PciLib.c
+++ b/MdeModulePkg/Bus/Pci/PciBusDxe/PciLib.c
@@ -1025,7 +1025,7 @@ PciAllocateBusNumber (
 {
   PCI_IO_DEVICE                      *RootBridge;
   EFI_ACPI_ADDRESS_SPACE_DESCRIPTOR  *BusNumberRanges;
-  UINT8                              NextNumber;
+  UINT16                             NextNumber;
   UINT64                             MaxNumberInRange;
 
   //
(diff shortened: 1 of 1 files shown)


MdeModulePkg/Bus/Pci/PciBusDxe/PciLib.c
```diff
@@ -1043,18 +1043,18 @@ PciAllocateBusNumber (
...
           return EFI_OUT_OF_RESOURCES;
         }
 
-        NextNumber       = (UINT8)(NextNumber + (BusNumberRanges->AddrRangeMin - (MaxNumberInRange + 1)));
+        NextNumber       = (UINT16)(NextNumber + (BusNumberRanges->AddrRangeMin - (MaxNumberInRange + 1)));
         MaxNumberInRange = BusNumberRanges->AddrRangeMin + BusNumberRanges->AddrLen - 1;
       }
 
-      *NextBusNumber = NextNumber;
+      *NextBusNumber = (UINT8)NextNumber;
```
mdkinney, line 1057: This strips the upper 8-bits of NextNumber, which means the return value NextBusNumber can be 0 or are small value.  Isn't that also an issue?

arkadiusz-filipowski-intel: Values greater than 8 bit would be caught by while loop "while (NextNumber > MaxNumberInRange)" in line 1047. 
As values greater than 255 are not possible this would never leave this loop and break with EFI_OUT_OF_RESOURCES.
So here we can cast to UINT8 as in this place we know that value is correct, meaning between <0-255>

lgao4:
@niruiyu , can you help confirm this change?

PR #11901, opened 2025-12-18
Add Visual Studio 2026 toolchain support

Daniel224455:
# Description

Add support for the VS2026 (VC180) toolchain.

Add tools_def for VS2026.
Update WindowsVsToolChain to support VS2026.
Update set_vsPrefix_envs and toolsetup and edksetup to support VS2026.
Add VS2026 linker flags to the EmulatorPkg WinHost INF.
Enable Address Sanitizer for VS2026 in UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
Add VS2026 linker flags to UnitTestFrameworkPkgHost.dsc.inc
Add VS2026 to the !if case in NetworkPkgHostTest.dsc

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Built EmulatorPkg, UefiPayloadPkg and MdePkg locally with VS2026 toolchain. Everything passed successfully.
CI build: Everything aside from the targets primarily for RISCV64 and AARCH64 passes.

## Integration Instructions

N/A

### DIFF
--- a/UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
+++ b/UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
@@ -49,7 +49,7 @@
 # +--------------------+--------+----------+------------+-----+----+--------+
 # | OS/Compiler        | VS2019 | CLANGPDB | CLANGDWARF |   GCC    | XCODE5 |
 # |                    | VS2022 |          |            |   GCC5   |        |
-# |                    |        |          |            | GCCNOLTO |        |
+# |                    | VS2026 |          |            | GCCNOLTO |        |
 # +--------------------+--------+----------+------------+----------+--------+
 # | Windows/VS         |IA32/X64|          |            |          |        |
 # | Windows/LLVM/VS    |        | IA32/X64 |            |          |        |
@@ -90,7 +90,7 @@
 # macOS/XCODE: macOS environment with XCODE5 installed.
 #
 !if $(WIN_MINGW32_BUILD)
-  !if $(TOOL_CHAIN_TAG) in "VS2019 VS2022"
+  !if $(TOOL_CHAIN_TAG) in "VS2019 VS2022 VS2026"
     !error Host-based Unit Tests not supported for Mingw/VS20xx builds
   !endif
   !if $(TOOL_CHAIN_TAG) in "CLANGPDB"
@@ -108,7 +108,7 @@
       !error Host-based Unit Tests not supported for Windows/CLANGDWARF builds
     !endif
   !else
-    !if $(TOOL_CHAIN_TAG) in "VS2019 VS2022"
+    !if $(TOOL_CHAIN_TAG) in "VS2019 VS2022 VS2026"
       !error Host-based Unit Tests not supported for Linux/VS20xx builds
     !endif
     !if $(TOOL_CHAIN_TAG) in "CLANGPDB"
@@ -149,10 +149,11 @@
 
 !if $(UNIT_TESTING_ADDRESS_SANITIZER_ENABLE)
   #
-  # Enable Address Sanitizer for VS2019, VS2022, and GCC
+  # Enable Address Sanitizer for VS2019, VS2022, VS2026, and GCC
   #
   MSFT:*_VS2019_*_CC_FLAGS = /fsanitize=address
   MSFT:*_VS2022_*_CC_FLAGS = /fsanitize=address
+  MSFT:*_VS2026_*_CC_FLAGS = /fsanitize=address
   GCC:*_*_*_CC_FLAGS       = -fsanitize=address
   CLANGPDB:*_*_*_CC_FLAGS  = -fsanitize=address
 !endif


BaseTools/Conf/tools_def.template
```diff
@@ -170,6 +177,11 @@ DEFINE DTC_BIN                 = ENV(DTC_PREFIX)dtc
 #                        Optional:
 #                             Required to build EBC drivers:
 #                               Intel(r) Compiler for Efi Byte Code (Intel(r) EBC Compiler)
+#   VS2026      -win32,win64-  Requires:
+#                             Microsoft Visual Studio 2026 version 18.0 or later
+#                        Optional:
+#                             Required to build EBC drivers:
+#                               Intel(r) Compiler for Efi Byte Code (Intel(r) EBC Compiler)
```
os-d: I would drop this

BaseTools/Conf/tools_def.template
```diff
@@ -816,6 +828,142 @@ NOOPT_VS2022_AARCH64_DLINK_FLAGS   = /NOLOGO /NODEFAULTLIB /IGNORE:4001 /IGNORE:
...
+# linked. We do not link the VCRuntime (except for HOST_APPLICATIONs) so this warning can be generated erroneously
+# whenever there are static initializers, because we will always fail the VCRuntime linking check
+  DEBUG_VS2026_X64_DLINK_FLAGS  = /NOLOGO /NODEFAULTLIB /IGNORE:4001 /IGNORE:4281 /OPT:REF /OPT:ICF=10 /MAP /ALIGN:32 /SECTION:.xdata,D /SECTION:.pdata,D /Machine:X64 /LTCG /DLL /ENTRY:$(IMAGE_ENTRY_POINT) /SUBSYSTEM:EFI_BOOT_SERVICE_DRIVER /SAFESEH:NO /BASE:0 /DRIVER /DEBUG /ALIGN:4096 /DLL /WX /IGNORE:4210
+RELEASE_VS2026_X64_DLINK_FLAGS  = /NOLOGO /NODEFAULTLIB /IGNORE:4001 /IGNORE:4281 /IGNORE:4254 /OPT:REF /OPT:ICF=10 /MAP /ALIGN:32 /SECTION:.xdata,D /SECTION:.pdata,D /Machine:X64 /LTCG /DLL /ENTRY:$(IMAGE_ENTRY_POINT) /SUBSYSTEM:EFI_BOOT_SERVICE_DRIVER /SAFESEH:NO /BASE:0 /DRIVER /MERGE:.rdata=.data /ALIGN:4096 /DLL /WX /IGNORE:4210
+NOOPT_VS2026_X64_DLINK_FLAGS    = /NOLOGO /NODEFAULTLIB /IGNORE:4001 /IGNORE:4281 /OPT:REF /OPT:ICF=10 /MAP /ALIGN:32 /SECTION:.xdata,D /SECTION:.pdata,D /Machine:X64 /LTCG /DLL /ENTRY:$(IMAGE_ENTRY_POINT) /SUBSYSTEM:EFI_BOOT_SERVICE_DRIVER /SAFESEH:NO /BASE:0 /DRIVER /DEBUG /ALIGN:4096 /DLL /WX /IGNORE:4210
```
os-d: We can drop the `/Align:32` for these, it gets overridden and not what we want. Previous versions had this, but shouldn't have had both.

BaseTools/Conf/tools_def.template
```diff
@@ -816,6 +828,142 @@ NOOPT_VS2022_AARCH64_DLINK_FLAGS   = /NOLOGO /NODEFAULTLIB /IGNORE:4001 /IGNORE:
...
+
+# BaseTools cannot handle XIP modules with different section and file alignment
+*_VS2026_X64_DLINK_XIPFLAGS     = /ALIGN:32
+
+#####################
+# AARCH64 definitions
```
os-d: TBH, I would be of the opinion that we don't add the MSVC ARM64 definitions to edk2 for VS2026. You can't build the repo with MSVC ARM64 because none of the asm files are in armasm64 format. For awhile we had a downstream workaround for this (essentially porting them over, but even that was a cludge). edk2 has new directly compiled PE target for ARM64 in clangpdb that actually works. Might as well not pollute this with toolchains that don't work and aren't tested. If any users come forward saying they want it and want to make it work, they are welcome to contribute.

EmulatorPkg/Win/Host/WinHost.inf
```diff
@@ -96,6 +96,7 @@
...
   MSFT:*_VS2019_IA32_DLINK_FLAGS     = /LIBPATH:"%VCToolsInstallDir%lib\x86" /LIBPATH:"%UniversalCRTSdkDir%lib\%UCRTVersion%\ucrt\x86" /LIBPATH:"%WindowsSdkDir%lib\%WindowsSDKLibVersion%\um\x86" /NOLOGO /SUBSYSTEM:CONSOLE /NODEFAULTLIB /IGNORE:4086 /MAP /OPT:REF /DEBUG /MACHINE:I386 /LTCG Kernel32.lib MSVCRTD.lib vcruntimed.lib ucrtd.lib Gdi32.lib User32.lib Winmm.lib Advapi32.lib
   MSFT:*_VS2022_IA32_DLINK_FLAGS     = /LIBPATH:"%VCToolsInstallDir%lib\x86" /LIBPATH:"%UniversalCRTSdkDir%lib\%UCRTVersion%\ucrt\x86" /LIBPATH:"%WindowsSdkDir%lib\%WindowsSDKLibVersion%\um\x86" /NOLOGO /SUBSYSTEM:CONSOLE /NODEFAULTLIB /IGNORE:4086 /MAP /OPT:REF /DEBUG /MACHINE:I386 /LTCG Kernel32.lib MSVCRTD.lib vcruntimed.lib ucrtd.lib Gdi32.lib User32.lib Winmm.lib Advapi32.lib
+  MSFT:*_VS2026_IA32_DLINK_FLAGS     = /LIBPATH:"%VCToolsInstallDir%lib\x86" /LIBPATH:"%UniversalCRTSdkDir%lib\%UCRTVersion%\ucrt\x86" /LIBPATH:"%WindowsSdkDir%lib\%WindowsSDKLibVersion%\um\x86" /NOLOGO /SUBSYSTEM:CONSOLE /NODEFAULTLIB /IGNORE:4086 /MAP /OPT:REF /DEBUG /MACHINE:I386 /LTCG Kernel32.lib MSVCRTD.lib vcruntimed.lib ucrtd.lib Gdi32.lib User32.lib Winmm.lib Advapi32.lib
```
os-d: A couple things:
- edk2 style (and why patchcheck is failing) is to have a minimum of one commit per package, e.g. one for EmulatorPkg, one for NetworkPkg, etc.
- For all of these build options that is exactly the same as the ones before it, let's not add a VS2026 specific one, let's actually make this future compatible and do a wildcard instead of VS<year>

os-d (commented):
This doesn't need to be done in this PR, but VS2015 and VS2017 should also be dropped. I can take that up, but certainly if you have interest to, you should feel free.

BaseTools/Conf/tools_def.template
Daniel224455: I think instead of removing the AARCH64 msvc definitions, it would be better to port [this commit](https://github.com/microsoft/mu_silicon_arm_tiano/commit/8aa5e41008e109acf80369ed8b0f1f740d8d0043) over to the edk2 codebase, but I figure thats up for another PR.

EmulatorPkg/Win/Host/WinHost.inf
Daniel224455: Before doing this, I guess VS2015 needs to be dropped first (like you mentioned in the comment above) because the LIBPATH is different

BaseTools/Conf/tools_def.template
Daniel224455: Wasn't probably very smart of me, but I copied that optional dropdown from VS2022, assuming they had the same requirement. Should I proceed with dropping this?

EmulatorPkg/Win/Host/WinHost.inf
Daniel224455: If it does end up being dropped, both `MSFT:*_*_X64_DLINK_FLAGS` and `MSFT:*_*_IA32_DLINK_FLAGS` would be enough. The emulator builds just fine with those.

BaseTools/Conf/tools_def.template
Daniel224455: Should the same be done for all toolchains using /ALIGN:32?

os-d: It can be for everything that also has `/ALIGN:4096`, but I think it's fine to leave the old toolchain as they are, we can just make the new toolchain correct.

EmulatorPkg/Win/Host/WinHost.inf
os-d: I thought (and I may be wrong) that BaseTools will take the first matching one, so that if you have

```c
*_VS2015 = <whatever>
*_* = <whatever>
```

That that will work, even before dropping VS2015

BaseTools/Conf/tools_def.template
os-d: That was reasonable to copy, it's just ancient information. You can't get the EBC compiler anymore (at least not through official channels), so no point carrying the comment forward.

BaseTools/Conf/tools_def.template
os-d: That was originally a plan, but unfortunately it is much more than just that commit that is needed to make MSVC ARM64 work. The biggest part are all of the ASM files don't work in edk2 and have to be written.

It is not worth the investment in MSVC ARM64, it is not used and CLANGPDB ARM64 (that I have a PR up to add support for now) is the better investment.

So, I would just drop MSVC ARM64 definitions here. As I said, it doesn't work in edk2, so no point in having the baggage there.

makubacki:
> This doesn't need to be done in this PR, but VS2015 and VS2017 should also be dropped. I can take that up, but certainly if you have interest to, you should feel free.

I filed a tracking issue for VS2015 - https://github.com/tianocore/edk2/issues/11913.

VS2017 technically has extended support until [April 13, 2027](https://learn.microsoft.com/lifecycle/products/visual-studio-2017). I think a stronger argument could be made for continuing to use that until extended supports ends, so I separated it out from VS2015 for now.

BaseTools/Conf/tools_def.template
ardbiesheuvel: > That was originally a plan, but unfortunately it is much more than just that commit that is needed to make MSVC ARM64 work. The biggest part are all of the ASM files don't work in edk2 and have to be written.
> 
> It is not worth the investment in MSVC ARM64, it is not used and CLANGPDB ARM64 (that I have a PR up to add support for now) is the better investment.
> 

This may be true for building entire platforms. But MSVC ARM64 may still be relevant for building 3rd party UEFI drivers / option ROMs. If porting an x86 driver also implies switching toolchains, it raises the barrier significantly.

This of course presupposes that building UEFI drivers currently works with MSVC ARM64 - could anyone confirm?

os-d: Fair enough. Building UEFI drivers does work with MSVC ARM64, though as always with bringing in different toolchains is going to raise different errors, so edk2 drivers may not compile as is (we also had a large set of fixes for that), but if you have your own UEFI driver and were building it, that should be fine.

Now, to be totally transparent, we have seen some gnarly bad code generation with MSVC ARM64 which was an additional reason for us to move away from it to CLANGPDB. But, again, for the third party case if it is what they want to use, we can leave it in. I wasn’t thinking about the third party driver case, so reasonable enough to leave it in. I also don’t mind doing the, let’s remove it and if anyone yells we can add it back. Because I really don’t think there are many users. It was broken in edk2 completely for awhile and one person a couple years ago did file a bug for it, so there was one user.

Daniel224455: > That was originally a plan, but unfortunately it is much more than just that commit that is needed to make MSVC ARM64 work. The biggest part are all of the ASM files don't work in edk2 and have to be written.
> 
> It is not worth the investment in MSVC ARM64, it is not used and CLANGPDB ARM64 (that I have a PR up to add support for now) is the better investment.
> 
> So, I would just drop MSVC ARM64 definitions here. As I said, it doesn't work in edk2, so no point in having the baggage there.

I'd be fine with removing MSVC ARM64 defs here, but I hope that when it gets upstreamed to Mu, that they get readded. Projects for example such as [Mu-Silicium](https://github.com/Project-Silicium/Mu-Silicium) are able to build with msvc arm64 just fine and boots from SEC -> BDS.

Daniel224455:
Just took care of the reviews, anything else to resolve?

BaseTools/Conf/tools_def.template
lgao4: I just want to confirm DXE_DRIVER also requires 4096 alignment. Right?

Daniel224455:
@os-d Any updates?

BaseTools/Conf/tools_def.template
lgao4: @os-d , can you confirm it?

os-d: @lgao4 yes, DXE_DRIVER requires 4K alignment for memory protections to be applied to the image. What this statement above will do is apply 4K section alignment for everything except for SEC, PEI_CORE, and PEIMs. Those will have 32 byte section alignment (from the XIPFLAGs argument and how that gets used in the build rules).

lgao4: DXE driver uses EfiBootServicesCode type memory. Is it also required to be protected?

lgao4:
@Daniel224455 , there is one CI failure on patch check.

Daniel224455:
> @Daniel224455 , there is one CI failure on patch check.

Resolved, is this considered ready to merge now?

BaseTools/Conf/tools_def.template
os-d: @lgao4 yes, the memory type is irrelevant, we protect all memory (for the most part, there are some gaps). These VS2026 definitions match what we do for VS2022 and GCC.

DXE images will be allocated with the EfiBootServicesCode type and then each image section (which needs to be 4KB aligned) gets RO+X or RW applied to it based on whether it is a code or a data section: https://github.com/tianocore/edk2/blob/4febebe9c2e8c6b53c2814b2ac113bac1ca335c6/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c#L315

lgao4 (approved):

Daniel224455:
> PR can not be merged due to conflict. Please rebase and resubmit

fixed, checks pass

Daniel224455:
@os-d Is there anything else left to do? all reviews are resolved

os-d:
> @os-d Is there anything else left to do? all reviews are resolved

@Daniel224455 nope, you have the appropriate approvals now, so I've added the push label and it will be merged. Thanks for the contribution!

PR #11902, opened 2025-12-18
Add CLANGPDB AArch64 Support

os-d:
# Description

This PR adds CLANGPDB AArch64 support to BaseTools. First, it fixes up the various errors that CLANGPDB catches that other toolchains do not.

These are described in each commit, but in general, the commits are of a couple types:
- ASM file changes
  - There are some ELF asm directives that aren't relevant or usable for PE targets, such as .type, .size, and .previous. These are all worked around
  - There are some relocation types that clang does not support for PE/COFF, including fixup_aarch64_ldr_pcrel_imm19 which is replaced with ADRP/LDR instead.
- Libraries not correctly linked in INF
- Build options that only specify GCC and use GCC style syntax, which breaks CLANGPDB, which uses MSVC style syntax

There is an [LLVM bug](https://github.com/llvm/llvm-project/issues/172660) that causes us to set 4KB section alignment for all CLANGPDB AARCH64 modules. Because XIP modules are required to have the same file alignment and section alignment, this means SEC/PEI CLANGPDB AARCH64 modules will also need a file alignment of 4KB for now.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

`stuart_ci_build -c .pytool\CISettings.py TOOL_CHAIN_TAG=CLANGPDB -a AARCH64` builds and ArmVirtQemu/ArmVirtQmeuKernel boot to shell. ArmVirtCloudHv and ArmVirtKvmtool build.

## Integration Instructions

CLANGPDB is now supported for AARCH64 targets.


os-d:
@ardbiesheuvel can you please help confirm that the ASM changes make sense? I believe these are reasonable approaches to using the same asm file for ELF and PE targets, but please feel free to suggest a different method.

os-d:
@mdkinney @mikebeaton can you please include CLANGPDB AARCH64 in the upcoming CLANGPDB CI changes you are making?

MdePkg/Include/AArch64/AArch64.h
```diff
@@ -150,9 +150,20 @@
...
+#ifdef __ELF__
 #define VECTOR_END(tbl)           \
   .org 0x800;                     \
   .previous
+#else
+#define VECTOR_END(tbl)           \
+  .org 0x800;                     \
+  .section .text.##tbl##,"ax";    \
+  .align 3
+#endif
```
ardbiesheuvel: I'd rather have one copy of the code, even if it means we need to work around the CLANGPDB limitation here as well.

MdePkg/Include/AArch64/AsmMacroLib.h
```diff
@@ -23,20 +23,37 @@
...
   .type     Name, %function       ; \
   Name:                           ; \
   AARCH64_BTI(c)
+#else // __ELF__
+#define _ASM_FUNC(Name, Section)    \
+  .global   Name                  ; \
+  .section  #Section, "ax"        ; \
+  Name:                           ; \
+  AARCH64_BTI(c)
+#endif // __ELF__
```
ardbiesheuvel: Please add a `_ASM_TYPE` macro that incorporates the `#ifdef` and call that from the common implementation of `_ASM_FUNC`

MdePkg/Library/CompilerIntrinsicsLib/AArch64/Atomics.S
```diff
@@ -32,12 +32,26 @@
...
         .section        .text.\name
         .globl          \name
+
+        /*
+         * PE targets (e.g. building with CLANGPDB) do not support ELF style
+         * .type directives
+         */
+#ifdef __ELF__
         .type           \name, %function
+#endif
```
ardbiesheuvel, line 42: Also use `_ASM_TYPE` here, or if that doesn't work inside the `.macro` for any reason, conditionally define another GAS macro and call that.

MdePkg/Library/CompilerIntrinsicsLib/memset.c
```diff
@@ -33,11 +33,29 @@ __memset (
...
   size_t  n
   );
+
+#else // __ELF__
+__attribute__ ((__weak__))
+void *
+memset (
+  void    *dest,
+  int     c,
+  size_t  n
```
ardbiesheuvel, line 45: We can have a single version of these, I don't think having the single extra jump instruction unconditionally is going to matter.

ArmPkg/Drivers/ArmPsciMpServicesDxe/MpFuncs.S
```diff
@@ -31,11 +31,17 @@ GCC_ASM_EXPORT (ApEntryPoint)
 //   );
 ASM_PFX(ApEntryPoint):
   // Configure the MMU and caches
-  ldr x0, gTcr
+  adrp x1, gTcr
+  ldr x1, [x1, :lo12:gTcr]
+  ldr x0, [x1]
```
ardbiesheuvel: This is wrong. You are replacing a single load with two loads.

ADRP+LDR is sufficient here, but the ADRP introduces a new restriction on this code, i.e., that it is aligned to 4k. In this case, it's probably fine because it is a DXE driver, but I don't want to impose this throughout the code base.

So please introduce a `LDR_LIT()` macro that only uses ADRP+LDR if `__ELF__` is not defined.

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
```diff
@@ -35,7 +35,9 @@ _SkipTransferList:
 // to install the stacks at the bottom of the Firmware Device (case the FD is located
 // at the top of the DRAM)
 _SystemMemoryEndInit:
-  ldr   x1, mSystemMemoryEnd
+  adrp  x1, mSystemMemoryEnd
+  ldr   x1, [x1, :lo12:mSystemMemoryEnd]
+  ldr   x1, [x1]
```
ardbiesheuvel: Use `LDR_LIT()` here too.

ArmVirtPkg/Library/ArmPlatformLibQemu/IdMap.S
```diff
@@ -20,7 +20,11 @@
   .set      BLOCK_MEM, TT_TYPE_BLOCK | TT_MT_MEM | TT_AF | TT_XN | TT_NG
 
   .globl    idmap
+#ifdef __ELF__
   .section  ".rodata.idmap", "a", %progbits
+#else
+  .section  ".rodata.idmap", "dr"
+#endif
```
ardbiesheuvel: Just drop the `%progbits` unless this breaks something on GCC/CLANGDWARF

mikebeaton:
> @mdkinney @mikebeaton can you please include CLANGPDB AARCH64 in the upcoming CLANGPDB CI changes you are making?

@os-d: Done - at least from the point of view of providing some (hopefully useful?) feedback as to what now does and doesn't build with your PR.

This is your commits with Azure Pipelines CLANGPDB CI on top: https://github.com/tianocore/edk2/pull/11906/commits

As you can see a lot passes, and there seem to be helpful errors in what doesn't.

The OvmfPkg jobs are not exercising your commits at all, i.e. they do not include AARCH64, but all the other jobs do.

os-d:
> > @mdkinney @mikebeaton can you please include CLANGPDB AARCH64 in the upcoming CLANGPDB CI changes you are making?
> 
> @os-d: Done - at least from the point of view of providing some (hopefully useful?) feedback as to what now does and doesn't build with your PR.
> 
> This is your commits with Azure Pipelines CLANGPDB CI on top: https://github.com/tianocore/edk2/pull/11906/commits
> 
> As you can see a lot passes, and there seem to be helpful errors in what doesn't.
> 
> The OvmfPkg jobs are not exercising your commits at all, i.e. they do not include AARCH64, but all the other jobs do.

Thanks for running this now, I just meant after it is merged to be included in the CI, but you caught me being lazy and not running the other ArmVirtPkg flavors other an ArmVirtQemu :). I have fixed up the additional issues and now ArmVirtQemuKernel also boots to shell and CloudHv and KvmTool build (which is all we do for them in CI). I believe if you pull this patch series now (which history is different on it now) that CI should pass.

mikebeaton:
> Thanks for running this now, I just meant after it is merged to be included in the CI, but you caught me being lazy and not running the other ArmVirtPkg flavors other an ArmVirtQemu :). I have fixed up the additional issues and now ArmVirtQemuKernel also boots to shell and CloudHv and KvmTool build (which is all we do for them in CI). I believe if you pull this patch series now (which history is different on it now) that CI should pass.

We live to serve! I actually thought you were probably asking for both!

You've got a typo /DRVIER for /DRIVER https://github.com/tianocore/edk2/pull/11906/commits :-/

Also, fwiw, there's a typo in subject line of final commit CLANDPDB -> CLANGPDB.

os-d:
> > Thanks for running this now, I just meant after it is merged to be included in the CI, but you caught me being lazy and not running the other ArmVirtPkg flavors other an ArmVirtQemu :). I have fixed up the additional issues and now ArmVirtQemuKernel also boots to shell and CloudHv and KvmTool build (which is all we do for them in CI). I believe if you pull this patch series now (which history is different on it now) that CI should pass.
> 
> We live to serve! I actually thought you were probably asking for both!
> 

Much appreciated!

> You've got a typo /DRVIER for /DRIVER https://github.com/tianocore/edk2/pull/11906/commits :-/
> 
> Also, fwiw, there's a typo in subject line of final commit CLANDPDB -> CLANGPDB.

Whoops :). And the number of times I've typed CLANGPDB in the past two days... :). Both are fixed now, hopefully should pass. I'm also testing locally with enabling linker warnings as errors, as I meant to also do that.

mikebeaton:
> Both are fixed now, hopefully should pass.

All green. Thank you.

> I'm also testing locally with enabling linker warnings as errors, as I meant to also do that.

That'd be good to have too.

os-d:
> > Both are fixed now, hopefully should pass.
> 
> All green. Thank you.
> 

Great, thanks!

> > I'm also testing locally with enabling linker warnings as errors, as I meant to also do that.
> 
> That'd be good to have too.

I pushed this up with a small fix to the StandaloneMmPkg commit. Tested the CI build and the armvirtpkg building/booting.

ArmPkg/Drivers/ArmPsciMpServicesDxe/MpFuncs.S
```diff
@@ -31,11 +31,14 @@ GCC_ASM_EXPORT (ApEntryPoint)
 //   );
 ASM_PFX(ApEntryPoint):
   // Configure the MMU and caches
-  ldr x0, gTcr
+  adrp x0, gTcr
+  ldr x0, [x0, :lo12:gTcr]
```
ardbiesheuvel: This introduces a 4k section alignment requirement that did not exist before, so I'd prefer if we could do this conditionally. I.e., introduce a `ldr_lit` GAS macro that implements one or the other depending on whether `__ELF__` is defined.

(I know it doesn't make any difference here, but in the SEC/PEIM context it does, and existing platforms might run out of flash space and fail to build due to this)

E.g.,

```
    .macro ldr_lit, dst:req, sym:req, tmp
#ifdef __ELF__
    ldr     \dst, \sym
#else
    .ifnb   \tmp
    adrp    \tmp, \sym
    ldr     \dst, [\tmp, :lo12:\sym]
    .else
    adrp    \dst, sym
    ldr     \dst, [\dst, :lo12:\sym]
#endif
    .endm
```

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
```diff
@@ -35,7 +35,8 @@ _SkipTransferList:
 // to install the stacks at the bottom of the Firmware Device (case the FD is located
 // at the top of the DRAM)
 _SystemMemoryEndInit:
-  ldr   x1, mSystemMemoryEnd
+  adrp  x1, mSystemMemoryEnd
+  ldr   x1, [x1, :lo12:mSystemMemoryEnd]
```
ardbiesheuvel: Use `ldr_lit`

mikebeaton:
> > > I'm also testing locally with enabling linker warnings as errors, as I meant to also do that.
> > 
> > That'd be good to have too.
> 
> I pushed this up

Thanks.

> with a small fix to the StandaloneMmPkg commit

cc @mhaeuser: in case you have any feedback, or just to notify you of the change/fix here https://github.com/tianocore/edk2/pull/11902/changes/9f856e7919489e6266082ca1cf9cd300c08392e2

> Tested the CI build and the armvirtpkg building/booting

Thanks. Pushed to temporary CI again in the same location: https://github.com/tianocore/edk2/pull/11906/commits - looks good.

mhaeuser:
> cc @mhaeuser: in case you have any feedback, or just to notify you of the change/fix here [9f856e7](https://github.com/tianocore/edk2/commit/9f856e7919489e6266082ca1cf9cd300c08392e2)

Well, the runtime code situation is still the same. Maybe there should be a macro or PCD or something, so the runtime code can either omit reloc when not needed, or emit and STATIC_ASSERT when built with a non-PIE toolchain.

Is there an issue with briefly making .text RWX to apply the relocs?

Or, a much cleaner yet much over-the-top solution, move all self-reloc code to its own section. This can double as a sanity-check that it’s reloc-free, solving both this issue, and ensuring no code pre-reloc relies on relocs.

MdePkg/Include/AArch64/ProcessorBind.h
```diff
@@ -179,10 +179,20 @@ typedef INT64 INTN;
 ///
 #define ASM_GLOBAL  .globl
 
+//
+// PE targets (i.e. building with CLANGPDB) do not support the ELF style
+// .type directive
+//
+  #ifdef __ELF__
 #define GCC_ASM_EXPORT(func__)  \
          .global  _CONCATENATE (__USER_LABEL_PREFIX__, func__)    ;\
```
ardbiesheuvel: While at it, mind fixing this macro? No need to open code the concatenation in the first line, and then use ASM_PFX() [which amounts to the same thing] in the second.

ArmVirtPkg/PrePi/ModuleEntryPoint.S
```diff
@@ -120,7 +124,9 @@ ASM_PFX(DiscoverDramFromDt):
   // The base of the runtime image has been preserved in x27. Check whether
   // the expected magic number can be found in the header.
   //
-  ldr   w8, .LArm64LinuxMagic
+  adrp  x8, .LArm64LinuxMagic
+  add   x8, x8, #:lo12:.LArm64LinuxMagic
+  ldr   w8, [x8]
```
ardbiesheuvel: Here you could use `ldr_lit w8, .LArm64LinuxMagic, x8`

ArmVirtPkg/ArmVirt.dsc.inc
```diff
@@ -30,10 +30,14 @@ DEFINE FD_SIZE_IN_MB    = 3
 !endif
 
 [BuildOptions.common.EDKII.DXE_CORE,BuildOptions.common.EDKII.DXE_DRIVER,BuildOptions.common.EDKII.UEFI_DRIVER,BuildOptions.common.EDKII.UEFI_APPLICATION]
-  GCC:*_*_*_DLINK_FLAGS = -z common-page-size=0x1000
+  GCC:*_GCC_*_DLINK_FLAGS = -z common-page-size=0x1000
+  GCC:*_CLANGDWARF_*_DLINK_FLAGS = -z common-page-size=0x1000
```
ardbiesheuvel, line 34: Is this needed? Is CLANGPDB in the GCC family too?

CryptoPkg/Library/OpensslLib/OpensslGen/AARCH64-GCC/crypto/aes/aesv8-armx.S
```diff
@@ -183,10 +185,15 @@ aes_v8_set_encrypt_key:
         mov     x0,x3                   // return value
         ldr     x29,[sp],#16
         ret
+#ifdef __ELF__  // PE targets (e.g. CLANGPDB) don't use the ELF style .size directive
```
ardbiesheuvel: This is terrible. These files are generated from OpenSSL 'perlasm' style sources, and so these patches need to be re-applied / carried forward each time.

It would be better if the upstream source generated a dialect that CLANGPDB can consume. Can we just disable these files for the time being? I don't think any of these are a bottleneck during boot, right?

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibReplaceEntry.S
```diff
@@ -97,10 +97,10 @@ ASM_FUNC_ALIGN(ArmReplaceLiveTranslationEntry, 0x200)
 ASM_GLOBAL ASM_PFX(ArmReplaceLiveTranslationEntrySize)
 
 ASM_PFX(ArmReplaceLiveTranslationEntrySize):
-  .long   . - ArmReplaceLiveTranslationEntry
+  .long   . - 0b
```
ardbiesheuvel: Why is this needed? Would it help to use ASM_PFX()?

UefiPayloadPkg/UefiPayloadPkg.dsc
```diff
@@ -189,7 +189,8 @@
   *_*_*_CC_FLAGS                 = $(APPEND_CC_FLAGS)
 
 [BuildOptions.AARCH64]
-  GCC:*_*_*_CC_FLAGS         = -mcmodel=tiny -mstrict-align
+  GCC:*_GCC_*_CC_FLAGS         = -mcmodel=tiny -mstrict-align
+  GCC:*_CLANGDWARF_*_CC_FLAGS  = -mcmodel=tiny -mstrict-align
```
ardbiesheuvel: I suspect that this is needed because the 4k alignment is not guaranteed, so wishing it away is not going to help.

ardbiesheuvel (dismissed):

os-d:
> > cc @mhaeuser: in case you have any feedback, or just to notify you of the change/fix here [9f856e7](https://github.com/tianocore/edk2/commit/9f856e7919489e6266082ca1cf9cd300c08392e2)
> 
> Well, the runtime code situation is still the same. Maybe there should be a macro or PCD or something, so the runtime code can either omit reloc when not needed, or emit and STATIC_ASSERT when built with a non-PIE toolchain.
> 
> Is there an issue with briefly making .text RWX to apply the relocs?
> 
> Or, a much cleaner yet much over-the-top solution, move all self-reloc code to its own section. This can double as a sanity-check that it’s reloc-free, solving both this issue, and ensuring no code pre-reloc relies on relocs.

Over [here](https://github.com/tianocore/edk2/pull/11886#issuecomment-3657899438) Ard mentioned that the self-reloc case was for platforms that run StMM as an application in OP-TEE. I am assuming OP-TEE only can run ELF images (maybe I am incorrect), in which case for the specific scenario, it would be a moot point for CLANGPDB support. However, in theory, the self reloc case could work for a standard system as well, but it isn't required. If you have a proposed change, I think that would be reasonable to explore, but I would prefer to take that in a separate PR.

MdePkg/Include/AArch64/AArch64.h
os-d: Totally fair. I'll change this and other places to have single copies of the code when possible.

ArmPkg/Drivers/ArmPsciMpServicesDxe/MpFuncs.S
os-d: Looks like you commented on an earlier commit that I had previously fixed. Agreed that was totally incorrect, that was during an experimentation of how best to fix it and I realized that was wrong and so fixed the instances. Please check the current commits and see if that looks right.

I will introduce the macro. I was concerned by the ADRP introduction in these various places as well, due to your explanation on the other PR of the 4K alignment requirement. We'll get that with the current CLANGPDB settings (and agreed that we don't need to impose that on other toolchains). Do you think that relaxing these to ADR in the SEC/PEI ASM files makes sense? Those modules should be < 1 MB and then we don't have the 4K alignment requirement. If ADR was used in a module that did have the symbol > 1 MB apart, I assume the build would fail so we could catch it?

ArmVirtPkg/ArmVirt.dsc.inc
os-d: Confusingly, yes, CLANGPDB is in the GCC family: https://github.com/tianocore/edk2/blob/4cf6f2f32d3b0bba5d4ea2dc14baa688fb50f583/BaseTools/Conf/tools_def.template#L1649.

At this point, it would probably be better to have clang be its own family or CLANGDWARF be GCC family and CLANGPDB be MSVC family. I think I recall @mdkinney had thoughts on this at one point, I may be misremembering.

CryptoPkg/Library/OpensslLib/OpensslGen/AARCH64-GCC/crypto/aes/aesv8-armx.S
os-d: I'm not sure what the perf gains are for a real platform, I assume of course that HW acceleration is going to be faster, but I don't know how much these paths are actually hit to make it matter. I didn't know these were generated from OpenSSL, I agree that having an override is not good, I can see about taking it upstream.

For this PR, I can remove the two accel libs (I think the only things that include them) from the CryptoPkg CI build for CLANGPDB AARCH64, which is why I applied these fixes in the first place, the CI build was failing on them.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibReplaceEntry.S
os-d: Let me try that. It was a build failure saying PE/COFF targets don't support that reloc type, I believe, but I'll try again to confirm.

UefiPayloadPkg/UefiPayloadPkg.dsc
os-d: i.e. that coreboot (or generically the pre-UEFI bootloader) isn't guaranteed to load UefiPayloadPkg at 4KB alignment? For CLANGPDB, currently, we are guaranteeing 4KB section alignment.

These build options were added with the addition of UefiPayloadPkg's AARCH64 support, with no mention of it, so I'm not sure if it was an intentional choice or copying from some place else without a specific need. @amosbu and @AjanZhong you added [the commit](https://github.com/tianocore/edk2/commit/f0a20153731073799cde14da8d362fa9d6529f6b) that added AARCH64 support to UefiPayloadPkg, can you comment on why you chose the tiny code model and -mstrict-align?

ArmPkg/Drivers/ArmPsciMpServicesDxe/MpFuncs.S
ardbiesheuvel: Yes, I noticed that you had uploaded a new set in the meantime, but this comment had completely disappeared from my view so I did not realize it was still staged somewhere in the bowels of GitHub.

ArmPlatformPkg/PeilessSec/AArch64/ModuleEntryPoint.S
ardbiesheuvel: Please use the GAS macro instead - this comment had also become invisible to me when you uploaded a new set while I was in the middle of review

ArmVirtPkg/ArmVirt.dsc.inc
ardbiesheuvel: Yeah, that needs fixing imho, or we will have many more of these warts going forward.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibReplaceEntry.S
ardbiesheuvel: Right, that would be due to the fact that `ArmReplaceLiveTranslationEntry` has external visibility, and the local label does not, allowing the assembler to perform the offset calculation directly, rather than relying on the linker.

Please use `.LArmReplaceLiveTranslationEntry` instead - numbered labels have a nasty habit of going out of sync.

UefiPayloadPkg/UefiPayloadPkg.dsc
ardbiesheuvel: Guaranteeing 4k section alignment in the build only helps if the loader adheres to it as well.

The `-mstrict-align` is likely needed because some of the code runs with the MMU and caches off, and so the CPU does not permit unaligned accesses.

ardbiesheuvel:
> > > cc @mhaeuser: in case you have any feedback, or just to notify you of the change/fix here [9f856e7](https://github.com/tianocore/edk2/commit/9f856e7919489e6266082ca1cf9cd300c08392e2)
> > 
> > 
> > Well, the runtime code situation is still the same. Maybe there should be a macro or PCD or something, so the runtime code can either omit reloc when not needed, or emit and STATIC_ASSERT when built with a non-PIE toolchain.
> > Is there an issue with briefly making .text RWX to apply the relocs?
> > Or, a much cleaner yet much over-the-top solution, move all self-reloc code to its own section. This can double as a sanity-check that it’s reloc-free, solving both this issue, and ensuring no code pre-reloc relies on relocs.
> 
> Over [here](https://github.com/tianocore/edk2/pull/11886#issuecomment-3657899438) Ard mentioned that the self-reloc case was for platforms that run StMM as an application in OP-TEE. I am assuming OP-TEE only can run ELF images (maybe I am incorrect), in which case for the specific scenario, it would be a moot point for CLANGPDB support. However, in theory, the self reloc case could work for a standard system as well, but it isn't required. If you have a proposed change, I think that would be reasonable to explore, but I would prefer to take that in a separate PR.

AFAIK the StMM pseudo-TA is not a loadable TA but one that gets incorporated into the OP-TEE build, and so I don't /think/ that it consumes the ELF but I'm not sure. @apalos?

ArmVirtPkg/ArmVirt.dsc.inc
mikebeaton: Don't forget there is also _BUILDRULEFAMILY:
```
$ grep -P "_[A-Z]*FAMILY" tools_def.template 
# *_*_*_ASL_FAMILY                   = INTEL
# *_*_*_ASL_FAMILY                   = MSFT
*_VS2015_*_*_FAMILY               = MSFT
*_VS2015x86_*_*_FAMILY        = MSFT
*_VS2017_*_*_FAMILY        = MSFT
*_VS2019_*_*_FAMILY        = MSFT
*_VS2022_*_*_FAMILY        = MSFT
*_GCC48_*_*_FAMILY               = GCC
*_GCC49_*_*_FAMILY               = GCC
*_GCCNOLTO_*_*_FAMILY               = GCC
*_GCC5_*_*_FAMILY                = GCC
*_GCC_*_*_FAMILY                = GCC
*_CLANGPDB_*_*_FAMILY                = GCC
*_CLANGPDB_*_*_BUILDRULEFAMILY       = CLANGPDB
*_CLANGDWARF_*_*_FAMILY             = GCC
*_CLANGDWARF_*_*_BUILDRULEFAMILY    = CLANGDWARF
*_XCODE5_*_*_FAMILY            = GCC
*_XCODE5_*_*_BUILDRULEFAMILY   = XCODE
```

mikebeaton: Presumably `*_CLANGPDB_*_*_BUILDRULEFAMILY       = CLANGPDB` means something a bit like `MSFT`, but not enough like it to actually use that.

mhaeuser:
> Over [here](https://github.com/tianocore/edk2/pull/11886#issuecomment-3657899438) Ard mentioned that the self-reloc case was for platforms that run StMM as an application in OP-TEE. I am assuming OP-TEE only can run ELF images (maybe I am incorrect), in which case for the specific scenario, it would be a moot point for CLANGPDB support. However, in theory, the self reloc case could work for a standard system as well, but it isn't required. If you have a proposed change, I think that would be reasonable to explore, but I would prefer to take that in a separate PR.

I would be very surprised if the environment was ELF-aware, but still doesn’t perform relocation or apply permissions (the comments state we start with RX). Sounds like a blob that is simply jumped into to me.

The last part was just thinking out loud on how one could support CLANGPDB if one wanted to. The first part I think is part of the immediate change set. I don’t think a comment about CLANGPDB is sufficient (that’s what I did the first time, because it wouldn’t build in the first place and thus this was easiest). I.e., I don’t think one should be able to compile builds that can fault (under the respective precondition fixed-/flexible-address, which needs some way to distinguish between).

ArmVirtPkg/ArmVirt.dsc.inc
os-d: Yeah, the problem is that fanily is used much more than buildrulefamily, and if you don’t specify the buildrulefamily for CLANGPDB it will end up inheriting the family. I _think_ that

*_CLANGPDB_* = <>
GCC:*_*_* = <something else>

Will work, but again, it is way too easy to accidentally loop in CLANGPDB and break the build.

os-d:
> > Over [here](https://github.com/tianocore/edk2/pull/11886#issuecomment-3657899438) Ard mentioned that the self-reloc case was for platforms that run StMM as an application in OP-TEE. I am assuming OP-TEE only can run ELF images (maybe I am incorrect), in which case for the specific scenario, it would be a moot point for CLANGPDB support. However, in theory, the self reloc case could work for a standard system as well, but it isn't required. If you have a proposed change, I think that would be reasonable to explore, but I would prefer to take that in a separate PR.
> 
> I would be very surprised if the environment was ELF-aware, but still doesn’t perform relocation or apply permissions (the comments state we start with RX). Sounds like a blob that is simply jumped into to me.
> 
> The last part was just thinking out loud on how one could support CLANGPDB if one wanted to. The first part I think is part of the immediate change set. I don’t think a comment about CLANGPDB is sufficient (that’s what I did the first time, because it wouldn’t build in the first place and thus this was easiest). I.e., I don’t think one should be able to compile builds that can fault (under the respective precondition fixed-/flexible-address, which needs some way to distinguish between).

Let me take a look, I think it would be easy to have a static assert as you mentioned.

UefiPayloadPkg/UefiPayloadPkg.dsc
os-d: If we think that these are hard requirements (which I think is fair, not knowing what the pre-UEFI bootloader is going to do) and we don’t have a good way to support this (the build fails if I try to use the tiny code model with CLANGPDB, support has not been added for PE targets), I’m fine to have UefiPayloadPkg be not compilable with CLANGPDB AARCH64 in this first PR and deal with it separately.

ArmVirtPkg/ArmVirt.dsc.inc
os-d: Before changing the family, I want to get @mdkinney ’s thoughts, because I know he has been working on a large set of changes for CLANGPDB for x86 and I don’t want to invalidate some of his efforts here. I would presume it would also benefit his side.

mdkinney: @ardbiesheuvel I do have a number of changes for clang enhancements queued up.  I prefer to get all of those reviewed and merged first along with Wiki write ups on the different clang compiler environments.  After that, I think a review of all the tool chains with a focus on clang can be done to see what simplifications can be done.  I suggest me move that topic to a sub-discussion from OSFC.

apalos:
> > > > cc @mhaeuser: in case you have any feedback, or just to notify you of the change/fix here [9f856e7](https://github.com/tianocore/edk2/commit/9f856e7919489e6266082ca1cf9cd300c08392e2)
> > > 
> > > 
> > > Well, the runtime code situation is still the same. Maybe there should be a macro or PCD or something, so the runtime code can either omit reloc when not needed, or emit and STATIC_ASSERT when built with a non-PIE toolchain.
> > > Is there an issue with briefly making .text RWX to apply the relocs?
> > > Or, a much cleaner yet much over-the-top solution, move all self-reloc code to its own section. This can double as a sanity-check that it’s reloc-free, solving both this issue, and ensuring no code pre-reloc relies on relocs.
> > 
> > 
> > Over [here](https://github.com/tianocore/edk2/pull/11886#issuecomment-3657899438) Ard mentioned that the self-reloc case was for platforms that run StMM as an application in OP-TEE. I am assuming OP-TEE only can run ELF images (maybe I am incorrect), in which case for the specific scenario, it would be a moot point for CLANGPDB support. However, in theory, the self reloc case could work for a standard system as well, but it isn't required. If you have a proposed change, I think that would be reasonable to explore, but I would prefer to take that in a separate PR.
> 
> AFAIK the StMM pseudo-TA is not a loadable TA but one that gets incorporated into the OP-TEE build, and so I don't /think/ that it consumes the ELF but I'm not sure. @apalos?

Exactly @ardbiesheuvel . We basically append it at the end of OP-TEE, jump to the first instruction and expect it to relocate it self and run. In fact StMM is not treated as a TA or a pseudo-TA. It's treated as a 'special' TA, which is sandboxed from the rest of the core/TAs etc

ArmVirtPkg/ArmVirt.dsc.inc
os-d: Resolving this for now to come back to, per Mike's request.

CryptoPkg/Library/OpensslLib/OpensslGen/AARCH64-GCC/crypto/aes/aesv8-armx.S
os-d: I was able to create a new AARCH64-CLANGPDB target in edk2's wrapper of the openssl configure script (since openssl already has the support) and generate the correct files. In the next set I push up I will have a new commit with this and dropping changing all the GCC files.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibReplaceEntry.S
os-d: I actually wasn't able to repro the build failure when I took this back, either on the UefiCpuPkg CI build (which I had to amend to add AARCH64 support) or the ArmVirtQemu build (which ran to shell and obviously uses this lib). I will drop the changes to this file for now and if it comes back later we can use the named label.

UefiPayloadPkg/UefiPayloadPkg.dsc
os-d: Oh, going back to why I wondered if this was actually needed: this is set in UefiPayloadPkg.dsc, not in any of the actual INFs. So the tiny code model and mstrict-align are only used when building this for CI. That's why I wondered if this was just a copy/paste from somewhere else. The actual FW builds would need to set these build flags, since they aren't enforced at the INF level. I don't see any edk2-platforms AARCH64 usage of UefiPayloadPkg, so I don't have any reference to what people are doing.

What I have done for now is preserve -mstrict-align in the CI build, because CLANGPDB does support that (I missed that before), but I have dropped the tiny code model, since CLANGPDB doesn't support that. In this way CLANGPDB can be used to be UefiPayloadPkg in CI and either it will "just work" for platforms, if they weren't overriding the build options already and if they were, they would need to ensure their pre-UEFI bootloader loads UefiPayloadPkg on a 4KB boundary (which is of course easy to do), but that's a platform decision.

os-d:
@mhaeuser @ardbiesheuvel forgive my ignorance on PIE modules, but returning the Standalone MM Core self relocation case, why does that require a PIE image? Isn't the point of PIE that it doesn't need relocation?

I would think that [this code](https://github.com/tianocore/edk2/blob/4febebe9c2e8c6b53c2814b2ac113bac1ca335c6/ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c#L1108-L1115) performs the StMM Core self relocation when required and all that that needs to work is for the .text section to be set as RW, apply the relocations, then update the image memory protections. Am I missing something on why that wouldn't work?

ArmPkg/Drivers/ArmPsciMpServicesDxe/MpFuncs.S
os-d: I ended up making this a C macro instead of a GAS macro, it's less flexible, but it was easier for me to share across the various places that needed to use it. With the GAS macro, I think I would need to introduce a new .inc file to share it. Let me know if this approach is acceptable, otherwise I can change it.

os-d:
@ardbiesheuvel I have addressed all of your comments (I have left some open if there are still open questions on them), can you please re-review?

mhaeuser:
> why does that require a PIE image? Isn't the point of PIE that it doesn't need relocation?

Nope. I thought the same thing when I started merging my revamped PE loader into our EDK II fork. So I looked at all the reference sites and always saw relocations, no matter what. I even sometimes read comments (no clue what was in EDK II and what was Google-foo) that indicated so, even if the code disagreed. Read explanations online, they contradicted, read more code, most of it undocumented, slammed my head on the desk. Oh well.

Anyway... a non-position-independent object will always load at its link address. So, in fact, there are no (runtime) relocations. PIC and PIEs are those that can be loaded at semi-arbitrary addresses and this is achieved by a combination of relative addressing and relocations. At this level of abstraction, it's the exact same deal as MSVC/PE (and this threw me off next...).

The special sauce is in one constraint we do not care about in EDK II: Page sharing. ELF toolchains implement PIE by forbidding (runtime) relocations that target read-only segments. This is so processes can share all code and read-only data pages. Anything that requires absolute addressing in these segments is indirected via the Global Offset Table (GOT). Basically (maybe a bit too simplified), it's a "middle-man" array that is accessed via relative addressing by the read-only segments and indexes all absolute addresses of the actual targets. So, all (runtime) relocations that would target read-only segments now target GOT. This compresses the non-sharable data of the read-only segments together, minimizing the amount of per-process pages. In UEFI, none of this exists, so we do not care. There is one special benefit though, which I will mention below.

> I would think that [this code](https://github.com/tianocore/edk2/blob/4febebe9c2e8c6b53c2814b2ac113bac1ca335c6/ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c#L1108-L1115) performs the StMM Core self relocation when required and all that that needs to work is for the .text section to be set as RW, apply the relocations, then update the image memory protections. Am I missing something on why that wouldn't work?

If the platform permits this, yes, that would work (though remember as it is self-relocation, it's RWX, not RW). Self-relocation is always ugly and should be avoided if possible. If unavoidable, ELF-style PIE has another benefit: We do not need to make .text writable, so W^X is preserved. I think at least Apple enforces W^X (at least past a point) on the hardware level now, so no clue how future-proof this is.

os-d:
> > why does that require a PIE image? Isn't the point of PIE that it doesn't need relocation?
> 
> Nope. I thought the same thing when I started merging my revamped PE loader into our EDK II fork. So I looked at all the reference sites and always saw relocations, no matter what. I even sometimes read comments (no clue what was in EDK II and what was Google-foo) that indicated so, even if the code disagreed. Read explanations online, they contradicted, read more code, most of it undocumented, slammed my head on the desk. Oh well.
> 
> Anyway... a non-position-independent object will always load at its link address. So, in fact, there are no (runtime) relocations. PIC and PIEs are those that can be loaded at semi-arbitrary addresses and this is achieved by a combination of relative addressing and relocations. At this level of abstraction, it's the exact same deal as MSVC/PE (and this threw me off next...).
> 
> The special sauce is in one constraint we do not care about in EDK II: Page sharing. ELF toolchains implement PIE by forbidding (runtime) relocations that target read-only segments. This is so processes can share all code and read-only data pages. Anything that requires absolute addressing in these segments is indirected via the Global Offset Table (GOT). Basically (maybe a bit too simplified), it's a "middle-man" array that is accessed via relative addressing by the read-only segments and indexes all absolute addresses of the actual targets. So, all (runtime) relocations that would target read-only segments now target GOT. This compresses the non-sharable data of the read-only segments together, minimizing the amount of per-process pages. In UEFI, none of this exists, so we do not care. There is one special benefit though, which I will mention below.
> 
> > I would think that [this code](https://github.com/tianocore/edk2/blob/4febebe9c2e8c6b53c2814b2ac113bac1ca335c6/ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.c#L1108-L1115) performs the StMM Core self relocation when required and all that that needs to work is for the .text section to be set as RW, apply the relocations, then update the image memory protections. Am I missing something on why that wouldn't work?
> 
> If the platform permits this, yes, that would work (though remember as it is self-relocation, it's RWX, not RW). Self-relocation is always ugly and should be avoided if possible. If unavoidable, ELF-style PIE has another benefit: We do not need to make .text writable, so W^X is preserved. I think at least Apple enforces W^X (at least past a point) on the hardware level now, so no clue how future-proof this is.

Thanks, this was a helpful description and I understand the difference now.

I have run into the first issue with just setting the memory permissions to RWX, that works without Hafnium, but Hafnium will block RWX from being set. So you have to play some more games here, but as Ard pointed out, the only use case for the self reloc code is for the OP-TEE case and he said that was acceptable to leave to the platform to ensure it is built with a pie toolchain. I'll play around a bit more here for a not too intrusive option, but for this PR I am going to leave it as is. The general use case (fixed address) works.

os-d:
FWIW, I did get a hackjob working with your idea @mhaeuser to put all the functions that needed fixing up in a separate image section. Then I could mark that section as RW, apply the fixups, then mark it as RO+X again. However, that is obviously incredibly brittle because a new call can be added in some function and a new fixup location is required. To make that work, you'd have to have a build plugin that checks the relocs and that none are in .text.

The other option would be to copy StMM Core to another location, mark that RW, apply fixups, then jump to there. You'd of course have to ensure that the code leading up to the jump doesn't have any fixup locations in it or else you would crash. That is an issue with the self relocation today, we can't have any references to the GOT before we fixup the GOT or else we'll be pointing to the wrong spot. I don't think anything about being a PIE image would help with that, because the toolchain doesn't know when we are applying fixups.

So, going back to, as long as Ard is okay with it, which it seems he is, I will leave this alone for now and the small use case for a self relocating StMM Core will need to know it has to build it as a PIE image. I'd be fine if we want to add a PCD that we can key off of and have a static assert that catches if you want self relocation and aren't building a PIE image.

mhaeuser:
@os-d I hope I wasn’t misleading, I did not suggest the separate section idea specifically to implement MSVC self-relocation (though yes, it would be helpful for that too). My main concern was what’s in your second paragraph, that ELF PIE is still vulnerable to bugs of running relocation-dependent code pre-relocation. And yes, the idea only makes proper sense with a separate tool that performs sanity checks. Though I mentioned this as a “this would be a nice and comprehensive approach to making self-relocation safer and easier to validate”, not as a suggestion to fix this specific issue. I believed and still believe a static assert is the way to go for now, especially as there seems to be no demand to build this with MSVC. Sorry for any confusion.

ardbiesheuvel:
As I said, we shouldn't obsess too much about the self relocation case, given that it is a bit of a hack anyway. The only way to do this safely is to implement the relocation logic in assembler, in order to be 100% sure that it does not rely on any absolute references itself before it is done fixing them up.

Building with `-z text` guarantees that the R-X region is free of absolute references subject to runtime relocation, and the GOT is placed here as well, and so the current code only works if the GOT is empty. On AArch64, this is usually the case when building with `-Bsymbolic` because the linker can relax all ADRP/LDR GOT references into ADRP/ADD direct references. Therefore, the resulting image should always tolerate self-relocation (modulo the issue where absolute references could theoretically be dereferenced before the relocation logic completes)

On x86_64, it is not always possible to relax GOT references, as they may appear anywhere as memory operands. So attempting to reuse this logic would likely result in permission faults on the R-X mapped GOT.

If we ever do this properly, I would suggest asm implementations that can apply PE/COFF relocations, rather than staying in the ELF domain.

Btw not sure if you spotted this but `ArmVirtQemuKernel` does something similar.

mhaeuser:
@ardbiesheuvel Huh, why would R-X GOT permission-fault? Accesses by the code would be strictly read-only and even with RELRO, GOT should start off as RW, so it should be turned RW before applying relocations (unless .got is merged into .text, which I have seen before, but that seems like a bug of its own; it should be merged into .rodata with write-toggling for applying relocations). I thought the “only” issue was accessing addresses before they were updated.

Regarding ASM, I don’t think I agree, the less ASM the better for everyone. At least with Clang and GCC, tooling should be able to make sure the generated image is safe (by leveraging link-time relocations to understand GOT and .data dependencies; I’m not familiar enough with MSVC to tell how hard solving this would be there). I’m doing something similar with Intel DOIT in LLVM right now, where after codegen I have an additional pass to ensure that all instructions flagged as constant-time are validated to have DOIT opcodes - no manual ASM required :).

ardbiesheuvel:
> @ardbiesheuvel Huh, why would R-X GOT permission-fault? Accesses by the code would be strictly read-only and even with RELRO, GOT should start off as RW, so it should be turned RW before applying relocations (unless .got is merged into .text, which I have seen before, but that seems like a bug of its own; it should be merged into .rodata with write-toggling for applying relocations).

`GccBase.lds` merges the GOT into the `.text` section. Note that we are talking about ELF to PE/COFF conversion here, so RELRO does not mean anything here. The best approximation of RELRO in our case is putting anything that needs to be relocated by the loader but is constant from the PoV of the program into .text.

MdePkg/Include/AArch64/AArch64.h
```diff
@@ -150,9 +150,21 @@
...
 
+//
+// PE targets (i.e. building with CLANGPDB) do not support the ELF style
+// .previous directive and so we need to manually define the section again.
+//
+#ifdef __ELF__
+#define _PREVIOUS_SECTION()  .previous
+#else // __ELF__
+#define _PREVIOUS_SECTION() \
+  .section .text.##tbl##,"ax";    \
```
ardbiesheuvel: Where is `tbl` defined?

ArmPkg/Drivers/ArmPsciMpServicesDxe/MpFuncs.S
```diff
@@ -31,11 +31,11 @@ GCC_ASM_EXPORT (ApEntryPoint)
 //   );
 ASM_PFX(ApEntryPoint):
   // Configure the MMU and caches
-  ldr x0, gTcr
+  LDR_LIT(x0, gTcr)
```
ardbiesheuvel, line 34: Code changes look good now but please update the commit log

BaseTools/Conf/tools_def.template
```diff
@@ -1757,6 +1760,47 @@ NOOPT_CLANGPDB_X64_DLINK_FLAGS      = /NOLOGO /NODEFAULTLIB /IGNORE:4001 /OPT:RE
...
+*_CLANGPDB_AARCH64_CC_PATH              = DEF(CLANGPDB_AARCH64_PREFIX)clang
+*_CLANGPDB_AARCH64_SLINK_PATH           = DEF(CLANGPDB_AARCH64_PREFIX)llvm-lib
+*_CLANGPDB_AARCH64_DLINK_PATH           = DEF(CLANGPDB_AARCH64_PREFIX)lld-link
+*_CLANGPDB_AARCH64_ASLDLINK_PATH        = DEF(CLANGPDB_AARCH64_PREFIX)lld-link
+*_CLANGPDB_AARCH64_ASM_PATH             = DEF(CLANGPDB_AARCH64_PREFIX)clang
+*_CLANGPDB_AARCH64_PP_PATH              = DEF(CLANGPDB_AARCH64_PREFIX)clang
+*_CLANGPDB_AARCH64_VFRPP_PATH           = DEF(CLANGPDB_AARCH64_PREFIX)clang
+*_CLANGPDB_AARCH64_ASLCC_PATH           = DEF(CLANGPDB_AARCH64_PREFIX)clang
+*_CLANGPDB_AARCH64_ASLPP_PATH           = DEF(CLANGPDB_AARCH64_PREFIX)clang
+*_CLANGPDB_AARCH64_RC_PATH              = DEF(CLANGPDB_IA32_PREFIX)llvm-rc
```
ardbiesheuvel: Is the prefix correct here?

BaseTools/Conf/build_rule.template
```diff
@@ -342,6 +342,8 @@
         "$(DLINK)" -o ${dst} $(DLINK_FLAGS) $(DLINK_XIPFLAGS) -Wl,--start-group,@$(STATIC_LIBRARY_FILES_LIST),--end-group $(CC_FLAGS) $(CC_XIPFLAGS) $(DLINK2_FLAGS)
```
ardbiesheuvel, line 342: The commit log says `CLANDPDB`

BaseTools/Conf/tools_def.template
```diff
@@ -1757,6 +1760,47 @@ NOOPT_CLANGPDB_X64_DLINK_FLAGS      = /NOLOGO /NODEFAULTLIB /IGNORE:4001 /OPT:RE
...
+*_CLANGPDB_AARCH64_ASLCC_FLAGS          = DEF(GCC_ASLCC_FLAGS) -m64 -fno-lto DEF(CLANGPDB_AARCH64_TARGET)
+*_CLANGPDB_AARCH64_ASM_FLAGS            = DEF(GCC_ASM_FLAGS) -m64 DEF(CLANGPDB_AARCH64_TARGET)
+*_CLANGPDB_AARCH64_OBJCOPY_FLAGS        =
+*_CLANGPDB_AARCH64_NASM_FLAGS           = -f win64
+*_CLANGPDB_AARCH64_PP_FLAGS             = DEF(GCC_PP_FLAGS) DEF(CLANGPDB_AARCH64_TARGET)
+*_CLANGPDB_AARCH64_ASLPP_FLAGS          = DEF(GCC_ASLPP_FLAGS) DEF(CLANGPDB_AARCH64_TARGET)
+*_CLANGPDB_AARCH64_VFRPP_FLAGS          = DEF(GCC_VFRPP_FLAGS) DEF(CLANGPDB_AARCH64_TARGET)
+*_CLANGPDB_AARCH64_CC_XIPFLAGS          = -mstrict-align
+
+DEBUG_CLANGPDB_AARCH64_CC_FLAGS         = DEF(CLANGPDB_ALL_CC_FLAGS) -m64 -mno-red-zone -mcmodel=small -Oz -flto DEF(CLANGPDB_AARCH64_TARGET) -gcodeview  -funwind-tables -Wno-unused-but-set-variable -Wno-deprecated-non-prototype -Wno-constant-conversion
```
ardbiesheuvel: What does `-mno-red-zone` do on arm64?

mhaeuser:
> Note that we are talking about ELF to PE/COFF conversion here, so RELRO does not mean anything here.

Sorry if I was unclear, but exactly that was the point (as in, “not even RELRO changes the fact we start off with RW”).

> The best approximation of RELRO in our case is putting anything that needs to be relocated by the loader but is constant from the PoV of the program into .text.

If there is no separate .rodata, then yes for the general case, but in this specific case this violates the well-formedness of the image (because we assume ELF semantics). Now I also understand why you said “GOT must be empty” rather than “GOT must not be accessed before relocation”…

benjamindoron:
Rather than treating Standalone MM as a flat binary that OP-TEE can just jump into the top of, I recommend the solution that we're using for coreboot's "payload MM," where coreboot allows its payload to install/run SMI handlers too (https://github.com/tianocore/edk2/pull/11370): we created platform specific hook functions in the MM IPL so that we can relocate the image in a temporary buffer (but matching its final destination. The `DestinationAddress` field of PeCoffLib's ImageContext helped enable that). Then, we call an SMI handler in coreboot to copy that buffer to its final destination, and call the entrypoint.

This allows dropping the self-relocation code, which means that there's no need to switch permissions between RW and RX. This is probably out of scope of this PR, but I believe it's a better design that I'd be glad to see OP-TEE adopt in the future.

MdePkg/Include/AArch64/AArch64.h
os-d: `VECTOR_END`, the only user of this macro, takes `tbl` as a param. I can update `_PREVIOUS_SECTION` to take `tbl` as a parameter.

In general, this section came from `VECTOR_BASE`, but looking at this now, I think this should this just be `.section .text`. i.e. VECTOR_BASE() is creating a new section of .text.<tbl> for this vector and so when the ELF code does `.previous` this should take us back to the regular `.text` section, not continue in .text.<tbl>.

I'll update this to `.section .text` unless my understanding is incorrect and I'll check the other places where `.previous` had to be worked around.

BaseTools/Conf/build_rule.template
os-d: Fixed.

ArmPkg/Drivers/ArmPsciMpServicesDxe/MpFuncs.S
os-d: Went through all the commits and made sure the messages were updated to the latest changes.

BaseTools/Conf/tools_def.template
os-d: It is not, thanks, fixed.

BaseTools/Conf/tools_def.template
os-d: It is a no-op. I'm surprised clang doesn't reject it. I've dropped it from all AARCH64 CLANGPDB flags.

MdePkg/Include/AArch64/AArch64.h
```diff
@@ -150,9 +150,21 @@
...
+#else // __ELF__
+#define _PREVIOUS_SECTION()       \
+  .section .text;                 \
+  .align 3
+#endif // __ELF__
+
 #define VECTOR_END(tbl)           \
   .org 0x800;                     \
-  .previous
+  _PREVIOUS_SECTION()
```
ardbiesheuvel: I propose we just drop VECTOR_END() entirely, and replace it with `.section .text` in the callers, as it doesn't seem that useful, and in `ArmPkg/Library/ArmExceptionLib/AArch64/ExceptionSupport.S` it actually seems to be used incorrectly (the `ExceptionHandlersEnd` symbol is placed in a different section than expected due to the `.previous` in VECTOR_END(). In fact, `ExceptionHandlersEnd` itself seems entirely unused too)

os-d: @ardbiesheuvel that's fine with me, my only question is: Is the `.org ASM_PFX(tbl) + 0x800` required as well? That was added with the original macros: https://github.com/tianocore/edk2/commit/d855b261d19e2c21c059293befe6ab37db1e8efb.

My understanding is that statement will create zeros in the rest of that section that was unused and move forward to that address for the next instructions, while causing a build failure if the vector in question was larger than 0x800 (which would be very large indeed). I'm not sure how important this is?

I do agree the `.align 3` can also be dropped, since the `.previous` also was not affecting that.

os-d: Oh, sorry, I was looking at the original commit, you then removed the `ASM_PFX(tbl)` part for clang support: https://github.com/tianocore/edk2/commit/e7e120133d5f5fa7871d20d3995da87826f172e6. So, question being do we need to keep `.org 0x800` in the callers as well as `.section .text`? If so, the macro may still be useful to enforce that?

ardbiesheuvel: The `.org 0x800` is supposed to ensure that the entire vector table does not exceed 2048, which is the architected size. Each individual entry is already limited to 128 bytes, so this should only fire if we inadvertently try to cram more than 16 entries into the table.

I'd like to keep it if we can, but given that a) nothing will happen if you forget to put VECTORS_END(), and b) only one or two of those entries are ever invoked to begin with, I don't think it would be a tragedy to get rid of it.

os-d: @ardbiesheuvel I addressed this by combining the VECTOR_BASE and VECTOR_END macros into a VECTOR_TABLE macro that takes the interleaving VECTOR_ENTRYs as an argument. In this way we can enforce that the VECTOR_END behavior is used if you used VECTOR_BASE. This does look a bit uglier because of having to take the VECTOR_ENTRYs as an argument to the macro. I added a new commit (MdePkg, ArmPkg, ArmPlatformPkg: Create VECTOR_TABLE Macro) for this, please take a look and let me know what you think. I don't have a preference between doing this or just removing VECTOR_END and having the callers do the .org/.section.

I also added another new commit (ArmPkg: ArmExceptionLib: Drop ExceptionHandlersEnd) to resolve the issue of ExceptionHandlersEnd being in the wrong section.

Finally, I amended MdePkg: Add PE Target Support to AArch64 to simply replace the `.previous` with `.section .text` and remove the macro that was switching between them for ELF vs PE.

BaseTools/Conf/tools_def.template
```diff
@@ -1783,9 +1784,11 @@ RELEASE_GCC_LOONGARCH64_CC_FLAGS       = DEF(GCC5_LOONGARCH64_CC_FLAGS) -Wno-unu
 
 DEFINE CLANGPDB_IA32_PREFIX          = ENV(CLANG_BIN)
 DEFINE CLANGPDB_X64_PREFIX           = ENV(CLANG_BIN)
+DEFINE CLANGPDB_AARCH64_PREFIX       = ENV(CLANG_BIN)
 
 DEFINE CLANGPDB_IA32_TARGET          = -target i686-pc-windows-msvc
 DEFINE CLANGPDB_X64_TARGET           = -target x86_64-pc-windows-msvc
+DEFINE CLANGPDB_AARCH64_TARGET       = -target aarch64-unknown-windows-gnu
```
os-d: While testing a debugger with it, I noticed that the .pdata section was missing, so we didn't have stack trace info. That turned out to be that this target is incorrect (it was incorrect when it was defined on our side). This should be aarch64-unknown-windows-msvc, not gnu. The gnu version is the MinGW configuration and does not generate a .pdata section as we expect. So, I will fix that, but unfortunately changing the target also causes some other things to break, so I will fix those up as well.

os-d: I resolved this, not too many issues, noted in general comment on PR for reviewer awareness.

os-d:
As I commented, I discovered that we had the incorrect target defined, aarch64-unknown-windows-gnu (the MinGW target) instead of aarch64-unknown-windows-msvc. This can have several odd effects, one of which is not producing the .pdata section in the image. When changing the target, I had to make a couple of other changes I'll note here for reviewers:
- Add -Werror (not required, but realized the CLANGPDB all flags doesn't define it)
- CryptoPkg: Add CLANGPDB AARCH64 Support -> Had to add -U_MSV_VER to not have OpenSSL think this is a Windows host build
- Added new commit (ArmPkg: ArmStandaloneMmCoreEntryPoint: Add CLANGPDB Support) to do the same scoping of `-fpie` for the entry point that I did for StMM core. I'm surprised this wasn't broken with the old target but the StMM core one was...

ardbiesheuvel (approved):

os-d:
@mikebeaton now that we are at the hopefully final version of this, do you mind rerunning it in your test CI branch and make sure it didn’t break something? Local CI and all flavors of ArmVirtPkg worked.

os-d:
This PR has been up for awhile, but checking in with the other reviewers/maintainers (at least for pkgs Ard isn't a maintainer of, though of course he is also a general AARCH64 maintainer, so it all gets a bit fuzzy):

@lgao4 @gapalomi @YuweiChen1110 @ponchofigueroa do you want to review for BaseTools?

@jyao1 @liyi77 do you want to review from the CryptoPkg side?

@niruiyu @jiaxinwu @LiuZhiguang001 @td36 @arunsbaskaran @kraxel @lzeng14 @ecuevasf do you want to review for UefiCpuPkg?

@gdong1 @Sean-StarLabs @jameslu8 @benjamindoron @gguo11837463 @ChaselChiu @shuoliu0 do you want to review for UefiPayloadPkg?

If not we can move forward with merging this.

BaseTools/Conf/build_rule.template
```diff
@@ -342,6 +342,8 @@
         "$(DLINK)" -o ${dst} $(DLINK_FLAGS) $(DLINK_XIPFLAGS) -Wl,--start-group,@$(STATIC_LIBRARY_FILES_LIST),--end-group $(CC_FLAGS) $(CC_XIPFLAGS) $(DLINK2_FLAGS)
         "$(OBJCOPY)" $(OBJCOPY_FLAGS) ${dst}
 
+    <Command.CLANGPDB>
+        "$(DLINK)" /OUT:${dst} $(DLINK_FLAGS) $(DLINK_SPATH) @$(STATIC_LIBRARY_FILES_LIST) $(DLINK2_FLAGS) $(DLINK_XIPFLAGS)
```
mdkinney, line 346: The intent for DLINK2_FLAGS for MSFT and CLANGPDB is to add /WHOLEARCHIVE to detect multiply defined symbols.  This is why the MSFT one does the link twice.  Once with /WHOLEARCHIVE to detect multiple defined symbols, and the final link without to minimize size.  We with need CLANGPDB to follow MSFT settings and rules, or we need CLANGPDB to drop DLINK2_FLAGS and match the final link rule.

CryptoPkg/Library/MbedTlsLib/MbedTlsLib.inf
```diff
@@ -129,7 +129,7 @@
   #
   GCC:*_*_IA32_CC_FLAGS    = -U_WIN32 -U_WIN64 -U_MSC_VER -Wno-error=maybe-uninitialized -Wno-error=unused-but-set-variable
   GCC:*_*_X64_CC_FLAGS     = -U_WIN32 -U_WIN64 -U_MSC_VER -Wno-error=maybe-uninitialized -Wno-error=format -Wno-format -Wno-error=unused-but-set-variable -DNO_MSABI_VA_FUNCS
-  GCC:*_*_AARCH64_CC_FLAGS =  -Wno-error=maybe-uninitialized -Wno-format -Wno-error=unused-but-set-variable -Wno-error=format
+  GCC:*_*_AARCH64_CC_FLAGS =  -U_WIN32 -U_WIN64 -Wno-error=maybe-uninitialized -Wno-format -Wno-error=unused-but-set-variable -Wno-error=format
```
mdkinney: Should -U_MSC_VER be added as well to match IA32/X64 settings?

CryptoPkg/Library/OpensslLib/OpensslLibAccel.inf
```diff
@@ -2159,19 +2159,45 @@
...
+  $(OPENSSL_GEN_PATH)/AARCH64-GCC/crypto/md5/md5-aarch64.S | CLANGDWARF
+  $(OPENSSL_GEN_PATH)/AARCH64-GCC/crypto/modes/aes-gcm-armv8-unroll8_64.S | CLANGDWARF
+  $(OPENSSL_GEN_PATH)/AARCH64-GCC/crypto/modes/aes-gcm-armv8_64.S | CLANGDWARF
+  $(OPENSSL_GEN_PATH)/AARCH64-GCC/crypto/modes/ghashv8-armx.S | CLANGDWARF
+  $(OPENSSL_GEN_PATH)/AARCH64-GCC/crypto/sha/keccak1600-armv8.S | CLANGDWARF
+  $(OPENSSL_GEN_PATH)/AARCH64-GCC/crypto/sha/sha1-armv8.S | CLANGDWARF
+  $(OPENSSL_GEN_PATH)/AARCH64-GCC/crypto/sha/sha256-armv8.S | CLANGDWARF
+  $(OPENSSL_GEN_PATH)/AARCH64-GCC/crypto/sha/sha512-armv8.S | CLANGDWARF
+  $(OPENSSL_GEN_PATH)/AARCH64-GCC/crypto/sm3/sm3-armv8.S | CLANGDWARF
+  $(OPENSSL_GEN_PATH)/AARCH64-CLANGPDB/crypto/aes/aesv8-armx.S | CLANGPDB
```
mdkinney: What is the difference between the GCC and CLANGPDB versions of the .S files?  Is it the assumed calling convention of the assembly code implementation?  This is why IA32/X64 added a PCD to select different files based on assumed calling convention.

CryptoPkg/Library/OpensslLib/OpensslLibAccel.inf
```diff
@@ -2230,7 +2256,7 @@
   GCC:*_*_IA32_CC_FLAGS    = -U_WIN32 -UWIN32 -U_WIN64 -U_MSC_VER $(OPENSSL_FLAGS) $(OPENSSL_FLAGS_IA32) -Wno-error=maybe-uninitialized -Wno-error=unused-but-set-variable
   GCC:*_*_X64_CC_FLAGS     = -U_WIN32 -UWIN32 -U_WIN64 -U_MSC_VER $(OPENSSL_FLAGS) $(OPENSSL_FLAGS_X64) -Wno-error=maybe-uninitialized -Wno-error=format -Wno-format -Wno-error=unused-but-set-variable -DNO_MSABI_VA_FUNCS
   GCC:*_CLANGDWARF_*_CC_FLAGS = -std=gnu99 -Wno-error=uninitialized -Wno-error=incompatible-pointer-types -Wno-error=pointer-sign -Wno-error=implicit-function-declaration -Wno-error=ignored-pragma-optimize
-  GCC:*_CLANGPDB_*_CC_FLAGS = -std=c99 -Wno-error=uninitialized -Wno-error=incompatible-pointer-types -Wno-error=pointer-sign -Wno-error=implicit-function-declaration -Wno-error=ignored-pragma-optimize -Wno-error=unused-function
+  GCC:*_CLANGPDB_*_CC_FLAGS = -std=c99 -Wno-error=uninitialized -Wno-error=incompatible-pointer-types -Wno-error=pointer-sign -Wno-error=implicit-function-declaration -Wno-error=ignored-pragma-optimize -Wno-error=unused-function -U_WIN32 -U_WIN64 -U_MSC_VER
```
mdkinney: Should -UWIN32 also be added to match IA32/X64 settings?

EmbeddedPkg/Library/PrePiHobLib/PrePiHobLib.inf
```diff
@@ -35,11 +35,13 @@
   BaseLib
   BaseMemoryLib
   DebugLib
+  PeCoffLib
```
mdkinney, line 38: How are the changes in this non arch specific INF file related to CLANGPDB AArch64 Support?

os-d: I noted that in the commit message, but CLANGPDB AArch64 CI fails to build this module because of the missing library class. Other toolchains don't catch it (due to implicit linking).

CryptoPkg/Library/OpensslLib/OpensslLibAccel.inf
os-d: Unfortunately not. AArch64 ELF targets and PE targets do not support the same assembly directives, in this case `.type` and `.size` (there may be use of `.previous` too, I don't recall). So CLANGPDB cannot build these assembly files because of the unsupported directives.

I originally had amended the existing files to support PE targets as well as ELF targets with conditional logic. Ard pointed out this was a bad idea because these files are generated with each new release (I had thought we generated them once and then updated from there) and the proper way to do this is to generate the CLANGPDB supported ones.

BaseTools/Conf/build_rule.template
os-d: I copied what the existing CLANGPDB rule already was: https://github.com/tianocore/edk2/blob/1588d5b9f0112a448ff76e80a872e35c1628176e/BaseTools/Conf/build_rule.template#L312.

I am fine to change the existing behavior, currently CLANGPDB [doesn't use the DLINK2_FLAGS at all](https://github.com/tianocore/edk2/blob/1588d5b9f0112a448ff76e80a872e35c1628176e/BaseTools/Conf/tools_def.template#L1859), so I would opt to remove them and rely on MSVC to catch the duplicate symbols (as it will cover a large portion of the code base and I think doubling the linking time for all toolchains feels like overkill).

CryptoPkg/Library/OpensslLib/OpensslLibAccel.inf
mdkinney: I agree that modifying the generated assembly is not a good idea.

So these are directives in the assembly source files that are either GCC ld compatible directives or CLANG lld-link compatible directives?

os-d: It's actually in the assembler when it knows it is building a PE image vs an ELF image. If I attempt to build one of the ELF style assembly files with CLANGPDB, I get this:

```text
INFO - 	"clang" @D:\r\edk2\Build\CryptoPkg\All\DEBUG_CLANGPDB\AARCH64\CryptoPkg\Library\OpensslLib\OpensslLibAccel\OUTPUT\asm_resp.txt -o D:\r\edk2\Build\CryptoPkg\All\DEBUG_CLANGPDB\AARCH64\CryptoPkg\Library\OpensslLib\OpensslLibAccel\OUTPUT\OpensslGen\AARCH64-GCC\crypto\aes\aesv8-armx.obj  D:\r\edk2\Build\CryptoPkg\All\DEBUG_CLANGPDB\AARCH64\CryptoPkg\Library\OpensslLib\OpensslLibAccel\OUTPUT\OpensslGen\AARCH64-GCC\crypto\aes\aesv8-armx.iiii
INFO - D:\r\edk2\Build\CryptoPkg\All\DEBUG_CLANGPDB\AARCH64\CryptoPkg\Library\OpensslLib\OpensslLibAccel\OUTPUT\OpensslGen\AARCH64-GCC\crypto\aes\aesv8-armx.iiii:12:1: error: unknown directive
INFO - .previous
INFO - ^
INFO - D:\r\edk2\Build\CryptoPkg\All\DEBUG_CLANGPDB\AARCH64\CryptoPkg\Library\OpensslLib\OpensslLibAccel\OUTPUT\OpensslGen\AARCH64-GCC\crypto\aes\aesv8-armx.iiii:14:7: error: expected absolute expression
INFO - .type aes_v8_set_encrypt_key,%function
INFO -       ^
INFO - D:\r\edk2\Build\CryptoPkg\All\DEBUG_CLANGPDB\AARCH64\CryptoPkg\Library\OpensslLib\OpensslLibAccel\OUTPUT\OpensslGen\AARCH64-GCC\crypto\aes\aesv8-armx.iiii:186:1: error: unknown directive
INFO - .size aes_v8_set_encrypt_key,.-aes_v8_set_encrypt_key
INFO - ^
INFO - D:\r\edk2\Build\CryptoPkg\All\DEBUG_CLANGPDB\AARCH64\CryptoPkg\Library\OpensslLib\OpensslLibAccel\OUTPUT\OpensslGen\AARCH64-GCC\crypto\aes\aesv8-armx.iiii:189:7: error: expected absolute expression
INFO - .type aes_v8_set_decrypt_key,%function

<many more of these>

INFO - NMAKE : fatal error U1077: '"clang" @D:\r\edk2\Build\CryptoPkg\All\DEBUG_CLANGPDB\AARCH64\CryptoPkg\Library\OpensslLib\OpensslLibAccel\OUTPUT\asm_resp.txt -o D:\r\edk2\Build\CryptoPkg\All\DEBUG_CLANGPDB\AARCH64\CryptoPkg\Library\OpensslLib\OpensslLibAccel\OUTPUT\OpensslGen\AARCH64-GCC\crypto\aes\aesv8-armx.obj  D:\r\edk2\Build\CryptoPkg\All\DEBUG_CLANGPDB\AARCH64\CryptoPkg\Library\OpensslLib\OpensslLibAccel\OUTPUT\OpensslGen\AARCH64-GCC\crypto\aes\aesv8-armx.iiii' : return code '0x1'
```

CryptoPkg/Library/OpensslLib/OpensslLibAccel.inf
os-d: Updated all instances of OpensslLib that have these flags.

CryptoPkg/Library/MbedTlsLib/MbedTlsLib.inf
os-d: Updated both MbedTlsLibs.

CryptoPkg/Library/OpensslLib/OpensslLibAccel.inf
mdkinney: Then it must be based one of the options passe into clang (likely the -target triple) that causes the assembler to process the assembly sources differently.

Which, BTW, is exactly that the PCD for IA32/X64 handles.  The -target triple determines the compiler default calling  convention.

os-d: Ah, sorry, I am following along now. Yes, this would be caused by the target triple. I thought that PCD was getting used within the asm files, I see what you are saying now about being used to select the files to build. I can update this commit to introduce a new PCD that determines which set of files are compiled, then we don't need the duplicate CLANGDWARF and GCC ones.

os-d: @mdkinney I updated this commit to use a new PCD. I can use the existing PCD if you want, but it should be renamed, because NASM is not at play here. Let me know what you think.

os-d:
@mdkinney wanted to circle back with you on the open comments that I responded back, are those answers reasonable to you?

CryptoPkg/Library/OpensslLib/OpensslLibAccel.inf
mdkinney: The only other feedback I have is in the dir names for the optimized .S files.  Should they be ELF and PE instead of GCC and CLANGPDB?

os-d: I updated the directory names to AARCH64-ELF and AARCH64-PE. I left the x86 ones alone, though they could really have the same naming.

os-d:
@mdkinney I addressed the final comment by updating the directory names, please re-review when you have a chance.

With your, Ard's (who maintains half of these packages plus AARCH64 all up), and Mike Beaton's reviews, are we good to merge this? This has been open ~a month without review from other package maintainers and I pinged them a couple days ago to see if any of them wanted to see the hopefully final state, without response.

ardbiesheuvel:
> With your, Ard's (who maintains half of these packages plus AARCH64 all up), and Mike Beaton's reviews, are we good to merge this?

Yes.

PR #11903, opened 2025-12-18
Prevent TPL inversion in ReportStatusCodeLib

Javagedes:
# Description

A consumer of ReportStatusCodeLib will assert when calling `ReportStatusCode` at a TPL higher then `TPL_NOTIFY` when the protocol has not yet been cached. This is because when `ReportStatusCode` is called and the Report Status Code protocol has not yet been cached, it will call a `LocateProtocol` to attempt to locate it. Access to the protocol database is behind a TPL lock at `TPL_NOTIFY`, so a `RaiseTpl` on `TPL_NOTIFY` will be called, resulting in an assert for raising the TPL to a lower TPL.

This pull-request fixes this issue by moving the caching of the protocol from "on-demand" calls to `ReportStatusCode` to an event callback for the protocol's installation (Unless the protocol was found during library construction, then no event callback is registered).

- [ ] Breaking change?
- [ ] Impacts security?
- [x] Includes tests?

## How This Was Tested

A unit test was written that will assert with the TPL inversion error in `CoreRaiseTpl`. This test no longer fails after the above changes.

## Integration Instructions

N/A


MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
```diff
@@ -611,3 +612,105 @@ ReportDebugCodeEnabled (
...
+  EFI_STATUS  Status;
+  VOID        *Registration;
+
+  InternalGetReportStatusCode ();
+
+  //
+  // Register for protocol installation notification if the protocol is not found
+  //
+  if (mReportStatusCodeLibStatusCodeProtocol == NULL) {
+    ASSERT (gBS != NULL);
```
AshrafAliS: how about release BIOS. should we return

Javagedes: Yes. Thanks. Updating.

lgao4:
The code change is good to me. But, I suggest to create UnitTest directory to include test module, and add test module into MdeModulePkgHostTest.dsc.

Javagedes:
> The code change is good to me. But, I suggest to create UnitTest directory to include test module, and add test module into MdeModulePkgHostTest.dsc.

@lgao4, I've updated the test from being a UEFI_APPLICATION to being a host application unit test. I created a GoogleTest directory and placed the test there, while also adding the test to MdeModulePkgHostTest.dsc.

Thanks!

lgao4 (approved):

Javagedes:
@mdkinney @makubacki would either of you like to review this PR? If not, maybe in a few days, @lgao4 could you had the push tag?

Thanks!

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
```diff
@@ -611,3 +612,103 @@ ReportDebugCodeEnabled (
 {
   return (BOOLEAN)((PcdGet8 (PcdReportStatusCodePropertyMask) & REPORT_STATUS_CODE_PROPERTY_DEBUG_CODE_ENABLED) != 0);
 }
+
+/**
+  Notification function called when the Status Code Runtime protocol is installed. Locates and caches the protocol.
+
+    @param    Event           Not Used.
+    @param    Context         Not Used.
```
makubacki: nit:

```suggestion
    @param[in]    Event           Not Used.
    @param[in]    Context         Not Used.
```

And similar cases.

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
```diff
@@ -611,3 +612,103 @@ ReportDebugCodeEnabled (
...
+                    OnStatusCodeRuntimeProtocolInstalledNotification,
+                    NULL,
+                    &mOnStatusCodeProtocolInstallEvent
+                    );
+    ASSERT_EFI_ERROR (Status);
+
+    //
+    // Register for protocol notifications on this event
+    //
+    Status = gBS->RegisterProtocolNotify (
```
makubacki, line 679: Technically, if the event creation fails, there is no need to install a protocol notification on it.

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
Javagedes: Updated this and all other functions to include `[in]`.

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
Javagedes: Im going to leave this the same unless you have a really strong opinion. This flow is pulled from the RuntimeDxe implementation

makubacki: Not a strong opinion. Because this is in a library instance typically linked against many modules, it's more likely to hit this case.

I saw that but didn't want to leave two comments for the same thing.

Javagedes:
@lgao4 Any other requests? If not, can we put the push tag on this.

PR #11914, opened 2025-12-21
Fix CLANGPDB issues with -U _MSC_VER by removing -U _MSC_VER from tools_def.txt

mdkinney:
# Description

Remove -U _MSC_VER from CLANGPDB at global scope from tools_def.txt that breaks host application builds for EmulatorPkg and host-based unit testing with UnitTestFrameworkPkg. Instead, add -U _MSC_VER in the few places that require it in library builds that use submodules from other projects. Most of these changes are to add -U _MSC_VER to GCC family in specific components to match the -U _MSC_VER that is already present for MSFT family.

* UnitTestFrameworkPkg/Include/Library: Remove _MSC_VER workaround
* BaseTools/Conf: Remove -U _MSC_VER from CLANGPDB
* MdeModulePkg/BrotliCustomDecompressLib: Undefine _MSC_VER for GCC
* CryptoPkg/Library/MbedTlsLib: Undefined _MSC_VER for GCC Family
* CryptoPkg/Library/OpensslLib: Undefined _MSC_VER for GCC Family

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verified all current firmware buildings using clang compilers work with this change applied.  And also verified along with other changes queued up that host application issues with EmulatorPkg and UnitTestFrameworkPkg are resolved with these changes.

## Integration Instructions

N/A


mikebeaton:
If I'm not wrong, we seem to remove `-U _MSC_VER` from CLANGPDB globally, but then add it in where needed for _all_ GCC family toolchains (not just CLANGPDB). It seems from comments this is as intended - but not fully clear why (to me, at least, apologies!). E.g. Did the other GCC family toolchains fail to build, in these cases?

mdkinney:
> If I'm not wrong, we seem to remove `-U _MSC_VER` from CLANGPDB globally, but then add it in where needed for _all_ GCC family toolchains (not just CLANGPDB). It seems from comments this is as intended - but not fully clear why (to me, at least, apologies!). E.g. Did the other GCC family toolchains fail to build, in these cases?

Yes.  This is intended. -U _MSC_VER should have never been added to CLANGPDB in tools_def.txt. This fixes all the side effects of that incorrect change and adds -U _MSC_VER to only the modules that actually need it.

os-d (approved):

makubacki (approved):

PR #11925, opened 2025-12-23
EmulatorPkg fix issues reported by CLANG

mdkinney:
# Description

Set of changes to address errors reported by the CLANG compiler in Windows and Linux environments. Broken up into the following commits so each issue identified by CLANG can be reviewed independently.

* EmulatorPkg/Sec: Port SwitchRam to NASM
* Unix/Host CLANG issues
  * Fix symbol loading with clang
  * Fix gPpiList multiply defined
* Windows/Host CLANG Issues 
  * Fix WinNtBlockIoError() parameter type
  * Fix WinNtOpenVolume() uninitialized variable
  * Fix main() return type and error paths
  * Fix Receive() parameter type
  * Fix GetProcessAffinityMask() param types
  * Fix CLANG windows include issues
  * Remove workarounds for older VS20xx/SDKs

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

EmulatorPkg builds with CLANG for Windows and Linux envs fail with errors before these changes.  CLANG Builds pass after these changes and EmulatorPkg boots in Windows and Linux envs.

## Integration Instructions

N/A


mdkinney:
Updated this PR to remove `#pragma warning` statements that are no longer required for any supported tool chains. This simplifies `WinInclude.h` and `EmulatorPkg/Win/Host` in general.

Also found many other statements in `WinInclude.h` that were there for support of older VS20xx/SDKs can also be removed.  Further simplifying `WinInclude.h` for VS20xx and all CLANG compilers.

Investigated why the `#undef  IMAGE_FILE_MACHINE_ARM64` was required for Mingw CLANG builds. Mingw CLANG uses Mingw ucrt includes for `winnt.h` that defines `IMAGE_FILE_MACHINE_ARM64` with lower case `0xaa64`.  Visual Studio/SDKs use upper case `0xAA64`.  `MdePkg` uses upper case `0xAA64`.  If a define is done more than one time for the exact same string, then no error occurs.  This is why Visual Studio/SDKs do no report redefinition error in `MdePkg`. However, when Mingw ucrt includes are used with `MdePkg`, the case of the value is not the same, so the compiler reports a redefinition error.

The rest of the `IMAGE_FILE_MACHINE_*` defines use the same case for values across Visual Studio/SDK includes, Mingw ucrt includes, and edk2 includes, which is why this error has not been observed before.

The fix implemented is to see if `IMAGE_FILE_MACHINE_ARM64` is defined with the case that does not match `MdePkg`, and only then undefine it and define it to the case that matches `MdePkg`.

EmulatorPkg/Unix/Host/Host.c
```diff
@@ -1123,6 +1121,17 @@ SecGdbScriptBreak (
...
   )
 {
+  volatile UINTN  Index;
+
+  //
+  // Disable inline and have this function do some work to prevent
+  // optimizing this function away. It must always be in symbol table to
+  // set breakpoint for gdb script to run, load symbols, and set pending
+  // breakpoints.
+  //
```
os-d, line 1131: Instead of this will `__attribute__((used, noinline))` get the same effect? Or will the linker still discard the function? I don't have a big problem with this, we have just have seen issues even with volatile getting optimized away (that was with MSVC, but still, if there is a declarative way of doing this, that would be preferrable).

EmulatorPkg/Win/Host/WinInclude.h
```diff
@@ -63,9 +53,23 @@ typedef UINT32 size_t;
...
-#pragma warning(default : 4115)
-#pragma warning(default : 4201)
+// Redefining a macro to the exact same string is not flagged an error.
+//
+// If IMAGE_FILE_MACHINE_ARM64 is defined, check if it matches the unexpected
+// value of "0xaa64". If "0xaa64" is detected, then undefine and redefine to
+// "0xAA64" which is the same value used in Windows SDKs and MdePkg includes.
+//
+#if defined (IMAGE_FILE_MACHINE_ARM64)
+  #if IMAGE_FILE_MACHINE_ARM64 == 0xaa64
```
os-d, line 44: Not a big thing, but why not just always undef and define to the correct one?

os-d (approved):

EmulatorPkg/Unix/Host/Host.c
mdkinney: It can still be optimized away if there are no references.

os-d: Okay, that's where whole archive or the equivalent comes in. I think this is fine to leave as is. If we encounter issues with different toolchain versions, we can deal with it then.

EmulatorPkg/Win/Host/WinInclude.h
mdkinney: If this define ever has a different value, then we want to build to break and be reviewed.

PR #11927, opened 2025-12-24
MdeModulePkg PciBus: Add bus check in PciAllocateBusNumber

lgao4:
# Description
Add the checker to make sure bus number not overflow in PciAllocateBusNumber.

- [ ] Breaking change?
If platform has the wrong usage, this change will not allocate the wrong bus number.  
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Build pass, and boot to shell on the real platform. 

## Integration Instructions
N/A

### DIFF
--- a/MdeModulePkg/Bus/Pci/PciBusDxe/PciLib.c
+++ b/MdeModulePkg/Bus/Pci/PciBusDxe/PciLib.c
@@ -1044,6 +1044,11 @@ PciAllocateBusNumber (
     MaxNumberInRange = BusNumberRanges->AddrRangeMin + BusNumberRanges->AddrLen - 1;
     if ((StartBusNumber >= BusNumberRanges->AddrRangeMin) && (StartBusNumber <=  MaxNumberInRange)) {
       NextNumber = (UINT8)(StartBusNumber + NumberOfBuses);
+      if (NextNumber < StartBusNumber) {
+        DEBUG ((DEBUG_ERROR, "%a: Bus number overflow. StartBusNumber: %d, NumberOfBuses: %d\n",  __func__, StartBusNumber, NumberOfBuses));
+        return EFI_OUT_OF_RESOURCES;
+      }
+
       while (NextNumber > MaxNumberInRange) {
         ++BusNumberRanges;
         if (BusNumberRanges->Desc == ACPI_END_TAG_DESCRIPTOR) {


MdeModulePkg/Bus/Pci/PciBusDxe/PciLib.c
```diff
@@ -1043,6 +1043,12 @@ PciAllocateBusNumber (
   while (BusNumberRanges->Desc != ACPI_END_TAG_DESCRIPTOR) {
     MaxNumberInRange = BusNumberRanges->AddrRangeMin + BusNumberRanges->AddrLen - 1;
     if ((StartBusNumber >= BusNumberRanges->AddrRangeMin) && (StartBusNumber <=  MaxNumberInRange)) {
+
+      if (StartBusNumber + NumberOfBuses > 255) {
+        DEBUG ((EFI_D_ERROR, "Required Bus Number larger than 255. StartBusNumber:%d,NumberOfBuses:%d\n", StartBusNumber, NumberOfBuses));
+        return EFI_OUT_OF_RESOURCES;
+      }
+
```
niruiyu: The new logic is actually to catch the uint_8 overflow.
So, how about adding a check after "NextNumber = (UINT8)(StartBusNumber + NumberOfBuses);"?

if (NextNumber < StartBusNumber) {
  DEBUG ((DEBUG_ERROR, "%a: Bus number overflow. StartBusNumber: %d, NumberOfBuses: %d\n", __func__, .....)
}

lgao4: My idea is to return errro status when overflow happens, not just report error message.

niruiyu: I agree to return errror status. But I want to use the check "NextNumber < StartBusNumber" instead of "255".

lgao4: I get your point.

MdeModulePkg/Bus/Pci/PciBusDxe/PciLib.c
```diff
@@ -1044,6 +1044,11 @@ PciAllocateBusNumber (
     MaxNumberInRange = BusNumberRanges->AddrRangeMin + BusNumberRanges->AddrLen - 1;
     if ((StartBusNumber >= BusNumberRanges->AddrRangeMin) && (StartBusNumber <=  MaxNumberInRange)) {
       NextNumber = (UINT8)(StartBusNumber + NumberOfBuses);
+      if (NextNumber < StartBusNumber) {
+        DEBUG ((DEBUG_ERROR, "%a: Bus number overflow. StartBusNumber: %d, NumberOfBuses: %d\n",  __func__, StartBusNumber, NumberOfBuses));
```
jkolakow, line 1048: I see this is a bit different approach than in https://github.com/tianocore/edk2/pull/11891
Why do we print the error message here when we don't have similar in line 1055, or 1069? Do you consider this case more important ?

lgao4: OK. They resolve the same problem. I will close this one.

PR #11928, opened 2025-12-24
MdeModulePkg: Fix issue that French package is unexpectedly cleaned

EricGao2015:
# Description

**MdeModulePkg: Fix issue that French package is unexpectedly cleaned**
The DriverHealthManagerStrings.uni file defines both English and French languages, resulting in the generation of two Unicode string packages: the first for English and the second for French. When cleaning the dynamic strings from the HII package list, the pointer which should point to dynamic string package incorrectly points to the static French package, causing it to be unexpectedly cleaned.
This patch fixes the pointer of EFI_HII_PACKAGE_END type PackageHeader. This way, there is no need for concern regarding the number of language packages available.

**MdeModulePkg: Optimize the process while cleaning dynamic string**
In the current code, the HiiPackageList will be created and destroyed each time the form is closed, which is unneccessary. This patch makes a function that produces the origin HiiPackageList. The function will only be called when the driver is initialized and the HiiPackageList will be directly updated in DriverHealthManagerCleanDynamicString function. This approach can avoid the unneccessary creation and destruction of the HiiPackageList.

## How This Was Tested

Change language from English to French, enter Device Manager page, Driver Health Manager page, and then return to Device Manager page, the Driver Health Manager is shown as French.

## Integration Instructions

N/A


MdeModulePkg/Universal/DriverHealthManagerDxe/DriverHealthManagerDxe.c
```diff
@@ -889,41 +928,14 @@ DriverHealthManagerCleanDynamicString (
...
-  PackageHeader         = (EFI_HII_PACKAGE_HEADER *)((UINT8 *)PackageHeader + PackageHeader->Length);
-  PackageHeader->Type   = EFI_HII_PACKAGE_END;
-  PackageHeader->Length = sizeof (EFI_HII_PACKAGE_HEADER);
+  EFI_STATUS  Status;
 
   Status = mDriverHealthManagerDatabase->UpdatePackageList (
                                            mDriverHealthManagerDatabase,
                                            mDriverHealthManagerHiiHandle,
-                                           HiiPackageList
+                                           mBackupHiiPackageList
```
lgao4, line 936: So, this is always the initial version of HII string package. If yes, can mBackupHiiPackageList be just pointer to the STRING_ARRAY_NAME?

EricGao2015: Probably not. The PackageList parameter of function `UpdatePackageList` must be a pointer to an EFI_HII_PACKAGE_LIST_HEADER package which should contains `EFI_GUID PackageListGuid` and `UINT32 PackageLength`, but STRING_ARRAY_NAME do not have this struct data.
This is STRING_ARRAY_NAME data that generated by AutoGen in Build directory.
```
//
//Unicode String Pack Definition
//
unsigned char DriverHealthManagerDxeStrings[] = {

// STRGATHER_OUTPUT_HEADER
  0x38,  0x03,  0x00,  0x00,

// PACKAGE HEADER

  0x99,  0x01,  0x00,  0x04,  0x34,  0x00,  0x00,  0x00,  0x34,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  
  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  
  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x00,  0x01,  0x00,  0x65,  0x6E,  
  0x2D,  0x55,  0x53,  0x00,

// PACKAGE DATA

// 0x0001: $PRINTABLE_LANGUAGE_NAME:0x0001
  0x14,  0x45,  0x00,  0x6E,  0x00,  0x67,  0x00,  0x6C,  0x00,  0x69,  0x00,  0x73,  0x00,  0x68,  0x00,  0x00,  
  0x00,
// 0x0002: STR_FORM_TITLE:0x0002
  0x14,  0x44,  0x00,  0x72,  0x00,  0x69,  0x00,  0x76,  0x00,  0x65,  0x00,  0x72,  0x00,  0x20,  0x00,  0x48,  
  0x00,  0x65,  0x00,  0x61,  0x00,  0x6C,  0x00,  0x74,  0x00,  0x68,  0x00,  0x20,  0x00,  0x4D,  0x00,  0x61,  
  0x00,  0x6E,  0x00,  0x61,  0x00,  0x67,  0x00,  0x65,  0x00,  0x72,  0x00,  0x00,  0x00,
  ...
}
```

lgao4:
The pointer which should point to dynamic string package incorrectly points to the static French package, causing it to be unexpectedly cleaned.

I don't understand this problem. Which pointer is wrong here?

EricGao2015:
> The pointer which should point to dynamic string package incorrectly points to the static French package, causing it to be unexpectedly cleaned.
> 
> I don't understand this problem. Which pointer is wrong here?

This is code snippet in function `DriverHealthManagerCleanDynamicString`.
https://github.com/tianocore/edk2/blob/85e2fef2a8dfcdc64235e783abd827f33f2664b3/MdeModulePkg/Universal/DriverHealthManagerDxe/DriverHealthManagerDxe.c#L909-L921
The target is to skip English package and French package, but actually only the English package is skipped after line 912 and then the french package is marked as`EFI_HII_PACKAGE_END`. The right thing is to copy the code in line 912 to the next line. In this way `PackageHeader` will point to dynamic added package in final. But this method is not flexible enough while another new language is added, so I  make the change in this patch. 
Can my explanation address your concerns?

MdeModulePkg/Universal/DriverHealthManagerDxe/DriverHealthManagerDxe.c
```diff
@@ -889,41 +928,14 @@ DriverHealthManagerCleanDynamicString (
...
-  HiiPackageList->PackageLength = (UINT32)BufferSize;
-  CopyMem (&HiiPackageList->PackageListGuid, &gEfiCallerIdGuid, sizeof (EFI_GUID));
-
-  PackageHeader = (EFI_HII_PACKAGE_HEADER *)(HiiPackageList + 1);
-  CopyMem (PackageHeader, STRING_ARRAY_NAME + sizeof (UINT32), FixedStringSize);
-
-  PackageHeader         = (EFI_HII_PACKAGE_HEADER *)((UINT8 *)PackageHeader + PackageHeader->Length);
-  PackageHeader->Type   = EFI_HII_PACKAGE_END;
-  PackageHeader->Length = sizeof (EFI_HII_PACKAGE_HEADER);
+  EFI_STATUS  Status;
```
lgao4, line 932: Here, PackageHeader->Length should be FixedStringSize, right?

EricGao2015: Do you mean the line 914 in unchanged file? The package of type EFI_HII_PACKAGE_END is to mark the end of the package list, so the Length should be sizeof (EFI_HII_PACKAGE_HEADER). I think it can't be FixedStringSize.

lgao4: Line 912 should be updated like below
```
PackageHeader = (EFI_HII_PACKAGE_HEADER *)((UINT8 *)PackageHeader + PackageHeader->Length);
==>
PackageHeader = (EFI_HII_PACKAGE_HEADER *)((UINT8 *)PackageHeader + FixedStringSize);
```

EricGao2015: > Line 912 should be updated like below
> 
> ```
> PackageHeader = (EFI_HII_PACKAGE_HEADER *)((UINT8 *)PackageHeader + PackageHeader->Length);
> ==>
> PackageHeader = (EFI_HII_PACKAGE_HEADER *)((UINT8 *)PackageHeader + FixedStringSize);
> ```

Yes, this change also works and the change is minimal. But the HiiPackageList will be created and destroy each time the form is closing. If you prefer the minimal change, I can consider following your suggestion.

lgao4: I agree your fix is better. I suggest to do them both. The first commit is to correct this length, the second commit is to your change for optimization.

EricGao2015: > I agree your fix is better. I suggest to do them both. The first commit is to correct this length, the second commit is to your change for optimization.

I have resubmited two commits that you suggested. Please review them again.

EricGao2015: @lgao4  Hi Liming, please continue to review if you have time, thanks.

PR #11905, opened 2025-12-19
CLANGDWARF CI - TEST ONLY

mikebeaton:
# Description

CLANGDWARF CI - TEST ONLY

- [ ] Breaking change?
  - NO.
- [ ] Impacts security?
  - NO.
- [ ] Includes tests?
  - NO.

## How This Was Tested

CI
## Integration Instructions

N/A

PR #11929, opened 2025-12-24
Base tools conf updates for clang

mdkinney:
# Description

Update BaseTools/Conf files for CLANG compatibility.

* BaseTools/Conf: Update CLANG NASM flags for IA32/X64
* BaseTools/Conf: Align CLANGDWARF with GCC .efi rules
* BaseTools/Conf: Use common rule for Hii-Binary-Package
* BaseTools/Conf: Remove CC_FLAGS/CC_XIPFLAGS from DLINK actions

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verify no regressions for supported tool chains with these changes.  

Resolves compatibility issues using clang for EmulatorPkg and UnitTestFrameworkPkg for source level debugging, support of HII resource sections, and linking modules of type HOST_APPLICATION.

## Integration Instructions

N/A


BaseTools/Conf/build_rule.template
```diff
@@ -313,10 +313,10 @@
         "$(DLINK)" /OUT:${dst} $(DLINK_FLAGS) $(DLINK_SPATH) @$(STATIC_LIBRARY_FILES_LIST) $(DLINK2_FLAGS)
 
     <Command.CLANGDWARF>
-        "$(DLINK)" -o ${dst} $(DLINK_FLAGS) -Wl,--start-group,@$(STATIC_LIBRARY_FILES_LIST),--end-group $(CC_FLAGS) $(DLINK2_FLAGS)
```
lgao4, line 316: I remember to add CC_FLAGS to DLINK for GCC LTO requirement. After remove CC_FLAGS, have you verify GCC tool chain?

@ardbiesheuvel , can you confirm whether CC_FLAGS be removed from DLINK?

BaseTools/Conf/build_rule.template
```diff
@@ -692,18 +692,40 @@
...
+    # not specify an output file and the default output file behavior is used.
+    #
+    # In order to generate the final {dst} in all cases, a copy command that
+    # is allowed to fail without error is used to copy the .res file to the
+    # .rc file so when the rc or llvm-rc command is used the result is the same
+    # as objcopy. Finally, the .rc file produced in all cases is copied to {dst}.
+    #
+    <Command.MSFT, Command.INTEL, Command.GCC, Command.CLANGPDB, Command.CLANGDWARF>
+        "$(GENFW)" -o $(OUTPUT_DIR)(+)$(MODULE_NAME)hii.rc -g $(MODULE_GUID) $(GENFWHII_FLAGS) $(HII_BINARY_PACKAGES) $(GENFW_FLAGS)
+        "$(RC)" $(RC_FLAGS) $(OUTPUT_DIR)(+)$(MODULE_NAME)hii.rc
```
lgao4, line 726: I think --hiipackage can still be used here, because XCODE has the separate rule.

mdkinney: I agree.  I have removed the XCODE specific changes from this PR.

BaseTools/Conf/build_rule.template
mdkinney: I have verified that GCC tool chain still works as expected.

This PR also passes edk2 CI which includes GCC coverage.

If there are regressions in GCC toolchain, those should be fixed in this PR by adding required options to DLINK flags.  This is why these changes include adding TARGET options to DLINK_FLAGS for the CLANGDWARF toolchain for IA32/X64.

ardbiesheuvel: GCC LTO definitely needs the compiler flags in the linker invocation too, otherwise, things like `-mstrict-align` may not be effective, resulting in crashes in XIP code that runs before the MMU is enabled.

mdkinney: Can those options be added to the GCCLTO DLINK_FLAGS?

I agree that there are some options that need to be present in both CC_FLAGS and DLINK_FLAGS, but those seem to be rare.  Adding all CC_FLAGS unconditionally to all DLINK actions does not sounds like a good approach and it actually breaks many of the host application builds for EmulatorPkg and host-based unit tests.

ardbiesheuvel: > I agree that there are some options that need to be present in both CC_FLAGS and DLINK_FLAGS

The point is really that code generation is done at link time, not at compile time. So all options that affect codegen should be passed to the linker too. Only things like `-include` or `-I`, or options that control the preprocessor can be safely omitted.

> but those seem to be rare.

How so?

> Adding all CC_FLAGS unconditionally to all DLINK actions does not sounds like a good approach and it actually breaks many of the host application builds for EmulatorPkg and host-based unit tests.

Why are we using target C flags for host tools? That sounds like a hack to me - our build rules describe how to build for the target, and I don't think any of the codegen related options are suitable for the host, although many may be benign.

mdkinney: The -D and -U options can also be removed.  It is cleaner to not pass in extra options to the linker that are not used.  I see other rules that use DLINK_XIPFLAGS.  Why are CC_XIPFLAGS also passed to linker instead of defining DLINK2_XIPFLAGS.  I find it confusing to use CC_* in a DLINK action.  It is not obvious to developers that add [BuildOptions] sections to INF/DSC files that appending CC_FLAGS or CC_XIPFLAGS will also add options to DLINK actions.

I am not referring to Host Tool builds.  I am referring to builds of modules of type HOST_APPLICATION in EmulatorPkg and host-based unit tests for the UnitTestFrameworkPkg.

I ran into a number of issues getting the 4 versions of CLANG tools to work with all combinations.  Removing the CC_FLAGS and CC_XIPFLAGS was a way to simplify the builds with what appeared to be no side effects for the combinations I was testing.

Now that I have all those CLANG combinations working I have run some basic tests with CC_FLAGS and CC_XIPLAGS added back into these rules, and everything is still working.  I will perform more extensive tests today to verify full functionality and if that passes I will remove the commit that removes CC_FLAGS and CC_XIPLAGS from DLINK actions from this PR.

ardbiesheuvel: > I find it confusing to use CC_* in a DLINK action.

Sure, but it is equally confusing that it is in fact the linker that is doing the compilation when LTO is in effect. It needs those flags one way or the other, and forcing users to add all codegen flags to both CC and DLINK flags is even worse.

> I am not referring to Host Tool builds. I am referring to builds of modules of type HOST_APPLICATION in EmulatorPkg and host-based unit tests for the UnitTestFrameworkPkg.
> 

OK.

> I ran into a number of issues getting the 4 versions of CLANG tools to work with all combinations. Removing the CC_FLAGS and CC_XIPFLAGS was a way to simplify the builds with what appeared to be no side effects for the combinations I was testing.
> 
> Now that I have all those CLANG combinations working I have run some basic tests with CC_FLAGS and CC_XIPLAGS added back into these rules, and everything is still working. I will perform more extensive tests today to verify full functionality and if that passes I will remove the commit that removes CC_FLAGS and CC_XIPLAGS from DLINK actions from this PR.

Excellent, let's hope that it works.

mdkinney: I have reverted the changes that removed CC_FLAGS from DLINK actions.

mikebeaton:
I'm not sure whether it adds a lot that you don't already know, but I can confirm that all of the commits from:

https://github.com/tianocore/edk2/pull/11925/commits
https://github.com/tianocore/edk2/pull/11935/commits
https://github.com/tianocore/edk2/pull/11936/commits
https://github.com/tianocore/edk2/pull/11929/commits (this PR)

build fine when all rebased together on top of my CLANG CI:

https://github.com/tianocore/edk2/pull/11905/commits (CLANGDWARF demo)
https://github.com/tianocore/edk2/pull/11906/commits (CLANGPDB demo)

In the case of the CLANGPDB demo, I placed @mdkinney's commits from the above PRs on top of @os-d's [AARCH64 CLANGPDB support](https://github.com/tianocore/edk2/pull/11902/commits), and again everything builds.

---

Note: There seems to be a bug in the GitHub UI whereby you have to manually navigate to the last commit of a PR with many commits in order to see its CI results (you can't just scroll down to it, 'below the fold' as they say, in the pages above). I have made a GitHub bug report.

BaseTools/Conf/build_rule.template
lgao4: I mean GENFWHII_FLAGS is not required if --hiipackage is specified here.

mdkinney: GENFWHII_FLAGS is not always --hiipackage.  It can be --hiipackage or --hiibinpackage.

lgao4: Got it.

lgao4 (approved):

PR #11685, opened 2025-10-31
MdeModulePkg: Core Static analysis changes

apop5:
# Description
Fixing CodeQl Scanning results for MdeModulePkg/Core.

Commits are split based upon the type of error that was detected. Individual commit messages contain links
to the documentation on the error types.

This is split from #11314 to try to make reviewing the changes simpler. 

- [ ] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [x] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

## How This Was Tested
Local CI.

## Integration Instructions
No Integration necessary.

PR #11933, opened 2025-12-25
MdeModulePkg/Core: Make PPI services spec complaint

khaliid2040:
# Description

As per spec and function documentation NotifyPpi() and InstallPpi() should return EFI_OUT_OF_RESOURCES if memory allocation fail. However the current implementation of those functions do assertions instead of returning.

Fix that by returning EFI_OUT_OF_RESOURCES if memory allocation fail.

- [ ] Breaking change?
  - NO
- [ ] Impacts security?
  - NO
- [ ] Includes tests?
  - NO

## How This Was Tested

OVMF X64 image run in QEMU

## Integration Instructions
N/A


MdeModulePkg/Core/Pei/Ppi/Ppi.c
```diff
@@ -493,7 +493,10 @@ InternalPeiInstallPpi (
       TempPtr = AllocateZeroPool (
                   sizeof (PEI_PPI_LIST_POINTERS) * (PpiListPointer->MaxCount + PPI_GROWTH_STEP)
                   );
-      ASSERT (TempPtr != NULL);
+      if (TempPtr == NULL) {
```
mdkinney, line 496: The ASSERT() is still useful for debug builds to detect out of memory conditions that should never occur in PEI phase. If an out of memory condition does occur, then changes must be made to remove the possibility of an out of memory condition. Adding the return status looks correct.  ASSERT() can be moved within if condition.

khaliid2040: Thanks, however most call sites of those two functions use ASSERT_EFI_ERROR() which will catch EFI_OUT_OF_RESOURCES when it return. But moving ASSERT() inside if block will make the change not take effect. 

As i understood from spec it is up to PEIM of that PPI to decide whether assertion is ideal because some PPIs aren't critical to boot process.

mdkinney: If a platform configures a release build to remove ASSERTS() then the logic added by this PR will be used and the error return status can be processed by the caller.

PEI is a very constrained environment.  Do not want PPI Installation and Notification failures to be silent if any caller does not check error return status.  This includes downstream projects outside the edk2 sources.  Would prefer that there is always enough memory for required and optional PPIs and that these out of memory conditions are always found and resolved before a production release filters out the ASSERT()s.

khaliid2040: So that means the code didn't check for NULL before. Since assert won't work on release. There are more than one place which assert is only used without NULL check. @apop5 added some null tests but wasn't complete. It seems like we need to add null check and return on every call to AllocatePool().

I will work on null tests since that will also fix complaince of InstallPpi() and NotifyPpi() also.

khaliid2040:
cc: @makubacki, @lgao4, @apop5, @ardbiesheuvel, @leiflindholm 

Hi, you were working on #11685 which added null checks. So, do you think this extra null checks i found useful.

apop5:
> cc: @makubacki, @lgao4, @apop5, @ardbiesheuvel, @leiflindholm
> 
> Hi, you were working on #11685 which added null checks. So, do you think this extra null checks I found useful.

Yes, I think the additional checks are useful.

khaliid2040:
> > cc: @makubacki, @lgao4, @apop5, @ardbiesheuvel, @leiflindholm
> > Hi, you were working on #11685 which added null checks. So, do you think this extra null checks I found useful.
> 
> Yes, I think the additional checks are useful.

Thanks, but what i don't understand is when it comes to null-checks why the project is completely relying on asserts even though they don't work on some builds . You have open PRs for this i will send you if you miss some.

makubacki:
> Thanks, but what i don't understand is when it comes to null-checks why the project is completely relying on asserts even though they don't work on some builds . You have open PRs for this i will send you if you miss some.

It's a known deficiency in error handling. Assert behavior can be configured to vary, but it is not sufficient as a sole mechanism to handle many of the error paths it is placed in throughout the codebase.

khaliid2040:
> > Thanks, but what i don't understand is when it comes to null-checks why the project is completely relying on asserts even though they don't work on some builds . You have open PRs for this i will send you if you miss some.
> 
> It's a known deficiency in error handling. Assert behavior can be configured to vary, but it is not sufficient as a sole mechanism to handle many of the error paths it is placed in throughout the codebase.

@makubacki I do have time to improve overall error handling of the project if you are okay with it. And i will expand this PR to resolve other deficiencies too.

Still, what i am not sure is if we need to improve, or if there is obvious reason to let it like that.

makubacki:
> @makubacki I do have time to improve overall error handling of the project if you are okay with it. And i will expand this PR to resolve other deficiencies too.
> 
> Still, what i am not sure is if we need to improve, or if there is obvious reason to let it like that.

I would start with cases you can easily reason about. For example, if the assert checks a condition that will clearly cause unintended behavior after the assert, then the error handling should be improved.

khaliid2040:
Then, in that case i will rollout to previous changes of the PR which was applying changes only to NotifyPpi and InstallPpi.

Because those two functions the spec defined them explicitly.

khaliid2040:
@lgao4 @mdkinney Is it possible this PR to get merged if there isn't anything else i could change.

Thanks

PR #11936, opened 2025-12-26
BaseTools/WindowsVsToolChain: Add CLANG support

mdkinney:
# Description

Add logic to detect CLANG environments and set the required environment variables.

The CLANGPDB tool chain tag in Windows environments can be used with Visual Studio compiler or Mingw CLANG compiler. Add detection of Mingw CLANG compiler first and if Mingw not detected, then use the existing logic for using Visual Studio.

The CLANGDWARF tool chain tag in Windows environments can only be used with Mingw CLANG compiler. Add detection of Mingw CLANG compiler first and if Mingw not detected, then fail with an error.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verify no regressions for supported toolchain using stuart builds.

Verify that the correct environment variables are set for Windows Mingw CLANG environments.

## Integration Instructions

N/A


mdkinney:
NOTE: The changes are easier to review if whitespace changes are ignored.

BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py
```diff
@@ -256,74 +256,160 @@ def do_pre_build(self, thebuilder):
...
-                    if HostInfo.bit == "32":
-                        HostType = "x86"
-                    elif HostInfo.bit == "64":
-                        HostType = "x64"
+            # Look for $(CLANG_BIN)/mingw32-make.exe to determine if we are in a
+            # mingw32 environment
+            #
+            is_mingw32_env = False
+            clang_bin = shell_environment.GetEnvironment().get_shell_var("CLANG_BIN")
+            if clang_bin is not None:
```
Javagedes: This can be simplified:

``` python
clang_bin = shell_environment.GetEnvironment().get_shell_var("CLANG_BIN")
is_mingw32_env = (clang_bin and os.path.exists(os.path.join(clang_bin, "mingw32-make.exe")))
            
if is_mingw32_env:
    self.Logger.debug("CLANG_BIN is set to a mingw32 toolchain.")
    # The rest of the logic
```

BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py
```diff
@@ -256,74 +256,160 @@ def do_pre_build(self, thebuilder):
...
+                    self.Logger.error("CLANG_BIN is set to a mingw32 toolchain but lld-link.exe not found.")
+                    return -1
+
+                #
+                # Set environment variables for mingw32 CLANGPDB environment
+                #
+                basetools_mingw_build = shell_environment.GetEnvironment().get_shell_var("BASETOOLS_MINGW_BUILD")
+                clang_host_bin = shell_environment.GetEnvironment().get_shell_var("CLANG_HOST_BIN")
+
+                if basetools_mingw_build is not None:
```
Javagedes: I think you should simplified this a bit, and group the two different variables setting into two different sections

```python
#
# Set environment variables for mingw32 CLANGPDB environment
#
basetools_mingw_build = shell_environment.GetEnvironment().get_shell_var("BASETOOLS_MINGW_BUILD") or 'TRUE'
if basetools_mingw_build.upper() != 'TRUE':
    self.Logger.error("BASETOOLS_MINGW_BUILD must be set to TRUE for CLANGPDB toolchain.")
    return -1
shell_environment.GetEnvironment().set_shell_var("BASETOOLS_MINGW_BUILD", basetools_mingw_build)

clang_host_bin = shell_environment.GetEnvironment().get_shell_var("CLANG_HOST_BIN")
if clang_host_bin is not None:
    self.Logger.debug("CLANG_HOST_BIN is already set.")
else:
    # The environment is mingw32 make add "mingw32-" to the end of the path.
    # The rest of the command is derived from definitions in tools.def.
    clang_host_bin = 'mingw32-'
shell_environment.GetEnvironment().set_shell_var("CLANG_HOST_BIN", clang_host_bin)
```

BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py
```diff
@@ -256,74 +256,160 @@ def do_pre_build(self, thebuilder):
...
-            for (k, v) in vs_vars.items():
-                shell_env.set_shell_var(k, v)
+                # check to see if host is configured
+                # HostType for VS tools should be (defined in tools_def):
+                # x86   == 32bit Intel
+                # x64   == 64bit Intel
+                # arm64 == 64bit Arm
+                #
+                HostType = shell_environment.GetEnvironment().get_shell_var("CLANG_VS_HOST")
+                if HostType is not None:
```
Javagedes: nit:


```suggestion
                if HostType:
```

BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py
```diff
@@ -256,74 +256,160 @@ def do_pre_build(self, thebuilder):
...
+                else:
+                    #figure it out based on host info
+                    if HostInfo.arch == "x86":
+                        if HostInfo.bit == "32":
+                            HostType = "x86"
+                        elif HostInfo.bit == "64":
+                            HostType = "x64"
+                    else:
+                        # anything other than x86 or x64 is not supported
+                        raise NotImplementedError()
```
Javagedes: probably better to log an error and return -1 here, just so there is more context for the end user. Something like

```python
self.Logger.error(f"CLANGPDB not supported for detected host [{HostType.arch}-{HostType.bit}]")
return -1
```

BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py
```diff
@@ -256,74 +256,160 @@ def do_pre_build(self, thebuilder):
...
-            # set the path to the VS tools like nmake.exe
+            # set the path to the mingw32 make utility
             ##
-            if shell_environment.GetEnvironment().get_shell_var("CLANG_HOST_BIN") is not None:
-                self.Logger.debug("CLANG_HOST_BIN is already set.")
 
-            else:
-                install_path = self._get_vs_install_path(None, None)
-                vc_ver = self._get_vc_version(install_path, None)
+            basetools_mingw_build = shell_environment.GetEnvironment().get_shell_var("BASETOOLS_MINGW_BUILD")
```
Javagedes: This is a nitpick as this is a personal stylistic preference rather than something terribly important. Feel free to close as won't fix.

I would prefer the opposite to the current approach. e.g.

Your current approach is:

1. Get all three variables
2. validate all three variables
3. Set all three variables

I would prefer if the approach is changed to:

1. For `BASETOOLS_MINGW_BUILD` - Get, validate, set
2. For `CLANG_HOST_BIN` - Get, validate, set
3. For `CLANG_BIN` - Get, validate, set

This is a nitpick as it is my own personal stylistic

mdkinney: One reason I did the grouping is to not change the global environment variable setting until all consistency checks have been made.  Perhaps that does not matter if and error fails the entire job and the env vars set are never evaluated.

BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py
mdkinney: agreed.  Will fix

BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py
mdkinney: agreed.  Will fix.

BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py
mdkinney: Agreed.  Will Fix.

BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py
mdkinney: Agreed.  Will fix.

BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py
mdkinney: I will make the change you suggest to make the logic consistent

mdkinney:
@Javagedes  Branch updated with all feedback addressed.

Javagedes (approved):

Javagedes:
Thanks @mdkinney. It looks good to me. Seeing as this is two weeks old with no other reviews, I don't forsee another review coming in. I am happy to add the push tag now (or you can). Just let me know.

PR #11935, opened 2025-12-26
BaseTools/build: Add defines for Windows build environments

mdkinney:
# Description

The stuart tools automatically add -D WIN_HOST_BUILD to edk2 build command line if a Windows build environment is detected. This behavior is added to build.py so that builds of the EmulatorPkg using build.py are not required to add the option -D WIN_HOST_BUILD when building in a Windows environment. This aligns Linux and Windows builds of the EmulatorPkg removing the need to specify extra defines.

In order to build the EmulatorPkg for Windows Mingw environments, EmulatorPkg DSC/FDF files require a way to detect if Windows Mingw environment is present. The Windows Mingw environment can be detected if
CLANG_BIN is set and mingw32-make.exe is detected
in the CLANG_BIN directory.

If a Windows Mingw environment is detected, add
-D WIN_MINGW32_BUILD to the edk2 build command line.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verify that EmulatorPkg builds in Windows using edk2 build command do not require -D WIN_HOST_BUILD

## Integration Instructions

N/A

PR #11938, opened 2025-12-26
CryptoPkg/UnitTestFrameworkPkg: Fix optimzied openssl library unit test builds

mdkinney:
# Description

Fix build failures when using optimized Openssl library instances for IA32/X64 when used in host-based unit tests under Windows. For this configuration, the NASM generated Openssl files and not the .S generated Opensll files must be used because the Windows host-based unit tests always use a Windows calling convention.

Depends on #11935

The optimized NASM assembly files from Openssl project for IA32/X64 assume a Windows calling convention. Those NASM files must be used for both firmware builds where the default compiler calling convention is the Windows calling convention and host-based unit tests built using any of the compilers supported in Windows environment.

All the VS20xx tool chains in the `MSFT` tool family use a Windows calling convention. The `CLANGPDB` tool chain under both Windows and Linux environments also use a Windows calling convention. Set `PcdOpensslLibAssemblySourceStyleNasm` to `TRUE` for these conditions.

Host-based unit tests built in a Windows environment also use a Windows calling convention. If `WIN_HOST_BUILD` is set indicating a Windows build environment and `HOST_BASED_UNIT_TESTING_ENABLED` is also set indicating host-based unit tests, then `PcdOpensslLibAssemblySourceStyleNasm` must be set to `TRUE`.

Otherwise a Linux calling convention is assumed and `PcdOpensslLibAssemblySourceStyleNasm` uses its DEC default value of `FALSE` that uses the Openssl project optimized .S files that assume a Linux calling convention.

 Also update `UnitTestFrameworkPkgHost.dsc.inc` to set `BUILD_TARGETS` to `NOOPT` because host-based unit tests only support `NOOPT`. If the DSC file that includes `UnitTestFrameworkPkgHost.dsc.inc` uses a `BUILD_TARGETS` other than `NOOPT`, this will override the DSC setting.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

No regressions for current unit test builds.

Verified using optimized openssl library instances in host-based unit tests under Windows passes build.

## Integration Instructions

N/A


liyi77 (approved):
LGTM

PR #11940, opened 2025-12-29
Update Smbiosview to display SMBIOS Table 45 name correctly

vbpandya:
This is the third PR focusing on Morello upstream activity 

[PR-798](https://github.com/tianocore/edk2-platforms/pull/798)  and [PR-877](https://github.com/tianocore/edk2-platforms/pull/877) are focusing on FVP and SoC changes for EDK-2 platforms

This PR has 1 commits,

- Type45 entry to SmbiosView

edk2 repository in morello project is hosted at https://git.morello-project.org/morello/edk2/-/tree/morello/master which is not regularly synced with upstream master and hence have fallen behind. The idea of the effort is to port the extra patches from the above downstream repo to here.

more information can be found at [ARM Architecture Reference Manual Supplement Morello for A-profile Architecture](https://developer.arm.com/documentation/ddi0606/latest)

### DIFF
--- a/MdeModulePkg/Bus/Pci/PciBusDxe/PciBus.h
+++ b/MdeModulePkg/Bus/Pci/PciBusDxe/PciBus.h
@@ -50,7 +50,6 @@ typedef struct _PCI_BAR        PCI_BAR;
 
 #define     EFI_PCI_IOV_POLICY_ARI    0x0001
 #define     EFI_PCI_IOV_POLICY_SRIOV  0x0002
-#define     EFI_PCI_IOV_POLICY_MRIOV  0x0004
 
 typedef enum {
   PciBarTypeUnknown = 0,
@@ -268,7 +267,6 @@ struct _PCI_IO_DEVICE {
   UINT8                                        PciExpressCapabilityOffset;
   UINT32                                       AriCapabilityOffset;
   UINT32                                       SrIovCapabilityOffset;
-  UINT32                                       MrIovCapabilityOffset;
   PCI_BAR                                      VfPciBar[PCI_MAX_BAR];
   UINT32                                       SystemPageSize;
   UINT16                                       InitialVFs;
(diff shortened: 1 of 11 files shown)


mikebeaton:
@vbpandya - Could you clarify how this relates to the contents of edk2-platforms Platform/ARM/Morello?

ArmPkg/Library/ArmExceptionLib/AArch64/ExceptionSupport.S
```diff
@@ -349,10 +353,22 @@ ASM_PFX(CommonExceptionEntry):
   isb
 
   EL1_OR_EL2(x28)
-1:msr      elr_el1, x29  // Exception Link Register
+1:
+#if !defined(ENABLE_MORELLO_CAP)
```
samimujawar: Can we avoid this macro in the core AArch64 code, please?

vbpandya:
> @vbpandya - Could you clarify how this relates to the contents of edk2-platforms Platform/ARM/Morello?

This PR does not add anything that the platforms will use, the Morello HW is a bit different in terms of how ESR_ELx, VBAR_ELx and other registers are configured compared to normal ARM HW.

Since this is part of the effort to try and upstream the changes in edk2 in general, I have referenced the edk2-platforms PR.

ArmPkg/Library/ArmExceptionLib/AArch64/ExceptionSupport.S
vbpandya: I'll try out few things and see we can avoid this altogether.

vbpandya: So far I have tested the UEFI image in the downstream setup on Morello FVP, I have tried 2 scenarios 
1. Using downstream UEFI with ENABLE_MORELLO_CAP=0 only for UEFI and the system boots to capabilities enabled busybox
2. Using upstream UEFI with my other PR in edk2-platforms which are under review and with this PR removed removed from edk2 and the system boots to capabilities enabled busybox

With this I think we don't need any capabilities related switching or handling in UEFI and can rely on capabilities enabled kernel to do that

I'll get rid of the commit first commit in this PR.

samimujawar:
Can you modify the PR title to reflect that the patch is to update "smbiosview" to display the SMBIOS type 45 table name correctly, please?

PR #11942, opened 2025-12-29
DynamicTablesPkg: Add SsdtCpuTopology tests

jbrasen:
# Description

Add Host-based tests for the AcpiSsdtCpuTopologyLib module.
There are multiple test modules to allow these to be built on any architecture. The common test mocks the architecture specific functions and then there are 2 test modules per architecture one that mocks the common function and one that tests some functions that need integration.

There is also a fix for missing _PSD generated with GICC only for Arm that was found during test development. 

- [ ] Breaking change?
- [ ] Impacts security?
- [X] Includes tests?

## How This Was Tested

This is primarily tests

## Integration Instructions

N/A

### DIFF
--- a/DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/Arm/ArmSsdtCpuTopologyGenerator.c
+++ b/DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/Arm/ArmSsdtCpuTopologyGenerator.c
@@ -270,6 +270,16 @@ CreateTopologyFromIntC (
       }
     }
 
+    // If a PSD info is associated with the
+    // GicCinfo, create an _PSD method returning them.
+    if (GicCInfo[Index].PsdToken != CM_NULL_TOKEN) {
+      Status = CreateAmlPsdNode (Generator, CfgMgrProtocol, GicCInfo[Index].PsdToken, CpuNode);
+      if (EFI_ERROR (Status)) {
+        ASSERT_EFI_ERROR (Status);
+        break;
+      }
+    }
+
     if (GicCInfo[Index].EtToken != CM_NULL_TOKEN) {
       Status = CreateAmlEtNode (
                  Generator,
--- a/DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/AmlTestHelper.cpp
+++ b/DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/AmlTestHelper.cpp
@@ -1,0 +1,1315 @@
+/** @file
+  AML Test Helper implementation for validating generated SSDT CPU Topology tables.
+
+  Provides utilities to parse and validate AML output against ACPI 6.6
+  specification requirements.
+
+  Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+  @par Reference(s):
+    - ACPI 6.6 Specification - s8.4 Declaring Processors
+    - ACPI 6.6 Specification - s8.4.7.1 _CPC (Continuous Performance Control)
+    - ACPI 6.6 Specification - s8.4.4 _LPI (Low Power Idle States)
+    - ACPI 6.6 Specification - s8.4.5.5 _PSD (P-State Dependency)
+    - ACPI 6.6 Specification - s8.4.1.1 _CST (C-State)
+    - ACPI 6.6 Specification - s8.4.1.2 _CSD (C-State Dependency)
+    - ACPI 6.6 Specification - s8.4.5.1 _PCT (Performance Control)
+    - ACPI 6.6 Specification - s8.4.5.2 _PSS (Performance Supported States)
+    - ACPI 6.6 Specification - s8.4.5.3 _PPC (Performance Present Capabilities)
+**/
+
+#include "AmlTestHelper.h"
+
+// =============================================================================
+// Basic AML Tree Navigation
+// =============================================================================
+
+EFI_STATUS
+AmlTestHelper::ParseSsdtTable (
+  IN  EFI_ACPI_DESCRIPTION_HEADER  *Table,
+  OUT AML_ROOT_NODE_HANDLE         *RootNode
+  )
+{
+  if ((Table == NULL) || (RootNode == NULL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  return AmlParseDefinitionBlock (Table, RootNode);
+}
+
+EFI_STATUS
+AmlTestHelper::FindDeviceInScope (
+  IN  AML_NODE_HANDLE  ScopeNode,
+  IN  CONST CHAR8      *DeviceName,
+  OUT AML_NODE_HANDLE  *OutNode
+  )
+{
+  AML_NODE_HEADER  *Child;
+
+  if (!IS_AML_OBJECT_NODE (ScopeNode)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  // Iterate through children of the scope
+  Child = AmlGetNextVariableArgument ((AML_NODE_HEADER *)ScopeNode, NULL);
+  while (Child != NULL) {
+    if (IS_AML_OBJECT_NODE (Child)) {
+      // Check if this is a Device node - we can look at the fixed args
+      AML_OBJECT_NODE  *ObjNode = (AML_OBJECT_NODE *)Child;
+
+      // Fixed argument 0 of a Device node should be the name (4-char NameSeg)
+      AML_NODE_HEADER  *NameArg = AmlGetFixedArgument (ObjNode, EAmlParseIndexTerm0);
+      if (IS_AML_DATA_NODE (NameArg)) {
+        AML_DATA_NODE  *DataNode = (AML_DATA_NODE *)NameArg;
+        // WARNING: AML names are NOT null-terminated! They are exactly 4 bytes.
+        // Use CompareMem instead of AsciiStrnCmp to avoid buffer overrun.
+        if ((DataNode->Buffer != NULL) && (DataNode->Size == 4) &&
+            (CompareMem (DataNode->Buffer, DeviceName, 4) == 0))
+        {
+          *OutNode = (AML_NODE_HANDLE)Child;
+          return EFI_SUCCESS;
+        }
+      }
+    }
+
+    Child = AmlGetNextVariableArgument ((AML_NODE_HEADER *)ScopeNode, Child);
+  }
+
+  return EFI_NOT_FOUND;
+}
+
+EFI_STATUS
+AmlTestHelper::FindDeviceByPath (
+  IN  AML_ROOT_NODE_HANDLE  RootNode,
+  IN  CONST CHAR8           *AslPath,
+  OUT AML_NODE_HANDLE       *OutNode
+  )
+{
+  EFI_STATUS       Status;
+  AML_NODE_HEADER  *Child;
+  AML_NODE_HANDLE  ScopeNode = NULL;
+
+  if ((RootNode == NULL) || (AslPath == NULL) || (OutNode == NULL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  // For path like "\_SB.C000", we need to:
+  // 1. Find the _SB_ scope as a child of root
+  // 2. Find C000 device as a child of _SB_
+
+  // First, find _SB_ scope in root's children
+  Child = AmlGetNextVariableArgument ((AML_NODE_HEADER *)RootNode, NULL);
+
+  while (Child != NULL) {
+    if (IS_AML_OBJECT_NODE (Child)) {
+      AML_OBJECT_NODE  *ObjNode = (AML_OBJECT_NODE *)Child;
+      AML_NODE_HEADER  *NameArg = AmlGetFixedArgument (ObjNode, EAmlParseIndexTerm0);
+
+      if ((NameArg != NULL) && IS_AML_DATA_NODE (NameArg)) {
+        AML_DATA_NODE  *DataNode = (AML_DATA_NODE *)NameArg;
+        // Check if this contains "_SB" (the name might be \_SB or _SB_)
+        // Look for the sequence '_' 'S' 'B' in the buffer (safely)
+        if ((DataNode->Buffer != NULL) && (DataNode->Size >= 3)) {
+          for (UINT32 i = 0; i <= DataNode->Size - 3; i++) {
+            if ((DataNode->Buffer[i] == '_') &&
+                (DataNode->Buffer[i+1] == 'S') &&
+                (DataNode->Buffer[i+2] == 'B'))
+            {
+              ScopeNode = (AML_NODE_HANDLE)Child;
+              break;
+            }
+          }
+
+          if (ScopeNode != NULL) {
+            break;
+          }
+        }
+      }
+    }
+
+    Child = AmlGetNextVariableArgument ((AML_NODE_HEADER *)RootNode, Child);
+  }
+
+  if (ScopeNode == NULL) {
+    return EFI_NOT_FOUND;
+  }
+
+  // Now find the device in the scope (extract device name from path)
+  // Path is like "\_SB.C000", device name is "C000"
+  CONST CHAR8  *DeviceName = AsciiStrStr (AslPath, ".");
+
+  if (DeviceName == NULL) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  DeviceName++;  // Skip the dot
+
+  Status = FindDeviceInScope (ScopeNode, DeviceName, OutNode);
+  return Status;
+}
+
+BOOLEAN
+AmlTestHelper::IsDeviceNode (
+  IN  AML_NODE_HANDLE  Node
+  )
+{
+  if (!IS_AML_OBJECT_NODE (Node)) {
+    return FALSE;
+  }
+
+  return AmlNodeCompareOpCode (
+           (AML_OBJECT_NODE *)Node,
+           AML_EXT_OP,
+           AML_EXT_DEVICE_OP
+           );
+}
+
+BOOLEAN
+AmlTestHelper::IsNameNode (
+  IN  AML_NODE_HANDLE  Node
+  )
+{
+  if (!IS_AML_OBJECT_NODE (Node)) {
+    return FALSE;
+  }
+
+  return AmlNodeCompareOpCode (
+           (AML_OBJECT_NODE *)Node,
+           AML_NAME_OP,
+           0
+           );
+}
+
+EFI_STATUS
+AmlTestHelper::FindNamedObjectInDevice (
+  IN  AML_NODE_HANDLE  DeviceNode,
+  IN  CONST CHAR8      *Name,
+  OUT AML_NODE_HANDLE  *OutNode
+  )
+{
+  AML_NODE_HEADER  *Child;
+
+  if ((DeviceNode == NULL) || (Name == NULL) || (OutNode == NULL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  if (!IS_AML_OBJECT_NODE (DeviceNode)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  // Iterate through variable arguments (children) of the device
+  Child = AmlGetNextVariableArgument ((AML_NODE_HEADER *)DeviceNode, NULL);
+  while (Child != NULL) {
+    if (IsNameNode ((AML_NODE_HANDLE)Child)) {
+      // Get the name from fixed argument 0
+      AML_NODE_HEADER  *NameArg = AmlGetFixedArgument ((AML_OBJECT_NODE *)Child, EAmlParseIndexTerm0);
+      if (IS_AML_DATA_NODE (NameArg)) {
+        AML_DATA_NODE  *DataNode = (AML_DATA_NODE *)NameArg;
+        // AML names are exactly 4 bytes, NOT null-terminated
+        // Use CompareMem to safely compare without buffer overrun
+        if ((DataNode->Buffer != NULL) && (DataNode->Size == 4) &&
+            (CompareMem (DataNode->Buffer, Name, 4) == 0))
+        {
+          *OutNode = (AML_NODE_HANDLE)Child;
+          return EFI_SUCCESS;
+        }
+      }
+    }
+
+    Child = AmlGetNextVariableArgument ((AML_NODE_HEADER *)DeviceNode, Child);
+  }
+
+  return EFI_NOT_FOUND;
+}
+
+EFI_STATUS
+AmlTestHelper::GetDeviceHid (
+  IN  AML_NODE_HANDLE  DeviceNode,
+  OUT CHAR8            *HidString,
+  IN  UINTN            BufferSize
+  )
+{
+  EFI_STATUS       Status;
+  AML_NODE_HANDLE  HidNode;
+  AML_NODE_HEADER  *ValueArg;
+
+  if ((DeviceNode == NULL) || (HidString == NULL) || (BufferSize < 9)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  Status = FindNamedObjectInDevice (DeviceNode, "_HID", &HidNode);
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  // Get the value from fixed argument 1
+  ValueArg = AmlGetFixedArgument ((AML_OBJECT_NODE *)HidNode, EAmlParseIndexTerm1);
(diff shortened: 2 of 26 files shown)


jbrasen:
@pierregondois @samimujawar Any thoughts on this? Was going to add CPC v4 support but wanted a testing baseline first

pierregondois:
Hello Jeff,
I will try to review the PR next week.
Sorry for the delay, the size of the PR is quite huge.

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/MockCommonFunctions.cpp
```diff
@@ -0,0 +1,124 @@
...
+MockCommonFunctions  *gMockCommonFunctions = nullptr;
+
+//
+// C function implementations that delegate to the mock
+//
+
+extern "C" {
+  EFI_STATUS
+  EFIAPI
+  WriteAslName (
```
pierregondois, line 27: Do we need to mock this functions ?
Isn't it possible to just use the common implementation as there is not much to test ?

Worst case it might be simpler to move the function to:
`DynamicTablesPkg/Include/Library/AcpiHelperLib.h`
as the function could be generalized

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/X64SsdtCpuTopologyGoogleTest.cpp
```diff
@@ -0,0 +1,1014 @@
...
+  return oss.str ();
+}
+
+INSTANTIATE_TEST_SUITE_P (
+  CpuCounts,
+  ApicCpuCountTest,
+  ::testing::Values (
+               1u,
+               2u,
+               4u,
```
pierregondois, line 592: Ideally it would be nice to also test odd numbers

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/X64SsdtCpuTopologyIntegrationGoogleTest.cpp
```diff
@@ -0,0 +1,1218 @@
...
+  ::testing::Values (
+               0u,     // Zero
+               63u,    // 6-bit max
+               64u,    // 7-bit
+               127u,   // ByteData boundary
+               255u,   // 8-bit max
+               256u,   // WordData boundary
+               32767u, // 16-bit signed max
+               65535u, // 16-bit max
+               65536u  // DWordData boundary
```
pierregondois, line 1202: Shouldn't we also test `0xFFFFFFFF` ?

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/X64SsdtCpuTopologyIntegrationGoogleTest.cpp
```diff
@@ -0,0 +1,1218 @@
...
+/**
+  @test X64Integration_MultipleCpus_AllHaveDevices
+
+  @brief Integration test validating multiple CPU generation.
+
+  @spec ACPI 6.6 s8.4: Each processor must have its own Processor Device object.
+
+  @expected One device generated per APIC entry.
+**/
+TEST_F (X64IntegrationTest, MultipleCpus_AllHaveDevices) {
```
pierregondois, line 523: IIUC, all arch have both:
- `Unit tests`, i.e. `X64SsdtCpuTopologyGoogleTest` for X64
- `Integration tests`, i.e. `X64SsdtCpuTopologyIntegrationGoogleTest` for X64

The `unit tests` are testing the arch-specific functions only, i.e.:
- GetIntCInfo()
- CreateTopologyFromIntC()
- AddArchAmlCpuInfo()
What is tested in these files is mainly:
- the return code of the above 3 functions
- the number of calls to sub-functions, like `CreateAmlPsdNode()`, `CreateAmlCpu()`, etc.

The `integration tests` try to build an SSDT table and then parse it to check the number of CPUs, their name, their attribute (e.g. presence of a _CPC, _PSD, ... object).

------------
------------

A. The `Unit tests` have a low granularity.
- It seems it will be quite some work to modify them if the Generator is refactored one day.
- What is tested is likely also tested by the `integration tests`. If one of the 3 arch-functions fails, this will show up in the `integration tests`
- I don't think that testing the number of calls to sub-functions is a sign of good health
IMO it would be better to remove them ...

B. In the  `Integration tests`, most of the tests are common to all archs. Unfortunately the initialization of all these tests is arch specific. E.g. common tests:
- DisabledCpu_StillGenerated
- MultipleCpus_AllHaveDevices
- CpcToken_ValidatesPackageStructure
- ...

Would it be possible to create a file common to all arch containing all the tests ? For instance:

`X64SsdtCpuTopologyIntegrationGoogleTest.cpp`:
```
TEST_F (X64IntegrationTest, DisabledCpu_StillGenerated) {
  const CM_OBJECT_TOKEN  StaToken = 8301;

  mApicInfo.resize (1);
  CreateApicInfo (&mApicInfo[0], 0, 100);
  mApicInfo[0].Flags    = 0;  // Disabled
  mApicInfo[0].StaToken = StaToken;
  SetupApicInfo (mApicInfo.data (), 1);

  // Set up STA info - device not present (0)
  mStaInfo.resize (1);
  CreateSimpleStaInfo (&mStaInfo[0], 0);
  SetupStaInfo (&mStaInfo[0], StaToken);

  EFI_ACPI_DESCRIPTION_HEADER  **Table    = nullptr;
  UINTN                        TableCount = 0;

  ASSERT_EQ (BuildSsdtCpuTopologyTable (&Table, &TableCount), EFI_SUCCESS);
  ASSERT_NE (Table, nullptr);

  AML_ROOT_NODE_HANDLE  RootNode;

  ASSERT_EQ (AmlTestHelper::ParseSsdtTable (Table[0], &RootNode), EFI_SUCCESS);

  // Verify CPU devices exist even when disabled
  UINT32  CpuCount = 0;

  EXPECT_EQ (AmlTestHelper::CountCpuDevices (RootNode, "\\_SB", &CpuCount), EFI_SUCCESS);
  EXPECT_GE (CpuCount, 1u) << "Disabled CPUs should still generate devices";

  AmlDeleteTree (RootNode);
  FreeSsdtCpuTopologyTable (Table, TableCount);
}
```

Would be changed to:
`X64SsdtCpuTopologyIntegrationGoogleTest.cpp`
```
TEST_F (X64IntegrationTest, DisabledCpu_StillGenerated) {
  const CM_OBJECT_TOKEN  StaToken = 8301;

  mApicInfo.resize (1);
  CreateApicInfo (&mApicInfo[0], 0, 100);
  mApicInfo[0].Flags    = 0;  // Disabled
  mApicInfo[0].StaToken = StaToken;
  SetupApicInfo (mApicInfo.data (), 1);

  // Set up STA info - device not present (0)
  mStaInfo.resize (1);
  CreateSimpleStaInfo (&mStaInfo[0], 0);
  SetupStaInfo (&mStaInfo[0], StaToken);

   // Run the common test
  CommonXXX_DisabledCpu_StillGenerated()

   // If necessary add some arch specific tests:
   ASSERT_EQ (XXX)
}
```
and:
`ArchCommonXXXSsdtCpuTopologyIntegrationGoogleTest.cpp`
```
CommonXXX_DisabledCpu_StillGenerated(XXX) {
  EFI_ACPI_DESCRIPTION_HEADER  **Table    = nullptr;
  UINTN                        TableCount = 0;

  ASSERT_EQ (BuildSsdtCpuTopologyTable (&Table, &TableCount), EFI_SUCCESS);
  ASSERT_NE (Table, nullptr);

  AML_ROOT_NODE_HANDLE  RootNode;

  ASSERT_EQ (AmlTestHelper::ParseSsdtTable (Table[0], &RootNode), EFI_SUCCESS);

  // Verify CPU devices exist even when disabled
  UINT32  CpuCount = 0;

  EXPECT_EQ (AmlTestHelper::CountCpuDevices (RootNode, "\\_SB", &CpuCount), EFI_SUCCESS);
  EXPECT_GE (CpuCount, 1u) << "Disabled CPUs should still generate devices";

  AmlDeleteTree (RootNode);
  FreeSsdtCpuTopologyTable (Table, TableCount);
}
```

This way all archs can use CommonXXX_DisabledCpu_StillGenerated()

C. There are some `integration tests` that are:
- Generating the same platform objects (or slighly different objects)
- Testing slightly different things
E.g., `SingleCpu_GeneratesCorrectAml()`, `MultipleCpus_AllHaveDevices()` generate N CPUs with a generic UID, and only the number of CPUs is tested. Their UID, the presence/absence of a `_STA`, `_PSD`, etc. object could also easily be tested.
`MultipleCpus_UidsMatchApicOrder()` then only check the UID value.

To increase the number things tested and also have a generic way to generate platforms, would it be possible to have, in each arch-specific `integration tests` file, a function setting up the arch objects ?
Something like:
```
GenerateXXX(
  Cpus, # Number of APICs to generate in mApicInfo
  ExpectedUids, # Uid to use for the CPUs
  GeneratePsd, # Boolean, true to generate Psd info
  GenerateCpc, # Boolean, true to generate Cpc info
  &mApicInfo, # Reference to mApicInfo to populate
  &mPsdInfo, # Reference to mPsdInfo to populate
  &mCpcInfo # Reference to mCpcInfo to populate
)
```
And a testing equivalent function:
```
TestXXX(
  Cpus, # Check that there are #Cpus cpus generated
  ExpectedUids, # Check their expected Uid
  GeneratePsd, # Boolean, if true, check the presence of Psd, otherwise check the absence
  GenerateCpc # Boolean, if true, check the presence of Cpc, otherwise check the absence
)
```
This would make the tests more platform/configuration oriented and less about testing a specific aspect. This is a design choice to use go this way or let it as is

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/X64SsdtCpuTopologyIntegrationGoogleTest.cpp
```diff
@@ -0,0 +1,1218 @@
...
+  X64CpuCountIntegrationTest,
+  ::testing::Values (
+               1u,
+               2u,
+               4u,
+               8u,
+               16u,
+               32u
+               ),
+  X64CpuCountIntegrationTestName
```
pierregondois, line 1135: Following the comment above about having a common file containing common integration tests, these test vectors seems generic to all archs

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/X64SsdtCpuTopologyGoogleTest.inf
```diff
@@ -0,0 +1,54 @@
...
+  FILE_GUID                      = 9b4d263f-cdf4-4dc4-b7c1-ac2ab09033d1
+  MODULE_TYPE                    = HOST_APPLICATION
+  VERSION_STRING                 = 1.0
+  ENTRY_POINT                    = GoogleTestCppMain
+
+[Sources]
+  X64SsdtCpuTopologyGoogleTest.cpp
+  AmlTestHelper.h
+  AmlTestHelper.cpp
+  SsdtCpuTopologyTestCommon.h
```
pierregondois, line 23: Not necessary I think

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/X64SsdtCpuTopologyGoogleTest.cpp
```diff
@@ -0,0 +1,1014 @@
...
+  CreateApicInfo (
+    OUT CM_X64_LOCAL_APIC_X2APIC_INFO  *ApicInfo,
+    IN  UINT32                         ApicId,
+    IN  UINT32                         AcpiProcessorUid
+    )
+  {
+    ZeroMem (ApicInfo, sizeof (*ApicInfo));
+    ApicInfo->ApicId           = ApicId;
+    ApicInfo->AcpiProcessorUid = AcpiProcessorUid;
+    ApicInfo->Flags            = 1;  // Enabled
```
pierregondois, line 125: Should be `EFI_ACPI_6_5_LOCAL_APIC_ENABLED`. The comment might apply to other places

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/ArmSsdtCpuTopologyGoogleTest.cpp
```diff
@@ -0,0 +1,1631 @@
...
+    IN  UINT32            AcpiProcessorUid,
+    IN  CM_OBJECT_TOKEN   CpcToken = CM_NULL_TOKEN,
+    IN  CM_OBJECT_TOKEN   PsdToken = CM_NULL_TOKEN,
+    IN  CM_OBJECT_TOKEN   EtToken  = CM_NULL_TOKEN
+    )
+  {
+    ZeroMem (GiccInfo, sizeof (*GiccInfo));
+    GiccInfo->AcpiProcessorUid   = AcpiProcessorUid;
+    GiccInfo->CPUInterfaceNumber = AcpiProcessorUid;
+    GiccInfo->Flags              = 1;  // Enabled
```
pierregondois, line 144: `(EFI_ACPI_6_5_GIC_ENABLED|EFI_ACPI_6_5_GIC_ONLINE_CAPABLE)` (the comment might apply to other places).

The online capable bit seems to have been added in ACPI 6.5, so maybe the online capable bit should be removed ?

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/X64SsdtCpuTopologyIntegrationGoogleTest.cpp
```diff
@@ -0,0 +1,1218 @@
...
+}
+
+/**
+  @test X64Integration_LargeCpuCount
+
+  @brief Tests handling of large CPU count.
+
+  @expected All CPUs generated correctly.
+**/
+TEST_F (X64IntegrationTest, LargeCpuCount) {
```
pierregondois, line 1044: Is it equivalent to remove this test and add `64` in the `X64CpuCountIntegrationTest` vector instead ?

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/X64SsdtCpuTopologyIntegrationGoogleTest.cpp
```diff
@@ -0,0 +1,1218 @@
...
+
+  @brief Validates all CPU UIDs match APIC order.
+
+  @expected Each CPU's _UID matches corresponding APIC AcpiProcessorUid.
+**/
+TEST_F (X64IntegrationTest, MultipleCpus_UidsMatchApicOrder) {
+  const UINT32  NumCpus = 4;
+
+  mApicInfo.resize (NumCpus);
+  UINT32  ExpectedUids[] = { 10, 20, 30, 40 };
```
pierregondois, line 687: For Uids or for other statically-written values (e.g. `CstInfo->Latency = 100`),
would it make sense to randomly generate values instead ? Or would it be too complicated ?

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/HostEnvStubs.c
```diff
@@ -0,0 +1,132 @@
...
+  )
+{
+  // For unit tests, assume no PPTT or other tables are present
+  // This simplifies test setup and validation
+  return FALSE;
+}
+
+/** Reset the metadata UID counter.
+
+  This function should be called between tests to ensure consistent
```
pierregondois, line 122: It doesn't seem to be called.

The handling of Metadata is mostly useful for cross-table generation, so it should be ok not to reset the Count awell. But it might be good to either use the function or remove it

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/CommonSsdtCpuTopologyGoogleTest.cpp
```diff
@@ -0,0 +1,2144 @@
...
+**/
+TEST_P (PsdDomainTest, FieldsEncodedCorrectly) {
+  const PsdDomainConfig  &config = GetParam ();
+
+  CM_ARCH_COMMON_PSD_INFO  PsdInfo;
+
+  ZeroMem (&PsdInfo, sizeof (PsdInfo));
+  PsdInfo.Revision  = 0;
+  PsdInfo.Domain    = config.Domain;
+  PsdInfo.CoordType = 0xFE;  // HW_ALL
```
pierregondois, line 2051: Would it be possible to use the existing macros whenever possible ?

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/AmlTestHelper.cpp
```diff
@@ -0,0 +1,1315 @@
...
+**/
+
+#include "AmlTestHelper.h"
+
+// =============================================================================
+// Basic AML Tree Navigation
+// =============================================================================
+
+EFI_STATUS
+AmlTestHelper::ParseSsdtTable (
```
pierregondois, line 29: It is necessary to have a wrapper ? Isn't it possible to just call `AmlParseDefinitionBlock()` directly ?

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/AmlTestHelper.cpp
```diff
@@ -0,0 +1,1315 @@
...
+{
+  if ((Table == NULL) || (RootNode == NULL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  return AmlParseDefinitionBlock (Table, RootNode);
+}
+
+EFI_STATUS
+AmlTestHelper::FindDeviceInScope (
```
pierregondois, line 42: Would it be possible to use `AmlFindNode()` instead if we know the absolute path of the Device Node ?
For instance this should allow to find the `C000` CPU:
```
Status = AmlFindNode(RootNode, "\\_SB_.C000", &OutNode);
```

If there is a need to check that the object found at this place in the AML namespace is of a specific type, e.g. a DeviceNode, a wrapper function could be created, but it looks like many functions are doing similar things regarding locating objects in the namespace

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/AmlTestHelper.cpp
```diff
@@ -0,0 +1,1315 @@
...
+    }
+
+    Child = AmlGetNextVariableArgument ((AML_NODE_HEADER *)ScopeNode, Child);
+  }
+
+  return EFI_NOT_FOUND;
+}
+
+EFI_STATUS
+AmlTestHelper::FindDeviceByPath (
```
pierregondois, line 83: Similar comment, the function looks similar to `AmlFindNode()`

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/AmlTestHelper.cpp
```diff
@@ -0,0 +1,1315 @@
...
+
+  return AmlNodeCompareOpCode (
+           (AML_OBJECT_NODE *)Node,
+           AML_NAME_OP,
+           0
+           );
+}
+
+EFI_STATUS
+AmlTestHelper::FindNamedObjectInDevice (
```
pierregondois, line 185: Similar comment, the function looks similar to `AmlFindNode()`

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/AmlTestHelper.cpp
```diff
@@ -0,0 +1,1315 @@
...
+    return EFI_INVALID_PARAMETER;
+  }
+
+  Status = FindNamedObjectInDevice (DeviceNode, "_HID", &HidNode);
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  // Get the value from fixed argument 1
+  ValueArg = AmlGetFixedArgument ((AML_OBJECT_NODE *)HidNode, EAmlParseIndexTerm1);
```
pierregondois, line 247: There is effectively no function in AmlApi.h to read the string or integer value of a a NameOp. This seems like a regular use case.
Maybe it would be worth putting this logic in AmlApi.h and using this new function in this test file to allow re-using it ? Ideally in a separate PR ?

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/AmlTestHelper.cpp
```diff
@@ -0,0 +1,1315 @@
...
+  // Package opcode is 0x12
+  return AmlNodeCompareOpCode (
+           (AML_OBJECT_NODE *)Node,
+           AML_PACKAGE_OP,
+           0
+           );
+}
+
+BOOLEAN
+AmlTestHelper::IsMethodNode (
```
pierregondois, line 503: The function doesn't seem to be used

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/CommonSsdtCpuTopologyGoogleTest.cpp
```diff
@@ -0,0 +1,2144 @@
...
+
+  if (TotalCpus == 0) {
+    GTEST_SKIP () << "No CPUs in configuration";
+  }
+
+  mProcHierarchyInfo.resize (TotalNodes);
+  UINT32  NodeIndex = 0;
+
+  // Create packages
+  for (UINT32 pkg = 0; pkg < config.PackageCount; pkg++) {
```
pierregondois, line 1680: Similarly to a comment made for the X64 tests, I think he creation of the platform to test could be made generic, i.e. a set of functions could be created to create platforms with N CPUs, M clusters, adding PSD information etc.
If necessary, the generated platform can be then patched in the test to have a more specific setup.

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/Arm/ArmSsdtCpuTopologyGenerator.c
```diff
@@ -270,6 +270,16 @@ CreateTopologyFromIntC (
       }
     }
 
+    // If a PSD info is associated with the
+    // GicCinfo, create an _PSD method returning them.
```
pierregondois, line 274: Would it also be possible to split this patch in 3:
1. Use of _PSD in Arm Ssdt generation
2. Modifications of the tests
3. Fix for CreateTopologyFromIntC()

Maybe 1. could be either placed first in the patch-set or sent separately, this would allow to merge 2. into `DynamicTablesPkg: Add ARM SSDT CPU Topology Unit and Integration Tests`

jbrasen: Will work on splitting this up.

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtCpuTopologyLib/GoogleTest/X64SsdtCpuTopologyIntegrationGoogleTest.cpp
jbrasen: Yeah will work on re-architecting these a bit.  Originally my plan was to have the generic test and then one per an architecture but then I needed stuff from the common files for testing so that layer broke a bit.

Issue #11797, opened 2025-11-24
[Bug]: Build error when STATIC Stack check is enabled on X86 Architectures on GCC Build


GowthamM9974:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

MdePkg

### Which targets are impacted by this bug?

DEBUG, RELEASE

### Current Behavior

When CUSTOM_STACK_CHECK_LIB is set to STATIC, the GCC build fails with undefined references to DebugPrintEnabled and StackCheckFailureHook in the function __stack_chk_fail located in MdePkg/Library/StackCheckLib/StackCheckLibCommonGcc.c.
The .inf file for StackCheckLib does not include DebugLib under the [LibraryClasses] section, although StackCheckFailureHookLib is listed. Despite this, the linker still reports missing symbols.
When CUSTOM_STACK_CHECK_LIB is set to DYNAMIC, the system builds successfully without errors.
This behavior particularly happens in X86 Architectures.

### Expected Behavior

When CUSTOM_STACK_CHECK_LIB is set to STATIC, if there is no buffer Overflow system should build successfully without any errors.

### Steps To Reproduce

Operating System: ubuntu24.04
Toolchain: GCC
Platform: Emulator
Acquired the source using the command: "git clone https://github.com/tianocore/edk2.git" and "git submodule update --init --recursive"
1.Define CUSTOM_STACK_CHECK_LIB = STATIC in EmulatorPkg.dsc
2.Set up the build environment using the command:
"source edksetup.sh"
3.Build the project with the following command:
"build -p EmulatorPkg/EmulatorPkg.dsc -t GCC -a X64"
4.Observe the Error message:
" In function `__stack_chk_fail:MdePkg/Library/StackCheckLib/StackCheckLibCommonGcc.c:37:(.text.__stack_chk_fail+0x47): undefined reference to DebugPrintEnabled'"
[...]

PR #11943, opened 2025-12-29
EmulatorPkg: Use HOST_APPLICATION and add MINGW CLANG

mdkinney:
# Description

* Update EmulatorPkg to use HOST_APPLICATION module type instead of USER_DEFINED to align with UnitTestFrameworkPkg that uses HOST_APPLICATION for all Host-based unit tests.
* Move [BuildOptions] from host module INFs files to EmulatorPkg DSC file. This is required to have finer control over build options for MINGW CLANG builds.
* Add support for MINGW CLANG builds.
* Use WIN_HOST_BUILD to simplify the selection of the Windows host module or the Linux host module.
* Add missing StackCheckLib mapping
* Override PcdOpensslLibAssemblySourceStyleNasm to TRUE for MINGW CLANG builds that use Windows calling convention.
* Add checks for support tool chains with error messages
* Set linker BASE to correct values for DLLs and Applications for 32-bit and 64-bit Windows builds.
* Update DEBUG, RELEASE, and NOOPT profiles with correct optimization settings and source level debug settings.
* Reduce Visual Studio support to VS2019 and VS2022
* Simplify linker settings for Visual Studio builds.
* Increase size of EmulatorPkg FLASH Device to support all supported EmulatorPkg builds. The configuration that produces the largest FD size is CLANGDWARF NOOPT with SECURE_BOOT_ENABLE that is ~10MB.  Set size to ~12 MB for future growth. 
* Some of the NOOPT builds require more FLASH Device space
* Update GdbRun.sh to load symbols as modules are loaded and executed in the EmulatorPkg by executing a command each time the function SecGdbScriptBreak is executed.
* Set PcdEmulatorLazyLoadSymbols to FALSE to disable lazy symbol loading.

Depends on #11925, #11929, #11935, and #11936

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verify build, boot, and expected source level functionality of EmulatorPkg for all currently supported compiler/linkers and the MINGW CLANG compilers/linkers.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
+++ b/BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
@@ -340,7 +340,7 @@ class HostBasedUnitTestRunner(IUefiBuildPlugin):
         if ret != 0:
             logging.error(f"clang_gen_lcov_xml: Failed to generate coverage lcov. {output_lcov}")
             return 1
-        ret = RunCmd("lcov_cobertura",f"{output_lcov} --excludes ^.*UnitTest\|^.*MU\|^.*Mock\|^.*DEBUG -o {output_xml}")
+        ret = RunCmd("lcov_cobertura",f'{output_lcov} --excludes "^.*UnitTest\|^.*MU\|^.*Mock\|^.*DEBUG" -o {output_xml}')
         if ret != 0:
             logging.error(f"clang_gen_lcov_xml: Failed generate filtered coverage XML. {output_xml}")
             return 1
--- a/UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
+++ b/UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
@@ -94,6 +94,7 @@ ArmMemoryAttributeToPageAttribute (
 // T0SZ can be below MIN_T0SZ when LPA2 is in use, meaning the page table starts at level -1
 #define MIN_T0SZ        16
 #define BITS_PER_LEVEL  9
+#define MAX_VA_BITS_48  48
 #define MAX_VA_BITS     52
 
 STATIC
@@ -658,8 +659,13 @@ ArmConfigureMmu (
   // into account the architectural limitations that result from UEFI's
   // use of 4 KB pages.
   //
-  MaxAddressBits = MIN (ArmGetPhysicalAddressBits (), MAX_VA_BITS);
-  MaxAddress     = LShiftU64 (1ULL, MaxAddressBits) - 1;
+  if (ArmHas52BitTgran4 ()) {
+    MaxAddressBits = MIN (ArmGetPhysicalAddressBits (), MAX_VA_BITS);
+  } else {
+    MaxAddressBits = MIN (ArmGetPhysicalAddressBits (), MAX_VA_BITS_48);
+  }
+
+  MaxAddress = LShiftU64 (1ULL, MaxAddressBits) - 1;
 
   T0SZ                = 64 - MaxAddressBits;
   RootTableEntryCount = GetRootTableEntryCount (T0SZ);
--- a/UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
+++ b/UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
@@ -40,9 +40,98 @@
   NULL|UnitTestFrameworkPkg/Library/UnitTestDebugAssertLib/UnitTestDebugAssertLibHost.inf
   StackCheckLib|MdePkg/Library/StackCheckLibNull/StackCheckLibNullHostApplication.inf
 
+#
+# Fail with error message if the OS/Compiler combination is not supported
+#
+# Operating System and Compiler Compatibility Matrix for Host-based unit tests
+# NOTE: Only NOOPT build target is supported for unit test builds
+#
+# +--------------------+--------+----------+------------+-----+----+--------+
+# | OS/Compiler        | VS2019 | CLANGPDB | CLANGDWARF |   GCC    | XCODE5 |
+# |                    | VS2022 |          |            |   GCC5   |        |
+# |                    |        |          |            | GCCNOLTO |        |
+# +--------------------+--------+----------+------------+----------+--------+
+# | Windows/VS         |IA32/X64|          |            |          |        |
+# | Windows/LLVM/VS    |        | IA32/X64 |            |          |        |
+# | Windows/LLVM/MSYS2 |        |          |    X64     |          |        |
+# | Windows/LLVM/MINGW |        |          |  IA32/X64  |          |        |
+# | Linux/LLVM         |        |          |  IA32/X64  |          |        |
+# | Linux/GCC          |        |          |            | IA32/X64 |        |
+# | macOS/XCODE5       |        |          |            |          |IA32/X64|
+# +--------------------+--------+----------+------------+----------+--------+
+#
+# * Windows/VS: Windows environment with Visual Studio installed
+# * Windows/LLVM/VS: Windows environment with Visual Studio and LLVM 20.1.8 or
+#   higher installed. The default version of LLVM installed with Visual Studio
+#   is not supported.
+#   * https://github.com/llvm/llvm-project/releases
+#   * Host-based unit tests depend on Visual Studio includes/libraries
+# * Windows/LLVM/MSYS2: Windows environment with MSYS2 based LLVM 20.1.8 or
+#   higher installed.
+#   * https://github.com/tianocore/edk2-edkrepo/releases?q=clang
+#   * Host-based unit tests do not use any Visual Studio includes/libraries
+#   * MSYS2 based LLVM 20.1.8 does not provide IA32 includes/libraries
+# * Windows/LLVM/MINGW: Windows environment with LLVM MINGW 20.1.8 or higher
+#   installed with UCRT includes/libraries.
+#   * https://github.com/mstorsjo/llvm-mingw/releases
+#   * Host-based unit tests do not use any Visual Studio includes/libraries
+#   * UCRT release provide IA32 and X64 includes/libraries
+# * Linux/LLVM or Linux/GCC: Linux environment with GCC 13 or higher and
+#   LLVM 20.1.8 or higher installed. Not all Linux distributions provide IA32
+#   includes/libraries for Host-based unit tests.
+#   * Host-based unit tests use GCC and LLVM includes/libraries
+#   * Ubuntu 24.04 apt modules for IA32 and X64 includes/libraries:
+#       build-essential uuid-dev lcov
+#       g++-13 gcc-13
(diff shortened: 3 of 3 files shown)


EmulatorPkg/EmulatorPkg.dsc
```diff
@@ -522,25 +527,243 @@
...
-  MSFT:DEBUG_*_*_DLINK_FLAGS = /BASE:0x10000
-  MSFT:NOOPT_*_*_DLINK_FLAGS = /BASE:0x10000
+  MSFT:*_*_IA32_DLINK_FLAGS  = /BASE:0x010000000
+  MSFT:*_*_X64_DLINK_FLAGS   = /BASE:0x180000000
+
+  #
+  # Windows/CLANGPDB using Visual Studio includes and libraries
+  #
+  CLANGPDB:RELEASE_*_*_CC_FLAGS = -g0
+  CLANGPDB:DEBUG_*_*_CC_FLAGS   = -O0 -fno-omit-frame-pointer -g3 -gcodeview -fno-lto
```
os-d: Some of these flags are already set for CLANGPDB DEBUG, e.g. -gcodeview. I think we should drop the ones that are already set in tools_def or change these to `==` (IIRC that's the form that overrides, not appends, right?).

EmulatorPkg/EmulatorPkg.dsc
```diff
@@ -522,25 +527,243 @@
...
+  #
+  # Windows/Mingw CLANGDWARF using Mingw CLANG includes and libraries
+  #
+  GCC:*_CLANGDWARF_IA32_PP_FLAGS    = -target i686-w64-mingw32
+  GCC:*_CLANGDWARF_X64_PP_FLAGS     = -target x86_64-w64-mingw32
+
+  GCC:*_CLANGDWARF_X64_CC_FLAGS     = -target x86_64-w64-mingw32 -mno-stack-arg-probe
+  GCC:*_CLANGDWARF_IA32_CC_FLAGS    = -target i686-w64-mingw32 -mno-stack-arg-probe
+  GCC:RELEASE_CLANGDWARF_*_CC_FLAGS = -g0
+  GCC:DEBUG_CLANGDWARF_*_CC_FLAGS   = -O0 -fno-omit-frame-pointer -fexceptions -g3 -gcodeview -fdebug-macro -fno-lto
```
os-d: Why are we setting -gcodeview for CLANGDWARF? That is what will generate PE/COFF images and PDBs, right?

mdkinney: Yes.  And Mingw  is a Windows env, so we want PDB files to be generated to use Windows debug tools.

os-d: Okay, I see. Seems like CLANGDWARF just isn't supported from this environment, since we are coercing it to be CLANGPDB essentially. I think that's fine, but odd because we have both definitions. Seems like CLANGDWARF MINGW users would be confused by this behavior. I don't have a use case I can think of for building EmulatorPkg with CLANGDWARF from MINGW, though, so if this meets a use case you have, I think that's fine.

mdkinney: Since it is possible for the clang compiler to produce both dwarf and pdb information for a single module, it would allow consumers to choose a debugger that supports pdb or a gdb/lldb debugger. It provides options with no impact to the .efi file size.  The .debug sections and pdb files are preserved in a file outside FD/FVs.

This is where the historical names of CLANGPDB and CLANGDWARF are no longer correct. Especially since the clang compiler can be used for FW modules and for modules of type HOST_APPLICATION.  The difference is really about which linker is used, and not the format of the symbol information. Especially since there are environments where clang can produce both symbols formats.

* CLANGPDB --> Use linker that generates PE/COFF.  Generate PDB symbol info.
* CLANGDWARF --> Use the native environment linker mode (PE/COFF for Windows, ELF or Linux). Generate DWARF symbol information and also generate PDB symbols information if possible.

When used for FW modules, the CLANGPDB means use lld-link that can directly generate PE/COFF.  CLANGDWARF mean use lld to generate ELF and then use GENFW ELF->PE/COFF conversion step.

When used for HOST_APPLICATION modules, CLANGPDB is only used for Windows Visual Studio + CLANG that uses the native lld-link linker that generates Windows Application PE/COFF images.  CLANGDWARF used with Windows Mingw compilers means uses the native environment lld linker that generates a Windows Application PE/COFF image.  CLANGDWARF used with Linux compilers means use the native lld linker that generated Linux Application ELF image.

os-d: I do think that CLANGPDB vs CLANGDWARF does need cleanup because it has gotten very muddy what the toolchain means. But that's independent of this and can come later. I'll resolve this.

EmulatorPkg/EmulatorPkg.dsc
```diff
@@ -521,25 +527,243 @@
...
+  GCC:DEBUG_CLANGDWARF_*_DLINK_FLAGS = -g -Wl,--pdb,$(DEBUG_DIR)/$(BASE_NAME).pdb
+  GCC:NOOPT_CLANGDWARF_*_DLINK_FLAGS = -g -Wl,--pdb,$(DEBUG_DIR)/$(BASE_NAME).pdb
+  #
+  # Set DLINK2_FLAGS to empyty string to disable use of linker script
+  #
+  GCC:*_CLANGDWARF_X64_DLINK2_FLAGS  ==
+  GCC:*_CLANGDWARF_IA32_DLINK2_FLAGS ==
+!endif
+
+  GCC:RELEASE_*_*_CC_FLAGS = -g0
```
khaliid2040, line 677: I wouldn't recommend completely disabling LTO, currently the only issue with LTO is this one #11797. Maybe something like this is better.


```suggestion
#
# Several link issues has been encoutered when linking with StackCheckLib, so disable it when linking with that library.
#
  !if $(CUSTOM_STACK_CHECK_LIB) == STATIC || $(CUSTOM_STACK_CHECK_LIB) == DYNAMIC
    GCC:*_*_*_CC_FLAGS = -fno-lto
  !endif
```

khaliid2040, line 677: I wouldn't recommend completely disabling LTO, currently the only issue with LTO is this one #11797. Maybe something like this is better.


```suggestion
#
# Several link issues has been encoutered when linking with StackCheckLib, so disable it when linking with that library.
#
  !if $(CUSTOM_STACK_CHECK_LIB) == STATIC || $(CUSTOM_STACK_CHECK_LIB) == DYNAMIC
    GCC:*_*_*_CC_FLAGS = -fno-lto
  !endif
```

mdkinney: I have been testing this PR with Linux GCC and Linux CLANG and I do not see any failures with -D CUSTOM_STATIC_CHECK_LIB=DYNAMIC or -D CUSTOM_STATIC_CHECK_LIB=STATIC, so I do not think this additional condition is required.

I will convert this PR from Draft to Normal later today, so I recommend we focus review and testing on this version of EmulatorPkg and consider closing PR #11959 because I think this PR fixes the issue raised in that PR.

EmulatorPkg/EmulatorPkg.dsc
mdkinney: I agree that some of these settings could be moved into tools_def.txt if they apply to both Firmware and Host Applications. Perhaps this can be done as further cleanup later?

Also, -gcodeview is an option that causes problems with the Mingw CLANG builds using CLANGDWARF because -gcodeview works with CC, but is considered an unrecognized option by DLINK. It is an example of why I had the previous change to remove CC_FLAGS from DLINK operations.

os-d: Well, my point is that some of these flags, such as `-gcodeview` are already set for all CLANGPDB DEBUG builds:

https://github.com/tianocore/edk2/blob/923a951e8a77ab62b089295a7e73b78b39940ab8/BaseTools/Conf/tools_def.template#L1707

https://github.com/tianocore/edk2/blob/923a951e8a77ab62b089295a7e73b78b39940ab8/BaseTools/Conf/tools_def.template#L1745

Same with `-fno-omit-frame-pointer` and I think some of the others.

https://github.com/tianocore/edk2/blob/923a951e8a77ab62b089295a7e73b78b39940ab8/BaseTools/Conf/tools_def.template#L1679

So it shouldn't need to be added here and my point was that I think we shouldn't have duplicates of the flags being passed, so either the EmulatorPkg build should drop the duplicates (and it will inherit them from tools_def and append the unique ones) or we should say EmulatorPkg has a unique set of requirements and so we completely don't use the tools_def ones and override with these ones (e.g. using `==`).

I think either of those options is fine, but I think it is confusing to re-add flags here, particularly because EmulatorPkg already has a mess of options for a bunch of different configurations. I think it makes a lot more sense to only add the ones that are specifically uniquely needed here.

mdkinney: I agree.  The example of adding -gcodeview to tools_def for CLANGDWARF will actually break Firmware builds because -gcodeview is not supported by Mingw CLANGDWARF link operation.  

-gcodeview is compatible with CC and DLINK action for CLANGPDB.  That is why it works in tools_def for CLANGPDB

os-d: Right, but what I'm saying is that the line this is commented on is for CLANGPDB. So for EmulatorPkg, CLANGPDB DEBUG will be getting `-gcodeview` and some other flags multiple times. I'd prefer to not add the extra noise, so that it is  clear what the additional flags EmulatorPkg needs to CLANGPDB (and the other toolchains) to work.

e.g. this particular line would become

```c
CLANGPDB:DEBUG_*_*_CC_FLAGS   = -O0 -g3-fno-lto
```

This makes it clear that these are the special flags EmulatorPkg needs for CLANGPDB DEBUG. It makes it easier to reason about and maintain in the future.

Again, if you wanted to split completely from tools_def, that's also fine, if there's enough justification, all I'm asking for is we don't duplicate the flags that we already inherit from tools def for these toolchains.

mdkinney: I agree with CLANGPDB feedback.  I will fix EmulatorPkg CLANGPDB to not add duplicate options.

EmulatorPkg/EmulatorPkg.dsc
mdkinney: @khaliid2040  This PR has been updated to latest and converted to a normal PR.  Please test this PR with the GCC builds that were failing.  If they are still failing please provide build command and options used.

EmulatorPkg/EmulatorPkg.dsc
mdkinney: @os-d  I have pushed an update that removes the redundant options and re-groups the options by category.

khaliid2040 (approved):
RELEASE - OK
DEBUG - OK
NOOPT - OK

EmulatorPkg/EmulatorPkg.dsc
mdkinney: This additional change is not required.  All tests pass with current set  of changes

Issue #11528, opened 2025-09-18
[Bug]: Use flexible array members consistently


makubacki:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

ArmPkg, ArmVirtPkg, FatPkg, IntelFsp2Pkg, MdeModulePkg, MdePkg, NetworkPkg, OvmfPkg, PrmPkg, RedfishPkg, SecurityPkg, ShellPkg, SourceLevelDebugPkg, StandaloneMmPkg, UefiCpuPkg

### Current Behavior

Currently, there are approximately 254 incorrect usages of flexible array member declarations in 111 files in the edk2 repository. These should be updated to use `[]` instead of `[1]`.

This is a tracking issue for all occurrences across the codebase. Subissues per package may need to be attached depending on how package owners want to address the changes.

```
ArmPkg\Drivers\ArmScmiDxe\ArmScmiBaseProtocolPrivate.h:
  24    // length array.
  25:   UINT8     Protocols[1];
  26  } BASE_DISCOVER_LIST;

ArmPkg\Drivers\ArmScmiDxe\ArmScmiClockProtocolPrivate.h:
  43    // length array.
  44:   CLOCK_RATE_DWORD    Rates[1];
  45  } CLOCK_DESCRIBE_RATES;

ArmPkg\Drivers\ArmScmiDxe\ArmScmiPerformanceProtocolPrivate.h:
  33    // length array.
  34:   SCMI_PERFORMANCE_LEVEL    PerfLevel[1]; // Offset to array of performance levels
  35  } PERF_DESCRIBE_LEVELS;

ArmPkg\Include\Library\ArmMtlLib.h:
  30    // length array.
  31:   UINT32    Payload[1]; // size less object gives offset to payload.
  32  } MTL_MAILBOX;

ArmVirtPkg\KvmtoolCfgMgrDxe\ConfigurationManager.h:
   95    ///
   96:   CM_ARM_ITS_IDENTIFIER                           ItsIdentifierArray[1];
   97  

  105    ///
  106:   CM_ARM_ID_MAPPING                               DeviceIdMapping[1];
  107  

FatPkg\EnhancedFatDxe\Fat.h:
  371    //
  372:   CHAR16                             RootFileString[1];
  373    FAT_OFILE                          *Root;

[...]

PR #11949, opened 2026-01-01
Use flexible array members consistently

khaliid2040:
# Description

Closes: #11528 
__________________________________________________

This PR introduces usage of flexible array members consistently to the project. Flexible array members are introduced in C99 and supported by all compilers.

Currently, the project uses predefined sizes like `[1]` to define flexible array, while somewhere else used by flexible array members.

This PR does following:
* All private structures of the project (C structures not defined in any spec) have been changed to use flexible array member, if any.
* All structures defined in the spec have been changed accordingly to match the spec definitions.

## Limitations and challenges

Several issues and limitations have arised from this task including following:

* C sizeof() operator doesn't work with flexible arrays when retrieving flexible array member size e.g  
`sizeof (Structure->Member)`  where `Member is flexible array won't build and not permitted by some compilers.
* When it comes to nested structures( C structure referencing another C structure) the structure being referenced should always be at the bottom of the structure. e.g:
 ```C
struct Foo {
   int A;
   int B[];
};

struct Bar {
  struct Foo foo;
  int C;
} ;
```
Above code won't able to compile :x:

```C
 struct Foo {
   int A;
   int B[];
};

struct Bar {
  int C;
  struct Foo foo;
} ;
``` 
will compile :white_check_mark: 

GCC compiler bypasses the later limitation with  unstandarized extensions, however CLANG and MSVC won't allow unless structure member is changed to be always at the bottom.

In order to resolve those issue, the first issue all usage of `sizeof()` operator to get flexible array members have been changed to use `offsetof()` macro which returns the offset of the member on a beginning of the structure as base.

For the second issue there is no work around, other than changing structure members. So, this PR skipped those structures.

- [x] Breaking change?
  - Yes, expected build failure and runtime issues like assertions
- [ ] Impacts security?
  - NO
- [ ] Includes tests?
  - NO

## How This Was Tested

MdeModulePkg, MdePkg, NetworkPkg, OvmfPkg, PrmPkg, SecurityPkg, RedFishPkg, shellPkg and UefiPayload built with GCC and CLANG. Build was successful.

Also, used project CI for other envirnment and build.
## Integration Instructions
N/A

### DIFF
--- a/UefiPayloadPkg/FvbRuntimeDxe/FvbSmmCommon.h
+++ b/UefiPayloadPkg/FvbRuntimeDxe/FvbSmmCommon.h
@@ -22,7 +22,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 typedef struct {
   UINTN         Function;
   EFI_STATUS    ReturnStatus;
-  UINT8         Data[1];
+  UINT8         Data[];
 } SMM_FVB_COMMUNICATE_FUNCTION_HEADER;
 
 ///


ardbiesheuvel:
We cannot do this: many of these struct types are defined in the UEFI, PI or other specifications.

khaliid2040:
> We cannot do this: many of these struct types are defined in the UEFI, PI or other specifications.

I am aware of that and i changed spec structures accordingly as spec defined. Could you please pinpoint to me specs define C structures other than UEFI and PI. Most of other spec don't define C structures directly.

MdePkg/Include/Protocol/Dhcp6.h
```diff
@@ -156,7 +156,7 @@ typedef struct {
   ///
   /// The data for the DHCPv6 option, stored in network order.
   ///
-  UINT8     Data[1];
```
lgao4, line 159: This structure is defined in UEFI spec. You can check it.

MdePkg/Include/Protocol/BlockIoCrypto.h
```diff
@@ -110,7 +110,7 @@ typedef struct {
   //
   // Array of supported capabilities.
   //
-  EFI_BLOCK_IO_CRYPTO_CAPABILITY    Capabilities[1];
+  EFI_BLOCK_IO_CRYPTO_CAPABILITY    Capabilities[];
 } EFI_BLOCK_IO_CRYPTO_CAPABILITIES;
```
lgao4, line 114: This one is defined in UEFI spec. You can check spec.

MdePkg/Include/Protocol/Dhcp6.h
khaliid2040: Thanks, i excluded from it

MdePkg/Include/Protocol/BlockIoCrypto.h
khaliid2040: Also excluded

jyao1:
This PR does not follow UEFI spec and will break the compatibility.

@mdkinney , is there any recommendation?

jyao1:
@khaliid2040 , please clarify how you do the test to ensure the original feature still works, after this change.

> Build was successful

That is not sufficient, because if the size_of() calculation is wrong, it will impact the runtime.

khaliid2040:
> This PR does not follow UEFI spec and will break the compatibility.
> 
> @mdkinney , is there any recommendation?

There are some structure UEFI defined as flexible array for the structure, however in edk2 implementation it is defined as `[1]`. I agree this is breaking change.

Flexible arrays don't have compile time size. So retrieving size of flexible array won't work.  using offsetof() macro is intended to work with flexible arrays.

> @khaliid2040 , please clarify how you do the test to ensure the original feature still works, after this change.
> 
> > Build was successful
> 
> That is not sufficient, because if the size_of() calculation is wrong, it will impact the runtime.

I will update PR discription to add detailed description.

khaliid2040:
@mdkinney, @lgao4 and @makubacki You were involved the issue linked to this PR. Anything you could add.

I think we should move this PR and the issue to a new direction. We should only change structures the spec defines as flexible array but in edk2 defined the opposite this is safer.

jyao1:
> Flexible arrays don't have compile time size. So retrieving size of flexible array won't work. using offsetof() macro is intended to work with flexible arrays.

The problem is: How we know the fix is sufficient?
We only know what is fixed, but we cannot know what is NOT fixed in other codebase which is consuming the data structure. Some code may even be close source. You never know.

Also this kind of issue (still using old sizeof for old structure) is hard detected at build time because build will pass. The people who does the porting might be even not aware of such trick. How can we prevent such porting issue from happening?

(And I did encounter such issue before, because one definition in MdePkg is changed differently from the one defined in Tool. It took me lots of time to check what was wrong.)

I hope we can get a clear overall picture for the smooth transition, before we jump to this direction.

khaliid2040:
> > Flexible arrays don't have compile time size. So retrieving size of flexible array won't work. using offsetof() macro is intended to work with flexible arrays.
> 
> The problem is: How we know the fix is sufficient? We only know what is fixed, but we cannot know what is NOT fixed in other codebase which is consuming the data structure. Some code may even be close source. You never know.
> 
> Also this kind of issue (still using old sizeof for old structure) is hard detected at build time because build will pass. The people who does the porting might be even not aware of such trick. How can we prevent such porting issue from happening?
> 
> (And I did encounter such issue before, because one definition in MdePkg is changed differently from the one defined in Tool. It took me lots of time to check what was wrong.)
> 
> I hope we can get a clear overall picture for the smooth transition, before we jump to this direction.

wouldn't be safe if we just update the structures defined in spec as flexible but not in the project defined differently.

There were much things we didn't discuss on the issue before. But spec structure, there update is mandatory in order to comply with spec. However, still that needs updates with some structures which could be dangerous.

I am waiting if @makubacki have to say anything because he was the one opened the issue.

> The problem is: How we know the fix is sufficient? We only know what is fixed, but we cannot know what is NOT fixed in other codebase which is consuming the data structure. Some code may even be close source. You never know.

It is not sufficient, that is why i am moving this PR into another direction which is changing structures that don't match the spec in terms of flexible array member.

IntelFsp2Pkg/Tools/Tests/ExpectedFspmUpd.h
niruiyu, line 1: This change impacts the sizeof (FSP_M_CONFIG), which will cause issues.
And by design, ReservedFspmUpd is 4-byte. You cannot just remove 4 here.

khaliid2040:
> Please do some experiment to better understand OFFSET_OF, sizeof and flex-array.

Thanks @niruiyu, i will get back this PR into draft. As i indicated above i will leave every structure and change and just focus on spec structures that don't match the spec when it comes to flexible array.

And honestly, i would suggest everyone to comment on the issue instead of this PR. We can even close the issue which will close this PR too.

mdkinney:
I think there are 2 categories to consider.
* Structures internal to the edk2 implementation whose only consumers and producers are within the edk2 repo. These could be safely converted to proper flexible array members with no impact to external consumers of the edk2 repo.
* Structures that are defined in industry standard specifications where the edk2 implementation does not match that spec. This is not constrained to the flexible array member attribute.  However, changes to follow specs have to be carefully evaluated because if the are used by content outside of the edk2 repo, then it may be a compatibility break. So each one will have to be evaluated independently.

I think these are 2 different categories of tasks and should be handled in independent PRs.

IntelFsp2Pkg/Tools/Tests/ExpectedFspmUpd.h
```diff
@@ -40,7 +40,7 @@ typedef struct {
 
   /** Offset 0x00D0
   **/
-  UINT8     ReservedFspmUpd[4];
```
AshrafAliS, line 43: this is not a flexy array.

Issue #48, opened 2016-01-26
RFE: please enable NVMe boot in OVMF


stamerlan:
Hi, guys!

It would be nice to add NVMe boot support (to install windows on NVMe disk for example).

Our company developing SSD and we have simulator for SSD HW (to debug FW and run tests). But currently we need to gather logs from NVMe controller while windows is installing (yes, it's stupid, but it isn't my decision). Those logs will be used to create tests for SSD FW. I've modified QEMU's NVMe controller to create logs in a suitable format, but I can't install Windows on NVMe disk. That's why I ask you to add this feature, it may be useful for someone else too.
It may be useful for OS developers, who are trying to write it's own bootloader and wanna check it on virtual NVMe SSD.
I'm not familiar with OVMF, so I'll try to add NVMe support for SeaBios (I'm not bios developer, but I read SeaBios code for a week, so I more familiar with it).

QEMU cmdline:
qemu-system-x86_64 \
  -drive file=ssd.raw,if=none,id=disk0,format=raw \
  -device nvme,drive=disk0,serial=1234

lersek:
What does NVMe enable that virtio-blk / virtio-scsi disks don't?

(Not arguing against the feature request, I'd just like to know.)

Please also provide a reasonably short QEMU command line that exposes a drive as an NVMe device to the guest. Thanks.

lersek:
The title for this entry should state "OVMF"; can you please edit it?

lersek:
@stamerlan haha, sorry, I was not clear enough. In more details:

This tracker is not just for OVMF, it is for the entirety of edk2. In sensible bug tracking systems, like Bugzilla, you have products, components, sub-components and such -- metadata for each bug report. When you report a bug in those sensible systems, you are expected to make an effort to categorize your bug report *up-front*, in order to save time for the people who *read* the bug reports.

In *this* bug tracker though, we have no such metadata, the best we can do is (apparently) "tags". So @jljusten tends to tag OVMF-related bug reports with the "OVMF" label. But, since you know up-front that you are reporting the issue for OVMF, not another platform built from edk2, you should immediately signal in the title of your report that it is for OVMF.

Therefore the title should contain *both* OVMF *and* NVMe boot.

[...]

PR #11961, opened 2026-01-05
ArmPkg, UefiCpuPkg: Fix boot failure on FEAT_LPA-only systems without LPA2

vishnu430:
# Description

Commit 9077163 added support for 52-bit PA/VA (LPA2) in EDK2. The previous change treated the presence of FEAT_LPA as sufficient to enable 52-bit VA for 4K page granularity. Some platforms advertise FEAT_LPA but do not implement full LPA2 support for 4K PAGE_SIZE; enabling 52-bit VA on those platforms produced an invalid MMU configuration and caused boot failures.

This patch tightens the detection logic so 52-bit PA/VA (LPA2) is enabled only when the platform explicitly advertises LPA2 support. When LPA2 is not present we fall back to the previous 48-bit address limit for 4K pages, preserving correct behavior on non-LPA2 systems.

Fixes: 9077163 ("UefiCpuPkg/ArmMmuLib: Add support for LPA2")

## How This Was Tested

Built and booted EDK2 on a board advertising FEAT_LPA without LPA2 for
  4K pages: system now completes early boot where it previously failed.

## Integration Instructions

N/A

### DIFF
--- a/UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
+++ b/UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
@@ -94,6 +94,7 @@ ArmMemoryAttributeToPageAttribute (
 // T0SZ can be below MIN_T0SZ when LPA2 is in use, meaning the page table starts at level -1
 #define MIN_T0SZ        16
 #define BITS_PER_LEVEL  9
+#define MAX_VA_BITS_48  48
 #define MAX_VA_BITS     52
 
 STATIC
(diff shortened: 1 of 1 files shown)


vbpandya:
I think you need to rebase the PR on latest tip and not add merge commits

vishnu430:
> I think you need to rebase the PR on latest tip and not add merge commits

Thanks @vbpandya. I rebased to latest master and all checks were passed.

nhle-ampere:
That commit should not merge because CPU vendors can fix their silicon issue by adjusting the function `ArmGetPhysicalAddressBits`

Exp:

```
ASM_FUNC(ArmGetPhysicalAddressBits)
    mrs   x0, id_aa64mmfr0_el1
    adr   x1, .LPARanges
    and   x0, x0, #0xf
    ldrb  w0, [x1, x0]
    cmp   w0, #48
    b.ne  .OverrideCpuReg
    ret
.OverrideCpuReg:
    mov   w0, #48
    ret
```

Let keep the firmware generic to ensure compatibility across different CPUs/device.
@vishnu430, @LeviYeoReum

ardbiesheuvel:
> That commit should not merge because CPU vendors can fix their silicon issue by adjusting the function `ArmGetPhysicalAddressBits`
> 
> Exp:
> 
> ```
> ASM_FUNC(ArmGetPhysicalAddressBits)
>     mrs   x0, id_aa64mmfr0_el1
>     adr   x1, .LPARanges
>     and   x0, x0, #0xf
>     ldrb  w0, [x1, x0]
>     cmp   w0, #48
>     b.ne  .OverrideCpuReg
>     ret
> .OverrideCpuReg:
>     mov   w0, #48
>     ret
> ```
> 

I am sorry, but what is the point of this example? This function always returns `48` right?

> Let keep the firmware generic to ensure compatibility across different CPUs/device. @vishnu430, @LeviYeoReum

What are you objecting to here, exactly? Generic code should not assume `LPA2` support based on a PA range of `52`, because that PA range might only be available when using 64k pages.

nhle-ampere:
My example is the possible fix for silicon vendors, who builds the register `id_aa64mmfr0_el1` in the wrong construction. In my perspective, the additional `MAX_VA_BITS_48` is not meaningful anyway.

nhle-ampere:
I opened the PR for discussion. (https://github.com/tianocore/edk2/pull/12071)

PR #11964, opened 2026-01-06
OvmfPkg/PlatformInitLib: Add SoftReserved memory type support

JerryHuang802:
Add support for E820 Soft Reserved Memory regions in OVMF firmware. 
This memory type is used by QEMU to expose Specific Purpose Memory (SPM).

Changes:
- Add EfiAcpiAddressRangeSoftReserved to EFI_ACPI_MEMORY_TYPE enum
- Handle SoftReserved in PlatformAddHobCB() by creating Resource HOBs 
  with standard memory attributes plus EFI_RESOURCE_ATTRIBUTE_SPECIAL_PURPOSE

## Description

SPM (Specific Purpose Memory) is defined in ACPI 6.4+ specification for 
memory intended for specific use cases such as High Bandwidth Memory (HBM) 
attached to accelerators/GPUs.

Related QEMU patches:
https://lore.kernel.org/qemu-devel/1b2acdb2-ad9d-4fdc-bd3b-a13249cc0ab9@kernel.org/T/

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- Built OVMF with this change
- Tested in QEMU with `-numa node,spm=on` configuration
- Verified SPM regions are recognized and reported to guest OS

## Integration Instructions

N/A

Cc: @ardbiesheuvel @jyao1 @kraxel


kraxel:
CI failures (which I guess is uncrustify codestyle check failing) need fixing.

Otherwise the update looks good to me.

Usual procedure for this kind of changes is to wait for the qemu changes being merged before merging the corresponding firmware changes too.

mikebeaton:
Is there a spec. doc. defining the `0xEFFFFFFF` value somewhere? Ofc indeed matches the linked QEMU code, but wondering where - apart from that - the value is defined.

mikebeaton:
> Is there a spec. doc. defining the `0xEFFFFFFF` value somewhere? Ofc indeed matches the linked QEMU code, but wondering where - apart from that - the value is defined.

Hi @JerryHuang802 - I spotted you already answered this in https://edk2.groups.io/g/devel/message/121742 - from my pov it would be useful to add most of the info which you put into that list message into the commit message here, for context and future reference - if you would agree Gerd, @kraxel?

JerryHuang802:
> > Is there a spec. doc. defining the `0xEFFFFFFF` value somewhere? Ofc indeed matches the linked QEMU code, but wondering where - apart from that - the value is defined.
> 
> Hi @JerryHuang802 - I spotted you already answered this in https://edk2.groups.io/g/devel/message/121742 - from my pov it would be useful to add most of the info which you put into that list message into the commit message here, for context and future reference - if you would agree Gerd, @kraxel?

Hi @mikebeaton,
Thank you for the suggestion! I've updated the commit message to include the background information.

JerryHuang802:
> CI failures (which I guess is uncrustify codestyle check failing) need fixing.
> 
> Otherwise the update looks good to me.
> 
> Usual procedure for this kind of changes is to wait for the qemu changes being merged before merging the corresponding firmware changes too.

Hi @kraxel,
Thank you for the review and feedback!
1. The CI failures have been fixed in the latest push.
2. I agree with the usual procedure. On the QEMU side, the discussion has reached consensus on the basic implementation logic. The current discussion is mainly about the parameter naming/expression (e.g., whether to use memmap_type or other forms), which should not affect the firmware-side implementation.
3. I completely understand and respect your decision. While I would hope this patch could potentially be merged earlier since the firmware implementation is independent of the QEMU parameter discussion, I fully respect your judgment on the timing. Please let me know how you'd like to proceed.

kraxel:
> > Usual procedure for this kind of changes is to wait for the qemu changes being merged before merging the corresponding firmware changes too.

>     3. I completely understand and respect your decision. While I would hope this patch could potentially be merged earlier since the firmware implementation is independent of the QEMU parameter discussion, I fully respect your judgment on the timing. Please let me know how you'd like to proceed.

There is no time pressure right now, one more month before code freeze for the next edk2 stable tag starts, which should be enough time to get the qemu changes merged given that there are no fundamental objections and only implementation details are discussed.  Should this not happen we can revisit options when the soft freeze deadline comes closer.

mikebeaton:
@kraxel - I think this one might be good to merge?

kraxel:
The qemu code is not yet merged.

Also checked the acpi spec (6.5a), looking for EfiAcpiAddressRangeSoftReserved and only found "13 to 0xEFFFFFFF -> Undefined" (Table 15.1).

JerryHuang802:
> The qemu code is not yet merged.
> 
Hi Gerd,
Thanks for checking. The QEMU code is not yet merged, but the reviewer is setting up tests, so it should be close.

> Also checked the acpi spec (6.5a), looking for EfiAcpiAddressRangeSoftReserved and only found "13 to 0xEFFFFFFF -> Undefined" (Table 15.1).
> 
Regarding the ACPI spec — yes, Table 15.1 lists 13 to 0xEFFFFFFF as "Undefined".
The value `0xEFFFFFFF` is not defined by ACPI itself. It is the Linux kernel's E820
translation of the `EFI_MEMORY_SP` attribute, which has been defined in the UEFI
specification since version 2.8.

The kernel already uses this throughout the codebase:
- Type definition: https://elixir.bootlin.com/linux/v6.18/source/arch/x86/include/asm/e820/types.h#L37
- EFI → E820 translation: https://elixir.bootlin.com/linux/v6.18/source/arch/x86/platform/efi/efi.c#L135
- E820 handling: https://elixir.bootlin.com/linux/v6.18/source/arch/x86/kernel/e820.c#L1293

QEMU reuses this same value to expose SPM regions via E820, and this OVMF patch
allows the firmware to recognize it so the guest kernel can properly handle them.

kraxel:
>     * Type definition: https://elixir.bootlin.com/linux/v6.18/source/arch/x86/include/asm/e820/types.h#L37

Ok, I think that is good enough to merge things even with the qemu patches not being merged yet.

ardbiesheuvel (approved):

Issue #11965, opened 2026-01-06
[Bug]: qemu-kvm GUEST start failed on RISC-V


yechao-w:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

OvmfPkg

### Which targets are impacted by this bug?

RELEASE, DEBUG

### Current Behavior

1. qemu-kvm use the RISCV_VIRT_CODE.fd to start, the xml configuration of libvirt:
```
  <os>
    <type arch='riscv64' machine='virt'>hvm</type>
    <loader readonly='yes' type='pflash'>/home/RISCV_VIRT_CODE.fd</loader>
    <nvram>/home/RISCV_VIRT_VARS.fd</nvram>
    <boot dev="hd"/>
    <boot dev="cdrom"/>
  </os>
```
2. the "EXCEPT_RISCV_ILLEGAL_INST" exception happened:
```
FSOpen: Open '\EFI\BOOT\BOOTRISCV64.EFI' Success
[Bds] Expand PciRoot(0x0)/Pci(0x1,0x3)/Pci(0x0,0x0) -> PciRoot(0x0)/Pci(0x1,0x3)/Pci(0x0,0x0)/HD(1,GPT,2DF34A60-A57D-488A-881E-812E3F647716,0x800,0xFA000)/\EFI\BOOT\BOOTRISCV64.EFI
[Security] 3rd party image[0] can be loaded after EndOfDxe: PciRoot(0x0)/Pci(0x1,0x3)/Pci(0x0,0x0)/HD(1,GPT,2DF34A60-A57D-488A-881E-812E3F647716,0x800,0xFA000)/\EFI\BOOT\BOOTRISCV64.EFI.
InstallProtocolInterface: 5B1B31A1-9562-11D2-8E3F-00A0C969723B 17EC0C040
Loading driver at 0x0017DE21000 EntryPoint=0x0017DE22000 
InstallProtocolInterface: BC62157E-3E33-4FEC-9920-2D3B36D750DF 17EC0CE98
ProtectUefiImageCommon - 0x7EC0C040
  - 0x000000017DE21000 - 0x00000000004B0000
SetUefiImageMemoryAttributes - 0x000000017DE21000 - 0x0000000000001000 (0x0000000000004000)
SetUefiImageMemoryAttributes - 0x000000017DE22000 - 0x0000000000018000 (0x0000000000020000)
[...]

PR #11966, opened 2026-01-06
MdePkg/BaseCpuLib: Add Vector initialization support for RISC-V

yechao-w:
Add InitializeVectorUnits function to Enable RISC-V vector.

# Description
https://github.com/tianocore/edk2/issues/11965
The edk2 firmware enables the CSR_SSTATUS.VS bit default on RISC-V, allowing the BOOTRISCV64.EFI program to use vector instructions.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Use the BOOTRISCV64.EFI  compiled by GCC14,  the qemu-kvm GUEST start success

## Integration Instructions

N/A

### DIFF
--- a/UefiCpuPkg/CpuDxeRiscV64/CpuDxe.c
+++ b/UefiCpuPkg/CpuDxeRiscV64/CpuDxe.c
@@ -363,6 +363,11 @@ InitializeCpu (
   InitializeFloatingPointUnits ();
 
   //
+  // Initialize Vector
+  //
+  InitializeVectorUnits ();
+
+  //
   // Install Boot protocol
   //
   Status = gBS->InstallProtocolInterface (
--- a/MdePkg/Include/Library/CpuLib.h
+++ b/MdePkg/Include/Library/CpuLib.h
@@ -122,4 +122,17 @@ DisableFloatingPointUnits (
 
 #endif
 
+#if defined (MDE_CPU_RISCV64)
+
+/**
+  Initialize the CPU Vector units.
+**/
+VOID
+EFIAPI
+InitializeVectorUnits (
+  VOID
+  );
+
+#endif
+
 #endif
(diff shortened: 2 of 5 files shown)


UefiCpuPkg/CpuDxeRiscV64/CpuDxe.c
```diff
@@ -362,6 +362,11 @@ InitializeCpu (
   //
   InitializeFloatingPointUnits ();
 
+  //
+  // Initialize Vector
+  //
+  InitializeVectorUnits ();
+
```
vlsunil, line 369: I don't think it is correct to initialize vector unit here. I think grub should be compiled such that it doesn't generate vector instructions using proper -march. There is no guarantee that vector unit is supported in a cpu.

yechao-w: If GRUB is compiled without enabling vector instructions, the system runs normally. However, different OS distributions may have their own -march compilation parameters, making it impossible for the EDK2 firmware to be compatible with all guest OS systems. Therefore, I want to modify EDK2 to enable vector support as a way to improve its compatibility. 

However, it indeed lacks a check to determine whether the CPU supports vector extensions. Is there a simple method in EDK2 to check for vector support? I attempted to read CSR_MISA to determine this, but found that it causes an instruction exception.

vlsunil: EDK2 should not enable the vector unit or leave it enabled by default. If GRUB does not use vector instructions, it should not be compiled with vector support across all OS distributions. If vector support is required, GRUB should first detect whether the CPU supports it, enable and use it as needed, and then restore any required state afterward. Otherwise, if the CPU does not support vector instructions, the same issue will occur again, right?

PR #11972, opened 2026-01-08
MdeModulePkg/Core/Dxe/Mem: Prevent allocations around bins

mdkinney:
# Description

Fix alignment issues in memory map entries returned by `GetMemoryMap()` when `RUNTIME_PAGE_ALLOCATION_GRANULARITY` is larger than `DEFAULT_PAGE_ALLOCATION_GRANULARITY`.

Alignment issues are addressed in the initial memory map
layout when Memory Type Information is provided with memory
bins that use RUNTIME_PAGE_ALLOCATION_GRANULARITY.

There are no issues in the Page/Pool Allocation/Free services.
* CoreSetMemoryTypeInformationRange() make sure there is room
  for all bins when accounting for alignment requirements.
  Allocate space for bins with base and length following
  alignment requirements.
* CoreSetMemoryTypeInformationRange() round up NumberOfPages in
  Memory Type Information based on alignment requirements.
  This is required so GetMemoryMap() will generate memory
  map entries that always follow alignment requirements.
* CoreAddMemoryDescriptor() round up NumberOfPages in
  Memory Type Information based on alignment requirements.
  This is required so GetMemoryMap() will generate memory
  map entries that always follow alignment requirements.

## Memory Layout
```
┌────────────────────┐
│   System Memory    │
│   (Above Bins)     │
├────────────────────┤
│ ~ Gap (Optional) ~ │
│┌──────────────────┐│
││  Memory Bin #N   ││
│└──────────────────┘│
│ ~ Gap (Optional) ~ │
│      ...           │
│ ~ Gap (Optional) ~ │
│┌──────────────────┐│
││  Memory Bin #1   ││
│└──────────────────┘│
│ ~ Gap (Optional) ~ │
│┌──────────────────┐│
││  Memory Bin #0   ││
│└──────────────────┘│
├────────────────────┤ <-- mDefaultMaximumAddress
│   System Memory    │
│   (Below Bins)     │
├────────────────────┤ <-- mDefaultBaseAddress
│    Guard Region    │
│     Size 4KB       │
└────────────────────┘
```
## Memory Allocation Rules

* Memory bins are allocated in a contiguous region that supports all the bins with their required alignments.
  * Each memory bin has a unique MemoryType
  * Alignment requirements must be met for each bin
  * Satisfying alignment requirements may introduce gaps between bins
* Never allocate from any Guard Region
  * Guard Region at address 0 is used to catch NULL pointer dereferences.
* Allocate from the highest possible memory region that satisfies the requested size with the following priority
  1) If MemoryType matches one of the memory bins, then attempt allocate from the matching memory bin
  2) If MemoryType matches one of the memory bins, and the matching memory bin is full, then attempt to allocate below bins
  3) If MemoryType does not match any of the memory bins, then attempt to allocate below memory bins

## GetMemoryMap() Updates

Logic issues issues are addressed in the memory map returned
by GetMemoryMap() due to missing cases for memory map entries
of type EfiConventionalMemory that overlap special memory bins.
Add logic to handle all possible memory map splits required to
convert internal memory map entries into an EFI Memory Map with
EFI Memory Map entries the follow alignment requirements when
the EFI Memory Map entries cover memory bins.

The four cases that must be handled are:
* Memory map entry contained within a bin.    [Already covered]
  Convert memory map entry type
* Memory map entry overlaps beginning of bin. [Added]
  Split memory map entry at beginning of bin.
* Memory map entry overlaps end of bin.       [Added]
  Split memory map entry at end of bin.
* Memory map entry overlaps entire bin.       [Added]
  Split memory map entry at both ends of bin.

## Case 1: Memory Map Entry Fits Entirely Within a Memory Bin

```
┌────────────┐
│ Memory Bin │
│   Type X   │
│            │  ┌───────────────────────┐
│            │  │    Memory Map Entry   │
│            │  │ EfiConventionalMemory │
│            │  └───────────────────────┘
└────────────┘
```
Convert Memory Map Entry to Type X and merge with adjacent entries of same type.

```
┌────────────┐
│ Memory Bin │
│   Type X   │
│            │  ┌───────────────────────┐
│            │  │    Memory Map Entry   │
│            │  │        Type X         │
│            │  └───────────────────────┘
└────────────┘
```

## Case 2: Memory Map Entry Overlaps with Lower Bound of a Memory Bin

```
┌────────────┐
│ Memory Bin │
│   Type X   │
│            │  ┌───────────────────────┐    
│            │  │    Memory Map Entry   │
│            │  │ EfiConventionalMemory │
│            │  │                       │
└────────────┘  │                       │
                │                       │
                │                       │
                └───────────────────────┘
```

Split Memory Map Entry at lower bound and convert upper portion of split to Type X and merge with adjacent entries of same type. Lower portion of split remains with type EfiConventionalMemory.

```
┌────────────┐
│ Memory Bin │
│   Type X   │
│            │  ┌───────────────────────┐    
│            │  │    Memory Map Entry   │
│            │  │         Type X        │
│            │  │                       │
└────────────┘  ├───────────────────────┤
                │    Memory Map Entry   │
                │ EfiConventionalMemory │
                └───────────────────────┘
```

## Case 3: Memory Map Entry Overlaps with Upper Bound of a Memory Bin

```
                ┌───────────────────────┐     
                │    Memory Map Entry   │
                │ EfiConventionalMemory │
┌────────────┐  │                       │
│ Memory Bin │  │                       │
│   Type X   │  │                       │
│            │  └───────────────────────┘
│            │
│            │
│            │
└────────────┘
```

Split Memory Map Entry at upper bound and convert lower portion of split to Type X and merge with adjacent entries of same type. Upper portion of split remains with type EfiConventionalMemory.

```
                ┌───────────────────────┐     
                │    Memory Map Entry   │
                │ EfiConventionalMemory │
┌────────────┐  ├───────────────────────┤
│ Memory Bin │  │    Memory Map Entry   │
│   Type X   │  │         Type X        │
│            │  └───────────────────────┘
│            │
│            │
│            │
└────────────┘
```

## Case 4: Memory Map Entry Overlaps entire Memory Bin

```
                ┌───────────────────────┐     
                │    Memory Map Entry   │
                │ EfiConventionalMemory │
┌────────────┐  │                       │
│ Memory Bin │  │                       │
│   Type X   │  │                       │
│            │  │                       │
│            │  │                       │
│            │  │                       │
│            │  │                       │
└────────────┘  │                       │
                │                       │
                │                       │
                └───────────────────────┘
```
Split Memory Map Entry at both bounds of Memory Bin. Convert middle portion of split to Type X and merge with adjacent entries of same type. Upper and lower portions of split remains with type EfiConventionalMemory. 

```
                ┌───────────────────────┐     
                │    Memory Map Entry   │
                │ EfiConventionalMemory │
┌────────────┐  ├───────────────────────┤
│ Memory Bin │  │    Memory Map Entry   │
│   Type X   │  │         Type X        │
│            │  │                       │
│            │  │                       │
│            │  │                       │
│            │  │                       │
└────────────┘  ├───────────────────────┤
                │    Memory Map Entry   │
                │ EfiConventionalMemory │
                └───────────────────────┘
```

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

TBD

## Integration Instructions

N/A


mdkinney:
@kuqin12 Can you please review and test this as an alternative to #11273

kuqin12:
> @kuqin12 Can you please review and test this as an alternative to #11273

@mdkinney thanks! I will do the test sometime today.

mdkinney:
@kuqin12 I fixed one logic bug that was breaking OVMF in CI.  That is resolved now.

I have also simplified the `AllocateAddress` logic.  It currently allows `AllocateAddress` to allocate pages if the address falls within a bin that is not tagged as `Special`. Since the bins are dynamically allocated based on resource descriptor HOBs passed into DXE, I do not think it makes sense to allow allocation requests at a specific address to be supported within the bins.  This could cause some of the bin statistics to be incorrect and change over time based on the region the bins are allocated from at DXE init.  It is simpler to not allow an `AllocateAddress` request at an address that falls within the bins.

mdkinney:
Pushed one more update to refine comments and to adjust logic to guarantee no math over/under flows.

Also simplifies the check for a request that spans/overlaps bins area.

kuqin12:
> @kuqin12 I fixed one logic bug that was breaking OVMF in CI. That is resolved now.
> 

@mdkinney I tried the change included by removing other changes in `CoreAddRange`, but still keeps the intersection check in `CoreGetMemoryMap`. And the system still asserts when the memory map is fetched.

To provide more information, this tested with QEMU-SBSA platform. In that case, the created bin has such:

```text
INFO - CoreGetMemoryMap: Memory Type Bin 0, Base Address 0x10006660000, Max Address 0x10006B64FFF, Number Of Pages 0x505, Special 1, Runtime 0
INFO - CoreGetMemoryMap: Memory Type Bin 5, Base Address 0x100061A0000, Max Address 0x100063FFFFF, Number Of Pages 0x260, Special 1, Runtime 1
INFO - CoreGetMemoryMap: Memory Type Bin 6, Base Address 0x10006400000, Max Address 0x10006657FFF, Number Of Pages 0x258, Special 1, Runtime 1
INFO - CoreGetMemoryMap: Memory Type Bin 9, Base Address 0x10006BAC000, Max Address 0x10006BEBFFF, Number Of Pages 0x40, Special 1, Runtime 0
INFO - CoreGetMemoryMap: Memory Type Bin 10, Base Address 0x10006B70000, Max Address 0x10006B97FFF, Number Of Pages 0x28, Special 1, Runtime 0
```

But the system still asserts on GetMemoryMap from ArmCpuDxe when it launches. The failing entry is `Type 7, Start 0x1000619F000, End 0x1000663FFFF`, which seems to start from where the guard page should be.

The problem seems that, if we allocate something from Type 5 bin, and then free it, I do not think there is any mechanism that prevents the CoreAddRange from merging this freed region into the Guard Region.

mdkinney:
This seems more like a bug in logic in GetMemoryMap().  If the bottom page of a bin is freed and the region merged with a free page below the bin, that should be acceptable and GetMemoryMap() should handle that case.  If we can fix GetMemoryMap(), then the guard pages could be removed.  I will review GetMemoryMap() logic next.

kuqin12:
> This seems more like a bug in logic in GetMemoryMap(). If the bottom page of a bin is freed and the region merged with a free page below the bin, that should be acceptable and GetMemoryMap() should handle that case. If we can fix GetMemoryMap(), then the guard pages could be removed. I will review GetMemoryMap() logic next.

In this case, are you suggesting that we should report the guard page below the lowest bin as if it were part of that bin? Along similar lines, what’s your perspective on how we should treat the gap regions between other bins?

mdkinney:
> > This seems more like a bug in logic in GetMemoryMap(). If the bottom page of a bin is freed and the region merged with a free page below the bin, that should be acceptable and GetMemoryMap() should handle that case. If we can fix GetMemoryMap(), then the guard pages could be removed. I will review GetMemoryMap() logic next.
> 
> In this case, are you suggesting that we should report the guard page below the lowest bin as if it were part of that bin? Along similar lines, what’s your perspective on how we should treat the gap regions between other bins?

Based on further analysis, I think the only changes required are in GetMemoryMap().  I have identified 3 cases where GetMemoryMap() could do an incorrect action. When comparing a memory map entry that is type EfiConventionalMemory against the bins, only a memory map entry that is completely contained within a bin is handled correctly.  It does not handle an EfiConventionalMemory entry that overlaps the beginning of a bin, overlaps the end of a bin, or completely covers an entire bin. I think if logic is added for these 3 cases, then the generated memory map should be correct.  I will work on an update for further testing.

Can you please provide the build command and QEMU launch command that was used for the QEMU-SBSA platform test case?

mdkinney:
New approach pushed that removes all the previous changes and focused on the init of the memory bins to make sure alignment requirements are followed.  I observed cases where the alignment requirements were not follow in the initial memory bin allocation so that the size of the bin was not aligned. That issue caused the memory map from GetMemoryMap() to not follow alignment requirements. The top down allocation of bins need to round size up to alignment and then round down the start of the bin for alignment. This will generate the expected gaps between the memory bins if RUNTIME_PAGE_ALLOCATION_GRANULARITY is larger than DEFAULT_PAGE_ALLOCATION_GRANULARITY.

Also fix logic gaps in GetMemoryMap() to handle all possible overlaps between memory map entries and bins that are required to be handled to cover all possible cases.  These cases only get exposed when RUNTIME_PAGE_ALLOCATION_GRANULARITY is larger than DEFAULT_PAGE_ALLOCATION_GRANULARITY.  The challenge is that GetMemoryMap() can not use any memory allocation services and has to use the buffer provided by the caller and GetMemoryMap() has to be able to tell the caller how many memory map entries are required. So the algorithm must determine the maximum number of splits required and then shuffle memory map entries so the number of entries required never exceeds the number of entries in the buffer provided by the caller.

# Case 1: Memory Map Entry Fits Entirely Within a Memory Bin

```
┌────────────┐
│ Memory Bin │
│   Type X   │
│            │  ┌───────────────────────┐
│            │  │    Memory Map Entry   │
│            │  │ EfiConventionalMemory │
│            │  └───────────────────────┘
└────────────┘
```
Convert Memory Map Entry to Type X and merge with adjacent entries of same type.

```
┌────────────┐
│ Memory Bin │
│   Type X   │
│            │  ┌───────────────────────┐
│            │  │    Memory Map Entry   │
│            │  │        Type X         │
│            │  └───────────────────────┘
└────────────┘
```

# Case 2: Memory Map Entry Overlaps with Lower Bound of a Memory Bin

```
┌────────────┐
│ Memory Bin │
│   Type X   │
│            │  ┌───────────────────────┐    
│            │  │    Memory Map Entry   │
│            │  │ EfiConventionalMemory │
│            │  │                       │
└────────────┘  │                       │
                │                       │
                │                       │
                └───────────────────────┘
```

Split Memory Map Entry at lower bound and convert upper portion of split to Type X and merge with adjacent entries of same type. Lower portion of split remains with type EfiConventionalMemory.

```
┌────────────┐
│ Memory Bin │
│   Type X   │
│            │  ┌───────────────────────┐    
│            │  │    Memory Map Entry   │
│            │  │         Type X        │
│            │  │                       │
└────────────┘  ├───────────────────────┤
                │    Memory Map Entry   │
                │ EfiConventionalMemory │
                └───────────────────────┘
```

# Case 3: Memory Map Entry Overlaps with Upper Bound of a Memory Bin

```
                ┌───────────────────────┐     
                │    Memory Map Entry   │
                │ EfiConventionalMemory │
┌────────────┐  │                       │
│ Memory Bin │  │                       │
│   Type X   │  │                       │
│            │  └───────────────────────┘
│            │
│            │
│            │
└────────────┘
```

Split Memory Map Entry at upper bound and convert lower portion of split to Type X and merge with adjacent entries of same type. Upper portion of split remains with type EfiConventionalMemory.

```
                ┌───────────────────────┐     
                │    Memory Map Entry   │
                │ EfiConventionalMemory │
┌────────────┐  ├───────────────────────┤
│ Memory Bin │  │    Memory Map Entry   │
│   Type X   │  │         Type X        │
│            │  └───────────────────────┘
│            │
│            │
│            │
└────────────┘
```

# Case 3: Memory Map Entry Overlaps entire Memory Bin

```
                ┌───────────────────────┐     
                │    Memory Map Entry   │
                │ EfiConventionalMemory │
┌────────────┐  │                       │
│ Memory Bin │  │                       │
│   Type X   │  │                       │
│            │  │                       │
│            │  │                       │
│            │  │                       │
│            │  │                       │
└────────────┘  │                       │
                │                       │
                │                       │
                └───────────────────────┘
```
Split Memory Map Entry at both bounds of Memory Bin. Convert middle portion of split to Type X and merge with adjacent entries of same type. Upper and lower portions of split remains with type EfiConventionalMemory. 

```
                ┌───────────────────────┐     
                │    Memory Map Entry   │
                │ EfiConventionalMemory │
┌────────────┐  ├───────────────────────┤
│ Memory Bin │  │    Memory Map Entry   │
│   Type X   │  │         Type X        │
│            │  │                       │
│            │  │                       │
│            │  │                       │
│            │  │                       │
└────────────┘  ├───────────────────────┤
                │    Memory Map Entry   │
                │ EfiConventionalMemory │
                └───────────────────────┘
```

kuqin12:
@mdkinney thanks for the update. I tested the latest change and it seems good now.

As per SBSA testing, I was using mu_tiano_platforms, but I pushed a platform level branch here: https://github.com/kuqin12/mu_tiano_platforms/tree/test_new_alloc, and the DXE core change is here: https://github.com/kuqin12/mu_basecore/tree/test_newallo.

The build is the same as CI process, but they are described here: https://github.com/microsoft/mu_tiano_platforms/blob/main/Platforms/Docs/Common/building.md.

I will try this one a physical platform tomorrow.

kuqin12:
@mdkinney I just tested on physical platforms as well, this also seems good in my case.

Functionally, I think we can go with this change, but I would prefer to keep the overlap check in the `GetMemoryMap`, at least guarded inside `DEBUG_CODE`. Thanks for putting this together.

mdkinney:
> @mdkinney I just tested on physical platforms as well, this also seems good in my case.
> 
> Functionally, I think we can go with this change, but I would prefer to keep the overlap check in the `GetMemoryMap`, at least guarded inside `DEBUG_CODE`. Thanks for putting this together.

Do you want to add that overlap check on top of this PR?  I think we can break these changes into multiple commits.
* DXE Core Memory Bin init with changes for aligning base and length
* GetMemoryMap() updates to handle all possible memory map and memory bin overlaps
* GetMemoryMap() consistency checks in DEBUG_CODE()

kuqin12:
> Do you want to add that overlap check on top of this PR? I think we can break these changes into multiple commits.
> 
> * DXE Core Memory Bin init with changes for aligning base and length
> * GetMemoryMap() updates to handle all possible memory map and memory bin overlaps
> * GetMemoryMap() consistency checks in DEBUG_CODE()

This sounds good to me! I will follow up with that change in a separate PR. Thanks

kuqin12:
Tested-by: Kun Qin <kun.qin@microsoft.com>

mdkinney:
@kuqin12 Do you want to see this PR opened for review with current content?

kuqin12:
> @kuqin12 Do you want to see this PR opened for review with current content?

Yes, please. Let's do that and I will follow up with a check PR next.

mdkinney:
Pushed update to split into 2 commits.  One for DXE init and one for GetMemoryMap().

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -1893,6 +1951,36 @@ MergeMemoryMapDescriptor (
...
+  Sets the memory type for an EFI memory descriptor and updates its attributes.
+
+  This function sets the Type field of the provided EFI_MEMORY_DESCRIPTOR and
+  then updates the memory map entry attributes based on the memory type statistics.
+
+  @param  MemoryMap  Pointer to the EFI_MEMORY_DESCRIPTOR to update
+  @param  Type       The memory type to set
+
+**/
+static
```
kuqin12, line 1964: Should this be `STATIC` for consistency?

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -2013,50 +2119,198 @@ CoreGetMemoryMap (
...
 
-    //
-    // Check to see if the new Memory Map Descriptor can be merged with an
-    // existing descriptor if they are adjacent and have the same attributes
-    //
-    MemoryMap = MergeMemoryMapDescriptor (MemoryMapStart, MemoryMap, Size);
+    BinStart = mMemoryTypeStatistics[Type].BaseAddress;
+    BinEnd   = mMemoryTypeStatistics[Type].MaximumAddress;
+
+    for (Modified = TRUE; Modified; ) {
```
kuqin12: Is this basically `while (Modified) {`

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -580,7 +583,33 @@ CoreSetMemoryTypeInformationRange (
       continue;
     }
 
-    Size += EFI_PAGES_TO_SIZE (gMemoryTypeInformation[Index].NumberOfPages);
+    if (gMemoryTypeInformation[Index].NumberOfPages != 0) {
+      Alignment = DEFAULT_PAGE_ALLOCATION_GRANULARITY;
+      if ((gMemoryTypeInformation[Index].Type == EfiReservedMemoryType) ||
```
os-d, line 588: This matches the existing behavior, but we should probably be doing this if mMemoryStatistics[Type].Special. This would also include EfiACPIReclaimMemory, which is typically in a bin and not always reclaimed by the OS. I suspect we haven't run into this because Linux may reclaim this (and cares about the runtime granularity > 4K) and Windows does not (and is only ever 4K pages). We don't need to change this here, but ruminating on that.

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -1951,6 +1951,36 @@ MergeMemoryMapDescriptor (
...
+
+**/
+static
+VOID
+SetEfiMemoryDescriptorType (
+  IN EFI_MEMORY_DESCRIPTOR  *MemoryMap,
+  IN EFI_MEMORY_TYPE        Type
+  )
+{
+  ASSERT (MemoryMap != NULL);
```
os-d: nit: return if this is the case? Won't happen in this PR, but for the future and this is the kind of thing CodeQL may flag

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -2013,50 +2119,198 @@ CoreGetMemoryMap (
   //
   ZeroMem (MemoryMap, BufferSize);
   MemoryMapStart = MemoryMap;
+  MemoryMapEnd   = MemoryMapStart;
   for (Link = gMemoryMap.ForwardLink; Link != &gMemoryMap; Link = Link->ForwardLink) {
     Entry = CR (Link, MEMORY_MAP, Link, MEMORY_MAP_SIGNATURE);
     ASSERT (Entry->VirtualStart == 0);
+    MemoryMapEnd->PhysicalStart = Entry->Start;
+    MemoryMapEnd->VirtualStart  = 0;
+    MemoryMapEnd->NumberOfPages = EFI_SIZE_TO_PAGES ((UINTN)(Entry->End - Entry->Start + 1));
```
os-d, line 2132: very small nit, but I think makes the intention clearer for the future reader

```suggestion
    MemoryMapEnd->NumberOfPages = EFI_SIZE_TO_PAGES ((UINTN)(Entry->End + 1 - Entry->Start));
```

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -2013,50 +2119,198 @@ CoreGetMemoryMap (
...
+            //
+            // Unexpected condition since max expected size for worst case splits was
+            // computed and returned.
+            //
+            Status = EFI_BUFFER_TOO_SMALL;
+            goto Done;
+          }
+
+          MemoryMap->PhysicalStart = BinStart;
+          MemoryMap->NumberOfPages = EFI_SIZE_TO_PAGES ((UINTN)(EntryEnd - BinStart + 1));
```
os-d, line 2234: Same very minor nit

```suggestion
          MemoryMap->NumberOfPages = EFI_SIZE_TO_PAGES ((UINTN)(EntryEnd + 1 - BinStart));
```

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -2013,50 +2119,198 @@ CoreGetMemoryMap (
...
+          SetEfiMemoryDescriptorType (MemoryMap, Type);
+
+          if (EntryEnd > BinEnd) {
+            //
+            // Shrink new entry to only cover the memory bin
+            //
+            MemoryMap->NumberOfPages = EFI_SIZE_TO_PAGES ((UINTN)(BinEnd - BinStart + 1));
```
os-d, line 2241:
```suggestion
            MemoryMap->NumberOfPages = EFI_SIZE_TO_PAGES ((UINTN)(BinEnd + 1 - BinStart));
```

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -2013,50 +2119,198 @@ CoreGetMemoryMap (
...
+
+        //
+        // If the memory map entry ends above the memory bin end, then split at the memory bin end
+        //
+        if (EntryEnd > BinEnd) {
+          //
+          // Shrink the original entry for part before the memory bin end
+          //
+          MemoryMap->PhysicalStart = EntryStart;
+          MemoryMap->NumberOfPages = EFI_SIZE_TO_PAGES ((UINTN)(BinEnd - EntryStart + 1));
```
os-d, line 2283:
```suggestion
          MemoryMap->NumberOfPages = EFI_SIZE_TO_PAGES ((UINTN)(BinEnd + 1 - EntryStart));
```

MdeModulePkg/Core/Dxe/Mem/Page.c
mdkinney: The latest recommendation for code style is to use `static`.  New code should use that style.  We will need a different PR to change STATIC -> static for the entire repo.

MdeModulePkg/Core/Dxe/Mem/Page.c
mdkinney: It would be 2 statements:

```
Modified = TRUE;
while (Modified) {
```

Do you prefer that style?

MdeModulePkg/Core/Dxe/Mem/Page.c
mdkinney: I also noticed that some places in DXE Core the Special flag is used and in other these specific memory types are used.  I followed the pattern that I saw for the Page allocation/free that does not use Special.

If you believe that the Special flag should be used for locations in the DXE Core, then that would be a different change. Please open a new PR or Issue for that topic.  We may want to have the set of memory types that require different special treatment to be in their own const global variable and remove Special flag from MemoryTypeStatistics so that special treatment is not limited to just the MemoryTypeInformation feature.

os-d: Yep, agreed that is a separate change. Let me think about that a bit more and I can put up a PR.

MdeModulePkg/Core/Dxe/Mem/Page.c
mdkinney: I decided to use ASSERT() here because this is an internal worker function that is `static` and the ASSERT() only catches misuse of the internal API by other code in the same file. If I change to an error return, then I have to add error handling to all the places this internal helper function is called.

os-d: I don't mean error return, just return. But if CodeQL didn't complain about it on this PR, it's also fine to leave.

MdeModulePkg/Core/Dxe/Mem/Page.c
mdkinney: That was the order of ops in the code before this change.  This order of ops avoids integer overflow if Entry->End is all Fs.

MdeModulePkg/Core/Dxe/Mem/Page.c
mdkinney: This order of ops avoids integer overflow if Entry->End is all Fs.

MdeModulePkg/Core/Dxe/Mem/Page.c
mdkinney: This order of ops avoids integer overflow if BinEnd is all Fs.

MdeModulePkg/Core/Dxe/Mem/Page.c
os-d: Fair enough

MdeModulePkg/Core/Dxe/Mem/Page.c
mdkinney: This order of ops avoids integer overflow if BinEnd is all Fs.

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: That would be how I would write it. But this is a very personal preference, so I will leave it to your call :)

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: I had comment here earlier but did not end up posting it.

Agree that the internal function is running in a controlled environment, so check here is just extra safeguard. But leaving the code to write to null pointer could trip on other exceptions and halt the system, which seems to be a worse error handling compared to not having a valid memory bin.

mdkinney: The point is that an ASSERT() in debug build will halt system with message and force a fix to be made.  If just return, it could be a silent error.

os-d: Well, I meant both

if (MemoryMap == NULL) {
  ASSERT (MemoryMap != NULL;
  return;
}

mdkinney: Update pushed with this change

MdeModulePkg/Core/Dxe/Mem/Page.c
mdkinney: Update pushed with this change

MdeModulePkg/Core/Dxe/Mem/Page.c
mdkinney: Can this conversation be resolved now?

kuqin12: Yep, just resolved.

mdkinney:
@lgao4 Please review

PR #11987, opened 2026-01-13
OvmfPkg/VirtioRngDxe: protect VirtioRngGetRNG against timer callbacks

tpressure:
# Description

This PR raises the TPL level for `VirtioRngGetRNG` to block timer callbacks from executing when we work with the virtio-rng device.

This fixes a [long standing bug](https://github.com/cloud-hypervisor/cloud-hypervisor/issues/6779) first discovered in cloud hypervisor. The issue itself is not limited to cloud-hypervisor though, we have also reports from customers that observed this in qemu.

In our cloud-hypervisor regression test suite, we sometimes see VMs that are stuck in the firmware. This happen roughly in 1 in 1000 VM starts. After a very long time of debugging, we found out that the vCPU is looping in `VirtioFlush` when this happens.

With this [small patch](https://github.com/tpressure/edk2/commit/2e285e9a3e3d4294c2ad7713b209676d3a4fab39), we could pinpoint the issue. For each successful boot, we could observe a log like this:

```
Enter RNG loop TPL:4: 11
Leaving RNG loop: 11
Enter RNG loop TPL:4: 12
Leaving RNG loop: 12
Enter RNG loop TPL:4: 13
Leaving RNG loop: 13
...
<vm boots>
```

In the case of a stuck boot, the log looks like this:

```
Enter RNG loop TPL:4: 11
Leaving RNG loop: 11
Enter RNG loop TPL:4: 12
Enter RNG loop TPL:8: 13 <---- note that TPL is 8 here
Leaving RNG loop: 13
<vcpu is stuck in VirtioFlush of loop 12 here>
```

I'm not an EDK2 expert, but I think this is backtrace (artificial) of the issue

VirtioRngGetRNG (call 1)
└─ VirtioFlush
└─ Timer interrupt
└─ [Notify callback]
└─ GetRandomNumber
└─ VirtioRngGetRNG (call 2)
└─ VirtioFlush

This means that while we're in VirtioRngGetRNG and are using the virtio-rng device, we can be preempted by a timer that just expired, and the callback is calling GetRandomNumber which ultimately calls VirtioRngGetRNG again. After we return to the original call to `VirtioFlush`, the state of the virtqueue is then corrupted and the loop [here](https://github.com/tianocore/edk2/blob/master/OvmfPkg/Library/VirtioLib/VirtioLib.c#L329) never completes.

I fixed this by temporarily raising the TPL to `TPL_NOTIFY` when we work with the device. I tested this and did not hit the issue again in 30.000 VM boots.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested with our [libvirt test suite](https://github.com/cyberus-technology/libvirt-tests), especially this [test](https://github.com/tpressure/libvirt-tests/blob/130d91e9f55d28cedf17936372ed5182a6b9bd03/tests/testscript.py#L1566).

### DIFF
--- a/MdeModulePkg/Bus/Pci/PciBusDxe/PciLib.c
+++ b/MdeModulePkg/Bus/Pci/PciBusDxe/PciLib.c
@@ -1025,7 +1025,7 @@ PciAllocateBusNumber (
 {
   PCI_IO_DEVICE                      *RootBridge;
   EFI_ACPI_ADDRESS_SPACE_DESCRIPTOR  *BusNumberRanges;
-  UINT8                              NextNumber;
+  UINT16                             NextNumber;
   UINT64                             MaxNumberInRange;
 
   //
@@ -1043,18 +1043,18 @@ PciAllocateBusNumber (
   while (BusNumberRanges->Desc != ACPI_END_TAG_DESCRIPTOR) {
     MaxNumberInRange = BusNumberRanges->AddrRangeMin + BusNumberRanges->AddrLen - 1;
     if ((StartBusNumber >= BusNumberRanges->AddrRangeMin) && (StartBusNumber <=  MaxNumberInRange)) {
-      NextNumber = (UINT8)(StartBusNumber + NumberOfBuses);
+      NextNumber = (StartBusNumber + NumberOfBuses);
       while (NextNumber > MaxNumberInRange) {
         ++BusNumberRanges;
         if (BusNumberRanges->Desc == ACPI_END_TAG_DESCRIPTOR) {
           return EFI_OUT_OF_RESOURCES;
         }
 
-        NextNumber       = (UINT8)(NextNumber + (BusNumberRanges->AddrRangeMin - (MaxNumberInRange + 1)));
+        NextNumber       = (UINT16)(NextNumber + (BusNumberRanges->AddrRangeMin - (MaxNumberInRange + 1)));
         MaxNumberInRange = BusNumberRanges->AddrRangeMin + BusNumberRanges->AddrLen - 1;
       }
 
-      *NextBusNumber = NextNumber;
+      *NextBusNumber = (UINT8)NextNumber;
       return EFI_SUCCESS;
     }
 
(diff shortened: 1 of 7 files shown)


OvmfPkg/VirtioRngDxe/VirtioRng.c
```diff
@@ -177,6 +178,8 @@ VirtioRngGetRNG (
     goto FreeBuffer;
   }
 
+  CurrentTpl = gBS->RaiseTPL (TPL_HIGH_LEVEL);
```
ardbiesheuvel: Would it be sufficient to use `TPL_NOTIFY` here? `TPL_HIGH_LEVEL` is really only intended for short critical sections, and such sections should not be making calls into this driver.

tpressure: Since I am not very familiar with EDK2 I cannot answer this question, but this is my guess too. I was hoping someone with more experience could give me this answer :)

 I can test it though.

tpressure: I studied more code and think `TPL_NOTIFY` is sufficient. Testing in progress...

tpressure: Changed to `TPL_NOTIFY`

OvmfPkg/VirtioRngDxe/VirtioRng.c
```diff
@@ -205,6 +208,8 @@ VirtioRngGetRNG (
     ASSERT (Len <= BufferSize);
   }
 
+  gBS->RestoreTPL (CurrentTpl);
+
```
ardbiesheuvel, line 213: You should also restore the TPL if the `goto` path is taken in case `VirtioFlush()` returns with an error.

tpressure: You are absolutely right, good catch.

kraxel (approved):

phip1611:
Hi @kraxel, could you please help us to classify the problem here? Have you seen this in other parts of edk2 as well? Have you experienced similar bugs before?

And has anyone experienced this specific issue?

kraxel:
It's essentially a locking bug.  The TPL is used in edk2 to protect critical sections against re-entry.
And, yes, there have been locking bugs before, for example in IoMmuDxe (see commits a52044a9e602bc168cdf5d73a48952bfc9edb521 and dfb941d32a2f38c9177729e39c6a6515abbbad48).

PR #11989, opened 2026-01-13
MdeModulePkg/Variable/RuntimeDxe: Do not clobber Attributes on error

3rror:
# Description

`FindVariableInSmm()` copied `Attributes` from the SMM response back to the caller even when `GetVariable` returned an error (e.g. `EFI_NOT_FOUND`). In the `EFI_NOT_FOUND` case, the returned `Attributes` value is 0, which overwrote the caller-provided `Attributes`.

This differs from the native `VariableServiceGetVariable()` path, which only updates `Attributes` on `EFI_SUCCESS` or `EFI_BUFFER_TOO_SMALL`. In the SMM path, clobbering `Attributes` on error can cause subsequent `SetVariable` calls to use incorrect attributes (e.g. losing `TIME_BASED_AUTH`), which breaks UEFI-SCT authenticated variable tests when using the SMM gateway backend.

Only copy `Attributes` back to the caller on `EFI_SUCCESS` or `EFI_BUFFER_TOO_SMALL`, preserving the caller-provided value on errors.

Allows [BlackBoxTest/AuthVariableServicesBBTestFunction - "SetVariable with one old timestamp, the return status should be EFI_SECURITY_VIOLATION"](https://github.com/tianocore/edk2-test/blob/master/uefi-sct/SctPkg/TestCase/UEFI/EFI/RuntimeServices/VariableServices/BlackBoxTest/AuthVariableServicesBBTestFunction.c#L1030) tests to pass correctly when using SMM gateway.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Manual run of UEFI SCT test as part of [Arm SystemReady ACS (v25.04_SR_3.0.1)](https://github.com/ARM-software/arm-systemready/tree/v25.04_SR_3.0.1).

Final output:

```
Returned Status Code: Success

AuthVar_Func: [PASSED]
Passes........... 16
Warnings......... 0
Errors........... 0
```

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
+++ b/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
@@ -13,6 +13,7 @@
 
   InitCommunicateBuffer() is really function to check the variable data size.
 
+Copyright (c) 2026, ARM Ltd. All rights reserved.<BR>
 Copyright (c) 2010 - 2024, Intel Corporation. All rights reserved.<BR>
 Copyright (c) Microsoft Corporation.<BR>
 SPDX-License-Identifier: BSD-2-Clause-Patent
@@ -760,10 +761,10 @@ FindVariableInSmm (
     // Only update DataSize when needed
     //
     *DataSize = SmmVariableHeader->DataSize;
-  }
 
-  if (Attributes != NULL) {
-    *Attributes = SmmVariableHeader->Attributes;
+    if (Attributes != NULL) {
+      *Attributes = SmmVariableHeader->Attributes;
+    }
   }
 
   if (EFI_ERROR (Status)) {


LeviYeoReum (approved):

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
```diff
@@ -13,6 +13,7 @@
 
   InitCommunicateBuffer() is really function to check the variable data size.
 
+Copyright (c) 2026 Arm Limited and/or its affiliates <open-source-office@arm.com><BR>
```
samimujawar: This is not the standard copyright notice format that we use in this project.
Please see other files for the copyright notice from Arm.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
```diff
@@ -760,10 +761,10 @@ FindVariableInSmm (
     // Only update DataSize when needed
     //
     *DataSize = SmmVariableHeader->DataSize;
-  }
 
-  if (Attributes != NULL) {
-    *Attributes = SmmVariableHeader->Attributes;
+    if (Attributes != NULL) {
```
samimujawar, line 765: Although I agree with this fix, I also think the SCT test case should not rely on the Attributes value if the return code is not EFI_SUCCESS.

In addition to this fix, can you also look into fixing the SCT test case, please?

samimujawar (changes requested):

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
3rror: Changed to `Copyright (c) 2026, ARM Ltd. All rights reserved.<BR>`

lgao4 (approved):

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
3rror: I raised the issue internally, but I still need to get a response for this.
How would you like to proceed? Are you happy to close this for now?

PR #11992, opened 2026-01-14
CryptoPkg: Add ED448 signature and verification functions

mhollan1:
# Description

- Implemented signature verification for ED448
- Updated documentation and unit tests to cover new verification functions

- [ ] Breaking change?
- [x] Impacts security?
- [x] Includes tests?

## How This Was Tested

Unit Tested with Google Unit Test Framework

## Integration Instructions

N/A

### DIFF
--- a/CryptoPkg/Include/Library/BaseCryptLib.h
+++ b/CryptoPkg/Include/Library/BaseCryptLib.h
@@ -4,7 +4,7 @@
   primitives (Hash Serials, HMAC, RSA, Diffie-Hellman, etc) for UEFI security
   functionality enabling.
 
-Copyright (c) 2009 - 2022, Intel Corporation. All rights reserved.<BR>
+Copyright (c) 2009 - 2026, Intel Corporation. All rights reserved.<BR>
 Copyright (c) Microsoft Corporation. All rights reserved.
 (c) Copyright 2026 HP Development Company, L.P.
 SPDX-License-Identifier: BSD-2-Clause-Patent
@@ -27,6 +27,9 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #define CRYPTO_NID_SECP384R1        0x0205
 #define CRYPTO_NID_SECP521R1        0x0206
 #define CRYPTO_NID_BRAINPOOLP512R1  0x03A5
+
+// EdDSA
+#define CRYPTO_NID_ED448  0x0440
 
 ///
 /// MD5 digest size in bytes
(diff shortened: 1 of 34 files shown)


jyao1:
some initial feedback:
1) The EdDSA/MLDSA/SLHDSA API design is different from the one in ECDSA or RSA. E.g. For ECDSA and RSA, the key can be get from X509 certificate. What is the use case here? Any reason to change? It is weird that the interface design is not aligned.
2) We should add new API to crypto protocol, and corresponding config.
3) The EdDSA/MLDSA/SLHDSA unit test is invalid. It should call EDKII Crypto API, not openssl API. Otherwise, the unit test cannot be used for mbedtls version.

mhollan1:
> some initial feedback:
> 
> 1. The EdDSA/MLDSA/SLHDSA API design is different from the one in ECDSA or RSA. E.g. For ECDSA and RSA, the key can be get from X509 certificate. What is the use case here? Any reason to change? It is weird that the interface design is not aligned.
> 2. We should add new API to crypto protocol, and corresponding config.
> 3. The EdDSA/MLDSA/SLHDSA unit test is invalid. It should call EDKII Crypto API, not openssl API. Otherwise, the unit test cannot be used for mbedtls version.

1.  The use case here is that the user requires authentication of a file.  As an input, they will provide the data, the Public Key and the Signature.  Is there a standard interface design documented?
2.  Where should this be added?
3.  I will update it to reflect the EDKII Crypto API.

jyao1:
> 1. The use case here is that the user requires authentication of a file. As an input, they will provide the data, the Public Key and the Signature. Is there a standard interface design documented?

Please refer to:
https://github.com/tianocore/edk2/blob/master/CryptoPkg/Library/BaseCryptLib/Pk/CryptX509.c
https://github.com/tianocore/edk2/blob/master/CryptoPkg/Library/BaseCryptLib/Pem/CryptPem.c
on how to construct the rsa_context or ec_context.
We need similar eddsa_context and mldsa_context, etc.

> 2. Where should this be added?

Please refer to:
https://github.com/tianocore/edk2/blob/master/CryptoPkg/Include/Pcd/PcdCryptoServiceFamilyEnable.h
https://github.com/tianocore/edk2/blob/master/CryptoPkg/Private/Protocol/Crypto.h

liyi77:
1. Please always use configure.py to update openssl config, instead of manually modifying it.
You can use this commit if you cannot run it: https://github.com/tianocore/edk2-staging/pull/520/changes/46496446219027a83e1f3b7224828632701f25f9

2. No need to run Uncrustify with CryptoPkg/Library/OpensslLib/OpensslStub/uefiprov.c, this file will be skipped by CI format checking.

jyao1:
I think we need DsaSign API besides DsaVerify API. They should follow the existing style and match.
X.509/PEM API for DSA can be added later, but please ensure the API design can be reused for X.509/PEM later. For example, the key_context. We do not want a totally different set of parameter for X.509.

CryptoPkg/Library/BaseCryptLib/Pem/CryptPem.c
```diff
@@ -44,6 +45,86 @@ PasswordCallback (
...
+  @param[in]  PemSize   Size of the PEM key data in bytes.
+  @param[in]  Password  NULL-terminated passphrase used for encrypted PEM key data.
+  @param[out] Pkey      Pointer to receive the EVP_PKEY structure containing the private key.
+
+  @retval TRUE   Private key was retrieved successfully.
+  @retval FALSE  Failed to create BIO, write data, or read private key.
+**/
+STATIC
+BOOLEAN
+GetPrivateKeyFromPemBios (
```
liyi77: GetPrivateKeyFromPem (
The input is PEM data instead of BIO.

CryptoPkg/Library/BaseCryptLib/Pk/CryptEdDsa.c
```diff
@@ -0,0 +1,712 @@
...
+  }
+
+  Result = EVP_PKEY_keygen_init (PkeyCtx);
+  if (Result <= 0) {
+    EVP_PKEY_CTX_free (PkeyCtx);
+    return NULL;
+  }
+
+  Pkey   = NULL;
+  Result = EVP_PKEY_keygen (PkeyCtx, &Pkey);
```
liyi77: EVP_PKEY_keygen will generate a random key pair and put it to the CTX, but in fact we only need an empty CTX here.

As I recall OpenSSL currently might not allow the creation of an empty CTX containing the EdDSA flag - correct? then I think we can put EdDSA flag to KEY_CONTEXT, such as add a new item:
typedef struct {
  INT32           OpensslNid;
  EVP_PKEY    *EvpPkey;
} KEY_CONTEXT;

In EdDsaNewByNid we only check whether it is a valid NID and put it to Nid, leave EvpPkey to NULL.
In SetKey we then enter the generated Pkey as needed.

CryptoPkg/Private/Protocol/Crypto.h
```diff
@@ -6130,6 +6439,17 @@ struct _EDKII_CRYPTO_PROTOCOL {
   EDKII_CRYPTO_EC_GET_PRIVATE_KEY_FROM_PEM            EcGetPrivateKeyFromPem;
   EDKII_CRYPTO_EC_DSA_SIGN                            EcDsaSign;
   EDKII_CRYPTO_EC_DSA_VERIFY                          EcDsaVerify;
+  // Ed-DSA
+  EDKII_CRYPTO_ED_DSA_NEW_BY_NID                      EdDsaNewByNid;
```
liyi77: 1. Always add new items to then end of protocol structure.
2. Update EDKII_CRYPTO_VERSION after protocol structure updating.

CryptoPkg/Library/BaseCryptLib/Pem/CryptPem.c
mhollan1: Remove Bios from GetPrivateKeyFromPemBios function name

CryptoPkg/Library/BaseCryptLib/Pk/CryptEdDsa.c
```diff
@@ -0,0 +1,712 @@
...
+
+  @retval TRUE   EdDSA signature verification succeeded.
+  @retval FALSE  EdDSA signature verification failed or invalid parameters.
+
+**/
+BOOLEAN
+EFIAPI
+EdDsaVerify (
+  IN     VOID         *EdDsaContext,
+  IN     INT32        HashNid,
```
liyi77: EdDsa didnt need a HashNid

CryptoPkg/Include/Library/BaseCryptLib.h
```diff
@@ -4703,7 +4706,6 @@ EcDsaSign (
   @param[in]  HashNid      hash NID
   @param[in]  MessageHash  Pointer to octet message hash to be checked.
   @param[in]  HashSize     Size of the message hash in bytes.
-  @param[in]  Signature    Pointer to EC-DSA signature to be verified.
```
liyi77, line 4706: should not change it

CryptoPkg/Library/BaseCryptLib/Pk/CryptEdDsa.c
```diff
@@ -0,0 +1,605 @@
...
+
+  if (!OpensslNidToKeySize (Ctx->Nid, &FinalPrivateKeySize)) {
+    return FALSE;
+  }
+
+  if (FinalPrivateKeySize != PrivateKeySize) {
+    return FALSE;
+  }
+
+  Ctx->EvpPkey = EVP_PKEY_new_raw_private_key (Ctx->Nid, NULL, PrivateKey, PrivateKeySize);
```
liyi77, line 218: Potential memory leak issue, check Ctx->EvpPkey and free it if not a null pointer.

CryptoPkg/Library/OpensslLib/OpensslStub/uefiprov.c
```diff
@@ -23,58 +23,73 @@
 #include "internal/nelem.h"
 #include "provider_local.h"
 
-OSSL_provider_init_fn ossl_uefi_provider_init;
-const OSSL_PROVIDER_INFO ossl_predefined_providers[] = {
-    { "default", NULL, ossl_uefi_provider_init, NULL, 1 },
-    { NULL, NULL, NULL, NULL, 0 }
+OSSL_provider_init_fn     ossl_uefi_provider_init;
+const OSSL_PROVIDER_INFO  ossl_predefined_providers[] = {
```
liyi77: should not change this file

CryptoPkg/Library/BaseCryptLib/Pk/CryptEdDsa.c
mhollan1: Should it be freed in the function or is it the caller's responsibility to free it?

liyi77: I think both.
A KeyContext may be re-used with different keys, then it is the responsibility of the SetKey function to check and free the original Pkey.
When the user no longer needs the KeyContext, it is their responsibility to free the context using the Free API.

mhollan1: Ok understood

CryptoPkg/Library/OpensslLib/OpensslStub/uefiprov.c
mhollan1: I copied the original file and made the updates.  Uncrustify keeps updating this file.  it should be back to the original.

liyi77 (approved):
LGTM. Lets wait few days to see if any feedback from other maintainer

PR #11993, opened 2026-01-14
MM variable PEI with v3

kuqin12:
# Description

The current variable service through MM communicate PPI v3 could run into random behavior.

The reason was that the input module would reserve the header in the size of v1 communicate structure, which will cause the payload to corrupt the already populated MM communicate v3 header.

This change refactored the MM communication invoking routine from MM variable PPI, by separating the header population and communication into 2 individual functions. The PPI will be located once in the init call and passed on to the communication routine in the same transaction.

Lastly, as the Microsoft representative for Trust Execution Environment (TEE), I intend to help review Standalone MM related changes and help drive industry alignment of TEE implementations in UEFI firmware.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested by changing STMM OVMF to use MmVariablePei module and verified the memory type info variable could be properly read in the second boot and NextVariableName could iterate through all variables, through either v1 or v3.

## Integration Instructions

N/A

### DIFF
--- a/StandaloneMmPkg/Core/Handle.c
+++ b/StandaloneMmPkg/Core/Handle.c
@@ -233,9 +233,15 @@ MmInstallProtocolInterfaceNotify (
   }
 
   //
-  // Print debug message
+  // Print debug message unless installing MM entry/exit notify protocol as that spams the log
   //
-  DEBUG ((DEBUG_LOAD | DEBUG_INFO, "MmInstallProtocolInterface: %g %p\n", Protocol, Interface));
+  if (!CompareGuid (Protocol, &gEfiMmEntryNotifyProtocolGuid) &&
+      !CompareGuid (Protocol, &gEfiMmExitNotifyProtocolGuid))
+  {
+    DEBUG ((DEBUG_LOAD | DEBUG_INFO, "MmInstallProtocolInterface: Installing non-entry/exit protocol %g %p\n", Protocol, Interface));
+  } else {
+    DEBUG ((DEBUG_VERBOSE, "MmInstallProtocolInterface: Installing entry/exit protocol %g %p\n", Protocol, Interface));
+  }
 
   Status = EFI_OUT_OF_RESOURCES;
   Prot   = NULL;
--- a/StandaloneMmPkg/Core/StandaloneMmCore.c
+++ b/StandaloneMmPkg/Core/StandaloneMmCore.c
@@ -512,8 +512,6 @@ MmEntryPoint (
   UINTN                         CommGuidOffset;
   UINTN                         CommHeaderSize;
 
-  DEBUG ((DEBUG_INFO, "MmEntryPoint ...\n"));
-
   //
   // Update MMST using the context
   //
@@ -642,8 +640,6 @@ MmEntryPoint (
   //
   // TBD: Do not use private data structure ?
   //
-
-  DEBUG ((DEBUG_INFO, "MmEntryPoint Done\n"));
 }
 
 /** Register the MM Entry Point provided by the MM Core with the


MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -98,77 +113,123 @@ PopulateHeaderAndCommunicate (
...
+    MmCommunicateHeaderV3->BufferSize = RequiredPages * EFI_PAGE_SIZE;
+    CopyGuid (&MmCommunicateHeaderV3->MessageGuid, &gEfiSmmVariableProtocolGuid);
+    MmCommunicateHeaderV3->MessageSize = DataSize + SMM_VARIABLE_COMMUNICATE_HEADER_SIZE;
+    MmVariableFunctionHeader           = (SMM_VARIABLE_COMMUNICATE_HEADER *)MmCommunicateHeaderV3->MessageData;
+    if (MmCommunicationPpiV3Stub != NULL) {
+      *MmCommunicationPpiV3Stub = MmCommunicationPpiV3;
+    } else {
+      DEBUG ((DEBUG_ERROR, "%a: System supports MM Communication v3 PPI, but caller does not want it...\n", __func__));
+      ASSERT (FALSE);
+      return EFI_UNSUPPORTED;
```
samimujawar: Free MmCommunicateBuffer allocated at line 122 ?

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -98,77 +113,123 @@ PopulateHeaderAndCommunicate (
...
+    MmCommunicateHeader = (EFI_MM_COMMUNICATE_HEADER *)MmCommunicateBuffer;
+    CopyGuid (&MmCommunicateHeader->HeaderGuid, &gEfiSmmVariableProtocolGuid);
+    MmCommunicateHeader->MessageLength = DataSize + SMM_VARIABLE_COMMUNICATE_HEADER_SIZE;
+    MmVariableFunctionHeader           = (SMM_VARIABLE_COMMUNICATE_HEADER *)MmCommunicateHeader->Data;
+    if (MmCommunicationPpiStub != NULL) {
+      *MmCommunicationPpiStub = MmCommunicationPpi;
+    } else {
+      DEBUG ((DEBUG_ERROR, "%a: System supports only MM Communication v1 PPI, but caller does not want it...\n", __func__));
+      ASSERT (FALSE);
+      return EFI_UNSUPPORTED;
```
samimujawar: Free MmCommunicateBuffer allocated at line 146?

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
kuqin12: This is fixed, thanks!

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
kuqin12: Updated. Thanks!

kuqin12:
@lgao4 could you please review the change for MM variable PEI module? Thanks.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -98,77 +113,125 @@ PopulateHeaderAndCommunicate (
...
-
-    // Use gEfiSmmVariableProtocolGuid to request the MM variable service in Standalone MM
-    CopyMem ((VOID *)&MmCommunicateHeaderV3->HeaderGuid, &gEfiMmCommunicateHeaderV3Guid, sizeof (GUID));
-    MmCommunicateHeaderV3->BufferSize = CommunicateBufferSize;
-
-    CopyMem ((VOID *)&MmCommunicateHeaderV3->MessageGuid, &gEfiSmmVariableProtocolGuid, sizeof (GUID));
+    RequiredPages       = EFI_SIZE_TO_PAGES (DataSize + SMM_COMMUNICATE_HEADER_SIZE_V3 + SMM_VARIABLE_COMMUNICATE_HEADER_SIZE);
+    MmCommunicateBuffer = (UINT8 *)AllocatePages (RequiredPages);
+    if (MmCommunicateBuffer == NULL) {
+      DEBUG ((DEBUG_ERROR, "%a: Failed to allocate memory: %r\n", __func__, Status));
```
lgao4: Seemly, here Status is always EFI_SUCCESS.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -98,77 +113,125 @@ PopulateHeaderAndCommunicate (
...
+      ASSERT (FALSE);
+      FreePages (MmCommunicateBuffer, RequiredPages);
+      return EFI_UNSUPPORTED;
+    }
+  } else {
+    // Use v1 communication header, if v3 protocol is not available.
+    RequiredPages       = EFI_SIZE_TO_PAGES (DataSize + SMM_COMMUNICATE_HEADER_SIZE + SMM_VARIABLE_COMMUNICATE_HEADER_SIZE);
+    MmCommunicateBuffer = (UINT8 *)AllocatePages (RequiredPages);
+    if (MmCommunicateBuffer == NULL) {
+      DEBUG ((DEBUG_ERROR, "%a: Failed to allocate memory: %r\n", __func__, Status));
```
lgao4: Seemly, here Status is always EFI_SUCCESS.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -230,27 +295,23 @@ PeiMmGetVariable (
...
-
-  if (MmCommunicateBuffer == NULL) {
-    Status = EFI_OUT_OF_RESOURCES;
-    DEBUG ((DEBUG_ERROR, "%a: Failed to allocate memory: %r\n", __func__, Status));
-    return Status;
+  Status = InitCommunicateBuffer (
+             (VOID **)&MmVarAccessHeader,
+             MessageSize,
+             SMM_VARIABLE_FUNCTION_GET_VARIABLE,
+             (VOID **)&MmCommunicateBuffer,
```
lgao4, line 442: MmCommunicateBuffer should be set to NULL as its initialized value before call InitCommunicateBuffer.On current logic, if InitCommunicateBuffer returns erros,  it will go to Exit and Free MmCommunicateBuffer. 

There are several similar issue. I suggest to update InitCommunicateBuffer () to set *DataBuffer = NULL in the begin of it.

kuqin12: Great catch. Thank you. Will update shortly.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
kuqin12: Thanks, the Status is not relevant here. It is removed.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
kuqin12: Correct. It is now removed.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
kuqin12: This is updated.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
```diff
@@ -189,7 +189,7 @@ Communicate3 (
     //
     // Check if the HeaderGuid is valid
     //
-    if (CompareGuid (&CommunicateHeader->HeaderGuid, &gEfiMmCommunicateHeaderV3Guid)) {
+    if (!CompareGuid (&CommunicateHeader->HeaderGuid, &gEfiMmCommunicateHeaderV3Guid)) {
```
leiflindholm: While the fix is clearly correct, I feel "== 0" is more logically clear for a comparison return value.

Also, with my slightly rubbish eyesight it's a lot easier to spot :)

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -45,51 +43,70 @@ PeiMmVariableInitialize (
...
+  @retval EFI_INVALID_PARAMETER     The data size is too big.
+  @retval EFI_SUCCESS               Find the specified variable.
 
 **/
 EFI_STATUS
-PopulateHeaderAndCommunicate (
-  IN OUT  UINT8  *CommunicateBuffer,
-  IN UINTN       CommunicateBufferSize,
-  IN UINTN       Function
+InitCommunicateBuffer (
```
leiflindholm, line 208: This turns a too-big 100-line function into a 150-line function. Can we break it up a bit with some helpers please?

Maintainers.txt
```diff
@@ -647,6 +647,7 @@ F: StandaloneMmPkg/
 M: Ard Biesheuvel <ardb+tianocore@kernel.org> [ardbiesheuvel]
 M: Sami Mujawar <sami.mujawar@arm.com> [samimujawar]
 M: Ray Ni <ray.ni@intel.com> [niruiyu]
+M: Kun Qin <kun.qin@microsoft.com> [kuqin12]
 R: Jiaxin Wu <jiaxin.wu@intel.com> [jiaxinwu]
```
leiflindholm, line 652: Very pleased with this, thank you!

However, it makes it look a little weird that the email address here and in the S-o-b is different from the previous two commits.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
kuqin12: Thank you! Good call. This is updated.

Maintainers.txt
kuqin12: This is likely some legacy from the mailing list ear... The commit messages are fixed to be consistent across the PR.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
kuqin12: This function has been refactored into three smaller functions: one that follows the v3 PPI flow, one that follows the v1 PPI flow, and this function, which now orchestrates between them.

leiflindholm:
Much oblliged for the refactoring.

Happy with the code.

For the PR, the "breaking change" label is set, but there are no integration instructions.
Was the flag set by accident?

lgao4:
@kuqin12 , I have no other comment. Please check CI failure first.

kuqin12:
> @kuqin12 , I have no other comment. Please check CI failure first.

Thanks @lgao4, seems like the failure was due to a failure to download the uncrustify file. Can we please requeue this failed build?

kuqin12:
> Much oblliged for the refactoring.
> 
> Happy with the code.
> 
Thank you for the review!
>
> For the PR, the "breaking change" label is set, but there are no integration instructions. Was the flag set by accident?
>
This might be added when I marked it incorrectly in the description during PR creation time. But the labelling action did not follow the to remove the tag later.

You are correct that this is not a breaking change.

ardbiesheuvel:
ok for soft freeze

ardbiesheuvel:
ok for soft freeze

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -45,130 +43,285 @@ PeiMmVariableInitialize (
...
+  SMM_VARIABLE_COMMUNICATE_HEADER  *MmVariableFunctionHeader;
 
-    // Send the MM request using MmCommunicationPei
-    Status = MmCommunicationPpi->Communicate (MmCommunicationPpi, CommunicateBuffer, &CommunicateBufferSize);
-  } else {
-    // We should never reach here, but just in case
-    Status = EFI_UNSUPPORTED;
-    DEBUG ((DEBUG_ERROR, "%a: No MM Communication PPI found!\n", __func__));
+  // Should not both stubs be NULL
+  if ((MmCommunicationPpiStub == NULL) || (MmCommunicationPpiV3Stub == NULL)) {
```
makubacki, line 222:
```suggestion
  if ((MmCommunicationPpiStub == NULL) && (MmCommunicationPpiV3Stub == NULL)) {
```

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -45,130 +43,285 @@ PeiMmVariableInitialize (
...
-    // We received an error from Variable Service.
-    // We cant do anymore so return Status
-    if (Status != EFI_BUFFER_TOO_SMALL) {
-      DEBUG ((DEBUG_ERROR, "%a - Variable Service Error: %r\n", __func__, Status));
+  *DataBuffer               = NULL;
+  *MmCommunicationPpiStub   = NULL;
+  *MmCommunicationPpiV3Stub = NULL;
+
+  // Attempt to locate PEI MM Communication3 PPI first
+  Status = InternalInitCommBufferV3 (
```
makubacki, line 249: Was there a reason to exclude `InternalInitCommBufferV3` from the null checks? It is currently not validated all the way to dereference in `InternalInitCommBufferV3 ()`.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -230,27 +385,23 @@ PeiMmGetVariable (
...
+             MessageSize,
+             SMM_VARIABLE_FUNCTION_GET_VARIABLE,
+             (VOID **)&MmCommunicateBuffer,
+             &RequiredPages,
+             &MmCommunicationPpi,
+             &MmCommunicationPpiV3
+             );
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "%a: InitCommunicateBuffer failed: %r\n", __func__, Status));
+    goto Exit;
```
makubacki, line 449: The `Exit` label sets some output parameters to this function based on `MmVarAccessHeader` values. I don't think you can trust those values in this failure case.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -45,130 +43,285 @@ PeiMmVariableInitialize (
...
-    ZeroMem (MmCommunicateHeaderV3, SMM_COMMUNICATE_HEADER_SIZE_V3);
+  if (DataBuffer == NULL) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  *DataBuffer = NULL;
+
+  Status = PeiServicesLocatePpi (&gEfiPeiMmCommunicationPpiGuid, 0, NULL, (VOID **)MmCommunicationPpiV1Stub);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_WARN, "%a: Unable to locate PEI MM Communication3 PPI: %r\n", __func__, Status));
```
makubacki: This is `InternalInitCommBufferV1()`.


```suggestion
    DEBUG ((DEBUG_WARN, "%a: Unable to locate PEI MM Communication PPI: %r\n", __func__, Status));
```

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -230,27 +385,23 @@ PeiMmGetVariable (
   }
 
   // Allocate required pages to send MM request
-  MessageSize = MM_VARIABLE_COMM_BUFFER_OFFSET +
-                OFFSET_OF (SMM_VARIABLE_COMMUNICATE_ACCESS_VARIABLE, Name) +
+  MessageSize = OFFSET_OF (SMM_VARIABLE_COMMUNICATE_ACCESS_VARIABLE, Name) +
                 StrSize (VariableName) + *DataSize;
```
makubacki: You should consider checking `DataSize` to prevent overflow in calculating the `MessageSize` here.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
kuqin12: This is an internal function. As the point here is to prevent misusage, I would not want the input parameters to be partially NULL.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
kuqin12: Thanks. I added more checks for the input pointers.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
kuqin12: Good call. This is updated to only read those fields when the status code is success or buffer too small.

kuqin12: and the same for get next variable function.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
kuqin12: Thanks, this is updated/

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
kuqin12: Indeed. This is updated to perform unitn addition through the safe int lib.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
makubacki: In that case, the comment needs to be updated to something like "// Either stub should not be null".

The confusing part right now is "both" means "all" which is an "AND" Boolean expression. "Partial" means "either" and is an "OR" Boolean expression.

kuqin12: Yes, the updated comment say `// Neither stub shall be NULL`. Thanks for the suggestion.

makubacki (approved):

kuqin12:
Pushed an empty amend to force trigger the build. Did not change anything in this push.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -276,20 +474,22 @@
...
+  if (IncomingDataSize < MmVarAccessHeader->DataSize) {
     Status = EFI_BUFFER_TOO_SMALL;
   }
 
 Exit:
-  // Check if we need to set Attributes
-  if (Attributes != NULL) {
-    *Attributes = MmVarAccessHeader->Attributes;
+  if ((Status == EFI_SUCCESS) || (Status == EFI_BUFFER_TOO_SMALL)) {
+    *DataSize = MmVarAccessHeader->DataSize;
```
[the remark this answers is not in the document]
kuqin12: This was checked at the entry of this function and branch here with a `EFI_INVALID_PARAMETER` error code. It would not end up in this case.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -276,20 +474,22 @@
...
   }
 
-  *DataSize = MmVarAccessHeader->DataSize;
-
   if (Status == EFI_SUCCESS) {
-    CopyMem ((VOID *)Data, (UINT8 *)MmVarAccessHeader->Name + MmVarAccessHeader->NameSize, *DataSize);
+    // Check if we need to set Attributes
+    if (Attributes != NULL) {
+      *Attributes = MmVarAccessHeader->Attributes;
```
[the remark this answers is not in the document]
kuqin12: `MmVarAccessHeader` is part of the allocated pages during the call. With the condition making sure this is a successful call, the `MmVarAccessHeader` will not be a null pointer.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -276,20 +474,22 @@
...
+    }
+
+    CopyMem ((VOID *)Data, (UINT8 *)MmVarAccessHeader->Name + MmVarAccessHeader->NameSize, MmVarAccessHeader->DataSize);
```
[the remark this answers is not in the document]
kuqin12: Same as above.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -403,8 +629,10 @@
   }
 
 Exit:
-  // Update the name size to be returned
-  *VariableNameSize = MmVarGetNextVarHeader->NameSize;
+  if ((Status == EFI_BUFFER_TOO_SMALL) || (Status == EFI_SUCCESS)) {
+    // Update the name size to be returned
+    *VariableNameSize = MmVarGetNextVarHeader->NameSize;
```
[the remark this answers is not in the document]
kuqin12: This was checked at the entry of this function and branched here with a "EFI_INVALID_PARAMETER". This case would not trigger.

PR #12006, opened 2026-01-17
RedfishPkg/RedfishPlatformConfig: Fix CLANGPDB build failure

changab:
# Description

CLANGPDB build fails because the intrinsic _memcpy is used by the compiler for the value structure passed via a function. Change the code to use the value reference.

- [X ] Breaking change?
  - **Breaking change** - Interface parameter change, but no impact on the existing driver. 
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
No

## Integration Instructions
N/A


nicklela (commented):
Hi @changab 
Since we change the protocol interface, we need to change protocol "Revision" for the backward compatibility.

changab:
> Hi @changab Since we change the protocol interface, we need to change protocol "Revision" for the backward compatibility.

Thanks, @nicklela. I resent the PR.

apop5 (approved):

apop5:
Now that it's a reference, can you verify that the passed reference is not null?

changab:
> Now that it's a reference, can you verify that the passed reference is not null?

Done!

mdkinney:
This is a non-backwards compatible change to an interface.

Is a small bump in the revision sufficient?  Can or are these interfaces consumed by code outside the edk2 repo?  Those external uses would fail to compile because of the interface change.

Also, since it is a interface change, I think the "Breaking Change" checkbox should be checked.

changab:
> This is a non-backwards compatible change to an interface.
> 
> Is a small bump in the revision sufficient? Can or are these interfaces consumed by code outside the edk2 repo? Those external uses would fail to compile because of the interface change.
> 
> Also, since it is a interface change, I think the "Breaking Change" checkbox should be checked.

@mdkinney , I don't think this would cause a backward compatibility issue. The only consumer is the library under Redfish client package, and it acts as an abstract layer for all of high-level Redfish feature drivers.
I updated the PR descriptor.

mdkinney:
> > This is a non-backwards compatible change to an interface.
> > Is a small bump in the revision sufficient? Can or are these interfaces consumed by code outside the edk2 repo? Those external uses would fail to compile because of the interface change.
> > Also, since it is a interface change, I think the "Breaking Change" checkbox should be checked.
> 
> @mdkinney , I don't think this would cause a backward compatibility issue. The only consumer is the library under Redfish client package, and it acts as an abstract layer for all of high-level Redfish feature drivers. I updated the PR descriptor.

Does this mean that changes to https://github.com/tianocore/edk2-redfish-client are also required?  Any chance that products have integrated the features from that repo that will break when run with edk2 platforms that have the new version of the Protocol? The reason I ask is that there are other methods that can be used to handle changes like these such as introducing a new version of the protocol with a different GUID.

mikebeaton:
Is there no way to update the intrinsics support, and avoid changing the interface?

mdkinney:
The only place intrinsics are used are in submodules that edk2 consumes "as is".  All other components must be implemented to not introduce the use of intrinsics. That is why the EDK II C Coding Standard Specification requires no pass by value structure parameters.

changab:
> > > This is a non-backwards compatible change to an interface.
> > > Is a small bump in the revision sufficient? Can or are these interfaces consumed by code outside the edk2 repo? Those external uses would fail to compile because of the interface change.
> > > Also, since it is a interface change, I think the "Breaking Change" checkbox should be checked.
> > 
> > 
> > @mdkinney , I don't think this would cause a backward compatibility issue. The only consumer is the library under Redfish client package, and it acts as an abstract layer for all of high-level Redfish feature drivers. I updated the PR descriptor.
> 
> Does this mean that changes to https://github.com/tianocore/edk2-redfish-client are also required? Any chance that products have integrated the features from that repo that will break when run with edk2 platforms that have the new version of the Protocol? The reason I ask is that there are other methods that can be used to handle changes like these such as introducing a new version of the protocol with a different GUID.

Yes, we already prepared the PR for edk2-redfish-client. We can merge edk2-redfish-client PR once this edk2 PR gets merged. 
The key is, I don't think there is any direct reference to this change. The only driver that invokes (not invoking directly) this interface is the BIOS Redfish feature driver under edk2-redfish-client via an abstract library. We don't have the use case under tianocore edk2-platform that has references with edk2 RedfishclinetPkg either.  We can keep the solution simple.

igorkulchytskyy (approved):
Sorry for the delay. I approved it

mdkinney:
@AbnerChang Is this ready for push label now?

changab:
> @AbnerChang Is this ready for push label now?

Yes, Push label is added.

PR #12007, opened 2026-01-17
RedfishPkg: Fix missing dependency when interigating with EmulatorPkg

khaliid2040:
# Description

When Redfish is enabled on EmulatorPkg it fails to build due missing instances of RedfishPlatformWantedDeviceLib, IpmiLib and RedfishContentCodingLib libraries. Add those library instances to DSC include file so EmulatorPkg could detect it.

Also, add gEfiRedfishClientPkgTokenSpaceGuid to DEC file of RedfishPkg which EmulatorPkg depends on.

Those two solutions fix following errors when building EmulatorPkg with Redfish:

Note: this below block contains multiple build fails with different library instances
```
build.py...
/home/khaalid/edk2/EmulatorPkg/EmulatorPkg.dsc(...): error 4000: Instance of library class [RedfishPlatformWantedDeviceLib] is not found for module [/home/khaalid/edk2/RedfishPkg/RedfishRestExDxe/RedfishRestExDxe.inf], [RedfishPlatformWantedDeviceLib] is:
        consumed by /home/khaalid/edk2/RedfishPkg/RedfishRestExDxe/RedfishRestExDxe.inf

build.py...
/home/khaalid/edk2/EmulatorPkg/EmulatorPkg.dsc(...): error 4000: Instance of library class [IpmiLib] is not found for module [/home/khaalid/edk2/RedfishPkg/RedfishCredentialDxe/RedfishCredentialDxe.inf], [IpmiLib] is:
        consumed by /home/khaalid/edk2/RedfishPkg/Library/RedfishPlatformCredentialIpmiLib/RedfishPlatformCredentialIpmiLib.inf
 
build.py...
/home/khaalid/edk2/EmulatorPkg/EmulatorPkg.dsc(...): error 4000: Instance of library class [RedfishContentCodingLib] is not found for module [/home/khaalid/edk2/RedfishPkg/RedfishHttpDxe/RedfishHttpDxe.inf], [RedfishContentCodingLib] is:
        consumed by /home/khaalid/edk2/RedfishPkg/RedfishHttpDxe/RedfishHttpDxe.inf

```

When the PCD is missing it fails with following:
```
build.py...
/home/khaalid/edk2/EmulatorPkg/EmulatorPkg.dsc(300): error F001: Pcd (gEfiRedfishClientPkgTokenSpaceGuid.PcdRedfishServiceEtagSupported) defined in DSC is not declared in DEC files referenced in INF files in FDF. Arch: ['X64']
```

- [ ] Breaking change?
  - NO
- [ ] Impacts security?
  - NO
- [ ] Includes tests?
  - NO

## How This Was Tested
```
build -a X64 -p EmulatorPkg/EmulatorPkg.dsc -t GCC5 -b DEBUG -DREDFISH_ENABLE=1
```
## Integration Instructions
N/A

### DIFF
--- a/RedfishPkg/RedfishPkg.dec
+++ b/RedfishPkg/RedfishPkg.dec
@@ -128,6 +128,7 @@
     MdeModulePkg/MdeModulePkg.dec
     RedfishPkg/RedfishPkg.dec
   }
+
   #
   # This PCD indicates the EFI REST EX access mode to Redfish service.
   # Default is set to out of band access.
--- a/RedfishPkg/RedfishLibs.dsc.inc
+++ b/RedfishPkg/RedfishLibs.dsc.inc
@@ -22,5 +22,8 @@
   HiiUtilityLib|RedfishPkg/Library/HiiUtilityLib/HiiUtilityLib.inf
   RedfishPlatformConfigLib|RedfishPkg/Library/RedfishPlatformConfigLib/RedfishPlatformConfigLib.inf
   RedfishHttpLib|RedfishPkg/Library/RedfishHttpLib/RedfishHttpLib.inf
+  RedfishPlatformWantedDeviceLib|RedfishPkg/Library/RedfishPlatformWantedDeviceLibNull/RedfishPlatformWantedDeviceLibNull.inf
+  IpmiLib|MdeModulePkg/Library/BaseIpmiLibNull/BaseIpmiLibNull.inf
+  RedfishContentCodingLib|RedfishPkg/Library/RedfishContentCodingLibNull/RedfishContentCodingLibNull.inf
 !endif
 


RedfishPkg/RedfishPkg.dec
```diff
@@ -128,6 +128,11 @@
     MdeModulePkg/MdeModulePkg.dec
     RedfishPkg/RedfishPkg.dec
   }
+
+  #
+  # This PCD indicates whether support of whether to enable Etag or not
+  #
+  gEfiRedfishClientPkgTokenSpaceGuid.PcdRedfishServiceEtagSupported|FALSE|BOOLEAN|0x0001
```
nicklela: This PCD is defined in RedfishClientPkg.dec here: https://github.com/tianocore/edk2-redfish-client/blob/0c3c6f7800b1576cd53eb623270a17be58ba0ddd/RedfishClientPkg/RedfishClientPkg.dec#L77

Should we include RedfishClientPkg in PACKAGES_PATH while building EmulatorPkg?

khaliid2040: > Should we include RedfishClientPkg in PACKAGES_PATH while building EmulatorPkg?

edk2-redfish-client is external reference and it requires some setup, which could specially difficult for people new to edk2 build infrastructure.

Also, by default PACKAGES_PATH is not set, it would have been easier if that env variable was set by `edksetup.sh` and we just append to it.

So this will introduce some complexity. This change the PR includes seems simpler.

Do you have any simpler way we could achieve your suggestion?

If not then i will copy the PCD from that file, since this one the value is incorrect.

changab: We shouldn't create a dependency with edk2-redfish-client from RedfishPkg, as  edk2-redfish-client is designed as the Redfish client, which is an application based on RedfishPkg. Having gEfiRedfishClientPkgTokenSpaceGuid in RedfishPkg also leads to a build failure with -p RedfishPkg.dsc assigned to edk2 build.  We can have gEfiRedfishClientPkgTokenSpaceGuid in EmulatorPKg when Redfish build is enabled.  Let EmulatorPKg pull in whatever it needs from edk2-redfish-client.

khaliid2040: > We shouldn't create a dependency with edk2-redfish-client from RedfishPkg, as edk2-redfish-client is designed as the Redfish client, which is an application based on RedfishPkg. 

I don't think we need that same dependency in EmulatorPkg, but i agree we can move gEfiRedfishClientPkgTokenSpaceGuid to EmulatorPkg.

> Having gEfiRedfishClientPkgTokenSpaceGuid in RedfishPkg also leads to a build failure with -p RedfishPkg.dsc assigned to edk2 build. We can have gEfiRedfishClientPkgTokenSpaceGuid in EmulatorPKg when Redfish build is enabled. Let EmulatorPKg pull in whatever it needs from edk2-redfish-client.

I didn't get any build failure when i added that PCD to the DEC file of RedfishPkg. I tested both on DEBUG and RELEASE with these changes.

mdkinney: This change does not make sense.  I do not see any references to that PCD in the edk2 repo.  Should not be added to DEC and the reference from DSC should be removed.

Any platforms in edk2 repo should only use content from the edk2 repo.  If there are Redfish features that require content from other packages to test, then another version of the EmulatorPkg build that uses those features can be added to those other repos.

khaliid2040: Yes, we should remove the reference from DSC file i submitted the changes. I made two commits for the two packages.

YangGangUEFI:
I've been researching Redfish recently and I just created a PR #12010 (and then saw this).

The difference is that i just modified `EmulatorPkg.dsc`:

1. Move `!include RedfishPkg/Redfish.dsc.inc` to the top of the .dsc file.
2. Add the missing libraries to the .dsc file.
3. Remove `gEfiRedfishClientPkgTokenSpaceGuid.PcdRedfishServiceEtagSupported`

I think 1. is important, as it can prevent specific libraries of EmulatorPkg from being override by `RedfishPkg/Redfish.dsc.inc`.

https://github.com/tianocore/edk2/blob/f82d1fa48eeae62581383c5f341bf9549e370e6c/EmulatorPkg/EmulatorPkg.dsc#L112-L115

khaliid2040:
> I've been researching Redfish recently and I just created a PR #12010 (and then saw this).
> 
> The difference is that i just modified `EmulatorPkg.dsc`:
> 
>     1. Move `!include RedfishPkg/Redfish.dsc.inc` to the top of the .dsc file.
> 
>     2. Add the missing libraries to the .dsc file.
> 
>     3. Remove `gEfiRedfishClientPkgTokenSpaceGuid.PcdRedfishServiceEtagSupported`
> 
> 
> I think 1. is important, as it can prevent specific libraries of EmulatorPkg from being override by `RedfishPkg/Redfish.dsc.inc`.
> 
> https://github.com/tianocore/edk2/blob/f82d1fa48eeae62581383c5f341bf9549e370e6c/EmulatorPkg/EmulatorPkg.dsc#L112-L115

There is no overrides, and we shouldn't put Redfish libraries into EmulatorPkg DSC file. But, i like that small refactor.
Let's work on this one which already have some review.

khaliid2040:
I moved EmulatorPkg changes to another to #12016  that fixes other issues with EmulatorPkg too. So this one should only involve with RedfishPkg.

YangGangUEFI:
> There is no overrides, and we shouldn't put Redfish libraries into EmulatorPkg DSC file. 

No, in your previous commit, `RedfishPlatformCredentialLib|EmulatorPkg/Library/RedfishPlatformCredentialLib/RedfishPlatformCredentialLib.inf` will be replaced by `RedfishPlatformCredentialLib|RedfishPkg/Library/RedfishPlatformCredentialIpmiLib/RedfishPlatformCredentialIpmiLib.inf` (What I mean by "override" is this.)

On the other hand, I'm **unsure** if any users have added `!include RedfishPkg/Redfish.dsc.inc` to the `[Components]` section. If so, the changes in this PR might introduce some ***NULL*** library instances, causing users to experience issues.

I think #12010 is enough(to resolve the build issue). So let's wait and see what other information we can get or hear from the reviewers.

khaliid2040:
@YangGangUEFI, btw the changes your PR introduced and this PR are relatively small, and i am thinking if we combine them into one PR and close #12010 and this PR.

I found other issues too like missing library instances for networking. It is better if we fix in #12016. 

> I think https://github.com/tianocore/edk2/pull/12010 is enough(to resolve the build issue). So let's wait and see what other information we can get or hear from the reviewers.

I will adjust #12016 with your changes. This makes review easier instead of fixing some random missing library instances in multiple PRs.

PR #12008, opened 2026-01-19
MdeModulePkg/Logo: Use actual logo and typography

howl:
# Description

BMP following ACPI BGRT spec created from actual TianoCore SVG logo with default size and setting white letters over black background.

- [ ] Breaking change?
  - No
- [ ] Impacts security?
  - No
- [ ] Includes tests?
  - No

## How This Was Tested

Straight forward image change.

## Integration Instructions

N/A


mikebeaton:
Hi - You've updated to the logo from here https://github.com/tianocore/tianocore.github.io/tree/master/images, used here https://www.tianocore.org/. 

We can see that the logo+text SVG is 9 years old, and the actual logo (logo.svg) is 11 years old.

While the logo used on the homepage of tianocore.org _ought_ to be the current logo and branding, I am not sure that it is. E.g. the logo used on Github https://github.com/tianocore - even though it is terribly low res - looks more modern, and I would assume was updated much more recently.

However, I don't know who updates this or might have access to more recent logo sources.

howl:
@mikebeaton I took the one from the tiano web page, but where you point there is a directly a white letter one version. I suppouse the resulting bmp is going to be identical but anyway I'm going to redo with the one from you point.

I was thinking about using only the logo but using the logo plus letters is more similar to the one is now and also I think is better to recognize tianocore for people who don't know anything about when starting with vm for example.

The doubts I have is if it could be interesting to set up the letters down the logo both centred as seems to be the most common bgrt to see over different vendors.

I will just leave this here and if one who could decide over this as you say looks this then just accept or reject.

Actually I'm looking to compile ovmf fd image but haven't been successful, also I messed up the labels of the PR impact:breaking-change impact:security impact:testing :sweat_smile:

howl:
@mikebeaton just did it from the white letters svg and just the same result, both bmps have b3sum 4124055ac1720bdd010bf31560ec7413c089e1fb8b46e5b6f9f4efd1ca7da2e3 so don't need to force push the PR.

mikebeaton:
@howl - Hi again - My opinion is that although the logo is old, the logo you are replacing it with is almost certainly also too old to be worth replacing it with.

Source for a newer (and newer looking) logo and font should be found, if the update is worth doing at all. IMO.

I hesitate to cc busy people, but do active and helpful reviewers @ardbiesheuvel and @leiflindholm have any idea who maintains these things (e.g. where the current Tianocore Github logo came from), and who could help this author obtain a more up-to-date logo - if indeed the update is worth doing?

howl:
@mikebeaton sorry I didn't understand you right. As far as I know there is no more actual logo, despite being so much years old there hasn't been a new redesign, this is the one used in the official web page both the logo and the typography.

https://www.tianocore.org/

About the positioning here the coreboot logo for reference.

https://www.coreboot.org/assets/images/banner.svg

The bgrt images tend to have this format.

lgao4:
I don't think it is necessary to change the default logo file. The project can have its own logo.

howl:
> I don't think it is necessary to change the default logo file. The project can have its own logo.

Could be possible to set with transparent background, black as per ACPI spec?

I haven't been able to find svg or at least a high res of the ovmf logo.

mikebeaton:
> > I don't think it is necessary to change the default logo file. The project can have its own logo.
> 
> Could be possible to set with transparent background, black as per ACPI spec?
> 
> I haven't been able to find svg or at least a high res of the ovmf logo.

@lgao4 is saying he does not think the change is worth making. I said I think it's not worth making unless we have a newer logo to base off, which we don't. It is certainly Liming's decision to make, and not mine at all! Sorry. Thanks for the effort of making the PR.

PR #12009, opened 2026-01-19
BaseTools: Prevent Subsection PCDs from polluting global expressions

PaddyDengAmi:
# Description

The PCD value defined in module subsections can be added to global PCD database. Therefore the unsolved expressions, even belongs to the global scope, can incorrectly refer to the value from module subsection.

This only happens when the referred PCD has no value assignment in the platform dsc file. Which also should raise an error.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Use test code from https://github.com/PaddyDengAmi/edk2/commit/8673d4046673634fc747dc794c51b033e14e24ad
Run `test_build.bat` after setting up edk2 build environment.

Check the produced `build_report.log`, the value of `PcdIpmiKcsIoBaseAddress` in both HelloWorld.inf instances are polluted by `gEfiMdePkgTokenSpaceGuid.PcdIpmiSsifSmbusSlaveAddr|2` in the subsection.

Compares to `build_report_gPcd.log`, which produced by `TestPkg_gPcd.dsc` that has a proper global value of `PcdIpmiSsifSmbusSlaveAddr`, the subsection override of `PcdIpmiSsifSmbusSlaveAddr|2` only affects this PCD itself.
All value evaluated by `gEfiMdePkgTokenSpaceGuid.PcdIpmiKcsIoBaseAddress|gEfiMdePkgTokenSpaceGuid.PcdIpmiSsifSmbusSlaveAddr` are not affected and correctly refer to `PcdIpmiSsifSmbusSlaveAddr|1` in global scope.

After applying the change of this PR, the build for `TestPkg.dsc` will fail due to `gEfiMdePkgTokenSpaceGuid.PcdIpmiKcsIoBaseAddress|gEfiMdePkgTokenSpaceGuid.PcdIpmiSsifSmbusSlaveAddr` is referring `PcdIpmiSsifSmbusSlaveAddr` which is not given a default value in this dsc and raising an error:

```
D:\work\edk2\TestPkg.dsc(36): error 3000: The PCD should be FeatureFlag type or FixedAtBuild type: [gEfiMdePkgTokenSpaceGuid.PcdIpmiSsifSmbusSlaveAddr].
        PCD [gEfiMdePkgTokenSpaceGuid.PcdIpmiKcsIoBaseAddress] Value "gEfiMdePkgTokenSpaceGuid.PcdIpmiSsifSmbusSlaveAddr"
```

## Integration Instructions

N/A


lgao4:
Yes. This is a problem. Do you meet with the real issue?

PaddyDengAmi:
Hi @lgao4,

Yes, we found this on real project when trying to connecting 2 PCDs value and found the assigned value is not what we expected.

lgao4:
> Hi @lgao4,
> 
> Yes, we found this on real project when trying to connecting 2 PCDs value and found the assigned value is not what we expected.

So, you expect the build tool should report error for this case, and require the developer to update DSC by setting PCD value.

PaddyDengAmi:
> > Hi @lgao4,
> > Yes, we found this on real project when trying to connecting 2 PCDs value and found the assigned value is not what we expected.
> 
> So, you expect the build tool should report error for this case, and require the developer to update DSC by setting PCD value.

Hi @lgao4 ,

Yes, in my understanding, .dsc file does not bind to any .dec file. Only in the scope of .inf we can certain which .dec file the PCD is declared and know its default value. So we need to explicit assign a value in .dsc file for PCDs used in .dsc expressions.
But PCDs defined in driver sub-scope shouldn't be used in these cases, so we should still report a PCD value error.

PR #12012, opened 2026-01-19
 Use common String utilities

pierregondois:
# Description

This is a trimmed version of https://github.com/tianocore/edk2/pull/11839. Functions checking non-numbers related Unicode chars/strings have been removed.

---

The edk2 repository contains multiple local implementations of string utilities:

- checking a string/char uppercase/lowercase/alphanumerical/hexadecimal compliance
- replacing uppercase string/char to lowercase (or conversely)

Add string/char utilities to BaseLib and remove some of the local implementations.
Not all implementations have been replaced, but this is a first step.

### DIFF
--- a/MdeModulePkg/Universal/Disk/UnicodeCollation/EnglishDxe/UnicodeCollationEng.c
+++ b/MdeModulePkg/Universal/Disk/UnicodeCollation/EnglishDxe/UnicodeCollationEng.c
@@ -106,7 +106,7 @@ InitializeUnicodeCollationEng (
     mEngLowerMap[Index] = (CHAR8)Index;
     mEngInfoMap[Index]  = 0;
 
-    if (((Index >= 'a') && (Index <= 'z')) || ((Index >= 0xe0) && (Index <= 0xf6)) || ((Index >= 0xf8) && (Index <= 0xfe))) {
+    if (AsciiCharIsLowerAlpha ((CHAR8)Index) || ((Index >= 0xe0) && (Index <= 0xf6)) || ((Index >= 0xf8) && (Index <= 0xfe))) {
       Index2               = Index - 0x20;
       mEngUpperMap[Index]  = (CHAR8)Index2;
       mEngLowerMap[Index2] = (CHAR8)Index;
--- a/MdePkg/Include/Library/BaseLib.h
+++ b/MdePkg/Include/Library/BaseLib.h
@@ -2667,6 +2667,100 @@ AsciiCharToUpper (
   IN      CHAR8  Chr
   );
 
+/*
+ * Unicode String / Char checker functions.
+ */
+
+/**
+  Determine if a Unicode Char is a number.
+
+  @param[in] Char         Char to analyze.
+
+  @retval TRUE            Unicode Char is a number.
+  @retval FALSE           Otherwise.
+**/
+BOOLEAN
+EFIAPI
+CharIsNum (
+  IN CHAR16  Char
+  );
+
+/**
+  Determine if a Unicode Char is a hexadecimal number.
+
+  @param[in] Char         Char to analyze.
+
+  @retval TRUE            Unicode Char is a hexadecimal number.
+  @retval FALSE           Otherwise.
+**/
+BOOLEAN
+EFIAPI
+CharIsHexNum (
+  IN CHAR16  Char
+  );
+
+/**
+  Determine if a Unicode Char is an uppercase hexadecimal number.
+
+  @param[in] Char         Char to analyze.
+
+  @retval TRUE            Unicode Char is an uppercase hexadecimal number.
+  @retval FALSE           Otherwise.
+**/
+BOOLEAN
+EFIAPI
+CharIsUpperHexNum (
+  IN CHAR16  Char
+  );
+
+/**
+  Determine if a Unicode String has only numbers.
+
+  @param[in] String       Pointer to the string to analyze.
+  @param[in] MaxSize      Maximum number of characters to analyze.
+
+  @retval TRUE            Unicode String has only numbers.
+  @retval FALSE           Unicode String has at least one other character.
+**/
+BOOLEAN
+EFIAPI
+StrnIsNum (
+  IN CONST CHAR16  *String,
+  IN       UINTN   MaxSize
+  );
+
+/**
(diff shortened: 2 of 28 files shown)


MdePkg/Include/Library/BaseLib.h
```diff
@@ -2668,6 +2668,94 @@ AsciiCharToUpper (
   IN      CHAR8  Chr
   );
 
+/*
```
leiflindholm, line 2884: I think commit subject line needs updating now?

leiflindholm: Did github always fail to separate comments on a per-commit basis or is this some new feature to tire maintainers into letting copilot do the reviewing instead?

This comment applies to both:
MdePkg/BaseLib: Add Unicode string case/apha/num check functions 
and
MdePkg/BaseLib: Add Ascii string case/apha/num check functions

MdeModulePkg/Universal/Disk/UnicodeCollation/EnglishDxe/UnicodeCollationEng.c
```diff
@@ -106,7 +106,7 @@ InitializeUnicodeCollationEng (
     mEngLowerMap[Index] = (CHAR8)Index;
     mEngInfoMap[Index]  = 0;
 
-    if (((Index >= 'a') && (Index <= 'z')) || ((Index >= 0xe0) && (Index <= 0xf6)) || ((Index >= 0xf8) && (Index <= 0xfe))) {
+    if (AsciiCharIsLowerAlpha ((CHAR8)Index) || ((Index >= 0xe0) && (Index <= 0xf6)) || ((Index >= 0xf8) && (Index <= 0xfe))) {
```
leiflindholm, line 109: The original code is a bit of a horror show. As far as I can tell, it does some manual inline-extraction of some additional UCS-2 <256 code points (but leaves out LATIN SMALL LETTER SHARP S - 0xdf - and LATIN SMALL LETTER Y WITH DIAERESIS - 0xff).

The change arguably corrects an error in the original code through the explicit cast to CHAR8. But wouldn't it make more sense to change Index and Index2 to UINT8 (or UCHAR8) anyway, given MAP_TABLE_SIZE is 0x100 and all other parts of this function assume they're both always <256?

MdePkg/Library/BaseLib/String.c
```diff
@@ -567,7 +545,7 @@ InternalAsciiIsHexaDecimalDigitCharacter (
   IN      CHAR8  Char
   )
 {
-  return (BOOLEAN)(InternalAsciiIsDecimalDigitCharacter (Char) ||
+  return (BOOLEAN)(AsciiCharIsNum (Char) ||
```
leiflindholm: Can collapse all three tests into AsciiCharIsHexNum?

leiflindholm (changes requested):

MdePkg/Include/Library/BaseLib.h
```diff
@@ -2668,6 +2668,294 @@ AsciiCharToUpper (
...
+
+  @retval TRUE            Unicode String has only uppercase hexadecimal numbers.
+  @retval FALSE           Unicode String has at least one other character.
+**/
+BOOLEAN
+StrnIsUpperHexNum (
+  IN CONST CHAR16  *String,
+  IN       UINTN   MaxSize
+  );
+
```
lgao4, line 2977: Library Function needs EFIAPI. 

And, those APIs are really used?

MdeModulePkg/Universal/Disk/UnicodeCollation/EnglishDxe/UnicodeCollationEng.c
pierregondois: Yes right, it seems that the conversion to upper case by subtracting 0x20 doesn't work for:
```
D7 - × - Multiplication sign
F7 - ÷ - Division sign

DF - ß - Latin small letter sharp s - ess-zed
FF - ÿ - Latin small letter y with diaeresis
```
so they are just tagged as invalid characters.

About changing `UINTN Index -> UINT8 Index`, we have `MAP_TABLE_SIZE == 0x100`,
so this might lead to a never ending loop as `UINT8 Index` will overflow to `0` before reaching `0x100`.
Otherwise the check itself should be casted to a bigger type.
I assume it is simpler to let the types as is.

MdePkg/Include/Library/BaseLib.h
pierregondois: Right now the following functions are used:
- AsciiCharIsLowerAlpha()
- AsciiCharIsUpperAlpha()
- AsciiCharIsAlpha()
- AsciiCharIsNum()
- AsciiCharIsHexNum()
- AsciiStrnIsAlpha ()
- AsciiStrnIsNum()
- AsciiStrnIsHexNum()
- CharIsNum()
- CharIsHexNum()
- CharIsUpperHexNum()
- StrnIsNum()
- StrnIsHexNum()
- StrnIsUpperHexNum()

The following functions are not used:
- AsciiCharIsUpperHexNum()
- AsciiCharIsAlphaNum()
- AsciiStrnIsLowerAlpha()
- AsciiStrnIsUpperAlpha()
- AsciiStrnIsUpperHexNum()
- AsciiStrnIsAlphaNum()

I stopped looking for places to use these functions to have a smaller patchset.
It can be assumed that most of places where using these functions have been spotted, but there are still some places that would benefit from it.

MdePkg/Include/Library/BaseLib.h
pierregondois: Hello Leif,
I updated `MdePkg/BaseLib: Add Unicode string case/apha/num check functions`,
but as your comment on the last PR was only concerning Unicode strings
I kept all the functions for Ascii strings, so the commit header should
still be valid for:
`MdePkg/BaseLib: Add Ascii string case/apha/num check functions`

MdePkg/Library/BaseLib/String.c
pierregondois: Yes right, this is done in 2 steps, `MdePkg/BaseLib: Replace InternalAsciiIsHexaDecimalDigitCharacter()` should remove the `InternalAsciiIsHexaDecimalDigitCharacter()` function itself

PR #12010, opened 2026-01-19
EmulatorPkg: Fix build issue when REDFISH_ENABLE is TRUE

YangGangUEFI:
# Description

Fix build issue when REDFISH_ENABLE is TRUE

- [ ] Breaking change?
  - N/A
- [ ] Impacts security?
  - N/A
- [ ] Includes tests?
  - N/A

## How This Was Tested

`DEFINE REDFISH_ENABLE = TRUE`

and

`build -p .\EmulatorPkg\EmulatorPkg.dsc -a X64 -t VS2022 -b DEBUG`

## Integration Instructions

N/A

PR #12016, opened 2026-01-19
EmulatorPkg: Add missing library instance when TLS enabled

khaliid2040:
# Description

Currently, build fails with TLS in emulatorPkg due missing library
instance of TlsLib. Add the library instance.

One of the library instances TlsLib depends on is OpensslLib, however the current library instance of OpensslLib which is OpensslLibCrypto.inf doesn't have TLS features. This causes link to fail when that library instance is linked because openssl (the upstream implemtation) symbols not get resolved.

Also, move Network.dsc.inc at the top of the file where other dsc.inc are declared.

- [ ] Breaking change?
  - NO
- [ ] Impacts security?
  - NO
- [ ] Includes tests?
  - NO

## How This Was Tested
Tested locally
## Integration Instructions
N/A

### DIFF
--- a/EmulatorPkg/EmulatorPkg.dsc
+++ b/EmulatorPkg/EmulatorPkg.dsc
@@ -51,6 +51,7 @@
 !include MdePkg/MdeLibs.dsc.inc
 !include CryptoPkg/CryptoPkgFeatureFlagPcds.dsc.inc
 !include RedfishPkg/Redfish.dsc.inc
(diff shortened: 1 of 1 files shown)


khaliid2040:
@YangGangUEFI Please review and test this PR. And i would suggest you close  #12010. So we fix it in one PR

khaliid2040:
@os-d I finished this PR, please review when you have time.

PR #12018, opened 2026-01-20
Ovmf build

lakshmi-h-ext:
# Description
This automated, on-demand GitHub Actions workflow for OVMF builds is now ready for use with the following capabilities:
Option to build from the latest code or from a specific Git tag
Ability to select the build type at trigger time:
Secure
Standard
Both (default)
Due to network isolation between GitHub-hosted runners and the internal environment, we are currently unable to upload artifacts to the GreenLake Harbor registry. As an interim solution, the build artifacts are being published to GitHub Actions Artifacts.


lakshmi-h-ext:
Ignore the Pull request

PR #12024, opened 2026-01-20
MdeModulePkg/Universal/RegularExpressionDxe: Fix VS2022 NOOPT IA32

mdkinney:
# Description

Add implementation of intrinsic __allmul that is generated with VS2022 NOOPT IA32 builds.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verified that VS2022 NOOPT IA32 builds that use RegularExpressionDxe pass with no unresolved externals.

## Integration Instructions

N/A


lgao4:
There is the library MdePkg\Library\CompilerIntrinsicsLib\CompilerIntrinsicsLib.inf. Can this API be added into it?

mdkinney:
> There is the library MdePkg\Library\CompilerIntrinsicsLib\CompilerIntrinsicsLib.inf. Can this API be added into it?

That library is currently only enabled for AARCH64 and LOONGARCH and is for edk2 components that can not prevent use of intrinsics. This lib is also not correct.  It declares LIBCLASS of CompilerIntrinsicLib but uses NULL mapping in MdeLibs.dsc.inc.  The NULL mapping is required because the intrinsics handled can be added to any module.  The problem with NULL lib mappings at MdeLibs.dsc.inc is that the lib is unconditionally linked against all modules which can grow module sizes, especially for unoptimized builds.

This use case is a submodule that causes an intrinsic to be used and the pattern followed for submodules is to all the intrinsics local to the submodule so the extra APIs are only linked against the one module that needs it.

I like the idea of a consolidated intrinsic lib with a proper LIBCLASS and lib dependencies declared in INF and proper lib mapping in a DSC file to only use it for modules that require it. However, that is a larger task and should be a feature request.

lgao4:
> > There is the library MdePkg\Library\CompilerIntrinsicsLib\CompilerIntrinsicsLib.inf. Can this API be added into it?
> 
> That library is currently only enabled for AARCH64 and LOONGARCH and is for edk2 components that can not prevent use of intrinsics. This lib is also not correct. It declares LIBCLASS of CompilerIntrinsicLib but uses NULL mapping in MdeLibs.dsc.inc. The NULL mapping is required because the intrinsics handled can be added to any module. The problem with NULL lib mappings at MdeLibs.dsc.inc is that the lib is unconditionally linked against all modules which can grow module sizes, especially for unoptimized builds.
> 
> This use case is a submodule that causes an intrinsic to be used and the pattern followed for submodules is to all the intrinsics local to the submodule so the extra APIs are only linked against the one module that needs it.
> 
> I like the idea of a consolidated intrinsic lib with a proper LIBCLASS and lib dependencies declared in INF and proper lib mapping in a DSC file to only use it for modules that require it. However, that is a larger task and should be a feature request.

I see. So, this API is added for the submodule. I am ok for this change.

lgao4 (approved):

Issue #10451, opened 2023-11-04
Consider removing SignedCapsulePkg from edk2 repository (Bugzilla Bug 4589)


tianocore-issues:
This issue was created automatically with [bugzilla2github](https://github.com/makubacki/bugzilla2github/tree/bz2gh_tianocore)
# Bugzilla Bug 4589
Date: 2023-11-04T21:17:39+00:00
From: @mdkinney
To:   unassigned &lt;<unassigned>&gt;
CC:   @ardbiesheuvel, @lgao4, @leiflindholm

Last updated: 2024-01-30T22:12:26+00:00

tianocore-issues:
## Comment 22129
Date: 2023-11-04 21:17:39 +0000
From: @mdkinney

- Industry Specification: ---
- Release Observed: EDK II Master
- Releases to Fix: EDK II Master
- Target OS: ---
- Bugzilla Assignee(s): unassigned &lt;<unassigned>&gt;

The FmpDevicePkg is the replacement for SignedCapsulePkg.

A few platforms in edk2-platforms depend on SignedCapsulePkg.

If those platforms are no longer maintained and are removed from
edk2-platforms, then SignedCapsulePkg can be safely removed from 
the edk2 repository.

The other option is to move SignedCapulePkg into the edk2-plaforms
repository.

tianocore-issues:
## Comment 22474
Date: 2024-01-30 22:12:26 +0000
From: @lgao4

Add Ard and Leif for disucssion first.

os-d:
@mdkinney @leiflindholm @ardbiesheuvel looks like in this PR: https://github.com/tianocore/edk2/pull/6475, there was agreement to archive the edk2-platforms platforms that depend on SignedCapsulePkg and then removing SignedCapsulePkg per this issue. Anything else needed to discuss or just have the removals done?

ardbiesheuvel:
So does this mean we'll get an `archived-after-edk2-stable-XXXX` branch in `edk2-platforms`, and the platforms are removed from `master`? That sounds fine to me.

mdkinney:
A tag in edk2-platforms may be better choice than a branch if there are no new commits expected.

PR #6475, opened 2024-11-27
SignedCapsulePkg: Allow hex FirmwareType values in ini files

bexcran:
# Description

Some platforms such as Ampere Altra have special values for the FirmwareType field: for example 2147483649 is for SCP updates.

To make such values easier to read/understand, allow them to be written in hex: in this case it would be 0x80000001. Implement this by attempting to get a hex integer from the ini data file if reading a decimal value fails.

- [ ] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [ ] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

## How This Was Tested

Write a SystemFirmwareUpdateConfig.ini file with a FirmwareType value of 0x80000002 and verified it was correctly read and interpreted when applying a capsule update.

## Integration Instructions

N/A

PR #12031, opened 2026-01-20
SignedCapsulePkg/Universal/RecoveryModuleLoadPei: 4KB alignment

mdkinney:
# Description

4KB alignment required because this module exceeds maximum size for some NOOPT build configurations. This is a post-memory PEIM that can be compressed, so size impacts for 4KB alignment can be minimized. Must be in DLINK2_FLAGS to override DLINK_XIPFLAGS for GCC PEIM build rules flags ordering.

Build failure detected with NOOPT AARCH64 builds.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verified that SignedCapsulePkg build passes for all NOOPT configurations after this change

## Integration Instructions

N/A


os-d:
@mdkinney I don't have a stake in this package, but perhaps we should just remove it from CI as part of the deprecation process? It looks like there is agreement to drop it: https://github.com/tianocore/edk2/issues/10451 and most of the edk2-platforms that use it can be archived: https://github.com/tianocore/edk2/pull/6475.

But as a first step, certainly could just remove it from CI, then separately work towards deleting it. Then it doesn't have to be maintained and does save >20 mins in our pipelines (and that's with only running it for IA32 + X64 and VS2022 + GCC. If we add CLANGPDB and AARCH64 to the mix we sink even more time building a package we are trying to remove):

Recent Windows CI:

image


Recent GCC CI:

image

mdkinney:
The failure was seen in a GCC NOOPT configuration, and just happened to trip across the size limit on AARCH64 first.  I believe the same issue could show up on the supported GCC NOOPT IA32/X64 configurations.

SignedCapulePkg has been on the list to remove from edk2 for a long time.  If that can be accelerated, then I would gladly drop this PR. What are the tasks remaining to remove SignedCapsulePkg?

os-d:
> SignedCapulePkg has been on the list to remove from edk2 for a long time. If that can be accelerated, then I would gladly drop this PR. What are the tasks remaining to remove SignedCapsulePkg?

Looks like retiring a set of platforms in edk2-platforms as agreed upon in: https://github.com/tianocore/edk2/pull/6475. Then either moving the other platforms to FmpDevicePkg or putting a local copy (or copies) of SignedCapsulePkg in edk2-platforms and let the platforms move on their own time but have to carry the maintenance burden. This would be:
- Platform/Intel/Vlv2TbltDevicePkg/
- Platform/AMD/VanGoghBoard
- Platform/AMD/OverdriveBoard
- Platform/Ampere/JadePkg
- Platform/Intel/QuarkPlatformPkg

However, my point was that if edk2 is saying this package is unsupported, I would say start by removing it from CI. Prevent having the maintenance burden on our side, save 20 mins of CI time, but don't have to try to fix up the platforms immediately. Whatever the next step is can be decided independent of the CI errors.

mdkinney:
I do not agree with removing from CI.  If it is in edk2 repo it is supported.

I agree with the goal of reducing CI time.  Lets focus on removing edk2-platform dependencies and delete as quickly as possible.

mdkinney:
The latest results I see for Platform/Silicon packages in edk2-platforms that reference SignedCapsulePkg are:

* Platform\ADLINK
* Platform\AMD\VanGoghBoard
* Platform\Ampere\JadePkg
* Platform\ASRockRack
* Platform\Intel\Vlv2TbltDevicePkg
* Silicon\Ampere\AmpereAltraPkg
* Silicon\Marvell\Armada7k8k

os-d:
> I do not agree with removing from CI. If it is in edk2 repo it is supported.
> 

That's fair.

> I agree with the goal of reducing CI time. Lets focus on removing edk2-platform dependencies and delete as quickly as possible.

Another option is to move SignedCapsulePkg into edk2-platforms. If the statement is that edk2 wants to delete SignedCapsulePkg but we don't want to break platforms, that would be a bridge, giving the platforms who are using the deprecated package the maintenance burden of it. Obviously, no movement has been made otherwise, so unless someone signs up to do the work of removing it from edk2-platforms (I don't have the knowledge of either SignedCapsulePkg or FmpDevicePkg to help), I don't think the situation is going to change unless edk2 says the package must live in edk2-platforms now, migrate to FmpDevicePkg to follow edk2.

SignedCapsulePkg had 3 commits in 2025, only one of which was actually to improve the code of the package. It had 6 commits in 2024, none of which were to improve the code in the package (i.e. adding StackCheckLib everywhere, etc.). This pattern continues back to ~2018 :).

But again, this doesn't need to affect this PR and I defer a decision to you who are much better informed in this space. I just remembered that SignedCapsulePkg has been slated for deletion for a long time and seeing a PR come in just to fix a CI break on it reminded me that it does take a long time to build and for that reason alone it would be nice to remove, beyond whatever the goals were to move to FmpDevicePkg.

> The latest results I see for Platform/Silicon packages in edk2-platforms that reference SignedCapsulePkg are:
> 
> * Platform\ADLINK
> * Platform\AMD\VanGoghBoard
> * Platform\Ampere\JadePkg
> * Platform\ASRockRack
> * Platform\Intel\Vlv2TbltDevicePkg
> * Silicon\Ampere\AmpereAltraPkg
> * Silicon\Marvell\Armada7k8k

FWIW, it seems Ard and Leif said Armada7k8k can be archived: https://github.com/tianocore/edk2/pull/6475#issuecomment-2551560151. Looks like the others mentioned either already were or dropped the SignedCapsulePkg dependency.

mdkinney:
> > I do not agree with removing from CI. If it is in edk2 repo it is supported.
> 
> That's fair.
> 
> > I agree with the goal of reducing CI time. Lets focus on removing edk2-platform dependencies and delete as quickly as possible.
> 
> Another option is to move SignedCapsulePkg into edk2-platforms. If the statement is that edk2 wants to delete SignedCapsulePkg but we don't want to break platforms, that would be a bridge, giving the platforms who are using the deprecated package the maintenance burden of it. Obviously, no movement has been made otherwise, so unless someone signs up to do the work of removing it from edk2-platforms (I don't have the knowledge of either SignedCapsulePkg or FmpDevicePkg to help), I don't think the situation is going to change unless edk2 says the package must live in edk2-platforms now, migrate to FmpDevicePkg to follow edk2.
> 
> SignedCapsulePkg had 3 commits in 2025, only one of which was actually to improve the code of the package. It had 6 commits in 2024, none of which were to improve the code in the package (i.e. adding StackCheckLib everywhere, etc.). This pattern continues back to ~2018 :).
> 
> But again, this doesn't need to affect this PR and I defer a decision to you who are much better informed in this space. I just remembered that SignedCapsulePkg has been slated for deletion for a long time and seeing a PR come in just to fix a CI break on it reminded me that it does take a long time to build and for that reason alone it would be nice to remove, beyond whatever the goals were to move to FmpDevicePkg.
> 
> > The latest results I see for Platform/Silicon packages in edk2-platforms that reference SignedCapsulePkg are:
> > 
> > * Platform\ADLINK
> > * Platform\AMD\VanGoghBoard
> > * Platform\Ampere\JadePkg
> > * Platform\ASRockRack
> > * Platform\Intel\Vlv2TbltDevicePkg
> > * Silicon\Ampere\AmpereAltraPkg
> > * Silicon\Marvell\Armada7k8k
> 
> FWIW, it seems Ard and Leif said Armada7k8k can be archived: [#6475 (comment)](https://github.com/tianocore/edk2/pull/6475#issuecomment-2551560151). Looks like the others mentioned either already were or dropped the SignedCapsulePkg dependency.

I would prefer to not add SignedCapsulePkg to edk2-platforms.  Let's reach out to the platform/si owners and see if we can move all to archived state and decide a date to do so with a tag on edk2-platforms just before that change.  One option is to create the edk2-platforms PRs that remove these plaforms/si components to see if the owners will accept the PRs. One PR per plaform/si area, so the scope of each PR will be the single owner.

I will also note that the change in this specific PR is interesting because it shows that large PEIMs require a different section alignment.  That becomes a constraint on the section alignment topic where alignment constraints can come from structures in compiled code, from security requirement to page align section to apply page protections to code/ro data/rw data.  And now additional linker restrictions that require larger alignments if a module to too big.

mdkinney:
Draft PR to remove VLV2: https://github.com/tianocore/edk2-platforms/pull/934

os-d:
> I would prefer to not add SignedCapsulePkg to edk2-platforms. Let's reach out to the platform/si owners and see if we can move all to archived state and decide a date to do so with a tag on edk2-platforms just before that change. One option is to create the edk2-platforms PRs that remove these plaforms/si components to see if the owners will accept the PRs. One PR per plaform/si area, so the scope of each PR will be the single owner.
> 

Sounds reasonable, though from a cursory look at the git history, it does seem like at least a few of these are somewhat actively updated. Those owners may need to update to FmpDevicePkg, in which case we are back at the same issue of how do we convince the owners to update when they haven't for however many years?

> I will also note that the change in this specific PR is interesting because it shows that large PEIMs require a different section alignment. That becomes a constraint on the section alignment topic where alignment constraints can come from structures in compiled code, from security requirement to page align section to apply page protections to code/ro data/rw data. And now additional linker restrictions that require larger alignments if a module to too big.

Yeah, that is a concern of how tangled it all gets with different requirements. Hopefully we end up with build time failures when we don't wire it up correctly.

mdkinney:
@makubacki We will continue to work to remove SignedCapsulePkg.  In the mean time, can we get this change reviewed and merged to address failures until it is removed.

PR #12034, opened 2026-01-20
EmbeddedPkg/Universal/MmcDxe: Fix VS2022 NOOPT IA32/X64 issues

mdkinney:
# Description

Fix type mismatches found with VS2022 NOOPT IA32 and X64 builds

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verified VS2022 NOOPT IA32/X64 builds of EmbeddedPkg pass after this change.

## Integration Instructions

N/A


EmbeddedPkg/Universal/MmcDxe/MmcIdentification.c
```diff
@@ -166,7 +166,7 @@ EmmcIdentificationMode (
   }
 
   // Assign a relative address value to the card
-  MmcHostInstance->CardInfo.RCA = ++mEmmcRcaCount; // TODO: might need a more sophisticated way of doing this
+  MmcHostInstance->CardInfo.RCA = (UINT16)++ mEmmcRcaCount; // TODO: might need a more sophisticated way of doing this
```
ardbiesheuvel: Did uncrustify reformat this? It looks positively weird. And given the way it is used here, perhaps `mEmmcRcaCount` should be a `UINT16` to begin with?

mdkinney: Passed uncrustify check.

I will evaluate changing mEmmRcaCount to UINT16.

mdkinney: I agree that the space character following ++ looks very strange.  Don't know if I introduced that in my edit and uncrustify accepted it.  Or if uncrustify requires that when a typecase is before ++.  We do not have many pre-increment or pre-decrement statements.

mdkinney: I pushed update that changes mEmmRcaCount to UINT16

PR #12035, opened 2026-01-21
BaseTools: Enable -Os optimization by default for GCC5_RISCV targets

tphan-qualcomm:
# Description
To match with other architectures, enabling -Os optimization flag by default for all GCC5_RISCV targets

- [ ] Breaking change?
- [ ] Impacts security?.
- [ ] Includes tests?

## How This Was Tested
Built EDk2 for RiscVVirt target and compared the size of generated DXEFV. 
Without -Os: DXEFV 7667712 (0x750000)
With -Os: DXEFV 5373952 (0x520000)

## Integration Instructions
N/A


EmbeddedPkg/Library/PrePiLib/FwVol.c
```diff
@@ -395,7 +395,7 @@ FfsProcessSection (
       return EFI_SUCCESS;
     } else if ((Section->Type == EFI_SECTION_COMPRESSION) || (Section->Type == EFI_SECTION_GUID_DEFINED)) {
       CHAR8   *CompressedData;
-      UINT32  CompressionSectionHeaderSize;
+      UINT32  CompressionSectionHeaderSize = 0;
```
lgao4: Please set its value after declare it to follow edk2 coding style.

leiflindholm: No symptom bashing, please.

I spent a fair amount of time simplifying this code which used to be even more unreadable (4a1dca59f673..615e5ca40a0d).

Let's not start making it worse again in order to pass ci.
If the code needs further refactoring, let's do that.

MdePkg/MdeLibs.dsc.inc
```diff
@@ -60,12 +60,12 @@
   # implements exception handling for SEC and PEI_CORE, it can use StackCheckLib for these phases in its DSC.
   StackCheckLib|MdePkg/Library/StackCheckLibNull/StackCheckLibNull.inf
 
-[LibraryClasses.AARCH64]
+[LibraryClasses.AARCH64, LibraryClasses.RISCV64]
   #
-  # It is not possible to prevent the AARCH64 compilers from inserting generic intrinsic functions.
+  # It is not possible to prevent the AARCH64/RISCV64 compilers from inserting generic intrinsic functions.
```
leiflindholm: I would suggest dropping "AARCH64" instead of enumerating all architectures.

MdePkg/MdeLibs.dsc.inc
```diff
@@ -60,12 +60,12 @@
...
 
-[LibraryClasses.AARCH64]
+[LibraryClasses.AARCH64, LibraryClasses.RISCV64]
   #
-  # It is not possible to prevent the AARCH64 compilers from inserting generic intrinsic functions.
+  # It is not possible to prevent the AARCH64/RISCV64 compilers from inserting generic intrinsic functions.
   # This library provides the intrinsic functions generated by these compilers.
   #
-  # Linking this here as a null library will cause all AARCH64 files to link against it and have
+  # Linking this here as a null library will cause all AARCH64/RISCV64 files to link against it and have
```
leiflindholm: Same again, drop AARCH64 instead.

leiflindholm (changes requested):

EmbeddedPkg/Library/PrePiLib/FwVol.c
tphan-qualcomm: Will do, tks

tphan-qualcomm: Agree, will need a little bit of refactoring.

MdePkg/MdeLibs.dsc.inc
tphan-qualcomm: Sure

MdePkg/MdeLibs.dsc.inc
tphan-qualcomm: Sure

EmbeddedPkg/Library/PrePiLib/FwVol.c
leiflindholm: Thanks, the refactoring is good.

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
```diff
@@ -88,8 +88,11 @@ PopulateHeaderAndCommunicate (
     goto Exit;
   }
 
-  MmCommunicationPpiV3 = NULL;
-  MmCommunicationPpi   = NULL;
+  MmCommunicationPpiV3  = NULL;
```
leiflindholm: This (including the code before this change) is just hiding errors.
MmCommunicationPpiV3 and MmCommunicationPpi should be initialized to NULL if the corresponding PeiServicesLocatePpi() calls fail.

The other warnings are caused because the bodies of
if (MmCommunicationPpiV3 != NULL) {
and
} else if (MmCommunicationPpi != NULL) {
should be inside helper functions. And because we've not done that, we're hiding actual control flow from the compiler.

MdeModulePkg/Bus/Spi/SpiNorFlashJedecSfdp/SpiNorFlashJedecSfdp.c
```diff
@@ -1502,6 +1507,7 @@ SpiReadSfdpPtp (
   UINT32      MaximumTransferBytes;
   UINT32      TransactionBufferLength;
 
+  Status               = EFI_SUCCESS;
```
leiflindholm: This merely hides a problem with the code. Or possibly more than one.

First of all, the error happens because it's not guaranteed that LengthInBytes is always greater than 0. If LengthInBytes is 0 (causing the for-loop to never be executed), we probably shouldn't be calling this function at all.

Secondly, there is a silly bit of success handling on (original) line 1521.
If Status is an error condition here, the result is that the function returns that error condition. So let's do that directly instead of treating normal operation as a corner case.

Thirdly, if the call to Transaction returns an error, we ASSERT but then (if ASSERTs don't cause a halt) happily continue around the loop and potentially return success if a subsequent call succeeds. This does not feel right.

So I think it would make more sense for that failure also causing a return.
At which point the function could simply return EFI_SUCCESS directly at the end.

@changab thoughts?

MdeModulePkg/Bus/Usb/UsbNetwork/NetworkCommon/DriverBinding.c
```diff
@@ -185,6 +185,7 @@ NetworkCommonDriverStart (
     return Status;
   }
 
+  TmpPxePointer = NULL;
```
leiflindholm: While the ideal solution here would be to rewrite this function to not have four separate cleanup blocks instead of a goto ErrorExit or similar ... the TmpPxePointer NULLing moving would still be the right thing to do (why in heck was it being NULLed on the line before an Alloc call anyway?).

However, we're not trying to make the compiler go quiet, we're trying to make the code more correct. Please update the commit message to reflect that TmpPxePointer was being referenced uninitialized in later cleanup blocks.

leiflindholm (changes requested):

MdeModulePkg/Universal/Variable/MmVariablePei/MmVariablePei.c
tphan-qualcomm: I think I overdone this code, only MmVarCommsHeader is causing the warning. There is nothing wrong with the MmCommunicationPpiV3 and MmCommunicationPpi, just a sense that all pointers should be initialized at NULL. Anyway, did a change to make it clearer.

MdeModulePkg/Bus/Spi/SpiNorFlashJedecSfdp/SpiNorFlashJedecSfdp.c
tphan-qualcomm: - Agree that if LengthInBytes is zero, the function shouldn't be called at all. But it is safe to return EFI_SUCCESS if Length is zero as major reading bytes from device return success in case of length is zero (assuming a lot of callers make mistake calling zero length during reading loop)
- Agree that ASSERT shouldn't be used to treat error as RELEASE build to force it move on and data may be corrupted. But let the owner of this code comment on that. If we treat an error as a condition to return in the loop then EFI_SUCCESS can be returned at the end.

MdeModulePkg/Bus/Usb/UsbNetwork/NetworkCommon/DriverBinding.c
tphan-qualcomm: To make it clearer, moved TmpPxePointer to else statement.

MdeModulePkg/Bus/Spi/SpiNorFlashJedecSfdp/SpiNorFlashJedecSfdp.c
```diff
@@ -331,12 +331,17 @@ GetEraseTypeRecord (
...
   //
   switch (SearchType) {
     case SearchEraseTypeByType:
     case SearchEraseTypeByCommand:
     case SearchEraseTypeBySize:
+      //
+      // Although ValueToCompare is not used for these types, it must be
+      // initialized to avoid "variable may be used uninitialized" compiler warning.
+      //
+      ValueToCompare = 0;
```
lgao4, line 344: How about init ValueToCompare to zero at the begin of this function?

leiflindholm: The warnings are useful and frequently highlight either excessive complexity or actual errors.

This change localizes the initalization to where it's needed, so won't hide bugs introduced by future changes.

NetworkPkg/HttpBootDxe/HttpBootImpl.c
```diff
@@ -395,9 +395,11 @@ HttpBootGetBootFileCaller (
         }
 
         //
-        // Load the boot file into Buffer
+        // Load the boot file into Buffer.
         //
-        for (Retries = 1; Retries <= PcdGet32 (PcdMaxHttpResumeRetries); Retries++) {
+        ASSERT (PcdGet32 (PcdMaxHttpResumeRetries) > 0);
```
SaloniKasbekar: cant we just initialize status to EFI_UNSUPPORTED instead of making this change?

NetworkPkg/HttpBootDxe/HttpBootImpl.c
```diff
@@ -397,6 +397,7 @@ HttpBootGetBootFileCaller (
         //
         // Load the boot file into Buffer
         //
+        Status = EFI_UNSUPPORTED;
```
SaloniKasbekar: Can you initialize this at the top? before the for() loop?

SaloniKasbekar (approved):
Approved for NetworkPkg

leiflindholm:
@tphan-qualcomm whoops, this one seems to have fallen over the horizon.
Could you rebase to current upstream please, and we'll have another go?

tphan-qualcomm:
> @tphan-qualcomm whoops, this one seems to have fallen over the horizon. Could you rebase to current upstream please, and we'll have another go?

Done. Just added one more patch to fix warning for a recent patch. Hopefully, this PR can be merged soon so new PR will be forced to check code carefully to avoid build failed.

MdeModulePkg/Bus/Spi/SpiNorFlashJedecSfdp/SpiNorFlashJedecSfdp.c
leiflindholm: Guess we had no comments from @changab, but I'd prefer to drop this initialization, add the error return, and the direct EFI_SUCCESS return at the end.

NetworkPkg/HttpBootDxe/HttpBootImpl.c
```diff
@@ -360,6 +360,8 @@ HttpBootGetBootFileCaller (
   EFI_STATUS                Status;
   UINT32                    Retries;
 
+  Status = EFI_UNSUPPORTED;
```
leiflindholm: I think there are real bugs in this code, under the LoadBootFile case.

How about this (compile-tested only) rework/untanglement.

```
@@ -447,39 +447,41 @@ HttpBootGetBootFileCaller (
                      Buffer,
                      ImageType
                      );
-          if (!EFI_ERROR (Status) ||
-              ((Status != EFI_TIMEOUT) && (Status != EFI_DEVICE_ERROR)) ||
-              (Retries >= PcdGet32 (PcdMaxHttpResumeRetries)))
-          {
-            break;
+          if (!EFI_ERROR (Status)) {
+            return Status;
           }
 
-          //
-          // HttpBootGetBootFile returned EFI_TIMEOUT or EFI_DEVICE_ERROR.
-          // We may attempt to resume the interrupted download.
-          //
-
-          Private->HttpCreated = FALSE;
-          HttpIoDestroyIo (&Private->HttpIo);
-          Status = HttpBootCreateHttpIo (Private);
-          if (EFI_ERROR (Status)) {
+          if ((Status != EFI_TIMEOUT) && (Status != EFI_DEVICE_ERROR)) {
             break;
-          }
+          } else {
 
-          DEBUG ((DEBUG_WARN | DEBUG_INFO, "HttpBootGetBootFileCaller: NBP file download interrupted, will try to resume the operation.\n"));
-          gBS->Stall (1000 * 1000 * PcdGet32 (PcdHttpDelayBetweenResumeRetries));
+            //
+            // HttpBootGetBootFile returned EFI_TIMEOUT or EFI_DEVICE_ERROR.
+            // We may attempt to resume the interrupted download.
+            //
+
+            Private->HttpCreated = FALSE;
+            HttpIoDestroyIo (&Private->HttpIo);
+            Status = HttpBootCreateHttpIo (Private);
+            if (EFI_ERROR (Status)) {
+              break;
+            }
+
+            DEBUG ((DEBUG_WARN | DEBUG_INFO, "HttpBootGetBootFileCaller: NBP file download interrupted, will try to resume the operation.\n"));
+            gBS->Stall (1000 * 1000 * PcdGet32 (PcdHttpDelayBetweenResumeRetries));
+          }
         }
 
-        if (EFI_ERROR (Status) && (Retries >= PcdGet32 (PcdMaxHttpResumeRetries))) {
+        if (Retries > PcdGet32 (PcdMaxHttpResumeRetries)) {
           DEBUG ((DEBUG_ERROR, "HttpBootGetBootFileCaller: Error downloading NBP file, even after trying to resume %d times.\n", Retries));
         }
 
-        return Status;
+        return EFI_HTTP_ERROR;
 
       case GetBootFileError:
       default:
         AsciiPrint ("\n  Error: Could not retrieve NBP file size from HTTP server.\n");
-        return Status;
+        return EFI_UNSUPPORTED;
     }
   }
 }
```

MdePkg/Include/Library/DxeRiscvMpxy.h
```diff
@@ -38,6 +38,8 @@ typedef enum {
   @retval  EFI_SUCCESS            Library is successfully initialized.
   @retval  EFI_ALREADY_STARTED    Library is already initialized.
   @retval  EFI_OUT_OF_RESOURCES   Not enough memory to allocate.
+  @retval  EFI_UNSUPPORTED        MPXY SBI extension is not supported or shared memory size is 0.
```
leiflindholm: This doesn't feel like it belongs to this PR.

MdePkg/Library/DxeRiscvMpxyLib/DxeRiscvMpxy.c
```diff
@@ -615,6 +615,8 @@ SbiMpxySendMessage (
   @retval  EFI_SUCCESS            Library is successfully initialized.
   @retval  EFI_ALREADY_STARTED    Library is already initialized.
   @retval  EFI_OUT_OF_RESOURCES   Not enough memory to allocate.
+  @retval  EFI_UNSUPPORTED        MPXY SBI extension is not supported or shared memory size is 0.
```
leiflindholm: This doesn't feel like it's part of this PR.

MdePkg/Library/DxeRiscvMpxyLib/DxeRiscvMpxy.c
```diff
@@ -635,11 +637,16 @@ SbiMpxyLibInit (
     return Status;
   }
 
+  ShmemSize = 0;
```
leiflindholm: This is a compiler bug though (or at least compiler being daft).

I'd prefer to hide this behind
```
#pragma GCC diagnostic push                                                                                                                              
#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"       
...
#pragma GCC diagnostic pop                                                                                                                               
```

MdePkg/Library/DxeRiscvMpxyLib/DxeRiscvMpxy.c
```diff
@@ -635,11 +637,16 @@ SbiMpxyLibInit (
...
   }
 
+  ShmemSize = 0;
   Status = SbiMpxyGetShmemSize (&ShmemSize);
   if (EFI_ERROR (Status)) {
     return Status;
   }
 
+  if (ShmemSize == 0) {
+    return EFI_UNSUPPORTED;
```
leiflindholm: This feels unrelated to this commit description?

leiflindholm (changes requested):

MdePkg/Library/DxeRiscvMpxyLib/DxeRiscvMpxy.c
```diff
@@ -155,6 +155,7 @@ SbiMpxyGetShmemSize (
     return EFI_SUCCESS;
   }
 
+  *ShmemSize = 0;
```
leiflindholm, line 158: I still say this is a compiler bug to warn about, but I approve of this change anyway.

leiflindholm (approved):
I don't like the state of the functions requiring those last two commits, but I also don't want to keep dragging this out, so I'll have a look at those separately.

leiflindholm:
OK, no sign of life from BaseTools maintainers. As you've pointed out, the longer it takes for this to go in, the more new bugs we seem to accumulate, so I'm setting the push label.

PR #12038, opened 2026-01-21
MdeModulePkg/UsbKbDxe: Handle Ctrl+Alt+Del reset at TPL_CALLBACK

kritpals:
Defer the Ctrl+Alt+Del ResetSystem invocation to TPL_CALLBACK in the UsbKbDxe override. Previously the reset was performed directly in the key-processing path, which could occur at an elevated task priority and lead to contention or multiple system-control calls.

This change introduces a dedicated event and a pending flag on the USB_KB_DEV instance. When Ctrl and Alt modifiers accompany Delete, the driver sets the pending flag and signals an event, the event callback consumes the request and performs a warm reset at TPL_CALLBACK.

@vishalo @leiflindholm

### DIFF
--- a/MdeModulePkg/Bus/Usb/UsbKbDxe/EfiKey.c
+++ b/MdeModulePkg/Bus/Usb/UsbKbDxe/EfiKey.c
@@ -563,6 +563,7 @@ USBKeyboardDriverBindingStop (
   gBS->CloseEvent (UsbKeyboardDevice->SimpleInput.WaitForKey);
   gBS->CloseEvent (UsbKeyboardDevice->SimpleInputEx.WaitForKeyEx);
   gBS->CloseEvent (UsbKeyboardDevice->KeyNotifyProcessEvent);
+  gBS->CloseEvent (UsbKeyboardDevice->CtrlAltDelEvent);
   KbdFreeNotifyList (&UsbKeyboardDevice->NotifyList);
 
   ReleaseKeyboardLayoutResources (UsbKeyboardDevice);
--- a/MdeModulePkg/Bus/Usb/UsbKbDxe/EfiKey.h
+++ b/MdeModulePkg/Bus/Usb/UsbKbDxe/EfiKey.h
@@ -153,6 +153,9 @@ typedef struct {
   USB_NS_KEY                           *CurrentNsKey;
   EFI_KEY_DESCRIPTOR                   *KeyConvertionTable;
   EFI_EVENT                            KeyboardLayoutEvent;
+
+  // Event to handle Ctrl + Alt + Del key
+  EFI_EVENT                            CtrlAltDelEvent;
 } USB_KB_DEV;
 
 //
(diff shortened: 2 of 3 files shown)


jkolakow (approved):

mdkinney:
In general, all code should run at the lowest possible TPL, so this change makes sense from that perspective.

What about all the other Simple Input or Simple Input Ex device drivers.  We should be consistent about CTRL-ALT-DEL processing across all Simple Input devices.

My larger concern is that this seems to be a workaround for a call to ResetSystem() at TPL_NOTIFY or higher and that could occur from many location when processing a critical/fatal error condition.  If there are ResetSystem() handlers that are not compatible with running at higher TPL levels, then that seems like a more significant issue and if that is resolved, then this change would not be required.

joschock:
>What about all the other Simple Input or Simple Input Ex device drivers. We should be consistent about CTRL-ALT-DEL processing across all Simple Input devices.

Is there a reason this needs to be implemented in the physical input processing in the device driver rather than having a platform that wants this behavior use SimpleTextIn->RegisterKeyNotify()  to setup a callback on Ctrl-Alt-Del and issuing reset in the callback? It seems weird for a separation-of-concerns standpoint to special-case this in all the input drivers when there is a public API that I think could be used to get this behavior for platforms that want it.

mdkinney:
> > What about all the other Simple Input or Simple Input Ex device drivers. We should be consistent about CTRL-ALT-DEL processing across all Simple Input devices.
> 
> Is there a reason this needs to be implemented in the physical input processing in the device driver rather than having a platform that wants this behavior use SimpleTextIn->RegisterKeyNotify() to setup a callback on Ctrl-Alt-Del and issuing reset in the callback? It seems weird for a separation-of-concerns standpoint to special-case this in all the input drivers when there is a public API that I think could be used to get this behavior for platforms that want it.

Good question.  The reason it started in keyboard drivers is that the original Simple Input Protocol did not support RegisterKeyNotify() so it had to be handled in the keyboard driver.  RegisterKeyNotify() was added in the Simple Input Ex Protocol, so keyboard drivers that implement both Simple Input for back compat and the newer Simple Input Ex could use your suggestion.

The challenge is that there could be keyboard drivers that only implement Simple Input protocol and not Simple Input Ex Protocol. In this case, your suggestion would be a common handler for all keyboard drivers that do support Simple Input Ex and those keyboard drivers that only implement Simple Input would require special processing.

For the edk2 repo, I believe all keyboard drivers implement both, so moving to a common platform driver makes sense.  Also need to consider how this would be handled by the Console Splitter.

kritpals:
Thank you everyone for the comments. 
Good suggestion, and I agree it’s conceptually cleaner for separation of concerns when all consumers can rely on Simple Input Ex. The historical reason Ctrl+Alt+Del started in keyboard drivers is that the original Simple Input protocol didn’t have RegisterKeyNotify(), so drivers handled it directly. While most keyboard drivers in edk2 do implement both protocols today, we also need to consider how this interacts with the Console Splitter and with any environments where only Simple Input is present.
Given the above, I’d like to keep this PR focused on making the existing driver behavior safer (deferring reset out of the key path). I can follow up with an alternative/cleaner approach using RegisterKeyNotify() (potentially via a common module/driver) once we align on the expected behavior across Console Splitter and the set of keyboard drivers that must be supported.

PR #12039, opened 2026-01-21
EmbeddedPkg/PrePiMemoryAllocationLib: Update InternalAllocatePages to satisfy MemoryType page alignment

antklein:
# Description

When allocating Runtime memory, the MemoryAllocation HOB's length needs to meet the alignment requirements of Runtime memory. On AARCH64 architectures this alignment is not the default of 4KB. Instead, it is set by the RUNTIME_PAGE_ALLOCATION_GRANULARITY definition, which is 64KB.

Per AARCH64 spec requirements (see below), runtime memory allocation requirements require alignment to 64KiB. In certain implementations, like the Patina (Rust) DXE Core, this alignment requirement is more strongly enforced and must be met. Patina will reject a memory map that violates those spec requirements, preventing system boot.

https://uefi.org/specs/UEFI/2.11/02_Overview.html#aarch64-platforms
>The ARM architecture allows mapping pages at a variety of granularities, including 4KiB and 64KiB. If a 64KiB physical page contains any 4KiB page with any of the following types listed below, then all 4KiB pages in the 64KiB page must use identical ARM Memory Page Attributes (as described in [Table 2.5](https://uefi.org/specs/UEFI/2.11/02_Overview.html#map-efi-cacheability-attributes-to-aarch64-memory-types)):
>- EfiRuntimeServicesCode
>- EfiRuntimeServicesData
>- EfiReserved
>- EfiACPIMemoryNVS
>
>Mixed attribute mappings within a larger page are not allowed.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

* Validated via a DXE Core (Patina) that validates MemoryAllocation HOB requirements.

## Integration Instructions

N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/code_first.yml
+++ b/.github/ISSUE_TEMPLATE/code_first.yml
@@ -13,7 +13,8 @@ body:
   - type: markdown
     attributes:
       value: |
-        👋 Only use this issue form for changes following the "code first" process described in [EDK II Code First Process](https://github.com/tianocore/tianocore.github.io/wiki/EDK-II-Code-First-Process).
+        👋 Only use this issue form for changes following the "code first" process described in
+           [EDK II Code First Process](https://www.tianocore.org/tianocore-wiki.github.io/development/contribution-guides/edk_ii_code_first_process.html).
 
         **Read that document before filing this issue.**
 
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -10,7 +10,7 @@
 
 contact_links:
   - name: Reporting Security Issues
-    url: https://github.com/tianocore/tianocore.github.io/wiki/Reporting-Security-Issues
+    url: https://www.tianocore.org/tianocore-wiki.github.io/security/processes/reporting_security_issues.html
     about: Read the wiki page that describes the process here
   - name: EDK II Development Mailing List
     url: https://edk2.groups.io/g/devel
--- a/.github/ISSUE_TEMPLATE/documentation_request.yml
+++ b/.github/ISSUE_TEMPLATE/documentation_request.yml
@@ -23,8 +23,9 @@ body:
         alongside the code implementation of a particularly complex feature or module.
 
 
-        - To file an issue for a TianoCore specification, refer to [these instructions](https://github.com/tianocore-docs/edk2-TemplateSpecification/wiki/TianoCore-Documents-GitBook-Overview)
-        and reference the specifications In the [TianoCore Docs organization](https://github.com/tianocore-docs).
+        - To file an issue for a TianoCore specification, refer to
+          [these instructions](https://www.tianocore.org/tianocore-wiki.github.io/reference/specs-standards/edk_ii_documentation.html)
+          and reference the specifications In the [TianoCore Docs organization](https://github.com/tianocore-docs).
 
         - For UEFI specifications, refer to the [UEFI Forum](https://uefi.org/specifications) website.
 
--- a/.pytool/Plugin/UncrustifyCheck/UncrustifyCheck.py
+++ b/.pytool/Plugin/UncrustifyCheck/UncrustifyCheck.py
@@ -577,7 +577,7 @@ class UncrustifyCheck(ICiBuildPlugin):
             logging.warning(
                 "Visit the following instructions to learn "
                 "more about uncrustify setup instructions and CI:"
-                "https://github.com/tianocore/tianocore.github.io/wiki/EDK-II-Code-Formatting\n")
+                "https://www.tianocore.org/tianocore-wiki.github.io/development/coding-standards/edk_ii_code_formatting.html\n")
 
             if self._output_file_diffs:
                 logging.info("Calculating file diffs. This might take a while...")
--- a/.pytool/Readme.md
+++ b/.pytool/Readme.md
@@ -4,7 +4,7 @@ This file focuses on information for those working with the `.pytools` directory
 directly or interested in lower-level details about how CI works.
 
 If you just want to get started building code, visit
-[Build Instructions](https://github.com/tianocore/tianocore.github.io/wiki/Build-Instruction)
+[Build Instructions](https://www.tianocore.org/tianocore-wiki.github.io/build-tooling/build-workflows/build_instructions.html)
 on the TianoCore wiki.
 
 ## Basic Status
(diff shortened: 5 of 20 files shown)


antklein:
@ardbiesheuvel, @changab, @leiflindholm looking for help reviewing this change.

kajal-agarwal-in:
The alignment logic here can over‑allocate past EfiFreeMemoryTop when Pages is small and RUNTIME_PAGE_ALLOCATION_GRANULARITY is large.

For example, with EfiFreeMemoryTop = 0x11000, a 1‑page request ends up aligned to start at 0x10000 and expanded to 64 KB, producing the range [0x10000, 0x20000). This goes well beyond the actual free top (0x11000), effectively overclaiming 0xF000 bytes that aren't free.

The allocation logic should ensure the final range stays fully within the available region; if it doesn’t fit, the allocation should fail rather than overrun the boundary.

antklein:
> The alignment logic here can over‑allocate past EfiFreeMemoryTop when Pages is small and RUNTIME_PAGE_ALLOCATION_GRANULARITY is large.
> 
> For example, with EfiFreeMemoryTop = 0x11000, a 1‑page request ends up aligned to start at 0x10000 and expanded to 64 KB, producing the range [0x10000, 0x20000). This goes well beyond the actual free top (0x11000), effectively overclaiming 0xF000 bytes that aren't free.
> 
> The allocation logic should ensure the final range stays fully within the available region; if it doesn’t fit, the allocation should fail rather than overrun the boundary.

@kajal-agarwal-in Thank you for the review and catching my mistake! I've rectified this with the latest commit. NewTop is now calculated after aligning the requested Pages.

ardbiesheuvel:
> > The alignment logic here can over‑allocate past EfiFreeMemoryTop when Pages is small and RUNTIME_PAGE_ALLOCATION_GRANULARITY is large.
> > For example, with EfiFreeMemoryTop = 0x11000, a 1‑page request ends up aligned to start at 0x10000 and expanded to 64 KB, producing the range [0x10000, 0x20000). This goes well beyond the actual free top (0x11000), effectively overclaiming 0xF000 bytes that aren't free.
> > The allocation logic should ensure the final range stays fully within the available region; if it doesn’t fit, the allocation should fail rather than overrun the boundary.
> 
> @kajal-agarwal-in Thank you for the review and catching my mistake! I've rectified this with the latest commit. NewTop is now calculated after aligning the requested Pages.

Please merge this fix into the original patch, and force-push the result.

kajal-agarwal-in:
The logic at Line 51 subtracts the 64 KB runtime‑granularity allocation before verifying that the free range actually has enough space. On edge EfiFreeMemoryTop values, this early subtraction can underflow and wrap to a large 64‑bit value (e.g., 0xFFFFxxxx), long before the later [NewTop < (Bottom + sizeof(HOB))] safety check runs.

Consider validating the aligned region has at least 64 KB before subtracting, as the current 64 KiB decrement can underflow EfiFreeMemoryTop and produce a wrapped NewTop.

makubacki:
@antklein submitted this PR well before soft freeze and has a few approvals but not from EmbeddedPkg maintainers. @antklein has requested that this PR be merged before the hard freeze.

Are EmbeddedPkg maintainers (@leiflindholm, @ardbiesheuvel, @changab) able to revisit this PR for review? Do you have feedback on its acceptance to merge for 2602?

vishalo:
Memory is being allocated (or reserved) at PEI, which in some cases needs to be reflected in the runtime memory map. Hence the alignment, even though we're at PEI, needs to respect the runtime alignment requirement - to adhere to AARCH64 specification. It would be helpful if this makes it to the stable tag.

leiflindholm:
OK, and presumably this is causing an error (be it in a test or at runtime) which neither the author or anyone else feels is relevant/safe to mention?

For something to be pulled into a stable tag, post freeze, it should be a real error not a theoretical one. Can someone confirm that, and @antklein update the PR description to reflect this?

antklein:
> OK, and presumably this is causing an error (be it in a test or at runtime) which neither the author or anyone else feels is relevant/safe to mention?
> 
> For something to be pulled into a stable tag, post freeze, it should be a real error not a theoretical one. Can someone confirm that, and @antklein update the PR description to reflect this?

Thank you for the comment @leiflindholm. I've updated the PR description to better describe the spec violation this PR intends to fix. Please let me know if there is any additional information you are looking for before considering merging this change.

leiflindholm:
I do wonder if there is some missing or unclear language in PI in this area. And we would probablt want DxeCore to kick back on this also. But this fix looks clear enough, and I think we should make it catch the stable tag.

Thank you for clarifying how it was discovered.

makubacki:
@antklein, your PR description update is great, I just wanted to let you know I modified it slightly for those less familiar with Patina.

ardbiesheuvel:
The UEFI spec does not require all allocations to be rounded up to 64k on AArch64 - that is a misconception. The actual wording is about the use of conflicting memory type attributes within the same 64k page frame when dealing with regions that are exposed as runtime regions to the OS.

So it is really not necessary to round up *every* allocation to 64k - the only requirement is that once the OS takes over, the memory map's layout complies with this requirement, and this could be achieved by other means too.

joschock:
> The UEFI spec does not require all allocations to be rounded up to 64k on AArch64 - that is a misconception. The actual wording is about the use of conflicting memory type attributes within the same 64k page frame when dealing with regions that are exposed as runtime regions to the OS.
> 
> So it is really not necessary to round up _every_ allocation to 64k - the only requirement is that once the OS takes over, the memory map's layout complies with this requirement, and this could be achieved by other means too.

Agreed, but I think trying to implement that flexibility at the level of this library and the HOB allocation list somehow would introduce a lot of complexity. Are you suggesting that is needed here? Most of the other art I see around this requirement also just rounds for simplicity.

leiflindholm:
> The UEFI spec does not require all allocations to be rounded up to 64k on AArch64 - that is a misconception. The actual wording is about the use of conflicting memory type attributes within the same 64k page frame when dealing with regions that are exposed as runtime regions to the OS.
> 
> So it is really not necessary to round up _every_ allocation to 64k - the only requirement is that once the OS takes over, the memory map's layout complies with this requirement, and this could be achieved by other means too.

For sure, but this is the trivial fix, and others would get a lot more hair splitting.
Are you mainly objecting to the description/commit message or the fix itself?

I like the simplicity of this fix, even if more efficient ones would certainly be possible.

lgao4:
If no other comment, I will add push label this week.

PR #12040, opened 2026-01-21
MdeModulePkg: PciBusDxe: Degrade MEM64 to PMEM64 when bridge lacks MEM64

kuqin12:
# Description

P2P bridge Memory Base/Limit registers (0x20-0x22) only support 32-bit addresses. Previously, MEM64 resources behind such bridges were always degraded to MEM32, forcing allocation below 4GB. This can lead to resource constraints when using massive storage (such as MPF drives) as MEM32 space is limited on most platforms.

Based on PCIe Base Specification 6.3+, assigning 64-bit resources irrespective of the prefetchable/non-prefetchable BAR bit is allowed.

In DegradeResource(), if an upstream bridge supports PMEM64, MEM64 resources are now degraded to PMEM64 first. This enables 64-bit BAR allocation for devices like NVMe controllers that declare non- prefetchable 64-bit BARs.

We still fall back to MEM32 degradation if the bridge lacks PMEM64 support as well.

Ref: PCI-SIG ECN "Removing Prefetchable Terminology" (2024-04-05)
Ref: PCIe Base Specification 6.3+

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on physical and virtual X64 and AArch64 platforms and booted to Windows desktop.

## Integration Instructions

For platforms needing such promotion behavior, set `gEfiMdeModulePkgTokenSpaceGuid.PcdPciDegradeMem64toPMem64` to **TRUE**


mdkinney:
Why is a PCD added?  If this is a recommended algorithm from PCI Spec, shouldn't that be the new behavior?

kuqin12:
> Why is a PCD added? If this is a recommended algorithm from PCI Spec, shouldn't that be the new behavior?

This is intended to preserve behavioral backward compatibility. While the spec does recommend this algorithm, my concern is that devices in the field may not fully support the change or may exhibit unexpected behavior. The PCD was therefore introduced to allow reverting to the previous behavior if needed. We can switch the default PCD to the recommended algorithm if desired.

mdkinney:
I prefer to not add PCDs and instead work with edk2 downstream consumers to help determine if the algorithm change will break or not.

kuqin12:
> I prefer to not add PCDs and instead work with edk2 downstream consumers to help determine if the algorithm change will break or not.

I understand and agree with the preference, which is why we validated this change across the platforms we have access to (server and client, x64 and AArch64). Given that this change is scoped to bridges that support only PMEM64, the expected impact should be limited, and based on our testing results, we do not see issues with removing the PCD.

That said, our testing does not cover the full range of PCIe device vendors in the field. To mitigate this risk, we are proposing to retain a PCD as a safeguard. How would you suggest we proceed?

kuqin12:
> That said, our testing does not cover the full range of PCIe device vendors in the field. To mitigate this risk, we are proposing to retain a PCD as a safeguard. How would you suggest we proceed?

The PCD was removed for simplicity.

kuqin12:
@lgao4 @jkolakow Could you please review and let me what you think of the change? Any input is appreciated.

lgao4 (approved):

mdkinney:
@kuqin12 Some OS compatibility issues have been observed with this change.  Recommend adding PCD.  DEC default setting should be for new behavior with option to fall back to previous behavior with Platform DSC setting.  Recommend PCD be added to DEC section [PcdsFixedAtBuild, PcdsPatchableInModule, PcdsDynamic, PcdsDynamicEx]/

PR #12041, opened 2026-01-22
MdeModulePkg: BdsDxe: Introduce infinite boot retries

berlin-with0ut-return:
# Description

This PR introduces a new feature to enable infinite boot retries based on a newly created PCD.
When true, the system will continuously loop over all boot options.

PCD default is FALSE to match existing functionality.

This change is tested on QEMU based virtual platforms and physical
platforms.

This change is useful for certain server cases. Infinite retries allows
a server to continuously attempt boot in case of network failure and
recovery, and for such attempts to be accurately recorded in the TCG
logs.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on QEMU based virtual platforms and physical platforms.

## Integration Instructions

N/A. 
Platforms which need infinite boot retries should enable the PCD.


berlin-with0ut-return:
@spbrogan and @kuqin12 also have context

mikebeaton:
Can you clarify the real-world examples where this is useful to enable?

apop5:
> Can you clarify the real-world examples where this is useful to enable?

A server in a data center where the system takes tens of minutes to boot.  The server should keep attempting to boot the boot options in the system until the watchdog resets the system, or the network comes back online and the server can boot.

The infinite retrying of boot options needs to exist in the Bds side so that TCG logs accurately reflect the boot attempts, etc.

mikebeaton:
> > Can you clarify the real-world examples where this is useful to enable?
> 
> A server in a data center where the system takes tens of minutes to boot. The server should keep attempting to boot the boot options in the system until the watchdog resets the system, or the network comes back online and the server can boot.
> 
> The infinite retrying of boot options needs to exist in the Bds side so that TCG logs accurately reflect the boot attempts, etc.

Makes sense, I wonder if some brief version of the above wording could be added to the commit message? (If you/other reviewers agree that might be helpful.)

makubacki:
@berlin-with0ut-return, thank you for pushing the updated commit message. Just fyi that I updated the PR description with it as well.

mikebeaton (approved):

berlin-with0ut-return:
@lgao4 @dandanbi please review the above changes

lgao4 (approved):

PR #12046, opened 2026-01-22
StandaloneMmPkg: StandaloneMmCore: Don't Print Entry/Exit Logs

os-d:
# Description

Currently, StandaloneMmCore prints for every MM entry and exit as well as producing protocols for MM entry and exit that get logged. So there are four logs for every MM entry/exit pair.

In a sample OVMF run with StMM enabled, this produced 6,800 logs to log MM entry/exit out of 16,400 logs. In other words, > 1/3 of logs were for these prints.

In general, entry and exit logs are not that useful outside of very specific debugging scenarios. Protocol install notifications are useful, but not for these ones that are just entry and exit notifies.

This commit drops the entry/exit prints and only logs at verbose level if the protocol installed is the entry or exit protocol.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Booting OVMF with StMM to shell.

## Integration Instructions

N/A.


StandaloneMmPkg/Core/Handle.c
```diff
@@ -233,9 +233,15 @@ MmInstallProtocolInterfaceNotify (
...
+  // Print debug message unless installing MM entry/exit notify protocol as that spams the log
   //
-  DEBUG ((DEBUG_LOAD | DEBUG_INFO, "MmInstallProtocolInterface: %g %p\n", Protocol, Interface));
+  if (!CompareGuid (Protocol, &gEfiMmEntryNotifyProtocolGuid) &&
+      !CompareGuid (Protocol, &gEfiMmExitNotifyProtocolGuid))
+  {
+    DEBUG ((DEBUG_LOAD | DEBUG_INFO, "MmInstallProtocolInterface: %g %p\n", Protocol, Interface));
+  } else {
+    DEBUG ((DEBUG_VERBOSE, "MmInstallProtocolInterface: %g %p\n", Protocol, Interface));
+  }
```
ardbiesheuvel, line 244: Why is this a single patch?

os-d: I'll split it into 2. To me it is the same thing, getting rid of entry/exit prints, but it's fair to say these are different versions of that and should be two commits.

StandaloneMmPkg/Core/Handle.c
```diff
@@ -233,9 +233,15 @@ MmInstallProtocolInterfaceNotify (
...
 
   //
-  // Print debug message
+  // Print debug message unless installing MM entry/exit notify protocol as that spams the log
   //
-  DEBUG ((DEBUG_LOAD | DEBUG_INFO, "MmInstallProtocolInterface: %g %p\n", Protocol, Interface));
+  if (!CompareGuid (Protocol, &gEfiMmEntryNotifyProtocolGuid) &&
+      !CompareGuid (Protocol, &gEfiMmExitNotifyProtocolGuid))
+  {
+    DEBUG ((DEBUG_LOAD | DEBUG_INFO, "MmInstallProtocolInterface: %g %p\n", Protocol, Interface));
```
kuqin12: nit: Do you want to print something different compared to the other generic case? (because you can)

niruiyu: Yes I agree to explicitly highlight that MmEntry and MmExit installation is not printed here otherwise a future debugger may be lost.

os-d: Sure, I can update the print

StandaloneMmPkg/Core/Handle.c
```diff
@@ -233,9 +233,15 @@ MmInstallProtocolInterfaceNotify (
   }
 
   //
-  // Print debug message
+  // Print debug message unless installing MM entry/exit notify protocol as that spams the log
```
ardbiesheuvel, line 236: Apologies for being slow, but could you explain why this is a special case wrt log spamming? Are you saying a protocol is being installed for every entry/exit into MM? If that is the case, surely there is a bigger problem with the design than the logging.

niruiyu: @ardbiesheuvel following code runs on every SMM entry:
```c
  Status   = MmInstallProtocolInterface (
               &MmHandle,
               &gEfiMmEntryNotifyProtocolGuid,
               EFI_NATIVE_INTERFACE,
               NULL
               );
  if (!EFI_ERROR (Status)) {
    MmUninstallProtocolInterface (
      MmHandle,
      &gEfiMmEntryNotifyProtocolGuid,
      NULL
      );
  }
```

Initially, I did not like this approach, but the request to be notified of SMI entry and exit is present.

ardbiesheuvel: Wow that is absolutely terrible. Every entry/exit involves pool (de)allocations and linked list traversals, to install a protocol that does not appear to have any users in the tree.

Could we perhaps optimize this? I.e., could we have a special case for `gEfiMmEntryNotifyProtocolGuid` where we emit the protocol notifications directly upon entry/exit, rather than going through the protocol database? Given that there is no actual protocol being installed, it should not matter for any existing users [if there are any] but it would remove this awful wart from the code.

niruiyu: This is an idea for optimization. But, we may need to measure how many us/ms are spent for the MmEntry/MmExit installation/uninstallation.

ardbiesheuvel: ok

os-d:
@niruiyu I updated the print, please re-review

StandaloneMmPkg/Core/Handle.c
os-d: Agreed this pattern should be fixed, but tbh this PR came out of just being annoyed at having to scroll through all these lines in the OVMF log, so this is a minimum annoyance fix PR. Certainly if a different direction is desired, I can do that.

ardbiesheuvel: Sure. Let's merge this first, and then think about a proper fix.

PR #12047, opened 2026-01-22
BaseTools: fix building with GCC16's new unused-but-set-variable

### LINKED ISSUE #12043: [Bug]: edk2:Shell won't build with GCC 16 with -Werror

### Bug Type

- [ ] Firmware
- [x] Tool
- [ ] Unit Test

ShellPkg

RELEASE

### Current Behavior

edk2:shell doesn't build with GCC 16 when using -Werror:

[   29s] StringFuncs.c: In function ‘SplitStringByWhitespace’:
[   29s] StringFuncs.c:113:15: error: variable ‘Item’ set but not used [-Werror=unused-but-set-variable=]
[   29s]   113 |   UINTN       Item;
[   29s]       |               ^~~~
[   29s] cc1: all warnings being treated as errors

Starting with GCC 16, var pre/postincrements and var @= expr no longer count as uses at some warning levels. Suggested fix:
 - Remove the unused variable.
 - Add some uses (void casts, etc.) 

See the porting to for more details: https://gcc.gnu.org/gcc-16/porting_to.html

Full build log: https://build.opensuse.org/package/live_build_log/home:rguenther:plgrnd/edk2:Shell/openSUSE_Factory/x86_64
This bug was originally opened in openSUSE Bugzilla: https://bugzilla.suse.com/show_bug.cgi?id=1256964

### Expected Behavior

No warnings during build

### Steps To Reproduce

Build the shell package with GCC 16, -Wunused-but-set-variable and -Werror

### Build Environment

```markdown
- OS(s): OpenSUSE Tumbleweed
- Tool Chain(s): GCC 16 git:r16-6733
```

### Version Information

```text
Tag: edk2-stable202502
```

### Urgency

Low

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

_No response_

dfaggioli:
GCC 16 has a strictier approach to unused variables (documented at https://gcc.gnu.org/gcc-16/porting_to.html). And, in fact, Item is never actually read, so we can just get rid of it.

Fixes: #12043

### DIFF
--- a/BaseTools/Source/C/Common/StringFuncs.c
+++ b/BaseTools/Source/C/Common/StringFuncs.c
@@ -110,7 +110,6 @@ SplitStringByWhitespace (
   CHAR8       *EndOfSubString;
   CHAR8       *EndOfString;
   STRING_LIST *Output;
-  UINTN       Item;
 
   String = CloneString (String);
   if (String == NULL) {
@@ -120,7 +119,8 @@ SplitStringByWhitespace (
 
   Output = NewStringList ();
 
-  for (Pos = String, Item = 0; Pos < EndOfString; Item++) {
+  Pos = String;
+  while (Pos < EndOfString) {
     while (isspace ((int)*Pos)) {
       Pos++;
     }


khaliid2040:
Please link the PR with the issue probably so no one forgets to close it later.

Change 
`Fixes issue #12043.`
To
`Fixes: #12043`

Also apply it on the commit message.

dfaggioli:
> Change `Fixes issue #12043.` To `Fixes: #12043`
> 
> Also apply it on the commit message.

Ok, done... I think :-)

BaseTools/Source/C/Common/StringFuncs.c
```diff
@@ -120,7 +119,7 @@ SplitStringByWhitespace (
 
   Output = NewStringList ();
 
-  for (Pos = String, Item = 0; Pos < EndOfString; Item++) {
+  for (Pos = String; Pos < EndOfString; ) {
```
mikebeaton: I think this would be more idiomatic as:
```
  Pos = String;
  while (Pos < EndOfString) {
```

dfaggioli: Done (and thanks for looking at the patch)

mikebeaton (approved):

luigix25 (approved):

mikebeaton:
This was fixed in https://github.com/tianocore/edk2/pull/12079 in the end, which fixed several similar issues.

mikebeaton:
This is fixed elsewhere, can it be closed to silence ongoing notifications?

Issue #11764, opened 2025-11-17
[Bug]: SmiHandlerUnRegister  is pended forever when an unregistered SMI get called


KaiYehHp:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

MdeModulePkg

### Which targets are impacted by this bug?

RELEASE

### Current Behavior

impacted file
MdeModulePkg/Core/PiSmmCore/Smi.c

In the commit https://github.com/tianocore/edk2/commit/74f6ce67343f81a842e6af3c599e4ae2a07ee51b
SmiHandlerUnRegister will hold the unregistration until all SMI get processed(mSmiManageCallingDepth reach 0).

**the problem is** 
When we generate a call to SmiManage()  with a HandlerType that haven't been registered or a random GUID. SmmCoreFindSmiEntry  will return NULL on line 167 and then return on line 173. and now the mSmiManageCallingDepth is "1".  There is no way mSmiManageCallingDepth back to 0 anymore. then the delayed unregistration from line 266 will never entered. 

It's because the mSmiManageCallingDepth  only get minus 1 when a valid SMI being processed.

and the consequence is the SMI we expected to be unregistered actually still alive and can be called.

### Expected Behavior

1. mSmiManageCallingDepth  should be only increased when a valid SMI allowed to be called. So the line 266 can be entered correctly when it's the on the outermost SmiManage()
2. SMI gets unregistered can NOT be called successfully anymore

### Steps To Reproduce

1. calling SMI with a random HandlerType. As long as the HandlerType  is not registered through the SmiHandlerRegister (). you will see mSmiManageCallingDepth   remain 1 forever
2. any SmiHandlerUnRegister () will return EFI_SUCCESS but actually it's queued and never processed because mSmiManageCallingDepth   never reach 0
3. You will be still able to call the SMI call even it has been calling SmiHandlerUnRegister() 

### Build Environment

```markdown
- OS(s): Windows
- Tool Chain(s):VS2019(should be all impacted)
```

### Version Information

```text
74f6ce67343f81a842e6af3c599e4ae2a07ee51b
b40c64ec25986d36e1fd02304daa9d72ac863da6
```

### Urgency

High

### Are you going to fix this?

Someone else needs to fix it

### Do you need maintainer feedback?

Maintainer feedback requested

lgao4:
[...]

PR #12056, opened 2026-01-24
MdeModulePkg,StandaloneMmPkg: Fix SMI handler unregistration issue

khaliid2040:
# Description

Fixes: #11764 

Currently, on both PiSmmCore and StandaloneMmPkg, always unconditionally increment mSmiManageCallingDepth and mMmiManageCallingDepth respectively. However it will decremented only if SMI is valid.

This causes those two variables to get incremented always even when SMI/MMI isn't installed, but never get decremented on return when SMI handler isn't registered . This makes those two variables to grow depending on the SMI as long as the handler didn't found by SmmCoreFindSmiEntry() .

Also when mSmiManageCallingDepth is greater than zero SmiHandlerUnregister () will skip removing SMI handler and let SmiManage ()  handle it. However, on SmiManage () when mSmiManageCallingDepth is greater than zero SMI unregisteration will be skip too. Because, if SMI is triggered however the handler isn't registered then mSmiManageCallingDepth will never get decremented. The same issue is observed on StandaloneMmPkg implementation of that function (MmiManage ()).

## Summary

If there is a single call to SmiManage () or MmiManage () and SMI handler didn't found, then SMIs will never get unregistered because the SmiManage () and SmiHandlerUnregister () both, the condition of :
```C
if (mSmiManageCallingDepth == 0) {
...
}
```
will evaluate as false.

Same for StandaloneMmPkg for those implementations condition:
```C
if (mMmiManageCallingDepth == 0) {
...
}
```

Will evaluate as false

## Solution
Move mSmiManageCallingDepth incremental before for-loop just below else block.

Please for more details read the linked issue.
- [ ] Breaking change?
  - NO
- [ ] Impacts security?
  - NO
- [ ] Includes tests?
  - NO

## How This Was Tested
On OVMF qemu by setting SMM_REQUIRE=TRUE and STANDALONE_MM_ENABLE=TRUE.

## Integration Instructions
N/A


khaliid2040:
@niruiyu I finished please review.

lgao4:
Seemly, this is a good fix. 

@kuqin12 , could you help review it?

kuqin12 (approved):

khaliid2040:
@lgao4, @ardbiesheuvel and @niruiyu 

I think the project entered soft freeze period. The urgency of the bug this PR is fixing is high. Is it possible to get merged before hard freeze ?

lgao4:
> @lgao4, @ardbiesheuvel and @niruiyu
> 
> I think the project entered soft freeze period. The urgency of the bug this PR is fixing is high. Is it possible to get merged before hard freeze ?

I agree.

khaliid2040:
@niruiyu and @ardbiesheuvel could this PR get merged if there is nothing to change.

niruiyu (approved):

PR #12057, opened 2026-01-24
UefiPayloadPkg,SecurityPkg: add OPAL S3 lockbox handoff

Sean-StarLabs:
This refreshed branch carries the current OPAL S3 unlock flow:

- add optional OPAL password support in `UefiPayloadPkg`
- build the OPAL S3 LockBox record at `ReadyToBoot`
- share the LockBox record definition between payload and driver code
- route the LockBox handoff through the payload SMM path
- refresh the LockBox contents when the OPAL password changes

The goal is to preserve the OPAL password across S3 in a form the firmware
resume path can use without prompting again.

### DIFF
--- a/UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
+++ b/UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
@@ -9,6 +9,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 
 #include "PlatformBootManager.h"
 #include "PlatformConsole.h"
+#include <Library/Tcg2PhysicalPresenceLib.h>
 #include <Protocol/FirmwareVolume2.h>
 
 /**
@@ -319,6 +320,11 @@ PlatformBootManagerAfterConsole (
     gST->ConOut->ClearScreen (gST->ConOut);
     BootLogoEnableLogo ();
   }
+
+  //
+  // Ensure TCG2 physical presence variables are initialized (required for OPAL BlockSID UI).
+  //
+  Tcg2PhysicalPresenceLibProcessRequest (NULL);
 
   EfiBootManagerConnectAll ();
   EfiBootManagerRefreshAllBootOption ();
--- a/UefiPayloadPkg/UefiPayloadPkg.dsc
+++ b/UefiPayloadPkg/UefiPayloadPkg.dsc
@@ -45,6 +45,7 @@
   DEFINE NVME_ENABLE                  = TRUE
   DEFINE LOCKBOX_SUPPORT              = FALSE
   DEFINE LOAD_OPTION_ROMS             = FALSE
+  DEFINE OPAL_PASSWORD_ENABLE         = FALSE
 
   #
   # Capsule updates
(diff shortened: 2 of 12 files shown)


gguo11837463 (dismissed):

gguo11837463:
Please don't touch multiple package in one commits

benjamindoron (changes requested):
First: For reference, this is supposed to be integrated against the patch series beginning with https://review.coreboot.org/c/coreboot/+/91042. Also, thanks for working on this.

Secondly: To my knowledge, the typical flow is implemented with the LockBox, wherein the DXE driver calls a SMI handler in the lockbox driver (not actually included by UefiPayload) to deposit the password, for it to be fetched by the PEIM on S3 resume. Therefore, OpalS3PasswordLibApmsmi is implementing a duplicate path to achieve what native code already has, and I think we should carefully consider whether that's necessary first.
- It's also possible that the LockBox is actually held outside of SMRAM until `DxeSmmReadyToLock`, for it to be easily modified by the LockBox library and driver, and is brought inside SMRAM when this event is fired. It can then continue to be worked with by SMM code, and on S3 resumes, PEIMs will simply read the now-unlocked SMRAM. It's been a while since I worked with the LockBox while I developed S3 support for MinPlatform, but this is basically the gist.

So, by the way, you don't actually need the real version of the lockbox library in the current PR. I'm not even sure why we'd ever use the real version of LockBoxLib without the SMM driver, and without something in the firmware to use it later.

Taken together, OpalS3PasswordLibApmsmi and LockBoxLib being duplicate approaches to achieve the same thing, and coreboot not needing a real LockBoxLib... why not (ab)use the LockBoxLib API to implement what's needed for coreboot? The function containing https://github.com/tianocore/edk2/blob/072ab3846c4fd86387dfd5d191dcb91e98202023/SecurityPkg/Tcg/Opal/OpalPassword/OpalDriver.c#L393-L398 builds up similar data to the library you wrote. If you filter on GUIDs you care about (so that we can present other callers with the behaviour of 'LockBoxLibNull'), we can isolate the coreboot-specific behaviour to a coreboot-specific library. And hey, the coreboot patches are kind of lockbox-esque.

UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
```diff
@@ -320,6 +321,11 @@ PlatformBootManagerAfterConsole (
     BootLogoEnableLogo ();
   }
 
+  //
+  // Ensure TCG2 physical presence variables are initialized (required for OPAL BlockSID UI).
```
benjamindoron, line 325: Did we never land upstream TPM support/PPI support? Apparently not. I suppose this comment (specifically "required for OPAL") is fine then.

UefiPayloadPkg/UefiPayloadPkg.dsc
```diff
@@ -353,13 +354,27 @@
 !endif
 
   DebugLib|MdeModulePkg/Library/PeiDxeDebugLibReportStatusCode/PeiDxeDebugLibReportStatusCode.inf
-!if $(LOCKBOX_SUPPORT) == TRUE
+!if $(LOCKBOX_SUPPORT) == TRUE || $(OPAL_PASSWORD_ENABLE) == TRUE
```
benjamindoron: Which instance of the LockBoxLib should be used comes down to platform support, not whether we have a feature that demands/wants it. Also, as I'm saying in the bigger comment, this is dead code anyways: all that this instance can do is try to call the module in SMM, which isn't included in any UefiPayload build.

benjamindoron:
First, thank you for making those changes, helping to localise coreboot-specific quirks here.

Secondly, from a functionality perspective, is [SecurityPkg/Tcg/Opal: keep LockBox in sync with password](https://github.com/tianocore/edk2/commit/9d45b8f8cc52f51f03e3bc647455cc2b189b4946) needed? It's not going to work for native platforms.

In the standard LockBox design, the following are capable of working with the LockBox:
- The SMM/MM instance of LockBoxLib. This can be seen as the main 'worker' of the feature. It's mainly consumed by the SmmLockBox driver, but other SMM/MM modules can consume it too
- The SmmLockBox driver. It installs an SMI handler for other phases to work with the LockBox
  - Critically, it locks itself at SmmReadyToLock, and will refuse any calls that come in after that
- The DXE instance of LockBoxLib. This uses the SmmCommunication protocol to call the module in SMM
- The PEI instance of LockBoxLib. It will use SmmCommunication if it can too, but if not, it will attempt to read SMRAM itself. This works on the S3 resume path, which is what the library instance for

Therefore, the OPAL driver can't use the LockBox after EndOfDxe, it's not supposed to work anymore. After ReadyToLock, only an SMM driver can update the LockBox.

With regards to [UefiPayloadPkg: coreboot LockBox via APMC SMI](https://github.com/tianocore/edk2/commit/18b871b9c02bb94c048157bd882ffc59ebc06752), I'm a little concerned about security: the LockBox design and the standard implementation are capable of offering a guarantee of confidentiality, integrity or both, to anything stored inside it, whereas maintaining an in-DXE LockBox means that any ring-0 code could read the contents. While you'll see that the SmmLockBox SMI handler allows the codepath into LockBoxLib to be entered for the `Restore*()` functions even after it's locked, LockBoxLib will check attributes on it, and the OPAL driver does set https://github.com/tianocore/edk2/blob/072ab3846c4fd86387dfd5d191dcb91e98202023/SecurityPkg/Tcg/Opal/OpalPassword/OpalDriver.c#L400-L404.

I'd much prefer to genuinely abuse the LockBoxLib API to accomplish only what's needed here:
- For `SaveLockBox()`, check if the Guid is `mOpalDeviceLockBoxGuid`, and if not, return. For OPAL, send it to coreboot
- For `SetLockBoxAttributes()`, eh, it's not really possible. Return unsupported, probably
- For `UpdateLockBox()`, well, what I said above applies, so the API isn't used by OPAL at the moment. But if we sort that out, then, if coreboot supports being given an update, then send OPAL lockboxes to coreboot. If not, always return immediately
- For the `Restore*()` functions, always return. At least `gS3StorageDeviceInitListGuid`  must be told that it's not found

I've inspected the code, and I believe this will work

PR #12058, opened 2026-01-24
OvmfPkg: Introduce Tcg2PhysicalPresencePlatformLib

Sean-StarLabs:
Move the QEMU TPM PPI config retrieval into a platform library class so Tcg2PhysicalPresenceLibQemu can be reused across platforms.

Also add a user-confirmation timeout to avoid hanging indefinitely on headless systems.

Update to more user-friendly names


kraxel:
One change per commit please.  Also the mostly unrelated help string fix is probably best moved to a separate PR.

PR #12063, opened 2026-01-27
MdeModulePkg HobPrintLib: Also print EFI_HOB_TYPE_UNUSED type HOB

lzeng14:
EFI_HOB_TYPE_UNUSED is not unknown HOB type, instead of printing "Unknown Hob type, ...", this patch updates code to also print EFI_HOB_TYPE_UNUSED type HOB.


MdeModulePkg/Library/HobPrintLib/HobPrintLib.c
khaliid2040, line 388: I think this array should be moved at the top of the source file. This where all global variables declared in the entire codebase.

lzeng14: Thanks for the comment. Putting it at the top of the source file will make all the referred functions need to be declared first.

Anyway, that is out of scope for this PR. :)

lzeng14:
@lgao4 Liming, could you help merge it ?

Issue #11113, opened 2025-05-23
[Bug]: InstallProtocolInterface() does not update the key of an existing handle


rpgrahamjr941:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

MdeModulePkg

### Which targets are impacted by this bug?

RELEASE, DEBUG

### Current Behavior

InstallProtocolInterface() sets the key of a new handle it creates but does not update the key of an existing handle. ReinstallProtocolInterface() and UninstallProtocolInterface() both update the key, so presumably Install should as well.

Using the attached driver, the debug output includes
```
Bug3EntryPoint called
  Image handle key initially is 96
InstallProtocolInterface: 5C198761-16A8-4E69-972C-89D67954F81D 440F9440
  Image handle key after installation is 96
  Image handle key after reinstallation is 97
  Image handle key after uninstallation is 98
```

### Expected Behavior

One would expect the key to be incremented by the install operation. If the expected behavior is _not_ to update the key, then the UEFI Specification should indicate this in Sections 7.3.1 and 7.3.2 and explain why.

### Steps To Reproduce

1. Create the folder `edk2/EmulatorPkg/Demo1_Bugs`
2. Add the attached files to `edk2/EmulatorPkg/Demo1_Bugs`
3. Add this line to the `[Components]` section of `EmulatorPkg/EmulatorPkg.dsc`:
```
EmulatorPkg/Demo1_Bugs/Demo1_Bug3.inf
```
4. Add this line to the `[FV.FvRecovery]` section of EmulatorPkg/EmulatorPkg.fdf:
```
INF EmulatorPkg/Demo1_Bugs/Demo1_Bug3.inf
```
5. cd to the `edk2` folder
6. run
```
   $ EmulatorPkg/build.sh
   $ EmulatorPkg/build.sh run 2> ../Debug_bugs.txt
```
7. Press `ENTER` or wait for the UEFI Shell screen
8. Press `ENTER` or wait for the UEFI Shell command-line prompt
9. Type `exit`
10. Select `Reset` from the EmulatorPkg screen
11. Examine `../Debug_bugs.txt` and look for `Bug3EntryPoint`

[Demo1_Bug3.inf.txt](https://github.com/user-attachments/files/20416799/Demo1_Bug3.inf.txt)
[...]

PR #12069, opened 2026-01-27
MdeModulePkg/Core: Fix key incremental issue on existing handle

khaliid2040:
# Description

   Fixes: #11113

Currently, the global handle key and key inside handle structure is
incremented only when a new handle is allocated for protocol interface
to be installed.
    
However, when caller already supplies a handle gHandleDatabaseKey never
get incremented to reflect .
    
Move handle key incremental outside if block, just below the else
statement which allows gHandleDatabaseKey to always incremented whether
handle is supplied or not.

- [ ] Breaking change?
 NO
- [ ] Impacts security?
  NO
- [ ] Includes tests?
NO

## How This Was Tested
N/A
## Integration Instructions
N/A


lgao4:
@mdkinney ,  Mike, I think this patch is a good fix to make sure handle be recorded when the same handle has been installed by the additional protocol. Can you help confirm this change?

khaliid2040:
@mdkinney could you please review this patch if you have chance.

khaliid2040:
@lgao4 this PR was open for a while now. I would like you to merge if you are okay with the changes.

MdeModulePkg/Core/Dxe/Hand/Handle.c
```diff
@@ -539,6 +533,12 @@ CoreInstallProtocolInterfaceNotify (
     }
   }
 
+  //
+  // Initialize the Key to show that the handle has been created/modified
```
mikebeaton: This comment should be updated to:

`Initialize/update the Key to show that the handle has been created/modified`

mikebeaton:
I believe UnitTestInstallProtocolInterfaceNotify needs modifying to match.

MdeModulePkg/Core/Dxe/Hand/Handle.c
khaliid2040: Actually, gHandleDatabaseKey is already pre-initialized on top. So, this is just update for gHandleDatabaseKey and handle key. It is better if we  replace to Update instead of initialize like:

`Update the Key to show that the handle has been created/modified`

All other places where gHandleDatabaseKey is incremented use that comment above. 

Let me know if you are okay with that comment above so i could update.

khaliid2040:
> I believe UnitTestInstallProtocolInterfaceNotify needs modifying to match.

Thanks, i updated UnitTestFrameworkPkg too.

MdeModulePkg/Core/Dxe/Hand/Handle.c
mikebeaton: I don't believe that in the comment "the Key" refers to `gHandleDatabaseKey`. It refers to `Handle->Key`. It makes sense if you look at how `->Key` is used (i.e. to decide which handles have been updated recently). So we are indeed "Initializing" `Handle->Key` in the case where `Handle` was "created", and "updating" it in the case where Handle was "modified", hence my suggestion.

khaliid2040: You are right, in this place there is a handle creation while other places just update the key of existig handle. 

I updated the comment too

lgao4 (approved):

PR #12070, opened 2026-01-28
Introducing host based unit tests on AArch64 systems

kuqin12:
# Description

This change adds the support for building and running host based unit tests on AArch64 Linux systems.

The change currently only supports the operations using GCC as the compiler. Support for CLANG is pending.

Known issues:

- Subhook does not support AArch64 binaries, thus function pointer patching are not supported when building on AArch64 platforms.
- `RngLib` for AArch64 target will check on to determine the availability of RDRAND instruction, which may or may not be supported depending on . It is suggested to use the timer based RngLib for host tests.
- We might need a legit `TimerLib` for host based unit test to resolve the issue above.
- The packages that depends on RngLib are not onboarded to support AArch64 host tests. Including `CryptoPkg`, `SecurityPkg`, and `UefiCpuPkg`.

- [ ] Breaking change?
- [ ] Impacts security?
- [x] Includes tests?

## How This Was Tested

This was tested on a AArch64 host system running Ubuntu 24.04.

## Integration Instructions

N/A


kuqin12:
Not stale

UnitTestFrameworkPkg/Test/GoogleTest/Sample/SampleGoogleTest/SampleGoogleTest.cpp
```diff
@@ -563,7 +563,12 @@ TEST (SanitizerTests, DivideByZeroDeathTest) {
   //
   // Divide by 0 should be caught by address sanitizer, log details, and exit
   //
+ #if !defined (MDE_CPU_AARCH64) || !defined (__GNUC__)
   EXPECT_DEATH (DivideWithNoParameterChecking (10, 0), "ERROR: AddressSanitizer: ");
```
mdkinney: What does AARCH64 application do when divide by 0?  x86 generated an exception which is caught with EXPECT_DEATH()

kuqin12: It just returns 0 on AArch64.

kuqin12: It seems that the trap is only available on M profiles. And only VC has the compiler injected artificial traps, whereas GCC does not have the counterpart.

mdkinney: @ardbiesheuvel This seems very odd to me that divide by zero does not raise an exception or signal for all compilers.  Especially for Host based unit test build env that build OS applications.  Are there specific compiler flags needed for VS, GCC, CLANG for AARCH application builds to guarantee that an exception/signal is raised that can be caught by EXPECT_DEATH().  These would be added to UnitTestFrameworkPkgHost.dsc.inc.

ardbiesheuvel: The underlying `UDIV`/`SDIV` instructions simply return `0` when the divisor is `0`. There is no portable way to force the compiler to emit such a check (`-mcheck-zero-division` exists in GCC but is not supported on the AArch64 target).

The upshot of that is that those instructions can never trap in the actual code either, and so adding our own trapping versions just for the unit tests seems rather pointless.

mdkinney: That is a very interesting difference in CPU archs.

With that info, I would claim that the unit test to verify EXPECT_DEATH() on divide by zero is invalid on AARCH64, and that specific unit test case should be split out into an IA32/X64 specific file.

UnitTestFrameworkPkg/Library/UnitTestDebugAssertLib/UnitTestDebugAssertLibHost.inf
```diff
@@ -30,3 +30,4 @@
   MSFT:*_*_*_CC_FLAGS   == /c /Zi /Od
   GCC:*_*_IA32_CC_FLAGS == -g -c -fshort-wchar -fexceptions -O0 -m32 -malign-double -fno-pie
   GCC:*_*_X64_CC_FLAGS  == -g -c -fshort-wchar -fexceptions -O0 -m64 -fno-pie "-DEFIAPI=__attribute__((ms_abi))"
+  GCC:*_*_AARCH64_CC_FLAGS  == -g -c -fshort-wchar -fexceptions -O0 -fno-pie "-DEFIAPI=__attribute__((ms_abi))"
```
ardbiesheuvel: `__attribute__((ms_abi))` is specific to x86_64, so I'm pretty sure we don't need it here.

UnitTestFrameworkPkg/Test/GoogleTest/Sample/SampleGoogleTest/SampleGoogleTest.cpp
kuqin12: Thanks, this is updated.

UnitTestFrameworkPkg/Library/UnitTestDebugAssertLib/UnitTestDebugAssertLibHost.inf
kuqin12: Good catch. This is now removed.

kuqin12: I also added a new commit to remove the one for X64 targets.

kuqin12: > I also added a new commit to remove the one for X64 targets.

Overlooked the flag usage. It is still needed for x64 targets. Removed the wrong commit.

MdePkg/Library/BaseLib/UnitTestHostBaseLib.inf
```diff
@@ -178,12 +177,10 @@
   Ebc/SetJumpLongJump.c
   Ebc/SwitchStack.c
   Ebc/SpeculationBarrier.c
-  Unaligned.c
   Math64.c
 
 [Sources.AARCH64]
   AArch64/InternalSwitchStack.c
-  AArch64/Unaligned.c
```
pierregondois, line 186: NIT:
It seems that this patch does 2 things:
1. Add preprocessor for AArch64 instances
2. Cleanup some unnecessary files

Maybe this patch should be split in 2 and also remove the following unused sections:
- [Sources.RISCV64]
- [Sources.EBC]

------

NIT2:
The following function are not supported by the UnitTestHost BaseLib implementation:
- AArch64/ArmReadCntPctReg.S
- AArch64/ArmReadIdAA64Isar0Reg.S

It doesn't seem it is required right now and it will still be possible to add it later if needed, so just making a note of it.

------
NIT3:
Also, the `VALID_ARCHITECTURES` might be updated to add AARCH64

UnitTestFrameworkPkg/Test/GoogleTest/Sample/SampleGoogleTest/SampleGoogleTest.cpp
```diff
@@ -544,28 +544,6 @@ TEST (SanitizerTests, InvalidPointerWriteDeathTest) {
...
-  //
-  // Perform integer division with no check for divide by zero
-  //
-  return (Dividend / Divisor);
-}
-
-/**
-  Sample unit test that performs a divide by 0
-**/
-TEST (SanitizerTests, DivideByZeroDeathTest) {
```
pierregondois, line 562: It seems that the same division by 0 test exists in:
`UnitTestFrameworkPkg/Test/GoogleTest/Sample/SampleGoogleTestGenerateException/ampleGoogleTestGenerateException.cpp`

The test should generate an exception, but instead it should fail as the result is `0` (checked against `MAX_UINTN`). Maybe the test should be IA32/X64 only.

pierregondois (commented):
Hello Kun,
Just a few comments, otherwise I tried the branch locally and it seems to work well.

MdePkg/Library/BaseLib/UnitTestHostBaseLib.inf
kuqin12: Thanks for your comments and testing this.

I have updated this patch per your feedback. Specifically, I split the majority of the inf change into a separate commit and also removed the EBC and RISCV sections.

It was a good callout, but I did not add the CntPctReg and IdAA64Isar0Reg support in this patch set. I would like to add the timer related changes in a separate PR to resolve the lack of timer support on host-based tests. But please let me know if you have other thoughts.

UnitTestFrameworkPkg/Test/GoogleTest/Sample/SampleGoogleTest/SampleGoogleTest.cpp
kuqin12: I left this feedback out for this round of push. The test is intended to generate an exception, but divide by 0 will not trip on any exception on AArch64 platforms. Should we switch to read from a null pointer instead in this case? @mdkinney any thoughts?

On the other hand, this test is not included in the unit test dsc file, so we do not have obvious way to mark this test as x86 specific, other than the inf's supported architecture, which is already configured as such.

mdkinney: > I left this feedback out for this round of push. The test is intended to generate an exception, but divide by 0 will not trip on any exception on AArch64 platforms. Should we switch to read from a null pointer instead in this case? @mdkinney any thoughts?
> 
> On the other hand, this test is not included in the unit test dsc file, so we do not have obvious way to mark this test as x86 specific, other than the inf's supported architecture, which is already configured as such.

Most of this file can work for all archs.  Just this one test should be broken out into an x86 only cpp file and INF updated with main cpp file for all archs and the one special cpp file for IA32/X64 only.

kuqin12: I think Pierre is referring to this test: `UnitTestFrameworkPkg/Test/GoogleTest/Sample/SampleGoogleTestGenerateException/ampleGoogleTestGenerateException.cpp`, which only has one test and it throws an exception using a divide by zero exception. I think I will create a specific cpp file for AArch64 that generates exceptions differently.

pierregondois: As you wish, but as you suggested, shouldn't a NULL pointer dereference trigger an exception for all archs ?
Like that there is no need for arch specific code.

Also as you said this test is not part of any dsc file, so it will be hard to follow/control it

kuqin12: Agreed. I just updated the exception test to touch null pointer in the latest push, so that it will be applicable to all supported arch.

kuqin12:
@mdkinney @makubacki @spbrogan Could you please help to review the unit test framework changes in this PR once you have a chance? Thanks in advance!

pierregondois (approved):

kuqin12:
@mdkinney @makubacki @spbrogan Do you have any feedback from UnitTestFrameworkPkg perspective? Any feedback is appreciated.

UnitTestFrameworkPkg/Test/UnitTestFrameworkPkgHostTestExpectFail.dsc
```diff
@@ -35,14 +35,16 @@
...
+
   #
   # Disable warning for divide by zero to pass build of unit tests
   # that generate a divide by zero exception.
   #
-  UnitTestFrameworkPkg/Test/GoogleTest/Sample/SampleGoogleTestGenerateException/SampleGoogleTestHostGenerateException.inf {
-    <BuildOptions>
-      MSFT:*_*_*_CC_FLAGS = /wd4723
-  }
   UnitTestFrameworkPkg/Test/UnitTest/Sample/SampleUnitTestGenerateException/SampleUnitTestHostGenerateException.inf {
```
makubacki, line 48: Is divide-by-zero an issue in `SampleUnitTestHostGenerateException`?

https://github.com/tianocore/edk2/blob/3a4556803d3a02dc0b298c489013e04082a9617e/UnitTestFrameworkPkg/Test/UnitTest/Sample/SampleUnitTestGenerateException/SampleUnitTestGenerateException.c#L90-L93

UnitTestFrameworkPkg/Include/Library/FunctionMockLib.h
makubacki, line 74: Was there a reason to not conditionalize `MOCK_FUNCTION_HOOK_DEFINITIONS` on `(MDE_CPU_IA32) || defined (MDE_CPU_X64)` like `MOCK_FUNCTION_HOOK_DECLARATIONS`?

MdePkg/Library/BaseLib/AArch64UnitTestHost.c
```diff
@@ -0,0 +1,30 @@
+/** @file
+  IA32/X64 specific Unit Test Host functions.
```
makubacki:
```suggestion
  AARCH64 specific Unit Test Host functions.
```

UnitTestFrameworkPkg/Test/GoogleTest/Sample/SampleGoogleTest/SampleGoogleTestHost.inf
makubacki: To match updates in files like `SampleGoogleTestHostGenerateException.inf` and `CmockaLib.inf`.

```suggestion
#  VALID_ARCHITECTURES           = IA32 X64 AARCH64
```

UnitTestFrameworkPkg/Library/CmockaLib/CmockaLib.inf
makubacki: To match other INFs.

```suggestion
#  VALID_ARCHITECTURES           = IA32 X64 AARCH64
```

makubacki (approved):

UnitTestFrameworkPkg/Include/Library/FunctionMockLib.h
kuqin12: Good catch. It was overlooked because there was no real callers somehow. This is fixed.

UnitTestFrameworkPkg/Test/GoogleTest/Sample/SampleGoogleTest/SampleGoogleTestHost.inf
kuqin12: Udpated.

UnitTestFrameworkPkg/Library/CmockaLib/CmockaLib.inf
kuqin12: Updated.

MdePkg/Library/BaseLib/AArch64UnitTestHost.c
kuqin12: Updated.

UnitTestFrameworkPkg/Test/UnitTestFrameworkPkgHostTestExpectFail.dsc
kuqin12: Good catch. Added a new commit https://github.com/tianocore/edk2/pull/12070/changes/7728ec9736f7f261d608b81a9f09a58c0b496898 to cover this test case for AArch64.

ardbiesheuvel (approved):

kuqin12:
@lgao4 @mdkinney @AshrafAliS @YuweiChen1110 @ponchofigueroa @xuweiintel Would you mind providing feedback to this PR? Any input is appreciated~

jbrasen:
@kuqin12 Thanks for working on this. I didn't see your pending change before I posted mine and it looks like we have a similar stack here. I'll close mine in a bit and move my testing on top of yours.

The packages this PR defers (SecurityPkg / CryptoPkg). You note these as not-yet-onboarded due to RngLib. We have small changes that bring them up on GCC/AArch64 host tests:

- Map the host RngLib to BaseRngLibNull for AArch64 in the SecurityPkg host DSC (IA32/X64 keep BaseRngLib), avoiding the RNDR path that doesn't link in the host environment — jbrasen/edk2@3ff8f68df6cd02322d5f916640f4ae1605e51344 and jbrasen/edk2@02d9e0c2a12980d8b5c000f7a30e8dc756ab11cd
- Add the openssl-backed Rand/CryptRand.c for AArch64 to UnitTestHostBaseCryptLib.inf so BaseCryptLib links in AArch64 host tests — jbrasen/edk2@dd508d23d843dc8e529c7b76d877142acd0cb6ca

With these, the SecurityPkg suites (DxeImageVerification, Tpm/Tpm2 MeasuredBoot, SecureBootVariable) and FmpDevicePkg build and pass on aarch64 Ubuntu 24.04 / GCC.

One thing worth reconciling first: those SecurityPkg GoogleTest suites mock via FunctionMockLib, which this PR disables on AArch64. We instead added a no-op subhook stub so FunctionMockLib still compiles (function-pointer patching is simply a no-op on AArch64, matching subhook's lack of an AArch64 backend) — that's what lets the mocked SecurityPkg tests build. Either approach works for me.

Also MemoryBinGoogleTest fails on AArch64. This PR onboards MdeModulePkg, but BaseMemoryBinLibTest.PopulatesFromValidHob hard-codes 4 KiB page counts. PopulateMemoryTypeInformation rounds the runtime memory types up to RUNTIME_PAGE_ALLOCATION_GRANULARITY (64 KiB on AArch64), so e.g. 5 pages becomes 16 and the assertion fails. Since upstream CI doesn't run AArch64 host tests, it won't surface in CI. 
A granularity-aware fix: jbrasen/edk2@a5c7d4fb67e67da3a0835df87e31c76db630a2d4

os-d:
> Also MemoryBinGoogleTest fails on AArch64. This PR onboards MdeModulePkg, but BaseMemoryBinLibTest.PopulatesFromValidHob hard-codes 4 KiB page counts. PopulateMemoryTypeInformation rounds the runtime memory types up to RUNTIME_PAGE_ALLOCATION_GRANULARITY (64 KiB on AArch64), so e.g. 5 pages becomes 16 and the assertion fails. Since upstream CI doesn't run AArch64 host tests, it won't surface in CI. A granularity-aware fix: [jbrasen/edk2@a5c7d4f](https://github.com/jbrasen/edk2/commit/a5c7d4fb67e67da3a0835df87e31c76db630a2d4)

Thanks for the fix here, I agree.

jbrasen:
I validated these locally as well. I have a branch that has the 4 additional changes I had ontop of this one here https://github.com/jbrasen/edk2/tree/pr12070-plus-delta

jbrasen (approved):
Changes look good to me as well

kuqin12:
@jbrasen thanks for running this locally and provide the feedback.

The memory bin change is what I hope the unit test could catch when I draft the PR. Thanks for pointing this out.

I agree with the SecurityPkg and CryptoPkg changes. They suffice the mid-term needs. Although I do want to bring up a host-based TimerLib so that we will have a non-null version of RngLib to test with.

As per moving forward with this PR, given I have not received reviews from the impacted packages, I would not mind cherry-picking your changes and try to get this in all at once. But if you prefer to file a follow up PR from your side, that is fine as well. Please let me know :)

jbrasen:
> @jbrasen thanks for running this locally and provide the feedback.
> 
> The memory bin change is what I hope the unit test could catch when I draft the PR. Thanks for pointing this out.
> 
> I agree with the SecurityPkg and CryptoPkg changes. They suffice the mid-term needs. Although I do want to bring up a host-based TimerLib so that we will have a non-null version of RngLib to test with.
> 
> As per moving forward with this PR, given I have not received reviews from the impacted packages, I would not mind cherry-picking your changes and try to get this in all at once. But if you prefer to file a follow up PR from your side, that is fine as well. Please let me know :)

Either way works for me, your call on how you want to handle that.

kuqin12:
> Either way works for me, your call on how you want to handle that.

Thanks, cherry-picked your changes. Will keep monitoring the pipeline status while I do some local testing from my end as well.

kuqin12:
Tested locally and tests worked as expected.

os-d (approved):
Approved for MdeModulePkg/Core

cfernald (approved):
Approving for the minor SecurityPkg changes for aarch64 execution.

Flickdm (approved):

Flickdm:
Approved for CryptoPkg

kuqin12:
@lgao4 @mdkinney would you please providing feedback on this PR? Specifically, this PR still needs reviews on MdePkg, MdeModulePkg, FmpDevicePkg and BaseTools. Any input is appreciated. Thanks in advance!

MdePkg/Library/BaseLib/UnitTestHostBaseLib.inf
```diff
@@ -199,18 +189,6 @@
   AArch64/SpeculationBarrier.asm    | MSFT
   AArch64UnitTestHost.c
 
-[Sources.RISCV64]
-  Math64.c
```
lgao4, line 202: I may suggest to keep RISCV64 arch for the future support.

lgao4: Other changes are good to me.

kuqin12: Thanks! The RISCV entries are preserved.

kuqin12: Please let me know if you have further feedback. Thank you~

kuqin12:
@lgao4 @mdkinney could you please let me know if you have further feedback on this change? Thanks in advance!

lgao4 (approved):

lgao4:
> @lgao4 @mdkinney could you please let me know if you have further feedback on this change? Thanks in advance!

I am OK.

kuqin12:
> I am OK.

Thanks @lgao4. It seems that we have the needed approval from all stakeholders. Unless there are new feedback, can we try to merge this so that we can have it for 2608?

I understand there are still pending discussions on #12626. But I think it is agreeable that we can set the foundation with this change and build the support for `RngLib`, `TimerLib` or others on the top.

makubacki:
> I understand there are still pending discussions on #12626. But I think it is agreeable that we can set the foundation with this change and build the support for `RngLib`, `TimerLib` or others on the top.

@LeviYeoReum & @samimujawar, do you have any concern with merging this PR, so it is available in the stable tag?

LeviYeoReum (approved):

LeviYeoReum:
> @LeviYeoReum

From my side, LGTM. Thanks!

pierregondois:
> @LeviYeoReum & @samimujawar, do you have any concern with merging this PR, so it is available in the stable tag?

Sami is on leave for the coming weeks,
FWIW I think it would be nice to have PR included in the stable tag.

makubacki:
Thanks all, I just wanted to make sure there was no problem with it going ahead while discussion in https://github.com/tianocore/edk2/pull/12626 continues. I've added the push label.

Issue #12073, opened 2026-01-28
[Bug]: UsbMassStorageDxe: UsbMassReadBlocks lacks retry mechanism for transient USB failures


pbgrimes:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

MdeModulePkg

### Which targets are impacted by this bug?

DEBUG, NO-TARGET, NOOPT, RELEASE

### Current Behavior

UsbMassReadBlocks() in MdeModulePkg/Bus/Usb/UsbMassStorageDxe/UsbMassImpl.c performs a single Reset Recovery on read failure but does not retry the operation. When USB mass storage devices experience transient read errors, the function returns failure immediately after reset, even though a retry would likely succeed.

This causes unnecessary boot failures or file system errors with USB devices that exhibit intermittent communication issues due to electrical noise, hub timing, or device firmware quirks.

### Expected Behavior

After a read failure and Reset Recovery, UsbMassReadBlocks() should retry the read operation before returning failure. Per USB Mass Storage Class Bulk-Only Transport Spec Section 3.1, the device is ready for the next CBW after Reset Recovery, so retrying is both allowed and recommended for robustness.

Expected behavior: retry the read operation up to 3 additional times (4 total attempts) after reset, only returning failure if all attempts fail.

### Steps To Reproduce

1. Environment: Any platform with USB mass storage boot support
2. Connect a USB flash drive or USB HDD that exhibits occasional transient read errors
3. Attempt to boot from the USB device or read files from it
4. Observe that transient failures cause immediate EFI_DEVICE_ERROR returns
5. With retry logic added, the same transient failures recover successfully

Note: This issue is timing/hardware dependent and may not reproduce on all USB devices. It is more commonly observed with:
- Lower-quality USB flash drives
- USB devices connected through hubs
- Systems with marginal USB signal integrity

### Build Environment

```markdown
- OS(s): Windows 11, Ubuntu 22.04
- Tool Chain(s): VS2022, GCC5
```

### Version Information

```text
Commit: Current edk2/master (issue exists in all known versions)

The current upstream code has never included retry logic in UsbMassReadBlocks(). The code performs reset on error but returns failure without retrying.
```

### Urgency

Medium

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

Maintainer feedback requested

## Proposed Fix

[...]

PR #12074, opened 2026-01-28
MdeModulePkg/UsbMassStorageDxe: Add retry mechanism to UsbMassReadBlocks

pbgrimes:
## Summary
Add retry mechanism to UsbMassReadBlocks() to improve robustness with USB mass storage devices that experience transient read failures.

## Details
- Saves original buffer parameters before read operation
- Retries up to 4 times (initial + 3 retries) on failure
- Performs Reset Recovery via UsbMassReset() before each retry
- Compliant with USB Mass Storage Class Bulk-Only Transport Spec Rev 1.0

## Test
Tested on AMD platforms with various USB mass storage devices.

Fixes: https://github.com/tianocore/edk2/issues/12073

### DIFF
--- a/.azurepipelines/templates/pr-gate-build-job.yml
+++ b/.azurepipelines/templates/pr-gate-build-job.yml
@@ -25,21 +25,32 @@ jobs:
   strategy:
     matrix:
       ${{ if eq(parameters.tool_chain_tag, 'GCC') }}:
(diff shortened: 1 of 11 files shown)


lgao4 (approved):

lgao4:
I agree this change.

PR #12075, opened 2026-01-28
NetworkPkg: Add TlsAuthConfig enablement define

antklein:
# Description
This change allows consumers of the NetworkPkg .inc files to enable TSL support and choose if they also want the TlsAuthConfigDxe driver support.

The current design requires both NETWORK_TLS_ENABLE and NETWORK_TLS_AUTH_CONFIG_ENABLE to be set to TRUE for TlsAuthConfigDxe to be enabled.

This change was made because implementation of TlsDxe should not require the optional enablement of TlsAuthConfigDxe.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Verified that setting TLS_AUTH_CONFIG_ENABLE to FALSE causes TlsAuthConfigDxe to not be included as part of the NetworkPkg inc includes.

## Integration Instructions
* N/A


ardbiesheuvel:
> This change was made because implementation of TlsDxe should not require the optional enablement of TlsAuthConfigDxe.
> 

Why not?

antklein:
> > This change was made because implementation of TlsDxe should not require the optional enablement of TlsAuthConfigDxe.
> 
> Why not?

Hi @ardbiesheuvel,

Currently, if utilizing the NetworkPkg's `Network.dsc.inc` and `Network.fdf.inc` files to enable TLS support the user only can fully enablement TLS support via `NETWORK_TLS_ENABLE`. The problem with this design is that the TlsDxe driver is independent of the TlsAuthConfigDxe driver. A user could want to enable TlsDxe without wanting to enable the "setup browser" support that TlsAuthConfigDxe provides.

With the current design this forces the user to set `NETWORK_TLS_ENABLE = FALSE` and then manually include TlsDxe in both their dsc and fdf files. From a glance this design causes the implementor to disable and enable Tls support just to avoid including TlsAuthConfigDxe.

Do you see issues with only enabling TlsDxe without TlsAuthConfigDxe?

SaloniKasbekar:
TlsAuthConfigDxe could be avoided by not having the gEfiHiiConfigAccessProtocolGuid protocol installed. Would that be an option? There are other HII related modules within NetworkPkg that dont have an enablement define. Should this be expanded to all those use cases as well?

PR #12076, opened 2026-01-28
MdeModulePkg: DxeMain: Check memory type overlap inside CoreGetMemoryMap

kuqin12:
# Description

This change adds validation to CoreGetMemoryMap to ensure that special memory bins are fully respected. Specifically, any memory map entry that falls within a special bin must be entirely contained within that bin, and its memory type must match the bin’s designated type.

This check helps preventing unintended changes that could cause the system memory map to cross bin boundaries unexpectedly.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This change was tested on QEMU SBSA platform and booted to Windows desktop.

## Integration Instructions

N/A


MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -141,6 +141,31 @@ RemoveMemoryMapEntry (
...
+  IN EFI_PHYSICAL_ADDRESS  Start1,
+  IN EFI_PHYSICAL_ADDRESS  End1,
+  IN EFI_PHYSICAL_ADDRESS  Start2,
+  IN EFI_PHYSICAL_ADDRESS  End2
+  )
+{
+  ASSERT (Start1 <= End1);
+  ASSERT (Start2 <= End2);
+
+  return ((Start1 < End2) && (Start2 < End1));
```
mdkinney: If first byte of one range is the same as the last byte of the other range, then I think this is a 1 byte overlap.

Should these be <= ?

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -141,6 +141,31 @@ RemoveMemoryMapEntry (
...
+  Helper function to evaluate if memory regions intersect.
+
+  @param  Start1     The address of the first byte in the first memory region.
+  @param  End1       The address of the last byte in the first memory region.
+  @param  Start2     The address of the first byte in the second memory region.
+  @param  End2       The address of the last byte in the second memory region.
+
+  @return TRUE if the memory regions intersect, FALSE otherwise.
+**/
+STATIC
```
mdkinney: use `static`

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -141,6 +141,31 @@ RemoveMemoryMapEntry (
...
+BOOLEAN
+MemoryRegionsIntersect (
```
mdkinney, line 156: This range overlap check is used in many places throughout edk2.  Wondering if we should add a macro to Base.h in area where MIN() and MAX() are defined to add IS_OVERLAP(Start1, End1, Start2, End2) or IS_OVERLAP (Start1, Length1, Start2, Lengths).  The advantage of using length instead of end is that the check for Start < End can be removed.

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -2488,6 +2513,66 @@ CoreGetMemoryMap (
   MergeMemoryMap (MemoryMapStart, &BufferSize, Size);
   MemoryMapEnd = (EFI_MEMORY_DESCRIPTOR *)((UINT8 *)MemoryMapStart + BufferSize);
 
+  DEBUG_CODE_BEGIN ();
```
mdkinney, line 2602: Can this check be moved into its own internal helper function?

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: You are correct. I will update the change shortly.

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: Thanks for the suggestion. I agree this would be a useful addition. I also share the preference for using length as the input, as it avoids potential off‑by‑one ambiguity (as you noted in the other comment).

That said, after evaluating this further, I’m not convinced that `Base.h` is the right place for this check. My concern is that making it too broadly available could encourage overly casual use. Historically, this exact pattern has been the source of security issues (see, for example: https://github.com/tianocore/edk2/commit/cab1f02565d3b29081dd21afb074f35fdb4e1fd6), so I think it’s important that any such interface enforces the same security constraints.

I considered adding a new function in `BaseLib`, but that would prevent reuse of `SafeIntLib` due to circular dependencies, meaning we would need to carefully re‑implement the necessary guards using lower‑level primitives. Another option would be to introduce a dedicated library specifically for this purpose.

I’d appreciate your thoughts on which direction makes the most sense.

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: This is updated. Thanks.

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: Updated, thanks.

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: Updated.

MdeModulePkg/Core/Dxe/Mem/Page.c
mdkinney: We can move this topic to a github issue feature request were we can discuss the options.  Go ahead with the internal function for now.

kuqin12: I created a github issue here: https://github.com/tianocore/edk2/issues/12100. We can continue discussion over there.

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -1985,6 +2010,92 @@ SetEfiMemoryDescriptorType (
...
+            )
+          )
+      {
+        // Verify that it is completely contained within the bin.
+        if ((EntryStart < mMemoryTypeStatistics[Type].BaseAddress) ||
+            (EntryEnd > mMemoryTypeStatistics[Type].MaximumAddress))
+        {
+          DEBUG ((
+            DEBUG_ERROR,
+            "%a: Memory Map entry (Type %d, Start 0x%lx, End 0x%lx) fit within special memory type bin (Type %d, Start 0x%lx, End 0x%lx).\n",
```
lgao4: here should be doesn't fit within, please confirm.

kuqin12: Thanks for the comment. This is updated.

kuqin12:
@lgao4 could you please review the change again as we exit the soft freeze? Thanks in advance!

PR #12020, opened 2026-01-20
Plugins enable noopt package build

mdkinney:
# Description

Update plugins to support GitHub Actions CI agents and enable NOOPT package builds and IA32 host based unit tests

* .pytool/Plugin/CompilerPlugin: Allow NOOPT builds
* .pytool/Plugin/CompilerPlugin: Skip TARGET not supported by DSC
* .pytool/Plugin/HostUnitTestCompilerPlugin: Allow non VS20xx IA32
* BaseTools/Plugin/WindowsVsToolChain: CLANGPDB add IA32 DLLs to Path
* BaseTools/Plugin/WindowsVsToolChain: Add WindowsSDKLibVersion

With the the addition of NOOPT package builds, the Azure Pipelines jobs need to be adjusted to make sure the CI coverage is identical before and after this change.  This means NOOPT is only used for unit tests, and that NOOPT is not used for package builds.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verify GitHub actions CI agents can work and cover NOOPT package builds and IA32 host based unit tests

## Integration Instructions

N/A

PR #12004, opened 2026-01-16
Move post-merge pipelines runs to a daily build

makubacki:
# Description

Pipelines currently run at two main points:

- Pushes to pull requests
- Post-merge (after a commit is merged to the master branch)

There are two main reasons this has historically been done:

1. To catch any issues that may have been missed during PR validation
   due to timing related factors (e.g., merge order into master).
2. To separate a more extensive set of tests from the PR validation
   to reduce the feedback time for contributors in PRs.

PRs are ordered in master with mergify's merge queue, which helps
reduce timing related issues for (1). For (2), the current
post-merge tests are not significantly different from the PR
validation tests, so the value is limited. The main differentiation
is that the PR eval tool is only run in PR triggered pipelines and
always runs in non-PR triggered runs:

For example:

```
# trim the package list if this is a PR
- task: CmdLine@2
  displayName: Check if ${{ parameters.build_pkgs }} need testing
  inputs:
    script: stuart_pr_eval ...
  condition: eq(variables['Build.Reason'], 'PullRequest')
```

It is rare for an issue not be found in the PR eval run. Therefore,
we are moving post-merge runs to a daily schedule. This will
still provide regular testing of the master branch while reducing
overall resource usage lending more capacity to PR validation runs
and the introduction of Clang pipelines.

These pipelines are scheduled to run daily with some degree of
staggering to reduce overlap:

- Non-platform - 7AM UTC
- ArmVirtPkg - 2AM UTC
- EmulatorPkg - 2AM UTC
- OvmfPkg - 5AM UTC

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- Run pipeline to confirm trigger syntax

## Integration Instructions

- N/A

PR #12079, opened 2026-01-29
fix gcc 16 warnings

kraxel:
# Description

gcc learned a new trick, it detects set but unused variables even
in case they get updated.  So it finds a few more unused variables
in the edk2 code base, typically some loop counters where the
counting result is never actually used.

Fix these warnings.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Build test with gcc 16, on fedora 44.

## Integration Instructions

N/A

### DIFF
--- a/ArmPlatformPkg/ArmPlatformPkg.ci.yaml
+++ b/ArmPlatformPkg/ArmPlatformPkg.ci.yaml
@@ -93,9 +93,42 @@
         "IgnoreFiles": [],           # use gitignore syntax to ignore errors
                                      # in matching files
         "ExtendWords": [
+            "afsel",
+            "ctsen",
+            "dataen",
+            "daten",
+            "deassert",
+            "hbackporch",
             "hdlcd",
+            "hsync",
             "icdsgir",
-            "primecells"
+            "inten",
+            "jedec",
+            "pcell",
+            "primecells",
+            "pxclk",
+            "resen",
+            "rtsen",
+            "tpidrurw",
+            "uartcr",
+            "uartdmacr",
+            "uartdr",
+            "uartecr",
+            "uartfbrd",
+            "uartfr",
+            "uartibrd",
+            "uarticr",
+            "uartifls",
+            "uartilpr",
+            "uartimsc",
+            "uartris",
+            "vbackporch",
+            "vsync",
+            "vsyncwidth",
+            "wdogmis",
+            "wdogris",
+            "wsvga",
+            "wsxga"
            ],           # words to extend to the dictionary for this package
         "IgnoreStandardPaths": [    # Standard Plugin defined paths that
             "*.asm", "*.s"          # should be ignore


kraxel:
Hmm, flipping the PR from draft to ready re-runs all CI jobs now (instead of only the assign reviewers job).  Is that intentional?

BaseTools/Source/C/Common/StringFuncs.c
```diff
@@ -120,7 +119,7 @@ SplitStringByWhitespace (
 
   Output = NewStringList ();
 
-  for (Pos = String, Item = 0; Pos < EndOfString; Item++) {
+  for (Pos = String; Pos < EndOfString;) {
```
mikebeaton, line 122: Hi, @dfaggioli already put up a PR to fix just this one of these https://github.com/tianocore/edk2/pull/12047 . Apparently there's a related issue https://github.com/tianocore/edk2/issues/12043, so might be worth adding a link to that here.

On that PR, I suggested that the following would be more idiomatic, once where is is no per-loop update in the for loop:
```
  Pos = String;
  while (Pos < EndOfString) {
```

MdeModulePkg/Library/VarCheckHiiLib/VarCheckHiiGenFromFv.c
```diff
@@ -315,7 +314,7 @@ ParseFv (
       Key = InternalVarCheckAllocateZeroPool (Fv2->KeySize);
       ASSERT (Key != NULL);
 
-      for (FfsIndex = 0; ; FfsIndex++) {
+      for ( ; ; ) {
```
mikebeaton: When fixing similar issues which have been thrown up by CLANG before, IIRC @leiflindholm has suggested to me that `while (TRUE) {` would be more idiomatic here.

mikebeaton:
> Hmm, flipping the PR from draft to ready re-runs all CI jobs now (instead of only the assign reviewers job). Is that intentional?

cc @makubacki (https://github.com/tianocore/edk2/pull/12004) or @mdkinney (https://github.com/tianocore/edk2/pull/12020)

MdeModulePkg/Library/VarCheckHiiLib/VarCheckHiiGenFromFv.c
kraxel: Yes, looks a bit nicer, especially with uncrustify insisting on the spaces so you can't write `for (;;) { ...`

BaseTools/Source/C/Common/StringFuncs.c
kraxel: The initialization is still there though, so when not using `for` you need one additional line.  I'd prefer to stick to `for` here.

mikebeaton: I don't have permissions to resolve the comment, but 'Resolved'.

MdeModulePkg/Library/VarCheckHiiLib/VarCheckHiiGenFromFv.c
mikebeaton: Resolved.

PR #12083, opened 2026-01-29
Replaced non-ASCII characters with ASCII characters

DhiineshM-AMI:
# Description

Replacing the non-ASCII characters with ASCII equivalents fixes build failures on Windows systems using non-English locales (Chinese, Japanese, Korean, etc.). These characters trigger MSVC warning C4819 (“file contains a character that cannot be represented in the current code page”), which becomes an error when /WX is enabled. Using ASCII ensures predictable behavior regardless of system code page, maintains consistent behavior across compilers, editors, and platforms, and aligns the file with EDK2 coding standards

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## Integration Instructions

N/A

### DIFF
--- a/ArmPlatformPkg/ArmPlatformPkg.ci.yaml
+++ b/ArmPlatformPkg/ArmPlatformPkg.ci.yaml
@@ -93,9 +93,42 @@
         "IgnoreFiles": [],           # use gitignore syntax to ignore errors
                                      # in matching files
         "ExtendWords": [
+            "afsel",
+            "ctsen",
+            "dataen",
+            "daten",
+            "deassert",
+            "hbackporch",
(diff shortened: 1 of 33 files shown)


ardbiesheuvel:
You forgot to explain why this is needed.

DhiineshM-AMI:
> You forgot to explain why this is needed.

You’re right I’ve updated the PR description to explain why this change is needed

LeviYeoReum:
I think it would be better to update commit message too.

DhiineshM-AMI:
> I think it would be better to update commit message too.

Updated the commit message as suggested. Please let me know if anything else needs adjustment

PR #12085, opened 2026-01-29
ArmPlatformPkg/PL011UartLib : Fix SetControl() SCT conformance

samerhaj:
# Description

The PL011UartLib SetControl() is failing the SCT test for SerialIoBBTestConformance (00605CBC-3965-4B61-A254-2B2B723172EA), which is trying to set bits that are not supported per UEFI spec.

Add proper argument check for valid bits, and confirm that test passes.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

EDK2 SCT test SerialIoBBTestConformance (00605CBC-3965-4B61-A254-2B2B723172EA) on RPi4, with failure without this patch [shown here](https://github.com/pftf/acs-reports/blob/master/v1.17%2B/sct-tianocore/Overall/Summary.log)

Also tested by @wangsim on NVIDIA GB300, test results: [SetControl_Conf_0_0_B7767573-DEE3-424C-991C-1F5581433F02.log](https://github.com/user-attachments/files/24942344/SetControl_Conf_0_0_B7767573-DEE3-424C-991C-1F5581433F02.log)

## Integration Instructions

N/A

### DIFF
--- a/ArmPlatformPkg/Library/PL011UartLib/PL011UartLib.c
+++ b/ArmPlatformPkg/Library/PL011UartLib/PL011UartLib.c
@@ -22,10 +22,22 @@
 #define FRACTION_PART_MASK          ((1 << FRACTION_PART_SIZE_IN_BITS) - 1)
 
 //
-// EFI_SERIAL_SOFTWARE_LOOPBACK_ENABLE is the only
-// control bit that is not supported.
+// Per UEFI spec, only these control bits are allowed to be set.
 //
-STATIC CONST UINT32  mInvalidControlBits = EFI_SERIAL_SOFTWARE_LOOPBACK_ENABLE;
+STATIC CONST UINT32  mAllowedControlBits =
+  EFI_SERIAL_CLEAR_TO_SEND |
+  EFI_SERIAL_DATA_SET_READY |
+  EFI_SERIAL_RING_INDICATE |
+  EFI_SERIAL_CARRIER_DETECT |
+  EFI_SERIAL_SOFTWARE_LOOPBACK_ENABLE |
+  EFI_SERIAL_OUTPUT_BUFFER_EMPTY |
+  EFI_SERIAL_INPUT_BUFFER_EMPTY;
+
+//
+// platform-specific bits that are not supported by this device.
+//
+STATIC CONST UINT32  mPlatformInvalidControlBits =
+  EFI_SERIAL_SOFTWARE_LOOPBACK_ENABLE;
 
 /**
 
@@ -274,7 +286,11 @@ PL011UartSetControl (
 {
   UINT32  Bits;
 
-  if ((Control & mInvalidControlBits) != 0) {
+  if ((Control & ~mAllowedControlBits) != 0) {
+    return RETURN_UNSUPPORTED;
+  }
+
+  if ((Control & mPlatformInvalidControlBits) != 0) {
     return RETURN_UNSUPPORTED;
   }
 


ardbiesheuvel:
Please stop creating new PRs for this change - it is impossible to keep track of things this way, and I have no clue what changed between the different versions.

anncheng-arm:
Hi @ardbiesheuvel @samerhaj @leiflindholm 

I've addressed all the errors reported by CI. Let me know if any other revie comments.

ArmPlatformPkg/ArmPlatformPkg.ci.yaml
```diff
@@ -95,7 +95,40 @@
         "ExtendWords": [
             "hdlcd",
             "icdsgir",
-            "primecells"
+            "primecells",
```
leiflindholm: This is an unrelated change, please submit as a separate PR.
And then please keep the list sorted rather than appending a sorted list to the sorted list.

ArmPlatformPkg/Library/PL011UartLib/PL011UartLib.c
```diff
@@ -274,6 +274,21 @@ PL011UartSetControl (
 {
   UINT32  Bits;
 
+  //
```
leiflindholm: This comment feels slightly excessive.
Could summarise as "Ensure only control bits are specified".

But then the flow of the function becomes a bit weird when we return EFI_UNSUPPORTED (which should be RETURN_UNSUPPORTED) if a set of bits including EFI_SERIAL_SOFTWARE_LOOPBACK_ENABLE is set only to go on and return RETURN_UNSUPPORTED if EFI_SERIAL_SOFTWARE_LOOPBACK_ENABLE is set (in mInvalidControlBits).

Maybe just add the additional bits to mInvalidControlBits?

anncheng-arm: I had the same concern initially as well, but after looking at it more closely, I think what Samer is trying to do is:

The first if check is to validate the control bits defined by the UEFI spec, and return unsupported if any non-spec bits are set.

The second if check is to handle EFI_SERIAL_SOFTWARE_LOOPBACK_ENABLE, which is defined by the spec but is invalid on this platform.

@samerhaj correct me if anything wrong.

And how about this:
```
STATIC CONST UINT32 AllowedControlBits =
    EFI_SERIAL_REQUEST_TO_SEND |
    EFI_SERIAL_DATA_TERMINAL_READY |
    EFI_SERIAL_HARDWARE_LOOPBACK_ENABLE |
    EFI_SERIAL_SOFTWARE_LOOPBACK_ENABLE |
    EFI_SERIAL_HARDWARE_FLOW_CONTROL_ENABLE;

// platform-specific bits that are not supported on this platform
STATIC CONST UINT32 mInvalidControlBits = EFI_SERIAL_SOFTWARE_LOOPBACK_ENABLE;

if ((Control & ~AllowedControlBits) != 0) {
    return RETURN_UNSUPPORTED;
}

// platform capability: spec-allowed but platform-invalid
if ((Control & mInvalidControlBits) != 0) {
    return RETURN_UNSUPPORTED;
}
```

ArmPlatformPkg/ArmPlatformPkg.ci.yaml
anncheng-arm: Hi @leiflindholm Thanks for the comments.

I think Samer included this change because these files have existed for a while before the CI checks were enabled, and once CI was turned on, they started to introduce a large number of CI failures in this pull request.
The additional extendwords are added specifically to get the cspell check to pass for the existing content.

Would it be acceptable if I updated the commit message to clearly describe the purpose of this change, or would you still prefer this to be split into a separate pull request?

I will sort the list after getting your feedback.

ArmPlatformPkg/Library/PL011UartLib/PL011UartLib.c
samerhaj: @anncheng-arm yes you are correct. But I think it is ok to combine the bits by updating mInvalidControlBits as Leif suggested. We could also update the comment on line 25 to say something like:

```
//
// EFI_SERIAL_SOFTWARE_LOOPBACK_ENABLE is not supported by the device.
// The others are not allowed by the UEFI spec.
//
STATIC CONST UINT32  mInvalidControlBits = 
    EFI_SERIAL_SOFTWARE_LOOPBACK_ENABLE |
    EFI_SERIAL_REQUEST_TO_SEND |
    EFI_SERIAL_DATA_TERMINAL_READY |
    EFI_SERIAL_HARDWARE_LOOPBACK_ENABLE |
    EFI_SERIAL_SOFTWARE_LOOPBACK_ENABLE |
    EFI_SERIAL_HARDWARE_FLOW_CONTROL_ENABLE;
```

ArmPlatformPkg/ArmPlatformPkg.ci.yaml
anncheng-arm: Move spellcheck change to PR #12105

ArmPlatformPkg/Library/PL011UartLib/PL011UartLib.c
anncheng-arm: @samerhaj , @leiflindholm 

Updated per suggestion, but this PR needs to wait for #12105 to be merged first for CI to pass. You can review the code in the meantime.

anncheng-arm:
Move spellcheck to other PR #12105, and this PR needs to wait for #12105 to be merged first.

ArmPlatformPkg/Library/PL011UartLib/PL011UartLib.c
leiflindholm: To be honest, I like the AllowedControlBits/mInvalidControlBits variant, but either is fine.

However if using that variant, both variables should be either global (and with the m prefix) or local (and without it).

anncheng-arm:
@leiflindholm, @samerhaj 

the CI got PASSED after rebase.

ArmPlatformPkg/Library/PL011UartLib/PL011UartLib.c
anncheng-arm: @leiflindholm, thanks a lot for the reminding, I've pushed a new proposal based on your suggestion. Let me know if anything else that needs to be improved.

anncheng-arm:
@ardbiesheuvel @samimujawar 

Looks like I need a more approval from reviewer with write access. Could you kindly have a look?

@leiflindholm @samerhaj for vis.

leiflindholm:
> @ardbiesheuvel @samimujawar
> 
> Looks like I need a more approval from reviewer with write access. Could you kindly have a look?
> 
> @leiflindholm @samerhaj for vis.

This is an artefact of the repo lockdown for soft freeze.

@lgao4 / @mdkinney could one of you please set your review flag on this PR so the gating requirement is fulfilled?

leiflindholm:
@lgao4 @mdkinney this one seems stuck

PR #12088, opened 2026-01-30
MdeModulePkg/ArmFfaLib: Implement base instance of ArmFfaLib

Raymond-MS:
# Description

Current usage of Rx/Tx buffer is unnecessarily pulling in UEFI framework related data usage, i.e. PEI, DXE, STMM, etc. This is undesirable for components that only needs to use the basic functionalities.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Ran CI locally

## Integration Instructions

Normal integration instructions. If wanting to use this library, add it to the corresponding .inf as well as the .dsc for your platform.

### DIFF
--- a/MdeModulePkg/Library/ArmFfaLib/ArmFfaLibBase.inf
+++ b/MdeModulePkg/Library/ArmFfaLib/ArmFfaLibBase.inf
@@ -1,0 +1,46 @@
+## @file
+#  This file is originally sourced from ArmFfaStandaloneMmLib, with
+#  Rx/Tx-related APIs removed to focus exclusively on FFA primitives.
+#  Provides FF-A ABI Library used in all environments.
+#
+#  Copyright (c) 2024, Arm Limited. All rights reserved.<BR>
+#
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+[Defines]
+  INF_VERSION                    = 0x0001001B
+  BASE_NAME                      = ArmFfaLibBase
+  FILE_GUID                      = D79FF75E-AD0A-4074-A335-2084533744AF
+  MODULE_TYPE                    = BASE
+  VERSION_STRING                 = 1.0
+  PI_SPECIFICATION_VERSION       = 0x00010032
+  LIBRARY_CLASS                  = ArmFfaLib
+  CONSTRUCTOR                    = ArmFfaLibBaseConstructor
+
+[Sources]
+  ArmFfaCommon.h
+  ArmFfaCommon.c
+  ArmFfaLibBase.c
+
+[Packages]
+  MdePkg/MdePkg.dec
+  MdeModulePkg/MdeModulePkg.dec
+
+[LibraryClasses]
+  ArmSmcLib
+  ArmSvcLib
+  BaseLib
+  BaseMemoryLib
+  DebugLib
+  HobLib
+  PcdLib
+  MemoryAllocationLib
+  TimerLib
+
+[Pcd]
+  gEfiMdeModulePkgTokenSpaceGuid.PcdFfaLibConduitSmc
+  gEfiMdeModulePkgTokenSpaceGuid.PcdFfaTxRxPageCount
+
+[Guids]
+  gArmFfaRxTxBufferInfoGuid
--- a/MdeModulePkg/Library/ArmFfaLib/ArmFfaLibBase.c
+++ b/MdeModulePkg/Library/ArmFfaLib/ArmFfaLibBase.c
@@ -1,0 +1,57 @@
+/** @file
+  This file is originally sourced from ArmFfaStandaloneMmLib, with
+  Rx/Tx-related APIs removed to focus exclusively on FFA primitives.
+
+  Provides FF-A ABI Library used in all environments.
+
+  Copyright (c) 2024, Arm Limited. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+   @par Glossary:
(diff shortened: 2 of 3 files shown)


LeviYeoReum:
Thanks for your patch @kuqin12. 
But while I'm reviewing, I found an problem on ArmFfaPeiLib: 
  - https://github.com/tianocore/edk2/pull/12096
  
So, I wonder whether we could make a BaseLib format.
For a purpose of skipping Rx/Tx related, I think we can add a Pcd so that determine it whether to skip it or not.

Am I missing something?

Raymond-MS:
> Thanks for your patch @kuqin12. But while I'm reviewing, I found an problem on ArmFfaPeiLib:
> 
> * [MdeModulePkg: ArmFfaLib: make ArmFfaPeiLib run early PEIM phase #12096](https://github.com/tianocore/edk2/pull/12096)
> 
> So, I wonder whether we could make a BaseLib format. For a purpose of skipping Rx/Tx related, I think we can add a Pcd so that determine it whether to skip it or not.
> 
> Am I missing something?

Hey @LeviYeoReum, I actually have a change for this in the works with plans to upstream it shortly. On our reference platform, I noticed issues where ArmFfaCommon is using these global/static variables which was causing our SEC phase to hang. I removed/moved them. The changes I made were similar to the PR you referenced with the slight difference that the globals have been removed entirely. I ended up moving them to their corresponding phase libs. (i.e. ArmFfaDxeLib, etc.) The SEC and PEI versions no longer use the globals and instead, query the information each time they are needed. Would you like me to push this change up ASAP so you can take a look?

LeviYeoReum:
> > Thanks for your patch @kuqin12. But while I'm reviewing, I found an problem on ArmFfaPeiLib:
> > 
> > * [MdeModulePkg: ArmFfaLib: make ArmFfaPeiLib run early PEIM phase #12096](https://github.com/tianocore/edk2/pull/12096)
> > 
> > So, I wonder whether we could make a BaseLib format. For a purpose of skipping Rx/Tx related, I think we can add a Pcd so that determine it whether to skip it or not.
> > Am I missing something?
> 
> Hey @LeviYeoReum, I actually have a change for this in the works with plans to upstream it shortly. On our reference platform, I noticed issues where ArmFfaCommon is using these global/static variables which was causing our SEC phase to hang. I removed/moved them. The changes I made were similar to the PR you referenced with the slight difference that the globals have been removed entirely. I ended up moving them to their corresponding phase libs. (i.e. ArmFfaDxeLib, etc.) The SEC and PEI versions no longer use the globals and instead, query the information each time they are needed. Would you like me to push this change up ASAP so you can take a look?

Hi Raymond. 

First, the `ArmFfaSecLib` was designed for "PeilessSec" where doesn't need to be executed in XIP enviroment.
But If you have a patchset for making a ArmFfaSecLib in normal SEC phase, That would be also good.
If you push your change, I'm glad to review it and let us talk which is better.

Thanks ;)

Raymond-MS:
> > > Thanks for your patch @kuqin12. But while I'm reviewing, I found an problem on ArmFfaPeiLib:
> > > 
> > > * [MdeModulePkg: ArmFfaLib: make ArmFfaPeiLib run early PEIM phase #12096](https://github.com/tianocore/edk2/pull/12096)
> > > 
> > > So, I wonder whether we could make a BaseLib format. For a purpose of skipping Rx/Tx related, I think we can add a Pcd so that determine it whether to skip it or not.
> > > Am I missing something?
> > 
> > 
> > Hey @LeviYeoReum, I actually have a change for this in the works with plans to upstream it shortly. On our reference platform, I noticed issues where ArmFfaCommon is using these global/static variables which was causing our SEC phase to hang. I removed/moved them. The changes I made were similar to the PR you referenced with the slight difference that the globals have been removed entirely. I ended up moving them to their corresponding phase libs. (i.e. ArmFfaDxeLib, etc.) The SEC and PEI versions no longer use the globals and instead, query the information each time they are needed. Would you like me to push this change up ASAP so you can take a look?
> 
> Hi Raymond.
> 
> First, the `ArmFfaSecLib` was designed for "PeilessSec" where doesn't need to be executed in XIP enviroment. But If you have a patchset for making a ArmFfaSecLib in normal SEC phase, That would be also good. If you push your change, I'm glad to review it and let us talk which is better.
> 
> Thanks ;)

Of course! I just finished pushing up the changes I have. Please see: https://github.com/tianocore/edk2/pull/12101

MdeModulePkg/Library/ArmFfaLib/ArmFfaLibBase.inf
```diff
@@ -0,0 +1,46 @@
...
+  FILE_GUID                      = D79FF75E-AD0A-4074-A335-2084533744AF
+  MODULE_TYPE                    = BASE
+  VERSION_STRING                 = 1.0
+  PI_SPECIFICATION_VERSION       = 0x00010032
+  LIBRARY_CLASS                  = ArmFfaLib
+  CONSTRUCTOR                    = ArmFfaLibBaseConstructor
+
+[Sources]
+  ArmFfaCommon.h
+  ArmFfaCommon.c
```
os-d, line 23: ArmFfaCommon calls `ArmFfaLibGetRxTxBuffer()` which CLANGPDB AARCH64 points out is not being defined in this new base lib. Perhaps you need to add `ArmFfaSecRxTxMap.c` here? Perhaps renaming that `Base` instead of `Sec`?

Raymond-MS:
Dropping for now. Plans to upstream after reevaluating this base lib.

PR #12089, opened 2026-01-30
MdePkg: Add FFA_NS_RES_INFO_GET to Arm header

Raymond-MS:
# Description

FFA_NS_RES_INFO_GET is a new command introduced in FF-A v1.3. Updating the ArmFfaSvc header to include the new command ID.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Ran CI locally

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
+++ b/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
@@ -1194,3 +1194,49 @@ ArmFfaLibIsFfaSupported (
 
   return TRUE;
 }
+
+/**
+  Send FF-A Non-secure Resoure Info Get Command.
+
+  @param [in]   TargetId         Partition ID to query info from
+  @param [in]   Flags            Additional flags
+  @param [out]  WrittenSize      How much data was written in the transaction
+  @param [out]  RemainingSize    How much data remains to be read
+
+  @retval EFI_SUCCESS               Success, info returned in Rx/Tx buffer
+  @retval EFI_INVALID PARAMETER     Invalid parameter(s)
+  @retval Others                    Error
+
+**/
+EFI_STATUS
+EFIAPI
+ArmFfaLibFfaNsResInfoGet (
+  IN UINT16   TargetId,
+  IN UINT64   Flags,
+  OUT UINT32  *WrittenSize,
+  OUT UINT32  *RemainingSize
+  )
+{
+  EFI_STATUS    Status;
+  ARM_FFA_ARGS  FfaArgs;
+
+  if ((WrittenSize == NULL) || (RemainingSize == NULL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  FfaArgs.Arg0 = ARM_FID_FFA_NS_RES_INFO_GET;
+  FfaArgs.Arg1 = TargetId;
+  FfaArgs.Arg2 = Flags;
+
+  ArmCallFfa (&FfaArgs);
+
+  Status = FfaArgsToEfiStatus (&FfaArgs);
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  *WrittenSize   = FfaArgs.Arg2 >> 32;
+  *RemainingSize = FfaArgs.Arg2;
+
+  return EFI_SUCCESS;
+}
--- a/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.h
+++ b/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.h
@@ -113,3 +113,25 @@ EFIAPI
 ArmFfaLibGetPartId (
   OUT UINT16  *PartId
   );
+
+/**
+  Send FF-A Non-secure Resoure Info Get Command.
+
+  @param [in]   TargetId         Partition ID to query info from
+  @param [in]   Flags            Additional flags
+  @param [out]  WrittenSize      How much data was written in the transaction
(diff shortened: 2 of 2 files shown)


MdePkg/Include/IndustryStandard/ArmFfaSvc.h
kuqin12: Please update this comment as well.

MdePkg/Include/IndustryStandard/ArmFfaSvc.h
```diff
@@ -50,6 +50,7 @@
 #define ARM_FID_FFA_MSG_SEND_DIRECT_RESP_AARCH64  0xC4000070
 #define ARM_FID_FFA_MSG_SEND_DIRECT_REQ2          0xC400008D
 #define ARM_FID_FFA_MSG_SEND_DIRECT_RESP2         0xC400008E
+#define ARM_FID_FFA_NS_RES_INFO_GET               0xC400008F
```
kuqin12: Also might need comment to separate the new one(s) from FF-A 1.2 defined function IDs.

LeviYeoReum:
This patch seems to add defines only without any usage.
I think it would be better to be added with its usage too.

Thanks.

Raymond-MS:
> This patch seems to add defines only without any usage. I think it would be better to be added with its usage too.
> 
> Thanks.

Added the implementation as well!

@LeviYeoReum Please take a look when you get a chance! Thanks!

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -1172,3 +1172,49 @@ GetRxTxBufferMinSizeAndAlign (
...
+  Send FF-A Non-secure Resoure Info Get Command
+
+  @param [in]   TargetId         Partition ID to query info from
+  @param [in]   Flags            Additional flags
+  @param [out]  WrittenSize      How much data was written in the transaction
+  @param [out]  RemainingSize    How much data remains to be read
+
+  @retval EFI_SUCCESS               Success
+  @retval EFI_INVALID PARAMETER     Invalid parameter(s)
+  @retval Others                    Error
```
makubacki: Can you please add more detail to the return value descriptions?

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -1172,3 +1172,49 @@ GetRxTxBufferMinSizeAndAlign (
...
+{
+  EFI_STATUS    Status;
+  ARM_FFA_ARGS  FfaArgs = { 0 };
+
+  if ((WrittenSize == NULL) || (RemainingSize == NULL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  FfaArgs.Arg0 = ARM_FID_FFA_NS_RES_INFO_GET;
+  FfaArgs.Arg1 = TargetId;
```
LeviYeoReum, line 1228: Do we need to check TargetId is for S-Endpoint acoording to Flags's bit 0 at lest?

Raymond-MS: We could do that here but I didn't think it was necessary. The caller should make sure that the TargetId is valid if the bit in the flag is set. The handler on the Hafnium side will catch this and return an error if that is not the case.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
Raymond-MS: Of course, added.

MdePkg/Include/IndustryStandard/ArmFfaSvc.h
```diff
@@ -11,7 +11,7 @@
     - FF-A - Firmware Framework for Arm A-profile
 
   @par Reference(s):
-    - FF-A Version 1.2 [https://developer.arm.com/documentation/den0077/latest/]
+    - FF-A Version 1.3 [https://developer.arm.com/documentation/den0077/latest/]
```
kuqin12: Can you please change the link to point to v1.3 specifically? The latest does not really mean much when it comes to pairing with the implementation.

Raymond-MS: It does point to v1.3. That is why it is marked latest. I even checked to verify that it is v1.3 when I made this change.

kuqin12: It will not be when v1.4 is available, right?

Raymond-MS: I understand the concern, I am fine to make it point directly to the latest v1.3. However, if a newer version of v.1.3 is released, this link will become stale. Also, the link below for FF-A memory management should be updated to not point to the latest in this instance but as I didn't make a change there, I won't touch it.

kuqin12: No need to touch the line below.

But this link, after pointing to a specific version, becomes stale actually reflects the code status in this file, which likely will go stale without any modification.

Raymond-MS: Sounds good.

kuqin12 (approved):

Raymond-MS:
CI is failing which is why I am force pushing. There are no changes in this commits, I am only amending the commit in order to force CI to re-run. There have been weird issues with GitHub and CI recently with timeouts. I am attempting to make sure CI is passing on this PR.

PR #12096, opened 2026-01-30
MdeModulePkg: ArmFfaLib: make ArmFfaPeiLib run early PEIM phase

LeviYeoReum:
# Description

ArmFfaLibCommonInit() caches several pieces of information
in global variables. Therefore, this function should only be called
when the PEIM is running in permanent memory.
Otherwise, It could make a problem while running in XIP.
    
To address this Call ArmFfaLibCommonInit() when it runs on the memory and
For the case of early PEIM phase which doesn't call ArmFfaLibCommonInit(),
make all ArmFfaLib* functions will perform the necessary FF-A requests themselves 
at runtime to obtain the required information.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Try to run FVP_RevC with EDK_SKIP_PEIM=FALSE

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
+++ b/MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
@@ -103,23 +103,38 @@ ArmFfaPeiLibConstructor (
   UINTN                      Property1;
   UINTN                      Property2;
 
-  Status = ArmFfaLibCommonInit ();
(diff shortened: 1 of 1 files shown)


LeviYeoReum:
PR #12101 will supersede this one.
Once #12101 is merged, I'll close this.

Thanks.

PR #12101, opened 2026-01-30
MdeModulePkg: Remove global usage in ArmFfaCommon

Raymond-MS:
# Description

Removed global variables in ArmFfaCommon. Moved the globals to locals in each phase's ArmFfaLib implementation. SEC and PEI  will query when necessary to avoid setting globals when memory is unavailable.

Add Unmap callback for when PEI and SEC need to invalidate the Rx/Tx buffer HOB on a call to Unmap.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Ran CI locally

## Integration Instructions

N/A


MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.h
```diff
@@ -86,4 +88,34 @@ GetRxTxBufferMinSizeAndAlign (
...
+
+  @param[out] IsSupported  If FF-A is supported
+
+  @retval EFI_SUCCESS            FF-A is supported
+  @retval EFI_UNSUPPORTED        FF-A isn't supported
+
+**/
+EFI_STATUS
+EFIAPI
+ArmFfaLibIsFfaSupported (
```
LeviYeoReum, line 99: Well. for a internal functions which used by ArmFfaLib inside,
I think it would be better to remove prefix "ArmFfaLib*".

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
```diff
@@ -220,3 +222,42 @@ ArmFfaPeiLibConstructor (
...
+  @retval Others       Errors
+
+**/
+EFI_STATUS
+EFIAPI
+ArmFfaLibGetPartId (
+  IN UINT16  *PartId
+  )
+{
+  return ArmFfaLibPartitionIdGet (PartId);
```
LeviYeoReum: Hmm. This means whenever FF-A APIs which need partition id, It should send FF-A request.
What do you think about applying like: 
https://github.com/tianocore/edk2/pull/12096/changes/d9c57da5b9a8625f35335ee5ee87f4c31fc078c2#diff-a653eb9a9085f9eefb8f1d28702c565dd49f982934729ef90957ede96398bf?

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.c
```diff
@@ -138,3 +138,33 @@ IsFfaSupported (
...
+   * See ArmFfaLibRxTxMap(). If HOB is not found, the Rx/Tx
+   * buffers were not successfully mapped.
+   */
+  RxTxBufferAllocationHob = FindRxTxBufferAllocationHob (TRUE);
+  if (RxTxBufferAllocationHob == NULL) {
+    return;
+  }
+
+  /*
+   * Invalidate the HOB.
```
LeviYeoReum, line 161: I think this is not enough.
Suppose the failure in Sec phase and triggers the `UnmapCallback()` while Hoblist has `RxTxBufferAllocationHob`.

In the ArmPeiLib, It checks whether Rx/Tx buffer is allocated from SEC only present of `RxTxBufferAllocationHob()` not care about its contents so ArmFfaPeiLib construtor doesn't call the
ArmFfaLibRxTxMap() and all of the subsequence API call related Rx/Tx buffer will be failed since its Rx/Tx buffer is nullified.

I think it would be better to add some field in ARM_FA_RX_TX_BUFFER_INFO whether its information is valid so that it could be checked subsequence phase of ArmFfaLib's construtor.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.h
Raymond-MS: I think the issue here is that there is an IsFfaSupported function within ArmFfaLib. That one should be ArmFfaLibIsFfaSupported, right? And then this instance can be swapped to IsFfaSupported. However, I think the implementation of both should remain as is. The library version should query as normal and the other should use a global if available.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -205,11 +216,50 @@ ArmFfaLibRxTxMap (
...
-  mTxBuffer = TxBuffer;
-  mRxBuffer = RxBuffer;
+  BufferInfo = BuildGuidHob (&gArmFfaRxTxBufferInfoGuid, sizeof (ARM_FFA_RX_TX_BUFFER_INFO));
+  if (BufferInfo == NULL) {
+    DEBUG ((DEBUG_ERROR, "%a: Failed to create Rx/Tx Buffer Info Hob\n", __func__));
+    Status = EFI_OUT_OF_RESOURCES;
+    goto Unmap;
+  }
+
+  RxTxBufferAllocationHob = GetRxTxBufferAllocationHob ((EFI_PHYSICAL_ADDRESS)((UINTN)TxBuffer), PcdGet64 (PcdFfaTxRxPageCount) * EFI_PAGE_SIZE * 2, FALSE);
```
LeviYeoReum: Thanks for this patch. Just for question. Could I ask whether  this patch is tested with PeilessSec?

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.h
LeviYeoReum: Fair enough. Thanks

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
Raymond-MS: Absolutely, this actually spawned because our reference platform (QEMU SBSA) was recently switched to be PEI-less and as such we are using PeilessSec. The issue we ran into was that our PeilessSec needs to initialize a TPM which uses the ARM FF-A library. This caused our platform to hang because the libraries were attempting to set these global variables both in ArmFfaCommon and in ArmFfaSecRxTxMap. With this change it is working for us.

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
Raymond-MS: I have been taking a look at this change and I think I understand it but I am still a little confused. Please correct me if I am wrong but my understanding is that this change is checking to see if we are in a permanent memory state. If so, it will call ArmCommonInit which will set the flag to use the local cached variables. If not, the flag will still be false and all calls to the functions will query every time. However, this constructor should only ever be called once which means if the check to see if we are in a permanent memory state fails, how will we know if FF-A is available? That is one of the checks of ArmFfaCommonInit. Wouldn't we still want to call it? I think what I can do is have local variables in the PEI version but only use them if the check passes, thoughts? Basically it's a mesh of both of our changes regarding PEI.

LeviYeoReum: > this constructor should only ever be called once which means if the check to see if we are in a permanent memory state fails, how will we know if FF-A is available? That is one of the checks of ArmFfaCommonInit. Wouldn't we still want to call it? I think what I can do is have local variables in the PEI version but only use them if the check passes, thoughts?

AFAIK, In PEI phase, the "Construtor" is called two time
  - When it runs at XIP
  - After Permenant memory can be avaiable and PEIMs are moved to ShadowCopy.

And the RegisterForShadow returns:
  - EFI_SUCCESS: in XIP
  - EFI_READY_ALREADY: in Permenant 
 
the EFI_NOT_GOUND as I usnderstand seems almost "BUG" situation.

IOW, the failure of checking should not happen and it should be a ASSERT..
So I think the asumeption "permanent memory state fails" could be ignorable.

>  Basically it's a mesh of both of our changes regarding PEI.

Yes. agree but one of the effort to reduce the cost of "SMC or HVC"

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.c
Raymond-MS: My understanding is most of the calls to query the HOB use the GUID to find it. If the flag to use the GUID is false, it finds the next empty HOB and uses that for the allocation HOB. When we invalidate the HOB we are clearing all contents of it which includes the GUID. If Unmap gets called in SEC the HOB won't be found in PEI and as such the HOB will need to be generated again which is what PEI handles.

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
Raymond-MS: In the case of running from Flash, we want to make sure we are not using any globals which the change you provided handles. However, you will only check if FF-A is enabled when permanent memory is available, why when running from Flash would we want to go through all the extra parts of the constructor such as acquiring the HOB if we could bail early? Wouldn't this be an oversight to attempt the other parts of the constructor if FF-A is not supported? I think that could lead to undefined or undesirable behavior, right?

My suggestion is adding the check for permanent memory and if successful, pass global variables to populate in ArmFfaCommonInit, otherwise, we will pass locals like we are now. Basically, when running from Flash it will query the necessary information and when permanent memory is available it will start using the globals like ArmFfaDxeLib/etc.

LeviYeoReum: Ah. I've missed one of my change (I have a change is the local but I didn't update pull request...)
This makes you so confused and my response is quite a obscure because of this. Sorry:

https://github.com/tianocore/edk2/pull/12096/changes#diff-a653eb9a9085f9eefb8f1d28702c565dd49f982934729ef90957ede96398bf06

> My suggestion is adding the check for permanent memory and if successful, pass global variables to populate in ArmFfaCommonInit, otherwise, we will pass locals like we are now.

That's what I wanted, even if wihtout calling check the FF-A support via FFA_VERSION that makes a problem
So Yes and the missing part of former was you pointed out.

Thanks :)

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.c
LeviYeoReum: Oh. I've misread a code. You invalidate the Hob header too.
Thanks!

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
Raymond-MS: Okay, I pushed up the change. Let me know what you think! Also, please let me know if you want me to resolve the other comments or if you want to do it. Thanks!

LeviYeoReum:
Thislooks good to me :)
Thanks!

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.h
```diff
@@ -86,4 +88,34 @@ GetRxTxBufferMinSizeAndAlign (
...
+/**
+  Determine if FF-A is supported
+
+  @param[out] IsSupported  If FF-A is supported
+
+  @retval EFI_SUCCESS            FF-A is supported
+  @retval EFI_UNSUPPORTED        FF-A isn't supported
+
+**/
+EFI_STATUS
```
makubacki: A function like this would usually return a `BOOLEAN`. If the function only returns `EFI_SUCCESS` (true) and `EFI_UNSUPPORTED` (false), as documented, did you consider a Boolean type and reject it for some reason?

---

That allows code like you have calling it right now as:

```c
  Status = ArmFfaLibIsFfaSupported ();
  if (EFI_ERROR (Status)) {
    return FALSE;
  }

  return TRUE;
```

To simply be written as:

```c
  return ArmFfaLibIsFfaSupported ();
```

MdeModulePkg/Library/ArmFfaLib/ArmFfaDxeLib.c
```diff
@@ -194,3 +196,53 @@ ArmFfaDxeLibConstructor (
...
+  Return partition or VM ID
+
+  @retval EFI_SUCCESS
+  @retval Others       Errors
+
+**/
+EFI_STATUS
+EFIAPI
+ArmFfaLibGetPartId (
+  IN UINT16  *PartId
```
makubacki: The header file identifies `PartId` as `OUT` right now (which is correct as the pointer is written):

```c
EFI_STATUS
EFIAPI
ArmFfaLibGetPartId (
  OUT UINT16  *PartId
  );
```

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
```diff
@@ -220,3 +252,81 @@ ArmFfaPeiLibConstructor (
...
+  if (mIsPermMemAvailable) {
+    return mIsFfaSupported;
+  }
+
+  Status = ArmFfaLibIsFfaSupported ();
+  if (EFI_ERROR (Status)) {
+    return FALSE;
+  }
+
+  return TRUE;
```
makubacki: nit - readability:

```suggestion
  return !EFI_ERROR (ArmFfaLibIsFfaSupported ());
```

But, as I suggested elsewhere, it would be more natural for `ArmFfaLibIsFfaSupported()` return a Boolean type.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.c
```diff
@@ -91,39 +91,80 @@ ArmFfaSecLibConstructor (
                &Property2
                );
     if (!EFI_ERROR (Status)) {
-      DEBUG ((DEBUG_INFO, "%a Rx/Tx buffer doesn't support.\n", __func__));
+      DEBUG ((DEBUG_INFO, "%a Rx/Tx buffer isn't supported.\n", __func__));
```
makubacki: To a casual reader (me), the `ArmFfaLibGetFeatures()` prototype in ArmFfaLib.h is:

```c
/**
  Get FF-A features.

  @param [in]   Id               Feature id or function id
  @param [in]   InputProperties  Input properties according to Id
  @param [out]  Property1        First Property value.
  @param [out]  Property2        Second Property value.

  @retval EFI_SUCCESS
  @retval Others                 Error

**/
EFI_STATUS
EFIAPI
ArmFfaLibGetFeatures (
  IN  UINT32  Id,
  IN  UINT32  InputProperties,
  OUT UINTN   *Property1,
  OUT UINTN   *Property2
  );
```

1. The success meaning is not provided in the API
2. Is it expected that a successful result means "Rx/Tx buffer isn't supported."?

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.c
```diff
@@ -91,39 +91,80 @@ ArmFfaSecLibConstructor (
...
-    DEBUG ((DEBUG_ERROR, "%a: Failed to create Rx/Tx Buffer Info Hob\n", __func__));
-    ArmFfaLibRxTxUnmap ();
-    return EFI_OUT_OF_RESOURCES;
+  return EFI_SUCCESS;
+}
+
+/**
+  Return partition or VM ID
+
+  @retval EFI_SUCCESS
```
makubacki: There needs to be a description for what `EFI_SUCCESS` means.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -55,24 +52,34 @@ ArmFfaLibGetRxTxBuffers (
...
+  /*
+   * Attempt to find the allocation HOB, if it exists then the
+   * Rx/Tx buffers have already been mapped.
+   */
+  RxTxBufferAllocationHob = FindRxTxBufferAllocationHob (TRUE);
+  if (RxTxBufferAllocationHob == NULL) {
     return EFI_NOT_READY;
   }
 
+  BufferSize = PcdGet64 (PcdFfaTxRxPageCount) * EFI_PAGE_SIZE;
```
makubacki: Was there a reason to avoid `EFI_PAGES_TO_SIZE()`?

```suggestion
  BufferSize = EFI_PAGES_TO_SIZE (PcdGet64 (PcdFfaTxRxPageCount));
```

Other places in the file perform a similar operation.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -205,11 +216,50 @@ ArmFfaLibRxTxMap (
...
-  mRxBuffer = RxBuffer;
+  BufferInfo = BuildGuidHob (&gArmFfaRxTxBufferInfoGuid, sizeof (ARM_FFA_RX_TX_BUFFER_INFO));
+  if (BufferInfo == NULL) {
+    DEBUG ((DEBUG_ERROR, "%a: Failed to create Rx/Tx Buffer Info Hob\n", __func__));
+    Status = EFI_OUT_OF_RESOURCES;
+    goto Unmap;
+  }
+
+  RxTxBufferAllocationHob = GetRxTxBufferAllocationHob ((EFI_PHYSICAL_ADDRESS)((UINTN)TxBuffer), PcdGet64 (PcdFfaTxRxPageCount) * EFI_PAGE_SIZE * 2, FALSE);
+  ASSERT (RxTxBufferAllocationHob != NULL);
```
makubacki: Do not continue to dereference `RxTxBufferAllocationHob` (next line) when asserts are disabled.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.h
Raymond-MS: You are correct, I see no reason why I can't just return a BOOLEAN. Fixed.

MdeModulePkg/Library/ArmFfaLib/ArmFfaDxeLib.c
Raymond-MS: Good catch, fixed!

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
Raymond-MS: Correct. Updated!

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.c
Raymond-MS: This is not something I wrote, I just updated the comment to read better. My understanding is that the feature being requested is dependent on the ID passed in. The comment mentions that if the request for ARM_FID_FFA_PARTITION_INFO_GET_REGS returns UNSUPPORTED it means the RX/TX buffers aren't supported but it's not considered an error because the RX/TX buffers may not be required.

The check is if (!EFI_ERROR (Status))

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.c
Raymond-MS: Fixed!

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
Raymond-MS: No, this was just how it was previously done in the file.  This is also how it is done in ArmFfaRxTxMap.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
Raymond-MS: This was how it was previously written in ArmFfaSecLib. What is your recommendation for resolving this?

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
makubacki: For the lines you've modified, I recommend using the macro.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
makubacki: Something to keep in mind is that the function is ~120 lines long with two labels so it makes reasoning about error handling more complicated than a function with less direct implementation.

- The function directly returns errors up to Line 189 (when `Buffers` is allocated)
- If `ArmCallFfa()` fails, the function goes to `ErrorHandler` which just frees `Buffers` before returning the status code
- If `ArmCallFfa()` succeeds, it looks like the expectation is that logic in the `Unmap` label needs to execute before returning

Based on that, I'd say something like:

```c
  RxTxBufferAllocationHob = GetRxTxBufferAllocationHob ((EFI_PHYSICAL_ADDRESS)((UINTN)TxBuffer), EFI_PAGES_TO_SIZE ((UINTN)PcdGet64 (PcdFfaTxRxPageCount)) * 2, FALSE);
  if (RxTxBufferAllocationHob == NULL) {
    Status = EFI_OUT_OF_RESOURCES;
    goto Unmap;
  }
```

makubacki: Another issue to consider is that this pattern which is used on what is currently [Line 222](https://github.com/Raymond-MS/edk2/blob/ad1b2a9ffa290f96605a85305f2bd047622bcf56/MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c#L219-L224) loses the `Status` value because the first thing that happens in the `Unmap` label is `Status` is replaced unconditionally:

```c
Unmap:
  Status = ArmFfaLibPartitionIdGet (&PartId);
  if (EFI_ERROR (Status)) {
    DEBUG ((DEBUG_ERROR, "%a, Failed to acquire partition ID. Status: %r\n", __func__, Status));
    goto ErrorHandler;
  }
```

With that in mind, maybe update Line 222 to not set `Status` since that is not useful and confusing and update this code to be:

```c
  RxTxBufferAllocationHob = GetRxTxBufferAllocationHob ((EFI_PHYSICAL_ADDRESS)((UINTN)TxBuffer), EFI_PAGES_TO_SIZE ((UINTN)PcdGet64 (PcdFfaTxRxPageCount)) * 2, FALSE);
  if (RxTxBufferAllocationHob == NULL) {
    goto Unmap;
  }
```

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.c
makubacki: If that's the logic, the check is wrong.

> if the request for ARM_FID_FFA_PARTITION_INFO_GET_REGS returns UNSUPPORTED it means the RX/TX buffers aren't supported

Would be:

```c
    if (Status == EFI_UNSUPPORTED) {
      DEBUG ((DEBUG_INFO, "%a Rx/Tx buffer isn't supported.\n", __func__));
    }
```

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
Raymond-MS: Sounds good! Updated!

I also noticed that updating Status is much worse than I thought because if the Unmap call is successful it will end up returning SUCCESS from Map which isn't what we want. Instead, I think we should force an EFI_OUT_OF_RESOURCES before getting to ErrorHandler to make sure that an error is still returned. We should only ever get to the Unmap label in those instances.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
Raymond-MS: Will do, thanks!

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
makubacki: Returning an error in that case in the main point, but you could also use a second status variable that is scoped within the `Unmap` label so the incoming `Status` value is preserved.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.c
Raymond-MS: It took me a bit of time but I believe I finally understand this line of code. I don't think it should change. Basically, this line of code is only called if Map fails with UNSUPPORTED which means Rx/Tx buffers are unsupported. The printout only happens if the GetFeatures call returns success which means INFO_REGS is an alternative to getting the partition info. That is why the printout shows that Rx/Tx buffers are not supported. I hope this clears things up!

makubacki: That makes sense. I think a reason this sticks out as confusing is that the determination that Rx/Tx buffers are not supported was made earlier after this:

```c
  Status = ArmFfaLibRxTxMap ();
  if (EFI_ERROR (Status)) {
    if (Status != EFI_UNSUPPORTED) {
      return Status;
    }
  }
```

In which case, the message would be naturally placed after that:

```c
  Status = ArmFfaLibRxTxMap ();
  if (EFI_ERROR (Status)) {
    if (Status != EFI_UNSUPPORTED) {
      return Status;
    }

    DEBUG ((DEBUG_INFO, "%a Rx/Tx buffer isn't supported.\n", __func__));
  }
```

Then, this message could say something relevant to the action just performed like:

```c
    Status = ArmFfaLibGetFeatures (
               ARM_FID_FFA_PARTITION_INFO_GET_REGS,
               0x00,
               &Property1,
               &Property2
               );
    if (!EFI_ERROR (Status)) {
      DEBUG ((DEBUG_INFO, "%a INFO_REGS is available as an alternative to Rx/Tx.\n", __func__));
    }
```

Raymond-MS: I am fine with adding that, it makes things clearer. :)

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -1172,3 +1149,48 @@ GetRxTxBufferMinSizeAndAlign (
 
   return EFI_SUCCESS;
 }
+
```
leiflindholm, line 1152: This breakout is useful, but unrelated to the described change. Please break it out as a separate preceding commit.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.h
```diff
@@ -37,16 +34,21 @@ FfaArgsToEfiStatus (
   );
 
 /**
-  Common ArmFfaLib Constructor.
+  Common ArmFfaLib init.
```
leiflindholm, line 37: If ArmFfaLibCommonInit is _not_ the constructor, then this is a correction of 
If ArmFfaLibCommonInit is the common part called by constructors of all variants of the lib, then this change discards useful information (although the previous text was sloppy and could do with clarification).

Regardless, that is an unrelated change that belongs in its own commit,

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
```diff
@@ -102,24 +106,48 @@ ArmFfaPeiLibConstructor (
...
-    if (Status == EFI_UNSUPPORTED) {
-      /*
-       * EFI_UNSUPPORTED return from ArmFfaLibCommonInit() means
-       * FF-A interface doesn't support.
-       * However, It doesn't make failure of loading driver/library instance
-       * (i.e) ArmPkg's MmCommunication Dxe/PEI Driver uses as well as SpmMm.
-       * So If FF-A is not supported the the MmCommunication Dxe/PEI falls
-       * back to SpmMm.
-       * For this case, return EFI_SUCCESS.
+  Status = (**PeiServices).RegisterForShadow (FileHandle);
```
leiflindholm, line 112: Does this non-trivial change belong in a separate commit following the one stopping all other code from directly accessing the globals?

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
```diff
@@ -102,24 +106,48 @@ ArmFfaPeiLibConstructor (
   EFI_HOB_MEMORY_ALLOCATION  *RxTxBufferAllocationHob;
   UINTN                      Property1;
   UINTN                      Property2;
+  UINT16                     PartId;
```
leiflindholm, line 109: Ugh, this is adding more live variables and ~20 more lines to an already too-long ~120 line function. Please consider simplifying the existing function before adding complexity.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -96,21 +103,25 @@ ArmFfaLibRxTxMap (
...
+  UINTN                      Property1;
+  UINTN                      Property2;
+  UINTN                      MinSizeAndAlign;
+  UINTN                      MaxSize;
+  VOID                       *Buffers;
+  VOID                       *TxBuffer;
+  VOID                       *RxBuffer;
+  ARM_FFA_RX_TX_BUFFER_INFO  *BufferInfo;
+  EFI_HOB_MEMORY_ALLOCATION  *RxTxBufferAllocationHob;
+  UINT16                     PartId;
```
leiflindholm, line 117: Ugh, this is adding more live variables and ~50 more lines to an already too-long ~120 line function. Please consider simplifying the existing function before adding complexity.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -123,7 +134,7 @@ ArmFfaLibRxTxMap (
   if (EFI_ERROR (Status)) {
     DEBUG ((
       DEBUG_ERROR,
-      "%a: Failed to get RX/TX buffer property... Status: %r\n",
```
leiflindholm, line 126: I don't mind the tidying, but it belongs in a separate commit.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -159,8 +170,8 @@ ArmFfaLibRxTxMap (
 
   MaxSize = ((MaxSize == 0) ? MAX_UINTN : (MaxSize * MinSizeAndAlign));
 
-  if ((MinSizeAndAlign > (PcdGet64 (PcdFfaTxRxPageCount) * EFI_PAGE_SIZE)) ||
-      (MaxSize < (PcdGet64 (PcdFfaTxRxPageCount) * EFI_PAGE_SIZE)))
+  if ((MinSizeAndAlign > (EFI_PAGES_TO_SIZE (PcdGet64 (PcdFfaTxRxPageCount)))) ||
```
leiflindholm, line 173: I don't mind the tidying, but it belongs in a separate commit.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -179,7 +190,7 @@ ArmFfaLibRxTxMap (
   }
 
   TxBuffer = Buffers;
-  RxBuffer = Buffers + (PcdGet64 (PcdFfaTxRxPageCount) * EFI_PAGE_SIZE);
+  RxBuffer = Buffers + (EFI_PAGES_TO_SIZE (PcdGet64 (PcdFfaTxRxPageCount)));
```
leiflindholm, line 193: I don't mind the tidying, but it belongs in a separate commit.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -222,8 +283,8 @@ ArmFfaLibRxTxMap (
   Rx/Tx buffer is registered only once per partition.
 
   @retval EFI_SUCCESS
-  @retval EFI_INVALID_PARAMETERS               Already unregistered
-  @retval EFI_UNSUPPORTED                      Not supported
+  @retval EFI_INVALID_PARAMETER  Already unregistered
```
leiflindholm, line 286: I don't mind the tidying, but it belongs in a separate commit.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -248,26 +327,18 @@ ArmFfaLibRxTxUnmap (
...
 
-  mTxBuffer = NULL;
-  mRxBuffer = NULL;
-
   return EFI_SUCCESS;
 }
 
 /**
-  Update Rx/TX buffer information.
+  Update Rx/Tx buffer information.
```
leiflindholm, line 341: I don't mind the tidying, but it belongs in a separate commit.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -248,26 +327,18 @@ ArmFfaLibRxTxUnmap (
...
 
-  @param  BufferInfo            Rx/Tx buffer information.
+  @param  BufferInfo  Rx/Tx buffer information.
```
leiflindholm, line 343: I don't mind the tidying, but it belongs in a separate commit.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -276,14 +347,24 @@ UpdateRxTxBufferInfo (
...
+    BufferInfo->TxBufferAddr = (UINTN)RxTxBufferAllocationHob->AllocDescriptor.MemoryBaseAddress;
+    BufferInfo->TxBufferSize = EFI_PAGES_TO_SIZE (PcdGet64 (PcdFfaTxRxPageCount));
+    BufferInfo->RxBufferAddr = (UINTN)RxTxBufferAllocationHob->AllocDescriptor.MemoryBaseAddress + BufferInfo->TxBufferSize;
+    BufferInfo->RxBufferSize = EFI_PAGES_TO_SIZE (PcdGet64 (PcdFfaTxRxPageCount));
+  }
 }
 
 /**
-  Find Rx/TX buffer memory allocation hob.
+  Find Rx/Tx buffer memory allocation hob.
```
leiflindholm, line 369: I don't mind the tidying, but it belongs in a separate commit.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -297,17 +378,18 @@ FindRxTxBufferAllocationHob (
...
+   * Mapping the Rx/Tx buffer should have generated the HOB, if not,
+   * then the allocation HOB will not be found and NULL will be returned.
+   * Note that BufferAddr and BufferSize are ignored when UseGuid is TRUE.
+   */
+  return GetRxTxBufferAllocationHob (0, 0, TRUE);
 }
 
 /**
-  Remap Rx/TX buffer with converted Rx/Tx Buffer address after
+  Remap Rx/Tx buffer with converted Rx/Tx Buffer address after
```
leiflindholm, line 394: I don't mind the tidying, but it belongs in a separate commit.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
Raymond-MS: This isn't a breakout and it is directly related to the described change regarding removal of the globals. ArmFfaCommonInit used to have this code and it set the global gFfaSupported. There was no mechanism to query if FF-A is supported. The IsFfaSupported function only returned the global, this function was necessary to remove the global. This is used in both SEC and PEI when permanent memory is not available.

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
Raymond-MS: I don't know if I fully agree. PEI is a special case because it can have permanent memory available which is where this check comes in. If it is available, it is able to cache the init information, otherwise, it needs to act like SEC and query the information each time. In my opinion, removing access to the globals would need to have both of these implementations for PEI because of this reason.

MdeModulePkg/Library/ArmFfaLib/ArmFfaPeiLib.c
Raymond-MS: I am confused. These are necessary for when permanent memory isn't available. We need to pass in temporary variables to populate in ArmCommonInit. Do you have any recommendations? The only other option would be to have passing in variables be optional but that would add extra checks in ArmCommonInit and its calling functions to make sure to check for NULL pointers. I feel as though that would add more lines than just a couple of variables at the top of the function.

Raymond-MS:
@leiflindholm I believe I have addressed most of your concerns. The commits were split out into more meaningful and easier to digest changes. Please let me know if there is anything else you need from me. Thanks!

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
leiflindholm: Did I get sniped by one of githubs wanting to show me yeat another new interface?
Because when I reviewed this I could have sworn this was part of a single megacommit. Totally happy with this standalone variant.

Raymond-MS: I ended up splitting it out anyway since it is possible to create the function but not use it. I ended up using it in the next commit when removing the globals from ArmFfaCommonInit. I should have split things out into more relevant commits related to their changes. This is much more organized and I appreciate the feedback. Hopefully this is to your liking. Thanks!

lgao4:
If no other comment, I will add push label this week.

PR #12104, opened 2026-01-31
Fix GCC 16 Uninitialized Variable Warnings

r1chard-lyu:
# Description
GCC16 has enhanced its static analysis to detect variables that are assigned
conditionally in certain code paths but may be used without initialization in all
execution paths. This PR addresses a series of uninitialized variable warnings
reported by GCC16 across multiple EDK2 packages.

The fixes involve initializing variables to appropriate default values to ensure 
the code is compatible with GCC16's stricter warnings while maintaining the 
original logic and behavior.

The patch series in https://github.com/tianocore/edk2/pull/12079 fixes
various unused variable issues in edk2. These patches must be applied
alongside this series to enable a successful OVMF build using GCC 16.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
1. git submodule update --init
2. make -C BaseTools
3. . edksetup.sh
4. build -a X64 -p OvmfPkg/OvmfPkgX64.dsc -t GCC5

## Integration Instructions
1. curl -L https://github.com/tianocore/edk2/pull/12079.patch -o 12079.patch
2. git apply 12079.patch
3. Switch to GCC 16

### DIFF
--- a/OvmfPkg/AmdSevDxe/AmdSevDxe.c
+++ b/OvmfPkg/AmdSevDxe/AmdSevDxe.c
@@ -204,6 +204,7 @@ AmdSevDxeEntryPoint (
   UINTN                                     Index;
   CONFIDENTIAL_COMPUTING_SNP_BLOB_LOCATION  *SnpBootDxeTable;
 
+  SnpBootDxeTable = NULL;
   //
   // Do nothing when SEV is not enabled
   //


ardbiesheuvel:
Please DO NOT merge back the master branch into your PR - just ignore the GitHub warning about this

r1chard-lyu:
> Please DO NOT merge back the master branch into your PR - just ignore the GitHub warning about this

Sorry about that! I’ve reset everything back to how it was.

UefiCpuPkg/Library/MpInitLib/X64/AmdSev.c
```diff
@@ -277,7 +277,7 @@ SevSnpCreateAP (
   UINTN                          MaxIndex;
   UINT32                         ApicId;
   EFI_HOB_GUID_TYPE              *GuidHob;
-  GHCB_APIC_IDS                  *GhcbApicIds;
+  GHCB_APIC_IDS                  *GhcbApicIds = NULL;
```
ardbiesheuvel: What is the point of this? I understand that dereferencing an uninitialized pointer is problematic, but how is dereferencing a NULL pointer any better? If the logic is flawed, it needs to be fixed. If the compiler is wrong, we need to fix the compiler - there is still time until GCC 16 is released.

tlendacky: I think it is a false positive. GhcbApicIds is set 'if CpuMpData->InitFlag == ApInitConfig' and then only referenced 'if (CpuMpData->InitFlag == ApInitConfig)'.

OvmfPkg/AmdSevDxe/AmdSevDxe.c
```diff
@@ -202,7 +202,7 @@ AmdSevDxeEntryPoint (
   EFI_GCD_MEMORY_SPACE_DESCRIPTOR           *AllDescMap;
   UINTN                                     NumEntries;
   UINTN                                     Index;
-  CONFIDENTIAL_COMPUTING_SNP_BLOB_LOCATION  *SnpBootDxeTable;
+  CONFIDENTIAL_COMPUTING_SNP_BLOB_LOCATION  *SnpBootDxeTable = NULL;
```
khaliid2040: Can we please seperate the declaration and assignment. This is how variables are initialized in edk2. 

Example:

```suggestion
  CONFIDENTIAL_COMPUTING_SNP_BLOB_LOCATION  *SnpBootDxeTable;

SnpBootDxeTable = NULL;
```

please do like that on all your patches. Also, consider about line alignment so uncrutify don't complain.

OvmfPkg/AmdSevDxe/AmdSevDxe.c
```diff
@@ -204,6 +204,7 @@ AmdSevDxeEntryPoint (
   UINTN                                     Index;
   CONFIDENTIAL_COMPUTING_SNP_BLOB_LOCATION  *SnpBootDxeTable;
 
+  SnpBootDxeTable = NULL;
```
ardbiesheuvel, line 207: Could this also be fixed by setting `*CcBlobPtr` to `NULL` on line 46?

UefiCpuPkg/Library/MpInitLib/X64/AmdSev.c
r1chard-lyu: @tlendacky It seems that the compiler is indicating that GhcbApicIds is set only when ProcessorNumber < 0 and CpuMpData->InitFlag == ApInitConfig, but it is dereferenced at line 327. If it is not set, its value would be undefined, so GCC 16 reports it as an error.

r1chard-lyu: @ardbiesheuvel I agree this is only a rough workaround to silence the GCC 16 error during compilation. I think GCC 16 is flagging this because GhcbApicIds is assigned only along one control flow path, but later it is used along another path that still appears possible, so the compiler conservatively concludes that it may be used uninitialized. I also agree that the underlying logic issue should be properly fixed. 

However, for this commit, would it be acceptable to initialize GhcbApicIds to NULL and add ASSERT (GhcbApicIds != NULL); immediately before it is dereferenced at line 327?

OvmfPkg/AmdSevDxe/AmdSevDxe.c
r1chard-lyu: Thanks for the heads-up. I've finished it.

ardbiesheuvel:
Which version of GCC 16 are you using? I cannot reproduce any of these issues with the build in Debian experimental (gcc version 16.0.0 202512140)

I am finding other issues, though, so I'll send out some fixes for those.

ardbiesheuvel:
#12106

OvmfPkg/AmdSevDxe/AmdSevDxe.c
khaliid2040: > Thanks for the heads-up. I've finished it.

Thanks, please resolve this. Only you have permission to resolve review comments of your PRs.

r1chard-lyu:
> Which version of GCC 16 are you using? I cannot reproduce any of these issues with the build in Debian experimental (gcc version 16.0.0 202512140)
> 
> I am finding other issues, though, so I'll send out some fixes for those.

Because we have to apply the patch series from [PR 12079](https://github.com/tianocore/edk2/pull/12079) first before we encounter these issues. My GCC version is the openSUSE experimental build, 16.0.1 (20260113).

r1chard-lyu:
Thank you to the reviewers. I suspect some of the issues might be false positives. 
I will investigate these instances, make the necessary adjustments, and resubmit.

UefiCpuPkg/Library/MpInitLib/X64/AmdSev.c
tlendacky: > @tlendacky It seems that the compiler is indicating that GhcbApicIds is set only when ProcessorNumber < 0 and CpuMpData->InitFlag == ApInitConfig, but it is dereferenced at line 327. If it is not set, its value would be undefined, so GCC 16 reports it as an error.

You are still in the 'ProcessorNumber < 0' path and it is only dereferenced if 'CpuMpData->InitFlag == ApInitConfig' which is also the same condition when it is assigned.

UefiCpuPkg/Library/MpInitLib/X64/AmdSev.c
```diff
@@ -281,7 +281,7 @@ SevSnpCreateAP (
   volatile MP_CPU_EXCHANGE_INFO  *ExchangeInfo;
 
   ASSERT (CpuMpData->MpCpuExchangeInfo->BufferStart < 0x100000);
-
+  GhcbApicIds  = NULL;
```
tlendacky, line 284: I still feel this is a false positive, but I'm fine if with doing this if desired.

UefiCpuPkg/Library/MpInitLib/PeiMpLib.c
```diff
@@ -75,6 +75,8 @@ SaveCpuMpData (
   MP_HAND_OFF_CONFIG  MpHandOffConfig;
   UINTN               MpHandOffSize;
 
+  HobBase       = 0;
+  MpHandOff     = NULL;
```
tlendacky, line 79: Both of these are set/initialized because Index = 0 at the start of the loop and the first iteration of the loop will satisfy the 'if (Index % MaxCpusPerHob == 0) {' conditional and assign HobBase and MpHandOff. So another false positive.

OvmfPkg/Library/CcExitLib/CcExitVcHandler.c
```diff
@@ -1439,6 +1439,7 @@ CpuidExit (
   UINT32   Ecx;
   UINT32   Edx;
 
+  XCr0  = 0;
```
tlendacky, line 1442: Please set XCr0 = 0 in the (new) else path of the 'if (EaxIn == CPUID_EXTENDED_STATE) {'

r1chard-lyu: Thank you for the suggestion. I resubmitted it [here](https://github.com/tianocore/edk2/pull/12116).

PR #12106, opened 2026-02-01
MdeModulePkg: Remove unused variables

ardbiesheuvel:
GCC 16 spots some occurrences where variables are initialized and updated, but never read back, breaking the build. Just drop them.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?


khaliid2040:
But this has been already fixed in #12079. Generally, i would prefer #12079 over this one because it is covering more.

ardbiesheuvel:
> But this has been already fixed in #12079. Generally, i would prefer #12079 over this one because it is covering more.

Thanks for pointing that out - I wasn't cc'ed on that so I had no idea.

PR #12111, opened 2026-02-02
Maintainers.txt: Update M and R for StandaloneMmPkg and UefiCpuPkg

jiaxinwu:
# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?


leiflindholm (changes requested):
Please add an actual PR description. We can figure out what's going on from the patch, but we shouldn't have to. And please write a sensible commit message too, the current one contains no information beyond what file is being modified.

Aside from that, you may want to consider adding an entry in .mailmap.

PR #12114, opened 2026-02-03
SecurityPkg: AuthVariableLib: Handle empty signature lists

kuqin12:
# Description

The current implementation fails to set authenticated variables when the signature list is empty. This can legitimately occur for dbx when no signatures are revoked after a certificate rotation.

Update the logic to explicitly handle empty signature lists, avoiding an implicit dependency on the variable being absent from variable storage.

- [ ] Breaking change?
- [x] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested on a physical platform using an empty signature list dbx input.

## Integration Instructions

N/A


SecurityPkg/Library/AuthVariableLib/AuthService.c
```diff
@@ -591,7 +591,7 @@ CheckSignatureListFormat (
       CertData   = (EFI_SIGNATURE_DATA *)((UINT8 *)SigList + sizeof (EFI_SIGNATURE_LIST) + SigList->SignatureHeaderSize);
       CertLen    = SigList->SignatureSize - sizeof (EFI_GUID);
       RsaContext = NULL;
-      if (!RsaGetPublicKeyFromX509 (CertData->SignatureData, CertLen, &RsaContext)) {
+      if ((CertLen > 0) && !RsaGetPublicKeyFromX509 (CertData->SignatureData, CertLen, &RsaContext)) {
```
jyao1, line 594: Shouldn't this be handled inside of `RsaGetPublicKeyFromX509()`?

https://github.com/tianocore/edk2/blob/master/CryptoPkg/Library/BaseCryptLib/Pk/CryptX509.c#L601
https://github.com/tianocore/edk2/blob/master/CryptoPkg/Library/BaseCryptLib/Pk/CryptX509.c#L901

As such, we do not need to change all callers.

kuqin12: Thanks for the suggestion. This PR is updated.

kuqin12:
@jyao1 @mxu9 @Flickdm @liyi77 could you please re-review the PR and let me know if you have any further feedback? Thanks in advance!

SecurityPkg/Library/AuthVariableLib/AuthService.c
liyi77: I'm not sure if this is the correct way, the function returning TRUE indicates that the key was retrieved succeed and RsaContext was initialized. Modifying the return logic might be a surprise for other use cases.

Is your purpose to provide a method for deleting dbx? The correct approach is to provide an empty payload instead of an empty sig list in current code.

Flickdm: The intention here is that a Empty Signature List is an edge case that should be supported. 

Specifically due to transitioning of Certificates (like the one occurring now) - when we rotate the KEK and the DB, we're in a position where there are no valid contents for the DBX.(since we haven't revoked anything)

We could either leave the variable unset, or as in this case - give it an empty signature list.

liyi77: I am okay with using an empty SIG list to revoke a key variable,
but I'd prefer to implement it in SecurityPkg rather than CryptoPkg, as it's just a specific use case for EFI_SIG_DATA and doesn't have much relationship with Crypto.

kuqin12: Hmm, I think if we just consider from the code abstract perspective, having `RsaGetPublicKeyFromX509` handle a zero-sized income should be a better way to go. This should be treated as a special but valid case. And as Jiewen suggested, it would allow all callers to have this case handled for free.

However, practically speaking, if `RsaGetPublicKeyFromX509` returns TRUE with a `NULL` in the `RsaContext`, this might be something the callers are not ready to handle and could cause other side effects.

@jyao1 would you be okay to take the original change in `SecurityPkg` rather than changing the lower-level `CryptoPkg` function for the reason above?

jyao1: Sorry, I think I misunderstood the fix before.
Now, I am thinking the following:
1) CryptoPkg: We need handle "CertSize == 0" explicitly, by returning FALSE, not TRUE. It is defensive programming for invalid case.
2) SecurityPkg: If the caller think "CertSize == 0" is a valid case, then the caller should handle it specially. With that, I agree that we should take the original change in SecurityPkg. And we need ensure to change all places.

kuqin12: @jyao1 @liyi77 Thanks for the input. This PR is updated per discussion. Please let me know if you have any further comments.

CryptoPkg/Library/BaseCryptLib/Pk/CryptX509.c
```diff
@@ -602,6 +602,14 @@ RsaGetPublicKeyFromX509 (
     return FALSE;
   }
 
+  //
+  // If CertSize is 0, return FALSE to be safe.
+  //
+  if (CertSize == 0) {
+    *RsaContext = NULL;
+    return TRUE;
```
jyao1: return FALSE.

kuqin12: Thank you for the catch! This is updated.

kuqin12:
@jyao1 @liyi77 @Flickdm @mxu9 The change is updated per discussion in this PR. Do you happen to have any chance to re-review the change? Thanks in advance!

PR #12115, opened 2026-02-03
StandaloneMmPkg/MmCommunicationDxe: Enforce CommSize validation

makubacki:
# Description

CommSize is an optional parameter to ProcessCommunicationBuffer().

If it is provided, the size it points to must match the total incoming comm buffer size (header + data) where that size is calculated as needed for the comm header version.

If the sizes do not match, the copy from the user provided comm buffer (pointed to by the CommBuffer parameter) to the shared comm buffer (mMmCommonBuffer.PhysicalStart) could be incorrect, for example, only copying part of the buffer.

This change returns EFI_INVALID_PARAMETER if the provided CommSize does not match the expected size instead of asserting to ensure this behavior is enforced in scenarios when asserts are disabled and to allow the caller to operate on the return status.

In addition, for the copy from the user provided buffer to the common buffer to be valid, the total size must be less than or equal to the size of the common buffer.  This change adds that validation as well.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- StandaloneMmPkg CI
- MM communication in QEMU and a physical Intel platform using StandaloneMmPkg/Drivers/MmCommunicationDxe

## Integration Instructions

- N/A


StandaloneMmPkg/Drivers/MmCommunicationDxe/MmCommunicationDxe.c
```diff
@@ -112,7 +112,9 @@ ProcessCommunicationBuffer (
   }
 
   if (CommSize != NULL) {
-    ASSERT (*CommSize == BufferSize);
+    if ((*CommSize != BufferSize) || (*CommSize > EFI_PAGES_TO_SIZE (mMmCommonBuffer.NumberOfPages))) {
```
kuqin12: Agree with the change. But would it just be `(*CommSize > EFI_PAGES_TO_SIZE (mMmCommonBuffer.NumberOfPages))`? A `BufferSize` larger than the common buffer size would not be able to get through, seemly.

StandaloneMmPkg/Drivers/MmCommunicationDxe/MmCommunicationDxe.c
```diff
@@ -112,7 +112,9 @@ ProcessCommunicationBuffer (
   }
```
kuqin12, line 113: I think we should also add a check for `BufferSize` to ensure that it is smaller than `EFI_PAGES_TO_SIZE (mMmCommonBuffer.NumberOfPages)`

StandaloneMmPkg/Drivers/MmCommunicationDxe/MmCommunicationDxe.c
makubacki: I considered these two separate pieces of validation of the given input size.

1. `(*CommSize != BufferSize)`

   Verifies that the communicate header size (in `CommBuffer`) is correct for the header format.

2. `(*CommSize > EFI_PAGES_TO_SIZE (mMmCommonBuffer.NumberOfPages)`

   Verifies that the size will fit in the common buffer.

---

Are you saying (2) negates the need for (1)?

kuqin12: Sorry, the comments were mixed up. I think the other comment of checking `BufferSize` will negate your second check here.

The need for `BufferSize` inspection is to make sure that the incoming buffer can be properly copied into the designated common buffer for communicate. If we have that in place, combined with `if (*CommSize != BufferSize)` check, the second point should be covered.

makubacki: Thanks for the clarification. That's a good suggestion and I've pushed the update.

StandaloneMmPkg/Drivers/MmCommunicationDxe/MmCommunicationDxe.c
makubacki: Update included in response to [this comment thread](https://github.com/tianocore/edk2/pull/12115#discussion_r2757489423).

PR #12117, opened 2026-02-03
MdeModulePkg: Optimize the code handling of hotkey registration

EricGao2015:
# Description

The return value of HiiGetString function should be freed by using FreePool().
By the way, add a check for NewString before calling RegisterHotKey().
Following Mike and Ashraf's susggestion, ASSERT is preserved.

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Universal/DisplayEngineDxe/FormDisplay.c
+++ b/MdeModulePkg/Universal/DisplayEngineDxe/FormDisplay.c
@@ -4330,14 +4330,18 @@ InitializeDisplayEngine (
     HotKey.ScanCode    = SCAN_F10;
     NewString          = HiiGetString (gHiiHandle, STRING_TOKEN (FUNCTION_TEN_STRING), NULL);
     ASSERT (NewString != NULL);
-    FormBrowserEx2->RegisterHotKey (&HotKey, BROWSER_ACTION_SUBMIT, 0, NewString);
-    FreePool (NewString);
+    if (NewString != NULL) {
+      FormBrowserEx2->RegisterHotKey (&HotKey, BROWSER_ACTION_SUBMIT, 0, NewString);
+      FreePool (NewString);
+    }
 
     HotKey.ScanCode = SCAN_F9;
     NewString       = HiiGetString (gHiiHandle, STRING_TOKEN (FUNCTION_NINE_STRING), NULL);
     ASSERT (NewString != NULL);
-    FormBrowserEx2->RegisterHotKey (&HotKey, BROWSER_ACTION_DEFAULT, EFI_HII_DEFAULT_CLASS_STANDARD, NewString);
-    FreePool (NewString);
+    if (NewString != NULL) {
+      FormBrowserEx2->RegisterHotKey (&HotKey, BROWSER_ACTION_DEFAULT, EFI_HII_DEFAULT_CLASS_STANDARD, NewString);
+      FreePool (NewString);
+    }
   }
 
   return EFI_SUCCESS;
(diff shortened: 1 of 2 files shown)


MdeModulePkg/Universal/DriverSampleDxe/DriverSample.c
```diff
@@ -2202,10 +2202,12 @@ DriverSampleInit (
     NewString       = HiiGetString (mPrivateData->HiiHandle[0], STRING_TOKEN (FUNCTION_TEN_STRING), NULL);
     ASSERT (NewString != NULL);
```
makubacki, line 2203: This sample driver also sets a bad example of showing asserts being misused in place of proper handling.

EricGao2015: Hi Michael, I have updated the commits, please check whether they align with your suggestion. 
The PR title should also be updated but it can't be modified now. I'll try to revise it after the realese of tag stable202602.

makubacki: It does. Thank you!

makubacki (approved):

MdeModulePkg/Universal/DisplayEngineDxe/FormDisplay.c
```diff
@@ -4329,15 +4329,17 @@ InitializeDisplayEngine (
     HotKey.UnicodeChar = CHAR_NULL;
     HotKey.ScanCode    = SCAN_F10;
     NewString          = HiiGetString (gHiiHandle, STRING_TOKEN (FUNCTION_TEN_STRING), NULL);
```
AshrafAliS, line 4331: we can still keep those ASSERT right to catch the issue in debug BIOS.

AshrafAliS (approved):

MdeModulePkg/Universal/DisplayEngineDxe/FormDisplay.c
EricGao2015: @AshrafAliS Hi Ashraf, I seen you have approved the changes, so can I ingore your suggestion? I think ASSERT is dispensable at that case.

EricGao2015: > @AshrafAliS Hi Ashraf, I seen you have approved the changes, so can I ignore your suggestion? I think ASSERT is dispensable at that case.

@AshrafAliS Hi Ashraf, please reply to me if you have time.

EricGao2015: @AshrafAliS ping again

mikebeaton: I think you should keep the ASSERTs. ASSERT was there for a reason, to flag up something that should not be happening, at least in DEBUG and NOOPT builds. The problem was that, in a lot of places, EDK 2 code previously treated ASSERT as if it 'handles' the error, when it doesn't because in RELEASE builds nothing happens, the code just continues. The pattern of a lot of recent updates has been to keep the ASSERT, but to back it up with some added reasonable behaviour for RELEASE builds.

EricGao2015: > I think you should keep the ASSERTs. ASSERT was there for a reason, to flag up something that should not be happening, at least in DEBUG and NOOPT builds. The problem was that, in a lot of places, EDK 2 code previously treated ASSERT as if it 'handles' the error, when it doesn't because in RELEASE builds nothing happens, the code just continues. The pattern of a lot of recent updates has been to keep the ASSERT, but to back it up with some added reasonable behaviour for RELEASE builds.

OK. Finally, I decide to preserve ASSERT.

EricGao2015:
@lgao4 @dandanbi Hi all, please review this PR, thank you.

mikebeaton (approved):

EricGao2015:
@lgao4 Hi Liming, this PR has submitted for a long time and it has been reviewed. Does it have any chance to be merged?

lgao4 (approved):

PR #12121, opened 2026-02-03
NetworkPkg/UefiPxeBcDxe: Initialize IPV4 token in IPV4 branch

makubacki:
# Description

EfiPxeBcStart() in PxeBcImpl.c has conditional code branches for IPV4 and IPV6.

- IPV4 branch should use EFI_IP4_COMPLETION_TOKEN which is Private->IcmpToken.
- IPV6 branch should use EFI_IP6_COMPLETION_TOKEN which is Private->Icmp6Token.

Right now, the IPv4 branch incorrectly initializes Private->Icmp6Token to EFI_NOT_READY. That is changed to Private->IcmpToken.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- PXE boot with server setup on a separate device (TFTP + DHCP

## Integration Instructions

- N/A

### DIFF
--- a/NetworkPkg/UefiPxeBcDxe/PxeBcImpl.c
+++ b/NetworkPkg/UefiPxeBcDxe/PxeBcImpl.c
@@ -193,14 +193,14 @@ EfiPxeBcStart (
     //
     // Create event and set status for token to capture ICMP error message.
     //
-    Private->Icmp6Token.Status = EFI_NOT_READY;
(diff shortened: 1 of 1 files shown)


SaloniKasbekar (approved):

pierregondois (approved):
I didn't run the code, but the patch seems sensible

mikebeaton (approved):

PR #12122, opened 2026-02-03
NetworkPkg/DxeNetLib: Initialize MediaState to EFI_NOT_READY

makubacki:
# Description

Initializes the `MediaState` output parameter to `EFI_NOT_READY` at the start of `DxeNetLibGetMediaState()`. This ensures that the status only reflects a successful media state after either `NetLibDetectMedia()` or `GetInfomation()` confirms media is present.

---

Upon extensive testing of the above change, it was found the some UNDI drivers do not report that they support media presence, yet they otherwise work. Second, it was found that if SNP media presence is used, it does not wait at all, causing premature failures. Thirdly, it was found that those premature failures can cause SNP shutdown and reinitialization to occur multiple times, unnecessarily.

That led to the second commit being added, described below. It maintains backward compatibility with UNDI drivers that do not indicate they support media presence correctly.

---

NetworkPkg: Add timeout polling for PXE media detection

Simplify the `NetLibDetectMediaWaitTimeout()` execution flow and add a
timeout on non-AIP media detection to allow NICs sufficient time to
initialize and report media state.

---

This single holistic change address two problems:

Problem 1: Network devices should properly report media presence.

The previous implementation checked `Aip->GetInformation()` and fell
back to `Snp->GetStatus()` and then returned media is present if SNP
returned that it did not support media detection.

In the end, this falsely reported media presence over the years has
led to even more UNDI drivers not supporting media presence properly
because of this behavior.

Problem 2: Multiple SNP shutdown and reinitialization processes can
occur impacting performance and stability.

`NetLibDetectMediaWaitTimeout()` in `DxeNetLib` is called ~15 places
in the codebase. Multiple callers might call into the function for
the same NIC such as `Dhcp4Dxe`, `PxeBcDxe`, etc.

For example:

```
  Snp->undi.initialize()  <-- By SimpleNetworkDriverStart

  Snp->undi.initialize()   <-- By NetLibDetectMedia()
  Snp->undi.initialize()  8000h:6h
  Snp->undi.initialize()

  Snp->undi.initialize()  <-- By NetLibDetectMedia()
  Snp->undi.initialize()  8000h:6h
  Snp->undi.initialize()
```

Initialization might also occur multiple times in the same flow, for
example, when `SnpUndi32Initialize()` first calls with
`PXE_OPFLAGS_INITIALIZE_DETECT_CABLE` and if that fails, again with
`PXE_OPFLAGS_INITIALIZE_DO_NOT_DETECT_CABLE`.

Collectively this can lead to SNP shutting down and reinitializing
multiple times when it might not need to impacting stability and
performance.

---

Notes:

1. The 20 second value (`PXEBC_CHECK_MEDIA_WAITING_TIME`) often given
   as the timeout (`Timeout` argument) to
   `NetLibDetectMediaWaitTimeout()` is not used for
   `NetLibDetectMedia()` flows, only `Aip->GetInformation()` calls.
2. We know some NICs take at least 1.5+ seconds to report back media
   presence after initialization when going through the
   `NetLibDetectMedia()` flow. That function supports two methods for
   media presence detection through SNP to the underlying UNDI driver
   (1) `GET_STATUS` command and (2) a flow to stop and restart SNP for
   drivers that do not support the `GET_STATUS` command.

---

This change solves the following problems:

1. Allows sufficient time for SNP to report media presence instead of
   simply returning.
2. Makes it clear to platform owners when media detection is not able
   to be determined so it is simply set to "media is present" to
   satisfy NIC FW that does not support actual media detection.
3. Partially due to (1) (and TPL guards recently introduced), prevent
   multiple SNP shutdown and reinitialization processes from occurring
   which thrash NIC stability and impact network boot performance.

This change intentionally maintains the same behavior when actual
media presence cannot be determined (i.e. AIP and SNP do not support
media detection) to be backward compatible.

---

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- PXE server setup on a separate device (TFTP + DHCP)

- Physical x86 client platform PXE boot
  - Does not need NIC OPROM workaround
- Physical x86 server platform PXE boot
  - Uses NIC OPROM workaround
 
## Integration Instructions

- N/A

### DIFF
--- a/NetworkPkg/Include/Library/NetLib.h
+++ b/NetworkPkg/Include/Library/NetLib.h
@@ -93,6 +93,16 @@ typedef UINT16  TCP_PORTNO;
 //
 #define MEDIA_STATE_DETECT_TIME_INTERVAL  1000000U
 
+//
+// Number of times to attempt to detect network media through SNP
+//
+#define SNP_MEDIA_DETECT_RETRY_ATTEMPTS  0
+
+//
+// Amount of time to wait for network media to be detected through SNP, in 100ns units
+//
+#define SNP_MEDIA_DETECT_WAITING_TIME  EFI_TIMER_PERIOD_SECONDS(2)
+
 #pragma pack(1)
 
 //
--- a/NetworkPkg/Library/DxeNetLib/DxeNetLib.c
+++ b/NetworkPkg/Library/DxeNetLib/DxeNetLib.c
@@ -2701,7 +2701,7 @@ NetLibDetectMediaWaitTimeout (
     return EFI_INVALID_PARAMETER;
   }
 
-  *MediaState = EFI_SUCCESS;
+  *MediaState = EFI_NOT_READY;
   MediaInfo   = NULL;
 
   //
--- a/NetworkPkg/Library/DxeNetLib/DxeNetLib.c
+++ b/NetworkPkg/Library/DxeNetLib/DxeNetLib.c
@@ -2438,7 +2438,168 @@ NetLibGetMacString (
 }
 
 /**
-  Detect media status for specified network device.
+  Get media status from Simple Network Protocol (SNP).
+
+  This function calls EFI_SIMPLE_NETWORK_PROTOCOL.GetStatus() to refresh
+  the media state and returns the current MediaPresent field from the SNP
+  mode data.
+
+  @param[in]  Snp           The Simple Network Protocol instance.
+  @param[out] MediaPresent  The pointer to receive the media present status.
+
+  @retval EFI_SUCCESS           MediaPresent is returned successfully.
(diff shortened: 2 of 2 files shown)


SaloniKasbekar (approved):

ardbiesheuvel:
Does this change affect platforms that do implement  `Aip->GetInformation()`?

makubacki:
> Does this change affect platforms that do implement `Aip->GetInformation()`?

It does not. I've tested it with platforms that support `Aip->GetInformation()`.

makubacki:
Please feel free to review, but I think I can simplify some of the loop logic in the current PR. I'm in the process of testing the change.

ardbiesheuvel (approved):

PR #12123, opened 2026-02-03
Acpi ffa

kuqin12:
# Description

Current Tcg2AcpiFfa will populate the partition ID in byte order of big-endian. This conflicts with the TCG ACPI Specification, which specifies the byte-order to be little-endian. Thus an industry defined structure is used to replace the byte arrays.

In addition, since the start method of FFA is only introduced in revision 5 of the TCG ACPI specification. A TPM2 table with FFA start method and lower than 5 revision should not be allowed. Hence a few checks are added based on new revision 5 assumptions.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested on proprietary hardware platform and booted to Windows desktop.

## Integration Instructions

N/A


kuqin12:
Superseded by https://github.com/tianocore/edk2/pull/12139

PR #12129, opened 2026-02-05
Upstream/redfishlib api

anushaKankipati:
# Description

Add a public API to create a Redfish service with explicit user
credentials. This allows callers to use user-provided credentials
instead of relying on automatic credential discovery (IPMI bootstrap).

The API supports HTTP Basic Auth and Redfish Session Login methods.
It performs comprehensive input validation:
- NULL checks for service info and credentials (when required)
- AuthMethod range validation (< AuthMethodMax)
- Empty username detection (empty passwords allowed)

Use cases:
- BMC management operations requiring administrative credentials
- Multi-tenant scenarios with different credential sets
- Testing with specific user accounts

Callers must:
- Securely handle credentials (no storage in plaintext)
- Free the service via RedfishCleanupService()
- Use AuthMethodNone when credentials are not required

The API delegates to RedfishCreateLibredfishService() which handles
the authentication flow automatically.

### DIFF
--- a/RedfishPkg/PrivateLibrary/RedfishLib/RedfishLib.c
+++ b/RedfishPkg/PrivateLibrary/RedfishLib/RedfishLib.c
@@ -100,6 +100,12 @@ ON_EXIT:
   @return     New created Redfish Service, or NULL if error happens.
 
 **/
+//
+// Maximum credential length for security validation.
+// This is a reasonable limit for HTTP Basic Auth credentials.
+//
+#define REDFISH_CREDENTIAL_MAX_LENGTH  256
+
 REDFISH_SERVICE
 EFIAPI
 RedfishCreateServiceWithCredential (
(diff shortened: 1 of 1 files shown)


nicklela (changes requested):
Thank you for upstreaming this change. I see the patch check failure. 

Image

The first commit message does not follow the format I think. Can you modify it to like second commit message? We also have to remove "Bug 5195725" and "Change-Id: I9764d2c2af08650dafd913160ade56fdbcc1552a" since they are company things.

changab:
Hi @anushaKankipati,
Instead of adding a new funciton and maintain it in RedfishLib,  I prefer to add a v2 EDKII_REDFISH_HTTP_PROTOCOL that provides an API to create the Redfish service with an user credential passed from the caller. This way, we can have a consistent implementation that acauires the Redfish service with a credential, and manage the free resource in the same driver. This also reduces the modificaiton on RedfishLib as it is an intermediary library to the opensource libredfish. @nicklela and @igorkulchytskyy, how do you think?

igorkulchytskyy:
First of all, I thought we replaced RedfishLib with RedfshHttpLib. And RedfishLib is deprecated.
So, we do need to use the EDK2_REDFISH_HTTP_PROTOCOL function to create a Redfish Service, and adding a function with credentials as parameters is a good idea.

PR #12133, opened 2026-02-05
RedfishPkg: fix enum type mismatch in payload.c

anushaKankipati:
JsonGetType() returns EDKII_JSON_TYPE but the local variable was declared as json_type, causing a CLANG-19 build error with -Werror.

# Description

JsonGetType() returns EDKII_JSON_TYPE but the local variable was declared as json_type, causing a CLANG-19 build error with -Werror.


changab:
Cast the returning value of JsonGetType to "json_type" doesn't work for CLANG-19? The reason I am asking this is that we don't want to update libredfish as it is from the opensource, although it is no longer maintained by Redfish WG.

PR #12137, opened 2026-02-06
ShellPkg/Debug1CommandsLib: Add human readable size to memmap summary

Damien-Chen:
# Description

Update the 'memmap' command to display the total size of each memory type in a human-readable format in the summary view. It provides a more readable information to developer that can help better understand memory usage of each memory type.

Changes include:
1. Modifying string tokens in UefiShellDebug1CommandsLib.uni to include a  placeholder for the formatted size string.
2. Implementing GetFormattedSizeString helper in MemMap.c to
    convert page   counts/bytes into readable units
3. Updating ParseMemoryDescriptors to calculate and print
    the formatted size strings for both standard and OEM/OS memory types.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Test memmap UEFI shell command in QEMU OvmfX64.dsc,

Before
```
  Reserved  :      3,215,521 Pages (13,170,774,016 Bytes)
  LoaderCode:            234 Pages (958,464 Bytes)
  LoaderData:              0 Pages (0 Bytes)
  BS_Code   :          1,136 Pages (4,653,056 Bytes)
  BS_Data   :          6,699 Pages (27,439,104 Bytes)
  RT_Code   :            256 Pages (1,048,576 Bytes)
  RT_Data   :            257 Pages (1,052,672 Bytes)
  ACPI_Recl :             18 Pages (73,728 Bytes)
  ACPI_NVS  :          4,876 Pages (19,972,096 Bytes)
  MMIO      :              0 Pages (0 Bytes)
  MMIO_Port :              0 Pages (0 Bytes)
  PalCode   :              0 Pages (0 Bytes)
  Unaccepted:              0 Pages (0 Bytes)
  Available :        506,459 Pages (2,074,456,064 Bytes)
  Persistent:              0 Pages (0 Bytes)
```

After:
```
  Reserved  :      3,215,521 Pages (13,170,774,016 Bytes) (12 GB)
  LoaderCode:            234 Pages (958,464 Bytes) (936 KB)
  LoaderData:              0 Pages (0 Bytes) (0 B)
  BS_Code   :          1,136 Pages (4,653,056 Bytes) (4 MB)
  BS_Data   :          6,699 Pages (27,439,104 Bytes) (26 MB)
  RT_Code   :            256 Pages (1,048,576 Bytes) (1 MB)
  RT_Data   :            257 Pages (1,052,672 Bytes) (1 MB)
  ACPI_Recl :             18 Pages (73,728 Bytes) (72 KB)
  ACPI_NVS  :          4,876 Pages (19,972,096 Bytes) (19 MB)
  MMIO      :              0 Pages (0 Bytes) (0 B)
  MMIO_Port :              0 Pages (0 Bytes) (0 B)
  PalCode   :              0 Pages (0 Bytes) (0 B)
  Unaccepted:              0 Pages (0 Bytes) (0 B)
  Available :        506,459 Pages (2,074,456,064 Bytes) (1 GB)
  Persistent:              0 Pages (0 Bytes) (0 B)
```

## Integration Instructions

N/A


vbpandya:
the CI is complaining about not having a sign off on the merge commit
You might want to get rid of that (merge) commit and rebase your patch or latest master.

Damien-Chen:
Yes, seems like the merge commit is not signed automatically. I will fix it.

ShellPkg/Library/UefiShellDebug1CommandsLib/MemMap.c
```diff
@@ -172,6 +195,8 @@ ParseMemoryDescriptors (
   LIST_ENTRY              MemoryList;
   LIST_ENTRY              *Link;
   MEMORY_LENGTH_ENTRY     *Entry;
+  CHAR16                  SizeStrings[EfiMaxMemoryType][20];
```
pierregondois: Small NIT:
Maybe create a define value instead of a hard coded value value

pierregondois (approved):
The code itself lgtm.
The other question is whether this is necessary or not, I m fine either way.

Damien-Chen:
> The code itself lgtm. The other question is whether this is necessary or not, I m fine either way.

For code change, I believe this doesn`t cause any side effect. 
This is such a simple enhance to make the information more readable for developer. The reason is that when I first use this command, I have to calculate the actual size of each type of memory. So I just wonder it would be better if it provides more readable information such like those memory tool in OS. That is the story why I create this PR enhancement.

ShellPkg/Library/UefiShellDebug1CommandsLib/MemMap.c
Damien-Chen: Thanks, I modify it so that it can be more maintainable.

benjamindoron:
The results are going to sometimes be misleading with integer maths. Consider this line:
>   Available :        506,459 Pages (2,074,456,064 Bytes) (1 GB)

We can eyeball that as about 2 GB, and a double take confirms that it is GB, not MB or TB. But it prints as 1 GB, because it's true, it's slightly less than 2 GB. personally, I don't think that this PR is that necessary (because performing floating point maths will be more costly in performance for platforms that must perform it in software, so it's not a simple fix)

Damien-Chen:
> The results are going to sometimes be misleading with integer maths. Consider this line:
> 
> > Available :        506,459 Pages (2,074,456,064 Bytes) (1 GB)
> 
> We can eyeball that as about 2 GB, and a double take confirms that it is GB, not MB or TB. But it prints as 1 GB, because it's true, it's slightly less than 2 GB. personally, I don't think that this PR is that necessary (because performing floating point maths will be more costly in performance for platforms that must perform it in software, so it's not a simple fix)

Yes, I agree "1 GB" for ~1.93 GB is misleading. I can fix this with integer-only math (no floating point needed), either by showing one decimal place or by rounding to nearest. The performance concern doesn't apply here since it's pure integer division and only runs once per summary line, not in a hot path.

PR #12143, opened 2026-02-07
BaseTools/WindowsVsToolChain.py: Minor code style updates

mikebeaton:
# Description

cc @mdkinney  - Assuming you agree with these? Apologies for not catching in the original PR. I was aiming to remove the path duplication (which I think is worth doing), so removed the extra space in the method call at the same time.

Fixes: 0b867b65a2c1b30a822e09572323495478778e81

- [ ] Breaking change
- [ ] Impacts security
- [ ] Includes tests

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py
+++ b/BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py
@@ -407,8 +407,8 @@ class WindowsVsToolChain(IUefiBuildPlugin):
                 vc_tools_install_dir = shell_env.get_shell_var("VCToolsInstallDir")
                 if vc_tools_install_dir:
                     ia32_dll_path = os.path.join(vc_tools_install_dir, 'bin', 'Hostx64', 'x86')
-                    if os.path.exists (ia32_dll_path):
-                        shell_env.append_path(os.path.join (vc_tools_install_dir, 'bin', 'Hostx64', 'x86'))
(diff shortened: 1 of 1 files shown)


mikebeaton:
@mdkinney - Hi - I can't remember why I was re-reading this PR in more detail, but I was and I spotted the duplication of the path `join` calculation. I don't believe there was any intentional reason for recalculating it, so this small PR suggests re-using the already calculated value - and also makes a whitespace change to what seems to be (more) standard for .py files round here at the same time. Does this look right?

Javagedes:
LGTM. Adding push tag. Sorry I missed this. There tends to be a lot of emails related to EDKII and things fall through the cracks.

mikebeaton:
@Javagedes - Thanks for your time, and apologies but this didn't push for whatever reason.

Javagedes:
@mikebeaton apologies, I should not have tried to resolve this from the GitHub app. I did a merge commit instead of a rebase & ff to get the PR up to date. Can you delete the merge commit and do a rebase & ff? I'll re apply the push tag after

mikebeaton:
> Can you delete the merge commit and do a rebase & ff? I'll re apply the push tag after

Okay, just done that. Thanks.

mikebeaton:
Appears to be stuck again - I'm not sure why? The only similar thing I've seen is when there are weird errors and activity from mergify, but there aren't any here.

Javagedes:
@mergifyio queue

PR #12145, opened 2026-02-08
EmulatorPkg: Add /NODEFAULTLIB:LIBCMT to remove warning

mdkinney:
# Description

Add /NODEFAULTLIB:LIBCMT to remove warning for Windows EmulatorPkg builds that use visual studio libraries

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verified warning no longer generated.

## Integration Instructions

N/A


os-d (approved):

mdkinney:
> A little more context on why this is being addressed now versus in the past would be helpful to capture in the commit message.

The warning was missed in the previous updates.  In general there are some compiler/linker warnings that are not elevated to an error status, so the warning message will appear in the log but the check will pass. The warning message was noticed after the recent updates to EmulatorPkg.

PR #12147, opened 2026-02-08
BaseTools/Source/Python/Trim: Add -f/--source-code-format option

mdkinney:
# Description

Add --source-code-format option that can be one of C, NASM, VFR, ASL, ASM. This can be used for file format specific actions when --source-code is used.

A NASM specific action is added to convert #line to %line to preserve reference the originating NASM source file for source level debug in NASM format.

Without this change, the source level debug if NASM files loads the generated intermediate file in the build output directory.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verified that source level debug of NASM files opens NASM source file from source directory and not the .iii files from the build output directory.

## Integration Instructions

N/A

### DIFF
--- a/ShellPkg/Library/UefiShellDebug1CommandsLib/SmbiosView/PrintInfo.c
+++ b/ShellPkg/Library/UefiShellDebug1CommandsLib/SmbiosView/PrintInfo.c
@@ -1282,6 +1282,11 @@ SmbiosPrintStructure (
     // Management Controller Host Interface (Type 42)
     //
     case 42:
+      MC_HOST_INTERFACE_PROTOCOL_RECORD  *MCHostInterfaceProtocolRecord;
+      UINT8                              *RecordsPointer;
+      UINT8                              MCHostInterfaceProtocolNumber;
+      UINT8                              ProtocolTypeDataLen;
+
       DisplayMCHostInterfaceType (Struct->Type42->InterfaceType, Option);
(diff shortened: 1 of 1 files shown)


mdkinney:
@gapalomi and @nate-desimone please review.  This improves source level debug of NASM source files.

BaseTools/Source/Python/Trim/Trim.py
```diff
@@ -529,6 +535,8 @@ def Options():
     OptionList = [
         make_option("-s", "--source-code", dest="FileType", const="SourceCode", action="store_const",
                           help="The input file is preprocessed source code, including C or assembly code"),
+        make_option("-f", "--source-code-format", dest="FileFormat", type="choice", choices=["C", "NASM", "VFR", "ASL", "ASM"], default="C",
```
lgao4: I don't see Trim usage for C source file in build_rule.txt. So, I don't suggest not to add C type file.

mdkinney: Those keywords are placeholders.  Only NASM perform performs a different action in Trim.  

Are you suggesting I remove all except NASM and have the default be ""?

lgao4: Yes. Other type can be added later when it is requested.

mdkinney: I have pushed an update to reduce options to just NASM

PR #12148, opened 2026-02-09
OvmfPkg/OvmfXen: Switch timer library to support x2APIC mode

d3athjest3r:
# Description

The SecPeiDxeTimerLibCpu library from the MdePkg accesses the Local APIC timer exclusively via MMIO and explicitly asserts that x2APIC mode is not enabled. This causes failures when vCPUs are pre-enabled in x2APIC mode, which might be necessary for domains with more than 128 vCPUs. So, this commit switches the TimerLib to SecPeiDxeTimerLibUefiCpu from the UefiCpuPkg, which auto-detects the APIC mode and transparently uses MSR-based access for x2APIC or MMIO for xAPIC.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Booted a Xen guest with more than 128 vCPUs and forced all vCPUs to X2APIC mode. Currently it fails with an assert in the  timer library. With the new timer library it passes because this one also supports X2APIC. While the old one doesn't.

## Integration Instructions
N/A

### DIFF
--- a/OvmfPkg/OvmfXen.dsc
+++ b/OvmfPkg/OvmfXen.dsc
@@ -124,7 +124,7 @@
 
 [LibraryClasses]
   PcdLib|MdePkg/Library/BasePcdLibNull/BasePcdLibNull.inf
-  TimerLib|MdePkg/Library/SecPeiDxeTimerLibCpu/SecPeiDxeTimerLibCpu.inf
+  TimerLib|UefiCpuPkg/Library/SecPeiDxeTimerLibUefiCpu/SecPeiDxeTimerLibUefiCpu.inf
   ResetSystemLib|OvmfPkg/Library/ResetSystemLib/BaseResetSystemLibXen.inf
   PrintLib|MdePkg/Library/BasePrintLib/BasePrintLib.inf
   BaseMemoryLib|MdePkg/Library/BaseMemoryLibRepStr/BaseMemoryLibRepStr.inf


TSnake41:
Fixes OVMF hang when booting with [[RFC PATCH v2] x86/hvm: Allow pre-enabling x2apic mode on BSP](https://lore.kernel.org/xen-devel/0cb4d1f91212a65baf924ed0ef825d8adb4b5423.1762958551.git.teddy.astie@vates.tech) (using `x2apic_mode = "pre_enable"`)

```
Tested-by: Teddy Astie <teddy.astie@vates.tech>
```

d3athjest3r:
So I have verified that it works with a PVH VM, with the following parameters:

```
type='pvh'
kernel='/usr/share/edk2/OVMF-debug.fd'

memory = 3072
vcpus = 4
sdl = 0
vnc = 0
serial='pty'
name = "ovmf"
disk = [
## This line works in OVMF, but Linux will not look at CDROM over PV interface
  '/mnt/debian-live-12.4.0-amd64-standard.iso,,hdc,cdrom',
## This works for both Linux and OVMF, and archlinux ISO can boot.
  '/mnt/debian-live-12.4.0-amd64-standard.iso,,hdd,r',
]
on_reboot = 'destroy'
```
and `xl create -c foo.pvh`. I see OVMF booting properly. Although I don't see my Linux booting afterwards. I think there is an issue with disk command?  It then jumps into the EDKII menu and BIOS settings. It doesn't fail with an error with the timer driver or so. So, I think we should be good for PVH as well.

PR #12149, opened 2026-02-09
Use pragma once in place of include guards

makubacki:
# Description

### Summary:

- Allows `#pragma once` to work with VfrCompiler (simple change to prevent unrecognized token)
- Removes header guards checks in the ECC tool
- Updates header files to use `#pragma once` instead of traditional include guards

Background: A concern was raised to edk2 stewards that the [EDK II C Coding Standards Specification - 5.3 Include Files](https://tianocore-docs.github.io/edk2-CCodingStandardsSpecification/release-2.20/5_source_files/53_include_files.html#53-include-files) specifies requirements on include guards which are not met throughout the codebase and make it difficult to enforce the coding standard in downstream code. For example, this requirement is inconsistently applied:

> Names starting with one or two underscores, such as _MACRO_GUARD_FILE_NAME_H_, must not be used.
> They are reserved for compiler implementation.

Upon review, it was determined instead of replacing all macro names, the project would adopt `#pragma once`. A similar PR will be submitted to the [edk2-CCodingStandardsSpecification](https://github.com/tianocore-docs/edk2-CCodingStandardsSpecification) repo to update the specification.

---

#### BaseTools/VfrCompile: Add #pragma once support

The C preprocessor turns each .vfr file into a pre-processed .i
file. At this step, the C preprocessor processes `#pragma once`.
Then, VfrCompile is called (with `-n` to prevent preprocessing)
to parse the pre-processed .i files.

The .i files may still contain `#pragma once` lines. Currently,
VfrCompile treats `once` as an unknown token, causing parse failures.

Originally, this change was going to add a `PragmaOnce` token rule
to the VFR lexer grammar (in VfrSyntax.g) that matched `#pragma once`
lines and silently skipped them using `skip()` and `newline()`. The
`newline()` call would keep line numbers stable for error reporting.
This was consistent with how other preprocessor artifacts were already
handled like `#line` directives (`LineDefinition` and
`GccLineDefinition` tokens) and `extern` declarations (skipped with
`mode(CPP_COMMENT)`).

Writing a regular expression to match `#pragma once` was simple
enough, but it makes overall pragma token recognition more fragile
at the lexer level. When the lexer is walking the DFA state table,
it could begin to match a `#pragma ` line but then not be able to
match remaining characters to recognize tokens other than `once`.

Instead, this change handles `#pragma once` lines in the VFR parser
grammar in `vfrPragmaDefinition` alongside where `pack` is already
handled.

---

#### BaseTools/Ecc: Remove #ifndef include guard checks

The codebase has moved from traditional `#ifndef` include guards to
`#pragma once`. Remove the ECC checks that validated include guard
presence and naming conventions since they are no longer applicable.

The following checks are removed:

- `IncludeFileCheckIfndefStatement`: Verified all header file contents
  were guarded by a `#ifndef` statement, that the `#ifndef` was the
  first line of code after the file header comment, and that the
  `#endif` appeared on the last line.

- `NamingConventionCheckIfndefStatement`: Verified that the `#ifndef`
  guard name at the start of an include file used a postfix underscore
  and no prefix underscore character.

Also removed related error codes and configuration settings that were
specific to these checks.

---

#### Replace include guards with #pragma once

Replace traditional `#ifndef`/`#define`/`#endif` include guards with `#pragma` once.

`#pragma once` is a widely supported preprocessor directive that prevents header files from being included multiple times. It is supported by all toolchains used to build edk2: GCC, Clang/LLVM, and MSVC.

Compared to macro-based include guards, `#pragma once`:

- Eliminates the risk of macro name collisions or copy/paste errors
  where two headers inadvertently use the same guard macro.
- Eliminate inconsistency in the way include guard macros are named
  (e.g., some files use `__FILE_H__`, others use `FILE_H_`, etc.).
- Reduces boilerplate (three lines replaced by one).
- Avoids polluting the macro namespace with guard symbols.
- Can improve build times as the preprocessor can skip re-opening the
  file entirely, rather than re-reading it to find the matching
  `#endif` ("multiple-include optimization").
  - Note that some compilers may already optimize traditional include
    guards, by recognzining the idiomatic pattern.

This change is made acknowledging that overall portability of the code will technically be reduced, as `#pragma once` is not part of the C/C++ standards.

However, this is considered acceptable given:

1. edk2 already defines a subset of supported compilers in
   `BaseTools/Conf/tools_def.template`, all of which have supported
   `#pragma once` for over two decades.
2. There have been concerns raised to the project about inconsistent
   include guard naming and potential macro collisions.

Approximate compiler support dates:

- MSVC: Supported since Visual C++ 4.2 (1996)
  - Docs: https://learn.microsoft.com/cpp/preprocessor/once
- GCC: Supported since 3.4 (2004)
  (http://gnu.ist.utl.pt/software/gcc/gcc-3.4/changes.html)
  - Docs: https://gcc.gnu.org/onlinedocs/cpp/Pragmas.html#index-_0023pragma-once
- Clang (LLVM based): Since initial release in 2007

---

#### Notes

- **BaseTools**: Some files in BaseTools are excluded from the change if they are autogenerated or directly related to a header from a subproject, etc. In particular, headers in these directories were ignored:
  - `BaseTools/Source/C/LzmaCompress/Sdk/`
  - `BaseTools/Source/C/VfrCompile/Pccts/`
- **CryptoPkg**: Does not include updates to OpenSSL generated header files checked
into the repo. Those in `CryptoPkg/Library/OpensslLib/OpensslGen/`.
- **OvmfPkg**: Headers taken directly from external projects, such as those in `OvmfPkg/Include/IndustryStandard/Xen/` were not modified since they may be periodically re-synced and do not follow other edk2 coding standards.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- edk2 BaseTools build
- edk2 CI

## Integration Instructions

- No special integration work is expected. A supported compiler should already be used in the platform build.

### DIFF
--- a/EmulatorPkg/PlatformCI/.azurepipelines/Windows-VS.yml
+++ b/EmulatorPkg/PlatformCI/.azurepipelines/Windows-VS.yml
@@ -59,27 +59,6 @@ jobs:
             Build.Target: "NOOPT"
             Run.Flags: $(run_flags)
             Run: $(should_run)
-          EmulatorPkg_IA32_DEBUG:
-            Build.File: "$(package)/PlatformCI/PlatformBuild.py"
-            Build.Arch: "IA32 "
-            Build.Flags: "BLD_*_NO_PLATFORM_BOOT_DELAYS"
-            Build.Target: "DEBUG"
-            Run.Flags: $(run_flags)
-            Run: $(should_run)
-          EmulatorPkg_IA32_RELEASE:
-            Build.File: "$(package)/PlatformCI/PlatformBuild.py"
-            Build.Arch: "IA32 "
-            Build.Flags: "BLD_*_NO_PLATFORM_BOOT_DELAYS"
-            Build.Target: "RELEASE"
-            Run.Flags: $(run_flags)
-            Run: $(should_run)
-          EmulatorPkg_IA32_NOOPT:
-            Build.File: "$(package)/PlatformCI/PlatformBuild.py"
-            Build.Arch: "IA32 "
-            Build.Flags: "BLD_*_NO_PLATFORM_BOOT_DELAYS"
-            Build.Target: "NOOPT"
-            Run.Flags: $(run_flags)
-            Run: $(should_run)
           EmulatorPkg_X64_FULL_DEBUG:
             Build.File: "$(package)/PlatformCI/PlatformBuild.py"
             Build.Arch: "X64"
@@ -97,27 +76,6 @@ jobs:
           EmulatorPkg_X64_FULL_NOOPT:
             Build.File: "$(package)/PlatformCI/PlatformBuild.py"
             Build.Arch: "X64"
-            Build.Flags: "BLD_*_SECURE_BOOT_ENABLE=TRUE BLD_*_NO_PLATFORM_BOOT_DELAYS"
-            Build.Target: "NOOPT"
-            Run.Flags: $(run_flags)
-            Run: $(should_run)
-          EmulatorPkg_IA32_FULL_DEBUG:
-            Build.File: "$(package)/PlatformCI/PlatformBuild.py"
-            Build.Arch: "IA32"
-            Build.Flags: "BLD_*_SECURE_BOOT_ENABLE=TRUE BLD_*_NO_PLATFORM_BOOT_DELAYS"
-            Build.Target: "DEBUG"
-            Run.Flags: $(run_flags)
-            Run: $(should_run)
-          EmulatorPkg_IA32_FULL_RELEASE:
-            Build.File: "$(package)/PlatformCI/PlatformBuild.py"
-            Build.Arch: "IA32"
-            Build.Flags: "BLD_*_SECURE_BOOT_ENABLE=TRUE BLD_*_NO_PLATFORM_BOOT_DELAYS"
-            Build.Target: "RELEASE"
-            Run.Flags: $(run_flags)
-            Run: $(should_run)
-          EmulatorPkg_IA32_FULL_NOOPT:
-            Build.File: "$(package)/PlatformCI/PlatformBuild.py"
-            Build.Arch: "IA32"
             Build.Flags: "BLD_*_SECURE_BOOT_ENABLE=TRUE BLD_*_NO_PLATFORM_BOOT_DELAYS"
             Build.Target: "NOOPT"
             Run.Flags: $(run_flags)


os-d (approved):
Approved for EmulatorPkg and FatPkg.

OvmfPkg/Bhyve/BhyveRfbDxe/VbeShim.h
```diff
@@ -1,912 +1,912 @@
 //
 // THIS FILE WAS GENERATED BY "VbeShim.sh". DO NOT EDIT.
 //
```
mikebeaton, line 3: Hi - There are some line endings issue in this file (see below). Also, if the file is still generated as stated above, that should be modified too.

@bexcran can hopefully comment on whether it is still expected that VbeShim.h will be regenerated using VbeShim.sh.

(Running VbeShim.sh in GfW Bash, I found that I had to make this change before it would run:
```
-nasm -o "$STEM".bin -- "$STEM".asm
+nasm -o "$STEM".bin "$STEM".asm
```
then the generated file was similar but not the same (sample of multiple similar changes):
```
-  /* 00000448 imul si,[si+0x69],word 0x7365  */  0x69, 0x74, 0x69, 0x65, 0x73,
+  /* 00000448 imul si,[si+0x69],0x7365       */  0x69, 0x74, 0x69, 0x65, 0x73,^M
   /* 0000044D or ax,0xa                      */  0x0D, 0x0A, 0x00,
   /* 00000450 jna 0x4a4                      */  0x76, 0x52,
-  /* 00000452 gs popaw                       */  0x65, 0x61,
+  /* 00000452 gs popa                        */  0x65, 0x61,^M
   /* 00000454 fs inc bp                      */  0x64, 0x45,
-  /* 00000456 imul sp,[fs:si+0xd],word 0xa   */  0x64, 0x69, 0x64, 0x0D, 0x0A, 0x00,
+  /* 00000456 imul sp,[fs:si+0xd],0xa        */  0x64, 0x69, 0x64, 0x0D, 0x0A, 0x00,^M
   /* 0000045C jna 0x4aa                      */  0x76, 0x4C,
-  /* 0000045E gs a32 popaw                   */  0x65, 0x67, 0x61,
+  /* 0000045E gs a32 popa                    */  0x65, 0x67, 0x61,^M
```
)

makubacki: I pushed an update to drop the file update. That was mostly due to uncrustify formatting (which was reverted in the past).

BaseTools/Source/C/VfrCompile/VfrCompiler.cpp
```diff
@@ -9,6 +9,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 
 #include "stdio.h"
 #include "stdlib.h"
+#include "ctype.h"
```
leiflindholm: Could this be inserted on line 10 so as to not reduce level of sortedness?

leiflindholm: Could this be inserted on line 10 so as to not reduce level of sortedness of includes?

makubacki: Pushed

Li-Aaron (approved):
Approve for PrmPkg.

niruiyu (approved):
You are the king of "include"!

BaseTools/Source/C/VfrCompile/VfrCompiler.cpp
```diff
@@ -18,6 +19,116 @@ PACKAGE_DATA  gCBuffer;
 PACKAGE_DATA  gRBuffer;
 CVfrStringDB  gCVfrStringDB;
 
+STATIC
+CONST CHAR8 *
+SkipWhitespace (
```
vishalo: Can the lexer handle this, instead of CPP code? BaseTools/Source/C/VfrCompile/VfrSyntax.g

AshrafAliS (approved):

pierregondois:
Just a side comment: there is a check in the Ecc tool regarding the naming of the include guards. Maybe it would be worth removing it aswell. Cf.
```
ERROR_NAMING_CONVENTION_CHECK_IFNDEF_STATEMENT
NamingConventionCheckIfndefStatement()
```

makubacki:
> Just a side comment: there is a check in the Ecc tool regarding the naming of the include guards. Maybe it would be worth removing it aswell. Cf.
> 
> ```
> ERROR_NAMING_CONVENTION_CHECK_IFNDEF_STATEMENT
> NamingConventionCheckIfndefStatement()
> ```

Thanks for pointing it out. I will remove it.

makubacki:
> > Just a side comment: there is a check in the Ecc tool regarding the naming of the include guards. Maybe it would be worth removing it aswell. Cf.
> > ```
> > ERROR_NAMING_CONVENTION_CHECK_IFNDEF_STATEMENT
> > NamingConventionCheckIfndefStatement()
> > ```
> 
> Thanks for pointing it out. I will remove it.

Added in commit https://github.com/tianocore/edk2/pull/12149/changes/36716d0ebc5889d78ea9421823779fc6c20d7565

BaseTools/Source/C/VfrCompile/VfrCompiler.cpp
makubacki: I made the update in https://github.com/tianocore/edk2/pull/12149/changes/0cff617294494f7ef1ad10024ff3cf23435bbadd.

samimujawar (approved):
Approved for ArmPkg, ArmPlatformPkg, ArmVirtPkg, DynamicTablesPkg, StandaloneMmPkg.

makubacki:
This PR has a high likelihood of merge conflicts due to the number of files modified.

I would like to get an approval from a BaseTools maintainer/reviewer before merging. @mikebeaton, you've reviewed it in the past.

Other BaseTools maintainers/reviewers - @lgao4, @gapalomi, @YuweiChen1110, @ponchofigueroa

gapalomi:
I see the PR it's removing the `IncludeFileCheckIfndefStatement` and `NamingConventionCheckIfndefStatement` checks but not replacing it with broader test that includes `#pragma once`

should we extend the test instead in a following PR?

makubacki:
> I see the PR it's removing the `IncludeFileCheckIfndefStatement` and `NamingConventionCheckIfndefStatement` checks but not replacing it with broader test that includes `#pragma once`
> 
> should we extend the test instead in a following PR?

@gapalomi, I think that's reasonable but requires additional assessment to determine how to handle external files copied into the repository that retain their original include guards. Since you’re comfortable addressing this in a follow-up PR, I’ve filed an issue to track it - https://github.com/tianocore/edk2/issues/12187.

PR #12150, opened 2026-02-09
UefiCpuPkg, MdePkg: Make X64 Page Table Definitions Public

os-d:
# Description

Currently, X64 page table structure definitions live in a private header in UefiCpuPkg's CpuPageTableLib. This moves them to a new CpuPageTable.h header file in MdePkg/Register/Intel, so that other code can use these standard architectural definitions.

The main use case is for test code to consume these definitions to run memory protection tests.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Running memory protection tests using these definitions.

## Integration Instructions

N/A.


os-d:
@mdkinney @lgao4 I could also see this being moved into MdePkg because these are architectural definitions not part of a library class.

AArch64 has a `MdePkg/Include/AArch64/AArch64.h` that has similar definitions: https://github.com/tianocore/edk2/blob/f0542ae07d5e5e4d311bf8ae33bf26b4b1acf9f4/MdePkg/Include/AArch64/AArch64.h and also includes AArch64Mmu.h in the same file, also with MMU related definitions. Do you think it would be appropriate to have a `MdePkg/Include/X64/X64.h` or something equivalent and have X64 architectural definitions there?

mdkinney:
> @mdkinney @lgao4 I could also see this being moved into MdePkg because these are architectural definitions not part of a library class.
> 
> AArch64 has a `MdePkg/Include/AArch64/AArch64.h` that has similar definitions: https://github.com/tianocore/edk2/blob/f0542ae07d5e5e4d311bf8ae33bf26b4b1acf9f4/MdePkg/Include/AArch64/AArch64.h and also includes AArch64Mmu.h in the same file, also with MMU related definitions. Do you think it would be appropriate to have a `MdePkg/Include/X64/X64.h` or something equivalent and have X64 architectural definitions there?

There is also MdePkg/Include/Register area for CPU specific content

os-d:
> There is also MdePkg/Include/Register area for CPU specific content

Ah, thanks. I will add a new CpuPageTable.h there with this content, I agree that looks to be an appropriate place.

os-d:
@mdkinney I moved these to `MdePkg/Register`, please re-review when you have a chance (obviously this will wait until after the stable tag).

mdkinney:
There are also rules for UefiCpuPkg for content that is common to all x86 CPUs.  Not sure if we should add a new subdir under Register to distinguish vendor specific content from content that is common to all vendors.

@niruiyu Can you provide feedback on best location for this content?

os-d:
> There are also rules for UefiCpuPkg for content that is common to all x86 CPUs. Not sure if we should add a new subdir under Register to distinguish vendor specific content from content that is common to all vendors.
> 
> @niruiyu Can you provide feedback on best location for this content?

Yeah, that's why I was originally wondering if we wanted to create the `MdePkg/Include/X64/X64.h`, like the AArch64 has. I don't have a preference on location, will wait for Ray's feedback.

os-d:
@niruiyu do you have feedback on the best location for this content?

mdkinney:
@os-d The goal of the PageTableLib in the UefiCpuPkg is to hide these types of details from components outside the UefiCpuPkg. What information is needed from the page table details to implement memory protection tests?  Could those tests be part of the UefiCpuPkg, or could the PageTableLib be extended to return the information needed?

os-d:
> @os-d The goal of the PageTableLib in the UefiCpuPkg is to hide these types of details from components outside the UefiCpuPkg. What information is needed from the page table details to implement memory protection tests? Could those tests be part of the UefiCpuPkg, or could the PageTableLib be extended to return the information needed?

@mdkinney we had talked through this on our side, originally I had been in the same camp that the details should be hidden from consumers of CpuPageTableLib to protect against improper use of the APIs. It was pointed out to me that these are all standard architectural definitions and that have the definitions available is a separate question than API contract and enforcement. That argument made sense to me, that we do expose other architectural definitions widely (and as you pointed out, MdePkg may better home for these than UefiCpuPkg because that's where most of the standard architectural definitions live).

As for the specific use case, I am updating a test application that is intended to be able to verify a lot of different environments memory safety. It directly included the page table definitions prior to CpuPageTableLib and then did switch over to it once it was created. I just had to remove CpuPageTableLib from it to be able to understand self-mapped page tables, which Patina uses. I don't think that understanding should be added to CpuPageTableLib unless edk2 decided to move to self-mapped page tables, which is not currently on the roadmap.

The general use case exists of accessing these structures, we've had cases where the test application discovered bugs in CpuPageTableLib because the parsing in the lib was incorrect and an independent parsing discovered the issue.

Let me know what you think, the main impetus is to have architectural definitions live in a standard accessible place. If that is not desired, I can close this and the test app can host duplicate copies of the architectural definitions.

mdkinney:
@osd Thanks for the details.  I agree the MdePkg is the package that contains include files for industry standard and public specs. PageTableLib is just one use case of using the CPU Page Table definitions and there could be many more use cases.  Adding this content to the MdePkg does make sense.

MdePkg/Include/Register/Intel is one candidate location.

If these definitions are common to all x86 vendors, then MdePkg/Include/Register/X86 may be a better location.

@niruiyu @AbnerChang @pbgrimes Please review and help select the correct path location for page table definitions in MdePkg.

os-d:
> @osd Thanks for the details. I agree the MdePkg is the package that contains include files for industry standard and public specs. PageTableLib is just one use case of using the CPU Page Table definitions and there could be many more use cases. Adding this content to the MdePkg does make sense.
> 
> MdePkg/Include/Register/Intel is one candidate location.
> 
> If these definitions are common to all x86 vendors, then MdePkg/Include/Register/X86 may be a better location.
> 
> @niruiyu @AbnerChang @pbgrimes Please review and help select the correct path location for page table definitions in MdePkg.

Yeah, I think Register/X86 may be a better spot, will wait for additional feedback.

niruiyu:
> > @osd Thanks for the details. I agree the MdePkg is the package that contains include files for industry standard and public specs. PageTableLib is just one use case of using the CPU Page Table definitions and there could be many more use cases. Adding this content to the MdePkg does make sense.
> > MdePkg/Include/Register/Intel is one candidate location.
> > If these definitions are common to all x86 vendors, then MdePkg/Include/Register/X86 may be a better location.
> > @niruiyu @AbnerChang @pbgrimes Please review and help select the correct path location for page table definitions in MdePkg.
> 
> Yeah, I think Register/X86 may be a better spot, will wait for additional feedback.

When I commit the CpuPageTableLib to edk2, my rule was to not expose the API (including the structure definitions) until it's necessary. Now, because more usages that require these definitions appear, and the structures moved to MdePkg are defined by Intel SDM, I am ok with the change. The structures invented only for the CpuPageTableLib APIs are not moved (e.g.: IA32_MAP_ATTRIBUTE, PAGING_MODE), which is good.

I am ok with the location MdePkg/Include/Register/X86. Future common definitions can be moved to here and Intel/Amd headers can refer to the common headers if needed.

MdePkg/Include/Register/Intel/CpuPageTable.h
```diff
@@ -0,0 +1,178 @@
...
+
+**/
+
+#pragma once
+
+#define IA32_PE_BASE_ADDRESS_MASK_40  0xFFFFFFFFFF000ull
+#define IA32_PE_BASE_ADDRESS_MASK_39  0xFFFFFFFFFE000ull
+
+#define MAX_PAE_PDPTE_NUM  4
+
```
niruiyu: Wait, what is this for?

os-d: Thanks, I missed that. That is for something internal to CpuPageTableLib. I will leave that in the internal header file.

os-d:
Thanks for reviewing, Ray, I have moved the definitions to MdePkg\Include\Register\X86. @mdkinney can you please re-review when you have a chance?

mdkinney:
This should be broken up into 2 commits.  MdePkg to add the new .h file, then UefiCpuPkg to use the new include file in MdePkg and remove definitions from UefiCpuPkg.  This avoids the need for the special tag in the commit message to allow modifications to multiple packages.

os-d:
> This should be broken up into 2 commits. MdePkg to add the new .h file, then UefiCpuPkg to use the new include file in MdePkg and remove definitions from UefiCpuPkg. This avoids the need for the special tag in the commit message to allow modifications to multiple packages.

@mdkinney agreed, fixed. When I was adding it to an existing header file that was already included in CpuPageTableLib.h, it needed to be one commit to prevent duplicate definitions. Now that this is a new header file it can be two commits.

MdePkg/Include/Register/Intel/CpuPageTable.h
niruiyu: Thanks!

os-d:
@mdkinney can you re-review when you have a chance?

os-d:
@mdkinney ping on this. I think we are all agreed and can add the push label?

PR #12153, opened 2026-02-10
ArmPlatformPkg/PL011UartLib: Fix incorrect control bit validation

KavipriyanKP:
# Description

REF: [e092122]
(PR #12085)

The current logic in ArmPlatformPkg/PL011UartLib performs control-bit checks intended to align with the UEFI specification. However, it incorrectly validates control bits that cannot be set through SetControl(), resulting in failures in the SetControl() SCT test.

According to the **UEFI specification 2.11, under section:12.8.3.3**
the valid control bits that can be set are:

EFI_SERIAL_REQUEST_TO_SEND
EFI_SERIAL_DATA_TERMINAL_READY
EFI_SERIAL_HARDWARE_LOOPBACK_ENABLE
EFI_SERIAL_SOFTWARE_LOOPBACK_ENABLE
EFI_SERIAL_HARDWARE_FLOW_CONTROL_ENABLE

This patch updates mAllowedControlBits so that only these valid control bits are accepted during validation.

**Important**:
This patch fixes an issue present in the current release candidate and must be merged **before** the **edk2-stable202602** release to ensure correctness and prevent regressions.

Cc: Sachin Ganesh <sachinganesh@ami.com>
Cc: @mdkinney 
Cc: @ardbiesheuvel 
Cc: @samimujawar 
Cc: @leiflindholm 
Cc: @lgao4 

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verified EDK2 SCT test SerialIoBBTestConformance on GB200 platform, 

SCT tool version: V24.11.3_0_0_BETA0

Test results:
1) With Fix: 
[SCT_ SetControl_Conf_with_Patch.log](https://github.com/user-attachments/files/25212421/SCT_.SetControl_Conf_with_Patch.log)

2) Without Fix: 
[SCT_ SetControl_Conf_without_Patch.log](https://github.com/user-attachments/files/25212426/SCT_.SetControl_Conf_without_Patch.log)

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/BinWrappers/PosixLike/BrotliCompress
+++ b/BaseTools/BinWrappers/PosixLike/BrotliCompress
@@ -36,13 +36,7 @@ do
 done
 
 
-if [ -n "$WORKSPACE" ] && [ -e "$WORKSPACE/BaseTools/Build/Source/C/bin" ]
-then
-  if [ -e "$WORKSPACE/BaseTools/Build/Source/C/bin/$cmd" ]
-  then
-    exec "$WORKSPACE/BaseTools/Build/Source/C/bin/$cmd" $QLT $ARGS
-  fi
-eif [ -n "$EDK_TOOLS_PATH" ] && [ -e "$EDK_TOOLS_PATH/Source/C/bin" ]
+if [ -n "$EDK_TOOLS_PATH" ] && [ -e "$EDK_TOOLS_PATH/Source/C/bin" ]
 then
   if [ -e "$EDK_TOOLS_PATH/Source/C/bin/$cmd" ]
   then
--- a/BaseTools/BinWrappers/PosixLike/GenericShellWrapper
+++ b/BaseTools/BinWrappers/PosixLike/GenericShellWrapper
@@ -4,13 +4,7 @@ full_cmd=${BASH_SOURCE[-1]:-$0} # see http://mywiki.wooledge.org/BashFAQ/028 for
 dir=$(dirname "$full_cmd")
 cmd=${full_cmd##*/}
 
-if [ -n "$WORKSPACE" ] && [ -e "$WORKSPACE/BaseTools/Build/Source/C/bin" ]
-then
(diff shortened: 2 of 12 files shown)


samimujawar (approved):
The relevant section in the UEFI spec can be seen at 

https://uefi.org/specs/UEFI/2.9_A/12_Protocols_Console_Support.html#:~:text=Only%20the%20REQUEST_TO_SEND%2C%20DATA_TERMINAL_READY%2C%20HARDWARE_LOOPBACK_ENABLE%2C%20SOFTWARE_LOOPBACK_ENABLE%2C%20and%20HARDWARE_FLOW_CONTROL_ENABLE%20bits%20can%20be%20set%20with%20SetControl().

ardbiesheuvel:
Can we just revert the offending change, and worry about this after the release?

leiflindholm:
> Can we just revert the offending change, and worry about this after the release?

It is literally a logical inversion compared to how the mask variable is named. So I don't see this fix as being "more risky" than a revert.

It's a bit of a weird protocol interface presumably based on an actual hw register in uart8250. Half the bits are writable, half are not. So the spec explicitly calls out that only the writable bits should be settable. *shrug*

ardbiesheuvel:
> > Can we just revert the offending change, and worry about this after the release?
> 
> It is literally a logical inversion compared to how the mask variable is named. So I don't see this fix as being "more risky" than a revert.
> 
> It's a bit of a weird protocol interface presumably based on an actual hw register in uart8250. Half the bits are writable, half are not. So the spec explicitly calls out that only the writable bits should be settable. _shrug_

But the original change fixed an SCT error rather than an actual problem, right? So how did that break SCT?

leiflindholm:
> But the original change fixed an SCT error rather than an actual problem, right? So how did that break SCT?

SCT checks that the function returns an error if any of the non-writable bits are set.

PR #12157, opened 2026-02-10
BaseTools: Add EDK_TOOLS_BUILD_PATH for tool output

mdkinney:
# Description

Update BaseTools to opt-in to the out-of-tree builds of tools. If the EDK_TOOLS_BUILD_PATH environment
variable is not defined, then binary artifacts of the BaseTools C tools builds are generated in the BaseTools directory. If EDK_TOOLS_BUILD_PATH is defined, then the binary artifacts BaseTools C tools builds are
generated in path defined by the EDK_TOOLS_BUILD_PATH environment variable.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verify build without EDK_TOOLS_BUILD_PATH set generates tools in BaseTools and build with
EDK_TOOLS_BUILD_PATH set generate tools in that path.

## Integration Instructions

No change to preserve current BaseTools output behavior.

Set EDK_TOOLS_BUILD_PATH to specify an out-of-tree output path for BaseTools.

### DIFF
--- a/BaseTools/BinWrappers/PosixLike/BrotliCompress
+++ b/BaseTools/BinWrappers/PosixLike/BrotliCompress
@@ -36,13 +36,13 @@ do
 done
 
 
-if [ -n "$WORKSPACE" ] && [ -e "$WORKSPACE/BaseTools/Build/Source/C/bin" ]
+if [ -n "$EDK_TOOLS_BUILD_PATH" ] && [ -e "$EDK_TOOLS_BUILD_PATH/Source/C/bin" ]
 then
-  if [ -e "$WORKSPACE/BaseTools/Build/Source/C/bin/$cmd" ]
+  if [ -e "$EDK_TOOLS_BUILD_PATH/Source/C/bin/$cmd" ]
   then
-    exec "$WORKSPACE/BaseTools/Build/Source/C/bin/$cmd" $QLT $ARGS
+    exec "$EDK_TOOLS_BUILD_PATH/Source/C/bin/$cmd" $QLT $ARGS
   fi
-eif [ -n "$EDK_TOOLS_PATH" ] && [ -e "$EDK_TOOLS_PATH/Source/C/bin" ]
+elif [ -n "$EDK_TOOLS_PATH" ] && [ -e "$EDK_TOOLS_PATH/Source/C/bin" ]
 then
   if [ -e "$EDK_TOOLS_PATH/Source/C/bin/$cmd" ]
   then
--- a/BaseTools/BinWrappers/PosixLike/GenericShellWrapper
+++ b/BaseTools/BinWrappers/PosixLike/GenericShellWrapper
@@ -4,11 +4,11 @@ full_cmd=${BASH_SOURCE[-1]:-$0} # see http://mywiki.wooledge.org/BashFAQ/028 for
 dir=$(dirname "$full_cmd")
 cmd=${full_cmd##*/}
 
-if [ -n "$WORKSPACE" ] && [ -e "$WORKSPACE/BaseTools/Build/Source/C/bin" ]
+if [ -n "$EDK_TOOLS_BUILD_PATH" ] && [ -e "$EDK_TOOLS_BUILD_PATH/Source/C/bin" ]
 then
-  if [ -e "$WORKSPACE/BaseTools/Build/Source/C/bin/$cmd" ]
+  if [ -e "$EDK_TOOLS_BUILD_PATH/Source/C/bin/$cmd" ]
   then
-    exec "$WORKSPACE/BaseTools/Build/Source/C/bin/$cmd" "$@"
+    exec "$EDK_TOOLS_BUILD_PATH/Source/C/bin/$cmd" "$@"
   fi
 elif [ -n "$EDK_TOOLS_PATH" ] && [ -e "$EDK_TOOLS_PATH/Source/C/bin" ]
 then
--- a/BaseTools/Source/C/GNUmakefile
+++ b/BaseTools/Source/C/GNUmakefile
@@ -13,8 +13,8 @@ else
   MAKEROOT := .
 endif
 
-ifneq ($(WORKSPACE),)
-  BUILDROOT=$(WORKSPACE)/BaseTools/Build/Source/C
+ifneq ($(EDK_TOOLS_BUILD_PATH),)
+  BUILDROOT=$(EDK_TOOLS_BUILD_PATH)/Source/C
 else
   BUILDROOT=$(abspath $(MAKEROOT))
 endif


mdkinney:
@gonzoua Please review

ardbiesheuvel:
I will leave it to @gonzoua to review the actual change, but it looks fine to me. But it wouldn't be better to revert the final change in the original PR, and bring it back (and his fix along with it) after the release?

mdkinney:
> I will leave it to @gonzoua to review the actual change, but it looks fine to me. But it wouldn't be better to revert the final change in the original PR, and bring it back (and his fix along with it) after the release?

I am open to that option as well. If this update is considered low risk and can be merged quickly, then it can go into release.

There are now at least 2 PRs that would have to be reverted if we go that path.

gonzoua:
I don't want to introduce new build-location variables without clear path to deprecate them. The relationship between WORKSPACE, BASE_TOOLS_PATH and various ways EDK2 can be built is more complex than they should be and adding new stuff in the mix.

My preferred fix would be to add an escape hatch  to opt-out of new behavior and deprecate the parameter in a reasonable amount of time. For example env var `BASE_TOOLS_BUILD_COMPAT` that reverts build to the previous behavior when set. This approach would allow downstream project to identify regressions, quickly address it and get some time to address it.

I'm OK with rolling back the current implementation from the stable branch and re-adding it in one of the next releases after discussion and rework too. I'll leave it to maintainers' team to decide as they have more holistic view of the project.

mdkinney:
> I don't want to introduce new build-location variables without clear path to deprecate them. The relationship between WORKSPACE, BASE_TOOLS_PATH and various ways EDK2 can be built is more complex than they should be and adding new stuff in the mix.
> 
> My preferred fix would be to add an escape hatch to opt-out of new behavior and deprecate the parameter in a reasonable amount of time. For example env var `BASE_TOOLS_BUILD_COMPAT` that reverts build to the previous behavior when set. This approach would allow downstream project to identify regressions, quickly address it and get some time to address it.
> 
> I'm OK with rolling back the current implementation from the stable branch and re-adding it in one of the next releases after discussion and rework too. I'll leave it to maintainers' team to decide as they have more holistic view of the project.

I also prefer to not add any new configuration settings.  However, we have to avoid introducing breaking changes.  What you suggest would require all downstream platforms that have one or more dependencies on the in-tree build of tools to add that setting before they can update to latest version of edk2. 

In this specific case, this breaking change is blocking ability to test hard freeze content for the edk2-stable202602 tag before the release.

If this feature needs this additional level of discussion, then my preference is to revert the feature.

mdkinney:
> > I don't want to introduce new build-location variables without clear path to deprecate them. The relationship between WORKSPACE, BASE_TOOLS_PATH and various ways EDK2 can be built is more complex than they should be and adding new stuff in the mix.
> > My preferred fix would be to add an escape hatch to opt-out of new behavior and deprecate the parameter in a reasonable amount of time. For example env var `BASE_TOOLS_BUILD_COMPAT` that reverts build to the previous behavior when set. This approach would allow downstream project to identify regressions, quickly address it and get some time to address it.
> > I'm OK with rolling back the current implementation from the stable branch and re-adding it in one of the next releases after discussion and rework too. I'll leave it to maintainers' team to decide as they have more holistic view of the project.
> 
> I also prefer to not add any new configuration settings. However, we have to avoid introducing breaking changes. What you suggest would require all downstream platforms that have one or more dependencies on the in-tree build of tools to add that setting before they can update to latest version of edk2.
> 
> In this specific case, this breaking change is blocking ability to test hard freeze content for the edk2-stable202602 tag before the release.
> 
> If this feature needs this additional level of discussion, then my preference is to revert the feature.

I will add that I do like the ability to configure where the out-of-tree tools are located.  I am working on github actions with caching features to reduce CI execution times.  Adding the ability to cache the pre-built BaseTools would help.  The CI Agent being able to configure the directory location that is cached would be an important feature.

makubacki:
The [new RFC process](https://github.com/tianocore/tianocore-wiki.github.io/blob/main/rfc/text/0001-rfc-process.md) might be useful to document this feature in more detail and discuss alternatives and compatibility options.

gonzoua:
> What you suggest would require all downstream platforms that have one or more dependencies on the in-tree build of tools to add that setting before they can update to latest version of edk2.

This is in no way  different from a downstream adopting to an upstream's library API change or module removal. We should minimize such occurrences but not at the expense of increased cost of maintenance. Adding a "revert to old behavior" knob as proposed in comment # alleviates some of the pains of fixing this regression downstream.

> In this specific case, this breaking change is blocking ability to test hard freeze content for the edk2-stable202602 tag before the release.

I appologize for the breakage and I'll prioritize fixing it. But I need to know the scope of work to understand whether the solution I proposed is feasible or not and can be done in timely fashion. I am new to the community and not familiar with the stable tag testing process or passing criteria. What's in the scope for testing? How can I reproduce the broken build scenario for tests? What is the deadline for starting testing?

> I will add that I do like the ability to configure where the out-of-tree tools are located. I am working on github actions with caching features to reduce CI execution times. Adding the ability to cache the pre-built BaseTools would help.

FWIW, this does not require additional build target dir variable. With the out-of-tree builds it can be achieved by building just BaseTools, caching `$WORKSPACE/BaseTools` or `$WORKSPACE/BaseTools/Build/Source/C/bin/` and implementing a Linux equivalent of `EDK_TOOLS_BIN` env var that already exists for Windows.

mdkinney:
> > What you suggest would require all downstream platforms that have one or more dependencies on the in-tree build of tools to add that setting before they can update to latest version of edk2.
> 
> This is in no way different from a downstream adopting to an upstream's library API change or module removal. We should minimize such occurrences but not at the expense of increased cost of maintenance. Adding a "revert to old behavior" knob as proposed in comment # alleviates some of the pains of fixing this regression downstream.
> 
> > In this specific case, this breaking change is blocking ability to test hard freeze content for the edk2-stable202602 tag before the release.
> 
> I appologize for the breakage and I'll prioritize fixing it. But I need to know the scope of work to understand whether the solution I proposed is feasible or not and can be done in timely fashion. I am new to the community and not familiar with the stable tag testing process or passing criteria. What's in the scope for testing? How can I reproduce the broken build scenario for tests? What is the deadline for starting testing?
> 
> > I will add that I do like the ability to configure where the out-of-tree tools are located. I am working on github actions with caching features to reduce CI execution times. Adding the ability to cache the pre-built BaseTools would help.
> 
> FWIW, this does not require additional build target dir variable. With the out-of-tree builds it can be achieved by building just BaseTools, caching `$WORKSPACE/BaseTools` or `$WORKSPACE/BaseTools/Build/Source/C/bin/` and implementing a Linux equivalent of `EDK_TOOLS_BIN` env var that already exists for Windows.

I am thinking bigger scope on the cache than just the BaseTools.  We also have binary tools that are pulled from Nuget.  I would like to have a way to put all of those tools under a single parent directory for CI build efficiencies.

I think Ard's suggestion to revert this feature is the best path to make the release on time.  We can discuss how to reduce the number of env vars in both Windows and Linux and remove generation of intermediate files in the source dirs after the release.

mdkinney:
Close this PR and resolve by reverting out-of-tree build in PR https://github.com/tianocore/edk2/pull/12161

PR #12165, opened 2026-02-12
OvmfPkg/IntelTdx: disable shell by default

katexochen:
Comment says it shouldn't be enabled for production. A secure default is likely more sane for TDX.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Applied patch downstream and omitted previously used `-DBUILD_SHELL=FALSE`.

## Integration Instructions

Users who previously built  without `-DBUILD_SHELL=FALSE` do need to explicitly enable shell via `-DBUILD_SHELL=TRUE` if they want to keep that behavior.

### DIFF
--- a/OvmfPkg/IntelTdx/IntelTdxX64.dsc
+++ b/OvmfPkg/IntelTdx/IntelTdxX64.dsc
@@ -34,7 +34,7 @@
   #
   # Shell can be useful for debugging but should not be enabled for production
   #
-  DEFINE BUILD_SHELL             = TRUE
+  DEFINE BUILD_SHELL             = FALSE
 
   #
   # Device drivers


kraxel:
The shell wouldn't be loaded in TDX mode.

The IntelTdx image has two DXE phase firmware volumes. One of them (`FV.NCCFV`) is simply not used when booted in a TDX guest.  The shell is in that bucket, together with other features and drivers which should not be used in a CoCo guest for various reasons.

katexochen:
> The shell wouldn't be loaded in TDX mode.
> 
> The IntelTdx image has two DXE phase firmware volumes. One of them (`FV.NCCFV`) is simply not used when booted in a TDX guest. The shell is in that bucket, together with other features and drivers which should not be used in a CoCo guest for various reasons.

Thanks for the clarification! I don't get how/why `BUILD_SHELL` is only used in `NCCFV`, but I'm also pretty new to how edk2 builds work.

I still wonder how to understand the comment on `BUILD_SHELL` then? Disable it in case you are using TDVF in a non-TDX guest in production?

kraxel:
Not fully sure what exactly the comment refers to.  One known issue is that efi shell is powerful enough that one can bypass secure boot.  Which why OVMF will include the shell binary and related components into the firmware image only in case the secure boot build option is turned off (see `OvmfPkg/Include/Fdf/ShellDxe.fdf.inc`).

katexochen:
> Not fully sure what exactly the comment refers to. One known issue is that efi shell is powerful enough that one can bypass secure boot. Which why OVMF will include the shell binary and related components into the firmware image only in case the secure boot build option is turned off (see `OvmfPkg/Include/Fdf/ShellDxe.fdf.inc`).

Thanks @kraxel, makes sense. I'll close this PR as it looks like it isn't needed.

PR #12167, opened 2026-02-17
Fix string comparisons in python

BMBurstein:
# Description

The old comparisons worked by luck that none of them were a sub-string
of another. This fixes that.

In python `'ABC' in 'ABC'` is True, but so is `'AB' in 'ABC'`.
The old code checked the type of data using (in some cases) code like
`Type in TypeStr`, which would erroneously evaluate to True if `Type`
was a substring of `TypeStr`.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

### DIFF
--- a/BaseTools/Source/Python/FMMT/core/BiosTree.py
+++ b/BaseTools/Source/Python/FMMT/core/BiosTree.py
@@ -28,9 +28,9 @@ PECOFF_TREE = 'PECOFF_TREE'
 RootType = [ROOT_TREE, ROOT_FV_TREE, ROOT_FFS_TREE, ROOT_SECTION_TREE]
 FvType = [FV_TREE, SEC_FV_TREE]
 FfsType = [FFS_TREE, FFS_PAD]
-SecType = SECTION_TREE
+SecType = [SECTION_TREE]
 ElfType = [ROOT_ELF_TREE, ELF_TREE]
-PeCoffTree = PECOFF_TREE
+PeCoffTree = [PECOFF_TREE]
 
 class BIOSTREE:
     def __init__(self, NodeName: str) -> None:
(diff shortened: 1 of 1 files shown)


mikebeaton:
You're talking about code which has changed and requires the further changes you are now making? Please can you reference the commit and line numbers which introduce that change.

BMBurstein:
> You're talking about code which has changed and requires the further changes you are now making? Please can you reference the commit and line numbers which introduce that change.

I am sorry, but I don't understand what you are asking. The commit and line numbers that introduce what change? I am proposing the change, it is not yet in the code

mikebeaton:
@BMBurstein - Apologies, I'm not quite sure what I was asking before either - I misread you as saying you were fixing a previous commit which broke something that was working, but the code has been like this since it was added.

I do agree that the items you're fixing look wrong - but they also seem to me to be unused, both in this file and in any other files that reference this one, that I can see? I'm not sure if there's some future-proofing reason - or existing use case I haven't manged to spot - which is a good reason to keep them (and fix them).

BMBurstein:
@mikebeaton I found this bug when I was using an older version of edk2 where this was the case in a variable that was actually used (https://github.com/tianocore/edk2/blob/stable/202508/BaseTools/Source/Python/FMMT/core/BiosTree.py#L29). When I came to fix it in using the master branch, I found that that particular variable was no longer a problem, but there were other ones that have this issue, so I opened this PR to future-proof it

mikebeaton (approved):
Someone else can state if they think it would be better to remove the unused items?

I agree that all the items where the changes are made are, were or presumably would be (?) used the way stated, so the change is correct - and the PR is good to merge as is.

PR #12170, opened 2026-02-18
SecurityPkg: Move Platform Lockdown to EndOfDxe event

Raymond-MS:
# Description

TPM update flows typically require Platform Auth and typically capsule processor runs in BDS, so deferring until ReadyToBoot in BOOT_ON_FLASH_UPDATE permits TPM capsule update to complete.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Ran CI locally

## Integration Instructions

N/A

### DIFF
--- a/SecurityPkg/Tcg/Tcg2PlatformDxe/Tcg2PlatformDxe.c
+++ b/SecurityPkg/Tcg/Tcg2PlatformDxe/Tcg2PlatformDxe.c
@@ -9,51 +9,36 @@
 #include <PiDxe.h>
 
 #include <Library/DebugLib.h>
+#include <Library/HobLib.h>
 #include <Library/UefiBootServicesTableLib.h>
 #include <Library/UefiLib.h>
 #include <Library/TpmPlatformHierarchyLib.h>
 #include <Protocol/DxeSmmReadyToLock.h>
 
 /**
-   This callback function will run at the SmmReadyToLock event.
+   This callback function will run at EndOfDxe or ReadyToBoot based on boot mode.
 
    Configuration of the TPM's Platform Hierarchy Authorization Value (platformAuth)
    and Platform Hierarchy Authorization Policy (platformPolicy) can be defined through this function.
 
   @param  Event   Pointer to this event
-  @param  Context Event hanlder private data
+  @param  Context Event handler private data
  **/
 VOID
 EFIAPI
-SmmReadyToLockEventCallBack (
+TpmReadyToLockEventCallBack (
   IN EFI_EVENT  Event,
   IN VOID       *Context
   )
 {
-  EFI_STATUS  Status;
-  VOID        *Interface;
-
-  //
-  // Try to locate it because EfiCreateProtocolNotifyEvent will trigger it once when registration.
-  // Just return if it is not found.
-  //
-  Status = gBS->LocateProtocol (
-                  &gEfiDxeSmmReadyToLockProtocolGuid,
-                  NULL,
-                  &Interface
-                  );
-  if (EFI_ERROR (Status)) {
-    DEBUG ((DEBUG_ERROR, "Failed to locate gEfiDxeSmmReadyToLockProtocolGuid.\n"));
-    return;
-  }
-
+  DEBUG ((DEBUG_INFO, "[%a] Disabling TPM Platform Hierarchy\n", __func__));
   ConfigureTpmPlatformHierarchy ();
 
   gBS->CloseEvent (Event);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


jyao1:
First, a "platform" lib usually means that it requires platform customization. As such, I do not suggest to change the "template". (That means you can create your own Tpm2PlatformLib)

Second, by definition, a ReadyToBoot is too late for security action. EndofDxe is right event. If a platform has such special requirement e.g. FIRMWARE_UPDATE, and the platform owner think it is SAFE to perform the security work late, I recommend the platform owner to **signal the EndOfDxe late**, instead of changing every EndOfDxe handler. (That means you can change the BDS on where the platform signal EndOfDxe)

Raymond-MS:
This is still on my list, it's just in my backlog. I need to get information from the team to provide information regarding why the change was made.

apop5:
> First, a "platform" lib usually means that it requires platform customization. As such, I do not suggest to change the "template". (That means you can create your own Tpm2PlatformLib)
> 
> Second, by definition, a ReadyToBoot is too late for security action. EndofDxe is right event. If a platform has such special requirement e.g. FIRMWARE_UPDATE, and the platform owner think it is SAFE to perform the security work late, I recommend the platform owner to **signal the EndOfDxe late**, instead of changing every EndOfDxe handler. (That means you can change the BDS on where the platform signal EndOfDxe)

If we want to treat this driver as a template, then there is good reason to include these changes. To be able to update TPM firmware, the TPM cannot be locked down at EndOfDxe. It requires that the TPM is still accessible after consoles, etc are available so that the update can occur.

This is why this change is only scoped to change to ReadyToBoot when the system boot mode is boot on flash update.
As for platforms needing to provide their own version of this file, I do not disagree that they are still free to do so. But we should give the platforms a baseline of functionality by including this in the Edk2 Version so they can reference this instead of creating even less secure custom implementations which never implement a security boundary.

Raymond-MS:
@jyao1 Is the information that Aaron provided enough to justify our reasoning in upstreaming this change? Please let me know what you think when you get a moment. Thanks!

SecurityPkg/Tcg/Tcg2PlatformDxe/Tcg2PlatformDxe.c
```diff
@@ -69,21 +54,29 @@ Tcg2PlatformDxeEntryPoint (
...
+                    EVT_NOTIFY_SIGNAL,
+                    TPL_CALLBACK,
+                    TpmReadyToLockEventCallBack,
+                    NULL,
+                    &gEfiEndOfDxeEventGroupGuid,
+                    &Event
+                    );
   }
 
+  ASSERT_EFI_ERROR (Status);
```
cfernald, line 79: I don't think we want to mask this. We should fail instead,

jyao1:
Some feedback:
1) EndOfDxe is an architecture point defined by PI specification. The goal is to introduce a known checkpoint for security actions.
2) This PR breaks the threat model defined EDK2, because it is trying to move a security lock from EndOfDxe to ReadyToBoot.

Some suggestion:
1) We still want the reference code to be secure by default. That is a golden rule.
2) A specific platform may choose to adjust the threat model based on its own need. What is why we create Tcg2Platform (but not make it in Tcg2Dxe). EDK2 allows the platform to override. Or if the override is considered as common, the alternative is to have a PCD to control the insecure override. But the default policy setting should still be secured.
3) If we really want to change it as default behavior in EDK2 reference, we need to raise to EDK2 community and let community agree the threat model change for TPM2 firmware update. (More voice is welcome)

apop5:
> Some feedback:
> 
> 1. EndOfDxe is an architecture point defined by PI specification. The goal is to introduce a known checkpoint for security actions.
> 2. This PR breaks the threat model defined EDK2, because it is trying to move a security lock from EndOfDxe to ReadyToBoot.
> 
> Some suggestion:
> 
> 1. We still want the reference code to be secure by default. That is a golden rule.
> 2. A specific platform may choose to adjust the threat model based on its own need. What is why we create Tcg2Platform (but not make it in Tcg2Dxe). EDK2 allows the platform to override. Or if the override is considered as common, the alternative is to have a PCD to control the insecure override. But the default policy setting should still be secured.
> 3. If we really want to change it as default behavior in EDK2 reference, we need to raise to EDK2 community and let community agree the threat model change for TPM2 firmware update. (More voice is welcome)

@jyao1 Thank you for the detailed feedback

Raymond-MS:
Closing for now. Plans to discuss further and decide on how to move forward.

PR #12181, opened 2026-02-20
UefiCpuPkg, MdeModulePkg, OvmfPkg: Use 256 IDT Descriptors In SEC, PEI, SMM

os-d:
# Description

Currently, SEC, PEI, and SMM only set up 32 - 34 IDT descs. This means that any exceptions received above that number will cause a #GP fault.

This commit updates these environments to follow DXE and set up the IDT with the architecturally defined 256 entries, each of which point to the common exception handler.

The SEC IDT is used throughout PEI, even though the PEI lib exists, it is not called. So even though we may not care about stack cookie verification in SEC, if we want it in PEI, we need the SEC IDT to be correct.

The IA32 DxeIpl needs an update because the IDT it creates is in use until CpuDxe comes up and sets up an IDT with 256 descriptors.

This does two things:
- Has well defined behavior for receiving exceptions in these environments instead of relying on a #GP fault to occur.
- Allows the stack cookie interrupt (0x42) to be routed correctly instead of appearing as a #GP fault.

This was discovered while working to enable stack cookie support for CLANGPDB, but this change is independent of other PRs I will put up for that. Neither rely on each other, this PR just makes a better experience for the other PRs (able to identify the stack cookie exception and print useful information instead of just seeing a #GP fault.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on OVMF IA32X64 and OVMF X64 by taking stack cookie exceptions in PEI, Traditional SMM, and Standalone MM.

## Integration Instructions

N/A.


os-d:
I can't add you as reviewers because my write access is disabled during the hard freeze, but @makubacki @kuqin12 please review as well

ardbiesheuvel (approved):

khaliid2040:
I was just wondering if S3 resume IDT needs to setup 256 also, instead first 32 descriptors. The IDT in [PiSmmCpuDxe](https://github.com/tianocore/edk2/blob/master/UefiCpuPkg/PiSmmCpuDxeSmm/CpuS3.c#L109)

os-d:
> I was just wondering if S3 resume IDT needs to setup 256 also, instead first 32 descriptors. The IDT in [PiSmmCpuDxe](https://github.com/tianocore/edk2/blob/master/UefiCpuPkg/PiSmmCpuDxeSmm/CpuS3.c#L109)

There were a couple of cases I chose not to update: IA32 PEI S3, IntelFsp2Pkg's SEC (used in API mode), and UefiPayloadPkg IA32 Entry. The reason I didn't was that I personally don't use any of these cases and so I am not informed enough to know if there is some reason this should not be done (tbh, I also don't use traditional SMM, for its inherent security issues, but that is updated more or less for free when I update StMM). I wanted to keep the blast radius of this down to what I could actually test. I almost didn't update the IA32 DxeIpl, but I did because that was the first case I hit, when testing my new stack cookie PR on OVMFIA32X64.

If any users of the above cases want me to update them as well, I certainly can. Without this change, exceptions above 32 will result in a #GP fault. With this change it will hit the default exception handler. The change is really just exchanging one fault for another (that is more actionable and defined). So we aren't missing something if we don't cover the other cases except for better debuggability.

khaliid2040:
> If any users of the above cases want me to update them as well, I certainly can. Without this change, exceptions above 32 will result in a #GP fault. With this change it will hit the default exception handler. The change is really just exchanging one fault for another (that is more actionable and defined). So we aren't missing something if we don't cover the other cases except for better debuggability.

I am not sure about others too. However, about S3 IDT i would suggest you update it. In edk2 exceptions are confusing. It was a while ago when i was debugging #UD and i had to dig deep into assembly to see if any invalid instructions used until i found out it was null pointer dereference, which i was expecting to throw #PF just like OS.

I already have OVMF image with traditional SMM and S3 enabled. Given that, i would assume it will be painful to debug #GP when the issue was due stack cookie.

If you don't have environment for it, then i can test for you with the above config but i am sure you won't need it.

os-d:
> If you don't have environment for it, then i can test for you with the above config but i am sure you won't need it.

@khaliid2040 I have updated S3 as well. Please do test at least normal operation and confirm the S3 boot works.

UefiCpuPkg/Library/CpuExceptionHandlerLib/SecPeiCpuException.c
```diff
@@ -130,11 +130,11 @@ InitializeCpuExceptionHandlers (
...
   AsmReadIdtr (&IdtDescriptor);
   IdtEntryCount = (IdtDescriptor.Limit + 1) / sizeof (IA32_IDT_GATE_DESCRIPTOR);
-  if (IdtEntryCount > CPU_EXCEPTION_NUM) {
+  if (IdtEntryCount > CPU_INTERRUPT_NUM) {
     //
-    // CPU exception library only setup CPU_EXCEPTION_NUM exception handler at most
+    // CPU exception library only setup CPU_INTERRUPT_NUM exception handler at most
     //
-    IdtEntryCount = CPU_EXCEPTION_NUM;
+    IdtEntryCount = CPU_INTERRUPT_NUM;
```
niruiyu: UefiCpuPkg/Library/CpuExceptionHandlerLib/X64/SecPeiExceptionHandlerAsm.nasm still sets up only 32 interrupt handlers.
There is a bug here that the IDT[32..255] points to invalid code.

niruiyu: The original intention of using only 32 entries was to reduce the code size in pre-mem XIP phase when the code cannot be compressed.
With this change, we would increase (256-32)*X bytes in SEC. Can you kindly check the exact number of bytes increased in SecMain module?
+ @mdkinney

os-d: Thank you, this is an important catch. It worked in my tests because `UpdateIdtTable()` copied off the end of the template map, so the IDT entries were correct, but we overwrote whatever was past that, which is dangerous. I have fixed up the ASM, testing again and will push.

For SecMain's size, on OVMF X64 with VS2022 it went from

image

to

image

os-d: @niruiyu I retested with VS2022 and CLANGPDB and both boot to shell and correctly trap the stack cookie exception if I add one. Please re-review when you have a chance.

niruiyu: thank you for the additional change and providing the size change data.
It seems now X64/SecPeiExceptionHandlerAsm.nasm and X64/ExceptionHandlerAsm.nasm are the same except:
1. X64/ExceptionHandlerAsm.nasm supports XCODE toolchain (XCODE toolchain defines NO_ABSOLUTE_RELOCS_IN_TEXT)
2. X64/ExceptionHandlerAsm.nasm supports CET

Then, do you mind taking an extra step to remove X64/SecPeiExceptionHandlerAsm.nasm?

@ardbiesheuvel please confirm if the above is doable as you added the XCODE support.

os-d: I was wondering the same thing. I tested with removing the SecPei version and OVMF IA32/64 and X64 boots to shell (and my other test cases injecting a stack cookie exception show up correctly). I don't have the facility to test XCODE though, so will wait to hear from Ard before pushing the change.

ardbiesheuvel: I don't understand the question sorry

os-d: @ardbiesheuvel the question is: Is it okay to get rid of SecPeiExceptionHandlerAsm.nasm that you added here: https://github.com/tianocore/edk2/commit/95f03309537d567655b90461b15d7cba0d6d1887 ?.

Ray was pointing out that the delta between it and X64/ExceptionHandlerAsm.nasm seems trivial now and indeed on VS2022/CLANGPDB it seems fine to only use X64/ExceptionHandlerAsm.nasm, but we don't know about for XCODE, which seems the impetus for adding it.

ardbiesheuvel: I see. So the difference is that those files have different approaches to work around the fact that XCODE does not permit absolute references in .text
- the SEC/PEI flavor just emits those entries in .data, on the basis that the whole image is mapped read-only and executable in practice anyway,
- the DXE flavor has some XCODE specific startup code that patches the .text section at runtime, and this only works if .text is not mapped read-only.

Those workarounds are mutually incompatible (.data is not mapped executable in DXE and .text is not writable in SEC/PEI) so merging the two files would involve some additional ifdeffery to distinguish the SEC/PEI workaround from the DXE one. Beyond that, I see no problems with combining the two.

os-d: @niruiyu @ardbiesheuvel I have pushed another commit that merges the two files. Please check to confirm this matches your expectations. For reference, I also attached the diff between the two files (before this commit). Please verify nothing else needs to change from there, I'm not familiar with the hook stub part and why that was different.

I did boot test this with VS2022/CLANGPDB on OVMF IA32/X64 and X64 and tested the stack cookie exception was working correctly in PEI.

[exception_handler.txt](https://github.com/user-attachments/files/25614783/exception_handler.txt)

UefiCpuPkg/Library/CpuExceptionHandlerLib/X64/ExceptionHandlerAsm.nasm
```diff
@@ -73,13 +81,18 @@ HookAfterStubHeaderBegin:
...
+%ifdef SMM_DXE_NO_ABSOLUTE_RELOCS_IN_TEXT
     mov     rax, strict qword 0 ;     mov     rax, HookAfterStubHeaderEnd
 JmpAbsoluteAddress:
 %else
     mov     rax, HookAfterStubHeaderEnd
 %endif
     jmp     rax
+
+%ifndef SEC_PEI_NO_ABSOLUTE_RELOCS_IN_TEXT
+SECTION .text
```
ardbiesheuvel, line 94: Isn't this redundant?

os-d: Redundant yes, but also incorrect. This should have been `%ifdef` not `%ifndef`. Thanks for catching will fix.

UefiCpuPkg/Library/CpuExceptionHandlerLib/X64/ExceptionHandlerAsm.nasm
```diff
@@ -4,14 +4,20 @@
...
 ;   x64 CPU Exception Handler
 ;
 ; Notes:
 ;
+; This module supports XCODE which does not permit absolute references in
+; .text sections. This must be handled differently in SEC/PEI vs SMM/DXE. In
+; SEC/PEI we emit the IDT vectors in .data because the entire image is R+X.
+; In SMM/DXE, we patch the .text section at runtime because .data is not executable,
+; but it requires a writeable .text section (which SEC/PEI does not have).
+;
```
niruiyu: Thanks for adding the comments to explain. It did take me a while to remember/understand why Ard did the changes. But I feel we could add more contexts here. So I did some change to your comments as the below. Please feel free to adjust the wording as you like.

This module uses the build options SEC_PEI_NOABSOLUTE_RELOCS_IN_TEXT and SMM_DXE_NO_ABSOLUTE_RELOCS_IN_TEXT to work around Xcode’s linker, which disallows absolute references in .text. For SEC/PEI the IDT stubs are emitted into .data (which Xcode accepts). In SMM/DXE, however, .data is not executable, so the IDT stubs remain in .text and are initially patched by GetTemplateAddressMap(). The SMM/DXE method cannot be applied to SEC/PEI because .text cannot be modified in place when the module runs XIP.

os-d:
@niruiyu I have updated the comment as requested, please re-review when you have a chance

UefiCpuPkg/Library/CpuExceptionHandlerLib/X64/ExceptionHandlerAsm.nasm
niruiyu: I think you can just add "SECTION .text" here, without checking the macro.

os-d: Dropped the macro

os-d:
@niruiyu I addressed the latest comment, please re-review

UefiCpuPkg/Library/CpuExceptionHandlerLib/X64/ExceptionHandlerAsm.nasm
tlendacky: One of the reasons for separating out the XCODE5 assembler is documented in commits ec94e97a6eb1 ("UefiCpuPkg/CpuExceptionHandler: Make XCODE5 changes toolchain specific") and 9378310dd877 ("UefiCpuPkg/CpuExceptionHandler: Revert CpuExceptionHandler binary patching"). This binary patching prevented use of the CPU exception handling under SEV-ES because it resulted in an infinite nested page fault.

Is the binary patching still there and do the ifdef conditions protect against that? If they don't, you can't combine these files.

os-d: @tlendacky thanks for the history here. The change you referenced was already combined into a common SEC/PEI file here: https://github.com/tianocore/edk2/commit/95f03309537d567655b90461b15d7cba0d6d1887. That accounts for not patching code in flash.

The current commit simply recombines the SEC/PEI one with the SMM/DXE one, it doesn't (hopefully) change any behavior w.r.t. binary patching. That was done in the commit referenced above. The ifdefs in place account for the differences between the PEI/SEC case (patch .data) and the SMM/DXE case (patch .text). Both of those cases only occur if XCODE is being used.

tlendacky: @os-d, cool, thanks for that history, too!

UefiCpuPkg/Library/CpuExceptionHandlerLib/X64/ExceptionHandlerAsm.nasm
```diff
@@ -4,14 +4,22 @@
...
-;   ExceptionHandlerAsm.Asm
+;   ExceptionHandlerAsm.nasm
 ;
 ; Abstract:
 ;
 ;   x64 CPU Exception Handler
 ;
 ; Notes:
 ;
+; This module uses the build options SEC_PEI_NOABSOLUTE_RELOCS_IN_TEXT and
```
khaliid2040: There is a typo for the macro name  it should be `SEC_PEI_NO_ABSOLUTE_RELOCS_IN_TEXT`

lgao4:
I see the same macro is defined in more than source files. How about define one MACRO in UefiCpuPkg public header file, then all modules except for DxeIpl consume it. If so, only MdeModulePkg DxeIpl needs to define another macro.

os-d:
> I see the same macro is defined in more than source files. How about define one MACRO in UefiCpuPkg public header file, then all modules except for DxeIpl consume it. If so, only MdeModulePkg DxeIpl needs to define another macro.

@lgao4 SEC_IDT_ENTRY_COUNT is defined in DxeIpl, UefiCpuPkg/Sec, and OvmfPkg/Sec. OvmfPkg/Sec is a forked version of UefiCpuPkg/Sec made for the considerations of a virtual platform. In the general case, it has special considerations it must do for being a virtual platform and may not want to depend on what UefiCpuPkg is doing (which is why it was forked of course). So usually I would be all for a single definition, but in this case I think it makes sense for OvmfPkg to decide on its own what it wants to do here.

Let me know what you think. If you would prefer the shared definition between UefiCpuPkg and OvmfPkg, I will do that.

UefiCpuPkg/Library/SmmRelocationLib/SmmRelocationLib.c
```diff
@@ -274,10 +274,10 @@ InitSmmIdt (
   CONST EFI_PEI_SERVICES  **PeiServices;
 
   //
-  // There are 32 (not 255) entries in it since only processor
-  // generated exceptions will be handled.
+  // Populate 255 entries to ensure that a known state occurs
+  // for all possible exception vectors.
   //
-  gcSmmInitIdtr.Limit = (sizeof (IA32_IDT_GATE_DESCRIPTOR) * 32) - 1;
+  gcSmmInitIdtr.Limit = (sizeof (IA32_IDT_GATE_DESCRIPTOR) * 256) - 1;
```
lgao4: Here , the hard code value can be used by the macro.

UefiCpuPkg/PiSmmCpuDxeSmm/CpuS3.c
```diff
@@ -105,9 +105,9 @@ SmmRestoreCpu (
     //
     // Setup X64 IDT table
     //
-    ZeroMem (IdtEntryTable, sizeof (IA32_IDT_GATE_DESCRIPTOR) * 32);
+    ZeroMem (IdtEntryTable, sizeof (IA32_IDT_GATE_DESCRIPTOR) * 256);
```
lgao4: Here , the hard code value can be used by the macro.

UefiCpuPkg/PiSmmCpuDxeSmm/PiSmmCpuCommon.c
```diff
@@ -138,10 +138,10 @@ InitializeSmmIdt (
   IA32_DESCRIPTOR  DxeIdtr;
 
   //
-  // There are 32 (not 255) entries in it since only processor
-  // generated exceptions will be handled.
+  // Populate 255 entries to ensure that a known state occurs
+  // for all possible exception vectors.
   //
-  gcSmiIdtr.Limit = (sizeof (IA32_IDT_GATE_DESCRIPTOR) * 32) - 1;
+  gcSmiIdtr.Limit = (sizeof (IA32_IDT_GATE_DESCRIPTOR) * 256) - 1;
```
lgao4: Here , the hard code value can be used by the macro.

lgao4:
> > I see the same macro is defined in more than source files. How about define one MACRO in UefiCpuPkg public header file, then all modules except for DxeIpl consume it. If so, only MdeModulePkg DxeIpl needs to define another macro.
> 
> @lgao4 SEC_IDT_ENTRY_COUNT is defined in DxeIpl, UefiCpuPkg/Sec, and OvmfPkg/Sec. OvmfPkg/Sec is a forked version of UefiCpuPkg/Sec made for the considerations of a virtual platform. In the general case, it has special considerations it must do for being a virtual platform and may not want to depend on what UefiCpuPkg is doing (which is why it was forked of course). So usually I would be all for a single definition, but in this case I think it makes sense for OvmfPkg to decide on its own what it wants to do here.
> 
> Let me know what you think. If you would prefer the shared definition between UefiCpuPkg and OvmfPkg, I will do that.

I understand. So, I suggest to define one MACRO in UefiCpuPkg, and use it for the modules in UefiCpuPkg. We can unify this macro in UefiCpuPkg. Now, there are three definitions of CPU_INTERRUPT_NUM in UefiCpuPkg. They can be combined to single one.

UefiCpuPkg/PiSmmCpuDxeSmm/PiSmmCpuCommon.c
lgao4: Can it be replaced by CPU_INTERRUPT_NUM?

os-d:
> > > I see the same macro is defined in more than source files. How about define one MACRO in UefiCpuPkg public header file, then all modules except for DxeIpl consume it. If so, only MdeModulePkg DxeIpl needs to define another macro.
> > 
> > 
> > @lgao4 SEC_IDT_ENTRY_COUNT is defined in DxeIpl, UefiCpuPkg/Sec, and OvmfPkg/Sec. OvmfPkg/Sec is a forked version of UefiCpuPkg/Sec made for the considerations of a virtual platform. In the general case, it has special considerations it must do for being a virtual platform and may not want to depend on what UefiCpuPkg is doing (which is why it was forked of course). So usually I would be all for a single definition, but in this case I think it makes sense for OvmfPkg to decide on its own what it wants to do here.
> > Let me know what you think. If you would prefer the shared definition between UefiCpuPkg and OvmfPkg, I will do that.
> 
> I understand. So, I suggest to define one MACRO in UefiCpuPkg, and use it for the modules in UefiCpuPkg. We can unify this macro in UefiCpuPkg. Now, there are three definitions of CPU_INTERRUPT_NUM in UefiCpuPkg. They can be combined to single one.

@lgao4 , I agree with consolidating the definitions. This was a little awkward because of the spread of the components using this and that CpuExceptionHandlerLib.h is in MdeModulePkg, which is shared across architectures. So I ended up defining `X86_CPU_INTERRUPT_NUM` there. Because this is now a standard definition and not a choice of UefiCpuPkg, I did use it across UefiCpuPkg, MdeModulePkg, and OvmfPkg. Let me know if this works. If this looks good to you (and CI passes of course), I think we can apply the push label because other maintainers have approved.

PR #12199, opened 2026-02-25
BaseTools/Source/C/VfrCompile: Fix parallel make failures

mdkinney:
# Description

Update `VfrCompile` makefile rules to run `antlr` and `dlg` to completion before compiling any of the generated `.cpp` files.

Without this change, parallel make may start compiling some of the `.cpp` files before both `antlr` and `dlg` have finished which produces syntax errors from compilation with partially generated files.

Also use `&:` in the makefile rule that runs `antlr` and `dlg` so the targets are treated as a group. This guarantees that the rule is only executed once for the entire group. Without this change, parallel make may run the rule actions more than once for each target and this may modify output files while they are being used in a `.cpp` compilation rule.

The regression in parallel make of `VfrCompiler` was introduced by PR https://github.com/tianocore/edk2/pull/11757.  Specifically commit https://github.com/tianocore/edk2/commit/4cfa0c4911a264b77f6f4fc73d75bef5e2daa1e1. However, the changes in this commit are a good change. Building the `VfrCompiler` should not depend on the extra make operation for `VfrCompile/VfrLexer.h` before the `VfrCompiler` make is run. This extra make action was actually masking a real issue in the `VfrCompiler` makefile. In order for `VfrCompiler` to build correctly with parallel make, the `antlr` and `dlg` actions must both be completed before any of the generated files are used in a `.cpp` compile action.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Reproduced build failure of `VfrCompiler` in Linux/GCC environment with `make -C . -j 16`. May take several attempts to reproduce the issue.  Then applied this change that resolves race condition for parallel make and verifies that the build failure does not occur in a large number of `make -C . -j 16` invocations.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Library/PeiDebugLibDebugPpi/DebugLib.c
+++ b/MdeModulePkg/Library/PeiDebugLibDebugPpi/DebugLib.c
@@ -150,7 +150,7 @@ VaListToBaseList (
     // Parse Flags and Width
     //
     for (Format++; TRUE; Format++) {
-      if ((*Format == '.') || (*Format == '-') || (*Format == '+') || (*Format == ' ')) {
+      if ((*Format == '.') || (*Format == '-') || (*Format == '+') || (*Format == ' ') || (*Format == ',')) {
         //
         // These characters in format field are omitted.
         //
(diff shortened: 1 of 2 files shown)


mdkinney:
This issue is a regression in the edk2-stable202602 release. 

@leiflindholm @ardbiesheuvel @makubacki @ajfish This change should be considered for a hot fix against the edk2-stable202602 release.

mdkinney:
> Looks reasonable. Nit: The PR description and commit message use "antl" in a few places instead of "antlr".

I have fixed the commit message and the PR description

PR #12200, opened 2026-02-25
OvmfPkg/IntelTdx: add TdxDxeCpuExceptionHandlerLib for IST skip

sgrams:
In a TDX guest, the LTR (Load Task Register) instruction causes an unconditional #UD.  The initial TD vCPU state per the Intel TDX architecture specification has TR.Selector = 0 and TR.Base = 0, so there is no usable TSS to write IST entries into.  IST-based separate exception stacks cannot be configured at all inside a TDX guest.
ArchSetupExceptionStack calls AsmWriteTr (LTR) unconditionally, which causes CpuDxe to crash before BDS is reached.  The same code path is triggered for every AP via InitializeMpExceptionStackSwitchHandlers in CpuDxe/CpuMp.c, which dispatches InitializeExceptionStackSwitchHandlers to all processors and calls InitializeSeparateExceptionStacks for each.

Fix: introduce OvmfPkg/IntelTdx/TdxCpuExceptionHandlerLib, a thin OvmfPkg-local override of DxeCpuExceptionHandlerLib where InitializeSeparateExceptionStacks unconditionally returns EFI_SUCCESS
with BufferSize left at 0. 
The wrapper in CpuMp.c treats this as a successful no-op (Status == EFI_SUCCESS, BufferSize == 0) for every processor and skips the second allocation and setup round entirely, so AsmWriteTr is never reached on BSP or any AP.

Point IntelTdxX64.dsc DXE_CORE and DXE_DRIVER entries at the new library rather than modifying the upstream UefiCpuPkg library directly.

Fixes: e67f405713f0 ("UefiCpuPkg: Always Initialize Separate AP Exception Stacks")
# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions

### DIFF
--- a/OvmfPkg/IntelTdx/IntelTdxX64.dsc
+++ b/OvmfPkg/IntelTdx/IntelTdxX64.dsc
@@ -252,7 +252,7 @@
   DebugLib|OvmfPkg/Library/PlatformDebugLibIoPort/PlatformDebugLibIoPort.inf
 !endif
   ExtractGuidedSectionLib|MdePkg/Library/DxeExtractGuidedSectionLib/DxeExtractGuidedSectionLib.inf
-  CpuExceptionHandlerLib|UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHandlerLib.inf
+  CpuExceptionHandlerLib|OvmfPkg/IntelTdx/TdxCpuExceptionHandlerLib/TdxDxeCpuExceptionHandlerLib.inf
   PcdLib|MdePkg/Library/DxePcdLib/DxePcdLib.inf
 
 [LibraryClasses.common.DXE_RUNTIME_DRIVER]
@@ -308,7 +308,7 @@
   PlatformBmPrintScLib|OvmfPkg/Library/PlatformBmPrintScLib/PlatformBmPrintScLib.inf
   QemuBootOrderLib|OvmfPkg/Library/QemuBootOrderLib/QemuBootOrderLib.inf
   PlatformBootManagerCommonLib|OvmfPkg/Library/PlatformBootManagerCommonLib/PlatformBootManagerCommonLib.inf
-  CpuExceptionHandlerLib|UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHandlerLib.inf
+  CpuExceptionHandlerLib|OvmfPkg/IntelTdx/TdxCpuExceptionHandlerLib/TdxDxeCpuExceptionHandlerLib.inf
   LockBoxLib|OvmfPkg/Library/LockBoxLib/LockBoxDxeLib.inf
   PciLib|OvmfPkg/Library/DxePciLibI440FxQ35/DxePciLibI440FxQ35.inf
   MpInitLib|UefiCpuPkg/Library/MpInitLib/DxeMpInitLib.inf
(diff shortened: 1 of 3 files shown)


OvmfPkg/Include/WorkArea.h
```diff
@@ -109,6 +109,23 @@ typedef struct _TDX_WORK_AREA {
...
+  /// pointers directly into the existing TSS, bypassing the LTR instruction
+  /// that is not permitted inside a TDX guest.
+  ///
+  /// Layout note: MailboxGdt is at offset 128 (fixed by ApRunLoop.nasm).
+  /// sizeof(MAILBOX_GDT) = sizeof(IA32_DESCRIPTOR /*packed, 10B*/) + 40 = 50.
+  /// The next UINT64-aligned boundary after 128+50=178 is 184, so
+  /// VcpuTssBase lives at offset 184 from PcdOvmfWorkAreaBase.
+  /// The assembly constant TDX_WORK_AREA_VCPU_TSS_BASE in IntelTdx.asm
+  /// MUST equal (PcdOvmfWorkAreaBase + 184).
+  ///
```
sgrams: long comment to be dropped

OvmfPkg/IntelTdx/TdxDxeCpuExceptionHandlerLib/TdxDxeCpuExceptionHandlerLib.inf
```diff
@@ -0,0 +1,78 @@
...
+#  For TDX guests, LTR causes #UD so a new GDT+TSS cannot be installed in the
+#  normal way.  ReloadFlat32 (OvmfPkg/ResetVector/Ia32/IntelTdx.asm) saves
+#  the base address of the TSS that the TDX module had TR pointing to into
+#  TDX_WORK_AREA.VcpuTssBase before lgdt replaces the GDT.
+#  InitializeSeparateExceptionStacks detects TDX, allocates IST stacks from
+#  the caller-supplied (or static) buffer, and writes the IST pointers
+#  directly into the existing TSS.  The IDT IST fields are updated in the
+#  same way as the normal path.  No LGDT or LTR is needed.
+#
+#  Copyright (c) 2013 - 2024, Intel Corporation. All rights reserved.<BR>
```
sgrams: - 2026

OvmfPkg/IntelTdx/TdxDxeCpuExceptionHandlerLib/TdxDxeException.c
```diff
@@ -0,0 +1,293 @@
...
+  For TDX guests, LTR (#UD in TDX guest mode) cannot be used to install a
+  new GDT+TSS.  Instead, ReloadFlat32 in OvmfPkg/ResetVector/Ia32/IntelTdx.asm
+  saves the base address of the TSS that the TDX module had TR pointing to
+  (TDX_WORK_AREA.VcpuTssBase) before the firmware's lgdt clobbers the GDT.
+  InitializeSeparateExceptionStacks detects TDX, allocates IST stacks from
+  its caller-supplied (or static) buffer, and writes the IST pointers directly
+  into the already-loaded TSS memory.  The IDT IST fields are updated in the
+  same way as the normal path.  No LGDT or LTR is needed.
+
+  Copyright (c) 2013 - 2022, Intel Corporation. All rights reserved.<BR>
```
sgrams: - 2026

sgrams:
My first assumption was that LTR is explicitly disallowed in TDX but now when testing with top of the tree OvmfPkg I cannot reproduce the failure. Trending to close this PR

PR #12202, opened 2026-02-25
BaseTools: Enable control flow guard for Windows applications

kuqin12:
# Description

This change introduces a build flag to enable control flow guard for Windows host applications. This is to prevent attackers from redirecting execution to unexpected, unsafe locations.

- [ ] Breaking change?
- [x] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested on both x64 and arm64 Windows host systems and built UEFI images properly.

## Integration Instructions

N/A


kuqin12:
@gapalomi @lgao4 @mikebeaton @ponchofigueroa @YuweiChen1110 Could you please review the change and let me know how you think of the change? Thanks in advance!

lgao4:
> @gapalomi @lgao4 @mikebeaton @ponchofigueroa @YuweiChen1110 Could you please review the change and let me know how you think of the change? Thanks in advance!

Which VS version has been verified?

kuqin12:
> Which VS version has been verified?

@lgao4 it was verified on VS2022. Do you have other targets in mind? I will test those and report back.

mikebeaton:
> @gapalomi @lgao4 @mikebeaton @ponchofigueroa @YuweiChen1110 Could you please review the change and let me know how you think of the change? Thanks in advance!

As far as I can tell, the first commit is a reasonable hardening of the Windows BaseTools.

However, isn't the second commit somewhat unrelated, and perhaps suitable as a separate PR? Also, would we not be better to fix the source of the warnings (e.g. add casts to (UINTN) etc. as needed) rather than suppress the warning?

kuqin12:
> However, isn't the second commit somewhat unrelated, and perhaps suitable as a separate PR? Also, would we not be better to fix the source of the warnings (e.g. add casts to (UINTN) etc. as needed) rather than suppress the warning?

The reason I suppressed the warning was because the issue has spread through a handful locations in BaseTools, so I followed the precedence by suppressing them. I will file a separate PR to fix them first as they are breaking the build with the latest VS2022.

kuqin12:
A separate PR is created to fix this specific build break: https://github.com/tianocore/edk2/pull/12272. Will let the other one merge first.

lgao4:
> > Which VS version has been verified?
> 
> @lgao4 it was verified on VS2022. Do you have other targets in mind? I will test those and report back.

tools_def.template lists VS2015, VS2017, VS2019, VS2022, and VS2026 tool chain. I think they should be verified.

kuqin12:
> tools_def.template lists VS2015, VS2017, VS2019, VS2022, and VS2026 tool chain. I think they should be verified.

@lgao4 I tried to build with VS2019, VS2022 and VS2026, these all work. But VS2017 was broken, yet not due to the control flow guard added. We _can_ try to suppress it as this is from a submodule, but it seems odd on why this only fails on VS2017.

```
brotli\c\dec\bit_reader.c(29): warning C4028: formal parameter 1 different from declaration
brotli\c\dec\bit_reader.c(34): warning C4028: formal parameter 1 different from declaration
brotli\c\dec\bit_reader.c(59): warning C4028: formal parameter 1 different from declaration
```

I did not attempt to make VS2015 as it was no longer supported in edk2-pytool: https://github.com/tianocore/edk2-pytool-library/blob/ea78c48153a9b1d8784f3d2c3cab92add0f438e6/edk2toollib/windows/locate_tools.py#L51, and does allow me to install when newer versions are available. Should we remove the VS2015 support already?

But please let me know you would like to proceed with these 2 issues.

kuqin12:
> A separate PR is created to fix this specific build break: #12272. Will let the other one merge first.

The warning change is merged. This PR is updated to drop the warning suppression commit.

mikebeaton:
nit: there are unrelated whitespace changes in ms.app - I think these should be in a separate commit, if they are required at all (unless uncrustify is requiring them, but I don't think it applies here?).

kuqin12:
> nit: there are unrelated whitespace changes in ms.app - I think these should be in a separate commit, if they are required at all (unless uncrustify is requiring them, but I don't think it applies here?).

This was unfortunately called out by PatchCheck.py because of the tab used in the changed location. Will split

mikebeaton:
> > nit: there are unrelated whitespace changes in ms.app - I think these should be in a separate commit, if they are required at all (unless uncrustify is requiring them, but I don't think it applies here?).
> 
> This was unfortunately called out by PatchCheck.py because of the tab used in the changed location. Will split

If it's called out by PatchCheck, I wouldn't bother to split.

mikebeaton (approved):

kuqin12:
> If it's called out by PatchCheck, I wouldn't bother to split.

Okay, I won't then :) Thanks for the approval~

lgao4:
> > tools_def.template lists VS2015, VS2017, VS2019, VS2022, and VS2026 tool chain. I think they should be verified.
> 
> @lgao4 I tried to build with VS2019, VS2022 and VS2026, these all work. But VS2017 was broken, yet not due to the control flow guard added. We _can_ try to suppress it as this is from a submodule, but it seems odd on why this only fails on VS2017.
> 
> ```
> brotli\c\dec\bit_reader.c(29): warning C4028: formal parameter 1 different from declaration
> brotli\c\dec\bit_reader.c(34): warning C4028: formal parameter 1 different from declaration
> brotli\c\dec\bit_reader.c(59): warning C4028: formal parameter 1 different from declaration
> ```
> 
> I did not attempt to make VS2015 as it was no longer supported in edk2-pytool: https://github.com/tianocore/edk2-pytool-library/blob/ea78c48153a9b1d8784f3d2c3cab92add0f438e6/edk2toollib/windows/locate_tools.py#L51, and does allow me to install when newer versions are available. Should we remove the VS2015 support already?
> 
> But please let me know you would like to proceed with these 2 issues.

I agree to suppress this warning for brotli. But, VS2015 is still in tools_def. I think some developers may still use them. Does VS2015 support control flow guard option?

kuqin12:
> I agree to suppress this warning for brotli. But, VS2015 is still in tools_def. I think some developers may still use them. Does VS2015 support control flow guard option?

I tested this with VS2015 and did not see the build flag itself introduce a failure. Instead, the build appears to be failing due to the same Brotli-related issue. There are also a few integer conversion warnings when targeting IA32. At this point, I haven’t found a non-intrusive way to selectively suppress these warnings based on compiler version.

I also checked internally with others at Microsoft. The general alignment is to support build tools for versions N and N‑1. With that in mind, developers who prefer to remain on older toolchains may want to consider carrying or upstreaming the corresponding build adjustments that best fit their needs.

Please let me know your thoughts.

lgao4:
> > I agree to suppress this warning for brotli. But, VS2015 is still in tools_def. I think some developers may still use them. Does VS2015 support control flow guard option?
> 
> I tested this with VS2015 and did not see the build flag itself introduce a failure. Instead, the build appears to be failing due to the same Brotli-related issue. There are also a few integer conversion warnings when targeting IA32. At this point, I haven’t found a non-intrusive way to selectively suppress these warnings based on compiler version.
> 
> I also checked internally with others at Microsoft. The general alignment is to support build tools for versions N and N‑1. With that in mind, developers who prefer to remain on older toolchains may want to consider carrying or upstreaming the corresponding build adjustments that best fit their needs.
> 
> Please let me know your thoughts.

I may suggest to hold this change until VS2015 and VS2017 are removed from tools_def.txt.

kuqin12:
> I may suggest to hold this change until VS2015 and VS2017 are removed from tools_def.txt.

But that does not resolve the existing build break though. As I mentioned, the control flow guard does not cause the break. Current base tools build is already failing when compiled with VS2015 and VS2017 toolchains.

kuqin12:
A fix for VS2017/VS2015 build break due to brotli was added.

The conversion error I mentioned earlier was a false positive because the compiler I used was not a native x86 compiler.

lgao4:
> A fix for VS2017/VS2015 build break due to brotli was added.
> 
> The conversion error I mentioned earlier was a false positive because the compiler I used was not a native x86 compiler.

Thanks for your fix.

kuqin12:
@lgao4 do you have more feedback on this change? Otherwise would you mind help pushing this through? Thanks in advance!

lgao4 (approved):

PR #12203, opened 2026-02-25
Fix pragma once issue with pcd value init

mdkinney:
# Description

PR #12149 introduced a regression for builds that depends on PcdValueInit for Structured PCDs. The failure is type redefinition. Unfortunately, the same include guard macro name is used by BaseTools application build include files and MdePkg/MdeModulePkg. When they were replaced with #pragma once, both include files can now be included.

Add include guards back to include files that use the same include guard macro in BaseTools/Source/C/Include/Common and MdePkg or MdeModulePkg.

This regression highlights that the PcdValueInit builds need to be reviewed to remove this need for these include guards and allow #pragma once to be sufficient.  A new issue will be opened for this topic.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Builds that uses Structured PCDs and buids and run the PcdValueInit application fails to build PcdValueInit application due to redefinition of types.  After this change is applied, these builds pass as expected.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.inf
+++ b/ArmPkg/Library/ArmStandaloneMmCoreEntryPoint/ArmStandaloneMmCoreEntryPoint.inf
@@ -53,6 +53,7 @@
   gEfiMmPeiMmramMemoryReserveGuid
   gEfiStandaloneMmNonSecureBufferGuid
   gEfiHobListGuid
+  gEfiMmCommunicateHeaderV3Guid
 
 [Protocols]
   gEdkiiPiMmCpuDriverEpProtocolGuid
--- a/ArmVirtPkg/PlatformCI/ReadMe.md
+++ b/ArmVirtPkg/PlatformCI/ReadMe.md
@@ -6,7 +6,7 @@ to use the same Pytools based build infrastructure locally.
 ## Supported Configuration Details
 
 This solution for building and running ArmVirtPkg has been validated with Fedora
-37 Linux and the GCC5 toolchain. The following different firmware builds are
+37 Linux and the GCC toolchain. The following different firmware builds are
 supported.
 
 | Configuration name      | Architecture       | DSC File               | Additional Flags |
--- a/BaseTools/Conf/tools_def.template
+++ b/BaseTools/Conf/tools_def.template
@@ -19,14 +19,15 @@
 #      - Add support for ARM and AARCH64 to CLANGDWARF
 #      - Remove VS2008, VS2010, VS2012, VS2013, CLANG35, CLANG38, EBC
 #      - Add GCC and GCCNOLTO
-#      - Deprecate GCC48, GCC49 and GCC5.
+#      - Deprecate GCC48, GCC49 and GCC5
 # 3.01 - Add toolchain for VS2022
 # 3.02 - Enable stack cookies for IA32, X64, ARM, and AARCH64 builds for GCC and MSVC
 # 3.03 - Drop ARM32 Support
 # 3.04 - Add toolchain for VS2026
 # 3.05 - Add CLANGPDB AARCH64 Support
+# 3.06 - Remove GCC48, GCC49 and GCC5
 #
-#!VERSION=3.05
+#!VERSION=3.06
 
 IDENTIFIER = Default TOOL_CHAIN_CONF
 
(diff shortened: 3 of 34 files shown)


mdkinney:
@makubacki This is a high priority change to address regression.

Not sure why this was not found by CI.  CryptoPkg uses Structured PCDs and should have failed CI.  May be due to reduced CI scope for crypto libs.

mdkinney:
Converted to draft to evaluate some other options.

makubacki:
@mdkinney, is it unfortunate that this dependency exists.

I suspect CryptoPkg did not run into this because the structured PCD only depends on CryptoPkg.dec and the structure only uses simple types like `UINT8` and `UINT32` in [Pcd/PcdCryptoServiceFamilyEnable.h](https://github.com/tianocore/edk2/blob/7746ec4ddeaabc96b1056025a24c69cbd9cd21b0/CryptoPkg/Include/Pcd/PcdCryptoServiceFamilyEnable.h), it does not need to link against the firmware headers to cause conflict.

```ini
  gEfiCryptoPkgTokenSpaceGuid.PcdCryptoServiceFamilyEnable|{0x00}|PCD_CRYPTO_SERVICE_FAMILY_ENABLE|0x00000002 {
    <Packages>
      CryptoPkg/CryptoPkg.dec
    <HeaderFiles>
      Pcd/PcdCryptoServiceFamilyEnable.h
  }
```

---

RedfishPkg.dec has a structured PCD where `<Packages>` includes `MdePkg.dec` and `MdeModulePkg.dec`. 

```ini
  gEfiRedfishPkgTokenSpaceGuid.PcdRedfishRestExServiceDevicePath|{0x0}|REST_EX_SERVICE_DEVICE_PATH_DATA|0x00001000 {
  <HeaderFiles>
    Pcd/RestExServiceDevicePath.h
  <Packages>
    MdePkg/MdePkg.dec
    MdeModulePkg/MdeModulePkg.dec
    RedfishPkg/RedfishPkg.dec
  }
```

My understanding is that this would then bring in `-I MdePkg/Include` to include the copies of the headers outside of BaseTools which could cause the type redefinition.

But I'm not sure this would catch it without the default value being modified in a DSC.

EmulatorPkg.dsc does that:

https://github.com/tianocore/edk2/blob/7746ec4ddeaabc96b1056025a24c69cbd9cd21b0/EmulatorPkg/EmulatorPkg.dsc#L300-L302

But `REDFISH_ENABLE` is set to `FALSE`:

https://github.com/tianocore/edk2/blob/7746ec4ddeaabc96b1056025a24c69cbd9cd21b0/EmulatorPkg/EmulatorPkg.dsc#L43-L47

(and [RedfishDefines.dsc.inc](https://github.com/tianocore/edk2/blob/7746ec4ddeaabc96b1056025a24c69cbd9cd21b0/RedfishPkg/RedfishDefines.dsc.inc) is not being included).

In any case, Structured PCDs (and PcdValueInit) need to be directly tested in CI **and** we need to try to make these less fragile.

ardbiesheuvel:
Surely, the root cause here is that code intended to run on the host is combined with code that is intended to run on the target? I suppose this works fine when building for X64 on X64, but archs that more commonly rely on cross compilation are on shaky ground here.

mdkinney:
> Surely, the root cause here is that code intended to run on the host is combined with code that is intended to run on the target? I suppose this works fine when building for X64 on X64, but archs that more commonly rely on cross compilation are on shaky ground here.

Works for all archs.  The rules for Structured PCDs is that UINTN, INTN, and pointers are not allowed in the structures, so the structure layout is the same for all.  In order for the host PcdValueInit application to work, it needs to know the structure to evaluate and that structure is in the firmware packages.

I agree that this design is not ideal and we need to remove the dependency that exists between BaseTools include files and the rest of the FW Package include files.  I will open an issue for that.  In the mean time, some of these changes are required.

mdkinney:
This PR is ready for review again. Some exploration into additional changes was made that are not required. The scope of this change is for Structured PCDs and issues related to building the PcdValueInit application when Structured PCDs are used.

mdkinney:
@lgao4 Please review

mdkinney:
Pushed changes to address uncrustify issues from adding back include guard #ifndef statements

makubacki:
@mdkinney, what is the simplest way to reproduce the original issue in the edk2 codebase (not a downstream platform)?

mdkinney:
> @mdkinney, what is the simplest way to reproduce the original issue in the edk2 codebase (not a downstream platform)?

Build the CryptoPkg with the `build` command.  Not `stuart_ci_build`.

PR #12205, opened 2026-02-26
OvmfPkg/RiscVVirt: Make SecureBootDefaultKeysInit driver configurable

r1chard-lyu:
# Description
When Secure Boot is enabled, the SecureBootDefaultKeysInit driver is
not always required. There are alternative methods for enrollment,
such as EnrollDefaultKeys.efi or virt-firmware. Therefore, make the
SecureBootDefaultKeysInit driver an optional build component.

Wrap SecureBootDefaultKeysInit.inf with the SECURE_BOOT_DEFAULT_KEYS
condition, allowing builds to optionally configure the driver.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
- build -a RISCV64 -p OvmfPkg/RiscVVirt/RiscVVirtQemu.dsc -b DEBUG -t GCC5 -D SECURE_BOOT_ENABLE
- build -a RISCV64 -p OvmfPkg/RiscVVirt/RiscVVirtQemu.dsc -b DEBUG -t GCC5 -D SECURE_BOOT_ENABLE -D SECURE_BOOT_DEFAULT_KEYS

## Integration Instructions
N/A

### DIFF
--- a/OvmfPkg/OvmfPkgX64.dsc
+++ b/OvmfPkg/OvmfPkgX64.dsc
@@ -447,7 +447,9 @@
   QemuBootOrderLib|OvmfPkg/Library/QemuBootOrderLib/QemuBootOrderLib.inf
   PlatformBootManagerCommonLib|OvmfPkg/Library/PlatformBootManagerCommonLib/PlatformBootManagerCommonLib.inf
   CpuExceptionHandlerLib|UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHandlerLib.inf
-!if $(SMM_REQUIRE) == TRUE
+!if $(STANDALONE_MM_ENABLE) == TRUE
+  LockBoxLib|MdeModulePkg/Library/LockBoxNullLib/LockBoxNullLib.inf
+!elseif $(SMM_REQUIRE) == TRUE
   LockBoxLib|MdeModulePkg/Library/SmmLockBoxLib/SmmLockBoxDxeLib.inf
 !else
   LockBoxLib|OvmfPkg/Library/LockBoxLib/LockBoxDxeLib.inf


OvmfPkg/RiscVVirt/RiscVVirtQemu.dsc
```diff
@@ -35,12 +35,13 @@
...
   #
-  DEFINE TTY_TERMINAL            = FALSE
-  DEFINE SECURE_BOOT_ENABLE      = FALSE
-  DEFINE QEMU_PV_VARS            = FALSE
-  DEFINE TPM2_ENABLE             = FALSE
-  DEFINE TPM2_CONFIG_ENABLE      = FALSE
-  DEFINE DEBUG_ON_SERIAL_PORT    = TRUE
+  DEFINE TTY_TERMINAL             = FALSE
+  DEFINE SECURE_BOOT_ENABLE       = FALSE
+  DEFINE SECURE_BOOT_DEFAULT_KEYS = FALSE
```
tphan-qualcomm, line 40: Adding an option to turn on/off default keys is fine but why leave its value as FALSE? Unless you introduce an alternative method to register default key for this platform, changing its behavior may not a good way.

kraxel: Building with `SECURE_BOOT_DEFAULT_KEYS=TRUE` does simply not work unless you add key databases to your source tree.  Which why I think it makes sense to keep it turned off by default.  

And there are alternatives to enroll keys. see PR description.

r1chard-lyu: Indeed, as described in the PR.

tphan-qualcomm: > Building with `SECURE_BOOT_DEFAULT_KEYS=TRUE` does simply not work unless you add key databases to your source tree. Which why I think it makes sense to keep it turned off by default.
> 
> And there are alternatives to enroll keys. see PR description.

Ok, forgot keys database need to be downloaded and put to source code to build.

PR #12210, opened 2026-02-27
UefiCpuPkg/BaseRiscV64CpuTimerLib: Fix crash during PEI booting

tphan-qualcomm:
# Description

Commit c27b5554 introduced a constructor that calls GetPerformanceCounterProperties() to help initializing mTimeBase early.
However, in the SEC and PEI phases, this constructor may run before the HOB list is available, leading to a crash.
These patches fix the issue by invoking this constructor only in DXE phase.

- [x] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?


Vogtinator:
Confirmed to work here, both with and without PEI. Thanks!

UefiCpuPkg/Library/BaseRiscV64CpuTimerLib/BaseRiscV64CpuTimerSecLib.inf
```diff
@@ -0,0 +1,36 @@
...
+#  Copyright (c) 2022, Ventana Micro Systems Inc. All rights reserved.<BR>
+#  Copyright (C) 2025 Qualcomm Technologies, Inc. All rights reserved.<BR>
+#
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+[Defines]
+  INF_VERSION                    = 0x0001001B
+  BASE_NAME                      = BaseRisV64CpuTimerSecLib
```
Vogtinator:
```suggestion
  BASE_NAME                      = BaseRiscV64CpuTimerSecLib
```

UefiCpuPkg/Library/BaseRiscV64CpuTimerLib/BaseRiscV64CpuTimerSecLib.inf
```diff
@@ -0,0 +1,36 @@
...
+  FILE_GUID                      = 8fe44dd6-7526-42b6-8723-2d47d4bc0596
+  MODULE_TYPE                    = SEC
+  VERSION_STRING                 = 1.0
+  LIBRARY_CLASS                  = TimerLib | SEC PEI_CORE PEIM
+  MODULE_UNI_FILE                = BaseRisV64CpuTimerLib.uni
```
Vogtinator:
```suggestion
  MODULE_UNI_FILE                = BaseRiscV64CpuTimerLib.uni
```

UefiCpuPkg/Library/BaseRiscV64CpuTimerLib/BaseRiscV64CpuTimerSecLib.inf
tphan-qualcomm: Thanks for catching that.

UefiCpuPkg/Library/BaseRiscV64CpuTimerLib/BaseRiscV64CpuTimerSecLib.inf
tphan-qualcomm: Thanks for catching that.

niruiyu (approved):
I understand Hob is not available in SEC, but I do not understand why it could not be available in PEI.
Anyway your patch fixes the issue.

tphan-qualcomm:
> I understand Hob is not available in SEC, but I do not understand why it could not be available in PEI. Anyway your patch fixes the issue.

HOB is available in PEI but not in construction functions as those functions are called at early of PEI_CORE.

UefiCpuPkg/Library/BaseRiscV64CpuTimerLib/BaseRiscV64CpuTimerSecLib.inf
```diff
@@ -0,0 +1,36 @@
...
+#  Copyright (c) 2022, Ventana Micro Systems Inc. All rights reserved.<BR>
+#  Copyright (C) 2025 Qualcomm Technologies, Inc. All rights reserved.<BR>
+#
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+[Defines]
+  INF_VERSION                    = 0x0001001B
+  BASE_NAME                      = BaseRiscV64CpuTimerSecLib
```
vlsunil: Now, this makes it necessary to be moved under the generic CpuTimerLib. BASE cannot be used for SEC/PEI only library.

tphan-qualcomm: @vlsunil  Which CpuTimerLib are you referring to?
If it's UefiCpuPkg/Library/CpuTimerLib, it is still classified as a BASE library. If a BASE library cannot be used, then the "BASE" type should be removed from this library.

vlsunil: Sorry, I missed the response.  Yes, I meant UefiCpuPkg/Library/CpuTimerLib. This directory doesn't have "Base" in its name. So, you can put SEC/PEI specific INF and DXE specific INFs separately inside this directory.

vlsunil (approved):

PR #12213, opened 2026-02-28
NetworkPkg/UefiPxeBcDxe: Add missing Token.Context initialization

makubacki:
# Description

Adds missing `Token.Context = Private` initialization in two MTFTP write-file functions. Without this, the `Context` field passed to the MTFTP `WriteFile` callback was uninitialized, which could lead to a page fault when the callback attempted to access the `Private` driver context. The corresponding read-file functions already set this field correctly.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Test file write to TFTP server with memory protections enabled.

## Integration Instructions

N/A


makubacki:
This got stuck in the mergify queue. I'm going to try rebasing and queuing it again.

PR #12217, opened 2026-03-02
StandaloneMmPkg Core: Treat no DEPEX as TRUE

lzeng14:
Treat no DEPEX as TRUE, then the DEPEX section can be eliminated if a driver can be assured to have no
DEPEX needed.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Declare build rule in fdf to eliminate DEPEX section. 

## Integration Instructions

N/A

### DIFF
--- a/StandaloneMmPkg/Core/Dependency.c
+++ b/StandaloneMmPkg/Core/Dependency.c
@@ -182,12 +182,10 @@ MmIsSchedulable (
 
   if (DriverEntry->Depex == NULL) {
     //
-    // A NULL Depex means that the MM driver is not built correctly.
-    // All MM drivers must have a valid depex expression.
+    // If there is no DEPEX, assume the module can be executed
     //
-    DEBUG ((DEBUG_DISPATCH, "  RESULT = FALSE (Depex is empty)\n"));
-    ASSERT (FALSE);
-    return FALSE;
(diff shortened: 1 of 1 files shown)


niruiyu (approved):
I agree with the change that allows an MM driver without a DEP section.

PEI drivers are also allowed NOT carry a depex section.
DXE and traditional SMM drivers without a DEP section mean the drivers depend on all Architecture Protocols. MM drivers run in a separate env which cannot test if all Architecture Protocols exist.

@ardbiesheuvel @samimujawar @kuqin12 comments?

ardbiesheuvel:
No objections in principle. Doesn't the PI spec have anything to say about this?

niruiyu:
> No objections in principle. Doesn't the PI spec have anything to say about this?

Spec does not require the MM_STANDALONE driver to have the depex section. It only says it should not have more than one depex section.

> 
> IV-9.2 File Type EFI_FV_FILETYPE_MM_STANDALONE
> ......
> • The file must contain no more than one EFI_SECTION_MM_DEPEX section.

ardbiesheuvel (approved):

lzeng14:
There were un-related CI failures, rebased to re-run CI targets.

PR #12218, opened 2026-03-02
Add Asm(Read|Write)FsBase to BaseLib for x86-64.

niruiyu:
# Description

The PR adds Asm(Read|Write)FsBase to BaseLib for x86-64.

- [N] Breaking change?
- [N] Impacts security?
- [N] Includes tests?

## How This Was Tested
Call AsmWriteFsBase() and use AsmReadFsBase() to check if the value is written to FSBASE.

## Integration Instructions

n/a

### DIFF
--- a/MdePkg/Library/BaseLib/X64/WriteFsBase.nasm
+++ b/MdePkg/Library/BaseLib/X64/WriteFsBase.nasm
@@ -1,0 +1,31 @@
+;------------------------------------------------------------------------------
+;
+; Copyright (c) 2026, Intel Corporation. All rights reserved.<BR>
+; SPDX-License-Identifier: BSD-2-Clause-Patent
+;
+; Module Name:
+;
+;   WriteFsBase.nasm
+;
+; Abstract:
+;
+;   InternalX86WriteFsBase function
+;
+; Notes:
+;
+;------------------------------------------------------------------------------
+
+    DEFAULT REL
+    SECTION .text
+
+;------------------------------------------------------------------------------
+; VOID
+; EFIAPI
+; InternalX86WriteFsBase (
+;   UINT64  FsBase
+;   );
+;------------------------------------------------------------------------------
+global ASM_PFX(InternalX86WriteFsBase)
+ASM_PFX(InternalX86WriteFsBase):
+    wrfsbase rcx
+    ret
--- a/MdePkg/Include/Library/BaseLib.h
+++ b/MdePkg/Include/Library/BaseLib.h
@@ -5491,6 +5491,42 @@ AsmRmpAdjust (
   IN      UINT64  Rdx
   );
 
+/**
+  Reads the current value of the FS segment base address.
+
+  Reads and returns the current value of the FS segment base address using
+  the RDFSBASE instruction. This function is only available on X64.
+
+  Note: The function requires that CPUID.(EAX=7,ECX=0):EBX.FSGSBASE=1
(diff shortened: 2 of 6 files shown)


MdePkg/Library/BaseLib/X64/WriteFsBase.nasm
```diff
@@ -0,0 +1,32 @@
...
+; EFIAPI
+; AsmWriteFsBase (
+;   UINT64  FsBase
+;   );
+;------------------------------------------------------------------------------
+global ASM_PFX(AsmWriteFsBase)
+ASM_PFX(AsmWriteFsBase):
+    wrfsbase rcx
+    ret
+
```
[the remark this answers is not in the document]
niruiyu: @mdkinney rdfsbase instruction was introduced since IvyBridge, more than 10 years ago. I prefer to not check the availability of the feature for simplicity. What's your opinion?

MdePkg/Library/BaseLib/X64/WriteFsBase.nasm
```diff
@@ -0,0 +1,32 @@
...
+; Notes:
+;
+;------------------------------------------------------------------------------
+
+    DEFAULT REL
+    SECTION .text
+
+;------------------------------------------------------------------------------
+; VOID
+; EFIAPI
```
mdkinney, line 23: Write APIs usually return the value written.  Is there a reason this one does not?

MdePkg/Library/BaseLib/X64/WriteFsBase.nasm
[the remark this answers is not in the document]
mdkinney: Seems safe, but I think the other question is where these APIs will be called from.  If we can guarantee they will only be called by platforms that support these instructions, then avoiding the CPUID check seems reasonable.  If these APIs will be used by a generic module that could be integrated into any platform, then the CPUID check may be required.

MdePkg/Library/BaseLib/X64/WriteFsBase.nasm
niruiyu: No reason. I will change.

niruiyu:
@mdkinney I've addressed 2 comments:
1. Return FsBase from AsmWriteFsBase()
2. Use ASSERT() to verify that the CPU supports and enables FS/GS instructions.

MdePkg/Library/BaseLib/X64/WriteFsBase.nasm
```diff
@@ -0,0 +1,38 @@
...
+; UINT64
+; EFIAPI
+; AsmWriteFsBase (
+;   UINT64  FsBase
+;   );
+;------------------------------------------------------------------------------
+global ASM_PFX(AsmWriteFsBase)
+ASM_PFX(AsmWriteFsBase):
+    sub      rsp, 8   ; Align stack to 16-byte boundary for x64 ABI
+    call     ASM_PFX(InternalAssertFsGsBaseEnabled)
```
mdkinney: Prefer to not call C function from ASM function.

Can make ASM function an internal function name that only performs the CPU instruction that is not available from C code and make new BaseLib API in C code with the FsGs check and call to internal ASM function.

niruiyu: Pushed new changes that follow your suggestion.

PR #12219, opened 2026-03-02
BaseTools: Reject the null body of commit message in PatchCheck script

EricGao2015:
# Description

Current PatchCheck script allows null body of commit message. The patch adds a check for such case. If the body of commit message is null, an error will be throwed.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Create a commit with subject line, Signed-off-by line and null body message:
```
Author: Qihang Gao <gaoqihang@loongson.cn>
Date:   Mon Mar 2 16:31:28 2026 +0800

    TEST NULL COMMIT MESSAGE
    
    Signed-off-by: Qihang Gao <gaoqihang@loongson.cn>

diff --git a/testfile b/testfile
new file mode 100644
index 00000000..e69de29b
```

Run Command:`python3 BaseTools/Scripts/PatchCheck.py 0001-TEST-NULL-COMMIT-MESSAGE.patch`, the output is:
```
Checking patch file: 0001-TEST-NULL-COMMIT-MESSAGE.patch
TEST NULL COMMIT MESSAGE
 * The commit body lines are not found
https://www.tianocore.org/tianocore-wiki.github.io/development/contribution-guides/commit_message_format.html
The code passed all checks.
```

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Scripts/PatchCheck.py
+++ b/BaseTools/Scripts/PatchCheck.py
@@ -4,6 +4,7 @@
 #  Copyright (c) 2015 - 2021, Intel Corporation. All rights reserved.<BR>
 #  Copyright (C) 2020, Red Hat, Inc.<BR>
 #  Copyright (c) 2020 - 2023, Arm Limited. All rights reserved.<BR>
+#  Copyright (c) 2026, Loongson Technology Corporation Limited. All rights reserved.<BR>
 #
 #  SPDX-License-Identifier: BSD-2-Clause-Patent
 #
@@ -304,6 +305,7 @@ class CommitMessageCheck:
             self.error('Second line of commit message should be ' +
                        'empty.')
 
+        found_body_line = False
         for i in range(2, count):
             if (len(lines[i]) >= 76 and
                 len(lines[i].split()) > 1 and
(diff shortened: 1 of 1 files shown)


mikebeaton:
If I've understood correctly, this would automatically rule out commit messages with a subject line and a Signed-off-by line, but nothing else in the body? In my PR https://github.com/tianocore/edk2/pull/11886/commits (just for example) some of my commits are like that, and no one has raised this as being wrong - and I am not sure that it _should_ always be marked as wrong.

Certainly, it would only be correct to have an empty body (if at all) in the case where the commit is a simple change which can be simply described in just the subject line. I guess an approving reviewer should always be checking whether the commit messages are appropriate. It would not make sense to force just any old commit message; even with this change, reviewers would of course still need to check the commit message body, as now.

I would have said allowing (or disallowing) such an 'empty body' commit message is a subset of the existing process of reviewing commit messages; with 'allow' being a reasonable response, in some cases.

YangGangUEFI:
I don't think it's necessary to enforce a check on the commit message body, as some commits only require a brief summary of the changes.

On the other hand, there are still some keywords that could potentially bypass this check:
```
Co-authored-by

Message-Id

Contributed-under

Continuous-integration-options:

Change-Id
```

PR #12226, opened 2026-03-02
UefiPayloadPkg: add optional 2x GOP scaling

Sean-StarLabs:
When coreboot passes a high-resolution framebuffer (FSP GOP /
RUN_FSP_GOP), UefiPayloadPkg exposes a single GOP mode at the physical
resolution. Some payload UIs assume 1:1 pixels, which makes them tiny on
HiDPI panels.

Add an optional 2x scaling mode to GraphicsOutputDxe. When enabled and
the physical resolution meets or exceeds both thresholds, and is even in
each dimension, expose an additional software-only mode with width and
height divided by 2.

Treat the thresholds as inclusive (`>=`) so 1920x1080 triggers with the
default values.

BlSupportDxe now sets the video resolution PCDs to the scaled logical
resolution when HiDPI is enabled so GraphicsConsoleDxe prefers the HiDPI
mode for setup menus.

The HiDPI mode reports `PixelFormat = PixelBltOnly` and
`FrameBufferBase = 0`, and scales BLT operations to the physical
framebuffer. Keep that mode active until `ExitBootServices`, then switch
back to the physical framebuffer mode and restore the physical video
resolution PCDs so direct framebuffer consumers continue to work and
late DXE clears cover the full screen.

Enable with the build define `FSP_GOP_BASIC_HIDPI_SUPPORT` (sets
`PcdFspGopBasicHiDpiSupport` to `TRUE`). The threshold is controlled by:
- `PcdFspGopBasicHiDpiScaleThresholdHorizontal` (default `1920`)
- `PcdFspGopBasicHiDpiScaleThresholdVertical` (default `1080`)


UefiPayloadPkg/GraphicsOutputDxe/GraphicsOutput.c
nate-desimone, line 703: This is a violation of the UEFI specification.

Per spec, it is required that the framebuffer provided in `EFI_GRAPHICS_OUTPUT_PROTOCOL.Mode.FrameBufferBase` have a data format that matches the description of the frame buffer in `EFI_GRAPHICS_OUTPUT_PROTOCOL.Mode.Info`.

This is particularly relevant given that most PC compatible GPUs set `EFI_GRAPHICS_OUTPUT_PROTOCOL.Mode.Info.PixelFormat = PixelBlueGreenRedReserved8BitPerColor`. That means that it is completely valid for any UEFI compatible driver, application, or OS to directly write to the framebuffer given by `EFI_GRAPHICS_OUTPUT_PROTOCOL.Mode.FrameBufferBase` and bypass the hook you have inserted on `EFI_GRAPHICS_OUTPUT_PROTOCOL.Blt()`. I have seen multiple examples of this. Perhaps the most important being the `efifb` kernel driver: https://docs.kernel.org/fb/efifb.html

Because direct writes to the framebuffer are allowed by spec, changing the values of `EFI_GRAPHICS_OUTPUT_PROTOCOL.Mode.Info.HorizontalResolution` and `EFI_GRAPHICS_OUTPUT_PROTOCOL.Mode.Info.VerticalResolution` will result in the math done to calculate the address of a pixel to be wrong.

The correct way to implement a feature like this would be to expose the 2x scaled graphics mode as a separate mode that can be chosen using `EFI_GRAPHICS_OUTPUT_PROTOCOL.SetMode()`. That separate mode would set `EFI_GRAPHICS_OUTPUT_PROTOCOL.Mode.Info.PixelFormat = PixelBltOnly` and `EFI_GRAPHICS_OUTPUT_PROTOCOL.Mode.FrameBufferBase == NULL`, indicating that this is a software-only graphics mode that does not support a direct framebuffer. You would want to implement some platform BDS logic to set the mode back to the 1:1 mode (so that `EFI_GRAPHICS_OUTPUT_PROTOCOL.Mode.FrameBufferBase != NULL`) before launching the OS so that drivers like `efifb` continue to work.

mdkinney: Another option to handle high resolution displays is to scale the existing fonts or to provide higher resolution fonts.

PR #12227, opened 2026-03-02
UefiCpuPkg: Add an option to disable MTRR programming

Sean-StarLabs:
UefiCpuPkg: add PcdCpuDisableMtrrProgramming
Add PcdCpuDisableMtrrProgramming to allow platforms to disable MTRR
programming when the bootloader already configured MTRRs.

When enabled, skip call sites that program MTRRs (MtrrSetAllMtrrs() and
MtrrSetMemoryAttribute()) in UefiCpuPkg modules.

### DIFF
--- a/UefiPayloadPkg/UefiPayloadPkg.dsc
+++ b/UefiPayloadPkg/UefiPayloadPkg.dsc
@@ -654,6 +654,10 @@
 !endif
 !endif
 
+!if $(BOOTLOADER) == "COREBOOT"
+  # Disable MTRR programming (already configured by coreboot)
+  gUefiCpuPkgTokenSpaceGuid.PcdCpuDisableMtrrProgramming|TRUE
+!endif
 
 !if $(SECURE_BOOT_ENABLE) == TRUE
   # Override the default values from SecurityPkg to ensure images from all sources are verified in secure boot


benjamindoron:
I don't see why the bootloader's MTRRs are any more 'objective' than the payload's. coreboot has handed over control of the platform, and with it, knowledge about the memory map (ideally). This should be enough for EDK2 to calculate good MTRRs too.

Does this address some bug? If not, I say we abandon this. If this can enhance performance, then I think that the solution is either improve the inputs to MtrrLib (perhaps, giving UefiPayload more information about the memory it has), or to improve the algorithm in MtrrLib.

Sean-StarLabs:
>  Does this address some bug?

Yeah, makes every platform before MTL actually boot. Pretty sure every single fork of edk2 used with coreboot has Patrick original patch so it actually works (IIRC, Kconfig has warnings about using `upstream/master` as it wont work)

benjamindoron:
> > Does this address some bug?
> 
> Yeah, makes every platform before MTL actually boot. Pretty sure every single fork of edk2 used with coreboot has Patrick original patch so it actually works (IIRC, Kconfig has warnings about using `upstream/master` as it wont work)

I remember hearing that, and taking it into my fork for that reason too, but I eventually dropped it (and everything was fine). I do most things on https://github.com/benjamindoron/edk2/tree/dev/uefipayload, and while my local copy will diverge from that until I can update it, or until it makes sense to push something out publicly, it generally reflects what I use and test. And I've checked my fork (just in case I forgot, since I haven't thought about this issue for a long while) and this patch isn't there.

Can we debug this as a specific issue to solve? Is there a log for this?

MrChromebox:
> I don't see why the bootloader's MTRRs are any more 'objective' than the payload's. coreboot has handed over control of the platform, and with it, knowledge about the memory map (ideally). This should be enough for EDK2 to calculate good MTRRs too.

counterpoint: edk2 when used as a coreboot payload shouldn't be redoing things coreboot has already done for no good reason

> Can we debug this as a specific issue to solve? Is there a log for this?

I'll test a see if the issue is still reproducible, and if so get a log

Sean-StarLabs:
Ben was right, not needed anymore

MrChromebox:
this is still needed for some older platforms like Intel Braswell, so we should figure out what edk2 is doing there that's breaking things

PR #12229, opened 2026-03-02
UefiPayloadPkg/CbParseLib: Use CRC32 function compatible with coreboot

Sean-StarLabs:
EDK2's CalculateCrc32() does not match the CRC32 implementation used by
coreboot (non-reflected, init 0, poly 0x04C11DB7, no final xor). As a
result, CRC32 fields produced by coreboot cannot be validated with the
EDK2 helper.

Add CbCalculateCrc32() to CbParseLib to compute CRC32 values compatible
with coreboot.

### DIFF
--- a/UefiPayloadPkg/Library/CbParseLib/CbParseLib.c
+++ b/UefiPayloadPkg/Library/CbParseLib/CbParseLib.c
@@ -76,6 +76,43 @@ CbCheckSum16 (
 }
 
 /**
+  Coreboot implements a CRC32 checksum which differs from the one in EDK2. This function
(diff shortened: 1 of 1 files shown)


benjamindoron:
Where is this going to be used? I believe that the answer is "CFR," but then I'd just have this commit in a PR for that. (CFR is our system to define a setup menu in coreboot and then display it in UEFI, for those who don't know. You can find the driver on the 9elements EDK2 fork, my fork, MrChromebox's, Sean's and I don't know who else's).

Sean-StarLabs:
Yuppo - will close :)

PR #12231, opened 2026-03-02
UefiPayloadPkg/GraphicsOutputDxe: support framebuffer BAR offset

Sean-StarLabs:
Some platforms place the linear framebuffer at an offset within the PCI
BAR (for example, Meteor Lake can use BAR2 + 0x800000). The existing
GraphicsOutputDxe validation requires the framebuffer base to match the
BAR base exactly, which rejects these configurations.

Relax the check to accept a framebuffer base within the BAR window, and
use the provided framebuffer base address when it falls within that
range.


mikebeaton (approved):
This looks like a good commit that should be merged,

PR #12232, opened 2026-03-02
MdeModulePkg: Improve USB error handling for faulting devices

Sean-StarLabs:
Some faulty USB devices can cause an xHCI endpoint to halt. In this
case XhciDxe may keep polling until timeout and attempt recovery, which
adds delays during boot.

Detect halted/non-running endpoints via the xHCI output context and
return EFI_DEVICE_ERROR with EFI_USB_ERR_STALL instead of timing out.
Also propagate dequeue failures from XhcTransfer().

Stop retrying USB mass storage boot commands when the transport returns
EFI_DEVICE_ERROR.


mikebeaton (approved):
I believe this looks like a good commit which should be merged.

PR #12234, opened 2026-03-03
CryptoPkg: Fix leaks and failure-path mutation of RSA-owned values

szsam:
# Description

RsaSetKey passed RSA-internal BIGNUMs (returned by RSA_get0_key /
RSA_get0_factors / RSA_get0_crt_params as const) directly to BN_bin2bn,
which mutates its destination in place. This violates the OpenSSL API
contract and can leave RsaContext in a partially modified state on
failure paths: BN_bin2bn has already overwritten one of
n / e / d / p / q / dp / dq / qInv, the function then returns FALSE
on a subsequent BN_dup or RSA_set0_* failure, and the caller has no
indication that the RSA object was silently changed. The same paths
also leaked temporary BIGNUMs allocated by BN_bin2bn or BN_new.

This patch splits RsaSetKey into a small input-validating dispatcher
plus three static helpers (RsaSetKeyNED, RsaSetKeyFactors,
RsaSetKeyCrtParams), one per RSA_set0_* setter. Each helper:

  - Allocates a fresh BIGNUM for the slot being set via
    BN_bin2bn(BigNumber, BnSize, NULL); the NULL destination forces
    BN_bin2bn to allocate so no RSA-owned BIGNUM is mutated.
  - Reads the current RSA state via RSA_get0_* purely for inspection.
  - For slots not being set, passes NULL when RSA already has a value
    (preserves it) or supplies an empty BN_new() placeholder when
    RSA's slot is still NULL.
  - Atomically installs via RSA_set0_*, which takes ownership of every
    non-NULL argument on success and of none on failure.
  - Routes all exits through a single label that frees any locally
    held BIGNUM. Memory leaks in the early-return paths are fixed as
    a consequence.

The previous BN_dup calls are no longer needed and are removed.

~~Before this patch, RsaSetKey freed temporary BN objects only on successful paths. Error handling used early return FALSE exits that bypassed cleanup, so allocated BN objects were leaked on those paths. In addition, BN_dup results were not checked for NULL.~~

~~This patch routes all exits through a shared Exit cleanup path, tracks temporary BN_dup outputs, checks duplication results before RSA_set0_* calls, and releases temporary objects unless ownership has been transferred to the RSA context.~~

- [ ] Breaking change?
- [x] Impacts security?
  - **Security** - Fixed memory leaks.
- [ ] Includes tests?

## How This Was Tested

`stuart_ci_build -c .pytool/CISettings.py -a X64 TOOL_CHAIN_TAG=GCC -p CryptoPkg -t
  NOOPT`

## Integration Instructions

N/A

### DIFF
--- a/CryptoPkg/Library/BaseCryptLib/Pk/CryptRsaBasic.c
+++ b/CryptoPkg/Library/BaseCryptLib/Pk/CryptRsaBasic.c
@@ -57,6 +57,311 @@ RsaFree (
 }
 
 /**
+  Install one of the n, e, d components of an RSA key.
+
+  @param[in, out]  RsaKey     OpenSSL RSA object.
+  @param[in]       KeyTag     One of RsaKeyN, RsaKeyE, RsaKeyD.
+  @param[in]       BigNumber  Octet integer buffer to install.
+  @param[in]       BnSize     Size of BigNumber in bytes.
+
+  @retval  TRUE   Component installed.
+  @retval  FALSE  Allocation or RSA_set0_key failed; RsaKey is unchanged.
+
+**/
+STATIC
+BOOLEAN
+RsaSetKeyNED (
+  IN OUT  RSA          *RsaKey,
+  IN      RSA_KEY_TAG  KeyTag,
+  IN      CONST UINT8  *BigNumber,
+  IN      UINTN        BnSize
+  )
+{
+  BIGNUM        *PassN;
+  BIGNUM        *PassE;
+  BIGNUM        *PassD;
+  const BIGNUM  *CurN;
+  const BIGNUM  *CurE;
+  const BIGNUM  *CurD;
+  BOOLEAN       Result;
+
+  PassN  = NULL;
+  PassE  = NULL;
+  PassD  = NULL;
+  Result = FALSE;
+
+  //
+  // Allocate the slot being set. BN_bin2bn allocates a fresh BIGNUM when
+  // its destination argument is NULL, so no RSA-owned BIGNUM is touched.
+  //
+  switch (KeyTag) {
+    case RsaKeyN:
+      PassN = BN_bin2bn (BigNumber, (UINT32)BnSize, NULL);
+      if (PassN == NULL) {
+        goto Exit;
+      }
+
+      break;
+    case RsaKeyE:
+      PassE = BN_bin2bn (BigNumber, (UINT32)BnSize, NULL);
+      if (PassE == NULL) {
+        goto Exit;
+      }
+
+      break;
+    case RsaKeyD:
+      PassD = BN_bin2bn (BigNumber, (UINT32)BnSize, NULL);
+      if (PassD == NULL) {
+        goto Exit;
+      }
+
+      break;
+    default:
+      goto Exit;
+  }
+
+  //
+  // Snapshot the current RSA state. Per the OpenSSL API contract the
+  // BIGNUMs that CurN, CurE, CurD point to are owned by RsaKey and must
+  // not be modified by the caller.
+  //
+  RSA_get0_key (RsaKey, &CurN, &CurE, &CurD);
+
+  //
+  // For each slot we are not setting:
+  //   - leave PassX as NULL when RSA already has a value there
+  //     (RSA_set0_key treats NULL as "preserve the current value in
+  //     RsaKey"), or
+  //   - install a fresh BIGNUM via BN_new() (which returns a BIGNUM with
+  //     value 0) when RSA has NULL there.
+  //
+  if ((PassN == NULL) && (CurN == NULL)) {
+    PassN = BN_new ();
+    if (PassN == NULL) {
+      goto Exit;
+    }
+  }
+
+  if ((PassE == NULL) && (CurE == NULL)) {
+    PassE = BN_new ();
+    if (PassE == NULL) {
+      goto Exit;
+    }
+  }
+
+  if ((PassD == NULL) && (CurD == NULL)) {
+    PassD = BN_new ();
+    if (PassD == NULL) {
+      goto Exit;
+    }
+  }
+
+  //
+  // Atomic install. On success ownership of every non-NULL argument
+  // transfers to RsaKey; on failure no ownership transfers and the
+  // RSA object is left unchanged.
+  //
+  if (RSA_set0_key (RsaKey, PassN, PassE, PassD) == 0) {
+    goto Exit;
+  }
+
+  //
+  // Ownership transferred to RsaKey. Null our handles so cleanup is a no-op.
+  //
+  PassN  = NULL;
+  PassE  = NULL;
+  PassD  = NULL;
+  Result = TRUE;
+
+Exit:
+  BN_free (PassN);
+  BN_free (PassE);
+  BN_clear_free (PassD);
+  return Result;
+}
+
+/**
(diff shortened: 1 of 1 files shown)


szsam:
@Flickdm @jyao1 @liyi77 @baranee Please review this PR. This is a follow-up of #11105 and fixes memory leaks in error-handling paths.

Flickdm:
So in reviewing the full code there are a few places where I suspect BN_bin2bn would modify context even on failure which is problematic but not related to your proposed change. I'll review further and if a problem open an issue.

As it stands, I think your PR is addressing memory leakage and is a good change.

Flickdm (approved):

szsam:
> So in reviewing the full code there are a few places where I suspect BN_bin2bn would modify context even on failure

I can confirm your suspicion.

In cases of KeyTag belongs to `{KeyN, KeyE, KeyD}`, after a successful BN_bin2bn (line 165-177), either
    - any of the three BN_dup calls returns NULL (lines 183-188), or
    - RSA_set0_key (line 190) returns 0

... we goto Exit and return FALSE. The Exit block frees `AllocatedBn[*]` and `DuplicatedBn[*]`, but
RsaContext's internal states have silently been overwritten with BigNumber.

Similar issues exist in other `KeyTag` values.

Needs a follow-up patch to fix the "modification on failure" bugs.

We should not pass RSA-owned BIGNUM to BN_bin2bn. Instead, BN_bin2bn should allocate its own memory for BIGNUM.

```c
// BN_bin2bn allocates when passing NULL to the third arg
NewBn = BN_bin2bn (BigNumber, (UINT32)BnSize, NULL);
if (NewBn == NULL) goto Exit;
```


Pseudo-code of fixed code:

```c
  case RsaKeyN:
  case RsaKeyE:
  case RsaKeyD: {
    BIGNUM        *PassN = NULL;   // what we hand to RSA_set0_key for n
    BIGNUM        *PassE = NULL;   // ...for e
    BIGNUM        *PassD = NULL;   // ...for d
    const BIGNUM  *CurN;
    const BIGNUM  *CurE;
    const BIGNUM  *CurD;

    // (1) Allocate the new BIGNUM.
    switch (KeyTag) {
      case RsaKeyN:
        PassN = BN_bin2bn (BigNumber, (UINT32)BnSize, NULL);
        if (PassN == NULL) goto Exit;
        break;
      case RsaKeyE:
        PassE = BN_bin2bn (BigNumber, (UINT32)BnSize, NULL);
        if (PassE == NULL) goto Exit;
        break;
      case RsaKeyD:
        PassD = BN_bin2bn (BigNumber, (UINT32)BnSize, NULL);
        if (PassD == NULL) goto Exit;
        break;
    }

    // (2) Snapshot current RSA state (read-only).
    RSA_get0_key (RsaKey, &CurN, &CurE, &CurD);

    // (3) For each slot we did NOT set in step (1):
    //       - If RSA already has a value there, leave PassX == NULL
    //         (RSA_set0_key treats NULL as "preserve current value").
    //       - If RSA has NULL there, allocate an empty BIGNUM so RSA_set0_key
    //         has something to install.
    if (PassN == NULL && CurN == NULL) {
      PassN = BN_new ();
      if (PassN == NULL) goto Exit;
    }
    if (PassE == NULL && CurE == NULL) {
      PassE = BN_new ();
      if (PassE == NULL) goto Exit;
    }
    if (PassD == NULL && CurD == NULL) {
      PassD = BN_new ();
      if (PassD == NULL) goto Exit;
    }

    // (4) Atomic install. On success RSA_set0_key takes ownership of every
    //     non-NULL argument; on failure it takes ownership of none.
    if (RSA_set0_key (RsaKey, PassN, PassE, PassD) == 0) goto Exit;

    // (5) Ownership transferred — null our handles so cleanup is a no-op.
    PassN  = NULL;
    PassE  = NULL;
    PassD  = NULL;
    Result = TRUE;
    break;
  }
```
  Exit cleanup for N/E/D
```c
  Exit:
    switch (KeyTag) {
      case RsaKeyN:
      case RsaKeyE:
      case RsaKeyD:
        BN_free (PassN);
        BN_free (PassE);
        BN_clear_free (PassD);
        break;
      ...
    }
```

Flickdm:
> > So in reviewing the full code there are a few places where I suspect BN_bin2bn would modify context even on failure
> 
> I can confirm your suspicion.
> 

Yeah I agree with your assessment - Thanks for following up here! It had completely slipped my mind. Would you like to create an issue or would you prefer if I do?

Flickdm:
@jyao1 @liyi77 can we get a second review here?

szsam:
> > > So in reviewing the full code there are a few places where I suspect BN_bin2bn would modify context even on failure
> > 
> > 
> > I can confirm your suspicion.
> 
> Yeah I agree with your assessment - Thanks for following up here! It had completely slipped my mind. Would you like to create an issue or would you prefer if I do?

Happy to take this on. I think the cleaner approach is to roll it into this PR and reframe the change as a broader refactor of `RsaSetKey` that addresses all three concerns: memory leaks, `RsaContext` mutation, and removing `BN_dup` in favor of passing `NULL` to `RSA_set0_key` / `RSA_set0_factors` / `RSA_set0_crt_params` for key components that should be left untouched.

That should make fixing memory leaks much simpler, since we won’t need to manage ownership of `BN_dup` outputs. The current PR adds a `DuplicatedBn[]` array to track those outputs, but if we take the refactor approach, the `BN_dup`s will be removed entirely.

Flickdm:
> > > > So in reviewing the full code there are a few places where I suspect BN_bin2bn would modify context even on failure
> > > 
> > > 
> > > I can confirm your suspicion.
> > 
> > 
> > Yeah I agree with your assessment - Thanks for following up here! It had completely slipped my mind. Would you like to create an issue or would you prefer if I do?
> 
> Happy to take this on. I think the cleaner approach is to roll it into this PR and reframe the change as a broader refactor of `RsaSetKey` that addresses all three concerns: memory leaks, `RsaContext` mutation, and removing `BN_dup` in favor of passing `NULL` to `RSA_set0_key` / `RSA_set0_factors` / `RSA_set0_crt_params` for key components that should be left untouched.
> 
> That should make fixing memory leaks much simpler, since we won’t need to manage ownership of `BN_dup` outputs. The current PR adds a `DuplicatedBn[]` array to track those outputs, but if we take the refactor approach, the `BN_dup`s will be removed entirely.

Works for me! feel free to ping me when you're ready for me to look again!

liyi77:
That is a great proposal for fixing the memory leak issue, please @ me also when PR ready to review.

TBH EDK2 RSA function is outdated; it relies heavily on APIs that were deprecated in openssl 3.0. We should better to transfer to EVP APIs in future.

Flickdm:
> That is a great proposal for fixing the memory leak issue, please @ me also when PR ready to review.
> 
> TBH EDK2 RSA function is outdated; it relies heavily on APIs that were deprecated in openssl 3.0. We should better to transfer to EVP APIs in future.

Agreed - I actually have some local changes to move entirely to EVP and remove the deprecation define if you want me to propose that - I can wait till this is in and then resolve the differences. Thats one of the many changes I've been experiementing with.

szsam:
I updated the PR to fix memory leaks and mutation of RSA-owned BIGNUMs in failure paths. Pls review. @Flickdm @liyi77

liyi77 (approved):
LGTM

Flickdm (approved):

PR #12235, opened 2026-03-03
Ffa vm id 1

shenghuliu:
# Description

According to section 4.3 of DEN0077A_Firmware_Framework_Arm_A-profile_1.3_ALP2, EDK2 can operate as either a non-secure physical or non-secure virtual FF-A instance. However, the current implementation only supports running EDK2 as an FF-A physical instance; support for a virtual instance is not available.

This PR introduces the following changes:

-    Defines a PCD to configure whether EDK2 runs as a physical or virtual FF-A instance, defaulting to physical.
-    Adds a helper function, GetFfaVmIdVal.
-    Updates required FF-A commands to retrieve the VM ID using GetFfaVmIdVal.

- [ ] Breaking change? 
- [ ] Impacts security?
- [ ] Includes tests?

# How This Was Tested

This change was tested on an internal device that complies with the Hafnium implementation.

# Integration Instructions

For real-world use cases:
-    If EDK2 runs as an FF-A virtual instance, configure the PCD to FALSE.
-    If EDK2 runs as an FF-A physical instance, no configuration changes are required.


MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -70,6 +72,37 @@ EfiStatusToFfaStatus (
...
+
+  @param [in] PartId   Partition ID
+  @param [in] Shift    Per Specs, TRUE will left shift 16 bits, FALSE will not.
+
+  @retval              Final VM Id with UINTN size.
+
+**/
+UINTN
+EFIAPI
+GetFfaVmIdVal (
```
LeviYeoReum, line 89: I'm totally get confused for this.

First, why does this require?

When you see the, 13.10 FFA_ID_GET
```
Returns 16-bit ID of calling FF-A component.
– ID value 0 must be returned at the Non-secure physical FF-A instance (see Chapter 6 Identification and
Discovery).
```

IOW, regardless of virtual instance or physical, PartId should be returned properly by SPMC.
SPMC should return "0" for Non-secure physical FF-A instance.
EDK2 doesn't need to specifiy whether it is a "Physical" or "Virtual".

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -329,7 +362,7 @@ ArmFfaLibRxRelease (
   ZeroMem (&FfaArgs, sizeof (ARM_FFA_ARGS));
 
   FfaArgs.Arg0 = ARM_FID_FFA_RX_RELEASE;
-  FfaArgs.Arg1 = PartId;
+  FfaArgs.Arg1 = GetFfaVmIdVal (PartId, FALSE);
```
kuqin12, line 359: Is this really right?

Spec 1.3 ALP4 specifically says this is "only valid at the Non-secure physical FF-A instance".

Image

shenghuliu: "MBZ otherwise."  So if it is not physical instance, w1 must be zero.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
shenghuliu: image

For Hyp and VM(EDK2) case which is non-secure virtual instance, has nothing to do with physical case which is between Hyp and SPMD.  Physical case must return 0, that doesn't address virtual instance case.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
kuqin12: Thanks. But it only explains the outer if/else condition where the PCD is checked.

Why is the shift flag needed at all? IIUC, 1. this is an internal interface, 2. only called here once, 3. the shift is only hardcoded to `FALSE`. This input seems redundant and does not comply with the spec snippet above.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: > For Hyp and VM(EDK2) case which is non-secure virtual instance, has nothing to do with physical case which is between Hyp and SPMD. Physical case must return 0, that doesn't address virtual instance case.

What I meant was about the behavior of FFA_GET_PART_ID in EDK2, regardless of whether the case is virtual or physical.

First, in the physical case, 0 must be returned by SPMD.
In the virtual machine case (whether the conduit is HVC or SMC), the appropriate VM_ID should be returned as the PartID by HYP (might SMC redirect to HYP), and that value would then be stored and used as the PartID (it should get return "VM id 1" in your case")

I don’t see a reason why EDK2 should have to determine at build time—using a PCD—whether it is running in a virtual or physical environment.

shenghuliu: In virtual case, return "VM id 1" is correct. When composing RX_RELEASE and RXTX_UNMAP commands in virtual case, what VM id value should be filled in w1?  Per Specs, w1 should be 0, because it is virtual instance case, even though "VM id 1" is returned.

How is it known that EDK2 run as a virtual instance or not? Based on Specs definition, it can be known.  How to pass this virtual info to code? PCD flag.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
shenghuliu: RX_RELEASE and RXTX_UNMAP both require VM ID in W1, but one is in lower 16-bit, another one is in upper 16-bit, that is where the shift flag for.  

"does not comply with the spec snippet above." i don't get it, would you clarify more ? thanks.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: Ah. I missed FFA_RXTX_UNMAP's w1 description.
Hmm But, let me check about this. I'm really confused for this description.
Since other components (e.x) linux-kernel ff-a driver 
seem to have the same problem...

Let's check about this... But I'm really strange why caller or at build time should know
I'm physical or I'm virtual...

LeviYeoReum: BTW question is there any case require "FFA_RXTX_UNMAP" of other "PartId" in edk2?
Currently, "ArmFfaRxTxUnmap()" is called for myself.

IOW, If edk2 is physical it always 0 and it always uses 0. It wouldn't be a problem.
In case of virtual, by checking PART_ID > 0, Could we remove PCD for this and this function?

LeviYeoReum: When I check about this, This VM_ID is for the usage of "HYPERVIOR" to request SPMC.
since hypervior already know its where it came from the VM, edk2 doesn't need to set anything but 0.
So edk2 case, regardless of physical and virtual just simply sets "0"  since pyhsical is always 0.

Therefore, we don't need to any PCD or any new helper function.
It's enough to set w1 as 0 in FFA_RXTX_UNMAP.

shenghuliu: In existing EDK2 release, our test shows:

-  non-zero PartId is returned by calling FFA_GET_PART_ID,  
-  this non-zero PartId is set to W1, and
-  the command fails with invalid parameters error.

I agree this Specs is a little tricky, but simply set "0" is not a good practice. let's follow the Specs and use helper function command by command. 

To maintain backward compatibility, introducing a PCD is needed. Because we don't know how the clients use this EDK2 and whether they modify this or not, introducing a PCD and defaulting it to existing implementation (set to PartId) will not break their code after they pick up a new EDK2 release with this fix, at least no behavior change to them. If the existing implementation is an issue for them, they can re-configure PCD per their use case requirement; if not, nothing bother them.

LeviYeoReum: > To maintain backward compatibility, introducing a PCD is needed. Because we don't know how the clients use this EDK2 and whether they modify this or not, introducing a PCD and defaulting it to existing implementation (set to PartId) will not break their code after they pick up a new EDK2 release with this fix, at least no behavior change to them. If the existing implementation is an issue for them, they can re-configure PCD per their use case requirement; if not, nothing bother them.

EDK2 requires FF-A v1.2 as the minimum implementation specification, and in v1.2 the requirements for w1 in FFA_RXTX_UNMAP are the same.
Therefore, there is no need to introduce a PCD here for the sake of maintaining backward compatibility.

Moreover, introducing a PCD would actually be undesirable, as it would prevent users from using the same built binary regardless of whether it is running in a physical or virtual environment.

This change does not violate the specification either. EDK2 is not used as a hypervisor, and in the physical case the value is always “0”. In the virtual case, sending 0 also complies with the specification. A simple comment should be sufficient.

If sending 0 were to cause any issue, it would mean that another component has not correctly implemented the FF-A v1.2 specification, and that is not something we need to account for.

Again, requiring a rebuild of the binary through the use of a PCD depending on whether it runs in a physical or virtual environment is not a good approach. Forcing users to be aware of this distinction and rebuild accordingly implies that the binary itself must somehow know whether it is running on physical hardware or in a virtual environment, which is neither desirable nor necessary.

leiflindholm: > Moreover, introducing a PCD would actually be undesirable, as it would prevent users from using the same built binary regardless of whether it is running in a physical or virtual environment.

@LeviYeoReum while it's doesn't seem possible to configure github to respect the diff options to treat .dec as a .inf, this definition is actually inside a 
`[PcdsFixedAtBuild, PcdsPatchableInModule, PcdsDynamic, PcdsDynamicEx]` block.
Meaning a platform could override this at runtime.

The default setting preserves the current (incorrect for a vm) value.

LeviYeoReum: Hi Leif,

> @LeviYeoReum while it's doesn't seem possible to configure github to respect the diff options to treat .dec as a .inf, this definition is actually inside a
[PcdsFixedAtBuild, PcdsPatchableInModule, PcdsDynamic, PcdsDynamicEx] block.
Meaning a platform could override this at runtime.
>
> The default setting preserves the current (incorrect for a vm) value.

What I want to point out is that, regardless of whether it can be overridden, the very existence of this PCD itself is not appropriate for EDK2.

At runtime, EDK2 should not need to determine whether it is running on physical hardware or in a virtual environment.

According to the actual intent of the ARM FF-A specification, if it is not running under a hypervisor, the intended behavior is to pass 0 to unmap.

In other words, unless it is a hypervisor, handling this field with a meaningless if condition does not really make sense.

Therefore, since EDK2 is not acting as a hypervisor, the ID should always be 0, and in the case of a VM, the hypervisor can simply provide 0.

For this reason, rather than introducing a meaningless PCD and wrapper function, I think it would be better to simply set it to 0 and clearly document its meaning in a comment.

shenghuliu: @LeviYeoReum 
Your overall understanding is right. But when implementing the Specs, we need to follow the Specs strictly. Take  FFA_RX_RELEASE as an example, the Specs says, "Only valid at the Non-secure physical FF-A instance.  MBZ otherwise.". So the logic should be per Specs:
   If( physical )
      w1=  FFA_ID_GET call;     // only valid
   else
     w1 = 0;      //otherwise

If you agree the above logic, can you let me know how to judge physical or virtual? Please keep in mind, EDK2 can run in both EL1 and EL2, hypervisor can exist or not, depending on the design. 

As for what is returned from FFA_ID_GET call, FFA_RX_RELEASE should not care, it just put whatever returned from FFA_ID_GET in physical case.

It shouldn't be just hardcoded 0 because for physical case FFA_ID_GET should return 0 per Specs. Even though both final results(physical/virtual) are 0, they are 0 just by coincidence, with two different concepts here. Say in future, Specs is changed for FFA_ID_GET to return non-zero in physical case, this code will break immediately. Hardcoded 0 will coupling these two commands for some specific value, it is not good practice.

As for default value for PCD, I think your input make sense, we can default it to virtual. Hopefully there will be no
complaint if some clients already fix the bug in different way.

shenghuliu: BTW, I have opened a case with ARM last week, and ask what is the use case of (w1 != 0), if never, why VM ID is needed any way.  But no matter what, I believe we provided solid implementation in following the Specs.

shenghuliu: forgot another point, in other FF-A command, physical/virtual cases are also checked. Since they are not the topic in our case, we just leave it alone. I also don't see those commands are implemented in EDK2 so far.

LeviYeoReum: > Your overall understanding is right. But when implementing the Specs, we need to follow the Specs strictly. Take FFA_RX_RELEASE as an example, the Specs says, "Only valid at the Non-secure physical FF-A instance. MBZ otherwise.". So the logic should be per Specs:
If( physical )
w1= FFA_ID_GET call; // only valid
else
w1 = 0; //otherwise

First, If the client should know whether it's physical or vitual, I definitely said "it's a spec fault".
That's why I continuously said "This is wrong for specification". endpoint shouldn't discern whether it's a physical or vitual and it doesn't need to. (Why does it acks itself?"

As I said last time. This "VM ID" field is for "Hypervisor" delivers to the "SPMC".
Since hyervisor knows the where it came from (vcpu), client doesn't need to pass the "VM ID".
So for enpoint perspective, It should be ignored and as the intention, It should be 0.

I know "SPEC" makes a lot of confusion. It gives hallucination the endpoint should know "physical" or "virtual" but its intention doesn't. (Might you got the same answer).

> , physical/virtual cases are also checked.

Where? if you said (Partition ID > 0), It doesn't discern the physical or virtual since VM id could be == 0.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
kuqin12: I missed the one in RxTx buffer unmap usage. Resolving this comment.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
shenghuliu: I agree with what you said, the Specs may need more clarifications, but this is not what we want to address here. We can discuss this topic in separate thread.

Again, when implementing FFA_RX_RELEASE, we should only follow what Specs says, not some result deduced from some logic. A good SW practice also suggests using the API to abstract the information (in this case it is FFA_ID_GET), assuming FFA_ID_GET always return 0 in physical case is not FFA_RE_RELEASE function should do.  

I plan to keep the same code because it complies what the Specs. I will take your point and defaulting it to virtual case. Please let me know your comments. Thank you for your time and valuable inputs, I really appreciate it.

Here is a couple of more examples:
FFA_MSG_WAIT  w2, 
FFA_MSG_SEND2  w1,
there are more trivial usage of physical/virtual concept in non-secure VM cases. You can search if you are interested.

LeviYeoReum: > Here is a couple of more examples:
FFA_MSG_WAIT w2,
FFA_MSG_SEND2 w1,
there are more trivial usage of physical/virtual concept in non-secure VM cases. You can search if you are interested.

This is not a proper example since it's end when they're set with "PART-ID". Definitely, This is not a reason 
We need to make the edk2 endpoint ack whether it's running on physical or virtual.

> I plan to keep the same code because it complies what the Specs. I will take your point and defaulting it to virtual case. Please let me know your comments. Thank you for your time and valuable inputs, I really appreciate it.

Again, the setting the 0 for RX-TX buffer is also "compiled" with the specification.
And I think *it's wrong" to let edk2 ack whether it physical or vitual (if something is there, It means "fault of spec").

So, mentioning compilance with the spec is not a rational reason to make *PCD" for virutal or physical but it gives a terrible "hallucination" client should know whether it runs on virtual or physical.

Thanks.

shenghuliu:
reopen

shenghuliu:
need to reopen this PR to resume review

PR #12240, opened 2026-03-03
Update Standalone MM PEI IPL hob handling

kuqin12:
# Description

This change guards the `gEfiAcpiVariableGuid` inside `PcdAcpiS3Enable` as this is only needed for S3 context related data. The current requires this hob to be created otherwise the module will assert.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested on QEMU Q35 platform and booted to UEFI shell.

## Integration Instructions

N/A

### DIFF
--- a/StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
+++ b/StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
@@ -1026,9 +1026,12 @@ CreateMmFoundationHobList (
   //
   // Build ACPI variable HOB
   //
-  HobLength = GetRemainingHobSize (*FoundationHobSize, UsedSize);
-  MmIplCopyGuidHob (FoundationHobList + UsedSize, &HobLength, &gEfiAcpiVariableGuid, FALSE);
-  UsedSize += HobLength;
+  if (PcdGetBool (PcdAcpiS3Enable)) {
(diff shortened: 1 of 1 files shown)


StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
niruiyu, line 1: @kuqin12 can you share more details regarding what extra steps are needed in your platform so that you choose to put these steps, plus the hob creation in a separate module?

I am asking because if the extra steps are always needed, I prefer to delete the hob creation logic from IPL.

kuqin12: @niruiyu the issue is that this region is unblocked spontaneously in the same routine, which means the STMM will treat this region the same as others when it comes to unblocking operation. However, a communication buffer is special, and we could apply stricter validations/page table attributes on this region. But I think it could be handled on the core side rather than IPL given we have all the information available. Will remove this change in the next push.

kuqin12: This change is removed.

niruiyu (approved):

PR #12242, opened 2026-03-04
UefiCpuPkg: MpInitLibUp: Honour OPTIONAL semantics in GetNumberOfProc…

sgrams:
# Description

`MpInitLibGetNumberOfProcessors()` in `MpInitLibUp` unconditionally dereferenced both `OPTIONAL` output parameters without any NULL guard:

```c
*NumberOfProcessors        = 1;
*NumberOfEnabledProcessors = 1;
```

`CpuMpPei` legitimately passes `NULL` for `NumberOfEnabledProcessors` from `InitializeMpExceptionStackSwitchHandlers()`:

```c
Status = MpInitLibGetNumberOfProcessors (&NumberOfProcessors, NULL);
```

Because the function is small enough for GCC to inline, the compiler proves that `*(NULL) = 1` is undefined behaviour and replaces the call site with a `ud2` instruction. This results in a `#UD` (Invalid Opcode, exception type 06) trap during the `MemoryDiscoveredPpiNotifyCallback` PPI notification.

The fix adds a NULL check around each dereference so that a `NULL` argument for either `OPTIONAL` output parameter is silently skipped. `EFI_INVALID_PARAMETER` is returned only when **both** pointers are `NULL`, matching the existing function documentation.

- [ ] Breaking change?
  N/A
- [ ] Impacts security?
  N/A
- [ ] Includes tests?
  N/A

## How This Was Tested

Booted TDVF Config-A OvmfX64 under TDX-enabled QEMU/KVM with a `DEBUG_GCC` build.
The `#UD` exception no longer occurs and `CpuMpPei` initialises successfully.

The crash was confirmed via the exception dump and resolved via `addr2line -i` against the debug binary:

```
MpInitLibGetNumberOfProcessors        MpInitLibUp.c:72
InitializeMpExceptionStackSwitchHandlers  CpuMpPei.c:90
InitializeCpuMpWorker                 CpuMpPei.c:357 / :319
MemoryDiscoveredPpiNotifyCallback     CpuPaging.c:411
```

## Integration Instructions

N/A

### DIFF
--- a/StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
+++ b/StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
@@ -1026,9 +1026,12 @@ CreateMmFoundationHobList (
   //
   // Build ACPI variable HOB
   //
-  HobLength = GetRemainingHobSize (*FoundationHobSize, UsedSize);
-  MmIplCopyGuidHob (FoundationHobList + UsedSize, &HobLength, &gEfiAcpiVariableGuid, FALSE);
-  UsedSize += HobLength;
+  if (PcdGetBool (PcdAcpiS3Enable)) {
+    // Only check on this variable when S3 needs it.
+    HobLength = GetRemainingHobSize (*FoundationHobSize, UsedSize);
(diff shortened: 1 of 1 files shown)


niruiyu (approved):

sgrams:
This was already fixed in c422b3b97b69de7a0e5a79632d7b69cfc5c462b42.
We were testing edk2-stable202602 that did not have it.

PR #12244, opened 2026-03-04
BaseMemoryLib: Add a compiler barrier for SetMem

kuqin12:
# Description

Current `SetMem` does not have compiler barriers to prevent being reordered. If this is used against a buffer containing sensitive data, it could lead to unintended data leakage.

This change adds a compiler specific barrier for memory setting operations to prevent compiler induced reordering.

- [ ] Breaking change?
- [x] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This change is tested on ARM64 physical platform and measured the boot performance in consistency with the original implementation.

## Integration Instructions

N/A


mdkinney:
I am confused why ZeroMem() is not sufficient.  I believe all the implementations use volatile loop variables for C implementations or ZeroMem() is mapped to an assembly language implementation that can not be optimized away.

What is an example where a call to ZeroMem() does not actually zero the requested buffer?

kuqin12:
> I am confused why ZeroMem() is not sufficient. I believe all the implementations use volatile loop variables for C implementations or ZeroMem() is mapped to an assembly language implementation that can not be optimized away.
>
There are 2 additions in this version:

1. the compiler barrier after the wipe, this is to prevent the compiler from reordering the calls during optimization process. It could be that compiler sees no dependency between ZeroMem and the next function, yet the next function never returns.
2. the noinline and no-lto. This will prevent the compiler from inlining the function and then optimizing out the block when reasoning at the call site.

Please let me know if that makes sense.

mdkinney:
> > I am confused why ZeroMem() is not sufficient. I believe all the implementations use volatile loop variables for C implementations or ZeroMem() is mapped to an assembly language implementation that can not be optimized away.
> 
> There are 2 additions in this version:
> 
> 1. the compiler barrier after the wipe, this is to prevent the compiler from reordering the calls during optimization process. It could be that compiler sees no dependency between ZeroMem and the next function, yet the next function never returns.
> 2. the noinline and no-lto. This will prevent the compiler from inlining the function and then optimizing out the block when reasoning at the call site.
> 
> Please let me know if that makes sense.

The compiler hint barriers could be added to the existing ZeroMem() implementations.  That would solve (1).  I would also like to see a reproducible test case that demonstrates a reordering like you are stating.

Do you have a real example of (2) where a call to ZeroMem() that uses volatiles or ASM was optimized away?  Please provide a reproducible test case.  

The optimizers are very good at removing uncalled code.  So if what you are observing is a higher function in the call stack be optimized away due to a const conditional evaluation at compile time and that removed the entire call tree including a ZeroMem() call, then that would be correct behavior of the compiler. Any call to ZeroMem() that is required to clear sensitive information must never be in a conditional path.

kuqin12:
> The compiler hint barriers could be added to the existing ZeroMem() implementations. That would solve (1). I would also like to see a reproducible test case that demonstrates a reordering like you are stating.
> 
Hmm, I would prefer to separate the security guarantee from this existing API. From how I understand the BaseMemoryLib, it is meant to provide the "optimized implementations for common memory-based operations", so optimization is all it cares which makes sense.

But the secure version of zero memory provides the API under a different promise. So I think we should avoid combining the 2 guarantees together into the single function (as what other system library has been doing).

I think it makes better sense to go through an RFC at this point. I will initiate that so that we can move away from in PR discussion.

mikebeaton:
@kuqin12 - Did you raise an RFC? Could not find, but realise it may be in progress.

While I don't have a very strong response, less so a strong reason why my response should be considered, if others more senior disagree, my immediate reaction - I think along with @mdkinney? - would be that ZeroMem itself should be guaranteed to zero the memory. Happy to transfer my brief comment to RFC itself, if/when it's in place.

kuqin12:
> @kuqin12 - Did you raise an RFC? Could not find, but realise it may be in progress.
> 
> While I don't have a very strong response, less so a strong reason why my response should be considered, if others more senior disagree, my immediate reaction - I think along with @mdkinney? - would be that ZeroMem itself should be guaranteed to zero the memory. Happy to transfer my brief comment to RFC itself, if/when it's in place.

I have not raised an RFC yet. Will post the link here once I have it. Thanks for your interest :)

mdkinney:
> > @kuqin12 - Did you raise an RFC? Could not find, but realise it may be in progress.
> > While I don't have a very strong response, less so a strong reason why my response should be considered, if others more senior disagree, my immediate reaction - I think along with @mdkinney? - would be that ZeroMem itself should be guaranteed to zero the memory. Happy to transfer my brief comment to RFC itself, if/when it's in place.
> 
> I have not raised an RFC yet. Will post the link here once I have it. Thanks for your interest :)

I did happen to notice the CompilerIntrinsicLib used for AARCH64 does not use `volatile` in the memcpy/memmove/memset implementations. Is this perhaps where you are observing some unexpected optimizations?

jaykrell:
"We" have "related" input/experience here to share. :)
Compiler developers are not willing to be blind to no-lto.
  This surprised me, but it was recently asserted clearly.
    And it feels fragile anyway, to be so dependent on compiler flags.
  This is leading to more assembly in related and unrelated areas.

Noinline certainly is not a barrier to compilers seeing into functions and learning their semantics, and optimizing based on that. I know this surprises many, but it makes sense.

It is not clear if volatile is a barrier to compilers in this regard -- if volatiles stores w/o visible loads are considered dead.

Where this leaves things is that in order to inhibit compilers we are getting single instruction assembly wrappers like:
volatile_memcpy:
  jmp memcpy

volatile_memset:
  jmp memset

I think that should be considered the solution.
Not entirely new functions necessarily, just single assembly instruction to stop the compiler from looking further.
At least with Visual C++.

That is though without volatile in memset/memcpy.

jaykrell:
Again, I don't think volatile is enough.
I know very security-conscious people insist on putting an assembly stub (jmp) around memset, to prevent the compiler from seeing through it. I thought no-lto C stub would suffice but we are told by compiler developers it might not.

Anything that is only C is not strong enough to defeat compilers.

jaykrell:
Ok I asked about the assembly we have.
The reason for, like, "secure_memset is assembly: jmp memset" is to keep "fast memset", i.e. without volatile and barriers.
So there is that tradeoff, of portability vs. performance.
volatile is correct, but slower.

kuqin12:
> Ok I asked about the assembly we have. The reason for, like, "secure_memset is assembly: jmp memset" is to keep "fast memset", i.e. without volatile and barriers. So there is that tradeoff, of portability vs. performance. volatile is correct, but slower.

Sure, there is going to be tradeoff when it comes to security. But this is supposed to erase sensitive areas after use, where we need to prioritize the security over performance.

Sorry for the delay btw. Thanks for keep tapping on this thread. I will pick up the topic on the mailing list shortly.

kuqin12:
RFC'd here: https://github.com/tianocore/tianocore-wiki.github.io/pull/29

kuqin12:
> RFC'd here: [tianocore/tianocore-wiki.github.io#29](https://github.com/tianocore/tianocore-wiki.github.io/pull/29)

The RFC has settled with a PR change into `SetMem` because the elision does not happen due to `volatile` usage and nasm function callout, depending on the instance. The ordering does not happen on optimized instances due to nasm callouts serving as natural barrier.

The only impacted case is this base instance of memory library being reordered. Thus filing an issue to fix that.

MdePkg/Library/BaseMemoryLib/SetMem.c
```diff
@@ -14,6 +14,38 @@
...
+/**
+  A helper function to emit a compiler barrier and treat Buffer as an input to
+  discourage reordering and call-site elimination.
+
+  @param  Buffer  Pointer to the buffer being zeroed.
+
+**/
+STATIC
+VOID
+InternalMemBarrier (
```
mdkinney: Given the source of this change is around clearing memory associated with ZeroMem(), would it make more sense to integrate this change into implementations of:

```c
VOID *
EFIAPI
InternalMemZeroMem (
  OUT     VOID   *Buffer,
  IN      UINTN  Length
  )
{
  return InternalMemSetMem (Buffer, Length, 0);
}
```

This way the strong ordering for the barrier is only applied to ZeroMem() APIs and not all SetMemxx() APIs.

MdePkg/Library/BaseMemoryLib/SetMem.c
```diff
@@ -14,6 +14,38 @@
...
+  IN VOID  *Buffer
+  )
+{
+ #if defined (_MSC_VER)
+  _ReadWriteBarrier ();
+  (VOID)Buffer;
```
mdkinney: What is the purpose of the `(VOID)Buffer;` statement?  Buffer should always be touched by the function called.

kuqin12: Good call. This was the leftover from an earlier refactoring. This is now removed.

MdePkg/Library/BaseMemoryLib/SetMem.c
kuqin12: Thanks. This is updated.

MdePkg/Library/BaseMemoryLib/MemLibGeneric.c
```diff
@@ -13,6 +13,40 @@
...
+  VOID
+  )
+{
+ #if defined (_MSC_VER)
+  _ReadWriteBarrier ();
+ #elif defined (__GNUC__) || defined (__clang__)
+  __asm__ __volatile__ ("" : : : "memory");
+ #else
+  //
+  // No portable compiler barrier is available for this toolchain.
```
mdkinney, line 45: Should this be a #error condition?

Also, it appears that _ReadWriteBarrier() is deprecated and the recommendation is to use: https://en.cppreference.com/c/atomic/atomic_signal_fence.  But that is likely always mapped to a compiler specific intrinsic by stdatomic.h, and we avoid use of compiler specific includes/libs except for submodules that have a libc dependency.

Does not have to be resolved for this PR because _ReadWriteBarrier() is used in other libs.

MdePkg/Library/BaseMemoryLib/MemLibGeneric.c
```diff
@@ -13,6 +13,41 @@
...
+  constraint holds even when the caller is inlined under LTO.
+
+**/
+STATIC
+VOID
+InternalMemBarrier (
+  VOID
+  )
+{
+ #if defined (_MSC_VER)
```
mdkinney: One additional comment on the order of #if checks.  I have seen clang set _MSC_VER.  Wondering of the GNUC and clang checks should be done first and only fall back to _MSC_VER if compiler is neither GCC nor clang?

MdePkg/Library/BaseMemoryLib/MemLibGeneric.c
kuqin12: The error message was added.

I did not take `atomic_signal_fence` for this change because of the header file as you mentioned. But agree that we need to resolve that for the whole repo.

MdePkg/Library/BaseMemoryLib/MemLibGeneric.c
os-d: Agreed. CLANGPDB sets _MSC_VER, as you mentioned. So if you don’t want CLANGPDB to hit that case, then the order needs to be flipped

mdkinney (approved):

MdePkg/Library/BaseMemoryLib/MemLibGeneric.c
kuqin12: Indeed. This is updated per feedback. Thanks.

MdePkg/Library/BaseMemoryLib/MemLibGeneric.c
```diff
@@ -13,6 +13,41 @@
 
 #include "MemLibInternals.h"
 
+#if defined (_MSC_VER) && !defined (__GNUC__) && !defined (__clang__)
+VOID
+_ReadWriteBarrier (
```
os-d, line 18: Should these live in BaseLib? We have MemoryFence today, which meets the definition of GCC/Clang below. Should we be centralizing on that? Or if that API is not correct, then add these ones there?

kuqin12: From what I can tell, `MemoryFence` is serving different purposes. The AArch64, RISC and LoongArch64 are implementing this function as a hardware barrier, which prevents the processor from out-of-order execution during runtime.

But this is a compiler barrier to prevent the compiler from reordering the compiled code and emits no extra instructions. We could add a new API in BaseLib, but that is likely more suitable as a separate task.

Created a github issue here: https://github.com/tianocore/edk2/issues/12819.

os-d: Ok, I’m fine to handle the redefinitions later.

os-d (approved):

PR #12249, opened 2026-03-06
OvmfPkg/XenPvBlkDxe: Advertise the correct IO alignment

rosslagerwall:
# Description

The function XenPvBlockAsyncIo expects a sector-aligned buffer but the advertised alignment is always 512. Consequently, booting from the Rocky Linux 9.6 disc (using a 2048 sector size) fails with an assertion in XenPvBlockAsyncIo.

Adverstise the correct IO alignment to fix the assertion failure.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Manually tested installation of Rocky Linux 9.6 on XenServer using a virtual DVD provided by iDRAC. Before, the change, it hit an assertion failure. After the change, it completed successfully.

## Integration Instructions

N/A


ardbiesheuvel:
@mergifyio queue

PR #12251, opened 2026-03-06
DynamicTablesPkg: AcpiIortLibArm: Support IORT revision 6

kuqin12:
# Description

Starting from IO Remapping Table rev E.f (IORT revision 6), the SMMU v3 node is revisioned to 5. (https://developer.arm.com/documentation/den0049/ef/?lang=en)

This PR updates the ACPI IORT generator to support IORT revision 6.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested on proprietary hardware platform.

## Integration Instructions

N/A

### DIFF
--- a/DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
+++ b/DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
@@ -1490,6 +1490,158 @@ AddSmmuV1V2Nodes (
   return EFI_SUCCESS;
 }
 
+/**
+  Helper function to find the first IORT node of a specific type.
+
+  @param [in] Iort Pointer to the IORT table.
+  @param [in] NodeType The type of the node to find.
+  @param [in] Offset The offset from the beginning of the IORT table to start searching from.
+  @param [out] Node Pointer to a variable that will receive the found node.
+
+  @retval EFI_SUCCESS The function completed successfully.
+  @retval EFI_INVALID_PARAMETER One or more parameters are invalid.
+  @retval EFI_NOT_FOUND The specified node type was not found.
+
+**/
+STATIC
+EFI_STATUS
+FindIortNodeByType (
+  IN CONST EFI_ACPI_6_0_IO_REMAPPING_TABLE  *Iort,
+  IN UINT8                                  NodeType,
+  IN UINTN                                  Offset,
+  OUT EFI_ACPI_6_0_IO_REMAPPING_NODE        **Node
+  )
+{
+  EFI_ACPI_6_0_IO_REMAPPING_NODE  *CurrentNode;
+  UINT32                          CurrentOffset;
+
+  if ((Iort == NULL) || (Node == NULL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  // Validate the offset is within the bounds of the IORT table and is not before the first node.
+  if ((Offset >= Iort->Header.Length) || (Offset < Iort->NodeOffset)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  // Even if the offset is within the range of IORT table, it may not be the start of a node.
+  // So we need to iterate through the nodes until we find the node with the specified type or reach the end of the table.
+  CurrentOffset = Iort->NodeOffset;
+  while (CurrentOffset < Iort->Header.Length) {
+    CurrentNode = (EFI_ACPI_6_0_IO_REMAPPING_NODE *)((UINT8 *)Iort + CurrentOffset);
+    if ((CurrentOffset == Offset) && (CurrentNode->Type == NodeType)) {
+      *Node = CurrentNode;
+      return EFI_SUCCESS;
+    } else if (CurrentOffset > Offset) {
+      // If we have passed the offset and still haven't found the node, it means the node with the specified type does not
+      // exist at the specified offset.
+      DEBUG ((
+        DEBUG_ERROR,
+        "%a: IORT: Failed to find node with type %d at offset 0x%x. Current node type = %d, Current offset = 0x%x.\n",
+        __func__,
+        NodeType,
+        Offset,
+        CurrentNode->Type,
+        CurrentOffset
+        ));
+      return EFI_NOT_FOUND;
+    }
+
+    if (CurrentNode->Length == 0) {
+      // To avoid infinite loop in case of invalid node length, we need to check if the node length is zero.
+      DEBUG ((
+        DEBUG_ERROR,
+        "%a: IORT: Invalid node length 0 at offset 0x%x. Current node type = %d.\n",
+        __func__,
+        CurrentOffset,
+        CurrentNode->Type
+        ));
+      return EFI_NOT_FOUND;
+    }
+
+    CurrentOffset += CurrentNode->Length;
+  }
+
+  return EFI_NOT_FOUND;
+}
+
+/**
+  Helper function to evaluate the Device ID mapping flag validity for a given SMMUv3 node.
+
+  @param [in] Iort Pointer to the IORT table.
+  @param [in] SmmuV3Node Pointer to the SMMUv3 node to evaluate.
+
+  @retval EFI_SUCCESS             The function completed successfully.
+  @retval EFI_INVALID_PARAMETER   One or more parameters are invalid.
+  @retval EFI_DEVICE_ERROR        The Device ID Mapping Index is invalid or the referenced ITS Group Node is not found in the IORT table.
+
+**/
+STATIC
+EFI_STATUS
+EvaluateSmmuV3NodeDeviceIdMappingFlag (
+  IN CONST EFI_ACPI_6_0_IO_REMAPPING_TABLE  *Iort,
+  IN EFI_ACPI_6_0_IO_REMAPPING_SMMU3_NODE   *SmmuV3Node
+  )
+{
+  EFI_ACPI_6_0_IO_REMAPPING_ID_TABLE  *DeviceId;
+  EFI_ACPI_6_0_IO_REMAPPING_NODE      *ItsGroupNode;
+  EFI_STATUS                          Status;
+  BOOLEAN                             DeviceIdMappingIndexValid;
+
+  if (SmmuV3Node == NULL) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  // Validate the Device ID Mapping Index if the platform indicate it is valid
+  DeviceIdMappingIndexValid = SmmuV3Node->Flags & EFI_ACPI_IORT_SMMUv3_FLAG_DEVICEID_VALID;
+  if (!DeviceIdMappingIndexValid) {
+    // DeviceIdMappingIndex != 0 implies that the DEVICEID_VALID flag should be set.
+    ASSERT (SmmuV3Node->DeviceIdMappingIndex == 0);
+    return EFI_SUCCESS;
+  }
+
+  // 1. The Device ID Mapping Index should be less than the number of Id Mappings for the SMMUv3 node.
+  if (SmmuV3Node->DeviceIdMappingIndex >= SmmuV3Node->Node.NumIdMappings) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a: IORT: Invalid Device ID Mapping Index %d for SMMUv3 Node with %d Id Mappings, .\n",
+      __func__,
+      SmmuV3Node->DeviceIdMappingIndex,
+      SmmuV3Node->Node.NumIdMappings
+      ));
+    return EFI_DEVICE_ERROR;
+  }
+
+  // 2. The Reference to ID Array should be non-zero and the Id Mapping Token should be valid.
+  if (SmmuV3Node->Node.IdReference == 0) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a: IORT: Invalid reference to ID Array for SMMUv3 Node with Device ID Mapping Index %d.\n",
+      __func__,
+      SmmuV3Node->DeviceIdMappingIndex
+      ));
+    return EFI_DEVICE_ERROR;
+  }
+
+  // 3. If the Device ID Mapping Index is still valid, the output reference should point to a valid ITS Group Node in the IORT.
+  DeviceId = (EFI_ACPI_6_0_IO_REMAPPING_ID_TABLE *)((UINT8 *)SmmuV3Node +
+                                                    SmmuV3Node->Node.IdReference) + SmmuV3Node->DeviceIdMappingIndex;
+
+  Status = FindIortNodeByType (Iort, EFI_ACPI_IORT_TYPE_ITS_GROUP, DeviceId->OutputReference, &ItsGroupNode);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a: IORT: Failed to find ITS Group Node in IORT from OutputReference 0x%x. Status = %r\n",
+      __func__,
+      DeviceId->OutputReference,
+      Status
+      ));
+    return Status;
+  }
+
+  return EFI_SUCCESS;
+}
+
 /** Update the SMMUv3 Node Information.
 
     This function updates the SMMUv3 node information in the IORT table.
@@ -1554,8 +1706,11 @@ AddSmmuV3Nodes (
     {
       SmmuV3Node->Node.Revision   = 2;
       SmmuV3Node->Node.Identifier = EFI_ACPI_RESERVED_DWORD;
-    } else {
+    } else if (AcpiTableInfo->AcpiTableRevision == EFI_ACPI_IO_REMAPPING_TABLE_REVISION_05) {
       SmmuV3Node->Node.Revision   = 4;
+      SmmuV3Node->Node.Identifier = NodeList->Identifier;
+    } else {
+      SmmuV3Node->Node.Revision   = 5;
       SmmuV3Node->Node.Identifier = NodeList->Identifier;
     }
 
@@ -1577,7 +1732,8 @@ AddSmmuV3Nodes (
       SmmuV3Node->ProximityDomain = 0;
     }
 
-    if ((SmmuV3Node->Event != 0) && (SmmuV3Node->Pri != 0) &&
+    if ((SmmuV3Node->Node.Revision < 5) &&
+        (SmmuV3Node->Event != 0) && (SmmuV3Node->Pri != 0) &&
         (SmmuV3Node->Gerr != 0) && (SmmuV3Node->Sync != 0))
     {
       // If all the SMMU control interrupts are GSIV based,
@@ -1614,6 +1770,18 @@ AddSmmuV3Nodes (
         DEBUG ((
           DEBUG_ERROR,
           "ERROR: IORT: Failed to add Id Mapping Array. Status = %r\n",
+          Status
+          ));
+        return Status;
+      }
+    }
+
+    if (SmmuV3Node->Node.Revision >= 5) {
+      Status = EvaluateSmmuV3NodeDeviceIdMappingFlag (Iort, SmmuV3Node);
+      if (EFI_ERROR (Status)) {
+        DEBUG ((
+          DEBUG_ERROR,
+          "ERROR: IORT: Evaluation of Device ID Mapping Index for SMMUv3 Node failed. Status = %r\n",
           Status
           ));
         return Status;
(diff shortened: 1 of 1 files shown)


LeviYeoReum (approved):

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
samimujawar: Is it possible to update the Spec reference, please?

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
```diff
@@ -1554,9 +1554,12 @@ AddSmmuV3Nodes (
     {
       SmmuV3Node->Node.Revision   = 2;
       SmmuV3Node->Node.Identifier = EFI_ACPI_RESERVED_DWORD;
-    } else {
+    } else if (AcpiTableInfo->AcpiTableRevision == EFI_ACPI_IO_REMAPPING_TABLE_REVISION_05) {
       SmmuV3Node->Node.Revision   = 4;
       SmmuV3Node->Node.Identifier = NodeList->Identifier;
+    } else {
+      SmmuV3Node->Node.Revision   = 5;
```
pierregondois, line 1713: Hello,
I think other modifications should be done in order to increase the supported version. The spec says:
```
E.e:
Added flags in SMMUv3 node for
validity of ID mappings for MSIs related
to control interrupts.
```
And the flag description:
```
Set to 1 if the DeviceID mapping index field is valid. Set to 0
otherwise.
```

So ideally we should also check / invalidate the DeviceID mapping
depending on the flag value.

kuqin12: Thanks. I made a change in mu and will test it. Once it is tested on physical platforms, I will bring it over here: https://github.com/microsoft/mu_basecore/pull/1707

kuqin12: But please leave feedback if you see anything unexpected in that PR in terms of validation.

kuqin12: Okay, this is updated and tested on physical platform.

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
```diff
@@ -1490,6 +1490,157 @@ AddSmmuV1V2Nodes (
...
+  IN CONST EFI_ACPI_6_0_IO_REMAPPING_TABLE  *Iort,
+  IN EFI_ACPI_6_0_IO_REMAPPING_SMMU3_NODE   *SmmuV3Node,
+  OUT BOOLEAN                               *DeviceIdMappingIndexValid
+  )
+{
+  EFI_ACPI_6_0_IO_REMAPPING_ID_TABLE  *DeviceId;
+  EFI_ACPI_6_0_IO_REMAPPING_NODE      *ItsGroupNode;
+  EFI_STATUS                          Status;
+
+  if ((DeviceIdMappingIndexValid == NULL) || (SmmuV3Node == NULL)) {
```
LeviYeoReum: I'm not sure whether it's valuable to return EFI_INVALID_PARAMETER for this *staitc* function since to me
it seems almost dead code.
By removing this, I think it would be better to change the rettype as boolean for whether mapping is valid rather than passing @DeviceIdMappingIndexValid.

kuqin12: Agree that the current composite seems like a waste for the check here. But I was expecting it to be used later by other code in this file/library because it seems to be a very rudimentary helper function.

kuqin12:
@pierregondois @samimujawar Could you please kindly provide feedback on this PR? Thanks in advance!

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
```diff
@@ -1620,6 +1775,24 @@ AddSmmuV3Nodes (
...
+    if (EFI_ERROR (Status)) {
+      DEBUG ((
+        DEBUG_ERROR,
+        "ERROR: IORT: Failed to evaluate Device ID Mapping Index for SMMUv3 Node. Status = %r\n",
+        Status
+        ));
+      return Status;
+    }
+
+    ASSERT (DeviceIdMappingIndexValid == ((SmmuV3Node->Flags & EFI_ACPI_IORT_SMMUv3_FLAG_DEVICEID_VALID) != 0));
```
pierregondois: The HwInfoParser component generates CmOjects from DT parsing.
Would it be possible to either:
- modify the logic under:
```
    SmmuV3Info->DeviceIdMappingIndex = SmmuV3Info->IdMappingCount;
    SmmuV3Info->IdMappingCount++;
```
at
```DynamicTablesPkg/Library/FdtHwInfoParserLib/Arm/Iort/SmmuV3Parser.c```

to generate the EFI_ACPI_IORT_SMMUv3_FLAG_DEVICEID_VALID when needed

- or make the ASSERT a DEBUG message instead and I ll update the HwInfoParser later ?

File: DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c [hunk 4]
```diff
@@ -1490,6 +1490,157 @@ AddSmmuV1V2Nodes (
...
+    return EFI_INVALID_PARAMETER;
+  }
+
+  // Start searching from the specified offset, if the offset is less than the NodeOffset, start from NodeOffset
+  if (Offset < Iort->NodeOffset) {
+    Offset = Iort->NodeOffset;
+  }
+
+  // Even if the offset is within the range of IORT table, it may not be the start of a node.
+  // So we need to iterate through the nodes until we find the node with the specified type or reach the end of the table.
```
pierregondois, line 1528: If `Offset` doesn't point to an actual node, we might want to treat it as an error.
This is done with the check:
```
if (DeviceId->OutputReference != (UINT32)((UINTN)ItsGroupNode - (UINTN)Iort)) {
 ...
}
```

But maybe this could be done in the loop instead as:
```
while (...) {
  if ((CurrentOffset == Offset) && ...) {
    // Success
  } else if (CurrentOffset > Offset) {
    // error + bail out
  }
}
```

------

Also, the IORT table is not completely written at this point, so it might be possible that `CurrentNode->Length == 0`, which might create an infinite loop. We might want to bail out in such case.

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
```diff
@@ -1490,6 +1490,157 @@ AddSmmuV1V2Nodes (
...
+      "%a: IORT: Failed to find ITS Group Node in IORT from OutputReference 0x%x. Status = %r\n",
+      __func__,
+      DeviceId->OutputReference,
+      Status
+      ));
+    *DeviceIdMappingIndexValid = FALSE;
+    goto Exit;
+  }
+
+  if (DeviceId->OutputReference != (UINT32)((UINTN)ItsGroupNode - (UINTN)Iort)) {
```
pierregondois: Cf. the comment above, it might be better to move the validation of the Node offset in the function iterating over nodes (`FindIortNodeByType()`) and let this function only validate the `EFI_ACPI_IORT_SMMUv3_FLAG_DEVICEID_VALID` flag

pierregondois (commented):
Hello Kun,
Just a few more comments, but nothing big hopefully

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
kuqin12: Thanks, this is updated by moving the offset check into the `FindIortNodeByType` function

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
kuqin12: Thanks, this is updated by separating the == and > cases as suggested. The 0 length check is also added.

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
kuqin12: Thanks for the feedback. But we are not using the `FdtHwInfoParserLib` to generate the input as of today. I would not be able to test the change if I make it. So I took the changing the assert to debug message approach. May I defer the real change to you for later? :)

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
```diff
@@ -1620,6 +1787,32 @@ AddSmmuV3Nodes (
...
+    if (EFI_ERROR (Status)) {
+      DEBUG ((
+        DEBUG_ERROR,
+        "ERROR: IORT: Failed to evaluate Device ID Mapping Index for SMMUv3 Node. Status = %r\n",
+        Status
+        ));
+      return Status;
+    }
+
+    if (DeviceIdMappingIndexValid != ((SmmuV3Node->Flags & EFI_ACPI_IORT_SMMUv3_FLAG_DEVICEID_VALID) != 0)) {
```
pierregondois: IIUC `EvaluateSmmuV3NodeDeviceIdMappingFlag()` should have already printed an error message if this condition is triggered.
Indeed:
- if `EFI_ACPI_IORT_SMMUv3_FLAG_DEVICEID_VALID` is set, then having `DeviceIdMappingIndexValid == FALSE` is only possible if an error was triggered
- if `EFI_ACPI_IORT_SMMUv3_FLAG_DEVICEID_VALID` is not set, then `DeviceIdMappingIndexValid` is set to `FALSE`

Maybe it is ok to remove the duplicate error message.

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
```diff
@@ -1490,6 +1490,169 @@ AddSmmuV1V2Nodes (
...
+  }
+
+  // Validate the offset is within the bounds of the IORT table
+  if (Offset >= Iort->Header.Length) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  // Start searching from the specified offset, if the offset is less than the NodeOffset, start from NodeOffset
+  if (Offset < Iort->NodeOffset) {
+    Offset = Iort->NodeOffset;
```
pierregondois: Shouldn't this be an error aswell ?
Maybe this should be move up with the `if (Offset >= Iort->Header.Length)` check

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
```diff
@@ -1490,6 +1490,169 @@ AddSmmuV1V2Nodes (
...
+      DeviceId->OutputReference,
+      Status
+      ));
+    *DeviceIdMappingIndexValid = FALSE;
+    goto Exit;
+  }
+
+  *DeviceIdMappingIndexValid = TRUE;
+
+Exit:
```
pierregondois: The `Exit` label doesn't seem to contain any error handling code.
It should be ok to just return direcly the `Status` instead of going here.
(I.e. maybe the `goto` should be replaced by `return` statements)

pierregondois (commented):
Hello Kun,
This should be the last comments I have

pierregondois:
Actually there are more comments.
If you can please wait a bit before any update.

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
pierregondois: If `SmmuV3Node->Node.Revision>= 5`, we should not clear `SmmuV3Node->DeviceIdMappingIndex` anymore. I thing the condition should be:
```
@@ -1744,7 +1734,8 @@ AddSmmuV3Nodes (
       SmmuV3Node->ProximityDomain = 0;
     }
 
-    if ((SmmuV3Node->Event != 0) && (SmmuV3Node->Pri != 0) &&
+    if ((SmmuV3Node->Node.Revision < 5) &&
+        (SmmuV3Node->Event != 0) && (SmmuV3Node->Pri != 0) &&
         (SmmuV3Node->Gerr != 0) && (SmmuV3Node->Sync != 0))
     {
       // If all the SMMU control interrupts are GSIV based,

```

This means that we are not checking anymore:
1. The Device ID Mapping Index should be less than the number of Id Mappings for the SMMUv3 node.
2. The Reference to ID Array should be non-zero and the Id Mapping Token should be valid.
3. If the Device ID Mapping Index is still valid, the output reference should point to a valid ITS Group Node in the IORT.

for the SmmuV3 nodes with revision 4, but it might be impossible to do so as there is no indication on whether the `DeviceIdMappingIndex` is valid or not, so it might be a lost cause.

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
```diff
@@ -1620,6 +1787,32 @@ AddSmmuV3Nodes (
...
+      DEBUG ((
+        DEBUG_ERROR,
+        "ERROR: IORT: Device ID Mapping Index validity mismatch for SMMUv3 Node. "
+        "DeviceIdMappingIndexValid = %d, DeviceID Valid Flag = %d.\n",
+        DeviceIdMappingIndexValid,
+        (SmmuV3Node->Flags & EFI_ACPI_IORT_SMMUv3_FLAG_DEVICEID_VALID) != 0
+        ));
+    }
+
+    if (!DeviceIdMappingIndexValid) {
```
pierregondois: Cf. the comment above:
- if `DeviceIdMappingIndexValid == TRUE`, there should not need to set `EFI_ACPI_IORT_SMMUv3_FLAG_DEVICEID_VALID`
- If `DeviceIdMappingIndexValid == FALSE`, this should have been an error and we should have exited aswell
Maybe this should be removed aswell.

This implies that `DeviceIdMappingIndexValid` is not actually need in this function.

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
```diff
@@ -1620,6 +1787,32 @@ AddSmmuV3Nodes (
       }
     }
 
+    Status = EvaluateSmmuV3NodeDeviceIdMappingFlag (Iort, SmmuV3Node, &DeviceIdMappingIndexValid);
```
pierregondois: I think this should be only called if `(SmmuV3Node->Node.Revision >= 5)`. Like that we can consider any invalid configuration as an error.

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
```diff
@@ -1490,6 +1490,169 @@ AddSmmuV1V2Nodes (
...
+  if (SmmuV3Node->DeviceIdMappingIndex >= SmmuV3Node->Node.NumIdMappings) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a: IORT: Invalid Device ID Mapping Index %d for SMMUv3 Node with %d Id Mappings, .\n",
+      __func__,
+      SmmuV3Node->DeviceIdMappingIndex,
+      SmmuV3Node->Node.NumIdMappings
+      ));
+    *DeviceIdMappingIndexValid = FALSE;
+    goto Exit;
```
pierregondois: I think this should be considered as an invalid state and return an error status.
Similar comment for the other `goto` statements

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
```diff
@@ -1490,6 +1490,169 @@ AddSmmuV1V2Nodes (
...
+  if ((DeviceIdMappingIndexValid == NULL) || (SmmuV3Node == NULL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  Status = EFI_SUCCESS;
+
+  // Validate the Device ID Mapping Index if the platform indicate it is valid
+  *DeviceIdMappingIndexValid = SmmuV3Node->Flags & EFI_ACPI_IORT_SMMUv3_FLAG_DEVICEID_VALID;
+  if (!(*DeviceIdMappingIndexValid)) {
+    goto Exit;
```
pierregondois: Maybe we can add an `ASSERT (SmmuV3Node->DeviceIdMappingIndex != 0);` just to catch if someone forgot to set the `EFI_ACPI_IORT_SMMUv3_FLAG_DEVICEID_VALID` flag.
Otherwise we can return `EFI_SUCCESS` here.

pierregondois (commented):
Hello Kun,
Sorry I added more requests in the end

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
kuqin12: Agreed, I saw this in the spec when I was trying to add the check but then the thoughts were got lost while I implemented the functionality. I added the revision check in the latest push. And `DeviceIdMappingIndex` is removed.

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
kuqin12: This is added.

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
kuqin12: Good call. This is removed.

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
kuqin12: Thanks. Replaced them with returns.

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
kuqin12: With the updated implementation of the function from our previous discussions, I agree this is a better route to go. This is updated.

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
kuqin12: As mentioned above, this is updated to reflect the suggested changes.

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
kuqin12: Sounds good. I return device_error on those instances now.

DynamicTablesPkg/Library/Acpi/Arm/AcpiIortLibArm/IortGenerator.c
kuqin12: I specifically left this one out. Is this `DeviceIdMappingIndex` being 0 really not a valid index? Or the index is 1 based and could never be 0 if there is real content in it?

pierregondois: FYIU, `DeviceIdMappingIndex != 0` implies that the `DEVICEID_VALID` flag should be set.
If `DeviceIdMappingIndex == 0` we cannot make any assumption on the `DEVICEID_VALID` flag.

kuqin12: > FYIU, `DeviceIdMappingIndex != 0` implies that the `DEVICEID_VALID` flag should be set. If `DeviceIdMappingIndex == 0` we cannot make any assumption on the `DEVICEID_VALID` flag.

I see, this implies that the assert should actually be `ASSERT (SmmuV3Node->DeviceIdMappingIndex == 0);` inside this block, correct?

kuqin12: I updated the flow with `ASSERT (SmmuV3Node->DeviceIdMappingIndex == 0);` before returning EFI_SUCCESS, which I think should serve the purpose of catching the error. Please let me know if you have any concerns or questions.

pierregondois (approved):

PR #12252, opened 2026-03-06
[RFC] Simple nested capsules as an alternative to `SignedCapsulePkg`

SergiiDmytruk:
# Description

This PR is meant to gather some feedback on whether an approach like this has a chance of landing.  I'm aware of [TianoCore RFC Process](https://www.tianocore.org/tianocore-wiki.github.io/rfc/text/0001-rfc-process.html), but this isn't a major change which is what makes it attractive.  If the changes are accepted in general, some clean up will happen after more testing, but the code is essentially complete.

The text below covers various details, but given that the code addition is quite small (if boilerplate parts are ignored) some may prefer to look at the changes in `MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleLib.c` and `MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleProcessLib.c` right away.

#### The motivation

The format of update capsules employs signing on a per-payload basis (excluding `EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER` at the start).  This means that neither embedded DXE drivers nor a capsule as a whole are covered by that signature.

The situation is fine if `FmpDxe` is part of the firmware and capsules contain nothing but a payload.  However, embedding `FmpDxe` into capsules is much more flexible as it avoids fixing how a particular firmware is updated.

A significant problem in the latter case is that `FmpDxe` stores the root key used for payload verification and making `FmpDxe` an embedded driver results in capsules validating themselves (still can work as an integrity check, but not for authentication).

Of course, embedded drivers could be taken care of by SecureBoot, but that's not the context it was designed for (e.g., it can be turned off and also uses unrelated set of keys) and it still can't address validation of a capsule as a whole which can otherwise be repacked to add/remove payloads or drivers (should be of minor concern, but still).

#### Existing solution

`SignedCapsulePkg` is one way of addressing the issue which is done by creating a firmware volume and using it as a payload in a capsule.  The volume then gets mounted and processed by a proxy FMP implementation that lives in a firmware (the real one is in a capsule).

This solution works but because `SignedCapsulePkg` is addressing many things at once (there is also a binary firmware descriptor and an INI config to specify what should be updated) using it is not that easy.  The main issue is that each firmware needs a bit of extra code and the process of building capsules is harder to integrate than a script like `GenerateCapsule`.

In other words, this approach is hard to use and it doesn't scale well, although it may be possible to circumvent this by generating the boilerplate on the fly.  Even if such practical concerns are addressed, the capsule as a whole is still subject to modifications.

#### Proposed solution

To work around a capsule not being signed as a whole some kind of a container is needed.  And since a capsule is sort of a container, why not put one capsule as a payload of another capsule?  The only extra step for the firmware is checking the signature and peeling off the outer capsule before handling the inner capsule.

That's very little extra code which can be guarded by a PCD that enforces processing only this kind of fully-signed capsules.  Capsules themselves are still of FMP type, they just have two layers to them.  And building such capsules merely requires a second invocation of `GenerateCapsule`.

**Pros**:
 - authenticates the capsule as a whole
 - quite easy to implement, manage and reason about
 - doesn't introduce any new entities, just reuses a familiar one
 - works about the same as capsules already do and older firmware will just reject new capsules
 - unlike `SignedCapsulePkg`, this is orthogonal to other facets of capsule updates and should be compatible with, for example, https://github.com/tianocore/edk2/pull/12053

**Cons**:
 - in a way, this is an abuse of the capsule format which wasn't meant to work in this manner; however, any solution which addresses shortcomings of the format will have to not "comply" with it in some way
 - as is, both inner and outer capsules need to sign payloads because `FmpDxe` rejects unsigned ones; this can be changed, but doesn't necessarily need to because inner capsule can simply use a test key with minimal overhead

#### Implementation/PoC

 - Targets integration between coreboot and EDK.  There are [scripts](https://github.com/Dasharo/coreboot/commit/db7f1badb3925ba0e1f6d5bb9de2dc2dd15b41cb) to build and test the whole thing in QEMU.
 - Uses PCD to enable the support.
 - Not consistent in its terminology:
   * mostly calls new capsules "nested", but because this doesn't specify level of nesting nor other properties; "sealed capsules" may be a better term
   * can refer to outer capsules as "top"
   * can refer to inner capsules as "nested"
 - Had to add dummy `CapsuleRootDxe` to pull in PCDs from `FmpDevicePkg`.

Here is "an illustration" of the data which may be helpful to understand the approach in more detail:

https://github.com/tianocore/edk2/blob/a4465a7a5553f143fe90f0372162aff010c89e1b/MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleLib.c#L1850-L1886

#### Building/Testing

```bash
git clone --branch up/uefi-capsules-q35-nested https://github.com/Dasharo/coreboot.git coreboot
cd coreboot
./nested-poc-build-firmware
./nested-poc-make-capsules
./nested-poc-run-qemu
```

`qemu-disk/bad.cap` is signed with keys from `other-keys/` and results in security violation error:

```
Shell> fs0:\CapsuleApp.efi fs0:\bad.cap
... reboot ...
Shell> fs0:\CapsuleApp.efi -S
CapsuleMax - CapsuleFFFF
CapsuleLast - Capsule0000
CapsuleName: Capsule0000
  Capsule Guid: 6DCBD5ED-E82D-4C44-BDA1-7194199AD92A
  Capsule ProcessedTime: 03/06/2026  19:26
  Capsule Status: Security Violation
```

`qemu-disk/good.cap` is signed with test keys (EDK uses these by default) and successfully updates firmware (variables aren't preserved, so successful status isn't reported):

```
Shell> fs0:\CapsuleApp.efi fs0:\good.cap
... reboot ...
Shell> fs0:\CapsuleApp.efi -S
CapsuleMax - CapsuleFFFF
```

#### Questions

- [ ] Breaking change?
  - No, the change is opt-in and should not cause any breakage.
- [ ] Impacts security?
  - To a degree.  This indirectly addresses potential security issues of processing capsules of unknown origin by providing means of rejecting such capsules.
- [ ] Includes tests?
  - No.

## How This Was Tested

So far in QEMU using Q35 firmware.

## Integration Instructions

* Setting `gEfiMdeModulePkgTokenSpaceGuid.PcdUseNestedFmpCapsuleFormat` to `TRUE`.
* Setting `gFmpDevicePkgTokenSpaceGuid.PcdFmpDevicePkcs7CertBufferXdr` to root key.
* Wrapping actual firmware update capsule into another one the way the first one is built.

### DIFF
--- a/MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleLib.c
+++ b/MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleLib.c
@@ -24,6 +24,7 @@
 #include <Guid/FmpCapsule.h>
 #include <Guid/SystemResourceTable.h>
 #include <Guid/EventGroup.h>
+#include <Guid/ImageAuthentication.h>
 
 #include <Library/BaseLib.h>
 #include <Library/DebugLib.h>
@@ -36,6 +37,9 @@
 #include <Library/DevicePathLib.h>
 #include <Library/UefiLib.h>
 #include <Library/BmpSupportLib.h>
+#include <Library/FmpAuthenticationLib.h>
+#include <Library/FmpPayloadLib.h>
+#include <Library/PcdLib.h>
 
 #include <Protocol/GraphicsOutput.h>
 #include <Protocol/EsrtManagement.h>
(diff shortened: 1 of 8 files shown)


SergiiDmytruk:
The labels are wrong...  Will resend the PR after CI finishes to not run it for nothing.  Also, draft PR didn't assign reviewers.

Replaced with https://github.com/tianocore/edk2/pull/12254 after addressing some issues identified by the CI.

PR #12254, opened 2026-03-07
Simple nested capsules for superior authentication

SergiiDmytruk:
## Description

This PR presents a slightly different take on capsule authentication than what's already available.  The idea is to embed one FMP capsule (inner) as a payload of another FMP capsule (outer) to take advantage of payload signing and in this way ensure authenticity of the embedded capsule as a whole with all payloads and embedded drivers in a specific order.  This approach is drastically simpler than `SignedCapsulePkg` [was](https://github.com/tianocore/edk2/pull/12537) and is convenient to use with first-stage firmware like `coreboot`.

#### Motivation

The format of update capsules employs signing on a per-payload basis (excluding `EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER` at the start).  This means that neither embedded DXE drivers nor capsule flags are covered by any signature.

The situation is fine if `FmpDxe` is part of the firmware and capsules contain nothing but a payload.  However, embedding `FmpDxe` into capsules is much more flexible as it avoids hard-coding how a particular firmware is updated.

A significant problem in the latter case is that `FmpDxe` stores the root key used for payload verification and making `FmpDxe` an embedded driver results in capsules validating themselves (still can work as an integrity check, but not for authentication).

Of course, embedded drivers could be taken care of by SecureBoot, but that's not the context it was designed for (e.g., it can be turned off and also uses unrelated set of keys) and it still can't address validation of a capsule as a whole allowing it to be repacked to add/remove payloads or drivers (should be of minor concern, but still).

#### SignedCapsulePkg

`SignedCapsulePkg` was one way of addressing the issue which was done by creating a firmware volume and using it as a payload in a capsule.  During an update a proxy FMP implementation that lived in the firmware authenticated the payload, mounted it as a volume, dispatched drivers in it and invoked real FMP implementation provided by one of those drivers.

#### Proposed solution

To work around capsules not being signed as a whole some kind of container is needed.  And since a capsule is sort of a container, why not put one capsule as a payload of another capsule?  The only extra step for the firmware is checking the signature and peeling off the outer capsule before handling the inner one.

That's very little extra code which was implemented by a new optional `CapsuleTransformation` protocol.  If an instance of the protocol is found, `DxeCapsuleLibFmp` passes each capsule through it for validation/unpacking and then processes the result.  The protocol is generic enough to have multiple implementations that do things differently, the only requirement is that capsules are of FMP type (both input and output) which is already the case.

This way capsule images are still of FMP type, they just have two layers to them and a few restrictions on the contents of the outer capsule.  Building such capsules merely requires a second invocation of `GenerateCapsule`, which is a trivial extra step.

**Pros**:
 - authenticates the capsule as a whole
 - the implementation and usage are quite straightforward
 - doesn't introduce any new entities, just reuses an existing one twice
 - works about the same as capsules already do
 - unlike `SignedCapsulePkg`, this approach is non-intrusive and is orthogonal to other facets of capsule updates and should be compatible with, for example, https://github.com/tianocore/edk2/pull/12053

**Cons**:
 - in a way, this is an abuse of the capsule format which wasn't meant to work in this manner; however, any solution which addresses shortcomings of the format will have to not "comply" with it in some way
 - at the moment both inner and outer capsules need to sign payloads because `FmpDxe` rejects unsigned ones; this can be changed, but doesn't necessarily need to because inner capsule can simply use a test key with minimal overhead

#### Implementation

Targets integration between coreboot and EDK.  There are [scripts](https://github.com/Dasharo/coreboot/commit/db7f1badb3925ba0e1f6d5bb9de2dc2dd15b41cb) to build and test the whole thing in QEMU in 3 simple steps.

The code addition is quite small (if one ignores boilerplate and moving some code around).  The most important changes are in [`MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleProcessLib.c`](https://github.com/Dasharo/edk2-upstream/blob/up/nested-capsules/MdeModulePkg/Library/DxeCapsuleLibFmp/DxeCapsuleProcessLib.c#L605-L637) where a new [`CapsuleTransformation` protocol](https://github.com/Dasharo/edk2-upstream/blob/up/nested-capsules/MdePkg/Include/Protocol/CapsuleTransformation.h) is hooked and in [`FmpDevicePkg/SealedCapsulesDxe/SealedCapsulesDxe.c`](https://github.com/Dasharo/edk2-upstream/blob/up/nested-capsules/FmpDevicePkg/SealedCapsulesDxe/SealedCapsulesDxe.c) where it's implemented.

~~Additionally, detection of a test root key has been extracted to a separate library and a DXE has been added to invoke it from the firmware in cases when `FmpDxe` is not part of the firmware itself.  Without this change not compiling `FmpDxe` into the firmware made this warning disappear.~~  Removed from this PR to make it smaller, the main functionality does not depend on these changes in any way.  Will be sent separately after this PR is handled (mainly useful combined with this PR, so not sending both at the same time).

Here is an "illustration" of outer and inner capsules which may be helpful to understand the approach in more detail:

https://github.com/tianocore/edk2/blob/88d3ec697a09ac16ad8ba91964b80c5c10137b1a/FmpDevicePkg/SealedCapsulesDxe/SealedCapsulesDxe.c#L52-L84

#### Building/Testing

```bash
git clone --branch up/uefi-capsules-q35-nested https://github.com/Dasharo/coreboot.git coreboot
cd coreboot
./nested-poc-build-firmware
./nested-poc-make-capsules
./nested-poc-run-qemu
```

`qemu-disk/bad.cap` is signed with keys from `other-keys/` and results in security violation error:

```
Shell> fs0:\CapsuleApp.efi fs0:\bad.cap
... reboot ...
Shell> fs0:\CapsuleApp.efi -S
CapsuleMax - CapsuleFFFF
CapsuleLast - Capsule0000
CapsuleName: Capsule0000
  Capsule Guid: 6DCBD5ED-E82D-4C44-BDA1-7194199AD92A
  Capsule ProcessedTime: 03/06/2026  19:26
  Capsule Status: Security Violation
```

`qemu-disk/good.cap` is signed with test keys (EDK uses these by default) and successfully updates firmware (variables aren't preserved, so successful status isn't reported):

```
Shell> fs0:\CapsuleApp.efi fs0:\good.cap
... reboot ...
Shell> fs0:\CapsuleApp.efi -S
CapsuleMax - CapsuleFFFF
```

#### Questions

- [ ] Breaking change?
  - No, the change is opt-in and should not cause any breakage.
- [ ] Impacts security?
  - In a sense that this is about authentication and processing of untrusted input.
- [ ] Includes tests?
  - No.
 
## How This Was Tested

This implementation is tested in downstream EDK2 using [OSFV](https://github.com/Dasharo/open-source-firmware-validation/blob/develop/dasharo-stability/capsule-update.robot) test suite.  Tests cover things like successful update, wrong GUID, invalid signatures and a few other aspects which are out of scope of this PR.  Robot's logs are available at https://github.com/Dasharo/dasharo-issues/issues/1443#issuecomment-4311229142 for MSI Z-690 DDR4, NovaCustom V540TND and QEMU targets.

## Integration Instructions

* Setting `gFmpDevicePkgTokenSpaceGuid.PcdFmpDevicePkcs7CertBufferXdr` to production root key(s) in the firmware.
* Setting `gFmpDevicePkgTokenSpaceGuid.PcdFmpDevicePkcs7CertBufferXdr` to any (likely just public EDK2 test key) root key for `FmpDxe`.  The exact key doesn't matter, as long as the inner capsule successfully validates itself.
* Wrapping actual firmware update capsule (signed with test key) into another one (signed with production key) the same way the first one is built.

### DIFF
--- a/FmpDevicePkg/SealedCapsulesDxe/SealedCapsulesDxe.c
+++ b/FmpDevicePkg/SealedCapsulesDxe/SealedCapsulesDxe.c
@@ -1,0 +1,412 @@
+/** @file
+  Sealed Capsules DXE driver implementing CAPSULE_TRANSFORMATION_PROTOCOL.
+
+  Caution: This module requires additional review when modified.
+  This module will have external input - capsule image.
+  This external input must be validated carefully to avoid security issues like
+  buffer overflow, integer overflow.
+
+  Copyright (c) 2026, 3mdeb Sp. z o.o. All rights reserved.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#include <PiDxe.h>
+
+#include <Guid/FmpCapsule.h>
+#include <Guid/ImageAuthentication.h>
+
+#include <Library/BaseLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/DebugLib.h>
+#include <Library/FmpAuthenticationLib.h>
+#include <Library/FmpPayloadLib.h>
+#include <Library/PcdLib.h>
+#include <Library/UefiBootServicesTableLib.h>
+
+#include <Protocol/CapsuleTransformation.h>
+#include <Protocol/FirmwareManagement.h>
+
+//
+// A sealed capsule is an FMP capsule that acts as a container for the real FMP
+// capsule.  The two-level structure is meant to reuse as much as possible of
+// the existing functionality but address the fact that FMP signs each payload
+// individually instead of a capsule as a whole.  Signing a capsule in its
+// entirety avoids the need to use a separate mechanism to authenticate embedded
+// drivers as well as prevents repacking the capsule to modify its unsigned
+// parts in any way.
+//
+// The outer capsule (container) can have no embedded drivers, no dependencies
+// and only one payload.  Its target GUID is ignored.  The only function it has
+// is authenticating its payload which is the real (inner) capsule.  For flags
+// of outer and inner capsules must match as capsule dispatching depends on
+// them.
+//
+// Following a successful authentication the outer capsule is essentially
+// discarded and the inner one is processed in its place.  This process is
+// transparent for the capsule processing mechanism which initially cares only
+// if an outer capsule looks like an FMP capsule and then sees the inner
+// capsule.
+//
+// Description of a sealed capsule in terms of structures and their properties:
+//
+//    Outer/top capsule
+//    |
+//    |-- EFI_CAPSULE_HEADER
+//    |   |-- GUID: FMP capsule
+//    |   `-- flags: same as for the inner/nested capsule
+//    |
+//    `-- EFI_FIRMWARE_MANAGEMENT_CAPSULE_HEADER
+//        |-- embedded drivers: none
+//        |-- dependencies: none
+//        `-- payloads (exactly one):
+//            |-- EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER
+//            |-- EFI_FIRMWARE_IMAGE_AUTHENTICATION
+//            |   `-- signed with root key embedded into BIOS
+//            |-- FMP_PAYLOAD_HEADER
+//            |   `-- fields aren't meaningful beyond size of payload's data
+//            `-- Inner/nested capsule
+//
+//    Inner/nested capsule
+//    |
+//    |-- EFI_CAPSULE_HEADER
+//    |   |-- GUID: FMP capsule
+//    |   `-- flags: same as for the outer/top capsule
+//    |
+//    `-- EFI_FIRMWARE_MANAGEMENT_CAPSULE_HEADER
+//        |-- embedded drivers: any
+//        |-- dependencies: any
+//        `-- payloads (any, showing just one):
+//            |-- EFI_FIRMWARE_MANAGEMENT_CAPSULE_IMAGE_HEADER
+//            |-- EFI_FIRMWARE_IMAGE_AUTHENTICATION
+//            |   `-- signed with a test key to satisfy FmpDxe
+//            |-- FMP_PAYLOAD_HEADER
+//            |   `-- fields describe system firmware to be updated
+//            `-- Firmware update image
+//
+
+/**
+  Checks if a GUID is an FMP capsule GUID or not.
+
+  @param[in] CapsuleGuid  A pointer to EFI_GUID.
+
+  @retval TRUE   It is an FMP capsule GUID.
+  @retval FALSE  It is not an FMP capsule GUID.
+**/
+STATIC
+BOOLEAN
+IsFmpCapsuleGuid (
+  IN CONST EFI_GUID  *CapsuleGuid
+  )
+{
+  return CompareGuid (&gEfiFmpCapsuleGuid, CapsuleGuid);
+}
+
+/**
+  Validate capsule header for basic sanity.
+
+  @param[in]  CapsuleHeader  Points to a capsule header.
+  @param[in]  CapsuleSize    Size of the whole capsule image.
+
+  @retval TRUE   It is a valid capsule header.
+  @retval FALSE  It is not a valid capsule header.
+**/
+STATIC
+BOOLEAN
+IsValidCapsuleHeader (
+  IN CONST EFI_CAPSULE_HEADER  *CapsuleHeader,
+  IN UINT64                    CapsuleSize
+  )
+{
+  return (CapsuleHeader->CapsuleImageSize == CapsuleSize) &&
+         (CapsuleHeader->HeaderSize < CapsuleHeader->CapsuleImageSize);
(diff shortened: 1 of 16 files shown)


SergiiDmytruk:
Extracted `DetectTestKeyLib` out of `FmpDxe` and added `DetectTestKeyDxe` to be able to use it in cases when `FmpDxe` is not included into the firmware and get warnings in the UI (the text of was also updated).  This has also allowed to get rid of `CapsuleRootDxe`.

SergiiDmytruk:
Slightly updated description and title now that SignedCapsulePkg is gone.

> Extracted `DetectTestKeyLib` out of `FmpDxe` and added `DetectTestKeyDxe` to be able to use it in cases when `FmpDxe` is not included into the firmware and get warnings in the UI (the text of was also updated).

Removed those commits from this PR to make it smaller and concentrate on a single area.  They will be sent separately.

"FmpDevicePkg/FmpPayloadLib: Add" commit may also be unnecessary with some changes to the code, keeping it so far in case `SealedCapsulesDxe` will be moved outside of `FmpDevicePkg`.

mdkinney:
@spbrogan This proposal seems to address challenges with use of embedded drivers.  Do you have any feedback on the nesting of capsules?  Seems allowed by the UEFI spec since a capsule payload can be anything.

I do see a new protocol is added to MdePkg.  Can this is moved to FmpDevicePkg or MdeModulePkg?  MdePkg content is for industry standard specs, and this idea of nested capsules for now is an edk2 implementation feature.  Adding to MdePkg would require a UEFI Spec change.

SergiiDmytruk:
> I do see a new protocol is added to MdePkg. Can this is moved to FmpDevicePkg or MdeModulePkg? MdePkg content is for industry standard specs, and this idea of nested capsules for now is an edk2 implementation feature. Adding to MdePkg would require a UEFI Spec change.

Moved it to `MdeModulePkg`, `MdePkg` was a wrong choice.  The protocol can't be defined in `FmpDevicePkg` as that won't pass CI checks configured in https://github.com/tianocore/edk2/blob/eeaac552f81111014ab77f8f6e0dea00e03b3c46/MdeModulePkg/MdeModulePkg.ci.yaml#L53-L67

The protocol was specifically introduced to satisfy CI by turning build-time dependency `MdeModulePkg -> FmpDevicePkg` into a run-time one.

FmpDevicePkg/SealedCapsulesDxe/SealedCapsulesDxe.c
```diff
@@ -0,0 +1,412 @@
...
+  UINTN                                         InnerCapsuleSize;
+  UINTN                                         OuterCapsulePayloadSize;
+
+  if ((CapsuleHeader == NULL) || (*CapsuleHeader == NULL)) {
+    DEBUG ((DEBUG_ERROR, "%a(): input pointer is NULL.\n", __func__));
+    return EFI_INVALID_PARAMETER;
+  }
+
+  OuterCapsuleHeader = *CapsuleHeader;
+  if (OuterCapsuleHeader->HeaderSize >= OuterCapsuleHeader->CapsuleImageSize) {
```
spbrogan: shouldn't you also check that headersize is large enough for your assumed header contents? On line 271 you check the guid but how do you know that memory is valid and/or part of the capsule?

FmpDevicePkg/SealedCapsulesDxe/SealedCapsulesDxe.c
```diff
@@ -0,0 +1,412 @@
...
+    PublicKeyDataXdr  = (UINT8 *)ALIGN_POINTER (PublicKeyDataXdr, sizeof (UINT32));
+  }
+
+  return EFI_SECURITY_VIOLATION;
+}
+
+/**
+  Optionally transform a capsule out of an outer one.
+
+  Caution: This function may receive untrusted input.
```
spbrogan: I think more checks are needed.

FmpDevicePkg/SealedCapsulesDxe/SealedCapsulesDxe.c
```diff
@@ -0,0 +1,412 @@
...
+    DEBUG ((DEBUG_ERROR, "%a(): multiple payloads in the outer capsule.\n", __func__));
+    return EFI_UNSUPPORTED;
+  }
+
+  if (OuterFmpHeader->EmbeddedDriverCount != 0) {
+    DEBUG ((DEBUG_ERROR, "%a(): embedded driver inside the outer capsule.\n", __func__));
+    return EFI_UNSUPPORTED;
+  }
+
+  OuterOffsetList  = (UINT64 *)&OuterFmpHeader[1];
```
spbrogan: Is this safe?  are there size fields you can reference?

PR #12257, opened 2026-03-08
SecurityPkg/Tcg2Smm: harden NVS/SMI state initialization

20000419:
**Description of the change and why this change was made:**

This patch proactively hardens the initialization state of NVS and SMI variables in `Tcg2Smm` to prevent potential SMM crashes (DoS) or the exposure of undefined SMI values.

1. **SMI Value Initialization**: Initializes `mPpSoftwareSmi` and `mMcSoftwareSmi` to `MAX_UINTN`. While UEFI implicitly initializes global variables to `0`, `0` can actually be a valid and routable Software SMI number on certain platforms. Using `MAX_UINTN` acts as a clear sentinel value for an uninitialized state, preventing accidental zero-SMI execution.

2. **Defensive NULL Check**: Adds a defensive `NULL` check for `mTcgNvs` at the beginning of `PhysicalPresenceCallback()`. If a Software SMI is triggered maliciously or unexpectedly before the MM/DXE NVS exchange completes, this prevents an unsafe null pointer dereference that would cause an SMM page fault.

3. **Reinitialization**: Ensures `mPpSoftwareSmi` and `mMcSoftwareSmi` are explicitly reset to `MAX_UINTN` during `InitializeTcgCommon()` to guarantee a consistent startup state.

**Describe the test(s) that were run to verify the changes:**

* Ran `git diff --check` to ensure no whitespace or formatting errors.
* Verified code logic via static analysis/code review, confirming that the defensive checks correctly complement existing SMM communication validation paths.
* Verified successful compilation of `SecurityPkg` with this patch applied.

**Describe how these changes should be integrated:**

N/A

**Checklist:**

* [x] BaseTools/Scripts/SetupGit.py executed successfully
* [x] BaseTools/Scripts/PatchCheck.py executed successfully
* [x] Code complies with the EDK II coding standard
* [x] Commit message complies with the EDK II commit message rules


jyao1:
> Initialize mPpSoftwareSmi and mMcSoftwareSmi to MAX_UINTN to avoid exposing uninitialized SMI values.

In UEFI, the PE LOADER will initialize the global value to zero. It is implicit initialization.
Would you please clarify, if there is any need to change the initial value from `0` to `MAX_UINTN`?

> which could cause an SMM crash/DoS or expose undefined SMI values.

Have you observed or created the case that caused `SMM crash/DoS or expose undefined SMI values`?
If so, would you please share the reproduce steps?

20000419:
> > Initialize mPpSoftwareSmi and mMcSoftwareSmi to MAX_UINTN to avoid exposing uninitialized SMI values.
> 
> In UEFI, the PE LOADER will initialize the global value to zero. It is implicit initialization. Would you please clarify, if there is any need to change the initial value from `0` to `MAX_UINTN`?
> 
> > which could cause an SMM crash/DoS or expose undefined SMI values.
> 
> Have you observed or created the case that caused `SMM crash/DoS or expose undefined SMI values`? If so, would you please share the reproduce steps?

Hi @jyao1, thank you for the review! Here are the clarifications:

> **1. Would you please clarify, if there is any need to change the initial value from 0 to MAX_UINTN?**

While the PE loader implicitly initializes global variables to zero, `0` can actually be a valid and routable Software SMI number on many platforms. If the SMI variables are evaluated or registered before the NVS exchange dynamically assigns the correct values, the system might accidentally treat `0` as the intended SMI number. 
Initializing them to `MAX_UINTN` explicitly sets a sentinel value that is clearly invalid, preventing any accidental zero-SMI execution before the proper initialization is completed.

> **2. Have you observed or created the case that caused SMM crash/DoS or expose undefined SMI values? If so, would you please share the reproduce steps?**

I have not dynamically reproduced an exploit or a crash for this. This patch is proposed as a **proactive, defensive hardening measure** based on code review. 
The theoretical vulnerability window exists here: if a malicious actor or an unexpected OS behavior triggers the Physical Presence Software SMI *before* `mTcgNvs` is properly populated via the MM/DXE communication, `PhysicalPresenceCallback()` will be invoked. Without the check, it will dereference the NULL `mTcgNvs` pointer, which would lead to an SMM page fault and consequently a system DoS. The NULL check and the `MAX_UINTN` initialization safely close this theoretical window.

jyao1:
> > > Initialize mPpSoftwareSmi and mMcSoftwareSmi to MAX_UINTN to avoid exposing uninitialized SMI values.
> > 
> > 
> > In UEFI, the PE LOADER will initialize the global value to zero. It is implicit initialization. Would you please clarify, if there is any need to change the initial value from `0` to `MAX_UINTN`?
> > > which could cause an SMM crash/DoS or expose undefined SMI values.
> > 
> > 
> > Have you observed or created the case that caused `SMM crash/DoS or expose undefined SMI values`? If so, would you please share the reproduce steps?
> 
> Hi @jyao1, thank you for the review! Here are the clarifications:
> 
> > **1. Would you please clarify, if there is any need to change the initial value from 0 to MAX_UINTN?**
> 
> While the PE loader implicitly initializes global variables to zero, `0` can actually be a valid and routable Software SMI number on many platforms. If the SMI variables are evaluated or registered before the NVS exchange dynamically assigns the correct values, the system might accidentally treat `0` as the intended SMI number. Initializing them to `MAX_UINTN` explicitly sets a sentinel value that is clearly invalid, preventing any accidental zero-SMI execution before the proper initialization is completed.

Thanks. Makes sense.

> 
> > **2. Have you observed or created the case that caused SMM crash/DoS or expose undefined SMI values? If so, would you please share the reproduce steps?**
> 
> I have not dynamically reproduced an exploit or a crash for this. This patch is proposed as a **proactive, defensive hardening measure** based on code review. The theoretical vulnerability window exists here: if a malicious actor or an unexpected OS behavior triggers the Physical Presence Software SMI _before_ `mTcgNvs` is properly populated via the MM/DXE communication, `PhysicalPresenceCallback()` will be invoked. Without the check, it will dereference the NULL `mTcgNvs` pointer, which would lead to an SMM page fault and consequently a system DoS. The NULL check and the `MAX_UINTN` initialization safely close this theoretical window.

OK. Let's talk about theoretical vulnerability. "if a malicious actor or an unexpected OS behavior triggers the Physical Presence Software SMI _before_ `mTcgNvs` is properly populated via the MM/DXE communication".

According to our threat model, any work before EndOfDxe/ReadyToLock is trusted. The work after EndOfDxe/ReadyToLock is untrust. 

`mTcgNvs` is initialized at `TpmNvsCommunciate()`, which is triggered at `mReadyToLockHandle` event.
That means `mTcgNvs` SHALL be initialized correctly by the trusted code.

Then how could if "a malicious actor or an unexpected OS behavior" triggers the Physical Presence Software SMI _before_ `mTcgNvs` is properly populated?

If you believe "a malicious actor or an unexpected OS behavior" can do something before mReadyToLockHandle, then this action already violates the Threat Model we have defined.

Would you please clarify, what is the your threat model?

20000419:
> > > > Initialize mPpSoftwareSmi and mMcSoftwareSmi to MAX_UINTN to avoid exposing uninitialized SMI values.
> > > 
> > > 
> > > In UEFI, the PE LOADER will initialize the global value to zero. It is implicit initialization. Would you please clarify, if there is any need to change the initial value from  to ?`0``MAX_UINTN`
> > > > which could cause an SMM crash/DoS or expose undefined SMI values.
> > > 
> > > 
> > > Have you observed or created the case that caused ? If so, would you please share the reproduce steps?`SMM crash/DoS or expose undefined SMI values`
> > 
> > 
> > Hi @jyao1, thank you for the review! Here are the clarifications:
> > > **1. Would you please clarify, if there is any need to change the initial value from 0 to MAX_UINTN?**
> > 
> > 
> > While the PE loader implicitly initializes global variables to zero,  can actually be a valid and routable Software SMI number on many platforms. If the SMI variables are evaluated or registered before the NVS exchange dynamically assigns the correct values, the system might accidentally treat  as the intended SMI number. Initializing them to  explicitly sets a sentinel value that is clearly invalid, preventing any accidental zero-SMI execution before the proper initialization is completed.`0``0``MAX_UINTN`
> 
> Thanks. Makes sense.
> 
> > > **2. Have you observed or created the case that caused SMM crash/DoS or expose undefined SMI values? If so, would you please share the reproduce steps?**
> > 
> > 
> > I have not dynamically reproduced an exploit or a crash for this. This patch is proposed as a **proactive, defensive hardening measure** based on code review. The theoretical vulnerability window exists here: if a malicious actor or an unexpected OS behavior triggers the Physical Presence Software SMI _before_  is properly populated via the MM/DXE communication,  will be invoked. Without the check, it will dereference the NULL  pointer, which would lead to an SMM page fault and consequently a system DoS. The NULL check and the  initialization safely close this theoretical window.`mTcgNvs``PhysicalPresenceCallback()``mTcgNvs``MAX_UINTN`
> 
> OK. Let's talk about theoretical vulnerability. "if a malicious actor or an unexpected OS behavior triggers the Physical Presence Software SMI _before_ is properly populated via the MM/DXE communication".`mTcgNvs`
> 
> According to our threat model, any work before EndOfDxe/ReadyToLock is trusted. The work after EndOfDxe/ReadyToLock is untrust.
> 
> `mTcgNvs` is initialized at , which is triggered at event. That means SHALL be initialized correctly by the trusted code.`TpmNvsCommunciate()``mReadyToLockHandle``mTcgNvs`
> 
> Then how could if "a malicious actor or an unexpected OS behavior" triggers the Physical Presence Software SMI _before_ is properly populated?`mTcgNvs`
> 
> If you believe "a malicious actor or an unexpected OS behavior" can do something before mReadyToLockHandle, then this action already violates the Threat Model we have defined.
> 
> Would you please clarify, what is the your threat model?

Hi @jyao1, thank you for pointing this out. You are absolutely right. I apologize for the imprecise wording in my previous explanation regarding a "malicious actor" before ReadyToLock. I completely agree that according to the EDK II threat model, everything before ReadyToLock is considered part of the Trusted Computing Base (TCB).

My intention with the mTcgNvs == NULL check is not to defend against an OS-level attacker, but rather to apply general defensive programming for system robustness. Since the Software SMI handler is registered and becomes "live" before mTcgNvs is populated, a buggy (though trusted) DXE driver, a flawed 3rd-party Option ROM, or an accidental IO port write could theoretically trigger this SMI prematurely. Without the check, this would lead to a NULL pointer dereference inside SMM and cause a hard system crash.

However, if you believe this kind of defensive check against trusted code adds unnecessary overhead or deviates from the standard EDK II design philosophy, I am perfectly fine with dropping the mTcgNvs == NULL check and only keeping the MAX_UINTN initialization (which addresses the valid 0 SMI routing issue).

Please let me know your preference, and I will update the patch accordingly!

jyao1:
ok. thanks to clarify. My intension is to check if we miss the check in some other places, besides this one.

PR #12259, opened 2026-03-08
MdeModulePkg/Variable: Avoid double VA conversion of FVB protocol

lizthegrey:
## Summary

The Variable driver and the FVB protocol producer both register
SetVirtualAddressMap callbacks, but the delivery order is unspecified.
If both convert the FVB function pointers, they get converted twice
and runtime variable access crashes.

This affects any ARM platform using a non-flash FVB with StandaloneMm
(e.g., OP-TEE + EepromFvb), where the FVB producer performs its own
SetVirtualAddressMap conversion.

The fix saves a shadow copy of the FVB protocol at discovery time and
converts only the shadow during SetVirtualAddressMap. On the first
runtime variable access, if the FVB producer hasn't performed its own
conversion yet, the converted pointers are swapped into the live
protocol.

Originally written by Ard Biesheuvel for SolidRun's edk2 fork
(c5ee0556b6). Ported to current master and tested on an NXP LX2160A
(HoneyComb LX2K) with OP-TEE 4.9.0 StandaloneMm providing Secure
Boot variable storage.

## Test plan

- Verified on NXP LX2160A (SolidRun HoneyComb LX2K) with:
  - ATF v2.12, OP-TEE 4.9.0, edk2-stable202602
  - StandaloneMm-backed EepromFvb for Secure Boot variable storage
  - Runtime variable access (GetVariable/SetVariable) works correctly
    after ExitBootServices
- Without the fix, runtime variable access crashes due to double
  pointer conversion

Cc: @ardbiesheuvel, @jnettlet

### DIFF
--- a/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableDxe.c
+++ b/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableDxe.c
@@ -43,6 +43,8 @@ EDKII_VAR_CHECK_PROTOCOL        mVarCheck = {
   VarCheckVariablePropertyGet
 };
 
+STATIC EFI_FIRMWARE_VOLUME_BLOCK_PROTOCOL  mFvbProtocolShadow;
+
 /**
   Some Secure Boot Policy Variable may update following other variable changes(SecureBoot follows PK change, etc).
   Record their initial State when variable write service is ready.
@@ -109,8 +111,28 @@ AcquireLockOnlyAtBootTime (
   IN EFI_LOCK  *Lock
   )
 {
+  EFI_FIRMWARE_VOLUME_BLOCK_PROTOCOL  *FvbInstance;
+
   if (!AtRuntime ()) {
     EfiAcquireLock (Lock);
+  } else {
+    //
+    // Check whether we need to swap in the converted pointer values for the
+    // FVB protocol methods. This handles the case where the FVB producer did
+    // not convert its own pointers during SetVirtualAddressMap.
+    //
+    FvbInstance = mVariableModuleGlobal->FvbInstance;
+    if ((FvbInstance != NULL) &&
+        (FvbInstance->GetBlockSize != mFvbProtocolShadow.GetBlockSize))
+    {
+      FvbInstance->GetBlockSize       = mFvbProtocolShadow.GetBlockSize;
+      FvbInstance->GetPhysicalAddress = mFvbProtocolShadow.GetPhysicalAddress;
+      FvbInstance->GetAttributes      = mFvbProtocolShadow.GetAttributes;
+      FvbInstance->SetAttributes      = mFvbProtocolShadow.SetAttributes;
+      FvbInstance->Read               = mFvbProtocolShadow.Read;
+      FvbInstance->Write              = mFvbProtocolShadow.Write;
+      FvbInstance->EraseBlocks        = mFvbProtocolShadow.EraseBlocks;
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


ardbiesheuvel:
@apalos

apalos:
> ## Summary
> The Variable driver and the FVB protocol producer both register SetVirtualAddressMap callbacks, but the delivery order is unspecified. If both convert the FVB function pointers, they get converted twice and runtime variable access crashes.
> 
> This affects any ARM platform using a non-flash FVB with StandaloneMm (e.g., OP-TEE + EepromFvb), where the FVB producer performs its own SetVirtualAddressMap conversion.
> 
> The fix saves a shadow copy of the FVB protocol at discovery time and converts only the shadow during SetVirtualAddressMap. On the first runtime variable access, if the FVB producer hasn't performed its own conversion yet, the converted pointers are swapped into the live protocol.
> 
> Originally written by Ard Biesheuvel for SolidRun's edk2 fork ([c5ee055](https://github.com/tianocore/edk2/commit/c5ee0556b64d5427e9b711cff60eb540091d9bf8)). Ported to current master and tested on an NXP LX2160A (HoneyComb LX2K) with OP-TEE 4.9.0 StandaloneMm providing Secure Boot variable storage.
> 
> ## Test plan
> * Verified on NXP LX2160A (SolidRun HoneyComb LX2K) with:
>   
>   * ATF v2.12, OP-TEE 4.9.0, edk2-stable202602
>   * StandaloneMm-backed EepromFvb for Secure Boot variable storage
>   * Runtime variable access (GetVariable/SetVariable) works correctly
>     after ExitBootServices
> * Without the fix, runtime variable access crashes due to double
>   pointer conversion
> 
> Cc: @ardbiesheuvel, @jnettlet Cc: Liming Gao [gaoliming@byosoft.com.cn](mailto:gaoliming@byosoft.com.cn)

@lizthegrey apologies for the late response was traveling. This only happens when the backing storage is the i2c (iirc) connected EEPROM right? There's also the possibility to use an RPMB with OP-TEE, but in that case the runtime calls are overwritten by the linux kernel. I also don't remember seeing any crashes when the backing storage is memory only

lizthegrey:
> @lizthegrey apologies for the late response was traveling. This only happens when the backing storage is the i2c (iirc) connected EEPROM right? There's also the possibility to use an RPMB with OP-TEE, but in that case the runtime calls are overwritten by the linux kernel. I also don't remember seeing any crashes when the backing storage is memory only

correct, this only affects the LX2160A platform as deployed by the Solidrun Honeycomb and Clearfog boards, which have EEPROM.

jnettlet:
Just to be clear about the bug that is triggered. This bug could exist under any runtime driver, it just so happens that on the LX2160a the second remapping ends up putting the pointer address in a memory space region that is allocated to the networking complex. The decision that needs to be made is what is the incorrect behavior that needs to be fixed. Is the double remapping of the of the virtual address incorrect, or the fact that we don't take into account mapping the memory into a region of memory that should be reserved.

I don't think using alternative storage options as a workaround makes a lot of sense, since it is a band-aid to a design problem that unfortunately seems to only be a problem on the LX2160a platform.

apalos:
> Just to be clear about the bug that is triggered. This bug could exist under any runtime driver, it just so happens that on the LX2160a the second remapping ends up putting the pointer address in a memory space region that is allocated to the networking complex. The decision that needs to be made is what is the incorrect behavior that needs to be fixed. Is the double remapping of the of the virtual address incorrect, or the fact that we don't take into account mapping the memory into a region of memory that should be reserved.
>

Both of them sound wrong :).
@ardbiesheuvel explained why the double mapping exists on the original  patchset. I do agree with Ard that, unfortunately we can't change the current behavior... @lizthegrey can you please tweak the commit message a bit and include the historical piece of info?
 
> I don't think using alternative storage options as a workaround makes a lot of sense, since it is a band-aid to a design problem that unfortunately seems to only be a problem on the LX2160a platform.

Yea, I wasn't suggesting using the RPMB. I was just trying to understand if it's reproducible in other cases.

lizthegrey:
> Both of them sound wrong :). @ardbiesheuvel explained why the double mapping exists on the original patchset. I do agree with Ard that, unfortunately we can't change the current behavior... @lizthegrey can you please tweak the commit message a bit and include the historical piece of info?

Updated.

jnettlet:
Originally I had submitted a patch that changed how this was handled based on a variable setting. I figured the least intrusive way to handle this issue was with a board specific mechanism that can tell the variable storage driver how to handle the mapping of the backend. That proposal was rejected and Ard provided the alternative patch, which was also rejected. If we don't want to use a UUID to specify the required mapping behavior then the other alternative is a build definition.

-Jon
________________________________

apalos:
> Originally I had submitted a patch that changed how this was handled based on a variable setting. I figured the least intrusive way to handle this issue was with a board specific mechanism that can tell the variable storage driver how to handle the mapping of the backend. That proposal was rejected and Ard provided the alternative patch, which was also rejected. If we don't want to use a UUID to specify the required mapping behavior then the other alternative is a build definition.

Thanks Jon. That was actually my next question. Having a config option would make it easier for future drivers to do the right thing and slowly deprecate the built in option slowly. Do you remember why that got rejected?

jnettlet:
I am pretty sure it was deemed an unnecessary usage of a UUID, but I would have to find that conversation.  It was quite a few years ago at this point.
________________________________

niruiyu (approved):

ardbiesheuvel (approved):

ardbiesheuvel:
@mergifyio requeue

lizthegrey:
Thanks much!

PR #12260, opened 2026-03-08
UefiPayloadPkg: align DXE images for page protections

Sean-StarLabs:
Align payload DXE images to 4 KiB boundaries so the DXE core can apply its
normal image protection policy to those images.

This keeps the payload in the same direction as other edk2 platforms that
opt into NX-capable image layout, without relying on a coreboot-specific
toolchain selector.

### DIFF
--- a/UefiPayloadPkg/UefiPayloadPkg.dsc
+++ b/UefiPayloadPkg/UefiPayloadPkg.dsc
@@ -200,6 +200,16 @@
   CLANGPDB:*_*_*_DLINK_FLAGS = /ALIGN:4096
   MSFT:*_*_*_DLINK_FLAGS     = /ALIGN:4096
 
+# Force PE/COFF sections to be aligned at 4KB boundaries so DXE Core can
+# apply page-level image protection, including NX, to IA32/X64 DXE images.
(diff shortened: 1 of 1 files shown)


benjamindoron:
I don't think this will work: as of https://github.com/coreboot/coreboot/commit/9e5234feeed8760ee01d088f9cfc8cfda8cff8ab, coreboot no longer overrides the tools_def.txt file, and so there is no longer a `COREBOOT` toolchain. And the first specifier to these argument lines is a toolchain. What you might be thinking of is:
```
!if $(BOOTLOADER) == COREBOOT
*_*_*_DLINK_FLAGS      = -z common-page-size=0x1000
!endif
```

But if we want this for UefiPayload on coreboot, then SBL should get it too.

Also, what this is really about is whether the DXE core applies its image protection policy to the modules, and we've had that conversation before (page aligned sections means that you can give the sections different entries in the page table. Theoretically, you could probably just align the section sizes up to the next page, but it's just not like that, and section alignment == page alignment has come to mean opting into NX in UEFI). For some reason, those commits keep getting abandoned.

I say that if it works (now), then this commit should be talking about how we want image protection, that it works, and then we'll wait to see if it breaks for anybody. If they're in the minority, they can have an override, and this will go in.

benjamindoron (changes requested):
(there is no COREBOOT toolchain, and this should be a larger discussion. See the comment above)

benjamindoron:
This looks better, thanks. If this works for coreboot and SBL, I vote to get this PR in (in general).

Could you apply the same to AARCH64 to keep the behaviour across architectures in sync (note that their page size is apparently larger)? I don't like the way we diverge already on unnecessary (read: non-architectural) PCDs, libraries and drivers, and I've got some changes staged, and others I want to work on sometime, to get them back together, but I'd rather not diverge further.

Also, note that only UefiPayloadEntry (SEC) can be IA32. Everything from the DXE core and onwards will always be 64-bit.

PR #12262, opened 2026-03-09
MdeModulePkg/PeiCore: Accept MemoryDiscoveredPpi from SecCore

niruiyu:
# Description

PeiCore assumes that it runs initially from the temporary memory and switches to physical memory after PS::InstallPeiMemory() reports the physical memory range for PEI.
This change enables PeiCore to run initially from the physical memory when MemoryDiscoveredPpi is passed from SEC.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Verify the PeiCore can run to DxeIpl->Entry() without assertion
`ASSERT (PrivateData.PeiMemoryInstalled == TRUE);`  in PeiCore().

## Integration Instructions
N/A


MdeModulePkg/Core/Pei/Ppi/Ppi.c
```diff
@@ -1052,6 +1053,15 @@ ProcessPpiListFromSec (
       //
       Status = InternalPeiInstallPpi (PeiServices, PpiList, TRUE);
       ASSERT_EFI_ERROR (Status);
+
+      //
+      // Set the PeiMemoryInstalled flag if SEC passes in MemoryDiscoveredPpi to support
+      // PeiCore runs initially from physical memory.
+      //
+      if (CompareGuid (PpiList->Guid, &gEfiPeiMemoryDiscoveredPpiGuid)) {
```
khaliid2040, line 1061: I think you should update InstallPeiMemory() too, to account if permanent memory already present. Currently InstallPeiMemory() if it is called after a physical memory discovered it will assert because InstallPeiMemory () doesn't want to be called when there is already installed physical memory (In this case SEC discovered physical memory).

This will trigger assertion in InstallPeiMemory() if a PEIM calls that function later during dispatch because SecCore discovered memory is present. My suggestion  is to remove assert so it returns if physical memory is already installed (i.e the SeCore discovered physical memory in this case).

MdeModulePkg/Core/Pei/Ppi/Ppi.c
```diff
@@ -1052,6 +1053,15 @@ ProcessPpiListFromSec (
       //
       Status = InternalPeiInstallPpi (PeiServices, PpiList, TRUE);
       ASSERT_EFI_ERROR (Status);
+
+      //
+      // Set the PeiMemoryInstalled flag if SEC passes in MemoryDiscoveredPpi to support
+      // PeiCore runs initially from physical memory.
+      //
```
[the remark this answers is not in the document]
khaliid2040: I would suggest you this comment instead:

```C
//
// Set the PeiMemoryInstalled flag if SEC passed MemoryDiscoveredPpi, so PeiCore initially get 
// loaded into physical memory.
//
```

MdeModulePkg/Core/Pei/Ppi/Ppi.c
niruiyu: Calling InstallPeiMemory is NOT expected for a PEI Core which already running in a physical-memory env. That assertion should be kept.

MdeModulePkg/Core/Pei/Ppi/Ppi.c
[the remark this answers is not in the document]
niruiyu: You comments are good. But it does not emphasize that the whole PEI is running in a physical-memory ready env.

How about follows?

```c
//
// Set the PeiMemoryInstalled flag if SEC passed MemoryDiscoveredPpi, so the PEI runs in
// physical memory.
//

MdeModulePkg/Core/Pei/Ppi/Ppi.c
khaliid2040: What i am curious is, wouldn't this change impact later the PEIM that is dasignated to discover the physical memory. PEI core is already aware whether it is running on physical memory or not, but are the PEIMs aware. I think This change may impact the PEIM that will discover physical memory later on.

niruiyu: The scope of the change is to only support the scenario that partial or whole physical memory is discovered before PeiCore, and a certain range of the physical memory is passed to PEI Core to be used by the PEI environment.
There might be more physical memory that will be discovered by PEIMs or DXE drivers. But these new physical memories will NOT be used by the PEI environment. They will be reported out by PEIMs or DXE drivers through resource HOBs or gDS->AddMemory() calls.

khaliid2040: Thanks for the explanations, i didn't know multiple PEIMs could discover physical memory with different ranges and report it. I was expecting one time discovery by the whole physical memory, which is where all my questions pointing out.

You could resolve the thread now.

PR #12264, opened 2026-03-09
Support AMD GetProcessorCoreType

jiaxinwu:
# Description

The series patches are to support the AMD CoreType in the GetProcessorCoreType ()
function. With the change, MP HOB will record the correct CoreType for AMD processors.

## How This Was Tested

Reviewed register bit layout against AMD64 Architecture Programmer's Manual. Tested on
the AMD real silicon. The value is reported correctly.

## Integration Instructions

No need Integration.

### DIFF
--- a/UefiCpuPkg/CpuMpPei/CpuMpPei.c
+++ b/UefiCpuPkg/CpuMpPei/CpuMpPei.c
@@ -9,6 +9,8 @@
 
 #include "CpuMp2Pei.h"
 
+#include <Library/CpuLib.h>
+
 EFI_PEI_PPI_DESCRIPTOR  mPeiCpuMpPpiList[] = {
   {
     EFI_PEI_PPI_DESCRIPTOR_PPI,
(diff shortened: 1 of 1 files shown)


jiaxinwu:
@niruiyu @lzeng14 @lgao4 , can you please review this change?

MdePkg/Include/Register/Amd/Cpuid.h
```diff
@@ -59,9 +59,147 @@ CPUID Signature Information
...
+    ///
+    /// [Bits 31:0] Extended APIC ID of the logical processor.
+    ///
+    UINT32    ExtendedApicId : 32;
+  } Bits;
+  ///
+  /// All bit fields as a 32-bit value
+  ///
+  UINT32    Uint32;
+} AMD_CPUID_EXTENDED_TOPOLOGY_EDX;
```
niruiyu: If the whole 32bits are only for one purpose - ExtendedApicId, can you consider the following 2 options:
1. Remove this structure definition
2. `typedef UINT32 AMD_CPUID_EXTENDED_TOPOLOGY_EDX`

I think it's simpler.

UefiCpuPkg/CpuMpPei/CpuMpPei.c
```diff
@@ -220,10 +236,15 @@ BuildMpInformationHob (
...
-  // Get Processors CoreType
+  // Get Processors CoreType.
+  // For Intel Hybrid CPUs, use standard CPUID leaf 0x1A (CPUID_HYBRID_INFORMATION).
+  // For AMD CPUs, use extended CPUID leaf 0x80000026 (AMD_CPUID_EXTENDED_CPU_TOPOLOGY).
   //
   AsmCpuid (CPUID_SIGNATURE, &CpuidMaxInput, NULL, NULL, NULL);
-  if (CpuidMaxInput >= CPUID_HYBRID_INFORMATION) {
+  AsmCpuid (CPUID_EXTENDED_FUNCTION, &CpuidMaxExtInput, NULL, NULL, NULL);
+  if ((CpuidMaxInput >= CPUID_HYBRID_INFORMATION) ||
+      (StandardSignatureIsAuthenticAMD () && (CpuidMaxExtInput >= AMD_CPUID_EXTENDED_TOPOLOGY)))
```
niruiyu: What if the code runs in AMD CPU and `CpuidMaxInput >= CPUID_HYBRID_INFORMATION` is a success?
Because 0x1A < 0x80000026 so the CoreTypes is allocated and StartupAllCPUs is called to query coretype for each core.
Then it ends up with Cpuid (0x80000026) call in AMD CPU but that specific AMD CPU may not support this leaf.

UefiCpuPkg/CpuMpPei/CpuMpPei.c
```diff
@@ -179,14 +181,27 @@ GetProcessorCoreType (
...
 
   CoreTypes = (UINT8 *)Buffer;
-  AsmCpuidEx (CPUID_HYBRID_INFORMATION, CPUID_HYBRID_INFORMATION_MAIN_LEAF, &NativeModelIdAndCoreTypeEax.Uint32, NULL, NULL, NULL);
-  CoreTypes[ProcessorIndex] = (UINT8)NativeModelIdAndCoreTypeEax.Bits.CoreType;
+
+  if (StandardSignatureIsAuthenticAMD ()) {
+    //
+    // AMD: CPUID_Fn80000026_EBX[31:28] = CoreType
+    //
+    AsmCpuidEx (AMD_CPUID_EXTENDED_TOPOLOGY, 0, NULL, &Ebx.Uint32, NULL, NULL);
```
lzeng14: Is it guaranteed that leaf 0 (ecx input 0) will return Core LevelType as CoreType is "Only valid for LevelType = 1h (Core)" ?

MdePkg/Include/Register/Amd/Cpuid.h
jiaxinwu: agree, fixed by remove it.

UefiCpuPkg/CpuMpPei/CpuMpPei.c
jiaxinwu: agree, fixed

UefiCpuPkg/CpuMpPei/CpuMpPei.c
jiaxinwu: agree, fixed

PR #12267, opened 2026-03-09
ArmPkg: Fix incorrect bit width mask for SP_EL0

cfernald:
The current code uses `0xFUL` for stack alignment. Since this is only specified to be an unsigned long, it may only be translated to 0xFFFFFFF0. If this occurs and the stack resides above 4GB, the high bits of the pointer will be truncated. This will cause a data abort immediately when pushing to the exception stack.

This has been observed to specifically cause a failure with the ClangPDB tool chain.

This commit simply changes the mask to `0xFULL` to ensure that the full 64-bit pointer is correctly masked for alignment.

# Description

The current code uses `0xFUL` for stack alignment. Since this is only specified to be an unsigned long, it may only be translated to 0xFFFFFFF0. If this occurs and the stack resides above 4GB, the high bits of the pointer will be truncated. This will cause a data abort immediately when pushing to the exception stack.

This commit simply changes the mask to `0xFULL` to ensure that the full 64-bit pointer is correctly masked for alignment.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on a QEMU SBSA platform, using ClangPDB tool chain

## Integration Instructions

N/A


leiflindholm:
The fix is correct, but it would be worth mentioning in description/commit message that this is an LP64/LLP64 issue.

cfernald:
> The fix is correct, but it would be worth mentioning in description/commit message that this is an LP64/LLP64 issue.

Good point, clarified in the commit message.

PR #12270, opened 2026-03-10
Add CompilerIntrinsicsLib to BasePrintLib.inf.

jaykrell:
# Description
This is to fix:
```
 /usr/lib/aarch64-linux-gnu/bin/ld: /tmp/cczqHYdu.ltrans0.ltrans.o: in function `UnicodeVSPrint':
 ...MdePkg/Library/BasePrintLib/PrintLib.c:73:(.text.UnicodeVSPrint+0x48): undefined reference to `memcpy'
```
```
  return BasePrintLibSPrintMarker ((CHAR8 *)StartOfBuffer, BufferSize >> 1, FORMAT_UNICODE | OUTPUT_UNICODE, (CHAR8 *)FormatString, Marker, NULL);
```
presumably passing `VA_LIST Marker` by value.

Maybe EDK2 should avoid copying struct/union/va_list, or passing/returning by value?

- [x] Breaking change?
  - I think so, maybe, it breaks people's builds, that do not have CompilerIntrinsicsLib in *.dsc?
- [ ] Impacts security?
  - No.
- [ ] Includes tests?
  - No.

## How This Was Tested

Local builds.

## Integration Instructions

N/A?

### DIFF
--- a/MdePkg/Library/BasePrintLib/BasePrintLib.inf
+++ b/MdePkg/Library/BasePrintLib/BasePrintLib.inf
@@ -34,6 +34,7 @@
 
 
 [LibraryClasses]
+  CompilerIntrinsicsLib
   DebugLib
   BaseLib
   PcdLib


YangGangUEFI:
We have already included `CompilerIntrinsicsLib` in `MdePkg\MdeLibs.dsc.inc`, can you share the steps to reproduce the issue?

https://github.com/tianocore/edk2/blob/29a66468cbabe602c29e770f4382a5979b2d2304/MdePkg/MdeLibs.dsc.inc#L63-L71

makubacki:
I think @jaykrell's initial assessment that `Marker` passed by value is the issue.

https://github.com/tianocore/edk2/blob/309464ec9a8bc5ce5e9c41f4d2300b3e83dd79f8/MdePkg/Include/Base.h#L580-L585

https://github.com/tianocore/edk2/blob/309464ec9a8bc5ce5e9c41f4d2300b3e83dd79f8/MdePkg/Include/Base.h#L592-L606

https://github.com/tianocore/edk2/blob/309464ec9a8bc5ce5e9c41f4d2300b3e83dd79f8/MdePkg/Include/Base.h#L616-L625

https://github.com/tianocore/edk2/blob/309464ec9a8bc5ce5e9c41f4d2300b3e83dd79f8/MdePkg/Include/Base.h#L637-L642

---

Where the [ARM Architecture Procedure Call Standard](https://github.com/ARM-software/abi-aa/releases) defines `va_list` (`__builtin_va_list `) as:

image

So:

```c
UINTN
EFIAPI
UnicodeVSPrint (
  OUT CHAR16        *StartOfBuffer,
  IN  UINTN         BufferSize,
  IN  CONST CHAR16  *FormatString,
  IN  VA_LIST       Marker
  )
{
  ASSERT_UNICODE_BUFFER (StartOfBuffer);
  ASSERT_UNICODE_BUFFER (FormatString);
  return BasePrintLibSPrintMarker ((CHAR8 *)StartOfBuffer, BufferSize >> 1, FORMAT_UNICODE | OUTPUT_UNICODE, (CHAR8 *)FormatString, Marker, NULL);
}
```

Results in a struct copy of `Marker`.

Another option might be taking a pointer:

```c
UINTN
EFIAPI
UnicodeVSPrint (
  OUT CHAR16        *StartOfBuffer,
  IN  UINTN         BufferSize,
  IN  CONST CHAR16  *FormatString,
  IN  VA_LIST       *Marker
  );
```

But that would modify the `PrintLib` interface.

mdkinney:
There are many APIs beside UnicodeVSPrint () that pass VA_LIST or BASE_LIST as a parameter.

Given this issue seems to be specific to AARCH64 and AARCH64 does have a CompilerIntrinsicLib that is required for other reasons, can this be resolved through that dependency?

os-d:
I spoke with @jaykrell offline and have confirmed that the AARCH64 problem he has is that his platform does not include MdeLibs.dsc.inc and so does not include CompilerIntrinsicsLib. Adding CompilerIntrinsicsLib to his DSC resolves the problem for AARCH64. I'll let him comment if he is seeing an issue on X64 as well.

jaykrell:
Right it was ARM64 specific and I've gotten past it, by only putting CompilerIntrinsicsLib in our system for ARM64 (we only build AMD64 and ARM64).
I wonder if we should, move in one direction or the other or another:
 - compiler intrinsics for all architectures?
 - no compiler intrinsics for any architecture?

Should we work to reduce intrinsics use?
In particular, this is an obvious example. We could, wrap va_list in struct, since it might be a struct or pointer or fixed size array, and then anywhere we'd copy it by value, copy with CopyMem or similar (or a typesafe wrapper CopyVaList). Is this a winning, losing, worthwhile battle? I'm not sure, but I think for our platform it is probably an easily won battle.
There's probably just 1-2 va_list copies, and few or no other intrinsics used.

mikebeaton:
> Right it was ARM64 specific and I've gotten past it, by only putting CompilerIntrinsicsLib in our system for ARM64 (we only build AMD64 and ARM64). I wonder if we should, move in one direction or the other or another:
> 
> * compiler intrinsics for all architectures?
> * no compiler intrinsics for any architecture?
> 
> Should we work to reduce intrinsics use? In particular, this is an obvious example. We could, wrap va_list in struct, since it might be a struct or pointer or fixed size array, and then anywhere we'd copy it by value, copy with CopyMem or similar (or a typesafe wrapper CopyVaList). Is this a winning, losing, worthwhile battle? I'm not sure, but I think for our platform it is probably an easily won battle. There's probably just 1-2 va_list copies, and few or no other intrinsics used.

The issue of whether we should be trying to reduce intrinsics use or accept it and standardize it in one place has been discussed before. This is the most recent I am aware of https://github.com/tianocore/edk2/pull/6169#issuecomment-2337788338 .

PR #12271, opened 2026-03-10
MdeModulePkg: Add RealTimeClockLibNull

Flickdm:
# Description

Add a NULL implementation of RealTimeClockLib in MdeModulePkg that returns EFI_UNSUPPORTED for all time operations. This allows packages to satisfy the RealTimeClockLib dependency without pulling in EmbeddedPkg.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Using to test CI against a CryptoPkg that uses RealTimeClockLib instead of gBs->GetTime

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Library/RealTimeClockLibNull/RealTimeClockLibNull.c
+++ b/MdeModulePkg/Library/RealTimeClockLibNull/RealTimeClockLibNull.c
@@ -12,8 +12,7 @@
 
 **/
 
-#include <Uefi/UefiBaseType.h>
-#include <Uefi/UefiSpec.h>
+#include <PiDxe.h>
 #include <Library/RealTimeClockLib.h>
 
 /**
--- a/MdeModulePkg/Library/RealTimeClockLibNull/RealTimeClockLibNull.inf
+++ b/MdeModulePkg/Library/RealTimeClockLibNull/RealTimeClockLibNull.inf
@@ -14,7 +14,7 @@
   INF_VERSION                    = 0x00010005
   BASE_NAME                      = RealTimeClockLibNull
   FILE_GUID                      = BF830151-16A8-4238-BB2D-466591703597
-  MODULE_TYPE                    = BASE
+  MODULE_TYPE                    = DXE_DRIVER
   VERSION_STRING                 = 1.0
   LIBRARY_CLASS                  = RealTimeClockLib
 


MdeModulePkg/Library/RealTimeClockLibNull/RealTimeClockLibNull.c
```diff
@@ -0,0 +1,119 @@
...
+
+  Copyright (c) Microsoft Corporation.
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include <Library/BaseLib.h>
+#include <Library/DebugLib.h>
+#include <Library/RealTimeClockLib.h>
```
mdkinney, line 16: There are no dependencies on BaseLib or DebugLib.  Do need to include <Base.h> for libs of type BASE.

MdeModulePkg/Library/RealTimeClockLibNull/RealTimeClockLibNull.inf
```diff
@@ -0,0 +1,30 @@
...
+
+[Sources]
+  RealTimeClockLibNull.c
+
+[Packages]
+  MdePkg/MdePkg.dec
+  MdeModulePkg/MdeModulePkg.dec
+
+[LibraryClasses]
+  BaseLib
```
mdkinney: No lib dependencies in this implementation

MdeModulePkg/Library/RealTimeClockLibNull/RealTimeClockLibNull.inf
```diff
@@ -0,0 +1,28 @@
...
+#
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+[Defines]
+  INF_VERSION                    = 0x00010005
+  BASE_NAME                      = RealTimeClockLibNull
+  FILE_GUID                      = BF830151-16A8-4238-BB2D-466591703597
+  MODULE_TYPE                    = BASE
```
mdkinney: I see Uefi.h. is required for the UEFI Time structures used by this lib.  I think this library should be type DXE_DRIVER or UEFI_DRIVER?  I see EmbeddedPkg implementations have same issue.

I also see EmbeddedPkg has a similar module call TemplateRealTimeClockLib.  Is this Null library intended to replace the TemplateRealTimeClockLib?

Also, the return status from this lib and TemplateRealTimeClockLib are not the same.  Is that on purpose?

MdeModulePkg/Library/RealTimeClockLibNull/RealTimeClockLibNull.inf
```diff
@@ -0,0 +1,28 @@
...
+  LIBRARY_CLASS                  = RealTimeClockLib
+
+[Sources]
+  RealTimeClockLibNull.c
+
+[Packages]
+  MdePkg/MdePkg.dec
+  MdeModulePkg/MdeModulePkg.dec
+
+[LibraryClasses]
```
makubacki: If no library classes are being linked, you can drop the empty section.


```suggestion

```

MdeModulePkg/Library/RealTimeClockLibNull/RealTimeClockLibNull.inf
Flickdm: So for the sake of transparency:
```
MdePkg/Include/Uefi/UefiSpec.h:825:} EFI_TIME_CAPABILITIES;
MdePkg/Include/Uefi/UefiSpec.h:2105:} EFI_SYSTEM_TABLE;
MdePkg/Include/Uefi/UefiBaseType.h:35:typedef VOID *EFI_HANDLE;
MdePkg/Include/Uefi/UefiBaseType.h:31:typedef RETURN_STATUS EFI_STATUS;
```

I think including these directly allow me to make this base right?

That being said, my challenge with making it a DXE_DRIVER / UEFI_DRIVER is that I need this for CI for building more phase agnostic libraries such as DXE_CORE or even SMM. For example, I would like to be able to replace `gBs->GetTime(..)` in `TimerWrapper.c` with `LibGetTime(..)` that a phase can implement however they need (for Dxe could simply back it with `gBS->GetTime(..))` as it does today.

I also saw that EmbeddedPkg has a similar module. Originally I considered simply moving it out of EmbeddedPkg and fixing it up / removing it (maybe that is the right move and I'm open to it). But in practice I thought there was an intended difference between the "template" and my null implementation that was trying to follow existing examples of null libraries. (The template includes a bunch of unused libraries today - but if you had to go implement the file you may need them)

 The return statuses are intentionally different - as I was trying to make a null implementation and just return EFI_UNSUPPORTED.

What are your thoughts on what I should do with the template?

MdeModulePkg/Library/RealTimeClockLibNull/RealTimeClockLibNull.c
```diff
@@ -0,0 +1,119 @@
...
+  real-time clock hardware can link against this instance to
+  satisfy the library class dependency.
+
+  Copyright (c) Microsoft Corporation.
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include <Uefi/UefiBaseType.h>
```
mdkinney: This is still incorrect.  You have to include the top level env include files from MdePkg/Include first.  Can this lib really be used from PEI in a meaningful way?  Suggest DXE_DRIVER with include of PiDxe.h.  That will get UefiBaseTypes. and UefiSpec.h

MdeModulePkg/Library/RealTimeClockLibNull/RealTimeClockLibNull.inf
mdkinney: No.  This is incorrect style. Must always include the top level env include files from MdePkg/Include as first include. Can this lib really be used from PEI in a meaningful way? Suggest DXE_DRIVER with include of PiDxe.h. That will get UefiBaseTypes. and UefiSpec.h.

You can also declare in the INF which modules types the modules is compatible with in the LIBRARY_CLASS statement.

You would have to discuss this PR and the existing Template with the EmbeddedPkg maintainer to see if your Null one can replace their Template one.

Flickdm: Thanks Mike let me reset - I've updated to `MODULE_TYPE` `DXE_DRIVER` and with `#include <PiDxe.h>` per your suggestion. I think this can work for my CI needs.

Mostly I was using the INF from [VirtualRealTimeClockLib](https://github.com/tianocore/edk2/blob/master/EmbeddedPkg/Library/VirtualRealTimeClockLib/VirtualRealTimeClockLib.inf) as an example. 

Regarding the template, I think they serve different purposes and can coexist. The "TemplateRealTimeClockLib" in EmbeddedPkg is meant as a starting point for platform developers implementing real hardware access (hence the IoLib dependency and EFI_DEVICE_ERROR returns). This null lib is just a stub for satisfying the dependency in CI/test builds where no RTC is needed. 

Unless someone EmbeddedPkg disagrees (@ardbiesheuvel @leiflindholm @AbnerChang)  - I would suggest just leaving the version in EmbeddedPkg as a template and the one I'm trying to bring in for CI.

mdkinney:
@leiflindholm  @ardbiesheuvel The only difference I see between the template in EmbeddedPkg and this one is the return status for a couple APIs being Device Error vs Unsupported.

@Flickdm The Null lib is being added to support CI builds of CryptoPkg which may make sense for CI.  May also make sense if Crypto services are used in PEI phase and gBS->GetTime() is not available. Won't we also need a version of this lib that layers on top of gBS->GetTime() as well for product builds?

Flickdm:
> @leiflindholm @ardbiesheuvel The only difference I see between the template in EmbeddedPkg and this one is the return status for a couple APIs being Device Error vs Unsupported.
> 
> @Flickdm The Null lib is being added to support CI builds of CryptoPkg which may make sense for CI. May also make sense if Crypto services are used in PEI phase and gBS->GetTime() is not available. Won't we also need a version of this lib that layers on top of gBS->GetTime() as well for product builds?

Mike, that is correct if I want to make this change in CryptoPkg (which I'm not there yet - at least in edk2..) I will need to make a layered version that uses gBS->GetTime(). I've been viewing that as a separate PR to be honest.

ardbiesheuvel:
The library class definition was moved into MdeModulePkg on the basis that it is 'architectural', yet the driver that consumes it is still in EmbeddedPkg. That makes no sense to me.

The architectural piece is the driver that produces the associated runtime services (as does `EmbeddedPkg/RealTimeClockRuntimeDxe`) and the library class is just the impedance matcher that ties it to a platform specific implementation.

So what other drivers would need to incorporate this library? Actual RTC hardware may not enjoy being driven by two drivers both consuming the same library. I think it was a mistake to move the library class to MdeModulePkg - it is basically a private library interface that accompanies `EmbeddedPkg/RealTimeClockRuntimeDxe`.

Flickdm:
> The library class definition was moved into MdeModulePkg on the basis that it is 'architectural', yet the driver that consumes it is still in EmbeddedPkg. That makes no sense to me.
> 
> The architectural piece is the driver that produces the associated runtime services (as does `EmbeddedPkg/RealTimeClockRuntimeDxe`) and the library class is just the impedance matcher that ties it to a platform specific implementation.
> 
> So what other drivers would need to incorporate this library? Actual RTC hardware may not enjoy being driven by two drivers both consuming the same library. I think it was a mistake to move the library class to MdeModulePkg - it is basically a private library interface that accompanies `EmbeddedPkg/RealTimeClockRuntimeDxe`.

So just to set the context. I mostly didn't want to dive into a lot of details yet because I'm still working on getting this to a state where I would be happy bringing this to edk2. 

I have been working on a phase agnostic crypto driver that can contain all of the crypto a platform needs. Today it works for DXE / MM / SMM / SupvMM on Intel platforms with some additional support needed for AARCH64 to support Secure World. The goal here is that all the crypto code that a platform can need is encapsulated within a single binary that can have better "drop in" / backwards compatible support and SBOM. 

From the EDK2/UEFI perspective this is the view that the firmware would see:

```mermaid
---
config:
  layout: elk
---
classDiagram

    namespace Protocol {
        class OneCrypto:::protocol {
            <<Protocol Interface>>
            UINT16 Major;
            UINT16 Minor;
            // v1.0 ------------------------------------------
            ONE_CRYPTO_PROVIDED_FUNC ONE_CRYPTO_PROVIDED_FUNC;
            // ...
        }
    }

    namespace Library {
        class BaseCryptLib:::library {
        }

        class TlsLib:::library {

        }

        class HashApiLib:::library {

        }
    }

    namespace LibraryInstance {
        class BaseCryptOnOneCrypto:::library_instance {
            <<Library Interface>>
            EFI_STATUS CryptoFunc(..)
        }
    }

    namespace Driver {
        class OneCryptoBinSupvMm:::driver {
            <<Driver>>
            Phase Agnostic Crypto Driver
            Shared across Phases

            Crypto Entry(..) // phase agnostic code
            SupvMm Entry(..) // stub entry
        }

        class OneCryptoLoaderSupvMm:::driver {
            <<Driver>>
        }

        class OneCryptoLoaderDxe:::driver {
            <<Driver>>
        }
    }

    namespace GenericModules {
        class GenericDriver:::module {
            <<Driver>>
            Any driver that needs crypto
        }

        class GenericLibrary:::module {
            <<Library Interface>>
            Any library that needs crypto
        }

    }

    OneCryptoLoaderSupvMm --> OneCryptoBinSupvMm : Loads Phase Agnostic Code
    OneCryptoLoaderDxe --> OneCryptoBinSupvMm : Loads Phase Agnostic Code

    OneCryptoBinSupvMm ..> OneCryptoLoaderSupvMm : Provides Crypto
    OneCryptoBinSupvMm ..> OneCryptoLoaderDxe : Provides Crypto

    OneCryptoLoaderSupvMm ..|> OneCrypto : Publishes
    OneCryptoLoaderDxe ..|> OneCrypto : Publishes

    BaseCryptOnOneCrypto ..|> OneCrypto : Locates

    BaseCryptOnOneCrypto <.. BaseCryptLib : Backed By
    BaseCryptOnOneCrypto <.. TlsLib : Backed By


    GenericLibrary ..> BaseCryptLib : Uses
    GenericDriver ..> BaseCryptLib : Uses
    GenericLibrary ..> TlsLib : Uses
    GenericDriver ..> HashApiLib : Uses

    classDef driver fill:#187286,stroke:#333,stroke-width:2px
    classDef protocol fill:#d63230,stroke:#333,stroke-width:2px
    classDef library fill:#1c77c3,stroke:#333,stroke-width:2px
    classDef module fill:#ff8c00,stroke:#333,stroke-width:2px

```

Internally this looks something like this:

```mermaid
---
config:
  layout: elk
---
classDiagram
    direction TB

    namespace StaticCryptoLibraries {
        class BaseCryptLib {
            <<Library Interface>>
            +Pkcs7Verify()
            +RsaPkcs1Verify()
            +Sha256HashAll()
            +X509GetSubjectName()
        }
        class TlsLib {
            <<Library Interface>>
            +TlsInitialize()
            +TlsSetVersion()
            +TlsDoHandshake()
        }
        class OpensslLibFull {
            <<Library Instance>>
            OpensslPkg/Library/OpensslLib
            Full OpenSSL crypto + TLS
        }
        class IntrinsicLib {
            <<Library Instance>>
            CryptoPkg/Library/IntrinsicLib
            memcpy, memmove, memset
        }
    }

    namespace OnOneCryptoShims {
        class DebugLibOnOneCrypto {
            <<Shim>>
            implements DebugLib
            DebugPrint → OneCryptoDebugPrint
        }
        class MemoryAllocationLibOnOneCrypto {
            <<Shim>>
            implements MemoryAllocationLib
            AllocatePool → OneCryptoAllocatePool
        }
        class RngLibOnOneCrypto {
            <<Shim>>
            implements RngLib
            GetRandomNumber64 → OneCryptoGetRandom
        }
        class TimerLibOnOneCrypto {
            <<Shim>>
            implements TimerLib
            MicroSecondDelay → OneCryptoDelay
        }
        class RealTimeClockLibOnOneCrypto {
            <<Shim>>
            implements RealTimeClockLib
            GetTime → OneCryptoGetTime
        }
    }

    namespace DependencyInjection {
        class OneCryptoCrtLib {
            <<Library Instance>>
            ONE_CRYPTO_DEPENDENCIES* mDeps
            +OneCryptoCrtSetup(deps)
            +OneCryptoAllocatePool()
            +OneCryptoFreePool()
            +OneCryptoGetTime()
            +OneCryptoGetRandomNumber64()
            +OneCryptoDebugPrint()
            +OneCryptoMicroSecondDelay()
        }
    }

    namespace OneCryptoBinary {
        class OneCryptoBinStandaloneMm {
            <<Driver>>
            Statically linked binary
            MM_STANDALONE module
            Publishes gOneCryptoPrivateProtocolGuid
        }
        class OneCryptoBinSupvMm {
            <<Driver>>
            Statically linked binary
            Uses OpensslLibFullAccel
        }
        class OneCryptoBinDxe {
            <<Driver>>
            Statically linked binary
            DXE_DRIVER module
        }
    }

    namespace Loader {
        class OneCryptoLoaderDxe {
            <<Driver>>
            Loads OneCryptoBin PE/COFF
            Populates ONE_CRYPTO_DEPENDENCIES
            Installs gOneCryptoProtocolGuid
        }
    }

    %% Static linking into OneCryptoBin
    OneCryptoBinStandaloneMm --> BaseCryptLib : statically links
    OneCryptoBinStandaloneMm --> TlsLib : statically links
    OneCryptoBinStandaloneMm --> IntrinsicLib : statically links
    OneCryptoBinDxe --> BaseCryptLib : statically links
    OneCryptoBinDxe --> TlsLib : statically links
    OneCryptoBinDxe --> IntrinsicLib : statically links
    OneCryptoBinSupvMm --> BaseCryptLib : statically links
    OneCryptoBinSupvMm --> TlsLib : statically links
    OneCryptoBinSupvMm --> IntrinsicLib : statically links

    %% BaseCryptLib depends on OpensslLib
    BaseCryptLib --> OpensslLibFull : calls OpenSSL APIs
    TlsLib --> OpensslLibFull : calls SSL/TLS APIs

    %% OpensslLib platform dependencies satisfied by shims
    OpensslLibFull ..> DebugLibOnOneCrypto : DebugLib
    OpensslLibFull ..> MemoryAllocationLibOnOneCrypto : MemoryAllocationLib
    OpensslLibFull ..> RngLibOnOneCrypto : RngLib
    BaseCryptLib ..> TimerLibOnOneCrypto : TimerLib
    BaseCryptLib ..> RealTimeClockLibOnOneCrypto : RealTimeClockLib
    BaseCryptLib ..> MemoryAllocationLibOnOneCrypto : MemoryAllocationLib
    BaseCryptLib ..> DebugLibOnOneCrypto : DebugLib

    %% All shims delegate to OneCryptoCrtLib
    DebugLibOnOneCrypto --> OneCryptoCrtLib : delegates
    MemoryAllocationLibOnOneCrypto --> OneCryptoCrtLib : delegates
    RngLibOnOneCrypto --> OneCryptoCrtLib : delegates
    TimerLibOnOneCrypto --> OneCryptoCrtLib : delegates
    RealTimeClockLibOnOneCrypto --> OneCryptoCrtLib : delegates

    %% Loader populates dependencies
    OneCryptoLoaderDxe ..> OneCryptoCrtLib : calls OneCryptoCrtSetup
    OneCryptoLoaderDxe ..> OneCryptoBinDxe : loads + dispatches
```


The goal here is that the OneCryptoBinSupv is really some small shim code that takes OpensslLib as it is today and links against it satisfying its needs. The challenge is that I need to make OpensslLib / BaseCryptLib "phase agnostic". For most things this isn't too difficult as I can use Edk2's built in libraries and back them by my own implementation that shims out the call.

However there are some calls like the following:

In Openssl 3.5.1 `gBS->GetTime(..)` was added in `TimerWapper.c`. Now I'm not certain exactly what this is being used for as I don't think we have need for time in the UEFI environment but to reduce the risk I deviced to just back it by shimming out GetTime. Fortunately the MdeModulePkg already had a library class that I could use that met the calling convention for GetTime(..)

I.E:

```c
EFI_STATUS
EFIAPI
LibGetTime (
  OUT EFI_TIME               *Time,
  OUT EFI_TIME_CAPABILITIES  *Capabilities
  )
```

Which I could back with:

```c

EFI_STATUS
EFIAPI
LibGetTime (
  OUT EFI_TIME               *Time,
  OUT EFI_TIME_CAPABILITIES  *Capabilities OPTIONAL
  )
{
  if (Time == NULL) {
    return EFI_INVALID_PARAMETER;
  }

  return OneCryptoGetTime (Time, Capabilities);
}
```
And the rest of the functions simply returning `EFI_UNSUPPORTED` 

In the case of my DXE driver it shims gBS->GetTime(..) and in the case of my SMM drivers returns 0. since it was already not supported in that phase. (That's a debate for another thread)

Now @ardbiesheuvel, I agree with you the rest of this file is a little strange as it implies I am interacting with hardware whereas in my case I'm not - I'm simply using it to shim `GetTime` using a file already in MdeModulePkg because I need the abstraction. 

Now I'm not opposed to adding this function definition to another file and updating my library implementation but I was attempting to use a library already in the upstream and what I thought was intentionally placed here.

@mdkinney sorry for diverging this conversation - this was meant to be a small change that allows me to "prepare" openssllib for my phase agnostic implementation and ensure CI still works - that I was hoping would be trivial and have a larger ecosystem conversation when it makes sense to really dive deeper into this

ardbiesheuvel:
> In Openssl 3.5.1 `gBS->GetTime(..)` was added in `TimerWapper.c`. Now I'm not certain exactly what this is being used for as I don't think we have need for time in the UEFI environment but to reduce the risk I deviced to just back it by shimming out GetTime. Fortunately the MdeModulePkg already had a library class that I could use that met the calling convention for GetTime(..)
> 

First of all, GetTime() is a runtime service not a boot-time service, although perhaps that makes little difference in the context of this discussion.

If the code you are refactoring/abstracting has a dependency on the GetTime() runtime service, you will need to provide a shim for that, but I find the use of the current date/time in our crypto code rather fishy to begin with, given that it is unreliable at early boot.

Using RealTimeClockLib because it has a handy LibGetTime() interface is really not an acceptable approach here.

Flickdm:
> > In Openssl 3.5.1 `gBS->GetTime(..)` was added in `TimerWapper.c`. Now I'm not certain exactly what this is being used for as I don't think we have need for time in the UEFI environment but to reduce the risk I deviced to just back it by shimming out GetTime. Fortunately the MdeModulePkg already had a library class that I could use that met the calling convention for GetTime(..)
> 
> First of all, GetTime() is a runtime service not a boot-time service, although perhaps that makes little difference in the context of this discussion.
> 
> If the code you are refactoring/abstracting has a dependency on the GetTime() runtime service, you will need to provide a shim for that, but I find the use of the current date/time in our crypto code rather fishy to begin with, given that it is unreliable at early boot.
> 
> Using RealTimeClockLib because it has a handy LibGetTime() interface is really not an acceptable approach here.

Sorry @ardbiesheuvel you are correct and I misspoke. 

I agree with you that the use case of `GetTime(..)` in Crypto code is rather fishy. In my initial exploration the only two places I could find it being used are in [QUIC](https://github.com/openssl/openssl/blob/refs/heads/master/ssl/quic/quic_port.c) and the UnitTests for QUIC. I was simply trying to use `GetTime(..)` abstraction to reduce the risk of regression given the amount of things I'm changing / refactoring. 

However rather than create a new interface that I'm not convinced I need in light of this abstraction being removed - maybe the correct path is to remove the `GetTime` dependency. Given we do not have a dependency on QUIC today.

PR #12272, opened 2026-03-10
BaseTools: Fix MSVC warning 4319

kuqin12:
# Description

This change fixes the warning 4319 that was recently fixed from MSVC.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested locally using Edk2ToolsBuild.py for IA32, X64 and AArch64 targets and fixed the build break.

## Integration Instructions

N/A


kuqin12:
> I am not sure if these need to be in separate commits?
>
I agree this might be overkilling. Will squash.

kuqin12:
> I agree this might be overkilling. Will squash.

This is updated.

makubacki:
nit: You could optionally provide a bit more detail in the commit message. For example, the `(UINT32)` cast narrows the `sizeof()` to `UINT32` which is consistent with the other types in the expression being assigned to a `UINT32` variable. The `(UINTN)` cast widens the `UINT32` value so the mask is applied properly.

kuqin12:
> nit: You could optionally provide a bit more detail in the commit message. For example, the `(UINT32)` cast narrows the `sizeof()` to `UINT32` which is consistent with the other types in the expression being assigned to a `UINT32` variable. The `(UINTN)` cast widens the `UINT32` value so the mask is applied properly.

Thanks, this is updated.

kuqin12:
@lgao4 @makubacki would you mind adding a push tag to this change when you have time? Thanks in advance!

kuqin12:
@mergifyio queue

PR #12282, opened 2026-03-11
MdeModulePkg/TerminalDxe: Fix handling of shift state in notifications

### LINKED ISSUE #12281: [Bug]: TerminalDxe ignores shift state when calling notification functions

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

MdeModulePkg

_No response_

### Current Behavior

The implementation of SimpleTextInputEx protocol in TerminalDxe allows registering a notification for a key with nonzero shift state or toggle state, e.g. Ctrl+N or Shift+F3.

However, the shift state and toggle state are ignored when TerminalDxe decides which notifications to trigger for a key event (see `IsKeyRegistered ()` in MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c).

If a user presses 'n', say, it UEFI shell after that, the notification function for Ctrl+N will be triggered, which is wrong. The same way, if they press F3, the notification for Shift+F3 will be triggered.

### Expected Behavior

Most kinds of the shift states are not passed via the terminal lines, and TerminalDxe does not support that.

So, one could expect that the notifications for Ctrl+N and Shift+F3 and other key events with non-trivial shift states or toggle states would never be triggered by the implementation of SimpleTextInputEx in TerminalDxe.

Whether to allow registering such notifications here or not, it could be tricky. TerminalDxe could reject such notifications but I don't know if it could break any existing components. I'd rather allow registering them but would add a debug message to `TerminalConInRegisterKeyNotify()` to make it easier to see what is going on.

### Steps To Reproduce

1. The attached DXE driver can be used to demonstrate the issue:
[KeyDebugDxe.c](https://github.com/user-attachments/files/25919473/KeyDebugDxe.c)
The driver tries to register key notifications for RCtrl+N and LShift+F3 in all instances of SimpleTextInputEx protocol.

2. Add this driver to ArmVirtPkg/ArmVirtQemu platform and build the OVMF firmware for Aarch64 as follows:
```
(CT):/work/edk2$ . ./edksetup.sh
(CT):/work/edk2$ make -C BaseTools
(CT):/work/edk2$ export GCC_AARCH64_PREFIX="aarch64-linux-gnu-"
(CT):/work/edk2$ build -DTTY_TERMINAL \
  --platform=ArmVirtPkg/ArmVirtQemu.dsc \
  --arch=AARCH64 \
  --buildtarget=DEBUG \
  --tagname=GCC
```

3. Run a QEMU VM with this UEFI FW.
4. In the serial console, press ESC when the system boots to "Bds wait" stage to enter the EFI boot menu. Select "EFI internal shell", press ESC there.
5. In the Shell, press 'n'. A debug message will confirm that the notification for RCtrl+N has been (erroneously) triggered each time 'n' has been pressed:
```
DbgKeyHanderN: key 110, shift state 0x00000000.
nDbgKeyHanderN: key 110, shift state 0x00000000.
n
```
Same for F3.

### Build Environment

```markdown
- OS(s):Ubuntu 24.04
- Tool Chain(s): GCC
```

### Version Information

```text
Commit: 309464ec9a8bc5ce5e9c41f4d2300b3e83dd79f8
```

### Urgency

Medium

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

euspectre:
# Description

This patch fixes https://github.com/tianocore/edk2/issues/12281.

Currently, `TerminalConInRegisterKeyNotify()` allows registering a notification with nonzero KeyShiftState or KeyToggleState. However, `IsKeyRegistered()` ignores these when checking if the keys match.

As a result, some firmware component may successfully register a notification for, say, RCtrl+n, but the notification function will be called each time the user presses 'n', which is wrong.

Most kinds of shift states and toggle states are not transferred via a terminal line, so the notification functions for the keys with nonzero KeyShiftState or KeyToggleState should not trigger with TerminalDxe.

`TerminalConInRegisterKeyNotify()` could reject such notifications but it is unclear if it could break the existing UEFI components.

Instead, this patch adds a debug message when someone tries to register a notification with nonzero states, it also updates `IsKeyRegistered()` to take KeyShiftState and KeyToggleState into account.

This way, `IsKeyRegistered()` will treat the notifications for 'n' and 'RCtrl+n' as different ones. So, the callback function for 'RCtrl+n' will not be called when the user presses 'n'.

As it is not prohibited to set only EFI_SHIFT_STATE_VALID flag in the shift state (leaving the remaining bits zeroed), IsKeyRegistered() ignores EFI_SHIFT_STATE_VALID.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

I tested the fix using the steps to reproduce the issue described in https://github.com/tianocore/edk2/issues/12281.

Without the fix, pressing 'n' in the EFI internal shell triggered the notification for 'RCtrl+n', pressing F3 triggered the notification for 'LShift+F3'.

With the fix, neither of the notifications were triggered.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
+++ b/MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
@@ -168,10 +168,23 @@ IsKeyRegistered (
   IN EFI_KEY_DATA  *InputData
   )
 {
+  UINT32  ShiftState;
+  UINT32  InputShiftState;
+
   ASSERT (RegsiteredData != NULL && InputData != NULL);
 
   if ((RegsiteredData->Key.ScanCode    != InputData->Key.ScanCode) ||
       (RegsiteredData->Key.UnicodeChar != InputData->Key.UnicodeChar))
+  {
+    return FALSE;
+  }
+
+  /* Do not take EFI_SHIFT_STATE_VALID flag into account when comparing shift states */
+  ShiftState      = RegsiteredData->KeyState.KeyShiftState | EFI_SHIFT_STATE_VALID;
+  InputShiftState = InputData->KeyState.KeyShiftState | EFI_SHIFT_STATE_VALID;
+
+  if ((ShiftState != InputShiftState) ||
+      (RegsiteredData->KeyState.KeyToggleState != InputData->KeyState.KeyToggleState))
   {
     return FALSE;
   }
@@ -337,6 +350,7 @@ TerminalConInRegisterKeyNotify (
   LIST_ENTRY                     *Link;
   LIST_ENTRY                     *NotifyList;
   TERMINAL_CONSOLE_IN_EX_NOTIFY  *CurrentNotify;
+  UINT32                         ShiftState;
 
   if ((KeyData == NULL) || (NotifyHandle == NULL) || (KeyNotificationFunction == NULL)) {
     return EFI_INVALID_PARAMETER;
@@ -361,6 +375,29 @@ TerminalConInRegisterKeyNotify (
         return EFI_SUCCESS;
       }
     }
+  }
+
+  //
+  // Some components might set only EFI_SHIFT_STATE_VALID flag in KeyShiftState.
+  // Treat such shift state as zero.
+  //
+  ShiftState = KeyData->KeyState.KeyShiftState & ~EFI_SHIFT_STATE_VALID;
+  if ((ShiftState != 0) || (KeyData->KeyState.KeyToggleState != 0)) {
+    //
+    // Shift state and toggle state are not transferred via serial lines, so
+    // the notification functions for the keys with nonzero KeyShiftState or
+    // KeyToggleState will never trigger with this implementation of SimpleTextInEx
+    // protocol.
+    //
+    DEBUG ((
+      DEBUG_INFO,
+      "%a: Attempt to register notifier for the key 0x%04x (scan code 0x%04x) with shift state 0x%08x and toggle state 0x%02x.\n",
+      __func__,
+      KeyData->Key.UnicodeChar,
+      KeyData->Key.ScanCode,
+      KeyData->KeyState.KeyShiftState,
+      KeyData->KeyState.KeyToggleState
+      ));
   }
 
   //
--- a/MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
+++ b/MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
@@ -149,9 +149,9 @@ TerminalConInReadKeyStroke (
 /**
   Check if the key already has been registered.
 
-  If both RegsiteredData and InputData is NULL, then ASSERT().
+  If both RegisteredData and InputData is NULL, then ASSERT().
 
-  @param  RegsiteredData           A pointer to a buffer that is filled in with the
+  @param  RegisteredData           A pointer to a buffer that is filled in with the
                                    keystroke state data for the key that was
                                    registered.
   @param  InputData                A pointer to a buffer that is filled in with the
@@ -164,27 +164,27 @@ TerminalConInReadKeyStroke (
 **/
 BOOLEAN
 IsKeyRegistered (
-  IN EFI_KEY_DATA  *RegsiteredData,
+  IN EFI_KEY_DATA  *RegisteredData,
   IN EFI_KEY_DATA  *InputData
   )
 {
   UINT32  ShiftState;
   UINT32  InputShiftState;
 
-  ASSERT (RegsiteredData != NULL && InputData != NULL);
+  ASSERT (RegisteredData != NULL && InputData != NULL);
 
-  if ((RegsiteredData->Key.ScanCode    != InputData->Key.ScanCode) ||
-      (RegsiteredData->Key.UnicodeChar != InputData->Key.UnicodeChar))
+  if ((RegisteredData->Key.ScanCode    != InputData->Key.ScanCode) ||
+      (RegisteredData->Key.UnicodeChar != InputData->Key.UnicodeChar))
   {
     return FALSE;
   }
 
   /* Do not take EFI_SHIFT_STATE_VALID flag into account when comparing shift states */
-  ShiftState      = RegsiteredData->KeyState.KeyShiftState | EFI_SHIFT_STATE_VALID;
+  ShiftState      = RegisteredData->KeyState.KeyShiftState | EFI_SHIFT_STATE_VALID;
   InputShiftState = InputData->KeyState.KeyShiftState | EFI_SHIFT_STATE_VALID;
 
   if ((ShiftState != InputShiftState) ||
-      (RegsiteredData->KeyState.KeyToggleState != InputData->KeyState.KeyToggleState))
+      (RegisteredData->KeyState.KeyToggleState != InputData->KeyState.KeyToggleState))
   {
     return FALSE;
   }
--- a/MdeModulePkg/Universal/Console/TerminalDxe/Terminal.h
+++ b/MdeModulePkg/Universal/Console/TerminalDxe/Terminal.h
@@ -241,7 +241,7 @@ TerminalConInReadKeyStroke (
 /**
   Check if the key already has been registered.
 
-  @param  RegsiteredData           A pointer to a buffer that is filled in with the
+  @param  RegisteredData           A pointer to a buffer that is filled in with the
                                    keystroke state data for the key that was
                                    registered.
   @param  InputData                A pointer to a buffer that is filled in with the
@@ -254,7 +254,7 @@ TerminalConInReadKeyStroke (
 **/
 BOOLEAN
 IsKeyRegistered (
-  IN EFI_KEY_DATA  *RegsiteredData,
+  IN EFI_KEY_DATA  *RegisteredData,
   IN EFI_KEY_DATA  *InputData
   );
 


lgao4:
@lgao4 , please review it.

mikebeaton:
Slightly confusing, you're saying that "the notification functions for the keys with nonzero KeyShiftState or KeyToggleState should not trigger", and then adding code to make sure that they trigger properly.

euspectre:
> Slightly confusing, you're saying that "the notification functions for the keys with nonzero KeyShiftState or KeyToggleState should not trigger", and then adding code to make sure that they trigger properly.

No, they will not trigger with serial.

After this patch, UEFI will still allow registering such functions (but will complain about that in debug mode). However, it will now correctly compare the key event that actually happened and the ones the callback functions have been registered for. It will see the difference in the shift state (0 for the actual event and non-zero for a registered callback) and will not call such notifier functions.

This check added to IsKeyRegistered makes sure the callback for Ctrl+F3 will not trigger when F3 is pressed:
https://github.com/tianocore/edk2/pull/12282/changes/da963bdedd04d8446e5951e0d655d2c2add56df7#diff-6bdbf864505f8b86109e1ed5546f42ee8617ee62189fc7785045d072e36d5132R186

mikebeaton:
> > Slightly confusing, you're saying that "the notification functions for the keys with nonzero KeyShiftState or KeyToggleState should not trigger", and then adding code to make sure that they trigger properly.
> 
> No, they will not trigger with serial.
> 
> After this patch, UEFI will still allow registering such functions (but will complain about that in debug mode). However, it will now correctly compare the key event that actually happened and the ones the callback functions have been registered for. It will see the difference in the shift state (0 for the actual event and non-zero for a registered callback) and will not call such notifier functions.
> 
> This check added to IsKeyRegistered makes sure the callback for Ctrl+F3 will not trigger when F3 is pressed: [da963bd#diff-6bdbf864505f8b86109e1ed5546f42ee8617ee62189fc7785045d072e36d5132R186](https://github.com/tianocore/edk2/commit/da963bdedd04d8446e5951e0d655d2c2add56df7#diff-6bdbf864505f8b86109e1ed5546f42ee8617ee62189fc7785045d072e36d5132R186)

No, I meant that your wording is confusing. It's a direct quote from your commit message that "the notification functions for the keys with nonzero KeyShiftState or KeyToggleState should not trigger". At least one reading of "should not trigger" is "ought not to trigger". Which presumably you don't mean, since you're fixing to make sure they trigger properly. I am only saying that it would help to change the wording to explain what happens and what should happen, on serial and not on serial, more clearly.

Apart from the (to me, currently) confusing way it is explained, that actual changes look correct to me.

euspectre:
> No, I meant that your wording is confusing. It's a direct quote from your commit message that "the notification functions for the keys with nonzero KeyShiftState or KeyToggleState should not trigger". At least one reading of "should not trigger" is "ought not to trigger". Which presumably you don't mean, since you're fixing to make sure they trigger properly. 

Ah, I see. You are probably right.

Any ideas how to reword that to avoid confusion?

mikebeaton:
> > No, I meant that your wording is confusing. It's a direct quote from your commit message that "the notification functions for the keys with nonzero KeyShiftState or KeyToggleState should not trigger". At least one reading of "should not trigger" is "ought not to trigger". Which presumably you don't mean, since you're fixing to make sure they trigger properly.
> 
> Ah, I see. You are probably right.
> 
> Any ideas how to reword that to avoid confusion?

How about just update this para of the commit message:

```
Shift state and toggle state are not transferred via a terminal line, so
the notification functions for the keys with nonzero KeyShiftState or
KeyToggleState should not trigger with TerminalDxe.
```

to

```
Shift state and toggle state are not transferred via a serial line, so
the notification functions for the keys with nonzero KeyShiftState or
KeyToggleState will never trigger in this case.
```

euspectre:
How about something like this:

"Shift state and toggle state are not transferred via a terminal line, so it is expected that TerminalDxe never triggers the notification functions for the keys with nonzero KeyShiftState or KeyToggleState."

?

euspectre:
> ```
> Shift state and toggle state are not transferred via a serial line, so
> the notification functions for the keys with nonzero KeyShiftState or
> KeyToggleState will never trigger in this case.
> ```

Yes, I'll use your variant.

Thanks!

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
mikebeaton: `Regsitered` is a typo in this local variable name. Could you possibly add a separate commit to the PR which fixes this?

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
```diff
@@ -363,6 +377,26 @@ TerminalConInRegisterKeyNotify (
     }
   }
 
+  //
+  // The notification might not be triggered by this implementation of SimpleTextInEx
+  // protocol if shift state or toggle state are set.
+  //
```
mikebeaton: This comment should go below, above the added DEBUG line - and should probably mention the fact that shift and toggle state will never be sent over serial lines.

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
euspectre: Yes, I noticed ;-)  IIRC, this typo was copied & pasted a few more times across EDK2. I'll try to correct that, OK.

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
```diff
@@ -363,6 +377,26 @@ TerminalConInRegisterKeyNotify (
...
   }
 
+  //
+  // The notification might not be triggered by this implementation of SimpleTextInEx
+  // protocol if shift state or toggle state are set.
+  //
+  // Some components might set only EFI_SHIFT_STATE_VALID flag in KeyShiftState.
+  // Treat such shift state as zero.
+  //
+  ShiftState = KeyData->KeyState.KeyShiftState & ~EFI_SHIFT_STATE_VALID;
```
mikebeaton, line 384: I would actually suggest adding back in EFI_SHIFT_STATE_VALID if the result is non-zero after removing it. That way you should not even need the lines mentioned in my next comment, in the code above.

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
```diff
@@ -176,6 +179,16 @@ IsKeyRegistered (
     return FALSE;
   }
 
+  /* Do not take EFI_SHIFT_STATE_VALID flag into account when comparing shift states */
+  ShiftState      = RegsiteredData->KeyState.KeyShiftState | EFI_SHIFT_STATE_VALID;
+  InputShiftState = InputData->KeyState.KeyShiftState | EFI_SHIFT_STATE_VALID;
```
mikebeaton, line 184: I believe you would not need these lines, OR-ing in EFI_SHIFT_STATE_VALID, if you OR it in automatically in the way I suggest in the register function.

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
mikebeaton: If there are several, could be a separate PR.

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
euspectre: I think the line `InputShiftState = InputData->KeyState.KeyShiftState | EFI_SHIFT_STATE_VALID;` is still required here. We might get InputData with or without EFI_SHIFT_STATE_VALID bit here.

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
mikebeaton: Actually, no scratch this.

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
mikebeaton: Actually, no scratch this.

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
mikebeaton: Resolved. I believe I cannot resolve it.

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
mikebeaton: Resolved. I believe I cannot resolve it.

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
euspectre: EFI_SHIFT_STATE_VALID bit is cleared here for comparison only. I'd store `KeyData->KeyState.KeyShiftState` as it is, as it was passed here, without changes.

mikebeaton: I was thinking that EFI_SHIFT_STATE_VALID would only be set when other accompanying shift bits are set, but looking at the code which sets it I was wrong. So scratch the comments about this. Apologies.

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
euspectre: I have just checked - it seems, at least 7 implementations of `IsKeyRegistered ()` have this issue.
I'd rather not do that in this PR.

Or, perhaps, - correct that in TerminalConIn* here.

mikebeaton: No agreed, it's better to do it at once, in a different PR, rather than piecemeal. Are you interested in doing that?

mikebeaton:
@euspectre - Sorry, I gave a bit of a hail of comments, but if you get a moment to move the comment I suggested, I'm happy to recommend approve.

mikebeaton:
Please could you add `Fixes: #12281` in exactly that format as an additional line in the commit message. It will correctly link the issue and the PR.

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
euspectre: Well, I have added the fix for the typo in TerminalDxe for now. It won't hurt, I guess.

As for other UEFI components, I'd rather touch as little as possible.

MdeModulePkg/Universal/Console/TerminalDxe/TerminalConIn.c
euspectre: Done.

euspectre:
> Fixes: #12281

Sure. Done.

mikebeaton (approved):

euspectre:
Gentle ping.
Comments? Suggestions?

mikebeaton:
@lgao4 - Hi. This PR looked good to me.

lgao4 (approved):

euspectre:
@mikebeaton, @lgao4 : 
Thank you!

Could you please merge this PR then?
I do not have appropriate rights to do so.

lgao4:
> @mikebeaton, @lgao4 : Thank you!
> 
> Could you please merge this PR then? I do not have appropriate rights to do so.

Now , we are in hard feature freeze phase. Can this one be merged after the stable tag is created on May 22th?

euspectre:
> > @mikebeaton, @lgao4 : Thank you!
> > Could you please merge this PR then? I do not have appropriate rights to do so.
> 
> Now , we are in hard feature freeze phase. Can this one be merged after the stable tag is created on May 22th?

Ah, I see.

Yes, this one can be merged when the next merge window opens, it is not that urgent.

lgao4 (approved):

EricGao2015:
@euspectre Hi Evgenii, did you test this patch on AARCH64 QEMU platform? It seems the layout is disordered when entering UEFI SHELL.
image

Build command: `build -b DEBUG -t CLANGDWARF -a AARCH64 -p ArmVirtPkg/ArmVirtQemu.dsc`
Run firmware with QEMU: `qemu-system-aarch64 -m 4G -M virt -smp 2 -cpu cortex-a72 -bios Build/ArmVirtQemu-AArch64/DEBUG_CLANGDWARF/FV/QEMU_EFI.fd -nographic`

Maybe the print level of such messages should be degraded.

euspectre:
> Hi Evgenii, did you test this patch on AARCH64 QEMU platform? It seems the layout is disordered when entering UEFI SHELL. 

Yes, as I described in https://github.com/tianocore/edk2/issues/12281, the key notifiers were tested in UEFI shell in QEMU with OVMF firmware for Aarch64.

However, in my tests, the messages about registered notifications appeared earlier and did not clobber the layout.

Might be something with terminal settings or QEMU options.

> Maybe the print level of such messages should be degraded.

Yes, if the messages affect the layout as you said, it is reasonable to change their print level to DEBUG_VERBOSE or similar.

EricGao2015:
> > Hi Evgenii, did you test this patch on AARCH64 QEMU platform? It seems the layout is disordered when entering UEFI SHELL.
> 
> Yes, as I described in #12281, the key notifiers were tested in UEFI shell in QEMU with OVMF firmware for Aarch64.
> 
> However, in my tests, the messages about registered notifications appeared earlier and did not clobber the layout.
> 
> Might be something with terminal settings or QEMU options.
> 
> > Maybe the print level of such messages should be degraded.
> 
> Yes, if the messages affect the layout as you said, it is reasonable to change their print level to DEBUG_VERBOSE or similar.

Sorry for reply so late. I found that the layout issue only happens when the mapping is not found. I'll fix this issue later. Even though the layout is normal, the debug message is still causing interference to Shell page. So I agree change their print level to DEBUG_VERBOSE.
image

PR #12283, opened 2026-03-11
OvmfPkg/PlatformPei: Skip pvalidate for IGVM pre-validated pages

AmeanAsad:
When booting an SEV-SNP guest via IGVM, pages loaded through `SNP_LAUNCH_UPDATE` are already validated by the VMM. Re-validating these pages causes a #VC exception. Collect IGVM data ranges from GUIDed HOBs and skip them during system RAM validation.

## Description

When a kernel or UKI is embedded inside an IGVM file, its pages are validated by the VMM during `SNP_LAUNCH_UPDATE`, the PSP marks them as private in the RMP. During PEI, `AmdSevSnpInitialize()` iterates all system memory HOBs and calls `MemEncryptSevSnpPreValidateSystemRam()` for each range below 4 GB. This includes address ranges where IGVM data was already placed and validated.

The re-validation triggers a Page State Change (PSC) VMGEXIT for pages that are already private. The PSC fails and OVMF calls `SnpPageStateFailureTerminate()` and the guest silently terminates with no serial output.

This happens because QEMU's `qigvm_x86_get_mem_map_entry()` in `target/i386/igvm.c` converts E820 entries to the IGVM memory map 1:1 without excluding `PageData` GPA ranges. The guest sees the full RAM range as needing acceptance, including regions already validated by IGVM.

This patch modifies `AmdSevSnpInitialize()` to:

1. Scan the HOB list for `gEfiIgvmDataHobGuid` entries (created by `PlatformIgvmDataHobs()` when IGVM data is loaded)
2. Collect the address ranges of all IGVM data blocks
3. When validating `EFI_RESOURCE_SYSTEM_MEMORY` ranges, use a new `ValidateRamSkipIgvm()` helper that walks the range and skips sub-ranges overlapping with IGVM data

When no IGVM data HOBs are present (non-IGVM boot or firmware-only IGVM), the original code path is used unchanged.

Files changed:

- `OvmfPkg/PlatformPei/AmdSev.c` — new `ValidateRamSkipIgvm()` helper, modified `AmdSevSnpInitialize()` to collect IGVM ranges and route through the helper
- `OvmfPkg/PlatformPei/PlatformPei.inf` — added `gEfiIgvmDataHobGuid` to `[Guids]`

- [ ] Breaking change?
- [x] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on AMD EPYC 8224P (Siena, Zen 4c) running Linux 6.17.0-14-generic (Ubuntu 25.10) with SEV firmware API 1.55 build 49.

### QEMU

IGVM support requires building QEMU from source. Built from upstream master (commit `ece408818d`, Feb 2026) with `--enable-igvm`:

```bash
git clone https://github.com/qemu/qemu.git && cd qemu
./configure --target-list=x86_64-softmmu --enable-igvm --enable-slirp
make -j$(nproc) && sudo make install
```

### OVMF

Built from this branch. Note: `AmdSev/AmdSevX64.dsc` crashes with `0x404` under IGVM regardless of generator so qw used `OvmfPkgX64.dsc`:

```bash
source edksetup.sh
build -a X64 -t GCC5 -p OvmfPkg/OvmfPkgX64.dsc -b RELEASE -DSMM_REQUIRE=FALSE
```

### IGVM

Generated with a [fork of virt-firmware-rs](https://gitlab.com/amean.asad1999/virt-firmware-rs) (`igvm-wrap` + `igvm-update`) with fixes for missing SNP page types, duplicate GPA handling, and SMP support. A PR to upstream virt-firmware-rs is pending.

```bash
igvm-wrap --snp --smp 2 -i OVMF.fd -o base.igvm
igvm-update -i base.igvm -k uki.efi -o guest.igvm
```

### UKI

Assembled with `ukify` using the host kernel and a minimal attestation initramfs:

```bash
ukify build \
  --linux /boot/vmlinuz \
  --initrd initramfs.cpio.gz \
  --cmdline "console=ttyS0 earlyprintk=serial" \
  --stub /usr/lib/systemd/boot/efi/linuxx64.efi.stub \
  --output uki.efi
```

### Boot

```bash
qemu-system-x86_64 \
  -enable-kvm -cpu EPYC-Genoa \
  -object igvm-cfg,id=igvm0,file=guest.igvm \
  -machine q35,confidential-guest-support=sev0,igvm-cfg=igvm0,memory-backend=ram1,kernel-irqchip=split \
  -object memory-backend-memfd,id=ram1,size=4G,share=true \
  -object sev-snp-guest,id=sev0,cbitpos=51,reduced-phys-bits=1 \
  -smp 2 -nographic -nodefaults -serial stdio -no-reboot
```

### Results

- **Without the patch:** Guest silently terminates during PEI (no serial output).
- **With the patch:** Full boot, SEV-SNP active at VMPL0, `sev-guest` driver initializes.
- **SMP 1 and SMP 2 verified:** Pre-computed `LAUNCH_DIGEST` matches hardware attestation report (SHA-384 at offset `0x90`).
- **Non-IGVM boot regression tested:** Standard IGVM firmware-only boot (no embedded kernel) still works — the new code path is only activated when `gEfiIgvmDataHobGuid` HOBs are present.

## Integration Instructions

This is non-blocking but leveraging the patch changes relies on an IGVM change to [virt-firmware-rs](https://gitlab.com/amean.asad1999/virt-firmware-rs) to handle missing SNP page types (`SevSvsmCca`, `SevHashes`) and duplicate GPA conflicts when embedding a kernel in IGVM. A PR to upstream virt-firmware-rs is pending. Without those fixes, `igvm-wrap`/`igvm-update` produce IGVM files that fail at `SNP_LAUNCH_UPDATE` before reaching OVMF.

QEMU must be built from upstream master with `--enable-igvm`.

cc @kraxel

### DIFF
--- a/OvmfPkg/Library/BaseMemEncryptSevLib/X64/PeiSnpSystemRamValidate.c
+++ b/OvmfPkg/Library/BaseMemEncryptSevLib/X64/PeiSnpSystemRamValidate.c
@@ -8,23 +8,37 @@
 
 **/
 
-#include <Uefi/UefiBaseType.h>
+#include <PiPei.h>
 #include <Library/BaseLib.h>
 #include <Library/PcdLib.h>
 #include <Library/DebugLib.h>
+#include <Library/HobLib.h>
 #include <Library/MemEncryptSevLib.h>
+
+#include <IndustryStandard/IgvmData.h>
 
 #include "SnpPageStateChange.h"
 #include "VirtualMemory.h"
 
 STATIC UINT8  mPscBufferPage[EFI_PAGE_SIZE];
 
+//
+// Maximum number of IGVM data ranges that can be registered as
+// pre-validated.  When booting via IGVM, pages loaded through
+// SNP_LAUNCH_UPDATE are already validated by the PSP and must
+// not be validated again.
+//
+#define MAX_IGVM_PRE_VALIDATED_RANGES  16
+
 typedef struct {
   UINT64    StartAddress;
   UINT64    EndAddress;
 } SNP_PRE_VALIDATED_RANGE;
 
-STATIC SNP_PRE_VALIDATED_RANGE  mPreValidatedRange[] = {
+//
+// Compile-time pre-validated ranges (hypervisor metadata + SEC).
+//
+STATIC SNP_PRE_VALIDATED_RANGE  mStaticRanges[] = {
   // The below address range was part of the SEV OVMF metadata, and range
   // should be pre-validated by the Hypervisor.
   {
@@ -38,6 +52,54 @@ STATIC SNP_PRE_VALIDATED_RANGE  mPreValidatedRange[] = {
   },
 };
 
+//
+// IGVM pre-validated ranges, populated at runtime from HOBs.
+//
+STATIC SNP_PRE_VALIDATED_RANGE  mIgvmRanges[MAX_IGVM_PRE_VALIDATED_RANGES];
+STATIC UINTN                    mIgvmRangeCount      = 0;
+STATIC BOOLEAN                  mIgvmRangesCollected  = FALSE;
+
+/**
+  Scan HOB list for gEfiIgvmDataHobGuid entries and populate
+  the IGVM pre-validated range array.  Called once on the first
+  invocation of MemEncryptSevSnpPreValidateSystemRam().
+
+**/
+STATIC
+VOID
+CollectIgvmPreValidatedRanges (
+  VOID
+  )
+{
+  EFI_PEI_HOB_POINTERS  Hob;
+  EFI_IGVM_DATA_HOB     *IgvmData;
+
+  mIgvmRangesCollected = TRUE;
+
+  for (Hob.Raw = GetFirstGuidHob (&gEfiIgvmDataHobGuid);
+       Hob.Raw != NULL;
+       Hob.Raw = GetNextGuidHob (&gEfiIgvmDataHobGuid, GET_NEXT_HOB (Hob)))
+  {
+    if (mIgvmRangeCount >= ARRAY_SIZE (mIgvmRanges)) {
+      DEBUG ((DEBUG_WARN, "%a: Too many IGVM ranges, some will be re-validated\n", __func__));
+      break;
+    }
+
+    IgvmData = GET_GUID_HOB_DATA (Hob.Guid);
+    mIgvmRanges[mIgvmRangeCount].StartAddress = IgvmData->Address;
+    mIgvmRanges[mIgvmRangeCount].EndAddress   = IgvmData->Address +
+                                                  ALIGN_VALUE (IgvmData->Length, EFI_PAGE_SIZE);
+    DEBUG ((
+      DEBUG_INFO,
+      "%a: IGVM pre-validated range [0x%lx, 0x%lx)\n",
+      __func__,
+      mIgvmRanges[mIgvmRangeCount].StartAddress,
+      mIgvmRanges[mIgvmRangeCount].EndAddress
+      ));
+    mIgvmRangeCount++;
+  }
+}
+
 STATIC
 BOOLEAN
 DetectPreValidatedOverLap (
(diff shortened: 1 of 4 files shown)


OvmfPkg/PlatformPei/AmdSev.c
```diff
@@ -153,10 +254,20 @@ AmdSevSnpInitialize (
           continue;
         }
 
-        MemEncryptSevSnpPreValidateSystemRam (
```
kraxel, line 156: MemEncryptSevSnpPreValidateSystemRam already has logic to skip pre-validated memory regions (see `OvmfPkg/Library/BaseMemEncryptSevLib/X64/PeiSnpSystemRamValidate.c`).

I think it is much simpler to let the `DetectPreValidatedOverLap` function loop over the ranges registered via EfiIgvmDataHobGuid (in addition to the compiled-in exclude ranges).

AmeanAsad: Thanks @kraxel for taking a look here. I reworked the changes. The change now moves the skip logic into `DetectPreValidatedOverLap` in `PeiSnpSystemRamValidate.c`. `AmdSev.c` and `PlatformPei.inf` are reverted to upstream.

On the first call to `MemEncryptSevSnpPreValidateSystemRam()`, a new `CollectIgvmPreValidatedRanges()` scans the HOB list for `gEfiIgvmDataHobGuid` entries and populates a separate `mIgvmRanges[]` array. `DetectPreValidatedOverLap()` then checks both `mStaticRanges[]` (compile-time) and `mIgvmRanges[]` (runtime), two independent arrays with no hardcoded count to keep in sync with the initializer list.

`mIgvmRanges` is a static array, so no allocator is needed in the early PEI validation path. 16 slots is generous for the current IGVM use case (typically 1-2 data blobs). If exceeded, a `DEBUG_WARN` is emitted and extra ranges get revalidated (safe, just slower). Happy to switch to dynamic allocation if preferred.

Re-verified on hardware (AMD EPYC 8224P, SEV firmware 1.55 build 49). SMP 1 and SMP 2 launch digests match attestation reports. Non-IGVM boot regression tested (no HOBs found, `mIgvmRangeCount` stays at 0).

OvmfPkg/Library/BaseMemEncryptSevLib/X64/PeiSnpSystemRamValidate.c
```diff
@@ -38,6 +52,54 @@ STATIC SNP_PRE_VALIDATED_RANGE  mPreValidatedRange[] = {
   },
 };
 
+//
+// IGVM pre-validated ranges, populated at runtime from HOBs.
+//
+STATIC SNP_PRE_VALIDATED_RANGE  mIgvmRanges[MAX_IGVM_PRE_VALIDATED_RANGES];
```
kraxel, line 58: Writing to global variables is not allowed in PEI phase.

I'd suggest to remove the CollectIgvmPreValidatedRanges() function and move the code which loops over the HOBs to DetectPreValidatedOverLap() instead.

kraxel: > I'd suggest to remove the CollectIgvmPreValidatedRanges() function and move the code which loops over the HOBs to DetectPreValidatedOverLap() instead.

Implemented in https://github.com/tianocore/edk2/pull/12376

AmeanAsad: @kraxel shall we close this then?

kraxel: Yes, I think this can be closed.  Can you double-check whenever  https://github.com/tianocore/edk2/pull/12376 works for you too?

AmeanAsad: for sure, let me verify

AmeanAsad:
closing in favor of https://github.com/tianocore/edk2/pull/12376

thanks @kraxel

PR #12284, opened 2026-03-12
Global: Move Azure Pipelines vmImage defaults to defaults.yml

mikebeaton:
# Description

The pool vmImage for running Azure Pipelines jobs is always either 'windows-2022' or 'ubuntu-24.02'. Move these default values to defaults.yml so they can be changed in one location in future.

This change was suggested by @Javagedes in [#11886](https://github.com/tianocore/edk2/pull/11886#discussion_r2907608054).

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

CI

## Integration Instructions

N/A

### DIFF
--- a/.azurepipelines/templates/basetools-build-steps.yml
+++ b/.azurepipelines/templates/basetools-build-steps.yml
@@ -1,5 +1,5 @@
 ## @file
-# File templates/basetools-build-job.yml
+# File templates/basetools-build-steps.yml
 #
 # template file to build basetools
 #
--- a/.azurepipelines/templates/platform-build-run-steps.yml
+++ b/.azurepipelines/templates/platform-build-run-steps.yml
@@ -1,6 +1,6 @@
 
 ## @file
-# File steps.yml
+# File templates/platform-build-run-steps.yml
 #
 # template file containing the steps to build
 #
@@ -41,9 +41,8 @@ parameters:
 steps:
 - bash: |
     echo "##vso[task.prependpath]${HOME}/.local/bin"
-    echo "new PATH=${PATH}"
   displayName: Set PATH
-  condition: eq('${{ parameters.tool_chain_tag }}', 'GCC')
+  condition: eq(variables['Agent.OS'], 'Linux')
 
 - checkout: self
   clean: true
--- a/.azurepipelines/templates/pr-gate-steps.yml
+++ b/.azurepipelines/templates/pr-gate-steps.yml
@@ -19,9 +19,8 @@ parameters:
 steps:
 - bash: |
     echo "##vso[task.prependpath]${HOME}/.local/bin"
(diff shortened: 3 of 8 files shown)


mikebeaton:
cc @Javagedes @makubacki - Note that after this PR as it currently is, we still have:

https://github.com/tianocore/edk2/blob/master/.azurepipelines/Ubuntu-PatchCheck.yml#L22
https://github.com/tianocore/edk2/blob/master/.azurepipelines/templates/pr-gate-build-job.yml#L137
https://github.com/tianocore/edk2/blob/master/.github/workflows/codeql.yml#L28

I am not sure if any of the above, most likely the use of `windows-2022` in `pr-gate-build-job.yml`, should be integrated into this PR? If so, it is not yet clear to me how to do so cleanly.

(EDIT: I guess I could re-load defaults.yml in pr-gate-build-job.yml, even though it has in fact already been loaded by the caller in all cases?)

mikebeaton:
> (EDIT: I guess I could re-load defaults.yml in pr-gate-build-job.yml, even though it has in fact already been loaded by the caller in all cases?)

I've added a third commit which does this, to serve as something more concrete to comment on. [EDIT: This does not work, see below.]

mikebeaton:
Apologies, my attempted third commit did not work. I believe reloading variables in a template used inside a matrix was the problem, in any event multiple steps did not run:

image

So I believe we either have to:

 1. Leave `windows-2022` specified separately once inside `pr-gate-build-job.yml`, even though it's clearly logically a re-use of the default value
 1. Access the `windows-2022` value inside `pr-gate-build-job.yml` via `variables` from `defaults.yml` already loaded by the callers (which presupposes that all callers will always load this)
 1. Pass in the `window-2022` value, from the defaults, from every caller as a parameter
 
None are ideal.

mikebeaton:
> So I believe we either have to:
> 
> 1. Leave `windows-2022` specified separately once inside `pr-gate-build-job.yml`, even though it's clearly logically a re-use of the default value
> 2. Access the `windows-2022` value inside `pr-gate-build-job.yml` via `variables` from `defaults.yml` already loaded by the callers (which presupposes that all callers will always load this)
> 3. Pass in the `window-2022` value, from the defaults, from every caller as a parameter
> 
> None are ideal.

I have tried implementing 2 and that _also_ does not work (I guess `variables` is evaluated before it is loaded). So I've gone back to leaving the `windows-2022` value manually specified inside `pr-gate-build-job.yml`, even though it's clearly logically a re-use of the default value.

Javagedes:
Thank you for the change. I will leave open for 1 day to let others review if they wish, then I will add the push tag. If I don't do it within about an hour of this time tomorrow just @ me :)

PR #12286, opened 2026-03-12
.azurepipelines: Make 'Set PATH' step check OS rather than toolchain

mikebeaton:
# Description

The 'Set PATH' step added by becff4f473e27ba0a0d8a40e0071531abdb67872 is explicitly Linux only, so here we update `pr-gate-steps.yml` and `platform-build-run-steps.yml` to directly check `Agent.OS` rather than relying on the toolchain as an indicator of the OS, since the existing logic amounts to assuming that toolchains are OS specific, which is incorrect and will matter more now that CLANGPDB is planned to be more extensively used on Windows.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

CI

## Integration Instructions

N/A

### DIFF
--- a/.azurepipelines/templates/platform-build-run-steps.yml
+++ b/.azurepipelines/templates/platform-build-run-steps.yml
@@ -43,7 +43,7 @@ steps:
     echo "##vso[task.prependpath]${HOME}/.local/bin"
     echo "new PATH=${PATH}"
   displayName: Set PATH
-  condition: eq('${{ parameters.tool_chain_tag }}', 'GCC')
+  condition: eq(variables['Agent.OS'], 'Linux')
 
 - checkout: self
   clean: true
--- a/.azurepipelines/templates/pr-gate-steps.yml
+++ b/.azurepipelines/templates/pr-gate-steps.yml
@@ -21,7 +21,7 @@ steps:
     echo "##vso[task.prependpath]${HOME}/.local/bin"
     echo "new PATH=${PATH}"
   displayName: Set PATH
-  condition: eq('${{ parameters.tool_chain_tag }}', 'GCC')
+  condition: eq(variables['Agent.OS'], 'Linux')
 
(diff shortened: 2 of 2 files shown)


mikebeaton:
@Javagedes - Please can you check that this is what you were thinking of for https://github.com/tianocore/edk2/pull/11886#discussion_r2907600444?

Javagedes:
@mikebeaton I think we can get away with just checking the predefined variable: Agent os

`eq(variables['Agent.OS'], 'Linux')`

https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#agent-variables-devops-services

mikebeaton:
@Javagedes - Updated - thanks!

mikebeaton:
I have just spotted that the `echo "new PATH=${PATH}"` line used in both `Set PATH` definitions is misleading: it echoes the existing PATH without the changes - which is apparently as expected, as `task.prependpath` [modifies PATH for subsequent steps](https://learn.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#prependpath-prepend-a-path-to-the--path-environment-variable), and not for the current step - but it means the existing echo is misleadingly worded, at least.

image

To log it, it seems we would either have to replicate the calculation (so not really logging what `prependpath` has set anyway, but hopefully logging what it would set...) or we have to add another step (which seems like overkill...?).

Arguably, removing the echo is the correct choice? Or at least changing the wording? (Or ignoring this....!) Does anyone have an opinion? (None seem like brilliant choices.)

mikebeaton:
> I have just spotted that the echo "new PATH=${PATH}" line used in both Set PATH definitions is misleading: it echoes the existing PATH without the changes

https://github.com/tianocore/edk2/pull/12287

PR #12287, opened 2026-03-12
.azurepipelines/templates: Remove `echo new PATH` from 'Set PATH' task

mikebeaton:
# Description

The `echo "new PATH=${PATH}"` line  used in the `Set PATH` tasks added by becff4f473e27ba0a0d8a40e0071531abdb67872 is misleading: it echoes the pre-existing PATH without any changes - which is as expected, as `task.prependpath` modifies PATH only for subsequent tasks: https://learn.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#prependpath-prepend-a-path-to-the--path-environment-variable

image

Alternative approaches would be:
 - Ignore this
 - Duplicate the calculation so that we can echo what will be set
 - Update the message to say that it echoes the path before the change
 - Add an extra step which displays the path after it has been set

This PR proposes just removing the echo, since it is misleading as is, and it seems that no proportionately lightweight fix is available.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

CI

## Integration Instructions

N/A

### DIFF
--- a/.azurepipelines/templates/platform-build-run-steps.yml
+++ b/.azurepipelines/templates/platform-build-run-steps.yml
@@ -41,7 +41,6 @@ parameters:
 steps:
 - bash: |
     echo "##vso[task.prependpath]${HOME}/.local/bin"
-    echo "new PATH=${PATH}"
   displayName: Set PATH
   condition: eq(variables['Agent.OS'], 'Linux')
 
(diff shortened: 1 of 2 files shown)


mikebeaton:
@Javagedes - It does not seem keen to push my stuff! Can you prompt mergify again? Many thanks.

mikebeaton:
Ah - rebase did it. I'll try that myself next time. Many thanks.

PR #12289, opened 2026-03-13
ShellPkg/UefiHandleParsingLib: support handle list growth

bobm21:
Refactor GetHandleListByProtocol() to support additions to the handle list during its execution. Replace LocateHandle() with LocateHandleBuffer() to avoid the possibility that the buffer allocated for LocateHandle() is too small if additional handles
are added during GetHandleListByProtocol() execution.

Note that the previous implementation did not detect the handle list growth and would cause memory
corruption when writing the terminating NULL handle to the allocated buffer.

# Description

Fixes an intermittent problem where the second LocateHandle call in the EFI_BUFFER_TOO_SMALL handling also failed with  EFI_BUFFER_TOO_SMALL due to a handle being added between the two LocateHandle calls, but since Status of the second call is not checked, the  writing of the NULL terminating entry using the new Size value in HandleList[Size/sizeof (EFI_HANDLE)] corrupted the allocated buffer trailer causing an ASSERT when the HandleList was freed.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Re-ran the repetitive test that had been hitting the intermittent failure.

## Integration Instructions

N/A

### DIFF
--- a/ArmVirtPkg/ArmVirtQemu.dsc
+++ b/ArmVirtPkg/ArmVirtQemu.dsc
@@ -33,6 +33,7 @@
   DEFINE TPM2_ENABLE             = FALSE
   DEFINE TPM2_CONFIG_ENABLE      = FALSE
   DEFINE CAVIUM_ERRATUM_27456    = FALSE
+  DEFINE DEBUG_TO_MEM            = FALSE
 
   #
   # Network definition
--- a/BaseTools/Source/C/Makefiles/ms.app
+++ b/BaseTools/Source/C/Makefiles/ms.app
@@ -10,8 +10,8 @@ APPLICATION = $(BIN_PATH)\$(APPNAME).exe
 all: $(APPLICATION) 
 
 $(APPLICATION) : $(OBJECTS) 
-	-@if not exist $(BIN_PATH) mkdir $(BIN_PATH)
-	$(LD) /nologo /debug /OPT:REF /OPT:ICF=10 /incremental:no /nodefaultlib:libc.lib /out:$@ $(LIBS) $**
+  -@if not exist $(BIN_PATH) mkdir $(BIN_PATH)
+  $(LD) /nologo /debug /DYNAMICBASE /guard:cf /OPT:REF /OPT:ICF=10 /incremental:no /nodefaultlib:libc.lib /out:$@ $(LIBS) $**
 
 $(OBJECTS) : $(SOURCE_PATH)\Include\Common\BuildVersion.h
 
@@ -19,10 +19,10 @@ $(OBJECTS) : $(SOURCE_PATH)\Include\Common\BuildVersion.h
 .PHONY:cleanall
 
 clean:
-	del /f /q $(OBJECTS) *.pdb > nul
+  del /f /q $(OBJECTS) *.pdb > nul
 
 cleanall:
-	del /f /q $(OBJECTS) $(APPLICATION) *.pdb $(BIN_PATH)\$(APPNAME).pdb > nul
+  del /f /q $(OBJECTS) $(APPLICATION) *.pdb $(BIN_PATH)\$(APPNAME).pdb > nul
 
 !INCLUDE $(SOURCE_PATH)\Makefiles\ms.rule
 
--- a/BaseTools/Source/C/Makefiles/ms.common
+++ b/BaseTools/Source/C/Makefiles/ms.common
@@ -46,6 +46,8 @@ SYS_BIN_PATH = $(EDK_TOOLS_PATH)\Bin\Win32
 SYS_LIB_PATH = $(EDK_TOOLS_PATH)\Lib\Win32
 # Note: Disable flexible array member warnings
 CFLAGS = $(CFLAGS) /wd4200
+# Note: Suppress function implementation and prototype inconsistency warnings
+CFLAGS = $(CFLAGS) /wd4028
 
 !ELSEIF "$(HOST_ARCH)"=="X64"
 ARCH_INCLUDE = $(EDK2_PATH)\MdePkg\Include\X64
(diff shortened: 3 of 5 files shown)


ShellPkg/Library/UefiHandleParsingLib/UefiHandleParsingLib.c
```diff
@@ -3771,45 +3771,37 @@ GetHandleListByProtocol (
...
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "%a: LocateHandleBuffer %a failed: %r\n", __func__, (ProtocolGuid == NULL) ? "AllHandles" : "ByProtocol", Status));
+    return NULL;
+  }
 
-      Status                               = gBS->LocateHandle (ByProtocol, (EFI_GUID *)ProtocolGuid, NULL, &Size, HandleList);
-      HandleList[Size/sizeof (EFI_HANDLE)] = NULL;
-    }
+  // create new list with one more slot for the NULL terminator
+  HandleList = AllocateZeroPool ((OriginalHandleCount + 1) * sizeof (EFI_HANDLE));
```
mdkinney: Recommend using ReallocatePool().  It contains all the logic including AllocateZeroPool() of new buffer.

bobm21: Sounds good, updated to use ReallocatePool.

ShellPkg/Library/UefiHandleParsingLib/UefiHandleParsingLib.c
```diff
@@ -3771,45 +3771,37 @@ GetHandleListByProtocol (
...
+  }
 
-      Status                               = gBS->LocateHandle (ByProtocol, (EFI_GUID *)ProtocolGuid, NULL, &Size, HandleList);
-      HandleList[Size/sizeof (EFI_HANDLE)] = NULL;
-    }
+  // create new list with one more slot for the NULL terminator
+  HandleList = ReallocatePool (OriginalHandleCount * sizeof (EFI_HANDLE), (OriginalHandleCount + 1) * sizeof (EFI_HANDLE), OriginalHandleList);
+  if (HandleList == NULL) {
+    DEBUG ((DEBUG_ERROR, "%a: reallocate failed\n", __func__));
+    goto Done;
```
mdkinney: No need for Done label and goto now and the FreePool() call below should not be done because ReallocatePool() handles that logic.

ShellPkg/Library/UefiHandleParsingLib/UefiHandleParsingLib.c
```diff
@@ -3771,45 +3771,37 @@ GetHandleListByProtocol (
...
   EFI_HANDLE  *HandleList;
-  UINTN       Size;
+  EFI_HANDLE  *OriginalHandleList;
+  UINTN       OriginalHandleCount;
   EFI_STATUS  Status;
 
-  Size       = 0;
-  HandleList = NULL;
+  HandleList         = NULL;
+  OriginalHandleList = NULL;
```
mdkinney, line 3778: Do not initialize local variables in declaration.  Set OriginalHandleList to NULL just before calling LocateHandleBuffer().  HandleList does not need to be set to NULL.  It is always assign the value returned from ReallocatePool().

ShellPkg/Library/UefiHandleParsingLib/UefiHandleParsingLib.c
```diff
@@ -3771,45 +3771,37 @@ GetHandleListByProtocol (
...
-      }
+  Status = gBS->LocateHandleBuffer (
+                  (ProtocolGuid == NULL) ? AllHandles : ByProtocol,
+                  (EFI_GUID *)ProtocolGuid,
+                  NULL,
+                  &OriginalHandleCount,
+                  &OriginalHandleList
+                  );
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "%a: LocateHandleBuffer %a failed: %r\n", __func__, (ProtocolGuid == NULL) ? "AllHandles" : "ByProtocol", Status));
```
mdkinney: Split DEBUG() call into multiple lines.

ShellPkg/Library/UefiHandleParsingLib/UefiHandleParsingLib.c
```diff
@@ -3771,45 +3771,37 @@ GetHandleListByProtocol (
...
+    return NULL;
+  }
 
-      Status                               = gBS->LocateHandle (ByProtocol, (EFI_GUID *)ProtocolGuid, NULL, &Size, HandleList);
-      HandleList[Size/sizeof (EFI_HANDLE)] = NULL;
-    }
+  // create new list with one more slot for the NULL terminator
+  HandleList = ReallocatePool (OriginalHandleCount * sizeof (EFI_HANDLE), (OriginalHandleCount + 1) * sizeof (EFI_HANDLE), OriginalHandleList);
```
mdkinney: Split ReallocatePool() call into multiple lines.

ShellPkg/Library/UefiHandleParsingLib/UefiHandleParsingLib.c
bobm21: I moved FreePool to ReallocatePool error handling for clarity.

ShellPkg/Library/UefiHandleParsingLib/UefiHandleParsingLib.c
```diff
@@ -3771,45 +3771,40 @@ GetHandleListByProtocol (
...
+  Status             = gBS->LocateHandleBuffer (
+                              (ProtocolGuid == NULL) ? AllHandles : ByProtocol,
+                              (EFI_GUID *)ProtocolGuid,
+                              NULL,
+                              &OriginalHandleCount,
+                              &OriginalHandleList
+                              );
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
```
makubacki: If `gBS->LocateHandleBuffer()` returns `EFI_NOT_FOUND`, I would consider not logging this message at `DEBUG_ERROR` level. That's more of a valid outcome than an error.

bobm21: conditioned error output on !EFI_NOT_FOUND

mdkinney:
@bobm21 Why forced push changes after review?

bobm21:
> @bobm21 Why forced push changes after review?

Sorry, wasn't sure if I needed to rebase on tip.

mdkinney:
> > @bobm21 Why forced push changes after review?
> 
> Sorry, wasn't sure if I needed to rebase on tip.

Mergify handles that automatically when push label is set.

PR #12291, opened 2026-03-13
Revert "Move RealTimeClockLib header"

ardbiesheuvel:
This reverts commit 57230fff6b39a665485be1bd43ec608d412ba6fb.

RealTimeClockLib is not a generic library class that is intended to be used widely to get access to the RTC when the associated runtime services are not available in the calling context.

The purpose of RealTimeClockLib is to abstract the underlying hardware access from the generic RealTimeClockRuntimeDxe driver, which backs the runtime services.

This means it does not belong in a different package; it belongs in the same package as the driver.


Flickdm (approved):

Flickdm:
Does this need to be marked a breaking change?

felixpgithub:
Ard,

I agree with your argumentation, but both RealTimeClockLib library class and RealTimeClockRuntimeDxe driver have no architectural dependencies. So, rather than moving RealTimeClockLib back to the EmbeddedPkg, shouldn't we instead move RealTimeClockRuntimeDxe driver to the MdeModulePkg?

leiflindholm:
> Ard,
> 
> I agree with your argumentation, but both RealTimeClockLib library class and RealTimeClockRuntimeDxe driver have no architectural dependencies. So, rather than moving RealTimeClockLib back to the EmbeddedPkg, shouldn't we instead move RealTimeClockRuntimeDxe driver to the MdeModulePkg?

I was about to suggest the same.

There are several users of the header in edk2-platforms, and not all of them especially embedded.

Flickdm:
@felixpgithub @leiflindholm 

Not sure if you saw the full background but this was what sparked the conversation - not that I think this changes your positions
https://github.com/tianocore/edk2/pull/12271

ardbiesheuvel:
> > Ard,
> > I agree with your argumentation, but both RealTimeClockLib library class and RealTimeClockRuntimeDxe driver have no architectural dependencies. So, rather than moving RealTimeClockLib back to the EmbeddedPkg, shouldn't we instead move RealTimeClockRuntimeDxe driver to the MdeModulePkg?
> 
> I was about to suggest the same.
> 
> There are several users of the header in edk2-platforms, and not all of them especially embedded.

Each of those platforms also includes `EmbeddedPkg/RealTimeClockRuntimeDxe/RealTimeClockRuntimeDxe.inf`, which should be the only consumer of the library implementation.

If that belongs in MdeModulePkg too, we can migrate it but /that/ would be a breaking change.

I fail to see the motivation for the original commit, to migrate a library class that is only intended for use by a driver in EmbeddedPkg, and it is inviting misuse.

felixpgithub:
> I fail to see the motivation for the original commit, to migrate a library class that is only intended for use by a driver in EmbeddedPkg, and it is inviting misuse.

Yes, I think the driver and the library should have been moved together.
It's similar to the ResetSystemLib/ResetSystemRuntimeDxe that are already in the MdeModulePkg.

Migrating RealTimeClockRuntimeDxe to MdeModulePkg would also help to reduce fragmentation between x64 and ARM projects. x64 projects are now using PcAtChipsetPkg/PcatRealTimeClockRuntimeDxe, which can be replaced with RealTimeClockRuntimeDxe.inf and PC/AT compatible instance of RealTimeClockLib.
It can be done without moving anything, but then project are getting dependency on EmbeddedPkg.

Flickdm:
Out of curiosity while I plan my next steps did we ever resolve this or are just letting this go?

ardbiesheuvel:
I'd like to get this revert in, and then we can fix this properly during the next cycle.

@lgao4

leiflindholm (approved):
Agree to revisit proper solution next cycle.

ardbiesheuvel:
@mdkinney @leiflindholm this looks like a spurious patchcheck failure

leiflindholm:
> @mdkinney @leiflindholm this looks like a spurious patchcheck failure

It's complaining that ArmPlatformPkg isn't enumerated in the subject line.

ardbiesheuvel:
> > @mdkinney @leiflindholm this looks like a spurious patchcheck failure
> 
> It's complaining that ArmPlatformPkg isn't enumerated in the subject line.

I removed it because it was complaining before that the subject line was too long.

leiflindholm:
> > > @mdkinney @leiflindholm this looks like a spurious patchcheck failure
> > 
> > 
> > It's complaining that ArmPlatformPkg isn't enumerated in the subject line.
> 
> I removed it because it was complaining before that the subject line was too long.

Ah, that case... 

So, I added magic treatment of Revert commits since this was a likely outcome ... but that only kicks in when the subject line starts with "Revert" (b55530ad4460).

lgao4:
> I'd like to get this revert in, and then we can fix this properly during the next cycle.
> 
> @lgao4

I am OK.

lgao4 (approved):

Issue #7580, opened 2018-10-19
IntelFsp2Pkg: FSP should not override IDT (Bugzilla Bug 1265)


tianocore-issues:
This issue was created automatically with [bugzilla2github](https://github.com/makubacki/bugzilla2github/tree/bz2gh_tianocore)
# Bugzilla Bug 1265
Date: 2018-10-19T03:23:45+00:00
From: knifeeight22
To:   @ChaselChiu
CC:   @ChaselChiu

Last updated: 2018-12-17T18:34:28+00:00

tianocore-issues:
## Comment 5330
Date: 2018-10-19 03:23:45 +0000
From: knifeeight22

- Industry Specification: ---
- Target OS: ---
- Bugzilla Assignee(s): @ChaselChiu

FSP should not override IDT table when it is initialized by boot loader.
IDT should be re-initialized in FSP only when it is invalid.

tianocore-issues:
## Comment 5385
Date: 2018-10-25 05:05:54 +0000
From: knifeeight22

commit b1cc6f672f3b924cdb190e5b92db3b47f46a8911 (HEAD -&gt; master, origin/master, origin/HEAD)
Author: Chasel, Chiu &lt;chasel.chiu@intel.com&gt;
Date:   Fri Oct 19 17:10:30 2018 +0800

    IntelFsp2Pkg: FSP should not override IDT

    REF: https://bugzilla.tianocore.org/show_bug.cgi?id=1265

    FSP should not override IDT table when it is initialized
    by boot loader. IDT should be re-initialized in FSP only
    when it is invalid.
    To mitigate temporary memory usage a PCD
    PcdFspMaxInterruptSupported created for platform to decide
    how many interrupts the FSP IDT table can support.

    Test: Verified on internal platform and boots successfully.

    Cc: Jiewen Yao &lt;Jiewen.yao@intel.com&gt;
    Cc: Desimone Nathaniel L &lt;nathaniel.l.desimone@intel.com&gt;
    Contributed-under: TianoCore Contribution Agreement 1.1
    Signed-off-by: Chasel Chiu &lt;chasel.chiu@intel.com&gt;
    Reviewed-by: Jiewen Yao &lt;Jiewen.yao@intel.com&gt;

PR #12295, opened 2026-03-16
IntelFsp2Pkg/FspSecCore: Do not hang when bootloader IDT is larger

niruiyu:
# Description

The current logic in FspSecCore is when the size of IDT created by the bootloader is larger than the size of IDT that's going to be created by FSP, CpuDeadLoop() is hit.

The dead-loop is not necessary. The patch updates the logic to only copy the first part of the IDT entries when bootloader's IDT is larger.

The change addresses a boot hang in FSP API mode introduced by commit 3454d7ab41.
- [N] Breaking change?
- [N] Impacts security?
- [N] Includes tests?

## How This Was Tested

Verify the system does not hang any more in FSP API mode.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Library/ArmLib/AArch64/ArmLibSupport.S
+++ b/ArmPkg/Library/ArmLib/AArch64/ArmLibSupport.S
@@ -15,6 +15,10 @@
 
 .set SCTLR_ELx_M_BIT_POS, (0)
 
+ASM_FUNC(ArmReadAidr)
+  mrs     x0, aidr_el1        // Read from Auxiliary ID Register (AIDR_EL1)
+  ret
+
 ASM_FUNC(ArmReadMidr)
   mrs     x0, midr_el1        // Read from Main ID Register (MIDR)
   ret
--- a/MdeModulePkg/Core/Pei/Ppi/Ppi.c
+++ b/MdeModulePkg/Core/Pei/Ppi/Ppi.c
@@ -1038,6 +1038,7 @@ ProcessPpiListFromSec (
   EFI_STATUS              Status;
   EFI_SEC_HOB_DATA_PPI    *SecHobDataPpi;
   EFI_HOB_GENERIC_HEADER  *SecHobList;
+  PEI_CORE_INSTANCE       *PrivateData;
 
   for ( ; ;) {
     if ((PpiList->Flags & EFI_PEI_PPI_DESCRIPTOR_NOTIFY_TYPES) != 0) {
@@ -1052,6 +1053,15 @@ ProcessPpiListFromSec (
       //
       Status = InternalPeiInstallPpi (PeiServices, PpiList, TRUE);
       ASSERT_EFI_ERROR (Status);
+
+      //
+      // Set the PeiMemoryInstalled flag if SEC passed MemoryDiscoveredPpi, so the PEI runs in
+      // physical memory.
+      //
+      if (CompareGuid (PpiList->Guid, &gEfiPeiMemoryDiscoveredPpiGuid)) {
+        PrivateData                     = PEI_CORE_INSTANCE_FROM_PS_THIS (PeiServices);
+        PrivateData->PeiMemoryInstalled = TRUE;
+      }
     }
 
     if ((PpiList->Flags & EFI_PEI_PPI_DESCRIPTOR_TERMINATE_LIST) == EFI_PEI_PPI_DESCRIPTOR_TERMINATE_LIST) {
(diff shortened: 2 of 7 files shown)


niruiyu:
> This change can cause issues with software debuggers. If the bootloader IDT exceeds the FSP IDT _**and**_ those IDT entries are used in the implementation of a software debugger, then we will cause a regression. We originally fixed this software debugger + FSP API mode issue in #7580 (commit [b1cc6f6](https://github.com/tianocore/edk2/commit/b1cc6f672f3b924cdb190e5b92db3b47f46a8911)).
> 
> I prefer that we simply set `PcdFspMaxInterruptSupported` to 256 in `IntelFsp2Pkg.dec` instead of this change. The IDT is architecturally defined to contain at most 256 entries, so setting that PCD to 256 will make it impossible to hit the dead loop.

@nate-desimone Setting `PcdFspMaxInterruptSupported` to 256 would cause more temporary RAM being used (0x10 bytes per entry, the change increases # of entry from 34 to 256, resulting extra +3552B temporary RAM usage). I am not sure if it could be "the last straw" in some platforms.
Current PCD value 34 matches to the maximum vector 33 chosen by SourceLevelDebugPkg.
The patch is to try-best inherit the interrupt handlers set up by the bootloader, instead of dead-loop. I think it's an improvement to current implementation which simply hangs.
To support other implementation of soft debuggers, the certain FSP implementation can increase the PCD value. The extra temporary RAM usage should not be the default implementation, but can be configured by the certain FSP implementation.

nate-desimone:
> > This change can cause issues with software debuggers. If the bootloader IDT exceeds the FSP IDT _**and**_ those IDT entries are used in the implementation of a software debugger, then we will cause a regression. We originally fixed this software debugger + FSP API mode issue in #7580 (commit [b1cc6f6](https://github.com/tianocore/edk2/commit/b1cc6f672f3b924cdb190e5b92db3b47f46a8911)).
> > I prefer that we simply set `PcdFspMaxInterruptSupported` to 256 in `IntelFsp2Pkg.dec` instead of this change. The IDT is architecturally defined to contain at most 256 entries, so setting that PCD to 256 will make it impossible to hit the dead loop.
> 
> @nate-desimone Setting `PcdFspMaxInterruptSupported` to 256 would cause more temporary RAM being used (0x10 bytes per entry, the change increases # of entry from 34 to 256, resulting extra +3552B temporary RAM usage). I am not sure if it could be "the last straw" in some platforms. Current PCD value 34 matches to the maximum vector 33 chosen by SourceLevelDebugPkg. The patch is to try-best inherit the interrupt handlers set up by the bootloader, instead of dead-loop. I think it's an improvement to current implementation which simply hangs. To support other implementation of soft debuggers, the certain FSP implementation can increase the PCD value. The extra temporary RAM usage should not be the default implementation, but can be configured by the certain FSP implementation.

There are many commits to IntelFsp2Pkg that have increased the usage of temporary RAM over time. For example, commit b1cc6f6 added 512B. FSP implementations where the extra ~3.5KB is "the last straw" are welcome to set it back to 34 in their own DSC file. Therefore, I would like to see the following:

1. The default value of `PcdFspMaxInterruptSupported` shall increase to 256. If an FSP wants to continue to use 34 then they can set it back to 34 in their own DSC file. There have been many edk2 changes recently that has required porting changes on the platform/silicon side, so this one does not concern me.
2. Because `PcdFspMaxInterruptSupported` is configurable, I agree that we should keep the logic to only copy the first part of the IDT entries when bootloader's IDT is larger.
3. If the  bootloader's IDT is larger than the FSP's, then we should output a `DEBUG ((DEBUG_WARN, ...))` message.

SaiChaganty:
I do like the idea of increasing PcdFspMaxInterruptSupported to a large value (max is also ok) and let FSP implementation control this as .dsc level. Won't hurt to leave a note in the .dec file. This would:
- Be minimal change, compared to current one
- ensures BL IDT entries can still be entertained, if desired.
- Hits CpuDeadLopp() for exception cases only.

I am not sure what side effects there will be if we do a MIN (BootLoader IDT size , FSP IDT size).

niruiyu:
> I do like the idea of increasing PcdFspMaxInterruptSupported to a large value (max is also ok) and let FSP implementation control this as .dsc level. Won't hurt to leave a note in the .dec file. This would:
> 
> * Be minimal change, compared to current one
> * ensures BL IDT entries can still be entertained, if desired.
> * Hits CpuDeadLopp() for exception cases only.
> 
> I am not sure what side effects there will be if we do a MIN (BootLoader IDT size , FSP IDT size).

MIN() takes the interrupt handlers set up by bootloader as many as possible.
Since .DEC already sets to 255 which is the maximum value supported by X86 CPU, only when .DSC overrides it to a smaller value and bootloader does provide larger number of interrupt handlers, the MIN() will take actual effect.

I do not think there will be any real issue.

niruiyu:
> Implementation looks good! Please change the default value of `PcdFspMaxInterruptSupported` in `IntelFsp2Pkg.dec` and I will approve the PR.

@nate-desimone DEC change was missed in previous commit. Now it's included. please re-check.

PR #12297, opened 2026-03-16
StandaloneMmPkg Core: Check PE section finding status

lzeng14:
StandaloneMmPkg Core: Check PE section finding status
The code should check PE section finding status instead of
DEPEX section finding status for the driver to be added into list.
It is a supplementary patch for https://github.com/tianocore/edk2/commit/37db80097dbe4d1487191422240bdc0a9ad8d963 to support no DEPEX case.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Create no depex case and make sure it can be dispatched correctly.

## Integration Instructions
N/A

### DIFF
--- a/StandaloneMmPkg/Core/FwVol.c
+++ b/StandaloneMmPkg/Core/FwVol.c
@@ -222,7 +222,12 @@ MmCoreFfsFindMmDriver (
       Status = FfsFindSectionData (EFI_SECTION_PE32, FileHeader, &Pe32Data, &Pe32DataSize);
       DEBUG ((DEBUG_INFO, "Find PE data - 0x%x\n", Pe32Data));
       DepexStatus = FfsFindSectionData (EFI_SECTION_MM_DEPEX, FileHeader, &Depex, &DepexSize);
-      if (!EFI_ERROR (DepexStatus)) {
+      if (DepexStatus == EFI_NOT_FOUND) {
+        Depex     = NULL;
+        DepexSize = 0;
+      }
+
+      if (!EFI_ERROR (Status)) {
         MmAddToDriverList (FwVolHeader, Pe32Data, Pe32DataSize, Depex, DepexSize, &FileHeader->Name);
       }
     }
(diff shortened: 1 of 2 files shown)


StandaloneMmPkg/Core/FwVol.c
```diff
@@ -221,8 +220,8 @@ MmCoreFfsFindMmDriver (
     if (!EFI_ERROR (Status)) {
       Status = FfsFindSectionData (EFI_SECTION_PE32, FileHeader, &Pe32Data, &Pe32DataSize);
       DEBUG ((DEBUG_INFO, "Find PE data - 0x%x\n", Pe32Data));
-      DepexStatus = FfsFindSectionData (EFI_SECTION_MM_DEPEX, FileHeader, &Depex, &DepexSize);
-      if (!EFI_ERROR (DepexStatus)) {
+      FfsFindSectionData (EFI_SECTION_MM_DEPEX, FileHeader, &Depex, &DepexSize);
```
niruiyu: DepexSize might be random value if Depex section cannot be found in the FFS file.
How about still use DepexStatus to receive the status and then initialize Depex/DepexSize to NULL/0 when DepexStatus is error.

StandaloneMmPkg/Core/FwVol.c
```diff
@@ -222,7 +222,11 @@ MmCoreFfsFindMmDriver (
       Status = FfsFindSectionData (EFI_SECTION_PE32, FileHeader, &Pe32Data, &Pe32DataSize);
       DEBUG ((DEBUG_INFO, "Find PE data - 0x%x\n", Pe32Data));
       DepexStatus = FfsFindSectionData (EFI_SECTION_MM_DEPEX, FileHeader, &Depex, &DepexSize);
-      if (!EFI_ERROR (DepexStatus)) {
+      if (EFI_ERROR (DepexStatus)) {
```
kuqin12: Would you want to just check for `EFI_NOT_FOUND`? As that will the actual "no depex" scenario, right? I understand that the current interface only calls out it will return not found and success. But specific check will prevent the interface expansion to fail here unexpectedly.

lzeng14:
@mergifyio queue

PR #12298, opened 2026-03-16
﻿MdeModulePkg/DxeCore: Fix CoreMemoryMapSanityCheck assert with HeapG…

AmyChan-Intel:
…uard

When HeapGuard is active, CoreConvertPagesWithGuard() calls AdjustMemoryA() which expands the allocation range by up to two guard pages (head and tail). This can push the final CoreConvertPages() range past a special memory type bin boundary, causing CoreMemoryMapSanityCheck() to assert.

For example, an EfiBootServicesData allocation near the EfiACPIReclaimMemory bin boundary at 0x77FFDFFF gets guard-page expanded to 0x77FFFFFF, crossing the bin end. The splitting loop in CoreGetMemoryMap() only processes EfiConventionalMemory entries, so the allocated entry is left intact and the sanity check fires.

Fix: skip the containment check for entries whose type is neither EfiConventionalMemory nor the bin type. These are legitimate allocated entries (e.g. pool pages, guard pages) that the splitting loop intentionally does not touch.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

## Integration Instructions

### DIFF
--- a/OvmfPkg/Library/PlatformInitLib/MemDetect.c
+++ b/OvmfPkg/Library/PlatformInitLib/MemDetect.c
@@ -157,8 +157,13 @@ PlatformGetFirstNonAddressCB (
 }
 
 /**
-  Store the low (below 4G) memory size in
-  PlatformInfoHob->LowMemory
+  Store the low memory size in PlatformInfoHob->LowMemory.
+
+  The first memory block with base address zero is considered "low memory".
+  Traditionally this has been all memory below 4G.  When running under SVSM
+  there are multiple memory blocks below 4G though, because SVSM caves out a
+  chunk of memory for itself.  Only the first of these blocks is considered
+  low memory.
 **/
 STATIC
 VOID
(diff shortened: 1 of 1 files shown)


os-d:
Thanks @AmyChan-Intel for submitting this PR!

If I'm understanding this correctly, the fix should actually be in the HeapGuard code. We shouldn't have guard pages for other memory types that are being allocated in bin memory. The adjustment for guard pages should not be allowed to cross the bin boundary.

@mdkinney @kuqin12 thoughts?

mdkinney:
> Thanks @AmyChan-Intel for submitting this PR!
> 
> If I'm understanding this correctly, the fix should actually be in the HeapGuard code. We shouldn't have guard pages for other memory types that are being allocated in bin memory. The adjustment for guard pages should not be allowed to cross the bin boundary.
> 
> @mdkinney @kuqin12 thoughts?

@os-d I agree that the Heap Guard feature should guarantee that the requested allocation with guard pages is either completely within a bin or completely outside the bin.  Should never allow guard pages to cross a bin boundary.  What memory type are the guard pages?  Need to make sure the initial allocation is for number of requested pages + pages for guard regions.

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
```diff
@@ -970,34 +970,77 @@ AdjustMemoryF (
...
 {
   //
   // FindFreePages() has already taken the Guard into account. It's safe to
   // adjust the start address and/or number of pages here, to make sure that
   // the Guards are also "allocated".
   //
+  // However, guard pages must not extend into a special memory type bin that
+  // does not match the allocation type, as this would violate memory map
+  // invariants checked by CoreMemoryMapSanityCheck().
+  //
```
mdkinney: How did a range with guard pages ever get allocated in the first place that crosses a bin boundary.  Shouldn't this condition be caught before this call?

PR #12299, opened 2026-03-16
ArmPkg: Add an ASM function to read register AIDR_EL1

kainng-ampere:
# Description

This provides access to the Auxilary ID register, AIDR_EL1.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

The register value was successfully fetched in Ampere firmware.

## Integration Instructions

N/A


leiflindholm:
CI error is
` * Subject line not in "package,package: description" format!`

Subject needs to be "ArmPkg, MdePkg: ..." since you're modifying both packages.
Which means you'll also need 
to the commit message.

PR #12300, opened 2026-03-17
Fix VariableSmmRuntimeDxe MM communicate v3 buffer sizing

kuqin12:
## How This Was Tested

Tested on real hardware, now work as expected.

# Description

Fix VariableSmmRuntimeDxe MM communicate v3 buffer sizing

This change fixes `VariableSmmRuntimeDxe` to size its runtime communication buffer correctly when `EFI_MM_COMMUNICATION3_PROTOCOL` is available.

Since 202511 branch, the variable runtime path can use MM communicate v3, which requires the larger `EFI_MM_COMMUNICATE_HEADER_V3` header (56KB vs 24KB), but the driver was still allocating the buffer using the legacy header size. When the caller invokes the get variable API with a large enough buffer, the runtime code check will fail due to the buffer size cap logic.

This update makes the allocation v3-aware so the runtime variable communication buffer matches the header format actually in use and avoids failures on larger variable transactions.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on real hardware, now work as expected.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
+++ b/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
@@ -1815,8 +1815,13 @@ SmmVariableReady (
   //
   Status = GetVariablePayloadSize (&mVariableBufferPayloadSize);
   ASSERT_EFI_ERROR (Status);
-  mVariableBufferSize = SMM_COMMUNICATE_HEADER_SIZE + SMM_VARIABLE_COMMUNICATE_HEADER_SIZE + mVariableBufferPayloadSize;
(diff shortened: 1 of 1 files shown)


antklein (approved):

apop5 (approved):

liqiqiii (approved):

kuqin12:
@lgao4 could you please review and provide feedback if any? Thanks in advance!

AshrafAliS (approved):

lgao4 (approved):

kuqin12:
Rebased the branch due to lack of CLANGPDB pipeline results. No content change.

PR #12302, opened 2026-03-17
CryptoPkg: Add digest-based RSA-PSS sign and verify APIs

baranee:
# Description
Add RsaPssSignDigest() and RsaPssVerifyDigest() to BaseCryptLib for signing/verifying precomputed digests. 
Provide OpenSSL/MbedTLS/Null implementations, expose via EDKII_CRYPTO_PROTOCOL (v24), and 
add PCD controls for independent service enabling. 
Include unit tests.

- [ ] Breaking change?
- [ ] Impacts security?
- [x] Includes tests?

## How This Was Tested

Verified by invoking the new API with OpenSSL(SMM) and MbedTLS(PEI) implementations

## Integration Instructions

N/A


CryptoPkg/Readme.md
```diff
@@ -227,6 +227,7 @@ also configured.
 | TlsGet                          |     N      |     N     |             |             |    C-Tls     |             |                 |
 | RsaPss.Sign                     |     N      |     N     |             |             |      C       |             |                 |
 | RsaPss.Verify                   |     N      |     N     |             |      C      |      C       |      C      |                 |
+| RsaPss.VerifyDigest             |     N      |     N     |             |      C      |      C       |      C      |                 |
```
jyao1, line 231: Usually, we add "Sign/Verify" as a pair.
In this case, I recommend also add RsaPss.SignDigest.

jyao1:
Also need to update PcdCryptoServiceFamilyEnable.h to add new API.

baranee:
> Also need to update PcdCryptoServiceFamilyEnable.h to add new API.

My current thought is that the existing RsaPss.Services.Sign and RsaPss.Services.Verify could still cover both the message (current APIs) and digest (new SignDigest/VerifyDigest) variants. I didn’t realize we expect a 1:1 mapping between every public API and a distinct PCD bit. That said, I’m open to enhancing PcdCryptoServiceFamilyEnable.h as needed.

jyao1:
> the existing RsaPss.Services.Sign and RsaPss.Services.Verify could still cover both the message (current APIs) and digest (new SignDigest/VerifyDigest) variants.

Question: If the existing API can support current Digest Variant, then why we need a new specific API for VerifyDigest?

baranee:
> Question: If the existing API can support current Digest Variant, then why we need a new specific API for VerifyDigest?

I’ve clarified that the Sign and Verify bits in PcdCryptoServiceFamilyEnable.h cover both message‑ and digest‑level variants (capabilities, not API names). Per feedback, I added explicit RsaPss.Services.SignDigest and RsaPss.Services.VerifyDigest bits in the latest commit.

baranee:
@jyao1 Let me know any other concern needs to be addressed for this PR?

PR #12310, opened 2026-03-18
UefiPayloadPkg: seed memory type information from vars

Sean-StarLabs:
This branch now carries two related payload fixes:

- seed the memory type information table from the variable store contents
- skip `EfiBootManagerConnectAll()` only on S4 resume

The first change keeps the payload memory-type table in sync with stored
variables. The second avoids the AMI-style S4 resume path from reconnecting
all devices before the OS resume handoff.

### DIFF
--- a/UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
+++ b/UefiPayloadPkg/Library/PlatformBootManagerLib/PlatformBootManager.c
@@ -320,8 +320,19 @@ PlatformBootManagerAfterConsole (
     BootLogoEnableLogo ();
   }
 
-  EfiBootManagerConnectAll ();
-  EfiBootManagerRefreshAllBootOption ();
+  if (GetBootModeHob () != BOOT_ON_S4_RESUME) {
+    EfiBootManagerConnectAll ();
+    EfiBootManagerRefreshAllBootOption ();
+  } else {
+    //
+    // Avoid a global ConnectAll()/RefreshAllBootOption() on S4 resume.
+    //
+    // Connecting removable media here makes the final DXE handle graph and memory
+    // allocation pattern depend on USB and SD presence. That can perturb the EFI
+    // memory map enough to break OS hibernation resume if devices are inserted or
+    // removed while the system is in S4.
+    //
+  }
 
   //
   // Active BOOT_ON_FLASH_UPDATE mode means that at least one capsule has been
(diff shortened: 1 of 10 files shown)


MdeModulePkg/Library/UefiBootManagerLib/BmMisc.c
```diff
@@ -205,6 +211,22 @@ BmSetMemoryTypeInformationVariable (
     return;
   }
 
+  //
+  // Preserve the quotas that were used for this boot so S4 resume can seed DXE
+  // from the last known-good memory type information instead of the most recent
+  // auto-tuned value.
+  //
+  Status = BmSetVariableAndReportStatusCodeOnError (
```
os-d: I don't believe this makes sense.

Here's my understanding of the flow you are proposing, let me know if I am misunderstanding:

Cold boot, no change in flow:
- Platform reads MEMORY_TYPE_INFORMATION_VARIABLE_NAME and populates the HOB with those values
- During boot, however much memory is used
- PREVIOUS variable set to exactly the memory used, standard var set to tuned value
- Optionally reset the system to apply changes to bin sizes
- Boot to OS

Resume flow:
- Platform reads either PREVIOUS or standard variable (all existing platforms use the standard, you are proposing UefiPayloadPkg use PREVIOUS) and sets HOB values
- Bins are populated with the HOB value (either PREVIOUS or STD)
- During boot, however much memory is used. If you overran the bins in the cold boot, you'll overrun the bins here, too, S4 resume fails. The other case is BOOT_WITH_DEFAULT_SETTINGS and memory > 150% needed for the bins. So if your bins were very large, the advertisement could still adjust bin sizes.

The recommended platform path is to reset when the bin sizes change and repopulate on another cold boot the correct bin sizes (and after that point it likely never changes again, without FW changes). In that case, the previous variable is not needed. In the case the bins are too small, this doesn't help, S4 resume will still fail. In the case of the bins being too large and it is a BOOT_WITH_DEFAULT_SETTINGS (which my understanding is that is typically after flash update, so you are a new FW who may want to adjust the bin size), then this change could help. Seems like a small use case for additional complexity, especially as it is against the recommended path (which would be to reset here and apply the bin changes).

The other piece here is that there has been a conscious decision to avoid depending on different boot modes, particularly S4 Resume. We don't want to add the complexity and potential for error on missing edge cases, etc. BOOT_ON_S4_RESUME is only set by proprietary code, we don't have any way from the core to tell if this is being set correctly or means anything.

I would recommend dropping this commit. I do not think it meaningfully helps S4 resume stability.

Sean-StarLabs: I dropped the `PreviousMemoryTypeInformation` / `BOOT_ON_S4_RESUME` part and restacked this PR down to the payload-only change.

What remains is just teaching `UefiPayloadPkg` to seed the standard `MemoryTypeInformation` HOB from the persisted variable store instead of always using the hardcoded defaults. The MdeModulePkg change and the removable boot-option policy change are both gone now.

PR #12311, opened 2026-03-18
UnitTestFrameworkPkg: Align C++ sign compare

mdkinney:
# Description

GCC enables `-Wsign-compare` by default but CLANG does not enable `-Wsign-compare` by default. This causes GCC unit test builds to fail with a signed/unsigned mismatch when using googletest macros `EXPECT_EQ()`, `ASSERT_EQ()`, `EXPECT_NE()`, or `ASSERT_NE()` with one of the operands being literal 0. However, CLANG does not fail for these builds.

Add `-Wsign-compare` to `CXX_FLAGS` for `CLANGDWARF` and `CLANGPDB` to align CLANG error checking with GCC.

Add `/w44388` and `/w44389` to MSFT family compiler `CXX_FLAGS` to align VS20xx error checking with GCC.

Add unit tests that use `EXPECT_EQ()`, `ASSERT_EQ()`, `EXPECT_NE()`, and `ASSERT_NE()` to check a local variables of type `UINTN`, `UINT8`, `UINT16`, `UINT32`, `UINT64`, `UINT32` bitfields, and `UINT64` bitfields against literal 0 values.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Added unit tests that pass for CLANG and MSFT but fail for GCC.  Added options to make sure CLANG, MSFT, and GCC catch the same errors.  Then update unit tests to pass CLANG, MSFT, and GCC.

## Integration Instructions

If unit test failures in CLANG or MSFT are generated by this change but same unit tests pass GCC, then replace `0` with `0u` in `EXPECT_EQ()`, `ASSERT_EQ()`, `EXPECT_NE()`, or `ASSERT_NE()` statements referenced in build logs.

### DIFF
--- a/UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
+++ b/UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
@@ -183,10 +183,10 @@
   #
   # Set C++ language standard version
   #
-  MSFT:*_*_*_CXX_FLAGS         = /std:c++20 /Zc:strictStrings- /wd4244
+  MSFT:*_*_*_CXX_FLAGS         = /std:c++20 /Zc:strictStrings- /wd4244 /w44388 /w44389
   GCC:*_*_*_CXX_FLAGS          = -std=c++20
-  GCC:*_CLANGDWARF_*_CXX_FLAGS = -std=c++17
-  CLANGPDB:*_*_*_CXX_FLAGS     = -std=c++17
+  GCC:*_CLANGDWARF_*_CXX_FLAGS = -std=c++17 -Wsign-compare
+  CLANGPDB:*_*_*_CXX_FLAGS     = -std=c++17 -Wsign-compare
   #
   # CLANGDWARF and CLANGPDB: Some gtest c++ sources generate undefined
   # behavior that is never triggered. Map trap functions to abort() to


UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
```diff
@@ -183,10 +183,10 @@
   #
   # Set C++ language standard version
   #
-  MSFT:*_*_*_CXX_FLAGS         = /std:c++20 /Zc:strictStrings- /wd4244
+  MSFT:*_*_*_CXX_FLAGS         = /std:c++20 /Zc:strictStrings- /wd4244 /w44388 /w44389
   GCC:*_*_*_CXX_FLAGS          = -std=c++20
-  GCC:*_CLANGDWARF_*_CXX_FLAGS = -std=c++17
-  CLANGPDB:*_*_*_CXX_FLAGS     = -std=c++17
+  GCC:*_CLANGDWARF_*_CXX_FLAGS = -std=c++17 -Wsign-compare
```
GeorgeLiao012, line 188: Just curious it, why we set different standard version in the different compiler? 
/std:c++20 -->  MSFT and GCC
/std:c++17 --> CLANGPDB and CLANGDWARF

mdkinney: I do not recall the exact reason.  I likely was some issue making them all 20.  However, that may have been resolved by subsequent changes because local testing did not show any specific issues changing them to 20.

We can handle aligning those in a different PR.

GeorgeLiao012: OK, let me create a PR for align standard version in the UnitTestFrameworkPkg.

mdkinney:
@makubacki Can you please review

mdkinney:
> nit: There is a typo in one of the commit descriptions "CANGDWARF"

I have fixed that typo in the commit message and rebased and pushed.

PR #12312, opened 2026-03-19
UefiCpuPkg, OvmfPkg: Rename .asm files to .nasm.inc in ResetVector

niruiyu:
# Description

The .asm files are included by other assembly files and are not standalone assembly files. The .asm extension may cause confusion and make it seem like they are standalone assembly files and can be compiled by MASM.

Rename them to .nasm.inc to avoid confusion and make it clear that they are NASM include files.

The .nasm.inc extension is to distinguish them from .inc files which are usually used for assembly include files holding macros.

- [N] Breaking change?
- [N] Impacts security?
- [N] Includes tests?

## How This Was Tested

As the change is simply a file name change. Will depend on CI to verify the build and boot.

## Integration Instructions

N/A


sgrams (approved):

niruiyu:
looks like everyone likes this change. @copilot what do you think?

ardbiesheuvel:
> looks like everyone likes this change. @copilot what do you think?

Please don't invite copilot

OvmfPkg/Bhyve/ResetVector/Ia16/Real16ToFlat32.nasm.inc
makubacki, line 1: Can you split the OvmfPkg and UefiCpuPkg changes into two separate commits? We have downstream consumers that depend on UefiCPuPkg but do not use OvmfPkg and that split would make this easier to pick up.

niruiyu:
```
[BuildOptions]
   *_*_IA32_NASMB_FLAGS = -I$(WORKSPACE)/UefiCpuPkg/ResetVector/Vtf0/
   *_*_X64_NASMB_FLAGS = -I$(WORKSPACE)/UefiCpuPkg/ResetVector/Vtf0/
```
Above code is from the reset vector INF file in OvmfPkg.
That means OvmfPkg/ResetVector(s) are not a true-modularized edk2 modules. They use the nasm files from UefiCpuPkg/ResetVector sometimes and sometimes they create their own.

If I create one commit that only change UefiCpuPkg, that will make OvmfPkg build broken.

makubacki: Since it appears the UefiCpuPkg change is only breaking to these isolated instances of a module referencing internal files from an external package, then I agree it's simplest to keep it as a single commit. I suggest mentioning this in the commit description to clarify why this deviated from the per-package commit convention.

mdkinney: Adding -I to a different package is not allowed and must be removed from OvmfPkg.  Either the required files need to be copied into OvmfPkg or OvmfPkg needs to use the ResetVector component from UefiCpuPkg

makubacki: Thanks Mike. I wasn't sure if there was some prearranged agreement for an exception here but it certainly violates normal package dependency rules.

niruiyu: @makubacki the commit message is updated to explain why 2 pkgs are changed in one commit.

makubacki: I created an issue to track the -I change in OvmfPkg - https://github.com/tianocore/edk2/issues/12333.

PR #12318, opened 2026-03-20
UnitTestFrameworkPkg : Align C++ std version

GeorgeLiao012:
# Description
The C++ standard version doesn't align in the current UnitTestFrameworkPkg, that may caused C++ compiler flow cannot be aligned. it may caused potential issue.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
None

## Integration Instructions
Ensure each compiler adopt same C++ language standard version.

### DIFF
--- a/UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
+++ b/UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
@@ -185,8 +185,8 @@
   #
   MSFT:*_*_*_CXX_FLAGS         = /std:c++20 /Zc:strictStrings- /wd4244
   GCC:*_*_*_CXX_FLAGS          = -std=c++20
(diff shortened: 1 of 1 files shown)


makubacki (approved):

makubacki:
The change itself is fine to me, but if there is any documentation or rationale for the different versions previously used, this commit message would be a great place to include that information.

mdkinney:
I agree more details need to be added on why the current c++ spec levels were set.

Some downstream unit test are seeing some build issues with this change. Recommend moving this into a draft PR for now.   May require an update to the latest release of googletest (1.17.0) as well to resolve issues.

GeorgeLiao012:
As Mike mentioned I encountered some downstream unit test build issues with c++ std 20, I will address those build issues first. Thanks.

mdkinney:
Not stale.

PR #12319, opened 2026-03-20
Nasm compiles "int 3" to cd 3 and "int3" to the familiar cc.

jaykrell:
# Description

I noticed this and was surprised porting masm to nasm and comparing.

- [ ] Breaking change? I am not sure. I am afraid it might be.
- [ ] Impacts security? Not sure.
- [ ] Includes tests? No.

## How This Was Tested

It was not.

## Integration Instructions

Not sure. I was aware you could do this but I was surprised to see in practise. Is it deliberate?


mdkinney:
No functional impact from using `int3` or `int 3`.  `int3` is a single byte form that is useful for debuggers setting software breakpoints so only a single byte needs to be modified.

khaliid2040:
They are different instructions with same functionality. `int n` is more generic instruction that recieves vector as argument when it recieves 3 as argument it behaves like `int3`. MASM doesn't support `int3`, don't know why but this is why `int 3` is used rather than `int3`. Also other mnemonic like `int0` are same for MASM ( doesn't support them).

jaykrell:
Right. Masm "optimizes" "int 3" to one byte.
Nasm assembles "int 3" to two bytes, and "int3" to one byte.
This may have been originally Masm source, but it is now Nasm.
Perhaps this was aspect was lost in the conversion?
Programmers are used to nop'ing out int3 with one 0x90.
Or just me?

There is also various debugging infrastructure that I believe incorrectly assumes "breakpoints" are single byte instructions, and this breaks that assumption.

mikebeaton:
It's very unclear to me that this change is needed. As people are saying above, `int3` and `int 3` are two different instructions (with the same effect, but taking a different number of bytes). What is the motivation for making the change - is it actually causing you any issues?

jaykrell:
Need is a strong word. But how many people see int 3 in code and jump to the conclusion that it is a single byte 0xcc?
I think almost everyone.
That is also what it does if you used Masm. Nasm is surprising. Like maybe the code at some point ported from masm and accidentally changed? I almost did that myself, in a recent port from Masm to Nasm but caught it, which is why I went looking for other similar.

I am pretty sure we have systems that depend on int 3 being one byte, but not necessarily these particular int 3, probably not these. It is an idiom to get a trap and skip over a breakpoint, and _assume_ that a breakpoint trap is one byte, and just skip one byte. It is more correct to do the runtime disasm and check the instruction but not everyone does that. Not all systems allow reading code pages, but then the intersection of factors is approaching zero.

PR #11840, opened 2025-12-04
introduce debug log bit for paging

kraxel:
edk2 log files became quite chatty in recent years, with the
extended use of paging for confidential computing and for
page attributes (using NX / WP to improve firmware security).

This series adds a new debug log bit (DEBUG_PAGING) for all
messages related to page table updates, to allow turning these
messages on and off independent from other messages.

A number of log messages have been changed from DEBUG_VERBOSE
to DEBUG_PAGING.  These are the paging-related messages I've
noticed when booting OVMF and ArmVirt.  Most likely there are
more in the code base, but I want gather some feedback on the
idea before going hunt them all down.

- **MdePkg/DebugLib: add DEBUG_PAGING**
- **OvmfPkg/BaseMemEncryptSevLib: DEBUG_VERBOSE -> DEBUG_PAGING**
- **UefiCpuPkg/CpuDxe: DEBUG_VERBOSE -> DEBUG_PAGING**
- **UefiCpuPkg/ArmMmuLib: DEBUG_VERBOSE -> DEBUG_PAGING**

PR #12321, opened 2026-03-23
UefiCpuPkg/MpLib: Demote print level for "5-Level Paging ="

abdattar:
The "5-Level Paging" debug message in FillExchangeInfoData() is printed every time APs are woken up. At DEBUG_INFO level, this creates unnecessary log noise during normal boot. Demote the print level from DEBUG_INFO to DEBUG_VERBOSE so the message is only visible when verbose logging is enabled.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on AMD platform.

## Integration Instructions

N/A

### DIFF
--- a/MdePkg/Include/Library/FdtLib.h
+++ b/MdePkg/Include/Library/FdtLib.h
@@ -202,6 +202,11 @@ typedef struct {
 #define FdtSetPropEmpty(Fdt, NodeOffset, Name) \
   FdtSetProp ((Fdt), (NodeOffset), (Name), NULL, 0)
 
+#define FdtForEachPropertyOffset(Property, Fdt, Node)   \
+  for (Property = FdtFirstPropertyOffset(Fdt, Node);    \
+       Property >= 0;                                   \
+       Property = FdtNextPropertyOffset(Fdt, Property))
+
 /**
   Convert UINT16 data of the FDT blob to little-endian
 
@@ -869,9 +874,9 @@ FdtPathOffset (
 CONST CHAR8 *
 EFIAPI
 FdtGetName (
-  IN VOID   *Fdt,
-  IN INT32  NodeOffset,
-  IN INT32  *Length
+  IN CONST VOID  *Fdt,
+  IN INT32       NodeOffset,
+  IN INT32       *Length
   );
 
 /**
@@ -887,10 +892,10 @@ FdtGetName (
 INT32
 EFIAPI
 FdtGetPath (
-  IN VOID    *Fdt,
-  IN INT32   NodeOffset,
-  IN VOID    *Buffer,
-  IN UINT32  BufferSize
+  IN CONST VOID  *Fdt,
+  IN INT32       NodeOffset,
+  IN VOID        *Buffer,
+  IN UINT32      BufferSize
   );
 
 /**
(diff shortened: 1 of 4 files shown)


niruiyu:
This debug message is printed only when APs are waken up through INIT-SIPI-SIPI.
Can we print the CR4.LA57 bit in other place, just once in a boot if your system does require wake up AP everytime through INIT-SIPI-SIPI?

kraxel:
Merge https://github.com/tianocore/edk2/pull/11840, then use DEBUG_PAGING for this message?

niruiyu:
> Merge #11840, then use DEBUG_PAGING for this message?

I like this idea. Can you ping @mdkinney and @lgao4 to review #11840 as they are the maintainers of MdePkg?

khaliid2040:
I am not sure if this will prevent it completely, when DEBUG_PAGING is set wouldn't the message get printed multple times.

My suggestion is to move the log to `UefiCpuPkg/Library/MpInitLib/X64/CreatePageTable.c` then change log level to DEBUG_PAGING.

kraxel:
Yes, the switch to DEBUG_PAGING will not prevent the message being printed multiple times.

I think there is some value in having APs print this too, to show that propagating this from BSP to APs worked properly.  Having the BSP + one AP print that message would be enough, not sure there is an easy way to implement that though.

niruiyu:
> I am not sure if this will prevent it completely, when DEBUG_PAGING is set wouldn't the message get printed multple times.
> 
> My suggestion is to move the log to `UefiCpuPkg/Library/MpInitLib/X64/CreatePageTable.c` then change log level to DEBUG_PAGING.

Moving to that file does not help in all cases.
The logic in that file is to create page table for APs after ExitBootServices.
How about print the LA57 bit status in MpInitLibInitialize()?

khaliid2040:
> > I am not sure if this will prevent it completely, when DEBUG_PAGING is set wouldn't the message get printed multple times.
> > My suggestion is to move the log to `UefiCpuPkg/Library/MpInitLib/X64/CreatePageTable.c` then change log level to DEBUG_PAGING.
> 
> Moving to that file does not help in all cases. The logic in that file is to create page table for APs after ExitBootServices. How about print the LA57 bit status in MpInitLibInitialize()?

It is good to me, but as @kraxel noted if paging is needed to be printed each for BSP and AP, then we are missing a log for AP paging mode.

We just need another place to print paging mode for AP, then it will good to go.

niruiyu:
> > > I am not sure if this will prevent it completely, when DEBUG_PAGING is set wouldn't the message get printed multple times.
> > > My suggestion is to move the log to `UefiCpuPkg/Library/MpInitLib/X64/CreatePageTable.c` then change log level to DEBUG_PAGING.
> > 
> > 
> > Moving to that file does not help in all cases. The logic in that file is to create page table for APs after ExitBootServices. How about print the LA57 bit status in MpInitLibInitialize()?
> 
> It is good to me, but as @kraxel noted if paging is needed to be printed each for BSP and AP, then we are missing a log for AP paging mode.
> 
> We just need another place to print paging mode for AP, then it will good to go.

The code being changed is to print the LA57 bit in BSP just when it propagates the bit to APs.
If we do need to print paging mode for AP every time it wakes up, the PR should be rejected because the code is doing correctly.
If we want to just print paging mode once and assume the LA57 bit is not changed later when waking up APs, we can move the print to MpInitLibInitialize().

abdattar:
Changed the macro to "DEBUG_PAGING", Please let me know your thoughts.
This will reduce the print messages for normal boot with DEBUG_INFO.

niruiyu (approved):

PR #12325, opened 2026-03-23
MdeModulePkg: Add NULL check for mSmmBootPerformanceTable

shkrc:
In GetFpdtRecordPtr(), there is a chance when mSmmBootPerformanceTable could be set to NULL by ReallocatePool() if there is not enough memory for reallocation.

At some point later FpdtSmiHandler() tries to dereference this table in one of the cases which results in an exception.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

### DIFF
--- a/MdeModulePkg/Library/SmmCorePerformanceLib/MmCorePerformanceLib.c
+++ b/MdeModulePkg/Library/SmmCorePerformanceLib/MmCorePerformanceLib.c
@@ -11,7 +11,8 @@
 
   Copyright (c) 2006 - 2023, Intel Corporation. All rights reserved.<BR>
   (C) Copyright 2016 Hewlett Packard Enterprise Development LP<BR>
-  Copyright (c) Microsoft Corporation.
+  Copyright (c) Microsoft Corporation.<BR>
+  Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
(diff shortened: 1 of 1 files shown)


shkrc:
@niruiyu can you help with a review

MdeModulePkg/Library/SmmCorePerformanceLib/MmCorePerformanceLib.c
```diff
@@ -243,6 +244,7 @@ GetFpdtRecordPtr (
                                    );
 
       if (mSmmBootPerformanceTable == NULL) {
+        ASSERT_EFI_ERROR (EFI_OUT_OF_RESOURCES);
```
khaliid2040: Please remove this assertion, instead callers should decide whether to assert or not.

shkrc: Thanks for the review, I have removed the assertion

mikebeaton:
Might it not make more sense for the error returned to be `EFI_OUT_OF_RESOURCES`?

More importantly, if an update is addressing one place where `mSmmBootPerformanceTable` can be used without checking whether it was successfully allocated, then it should probably handle all the different places where this applies, in this file?

shkrc:
> Might it not make more sense for the error returned to be `EFI_OUT_OF_RESOURCES`?

My reason behind using EFI_NOT_FOUND is at this point in time the caller only understands that the table doesn't exist and doesn't know if its due to out of memory or something else. 

The actual check is done here, which already returns EFI_OUT_OF_RESOURCES. 
```
      mSmmBootPerformanceTable = ReallocatePool (
                                   mPerformanceLength,
                                   mPerformanceLength + RecordSize + FIRMWARE_RECORD_BUFFER,
                                   mSmmBootPerformanceTable
                                   );

      if (mSmmBootPerformanceTable == NULL) {
        return EFI_OUT_OF_RESOURCES;
      }
```

> More importantly, if an update is addressing one place where `mSmmBootPerformanceTable` can be used without checking whether it was successfully allocated, then it should probably handle all the different places where this applies, in this file?

I agree, let me check the code for more changes

PR #12332, opened 2026-03-24
NetworkPkg: Remove redundant Status initializations and fix unhandled errors

AbuthahirM11:
## Summary

This patch series cleans up redundant `EFI_STATUS` variable initializations
across 14 NetworkPkg drivers, and fixes two cases where function return
values were not checked.

### Bug Fixes
- **DnsDxe**: Add missing NULL check and error return for `Dhcp4->Build()`
  in `GetDns4ServerFromDhcp4`. Without this check, a NULL `Token.Packet`
  could be dereferenced.
- **HttpDxe**: Add missing error check with proper resource cleanup after
  `TlsCommonTransmit()` in `HttpsReceive`. A failure was previously silently
  ignored, leaking `PacketOut` and `BufferOut`.

### Code Cleanup
The following drivers had local `Status` variables initialized to a fixed
value (`EFI_SUCCESS` or `EFI_DEVICE_ERROR`) at declaration, then
unconditionally overwritten before any use. These redundant initializations
have been removed:

- Dhcp6Dxe
- DnsDxe
- HttpBootDxe
- HttpDxe
- Ip4Dxe
- Ip6Dxe
- MnpDxe
- Mtftp4Dxe
- Mtftp6Dxe
- SnpDxe
- TcpDxe
- TlsAuthConfigDxe
- UefiPxeBcDxe
- WifiConnectionManagerDxe

## Cc

### DIFF
--- a/NetworkPkg/DnsDxe/DnsDhcp.c
+++ b/NetworkPkg/DnsDxe/DnsDhcp.c
@@ -474,6 +474,14 @@ GetDns4ServerFromDhcp4 (
   ParaList[1]->Data[0] = DHCP4_TAG_DNS_SERVER;
 
   Status = Dhcp4->Build (Dhcp4, &SeedPacket, 0, NULL, 2, ParaList, &Token.Packet);
+  if (EFI_ERROR (Status)) {
+    goto ON_EXIT;
+  }
+
+  if (Token.Packet == NULL) {
+    Status = EFI_NOT_FOUND;
+    goto ON_EXIT;
+  }
 
   Token.Packet->Dhcp4.Header.Xid = Random;
 
--- a/NetworkPkg/HttpDxe/HttpsSupport.c
+++ b/NetworkPkg/HttpDxe/HttpsSupport.c
@@ -818,9 +818,6 @@ TlsCommonTransmit (
       );
 
     HttpInstance->Tcp4TlsTxToken.Packet.TxData = (EFI_TCP4_TRANSMIT_DATA *)Data;
-
-    Status = EFI_DEVICE_ERROR;
-
     //
     // Transmit the packet.
     //
@@ -852,8 +849,6 @@ TlsCommonTransmit (
       );
 
     HttpInstance->Tcp6TlsTxToken.Packet.TxData = (EFI_TCP6_TRANSMIT_DATA *)Data;
-
-    Status = EFI_DEVICE_ERROR;
 
     //
     // Transmit the packet.
@@ -1452,7 +1447,6 @@ TlsCloseSession (
   NET_BUF  *PacketOut;
   UINT8    *DataOut;
 
-  Status    = EFI_SUCCESS;
   BufferOut = NULL;
   PacketOut = NULL;
   DataOut   = NULL;
@@ -1574,7 +1568,6 @@ TlsProcessMessage (
   EFI_TLS_FRAGMENT_DATA  *OriginalFragmentTable;
   UINTN                  Index;
 
-  Status                = EFI_SUCCESS;
   Buffer                = NULL;
   BufferSize            = 0;
   BytesCopied           = 0;
@@ -1704,7 +1697,6 @@ HttpsReceive (
   UINT8              *GetSessionDataBuffer;
   UINTN              GetSessionDataBufferSize;
 
-  Status                   = EFI_SUCCESS;
   Pdu                      = NULL;
   BufferIn                 = NULL;
   BufferInSize             = 0;
--- a/NetworkPkg/HttpDxe/HttpsSupport.c
+++ b/NetworkPkg/HttpDxe/HttpsSupport.c
@@ -1909,6 +1909,11 @@ HttpsReceive (
       CopyMem (DataOut, BufferOut, BufferOutSize);
 
       Status = TlsCommonTransmit (HttpInstance, PacketOut);
+      if (EFI_ERROR (Status)) {
+        NetbufFree (PacketOut);
+        FreePool (BufferOut);
+        return Status;
+      }
 
       NetbufFree (PacketOut);
     }
--- a/NetworkPkg/Ip4Dxe/Ip4Config2Impl.c
+++ b/NetworkPkg/Ip4Dxe/Ip4Config2Impl.c
@@ -947,16 +947,7 @@ Ip4StartAutoConfig (
                   EFI_OPEN_PROTOCOL_BY_DRIVER
                   );
   if (EFI_ERROR (Status)) {
-    NetLibDestroyServiceChild (
-      IpSb->Controller,
-      IpSb->Image,
-      &gEfiDhcp4ServiceBindingProtocolGuid,
-      Instance->Dhcp4Handle
-      );
-
-    Instance->Dhcp4Handle = NULL;
-
-    return Status;
+    goto Error;
   }
 
   //
@@ -965,6 +956,10 @@ Ip4StartAutoConfig (
   //
   Dhcp4  = Instance->Dhcp4;
   Status = Dhcp4->GetModeData (Dhcp4, &Dhcp4Mode);
+  if (EFI_ERROR (Status)) {
+    goto Error;
+  }
+
   if (Dhcp4Mode.State == Dhcp4Bound) {
     Ip4Config2OnDhcp4Complete (NULL, Instance);
 
@@ -987,25 +982,7 @@ Ip4StartAutoConfig (
 
   Status = Dhcp4->Configure (Dhcp4, &Dhcp4Mode.ConfigData);
   if (EFI_ERROR (Status)) {
-    gBS->CloseProtocol (
-           Instance->Dhcp4Handle,
-           &gEfiDhcp4ProtocolGuid,
-           IpSb->Image,
-           IpSb->Controller
-           );
-
-    NetLibDestroyServiceChild (
-      IpSb->Controller,
-      IpSb->Image,
-      &gEfiDhcp4ServiceBindingProtocolGuid,
-      Instance->Dhcp4Handle
-      );
-
-    Instance->Dhcp4 = NULL;
-
-    Instance->Dhcp4Handle = NULL;
-
-    return Status;
+    goto Error;
   }
 
   //
@@ -1036,6 +1013,29 @@ Ip4StartAutoConfig (
   DispatchDpc ();
 
   return EFI_SUCCESS;
+
+Error:
+  if (Instance->Dhcp4 != NULL) {
+    gBS->CloseProtocol (
+           Instance->Dhcp4Handle,
+           &gEfiDhcp4ProtocolGuid,
+           IpSb->Image,
+           IpSb->Controller
+           );
+
+    Instance->Dhcp4 = NULL;
+  }
+
+  NetLibDestroyServiceChild (
+    IpSb->Controller,
+    IpSb->Image,
+    &gEfiDhcp4ServiceBindingProtocolGuid,
+    Instance->Dhcp4Handle
+    );
+
+  Instance->Dhcp4Handle = NULL;
+
+  return Status;
 }
 
 /**
--- a/NetworkPkg/Ip4Dxe/Ip4Config2Nv.c
+++ b/NetworkPkg/Ip4Dxe/Ip4Config2Nv.c
@@ -535,7 +535,6 @@ Ip4Config2ConvertIfrNvDataToConfigNvData (
   UINTN          DataSize;
   EFI_INPUT_KEY  Key;
 
-  Status    = EFI_SUCCESS;
   Ip4Cfg2   = &Instance->Ip4Config2;
   Ip4NvData = &Instance->Ip4NvData;
 
@@ -878,7 +877,6 @@ Ip4FormExtractConfig (
     return EFI_INVALID_PARAMETER;
   }
 
-  Status             = EFI_SUCCESS;
   IfrFormNvData      = NULL;
   ConfigRequest      = NULL;
   FormResult         = NULL;
--- a/NetworkPkg/MnpDxe/MnpConfig.c
+++ b/NetworkPkg/MnpDxe/MnpConfig.c
@@ -1199,8 +1199,6 @@ MnpStart (
   MNP_DEVICE_DATA  *MnpDeviceData;
 
   NET_CHECK_SIGNATURE (MnpServiceData, MNP_SERVICE_DATA_SIGNATURE);
-
-  Status        = EFI_SUCCESS;
   MnpDeviceData = MnpServiceData->MnpDeviceData;
 
   if (!IsConfigUpdate) {
@@ -1330,19 +1328,19 @@ MnpStop (
     //
     //  The system poll in on, cancel the poll timer.
     //
-    Status                          = gBS->SetTimer (MnpDeviceData->PollTimer, TimerCancel, 0);
+    gBS->SetTimer (MnpDeviceData->PollTimer, TimerCancel, 0);
     MnpDeviceData->EnableSystemPoll = FALSE;
   }
 
   //
   // Cancel the timeout timer.
   //
-  Status = gBS->SetTimer (MnpDeviceData->TimeoutCheckTimer, TimerCancel, 0);
+  gBS->SetTimer (MnpDeviceData->TimeoutCheckTimer, TimerCancel, 0);
 
   //
   // Cancel the media detect timer.
   //
-  Status = gBS->SetTimer (MnpDeviceData->MediaDetectTimer, TimerCancel, 0);
+  gBS->SetTimer (MnpDeviceData->MediaDetectTimer, TimerCancel, 0);
 
   //
   // Stop the simple network.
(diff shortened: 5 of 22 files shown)


NetworkPkg/Ip4Dxe/Ip4Config2Impl.c
```diff
@@ -963,8 +963,8 @@ Ip4StartAutoConfig (
   // Check the current DHCP status, if the DHCP process has
   // already finished, return now.
   //
-  Dhcp4  = Instance->Dhcp4;
-  Status = Dhcp4->GetModeData (Dhcp4, &Dhcp4Mode);
+  Dhcp4 = Instance->Dhcp4;
+  Dhcp4->GetModeData (Dhcp4, &Dhcp4Mode);
```
mikebeaton: I believe we should still get the Status here, and return the error if there is one. (Which would require adding appropriate teardown.)

NetworkPkg/Ip4Dxe/Ip4Config2Nv.c
mikebeaton, line 591: This is related, though possibly a separate PR, shouldn't the first of these calls to Ip4Cfg2->SetData return the Status, especially as the second one does?

NetworkPkg/MnpDxe/MnpConfig.c
```diff
@@ -1330,19 +1328,19 @@ MnpStop (
...
   // Cancel the timeout timer.
   //
-  Status = gBS->SetTimer (MnpDeviceData->TimeoutCheckTimer, TimerCancel, 0);
+  gBS->SetTimer (MnpDeviceData->TimeoutCheckTimer, TimerCancel, 0);
 
   //
   // Cancel the media detect timer.
   //
-  Status = gBS->SetTimer (MnpDeviceData->MediaDetectTimer, TimerCancel, 0);
+  gBS->SetTimer (MnpDeviceData->MediaDetectTimer, TimerCancel, 0);
```
mikebeaton, line 1343: Is it okay to discard these Statuses, should we be stopping and returning them if they are errors?

NetworkPkg/DnsDxe/DnsImpl.c
mikebeaton: Again possibly a separate PR, though related: this `while` is actually a `for` loop, and should probably be made into one - and the preceding `Status = EFI_NOT_FOUND` assignment is only needed in the case where the `for` loop is never entered. (So could be possibly made a conditional assignment, assuming that part of the point of this exercise is to help spot future coding errors where the Status is not initialised on a certain paths when it should be.)

mikebeaton (changes requested):
Apart from the places I commented, I did scan through all of this, and I believe the other changes are correct.

NetworkPkg/MnpDxe/MnpConfig.c
AbuthahirM11: Across other EDK II drivers, the pattern is the same: SetTimer with TimerCancel is called without checking the return value. This is because EFI_SetTimer can only fail with EFI_INVALID_PARAMETER if the event handle or timer type is invalid, which cannot happen here since these timers were created earlier and are guaranteed to be valid. As a result, canceling a timer is effectively guaranteed to succeed and is treated as a fire‑and‑forget operation in EDK II. In contrast, when starting a periodic or relative timer, drivers do check the status because those operations involve allocating timer state, which can legitimately fail in some implementations.

NetworkPkg/Ip4Dxe/Ip4Config2Impl.c
AbuthahirM11: You're right, thank you. This change was unintentional — the original code did capture the Status from Dhcp4->GetModeData(), and this PR accidentally dropped it. I'll restore the status check and add the proper teardown on failure

NetworkPkg/Ip4Dxe/Ip4Config2Nv.c
AbuthahirM11: The first Ip4Cfg2->SetData call should also check and return Status, just like the second one does. Since this is a pre-existing bug independent of this PR's scope, I've raised a separate PR to fix it.

NetworkPkg/DnsDxe/DnsImpl.c
AbuthahirM11: Agreed — the loop is really a for loop in disguise (counter from 0 to DnsHeader->AnswersNum). Converting it would also let us remove the Status = EFI_NOT_FOUND pre-assignment, replacing it with a conditional that only applies when the loop body is never entered, which makes uninitialized paths easier to spot. I've opened a separate PR for this refactoring.

NetworkPkg/MnpDxe/MnpConfig.c
mikebeaton: I can't currently Resolve this - but resolved, thanks.

NetworkPkg/Ip4Dxe/Ip4Config2Nv.c
mikebeaton: I assume "will have soon raised", since I can't see! Can you ping a link here when ready? Thank you.

NetworkPkg/DnsDxe/DnsImpl.c
mikebeaton: Great - can you ping up a link here when the PR is up on EDK 2? (I don't think it's open already? I can't see it.) Thank you.

AbuthahirM11: Refactored in a separate PR: https://github.com/tianocore/edk2/pull/12337

NetworkPkg/Ip4Dxe/Ip4Config2Nv.c
AbuthahirM11: Fixed in a separate PR: https://github.com/tianocore/edk2/pull/12337

mikebeaton: Resolved as regards this PR - thanks.

NetworkPkg/DnsDxe/DnsImpl.c
mikebeaton: Resolved as regards this PR - thanks.

NetworkPkg/DnsDxe/DnsDhcp.c
```diff
@@ -474,6 +474,13 @@ GetDns4ServerFromDhcp4 (
   ParaList[1]->Data[0] = DHCP4_TAG_DNS_SERVER;
 
   Status = Dhcp4->Build (Dhcp4, &SeedPacket, 0, NULL, 2, ParaList, &Token.Packet);
+  if (EFI_ERROR (Status)) {
```
leiflindholm, line 477: Since this is a functional change, could you please break it out into a separate commit?

NetworkPkg/HttpDxe/HttpsSupport.c
```diff
@@ -1917,6 +1909,11 @@ HttpsReceive (
       CopyMem (DataOut, BufferOut, BufferOutSize);
 
       Status = TlsCommonTransmit (HttpInstance, PacketOut);
+      if (EFI_ERROR (Status)) {
```
leiflindholm, line 1912: Since this is a functional change, can you please break it out into a separate commit?

NetworkPkg/MnpDxe/MnpConfig.c
leiflindholm: There are situations where certain tools will complain about ignored return values, which may be why the Status assignments were added in the first place.

Personally, I'm a fan of casting to (VOID) to make it clear we are explicitly ignoring the return value (as opposed to either blindly assigning or accidentally ignoring).

leiflindholm (dismissed):
Big fan of this change, with a small number of comments.

I'd like to say there's no need to split the changes up per module like you've done here. One commit for all "drop needless Status assignments" in NetworkPkg would have been sufficient, with a couple more commits for the other changes.

NetworkPkg/MnpDxe/MnpConfig.c
mikebeaton: @leiflindholm - It is a point that there are already multiple `gBS->SetTimer` which don't assign return value to anything in NetworkPkg (as well as many which do, and the value is then checked). So in this case it seemingly does match the rest of the package as @AbuthahirM11 argues to just not assign to anything - and any casting to (VOID) might be more like a separate PR, perhaps?

leiflindholm: @mikebeaton It's not a hill I will die on, but I feel if we're changing the code to be more clear on its intent, existing adjacent code being vague is not a strong argument to conform to it.

AbuthahirM11: @leiflindholm I understand the concern about silently ignoring return values. However, the gBS->SetTimer calls being changed here are specifically TimerCancel operations — not timer starts. Across NetworkPkg (and other EDK II packages), the existing convention is that TimerCancel calls don't check or assign the return value, but status is used for during set timer for a event.

leiflindholm: @AbuthahirM11 Unless you can find somewhere documenting this as a convention, I think we can safely assume we're dealing with laziness or slobbiness. Neither which feels like something to aspire to.

Alternatively, if we desperately want to have a way to cancel a timer without checking (or explicitly) the documented return value, the clear way of doing that would be by implementing a wrapper function for that, descriptively named.

NetworkPkg/DnsDxe/DnsDhcp.c
AbuthahirM11: Thank you for the review! I've reorganized the commits. The functional changes are now in separate commits

NetworkPkg/HttpDxe/HttpsSupport.c
AbuthahirM11: Thank you for the review! I've reorganized the commits. The functional changes are now in separate commits

NetworkPkg/MnpDxe/MnpConfig.c
mikebeaton: @leiflindholm - If this is not a hill you will die on, is this something that can be left for another commit?

NetworkPkg/Ip4Dxe/Ip4Config2Impl.c
mikebeaton: @AbuthahirM11 - Were you going to resolve this? I think the teardown used below here (rather than above) is the correct one, or perhaps there could be a goto Exit or goto Teardown to a shared teardown handler for all three (above, below, and this).

NetworkPkg/MnpDxe/MnpConfig.c
leiflindholm: Apologies, last week was all travel and meetings.

I'm happy to move this change out into its own PR to discuss in isolation. Whether involving wrapper functions or not.

NetworkPkg/Ip4Dxe/Ip4Config2Impl.c
```diff
@@ -965,6 +965,11 @@ Ip4StartAutoConfig (
   //
   Dhcp4  = Instance->Dhcp4;
   Status = Dhcp4->GetModeData (Dhcp4, &Dhcp4Mode);
+  if (EFI_ERROR (Status)) {
+    Ip4Config2DestroyDhcp4 (Instance);
+    return Status;
+  }
+
```
mikebeaton, line 962: Hi @AbuthahirM11 -

Looking at the cleanup above this:

```
    NetLibDestroyServiceChild (
      IpSb->Controller,
      IpSb->Image,
      &gEfiDhcp4ServiceBindingProtocolGuid,
      Instance->Dhcp4Handle
      );

    Instance->Dhcp4Handle = NULL;
```

and the cleanup below this:

```
    gBS->CloseProtocol (
           Instance->Dhcp4Handle,
           &gEfiDhcp4ProtocolGuid,
           IpSb->Image,
           IpSb->Controller
           );

    NetLibDestroyServiceChild (
      IpSb->Controller,
      IpSb->Image,
      &gEfiDhcp4ServiceBindingProtocolGuid,
      Instance->Dhcp4Handle
      );

    Instance->Dhcp4 = NULL;

    Instance->Dhcp4Handle = NULL;
```

I would have expected the required cleanup here to be like the cleanup below, and not just what you have here.

If you agree, the best approach might be to revert to my previous suggestion of adding an `Error:` goto label to handle all three cleanups with shared code, depending on what is null.

AbuthahirM11: @mikebeaton — I've replaced the three inline cleanup paths into a shared Error: label that conditionally cleans up based on non‑NULL state (close Dhcp4, destroy Dhcp4Handle). This gives GetModeData the same full teardown as Configure and removes duplicated logic.

NetworkPkg/Ip4Dxe/Ip4Config2Impl.c
```diff
@@ -1036,6 +1013,31 @@ Ip4StartAutoConfig (
...
+           Instance->Dhcp4Handle,
+           &gEfiDhcp4ProtocolGuid,
+           IpSb->Image,
+           IpSb->Controller
+           );
+
+    Instance->Dhcp4 = NULL;
+  }
+
+  if (Instance->Dhcp4Handle != NULL) {
```
mikebeaton: This can't be non-NULL, so no need for this wrapping `if`?

NetworkPkg/Ip4Dxe/Ip4Config2Impl.c
mikebeaton: Outdated/resolved.

NetworkPkg/Ip4Dxe/Ip4Config2Impl.c
mikebeaton: Apart from minor nit comment below, LGTM - resolved.

NetworkPkg/Ip4Dxe/Ip4Config2Impl.c
AbuthahirM11: @mikebeaton — You're right. Every goto Error path is only reachable after NetLibCreateServiceChild has already set Instance->Dhcp4Handle (a failure there returns early before any goto). So Dhcp4Handle is unconditionally non-NULL at the label. I've removed the redundant if wrapper.

mikebeaton: Great thanks - I don't seem to be able to click Resolved on my own comments (red error message), but ... resolved.

mikebeaton (approved):

AbuthahirM11:
@leiflindholm @ajfish @ardbiesheuvel @makubacki @mdkinney @SaloniKasbekar @Zclarkwilliams — Could you please take a look and approve this PR when you get a chance?

makubacki:
The review state of this PR is cluttered at the moment. @mikebeaton, are you able to resolve the threads you're involved in now? @leiflindholm, can you please revisit and remove your "changes requested" status?

mikebeaton:
> @mikebeaton, are you able to resolve the threads you're involved in now?

@makubacki - Unfortunately, still no:
image
=>
image

mikebeaton:
https://github.com/orgs/community/discussions/8327

makubacki:
> https://github.com/orgs/community/discussions/8327

Thanks. The simplest solution would be to sign up as a maintainer for a package if you're interested :)

@leiflindholm, can you please confirm the conversation threads you started are resolved?

mikebeaton:
> The simplest solution would be to sign up as a maintainer for a package if you're interested :)

I'll stick as a reviewer for now, though thanks for the vote of confidence. ;)

Just to clarify for anyone who doesn't have time to read the link, @AbuthahirM11 as PR author or anyone with write access to the repo can mark these as resolved. (But a reviewer without write access can't. :-( )

AbuthahirM11:
Hi @makubacki and @mikebeaton, I've resolved all the conversation threads. @leiflindholm, could you please confirm if your requested changes are addressed so we can mark those threads as resolved as well?

mdkinney:
@SaloniKasbekar Please review

AbuthahirM11:
@leiflindholm — all your review comments have been addressed and threads resolved. Could you please re-review and update your "changes requested" status so this PR can move forward? 
@SaloniKasbekar — would you also be able to take a look when you get a chance?

AbuthahirM11:
@leiflindholm @ajfish @ardbiesheuvel @mdkinney @SaloniKasbekar @Zclarkwilliams — Could you please take a look and approve this PR when you get a chance?

AbuthahirM11:
Hi @leiflindholm — all the review threads you opened have been addressed and resolved. The SetTimer return value concern is being handled separately in another PR.

Could you please revisit your review and dismiss the "changes requested" status so this PR can move forward? @makubacki, if possible, could you also help dismiss the stale review?

makubacki:
@AbuthahirM11, I've dismissed Leif's review. I know his availability is limited at the moment.

makubacki (approved):

PR #12334, opened 2026-03-25
UnitTestFrameworkPkg: Allow platform DSC to override address sanitizer

abdattar:
UNIT_TESTING_ADDRESS_SANITIZER_ENABLE is unconditionally set to TRUE in UnitTestFrameworkPkgHost.dsc.inc, preventing platform DSC files from disabling it. Wrap the assignment with !ifndef so platforms can override the value by defining UNIT_TESTING_ADDRESS_SANITIZER_ENABLE before the !include directive.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

AMD platform

## Integration Instructions

N/A

### DIFF
--- a/ArmVirtPkg/ArmVirtQemu.dsc
+++ b/ArmVirtPkg/ArmVirtQemu.dsc
@@ -33,6 +33,7 @@
   DEFINE TPM2_ENABLE             = FALSE
   DEFINE TPM2_CONFIG_ENABLE      = FALSE
   DEFINE CAVIUM_ERRATUM_27456    = FALSE
+  DEFINE DEBUG_TO_MEM            = FALSE
 
   #
   # Network definition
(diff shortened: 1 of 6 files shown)


mdkinney:
The intent here is that ASAN should always be enabled, and only disabled to debug issues.  The use of -D UNIT_TESTING_ADDRESS_SANITIZER_ENABLE=FALSE on the command line is the expected way to disable ASAN temporarily.

I do not disagree with this change in general, because if a specific Unit Test DSC needs to disable ASAN for a while to sort out issues with ASAN, then that is a reasonable use case.  

The same could be done for UNIT_TESTING_CODE_COVERAGE_ENABLE to have flexibility to disable code coverage collection if it is causing issues.  Do you want to update this PR to add !if around both?

abdattar:
> The intent here is that ASAN should always be enabled, and only disabled to debug issues. The use of -D UNIT_TESTING_ADDRESS_SANITIZER_ENABLE=FALSE on the command line is the expected way to disable ASAN temporarily.
> 
> I do not disagree with this change in general, because if a specific Unit Test DSC needs to disable ASAN for a while to sort out issues with ASAN, then that is a reasonable use case.
> 
> The same could be done for UNIT_TESTING_CODE_COVERAGE_ENABLE to have flexibility to disable code coverage collection if it is causing issues. Do you want to update this PR to add !if around both?

Thank you, @mdkinney, for your prompt review. I have addressed the review comment, could you please take another look?

PR #12337, opened 2026-03-25
NetworkPkg: Fix unhandled Status and refactor answer loop

AbuthahirM11:
## Summary

This patch series addresses two issues raised as review comments on PR #12332.

### Bug Fix

- **Ip4Dxe**: In `Ip4Config2ConvertIfrNvDataToConfigNvData()`, when the
  policy is changed away from a static configuration (`PolicyChanged` is
  `TRUE`), the first call to `Ip4Cfg2->SetData()` was not checking its
  return value and unconditionally returned `EFI_SUCCESS`. Fixed to check
  and return `Status` on error, consistent with the second `SetData` call
  in the same function.

### Code Cleanup

- **DnsDxe**: In `ParseDnsResponse()`, the answer section loop used a
  `while` loop with a manual counter increment at the bottom. Converted to
  an equivalent `for` loop to make initialization, condition, and increment
  explicit. Also replaced the unconditional `Status = EFI_NOT_FOUND`
  pre-assignment with a conditional that only sets `Status` when the loop
  is never entered (`DnsHeader->AnswersNum == 0`), making uninitialized
  `Status` paths easier to spot.

## Ref

Raised as review comments on: #12332

## Cc

### DIFF
--- a/NetworkPkg/DnsDxe/DnsImpl.c
+++ b/NetworkPkg/DnsDxe/DnsImpl.c
@@ -1165,10 +1165,9 @@ ParseDnsResponse (
   Dns4TokenEntry = NULL;
   Dns6TokenEntry = NULL;
 
-  IpCount          = 0;
-  RRCount          = 0;
-  AnswerSectionNum = 0;
-  CNameTtl         = 0;
+  IpCount  = 0;
+  RRCount  = 0;
+  CNameTtl = 0;
 
   HostAddr4 = NULL;
   HostAddr6 = NULL;
@@ -1378,8 +1377,6 @@ ParseDnsResponse (
     }
   }
 
-  Status = EFI_NOT_FOUND;
-
   //
   // Get Answer name
   //
@@ -1388,7 +1385,11 @@ ParseDnsResponse (
   //
   // Processing AnswerSection.
   //
-  while (AnswerSectionNum < DnsHeader->AnswersNum) {
+  if (DnsHeader->AnswersNum == 0) {
+    Status = EFI_NOT_FOUND;
+  }
+
+  for (AnswerSectionNum = 0; AnswerSectionNum < DnsHeader->AnswersNum; AnswerSectionNum++) {
     //
     // Check whether the remaining packet length is available or not.
     //
@@ -1635,7 +1636,6 @@ ParseDnsResponse (
     // Find next one
     //
     AnswerName = (CHAR8 *)AnswerSection + sizeof (*AnswerSection) + AnswerSection->DataLength;
-    AnswerSectionNum++;
   }
 
   if (Instance->Service->IpVersion == IP_VERSION_4) {
--- a/NetworkPkg/Ip4Dxe/Ip4Config2Nv.c
+++ b/NetworkPkg/Ip4Dxe/Ip4Config2Nv.c
@@ -575,7 +575,7 @@ Ip4Config2ConvertIfrNvDataToConfigNvData (
                                    sizeof (EFI_IP4_CONFIG2_POLICY),
                                    &Ip4NvData->Policy
                                    );
-    return EFI_SUCCESS;
+    return Status;
   }
 
   if (IfrFormNvData->DhcpEnable == TRUE) {


NetworkPkg/DnsDxe/DnsImpl.c
mikebeaton: ^^^

NetworkPkg/Ip4Dxe/Ip4Config2Nv.c
```diff
@@ -611,7 +611,9 @@ Ip4Config2ConvertIfrNvDataToConfigNvData (
                                    sizeof (EFI_IP4_CONFIG2_POLICY),
                                    &Ip4NvData->Policy
                                    );
-    return EFI_SUCCESS;
+    if (EFI_ERROR (Status)) {
+      return Status;
+    }
```
mikebeaton: No, I think this is wrong!

As per my comment on the other PR, this should just be `return Status;` - no surrounding `if`.

mikebeaton (changes requested):

NetworkPkg/DnsDxe/DnsImpl.c
mikebeaton: (This initialization should be removed, now that it's in the for loop.)

AbuthahirM11: Done. Removed the redundant AnswerSectionNum = 0; initialization since it's now handled in the for loop initializer.

NetworkPkg/Ip4Dxe/Ip4Config2Nv.c
AbuthahirM11: Good point — I changed it to always return Status because the function should return a value every time, so the if check wasn’t needed.

NetworkPkg/Ip4Dxe/Ip4Config2Nv.c
```diff
@@ -611,7 +611,7 @@ Ip4Config2ConvertIfrNvDataToConfigNvData (
                                    sizeof (EFI_IP4_CONFIG2_POLICY),
                                    &Ip4NvData->Policy
                                    );
-    return EFI_SUCCESS;
+      return Status;
```
mikebeaton: Hi - unwanted whitespace change here now. Otherwise LGTM.

AbuthahirM11: Fixed — removed the extra whitespace. Thank you!

mikebeaton (approved):

NetworkPkg/DnsDxe/DnsImpl.c
```diff
@@ -1167,7 +1167,6 @@ ParseDnsResponse (
 
   IpCount          = 0;
   RRCount          = 0;
-  AnswerSectionNum = 0;
   CNameTtl         = 0;
```
mikebeaton: Uncrustify is asking you to change the alignment of these lines, now that the long var name is removed:

```
ERROR -    Dns4TokenEntry = NULL;
ERROR -    Dns6TokenEntry = NULL;
ERROR -  
ERROR - -  IpCount          = 0;
ERROR - -  RRCount          = 0;
ERROR - -  CNameTtl         = 0;
ERROR - +  IpCount  = 0;
ERROR - +  RRCount  = 0;
ERROR - +  CNameTtl = 0;
ERROR -  
ERROR -    HostAddr4 = NULL;
ERROR -    HostAddr6 = NULL;
ERROR - 
```

AbuthahirM11: Thanks @mikebeaton — good catch. I've fixed the alignment as Uncrustify requires

NetworkPkg/DnsDxe/DnsImpl.c
mikebeaton: Resolved.

NetworkPkg/Ip4Dxe/Ip4Config2Nv.c
mikebeaton: Resolved.

NetworkPkg/Ip4Dxe/Ip4Config2Nv.c
mikebeaton: Resolved.

mikebeaton:
I think this one is okay to merge.

NetworkPkg/DnsDxe/DnsImpl.c
mdkinney: Should local variable AnswerSectionNum also be removed.  I do not think it is referenced anymore

AbuthahirM11: Thanks for the review, @mdkinney! The standalone AnswerSectionNum = 0; line was removed from the explicit variable initialization block at the top of the function because the initialization is now handled inline in the for statement initializer. The variable declaration itself must remain since it is still the loop control variable.

AbuthahirM11:
@leiflindholm @ajfish @ardbiesheuvel @makubacki @mdkinney @SaloniKasbekar @Zclarkwilliams — Could you please take a look and approve this PR when you get a chance?

makubacki:
Can you clarify what is meant by "making uninitialized Status paths on certain code paths easier to spot" in this commit message?

> Also replace the unconditional Status = EFI_NOT_FOUND assignment
before the loop with a conditional that only sets Status when the loop
is never entered (DnsHeader->AnswersNum == 0), making uninitialized
Status paths on certain code paths easier to spot.

This is confusing because `Status` is clearly initialized to `EFI_SUCCESS` at the start of the function (`ParseDnsResponse()`). Can you please add more detail on how this achieves the stated goal? In this request I'm not questioning if it does, but that the detail be added to the message so it can more easily be related to the code.

AbuthahirM11:
Good point - @makubacki The wording was misleading. ParseDnsResponse() initializes Status = EFI_SUCCESS at function entry and the loop body always overwrites it, so the old pre-assignment was dead code in the common case. Updated the commit message to clarify this.

makubacki:
I closed @mikebeaton's conversation threads since they are acknowledged as resolved. I've also added the push label given @mikebeaton's approval.

makubacki:
@Mergifyio queue

mikebeaton:
This is also stuck not merging.

mikebeaton:
Quick ping. This is still stuck not merging. 🤷‍♂️

leiflindholm:
@mergify requeue

leiflindholm (approved):

leiflindholm:
> Quick ping. This is still stuck not merging. 🤷‍♂️

Oh, right. Maybe I should actually read the message that mergify unhelpfully collapses in its status.

PR #12338, opened 2026-03-25
NetworkPkg/HttpBootDxe: Print TLS errors on screen during HTTP boot

SaloniKasbekar:
# Description

Currently an end user does not have any information about TLS failures if attempting a HTTP Boot. This change allows for more user-friendly HTTP Boot process.

Install EDKII_HTTP_CALLBACK_PROTOCOL in HttpBootDxe to receive TLS events from HttpDxe during the HTTP Boot process. When a TLS error occurs (failed TLS session connection or TLS configuration), print an error message on screen using AsciiPrint().

The callback is installed in HttpBootInstallCallback() and uninstalled in HttpBootUninstallCallback(), ensuring TLS errors are only printed during the HTTP Boot process and not for other TLS accesses.

- [ ] Breaking change?
None

- [ ] Impacts security?
None

- [ ] Includes tests?
None

## How This Was Tested

Verified prints during TLS failure with a release BIOS

## Integration Instructions

N/A

### DIFF
--- a/NetworkPkg/HttpBootDxe/HttpBootDxe.h
+++ b/NetworkPkg/HttpBootDxe/HttpBootDxe.h
@@ -59,6 +59,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 //
 #include <Protocol/LoadFile.h>
 #include <Protocol/HttpBootCallback.h>
+#include <Protocol/HttpCallback.h>
 
 //
 // Consumed Guids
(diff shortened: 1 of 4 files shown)


SaloniKasbekar:
@mdkinney Would you be able to help merge the changes? Not sure why the pending check is still queued up.

makubacki:
> @mdkinney Would you be able to help merge the changes? Not sure why the pending check is still queued up.

@SaloniKasbekar, it looks like this PR has not been rebased since the Windows CLANGPDB workflow was added. I rebased it to trigger the workflow. Given the approvals on the PR, I added the push label as well.

PR #12340, opened 2026-03-25
ArmPkg,ArmVirtPkg,UefiPayloadPkg,UefiCpuPkg: Relocate library to Uefi…

vishalo:
# Description

Move AARCH64 exception handling from ArmPkg to UefiCpuPkg to provide
consistent exception handling across all CPU architectures (IA32, X64,
LoongArch64, RISCV64, AARCH64).

Key changes:
- Merge ArmExceptionLib into CpuExceptionHandlerLib
- Merge DefaultExceptionHandlerLib into CpuExceptionHandlerLib
- Rename DefaultExceptionHandler() API to DumpCpuContext() for
  consistency with existing CpuExceptionLib.h interface
- Update ArmCrashDumpDxe and CpuDxe to use unified library
- Remove ArmPkg/Include/Library/DefaultExceptionHandlerLib.h

  
- [X] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Built and tested on SBSA Qemu.

## Integration Instructions

- For out of tree modules using DefaultExceptionHandlerLib, add UefiCpuPkg.dec as dependency and update dsc to point to CpuExceptionHandlerLib. 

- Remove all references of  ArmExceptionLib from dsc file.

- Replace DefaultExceptionHandler() calls with DumpCpuContext().


UefiCpuPkg/UefiCpuPkg.dec
```diff
@@ -95,6 +95,7 @@
   ##  @libraryclass  Provides a Mmu interface.
   #
   ArmMmuLib|Include/Library/ArmMmuLib.h
+  DefaultExceptionHandlerLib|Include/Library/DefaultExceptionHandlerLib.h
```
os-d: I agree with moving code out of ArmPkg, but should DefaultExceptionHandlerLib still be retained as a library class? Can we converge on CpuExceptionHandlerLib, which is what all other architectures use?

The answer may be no in this case, but as part of dismantling ArmPkg/ArmPlatformPkg, I think we need to decide on what the correct API surface should be. ArmPkg has been allowed to be the wild west and have any APIs it wants (looking at you ArmLib :) and not conform to existing library classes in many cases, such as this one.

From a brief look and my work in both before, I think the library classes could be merged? That is more of a breaking change, but imo, it doesn't make sense to just move the components out of ArmPkg, we also need to make sure we have the right abstractions.

vishalo: Thank you for reviewing @os-d . 

I was looking at merging DefaultExceptionHandlerLib as well. 

DefaultExceptionHandlerLib is currently used by ArmCrashdumpDxe. That's why I kept it.  Let me check further. I prefer not to block this change; would you agree to addressing this in a follow-on PR?

If you have any thoughts on how to handle ArmCrashdumpDxe, please let me know.

os-d: I'm fine with it being in a follow-on PR, but I will note that it would be two breaking changes for folks (the first one introduced in this PR is trivially changing a path, but still a bit of churn for consumers if it isn't completed by the next stable tag). I'm not overly concerned, more just noting because it has been a subject of discussion of late.

vishalo: Thank you @os-d . I've updated it - removed DefaultExceptionHandlerLib.

os-d: Thanks @vishalo ! I'll take a look now

os-d:
UefiCpuPkg.dsc has the various CpuExceptionHandlerLibs under a Components.IA32,Components.X64 section, the DXE one at a minimum needs to be moved out to the common section so it is built for AARCH64 during CI.

To that point, this PR only changes the DXE version, I think it should also update the SecPei and Pei versions to add AARCH64 support? Standalone MM is needed, too. It might be easier to introduce a BASE version of the lib? I'd have to review what the SecPei one does again, perhaps that could be converted to BASE?

UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHandlerLib.inf
```diff
@@ -83,8 +91,14 @@
 [LibraryClasses.LoongArch64]
   CpuLib
 
+[LibraryClasses.AARCH64]
+  ArmLib
```
os-d, line 97: I'm concerned about adding ArmLib here, so far UefiCpuPkg has not had a dependency on ArmPkg (and the only reason this isn't failing CI about it is this module isn't built for AARCH64 in CI, as noted in the other comment) and maybe because ArmPkg.dec isn't added to the Packages section here.

I'm trying this out locally to see what is actually being used from ArmLib, hopefully it could be isolated easily.

os-d: The only thing that is used from ArmLib is `ArmWriteVbar()`, which is a 6 line ASM function that is only used in this lib and one other place. The x86 equivalent is in BaseLib: https://github.com/tianocore/edk2/blob/c4d29cb62187060493a1f595083ddfb6dd346f39/MdePkg/Include/Library/BaseLib.h#L7228. I would recommend moving `ArmWriteVbar()` to BaseLib and then the ArmLib dependency can be dropped.

os-d: TBH, I think that most if not all of ArmLib actually should be in BaseLib, like other architectures. Let me send an email on the mailing list.

leiflindholm: https://github.com/tianocore/edk2/issues/10289

So, yes, agree, but I think that's a slightly longer play.

For now, ArmLib is declared in MdePkg, so the dependency should not be problematic.

os-d: The header file is declared in MdePkg, but the only implementation is still in ArmPkg. Unless I missed an ArmLibNull in MdePkg.

So you still have the problem that UefiCpuPkg now needs to depend on ArmPkg to satisfy the library dependency.

I'm sending out a proposal on the mailing list now. I do think ArmLib could be removed relatively easily. Maybe I'm being naiive :)

leiflindholm:
```
/work/git/edk2/UefiCpuPkg$ git grep ArmLib
Include/Library/ArmMmuLib.h:#include <Library/ArmLib.h>
Library/ArmMmuLib/AArch64/ArmMmuLibCore.c:#include <Library/ArmLib.h>
Library/ArmMmuLib/AArch64/ArmMmuPeiLibConstructor.c:#include <Library/ArmLib.h>
Library/ArmMmuLib/ArmMmuBaseLib.inf:  ArmLib
Library/ArmMmuLib/ArmMmuPeiLib.inf:  ArmLib
Library/BaseArchSupportLib/AArch64/AArch64LibSupport.c:#include <Library/ArmLib.h>
```

leiflindholm: But don't get me wrong, if you're willing to take on moving ArmLib altogether, I'll grab my pom-poms.

os-d: Fair enough :). I just saw UefiCpuPkg.ci.yaml didn't have ArmPkg as an acceptable dependency, but that makes sense it is a header dependency check, not the implemenation.

I drop my request to not use ArmLib in this PR. Let's address the overall ArmLib issue separately.

> But don't get me wrong, if you're willing to take on moving ArmLib altogether, I'll grab my pom-poms.

Let's see, I sent a proposal on the mailing list. Depending on what folks think on that or a version of that, I'm happy to take it up. I also would love to see ArmPkg homogenized with the rest of edk2 :)

vishalo:
@os-d :
> UefiCpuPkg.dsc has the various CpuExceptionHandlerLibs under a Components.IA32,Components.X64 section, the DXE one at a minimum needs to be moved out to the common section so it is built for AARCH64 during CI.
> 
Sure, I can move it out to common.

> To that point, this PR only changes the DXE version, I think it should also update the SecPei and Pei versions to add AARCH64 support? Standalone MM is needed, too. It might be easier to introduce a BASE version of the lib? I'd have to review what the SecPei one does again, perhaps that could be converted to BASE?

I will update the SecPei/Pei version.

What is the Standalone MM requirement ?

os-d:
> What is the Standalone MM requirement ?

Thinking through it more, we don't need this in StMM. If StMM is running S-EL1, it can handle exceptions, but edk2 isn't set up to support that (except where it is running as a pseudo-TA in OP-TEE and I don't know that scenario at all, presumably OP-TEE would handle exceptions). In the FF-A case with StMM running in S-EL0, Hafnium at S-EL2 will handle exceptions. For the SPM_MM case, StMM was still using S-EL0, so exceptions were getting handled by TF-A.

vishalo:
I moved it to draft to address CI failures - moving it back to review.
@os-d  - this has SecPei update as well.

ArmVirtPkg/ArmVirt.dsc.inc
```diff
@@ -97,8 +97,7 @@ DEFINE FD_SIZE_IN_MB    = 3
 
   # ARM Architectural Libraries
   CacheMaintenanceLib|ArmPkg/Library/ArmCacheMaintenanceLib/ArmCacheMaintenanceLib.inf
-  DefaultExceptionHandlerLib|ArmPkg/Library/DefaultExceptionHandlerLib/DefaultExceptionHandlerLib.inf
-  CpuExceptionHandlerLib|ArmPkg/Library/ArmExceptionLib/ArmExceptionLib.inf
+  CpuExceptionHandlerLib|UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHandlerLib.inf
```
os-d, line 101: Should this use the SecPei lib as well?

os-d (approved):

ArmVirtPkg/ArmVirt.dsc.inc
vishalo: Thank you @os-d!

Good catch. While ArmVirtQemu's SEC phase sets up exception vectors directly via [ExceptionSupport.S](https://github.com/tianocore/edk2/blob/master/ArmPlatformPkg/Sec/AArch64/Exception.S) (so SecPeiCpuExceptionHandlerLib is not actively invoked there), it is still correct to map the right library class for SEC and PEI module types. I'll add the following overrides to ArmVirtPkg/ArmVirt.dsc.inc:

[LibraryClasses.common.SEC]
  CpuExceptionHandlerLib|UefiCpuPkg/Library/CpuExceptionHandlerLib/SecPeiCpuExceptionHandlerLib.inf

[LibraryClasses.common.PEI_CORE, LibraryClasses.common.PEIM]
  CpuExceptionHandlerLib|UefiCpuPkg/Library/CpuExceptionHandlerLib/SecPeiCpuExceptionHandlerLib.inf

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/CpuExceptionHandlerLib.c
```diff
@@ -1,25 +1,32 @@
...
   SPDX-License-Identifier: BSD-2-Clause-Patent
-
 **/
 
 #include <Uefi.h>
-#include <Library/UefiLib.h>
+#include <Library/CpuExceptionHandlerLib.h>
+#include <Library/ArmLib.h>
+#include <Library/PcdLib.h>
+#include <Library/CacheMaintenanceLib.h>
```
leiflindholm: This was always going to be messy, but this looks messier than it needs to be.
Can you try a bit harder to insert includes alphabetically sorted?

leiflindholm (changes requested):
As your commit message already highlights, this should be split over three commits:
  - ArmExceptionLib (merge with common CpuExceptionHandlerLib)
  - Merge DefaultExceptionHandlerLib in to CpuExceptionHandlerLib
  - Rename DefaultExceptionHandler() API to DumpCpuContext(). This
    API is already present in CpuExceptionLib.h
Although probably with the order of 2 and 3 reversed.

vishalo:
Moving to draft to split as separate commit.

UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHandlerLib.inf
```diff
@@ -66,6 +72,7 @@
   MdePkg/MdePkg.dec
   MdeModulePkg/MdeModulePkg.dec
   UefiCpuPkg/UefiCpuPkg.dec
+  ArmPkg/ArmPkg.dec
```
leiflindholm: Please add packages somewhat alphabetically.

UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHandlerLib.inf
```diff
@@ -83,6 +90,8 @@
 [LibraryClasses.LoongArch64]
   CpuLib
 
+[LibraryClasses.AARCH64]
+  DefaultExceptionHandlerLib
```
leiflindholm: Add a blank line after to match surroundings.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/CpuExceptionSupport.h
```diff
@@ -0,0 +1,22 @@
+/** @file
+
+  Copyright (c) 2008 - 2009, Apple Inc. All rights reserved.<BR>
+  Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries. All rights reserved.
```
leiflindholm: `<BR>`

UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHelpers.c
```diff
@@ -1,15 +1,16 @@
...
 
   Copyright (c) 2008 - 2009, Apple Inc. All rights reserved.<BR>
+  Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries. All rights reserved. <BR>
 
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
 
 #include <Uefi.h>
-#include <Library/PeCoffGetEntryPointLib.h>
```
leiflindholm, line 10: Why move this?

UefiCpuPkg/Library/CpuExceptionHandlerLib/CpuExceptionHelpers.h
```diff
@@ -0,0 +1,22 @@
...
+
+  Copyright (c) 2008 - 2009, Apple Inc. All rights reserved.<BR>
+  Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries. All rights reserved.
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#pragma once
+
```
leiflindholm, line 12: This header is missing a Base.h include.

UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHelpers.c
leiflindholm, line 10: Why move this line?

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/SecPeiCpuExceptionSupport.c
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  return LastSlash;
+}
+
+/**
+  Validate that the address returned by PeCoffSearchImageBase() is
+  actually a valid PE/COFF or TE image header.
+
+  PeCoffSearchImageBase() only checks 2 bytes at each page boundary,
```
leiflindholm: Please stop letting the bot write your comments. A function description is to describe what a function does, not to describe why you expect it to be called and why.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/SecPeiCpuExceptionSupport.c
```diff
@@ -0,0 +1,181 @@
+/** @file
+  Provides image name lookup for SEC/PEI phase exception handling.
+
+  Uses PeCoffSearchImageBase() to locate the faulting image, then
```
leiflindholm: Please rewrite using human brain.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/SecPeiCpuExceptionSupport.c
```diff
@@ -0,0 +1,181 @@
...
+
+  if (DosHdr->e_magic == EFI_IMAGE_DOS_SIGNATURE) {
+    //
+    // DOS header found. Validate e_lfanew:
+    //   - Must be beyond the DOS header itself
+    //   - Must be within a reasonable range (< 4KB) to avoid
+    //     reading from a wildly wrong address
+    //
+    PeOffset = DosHdr->e_lfanew;
+    if ((PeOffset < sizeof (EFI_IMAGE_DOS_HEADER)) || (PeOffset > 0x1000)) {
```
leiflindholm: I was going to complain about the live-coded value, but what is this even intending to do? Put a finger in the air and guess whether we trust the value? Where does this idea come from?

In fact, apart from this test, this only re-checks what was already checked by PeCoffSearchImageBase (). This would suggest that if there is some merit to this additional check, that is where it should be added.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/SecPeiCpuExceptionSupport.c
```diff
@@ -0,0 +1,181 @@
...
+    return TRUE;
+  }
+
+  return FALSE;
+}
+
+/**
+  Return the image name for the given fault address.
+
+  Locates the PE/COFF image base by scanning backwards from FaultAddress,
```
leiflindholm: No it doesn't it calls PeCoffSearchImageBase.
Apply human, please.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/SecPeiCpuExceptionSupport.c
```diff
@@ -0,0 +1,181 @@
...
+  *ImageBase           = PeCoffSearchImageBase (FaultAddress);
+
+  if (*ImageBase == 0) {
+    return NULL;
+  }
+
+  //
+  // Validate the PE/COFF header before calling PeCoffLoaderGetPdbPointer().
+  // PeCoffSearchImageBase() only checks 2 bytes and can return false
+  // positives. Processing an invalid header in PeCoffLoaderGetPdbPointer()
```
leiflindholm: This is not a useful comment. No one reading this code should need to care about bytewise checks some called function makes or not. And no one reading this code needs to know what errors might happen if we pass bad data to functions we call.

What would be useful would be to point out that contrary to what the name of the function might lead you to expect, PeCoffSearchImageBase doesn't always return a valid address.

Which would be better addressed in PeCoffSearchImageBase.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/CpuExceptionHandlerLib.c
```diff
@@ -3,13 +3,13 @@
 
   Copyright (c) 2008 - 2010, Apple Inc. All rights reserved.<BR>
   Copyright (c) 2011 - 2014, ARM Ltd. All rights reserved.<BR>
-
```
leiflindholm, line 6: Please add Qualcomm statement adjacent to the other copyright statements without deleting the blank line between statements and SPDX line.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/CpuExceptionHandlerLib.c
```diff
@@ -3,13 +3,13 @@
...
   Copyright (c) 2011 - 2014, ARM Ltd. All rights reserved.<BR>
-
+  Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries. All rights reserved.<BR>
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
 
 #include <Uefi.h>
-
+#include <Library/CpuExceptionHandlerLib.h>
```
leiflindholm: Down one.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/CpuExceptionHandlerLib.c
```diff
@@ -10,13 +10,15 @@
...
 #include <Library/CpuExceptionHandlerLib.h>
+#include <Library/ArmLib.h>
 #include <Library/BaseLib.h>
 #include <Library/DebugLib.h>
 #include <Library/PeCoffGetEntryPointLib.h>
 #include <Library/PrintLib.h>
 #include <Library/SerialPortLib.h>
 #include <Protocol/DebugSupport.h>
 
+#include <ExceptionSupport.h>
```
leiflindholm: Down one.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/CpuExceptionHandlerLib.c
```diff
@@ -25,6 +27,13 @@
 //
 #define MAX_PRINT_CHARS  100
 
+UINTN                   gMaxExceptionNumber                           = MAX_AARCH64_EXCEPTION;
```
leiflindholm: I don't see this used outside of this module - should it be 'm'?
Same for next two lines.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/ExceptionSupport.h
```diff
@@ -0,0 +1,20 @@
...
+
+ Exception handling helper assembly functions
+
+ Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries. All rights reserved.<BR>
+ SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#pragma once
+
```
leiflindholm, line 15: Missing Base.h

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/CpuExceptionHandlerLib.c
```diff
@@ -175,6 +209,17 @@ BaseName (
   return Str;
 }
 
+/**
+  Configures the exception vector environment for the current execution
+  level. Registers a 16-byte aligned EL0 stack derived from mNewStackBase,
```
leiflindholm: Why is this function describing how it's implemented instead of what it does?
Apply human brain.

leiflindholm (changes requested):

UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHelpers.c
vishalo: Move it back.

vishalo: Moved it back.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/SecPeiCpuExceptionSupport.c
vishalo: Change is not required. I was trying to debug/fix crash due to XPU - I picked the wrong address to test this change!

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/SecPeiCpuExceptionSupport.c
vishalo: Change is not required.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/CpuExceptionHandlerLib.c
vishalo: Done

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/CpuExceptionHandlerLib.c
vishalo: Updated to m*.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/ExceptionSupport.h
vishalo: Updated !

vishalo:
Thanks for reviewing @leiflindholm  ! Moved to draft to split to separate commits, and fix CI failures. This is ready for review.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/DxeCpuExceptionSupport.c
leiflindholm, line 32: Digging around the tree, I notice we have CpuExceptionHandlerLib/RiscV/BacktraceHelperSec.c and CpuExceptionHandlerLib/RiscV/BacktraceHelper.c.

Could we perhaps unify these at the top level of the module instead? That Dxe GetImageName is a copy of this one to begin with.

vishalo: @leiflindholm : updated in latest commit. Validated on RiscVVirtQemu.

vishalo:
Rebased and fix CI ECC errors.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/CpuExceptionHandlerLib.c
```diff
@@ -259,7 +259,7 @@ CommonCExceptionHandler (
   )
 {
   if ((UINTN)ExceptionType <= mMaxExceptionNumber) {
-    if (mExceptionHandlers[ExceptionType]) {
+    if (mExceptionHandlers[ExceptionType] != NULL) {
```
leiflindholm, line 262: Better to fix in ArmExceptionLib before all of the merging and moving shenanigans.

ArmPkg/Drivers/ArmCrashDumpDxe/ArmCrashDumpDxe.inf
```diff
@@ -20,10 +20,11 @@
 [Packages]
   ArmPkg/ArmPkg.dec
   MdePkg/MdePkg.dec
+  UefiCpuPkg/UefiCpuPkg.dec
 
 [LibraryClasses]
   DebugLib
-  DefaultExceptionHandlerLib
+  CpuExceptionHandlerLib
```
leiflindholm, line 27: C comes before D, so please move up as part of renaming.

UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeCpuExceptionHandlerLib.inf
```diff
@@ -78,6 +78,7 @@
 
 [LibraryClasses.common]
   BaseLib
+  UefiLib
```
leiflindholm: And U comes last of these, please move to the end.

ArmPkg/Drivers/ArmCrashDumpDxe/ArmCrashDumpDxe.c
```diff
@@ -1,14 +1,15 @@
...
+  Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries. All rights reserved.<BR>
 
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
 
 #include <PiDxe.h>
 #include <Library/DebugLib.h>
-#include <Library/DefaultExceptionHandlerLib.h>
+#include <Library/CpuExceptionHandlerLib.h>
```
leiflindholm, line 11: Move up when renaming.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/CpuExceptionHandlerLib.c
```diff
@@ -19,14 +20,22 @@
 #include <Protocol/DebugSupport.h>
 
 #include <CpuExceptionSupport.h>
+#include <ExceptionSupport.h>
```
leiflindholm: Looks a little bit confusing with an ExceptionSupport next to a CpuExceptionSupport.

I think it would be more idiomatic to call the local include something like CpuExceptionHelpers.h (especially as it has no relation at all to ExceptionSupport.S) and include it with "" instead of <> for clarity.

leiflindholm (changes requested):
A few style/naming/sorting comments - then it should be ready to go in.

leiflindholm:
https://github.com/tianocore/edk2-platforms/pull/988

vlsunil (approved):
RISC-V changes LGTM.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/ExceptionSupport.h
leiflindholm: Why did you delete this again?

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/ExceptionSupport.h
```diff
@@ -0,0 +1,20 @@
+/** @file
```
leiflindholm, line 1: OK, let's try again:
Please name this file whatever you want except for ExceptionSupport.h, which carries the implicit suggestion that it is declaring C interfaces for things defined in ExceptionSupport.S (which is not what it's doing).

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/CpuExceptionHandlerLib.c
```diff
@@ -19,7 +19,7 @@
 #include <Library/SerialPortLib.h>
 #include <Protocol/DebugSupport.h>
 
-#include "CpuExceptionSupport.h"
+#include <CpuExceptionHelpers.h>
```
leiflindholm, line 22: This wasn't what I meant (see previous comment on ExceptionSupport.h), so please drop this commit.

(However, if this had been what I meant, I would still have wanted this folded into the existing commits, not tagged onto the end.)

leiflindholm (changes requested):

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/ExceptionSupport.h
vishalo: ExceptionSupport.h is declaring C interfaces for functions in ExceptionSupport.S.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/CpuExceptionHandlerLib.c
vishalo: I can fold this into existing commit - that makes sense ... is this what you meant?

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/ExceptionSupport.h
vishalo: oops, will fix ...

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/ExceptionSupport.h
leiflindholm: Err, I stand corrected. I told you this was confusing :)

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/CpuExceptionHandlerLib.c
leiflindholm: Given my ExceptionSupport.h confusion, yes please.

UefiCpuPkg/Library/CpuExceptionHandlerLib/AArch64/CpuExceptionHandlerLib.c
```diff
@@ -9,17 +9,15 @@
...
 #include <Library/PeCoffGetEntryPointLib.h>
 #include <Library/PrintLib.h>
 #include <Library/SerialPortLib.h>
-#include <Library/UefiBootServicesTableLib.h>
-
-#include <Guid/DebugImageInfoTable.h>
 #include <Protocol/DebugSupport.h>
-#include <Protocol/LoadedImage.h>
+
+#include <CpuExceptionHelpers.h>
```
leiflindholm, line 22: I won't ask for another rework since this ends up the way I want it by the end of the PR anyway, but idiomatically:
Since at this point, CpuExceptionHelpers.h is a local include, it would have been more clear in "", not <>.

leiflindholm (approved):

leiflindholm:
@mergify requeue

PR #12343, opened 2026-03-26
Global: Fix the issue that Depex sections exist in the INF files of UEFI_DRIVER or UEFI_APPLICATION

### LINKED ISSUE #12276: [Bug]:  [Depex]  sections exist in the INF files of UEFI_DRIVER or UEFI_APPLICATION

### Bug Type

- [x] Firmware
- [x] Tool
- [ ] Unit Test

MdeModulePkg, BaseTools, RedfishPkg, ShellPkg

DEBUG, NOOPT, RELEASE

### Current Behavior

In the current code, some inf files of UEFI_DRIVER or UEFI_APPLICATION include [Depex] section, which breaks the rule of [INF specification](https://tianocore-docs.github.io/edk2-InfSpecification/draft/3_edk_ii_inf_file_format/314_[depex]_sections.html#314-depex-sections ) .
In
 `MdeModulePkg/Bus/Usb/UsbNetwork/NetworkCommon/NetworkCommon.inf`
 `MdeModulePkg/Bus/Usb/UsbNetwork/UsbCdcEcm/UsbCdcEcm.inf`
 `MdeModulePkg/Bus/Usb/UsbNetwork/UsbCdcNcm/UsbCdcNcm.inf`
 `MdeModulePkg/Bus/Usb/UsbNetwork/UsbRndis/UsbRndis.inf`
 `MdeModulePkg/Universal/EbcDxe/EbcDebuggerConfig.inf`
 `MdeModulePkg/Universal/FaultTolerantWriteDxe/FaultTolerantWriteStandaloneMm.inf`
 `ShellPkg/DynamicCommand/VariablePolicyDynamicCommand/VariablePolicyApp.inf `, there are
```
[Depex]
  TRUE
```
In `RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.inf`， there is
```
[Depex]
  gEdkIIRedfishCredentialProtocolGuid
```
In `NetworkPkg/TcpDxe/TcpDxe.inf`, there is
```
[Depex]
  gEfiHash2ServiceBindingProtocolGuid
```
During the building of the modules, BaseTools does not generate any error messages.

### Expected Behavior

1. If there is [Depex] section in the UEFI_DRIVER or UEFI_APPLICATION INF file, BaseTools should throw error message to break the building.
2. [Depex] section in the UEFI_DRIVER or UEFI_APPLICATION INF file should be removed from the current code.

### Steps To Reproduce

Add a check in `BaseTools/Source/Python/Workspace/InfBuildData.py`, it will throw error message during the building while there is invalid [Depex] section in the INF files of UEFI_DRIVER or UEFI_APPLICATION.
```
diff --git a/BaseTools/Source/Python/Workspace/InfBuildData.py b/BaseTools/Source/Python/Workspace/InfBuildData.py
index cab2231d..b08aa4e7 100644
--- a/BaseTools/Source/Python/Workspace/InfBuildData.py
+++ b/BaseTools/Source/Python/Workspace/InfBuildData.py
@@ -772,6 +772,15 @@ class InfBuildData(ModuleBuildClassObject):
             return RetVal
 
         RecordList = self._RawData[MODEL_EFI_DEPEX, self._Arch]
+
+        # If the module is not a library (no LIBRARY_CLASS in the [Defines] section) and the MODULE_TYPE is
+        # SEC, SMM_CORE, DXE_CORE, PEI_CORE, UEFI_DRIVER, UEFI_APPLICATION or HOST_APPLICATION a Depex section
+        # is not permitted.
+        if len(self.LibraryClass) == 0 and len(RecordList) != 0:
+            if self.ModuleType in [SUP_MODULE_SEC, SUP_MODULE_SMM_CORE, SUP_MODULE_DXE_CORE, SUP_MODULE_PEI_CORE, SUP_MODULE_UEFI_DRIVER, \
+                SUP_MODULE_UEFI_APPLICATION, SUP_MODULE_HOST_APPLICATION]:
+                EdkLogger.error('build', RESOURCE_NOT_AVAILABLE, "[Depex] section is not permitted for [%s] module" % self.ModuleType, File=self.MetaFile)
+
         # PEIM and DXE drivers must have a valid [Depex] section
         if len(self.LibraryClass) == 0 and len(RecordList) == 0:
             if self.ModuleType == SUP_MODULE_DXE_DRIVER or self.ModuleType == SUP_MODULE_PEIM or self.ModuleType == SUP_MODULE_DXE_SMM_DRIVER or \
```

### Build Environment

```markdown
- OS(s): Any
- Tool Chain(s):GCC
```

### Version Information

```text
Tag stable202602
```

### Urgency

Low

I will fix it

Maintainer feedback requested

### Anything else?

_No response_

EricGao2015:
# Description

Fixes https://github.com/tianocore/edk2/issues/12276

The INF specification said:

> If the module is not a library (no LIBRARY_CLASS in the [Defines] section) and the MODULE_TYPE is SEC, SMM_CORE, DXE_CORE, PEI_CORE, UEFI_DRIVER, UEFI_APPLICATION or HOST_APPLICATION a Depex section is not permitted. 

Thus, this series of patches does the following:
- Remove improper Depex section in UEFI_DRIVER and UEFI_APPLICATION.
- Adds a check when retrieving dependency expression in BaseTools. If the module has [Depex] section but it's not permitted by INF specification, just throw warning message t instead of break the building proccess.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- github CI check is passed. 
- Build every platform, check if there is building failure. Run the following commands for instance: 
`build -a X64 -t GCC -p RedfishPkg/RedfishPkg.dsc`
`build -a X64 -t GCC -p NetworkPkg/NetworkPkg.dsc`
`build -a X64 -t GCC -p MdeModulePkg/MdeModulePkg.dsc`
`build -a X64 -t GCC -p MdePkg/MdePkg.dsc`
`build -a X64 -t GCC -p ShellPkg/ShellPkg.dsc`
`...`
- Add [Depex] in UEFI_DRIVER or UEFI_APPLICATION, building process will not be breaken in BaseTools and a warning message will be throwed when build the driver:
`[Depex] section is not permitted for UEFI_DRIVER/UEFI_APPLICATION module`

## Integration Instructions

Once the PR is merged, all UEFI drivers and uefi applications with [Depex] section will not be blocked by BaseTools when building but only warning message is throwed. If the driver is UEFI driver, remove Depex from INF file and check dependent protocol in driver binding Start() function or notify function if necessary.

### DIFF
--- a/NetworkPkg/TcpDxe/TcpDxe.inf
+++ b/NetworkPkg/TcpDxe/TcpDxe.inf
@@ -88,8 +88,5 @@
   gEfiHashAlgorithmMD5Guid                      ## CONSUMES
   gEfiHashAlgorithmSha256Guid                   ## CONSUMES
 
-[Depex]
-  gEfiHash2ServiceBindingProtocolGuid
-
 [UserExtensions.TianoCore."ExtraFiles"]
   TcpDxeExtra.uni
--- a/BaseTools/Source/Python/Workspace/InfBuildData.py
+++ b/BaseTools/Source/Python/Workspace/InfBuildData.py
@@ -772,6 +772,15 @@ class InfBuildData(ModuleBuildClassObject):
             return RetVal
 
         RecordList = self._RawData[MODEL_EFI_DEPEX, self._Arch]
+
+        # If the module is not a library (no LIBRARY_CLASS in the [Defines] section) and the MODULE_TYPE is
+        # SEC, SMM_CORE, DXE_CORE, PEI_CORE, UEFI_DRIVER, UEFI_APPLICATION or HOST_APPLICATION a Depex section
+        # is not permitted.
+        if len(self.LibraryClass) == 0 and len(RecordList) != 0:
+            if self.ModuleType in [SUP_MODULE_SEC, SUP_MODULE_SMM_CORE, SUP_MODULE_DXE_CORE, SUP_MODULE_PEI_CORE, SUP_MODULE_UEFI_DRIVER, \
+                SUP_MODULE_UEFI_APPLICATION, SUP_MODULE_HOST_APPLICATION]:
+                EdkLogger.warn(None, "[Depex] section is not permitted for [%s] module" % self.ModuleType, File=self.MetaFile)
+
         # PEIM and DXE drivers must have a valid [Depex] section
         if len(self.LibraryClass) == 0 and len(RecordList) == 0:
             if self.ModuleType == SUP_MODULE_DXE_DRIVER or self.ModuleType == SUP_MODULE_PEIM or self.ModuleType == SUP_MODULE_DXE_SMM_DRIVER or \


leiflindholm:
Since it's a breaking change, PR description should have Integration Instructions.

NetworkPkg/TcpDxe/TcpDxe.inf
```diff
@@ -16,7 +16,7 @@
   INF_VERSION                    = 0x00010005
   BASE_NAME                      = TcpDxe
   FILE_GUID                      = 1A7E4468-2F55-4a56-903C-01265EB7622B
-  MODULE_TYPE                    = UEFI_DRIVER
+  MODULE_TYPE                    = DXE_DRIVER
```
mdkinney: I think the intent of all the network stack drivers is to be UEFI Drivers.  Converting to DXE_DRIVER does not lool right.  I think the bug may be that there is a depex for gEfiHash2ServiceBindingProtocolGuid.  There are ways for a UEFI Driver to be notified when a dependent protocol is available or it can fail a driver binding start if its dependent services are not available.

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.inf
```diff
@@ -14,7 +14,7 @@
   INF_VERSION               = 0x0001000b
   BASE_NAME                 = RedfishConfigHandlerDriver
   FILE_GUID                 = 6e881000-5749-11e8-9bf0-8cdcd426c973
-  MODULE_TYPE               = UEFI_DRIVER
+  MODULE_TYPE               = DXE_DRIVER
```
mdkinney: Same comment as TcpDxe.  Need input from @AbnerChang to know if this is intended to be a UEFI Driver or if it should be a DXE Driver.  If UEFI Driver, then Depex need to be removed and instead, the driver implementation needs to handle cases where dependent services are available or not.

EricGao2015:
> Since it's a breaking change, PR description should have Integration Instructions.

Updated, please check again. It's my understand for Integration Instructions.

niruiyu:
I think it may cause combability issue as some downstream usages may already there.

EricGao2015:
> I think it may cause combability issue as some downstream usages may already there.

Yes, this is also the aspect that I conern about. This issue has persisted for more than a decade, the PR will cause breaking change to downstream. So I would like to know the opinions of all stewards and reviewers.

mdkinney:
I have 2 open comments above that need to be addressed before this PR can be completed.

For the INFs that have a [Depex] of TRUE for module types that do not support [Depex].  I agree those are incorrect, but are effectively harmless.  For example a UEFI Driver has an implied depex of all PI Arch Protocols.  ANDing in an additional TRUE does not change the depex.  I would like to see those cleaned up. So the patch to remove [Depex]  TRUE for those cases looks correct.

The BaseTools change to raise this to an error may be good long term.  Short term, raising it to a WARNING so downstream consumers can have a chance to clean up the warnings is likely a better first step.  The challenge is deciding when it can be promoted from a WARNING to an ERROR.

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.inf
EricGao2015: > Same comment as TcpDxe. Need input from @AbnerChang to know if this is intended to be a UEFI Driver or if it should be a DXE Driver. If UEFI Driver, then Depex need to be removed and instead, the driver implementation needs to handle cases where dependent services are available or not.

@changab Hi Abner, please help to confirm this issue, thanks.

EricGao2015:
> I have 2 open comments above that need to be addressed before this PR can be completed.

I will handle the 2 comments after getting Anber's ack.
> 
> For the INFs that have a [Depex] of TRUE for module types that do not support [Depex]. I agree those are incorrect, but are effectively harmless. For example a UEFI Driver has an implied depex of all PI Arch Protocols. ANDing in an additional TRUE does not change the depex. I would like to see those cleaned up. So the patch to remove [Depex] TRUE for those cases looks correct.
> 
> The BaseTools change to raise this to an error may be good long term. Short term, raising it to a WARNING so downstream consumers can have a chance to clean up the warnings is likely a better first step. The challenge is deciding when it can be promoted from a WARNING to an ERROR.

I agree with the gentle approach that raises warning instead of error. I have changed the print level form EdkLogger.error() to EdkLogger.warn(), so that the building will not be breaken.

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.inf
changab: This is an UEFI driver, we have to remove the DEPEX. Let me confirm this.

felixpgithub:
The PR does not produce warnings on transitive dependencies (UEF driver uses a library instance with [Depex]). Is this intentional?

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.inf
EricGao2015: > This is an UEFI driver, we have to remove the DEPEX. Let me confirm this.

Hi Anber, I have removed Depex from RedfishConfigHandler driver and use notify function to handle the dependency. Please review it again. Thank you.

EricGao2015:
> The PR does not produce warnings on transitive dependencies (UEF driver uses a library instance with [Depex]). Is this intentional?

Yes. This patch only handles the case that Depex in INF file which MODULE_TYPE is UEFI_DRIVER or UEFI_APPLICATION, because this constitutes a clear violation of the INF specification.
I think the case you mentioned is valid because the INF specification does not forbid this behavior. Actually, if transitive dependencies really exist in UEFI_DRIVER, the driver may be treated as a DXE_DRIVER in BaseTools AutoGen code:
https://github.com/tianocore/edk2/blob/b7cf7e465c402f9b32dcadc791ba37f2db750522/BaseTools/Source/Python/AutoGen/ModuleAutoGen.py#L1342

One example is: `NetworkPkg/Ip4Dxe/Ip4Dxe.inf` is a UEFI_DRIVER without Depex, but this driver uses IoLib(`NetworkPkg/Library/DxeNetLib/DxeNetLib.inf`) which has dependency on gEfiRngProtocolGuid. After processing by AutoGen, the MODULE_TYPE becomes DXE_DRIVER in `Build/NetworkPkg/DEBUG_GCC/X64/NetworkPkg/Ip4Dxe/Ip4Dxe/OUTPUT/Ip4Dxe.inf`.

mdkinney:
> > The PR does not produce warnings on transitive dependencies (UEF driver uses a library instance with [Depex]). Is this intentional?
> 
> Yes. This patch only handles the case that Depex in INF file which MODULE_TYPE is UEFI_DRIVER or UEFI_APPLICATION, because this constitutes a clear violation of the INF specification. I think the case you mentioned is valid because the INF specification does not forbid this behavior. Actually, if transitive dependencies really exist in UEFI_DRIVER, the driver may be treated as a DXE_DRIVER in BaseTools AutoGen code:
> 
> https://github.com/tianocore/edk2/blob/b7cf7e465c402f9b32dcadc791ba37f2db750522/BaseTools/Source/Python/AutoGen/ModuleAutoGen.py#L1342
> 
> One example is: `NetworkPkg/Ip4Dxe/Ip4Dxe.inf` is a UEFI_DRIVER without Depex, but this driver uses IoLib(`NetworkPkg/Library/DxeNetLib/DxeNetLib.inf`) which has dependency on gEfiRngProtocolGuid. After processing by AutoGen, the MODULE_TYPE becomes DXE_DRIVER in `Build/NetworkPkg/DEBUG_GCC/X64/NetworkPkg/Ip4Dxe/Ip4Dxe/OUTPUT/Ip4Dxe.inf`.

Library INFs that support multiple module types and Depex need to make sure their [Depex] section is correct.  The [Depex] section does allow syntax to provide a module type specific depex such as:

```
[Depex.common.DXE_DRIVER, Depex.common.DXE_RUNTIME_DRIVER, Depex.common.DXE_SMM_DRIVER]
  gEfiFileExplorerProtocolGuid
```

The DxeNetLib lists the following module compatibiliy:

```
  LIBRARY_CLASS                  = NetLib|DXE_CORE DXE_DRIVER DXE_RUNTIME_DRIVER DXE_SMM_DRIVER UEFI_APPLICATION UEFI_DRIVER
```

But has a depex section that is applied to all module types the library supports

```
[Depex]
  gEfiRngProtocolGuid
```

So either DxeNetLib is listing extra module types that need to be removed or the [Depex] statement needs to be updated.  I do not think that `DXE_CORE` or `DXE_RUNTIME_DRIVER` or `DXE_SMM_DRIVER`  make much sense for this library.  It would help to remove ones that can never be used.

Also, the dependency on The RNG Protocol can be handled differently for DXE_DRIVER and UEFI_DRIVER and UEFI_APPLICATION.  For UEFI_DRIVER and UEFI_APPLICATION, the depex must be empty and the constructor can look for the RNG Protocol or use a RegisterProtocolNotify().  Only DXE_DRIVER can use a direct [Depex] on RNG Protocol.  The most compatible lib would be to remove the [Depex] all together and use RegisterProtocolNotify() and have services that do not work until the RNG Protocol is available.

In general, the mix of UEFI_DRIVER and DXE_DRIVER in the network stack does not make sense.  The network stack should be pure UEFI Drivers that can be loaded from Boot Manager or the Shell as well as embedded in platform FW.

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.inf
changab: @EricGao2015, I didn't know you were working on this. Please refer to https://github.com/tianocore/edk2/pull/12395, I have tested my fix on AMD platform.  We don't need the notification for this driver, because the Redfish credential driver is a DXE driver. The Redfish credential must be ready before the ConnectController(). You can update your PR, remove the changes for Redfish, and resend the PR  once 12395 gets merged. Thanks.

EricGao2015:
> > > The PR does not produce warnings on transitive dependencies (UEF driver uses a library instance with [Depex]). Is this intentional?
> > 
> > 
> > Yes. This patch only handles the case that Depex in INF file which MODULE_TYPE is UEFI_DRIVER or UEFI_APPLICATION, because this constitutes a clear violation of the INF specification. I think the case you mentioned is valid because the INF specification does not forbid this behavior. Actually, if transitive dependencies really exist in UEFI_DRIVER, the driver may be treated as a DXE_DRIVER in BaseTools AutoGen code:
> > https://github.com/tianocore/edk2/blob/b7cf7e465c402f9b32dcadc791ba37f2db750522/BaseTools/Source/Python/AutoGen/ModuleAutoGen.py#L1342
> > 
> > One example is: `NetworkPkg/Ip4Dxe/Ip4Dxe.inf` is a UEFI_DRIVER without Depex, but this driver uses IoLib(`NetworkPkg/Library/DxeNetLib/DxeNetLib.inf`) which has dependency on gEfiRngProtocolGuid. After processing by AutoGen, the MODULE_TYPE becomes DXE_DRIVER in `Build/NetworkPkg/DEBUG_GCC/X64/NetworkPkg/Ip4Dxe/Ip4Dxe/OUTPUT/Ip4Dxe.inf`.
> 
> Library INFs that support multiple module types and Depex need to make sure their [Depex] section is correct. The [Depex] section does allow syntax to provide a module type specific depex such as:
> 
> ```
> [Depex.common.DXE_DRIVER, Depex.common.DXE_RUNTIME_DRIVER, Depex.common.DXE_SMM_DRIVER]
>   gEfiFileExplorerProtocolGuid
> ```
> 
> The DxeNetLib lists the following module compatibiliy:
> 
> ```
>   LIBRARY_CLASS                  = NetLib|DXE_CORE DXE_DRIVER DXE_RUNTIME_DRIVER DXE_SMM_DRIVER UEFI_APPLICATION UEFI_DRIVER
> ```
> 
> But has a depex section that is applied to all module types the library supports
> 
> ```
> [Depex]
>   gEfiRngProtocolGuid
> ```
> 
> So either DxeNetLib is listing extra module types that need to be removed or the [Depex] statement needs to be updated. I do not think that `DXE_CORE` or `DXE_RUNTIME_DRIVER` or `DXE_SMM_DRIVER` make much sense for this library. It would help to remove ones that can never be used.
> 
> Also, the dependency on The RNG Protocol can be handled differently for DXE_DRIVER and UEFI_DRIVER and UEFI_APPLICATION. For UEFI_DRIVER and UEFI_APPLICATION, the depex must be empty and the constructor can look for the RNG Protocol or use a RegisterProtocolNotify(). Only DXE_DRIVER can use a direct [Depex] on RNG Protocol. The most compatible lib would be to remove the [Depex] all together and use RegisterProtocolNotify() and have services that do not work until the RNG Protocol is available.
> 
> In general, the mix of UEFI_DRIVER and DXE_DRIVER in the network stack does not make sense. The network stack should be pure UEFI Drivers that can be loaded from Boot Manager or the Shell as well as embedded in platform FW.

Thanks for your explanation, I understood your thought. The DxeNetLib issue is different with the subject of this PR, so I'll submit another PR to fix the issue later.

mikebeaton:
> In general, the mix of UEFI_DRIVER and DXE_DRIVER in the network stack does not make sense. The network stack should be pure UEFI Drivers that can be loaded from Boot Manager or the Shell as well as embedded in platform FW.

@mdkinney - Hi. I'm not sure what _should_ happen, but if not for your comment I would have said I am almost certain that DXE Drivers can be loaded from the Shell. A quick Google seems to confirm, though I haven't had a chance to retest what I thought I knew.

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.inf
changab: https://github.com/tianocore/edk2/pull/12395 has been merged.

NetworkPkg/TcpDxe/TcpDxe.inf
EricGao2015: About TcpDxe.inf, I only removed the [Depex] section and the MOULE_TYPE is still UEFI_DRIVER. The dependency of gEfiHash2ServiceBindingProtocolGuid has been guaranteed. In Tcp6DriverBindingStart() and Tcp4DriverBindingStart() function, gEfiHash2ServiceBindingProtocol has been checked by calling TcpCreateService function if it is available by gBS->LocateProtocol().
https://github.com/tianocore/edk2/blob/2fb6e81c06b63cb8ab96ab82f1df2674492530a4/NetworkPkg/TcpDxe/TcpDriver.c#L310-L317
So the conversation can be resolved.

EricGao2015:
> > I have 2 open comments above that need to be addressed before this PR can be completed.
> 
> I will handle the 2 comments after getting Anber's ack.

About TcpDxe.inf, I only removed the [Depex] section and the MOULE_TYPE is still UEFI_DRIVER. The dependency of gEfiHash2ServiceBindingProtocolGuid has been guaranteed. In Tcp6DriverBindingStart() and Tcp4DriverBindingStart() function, gEfiHash2ServiceBindingProtocol has been checked by calling TcpCreateService function if it is available by gBS->LocateProtocol().
About RedfishConfigHandlerDriver.inf, Anber has fixed this issue in https://github.com/tianocore/edk2/pull/12395 and the PR is already merged.

> 
> > For the INFs that have a [Depex] of TRUE for module types that do not support [Depex]. I agree those are incorrect, but are effectively harmless. For example a UEFI Driver has an implied depex of all PI Arch Protocols. ANDing in an additional TRUE does not change the depex. I would like to see those cleaned up. So the patch to remove [Depex] TRUE for those cases looks correct.
> > The BaseTools change to raise this to an error may be good long term. Short term, raising it to a WARNING so downstream consumers can have a chance to clean up the warnings is likely a better first step. The challenge is deciding when it can be promoted from a WARNING to an ERROR.
> 
> I agree with the gentle approach that raises warning instead of error. I have changed the print level from EdkLogger.error() to EdkLogger.warn(), so that the building will not be broken.

@mdkinney Hi Michael, the two comments have solved and the print level has adjusted. could you please review this PR again?

EricGao2015:
@lgao4 @mdkinney Dear maintainers, This PR is under reviwing for a long time, do you have any comment on this PR?

PR #12344, opened 2026-03-26
NetworkPkg: Fix multiple security vulnerabilities in DHCP/PXE/HTTP Boot

AbuthahirM11:
Fix 9 security vulnerabilities found in NetworkPkg DHCP, PXE Boot,
and HTTP Boot option parsing code. These issues include out-of-bounds
reads/writes, integer overflow, infinite loop, and double-free
conditions that can be triggered by a malicious DHCP server.

Fixes:
- HttpBootDxe: Fix double-free in HttpBootGetBootFile error path
- DxeHttpLib: Fix infinite loop in HttpParseMessageBody
- UefiPxeBcDxe: Fix OOB access in DHCPv6 URL percent-decoding
- UefiPxeBcDxe: Fix OOB write via BOOTFILE option in PxeBcParseDhcp4Packet
- UefiPxeBcDxe: Fix UINT8 Offset overflow in PxeBcParseVendorOptions
- UefiPxeBcDxe: Add length validation to PXE vendor option handlers
- HttpBootDxe: Fix OOB write via BOOTFILE option in HttpBootParseDhcp4Packet
- HttpBootDxe: Fix OOB read in HttpBootParseDhcp4Options
- HttpBootDxe: Fix OOB write via GET_OPTION_BUFFER_LEN underflow

Cc: Michael D Kinney <michael.d.kinney@intel.com>

Signed-off-by: Abuthahir M <abuthahirm@ami.com>

### DIFF
--- a/NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
+++ b/NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
@@ -280,6 +280,9 @@ HttpBootParseDhcp4Packet (
 
   Offer   = &Cache4->Packet.Offer;
   Options = Cache4->OptList;
+  if (Offer->Length < sizeof (EFI_DHCP4_HEADER) + 4) {
+    return EFI_DEVICE_ERROR;
+  }
 
   //
   // Parse DHCPv4 options in this offer, and store the pointers.
@@ -617,6 +620,10 @@ HttpBootDhcp4CallBack (
   }
 
   Private = (HTTP_BOOT_PRIVATE_DATA *)Context;
+
+  if (Packet->Length < sizeof (EFI_DHCP4_HEADER) + 4) {
+    return EFI_SUCCESS;
+  }
 
   //
   // Override the Maximum DHCP Message Size.
--- a/NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
+++ b/NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
@@ -353,8 +353,16 @@ HttpBootParseDhcp4Packet (
     // RFC 2132, Section 9.5 does not strictly state Bootfile name (option 67) is null
     // terminated string. So force to append null terminated character at the end of string.
     //
+    if (Options[HTTP_BOOT_DHCP4_TAG_INDEX_BOOTFILE]->Length == 0) {
+      return EFI_DEVICE_ERROR;
+    }
+
     Ptr8  =  (UINT8 *)&Options[HTTP_BOOT_DHCP4_TAG_INDEX_BOOTFILE]->Data[0];
     Ptr8 += Options[HTTP_BOOT_DHCP4_TAG_INDEX_BOOTFILE]->Length;
+    if (Ptr8 > (UINT8 *)Offer->Dhcp4.Option + GET_OPTION_BUFFER_LEN (Offer)) {
+      return EFI_DEVICE_ERROR;
+    }
+
     if (*(Ptr8 - 1) != '\0') {
       *Ptr8 = '\0';
     }
--- a/NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
+++ b/NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
@@ -199,7 +199,15 @@ HttpBootParseDhcp4Options (
     if (Option->OpCode == DHCP4_TAG_PAD) {
       Offset++;
     } else {
+      if (Offset + Option->Length + 2 > Length) {
+        return NULL;
+      }
+
       Offset += Option->Length + 2;
+    }
+
+    if (Offset >= Length) {
+      return NULL;
     }
 
     Option = (EFI_DHCP4_PACKET_OPTION *)(Buffer + Offset);
--- a/NetworkPkg/Library/DxeHttpLib/DxeHttpLib.c
+++ b/NetworkPkg/Library/DxeHttpLib/DxeHttpLib.c
@@ -1184,6 +1184,8 @@ HttpParseMessageBody (
     switch (Parser->State) {
       case BodyParserStateMax:
         return EFI_ABORTED;
+      case BodyParserComplete:
+        return EFI_SUCCESS;
 
       case BodyParserBodyIdentity:
         //
--- a/NetworkPkg/UefiPxeBcDxe/PxeBcDhcp4.c
+++ b/NetworkPkg/UefiPxeBcDxe/PxeBcDhcp4.c
@@ -90,7 +90,7 @@ PxeBcParseVendorOptions (
   UINT32                   *BitMap;
   UINT8                    VendorOptionLen;
   EFI_DHCP4_PACKET_OPTION  *PxeOption;
-  UINT8                    Offset;
+  UINT32                   Offset;
 
   BitMap          = VendorOption->BitMap;
   VendorOptionLen = Dhcp4Option->Length;
@@ -194,7 +194,11 @@ PxeBcParseVendorOptions (
     if (PxeOption->OpCode == DHCP4_TAG_PAD) {
       Offset++;
     } else {
-      Offset = (UINT8)(Offset + PxeOption->Length + 2);
+      if (Offset + (UINT32)PxeOption->Length + 2 > VendorOptionLen) {
+        break;
+      }
+
+      Offset += (UINT32)PxeOption->Length + 2;
     }
 
     PxeOption = (EFI_DHCP4_PACKET_OPTION *)(Dhcp4Option->Data + Offset);
--- a/NetworkPkg/UefiPxeBcDxe/PxeBcDhcp4.c
+++ b/NetworkPkg/UefiPxeBcDxe/PxeBcDhcp4.c
@@ -545,8 +545,16 @@ PxeBcParseDhcp4Packet (
     // RFC 2132, Section 9.5 does not strictly state Bootfile name (option 67) is null
     // terminated string. So force to append null terminated character at the end of string.
     //
+    if (Options[PXEBC_DHCP4_TAG_INDEX_BOOTFILE]->Length == 0) {
+      return EFI_DEVICE_ERROR;
+    }
+
     Ptr8  =  (UINT8 *)&Options[PXEBC_DHCP4_TAG_INDEX_BOOTFILE]->Data[0];
     Ptr8 += Options[PXEBC_DHCP4_TAG_INDEX_BOOTFILE]->Length;
+    if (Ptr8 > (UINT8 *)Offer->Dhcp4.Option + GET_OPTION_BUFFER_LEN (Offer)) {
+      return EFI_DEVICE_ERROR;
+    }
+
     if (*(Ptr8 - 1) != '\0') {
       *Ptr8 = '\0';
     }
--- a/NetworkPkg/UefiPxeBcDxe/PxeBcDhcp4.c
+++ b/NetworkPkg/UefiPxeBcDxe/PxeBcDhcp4.c
@@ -105,38 +105,52 @@ PxeBcParseVendorOptions (
     //
     switch (PxeOption->OpCode) {
       case PXEBC_VENDOR_TAG_MTFTP_IP:
+        if (PxeOption->Length >= sizeof (EFI_IPv4_ADDRESS)) {
+          CopyMem (&VendorOption->MtftpIp, PxeOption->Data, sizeof (EFI_IPv4_ADDRESS));
+        }
 
-        CopyMem (&VendorOption->MtftpIp, PxeOption->Data, sizeof (EFI_IPv4_ADDRESS));
         break;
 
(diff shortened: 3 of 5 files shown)


mikebeaton:
Hi - You're probably aware, but to fix the patch check errors, you need to click on the failed job, click to 'view more details on Azure Pipelines', keep clicking through to the failed stage... eventually you get to the actual error (your commit subject lines are too long).

AbuthahirM11:
Thanks - I’ve gone through the logs and fixed the patch check errors.

NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
SaloniKasbekar, line 176: This function is called multiple times where NULL is not checked. Can you update the PR to ensure returning NULL doesnt cause unintended behavior by the callers?

AbuthahirM11: Thank you for the review, @SaloniKasbekar! I've traced all call sites of HttpBootParseDhcp4Options and confirmed that NULL is safely handled at each one — HttpBootDhcp4CallBack guards the result with if (MaxMsgSize != NULL), and all Options[Index] entries in HttpBootParseDhcp4Packet are checked before use. The new NULL returns added by this PR (on bounds-check failure) are treated identically to the existing "option not found" NULL path, so no unintended behavior is introduced. Please let me know if you spot a path I missed!

AbuthahirM11:
@leiflindholm @ajfish @ardbiesheuvel @makubacki @mdkinney @mikebeaton  @Zclarkwilliams — Could you please take a look and approve this PR when you get a chance?

NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
```diff
@@ -199,9 +199,17 @@ HttpBootParseDhcp4Options (
     if (Option->OpCode == DHCP4_TAG_PAD) {
       Offset++;
     } else {
+      if ((Offset + 2 > Length) || (Offset + Option->Length + 2 > Length)) {
```
mikebeaton: Why do we need both tests here?

AbuthahirM11: @mikebeaton Good catch. Offset + Option->Length + 2 > Length already covers the case where fewer than 2 bytes remain (when Option->Length == 0, it reduces to Offset + 2 > Length). The redundant first check has been removed. Updated.

AbuthahirM11:
@leiflindholm @ajfish @ardbiesheuvel @makubacki @mdkinney @mikebeaton @Zclarkwilliams — Could you please take a look and approve this PR when you get a chance?

NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
makubacki: @AbuthahirM11, please resolve the thread if no further work remains.

NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
makubacki: @SaloniKasbekar & @AbuthahirM11, please resolve this thread when the open is closed.

NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
AbuthahirM11: Hi @makubacki, the thread has been resolved. Thank you!

SaloniKasbekar (approved):

AbuthahirM11:
@leiflindholm @ajfish @ardbiesheuvel @makubacki @mdkinney @mikebeaton @Zclarkwilliams — Could you please take a look and approve this PR when you get a chance?

NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
```diff
@@ -199,9 +199,17 @@ HttpBootParseDhcp4Options (
     if (Option->OpCode == DHCP4_TAG_PAD) {
       Offset++;
     } else {
+      if (Offset + Option->Length + 2 > Length) {
+        return NULL;
+      }
```
mikebeaton, line 204: Do we need this? Won't the new test immediately below also catch this?

mikebeaton (changes requested):
Apart from the comment (which is a nit and wouldn't change behaviour) these are genuine fixes to genuine errors.

NetworkPkg/Library/DxeHttpLib/DxeHttpLib.c
```diff
@@ -1184,6 +1184,8 @@ HttpParseMessageBody (
     switch (Parser->State) {
       case BodyParserStateMax:
         return EFI_ABORTED;
+      case BodyParserComplete:
```
leiflindholm, line 1187: This is the minimal possible fix, but I'd prefer having the two cases below return directly instead of going another round through the loop in order to return.

And adding some error handling of unhandled parser states instead of silently ignoring them as is currently done at the `default:` label.

NetworkPkg/UefiPxeBcDxe/PxeBcDhcp6.c
```diff
@@ -545,12 +545,18 @@ PxeBcExtractBootFileUrl (
     //
     while (*BootFileNamePtr != '\0') {
       if (*BootFileNamePtr == '%') {
-        TmpChar               = *(BootFileNamePtr+ 3);
```
leiflindholm, line 548: I agree this is a bug, but why is the correct fix to try to reinterpret an invalid URL as a different valid one instead of returning an error?

NetworkPkg/UefiPxeBcDxe/PxeBcDhcp4.c
```diff
@@ -194,7 +194,11 @@ PxeBcParseVendorOptions (
     if (PxeOption->OpCode == DHCP4_TAG_PAD) {
       Offset++;
     } else {
-      Offset = (UINT8)(Offset + PxeOption->Length + 2);
+      if (Offset + (UINT32)PxeOption->Length + 2 > VendorOptionLen) {
+        break;
```
leiflindholm, line 198: Doesn't this mean we'd be silently ignoring misformed packets?
Since this function _can_ encounter error conditions, I think it's suboptimal that it returns VOID.
Maybe we should change that?

NetworkPkg/UefiPxeBcDxe/PxeBcDhcp4.c
```diff
@@ -105,38 +105,52 @@ PxeBcParseVendorOptions (
     //
     switch (PxeOption->OpCode) {
       case PXEBC_VENDOR_TAG_MTFTP_IP:
+        if (PxeOption->Length >= sizeof (EFI_IPv4_ADDRESS)) {
```
leiflindholm, line 108: Yeah, I think this strengthens the case for making the function return a status.
Then we can use error handling instead of success handling as well.

NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
```diff
@@ -353,8 +353,16 @@ HttpBootParseDhcp4Packet (
     // RFC 2132, Section 9.5 does not strictly state Bootfile name (option 67) is null
     // terminated string. So force to append null terminated character at the end of string.
     //
+    if (Options[HTTP_BOOT_DHCP4_TAG_INDEX_BOOTFILE]->Length == 0) {
```
leiflindholm, line 356: Can we merge this commit with the one with the identical subject line?

In fact, the fixes are to duplicate identical code sequences.
Can you break it out into a helper function and fix the pattern once?

NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
leiflindholm: Also, why one comparison on `> Length` and one on `>= Length`?

leiflindholm (changes requested):

NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
```diff
@@ -280,6 +280,9 @@ HttpBootParseDhcp4Packet (
 
   Offer   = &Cache4->Packet.Offer;
   Options = Cache4->OptList;
+  if (Offer->Length < sizeof (EFI_DHCP4_HEADER) + 4) {
```
leiflindholm, line 283: Better to use `sizeof (Offer->Packet.Dhcp4.Magik)` than live-coding an integer.

NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
```diff
@@ -618,6 +621,10 @@ HttpBootDhcp4CallBack (
 
   Private = (HTTP_BOOT_PRIVATE_DATA *)Context;
 
+  if (Packet->Length < sizeof (EFI_DHCP4_HEADER) + 4) {
```
leiflindholm, line 624: Same as above, replace the 4 with the appropriate sizeof.

NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
```diff
@@ -618,6 +621,10 @@ HttpBootDhcp4CallBack (
 
   Private = (HTTP_BOOT_PRIVATE_DATA *)Context;
 
+  if (Packet->Length < sizeof (EFI_DHCP4_HEADER) + 4) {
+    return EFI_SUCCESS;
```
leiflindholm, line 625: Is there any reason why we shouldn't abort attempt if a response is undersized?

First of all, we should only end up here if the DHCP driver has determined there's a valid packet to process. If we assume (which may be wrong) that this driver is used with our own Dhcp4Dxe, this happens at
https://github.com/tianocore/edk2/blob/f1f89f454cd8c9aabae7ce5e690f612bd969f5ff/NetworkPkg/Dhcp4Dxe/Dhcp4Io.c#L963

I guess my question is, what error case is this test attempting to protect against? (This of course applies to the other change in this commit also.)

leiflindholm (changes requested):

NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
mikebeaton: @leiflindholm - Should GET_OPTION_BUFFER_LEN be modified too, if changing this?

leiflindholm: @mikebeaton I'm less fussed about that since it's hidden from view other than if someone's curious about the implementation, but in a perfect world, yes.

mikebeaton: It might be easier to understand the check afterwards, if the way it's expressed matches what it is - in the end - checking for.

NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
AbuthahirM11: @mikebeaton Yes, both checks are needed — they guard different things. The first (> Length) validates that Option->Length itself is sane before using it to advance Offset. Without it, a maliciously large Option->Length could cause Offset to wrap or jump far past the buffer, potentially bypassing the second check entirely. The second (>= Length) then guards the pointer dereference after the advance.

@leiflindholm The difference is intentional. > Length is used before the advance because Offset landing exactly at Length is valid — the option fits flush to the end of the buffer. >= Length is used after the advance because even Offset == Length means zero bytes remain to dereference as an option header, making it out-of-bounds.

NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
AbuthahirM11: @leiflindholm @mikebeaton Thanks for the review, Fixed in that new PR [#12809](https://github.com/tianocore/edk2/pull/12809)

NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
AbuthahirM11: @leiflindholm HttpBootDxe and UefiPxeBcDxe are separate driver modules compiled into separate binaries — they cannot share a function without modifying a common library like NetLib, which is out of scope for a security fix. Added a STATIC helper in each file to consolidate the pattern within each module. The commits have distinct subject lines (NetworkPkg/HttpBootDxe: vs NetworkPkg/UefiPxeBcDxe:) so they cannot be merged — edk2 convention requires one logical change per package prefix per commit.

NetworkPkg/HttpBootDxe/HttpBootDhcp4.c
AbuthahirM11: @leiflindholm Fixed in that new PR [#12809](https://github.com/tianocore/edk2/pull/12809)

PR #12345, opened 2026-03-26
MdeModulePkg/AcpiTableDxe: Free the Hob memory

GeorgeLiao012:
# Description
Some ACPI tables are allocated the memory and full initialized during the PEI phase. After these tables are passed to 
the DXE phase and installed in the ACPI-type memory, the original memory allocated for ACPI tables in the PEI phase 
is no longer needed and should be be freed.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Boots to the OS.

## Integration Instructions
None

### DIFF
--- a/MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiTableProtocol.c
+++ b/MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiTableProtocol.c
@@ -2197,6 +2197,11 @@ InstallAcpiTableFromAcpiSiliconHob (
         if (FeaturePcdGet (PcdInstallAcpiSdtProtocol)) {
           SdtNotifyAcpiList (AcpiTableInstance, Version, TableKey);
         }
+
+        //
+        // Free the memory allocated for the ACPI table during the PEI phase.
+        //
+        FreePool (SocEntryTable);
       }
     }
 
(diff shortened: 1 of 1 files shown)


GeorgeLiao012:
@mdkinney @ankit13s Could you please review it? Thanks.

MdeModulePkg/Universal/Acpi/AcpiTableDxe/AcpiTableProtocol.c
```diff
@@ -2197,6 +2197,11 @@ InstallAcpiTableFromAcpiSiliconHob (
         if (FeaturePcdGet (PcdInstallAcpiSdtProtocol)) {
           SdtNotifyAcpiList (AcpiTableInstance, Version, TableKey);
         }
+
+        //
+        // Free the memory allocated for the ACPI table during the PEI phase.
+        //
+        FreePool (SocEntryTable);
```
niruiyu, line 2204: it's not possible to free PEI-allocated memory in DXE phase.

GeorgeLiao012: Yes, thank you for pointing that out. The memory used in the PEI and DXE phases is different. 
I overlooked the fact that the memory type was changed to "Boot Services Data" type when allocating memory during the PEI phase. As a result, there is no need to free the ACPI table memory in the DXE phase.

GeorgeLiao012:
This PR is not required.

PR #12346, opened 2026-03-26
ArmPkg/ArmTransferListLib: Integrate libtl for transfer list handling

HarishManchikanti:
# Description

This series of commits introduces support for the Transfer List Library (libtl) in ArmPkg/ArmTransferListLib.

The goal is to align EDK2 with the Firmware Handoff Specification by reusing the common libtl implementation instead of maintaining separate custom logic.
ArmTransferListLib now acts as a thin wrapper around libtl, exposing the same functionality via an EDK2-style API so platform code can work with Transfer Lists consistently.

Summary of changes

Add libtl as a git submodule under ArmPkg/Library/ArmTransferListLib/libtl.
Update ArmTransferListLib.inf to build with libtl sources.
Replace previous in-tree custom Transfer List handling with calls into libtl.
Add STATIC_ASSERTs to guarantee structure size/layout matches libtl.
Guard register-convention macros in ArmTransferList.h to avoid conflicts with libtl headers.
Extend ArmTransferListLib with additional wrapper APIs and helpers.
This ensures EDK2 stays aligned with the libtl reference implementation and avoids duplication.

Breaking change?
No. Existing ArmTransferListLib API continues to be provided, now backed by libtl.

Impacts security?
No direct security changes. Transfer List parsing continues to validate headers and sizes.

Includes tests?
No new unit tests. Verified manually with builds and integration tests on platform firmware.

How This Was Tested

Built ArmPkg successfully for AARCH64 target.
Validated that ArmTransferListLib links against libtl sources.
Verified structure layout compatibility with libtl using STATIC_ASSERT.
Tested with ftpm branch for any issues .
Integration Instructions

Consumers of ArmTransferListLib continue to include <Library/ArmTransferListLib.h> and link as before.

Also updated edk2-platforms related PR to update platform DSC files to consume ArmTransferListLib from MdePkg:
https://github.com/tianocore/edk2-platforms/pull/1010


Javagedes (approved):
Approving for `.pytool/*` updates.

LeviYeoReum:
I think at first libtl should fix below build error:
```
INFO - /__w/1/s/ArmPkg/Library/ArmTransferListLib/libtl/include/transfer_list.h:126:56: note: expression evaluates to '9 == 8'
INFO -   126 | LIBTL_STATIC_ASSERT(sizeof(struct transfer_list_entry) == 0x8U,
```
But, This would require to fix libtl with below patch:

```
diff --git a/include/transfer_list.h b/include/transfer_list.h
index 89a54e3..db53de9 100644
--- a/include/transfer_list.h
+++ b/include/transfer_list.h
@@ -98,7 +98,7 @@ struct transfer_list_header {
 
 struct __attribute__((packed)) transfer_list_entry {
        uint32_t tag_id : 24;
-       uint8_t hdr_size;
+       uint32_t hdr_size: 8;
        uint32_t data_size;
        /*
```

HarishManchikanti:
> I think at first libtl should fix below build error:
> 
> ```
> INFO - /__w/1/s/ArmPkg/Library/ArmTransferListLib/libtl/include/transfer_list.h:126:56: note: expression evaluates to '9 == 8'
> INFO -   126 | LIBTL_STATIC_ASSERT(sizeof(struct transfer_list_entry) == 0x8U,
> ```
> 
> But, This would require to fix libtl with below patch:
> 
> ```
> diff --git a/include/transfer_list.h b/include/transfer_list.h
> index 89a54e3..db53de9 100644
> --- a/include/transfer_list.h
> +++ b/include/transfer_list.h
> @@ -98,7 +98,7 @@ struct transfer_list_header {
>  
>  struct __attribute__((packed)) transfer_list_entry {
>         uint32_t tag_id : 24;
> -       uint8_t hdr_size;
> +       uint32_t hdr_size: 8;
>         uint32_t data_size;
>         /*
> ```

Is there a designated contact for libtl changes?
Or am I allowed to propose and submit this fix myself?

LeviYeoReum:
> Is there a designated contact for libtl changes?
> Or am I allowed to propose and submit this fix myself?

You can submit the patch to libtl from youself. but Cc me or add tag with Suggested-by tag please.

Thanks.

os-d:
Is libtl buildable natively in Windows? ArmVirtPkg is, with CLANGPDB, it would be a large limitation if a submodule is added that isn’t. I ask because other Arm projects, like TF-A, are not.

LeviYeoReum:
> Is libtl buildable natively in Windows? ArmVirtPkg is, with CLANGPDB, it would be a large limitation if a submodule is added that isn’t. I ask because other Arm projects, like TF-A, are not.

I think I *should* be buildable and there is no reason not to support in Windows build for libtl
(even there is no excuse since TF-A is buildable in Windowes).

os-d:
> > Is libtl buildable natively in Windows? ArmVirtPkg is, with CLANGPDB, it would be a large limitation if a submodule is added that isn’t. I ask because other Arm projects, like TF-A, are not.
> 
> I think I _should_ be buildable and there is no reason not to support in Windows build for libtl (even there is no excuse since TF-A is buildable in Windowes).

TF-A is not buildable natively in Windows, it is buildable through msys2, but that is a pain to add to existing flows. That's what I want to ensure here, because we have partners who build ARM platforms on Windows and want to keep those flows, which I think is reasonable we don't break development flows.

I will check if we have CLANGPDB building on Windows for CI (either in edk2 all up or ArmVirtPkg) and propose adding a pipeline (or converting an existing one) to do the CLANGPDB Windows build.

os-d:
I was able to build libtl on Windows (through `stuart_ci_build -c .pytool\CISettings.py -p ArmPkg TOOL_CHAIN_TAG=CLANGPDB -t DEBUG BUILDMODULE=ArmPkg\Library\ArmTransferListLib\ArmTransferListLib.inf`). There are a couple of issues CLANGPDB finds, which are likely the same as why the pipelines failed here.

LeviYeoReum:
> struct __attribute__((packed)) transfer_list_entry {
>         uint32_t tag_id : 24;
> -       uint8_t hdr_size;
> +       uint32_t hdr_size: 8;
>         uint32_t data_size;
>         /*

I think the problem is CLANGPDB doesn't deploy on hdr_size in the tag_id
(That's why the sizeof(transfer_list_entry) becomes 9).
I expect above patch would deploy the tag_id & hdr_size in the one uint32_t but need to be tested.

ArmPkg/Include/IndustryStandard/ArmTransferList.h
```diff
@@ -128,19 +139,18 @@ typedef struct TransferListHeader {
...
 typedef struct TransferEntryHeader {
   /// The entry type identifier.
-  UINT16    TagId;
-
-  /// Reserved.
-  UINT8     Reserved0;
+  UINT32    TagId : 24;
 
   /// The size of this entry header in bytes.
   UINT8     HeaderSize;
```
LeviYeoReum: For CLANGPDB, `HeaderSize` should be:
```
UINT32 HeaderSize :8;
```

LeviYeoReum:
@HarishManchikanti 
JFYI for relevant fix in libtl: 
   - https://review.trustedfirmware.org/c/shared/transfer-list-library/+/49829

With this fix and what I commented, I've confirmed to build with CLANGPDB and work.

Once you resolve comments, feel free to add:

Reviewed-by: Yeoreum Yun <yeoreum.yun@arm.com>

LeviYeoReum:
> I was able to build libtl on Windows (through `stuart_ci_build -c .pytool\CISettings.py -p ArmPkg TOOL_CHAIN_TAG=CLANGPDB -t DEBUG BUILDMODULE=ArmPkg\Library\ArmTransferListLib\ArmTransferListLib.inf`). There are a couple of issues CLANGPDB finds, which are likely the same as why the pipelines failed here.

Interesting. But when I try to build with llvm-22 and CLANGPDB, I can see the same errors.
After commit on libtl:
    * https://review.trustedfirmware.org/c/shared/transfer-list-library/+/49829
    
I've confirmed it doesn't complain any problem.

So, I think @HarishManchikanti should modify the libtl base on that commit.

ArmPkg/Include/IndustryStandard/ArmTransferList.h
HarishManchikanti: Addressed review comments in latest PR ,https://github.com/tianocore/edk2/pull/12346/changes/337c496bf018da2c5c2283c25e108f85fbed4255#diff-89b44e11be7607d9540d03350cdc1bbc167d0cd2ef91eedc68732387c91bd720

HarishManchikanti:
@LeviYeoReum  Review comments addressed .

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -68,73 +80,67 @@ TransferListVerifyChecksum (
...
+  @param[in]  ops                     libtl Operation modes.
+
+  @retval TRANSFER_LIST_OPS_ALL       Header is valid and usable (read/write).
+  @retval TRANSFER_LIST_OPS_RO        Header is valid but only supports read-only
+                                      operations.
+  @retval TRANSFER_LIST_OPS_CUSTOM    Header is valid but requires custom handling
+                                      due to a version mismatch.
+  @retval TRANSFER_LIST_OPS_INVALID   Header is invalid or unsupported.
+**/
+static TRANSFER_LIST_OPS
```
LeviYeoReum: static -> STATIC?

and I think

STATIC
TRANSFER_LIST_OPS
EFIAPI
MapLibtlOps (...

HarishManchikanti: solved

LeviYeoReum (approved):

.gitmodules
```diff
@@ -38,3 +38,6 @@
 [submodule "TcgTpmPkg/Library/TpmLib/TPM"]
 	path = TcgTpmPkg/Library/TpmLib/TPM
 	url = https://github.com/TrustedComputingGroup/TPM
+[submodule "ArmPkg/Library/ArmTransferListLib/libtl"]
+	path = ArmPkg/Library/ArmTransferListLib/libtl
+	url = https://review.trustedfirmware.org/shared/transfer-list-library.git
```
samimujawar, line 43: I think the License section needs to be updated at https://github.com/tianocore/edk2/blob/master/ReadMe.rst?plain=1#L68

ReadMe.rst
```diff
@@ -98,6 +98,7 @@ that are covered by additional licenses.
 -  `MdePkg/Library/MipiSysTLib/mipisyst <https://github.com/MIPI-Alliance/public-mipi-sys-t/blob/aae857d0d05ac65152ed24992a4acd834a0a107c/LICENSE>`__
 -  `SecurityPkg/DeviceSecurity/SpdmLib/libspdm <https://github.com/DMTF/libspdm/blob/main/LICENSE.md>`__
 -  `TcgTpmPkg/Library/TpmLib/TPM <https://github.com/TrustedComputingGroup/TPM/blob/main/LICENSE>`__
+-  `ArmPkg/Library/ArmTransferListLib/libtl <https://review.trustedfirmware.org/plugins/gitiles/shared/transfer-list-library/+/refs/heads/main/LICENSE.md>`__
```
samimujawar, line 101: I think this change should be part of the previous patch?

ArmPkg/Library/ArmTransferListLib/TransferListLibSupport.h
```diff
@@ -0,0 +1,150 @@
...
+  Root include file of C runtime library to support building the third-party
+  Transfer List library.
+
+  Copyright (c) 2025, Arm Limited. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef TRANSFER_LIST_LIB_SUPPORT_H_
+#define TRANSFER_LIST_LIB_SUPPORT_H_
```
samimujawar, line 11: Use #pragma once

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -17,13 +17,17 @@
 #include <Library/ArmTransferListLib.h>
 #include <Library/HobLib.h>
 
+typedef struct transfer_list_header  LIBTL_TRANSFER_LIST_HEADER;
```
samimujawar, line 20: Should these typedefs be in the header file?

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -68,73 +79,67 @@ TransferListVerifyChecksum (
...
+
+  @retval TRANSFER_LIST_OPS_ALL       Header is valid and usable (read/write).
+  @retval TRANSFER_LIST_OPS_RO        Header is valid but only supports read-only
+                                      operations.
+  @retval TRANSFER_LIST_OPS_CUSTOM    Header is valid but requires custom handling
+                                      due to a version mismatch.
+  @retval TRANSFER_LIST_OPS_INVALID   Header is invalid or unsupported.
+**/
+static TRANSFER_LIST_OPS
+MapLibtlOps (
```
samimujawar, line 98: I think this function should be removed. The only place where transfer_list_ops is ever used is in transfer_list_check_header. I understand that TRANSFER_LIST_OPS_xxx was added to give a meaningful prefix to the enums in transfer_list_ops i.e. instead of TL_OPS_xxx. But both are exactly the same, and must be.

So instead, it would be good to drop this function and add STATIC_ASSERTs in TransferListCheckHeader() before calling transfer_list_check_header() and cast the return value to TRANSFER_LIST_OPS.

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -261,7 +230,7 @@ TransferListFindNextEntry (
     Entry = TransferListGetNextEntry (TransferListHeader, CurrentEntry);
   }
 
-  while ((Entry != NULL) && ((Entry->TagId != TagId) || Entry->Reserved0 != 0)) {
+  while ((Entry != NULL) && ((Entry->TagId & 0xFFFFFF) != (TagId & 0xFFFFFF))) {
```
samimujawar, line 233: TRANSFER_LIST_TAGID_MASK instead of magic numbers?

ArmPkg/Include/Library/ArmTransferListLib.h
```diff
@@ -9,7 +9,8 @@
     - https://github.com/FirmwareHandoff/firmware_handoff
 **/
 
-#pragma once
+#ifndef ARM_TRANSFER_LIST_LIB_
```
samimujawar, line 12: why not use #pragma once?

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
samimujawar: Same comment for line 280. Please also check at other locations.

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -377,3 +381,303 @@ TransferListDump (
...
+  )
+{
+  TRANSFER_ENTRY_HEADER  *Entry;
+
+  Entry = TransferListFindEntryByTag (TransferListHeader, TagId);
+  if (Entry == NULL) {
+    return EFI_NOT_FOUND;
+  }
+
+  *EntryOut = Entry;
```
samimujawar, line 586: what if EntryOut == NULL? This should be checked at line 580.

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -402,9 +381,311 @@ TransferListDump (
     }
 
     DEBUG ((DEBUG_INFO | DEBUG_LOAD, "Entry %d:\n", Idx++));
-    DEBUG ((DEBUG_INFO | DEBUG_LOAD, "tag_id     0x%x\n", Entry->TagId));
+    DEBUG ((DEBUG_INFO | DEBUG_LOAD, "TagId     0x%x\n", Entry->TagId));
```
pierregondois: NIT: no need to update that I think + it breaks the nice alignment

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -80,13 +84,13 @@ TransferListVerifyChecksum (
   IN TRANSFER_LIST_HEADER  *TransferListHeader
   )
 {
-  return transfer_list_verify_checksum ((struct transfer_list_header *)TransferListHeader);
```
pierregondois: Is this patch actually needed ? Why not doing these changes in:
- `ArmPkg/ArmTransferListLib: Additional TL wrappers and helpers`

or where these structures are first introduced in this patch serie ?

MdePkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -10,11 +10,27 @@
...
+typedef struct transfer_list_header  LIBTL_TRANSFER_LIST_HEADER;
+typedef struct transfer_list_entry   LIBTL_TRANSFER_LIST_ENTRY;
+typedef enum transfer_list_ops       LIBTL_TRANSFER_LIST_OPS;
+
+STATIC_ASSERT (
+  sizeof (TRANSFER_LIST_HEADER) == sizeof (LIBTL_TRANSFER_LIST_HEADER),
+  "TRANSFER_LIST_HEADER size mismatch"
+  );
+
+STATIC_ASSERT (
```
pierregondois, line 28: It would be nice to also do this for:
- `TRANSFER_LIST_EVENTLOG`
- `TRANSFER_ENTRY_HEADER`
- `TRANSFER_LIST_HEADER`

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -178,7 +179,10 @@ TransferListGetNextEntry (
   IN TRANSFER_ENTRY_HEADER  *CurrentEntry
   )
 {
-  return (TRANSFER_ENTRY_HEADER *)transfer_list_next ((struct transfer_list_header *)TransferListHeader, (struct transfer_list_entry *)CurrentEntry);
+  return (TRANSFER_ENTRY_HEADER *)transfer_list_next (
```
pierregondois: This should be in another patch no ?

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -83,7 +84,7 @@ TransferListVerifyChecksum (
 }
 
 /**
-  This helper maps Transfer List operation modes to corresponding libtl ops.
+  This helper maps corresponding libtl ops to Transfer List operation modes.
```
pierregondois: This should be in another patch no ?

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -377,3 +381,303 @@ TransferListDump (
...
+            DataSize
+            ))
+        )
+    {
+      return EFI_BUFFER_TOO_SMALL;
+    }
+  }
+
+  if ((Data != NULL) && (DataSize != 0)) {
+    VOID  *EntryData = transfer_list_entry_data ((struct transfer_list_entry *)Entry);
```
pierregondois: NIT: IIRC all the variables should be declared at the beginning of the function for the edk2 project (even though this definition is correct).

pierregondois (commented):
Hello Harish,
I added some small comments

.gitmodules
HarishManchikanti: Updated the license section in ReadMe.rst to include libtl, and moved that change into the same commit that adds the libtl submodule.

ReadMe.rst
HarishManchikanti: Agreed, moved  the ReadMe.rst license update into the same commit that adds the libtl submodule.

ArmPkg/Library/ArmTransferListLib/TransferListLibSupport.h
HarishManchikanti: Updated this header to use #pragma once.

MdePkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -17,13 +17,17 @@
 #include <Library/ArmTransferListLib.h>
 #include <Library/HobLib.h>
 
+typedef struct transfer_list_header  LIBTL_TRANSFER_LIST_HEADER;
```
HarishManchikanti: Thanks for the suggestion. These aliases are used only inside the wrapper implementation and are not part of the public API, so I kept them local to [ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c](https://github.com/tianocore/edk2/pull/12346/files/87cabc86d984e41608bf53c3438d94f1d4d415a5#diff-590d789cb73a55b2c1e93d63e72367a6b283b3e8645e169ec05be5b194b43ac9) to avoid exposing libtl internals through public headers.

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
HarishManchikanti: Done. Removed MapLibtlOps(), added STATIC_ASSERT checks in TransferListCheckHeader(), and cast the return value of transfer_list_check_header() to TRANSFER_LIST_OPS.

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
HarishManchikanti: Updated these comparisons to use TRANSFER_LIST_TAGID_MASK instead of the magic value, and applied the same change to the other relevant locations as well.

ArmPkg/Include/Library/ArmTransferListLib.h
HarishManchikanti: Updated this header to use #pragma once.

MdePkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -377,3 +381,303 @@ TransferListDump (
...
+  )
+{
+  TRANSFER_ENTRY_HEADER  *Entry;
+
+  Entry = TransferListFindEntryByTag (TransferListHeader, TagId);
+  if (Entry == NULL) {
+    return EFI_NOT_FOUND;
+  }
+
+  *EntryOut = Entry;
```
HarishManchikanti: Added a NULL check for EntryOut and return EFI_INVALID_PARAMETER when it is not provided.

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
HarishManchikanti: Kept the original tag_id debug label .

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
HarishManchikanti: This was moved into the dedicated ECC cleanup patch, together with the local libtl type aliases used by these casts.

MdePkg/Library/ArmTransferListLib/ArmTransferListLib.c
HarishManchikanti: Added the missing compile-time checks. For TRANSFER_LIST_EVENTLOG, I used a layout check on EventLog offset rather than a sizeof check, since it has a flexible array member.

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
HarishManchikanti: Yes, this was moved into the dedicated ECC cleanup patch.

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
HarishManchikanti: Done. Moved EntryData to the beginning of the function to match edk2 style.

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
HarishManchikanti: Yes, this was moved into the appropriate patch.

MdePkg/Library/ArmTransferListLib/ArmTransferListLib.c
pierregondois: Could they be placed in `ArmPkg/Library/ArmTransferListLib/TransferListLibSupport.h` instead, so they remain local to the module ?

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
pierregondois: What I don't really understand is that it seems that `ArmPkg/ArmTransferListLib: Fix ECC issues in wrapper code` could be merged into `ArmPkg/ArmTransferListLib: Replace ArmTransferListLib content with libtl` no ?

I.e. :
```
typedef struct transfer_list_header  LIBTL_TRANSFER_LIST_HEADER;
typedef struct transfer_list_entry   LIBTL_TRANSFER_LIST_ENTRY;
```
was introduced earlier in the serie, there would not need to fix the Ecc errors no ?

MdePkg/Include/IndustryStandard/ArmTransferList.h
```diff
@@ -128,19 +144,18 @@ typedef struct TransferListHeader {
...
  * Transfer entry in transfer list starts with the following header.
  */
+#pragma pack(1)
 typedef struct TransferEntryHeader {
   /// The entry type identifier.
-  UINT16    TagId;
-
-  /// Reserved.
-  UINT8     Reserved0;
+  UINT32    TagId      : 24;
```
pierregondois, line 149: I thought it was not ideal to rely on bitfields for interfaces.
It seems the spec describes that [1] the first 3 bytes of the struct are for the tag_id, but I m not sure this is guaranteed by bit-fields.
This seems the issue originates from libtl. I guess we can update it later.

[1] https://firmwarehandoff.github.io/firmware_handoff/main/transfer_list.html

pierregondois: @LeviYeoReum @samimujawar 
As you suggested, it might actually be better to define this as:
```
UINT8    TagId[3];
UINT8    HeaderSize[1];
```
at least in edk2 ?

HarishManchikanti: Good point, agreed. Relying on bitfields in a public interface is not ideal.
At the moment we are constrained by the current libtl interface (no dedicated accessors for TagId/HeaderSize), so this series keeps compatibility with it.
I agree this should be improved in a follow-up once libtl exposes proper wrappers/accessors.

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
HarishManchikanti: Agree, it can be folded into the replace-libtl patch. I kept it separate for review clarity, but I can squash if preferred.

pierregondois: Yes ideally if it is possible to squash the `ArmPkg/ArmTransferListLib: Fix ECC issues in wrapper code` into the other commits

MdePkg/Include/IndustryStandard/ArmTransferList.h
pierregondois: ok right, I guess we can update it once the libtl is updated

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
pierregondois: Hello @HarishManchikanti,
I think the ECC patch is still present

samimujawar:
>ArmPkg/ArmTransferListLib: Fix ECC issues in wrapper code could be merged into ArmPkg/ArmTransferListLib: Replace ArmTransferListLib content with libtl 

@HarishManchikanti can you address this, please?

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.c
HarishManchikanti: fixed now

HarishManchikanti:
> > ArmPkg/ArmTransferListLib: Fix ECC issues in wrapper code could be merged into ArmPkg/ArmTransferListLib: Replace ArmTransferListLib content with libtl
> 
> @HarishManchikanti can you address this, please?

Done

samimujawar (approved):

ArmPkg/Include/IndustryStandard/ArmTransferList.h
```diff
@@ -128,19 +144,18 @@ typedef struct TransferListHeader {
 /*
  * Transfer entry in transfer list starts with the following header.
  */
+#pragma pack(1)
```
leiflindholm: From https://github.com/FirmwareHandoff/firmware_handoff/blob/main/source/transfer_list.rst:
"The TL header and all the TEs are 8-byte aligned"

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.inf
```diff
@@ -16,8 +16,15 @@
   VERSION_STRING                 = 1.0
   LIBRARY_CLASS                  = ArmTransferListLib
 
+  DEFINE TL_LIB_PATH             = libtl
+  DEFINE TL_LIB_INCLUDE_PATH     = $(TL_LIB_PATH)/include
```
leiflindholm: I'm not sure I feel like these defines help more than they confuse.

leiflindholm (changes requested):
Two minor comments, but here's a bigger one:
ArmPkg should be going away, as per https://github.com/tianocore/edk2/issues/10289.
If we consider this "industry standard", it should be going into MdePkg.

HarishManchikanti:
> Two minor comments, but here's a bigger one: ArmPkg should be going away, as per #10289. If we consider this "industry standard", it should be going into MdePkg.

Thanks @leiflindholm . Before I respin this, I’d like to better understand the intended direction here.

Are you suggesting that only the public Transfer List definitions belong under MdePkg/Include/IndustryStandard, or that the library class/implementation and related GUIDs should also be placed outside ArmPkg?

Looping in @samimujawar  for input as well.

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.inf
HarishManchikanti: The defines were added to keep the libtl root/include paths centralized and avoid repeating the same prefix in multiple source entries. The intention was to improve readability and make the INF easier to adjust if the embedded libtl location changes.

samimujawar:
> > Two minor comments, but here's a bigger one: ArmPkg should be going away, as per #10289. If we consider this "industry standard", it should be going into MdePkg.
> 
> Thanks @leiflindholm . Before I respin this, I’d like to better understand the intended direction here.
> 
> Are you suggesting that only the public Transfer List definitions belong under MdePkg/Include/IndustryStandard, or that the library class/implementation and related GUIDs should also be placed outside ArmPkg?
> 
> Looping in @samimujawar for input as well.

I agree with @leiflindholm we should just move this to MdePkg, this includes both the headers and the library.

ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.inf
leiflindholm: I understand that, but ... why would it? And if it would, why would we expect the include directory to move relative to that?

HarishManchikanti: Thanks. I removed those local path defines and used the paths directly in the INF to make it clearer.

ArmPkg/Include/IndustryStandard/ArmTransferList.h
HarishManchikanti: Addressed in latest change set .

HarishManchikanti:
> > > Two minor comments, but here's a bigger one: ArmPkg should be going away, as per #10289. If we consider this "industry standard", it should be going into MdePkg.
> > 
> > 
> > Thanks @leiflindholm . Before I respin this, I’d like to better understand the intended direction here.
> > Are you suggesting that only the public Transfer List definitions belong under MdePkg/Include/IndustryStandard, or that the library class/implementation and related GUIDs should also be placed outside ArmPkg?
> > Looping in @samimujawar for input as well.
> 
> I agree with @leiflindholm we should just move this to MdePkg, this includes both the headers and the library.

@leiflindholm Thanks for the review comments.

I have addressed the feedback and updated the series. Could you please take another look when you get a chance?

pierregondois:
Hello Harish,
2 last comments for me:
A.
Is it possible to split `ArmPkg: Consume ArmTransferListLib from MdePkg ` in 2:
- Changing the source of `ArmTransferListLib` to the MdePkg implementation
- Removing the `ArmPkg` implementation

and move the ArmPkg implementation removal after the `StandaloneMmPkg` patch, otherwise there are a few commits where `StandaloneMmPkg` and `ArmVirtPkg` don't build
```
StandaloneMmPkg: Consume ArmTransferListLib from MdePkg
ArmVirtPkg: Consume ArmTransferListLib from MdePkg
ArmPkg: Consume ArmTransferListLib from MdePkg      // ArmVirtPkg and StandaloneMmPkg won't build here
ArmPlatformPkg: Consume ArmTransferListLib from MdePkg
```


B.
Similarly, the `edk2-platfoms` repo relies on the ArmPkg implementation. Some platforms needs to change to the `MdePkg` one. It would be nice to have a PR ready for that

HarishManchikanti:
> Hello Harish, 2 last comments for me: A. Is it possible to split `ArmPkg: Consume ArmTransferListLib from MdePkg ` in 2:
> 
> * Changing the source of `ArmTransferListLib` to the MdePkg implementation
> * Removing the `ArmPkg` implementation
> 
> and move the ArmPkg implementation removal after the `StandaloneMmPkg` patch, otherwise there are a few commits where `StandaloneMmPkg` and `ArmVirtPkg` don't build
> 
> ```
> StandaloneMmPkg: Consume ArmTransferListLib from MdePkg
> ArmVirtPkg: Consume ArmTransferListLib from MdePkg
> ArmPkg: Consume ArmTransferListLib from MdePkg      // ArmVirtPkg and StandaloneMmPkg won't build here
> ArmPlatformPkg: Consume ArmTransferListLib from MdePkg
> ```
> 
> B. Similarly, the `edk2-platfoms` repo relies on the ArmPkg implementation. Some platforms needs to change to the `MdePkg` one. It would be nice to have a PR ready for that

Thanks @pierregondois.

I split the ArmPkg change into two commits:
- ArmPkg: Consume ArmTransferListLib from MdePkg
- ArmPkg: Remove ArmTransferListLib implementation

The removal commit is now placed after the edk2 consumers have moved to the MdePkg implementation.

I also prepared an edk2-platforms related PR to update platform DSC files to consume ArmTransferListLib from MdePkg:
https://github.com/tianocore/edk2-platforms/pull/1010

pierregondois (approved):

.gitmodules
```diff
@@ -38,3 +38,6 @@
 [submodule "TcgTpmPkg/Library/TpmLib/TPM"]
 	path = TcgTpmPkg/Library/TpmLib/TPM
 	url = https://github.com/TrustedComputingGroup/TPM
+[submodule "MdePkg/Library/ArmTransferListLib/libtl"]
```
kuqin12, line 41: Is it right to use the "review" instance of the link? Also, if we are going to add the submodule, is there a github mirror we can pull in for the sake of more stable reachability?

kuqin12: i.e. is this repo equivalently supported? https://github.com/TF-Shared/transfer-list-library/commits/main/

MdePkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -0,0 +1,703 @@
...
+EFIAPI
+TransferListGetFromHobList (
+  OUT TRANSFER_LIST_HEADER  **TransferList
+  )
+{
+  VOID               *HobList;
+  EFI_HOB_GUID_TYPE  *GuidHob;
+  UINTN              *GuidHobData;
+
+  *TransferList = NULL;
```
kuqin12, line 107: null pointer check?

MdePkg/Library/ArmTransferListLib/ArmTransferListLib.c
```diff
@@ -0,0 +1,703 @@
...
+TRANSFER_LIST_OPS
+EFIAPI
+TransferListCheckHeader (
+  IN TRANSFER_LIST_HEADER  *TransferListHeader
+  )
+{
+  STATIC_ASSERT ((UINT32)TL_OPS_NON == (UINT32)TRANSFER_LIST_OPS_INVALID, "TL_OPS_NON mismatch");
+  STATIC_ASSERT ((UINT32)TL_OPS_ALL == (UINT32)TRANSFER_LIST_OPS_ALL, "TL_OPS_ALL mismatch");
+  STATIC_ASSERT ((UINT32)TL_OPS_RO  == (UINT32)TRANSFER_LIST_OPS_RO, "TL_OPS_RO mismatch");
+  STATIC_ASSERT ((UINT32)TL_OPS_CUS == (UINT32)TRANSFER_LIST_OPS_CUSTOM, "TL_OPS_CUS mismatch");
```
kuqin12: Any specific reason why these static_assert are here?

.gitmodules
HarishManchikanti: @kuqin12  Thanks for checking.

The `review.trustedfirmware.org` URL is intentional here because it is the canonical TrustedFirmware review repository for libtl. The GitHub repository appears to be a read-only mirror, useful for visibility/cloning, but not equivalent for submitting or reviewing libtl changes.

Keeping the submodule pointed at the review URL keeps it aligned with the upstream TrustedFirmware contribution flow. TF-A also uses the same URL for its libtl submodule https://git.trustedfirmware.org/plugins/gitiles/TF-A/trusted-firmware-a.git/+/refs/heads/master/.gitmodules .

kuqin12: I understand the semantics behind using review.trustedfirmware.org in TFA repo. But MdePkg will likely be used for almost any PRs in this repo. My concern is that the traffic will get throttled if we pick up the checkouts from this repo on trustedfirmware.org, during repetitive CI builds.

We had to do something like this before to make the pipeline more stable: https://github.com/microsoft/mu_tiano_platforms/blob/main/.azurepipelines/Platform-Build-Job.yml#L91

HarishManchikanti: @kuqin12  Thanks, agreed.
I kept .gitmodules pointing to the TrustedFirmware review repo, and added CI-only redirects so Azure/GitHub CI fetch libtl from the GitHub mirror.

MdePkg/Library/ArmTransferListLib/ArmTransferListLib.c
HarishManchikanti: Fixed

MdePkg/Library/ArmTransferListLib/ArmTransferListLib.c
HarishManchikanti: Thanks, I moved these to file scope and added a comment.

kuqin12 (approved):

samimujawar:
@leiflindholm, @ardbiesheuvel if there are no further comments, can we merge this PR, please?

pierregondois:
@leiflindholm as you made some comments (and @ardbiesheuvel in case): gentle ping for the PR

pierregondois:
@leiflindholm please let us know if some additional changes are needed.
Otherwise it would be nice if the PR could make it to the stable tag

pierregondois:
Hi @leiflindholm ,
is it ok to merge the PR since the library was merged to the MdePkg ?
(The PR would need to be rebased first)

samimujawar:
@leiflindholm can you please provide your feedback for this PR, please?

samimujawar:
@leiflindholm apparently the merge for this PR is blocked as you had requested changes. I believe your concerns have been addressed, can you kindly help to progress this PR, please?

PR #12349, opened 2026-03-26
UefiCpuPkg/MpInitLib: Skip ucode rev dump if not loaded in the library

philnoh2:
MicrocodeDetect() loads microcode in the first thread of each core and populates MicrocodeRevision and MicrocodeEntryAddr fields. AMD CPUs use a separate function to load microcode, leaving MicrocodePatchRegionSize as zero. In the case, the function returns early without populating these fields, leaving them as zero. As a result, the misleading debug output (Microcode revision = 00000000, expected = 00000000) may appear on AMD systems, implying microcode was not applied when it actually was. To avoid the confusion, skip the microcode revision dump when it is not loaded in the library.

e.g.
CpuMpPei.efi : 
MpInitLib: Find 20 processors in system.
GetMicrocodePatchInfoFromHob: Microcode patch cache HOB is not found.
CPU[0000]: Microcode revision = 00000000, expected = 00000000
CPU[0002]: Microcode revision = 00000000, expected = 00000000
CPU[0004]: Microcode revision = 00000000, expected = 00000000
CPU[0006]: Microcode revision = 00000000, expected = 00000000
CPU[0008]: Microcode revision = 00000000, expected = 00000000
CPU[0010]: Microcode revision = 00000000, expected = 00000000
CPU[0012]: Microcode revision = 00000000, expected = 00000000
CPU[0014]: Microcode revision = 00000000, expected = 00000000
CPU[0016]: Microcode revision = 00000000, expected = 00000000
CPU[0018]: Microcode revision = 00000000, expected = 00000000

CpuDxe.efi : 
GetMicrocodePatchInfoFromHob: **MicrocodeBase = 0x0, MicrocodeSize = 0x0**
WakeupBufferStart = 87000, WakeupBufferSize = E5
AP Vector: 16-bit = 87000/39, ExchangeInfo = 87040/A5
CPU[0000]: Microcode revision = 00000000, expected = 00000000
CPU[0002]: Microcode revision = 00000000, expected = 00000000
CPU[0004]: Microcode revision = 00000000, expected = 00000000
CPU[0006]: Microcode revision = 00000000, expected = 00000000
CPU[0008]: Microcode revision = 00000000, expected = 00000000
CPU[0010]: Microcode revision = 00000000, expected = 00000000
CPU[0012]: Microcode revision = 00000000, expected = 00000000
CPU[0014]: Microcode revision = 00000000, expected = 00000000
CPU[0016]: Microcode revision = 00000000, expected = 00000000
CPU[0018]: Microcode revision = 00000000, expected = 00000000

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

The update has been tested and verified on AMD platforms.

## Integration Instructions

N/A

### DIFF
--- a/.azurepipelines/Windows-CLANGPDB.yml
+++ b/.azurepipelines/Windows-CLANGPDB.yml
@@ -1,0 +1,35 @@
+## @file
+# Azure Pipeline build file for a build using Windows and CLANGPDB
+#
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+trigger: none
+
+schedules:
+- cron: "0 7 * * *"
+  displayName: Daily Build
+  branches:
+    include:
+    - master
+    - stable/*
+  always: true
+
+pr:
+- master
+- stable/*
+
+variables:
+  - template: templates/defaults.yml
+
+jobs:
+- template: templates/pr-gate-build-job.yml
+  parameters:
+    tool_chain_tag: 'CLANGPDB'
+    vm_image: ${{ variables.default_windows_vm }}
+    arch_list: "X64,AARCH64"
+    usePythonVersion: ${{ variables.default_python_version }}
+    extra_install_step:
+    - powershell: choco install opencppcoverage; Write-Host "##vso[task.prependpath]C:\Program Files\OpenCppCoverage"
+      displayName: Install Code Coverage Tool
+      condition: and(gt(variables.pkg_count, 0), succeeded())
--- a/BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
+++ b/BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
@@ -182,8 +182,13 @@ class HostBasedUnitTestRunner(IUefiBuildPlugin):
         logging.info(f"Got lcov version {lcov_version_major}")
 
         # Generate base code coverage for all source files
-        # `--ignore-errors mismatch` needed to make lcov v2.0+/gcov work.
-        lcov_error_settings = "--ignore-errors mismatch" if lcov_version_major >= 2 else ""
+        # lcov v2.0+ with gcov requires ignoring several non-fatal error
+        # classes that commonly occur with large vendored third-party trees:
+        #   mismatch - gcov/gcno version or function line mismatches
+        #   source   - source files not found or newer than gcno notes
+        #   format   - unexpected gcov output (e.g. line numbers beyond EOF)
+        #   gcov     - gcov tool returned with a non-zero return code
+        lcov_error_settings = "--ignore-errors mismatch,source,format,gcov" if lcov_version_major >= 2 else ""
         ret = RunCmd("lcov", f"--no-external --capture --initial --directory {buildOutputBase} --output-file {buildOutputBase}/cov-base.info --rc lcov_branch_coverage=1 {lcov_error_settings}")
         if ret != 0:
             logging.error("UnitTest Coverage: Failed to build initial coverage data.")
--- a/Maintainers.txt
+++ b/Maintainers.txt
@@ -242,7 +242,6 @@ R: Wei6 Xu <wei6.xu@intel.com> [xuweiintel]
 IntelFsp2Pkg
 F: IntelFsp2Pkg/
 W: https://www.tianocore.org/tianocore-wiki.github.io/platforms-packages/core-packages/intel_fsp2_pkg.html
-M: Chasel Chiu <chasel.chiu@intel.com> [ChaselChiu]
 M: Nate DeSimone <nathaniel.l.desimone@intel.com> [nate-desimone]
 M: Ray Ni <ray.ni@intel.com> [niruiyu]
 R: Star Zeng <star.zeng@intel.com> [lzeng14]
@@ -252,7 +251,6 @@ R: Ashraf Ali S <ashraf.ali.s@intel.com> [AshrafAliS]
 IntelFsp2WrapperPkg
 F: IntelFsp2WrapperPkg/
 W: https://www.tianocore.org/tianocore-wiki.github.io/platforms-packages/core-packages/intel_fsp2_wrapper_pkg.html
-M: Chasel Chiu <chasel.chiu@intel.com> [ChaselChiu]
 M: Nate DeSimone <nathaniel.l.desimone@intel.com> [nate-desimone]
 R: Star Zeng <star.zeng@intel.com> [lzeng14]
 R: Ted Kuo <ted.kuo@intel.com> [tedkuo1]
@@ -449,7 +447,6 @@ F: MdePkg/Include/Library/FdtLib.h
 M: Benny Lin <benny.lin@intel.com> [Benny3345678]
 M: Leif Lindholm <leif.lindholm@oss.qualcomm.com> [leiflindholm]
 R: Gua Guo <gua.guo@intel.com> [gguo11837463]
-R: Chasel Chiu <chasel.chiu@intel.com> [ChaselChiu]
 R: James Lu <james.lu@intel.com> [jameslu8]
 
 MdePkg: Manageability industryStandard standard C header files
@@ -702,7 +699,6 @@ M: Sean Rhodes <sean@starlabs.systems> [Sean-StarLabs]
 M: James Lu <james.lu@intel.com> [jameslu8]
 M: Benjamin Doron <benjamin.doron@9elements.com> [benjamindoron]
 R: Gua Guo <gua.guo@intel.com> [gguo11837463]
-M: Chasel Chiu <chasel.chiu@intel.com> [ChaselChiu]
 R: Shuo Liu <shuo.liu@intel.com> [shuoliu0]
 S: Maintained
 
--- a/NetworkPkg/Dhcp4Dxe/Dhcp4Impl.c
+++ b/NetworkPkg/Dhcp4Dxe/Dhcp4Impl.c
@@ -1087,6 +1087,7 @@ EfiDhcp4Stop (
 
   DhcpSb->DhcpState    = Dhcp4Stopped;
   DhcpSb->ServiceState = DHCP_UNCONFIGED;
+  DhcpSb->IoStatus     = EFI_NOT_STARTED;
 
   gBS->RestoreTPL (OldTpl);
   return EFI_SUCCESS;
--- a/NetworkPkg/UefiPxeBcDxe/PxeBcMtftp.c
+++ b/NetworkPkg/UefiPxeBcDxe/PxeBcMtftp.c
@@ -391,6 +391,7 @@ PxeBcMtftp6WriteFile (
   Token.CheckPacket     = PxeBcMtftp6CheckPacket;
   Token.TimeoutCallback = NULL;
   Token.PacketNeeded    = NULL;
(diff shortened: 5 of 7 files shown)


UefiCpuPkg/Library/MpInitLib/MpLib.c
```diff
@@ -2479,7 +2479,12 @@ MpInitLibInitialize (
       //
       // MicrocodeDetect() loads microcode in first thread of each core, so,
       // CpuMpData->CpuData[Index].MicrocodeEntryAddr is initialized only for first thread of each core.
+      // Skip the revision dump when both the revision and entry address are zero.
       //
+      if ((CpuMpData->CpuData[Index].MicrocodeRevision == 0) && (CpuMpData->CpuData[Index].MicrocodeEntryAddr == 0)) {
+        continue;
+      }
+
```
niruiyu: The side effect of this change is that when no proper uCode is found and CPU is not applied with the microcode, no debug message at all.

CpuData[Index].MicrocodeRevision is read from MSR 8Bh.
Do you mean that such MSR does not return the microcode revision in AMD CPU? But https://github.com/torvalds/linux/blob/46b513250491a7bfc97d98791dbe6a10bcc8129d/arch/x86/kernel/cpu/microcode/amd.c#L340 shows the AMD CPU also returns the revision in the same MSR.

Can you explain more?

philnoh2: Thank you for your review. Like the log example in the description, there is no microcode patch version to be displayed on AMD platforms (e.g. MicrocodeBase = 0x0, MicrocodeSize = 0x0). 

In MicrocodeDetect() function, if CpuMpData->MicrocodePatchRegionSize is zero, it just returns as there are no microcode patches. In this case, there is no chance to update CpuMpData->CpuData[ProcessorNumber].MicrocodeRevision through GetProcessorMicrocodeSignature() that accesses the MSR (MSR_IA32_BIOS_SIGN_ID  0x0000008B). Basically the MSR is available for AMD CPUs. But, when we validated GetProcessorMicrocodeSignature(), we found that there are a couple of issues below.

(1) "AsmWriteMsr64 (MSR_IA32_BIOS_SIGN_ID, 0)" in the function causes a CPU exception. The code is not compatible with AMD CPUs.
(2) "AsmReadMsr64 (MSR_IA32_BIOS_SIGN_ID)" works on AMD CPUs. But the function doesn't return the right data due to a difference of MSR_IA32_BIOS_SIGN_ID_REGISTER structure (AMD - MicrocodeUpdateSignature : Bit 0 ~ 31, not 32 ~ 63).

Based on that, it could be possible for the related drivers (CpuMpPei.int and CpuDxe.inf) to show the zero-based debug message in case there is no microcode patch, regardless of the loading status. If a system has 256 cores, there will be 128 lines of debug messages of 0 from both PEI and DXE drivers, total of 256 lines. 

If we don't see the debug message on the update, we can assume that there are no microcode patches to load, as "GetMicrocodePatchInfoFromHob: MicrocodeBase = 0x0, MicrocodeSize = 0x0" already indicates the case of CPU[0000]: Microcode revision = 00000000, expected = 00000000...

niruiyu: > In MicrocodeDetect() function, if CpuMpData->MicrocodePatchRegionSize is zero, it just returns as there are no microcode patches. In this case, there is no chance to update CpuMpData->CpuData[ProcessorNumber].MicrocodeRevision through GetProcessorMicrocodeSignature() that accesses the MSR (MSR_IA32_BIOS_SIGN_ID 0x0000008B). 

I agree with you. CpuMpData->CpuData[ProcessorNumber].MicrocodeRevision update is skipped when CpuMpData->MicrocodePatchRegionSize == 0, which might be the case you saw in AMD platforms.
But I think it's still valuable to print the uCode revision, which could be updated from reset or by other code before MpInitLib.
How about reading the uCode revision when CpuMpData->MicrocodePatchRegionSize == 0?


> Basically the MSR is available for AMD CPUs. But, when we validated GetProcessorMicrocodeSignature(), we found that there are a couple of issues below.
> 
> (1) "AsmWriteMsr64 (MSR_IA32_BIOS_SIGN_ID, 0)" in the function causes a CPU exception. The code is not compatible with AMD CPUs.

Intel SDM recommends to preload this MSR with zero, then execute CPUID, then read this MSR to get the revision. Maybe AMD CPU does not require so. 

> (2) "AsmReadMsr64 (MSR_IA32_BIOS_SIGN_ID)" works on AMD CPUs. But the function doesn't return the right data due to a difference of MSR_IA32_BIOS_SIGN_ID_REGISTER structure (AMD - MicrocodeUpdateSignature : Bit 0 ~ 31, not 32 ~ 63).
> 
> Based on that, it could be possible for the related drivers (CpuMpPei.int and CpuDxe.inf) to show the zero-based debug message in case there is no microcode patch, regardless of the loading status. If a system has 256 cores, there will be 128 lines of debug messages of 0 from both PEI and DXE drivers, total of 256 lines.

OK then there are two diffs between Intel and AMD processors: 1). 8Bh MSR cannot be written; 2). low-32bits of 8Bh returns revision.
So can you use if-AMD check to run a different code path? As I think it's valuable to show the uCode revision.

> 
> If we don't see the debug message on the update, we can assume that there are no microcode patches to load, as "GetMicrocodePatchInfoFromHob: MicrocodeBase = 0x0, MicrocodeSize = 0x0" already indicates the case of CPU[0000]: Microcode revision = 00000000, expected = 00000000...

I agree the "expected" word is confusing. If the processor already contains a revision "1" uCode and platform BIOS does not supply the uCode patch file, it's wrong to "expect" the uCode revision to be 0.
How about "Microcode revision = 00000001 (in cpu) / 00000000 (in patch)"?

philnoh2: Thank you for your additional review and suggestion. Let me comment.

> So can you use if-AMD check to run a different code path? As I think it's valuable to show the uCode revision.

I understand that you suggest updating GetProcessorMicrocodeSignature function in MicrocodeLib library. If then, we thought that updating the library for AMD CPUs would not be proper because we found that it is dedicated to Intel CPUs. Adding some AMD code to the library would cause a confusion that it is compatible with AMD CPUs.

> How about reading the uCode revision when CpuMpData->MicrocodePatchRegionSize == 0?

I think that generally it is reasonable. But, due to the function/library limitation, the update couldn't be applicable to AMD CPUs.

> How about "Microcode revision = 00000001 (in cpu) / 00000000 (in patch)"?

I think that unmatching microcode revision in CPU and in patch would still give more feel of failing. I understand that the dump message (e.g. CPU[0000]: Microcode revision = AAAA_AAAA, expected = AAAA_AAAA) is to ensure that the microcode loaded matches the microcode expected to load. Keeping it would be better if not all zero.

---
Currently AMD-based Microcodelib library is not available. AMD SOC module supports loading microcode and displaying the information. Like MpInitLib library, we prefer to display the information in the same place where microcode is loaded. AMD SOC module does it. 

Based on that, the PR is intended for suppressing the information in the specific case (all 0). We think that it might not affect other CPUs (e.g. Intel) or next updates when considering the current code flow. Therefore, it focuses on the case of Microcode revision = 0, expected = 0 in which microcode is not loaded from MP init and does not exclude further update for AMD SOC to display its microcode revision here in a separate PR.

niruiyu: I understand your changes can meet your goal. But it hurts the readability of the code logic. For example, why the debug message should be suppressed when there is no patch file and CPU internal revision is 0? Should this a bug as CPU internal revision is not 0? (The 0 is because the code to read the revision is skipped.)
From my perspective, the change is to leverage an existing bug and add another bug(or confusing changes).

My past experience told me that if the code logic is based on a bug to work as expected, it is not a stable logic and will break again sooner or later.

If you think MicrocodeLib is not suitable for AMD, why not skip calling MicrocodeDetect() for AMD CPUs? Since you skip the call, it makes sense to skip the debug messages for AMD CPUs as well. I agree the change is not that simple and introduces some if-checks. But it gives the consistent meaning to future code maintainers.

philnoh2: I don't think that there is a bug in the existing code. I agree that using CpuMpData->CpuData[ProcessorNumber].MicrocodeRevision to skip message dump is confusing. CpuMpData->CpuData[ProcessorNumber].MicrocodeRevision is zero if there is no microcode patch file available to be loaded regardless of what CPU internal revision is. This existing logic has no problem.
 
This PR is focused on improving debug message for the case there is no microcode patch file to be loaded in MPinit regardless of Intel or AMD CPU. In this case, CpuMpData->CpuData[Index].MicrocodeRevision = 0 and ExpectedMicrocodeRevision = 0 for all cores by existing logic. Would replacing up to hundred of lines of "Microcode revision = 00000000, expected = 00000000" with single line of "No Microcode patch file found" be better ?
 
I revised the PR to use CpuMpData->MicrocodePatchRegionSize to check microcode patch file existence, same as the one at beginning of MicrocodeDetect(). Please help to review it. Thank you.

niruiyu: The bug is that when there is no patch file, MpInitLib prints the cpu revision as 0.
I am ok with your current patch. I can fix that bug later.

abdattar (approved):

UefiCpuPkg/Library/MpInitLib/MpLib.c
philnoh2: @niruiyu :

Thank you for your approval.
 
In considering your future fix for the bug of "cpu revision as 0", I would like to present architectural issue around microcode in MPInit.
 
Dependency chain: 
UefiCpuPkg\Library\MpInitLib\MpLib.c->UefiCpuPkg\Library\MpInitLib\Microcode.c->UefiCpuPkg\Library\MicrocodeLib\MicrocodeLib.inf.
Microcode.c is part of common code but it is 100% Intel specific and there is no NULL lib for MicrocodeLib.inf. Luckily, "if (CpuMpData->MicrocodePatchRegionSize == 0)->return" at beginning of MicrocodeDetect() helps AMD CPU to skip the everything in Microcode.c->UefiCpuPkg\Library\MicrocodeLib\MicrocodeLib.inf. However, your future fix for "cpu revision as 0" to let MicrocodeDetect() to call GetProcessorMicrocodeSignature() in MicrocodeLib would kill AMD CPU. AsmWriteMsr64 (MSR_IA32_BIOS_SIGN_ID, 0) in the function causes AMD CPU exception.
 
It is our idea/proposal to address the architectural issue.
1. Move Microcode.c into MicrocodeLib.inf where it belongs.
2. Adjust UefiCpuPkg\Include\Library\MicrocodeLib.h to define 4 public routines that are called from common MpLib.c: GetMicrocodePatchInfoFromHob, ShadowMicrocodeUpdatePatch, MicrocodeDetect and new one DumpMicrocodeRev() to move "Dump the microcode revision for each core" into the LIB so that Intel and AMD can choose independent way to dump Microcode rev. Further more CPU_MICROCODE_HEADER referenced the dump code is Intel specific.
3. Create NULL LIB or AMD version of the LIB for MicrocodeLib class.
 
I understand that StandardSignatureIsAuthenticAMD() can be used in existing Microcode.c/MicrocodeLib.c to make things easier but it is not architecturally solid. I will be happy to work with you to come out solid solution good for both Intel and AMD. Let me know your thoughts.

PR #12350, opened 2026-03-27
BaseTools: Only check for GCC prefixes when targeting

apop5:
# Description

LinuxGccToolChain is checking for the environment variable GCC_AARCH64_PREFIX when GCC_AARCH64_INSTALL is set in the environment variables. GCC_AARCH64_INSTALL is set  when any gcc aarch64 compiler is installed (i.e. aarch64-none-elf, aarch64-linux-gnu, aarch64-unknown-elf all result in a GCC_AARCH64_INSTALL environment variable).

When compiling for an X86 target, if an AARCH64 tool chain is installed in the system, this will result in an error due to the GCC_AARCH64_PREFIX not being set.

Add a check based upon TARGET_ARCH and and only verify the prefixes when attempting to build AARCH64.

Replicate the same check for RISCV and LOONGARCH64 architectures as well.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Ubuntu image compiling for IA32,X64 arch. 
Installed AARCH64 gcc tool chain (gcc-12-aarch64-linux-gnu)
Attempting to build resulted in plugin failure.
After modifications, build was able to complete for IA32,X64.

## Integration Instructions
No integration necessary.

### DIFF
--- a/ArmVirtPkg/KvmtoolCfgMgrDxe/ConfigurationManager.c
+++ b/ArmVirtPkg/KvmtoolCfgMgrDxe/ConfigurationManager.c
@@ -1,7 +1,7 @@
 /** @file
   Configuration Manager Dxe
 
-  Copyright (c) 2021 - 2022, Arm Limited. All rights reserved.<BR>
+  Copyright (c) 2021 - 2024, Arm Limited. All rights reserved.<BR>
 
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
@@ -716,6 +716,22 @@ GetStandardNameSpaceObject (
           // IORT is only required for GicV3/4
           //
           AcpiTableCount -= 1;
+        } else {
+          //
+          // Check if we have support for ITS.
+          //
+          Status = DynamicPlatRepoGetObject (
+                     PlatformRepo->DynamicPlatformRepo,
+                     CREATE_CM_ARM_OBJECT_ID (EArmObjGicItsInfo),
+                     CM_NULL_TOKEN,
+                     &CmObjDesc
+                     );
+          if (EFI_ERROR (Status)) {
+            //
+            // IORT is only required for GicV3/4 if ITS is present.
+            //
+            AcpiTableCount -= 1;
+          }
         }
       }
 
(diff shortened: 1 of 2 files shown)


apop5:
@mdkinney 

Would you please look this over? If you are okay with it, could you add the push tag?

BaseTools/Plugin/LinuxGccToolChain/LinuxGccToolChain.py
```diff
@@ -28,22 +28,25 @@ def do_pre_build(self, thebuilder):
         #
         if thebuilder.env.GetValue("TOOL_CHAIN_TAG") == "GCC":
             # Start with AARACH64 compiler
```
makubacki: nit: Not something that you introduced and non-functional, but I wanted to point it out in case you care to fix it.

```suggestion
            # Start with AARCH64 compiler
```

BaseTools/Plugin/LinuxGccToolChain/LinuxGccToolChain.py
```diff
@@ -28,22 +28,25 @@ def do_pre_build(self, thebuilder):
         #
         if thebuilder.env.GetValue("TOOL_CHAIN_TAG") == "GCC":
             # Start with AARACH64 compiler
-            ret = self._check_aarch64()
-            if ret != 0:
-                self.Logger.critical("Failed in check aarch64")
-                return ret
+            if "AARCH64" in thebuilder.env.GetValue("TARGET_ARCH", "AARCH64"):
```
makubacki, line 31: Are these intentionally falling back to a default value of the architecture? That would cause a match in each condition and effectively go back to the previous behavior.

apop5: Yes, it was intentional to default back to the previous behavior.

TARGET_ARCH _should_ be set. If it is not, then it would return None, which could cause a exception. Defaulting to the existing behavior seemed desirable in that case.

BaseTools/Plugin/LinuxGccToolChain/LinuxGccToolChain.py
apop5: fixed in latest push.

BaseTools/Plugin/LinuxGccToolChain/LinuxGccToolChain.py
makubacki: I think that's fine with the other option being to enforce the assumption that TARGET_ARCH is set.

kuqin12:
@mergify queue

PR #4906, opened 2023-10-11
RedfishPkg: Use base version SortLib for the specific modules

changab:
BZ #: 4566

Update Redfish modules to use the small footprint
version of base SortLib by the means of module scoped subelement <LibraryClass>. With this the platform
level SortLib (full version) is not impacted if
Redfish.dsc.inc is included in platform DSC.

Cc: Nickle Wang <nicklew@nvidia.com>
Cc: Igor Kulchytskyy <igork@ami.com>
Cc: Nhi Pham <nhi@os.amperecomputing.com>

PR #12355, opened 2026-03-28
Enable cast-function-type-mismatch warning on CLANG toolchains

mikebeaton:
# Description

This PR continues the approach of trying to enable warnings which can be enabled, in order to catch more real errors.

Based on the fixes needed for EDK 2 to compile with the warning enabled, it is believed that at least some of these are genuine issues which we would like to catch automatically.

This warning is enabled by default on recent versions of Apple Clang, but not LLVM Clang.

Add commits which fix reliance on undefined behaviour caught by this warning, then enable for CLANG toolchains.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

CI

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
+++ b/MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
@@ -35,6 +35,69 @@ EFI_SMM_BASE2_PROTOCOL  *gSmmBase2 = NULL;
 EFI_GUID                   *gDxeCoreFileName;
 EFI_LOADED_IMAGE_PROTOCOL  *gDxeCoreLoadedImage;
 
+/**
+  Copies a source buffer to a destination buffer, and returns the destination buffer.
+
+  This function copies Length bytes from SourceBuffer to DestinationBuffer, and returns
+  DestinationBuffer.  The implementation must be reentrant, and it must handle the case
+  where SourceBuffer overlaps DestinationBuffer.
+
+  If Length is greater than (MAX_ADDRESS - DestinationBuffer + 1), then ASSERT().
+  If Length is greater than (MAX_ADDRESS - SourceBuffer + 1), then ASSERT().
+
+  @param  DestinationBuffer   The pointer to the destination buffer of the memory copy.
+  @param  SourceBuffer        The pointer to the source buffer of the memory copy.
+  @param  Length              The number of bytes to copy from SourceBuffer to DestinationBuffer.
+
+  @return None.
+
+**/
+STATIC
+VOID
+EFIAPI
+EfiCopyMem (
+  IN VOID   *DestinationBuffer,
+  IN VOID   *SourceBuffer,
+  IN UINTN  Length
+  )
+{
+  CopyMem (
+    DestinationBuffer,
+    SourceBuffer,
+    Length
+    );
+}
+
+/**
+  Fills a target buffer with a byte value, and returns the target buffer.
+
+  This function fills Length bytes of Buffer with Value.
+
+  If Length is greater than (MAX_ADDRESS - Buffer + 1), then ASSERT().
+
+  @param  Buffer    The memory to set.
+  @param  Length    The number of bytes to set.
+  @param  Value     The value with which to fill Length bytes of Buffer.
+
+  @return None.
+
+**/
+STATIC
+VOID
+EFIAPI
+EfiSetMem (
+  IN VOID   *Buffer,
+  IN UINTN  Length,
+  IN UINT8  Value
+  )
+{
+  SetMem (
+    Buffer,
+    Length,
+    Value
+    );
+}
+
 //
 // DXE Core Module Variables
 //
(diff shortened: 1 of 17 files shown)


mikebeaton:
@leiflindholm - If you don't mind, is it okay if I cc you here on this, since you've shown an interest in these types of changes before, and aren't popping up automatically on the reviewers list.

MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
```diff
@@ -87,8 +150,8 @@ EFI_BOOT_SERVICES  mBootServices = {
   (EFI_INSTALL_MULTIPLE_PROTOCOL_INTERFACES)CoreInstallMultipleProtocolInterfaces,        // InstallMultipleProtocolInterfaces
   (EFI_UNINSTALL_MULTIPLE_PROTOCOL_INTERFACES)CoreUninstallMultipleProtocolInterfaces,    // UninstallMultipleProtocolInterfaces
   (EFI_CALCULATE_CRC32)CoreEfiNotAvailableYetArg3,                                        // CalculateCrc32
-  (EFI_COPY_MEM)CopyMem,                                                                  // CopyMem
-  (EFI_SET_MEM)SetMem,                                                                    // SetMem
+  (EFI_COPY_MEM)EfiCopyMem,                                                               // CopyMem
```
leiflindholm, line 153: This may sound a bit unhinged, but ... could we consider changing the CopyMem/SetMem implementations instead?
If you search through edk2/edk2-platforms nearly nowhere ever looks at the return value.

RedfishPkg/PrivateLibrary/RedfishCrtLib/RedfishCrtLib.c
```diff
@@ -26,6 +25,98 @@ char  errnum_message[] = "We don't support to map errnum to the error message on
 #endif
 int  GLOBAL_USED  _fltused;
 
+typedef
```
leiflindholm, line 28: This is clearly a recurring issue, so maybe we could solve it once and properly instead? How about a PerformQuickSortCrt () wrapper function in SortLib that accepts a SORT_COMPARE_CRT*?

leiflindholm (changes requested):

MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
mikebeaton: It sounds sensible to me, but ... should we wait for anyone else to offer an opinion? (I feel like @ardbiesheuvel, @mdkinney, @makubacki, @os-d for a start might possibly have opinions to add either way?)

As you mention, and as with the general step of enabling this warning in the first place, this would also need at least some degree of reasonable due diligence checking against edk2-platforms as well (which I haven't done yet, but was planning to if this is getting close to looking acceptable).

mikebeaton: BTW, the other fix, possibly cleaner than what I have now, and if we don't do the above, would be to add eight wrappers for EfiSetMem/EfiCopyMem (perhaps better named) to the eight libraries mentioned in this PR https://github.com/tianocore/edk2/pull/12357 (i.e. all the libraries which already have duplicated CopyMem/SetMem wrappers).

While this would replace four copies with eight, it was also make the duplication explicit and move it into a standard place (where the same type of duplication, for more or less the same reason, already explicitly exists).

leiflindholm: For edk2-platforms, I can find no users of SetMem return value, and the only user of CopyMem return value is entirely pointless.

leiflindholm: And yes, we should definitely wait for other opinions :)

os-d: I think the change to align to the spec is correct, how we get there I don't have a big opinion on. I think to Leif's point, most callers don't check the return from [Copy/Set]Mem and tbh, even though it is a standard pattern, I don't think it is that valuable to return destination. So I'd personally be fine with Leif's suggestion.

mikebeaton: FWIW I am certain @leiflindholm you are aware but just to mention here, all the CopyMem and SetMem which are part of various UEFI interface definitions either return VOID, or return EFI_STATUS if they do some pre-checking of their params (e.g. for valid memory ranges, for that interface). Just for comparison with whatever change we do or don't make here.

mikebeaton: I think I may be able to clean up the duplication in the wrapper functions in https://github.com/tianocore/edk2/pull/12357 in such a way that, if we don't really feel the need to remove the return values from CopyMem and SetMem, we can get variants with no return value (VoidCopyMem, VoidSetMem?) cleanly, without much code duplication, from the central memory libraries.

mdkinney: I recommend we focus on addressing the specific warnings here without changing any public APIs definitions, so this is a 100% compatible change.  

We can discuss Leif's suggestion in a separate issue.

RedfishPkg/PrivateLibrary/RedfishCrtLib/RedfishCrtLib.c
mikebeaton: @leiflindholm - Sorry, I don't know if you have a moment to look at this - but apart from these issues of reducing duplication, I think this PR would be ready - or at least close to.

In the case of this issue, this would need adding to BaseSortLib and UefiSortLib, I believe - is that what you had in mind?

I also have an unanswered question here which is somewhat confusing me as to the right way to go (i.e. what I've done now doesn't change whatever decisions were made here, but what I do next might): https://github.com/tianocore/edk2/pull/4906#issuecomment-4172800659

leiflindholm: > In the case of this issue, this would need adding to BaseSortLib and UefiSortLib, I believe - is that what you had in mind?

Debating whether this makes sense only in BaseSortLib and that's what components relying on this behaviour shold use.
But, yes.

MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
ardbiesheuvel: Agree that changing the SetMem/CopyMem prototypes is fine, given that nothing relies on those.

PR #12357, opened 2026-03-28
MdePkg: Normalize and unify memory wrappers

mikebeaton:
# Description

All of these wrapper files state in their own comments that they appear as eight identical copies.

Fix minor deviations from this:
 - Various files had picked up minor deviations in comments; fix these, adopting the majority decision
 - Additionally, SetMem and SetMemN were defined together in PeiMemoryLib, but in separate wrapper files in the other seven libraries, so again adopt the majority convention and split these in PeiMemoryLib

After ensuring that the wrapper files in each MemoryLib instance are identical, move them to a shared location to avoid duplication and make future changes or additions simpler.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

`find . -name SetMemWrapper.c | xargs -n1 diff ./MdePkg/Library/BaseMemoryLibRepStr/SetMemWrapper.c`
and similar.

CI

## Integration Instructions

N/A


mikebeaton:
I can see more work that can usefully be done here - moving back to draft.

mikebeaton:
cc @leiflindholm @mdkinney - If this PR looks okay, I can build on it to cleanly add the extra wrappers for VOID-returning SetMem and CopyMem variants for https://github.com/tianocore/edk2/pull/12355.

pierregondois:
Hello Mike,
Shouldn't the source files of modules be inside the module ? Cf.
```
All files (except those listed in the Packages sections) must reside in the directory containing the
INF file or in sub-directories of the directory containing the INF file
```
and
```
All paths are relative to the directory containing the INF file. If the filename is listed as myfile.c, the file
must be located in the same directory as the INF file. Absolute paths in the filename are prohibited.
```

Maybe the PR could do as in `OvmfPkg/Library/QemuFwCfgLib/` where the module contains all the various implementations:

```
MdePkg/Library/MemoryLib/
\-PeiMemoryLib.inf     # Pei Mem Lib
\-UefiMemoryLib.inf    # Uefi Mem Lib
\-...
\-ScanMem64Wrapper.c   # Common file
\-ScanMem32Wrapper.c   # Common file
\-...
\-Pei                  # Pei specific files
  \-PeiMemLibGeneric.c
  \-...
\-Uefi                 # Uefi specific files
  \-MemLibGeneric.c
  \-...
```

The issue of doing the above is that all the paths will have to be renamed. E.g.
`MdePkg/Library/PeiMemoryLib/PeiMemoryLib.inf` -> `MdePkg/Library/MemoryLib/PeiMemoryLib.inf`

Given the amount of deleted code, maybe it's worth checking with the maintainters.

mikebeaton:
@pierregondois -

Hi! Thank you for providing those quotes. I guess I could argue that I barely match the second; not the first.

It sounds as if you're right that the only correct way to do it is what you say. Which is obviously a breaking change (in path name) for everything that consumes these libraries.

I'm not too fussed - I could just revert to the first commit and normalise the files (to be done again, presumably, now and then... :-( ).

Who were you suggesting I contact? I've already cc'd a couple of EDK 2 stewards (who were interested in the PR related to this) above, and ofc the automatically added reviewer list for this PR includes all listed maintainers for these files in MdePkg already. (Feel free to ping anyone else who you think it would be appropriate to do so.)

MdePkg/Library/PeiMemoryLib/SetMem32Wrapper.c
```diff
@@ -11,7 +11,7 @@
     PeiMemoryLib
     UefiMemoryLib
 
-  Copyright (c) 2006 - 2009, Intel Corporation. All rights reserved.<BR>
+  Copyright (c) 2006 - 2010, Intel Corporation. All rights reserved.<BR>
```
pierregondois: It seems the date is updated without actual modification to the file

pierregondois:
> I'm not too fussed - I could just revert to the first commit and normalise the files (to be done again, presumably, now and then... :-( ).

IMO it would be useful to factorize all the functions, given the amount of code deleted.
But as discussed this would be a breaking change.

> Who were you suggesting I contact? I've already cc'd a couple of EDK 2 stewards (who were interested in the PR related to this) above, and ofc the automatically added reviewer list for this PR includes all listed maintainers for these files in MdePkg already. (Feel free to ping anyone else who you think it would be appropriate to do so.)

I think the right person are in CC, I just wanted to avoid that you start making the suggested modifications and later someone else says that it was a bad idea.

MdePkg/Library/PeiMemoryLib/SetMem32Wrapper.c
mikebeaton: This was from normalising the files. Since the date is 2010 in some of the (near) duplicates of the same file, I figure that the only possible correct date in a unified version must be 2010.

mikebeaton: To put it another way, the only point of the first commit is to show how each file, as it was before, differs from what will be the unified file, in the next commit. This is how this one differs.

I can see the argument which says that it was never valid for this copy of the file, in this location, to have this copyright date - but I thought it was useful to make the above mentioned information easily visible explicitly in the commit history.

MdePkg/Library/BaseMemoryLib/BaseMemoryLib.inf
```diff
@@ -26,19 +26,19 @@
...
-  IsZeroBufferWrapper.c
+  ../Common/MemoryLibWrappers/ScanMem64Wrapper.c
+  ../Common/MemoryLibWrappers/ScanMem32Wrapper.c
+  ../Common/MemoryLibWrappers/ScanMem16Wrapper.c
+  ../Common/MemoryLibWrappers/ScanMem8Wrapper.c
+  ../Common/MemoryLibWrappers/ZeroMemWrapper.c
+  ../Common/MemoryLibWrappers/CompareMemWrapper.c
+  ../Common/MemoryLibWrappers/SetMemNWrapper.c
+  ../Common/MemoryLibWrappers/SetMem64Wrapper.c
+  ../Common/MemoryLibWrappers/SetMem32Wrapper.c
```
mdkinney, line 37: Use of .. in library or module INF is not allowed.  

The only way to achieve the shared code would be to introduce yet another lib class.

mikebeaton: @pierregondois suggests a way to achieve it above

mdkinney: I agree that is functional but is a non-backwards compatible change to platform DSC files that select different memory libs.  This can not be solved the MdeLibs.dsc.inc because there are many instances.  Platforms select different instances based on CPU capabilities, size, and perf.

mikebeaton: Yet another lib class would be non-backwards compatible too, if I understood correctly? (Everything that referenced this lib and wanted these wrappers would have to reference this yet another lib class.)

mdkinney: If it is a single lib class -> lib instance and it is in MdePkg, that mapping can be added to MdeLibs.dsc.inc.

mdkinney: I am not recommending adding a new lib class.  These files have not been updated in a long time.  So the fact there is some duplication is not perfect but workable.  What is the specific harm in leaving things as they are?

mikebeaton: > If it is a single lib class -> lib instance and it is in MdePkg, that mapping can be added to MdeLibs.dsc.inc.

Right, I see.

> What is the specific harm in leaving things as they are?

Probably nothing. I was adding yet more duplication in https://github.com/tianocore/edk2/pull/12355 - specifically in its first commit, where I ended up with four copies each of new, VOID returning SetMem and CopyMem wrappers. I was trying to find one place to add something without duplication, and ended up here.

But I agree, there's probably nothing to be done that's anything like worth the upset involved.

leiflindholm: The specific harm in leaving things like they are is that it provides a poor example for developers trying to learn how edk2 modules should be designed. And it leaves a maintenance hazard if any of these ever need to be updated for whatever reason.

Getting rid of the duplication is indeed not critical, but the cost of needing to update some lines of .dsc/.fdf is low.

mikebeaton: > If it is a single lib class -> lib instance and it is in MdePkg, that mapping can be added to MdeLibs.dsc.inc.

If it is reasonably low effort to put up a PR with a working change, I though I might do it - just as something more concrete on which to base any further discussion.

But @mdkinney I just double checked, and because this duplication occurs between different versions of BaseMemoryLib, it still looks to me as if multiple consumers need to be changed? E.g. just searching the project for `BaseMemoryLib|` finds 54 results in .dsc files, and BaseMemoryLib doesn't seem to be referenced in MdeLibs.dsc.inc. Apologies if I'm probably still misunderstanding the suggestion.

pierregondois: IIUC, the suggestion was to create another library containing the common code (let s call it `CommonBaseMemoryLib`). The modules would just call this library for the functions that have no specificities.

For instance, the library would contain a generic implementation of `CopyMem()`.
`BaseMemoryLibOptDxe.inf`, `BaseMemoryLibMmx.inf` and other libraries all seem to use the same implementation, so they would just have to call the `CommonBaseMemoryLib` implementation of `CopyMem()`.
If a module (e.g. `BaseMemoryLibOptDxe.inf`) wants to instead have a specific implementation, it would just not rely on the common library.

`CommonBaseMemoryLib` would only have one implementation and be in `MdePkg.dsc.inc`.
This has the advantage of being backward compatible, but another meta-library is created, which might be complicated for a simple library as the BaseMemoryLib.

mikebeaton: @pierregondois - Got it, yes, the vast majority (though not quite all) consumers of `BaseMemoryLib|` within edk2 and edk2-platforms do include `MdeLibs.dsc.inc`. Thank you.

PR #12358, opened 2026-03-29
MdeModulePkg/Core: Prevent possible premature handle removal

### LINKED ISSUE #11114: [Bug]: UninstallMultipleProtocolInterfaces() can fail to reinstall protocols on failure

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

MdeModulePkg

RELEASE, DEBUG

### Current Behavior

If UninstallMultipleProtocolInterfaces() fails to uninstall an interface, it reinstalls what it uninstalled so far, but if it uninstalls the last interface on the handle, the handle is removed, and any failure after that is unable to restore anything because the handle is invalid.

Using the attached driver, the debug output includes
```
Bug4EntryPoint called
InstallProtocolInterface: 5C198761-16A8-4E69-972C-89D67954F81D 44116490
Bug4EntryPoint installation succeeded
  1 protocols installed on handle
    5C198761-16A8-4E69-972C-89D67954F81D
InstallProtocolInterface: 5C198761-16A8-4E69-972C-89D67954F81D 44116490
InstallProtocolInterface: input handle at 0x44153718 is invalid
InstallProtocolInterface: 5C198761-16A8-4E69-972C-89D67954F81D 44116490 failed with Invalid Parameter
Bug4EntryPoint uninstallation failed
  Call to ProtocolsPerHandle() failed
```

### Expected Behavior

One would expect the reinstallation to succeed. For example, the removal of the handle could be delayed until the entire operation has succeeded.

### Steps To Reproduce

1. Create the folder `edk2/EmulatorPkg/Demo1_Bugs`
2. Add the attached files to `edk2/EmulatorPkg/Demo1_Bugs`
3. Add this line to the `[Components]` section of `EmulatorPkg/EmulatorPkg.dsc`:
```
EmulatorPkg/Demo1_Bugs/Demo1_Bug4.inf
```
4. Add this line to the `[FV.FvRecovery]` section of EmulatorPkg/EmulatorPkg.fdf:
```
INF EmulatorPkg/Demo1_Bugs/Demo1_Bug4.inf
```
5. cd to the `edk2` folder
6. run
```
   $ EmulatorPkg/build.sh
   $ EmulatorPkg/build.sh run 2> ../Debug_bugs.txt
```
7. Press `ENTER` or wait for the UEFI Shell screen
8. Press `ENTER` or wait for the UEFI Shell command-line prompt
9. Type `exit`
10. Select `Reset` from the EmulatorPkg screen
11. Examine `../Debug_bugs.txt` and look for `Bug4EntryPoint`

[Demo1_Bug4.inf.txt](https://github.com/user-attachments/files/20416875/Demo1_Bug4.inf.txt)
[Demo1_Bug4.c.txt](https://github.com/user-attachments/files/20417524/Demo1_Bug4.c.txt)

### Build Environment

```markdown
- OS(s): Ubuntu 22.04
- Tool Chain(s): GCC5
```

### Version Information

```text
Commit: af8859bc
```

### Urgency

Low

Someone else needs to fix it

Maintainer feedback requested

### Anything else?

_No response_

khaliid2040:
# Description
    
fixes: #11114
    
According to UEFI spec function UinstallMultipleProtocolInterfaces() dictate if any issues happen after one or more protocols uninstalled with UinstallProtocolInterface, all protocols uninstalled prior the error should be reinstalled with InstallProtocolInterface().
    
However, the current implementation of UninstallProtocolInterface() if there is no interface left on the handle it will remove that handle and free it. This later causes UninstallMultipleProtocolInterfaces() to get back invalid freed handle and fail to reinstall protocols.

To fix that, these patches introduce an internal function for UinstallProtocolInterface() which both implementations could use. This preserves the API while making each function handle it's own handle removal independent. Thus both UinstallProtocolInterface() and UinstallMultipleProtocolInterfaces() will use those internal functions.

For more information please refer to linked issue which provides test case and steps to reproduce.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

On OVMF X64. For patch reproduction and test case has been provided in the linked issue.

## Integration Instructions
N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Hand/Handle.c
+++ b/MdeModulePkg/Core/Dxe/Hand/Handle.c
@@ -131,6 +131,43 @@ CoreValidateHandle (
 }
 
 /**
+  Remove handle from handle database and free it.
+
+  @param UserHandle    Handle to be removed from list
+
+  @return EFI_SUCCESS   the handle is removed
+  @return EFI_NOT_READY the handle has one or more protocols installed
+ **/
+STATIC
+EFI_STATUS
+CoreRemoveHandleFromHandleList (
+  IN EFI_HANDLE  UserHandle
+  )
+{
+  IHANDLE  *Handle;
+
+  Handle = UserHandle;
+
+  if (!IsListEmpty (&Handle->Protocols)) {
+    return EFI_NOT_READY;
+  }
+
+  //
+  // If there are no more interfaces for the handle, free the handle
+  //
+  Handle->Signature = 0;
+  OrderedCollectionDelete (
+    gOrderedHandleList,
+    OrderedCollectionFind (gOrderedHandleList, Handle),
+    NULL
+    );
+  RemoveEntryList (&Handle->AllHandles);
+  CoreFreePool (Handle);
+
+  return EFI_SUCCESS;
+}
+
+/**
   Finds the protocol entry for the requested protocol.
   The gProtocolDatabaseLock must be owned
 
@@ -788,9 +825,7 @@ CoreDisconnectControllersUsingProtocolInterface (
 }
 
 /**
-  Uninstalls all instances of a protocol:interfacer from a handle.
-  If the last protocol interface is remove from the handle, the
-  handle is freed.
+  Internal function that removes protocol interface from handle.
 
   @param  UserHandle             The handle to remove the protocol handler from
   @param  Protocol               The protocol, of protocol:interface, to remove
@@ -798,11 +833,9 @@ CoreDisconnectControllersUsingProtocolInterface (
 
   @retval EFI_INVALID_PARAMETER  Protocol is NULL.
   @retval EFI_SUCCESS            Protocol interface successfully uninstalled.
-
-**/
+ **/
 EFI_STATUS
-EFIAPI
-CoreUninstallProtocolInterface (
+CoreUninstallProtocolInterfacesInternal (
   IN EFI_HANDLE  UserHandle,
   IN EFI_GUID    *Protocol,
   IN VOID        *Interface
@@ -882,25 +915,44 @@ CoreUninstallProtocolInterface (
     Status = EFI_SUCCESS;
   }
 
-  //
-  // If there are no more handlers for the handle, free the handle
-  //
-  if (IsListEmpty (&Handle->Protocols)) {
-    Handle->Signature = 0;
-    OrderedCollectionDelete (
-      gOrderedHandleList,
-      OrderedCollectionFind (gOrderedHandleList, Handle),
-      NULL
-      );
-    RemoveEntryList (&Handle->AllHandles);
-    CoreFreePool (Handle);
-  }
-
 Done:
   //
   // Done, unlock the database and return
   //
   CoreReleaseProtocolLock ();
+  return Status;
+}
+
+/**
+  Uninstalls all instances of a protocol:interface from a handle.
+  If the last protocol interface is removed from the handle, the
+  handle is freed.
+
+  @param  UserHandle             The handle to remove the protocol handler from
+  @param  Protocol               The protocol, of protocol:interface, to remove
+  @param  Interface              The interface, of protocol:interface, to remove
+
+  @retval EFI_INVALID_PARAMETER  Protocol is NULL.
+  @retval EFI_SUCCESS            Protocol interface successfully uninstalled.
+
+**/
+EFI_STATUS
+EFIAPI
+CoreUninstallProtocolInterface (
+  IN EFI_HANDLE  UserHandle,
+  IN EFI_GUID    *Protocol,
+  IN VOID        *Interface
+  )
+{
+  EFI_STATUS  Status;
+
+  Status = CoreUninstallProtocolInterfacesInternal (UserHandle, Protocol, Interface);
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  Status = CoreRemoveHandleFromHandleList (UserHandle);
+
   return Status;
 }
 
@@ -954,7 +1006,7 @@ CoreUninstallMultipleProtocolInterfaces (
     //
     // Uninstall it
     //
-    Status = CoreUninstallProtocolInterface (Handle, Protocol, Interface);
+    Status = CoreUninstallProtocolInterfacesInternal (Handle, Protocol, Interface);
   }
 
   VA_END (Args);
@@ -976,6 +1028,8 @@ CoreUninstallMultipleProtocolInterfaces (
 
     VA_END (Args);
     Status = EFI_INVALID_PARAMETER;
+  } else {
+    Status = CoreRemoveHandleFromHandleList (Handle);
   }
 
   return Status;
--- a/MdeModulePkg/Core/Dxe/Hand/Handle.c
+++ b/MdeModulePkg/Core/Dxe/Hand/Handle.c
@@ -131,6 +131,50 @@ CoreValidateHandle (
 }
 
 /**
+  Remove handle from handle database and free it.
+
+  @param UserHandle    Handle to be removed from list
+
+  @return EFI_SUCCESS   the handle is removed
+  @return EFI_NOT_READY the handle has one or more protocols installed
+ **/
+STATIC
+EFI_STATUS
+CoreRemoveHandleFromHandleList (
+  IN EFI_HANDLE  UserHandle
+  )
+{
+  IHANDLE  *Handle;
+
+  Handle = UserHandle;
+
+  //
+  // Lock the protocol database
+  //
+  CoreAcquireProtocolLock ();
+
+  if (!IsListEmpty (&Handle->Protocols)) {
+    CoreReleaseProtocolLock ();
+    return EFI_INVALID_PARAMETER;
+  }
+
+  //
+  // If there are no more interfaces for the handle, free the handle
+  //
+  Handle->Signature = 0;
+  OrderedCollectionDelete (
+    gOrderedHandleList,
+    OrderedCollectionFind (gOrderedHandleList, Handle),
+    NULL
+    );
+  RemoveEntryList (&Handle->AllHandles);
+  CoreFreePool (Handle);
+  CoreReleaseProtocolLock ();
+
+  return EFI_SUCCESS;
+}
+
+/**
   Finds the protocol entry for the requested protocol.
   The gProtocolDatabaseLock must be owned
 
@@ -788,9 +832,7 @@ CoreDisconnectControllersUsingProtocolInterface (
 }
 
 /**
-  Uninstalls all instances of a protocol:interfacer from a handle.
-  If the last protocol interface is remove from the handle, the
-  handle is freed.
+  Internal function that removes protocol interface from handle.
 
   @param  UserHandle             The handle to remove the protocol handler from
   @param  Protocol               The protocol, of protocol:interface, to remove
@@ -798,11 +840,9 @@ CoreDisconnectControllersUsingProtocolInterface (
 
   @retval EFI_INVALID_PARAMETER  Protocol is NULL.
   @retval EFI_SUCCESS            Protocol interface successfully uninstalled.
-
-**/
+ **/
 EFI_STATUS
-EFIAPI
-CoreUninstallProtocolInterface (
+CoreUninstallProtocolInterfacesInternal (
   IN EFI_HANDLE  UserHandle,
   IN EFI_GUID    *Protocol,
   IN VOID        *Interface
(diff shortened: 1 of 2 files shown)


khaliid2040:
Do we need a synchronization to prevent a race condition?

MdeModulePkg/Core/Dxe/Hand/Handle.c
```diff
@@ -883,9 +884,11 @@ CoreUninstallProtocolInterface (
   }
 
   //
-  // If there are no more handlers for the handle, free the handle
+  // If there are no more interfaces for the handle, free the handle. Except when we get
+  // called by CoreUninstallMultipleProtocolInterfaces() which may reinstall protocols on
+  // the handle later.
   //
-  if (IsListEmpty (&Handle->Protocols)) {
+  if (IsListEmpty (&Handle->Protocols) && !mDelayHandleRemoval) {
```
YangGangUEFI: Always `FALSE` when called by `CoreUninstallMultipleProtocolInterfaces()`?

So, we might need to add a separate logic for `CoreUninstallMultipleProtocolInterfaces()` to handle these cases?

khaliid2040: Yes, that an issue. Handle will not be removed  when CoreUninstallMultipleProtocolInterfaces() becomes successful. It is one limitation of this patch.

> So, we might need to add a separate logic for CoreUninstallMultipleProtocolInterfaces() to handle these cases?

It would have been better with seperate change but spec dictate it should be called UinstallProtocolInterface() and handle to be reinstalled which this implementation remove it. Look at the spec function documentation [here](https://uefi.org/specs/UEFI/2.11/07_Services_Boot_Services.html#efi-boot-services-uninstallmultipleprotocolinterfaces).

I will think of approach to implement without violating spec requirement and preserving handle.

khaliid2040:
@mdkinney and @lgao4 Could you please review this patch when you get a chance?

This patch has multiple issues and it may require deep discussions.

khaliid2040:
@mdkinney and @lgao4 Could you please get a chance for this PR. Thanks

mdkinney:
I want to clarify the specific case being addressed here.

If a handle contains multiple protocols and one of the protocols fails to be uninstalled, then everything works as expected.

It is only the condition that the very last protocol on the handle fails the uninstall operation.  In that case, the handle is freed, and there is no handle to process the reinstall protocol interface operations on.  And we want to preserve the original handle.

mdkinney:
The global variable does not look like the right implementation due to the race conditions.  This PR will need some changes to guarantee no race conditions.

MdeModulePkg/Core/Dxe/Hand/Handle.c
mdkinney: I am wondering of the state of Prot can be used instead of global variable.   If CoreRemoveInterfaceFromProtocol() returns NULL, then the removal failed and the handle can be preserved.  However, we need to know if this is being called from UninstallMultipleProtocolInterfaces() or not.  May need to pass in a flag to know if called from UninstallMultipleProtocolInterfaces() or from UninstallProtocolInterface().  By passing in a parameter, it can remove the use of a global variable.

There may still be a corner case if the last uninstall fails and no protocols can be reinstalled, the we would have a handle with no protocols, which is an unexpected state.  Almost requires garbage collection on the handle database.  We could defer the removal of any empty handle and the last action from all APIs that can remove protocols can call an internal API to check for empty handles and free empty handles.  But that may be a much more complex change and may be slower for the additional scan.

khaliid2040: @mdkinney We cannot preserve the handle. This is a fundamental flaw between this implementation and the spec.

The spec document of function UinstallProtocolInterface() dictate the handle to be removed if the last interface uninstalled and in UinstallMultipleProtocolInterface() the interface should be reinstalled if the last interface removal failed.

I did some refactor.  Could you please check it. I will fix CI and add/update comments after discussions.

The updated change added two central functions one is internal implementation of UinstallProtocolInterface() which handles uninstallation of the inteface from the handle and the other is for removing the handle if it is empty. This fixed the bug without effecting the API.

Thus both UinstallprotocolInterface and UinstallMultipleProtocol() could use that central implementation. However each one will handle the handle removal independently.

khaliid2040:
@mdkinney Please review the new approach which is more simpler than previous approach(preserving the handle) when you get chance.

khaliid2040:
@lgao4 and @mdkinney polite ping. Please review this PR.

The PR became little larger than i am planned initially, but this is the best fix i could get for this bug. There is a contradiction between two function documentations in the spec. So i proposing each function to handle the handle removal independtly.

The CI failed due grammar, spelling and uncurtify issues which i will fix it after review.

UnitTestFrameworkPkg/Library/UnitTestUefiBootServicesTableLib/UnitTestUefiBootServicesTableLibProtocol.c
```diff
@@ -48,6 +48,35 @@ UnitTestValidateHandle (
...
+  //
+  // If there are no more handlers for the handle, free the handle
+  //
+  if (IsListEmpty (&Handle->Protocols)) {
+    Handle->Signature = 0;
+    RemoveEntryList (&Handle->AllHandles);
+    FreePool (Handle);
+  }
+
+  return EFI_NOT_READY;
```
lgao4: Here always return EFI_NOT_READY?

lgao4:
> @lgao4 and @mdkinney polite ping. Please review this PR.
> 
> The PR became little larger than i am planned initially, but this is the best fix i could get for this bug. There is a contradiction between two function documentations in the spec. So i proposing each function to handle the handle removal independtly.
> 
> The CI failed due grammar, spelling and uncurtify issues which i will fix it after review.

The code in DxeCore looks good to me .

MdeModulePkg/Core/Dxe/Hand/Handle.c
```diff
@@ -130,6 +130,41 @@ CoreValidateHandle (
...
+  )
+{
+  IHANDLE  *Handle;
+
+  Handle = UserHandle;
+
+  //
+  // If there are no more handlers for the handle, free the handle
+  //
+  if (IsListEmpty (&Handle->Protocols)) {
```
mdkinney: Change to if (!IsListEmpt()) and return EFI_NOT_READY.   Can reduce indent of success path.

MdeModulePkg/Core/Dxe/Hand/Handle.c
```diff
@@ -976,6 +1026,8 @@ CoreUninstallMultipleProtocolInterfaces (
 
     VA_END (Args);
     Status = EFI_INVALID_PARAMETER;
+  } else {
+    CoreRemoveHandleFromHandleList (Handle);
```
mdkinney: Should add check of status return.  The handle is expected to be empty at this point, but if it isn't, then an error is returned from this call and the uninstall multiple should return an error.

UnitTestFrameworkPkg/Library/UnitTestUefiBootServicesTableLib/UnitTestUefiBootServicesTableLibProtocol.c
```diff
@@ -48,6 +48,36 @@ UnitTestValidateHandle (
...
+  )
+{
+  IHANDLE  *Handle;
+
+  Handle = UserHandle;
+
+  //
+  // If there are no more handlers for the handle, free the handle
+  //
+  if (IsListEmpty (&Handle->Protocols)) {
```
mdkinney: Same comment as above to check error and return immediately.

UnitTestFrameworkPkg/Library/UnitTestUefiBootServicesTableLib/UnitTestUefiBootServicesTableLibProtocol.c
```diff
@@ -1644,6 +1694,8 @@ UnitTestUninstallMultipleProtocolInterfaces (
 
     VA_END (Args);
     Status = EFI_INVALID_PARAMETER;
+  } else {
+    Status = UnitTestRemoveHandleFromHandleList (Handle);
```
mdkinney, line 1700: This one looks correct.  The one in DXE Core should be updated to match.  Do need to check if EFI_NOT_READY is an expected return status.  May need to return a different error status code.

mdkinney (changes requested):

UnitTestFrameworkPkg/Library/UnitTestUefiBootServicesTableLib/UnitTestUefiBootServicesTableLibProtocol.c
khaliid2040: > This one looks correct. The one in DXE Core should be updated to match. Do need to check if EFI_NOT_READY is an expected return status. May need to return a different error status code.

From all EFI status codes, i found `EFI_NOT_READY` is the best fit (i.e the handle is not ready for removal, all interfaces must be removed before handle is deleted).

khaliid2040:
@mdkinney and @lgao4 friendly ping. I addressed your last reviews could you please re-review. Thanks

lgao4 (approved):

khaliid2040:
@lgao4 @mdkinney and @os-d A friendly ping, this PR was open for quite long could you please review and possibly merge it if there is no other reviews i should address.

MdeModulePkg/Core/Dxe/Hand/Handle.c
```diff
@@ -130,6 +130,42 @@ CoreValidateHandle (
...
+/**
+  Remove handle from handle database and free it.
+
+  @param UserHandle    Handle to be removed from list
+
+  @return EFI_SUCCESS   the handle is removed
+  @return EFI_NOT_READY the handle has one or more protocols installed
+ **/
+EFI_STATUS
+CoreRemoveHandleFromHandleList (
```
os-d, line 143: nit: static

MdeModulePkg/Core/Dxe/Hand/Handle.c
```diff
@@ -904,6 +922,39 @@ CoreUninstallProtocolInterface (
   return Status;
 }
 
+/**
+  Uninstalls all instances of a protocol:interfacer from a handle.
+  If the last protocol interface is remove from the handle, the
+  handle is freed.
```
os-d:
```suggestion
  Uninstalls all instances of a protocol:interface from a handle.
  If the last protocol interface is removed from the handle, the
  handle is freed.
```

UnitTestFrameworkPkg/Library/UnitTestUefiBootServicesTableLib/UnitTestUefiBootServicesTableLibProtocol.c
```diff
@@ -48,6 +48,37 @@ UnitTestValidateHandle (
...
 
+/**
+  Remove handle from handle database and free it.
+
+  @param UserHandle    Handle to be removed from list
+
+  @return EFI_SUCCESS   the handle is removed
+  @return EFI_NOT_READY the handle has one or more protocols installed
+ **/
+EFI_STATUS
```
os-d, line 60: nit: static

os-d (approved):

os-d:
A couple of nits, otherwise looks good. We'll need to wait for @mdkinney to approve (both because of his current request for changes and as UnitTestFrameworkPkg maintainer).

khaliid2040:
@mdkinney friendly ping, could you please review this PR since you requested changes before.

khaliid2040:
@mdkinney friendly remind, please review this when you get chance. It was sittting here for too long and i really need to be merged before next release cycle.

MdeModulePkg/Core/Dxe/Hand/Handle.c
```diff
@@ -904,6 +923,39 @@ CoreUninstallProtocolInterface (
...
+  )
+{
+  EFI_STATUS  Status;
+
+  Status = CoreUninstallProtocolInterfacesInternal (UserHandle, Protocol, Interface);
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  Status = CoreRemoveHandleFromHandleList (UserHandle);
```
makubacki, line 954: This directly returns the status from `CoreRemoveHandleFromHandleList()` which returns `EFI_NOT_READY`:

```c
  if (!IsListEmpty (&Handle->Protocols)) {
    return EFI_NOT_READY;
  }
```

Does that mean that any time a handle has remaining protocols, `EFI_NOT_READY` is returned with this change?

makubacki, line 954: `CoreRemoveHandleFromHandleList()` is running without the protocol database lock being held. You replaced the original `CoreUninstallProtocolInterface()` implementation with this function that makes the two function calls, but the lock is now only held inside `CoreUninstallProtocolInterfacesInternal()`.

khaliid2040: Yes, the code flow change of this PR is to seperate the handle removal of these two functions so each one handles it separately while using common functions. 

The issue is in spec too dictating impossible conditions like remove handle if it is empty in CoreUninstallProtocolInterface() and reinstall removed protocols to handle if there is error in protocol removal in CoreUninstallMultipleProtocolInterfaces(). If later function tries to reinstall protocols then it will get invalid freed handle. This PR created workaround to resolve that while preserving the original API code flow

khaliid2040: Thanks for catching, i will fix it.

makubacki: My concern is that this function (`CoreUninstallProtocolInterface()`) is the `UninstallProtocolInterface()` Boot Services implementation and it's not expected to return `EFI_NOT_READY` just because there are remaining protocols on the handle.

khaliid2040: I understand your concern, but we can't let CoreUninstallProtocolInterface() delete used handles. The function will return status codes documented in UEFI but this one is extra.

There are numerous cases where UEFI api implementations return status codes other than documented ones.

os-d: I agree with Michael, we have to return spec defined errors. There may well be cases where other spec defined APIs end up returning other errors, but when they are discovered, they should be fixed. The SCTs test for this, now of course they can’t test every possible scenario.

If there is a need for a new error code in any spec defined API, we can do code first to add it. I don’t think this case warrants it, can we simply return one of the defined errors at the spec defined API level?

khaliid2040: Okay, updated to return one of the defined status codes.

khaliid2040:
@mdkinney @os-d and @makubacki I addressed latest review comments, could you please check it.

makubacki:
@khaliid2040, have you run UEFI Self-Certification Test (SCT) with this change?

PR #12361, opened 2026-03-30
ArmPkg/CpuDxe: Support multiple entries in RegionIsSystemMemory check

### LINKED ISSUE #12363: [Bug]: OVMF crashes when exiting from aarch64 GRUB

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

ArmPkg

DEBUG

### Current Behavior

Ever since I updated OVMF to edk2-stable202602 in aarch64, OVMF crashes whenever I try to exit the GRUB menu using the 'c' then 'exit' command sequence.

### Expected Behavior

It should return to the menu normally.

### Steps To Reproduce

1. Install the latest openSUSE Tumbleweed using Libvirt.
2. Enter the GRUB2 menu, press "c", and then type "exit".

### Build Environment

```markdown
- OS(s): openSUSE Tumbleweed
- Tool Chain(s): GCC
```

### Version Information

```text
Tag: edk2-stable202602
```

### Urgency

Low

I will fix it

No maintainer feedback needed

### Anything else?

It has been observed that in the GetMemoryAttributes function, if the requested BaseAddress + Length exceeds a single entry of type EfiGcdSystemMemory in the GCD memory map, RegionIsSystemMemory returns FALSE, which then results in EFI_UNSUPPORTED.

r1chard-lyu:
# Description
The check performed by RegionIsSystemMemory is not necessarily limited to a single entry of type EfiGcdSystemMemory in the GCD memory map. For example, when a memory region spans multiple contiguous GCD entries, the current implementation returns False even though the entire range is system memory.

Therefore, this modification expands the RegionIsSystemMemory check to support multiple contiguous entries.

Fixes: https://github.com/tianocore/edk2/issues/12363

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Use the GetMemoryAttributes function for testing.

## Integration Instructions
N/A

### DIFF
--- a/ArmPkg/Drivers/CpuDxe/MemoryAttribute.c
+++ b/ArmPkg/Drivers/CpuDxe/MemoryAttribute.c
@@ -9,7 +9,7 @@
 #include "CpuDxe.h"
 
 /**
-  Check whether the provided memory range is covered by a single entry of type
+  Check whether the provided memory range is covered by one or more entries of type
   EfiGcdSystemMemory in the GCD memory map.
 
   @param  BaseAddress       The physical address that is the start address of
@@ -26,22 +26,25 @@ RegionIsSystemMemory (
   )
 {
   EFI_GCD_MEMORY_SPACE_DESCRIPTOR  GcdDescriptor;
-  EFI_PHYSICAL_ADDRESS             GcdEndAddress;
+  EFI_PHYSICAL_ADDRESS             CurrentAddress;
+  EFI_PHYSICAL_ADDRESS             EndAddress;
   EFI_STATUS                       Status;
 
-  Status = gDS->GetMemorySpaceDescriptor (BaseAddress, &GcdDescriptor);
-  if (EFI_ERROR (Status) ||
-      (GcdDescriptor.GcdMemoryType != EfiGcdMemoryTypeSystemMemory))
-  {
-    return FALSE;
+  CurrentAddress = BaseAddress;
+  EndAddress     = BaseAddress + Length;
+
+  while (CurrentAddress < EndAddress) {
+    Status = gDS->GetMemorySpaceDescriptor (CurrentAddress, &GcdDescriptor);
+    if (EFI_ERROR (Status) ||
+        (GcdDescriptor.GcdMemoryType != EfiGcdMemoryTypeSystemMemory))
+    {
+      return FALSE;
+    }
+
+    CurrentAddress = GcdDescriptor.BaseAddress + GcdDescriptor.Length;
   }
 
-  GcdEndAddress = GcdDescriptor.BaseAddress + GcdDescriptor.Length;
-
-  //
-  // Return TRUE if the GCD descriptor covers the range entirely
-  //
-  return GcdEndAddress >= (BaseAddress + Length);
+  return TRUE;
 }
 
 /**


ardbiesheuvel:
@os-d any thoughts?

LeviYeoReum (approved):

os-d:
> @os-d any thoughts?

I think that this is reasonable to not be constrained to a single GCD descriptor and that is an issue we have to fix all over the place. That is mixing two specs, anyway, the PI spec where the GCD is described and the UEFI spec where the memory attributes protocol is described. The memory attribute protocol shouldn't have a dependency on the state of something described in the PI spec (when it is as arbitrary as are there two descriptors here or one). The use case is also reasonable because what I'm guessing happened here (or could happen anyway) is that grub loaded an image through load image, which sets protection attributes, then grub unloaded it and set the whole region back to NX itself. The last part is what will fail if we constrain to a single GCD descriptor, but that is a valid scenario and no reason to force it to happen section by section.

However, I also think that the restricting of the memory attribute protocol to system memory only is questionable. The spec doesn't say it should be system memory only (i.e. not MMIO). The spec does envision returning EFI_ACCESS_DENIED in the case that system FW controls this memory region and it can't be changed, but currently this returns EFI_UNSUPPORTED. x86's memory attribute protocol lets any memory region be updated.

I don't think there is a security boundary here, if you can call the memory attribute protocol you can read TTBR0_EL2 and do whatever you want to the page table. I do think it is a good idea to not blindly map something that doesn't have a physical backing. So, checking that the GCD memory type is not NonExistent may be a better check. If the current behavior is kept, then I think the return status from the memory attribute protocol functions when this function fails should be updated to EFI_ACCESS_DENIED.

TL;DR: I would change the function to check if any part of the range is EfiGcdMemoryTypeNonExistent or EfiGcdMemoryTypeUnaccepted and return FALSE if so, returning EFI_UNSUPPORTED from the caller. I think x86 would benefit from having this check, too.

ardbiesheuvel:
> > @os-d any thoughts?
> 
> I think that this is reasonable to not be constrained to a single GCD descriptor and that is an issue we have to fix all over the place. That is mixing two specs, anyway, the PI spec where the GCD is described and the UEFI spec where the memory attributes protocol is described. The memory attribute protocol shouldn't have a dependency on the state of something described in the PI spec (when it is as arbitrary as are there two descriptors here or one). The use case is also reasonable because what I'm guessing happened here (or could happen anyway) is that grub loaded an image through load image, which sets protection attributes, then grub unloaded it and set the whole region back to NX itself. The last part is what will fail if we constrain to a single GCD descriptor, but that is a valid scenario and no reason to force it to happen section by section.
> 

Agree that the single GCD descriptor is problematic, and needs to be addressed.

> However, I also think that the restricting of the memory attribute protocol to system memory only is questionable. The spec doesn't say it should be system memory only (i.e. not MMIO). The spec does envision returning EFI_ACCESS_DENIED in the case that system FW controls this memory region and it can't be changed, but currently this returns EFI_UNSUPPORTED. x86's memory attribute protocol lets any memory region be updated.
> 

On ARM, removing the XN attribute from device mappings of MMIO regions might result in speculative I-side fetches, which usually ends as well as you might imagine (e.g., the I-cache prefetcher acking interrupts behind your back - true story). Beyond that, there are no compelling reasons to be overly strict when it comes to manipulating permission attributes on non-RAM regions.

> I don't think there is a security boundary here, if you can call the memory attribute protocol you can read TTBR0_EL2 and do whatever you want to the page table. I do think it is a good idea to not blindly map something that doesn't have a physical backing. So, checking that the GCD memory type is not NonExistent may be a better check. If the current behavior is kept, then I think the return status from the memory attribute protocol functions when this function fails should be updated to EFI_ACCESS_DENIED.
> 

I don't think we should permit the creation of new mappings using this interface, only the manipulation of permission attributes on existing, valid mappings. So perhaps the backing in the GCD map is irrelevant to begin with? So on ARM, the only check needed would be that device mappings retain their XN attribute.

os-d:
> On ARM, removing the XN attribute from device mappings of MMIO regions might result in speculative I-side fetches, which usually ends as well as you might imagine (e.g., the I-cache prefetcher acking interrupts behind your back - true story). Beyond that, there are no compelling reasons to be overly strict when it comes to manipulating permission attributes on non-RAM regions.
> 

Unfortunately, I learned this the hard way :). In my case it was speculative reads on sparse MMIO regions which caused stuck transactions in the memory controller and killed the system. That's why in Patina we follow what the Cpu Arch protocol in edk2 does, which is always set XN for device memory: https://github.com/tianocore/edk2/blob/f5c6c98d89e4d56e96804c2808c5a17e1558d025/ArmPkg/Drivers/CpuDxe/AArch64/Mmu.c#L377-L382.

> I don't think we should permit the creation of new mappings using this interface, only the manipulation of permission attributes on existing, valid mappings. So perhaps the backing in the GCD map is irrelevant to begin with? So on ARM, the only check needed would be that device mappings retain their XN attribute.

I agree with that statement. I think you could go either way whether to reject updates that remove XN from device memory or to silently enforce it like in the CPU arch protocol. For the memory attribute protocol, given that it can't update caching attributes, it may be better to return an error here because as you say, this will already be an existing device memory region so the caller would be explicitly trying to clear XN from it. In which case an explicit failure is better to catch the error in the caller.

r1chard-lyu:
> > I don't think we should permit the creation of new mappings using this interface, only the manipulation of permission attributes on existing, valid mappings. So perhaps the backing in the GCD map is irrelevant to begin with? So on ARM, the only check needed would be that device mappings retain their XN attribute.

> I agree with that statement. I think you could go either way whether to reject updates that remove XN from device memory or to silently enforce it like in the CPU arch protocol. For the memory attribute protocol, given that it can't update caching attributes, it may be better to return an error here because as you say, this will already be an existing device memory region so the caller would be explicitly trying to clear XN from it. In which case an explicit failure is better to catch the error in the caller.

If we intend to enforce XN for device memory, should we include this in the current PR or open a separate one?

> TL;DR: I would change the function to check if any part of the range is EfiGcdMemoryTypeNonExistent or EfiGcdMemoryTypeUnaccepted and return FALSE if so, returning EFI_UNSUPPORTED from the caller. I think x86 would benefit from having this check, too.

If we modify this function, should the function name also be updated, given that it would no longer be checking for system memory? Or, based on the discussion, is this change no longer necessary?

ardbiesheuvel:
Please drop the GCD check entirely. Instead, ensure that the API refuses to remove the XN attribute from device type mappings, as they are architecturally required. Beyond that, we can remove the guardrails from this API to match the UEFI spec and other implementations, and permit any memory mapping to be manipulated. Note that this implies that only valid mappings may be manipulated.

r1chard-lyu:
> Please drop the GCD check entirely. Instead, ensure that the API refuses to remove the XN attribute from device type mappings, as they are architecturally required. Beyond that, we can remove the guardrails from this API to match the UEFI spec and other implementations, and permit any memory mapping to be manipulated. Note that this implies that only valid mappings may be manipulated.

Thanks for the feedback. I'll create a new PR to implement these changes.

PR #12368, opened 2026-03-31
Check Signature when validate Handle

YangGangUEFI:
# Description

Check Signature when validate Handle

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Build pass and boot success.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Hand/Handle.c
+++ b/MdeModulePkg/Core/Dxe/Hand/Handle.c
@@ -116,7 +116,7 @@ CoreValidateHandle (
 {
   ORDERED_COLLECTION_ENTRY  *Entry;
 
-  if (UserHandle == NULL) {
+  if ((UserHandle == NULL) || (((IHANDLE *)UserHandle)->Signature != EFI_HANDLE_SIGNATURE)) {
     return EFI_INVALID_PARAMETER;
   }
 
--- a/UnitTestFrameworkPkg/Library/UnitTestUefiBootServicesTableLib/UnitTestUefiBootServicesTableLibProtocol.c
+++ b/UnitTestFrameworkPkg/Library/UnitTestUefiBootServicesTableLib/UnitTestUefiBootServicesTableLibProtocol.c
@@ -34,7 +34,7 @@ UnitTestValidateHandle (
   IHANDLE     *Handle;
   LIST_ENTRY  *Link;
 
-  if (UserHandle == NULL) {
+  if ((UserHandle == NULL) || (((IHANDLE *)UserHandle)->Signature != EFI_HANDLE_SIGNATURE)) {
     return EFI_INVALID_PARAMETER;
   }
 


StandaloneMmPkg/Core/Handle.c
```diff
@@ -247,7 +247,7 @@ MmInstallProtocolInterfaceNotify (
   Prot   = NULL;
   Handle = NULL;
 
-  if (*UserHandle != NULL) {
+  if (*UserHandle != NULL && ((IHANDLE *)*UserHandle)->Signature == EFI_HANDLE_SIGNATURE) {
```
kuqin12: Can you please add more explanation on why this might be needed? I would have thought the handle should be nullified by the caller. This change seems to indicate that you would like the logic to be able to process a dangling or uninitialized pointer? 

Note that the spec defined this argument as "IN/OUT" and "If * Handle is not a valid handle, then EFI_INVALID_PARAMETER is returned." So I do think the caller should be responsible for passing in a valid handler or NULL.

kuqin12 (changes requested):

MdeModulePkg/Core/Dxe/Hand/Handle.c
```diff
@@ -464,7 +464,7 @@ CoreInstallProtocolInterfaceNotify (
   Prot   = NULL;
   Handle = NULL;
 
-  if (*UserHandle != NULL) {
+  if (*UserHandle != NULL && ((IHANDLE *)*UserHandle)->Signature == EFI_HANDLE_SIGNATURE) {
```
khaliid2040: Redundant! CoreHandleProtocol() calls CoreOpenProtocol() which in turn calls CoreValidateHandle(). 

My suggestion is only add signature checking in CoreValidateHandle() other places are unneeded because they call CoreValidateHandle().

MdeModulePkg/Core/Dxe/Hand/Handle.c
```diff
@@ -494,10 +494,10 @@ CoreInstallProtocolInterfaceNotify (
   }
 
   //
-  // If caller didn't supply a handle, allocate a new one
+  // If caller didn't supply a valid handle, allocate a new one
   //
   Handle = (IHANDLE *)*UserHandle;
-  if (Handle == NULL) {
+  if (Handle == NULL || Handle->Signature != EFI_HANDLE_SIGNATURE) {
```
khaliid2040: There is two logic here:
 * Create new handle if `Handle` is NULL.
 * If handle supplied then call CoreValidateHandle() which  already checks signature (As you added above).

The suggestion i gave you before will apply here too.

MdeModulePkg/Core/PiSmmCore/Handle.c
```diff
@@ -240,7 +240,7 @@ SmmInstallProtocolInterfaceNotify (
   Prot   = NULL;
   Handle = NULL;
 
-  if (*UserHandle != NULL) {
+  if (*UserHandle != NULL && ((IHANDLE *)*UserHandle)->Signature == EFI_HANDLE_SIGNATURE) {
```
khaliid2040: SmmHandleProtocol() calls SmmValidateHandle() which already checks signature.

MdeModulePkg/Core/PiSmmCore/Handle.c
```diff
@@ -265,10 +265,10 @@ SmmInstallProtocolInterfaceNotify (
   }
 
   //
-  // If caller didn't supply a handle, allocate a new one
+  // If caller didn't supply a valid handle, allocate a new one
   //
   Handle = (IHANDLE *)*UserHandle;
-  if (Handle == NULL) {
+  if (Handle == NULL || Handle->Signature != EFI_HANDLE_SIGNATURE) {
```
khaliid2040: Same as i explained to you above in Dxe Handle source file.

StandaloneMmPkg/Core/Handle.c
```diff
@@ -247,7 +247,7 @@ MmInstallProtocolInterfaceNotify (
   Prot   = NULL;
   Handle = NULL;
```
khaliid2040, line 248: Same will apply here. In both cases MmValidateHandle() will be called which already checks signature.

khaliid2040 (changes requested):
Most of this signature checking is redandunt, all of this places end up in ValidateHandle() functions.

Only DXE handle (CoreValidateHandle) and it's unit test is missing signature checking which you already added. So other places are unnecessary.

StandaloneMmPkg/Core/Handle.c
YangGangUEFI: These changes were made in `MdeModulePkg\Core\Dxe\Hand\Handle.c` to handle a special case in `CoreUninstallMultipleProtocolInterfaces`, and then simply synced to Smm/StandaloneMm.

Since Smm/StandaloneMm does not have `***MultipleProtocolInterfaces` interface, this logic is unnecessary and I will remove it.

MdeModulePkg/Core/Dxe/Hand/Handle.c
YangGangUEFI: Ah, These changes are only based on current behavior and overthinking.

> When `CoreInstallProtocolInterfaceNotify()` called by `CoreUninstallMultipleProtocolInterfaces()`, `*UserHandle` is not `NULL`, but `Signature` is `0`. And not allocate a new one(overthinking, current logic is the same as the spec.)

> After https://github.com/tianocore/edk2/pull/12358 update, `Handle` passed into `CoreInstallProtocolInterface` should be a valid handle.

And this change will be removed.

khaliid2040 (approved):

YangGangUEFI:
@mdkinney @lgao4 Could you please help review these changes when you have a chance?

mdkinney:
The CR() macro also checks the signature.  So this change adds an additional check before the CR() macro is used, so we get an error return status instead of an ASSERT().  Right?

The ASSERT() behavior in DEBUG builds could be considered valuable because it halts.  With this change the ASSERT() will never occur and callers that do not evaluate the return status will keep running.

Just want to make sure this change does not introduce an unexpected behavior change that allows failure to be silent.

YangGangUEFI:
The change is simply to check if the handle is valid *a little faster*.

> The CR() macro also checks the signature. So this change adds an additional check before the CR() macro is used, so we get an error return status instead of an ASSERT(). Right?

Therefore, yes.

Based on this, I think this PR should be closed, or add `ASSERT` and update the behavior of `SmmValidateHandle` and `MmValidateHandle`?

khaliid2040:
I think these changes are useful, we should check the handle signature. The CR() macro and CoreValidateHandle() should be independent (i.e CR() is not always used on places called by CoreValidateHandle()).

These changes don't have any effect to callers, if the callers don't check the return then the callers should be fixed. What this change does is return EFI_INVALID_PARAMETER if signature is invalid too, so callers that use ASSERT_EFI_ERROR() will catch.

Also, what i observed is on multiple places called CoreValidateHandle() there was no CR() macro.

PR #12369, opened 2026-03-31
MdePkg/PciSegmentLib: Mask reserved bits in PCI_SEGMENT_LIB_ADDRESS macro

gapandya:
Apply final AND masks to clear reserved bit ranges:
- Bits 28-31 (true branch): 0xFFFF0FFFFFFF
- Bits 48-63 (both branches): 0xFFFFFFF (false) / implicit (true)

# Description

This adds precautionary reserved bits masking and macro safety.
Ensures reserved bits remain zero even if per-field masks are removed. 

- [ ] Breaking change?
  - No.
- [ ] Impacts security?
  - No.
- [ ] Includes tests?
  - No.

## How This Was Tested

Booted server platform with change.

## Integration Instructions
NA

### DIFF
--- a/MdePkg/Include/Library/PciSegmentLib.h
+++ b/MdePkg/Include/Library/PciSegmentLib.h
@@ -24,6 +24,7 @@
   accesses when PCI Segments other than Segment #0 must be accessed.
 
 Copyright (c) 2006 - 2018, Intel Corporation. All rights reserved.<BR>
+Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.<BR>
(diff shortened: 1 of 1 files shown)


gapandya:
Hi @lgao4 , @mdkinney ,

Can you please help to review the code. 
Thanks.

lgao4 (approved):

mdkinney:
Why is this required at all?  The construction of the value from the individual fields would never set those bits.

Please add an example where the reserved bits would be set before this change is applied.

mdkinney (changes requested):
Add same code the demonstrates reserved bits being set before this change

PR #12370, opened 2026-03-31
BaseTools/build.py: Use full source file path for dependency generation

Kowsik-003:
# Description

During Silent build,NMAKE suppresses the command echo entirely. As a result, the only output in ProcOut is the MSVC compiler’s output lines. Without the command echo, there is no full path in the output to identify which source file is currently being compiled. For unique basenames this is not an issue, but for namesake files (for example, AmdSev.c located in different directories), it is impossible to determine which file’s includes are being listed. This change improves dependency generation for MSVC builds by introducing explicit handling for source files with duplicate basenames (namesake sources). A new variable current_source_abs is added to consistently track the resolved absolute path of the active source file instead of repeatedly recomputing it from SourceFileAbsPathMap. To correctly resolve namesake files in silent builds (where compiler commands are not echoed), a namesake_queue is introduced, which preserves source ordering and sequentially maps basename occurrences to their corresponding full paths. Additionally, a cc_cmd_in_output flag is implemented to detect the presence of compiler command lines in the output stream; when present, source paths are derived directly from command-line arguments, otherwise the queue-based resolution is used. This ensures correct mapping of basenames to absolute paths across the silent builds, fixing incorrect dependency generation when multiple source files share the same name.

- [ ] Breaking change?
No
- [ ] Impacts security?
No
- [ ] Includes tests?
No

## How This Was Tested

When building UefiCpuPkg, can't find that deps file created for all the AmdSev.c files in the respective paths. Where the last occurence AmdSev.c  deps file is generated.

Example: AmdSev.c
1.UefiCpuPkg/Library/MpInitLib/AmdSev.c
2.UefiCpuPkg/Library/MpInitLib/Ia32/AmdSev.c
3.UefiCpuPkg/Library/MpInitLib/X64/AmdSev.c

## Integration Instructions

NA


mikebeaton:
> This issue was caused by ExpandNmakeSourcePaths() and SourceFileRelativePath() using only the base filename as the key when tracking source files.

This sounds exactly as if you are saying that was a problem in ExpandNmakeSourcePaths() and SourceFileRelativePath() which you have fixed. But that does not make sense, as you have added ExpandNmakeSourcePaths() and SourceFileRelativePath() yourself.

BaseTools/Source/Python/build/build.py
```diff
@@ -279,7 +308,17 @@ def LaunchCommand(Command, WorkingDir,ModuleAuto = None):
...
         iau = IncludesAutoGen(WorkingDir,ModuleAuto)
         if ModuleAuto.ToolChainFamily == TAB_COMPILER_MSFT:
-            iau.CreateDepsFileForMsvc(Proc.ProcOut)
+            if isinstance(Command, str):
+                CmdStr = Command
+            else:
+                CmdStr = ' '.join(Command)
+            CmdTokens = CmdStr.split()
+            IsSilentBuild = '-s' in CmdTokens
+            if not IsSilentBuild:
```
mikebeaton: Please explain why you only need to track relative paths in silent build.

Kowsik-003: NMAKE suppresses the command echo entirely. As a result, the only output in ProcOut is the MSVC compiler’s output lines. Without the command echo, there is no full path in the output to identify which source file is currently being compiled. For unique basenames this is not an issue, but for namesake files (for example, AmdSev.c located in different directories), it is impossible to determine which file’s includes are being listed.
ExpandNmakeSourcePaths() resolves this by replacing each basename occurrence in the output with the corresponding full path from NameSakeSourceFileMap, sequentially. This ensures that CreateDepsFileForMsvc receives a full‑path key. 
But in non silent build it echoes the command and the full path So deps generation is proper even in repeated basefiles name.

Kowsik-003:
@mikebeaton 
To clarify, this was not a defect or bug in ExpandNmakeSourcePaths() or SourceFileRelativePath(). These functions were newly introduced as part of this change.
The underlying issue lies with NMAKE: it suppresses command echoing, so the MSVC compiler output does not include the full path of the source file being compiled. As a result, for namesake source files (for example, AmdSev.c located in different directories), successive occurrences overwrite earlier ones. Consequently, the dependency file ends up being generated only for the last occurrence of the namesake file.
ExpandNmakeSourcePaths() and SourceFileRelativePath() were added specifically to map each basename occurrence in the output to its resolved full path, allowing CreateDepsFileForMsvc to operate on full‑path keys.

mikebeaton:
> @mikebeaton To clarify, this was not a defect or bug in ExpandNmakeSourcePaths() or SourceFileRelativePath(). These functions were newly introduced as part of this change. The underlying issue lies with NMAKE: it suppresses command echoing, so the MSVC compiler output does not include the full path of the source file being compiled. As a result, for namesake source files (for example, AmdSev.c located in different directories), successive occurrences overwrite earlier ones. Consequently, the dependency file ends up being generated only for the last occurrence of the namesake file. ExpandNmakeSourcePaths() and SourceFileRelativePath() were added specifically to map each basename occurrence in the output to its resolved full path, allowing CreateDepsFileForMsvc to operate on full‑path keys.

Thank you - I see (more or less...) the issue you are describing. But the commit message needs updating.

You should not say "This issue was caused by ExpandNmakeSourcePaths() and SourceFileRelativePath() using only the base filename" as it is not correct that the issue **was caused by** anything to do with ExpandNmakeSourcePaths() and SourceFileRelativePath(), since you just added them. Do you understand the issue I am trying to raise? It is not that your fix is wrong, but that the commit description of it is wrong/misleading (for people now or in future trying to understand the change).

You have just provided a lot of helpful info to me about how and why this issue happens. Please can you add this info (instead of the wording I noted above) into the commit message.

It would help to place a short (1-3 line) comment into the py code as well to clarify what is happening.

Kowsik-003:
@mikebeaton 
Thank you for the inputs, I've added some comments and changes in commit description.
Could you please review it.

mikebeaton:
> @mikebeaton Thank you for the inputs, I've added some comments and changes in commit description. Could you please review it.

Thank you for the source code changes. The commit message has not changed though, please update. Thanks.

Kowsik-003:
> > @mikebeaton Thank you for the inputs, I've added some comments and changes in commit description. Could you please review it.
> 
> Thank you for the source code changes. The commit message has not changed though, please update. Thanks.

The commit message is updated could you please review it.

mikebeaton:
Could you add this line at the end of the second paragraph of your commit message:

"In non silent build NMAKE echoes the command and the full path so deps generation is correct even for duplicate basefile names."

Kowsik-003:
> Could you add this line at the end of the second paragraph of your commit message:
> 
> "In non silent build NMAKE echoes the command and the full path so deps generation is correct even for duplicate basefile names."

@mikebeaton 
Updated.

mikebeaton:
After the changes I requested it is clearer to me now what the author of the PR has done, and why - but I don't know if anybody who works on Windows build more than I do could offer an opinion as to whether the changes make sense? @os-d @mdkinney

Kowsik-003:
> After the changes I requested it is clearer to me now what the author of the PR has done, and why - but I don't know if anybody who works on Windows build more than I do could offer an opinion as to whether the changes make sense? @os-d @mdkinney
@mikebeaton 

Ok, Thank you for the insights but in patchcheck is getting failed even though there is no trailing whitespaces.

mikebeaton:
> Ok, Thank you for the insights but in patchcheck is getting failed even though there is no trailing whitespaces.

I don't know if I offered any insights. Just trying to get the PR described clearly. Thanks for the updates.

Kowsik-003:
@mikebeaton @os-d @mdkinney 
Any updates regarding this PR.

os-d:
Sorry, I missed the @ last week. I don't have opinions on this. A mild question: instead of adding this extra logic, if we already have ways (seemingly) of doing silent or not silent nmake builds, can we just use that instead? i.e. most of the time you don't need this logic. In the rare case you do, can you just do a non-silent nmake build and get all the info you need instead of special casing this?

Kowsik-003:
> Sorry, I missed the @ last week. I don't have opinions on this. A mild question: instead of adding this extra logic, if we already have ways (seemingly) of doing silent or not silent nmake builds, can we just use that instead? i.e. most of the time you don't need this logic. In the rare case you do, can you just do a non-silent nmake build and get all the info you need instead of special casing this?

@os-d @mikebeaton @mdkinney 
edk2 supports silent build, there is a bug in it, users use silent build.
Hence this bug has to be fixed.

os-d:
> > Sorry, I missed the @ last week. I don't have opinions on this. A mild question: instead of adding this extra logic, if we already have ways (seemingly) of doing silent or not silent nmake builds, can we just use that instead? i.e. most of the time you don't need this logic. In the rare case you do, can you just do a non-silent nmake build and get all the info you need instead of special casing this?
> 
> @os-d @mikebeaton @mdkinney edk2 supports silent build, there is a bug in it, users use silent build. Hence this bug has to be fixed.

I don't think this is a bug, right? It sounds like the scenario is exactly as intended: a silent build is run, so it doesn't produce the full output. When you don't have the full output, it is hard to tell why a failure occurred.

So, I go back to my original question: can you not just run a non-silent build and get all the information? That seems much preferable to having this custom logic added and gets all of nmake's output, not just this part.

I'm not sure if the non-silent builds are already piped through to be able to run easily, but if not, then I would suggest adding the ability to do so. I don't see the -s in the NMAKE flags in tools_def.template, are you manually adding it? If not, where does it get added?

mikebeaton:
> > > Sorry, I missed the @ last week. I don't have opinions on this. A mild question: instead of adding this extra logic, if we already have ways (seemingly) of doing silent or not silent nmake builds, can we just use that instead? i.e. most of the time you don't need this logic. In the rare case you do, can you just do a non-silent nmake build and get all the info you need instead of special casing this?
> > 
> > 
> > @os-d @mikebeaton @mdkinney edk2 supports silent build, there is a bug in it, users use silent build. Hence this bug has to be fixed.
> 
> I don't think this is a bug, right? It sounds like the scenario is exactly as intended: a silent build is run, so it doesn't produce the full output. When you don't have the full output, it is hard to tell why a failure occurred.

@os-d - I understood the author to be saying that some aspect of dependency generation (which, again, I'm understanding to be a part of 'how the build actually builds', not just a build text log) was not working correctly for duplicate files in a silent build. Does that make any sense?

os-d:
> > > > Sorry, I missed the @ last week. I don't have opinions on this. A mild question: instead of adding this extra logic, if we already have ways (seemingly) of doing silent or not silent nmake builds, can we just use that instead? i.e. most of the time you don't need this logic. In the rare case you do, can you just do a non-silent nmake build and get all the info you need instead of special casing this?
> > > 
> > > 
> > > @os-d @mikebeaton @mdkinney edk2 supports silent build, there is a bug in it, users use silent build. Hence this bug has to be fixed.
> > 
> > 
> > I don't think this is a bug, right? It sounds like the scenario is exactly as intended: a silent build is run, so it doesn't produce the full output. When you don't have the full output, it is hard to tell why a failure occurred.
> 
> @os-d - I understood the author to be saying that some aspect of dependency generation (which, again, I'm understanding to be a part of 'how the build actually builds', not just a build text log) was not working correctly for duplicate files in a silent build. Does that make any sense?

Ah, that's make sense, appreciate the explanation. I had misread the first iteration of the PR description and thought a build was failing and it couldn't be identified which file was failing. This seems reasonable to me.

Kowsik-003:
> > > > > Sorry, I missed the @ last week. I don't have opinions on this. A mild question: instead of adding this extra logic, if we already have ways (seemingly) of doing silent or not silent nmake builds, can we just use that instead? i.e. most of the time you don't need this logic. In the rare case you do, can you just do a non-silent nmake build and get all the info you need instead of special casing this?
> > > > 
> > > > 
> > > > @os-d @mikebeaton @mdkinney edk2 supports silent build, there is a bug in it, users use silent build. Hence this bug has to be fixed.
> > > 
> > > 
> > > I don't think this is a bug, right? It sounds like the scenario is exactly as intended: a silent build is run, so it doesn't produce the full output. When you don't have the full output, it is hard to tell why a failure occurred.
> > 
> > 
> > @os-d - I understood the author to be saying that some aspect of dependency generation (which, again, I'm understanding to be a part of 'how the build actually builds', not just a build text log) was not working correctly for duplicate files in a silent build. Does that make any sense?
> 
> Ah, that's make sense, appreciate the explanation. I had misread the first iteration of the PR description and thought a build was failing and it couldn't be identified which file was failing. This seems reasonable to me.

@mikebeaton @os-d 
Please let me know if any further explanation is required or whether this PR can be merged.

os-d:
> @mikebeaton @os-d Please let me know if any further explanation is required or whether this PR can be merged.

Neither of us is a BaseTools maintainer (@mikebeaton  is a reviewer, but a very active one, perhaps you’d like to take on a maintainer role here? :). So you need approval from @lgao4 or @gapalomi . Then the additional consideration is that we are in the soft freeze and whether this should be merged now or after.

mikebeaton:
> > @mikebeaton @os-d Please let me know if any further explanation is required or whether this PR can be merged.
> 
> Neither of us is a BaseTools maintainer (@mikebeaton is a reviewer, but a very active one, perhaps you’d like to take on a maintainer role here? :). So you need approval from @lgao4 or @gapalomi . Then the additional consideration is that we are in the soft freeze and whether this should be merged now or after.

I think definitely after. I do not know this area of the code well enough to agree or disagree to merge, without spending considerably more time investigating. I was hoping someone who uses these builds would be able to see straight away whether these changes made sense or not.

mikebeaton:
Hi @Kowsik-003 -

I did some further tests. I can confirm that the issue is real, and that the PR goes some way to fixing it.

When I tested the patch with `UefiCpuPkg/UefiCpuPkg.dsc`, `-a X64 -b DEBUG -t VS2019`, it does restore creation of both `OUTPUT/AmdSev.c.deps` and `OUTPUT/X64/AmdSev.c.deps` in silent mode (without the patch  `OUTPUT/X64/AmdSev.c.deps` is missing in silent mode, in the Build output for both DxeMpInitLib and PeiMpInitLib).

However, the generated dependency contents do not exactly match the non-silent build. In particular, `AmdSvsmLib.h` is attributed to `OUTPUT/AmdSev.c.deps` in the patched silent build, while the non-silent build attributes it to `OUTPUT/X64/AmdSev.c.deps`. That would seem to need a further fix.

Additionally, I'd still like someone who knows the Windows build system better than me (I usually use/work on macOS or Linux) to confirm whether they can see a cleaner approach to the fix, but I do agree a fix is needed.

mikebeaton (changes requested):

Kowsik-003:
> Hi @Kowsik-003 -
> 
> I did some further tests. I can confirm that the issue is real, and that the PR goes some way to fixing it.
> 
> When I tested the patch with `UefiCpuPkg/UefiCpuPkg.dsc`, `-a X64 -b DEBUG -t VS2019`, it does restore creation of both `OUTPUT/AmdSev.c.deps` and `OUTPUT/X64/AmdSev.c.deps` in silent mode (without the patch `OUTPUT/X64/AmdSev.c.deps` is missing in silent mode, in the Build output for both DxeMpInitLib and PeiMpInitLib).
> 
> However, the generated dependency contents do not exactly match the non-silent build. In particular, `AmdSvsmLib.h` is attributed to `OUTPUT/AmdSev.c.deps` in the patched silent build, while the non-silent build attributes it to `OUTPUT/X64/AmdSev.c.deps`. That would seem to need a further fix.
> 
> Additionally, I'd still like someone who knows the Windows build system better than me (I usually use/work on macOS or Linux) to confirm whether they can see a cleaner approach to the fix, but I do agree a fix is needed.

Hi @mikebeaton 

You are correct, and the root cause is an ordering assumption in the current fix. The namesake_queue is built from module_autogen.SourceFileList in its list iteration order, and the fix assumes the compiler processes namesake files in that same order. The non-silent build (where compiler command lines are echoed) shows the OUTPUT/X64/ variant is compiled second, but our queue may position it first, causing AmdSvsmLib.h (which belongs to the X64 variant) to be incorrectly attributed to the first queued entry (OUTPUT/AmdSev.c.deps) instead.

A further fix would be to cross-reference the queue order against the order in which sources appear in the generated Makefile (via GenMake.py/TargetFileList. That would ensure the silent-mode queue matches the actual compiler invocation order.

Screenshot 2026-05-19 183536

Steps to reproduce:
Please use the changes included in IncludeAutoGen.py in master source.
In edk2 source location open Command Prompt and enter edksetup.bat.

Use this command for silent:
build -a X64 -b DEBUG -p UefiCpuPkg/UefiCpuPkg.dsc -t VS2022 -s

Non-silent:
build -a X64 -b DEBUG -p UefiCpuPkg/UefiCpuPkg.dsc -t VS2022 

Compare the build folder.

Kowsik-003:
@gapalomi @lgao4 @ponchofigueroa @YuweiChen1110 @mikebeaton 
Could you please review this.

lgao4:
> > Hi @Kowsik-003 -
> > I did some further tests. I can confirm that the issue is real, and that the PR goes some way to fixing it.
> > When I tested the patch with `UefiCpuPkg/UefiCpuPkg.dsc`, `-a X64 -b DEBUG -t VS2019`, it does restore creation of both `OUTPUT/AmdSev.c.deps` and `OUTPUT/X64/AmdSev.c.deps` in silent mode (without the patch `OUTPUT/X64/AmdSev.c.deps` is missing in silent mode, in the Build output for both DxeMpInitLib and PeiMpInitLib).
> > However, the generated dependency contents do not exactly match the non-silent build. In particular, `AmdSvsmLib.h` is attributed to `OUTPUT/AmdSev.c.deps` in the patched silent build, while the non-silent build attributes it to `OUTPUT/X64/AmdSev.c.deps`. That would seem to need a further fix.
> > Additionally, I'd still like someone who knows the Windows build system better than me (I usually use/work on macOS or Linux) to confirm whether they can see a cleaner approach to the fix, but I do agree a fix is needed.
> 
> Hi @mikebeaton
> 
> You are correct, and the root cause is an ordering assumption in the current fix. The namesake_queue is built from module_autogen.SourceFileList in its list iteration order, and the fix assumes the compiler processes namesake files in that same order. The non-silent build (where compiler command lines are echoed) shows the OUTPUT/X64/ variant is compiled second, but our queue may position it first, causing AmdSvsmLib.h (which belongs to the X64 variant) to be incorrectly attributed to the first queued entry (OUTPUT/AmdSev.c.deps) instead.
> 
> A further fix would be to cross-reference the queue order against the order in which sources appear in the generated Makefile (via GenMake.py/TargetFileList. That would ensure the silent-mode queue matches the actual compiler invocation order.
> 
> Screenshot 2026-05-19 183536
> Steps to reproduce: Please use the changes included in IncludeAutoGen.py in master source. In edk2 source location open Command Prompt and enter edksetup.bat.
> 
> Use this command for silent: build -a X64 -b DEBUG -p UefiCpuPkg/UefiCpuPkg.dsc -t VS2022 -s
> 
> Non-silent: build -a X64 -b DEBUG -p UefiCpuPkg/UefiCpuPkg.dsc -t VS2022
> 
> Compare the build folder.

I also think current fix is not full fix. It has ordering assumption. What's your next plan? I think another proposal to rename source file name to avoid the same file name.

Kowsik-003:
> > > Hi @Kowsik-003 -
> > > I did some further tests. I can confirm that the issue is real, and that the PR goes some way to fixing it.
> > > When I tested the patch with `UefiCpuPkg/UefiCpuPkg.dsc`, `-a X64 -b DEBUG -t VS2019`, it does restore creation of both `OUTPUT/AmdSev.c.deps` and `OUTPUT/X64/AmdSev.c.deps` in silent mode (without the patch `OUTPUT/X64/AmdSev.c.deps` is missing in silent mode, in the Build output for both DxeMpInitLib and PeiMpInitLib).
> > > However, the generated dependency contents do not exactly match the non-silent build. In particular, `AmdSvsmLib.h` is attributed to `OUTPUT/AmdSev.c.deps` in the patched silent build, while the non-silent build attributes it to `OUTPUT/X64/AmdSev.c.deps`. That would seem to need a further fix.
> > > Additionally, I'd still like someone who knows the Windows build system better than me (I usually use/work on macOS or Linux) to confirm whether they can see a cleaner approach to the fix, but I do agree a fix is needed.
> > 
> > 
> > Hi @mikebeaton
> > You are correct, and the root cause is an ordering assumption in the current fix. The namesake_queue is built from module_autogen.SourceFileList in its list iteration order, and the fix assumes the compiler processes namesake files in that same order. The non-silent build (where compiler command lines are echoed) shows the OUTPUT/X64/ variant is compiled second, but our queue may position it first, causing AmdSvsmLib.h (which belongs to the X64 variant) to be incorrectly attributed to the first queued entry (OUTPUT/AmdSev.c.deps) instead.
> > A further fix would be to cross-reference the queue order against the order in which sources appear in the generated Makefile (via GenMake.py/TargetFileList. That would ensure the silent-mode queue matches the actual compiler invocation order.
> > Screenshot 2026-05-19 183536
> > Steps to reproduce: Please use the changes included in IncludeAutoGen.py in master source. In edk2 source location open Command Prompt and enter edksetup.bat.
> > Use this command for silent: build -a X64 -b DEBUG -p UefiCpuPkg/UefiCpuPkg.dsc -t VS2022 -s
> > Non-silent: build -a X64 -b DEBUG -p UefiCpuPkg/UefiCpuPkg.dsc -t VS2022
> > Compare the build folder.
> 
> I also think current fix is not full fix. It has ordering assumption. What's your next plan? I think another proposal to rename source file name to avoid the same file name.

@lgao4 
1:
The latest changes that were already pushed is working fine. 
I have performed unit test, and both the silent and non-silent build folders are identical. (Unit test details [above](https://github.com/tianocore/edk2/pull/12370#issuecomment-4488063154))
If you are still facing any issues, please point out the bug, and I will take a look. 
2:
No, renaming source file name is not ok for me.
we cannot restrict users to follow a source file naming convention.
At present we can rename it but it will be difficult to enforce / review in the future.

lgao4:
> > > > Hi @Kowsik-003 -
> > > > I did some further tests. I can confirm that the issue is real, and that the PR goes some way to fixing it.
> > > > When I tested the patch with `UefiCpuPkg/UefiCpuPkg.dsc`, `-a X64 -b DEBUG -t VS2019`, it does restore creation of both `OUTPUT/AmdSev.c.deps` and `OUTPUT/X64/AmdSev.c.deps` in silent mode (without the patch `OUTPUT/X64/AmdSev.c.deps` is missing in silent mode, in the Build output for both DxeMpInitLib and PeiMpInitLib).
> > > > However, the generated dependency contents do not exactly match the non-silent build. In particular, `AmdSvsmLib.h` is attributed to `OUTPUT/AmdSev.c.deps` in the patched silent build, while the non-silent build attributes it to `OUTPUT/X64/AmdSev.c.deps`. That would seem to need a further fix.
> > > > Additionally, I'd still like someone who knows the Windows build system better than me (I usually use/work on macOS or Linux) to confirm whether they can see a cleaner approach to the fix, but I do agree a fix is needed.
> > > 
> > > 
> > > Hi @mikebeaton
> > > You are correct, and the root cause is an ordering assumption in the current fix. The namesake_queue is built from module_autogen.SourceFileList in its list iteration order, and the fix assumes the compiler processes namesake files in that same order. The non-silent build (where compiler command lines are echoed) shows the OUTPUT/X64/ variant is compiled second, but our queue may position it first, causing AmdSvsmLib.h (which belongs to the X64 variant) to be incorrectly attributed to the first queued entry (OUTPUT/AmdSev.c.deps) instead.
> > > A further fix would be to cross-reference the queue order against the order in which sources appear in the generated Makefile (via GenMake.py/TargetFileList. That would ensure the silent-mode queue matches the actual compiler invocation order.
> > > Screenshot 2026-05-19 183536
> > > Steps to reproduce: Please use the changes included in IncludeAutoGen.py in master source. In edk2 source location open Command Prompt and enter edksetup.bat.
> > > Use this command for silent: build -a X64 -b DEBUG -p UefiCpuPkg/UefiCpuPkg.dsc -t VS2022 -s
> > > Non-silent: build -a X64 -b DEBUG -p UefiCpuPkg/UefiCpuPkg.dsc -t VS2022
> > > Compare the build folder.
> > 
> > 
> > I also think current fix is not full fix. It has ordering assumption. What's your next plan? I think another proposal to rename source file name to avoid the same file name.
> 
> @lgao4 1: The latest changes that were already pushed is working fine. I have performed unit test, and both the silent and non-silent build folders are identical. (Unit test details [above](https://github.com/tianocore/edk2/pull/12370#issuecomment-4488063154)) If you are still facing any issues, please point out the bug, and I will take a look. 2: No, renaming source file name is not ok for me. we cannot restrict users to follow a source file naming convention. At present we can rename it but it will be difficult to enforce / review in the future.

How do you avoid the ordering issue?

Kowsik-003:
> > > > > Hi @Kowsik-003 -
> > > > > I did some further tests. I can confirm that the issue is real, and that the PR goes some way to fixing it.
> > > > > When I tested the patch with `UefiCpuPkg/UefiCpuPkg.dsc`, `-a X64 -b DEBUG -t VS2019`, it does restore creation of both `OUTPUT/AmdSev.c.deps` and `OUTPUT/X64/AmdSev.c.deps` in silent mode (without the patch `OUTPUT/X64/AmdSev.c.deps` is missing in silent mode, in the Build output for both DxeMpInitLib and PeiMpInitLib).
> > > > > However, the generated dependency contents do not exactly match the non-silent build. In particular, `AmdSvsmLib.h` is attributed to `OUTPUT/AmdSev.c.deps` in the patched silent build, while the non-silent build attributes it to `OUTPUT/X64/AmdSev.c.deps`. That would seem to need a further fix.
> > > > > Additionally, I'd still like someone who knows the Windows build system better than me (I usually use/work on macOS or Linux) to confirm whether they can see a cleaner approach to the fix, but I do agree a fix is needed.
> > > > 
> > > > 
> > > > Hi @mikebeaton
> > > > You are correct, and the root cause is an ordering assumption in the current fix. The namesake_queue is built from module_autogen.SourceFileList in its list iteration order, and the fix assumes the compiler processes namesake files in that same order. The non-silent build (where compiler command lines are echoed) shows the OUTPUT/X64/ variant is compiled second, but our queue may position it first, causing AmdSvsmLib.h (which belongs to the X64 variant) to be incorrectly attributed to the first queued entry (OUTPUT/AmdSev.c.deps) instead.
> > > > A further fix would be to cross-reference the queue order against the order in which sources appear in the generated Makefile (via GenMake.py/TargetFileList. That would ensure the silent-mode queue matches the actual compiler invocation order.
> > > > Screenshot 2026-05-19 183536
> > > > Steps to reproduce: Please use the changes included in IncludeAutoGen.py in master source. In edk2 source location open Command Prompt and enter edksetup.bat.
> > > > Use this command for silent: build -a X64 -b DEBUG -p UefiCpuPkg/UefiCpuPkg.dsc -t VS2022 -s
> > > > Non-silent: build -a X64 -b DEBUG -p UefiCpuPkg/UefiCpuPkg.dsc -t VS2022
> > > > Compare the build folder.
> > > 
> > > 
> > > I also think current fix is not full fix. It has ordering assumption. What's your next plan? I think another proposal to rename source file name to avoid the same file name.
> > 
> > 
> > @lgao4 1: The latest changes that were already pushed is working fine. I have performed unit test, and both the silent and non-silent build folders are identical. (Unit test details [above](https://github.com/tianocore/edk2/pull/12370#issuecomment-4488063154)) If you are still facing any issues, please point out the bug, and I will take a look. 2: No, renaming source file name is not ok for me. we cannot restrict users to follow a source file naming convention. At present we can rename it but it will be difficult to enforce / review in the future.
> 
> How do you avoid the ordering issue?

@lgao4 
Existing [FileListMacros](https://github.com/tianocore/edk2/blob/1f28f1e8198c7d97c6a080e7aa34359d4dddac62/BaseTools/Source/Python/AutoGen/GenMake.py#L1060) has OBJLIST to be updated in build time generated Makefile(UefiCpuPkg). It was already sorted using sort(). 

image

In modified [TargetFileList](https://github.com/Kowsik-003/edk2/blob/00207ac32098c4abb5adf17b2ea43d38459d86f6/BaseTools/Source/Python/AutoGen/IncludesAutoGen.py#L202) changes inf's source file list order is also sorted (originally it was not sorted), so the order of the dependency generation will be identical between Makefile OBJLIST and .deps generation list.

lgao4:
This change is to fix the issue in silent build. Has it any impact on the normal build?

Kowsik-003:
> This change is to fix the issue in silent build. Has it any impact on the normal build?

@lgao4 
No, this will not impact the non-silent build.
The variable [cc_cmd_in_output](https://github.com/Kowsik-003/edk2/blob/00207ac32098c4abb5adf17b2ea43d38459d86f6/BaseTools/Source/Python/AutoGen/IncludesAutoGen.py#L217) is the key here. 
In a non-silent build, it becomes True, whereas in a silent build, it remains False.

lgao4 (approved):

lgao4:
> > This change is to fix the issue in silent build. Has it any impact on the normal build?
> 
> @lgao4 No, this will not impact the non-silent build. The variable [cc_cmd_in_output](https://github.com/Kowsik-003/edk2/blob/00207ac32098c4abb5adf17b2ea43d38459d86f6/BaseTools/Source/Python/AutoGen/IncludesAutoGen.py#L217) is the key here. In a non-silent build, it becomes True, whereas in a silent build, it remains False.

OK. I have no other comment.

Kowsik-003:
> > > This change is to fix the issue in silent build. Has it any impact on the normal build?
> > 
> > 
> > @lgao4 No, this will not impact the non-silent build. The variable [cc_cmd_in_output](https://github.com/Kowsik-003/edk2/blob/00207ac32098c4abb5adf17b2ea43d38459d86f6/BaseTools/Source/Python/AutoGen/IncludesAutoGen.py#L217) is the key here. In a non-silent build, it becomes True, whereas in a silent build, it remains False.
> 
> OK. I have no other comment.

@lgao4 
Thank you.

Kowsik-003:
@mikebeaton 
Could please provide some friendly comment on this PR.

Kowsik-003 (commented):
Hi @mikebeaton 
Please help to review and merge the changes as well>
refer: https://github.com/tianocore/edk2/pull/12370#issuecomment-4488063154

Kowsik-003:
Hi @mikebeaton 
Can you please provide some comments on this PR, and the requested changes are done.

Kowsik-003:
Hi @mikebeaton ,
The requested changes are done could you please give a review or a comment on your side.

Kowsik-003:
Hi @mikebeaton
I've done the requested changes its working completely fine could you please provide some comments.

Kowsik-003:
Hi @lgao4,
Before the changes were approved by you, @mikebeaton requested some changes to this PR. I have made those changes, and you have reviewed and approved them. However, since there is "1 requested change", this PR is currently on hold and has not yet been merged. Additionally, Mikebeaton has been removed from the maintainer list [(12783)](https://github.com/tianocore/edk2/pull/12783). So could you kindly help merge this PR.

mikebeaton (approved):

mikebeaton:
Sorry for the delay.

PR #12373, opened 2026-03-31
BaseTools: Only Override FAMILY with TOOLCHAIN per BuildOptions Section

os-d:
# Description

Currently, all BuildOptions sections from all sources are merged together and then precedence is determined by choosing TOOLCHAIN matches over FAMILY matches and all FAMILY options are ignored.

This means that including a dsc.inc with a BuildOptions section that specifies a TOOLCHAIN instead of FAMILY will break a DSC that specifies by FAMILY.

This fixes this behavior by evaluating precedence rules per BuildOptions section, not on the combined section.

This is not an issue in INFs because they do not allow including other files.

Note: This change does not change the semantics that if multiple BuildOptions sections exist within a given DSC, DSC.inc, or INF, they will get merged and the precedence rules will take affect. This only helps the precedence to be done correctly within a given file before merging with another file.

An example of the pattern that was broken before this change:

```inf
# Some.dsc.inc

[BuildOptions]
  CLANGPDB:*_*_*_CC_FLAGS = -D SOMETHING
```

```inf
# Platform.dsc

!include Some.dsc.inc

...

[BuildOptions]
  GCC:*_*_*_CC_FLAGS = -D SOMETHING_ELSE
```

The merged build option precedence taking caused the `Platform.dsc` option to be dropped only `-D SOMETHING` gets applied. After this change, both `-D SOMETHING and -D SOMETHING_ELSE` are added to the `CC_FLAGS`.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested with various versions of BuildOptions in MdeLibs.dsc.inc using `CLANGPDB:` vs `GCC:` in OVMF DSCs. I also tested having the below:

```inf
!if $(TOOL_CHAIN_TAG) == CLANGPDB
  [BuildOptions]
    CLANGPDB:*_*_*_CC_FLAGS = -D SOMETHING
!endif
```

 The failures observed before these changes are not observed after. The conditional inclusion of the build options section works; however as noted above if this is in a file that has another BuildOptions section, the two will get merged together.

I am not a BaseTools nor Python expert, so additional testing would be greatly appreciated on this. There are a lot of corner cases that I easily could have missed testing.

## Integration Instructions

N/A. It is possible this fixes some issues or uncovers some hidden incompatibilities, but should match the original author intent now.

### DIFF
--- a/BaseTools/Source/Python/AutoGen/PlatformAutoGen.py
+++ b/BaseTools/Source/Python/AutoGen/PlatformAutoGen.py
@@ -12,7 +12,7 @@ from __future__ import print_function
 from __future__ import absolute_import
 import os.path as path
 import copy
-from collections import defaultdict
+from collections import defaultdict, OrderedDict
 
 from .BuildEngine import BuildRule,gDefaultBuildRuleFile,AutoGenReqBuildRuleVerNum
 from .GenVar import VariableMgr, var_info
@@ -984,15 +984,15 @@ class PlatformAutoGen(AutoGen):
     ## Return the build options specific for all modules in this platform
     @cached_property
     def BuildOption(self):
-        return self._ExpandBuildOption(self.Platform.BuildOptions)
+        return self._ExpandGroupedBuildOption(self.Platform.BuildOptions)
 
     def _BuildOptionWithToolDef(self, ToolDef):
-        return self._ExpandBuildOption(self.Platform.BuildOptions, ToolDef=ToolDef)
+        return self._ExpandGroupedBuildOption(self.Platform.BuildOptions, ToolDef=ToolDef)
 
     ## Return the build options specific for EDKII modules in this platform
     @cached_property
     def EdkIIBuildOption(self):
-        return self._ExpandBuildOption(self.Platform.BuildOptions, EDKII_NAME)
+        return self._ExpandGroupedBuildOption(self.Platform.BuildOptions, EDKII_NAME)
 
     ## Parse build_rule.txt in Conf Directory.
     #
@@ -1445,6 +1445,36 @@ class PlatformAutoGen(AutoGen):
                 for guid,mpath in Guid_Path:
                     retVal[(name,mpath)] = '%s_%s' % (name,guid)
         return retVal
+
+    ## Expand build options from grouped sections independently.
+    #
+    #  The BUILDRULEFAMILY vs FAMILY matching in _ExpandBuildOption is
+    #  designed so that within a single [BuildOptions] section,
+    #  BUILDRULEFAMILY takes precedence over FAMILY. When multiple
+    #  [BuildOptions] sections are merged (e.g. from !include files),
+    #  each section must be evaluated independently to avoid a
+    #  BUILDRULEFAMILY match in one section suppressing FAMILY matches
+    #  in another.
+    #
+    def _ExpandGroupedBuildOption(self, OptionGroups, ModuleStyle=None, ToolDef=None):
+        Merged = {}
+        for Options in OptionGroups:
+            # Pass a copy since _ExpandBuildOption may mutate the dictionary
+            Result = self._ExpandBuildOption(OrderedDict(Options), ModuleStyle, ToolDef)
+            for Tool in Result:
+                if Tool not in Merged:
+                    Merged[Tool] = {}
+                for Attr in Result[Tool]:
+                    Value = Result[Tool][Attr]
+                    if Attr != "FLAGS" or Attr not in Merged[Tool] or Value.startswith('='):
+                        Merged[Tool][Attr] = Value
+                    else:
+                        if Attr != 'PATH':
+                            Merged[Tool][Attr] += " " + Value
+                        else:
+                            Merged[Tool][Attr] = Value
+        return Merged
+
     ## Expand * in build option key
     #
     #   @param  Options     Options to be expanded
--- a/BaseTools/Source/Python/GenFds/GenFdsGlobalVariable.py
+++ b/BaseTools/Source/Python/GenFds/GenFdsGlobalVariable.py
@@ -950,12 +950,17 @@ def FindExtendTool(KeyStringList, CurrentArchList, NameGuid):
         MatchItem = None
         MatchPathItem = None
         MatchOptionsItem = None
+        # Flatten grouped BuildOptions into a single dict for key lookup
+        FlatBuildOptions = {}
         for KeyString in KeyStringList:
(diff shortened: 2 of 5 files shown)


os-d:
@mdkinney this fixed the issue I discovered in the CLANGPB stack check lib PR. Can you help give this a close look and any additional testing you have bandwidth for?

It fixes the scenarios I encountered, but I'm not confident enough in my knowledge of the BaseTools architecture that it completely covers all cases that could be hit.

As I noted in the description, what it does not fix is if you have multiple BuildOptions sections within a given file. The precedence will still be on the merged group. But it no longer overrides precedence across files, which is the more common case and dangerous.

mdkinney:
> @mdkinney this fixed the issue I discovered in the CLANGPB stack check lib PR. Can you help give this a close look and any additional testing you have bandwidth for?
> 
> It fixes the scenarios I encountered, but I'm not confident enough in my knowledge of the BaseTools architecture that it completely covers all cases that could be hit.
> 
> As I noted in the description, what it does not fix is if you have multiple BuildOptions sections within a given file. The precedence will still be on the merged group. But it no longer overrides precedence across files, which is the more common case and dangerous.

@osd This breaks the EDK II Build Spec rules. I think BaseTools should provide consistent behavior no matter if there are multiple [BuildOptions] sections in one file or across included files.  DSC includes and changing include boundaries must never change behavior of a build.

os-d:
> > @mdkinney this fixed the issue I discovered in the CLANGPB stack check lib PR. Can you help give this a close look and any additional testing you have bandwidth for?
> > It fixes the scenarios I encountered, but I'm not confident enough in my knowledge of the BaseTools architecture that it completely covers all cases that could be hit.
> > As I noted in the description, what it does not fix is if you have multiple BuildOptions sections within a given file. The precedence will still be on the merged group. But it no longer overrides precedence across files, which is the more common case and dangerous.
> 
> @osd This breaks the EDK II Build Spec rules. I think BaseTools should provide consistent behavior no matter if there are multiple [BuildOptions] sections in one file or across included files. DSC includes and changing include boundaries must never change behavior of a build.

Sure. Right now BaseTools is breaking the EDK II Build Spec rules, though. Because DSC includes and changing include boundaries will change the behavior of a build.

This change fixes one of those cases, not both. I'll take a look at what it would take to fix the second case as well. I may have thought of a way to handle it, because the record does contain LineNo, so maybe I can have distinct IDs for that. Let me test.

os-d:
@mdkinney I did fix the second issue as well, where precedence is evaluated per BuildOptions section in a given DSC/dsc.inc before they are merged.

However, I took a look at the DSC spec (and also the INF spec to see if the same issue was present there; it doesn't have includes, but can have multiple BuildOption sections). Both of these specs explicitly say that all BuildOptions sections will be merged.

[INF Spec 1.27](https://tianocore-docs.github.io/edk2-InfSpecification/release-1.27/edk2-InfSpecification-release-1.27.pdf), see section 3.1 General Rules:

image

[DSC Spec 1.28](https://tianocore-docs.github.io/edk2-DscSpecification/release-1.28/edk2-DscSpecification-release-1.28.pdf) see section 2.2.1 Section Entries:

image

So to change this behavior, perhaps we need INF/DSC spec updates? I don't think the current behavior is intentional or intuitive. But as it currently reads, the specs seem to envision that all sections (implicitly indicating this for included dsc files) are merged and that any precedence would be applied to a merged section.

However, I also don't see anywhere in the DSC spec that defines using anything other than a FAMILY for the part before the `:` in a BuildOption section:

image

And I don't see anything then about precedence of a BUILDRULEFAMILY (or TOOLCHAIN) over a FAMILY here. So I think in general BaseTools has moved beyond the DSC spec; clearly edk2 has taken dependence on this behavior. Perhaps I am missing where this is documented.

The only thing mentioning precedence for BuildOptions is:

image

Which doesn't envision a matching line not being taken.

I also took a look through the main EDK II Build spec and didn't see anything defining the precedence or using anything other than a FAMILY there.

So, what do you want to do here? I would propose we move forward with these changes that fix the current non-intuitive behavior that is breaking platforms. We can update the specs to reflect the current state of edk2 afterwards, since they are already out of date. I can take a look at what it would take for the INF parser to also support this, presumably it would be similar, though I don't know if it collects the same data as the DSC parser.

I have pushed up the changes that fix the two current issues in the DSC parser, so you can take a look if you want. I tested a variety of other scenarios like

```inf
# MdeLibs.dsc.inc

[BuildOptions]
  CLANGPDB:*_*_*_CC_FLAGS = -D SOMETHING_ELSE
  GCC:*_*_*_CC_FLAGS = -D NOT_THIS

[BuildOptions]
  GCC:*_*_*_CC_FLAGS = -D SOMETHING_ELSE_ENTIRELY
```
```inf
# IntelTxdX64.dsc
[BuildOptions]
  GCC:*_*_*_CC_FLAGS = -D TDX_PEI_LESS_BOOT # Critical, don't link without

[BuildOptions]
  CLANGPDB:*_*_*_CC_FLAGS = -D SOMETHING

[Components]
  MdeModulePkg/Core/Dxe/DxeMain.inf {
    <LibraryClasses>
      NULL|MdeModulePkg/Library/LzmaCustomDecompressLib/LzmaCustomDecompressLib.inf
      DevicePathLib|MdePkg/Library/UefiDevicePathLib/UefiDevicePathLib.inf
   <BuildOptions>
     GCC:*_*_*_CC_FLAGS = -D A_DEFINE_HERE
  }
```

I saw the expected `-D SOMETHING_ELSE -D SOMETHING_ELSE_ENTIRELY -D TDX_PEI_LESS_BOOT -D A_DEFINE_HERE` with these changes.

os-d:
Not stale, awaiting further discussion, but lower priority than other PRs currently being discussed. Will return to it after PEI memory bins have been merged.

os-d:
@mdkinney thanks for all the reviews on PEI memory bins. Now that that is approved, can we revisit this? I have followed your example in the XIP PR and added a BaseTools test file for this. I also put up a PR to amend the DSC spec to match what edk2 is already doing + this bug fix: https://github.com/tianocore-docs/edk2-DscSpecification/pull/14.

os-d:
@mdkinney can we revisit this one?

os-d:
@mdkinney I'd like to get this merged and the edk2 specs updated to reflect the current behavior and fix the errors here. This also unblocks the stack cookie proposal you had. Can you please review?

os-d:
@AshrafAliS @lgao4 @gapalomi can you also review, please?

os-d:
@mdkinney @AshrafAliS @lgao4 @gapalomi this has been open for getting on 6 months, can you review please?

PR #12374, opened 2026-04-01
BaseTools: FMMT: Fix FFS replace no-op in LZMA guided section

AshrafAliS:
# Description

When replacing an FFS whose LZMA recompression delta exactly fits within the existing PadData (Remain_New_Free_Space == 0), the output binary was byte-for-byte identical to the input, the merge appeared to succeed but no change was written.

Root cause: in the 'CompressedData > OriData' branch of 'CompressSectionData', 'TargetTree.Data.OriData' was never updated before calling 'ModifyTest'. With 'Remain_New_Free_Space == 0', 'ModifyTest' falls into the 'else' branch and calls 'CompressData()', which is a no-op because 'Status' is already 'True', leaving the stale LZMA blob in place.

Fix: pre-update 'OriData', 'ChangeSize', and 'PadData' when 'Remain_New_Free_Space <= 0', guarded so the '> 0' path (which relies on 'ModifyTest' to propagate the update up the tree) is unaffected.

| OriData   | CompressedData | Delta | Remain_New_Free_Space | Before Fix                        | After Fix         |
|-----------|----------------|-------|-----------------------|-----------------------------------|-------------------|
| 1,441,373 | 1,441,376      | +3    | 0 (absorbed by Pad)   | no-op: output identical to input  | driver updated    |
| 1,441,373 | 1,441,579      | +206  | +204                  | working: delta propagated up tree | no regression     |

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

**Before fix:** input BIOS and output BIOS were byte-for-byte identical during FMMT replace operation (verified with binary comparison).
**After fix:** the files differ and the updated driver is present in the output image as confirmed by re-parsing output with FMMT.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Source/Python/FMMT/core/FvHandler.py
+++ b/BaseTools/Source/Python/FMMT/core/FvHandler.py
@@ -259,6 +259,17 @@ class FvHandler:
             elif len(CompressedData) > len(TargetTree.Data.OriData):
                 New_Pad_Size = GetPadSize(len(CompressedData), SECTION_COMMON_ALIGNMENT)
                 self.Remain_New_Free_Space = len(CompressedData) + New_Pad_Size - len(TargetTree.Data.OriData) - len(TargetTree.Data.PadData)
+                if self.Remain_New_Free_Space <= 0:
+                    # Update section header size to reflect the new compressed data length.
+                    Size_delta = len(CompressedData) - len(TargetTree.Data.OriData)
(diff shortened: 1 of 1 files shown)


AshrafAliS:
@lgao4 @YuweiChen1110 help to reviw

PR #12377, opened 2026-04-01
OvmfPkg/IgvmSecureBootDxe: Do not return with CustomMode enabled.

kraxel:
# Description

In case some failure happens while CustomMode is enabled call
CpuDeadLoop() instead of returning an error code.  This makes
sure the firmware will not continue running with CustomMode
left enabled unintentionally.

I don't think it is possible to trigger this without also
changing the confidential VM launch measurement.  Nevertheless
it does not hurt to have some extra safeguards.

## How This Was Tested

Build tested only.

## Integration Instructions

N/A

### DIFF
--- a/OvmfPkg/IgvmSecureBootDxe/IgvmSecureBootDxe.c
+++ b/OvmfPkg/IgvmSecureBootDxe/IgvmSecureBootDxe.c
@@ -116,35 +116,6 @@ IgvmSecureBootSetVariable (
   return Status;
 }
 
-EFI_STATUS
-EFIAPI
-IgvmSecureBootCustomMode (
-  UINT8  Value
-  )
-{
-  EFI_STATUS  Status;
-
-  Status = gRT->SetVariable (
-                  EFI_CUSTOM_MODE_NAME,
-                  &gEfiCustomModeEnableGuid,
-                  (EFI_VARIABLE_NON_VOLATILE |
-                   EFI_VARIABLE_BOOTSERVICE_ACCESS),
-                  sizeof (Value),
-                  &Value
-                  );
-  if (EFI_ERROR (Status)) {
-    DEBUG ((
-      DEBUG_ERROR,
-      "%a: %a CustomMode: %r\n",
-      __func__,
-      Value ? "enable" : "disable",
-      Status
-      ));
-  }
-
-  return Status;
-}
-
 BOOLEAN
 EFIAPI
 IgvmSecureBootIsConfigured (
@@ -207,6 +178,11 @@ IgvmSecureBootDxeEntrypoint (
     return EFI_UNSUPPORTED;
   }
 
+  if (FeaturePcdGet (PcdRequireSelfSignedPk) != FALSE) {
+    DEBUG ((DEBUG_ERROR, "%a: ERROR: this driver requires PcdRequireSelfSignedPk=FALSE.\n", __func__));
+    return EFI_UNSUPPORTED;
+  }
+
   if (IgvmSecureBootIsConfigured ()) {
     DEBUG ((DEBUG_ERROR, "%a: secure boot is already configured.\n", __func__));
     return EFI_ALREADY_STARTED;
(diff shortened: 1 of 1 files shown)


ardbiesheuvel:
Does CustomMode have to be a non-volatile variable?

OvmfPkg/IgvmSecureBootDxe/IgvmSecureBootDxe.c
```diff
@@ -240,12 +240,12 @@ IgvmSecureBootDxeEntrypoint (
 
   Status = IgvmSecureBootSetVariable (L"PK", &gEfiGlobalVariableGuid, mPK);
   if (EFI_ERROR (Status)) {
-    return Status;
+    CpuDeadLoop ();
```
luigix25: what about adding a log message here?

kraxel: IgvmSecureBootSetVariable does that already.

kraxel:
> Does CustomMode have to be a non-volatile variable?

Not sure if that is a hard requirement.  Other places such as `EnrollDefaultKeys` set it as non-volatile, so I did the same.  IIRC I've also seen uefi firmware on physical hardware which offered a 'turn on custom mode for next boot' option, which can also only work if `CustomMode` is non-volatile.

ardbiesheuvel:
> > Does CustomMode have to be a non-volatile variable?
> 
> Not sure if that is a hard requirement. Other places such as `EnrollDefaultKeys` set it as non-volatile, so I did the same. IIRC I've also seen uefi firmware on physical hardware which offered a 'turn on custom mode for next boot' option, which can also only work if `CustomMode` is non-volatile.

OK, so it /can/ be non-volatile, but it is not required. In that case, wouldn't it make more sense to create it as volatile here, given that it is never the intent to leave it enabled?

kraxel:
> OK, so it /can/ be non-volatile, but it is not required.

Looking at `SecurityPkg/Library/AuthVariableLib/AuthVariableLib.c` I see an entry for `EFI_CUSTOM_MODE_NAME` which mandates non-volatile (`VARIABLE_ATTRIBUTE_NV_BS`), and `MdeModulePkg/Universal/Variable/RuntimeDxe/Variable.c` actually verifies that via `VarCheckLibVariablePropertySet`.

So, I don't think we can just drop the NV flag.

ardbiesheuvel:
> > OK, so it /can/ be non-volatile, but it is not required.
> 
> Looking at `SecurityPkg/Library/AuthVariableLib/AuthVariableLib.c` I see an entry for `EFI_CUSTOM_MODE_NAME` which mandates non-volatile (`VARIABLE_ATTRIBUTE_NV_BS`), and `MdeModulePkg/Universal/Variable/RuntimeDxe/Variable.c` actually verifies that via `VarCheckLibVariablePropertySet`.
> 
> So, I don't think we can just drop the NV flag.

Fair enough.

But do we still need this change? Is there now a use case for building this driver with `gEfiMdeModulePkgTokenSpaceGuid.PcdRequireSelfSignedPk` set to `TRUE` ?

I'm not keen on adding `CpuDeadloop()` calls if we can avoid it.

kraxel:
> > So, I don't think we can just drop the NV flag.
> 
> Fair enough.
> 
> But do we still need this change? Is there now a use case for building this driver with `gEfiMdeModulePkgTokenSpaceGuid.PcdRequireSelfSignedPk` set to `TRUE` ?

I don't see compelling reason.  But I suspect there is some reason why the PCD exists in the first place.  That said I can live with it if you feel like vetoing that change.

ardbiesheuvel:
> > > So, I don't think we can just drop the NV flag.
> > 
> > 
> > Fair enough.
> > But do we still need this change? Is there now a use case for building this driver with `gEfiMdeModulePkgTokenSpaceGuid.PcdRequireSelfSignedPk` set to `TRUE` ?
> 
> I don't see compelling reason. But I suspect there is some reason why the PCD exists in the first place. That said I can live with it if you feel like vetoing that change.

Given that we are using CustomMode only to get around this requirement, I'd prefer it if we would just require `PcdRequireSelfSignedPk` to be FALSE for this driver to load. If someone needs this changed in the future, they can explain why and we can reconsider.

kraxel:
> I'd prefer it if we would just require `PcdRequireSelfSignedPk` to be FALSE for this driver to load.

Yes, that makes sense.
/me goes delete code.

ardbiesheuvel (approved):

PR #12379, opened 2026-04-01
3945_arm_cca_v5_s2: Refactor Kvmtool MMIO memory mapping

samimujawar:
# Refactor Kvmtool MMIO memory mapping

Introduce a common MMIO mapping helper library MemoryMapLib and refactor the ArmVirt Kvmtool guest memory map setup so that only the memory regions required during early boot are mapped by KvmtoolVirtMemInfoLib.

This series adds MemoryMapLib to centralise the process of adding MMIO ranges to the GCD memory map and programming memory attributes. It also enables the library for DXE and runtime drivers, wires it into ArmVirt.dsc.inc, and updates existing users to avoid duplicating AddMemorySpace() and SetMemorySpaceAttributes() handling.

The series also updates the ArmVirt Kvmtool firmware flow so device MMIO ranges are mapped closer to the driver or probe library that needs them. This includes mapping GIC ranges before ArmGicDxe configures the interrupt controller, using MemoryMapLib for RTC and PCI MMIO mappings, mapping PCI MMIO windows in the GCD memory map, and adding an early FDT 16550 serial probe library so the serial port base address can be discovered early enough for mapping the serial port MMIO range to enable debug logging.

Finally, the Kvmtool guest memory map configuration is refactored to map only System Memory, FV, and the Serial Port during early memory-map setup. Other device mappings are expected to be configured by the corresponding DXE driver or probe library.

- [ ] Breaking change?
  - NO
- [ ] Impacts security?
  - NO
- [ ] Includes tests?

## How This Was Tested
Tested by launching a guest using Kvmtool VMM and booting the ArmVirtKvmtool guest firmware. 

## Integration Instructions
NA

---

# PR History
## Introduce a PlatformDeviceInfo lib

Introduce a platform device info library that parses the VMM provided FDT and extracts the MMIO base address
and the range for the devices present in the FDT.

This is useful for populating the memory map for a platform. e.g. the Kvmtool memory map is currently
configured such that the address range outside the DRAM range is considered as peripheral memory.

Although this currently works, starting with the RMM 1.0 - rel0 specification, it is expected
that the firmware queries the RMM to get the IPA state for the MMIO address ranges for the devices
to determine if the shared attribute must be set. If the IPA state is returned as RIPAS device,
the shared attribute must not be set for the device. Otherwise the shared attribute needs to be
set for the MMIO address range covering the device as it is emulated by the VMM.

Therefore, introduce a PlatformDeviceInfo lib to get the MMIO address ranges for the devices presented
to the VM and to populate the memory map accordingly.


OvmfPkg/Include/Library/PlatformDeviceInfoLib.h
```diff
@@ -0,0 +1,55 @@
...
+**/
+
+#pragma once
+
+#include <Base.h>
+#include <Uefi/UefiBaseType.h>
+
+/* Maximum number of platform devices that can be discovered from a FDT.
+*/
+#define MAX_PLAT_DEVICE_COUNT  FixedPcdGet64 (PcdMaxPlatDeviceCount)
```
ardbiesheuvel: Should this be part of the library class's public API?

ardbiesheuvel:
There is already FDT parsing logic/abstraction in EmbeddedPkg, in FdtClientDxe, and ArmVirtKvmTool already incorporates it.

Could you please explain why that is not sufficient, and why fixing it is not an option here?

pierregondois:
Hello Ard,
The `FdtClientDxe` is providing utility functions to parse the DT.
The `PlatformDeviceInfo` contains some information about which nodes are useful (or not). It also contains utility functions to parse the relevant information of the useful nodes.

Just to be sure, you are suggesting to update the `FDT_CLIENT_PROTOCOL` to remove/trim the parsing logic in `PlatformDeviceInfo` ? I.e. most of the content of `OvmfPkg/Library/PlatformDeviceInfoLib/HwInfoParser/FdtUtility.c`.
Is that correct ?

OvmfPkg/Include/Library/PlatformDeviceInfoLib.h
samimujawar: This is later being used in KvmtoolVirtMemInfoLib.c (https://github.com/tianocore/edk2/pull/12379/commits/5ef974c0b7304f4bb938c54bd02ab85c3d4dface) to populate the memory map, as shown below.
   // Number of Virtual Memory Map Descriptors
   #define MAX_VIRTUAL_MEMORY_MAP_DESCRIPTORS  (MAX_PLAT_DEVICE_COUNT + 4)

I can drop the macro and only use the PcdMaxPlatDeviceCount. Please let me know if this would be a preferred approach.

ardbiesheuvel: If clients of the library need to access this information, it should be a library API function, not a \#define

ardbiesheuvel:
> Hello Ard, The `FdtClientDxe` is providing utility functions to parse the DT. The `PlatformDeviceInfo` contains some information about which nodes are useful (or not). It also contains utility functions to parse the relevant information of the useful nodes.
> 

Indeed, so there is duplication, right?

> Just to be sure, you are suggesting to update the `FDT_CLIENT_PROTOCOL` to remove/trim the parsing logic in `PlatformDeviceInfo` ? I.e. most of the content of `OvmfPkg/Library/PlatformDeviceInfoLib/HwInfoParser/FdtUtility.c`. Is that correct ?

You are introducing another, completely separate driver that parses the FDT and exposes the information, and what I am asking is whether those two could be merged in one way or another.

samimujawar:
Hi @ardbiesheuvel The PlatformDeviceInfoLib is used in the very early stages of the firmware boot i.e. in the call to ArmVirtGetMemoryMap(). Since FDT_CLIENT_PROTOCOL is available in the DXE phase we cannot use it in the SEC/PEI phase.

OvmfPkg/Include/Library/PlatformDeviceInfoLib.h
samimujawar: Ack, I will update accordingly.

samimujawar:
@ardbiesheuvel please let me know if you have any further comments or queries for this series.

samimujawar:
Hi Ard (@ardbiesheuvel), Can you let me know your thoughts on this series, please?

leiflindholm:
I can't help but think that this library provides abstraction at the wrong level.

A PlatformInfoLib (the fact the memory map is included means I don't think it should contain the word Device, because it is more than that) feels like a useful thing to have.
Reducing the duplication between platforms of MemoryInitPeim/MemoryInitPeiLib would be a brilliant improvement.

But here we seem to be duplicating DynamicTablesPkg functionality in a platform-specific way, that still requires information to be passed the same way as happens for it.
If we need a simplified subset of DynamicTablesPkg functionality, could that be more cleanly provided by DynamicTablesPkg directly?

Because for a platform-specific library, this looks massively more complex than needed for the functionality provided.

And for a generic platform discovery library, it fails to provide any actual abstraction.

pierregondois:
Hello Leif,
Most ArmVirtGetMemoryMap() implementations seem to do similar things indeed.
To initialize the VirtualMemoryMap:
- Advertise PcdFvBaseAddress/PcdFvSize [KvmTool, CloudHv, Qemu]
- Advertise PcdSystemMemoryBase/PcdSystemMemorySize [KvmTool, CloudHv, Qemu]
Various:
- Advertise other memory nodes [CloudHv]
- Advertise various peripherals [KvmTool for Arm CCA]
- Advertise all peripherals at once with a fixed address MACH_VIRT_PERIPH_BASE [CloudHv, Qemu]
- Advertise TPM with a fixed address TOP_32BIT_DEVICE_BASE [CloudHv]

Note:
- Qemu advertises PcdFvBaseAddress with ARM_MEMORY_REGION_ATTRIBUTE_WRITE_BACK_RO
- Qemu parses multiple memory regions, but only maps the lowest one.

Xen is a bit off as all of the memory space is advertised without any distinction.

------

It seems possible to unify all the ArmVirtGetMemoryMap() implementations, but if we want to have some Platform specific behaviour, we should try to dynamically apply depending on the underlying platform.

We would need to detect the underlying platform from the information coming from the DTB. The issue is that it seems that only QemuVirt advertises at the top level of the DTB:
- compatible = "linux,sbsa-ref"
For kvmtool/CloudHv:
- compatible = "linux,dummy-virt";

------

Kvmtool seems to have the most generic requirement/implementation of ArmVirtGetMemoryMap().
Would it make sense to integrate the PlatformDeviceInfo to KvmtoolVirtMemInfoLib.inf and then try to make PlatformDeviceInfo generic for other virtual platforms ? (or PlatformDeviceInfo -> PlatformInfo if the name is changed)

------

Another point is that the DTB is somehow re-parsed to find some
addresses/interrupts at other places, like in:
- ArmVirtTimerFdtClientLib.inf
- ArmVirtGicArchLib.inf
- NorFlashKvmtoolLib.inf
- ArmVirtPL031FdtClientLib.inf
- KvmtoolRtcFdtClientLib.inf
Wouldn't the code benefit from:
- parsing the DTB once in the SEC/PrePi phase,
- populate a HOB
- when PEI/DXE starts, update the dynamic PCDs with the HOB values if needed
It would allow to avoid re-parsing the DTB (and have FDT-parsing libraries).

------

In 2 lines:
- Would it make sense to integrate PlatformDeviceInfo in KvmtoolVirtMemInfoLib.inf and try to make KvmtoolVirtMemInfoLib.inf generic ?
- Ideally, wouldn't it be better to have a library which would parse the input DTB if present to later populate dynamic PCDs (of any kind: address/size/interrupts/etc) ?

leiflindholm:
> We would need to detect the underlying platform from the information coming from the DTB. 

Why should a platform discovery library assume there is a DTB? This is the job of two libraries.

> Would it make sense to integrate PlatformDeviceInfo in KvmtoolVirtMemInfoLib.inf and try to make KvmtoolVirtMemInfoLib.inf generic?

No, I (think we) want a truly generic PlatformDeviceInfo library, and (possibly) _also_ try to make KvmtoolVirtMemInfoLib.inf generic.

Ideally in a way that benefits physical and virtual platforms both.

pierregondois:
> > We would need to detect the underlying platform from the information coming from the DTB.
> 
> Why should a platform discovery library assume there is a DTB? This is the job of two libraries.

So IIUC the 2 libraries would:
1- Create a structure containing the relevant information.
The information could be hard-coded, parsed from a DTB or any other source. No assumption is made on how the information will be consumed, whether the lowest DRAM region will only be conserved (cf. Qemu), or all the memory regions (cf. CloudHc).
2- Populate ARM_MEMORY_REGION_DESCRIPTOR from the above source of information, i.e. implement ArmVirtGetMemoryMap(). This is the platform specific code. It would be ideal if this was generic to all platforms.

About 1., the PlatformDeviceInfo returns a PLATFORM_DEVICE_INFO[].
FWIU, the library only misses:
- parsing memory node, which is required by Qemu/CloudHv
- tagging the type of memory parsed (is it a peripheral, memory, serial port, etc)

Would it make it generic enough ?

------

Coming back to the fdt parsing help libraries:
- ArmVirtTimerFdtClientLib.inf
- ArmVirtGicArchLib.inf
- etc.

if the PlatformDeviceInfo was also parsing additional information, would it help re-using the same libraries accross platforms without modifications ?

> 
> > Would it make sense to integrate PlatformDeviceInfo in KvmtoolVirtMemInfoLib.inf and try to make KvmtoolVirtMemInfoLib.inf generic?
> 
> No, I (think we) want a truly generic PlatformDeviceInfo library, and (possibly) _also_ try to make KvmtoolVirtMemInfoLib.inf generic.
> 
> Ideally in a way that benefits physical and virtual platforms both.

ardbiesheuvel:
Would it be better for MMIO ranges not to be mapped at all from the outset, and letting drivers that need them create the mappings on demand? I implemented such a change fairly recently for the GIC driver, because UPL cannot rely on such a blanket MMIO mapping either (but this is DXE stage)

That would put the burden on the individual drivers to determine whether or not the shared bit needs to be set, which may be either 'never' or 'always' in many cases, and thus only complicates drivers where the distinction actually matters.

samimujawar:
Hi Ard,
The approach you suggest also occurred to me, but in a different context i.e. consolidating platform code. We will do some prototyping to see how this can be achieved and get back.

os-d:
> Would it be better for MMIO ranges not to be mapped at all from the outset, and letting drivers that need them create the mappings on demand? I implemented such a change fairly recently for the GIC driver, because UPL cannot rely on such a blanket MMIO mapping either (but this is DXE stage)
> 
> That would put the burden on the individual drivers to determine whether or not the shared bit needs to be set, which may be either 'never' or 'always' in many cases, and thus only complicates drivers where the distinction actually matters.

FWIW, I agree with this. And in general I wish edk2/silicon drivers did not have the expectation that MMIO will just be mapped. This leads the core (and this came up again when doing Patina) to have to map huge regions because they might be accessed.

I agree that in this case then the drivers can make the decision on the shared bit.

samimujawar:
@ardbiesheuvel I have updated the PR based on your suggestion to avoid mapping all MMIO ranges up front. The series now introduces a common MemoryMapLib helper and moves the MMIO mappings to the relevant DXE drivers/probe libraries, with KvmtoolVirtMemInfoLib only setting up System Memory, FV, and the serial port needed for early logging.
Please let me know if this approach is in the right direction. If so, I will rework and re-spin the rest of the series accordingly.

ArmVirtPkg/ArmVirt.dsc.inc
```diff
@@ -279,13 +279,15 @@ DEFINE FD_SIZE_IN_MB    = 3
...
   MemoryAllocationLib|MdePkg/Library/UefiMemoryAllocationLib/UefiMemoryAllocationLib.inf
+  MemoryMapLib|MdeModulePkg/Library/MemoryMapLib/MemoryMapLib.inf
 
 [LibraryClasses.common.UEFI_APPLICATION]
   PerformanceLib|MdeModulePkg/Library/DxePerformanceLib/DxePerformanceLib.inf
   MemoryAllocationLib|MdePkg/Library/UefiMemoryAllocationLib/UefiMemoryAllocationLib.inf
 
 [LibraryClasses.common.DXE_RUNTIME_DRIVER]
   MemoryAllocationLib|MdePkg/Library/UefiMemoryAllocationLib/UefiMemoryAllocationLib.inf
+  MemoryMapLib|MdeModulePkg/Library/MemoryMapLib/MemoryMapLib.inf
```
pierregondois: Should this be part of next patch `MdeModulePkg/MemoryMapLib: Support runtime drivers` ?

OvmfPkg/Library/MapMmioLib/MapMmioLib.c
```diff
@@ -0,0 +1,200 @@
...
+
+  // Check if aligning RegionEnd would overflow
+  if (RegionEnd > MAX_UINT64 - ALIGN_VALUE_ADDEND (RegionEnd, EFI_PAGE_SIZE)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  RegionEnd = ALIGN_VALUE (RegionEnd, EFI_PAGE_SIZE);
+
+  // Align down Base to page boundary
+  Base = Base & ~(EFI_PAGE_SIZE - 1);
```
pierregondois, line 200: Maybe `ALIGN_VALUE()`.

ArmVirtPkg/Library/KvmtoolVirtMemInfoLib/KvmtoolVirtMemInfoLib.c
```diff
@@ -66,20 +61,12 @@ ArmVirtGetMemoryMap (
   VirtualMemoryTable[Idx].Length       = PcdGet64 (PcdSystemMemorySize);
   VirtualMemoryTable[Idx].Attributes   = ARM_MEMORY_REGION_ATTRIBUTE_WRITE_BACK;
 
-  // Peripheral space before DRAM
-  VirtualMemoryTable[++Idx].PhysicalBase = 0x0;
-  VirtualMemoryTable[Idx].VirtualBase    = 0x0;
-  VirtualMemoryTable[Idx].Length         = PcdGet64 (PcdSystemMemoryBase);
+  // Map the UART
+  VirtualMemoryTable[++Idx].PhysicalBase = PcdGet64 (PcdSerialRegisterBase);
```
pierregondois, line 66: Might be safer to add an assert to check that `PcdSerialRegisterBase` is aligned.
The default value of the Pcd is `0x03F8`

OvmfPkg/Library/MapMmioLib/MapMmioLib.c
```diff
@@ -0,0 +1,200 @@
...
+  @retval EFI_SUCCESS      The range is backed by compatible MMIO descriptors.
+  @retval EFI_UNSUPPORTED  The range overlaps an existing non-MMIO descriptor,
+                           or an MMIO descriptor without the requested
+                           attributes.
+  @retval EFI_ABORTED      An existing GCD descriptor is malformed.
+  @retval Others           The GCD memory services returned an error.
+**/
+STATIC
+EFI_STATUS
+AddMmioMemorySpace (
```
pierregondois, line 48: This looks like `AddMemoryMappedIoSpace() + IntersectMemoryDescriptor()` in:
- UefiCpuPkg/CpuDxe/CpuDxe.c
- MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciHostBridge.c

Maybe they can use the library aswell ?

ArmVirtPkg/Library/ArmVirtGicArchLib/ArmVirtGicArchLib.c
```diff
@@ -96,11 +103,38 @@ ArmVirtGicArchLibConstructor (
...
+        DEBUG_INFO,
+        "Found GIC v3 Redistributor @ 0x%Lx, Len 0x%Lx\n",
+        RedistBase,
+        RedistSize
+        ));
+
+      Status = MapMmioMemory (
+                 DistBase,
+                 DistSize,
+                 (EFI_MEMORY_UC | EFI_MEMORY_XP)
```
pierregondois, line 121: Ideally this would be done in a separate library to have distinct functionalities:
- Parse the DTB
- Add the memory to the GCD

but I don't see how we could do that.
Also, I cannot think of a case where this memory would not need to be mapped + `MapMmioMemory()` has no effect if the memory is already mapped, so it should be ok I think

ArmVirtPkg/Library/ArmVirtGicArchLib/ArmVirtGicArchLib.c
```diff
@@ -96,11 +103,38 @@ ArmVirtGicArchLibConstructor (
...
+                 );
+      if (EFI_ERROR (Status)) {
+        ASSERT_EFI_ERROR (Status);
+        return (RETURN_STATUS)Status;
+      }
+
+      Status = MapMmioMemory (
```
pierregondois, line 128: Not here, but I think this should also be done in:
OvmfPkg/Fdt/VirtioFdtDxe/VirtioFdtDxe.inf
or OvmfPkg/Library/VirtioMmioDeviceLib/VirtioMmioDeviceLib.inf

when launching kvmtool with `--virtio-transport mmio` there are some issues it seems

OvmfPkg/Library/MapMmioLib/MapMmioLib.c
samimujawar: I thought of unifying these, but AddMmioMemorySpace() also sets the attributes, while the CpuDxe and PciHostBridge implementations do not.

OvmfPkg/Library/MapMmioLib/MapMmioLib.c
samimujawar: ALIGN_VALUE() aligns at the next upper boundary, see https://github.com/tianocore/edk2/blob/master/MdePkg/Include/Base.h#L949

There is no macro to align down.

ArmVirtPkg/ArmVirt.dsc.inc
samimujawar: Ack

ArmVirtPkg/Library/KvmtoolVirtMemInfoLib/KvmtoolVirtMemInfoLib.c
samimujawar: Ack

ArmVirtPkg/Library/ArmVirtGicArchLib/ArmVirtGicArchLib.c
samimujawar: What we are doing here is similar to a probe function.

ArmVirtPkg/Library/ArmVirtGicArchLib/ArmVirtGicArchLib.c
samimujawar: Thanks for reporting this. I will recreate and fix the issue.

samimujawar: The last 2 patches in the updated sub series i.e. [OvmfPkg: Introduce a Virtio MMIO probe lib](https://github.com/tianocore/edk2/pull/12379/changes/2877e9739388a082f1ba32cc3e62bff2390d0554)
and  [ArmVirtPkg: Link VirtioMmioProbeLib to VirtioFdtDxe](https://github.com/tianocore/edk2/pull/12379/changes/fd6a33e76a6b99ef79f0c1c202ccb2df7a84fca1) should fix the issue.

OvmfPkg/Library/VirtioMmioProbeLib/VirtioMmioProbeLib.c
```diff
@@ -0,0 +1,111 @@
...
+      Range
+      ));
+
+    Status = MapMmioMemory (
+               RegBase,
+               Range,
+               (EFI_MEMORY_UC | EFI_MEMORY_XP)
+               );
+    if (EFI_ERROR (Status)) {
+      return Status;
```
pierregondois, line 112: NIT: Might be worth adding an error message

OvmfPkg/Library/VirtioMmioProbeLib/VirtioMmioProbeLib.c
```diff
@@ -0,0 +1,111 @@
...
+    return Status;
+  }
+
+  for (FindNodeStatus = FdtClient->FindCompatibleNode (
+                                     FdtClient,
+                                     "virtio,mmio",
+                                     &Node
+                                     );
+       !EFI_ERROR (FindNodeStatus);
+       FindNodeStatus = FdtClient->FindNextCompatibleNode (
```
pierregondois, line 62: NIT:
- `FindNextCompatibleNode()` can only return Success/NotFound, so `!EFI_ERROR (FindNodeStatus)` works, but it might be easier to understand with a check against `EFI_NOT_FOUND` 
- `FindCompatibleNode() ` seems to be equivalent to `FindNextCompatibleNode()` with `PrevNode = 0`

It might be easier to read the code with:
```
while (TRUE) {
  FindNodeStatus = FdtClient->FindNextCompatibleNode()
  if (FindNodeStatus == EFI_NOT_FOUND) {
    break;
  } else if (EFI_ERROR (FindNodeStatus)) {
    // Error
  }

  // Else success and continue
}
```

ArmVirtPkg/ArmVirt.dsc.inc
```diff
@@ -480,7 +482,11 @@ DEFINE FD_SIZE_IN_MB    = 3
   #
   EmbeddedPkg/Drivers/FdtClientDxe/FdtClientDxe.inf
 !if $(ENABLE_VIRTIO)
-  OvmfPkg/Fdt/VirtioFdtDxe/VirtioFdtDxe.inf
+  OvmfPkg/Fdt/VirtioFdtDxe/VirtioFdtDxe.inf {
```
pierregondois, line 486: There is a dependency over `gEfiCpuArchProtocolGuid` through:

```
VirtioFdtDxe.inf
\-VirtioMmioProbeLib.inf
  \-MapMmioMemory()
    \-gEfiCpuArchProtocolGuid
```

FWIU, Depex sections in libraries are just informative. So `VirtioFdtDxe.inf` would need to have depex over `gEfiCpuArchProtocolGuid`. Unfortunately `VirtioFdtDxe.inf` is used by other platforms aswell, so we need to check this is works

ArmVirtPkg/Library/ArmVirtGicArchLib/ArmVirtGicArchLib.c
pierregondois: Ok thanks for the fix

OvmfPkg/Library/VirtioMmioProbeLib/VirtioMmioProbeLib.c
samimujawar: Ack

OvmfPkg/Library/VirtioMmioProbeLib/VirtioMmioProbeLib.c
samimujawar: I would prefer to keep this as this pattern is can also be seen at https://github.com/tianocore/edk2/blob/master/OvmfPkg/Fdt/VirtioFdtDxe/VirtioFdtDxe.c#L54-L66

ArmVirtPkg/ArmVirt.dsc.inc
samimujawar: I think the library files can be used to setup the dependencies. Please see 
1. This section is used for specifying a Depex expression, not a binary file. In the "As Built" INF files, this section contains a comment that lists the full dependency expression, including Depex statements AND'd from library instances linked against a module. (https://github.com/tianocore-docs/edk2-InfSpecification/blob/main/2_inf_overview/215_%5Bdepex%5D_section.md)
2. The EDK II build system will generate an INF file for every module that is built from source files. 
 and further the '8.7.10 [Depex] Section' states
_The complete dependency expression including all dependencies from the libraries linked with the module must be included in comments in this section. The format for this dependency expression is defined in the Module Information (INF) File Specification._ (https://tianocore-docs.github.io/edk2-BuildSpecification/draft/8_pre-build_autogen_stage/87_generated_asbuilt_inf_files.html)

So, if you see the "As Built" INF file for VirtioFdtDxe at _Build/ArmVirtKvmTool/DEBUG_GCC/AARCH64/OvmfPkg/Fdt/VirtioFdtDxe/VirtioFdtDxe/OUTPUT/VirtioFdtDxe.inf_
it has the following content

> [Depex.AARCH64]
 > \#  (gFdtClientProtocolGuid) AND (gEfiPcdProtocolGuid) AND (gEfiPcdProtocolGuid) AND (gEfiPcdProtocolGuid) AND (gEfiDevicePathUtilitiesProtocolGuid) AND (gEfiDevicePathUtilitiesProtocolGuid) AND (gEfiDevicePathUtilitiesProtocolGuid) AND (gEfiBdsArchProtocolGuid AND **gEfiCpuArchProtocolGuid** AND gEfiMetronomeArchProtocolGuid AND gEfiMonotonicCounterArchProtocolGuid AND gEfiRealTimeClockArchProtocolGuid AND gEfiResetArchProtocolGuid AND gEfiRuntimeArchProtocolGuid AND gEfiSecurityArchProtocolGuid AND gEfiTimerArchProtocolGuid AND gEfiVariableWriteArchProtocolGuid AND gEfiVariableArchProtocolGuid AND gEfiWatchdogTimerArchProtocolGuid) AND (gFdtClientProtocolGuid)

samimujawar: >Unfortunately VirtioFdtDxe.inf is used by other platforms aswell, so we need to check this is works

Since we add the dependency using the module-specific library instance override in the DSC file, i.e. linking a NULL library instance as in this case.

> OvmfPkg/Fdt/VirtioFdtDxe/VirtioFdtDxe.inf {
> <LibraryClasses>
>    NULL|OvmfPkg/Library/VirtioMmioProbeLib/VirtioMmioProbeLib.inf
>  }

Only the concerned platform should be affected.

OvmfPkg/Library/VirtioMmioProbeLib/VirtioMmioProbeLib.c
pierregondois: ok understood

OvmfPkg/Library/MapMmioLib/MapMmioLib.inf
```diff
@@ -0,0 +1,30 @@
...
+[Packages]
+  MdeModulePkg/MdeModulePkg.dec
+  MdePkg/MdePkg.dec
+
+[LibraryClasses]
+  BaseLib
+  DebugLib
+  DxeServicesTableLib
+  MemoryAllocationLib
+  UefiBootServicesTableLib
```
pierregondois, line 32: I think the library should need a depex over `gEfiCpuArchProtocolGuid`

ArmVirtPkg/Library/KvmtoolRtcFdtClientLib/KvmtoolRtcFdtClientLib.inf
```diff
@@ -23,16 +23,19 @@
...
 [LibraryClasses]
   BaseLib
   DebugLib
+  MemoryMapLib
   PcdLib
   UefiBootServicesTableLib
   DxeServicesTableLib
 
 [Protocols]
+  gEfiCpuArchProtocolGuid                               ## CONSUMES
```
pierregondois: not needed I think

ArmVirtPkg/Library/ArmVirtGicArchLib/ArmVirtGicArchLib.inf
```diff
@@ -24,15 +25,18 @@
...
 
 [Packages]
   ArmPkg/ArmPkg.dec
   ArmVirtPkg/ArmVirtPkg.dec
   EmbeddedPkg/EmbeddedPkg.dec
+  MdeModulePkg/MdeModulePkg.dec
   MdePkg/MdePkg.dec
 
 [Protocols]
+  gEfiCpuArchProtocolGuid
```
pierregondois: Not needed I think, same comment for the depex section

ArmVirtPkg/Library/KvmtoolRtcFdtClientLib/KvmtoolRtcFdtClientLib.inf
pierregondois: The depex over `gEfiCpuArchProtocolGuid` should be removed I think

ArmVirtPkg/ArmVirt.dsc.inc
pierregondois: Ok right, yes indeed.
In that case I think the dependency over gEfiCpuArchProtocolGuid should be removed in some modules, cf new comments

ArmVirtPkg/Library/ArmVirtGicArchLib/ArmVirtGicArchLib.inf
samimujawar: ack

ArmVirtPkg/Library/KvmtoolRtcFdtClientLib/KvmtoolRtcFdtClientLib.inf
samimujawar: ack

OvmfPkg/Library/MapMmioLib/MapMmioLib.inf
samimujawar: ack

ArmVirtPkg/ArmVirt.dsc.inc
samimujawar: ack

ArmVirtPkg/Library/KvmtoolRtcFdtClientLib/KvmtoolRtcFdtClientLib.inf
samimujawar: ack

pierregondois (approved):

samimujawar:
@ardbiesheuvel, I believe I have addressed your feedback. Could you please review the updated changes in this PR and let me know your thoughts? I will then re-spin the remaining patches in the series accordingly.

OvmfPkg/Library/MapMmioLib/MapMmioLib.c
```diff
@@ -0,0 +1,200 @@
...
+                                 overflows the physical address space.
+  @retval EFI_UNSUPPORTED        The range overlaps an existing non-MMIO
+                                 descriptor, or an MMIO descriptor without the
+                                 requested attributes.
+  @retval EFI_ABORTED            An existing GCD descriptor is malformed.
+  @retval Others                 The GCD memory services returned an error.
+**/
+EFI_STATUS
+EFIAPI
+MapMmioMemory (
```
ardbiesheuvel, line 157: If this library is going to be callable from runtime drivers, these routines may be invoked at runtime. This means  that a check is required to ensure that we are not trying to call DXE services at runtime.

MdeModulePkg/Library/MemoryMapLib/MemoryMapLib.inf
```diff
@@ -0,0 +1,33 @@
...
+#
+#  Copyright (c) 2026, Arm Ltd. All rights reserved.<BR>
+#
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+[Defines]
+  INF_VERSION                    = 0x0001001C
+  BASE_NAME                      = MemoryMapLib
```
ardbiesheuvel: Does this need to live in MdeModulePkg at this point?

Also, please call this `MapMmioLib` or something like that, so it is clear that `map` is a verb, and it is obvious that it deals with MMIO only.

samimujawar: Ack

OvmfPkg/Library/MapMmioLib/MapMmioLib.c
samimujawar: Ack

samimujawar:
@ardbiesheuvel Thank you for the review feedback. I believe I have now addressed all your comments. Could you please take another look?

ardbiesheuvel:
> @ardbiesheuvel Thank you for the review feedback. I believe I have now addressed all your comments. Could you please take another look?

Did you add the runtime check as well?

Also, does adding all these MMIO regions to the GCD map make the appear in the EFI memory map as well?

samimujawar:
> > @ardbiesheuvel Thank you for the review feedback. I believe I have now addressed all your comments. Could you please take another look?
> 
> Did you add the runtime check as well?

I have added a runtime check to return EFI_ACCESS_DENIED if the MapMmioMem() is called after ExitBootServices, see https://github.com/tianocore/edk2/pull/12379/changes/ab805c0f971bd2bfdc66339fd81763c6913dccc4#diff-41548d0a5fb3fac0e069d0f077180be1b7a9e297842edae60aa67757866cb553R166-R172

> Also, does adding all these MMIO regions to the GCD map make the appear in the EFI memory map as well?

I ran the Shell memmap command and only the RTC controller memory appears in the EFI memory map, see the output below.

```
Shell> mmemmap
Type          Start                            End                              # Pages                       Attributes
Available  0000000080000000-000000009DE76FFF 000000000001DE77 0000000000000008
...
MMIO       0000000001010000-0000000001010FFF 0000000000000001 8000000000000001    <----- RTC memory

  Reserved  :              0 Pages (0 Bytes)
  LoaderCode:            261 Pages (1,069,056 Bytes)
  LoaderData:              0 Pages (0 Bytes)
  BS_Code   :            658 Pages (2,695,168 Bytes)
  BS_Data   :          5,951 Pages (24,375,296 Bytes)
  RT_Code   :          1,072 Pages (4,390,912 Bytes)
  RT_Data   :            416 Pages (1,703,936 Bytes)
  ACPI_Recl :              4 Pages (16,384 Bytes)
  ACPI_NVS  :             64 Pages (262,144 Bytes)
  MMIO      :              1 Pages (4,096 Bytes)  <----- RTC memory
  MMIO_Port :              0 Pages (0 Bytes)
  PalCode   :              0 Pages (0 Bytes)
  Unaccepted:              0 Pages (0 Bytes)
  Available :        122,646 Pages (502,358,016 Bytes)
  Persistent:              0 Pages (0 Bytes)
              -------------- 
Total Memory:            512 MB (536,870,912 Bytes)
```

Following is the EFI memory map as seen by Linux kernel
```

[    0.000000] random: crng init done
[    0.000000] efi: Processing EFI memory map:
[    0.000000] efi:   0x000080000000-0x0000957bffff [Conventional|   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x0000957c0000-0x00009809ffff [Loader Code |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x0000980a0000-0x000098d6dfff [Conventional|   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x000098d6e000-0x00009b65dfff [Loader Code |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009b65e000-0x00009dc6ffff [Conventional|   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009dc70000-0x00009de6ffff [Loader Data |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009de70000-0x00009de76fff [Conventional|   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009de77000-0x00009df7bfff [Loader Code |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009df7c000-0x00009dfe7fff [Conventional|   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009dfe8000-0x00009e00afff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e00b000-0x00009e01efff [Conventional|   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e01f000-0x00009e023fff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e024000-0x00009e024fff [Loader Data |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e025000-0x00009e649fff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e64a000-0x00009e65bfff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e65c000-0x00009e66efff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e66f000-0x00009e67bfff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e67c000-0x00009e682fff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e683000-0x00009e6cefff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e6cf000-0x00009e6d4fff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e6d5000-0x00009e734fff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e735000-0x00009e73cfff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e73d000-0x00009e75ffff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e760000-0x00009e7affff [Runtime Data|RUN|  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e7b0000-0x00009e7cffff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e7d0000-0x00009e7effff [Runtime Data|RUN|  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e7f0000-0x00009e7f0fff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e7f1000-0x00009e7f5fff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e7f6000-0x00009e7fdfff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e7fe000-0x00009e82dfff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e82e000-0x00009e833fff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e834000-0x00009e839fff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e83a000-0x00009e840fff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e841000-0x00009e8cdfff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e8ce000-0x00009e913fff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e914000-0x00009e91cfff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e91d000-0x00009e926fff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e927000-0x00009e945fff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e946000-0x00009e94afff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e94b000-0x00009e954fff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e955000-0x00009e955fff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e956000-0x00009e95cfff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e95d000-0x00009e965fff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e966000-0x00009e97dfff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e97e000-0x00009e97efff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e97f000-0x00009e99efff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e99f000-0x00009e99ffff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e9a0000-0x00009e9a4fff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009e9a5000-0x00009ef2ffff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009ef30000-0x00009f35ffff [Runtime Code|RUN|  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009f360000-0x00009f48ffff [Runtime Data|RUN|  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009f490000-0x00009f4cffff [ACPI Mem NVS|   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009f4d0000-0x00009f4d3fff [ACPI Reclaim|   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009f4d4000-0x00009f4dcfff [Conventional|   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009f4dd000-0x00009f4fdfff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009f4fe000-0x00009f543fff [Boot Code   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009f544000-0x00009fffbfff [Boot Data   |   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x00009fffc000-0x00009fffffff [Conventional|   |  |  |  |  |  |  |  |  |  |   |WB|  |  |  ]
[    0.000000] efi:   0x000001010000-0x000001010fff [MMIO        |RUN|  |  |  |  |  |  |  |  |  |   |  |  |  |UC]
[    0.000000] ACPI: Early table checksum verification disabled
[    0.000000] ACPI: RSDP 0x000000009F4D3018 000024 (v02 ARMLTD)
[    0.000000] ACPI: XSDT 0x000000009F4D3F18 00008C (v01 ARMLTD ARMLFACP 00010000      01000013)

```

PR #12224, opened 2026-03-02
Support for Arm CCA guest firmware v5

samimujawar:
### Consolidated branch tracking the latest changes: https://github.com/samimujawar/edk2/tree/3945_arm_cca_v5_latest

# PR Tracking

|  # | Description                                                                 | PR Series | PR Link                                      | Comments        | Merge Status |
| -: | --------------------------------------------------------------------------- | :-------: | -------------------------------------------- | --------------- | ------------ |
|  1 | ArmVirtPkg: Kvmtool: Only install IORT if ITS is present                    |     S1    | https://github.com/tianocore/edk2/pull/12375 |                 | Merged       |
|  2 | DynamicTablesPkg: handle missing 'iommu-map' in root complex parser         |     S1    | https://github.com/tianocore/edk2/pull/12375 |                 | Merged       |
|  3 | ArmVirtPkg: Clear TPIDR_EL0 before early HOB lookups                        |     S2    | https://github.com/tianocore/edk2/pull/12379 |                 | Merged       |
|  4 | OvmfPkg/MapMmioLib: Introduce a library for mapping MMIO regions            |     S2    | https://github.com/tianocore/edk2/pull/12379 |                 | Merged       |
|  5 | OvmfPkg/MapMmioLib: Support runtime drivers                                 |     S2    | https://github.com/tianocore/edk2/pull/12379 |                 | Merged       |
|  6 | ArmVirtPkg: Include MapMmioLib in ArmVirt.dsc.inc                           |     S2    | https://github.com/tianocore/edk2/pull/12379 |                 | Merged       |
|  7 | ArmVirtPkg: Map GIC MMIO ranges before configuring GIC                      |     S2    | https://github.com/tianocore/edk2/pull/12379 |                 | Merged       |
|  8 | ArmVirtPkg/KvmtoolRtcFdtClientLib: Use MapMmioLib for MMIO mapping          |     S2    | https://github.com/tianocore/edk2/pull/12379 |                 | Merged       |
|  9 | OvmfPkg/FdtPciHostBridgeLib: Use MapMmioLib for MMIO mapping                |     S2    | https://github.com/tianocore/edk2/pull/12379 |                 | Merged       |
| 10 | OvmfPkg/FdtPciHostBridgeLib: Map PCI MMIO windows in GCD                    |     S2    | https://github.com/tianocore/edk2/pull/12379 |                 | Merged       |
[...]

PR #12380, opened 2026-04-01
3945_arm_cca_v5_s3: Initial patches for enabling RME support

samimujawar:
# Initial patches for enabling RME support

This is an initial set of patches that:
- Implement a helper function to detect RME support
- Add an API to set/clear the Arm CCA memory protect attribute bit in the page tables
- Preserve the Realm protection attribute bit
- Add a missing implementation of FreeAlignedPages() in PrePiMemoryAllocationLib.

Note: This is a PR series that includes patches from PR https://github.com/tianocore/edk2/pull/12224

- [ ] Breaking change?
  - NO
- [ ] Impacts security?
  - NO
- [ ] Includes tests?

## How This Was Tested
Only build tested.  Actual functionality can only be tested when the remaining patches from PR https://github.com/tianocore/edk2/pull/12224 are integrated in the future.

## Integration Instructions
N/A

### DIFF
--- a/EmbeddedPkg/Library/PrePiMemoryAllocationLib/MemoryAllocationLib.c
+++ b/EmbeddedPkg/Library/PrePiMemoryAllocationLib/MemoryAllocationLib.c
@@ -282,6 +282,40 @@ FreePages (
 }
 
 /**
+  Frees one or more 4KB pages that were previously allocated with one of the aligned page
+  allocation functions in the Memory Allocation Library.
+
+  Frees the number of 4KB pages specified by Pages from the buffer specified by Buffer.  Buffer
+  must have been allocated on a previous call to the aligned page allocation services of the Memory
+  Allocation Library.  If it is not possible to free allocated pages, then this function will
+  perform no actions.
+
+  If Buffer was not allocated with an aligned page allocation function in the Memory Allocation
+  Library, then ASSERT().
+  If Pages is zero, then ASSERT().
+
+  @param  Buffer                Pointer to the buffer of pages to free.
+  @param  Pages                 The number of 4 KB pages to free.
+
+**/
+VOID
+EFIAPI
+FreeAlignedPages (
+  IN VOID   *Buffer,
+  IN UINTN  Pages
+  )
+{
+  ASSERT (ADDRESS_IS_ALIGNED (Buffer, EFI_PAGE_SIZE));
+  ASSERT (Pages != 0);
+  // For now, we do not support the ability to free pages in the PrePei Memory Allocator.
+  // The allocated memory is lost.
+  DEBUG ((
+    DEBUG_INFO,
+    "PrePei allocator cannot free pages; allocated memory is not reclaimed.\n"
+    ));
+}
+
+/**
   Allocates a buffer of type EfiBootServicesData.
 
   Allocates the number bytes specified by AllocationSize of type EfiBootServicesData and returns a
--- a/UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
+++ b/UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
@@ -523,10 +523,26 @@ FillTranslationTable (
   IN  BOOLEAN                       Lpa2Enabled
   )
 {
+  UINT64  CcaProtectionAttribute;
+
+  //
+  // The CCA protection attribute corresponds to the (IPA_WIDTH - 1) bit of the
+  // Realm address space. The VA and PA for a region are expected to differ only
+  // in this bit. Derive the attribute by XORing the two addresses.
+  //
+  // NOTE: If more than one bit differs, the memory map is misconfigured.
+  //
+  CcaProtectionAttribute = MemoryRegion->VirtualBase ^ MemoryRegion->PhysicalBase;
+
+  //
+  // Ensure only one bit is set.
+  //
+  ASSERT (((CcaProtectionAttribute & (CcaProtectionAttribute - 1)) == 0));
+
   return UpdateRegionMapping (
            MemoryRegion->VirtualBase,
            MemoryRegion->Length,
-           ArmMemoryAttributeToPageAttribute (MemoryRegion->Attributes) | TT_AF,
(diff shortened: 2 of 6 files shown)


EmbeddedPkg/Library/PrePiMemoryAllocationLib/MemoryAllocationLib.c
```diff
@@ -281,6 +281,34 @@ FreePages (
...
+
+**/
+VOID
+EFIAPI
+FreeAlignedPages (
+  IN VOID   *Buffer,
+  IN UINTN  Pages
+  )
+{
+  // For now, we do not support the ability to free pages in the PrePei Memory Allocator.
```
leiflindholm, line 310: Even if we don't implement the proper deallocation, we should probably still do the ASSERTs listed in the function header.

Would it be worth adding a DEBUG_INFO message or similar?

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -484,10 +484,14 @@ FillTranslationTable (
   IN  BOOLEAN                       Lpa2Enabled
   )
 {
+  UINT64  CcaNsBit;
+
+  CcaNsBit = MemoryRegion->VirtualBase ^ MemoryRegion->PhysicalBase;
```
leiflindholm: CcaNsBit sounds like an architecturally defined thing, but my understanding is this is something that is made up just for this function. If so, it should be more traditionally CamelCase.

But also, this feels like it could lead to very difficult to debug issues if someone made a typo in their virtual memory map (or PA size config). Should we not verify that any bit set in the result of that Xor is the top PA bit?

Regardless, this is a tricky area that might benefit from a code comment.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -495,7 +521,8 @@ FillTranslationTable (
            0,
            RootTable,
            FALSE,
-           Lpa2Enabled
+           Lpa2Enabled,
+           CcaNsBit
```
leiflindholm: (referencing comment on previous commit) Is this not in fact the same thing referred to as CcaProtectionAttribute everywhere else?

leiflindholm (changes requested):

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
samimujawar: Thank you for the feedback.

I am updating the variable name to CcaProtectionAttribute, and I will add a comment here as well to make the intent clearer.

On the validation point, I do not think we can reliably verify here that the bit set in the XOR result is the CCA protection attribute bit. That bit corresponds to the Realm IPA_WIDTH - 1 bit, and determining it would require reading the Realm configuration. In practice, that would introduce dependencies on additional Arm CCA libraries, such as ArmCcaLib and ArmCcaRsiLib, which I was trying to avoid in this code path.

EmbeddedPkg/Library/PrePiMemoryAllocationLib/MemoryAllocationLib.c
samimujawar: Ack.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
samimujawar: Ack, I will update this to CcaProtectionAttribute

leiflindholm (approved):

samimujawar:
@leiflindholm If there are no further comments, can we merge this branch, please?

leiflindholm:
> @leiflindholm If there are no further comments, can we merge this branch, please?

Yeah, I think others have had enough of an opportunity to comment.

ardbiesheuvel (approved):

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -677,7 +704,8 @@ ArmSetMemoryAttributes (
            PageAttributeMask,
            ArmGetTTBR0BaseAddress (),
            TRUE,
-           ArmLpa2Enabled ()
+           ArmLpa2Enabled (),
+           0
```
hajirazaman, line 708: Is there a reason we are setting memory private during permission updates?

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -386,7 +408,8 @@ UpdateRegionMappingRecursive (
                        Level + 1,
                        FALSE,
                        FALSE,
-                       Lpa2Enabled
+                       Lpa2Enabled,
+                       CcaProtectionAttribute
```
hajirazaman, line 412: The comment above states we are populating the new child table with the attributes of the block entry it replaces. However it looks like the recursive call populates the new child tables using the passed in CcaProtectionAttribute, rather than preserving the block's original attribute.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -450,7 +474,7 @@ UpdateRegionMappingRecursive (
     } else {
       EntryValue = (*Entry & AttributeClearMask) | AttributeSetMask;
       // Below clears shareability bits when LPA2 is in use
-      SetOutputAddress (&EntryValue, RegionStart, Lpa2Enabled);
+      SetOutputAddress (&EntryValue, RegionStart, Lpa2Enabled, CcaProtectionAttribute);
```
hajirazaman, line 477: Does the upstream design support mixed private/shared memory regions?

Looks like we are passing a single CcaProtectionAttribute down in this recursive function, and for example, in the case of a permission change for a memory range, overwriting the shared/private attribute for every individual page in that range.

PR #12384, opened 2026-04-02
MdeModulePkg/Core/Dxe: Fix HeapGuard mem bin conflict

AmyChan-Intel:
# Description

HeapGuard should not make memory type bin decisions.
Update AdjustMemoryS() to return the guard-expanded allocation range,
and let Page.c validate that range against special memory type bins.
Also apply the same validation to guarded AllocateAddress requests.

- [ ] Breaking change?
- [ ] Impacts security?
- [x] Includes tests?

## How This Was Tested

CI workflows are running as expected and all checks pass.
Special pool tests also pass for all memory types.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
+++ b/MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
@@ -817,18 +817,30 @@ UnsetGuardForMemory (
   @param[in]  Start           Start address of free memory block.
   @param[in]  Size            Size of free memory block.
   @param[in]  SizeRequested   Size of memory to allocate.
+  @param[out] GuardedStart    Start address of required range including
+                              any needed guard page.
+  @param[out] GuardedEnd      End address of required range including
+                              any needed guard page.
 
   @return The end address of memory block found.
   @return 0 if no enough space for the required size of memory and its Guard.
 **/
 UINT64
 AdjustMemoryS (
-  IN UINT64  Start,
-  IN UINT64  Size,
-  IN UINT64  SizeRequested
+  IN UINT64   Start,
+  IN UINT64   Size,
+  IN UINT64   SizeRequested,
+  OUT UINT64  *GuardedStart,
+  OUT UINT64  *GuardedEnd
   )
 {
-  UINT64  Target;
+  UINT64   Target;
+  BOOLEAN  NeedHeadGuard;
+  BOOLEAN  NeedTailGuard;
+
+  if ((GuardedStart == NULL) || (GuardedEnd == NULL)) {
+    return 0;
+  }
 
   //
   // UEFI spec requires that allocated pool must be 8-byte aligned. If it's
@@ -845,7 +857,8 @@ AdjustMemoryS (
     return 0;
   }
 
-  if (!IsGuardPage (Start + Size)) {
+  NeedTailGuard = !IsGuardPage (Start + Size);
+  if (NeedTailGuard) {
     // No Guard at tail to share. One more page is needed.
     Target -= EFI_PAGES_TO_SIZE (1);
   }
@@ -862,6 +875,11 @@ AdjustMemoryS (
       return 0;
     }
   }
+
+  NeedHeadGuard = !IsGuardPage (Target - EFI_PAGES_TO_SIZE (1));
+
+  *GuardedStart = NeedHeadGuard ? (Target - EFI_PAGES_TO_SIZE (1)) : Target;
+  *GuardedEnd   = NeedTailGuard ? (Target + SizeRequested) : (Target + SizeRequested - 1);
 
   // OK, we have enough pages for memory and its Guards. Return the End of the
   // free space.
--- a/MdeModulePkg/Core/Dxe/Mem/Page.c
+++ b/MdeModulePkg/Core/Dxe/Mem/Page.c
@@ -11,18 +11,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include "HeapGuard.h"
 #include <Pi/PiDxeCis.h>
 
-//
-// Entry for tracking the memory regions for each memory type to coalesce similar memory types
-//
-typedef struct {
-  EFI_PHYSICAL_ADDRESS    BaseAddress;
-  EFI_PHYSICAL_ADDRESS    MaximumAddress;
-  UINT64                  CurrentNumberOfPages;
-  UINT64                  NumberOfPages;
-  UINTN                   InformationIndex;
-  BOOLEAN                 Special;
-  BOOLEAN                 Runtime;
-} EFI_MEMORY_TYPE_STATISTICS;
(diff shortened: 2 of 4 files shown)


os-d:
Can you update the PR title and use the PR template for the description?

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -1622,6 +1628,28 @@ CoreInternalAllocatePages (
...
+      // allocation succeeds.  Guard pages may cross bin boundaries,
+      // which can trigger CoreMemoryMapSanityCheck ASSERT in DEBUG builds
+      // but is functionally harmless.  Platforms can add DXE headroom to
+      // avoid this fallback.  (HSD 16030160228)
+      //
+      DEBUG ((DEBUG_WARN, "[HeapGuard] Bin-aware search failed: Type=%d Pages=%d. "
+        "Guard may cross bin boundary. Add DXE headroom to fix.\n",
+        MemoryType, (UINT32)NumberOfPages));
+      mGuardBinFallbackCount++;
+      Start = FindFreePages (
```
os-d: I'm not following the need for this second commit. FindFreePages() will already go and search the default bin if the preferred bin fails, then if the default bin fails, it will try in all memory.

So your new allocation would only be needed if the memory was completely maxed out and two pages outside of the bins could not be found for allocations. That seems a pretty unlikely scenario.

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
```diff
@@ -863,6 +902,23 @@ AdjustMemoryS (
     }
   }
 
+  //
+  // Reject if guard pages would extend into a non-matching special memory
+  // type bin.  This prevents CoreMemoryMapSanityCheck violations when
+  // HeapGuard is enabled for all types (PCD 0x7FFF).  (HSD 16030160228)
```
os-d: Please drop the HSD number from all the changes, looks like an internal bug number. This issue could also occur in more cases than when HeapGuard is enabled for all types, you can drop that and the PCD values. This is just a change to ensure guard pages don't cross bin boundaries, regardless of how many types have heap guard enabled.

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
```diff
@@ -863,6 +902,23 @@ AdjustMemoryS (
     }
   }
 
+  //
+  // Reject if guard pages would extend into a non-matching special memory
+  // type bin.  This prevents CoreMemoryMapSanityCheck violations when
+  // HeapGuard is enabled for all types (PCD 0x7FFF).  (HSD 16030160228)
+  //
+  if (!IsGuardPage (Target - EFI_PAGES_TO_SIZE (1))) {
```
os-d: I don't think it is ideal that this brings bin knowledge into the heap guard code, I would prefer that stay in the Page.c code (though Gcd.c has some knowledge also). This check could be done in Page.c because we can easily tell when the head and tail guard have been added and can check against the bins there. However, that then leaks heap guard implementation details into Page.c, which is also not great.

Another possibility would be to ensure that the memory bins always have distinct gMemoryMap descriptors (or one that covers all bins), even though they are free memory. This could be done by adding a new attribute/capability bit, since gMemoryMap is not spec defined, it is an edk2 structure. But if the memory bins have their own gMemoryMap descriptors, this logic would already be okay, because it makes sure that the gMemoryMap desc boundary is not crossed. It also may prevent other issues like this from occurring in other places. Then, if memory allocation fails in all cases, the bin bit could be removed, let them merge with other free descs, and allocate again.

I think changing the heap guard code to be bin aware will work, though we should consider whether it is worthwhile to try a more holistic route, since we already have a lot of code that respects descriptor boundaries.

@mdkinney @kuqin12 what are your thoughts? I mostly want to make sure we are covering all cases and not spot fixing here and there. Again, having more of the code base need to know about the bins is not ideal, but heap guard is a memory feature, so not so bad for it to know about the bins.

MdeModulePkg/Core/Dxe/Mem/Page.c
AmyChan-Intel: The 2nd commit prevents allocation failures on platforms without sufficient DXE headroom while preserving guard protection, I can remove it if this is unnecessary.

mdkinney: @AmyChan-Intel Does this introduce new behavior?  Let's keep the behavior the same and guarantee that space for guard pages allocated as part of the main allocation and guarantee that no guard page allocation or free operation ever crosses a bin boundary

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
mdkinney: @os-d Does heap guard need to be bin aware?  The page allocator is already bin aware.  Why can't heap guard effectively be a layer on top of page allocation that grows the request as needed and the enlarged page allocation can succeed or fail with the page allocation guaranteeing no bin boundary crossings.

Does the current heap guard design need to be reconsidered?

mdkinney:
@AmyChan-Intel   The PR title and PR description needs to be updated and follow the PR template.

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
os-d: This was my original thought as well. The heap guard API would need to be redesigned slightly to return the updated range (that was my concern above if we don't change the API, then the page allocator would need to know heap guard internals). Then the page allocator can accept or reject the grow, as you said.

Because the heap guard API is only used by the page allocator, I think it would be appropriate and simple to change the API to return the new range and then let the allocator make the decision on whether that is appropriate or not.

kuqin12: @mdkinney can we make the gap between different memory bins (and the page immediately before and after the first and last bins) mandatory when the corresponding heap guard is enabled? And mark the applicable gaps to be guard pages. That should make the heap guard treat this gap as a shared guard page, but also bound the memory allocation of special types inside the bin, correct?

mdkinney: Are you suggesting that the init of the DXE Core when Heap Guard is enabled should do guard page allocations at specific locations around the bins to prevent the condition that this PR is attempting to address?

There is no guarantee there will be any gaps unused pages between the bins.  The size/alignment of the bins may match up exactly.  In which case, we would have to artificially increase the size of the bins when Heap Guard is enabled to guarantee there is unused space for a guard page.  Correct?

os-d:
As mentioned in the other PR, I put up a PR that proposes to solve this in a different way after an offline discussion: https://github.com/tianocore/edk2/pull/12491.

PR #12387, opened 2026-04-02
MdeModulePkg: Add a check and close Blocko2 protocol

kanagavels97:

# Description

During the DiskIoDriverBindingStart, both the BlockIo and BlockIo2 protocols are opened for the specified controller handle. The Instance->SharedWorkingBuffer is then allocated based on the BlockSize. If the BlockSize is zero, indicating that the handle does not support Read/Write operations, the buffer will not be allocated. A NULL check is performed on the SharedWorkingBuffer, and if it is NULL, the process jumps to the ErrorExit. In the ErrorExit section, only the BlockIo protocol is closed, while the BlockIo2 protocol remains open.

## How This Was Tested

Added the above changes and verified disconnect UFS device supports WLUN (44h), which is used for the RPMB feature.

### DIFF
--- a/MdeModulePkg/Universal/Disk/DiskIoDxe/DiskIo.c
+++ b/MdeModulePkg/Universal/Disk/DiskIoDxe/DiskIo.c
@@ -121,9 +121,11 @@ DiskIoDriverBindingStart (
   IN EFI_DEVICE_PATH_PROTOCOL     *RemainingDevicePath OPTIONAL
   )
 {
-  EFI_STATUS            Status;
-  DISK_IO_PRIVATE_DATA  *Instance;
-  EFI_TPL               OldTpl;
+  EFI_STATUS              Status;
+  DISK_IO_PRIVATE_DATA    *Instance;
+  EFI_TPL                 OldTpl;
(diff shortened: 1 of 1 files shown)


kanagavels97:
Hi @mdkinney 

https://github.com/tianocore/edk2/pull/10762#discussion_r1979831228

Changes are made according to the above comment. Please review the changes.

Thanks,
Kanagavel

MdeModulePkg/Universal/Disk/DiskIoDxe/DiskIo.c
```diff
@@ -216,16 +221,25 @@ DiskIoDriverBindingStart (
...
+    if (Instance->BlockIo2 != NULL) {
+      gBS->CloseProtocol (
+             ControllerHandle,
+             &gEfiBlockIo2ProtocolGuid,
+             This->DriverBindingHandle,
+             ControllerHandle
+             );
+    }
+
+    if (Instance != NULL) {
```
spbrogan, line 240: you moved your Instance NULL check to here but reference Instance above.  Do you have a safety check?

PR #12391, opened 2026-04-02
UefiCpuPkg: ArmMmuLib: Check if Block Split Following Page Alloc

os-d:
# Description

Currently, there is a bug in UpdateRegionMappingRecursive() when guard pages are enabled and a large page is being split.

The code checks whether the page table is a block or table and seeing that it is a block, allocates a new page table for the next level. However, when it does this, it will call an additional recursive call into the page table updating logic to make sure the new page table page is mapped. In addition, when guard pages are enabled, it will mark the guard page as RP. If the guard page is in the same block as we are already trying to split, the recursive call will split the block and mark the guard page as RP.

When we return to the original call, it will fill out the now orphaned page table but never install it into the page table hierarchy (and if it did, it would lose the guard page). This has been observed to cause a driver's code section to still have NX set on it and so crash when trying to execute.

This commit resolves the issue by checking if the block has already been split when we return from the new page table allocation. If it has, we simply update the existing table mapping instead of trying to split the block.

The allocated page table page cannot be immediately freed, because this might trigger the block to get re-merged, so a reference to it is held until the end of updating this level and subsequent levels, when it can be safely freed.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on two physical ARM64 platforms from two different silicon vendors who were hitting this crash and now boot after this patch.

## Integration Instructions

N/A.


UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -338,44 +341,59 @@ UpdateRegionMappingRecursive (
...
+        // Allocating a page may have split this block if a guard page
+        // was allocated in this block. Check if this is already split
+        // and if so skip the splitting logic
+        //
+        if (IsTableEntry (*Entry, Level)) {
           //
-          // Make sure we are not inadvertently hitting in the caches
-          // when populating the page tables.
+          // Don't free the page table here, we may end up recreating the
+          // large page
```
ardbiesheuvel: Why is this true? Is it because the guard page will be freed as well, freeing up the whole block? Is there a way to free the page but not the guard page?

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -433,6 +451,15 @@ UpdateRegionMappingRecursive (
     }
   }
 
+  //
+  // We may have left an orphaned page table page if we discovered a
+  // recursive call already split a block.
+  //
```
ardbiesheuvel, line 464: Couldn't this happen more than once, i.e., once for the misaligned start and once for the misaligned end? So having a single variable would not be sufficient then.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
os-d: Yeah, if we allocated a new guard page in the last allocation, when we go to free the page table, it will free the guard page as well. The guard page only gets preserved when it is being shared with another allocation (e.g. head of one, tail of the other), but in this case it would be a newly allocated guard page, otherwise the block would already have been split.

There is no way from CpuDxe to free the page but not the guard page. The core can, but only by setting a global variable.

x86 gets around this issue (and related issues) by preallocating page table pages: https://github.com/tianocore/edk2/blob/76406ef119625ee3185d4ae3bf7701147dba50ac/UefiCpuPkg/CpuDxe/CpuPageTable.c#L1218. In Patina these kind of issues led to our design of having the core own the page table and doing the self map, so that the page table pages are not identity mapped (to be fair, Patina also doesn't have heap guard enabled yet, but having the core own the page table and the special logic for the page allocation prevents the issue).

A long while ago I had brought a PR to edk2 to add a page table pool for ARM64 like x86 had that was rejected, as a result on our side we dropped that commit and reworked some memory protections to accommodate.

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
os-d: That's a good point, thanks. Yes, if the allocation crossed a large page boundary and both sides were still mapped as large pages then you could have two tables to free. I'll update this to account for that.

os-d:
@ardbiesheuvel I have addressed your feedback, please re-review when you have a chance.

ardbiesheuvel (approved):
I'm not thrilled that we need this, but the logic is sound, and the problem is real.

os-d:
> I'm not thrilled that we need this, but the logic is sound, and the problem is real.

Agreed. I don’t like that this is needed.

PR #12314, opened 2026-03-19
UefiCpuPkg: Add RISC-V IOMMU driver

benjamindoron:
# Description

This driver follows the RISC-V IOMMU Architecture Spec v1.0[1] and detects IOMMUs
declared in either an FDT or ACPI RIMT[2] table, then installs the EDK2 IOMMU protocol
on them. It handles PCI and system IOMMUs, as well as PCI and system devices
(for a system device, the platform policy protocol is needed to match DeviceHandle).

The driver may need some cleanup, but is otherwise ready.

We may also now want to discuss the best way to integrate BaseRiscVMmuLib with its consumers
in light of the new driver. I recommend some context object: is this good?

- [ ] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [x] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

1. https://docs.riscv.org/reference/hardware/iommu/_attachments/riscv-iommu.pdf
2. https://docs.riscv.org/reference/platform-software/acpi-rimt/_attachments/rimt-spec.pdf

## How This Was Tested

This was built against https://github.com/9elements/rise-riscv-iommu-edk2-platforms
and executed with the following:

```bash
export WORKSPACE=$(pwd)
export GCC5_RISCV64_PREFIX=riscv64-linux-gnu-
export PACKAGES_PATH=$WORKSPACE/edk2:$WORKSPACE/edk2-platforms:$WORKSPACE/edk2-platforms/Platform/Intel
export EDK_TOOLS_PATH=$WORKSPACE/edk2/BaseTools

# Set up the build environment
pushd edk2
source edksetup.sh --reconfig
make -C BaseTools
popd

# Clean and build
build -a RISCV64 --buildtarget DEBUG \
  -p edk2-platforms/Platform/SiFive/U5SeriesPkg/FreedomU540HiFiveUnleashedBoard/U540.dsc \
  -t GCC5 cleanall

build -a RISCV64 --buildtarget DEBUG \
  -p edk2-platforms/Platform/SiFive/U5SeriesPkg/FreedomU540HiFiveUnleashedBoard/U540.dsc \
  -t GCC5

# Boot
[...]

PR #12393, opened 2026-04-03
UefiCpuPkg/BaseRiscVMmuLib: Support MMU page unmap

benjamindoron:
# Description

Add support for the EFI_MEMORY_RP GCD attribute, completing the support
for the standardised page attributes and thereby allowing common code to
unmap pages.

This fixes potential security vulnerabilities, as out of bounds memory
accesses can now be caught by the MMU.

- [ ] Breaking change?
- [x] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested as part of https://github.com/tianocore/edk2/pull/12314

## Integration Instructions

N/A

### DIFF
--- a/UefiCpuPkg/Library/BaseRiscVMmuLib/BaseRiscVMmuLib.c
+++ b/UefiCpuPkg/Library/BaseRiscVMmuLib/BaseRiscVMmuLib.c
@@ -440,7 +440,18 @@ UpdateRegionMappingRecursive (
       }
 
       EntryValue = SetPpnToPte (EntryValue, RegionStart);
-      EntryValue = SetValidPte (EntryValue);
+
+      //
+      // Unmapped memory regions must have the 'valid' bit cleared.
+      // This is because entries with 'valid' set and RWX clear are considered as "table entries"
+      // (non-leaves), and trigger the ASSERT above for reaching the leaf-level with a non-leaf.
+      //
+      if ((AttributeClearMask == PTE_ATTRIBUTES_MASK) && (AttributeSetMask == 0)) {
+        EntryValue &= ~RISCV_PG_V;
+      } else {
+        EntryValue = SetValidPte (EntryValue);
+      }
+
       ReplaceTableEntry (Entry, EntryValue, RegionStart, TableIsLive);
     }
   }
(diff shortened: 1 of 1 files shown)


UefiCpuPkg/Library/BaseRiscVMmuLib/BaseRiscVMmuLib.c
```diff
@@ -440,7 +440,18 @@ UpdateRegionMappingRecursive (
...
 
       EntryValue = SetPpnToPte (EntryValue, RegionStart);
-      EntryValue = SetValidPte (EntryValue);
+
+      //
+      // Unmapped memory regions must have the 'valid' bit cleared.
+      // This is because entries with 'valid' set and RWX clear are considered as "table entries"
+      // (non-leaves), and trigger the ASSERT above for reaching the leaf-level with a non-leaf.
+      //
+      if ((AttributeClearMask == PTE_ATTRIBUTES_MASK) && (AttributeSetMask == 0)) {
```
tphan-qualcomm, line 449: How AttributeSetMask is zero in this function?
AttributeSetMask is made by coverting GCD attribute in GcdAttributeToPageAttribute function.
In that function, AttributeSetMask cannot be zero.

benjamindoron: The IOMMU driver can pass 0. We pass either `R`, `RW`, or nothing, if we want the pages unmapped.

I see that `GcdAttributeToPageAttribute` always returns at least `R` (other than the cacheability bits), but what do you do if you want something unmapped? We even have a bit for that in UEFI, `RP` (I forget all of the prefixes). Presumably, it's granted the read permission again, which is wrong. 

Ah. This PR should have a commit that adds support for `RP` into `GcdAttributeToPageAttribute`. Anyways, this explains my observation while I debugged this in https://github.com/tianocore/edk2/pull/12314, that the MMU driver seems not to trigger the map->unmap->map behaviour.

tphan-qualcomm: Yes, you need to have commit to support unmask attribute to GcdAttributeToPageAttribute function then this fix will make sense.

benjamindoron: Done. Thanks for the help catching this issue (that potentially has security impact by allowing out of bounds memory accesses, but nonetheless helps catch programming errors).

tphan-qualcomm:
Thanks for the patch. LGTM.

vlsunil (approved):

PR #12395, opened 2026-04-04
RedfishPkg/ConfigHandler: Remove [Depex] section from INF file

### LINKED ISSUE #12394: [Bug]: Remove [DEPEX] section from RedfishConfigHandler driver

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

RedfishPkg

DEBUG, NO-TARGET, NOOPT, RELEASE

### Current Behavior

Remove [DEPEX] section from RedfishConfigHandler driver, as it is a UEFI driver.

### Expected Behavior

No dependency section in the INF file.

### Steps To Reproduce

N/A

### Build Environment

```markdown
- OS(s): All
- Tool Chain(s): All
```

### Version Information

```text
b7cf7e465c402f9b32dcadc791ba37f2db750522
```

### Urgency

Medium

I will fix it

Maintainer feedback requested

### Anything else?

_No response_

changab:
UEFI driver shouldn't have the [Depex] section in INF file. Remove Redfish Credential Protocol in the dependency section and locate it when the UEFI driver is connected to execute.

# Description
fixes: #12394

UEFI driver shouldn't have a [Depex] section.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Verified the functionality on AMD platform.

## Integration Instructions
N/A

### DIFF
--- a/RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.h
+++ b/RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.h
@@ -3,6 +3,7 @@
   and DXE driver.
 
   (C) Copyright 2021 Hewlett Packard Enterprise Development LP<BR>
+  Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.<BR>
 
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
@@ -63,16 +64,12 @@ RedfishConfigDriverCommonUnload (
   This is the common code for Redfish configuration UEFI and DXE driver
   initialization.
 
-  @param[in]  ImageHandle       The firmware allocated handle for the UEFI image.
-  @param[in]  SystemTable       A pointer to the EFI System Table.
-
   @retval EFI_SUCCESS           The operation completed successfully.
   @retval Others                An unexpected error occurred.
 **/
 EFI_STATUS
 RedfishConfigCommonInit (
-  IN EFI_HANDLE        ImageHandle,
-  IN EFI_SYSTEM_TABLE  *SystemTable
+  VOID
   );
 
 /**
--- a/RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.c
+++ b/RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.c
@@ -5,6 +5,7 @@
 
   Copyright (c) 2019, Intel Corporation. All rights reserved.<BR>
   (C) Copyright 2021 Hewlett Packard Enterprise Development LP<BR>
+  Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.<BR>
 
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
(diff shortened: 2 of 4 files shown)


khaliid2040:
Please link the issue with the PR probably. Suggestion: 

Use `fixes: #12394` instead of  "Fix the issue #12394."

This auto-closes the issue when PR is merged.

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.c
```diff
@@ -171,6 +172,20 @@ RedfishConfigDriverBindingSupported (
     &gEfiRestExServiceBindingProtocolGuid,
     ChildHandle
     );
+
+  //
+  // Check if the Redfish credential protocol is installed or not.
+  //
+  Status = RedfishConfigCommonInit ();
```
nicklela: Hmm... I am not a fan of calling init function in supported function. The end-of-dxe event and exit-boot-service event are created multiple times.

Could we just test and see if gEdkIIRedfishCredentialProtocolGuid is installed or not?

changab: Yes, we can test this procotol in the Support function and locate it in the Start().
Also, thanks for catching the multiple event registration issue.

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.h
```diff
@@ -3,12 +3,14 @@
...
 
   (C) Copyright 2021 Hewlett Packard Enterprise Development LP<BR>
+  Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.<BR>
 
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
 
-#pragma once
+#ifndef EFI_REDFISH_CONFIG_HANDLER_COMMON_H_
```
nicklela: is this intentional?

nicklela (commented):
Looks good to me but I have one minor comment.

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerCommon.h
EricGao2015: Edk2 has replaced include guards with pragma once, please refer https://github.com/tianocore/edk2/pull/12149.

EricGao2015:
Look good for me. But it may be better that moves the locate protocol operation from Supported() to start(), because Supported() is called frequently.

changab:
@EricGao2015, @nicklela  and @igorkulchytskyy. Please check this out again.

igorkulchytskyy:
> Look good for me. But it may be better that moves the locate protocol operation from Supported() to start(), because Supported() is called frequently.

Since right after the success of Supported() function, Start() function is called, and we call LocateProtocol at the end of Supported() function, it would be the same number of calls.
But it is better to check all required conditions in Supported() function.

igorkulchytskyy (approved):

EricGao2015 (approved):

changab:
> > Look good for me. But it may be better that moves the locate protocol operation from Supported() to start(), because Supported() is called frequently.
> 
> Since right after the success of Supported() function, Start() function is called, and we call LocateProtocol at the end of Supported() function, it would be the same number of calls. But it is better to check all required conditions in Supported() function.

HI @igorkulchytskyy, we check gEdkIIRedfishCredentialProtocolGuid installation in Support() to ensure we are good to start the driver. I think this logic is fine by itself, or would you like to have a variable to record the protocol interface and skip LocateProtocol() when the protocol interface is not NULL?

igorkulchytskyy:
> HI @igorkulchytskyy, we check gEdkIIRedfishCredentialProtocolGuid installation in Support() to ensure we are good to start the driver. I think this logic is fine by itself, or would you like to have a variable to record the protocol interface and skip LocateProtocol() when the protocol interface is not NULL?
I think the current logic is good

nicklela (approved):

changab:
> > HI @igorkulchytskyy, we check gEdkIIRedfishCredentialProtocolGuid installation in Support() to ensure we are good to start the driver. I think this logic is fine by itself, or would you like to have a variable to record the protocol interface and skip LocateProtocol() when the protocol interface is not NULL?
> > I think the current logic is good

Thanks!

PR #12410, opened 2026-04-07
Use (VOID) cast for intentionally ignored gBS->SetTimer() return values

AbuthahirM11:
When gBS->SetTimer() is called and the return value is intentionally
not used, cast to (VOID) to make the intent explicit to static analysis
tools and code reviewers, rather than silently discarding the return value.

This fixes all 66 bare gBS->SetTimer() call sites across 6 packages
(39 files):

- NetworkPkg (19 files)
- MdeModulePkg (12 files)
- ShellPkg (5 files)
- EmulatorPkg (1 file)
- ArmPkg (1 file)
- OvmfPkg (1 file)

Each package is in a separate commit for independent review.

Cc: Michael D Kinney [michael.d.kinney@intel.com](mailto:michael.d.kinney@intel.com)

Signed-off-by: Abuthahir M [abuthahirm@ami.com](mailto:abuthahirm@ami.com)

### DIFF
--- a/ArmPkg/Drivers/ArmPsciMpServicesDxe/ArmPsciMpServicesDxe.c
+++ b/ArmPkg/Drivers/ArmPsciMpServicesDxe/ArmPsciMpServicesDxe.c
@@ -1147,11 +1147,11 @@ CheckAllAPsStatus (
     return;
   }
 
-  gBS->SetTimer (
-         mCpuMpData.CheckAllAPsEvent,
-         TimerCancel,
-         0
-         );
+  (VOID)gBS->SetTimer (
+               mCpuMpData.CheckAllAPsEvent,
+               TimerCancel,
+               0
+               );
 
   if (mCpuMpData.FailedListIndex == 0) {
     if (mCpuMpData.FailedList != NULL) {
--- a/EmulatorPkg/CpuRuntimeDxe/MpService.c
+++ b/EmulatorPkg/CpuRuntimeDxe/MpService.c
@@ -727,11 +727,11 @@ CpuMpServicesStartupThisAP (
   if (WaitEvent != NULL) {
     // Non Blocking
     gMPSystem.WaitEvent = WaitEvent;
-    gBS->SetTimer (
-           gMPSystem.ProcessorData[ProcessorNumber].CheckThisAPEvent,
-           TimerPeriodic,
-           gPollInterval
-           );
+    (VOID)gBS->SetTimer (
+                 gMPSystem.ProcessorData[ProcessorNumber].CheckThisAPEvent,
+                 TimerPeriodic,
+                 gPollInterval
+                 );
     return EFI_SUCCESS;
   }
 
@@ -1120,11 +1120,11 @@ CpuCheckAllAPsStatus (
     return;
   }
 
-  gBS->SetTimer (
-         gMPSystem.CheckAllAPsEvent,
-         TimerCancel,
-         0
-         );
+  (VOID)gBS->SetTimer (
+               gMPSystem.CheckAllAPsEvent,
+               TimerCancel,
+               0
+               );
 
   if (gMPSystem.FailedListIndex == 0) {
     if (gMPSystem.FailedList != NULL) {
--- a/MdeModulePkg/Bus/Isa/Ps2MouseDxe/Ps2Mouse.c
+++ b/MdeModulePkg/Bus/Isa/Ps2MouseDxe/Ps2Mouse.c
@@ -492,7 +492,7 @@ PS2MouseDriverStop (
   //
   // Cancel mouse data polling timer, close timer event
   //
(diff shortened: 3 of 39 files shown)


mikebeaton:
My thought in the other issue was that it would make sense to *globally* fix this in a separate PR. If we're only going to fix the three touched by the other PR, then an extra commit in that other PR seems fine. A very quick grep finds me 63 instances (~including these three~ EDIT: plus these three, which are formatted slightly differently) that need fixing.

AbuthahirM11:
Thanks for the suggestion @mikebeaton! I've updated this PR to globally fix all bare gBS->SetTimer() calls across the codebase. All instances now use the (VOID)gBS->SetTimer(...) pattern to make the intentionally ignored return values explicit.

mikebeaton:
@AbuthahirM11 - Many thanks for doing this!

@leiflindholm - This does concretize my reservations about doing this in the first place, as we end up casting to (VOID) the return values of various of these calls to SetTimer right next to calls to CreateEvent, CloseEvent, etc. which also have return values which we are ignoring, and not casting to (VOID).

a) It seems pretty clear that no tooling which balks at this can work with EDK 2 for now (and thus I think it seems pretty unlikely that the three cases we saw before - where the result is assigned to Status and then ignored - were like that to keep tooling happy), but

b) More importantly, this simply isn't something that is done in the EDK 2 codebase, and hasn't been since the start - I think that is clear to see? (I can find a _very_ small number of instances of casting to (VOID) in this way, in the codebase - but you can find almost any code pattern once or twice in edk2 if you look hard!) So I feel any changes to this would be at best piecemeal, at worst confusing for people adding to the codebase in future?

Just my tuppenyworth.

@AbuthahirM11, you certainly correctly catch the 63 instances of this that can quickly be found with a grep, plus the three which assign to a `Status` variable which is then unused, which were near the code you changed before. (I haven't checked with a fine toothcomb for anything else which might be like that.)

leiflindholm:
@mikebeaton what are we trying to do here? Improve the codebase or make analysis tool go quiet?

I am happy(ish) with:
- leaving the calls untouched
- explicitly casting to (VOID) to clarify intent
- creating a wrapper function that itself explicitly discards return value to clarify intent

I am unhappy with:
- changing one less-than-ideal pattern to another less-than-ideal pattern and calling that an improvement (not appliccable to this PR, but to the one that spawned it)

I am unconvinced by:
- other code in the tree also being unclear about intent or sloppiness

If we're convinced there is no way these calls could ever fail, I feel the cleanest way would be the wrapper function route. But that still means we assume that the EFI_EVENT argument is _always_ valid. And that to me looks and feels like trying to explain away bugs rather than an opinion about code patterns. The function has a return value for a reason.

mikebeaton:
> @mikebeaton what are we trying to do here? Improve the codebase or make analysis tool go quiet?
> 
> I am happy(ish) with:
> 
> * leaving the calls untouched
> * explicitly casting to (VOID) to clarify intent
> * creating a wrapper function that itself explicitly discards return value to clarify intent
> 
> I am unhappy with:
> 
> * changing one less-than-ideal pattern to another less-than-ideal pattern and calling that an improvement (not appliccable to this PR, but to the one that spawned it)
> 
> I am unconvinced by:
> 
> * other code in the tree also being unclear about intent or sloppiness
> 
> If we're convinced there is no way these calls could ever fail, I feel the cleanest way would be the wrapper function route. But that still means we assume that the EFI_EVENT argument is _always_ valid. And that to me looks and feels like trying to explain away bugs rather than an opinion about code patterns. The function has a return value for a reason.

@leiflindholm - It looks to me as if the original three lines, which assign to Status then ignore it, were just a copy-paste error or something similar, and if they'd be written the way the rest of the codebase is in the first place, they would just have had no assignment to Status, so ending up as @AbuthahirM11 originally had it in his PR.

Changing just those three to a cast to (VOID) is somewhat okay, but a little bit confusing to future coders, due to the inconsistent approach. Changing every single SetTimer call with unused status to a cast to (VOID) is also somewhat okay, but also a little bit confusing to future coders, since lots of other very similar stuff remains unconverted. And in both cases, it also doesn't actually deal with any error; as the code currently doesn't. But it would perhaps (and perhaps unfortunately) make it look as if any issues have been resolved.

So it still looks to me as if the best simple solution is leaving the original PR as @AbuthahirM11 had it, and not adding any casts to (VOID) (neither there nor here).

In terms of the best solution, what would a wrapper function do? I might be misunderstanding what you were thinking of, but perhaps the most obvious useful behaviour would be to ASSERT on any error, so warning of problems in DEBUG and NOOPT builds, but then to carry on anyway, so not actually changing anything in RELEASE builds.

AbuthahirM11:
@leiflindholm @mikebeaton Thank you both for the detailed feedback — should I close this PR and revert to the original approach (no casts), or would you prefer I implement the wrapper function with ASSERT on error as suggested?

PR #12412, opened 2026-04-07
NetworkPkg/HttpBootDxe: Add all events to HttpBootHttpCallback()

SaloniKasbekar:
Include other events from EDKII_HTTP_CALLBACK_EVENT to print failure information in the event of HTTP Boot failure

# Description

Add other non-TLS events to the print failure information during HTTP Boot flow for better end user experience

- [ ] Breaking change?

- [ ] Impacts security?

- [ ] Includes tests?

## How This Was Tested

Verified failure prints during HTTP Boot flow

## Integration Instructions

N/A

### DIFF
--- a/NetworkPkg/HttpBootDxe/HttpBootImpl.c
+++ b/NetworkPkg/HttpBootDxe/HttpBootImpl.c
@@ -978,9 +978,8 @@ EFI_HTTP_BOOT_CALLBACK_PROTOCOL  gHttpBootDxeHttpBootCallback = {
 /**
   Callback function that is invoked when an HTTP event occurs during HTTP Boot.
 
-  This function handles TLS-related events (HttpEventTlsConnectSession and
-  HttpEventTlsConfigured) and prints error messages to the screen when a TLS
-  error is encountered during the HTTP Boot process.
+  This function handles all HTTP callback events and prints error messages
+  to the screen when an error is encountered during the HTTP Boot process.
 
   @param[in]  This              Pointer to the EDKII_HTTP_CALLBACK_PROTOCOL instance.
   @param[in]  Event             The event that occurs in the current state.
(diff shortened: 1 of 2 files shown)


NetworkPkg/HttpBootDxe/HttpBootImpl.c
mikebeaton, line 1020: This would be an illegal value (now), all the same might it be a good idea to print something here too?

SaloniKasbekar: Yeah, would be no harm to print out something here. I'll add it in.

NetworkPkg/HttpBootDxe/HttpBootImpl.c
```diff
@@ -996,15 +995,28 @@ HttpBootHttpCallback (
...
+      case HttpEventInitSession:
+        AsciiPrint ("\n  Error: HTTP session initialization failed - %r.\n", EventStatus);
+        break;
+
       case HttpEventTlsConfigured:
         AsciiPrint ("\n  Error: TLS configuration failed - %r.\n", EventStatus);
         break;
 
       default:
+        AsciiPrint ("\n  Error: Unknown HTTP event - %r.\n", EventStatus);
```
mikebeaton, line 1019: Sorry - I wonder if we could add the unknown event code as an int to this message, in the unlikely event that it occurs?

Though it seems to me we need a horrible cast to do it safely:
```
        AsciiPrint ("\n  Error: Unknown HTTP event (%ld) - %r.\n", (INT64)Event, EventStatus);
````

SaloniKasbekar: I think printing the undefined event code in the AsciiPrint() might not be needed since an end user looking at this print wouldnt be able to decode it in any case. Ideally a developer should be able to see calls to the callback to identify the unknown event in any case.

If we see this as a need, we could add it to a Debug() API in the future?

mikebeaton: Unless anyone else wants to chime in, I'm happy to go with the decision to leave it as you have it now.

mikebeaton (approved):

NetworkPkg/HttpBootDxe/HttpBootImpl.c
mikebeaton: Again, I can't resolve my own comments - red error message - but resolved from me. Thank you.

Zclarkwilliams (approved):

SaloniKasbekar:
@mdkinney @makubacki Could you help review/push this change?

SaloniKasbekar:
@mdkinney @makubacki  Could you help review/push this change?

mdkinney (approved):

mdkinney:
@mergify rebase

PR #12418, opened 2026-04-08
EmulatorPkg: Revert Remove EmulatorPkg from CLANGPDB CI

mikebeaton:
# Description

This reverts commit 6e39689c659dd79c3c201832d9c7a83f84536653.

EmulatorPkg was explicitly excluded from CLANGPDB CI, since it was not supported under CLANGPDB at the time the CLANG CI PR was prepared.

It is now supported so revert the commit which disabled it.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

CI

## Integration Instructions

N/A

### DIFF
--- a/EmulatorPkg/PlatformCI/.azurepipelines/Ubuntu-CLANGPDB.yml
+++ b/EmulatorPkg/PlatformCI/.azurepipelines/Ubuntu-CLANGPDB.yml
@@ -1,0 +1,102 @@
+## @file
+# Azure Pipeline build file for building a platform.
+#
+# Platform: EmulatorPkg
+# OS: Ubuntu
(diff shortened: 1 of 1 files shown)


mikebeaton:
@makubacki - Hello! Apologies, but this would also need the respective pipeline adding, once again - if by any chance you have the time to do so! Many thanks.

os-d:
@mikebeaton this would need to run on Windows. The host emulation binary is also built by the target toolchain and the PE binary won’t run on Ubuntu, of course.

mikebeaton:
Oh - right, I think I was assuming it could at least build here. Wrongly, it seems.

PR #12422, opened 2026-04-09
BaseTools: silence make

kraxel:
# Description

Instead of using the '-' prefix append '|| true' to the commands where
errors should be ignored.  With the '-' prefix make will clutter the
build output with the messages about the (ignored) errors.

Changed only the GCC and CLANGDWARF rules for linux systems.


EricGao2015:
@Gred Hi Gred, could you please resubmit this PR? I believe that's the easiest way to eliminate the build error meassage.

PR #12423, opened 2026-04-09
NvmExpressMediaClear: Fix BufferSize value in writeblocks call.

### LINKED ISSUE #12397: [Bug]: NvmeDxe:: NvmExpressMediaClear: Incorrect buffersize for write command

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

MdeModulePkg

RELEASE, DEBUG

### Current Behavior

NvmExpressMediaClear: BufferSize passed as 1 instead of block size → function is completely non-functional

Why this is a real bug:
WriteBlocks expects BufferSize in bytes. The call passes 1:

NvmeBlockIoWriteBlocks checks (BufferSize % BlockSize) != 0. Since 1 % 512 != 0 (for any standard block size ≥ 2), this always returns EFI_BAD_BUFFER_SIZE.

Consequence:
The NIST 800-88r1 media clear operation is entirely non-functional. No overwrite data is written to any sector. A caller relying on this for data sanitization receives no actual sanitization.

Why this is NOT a false positive:
The comment says "Write one sector per write" — the intent was BlockSize bytes. The EFI_BLOCK_IO_PROTOCOL specification defines BufferSize as bytes, not sectors. There is no code path where BufferSize=1 would pass the block-alignment check.

### Expected Behavior

Block is overwritten

### Steps To Reproduce

Trigger path:

Caller invokes NvmExpressMediaClear with valid parameters.
Inner loop calls WriteBlocks with BufferSize = 1.
NvmeBlockIoWriteBlocks returns EFI_BAD_BUFFER_SIZE.
Loop continues (no error check), every iteration fails.
Function returns last error status. No data is ever written.

### Build Environment

```markdown
any
```

### Version Information

```text
edk2/masater
```

### Urgency

High

I will fix it

No maintainer feedback needed

### Anything else?

Found in Project Mu https://github.com/microsoft/mu_basecore/issues/1739

spbrogan:
NvmExpressMediaClear requires that the overwrite buffer be the same size as the block size of the media but then WriteBlocks expects BufferSize in bytes. NvmExpressMediaClear was hardcoded to call using 1.  This causes the WriteBlocks to fail.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Host based unit tests.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressMediaSanitize.c
+++ b/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressMediaSanitize.c
@@ -355,13 +355,17 @@ NvmExpressMediaClear (
   }
 
   //
-  // If an invalid buffer or buffer size is sent, the Media Clear operation
-  // cannot be performed as it requires a native WRITE command. The overwrite
-  // buffer must have granularity of a namespace block size.
+  // If an invalid buffer is sent, the Media Clear operation
+  // cannot be performed as it requires a native WRITE command.
   //
   if (SectorOwBuffer == NULL) {
     return EFI_INVALID_PARAMETER;
   }
+
+  //
+  // Since the parameters do not include OwBuffer Size, the size of the overwrite buffer is assumed to be equal
+  // to the block size of the media.
+  //
 
   Status = EFI_SUCCESS;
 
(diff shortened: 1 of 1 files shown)


spbrogan:
fixes #12397

lgao4 (approved):

lgao4:
@spbrogan , this is a bug fix. Do you request to merge it for stable tag 202605?

spbrogan:
this is ready to merge.  It is linked to a bug and yes it is a bugfix.

PR #12424, opened 2026-04-09
Last LBA missed on NvmExpressMediaClear

### LINKED ISSUE #12396: [Bug]: NvmeDxe::NvmExpressMediaClear: off-by-one skips last LBA

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

MdeModulePkg

DEBUG, RELEASE

### Current Behavior

NvmExpressMediaClear: off-by-one skips last LBA

Media->LastBlock is the 0-based index of the last valid LBA. The < operator excludes it. The loop writes sectors 0 through LastBlock - 1, missing the final sector.

Consequence:
The final sector of the namespace is not overwritten during a media clear. For security-sensitive sanitization, this leaves residual data on media.

Why this is NOT a false positive:
The NVMe spec and EFI_BLOCK_IO_MEDIA specification define LastBlock as the LBA of the last block (inclusive). Correct iteration would use <=.

### Expected Behavior

Last block should be cleared

### Steps To Reproduce

Trigger path:

Namespace has Nsze = 100 → Media->LastBlock = 99.
Loop: SectorOffset from 0 to 98 → sector 99 is never written.

### Build Environment

```markdown
any
```

### Version Information

```text
edk2/master
```

### Urgency

Medium

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

spbrogan:
Off by one error in the code of NvmExpressMediaClear() causes the last LBA to be missed when clearing the media. This patch fixes the issue by adjusting the loop condition to ensure that all LBAs are cleared properly.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Host based unit tests and code inspection

## Integration Instructions

n/a

### DIFF
--- a/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressMediaSanitize.c
+++ b/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressMediaSanitize.c
@@ -373,7 +373,7 @@ NvmExpressMediaClear (
   // Per NIST 800-88r1, one or more pass of writes may be alteratively used.
   //
   for (TotalPassCount = 0; TotalPassCount < PassCount; TotalPassCount++) {
-    for (SectorOffset = 0; SectorOffset < Media->LastBlock; SectorOffset++ ) {
+    for (SectorOffset = 0; SectorOffset <= Media->LastBlock; SectorOffset++ ) {
       Status = Device->BlockIo.WriteBlocks (
                                  &Device->BlockIo,
                                  MediaId,
--- a/MdePkg/Include/Protocol/BlockIo.h
+++ b/MdePkg/Include/Protocol/BlockIo.h
@@ -171,7 +171,8 @@ typedef struct {
   UINT32     IoAlign;
 
   ///
-  /// The last logical block address on the device.
+  /// The last logical block address on the device (inclusive).
+  /// This is generally the total number of user addressable blocks minus one.
   /// If the media changes, then this field is updated.
   ///
   EFI_LBA    LastBlock;


spbrogan:
resolves https://github.com/tianocore/edk2/issues/12396

mikebeaton (changes requested):
It is correct that LastBlock is (almost, see below) universally initialized to Size -1 (or copied from another last block value) in the codebase.

Could the comment be updated?
https://github.com/tianocore/edk2/blob/master/MdePkg/Include/Protocol/BlockIo.h#L173-L177
e.g.
```
/// The last logical block address on the device (inclusive).
```

Additionally, Is this test wrong?

https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Bus/Pci/NvmExpressDxe/UnitTest/MediaSanitizeUnitTest.c#L589

spbrogan:
> It is correct that LastBlock is (almost, see below) universally initialized to Size -1 (or copied from another last block value) in the codebase.
> 
> Could the comment be updated? https://github.com/tianocore/edk2/blob/master/MdePkg/Include/Protocol/BlockIo.h#L173-L177 e.g.
> 
> ```
> /// The last logical block address on the device (inclusive).
> ```
> 
> Additionally, Is this test wrong?
> 
> https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Bus/Pci/NvmExpressDxe/UnitTest/MediaSanitizeUnitTest.c#L589

This PR #12426 makes an update.  I don't think it really mattered for the host-based unit test as the device was essentially mocked but I think it should be clear for intent.

mikebeaton:
> > It is correct that LastBlock is (almost, see below) universally initialized to Size -1 (or copied from another last block value) in the codebase.
> > Could the comment be updated? https://github.com/tianocore/edk2/blob/master/MdePkg/Include/Protocol/BlockIo.h#L173-L177 e.g.
> > ```
> > /// The last logical block address on the device (inclusive).
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Additionally, Is this test wrong?
> > https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Bus/Pci/NvmExpressDxe/UnitTest/MediaSanitizeUnitTest.c#L589
> 
> This PR #12426 makes an update. I don't think it really mattered for the host-based unit test as the device was essentially mocked but I think it should be clear for intent.

Thank you. 

It is true that the UEFI spec. for EFI_BLOCK_IO_MEDIA is also unambiguous. The comment for LastBlock in the header file is not, however. Could/should it be updated, e.g. to:

```
  ///
  /// The last logical block address on the device (i.e. the
  /// total number of user addressable blocks minus one).
  /// If the media changes, then this field is updated.
  ///
  EFI_LBA    LastBlock;
```

spbrogan:
> > > It is correct that LastBlock is (almost, see below) universally initialized to Size -1 (or copied from another last block value) in the codebase.
> > > Could the comment be updated? https://github.com/tianocore/edk2/blob/master/MdePkg/Include/Protocol/BlockIo.h#L173-L177 e.g.
> > > ```
> > > /// The last logical block address on the device (inclusive).
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Additionally, Is this test wrong?
> > > https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Bus/Pci/NvmExpressDxe/UnitTest/MediaSanitizeUnitTest.c#L589
> > 
> > 
> > This PR #12426 makes an update. I don't think it really mattered for the host-based unit test as the device was essentially mocked but I think it should be clear for intent.
> 
> Thank you.
> 
> It is true that the UEFI spec. for EFI_BLOCK_IO_MEDIA is also unambiguous. The comment for LastBlock in the header file is not, however. Could/should it be updated, e.g. to:
> 
> ```
>   ///
>   /// The last logical block address on the device (i.e. the
>   /// total number of user addressable blocks minus one).
>   /// If the media changes, then this field is updated.
>   ///
>   EFI_LBA    LastBlock;
> ```

I have updated the BlockIo.h file as well.

spbrogan:
@lgao4  - This is done and very simple.  Can you please review and merge this?

lgao4 (approved):

spbrogan:
@mergifyio queue

PR #12430, opened 2026-04-09
Global: optimise StandaloneMmMmuLib with FF-A v1.3 and change ArmFfaGetVersion()'s argument type

LeviYeoReum:
# Description

commit 4ca452c ("ArmPkg/Library: Update StandaloneMmuLib with FF-A v1.2 with page granularity") introduced page-granular memory permission updates to fix the StandaloneMm load failure.

However, this approach results in issuing an SMC call per page when updating permissions over a region.

Since FF-A Memory Management Protocol v1.3 [0],
FFA_MEM_PERM_GET has been updated to return the number of contiguous pages that share the same memory permission starting from a base address.

With this change, update memory permissions in bulk (per contiguous page range) to reduce the number of SMC calls, leveraging FF-A Memory Management Protocol v1.3 [0].

Also, This patch includes to change ArmFfaGetVersion()'s argumtents type to remove
some impression treating major and minor version of the ABI as two unrelated 16-bit variables
as opposed to the upper and lower 16-bits of a 32-bit version variable.

Link: https://developer.arm.com/documentation/den0140/latest [0]

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Boot testing StandaloneMm in FVP revC

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
+++ b/ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
@@ -35,6 +35,9 @@
   StMM Core invokes this library before constructors are called and before the
   StMM image itself is relocated.
 
+  @param[out]     MajorVersion        Major Version of ABI.
+  @param[out]     MinorVersion        Minor Version of ABI.
+
   @retval TRUE            Use FF-A MemPerm ABIs.
   @retval FALSE           Use MM MemPerm ABIs.
 
@@ -43,20 +46,21 @@ STATIC
 BOOLEAN
 EFIAPI
 IsFfaMemoryAbiSupported (
-  IN VOID
+  OUT UINT16  *MajorVersion,
+  OUT UINT16  *MinorVersion
   )
 {
   EFI_STATUS  Status;
-  UINT16      CurrentMajorVersion;
-  UINT16      CurrentMinorVersion;
+
+  *MajorVersion = 0;
+  *MinorVersion = 0;
 
   Status = ArmFfaLibGetVersion (
              ARM_FFA_MAJOR_VERSION,
              ARM_FFA_MINOR_VERSION,
-             &CurrentMajorVersion,
-             &CurrentMinorVersion
+             MajorVersion,
+             MinorVersion
              );
-
   if (EFI_ERROR (Status)) {
     return FALSE;
   }
@@ -147,8 +151,13 @@ SendMemoryPermissionRequest (
 /** Request the permission attributes of a memory region from S-EL0.
 
   @param [in]   UseFfaAbis           Use FF-A abis or not.
+  @param [in]   AbiMajorVersion      ABI Major Version
+  @param [in]   AbiMinorVersion      ABI Minor Version
   @param [in]   BaseAddress          Base address for the memory region.
+  @param [in]   Length               Size of memory region.
   @param [out]  MemoryAttributes     Pointer to return the memory attributes.
+  @param [out]  PageCount            Number of page sharing the same attribute from
+                                     BaseAddress
 
   @retval EFI_SUCCESS             Request successfull.
   @retval EFI_INVALID_PARAMETER   A parameter is invalid.
@@ -165,8 +174,12 @@ STATIC
 EFI_STATUS
 GetMemoryPermissions (
   IN     BOOLEAN               UseFfaAbis,
+  IN     UINT16                AbiMajorVersion,
+  IN     UINT16                AbiMinorVersion,
   IN     EFI_PHYSICAL_ADDRESS  BaseAddress,
-  OUT    UINT32                *MemoryAttributes
+  IN     UINT64                Length,
+  OUT    UINT32                *MemoryAttributes,
+  OUT    UINT32                *PageCount
   )
 {
   EFI_STATUS    Status;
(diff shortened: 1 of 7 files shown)


ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
```diff
@@ -28,19 +28,26 @@
 #include <Library/DebugLib.h>
 #include <Library/PcdLib.h>
 
+typedef enum {
+  ARM_MEM_PERM_SPM_MM,           // Set/Get permission via SPM_MM ABI
+  ARM_MEM_PERM_FFA_PAGE_ONLY,    // Set/Get permission via FF-A ABI with page only.
+  ARM_MEM_PERM_FFA_PAGE_BULK,    // Set/Get permission via FF-A ABI with bulk pages
+  ARM_MEM_PERM_ABI_MAX,
+} ARM_MEM_PERM_ABI;
```
samimujawar: Can we avoid introducing this additional data structure and instead check against the FF-A version itself, please?

samimujawar (changes requested):

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
LeviYeoReum: Okay .. let me try..

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
```diff
@@ -187,11 +200,33 @@ GetMemoryPermissions (
...
   } else {
     *MemoryAttributes = Ret;
+    if (UseFfaAbis && ((AbiMajorVersion > 1) || (AbiMinorVersion > 2))) {
+      /*
+       * The output page count is encoded as (page count - 1),
+       * so add 1 to get the actual page count.
+       * See the FF-A Memory Management Protocol Spec version 1.3-ALP1
+       * 2.8 FFA_MEM_PERM_GET
+       */
+      *PageCount = SvcArgs.Arg3 + 1;
```
samimujawar, line 226: Maybe adding an assert to check that the returned page count is not greater than the number of pages that cover the length would be useful.

samimujawar (approved):
Minor suggestion, otherwise this patch looks good to me.

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
LeviYeoReum: Okay. I'll add the assert. Thanks!

LeviYeoReum:
Gentle ping @kuqin12

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
```diff
@@ -35,6 +35,9 @@
   StMM Core invokes this library before constructors are called and before the
   StMM image itself is relocated.
 
+  @param[out]     MajorVersion        Major Version of ABI.
```
leiflindholm: OK, arguably it's unfair of me to call this out in this PR, but it does somewhat highlight the fact that we've come to treat major and minor version of the ABI as two unrelated 16-bit variables as opposed to the upper and lower 16-bits of a 32-bit version variable.

Is there any way we can nip this in the bud, update the ArmFfaLibGetVersion () definition and use the ARM_FFA_<MAJOR|MINOR>_VERSION_GET macros instead?

Yes, that does turn the
` if (UseFfaAbis && ((AbiMajorVersion > 1) || (AbiMinorVersion > 2))) {`
style checks into a bit of a handful _but_ that's already an error-prone pattern that would be better hidden behind a helper function or macro ->
` if (UseFfaAbis && (ArmFfaAbiMinimum (1, 3)) {`.

leiflindholm (changes requested):

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
LeviYeoReum: Do you mean you want to change the input output of ArmFfaGetVersion() like?:
```
EFI_STATUS
EFIAPI
ArmFfaLibGetVersion (
  IN UINT32 RequestVersion,
  OUT UINT32 *CurrentVersion
  );
```

TBH I agree for this and it might be good to make macro like for `ARM_FFA_ABI_MINIMUM(version, 1, 3)`.
unless other diagress this.

@kuqin12 ?

leiflindholm: > Do you mean you want to change the input output of ArmFfaGetVersion() like?:

Yeah, exactly.

leiflindholm (approved):
Thanks for the late rework!

kuqin12 (approved):

mikebeaton:
Heads up, it seems the push failed here.

leiflindholm:
@mergify requeue

leiflindholm:
Apologies, the UI tricked me and snuck in a merge commit. Rebased and trying again.

leiflindholm:
@mergify requeue

ardbiesheuvel:
Apparently mergify is having issues, so I merged this by hand.

PR #12436, opened 2026-04-10
Remove carriage returns

jeremybobbin:
# Description

In this patch, most of the carriage return line-endings are removed like this:

```
find -type f | xargs file | awk -F: '/CRLF/ {print $1}' | xargs dos2unix
```

The remaining were found with `grep -REl '
$'` & reviewed individually.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

I built OVMF & ran it in QEMU to make sure that the shell, the menu & bootable ISOs worked the same.

## Integration Instructions

N/A


osteffenrh:
Hi @jeremybobbin,
may I ask why?
Maybe I missed something.

 In any case it is always very useful to explain the motivations behind changes in the PR message. thanks.

kilaterlee:
> Hi @jeremybobbin, may I ask why? Maybe I missed something.
> 
> In any case it is always very useful to explain the motivations behind changes in the PR message. thanks.

I guess he can't use the `git am` merge patch on Linux. I'd like to say: you can run the command `git config am.keepcr 1` under the edk2 folder, and then you can use the `git am` merge patch.

osteffenrh:
See also `BaseTools/Scripts/SetupGit.py`

ardbiesheuvel:
Please don't throw 3 million line PRs around without discussing the topic first - thanks.

jeremybobbin:
Thanks for your feedback & sorry for the extra pings - that should have been a draft PR.
My preferred editor sadly doesn't handle carriage-returns well.

ardbiesheuvel:
> Thanks for your feedback & sorry for the extra pings - that should have been a draft PR. My preferred editor sadly doesn't handle carriage-returns well.

No worries. Most of us would prefer the repo to use LF only, and let Git handle the LF to CR-LF conversion on the fly on systems that need it. Unfortunately, the associated churn is slightly prohibitive at this point.

PR #12439, opened 2026-04-10
NetworkPkg/IScsiDxe: Fix AttemptName cleared on loading defaults

AbuthahirM11:
The AttemptName field in the iSCSI configuration form gets wiped out
when optimized defaults are loaded (F3), replacing names like
"Attempt 1", "Attempt 2" with blank strings.

**Root Cause:**
The AttemptName string field in IScsiConfigVfr.vfr had default = STRING_TOKEN(STR_NULL), which resolves to an empty string. When defaults are loaded, the form browser resets every field to its defined default, clearing all attempt names.

The AttemptName is driver-assigned as the identifier for the attempt — it is not a user-configurable setting like target IP or port. The STR_NULL default was a VFR formality (string fields require a default token in VFR syntax), not the intended behaviour. An empty AttemptName leaves the form with a blank title, making it impossible to distinguish between attempts in the UI.

**Fix:**

Add the INTERACTIVE flag to the AttemptName string field in VFR so the form browser invokes the driver callback when loading defaults instead of applying the hardcoded empty string.
Introduce a helper function IScsiGetAttemptNameDefault() that validates Type == EFI_IFR_TYPE_STRING, converts the stored attempt name to a HII string, and sets Value->string.
In IScsiFormCallback, handle EFI_BROWSER_ACTION_DEFAULT_STANDARD, EFI_BROWSER_ACTION_DEFAULT_MANUFACTURING, and EFI_BROWSER_ACTION_DEFAULT_SAFE for KEY_ATTEMPT_NAME by delegating to the helper. This ensures loading any class of defaults always restores the correct attempt name.

Cc: Saloni Kasbekar saloni.kasbekar@intel.com
Cc: Zachary Clark-williams zachary.clark-williams@intel.com

Signed-off-by: Abuthahir M abuthahirm@ami.com

### DIFF
--- a/NetworkPkg/IScsiDxe/IScsiConfig.c
+++ b/NetworkPkg/IScsiDxe/IScsiConfig.c
@@ -3416,6 +3416,22 @@ IScsiFormCallback (
     return EFI_SUCCESS;
   }
 
+  if (Action == EFI_BROWSER_ACTION_DEFAULT_STANDARD) {
+    if ((QuestionId == KEY_ATTEMPT_NAME) && (Value != NULL) && (mCallbackInfo->Current != NULL)) {
+      CHAR16         TmpName[ATTEMPT_NAME_SIZE];
+      EFI_STRING_ID  StringId;
+
+      AsciiStrToUnicodeStrS (mCallbackInfo->Current->AttemptName, TmpName, ATTEMPT_NAME_SIZE);
+      StringId = HiiSetString (mCallbackInfo->RegisteredHandle, 0, TmpName, NULL);
+      if (StringId == 0) {
+        return EFI_OUT_OF_RESOURCES;
+      }
+
+      Value->string = StringId;
+      return EFI_SUCCESS;
+    }
+  }
+
   if ((Action != EFI_BROWSER_ACTION_CHANGING) && (Action != EFI_BROWSER_ACTION_CHANGED)) {
     //
     // All other type return unsupported.
--- a/NetworkPkg/IScsiDxe/IScsiConfig.c
+++ b/NetworkPkg/IScsiDxe/IScsiConfig.c
@@ -3357,6 +3357,44 @@ Exit:
 }
 
 /**
+  Restore the current attempt name when the form browser loads defaults.
+  The AttemptName is driver-assigned and should not be reset to empty
+  when defaults are loaded.
+
+  @param[in]      Type    The type of the question value.
+  @param[in, out] Value   On return, contains the HII string ID of the
+                          current attempt name.
+
+  @retval EFI_SUCCESS           AttemptName was restored successfully.
+  @retval EFI_UNSUPPORTED       Type is not EFI_IFR_TYPE_STRING, or no
+                                current attempt is selected.
+  @retval EFI_OUT_OF_RESOURCES  Failed to register the HII string.
+**/
+EFI_STATUS
+IScsiGetAttemptNameDefault (
+  IN      UINT8               Type,
+  IN OUT  EFI_IFR_TYPE_VALUE  *Value
+  )
+{
+  CHAR16         TmpName[ATTEMPT_NAME_SIZE];
+  EFI_STRING_ID  StringId;
+
+  if (Type != EFI_IFR_TYPE_STRING) {
+    return EFI_UNSUPPORTED;
+  }
+
+  AsciiStrToUnicodeStrS (mCallbackInfo->Current->AttemptName, TmpName, ATTEMPT_NAME_SIZE);
+
+  StringId = HiiSetString (mCallbackInfo->RegisteredHandle, 0, TmpName, NULL);
+  if (StringId == 0) {
+    return EFI_OUT_OF_RESOURCES;
+  }
+
+  Value->string = StringId;
+  return EFI_SUCCESS;
+}
+
+/**
 
   This function is called to provide results data to the driver.
   This data consists of a unique key that is used to identify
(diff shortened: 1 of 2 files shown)


AbuthahirM11:
@leiflindholm @ajfish @ardbiesheuvel @makubacki @mdkinney @mikebeaton @Zclarkwilliams — Could you please take a look and approve this PR when you get a chance?

AbuthahirM11:
@leiflindholm @ajfish @ardbiesheuvel @makubacki @mdkinney @mikebeaton @Zclarkwilliams — Could you please take a look and approve this PR when you get a chance?

mikebeaton:
> @leiflindholm @ajfish @ardbiesheuvel @makubacki @mdkinney @mikebeaton @Zclarkwilliams — Could you please take a look and approve this PR when you get a chance?

In this case, I don't know my way round these forms well enough to know the desired behaviour - specifically, IDK whether and why attempt name should be kept when defaults are loaded...

AbuthahirM11:
@mikebeaton The AttemptName (e.g., "Attempt 1") is auto-assigned by the driver as the identifier for the attempt — it is not a user-configurable setting in the same sense as target IP or port. The VFR defined default = STRING_TOKEN(STR_NULL) (empty string) only because string fields require a default token in VFR syntax, not because an empty name is the intended state.

Pressing F3 is meant to restore configuration parameters to factory defaults, not erase the identity of the attempt. An empty AttemptName leaves the form page with a blank title, making it impossible to distinguish between attempts in the UI.

So the fix preserves the name on "Load Defaults" because the name was never meant to be part of what gets defaulted — the empty STR_NULL default was a VFR formality, not the intended behaviour.

mikebeaton (approved):

AbuthahirM11:
@leiflindholm @ajfish @ardbiesheuvel @makubacki @mdkinney @Zclarkwilliams — Could you please take a look and approve this PR when you get a chance?

NetworkPkg/IScsiDxe/IScsiConfig.c
```diff
@@ -3416,6 +3416,22 @@ IScsiFormCallback (
     return EFI_SUCCESS;
   }
 
+  if (Action == EFI_BROWSER_ACTION_DEFAULT_STANDARD) {
```
makubacki: Should this also cover other classes like `EFI_BROWSER_ACTION_DEFAULT_MANUFACTURING` and `EFI_BROWSER_ACTION_DEFAULT_SAFE`?

NetworkPkg/IScsiDxe/IScsiConfig.c
```diff
@@ -3416,6 +3416,22 @@ IScsiFormCallback (
...
+      CHAR16         TmpName[ATTEMPT_NAME_SIZE];
+      EFI_STRING_ID  StringId;
+
+      AsciiStrToUnicodeStrS (mCallbackInfo->Current->AttemptName, TmpName, ATTEMPT_NAME_SIZE);
+      StringId = HiiSetString (mCallbackInfo->RegisteredHandle, 0, TmpName, NULL);
+      if (StringId == 0) {
+        return EFI_OUT_OF_RESOURCES;
+      }
+
+      Value->string = StringId;
```
makubacki: Afaict, this assumes `Type` is  `EFI_IFR_TYPE_STRING` so it considered safe to assign `StringId` to the `string` member of the `Value` (`EFI_IFR_TYPE_VALUE`) union.

If that's the case, it would be better to explicitly check that. For example:

```c
if (Type != EFI_IFR_TYPE_STRING) {
  return EFI_UNSUPPORTED;
}
```

NetworkPkg/IScsiDxe/IScsiConfig.c
```diff
@@ -3416,6 +3416,22 @@ IScsiFormCallback (
     return EFI_SUCCESS;
   }
 
+  if (Action == EFI_BROWSER_ACTION_DEFAULT_STANDARD) {
+    if ((QuestionId == KEY_ATTEMPT_NAME) && (Value != NULL) && (mCallbackInfo->Current != NULL)) {
```
makubacki, line 3461: You might want to consider moving most of this to a helper function. That will allow the declarations to remain at the top of the (helper) function and prevent growing this already long function by as much.

NetworkPkg/IScsiDxe/IScsiConfig.c
AbuthahirM11: @makubacki Done — updated to handle all three: EFI_BROWSER_ACTION_DEFAULT_STANDARD, EFI_BROWSER_ACTION_DEFAULT_MANUFACTURING, and EFI_BROWSER_ACTION_DEFAULT_SAFE.

NetworkPkg/IScsiDxe/IScsiConfig.c
AbuthahirM11: @makubacki Done — added the Type != EFI_IFR_TYPE_STRING guard in the helper function before accessing Value->string.

NetworkPkg/IScsiDxe/IScsiConfig.c
AbuthahirM11: @makubacki Done — extracted the logic into a new helper function IScsiGetAttemptNameDefault() placed just before IScsiFormCallback. The helper handles the type check, string conversion, HII string registration, and sets Value->string. IScsiFormCallback now just delegates to it.

AbuthahirM11:
@leiflindholm @ajfish @ardbiesheuvel @makubacki @mdkinney @Zclarkwilliams — Could you please take a look and approve this PR when you get a chance?

makubacki:
@AbuthahirM11, thanks for addressing the review feedback.

makubacki (approved):

PR #12443, opened 2026-04-10
Fix invalid UTF-8

jeremybobbin:
# Description

This patch fixes the invalid UTF-8 sequences which caused me some problems with CI(in the attempt to remove the carriage returns).

Some of the CI scripts(like the following) read the diff as UTF-8:
- `.pytool\Plugin\EccCheck\EccCheck.py`
- `.pytool\Plugin\LicenseCheck\LicenseCheck.py`

This means that if invalid UTF-8 was removed in a patch, it will raise an exception.

I do not anticipate this passing CI, however, if merged, it should prevent future CI problems.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested with OVMF in QEMU. I ran the affected make files to assert that there were no new parsing errors. I built `antlr` & `dlg` with no issues.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Source/C/VfrCompile/Pccts/CHANGES_FROM_133.txt
+++ b/BaseTools/Source/C/VfrCompile/Pccts/CHANGES_FROM_133.txt
@@ -40,7 +40,7 @@ List of Implemented Fixes and Changes for Maintenance Releases of PCCTS
 #309. (Changed in MR32) Renamed baseName because of VMS name conflict
 
     Renamed baseName to pcctsBaseName to avoid library name conflict with
-    VMS library routine.  Reported by Jean-François PIÉRONNE.
+    VMS library routine.  Reported by Jean-FranÃ§ois PIÃ‰RONNE.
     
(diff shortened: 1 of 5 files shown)


spbrogan:
we have a plugin that checks for this automatically across files in packages.
https://github.com/tianocore/edk2/tree/master/.pytool/Plugin/CharEncodingCheck
  
Could that be used somehow to not only catch these issues but prevent regressions?  

At first glance i think there are two issues. 
1. Basetools isn't a "package".  But maybe it should be somehow added to CI so that these tools can be run. 
2. These files are not standard file extensions and thus may not have a rule in the CharEncodingCheck tool.  

Anyway, look forward to seeing if there is a way to avoid these problems coming back.

jeremybobbin:
It seems that the existing CI prevents these issues, though not elegantly. For me, finding the bad text involved searching for the byte offset(mentioned in the decoding error) in the diff.

I think this could be useful though I'm not sure it will be worth the effort.

PR #12447, opened 2026-04-13
CryptoPkg: Add AES-CTR APIs to BaseCryptLib

baranee:
# Description

Add AesCtrEncrypt() & AesCtrDecrypt() for AES-CTR (128/192/256, IV=16). 
Provide OpenSSL/MbedTLS/Null implementations via EDKII_CRYPTO_PROTOCOL (v26). 
Add per-service PCD enables and unit tests (NIST SP 800-38A vectors).

- [ ] Breaking change?
- [ ] Impacts security?
- [x] Includes tests?

## How This Was Tested

Verified by calling AES CTR APIs within a SMM module in OVMF for test

## Integration Instructions

N/A


baranee:
@Flickdm @jyao1 @liyi77 - checking if you had a chance to review this PR and any feedback that I need to address? All the CI checks are passing. Thanks.

jyao1:
Would you please describe the use case for AES-CTR?
I am not sure why this API is needed.

I ask this question, because I do not believe we need to add ALL possible AES mode just for completeness reason.
APIs are added only because of a valid use case. And we would like to ensure using AES-CTR is a right choice in this specific use case.

Please help us understand more on that.

baranee:
AES-CTR is chosen to provide confidentiality within our existing design, where integrity and authentication are handled via HMAC, so this change adds only confidentiality.

baranee:
Is there any other concern on this PR?

baranee:
Since there is no feedback from @jyao1, 

@Flickdm @liyi77 - Could you please help to review and provide feedback if there are any concerns?

liyi77:
> AES-CTR is chosen to provide confidentiality within our existing design, where integrity and authentication are handled via HMAC, so this change adds only confidentiality.

This feels too outdated to me. Given that AEAD implementations already exist, I don't think there is any need to introduce such a feature.

baranee:
This is required for compatibility with our existing AES‑CTR + HMAC design used in current platforms.
Are there specific technical concerns with the implementation?

baranee:
Any chance of accepting AES-CTR support in CryptoPkg? We currently maintain this out of tree, and upstreaming it would reduce our ongoing maintenance effort.

PR #12457, opened 2026-04-14
SecurityPkg/Tpm2DeviceLibFfa: add FF-A direct request v1 support

artkat-arm:
Detect whether the TPM service supports FF-A direct request v1 or v2, and route requests through the matching FF-A ABI.

Use cached partition information when available, and keep using direct request v2 when PcdTpmServiceFfaPartitionId already identifies the TPM service partition. Refactor the request path so the transport logic is shared across interface version, feature, start, and notification operations.

# Description

## How This Was Tested

## Integration Instructions

N/A


samimujawar:
@artkat-arm I don’t believe we should enable this support.

In EDK2, FF-A and Standalone MM are implemented with specific constraints:
1. The supported FF-A version is >= 1.2
2. Direct Request V1 is strictly reserved for MM_COMMUNICATE
3. Direct Request V2 is used for targeting specific services

Given that Standalone MM can host multiple services within a single partition, using Direct Request V1 would not allow us to reliably identify the intended target service. As such, enabling this support would be inconsistent with the current design and constraints.

PR #12459, opened 2026-04-14
MdePkg: Fix Spelling Errors and Enforce Spell Check

os-d:
# Description

Currently, there are some thousands of spell check failures in MdePkg. The spell check plugin is also in audit only mode, so new failures are not caught. Thousands of lines of the CI build log are dedicated to spelling failures and are very annoying to scroll through. After this change, the MdePkg CI build log is 7,500 lines shorter.

Each associated area is split up into a separate commit where actual code changes are made (e.g. defines, structure fields, etc.). Where relevant, consumers are updated in the same commit.

All non-code change spelling fixes (i.e. comments) are in the final commit, which also adds all the words to ignore to the global cspell dictionary. The final commit disables AuditOnly in MdePkg so that new spelling errors will break the build and be caught.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

N/A.

## Integration Instructions

Some defines/structures/enums, etc. have had spelling fixes applied to them. Consumers should update accordingly. None of these are currently breaking changes, but will be removed in the future.


leiflindholm:
OK, the tree just changed while I was reviewing.
Please flip to draft and bring it out when you're done.

os-d:
> OK, the tree just changed while I was reviewing. Please flip to draft and bring it out when you're done.

@leiflindholm my apologies, I didn't think anyone had started reviewing it yet, so I fixed a small CI failure. I've switched it to draft, when CI is passing I'll flip it back out.

os-d:
Apparently the newer version of cspell I have (9.2.1) catches fewer words than the version the pipeline installs (5.20). I've installed the pipeline's version and going through the extra failures.

os-d:
@leiflindholm CI is passing so I've moved this out of draft. Please continue your review

.pytool/Plugin/SpellCheck/cspell.base.yaml
```diff
@@ -24,274 +24,1162 @@
     "allowCompoundWords": true,
     "maxNumberOfProblems": 200,
     "maxDuplicateProblems": 200,
-    "ignoreWords": [
-        "muchange"
-    ],
+    "ignoreWords": [],
     "words": [
+        "AABBCCDD",
```
pierregondois, line 29: Should this patch be split in 3 ?
- .pytool/Plugin/SpellCheck/cspell.base.yaml
- MdePkg/MdePkg.ci.yaml
- the rest

?

Also, I assume many of the ignored words added in `.pytool/Plugin/SpellCheck/cspell.base.yaml` are common to many packages, but shouldn't they normally be added to `MdePkg/MdePkg.ci.yaml` ? Otherwise we can just merge all the ignored words from all packages in `.pytool/Plugin/SpellCheck/cspell.base.yaml`.

pierregondois (commented):
Hello Oliver,
Just one comment, also I made https://github.com/tianocore/edk2/pull/12470 to help reducing the output of cspell

.pytool/Plugin/SpellCheck/cspell.base.yaml
os-d: I'm fine to split that patch into three. I'll wait until others have completed their review or paused review to wait for the split.

Yeah, I debated adding to the global ignore list or to the MdePkg specific one. I don't have a big preference, but I chose the global one for two reasons: MdePkg is the foundational package, so it is consumed by every other package. Any words that should be ignored there have a reasonable chance of showing up in other packages. Secondly, I don't think it is valuable to have to keep adding words we've already decided are fine (e.g. that first one AABBCCDD is part of an example GUID, that could show up anywhere in the codebase, no point to have to add it to each ci.yaml file).

So in general, I'm in favor of using the global ignore list, unless something is really just for a specific package (e.g. if a maintainer wants to disable audit mode but not fix all of the errors immediately or some spec has a misspelling in a definition and so that package needs to keep the misspelling until the spec is fixed).

pierregondois: Ok sure, works for me.
Just one other remark would be for the `edk2-platforms` repository: would there be a similar update ?

os-d: Yeah, I will put up a PR.

os-d:
Put up a PR to fix edk2-platforms after this goes in: https://github.com/tianocore/edk2-platforms/pull/977. Despite all the changes, the blast radius is pretty low, a lot of the definition changes are not widely used.

.pytool/Plugin/SpellCheck/cspell.base.yaml
```diff
@@ -24,274 +24,1162 @@
...
+        "otech",
+        "outpd",
+        "outpw",
+        "outsb",
+        "outsd",
+        "outsw",
         "OVMF",
+        "ovrpat",
+        "owpass",
+        "oxffffffff",
```
mikebeaton: Hi. Do we need to add this? I can find two instances of it in comments, but they should be changed to 0xffffffff.

MdePkg/Include/Library/UefiApplicationEntryPoint.h
```diff
@@ -19,7 +19,7 @@ extern CONST UINT32  _gUefiDriverRevision;
   This function is the entry point for a UEFI Application. This function must call
   ProcessLibraryConstructorList(), ProcessModuleEntryPointList(), and ProcessLibraryDestructorList().
   The return value from ProcessModuleEntryPointList() is returned.
-  If _gUefiDriverRevision is not zero and SystemTable->Hdr.Revision is less than _gUefiDriverRevison,
+  If _gUefiDriverRevision is not zero and SystemTable->Hdr.Revision is less than _gUefiDriverrevision,
```
mikebeaton: We've lost a capital R in the second _gUefiDriverrevision.

MdePkg/Include/Protocol/DiskIo2.h
```diff
@@ -69,7 +69,7 @@ EFI_STATUS
                                                                          Event will be signaled upon completion.
   @retval EFI_DEVICE_ERROR      The device reported an error while performing the write.
   @retval EFI_NO_MEDIA          There is no medium in the device.
-  @retval EFI_MEDIA_CHNAGED     The MediaId is not for the current medium.
+  @retval EFI_MEDIA_changed     The MediaId is not for the current medium.
```
mikebeaton: Case is wrong.

MdePkg/Include/Protocol/DiskIo2.h
```diff
@@ -102,7 +102,7 @@ EFI_STATUS
   @retval EFI_WRITE_PROTECTED   The device cannot be written to.
   @retval EFI_DEVICE_ERROR      The device reported an error while performing the write operation.
   @retval EFI_NO_MEDIA          There is no medium in the device.
-  @retval EFI_MEDIA_CHNAGED     The MediaId is not for the current medium.
+  @retval EFI_MEDIA_changed     The MediaId is not for the current medium.
```
mikebeaton: Case is wrong.

MdePkg/Include/Protocol/DiskIo2.h
```diff
@@ -132,7 +132,7 @@ EFI_STATUS
...
```
mikebeaton: Case is wrong

MdePkg/Include/Protocol/Dns4.h
```diff
@@ -428,7 +428,7 @@ EFI_STATUS
 
   @param[in]  This                Pointer to EFI_DNS4_PROTOCOL instance.
   @param[in]  DeleteFlag          If FALSE, this function is to add one entry to the
-                                  DNS Cahce. If TRUE, this function will delete
+                                  DNS Cache. If TRUE, this function will delete
                                   matching DNS Cache entry.
```
mikebeaton: Not quite what you are changing, but probably (nit, at most) Cache should have a lower-case c in these two lines, as it already is in the immediately following comment for the Override param.

MdePkg/Include/Protocol/Dns6.h
```diff
@@ -421,7 +421,7 @@ EFI_STATUS
 
   @param[in]  This                Pointer to EFI_DNS6_PROTOCOL instance.
   @param[in]  DeleteFlag          If FALSE, this function is to add one entry to the
-                                  DNS Cahce. If TRUE, this function will delete
+                                  DNS Cache. If TRUE, this function will delete
                                   matching DNS Cache entry.
```
mikebeaton: Same again, nit, but both Cache in these two lines to cache.

MdePkg/Include/Protocol/SerialIo.h
```diff
@@ -102,13 +102,13 @@ EFI_STATUS
...
 /**
-  Sets the baud rate, receive FIFO depth, transmit/receice time out, parity,
+  Sets the baud rate, receive FIFO depth, transmit/receive time out, parity,
   data bits, and stop bits on a serial device.
 
   @param  This             Protocol instance pointer.
   @param  BaudRate         The requested baud rate. A BaudRate value of 0 will use the
                            device's default interface speed.
-  @param  ReveiveFifoDepth The requested depth of the FIFO on the receive side of the
+  @param  receiveFifoDepth The requested depth of the FIFO on the receive side of the
```
mikebeaton: Initial r should be caps

MdePkg/Library/BaseSerialPortLibRiscVSbiLib/BaseSerialPortLibRiscVSbiLib.c
```diff
@@ -161,13 +161,13 @@ SerialPortGetControl (
...
 /**
-  Sets the baud rate, receive FIFO depth, transmit/receice time out, parity,
+  Sets the baud rate, receive FIFO depth, transmit/receive time out, parity,
   data bits, and stop bits on a serial device.
 
   @param BaudRate           The requested baud rate. A BaudRate value of 0 will use the
                             device's default interface speed.
                             On output, the value actually set.
-  @param ReveiveFifoDepth   The requested depth of the FIFO on the receive side of the
+  @param receiveFifoDepth   The requested depth of the FIFO on the receive side of the
```
mikebeaton: Should be initial caps R

MdePkg/Library/BaseSerialPortLibRiscVSbiLib/BaseSerialPortLibRiscVSbiLibRam.c
```diff
@@ -242,13 +242,13 @@ SerialPortGetControl (
...
 /**
-  Sets the baud rate, receive FIFO depth, transmit/receice time out, parity,
+  Sets the baud rate, receive FIFO depth, transmit/receive time out, parity,
   data bits, and stop bits on a serial device.
 
   @param BaudRate           The requested baud rate. A BaudRate value of 0 will use the
                             device's default interface speed.
                             On output, the value actually set.
-  @param ReveiveFifoDepth   The requested depth of the FIFO on the receive side of the
+  @param receiveFifoDepth   The requested depth of the FIFO on the receive side of the
```
mikebeaton: Initial caps R

.pytool/Plugin/SpellCheck/cspell.base.yaml
os-d: Good call, I will drop this and fix the usages of it.

MdePkg/Include/Library/UefiApplicationEntryPoint.h
os-d: Thanks for this and the other ones. The script I ran didn't do a good job about maintaining case and I tried to go and fix all of them, clearly missed some.

mikebeaton (changes requested):

os-d:
@mikebeaton thanks for the reviews. Let me know when you are done and I will apply the requested changes from you and @pierregondois to not change the tree out from under you while reviewing.

mikebeaton:
> @mikebeaton thanks for the reviews. Let me know when you are done and I will apply the requested changes from you and @pierregondois to not change the tree out from under you while reviewing.

Done. Thanks.

os-d:
> > @mikebeaton thanks for the reviews. Let me know when you are done and I will apply the requested changes from you and @pierregondois to not change the tree out from under you while reviewing.
> 
> Done. Thanks.

Thanks, for any other reviewers, I recommend pausing until I push up the current requested changes. If anyone else is actively reviewing, let me know, I will hold off pushing changes until your review is finished.

os-d:
@mikebeaton @pierregondois I have addressed your feedback, please re-review when you have a chance.

MdePkg/Include/Library/SerialPortLib.h
```diff
@@ -126,13 +126,13 @@ SerialPortGetControl (
...
 /**
-  Sets the baud rate, receive FIFO depth, transmit/receice time out, parity,
+  Sets the baud rate, receive FIFO depth, transmit/receive time out, parity,
   data bits, and stop bits on a serial device.
 
   @param BaudRate           The requested baud rate. A BaudRate value of 0 will use the
                             device's default interface speed.
                             On output, the value actually set.
-  @param ReveiveFifoDepth   The requested depth of the FIFO on the receive side of the
+  @param receiveFifoDepth   The requested depth of the FIFO on the receive side of the
```
mikebeaton: Another one.

MdePkg/Library/BaseSerialPortLibNull/BaseSerialPortLibNull.c
```diff
@@ -141,13 +141,13 @@ SerialPortGetControl (
...
 /**
-  Sets the baud rate, receive FIFO depth, transmit/receice time out, parity,
+  Sets the baud rate, receive FIFO depth, transmit/receive time out, parity,
   data bits, and stop bits on a serial device.
 
   @param BaudRate           The requested baud rate. A BaudRate value of 0 will use the
                             device's default interface speed.
                             On output, the value actually set.
-  @param ReveiveFifoDepth   The requested depth of the FIFO on the receive side of the
+  @param receiveFifoDepth   The requested depth of the FIFO on the receive side of the
```
mikebeaton: And another

mikebeaton: The rest of the changes look good - thanks

ardbiesheuvel:
Does enabling cspell result in yet another unfathomable error deep in the bowels of Azure, resulting in another hurdle for casual contributors? Is it possible for CI errors to be reported more clearly so mere mortals can figure why GitHub said no?

os-d:
> Does enabling cspell result in yet another unfathomable error deep in the bowels of Azure, resulting in another hurdle for casual contributors? Is it possible for CI errors to be reported more clearly so mere mortals can figure why GitHub said no?

@ardbiesheuvel we already have cspell running and reporting errors. For some packages it already fails the build as well; MdePkg is not currently one of those.

The top level CI log gives the details on what words failed, for example, I missed a word in this PR and here is what it shows:

image

https://dev.azure.com/tianocore/edk2-ci/_build/results?buildId=242141&view=logs&j=5f93e2ab-c680-598c-a6fe-4cf82ad4602b&t=3474e429-5030-58a2-a3d8-8017216fd2c4

From there it is easy to either add the word(s) to the global dictionary or realize it is a spelling error and fix it.

Let me know if that seems reasonable to diagnose. It is reproducible locally, too, of course.

ardbiesheuvel:
> > Does enabling cspell result in yet another unfathomable error deep in the bowels of Azure, resulting in another hurdle for casual contributors? Is it possible for CI errors to be reported more clearly so mere mortals can figure why GitHub said no?
> 
> @ardbiesheuvel we already have cspell running and reporting errors. For some packages it already fails the build as well; MdePkg is not currently one of those.
> 


> The top level CI log gives the details on what words failed, for example, I missed a word in this PR and here is what it shows:
> 
...
> 
> From there it is easy to either add the word(s) to the global dictionary or realize it is a spelling error and fix it.
> 
> Let me know if that seems reasonable to diagnose. It is reproducible locally, too, of course.

Fair enough. I'm not a huge fan (as you might have guessed :-)) but if others feel this has value, I am not going to oppose.

os-d:
> Fair enough. I'm not a huge fan (as you might have guessed :-)) but if others feel this has value, I am not going to oppose.

I know :). The value to me is ensuring that all defines, structure fields, etc., are correctly spelled so they make sense to use. For example, every time I try to use the PCI_CAPABILITY_POINTER_OFFSET by memory and then have the build fail because that's not defined, I have to look it up, and go, oh, right, its misspelled and I need to use PCI_CAPBILITY_POINTER_OFFSET. Particularly for MdePkg, which has a lot of spec defined structures, if someone is looking up something from a spec and can't find it because it is misspelled, that is bad.

Also, I think the current state is the worst state, where we run cspell then ignore it. As you may have seen, I really do not like build logs filled with useless stuff :). And so, this PR trimming 7500 lines from the build log is another reason to enable it, imo.

Javagedes (approved):
Approve for .pytool/*

kilaterlee (approved):
Approve for LoongArch*. Thanks!

pierregondois (approved):

pierregondois:
> > Does enabling cspell result in yet another unfathomable error deep in the bowels of Azure, resulting in another hurdle for casual contributors? Is it possible for CI errors to be reported more clearly so mere mortals can figure why GitHub said no?
> 
> @ardbiesheuvel we already have cspell running and reporting errors. For some packages it already fails the build as well; MdePkg is not currently one of those.
> 
> The top level CI log gives the details on what words failed, for example, I missed a word in this PR and here is what it shows:
> image
> 
> https://dev.azure.com/tianocore/edk2-ci/_build/results?buildId=242141&view=logs&j=5f93e2ab-c680-598c-a6fe-4cf82ad4602b&t=3474e429-5030-58a2-a3d8-8017216fd2c4
> 
> From there it is easy to either add the word(s) to the global dictionary or realize it is a spelling error and fix it.
> 
> Let me know if that seems reasonable to diagnose. It is reproducible locally, too, of course.

It's true that it is a bit annoying to add cspell exceptions to un-related PRs.
As most words that are triggering the tool are likely to be found in multiple packages, it would be nice to only modify `.pytool/Plugin/SpellCheck/cspell.base.yaml` for the whole project instead of relying on the per-package files.

mikebeaton (approved):
I agree with @pierregondois that it would be nice to unify this a bit more - e.g. the extra word that sneaked into this PR was required, I believe, because of code moved from ArmPkg, where it was already marked as an exception. (And possibly is now marked as an exception even though no longer needed.)

But I do agree this has value, since it avoids genuinely unwanted (and worth fixing) typos exactly like the ones fixed here.

os-d:
> I agree with @pierregondois that it would be nice to unify this a bit more - e.g. the extra word that sneaked into this PR was required, I believe, because of code moved from ArmPkg, where it was already marked as an exception. (And possibly is now marked as an exception even though no longer needed.)
> 
> But I do agree this has value, since it avoids genuinely unwanted (and worth fixing) typos exactly like the ones fixed here.

Yeah, that’s my goal with adding the words to the global exception list and moving forward I think we should continue to do that (unless there is a good reason to scope it to a package).

I think a follow on PR could certainly move the per package words to the global dictionary.

os-d:
@mdkinney @lgao4 can you review when you have a chance? Looks like other reviews have finished.

mdkinney:
No issues with changes to comments.

My biggest concern is changes to structure field names and define names.  This is marked as a breaking change, but this could break 100% of the downstream consumers with this 1 PR.

os-d:
> No issues with changes to comments.
> 
> My biggest concern is changes to structure field names and define names. This is marked as a breaking change, but this could break 100% of the downstream consumers with this 1 PR.

I made a PR in edk2-platforms to fix this after this goes in: https://github.com/tianocore/edk2-platforms/pull/977. As you can  see, at least breakage there is pretty light. I would believe that other downstream platforms would have similar usages to what we see there, i.e. that the usage should be relatively light. A lot of these misspellings are in definitions from specs in fields that aren't widely used.

This will break downstream platforms, but it also adds the enforcement to make sure we get it right in the future. Right now, having all of these spec defined structures, etc. be misspelled makes it hard to correlate between the given spec and the edk2 implementation.

The fix is also simple: take the correct spelling. No changes to any code flows. I believe this is a worthwhile one time breaking change to make sure we get it right in the future.

makubacki:
@mdkinney, in your opinion, what is the right way to stage these fixes for platforms to integrate the breaking changes?

mdkinney:
> > No issues with changes to comments.
> > My biggest concern is changes to structure field names and define names. This is marked as a breaking change, but this could break 100% of the downstream consumers with this 1 PR.
> 
> I made a PR in edk2-platforms to fix this after this goes in: [tianocore/edk2-platforms#977](https://github.com/tianocore/edk2-platforms/pull/977). As you can see, at least breakage there is pretty light. I would believe that other downstream platforms would have similar usages to what we see there, i.e. that the usage should be relatively light. A lot of these misspellings are in definitions from specs in fields that aren't widely used.
> 
> This will break downstream platforms, but it also adds the enforcement to make sure we get it right in the future. Right now, having all of these spec defined structures, etc. be misspelled makes it hard to correlate between the given spec and the edk2 implementation.
> 
> The fix is also simple: take the correct spelling. No changes to any code flows. I believe this is a worthwhile one time breaking change to make sure we get it right in the future.

That makes it even worse because downstreams that use both edk2 and edk2-platforms content have to sync 2 repos simultaneously.  If they sync one at a time, CI fails.  Torn state.

mdkinney:
> @mdkinney, in your opinion, what is the right way to stage these fixes for platforms to integrate the breaking changes?

The ideal method is to have 1 PR that introduces new names and preserves compatibility with old names. Warning messages about use of old names preferred.  This give downstream time to migrate from old names to new names until all warnings addressed.  Then after a period of time determine by community feedback, a different PR can remove the old name support and the warning message support.

ardbiesheuvel:
For struct fields, there is not a lot we can do, but couldn't we easily provide both the old and the new (fixed) definitions for enums etc? That way, we can deprecate the old ones at a later date (if we decide to bother)

makubacki:
> The ideal method is to have 1 PR that introduces new names and preserves compatibility with old names. Warning messages about use of old names preferred. This give downstream time to migrate from old names to new names until all warnings addressed. Then after a period of time determine by community feedback, a different PR can remove the old name support and the warning message support.

Thanks. This is what I was going to recommend as well with an update after the upcoming stable being a candidate for the point to remove old name support.

os-d:
I was going to suggest the same thing (with Ard's note that for struct fields I don't think there is anything differently we can do). But I can update this PR to add new definitions where possible (just ignoring the spelling errors in MdePkg.ci.yaml for now so we can still enable the check to catch future errors).

mikebeaton:
We've had PRs to edk2 before which have broken edk2-platforms. No one seemed to mind too much (as long as it was clear it was a sane and worthwhile fix and) as long as edk2-platforms gets updated to match straight away. It is not completely clear to me that the gold-plated approach is needed here.

gguo11837463 (approved):

os-d:
@mdkinney I have updated to have backwards compatibility for everything outside of struct field names (which I have listed the set in the Integration Instructions section to make it easy for platforms to integrate if they use any these, so it seems usage is light). Please re-review when you have a chance

MdePkg/Include/IndustryStandard/TcgStorageOpal.h
```diff
@@ -115,7 +115,7 @@ typedef enum {
   Unmap,
   ResetWritePointers,
   VendorSpecificErase,
-  ResearvedMechanism
+  ReservedMechanism
```
ardbiesheuvel: Please make this

```diff
-  ResearvedMechanism
+  ReservedMechanism,
+  ResearvedMechanism = ReservedMechanism

```

MdePkg/Include/IndustryStandard/Tpm12.h
```diff
@@ -1740,7 +1740,7 @@ typedef struct tdTPM_DELEGATIONS {
 #define TPM_DELEGATE_CreateMaintenanceArchive        (BIT21)
 #define TPM_DELEGATE_LoadMaintenanceArchive          (BIT20)
 #define TPM_DELEGATE_KillMaintenanceFeature          (BIT19)
-#define TPM_DELEGATE_OwnerReadInteralPub             (BIT18)
+#define TPM_DELEGATE_OwnerReadInternalPub            (BIT18)
```
ardbiesheuvel:
```diff
 #define TPM_DELEGATE_OwnerReadInteralPub             (BIT18)
+#define TPM_DELEGATE_OwnerReadInternalPub            (BIT18)
```

(i.e., keep the old definition)

MdePkg/Include/IndustryStandard/Atapi.h
```diff
@@ -456,7 +456,7 @@ typedef union {
   UINT16                       Data16[6];
   ATAPI_TEST_UNIT_READY_CMD    TestUnitReady;
   ATAPI_READ10_CMD             Read10;
-  ATAPI_REQUEST_SENSE_CMD      RequestSence;
+  ATAPI_REQUEST_SENSE_CMD      RequestSense;
```
ardbiesheuvel, line 462: This is union, not a struct, so we can keep both fields.

MdePkg/Include/IndustryStandard/SmBios.h
```diff
@@ -1815,15 +1815,15 @@ typedef enum {
...
-  EventLogVariableOEMAssigned                = 0x80
+  EventLogVariableNone                        = 0x00,
+  EventLogVariableHandle                      = 0x01,
+  EventLogVariableMultiEvent                  = 0x02,
+  EventLogVariableMultiEventHandle            = 0x03,
+  EventLogVariablePOSTResultBitmap            = 0x04,
+  EventLogVariableSysManagementType           = 0x05,
+  EventLogVariableMultiEventSysManagementType = 0x06,
+  EventLogVariableUnused                      = 0x07,
+  EventLogVariableOEMAssigned                 = 0x80
```
ardbiesheuvel: Please keep the old ones as aliases (below as well)

ardbiesheuvel (commented):
I've commented on a couple of patches, but the general gist is that we should try much harder to retain the old definitions.

os-d:
> I've commented on a couple of patches, but the general gist is that we should try much harder to retain the old definitions.

Thanks, I will fix these and go back through and make sure I didn't miss others.

mdkinney:
> @mdkinney I have updated to have backwards compatibility for everything outside of struct field names (which I have listed the set in the Integration Instructions section to make it easy for platforms to integrate if they use any these, so it seems usage is light). Please re-review when you have a chance

I agree with @ardbiesheuvel.  Have to maintain backwards compatibility for initial version of these changes.  I recommend this large PR be reduced to only the changes to comments and changes that can guarantee backward compatibility.  Items where backward compatibility is hard or impossible should be dropped from this PR and moved to an open issue for further discussion.

os-d:
> > I've commented on a couple of patches, but the general gist is that we should try much harder to retain the old definitions.
> 
> Thanks, I will fix these and go back through and make sure I didn't miss others.

Ah, I see what I happened. I did not push the changes I made to retain the old definitions. My apologies for wasting your time @ardbiesheuvel . I'll review my changes again locally and make sure I didn't miss any, then push it up.

os-d:
@mdkinney @ardbiesheuvel I have removed all breaking changes from this PR and will defer them to go through the new breaking change process (after it is ratified). Please re-review when you have a chance.

MdePkg/Include/IndustryStandard/PciExpress21.h
```diff
@@ -624,8 +624,8 @@ typedef struct {
   UINT8                                       EgressControlVectorArray[1];
 } PCI_EXPRESS_EXTENDED_CAPABILITIES_ACS_EXTENDED;
 
-#define PCI_EXPRESS_EXTENDED_CAPABILITY_ACS_EXTENDED_GET_EGRES_CONTROL(ACS_EXTENDED)      (UINT8)(((ACS_EXTENDED->AcsCapability)&0x00000020))
-#define PCI_EXPRESS_EXTENDED_CAPABILITY_ACS_EXTENDED_GET_EGRES_VECTOR_SIZE(ACS_EXTENDED)  (UINT8)(((ACS_EXTENDED->AcsCapability)&0x0000FF00))
+#define PCI_EXPRESS_EXTENDED_CAPABILITY_ACS_EXTENDED_GET_EGRESS_CONTROL(ACS_EXTENDED)      (UINT8)(((ACS_EXTENDED->AcsCapability)&0x00000020))
```
vishalo, line 627: Do we need to keep the old defines here as well ?

`#define PCI_EXPRESS_EXTENDED_CAPABILITY_ACS_EXTENDED_GET_EGRES_CONTROL \
    PCI_EXPRESS_EXTENDED_CAPABILITY_ACS_EXTENDED_GET_EGRESS_CONTROL`
`#define PCI_EXPRESS_EXTENDED_CAPABILITY_ACS_EXTENDED_GET_EGRES_VECTOR_SIZE \
    PCI_EXPRESS_EXTENDED_CAPABILITY_ACS_EXTENDED_GET_EGRESS_VECTOR_SIZE`

os-d: Yes thanks, missed this one

os-d:
@mdkinney @ardbiesheuvel @lgao4 can we consider this for the stable tag? It is now a non-breaking change with no behavior change.

ardbiesheuvel (approved):

ardbiesheuvel:
> @mdkinney @ardbiesheuvel @lgao4 can we consider this for the stable tag? It is now a non-breaking change with no behavior change.

Approved for stable tag

os-d:
Rebased on master to fix the last couple things that have crept in while this has been up (and a couple things because my local cspell version crept forward again, so reset back to what the pipelines install). Only changes are adding 6 words to the global cspell list and fixing one misspelling in a comment. Fixes the CI failure.

mdkinney:
I see aliases of #defines from misspelled name to correctly spelled name being added.  However, there is no comment that indicates that this is only present for backward compatibility for usage of the misspelled define name and providing a notice that the misspelled name will be removed in a future PR.  Adding a standard comment block for these additions for compatibility will make it easier to notice when looking at the code on which one to use and not to use and will also make it easier to find these when the future PR needs to be generated to removed the support for the misspelled defines.

os-d:
> I see aliases of #defines from misspelled name to correctly spelled name being added. However, there is no comment that indicates that this is only present for backward compatibility for usage of the misspelled define name and providing a notice that the misspelled name will be removed in a future PR. Adding a standard comment block for these additions for compatibility will make it easier to notice when looking at the code on which one to use and not to use and will also make it easier to find these when the future PR needs to be generated to removed the support for the misspelled defines.

Yeah, I added that in some places, but not globally. Agreed that is helpful and I will add that uniformly with the same message.

os-d:
@mdkinney I have added a uniform comment everywhere, please re-review. Also, when I removed the breaking change label, I forgot to update the PR description and so the GH bot re-added the label. I can't remove it because of the soft freeze, would you please remove it? I have fixed the PR description.

mdkinney (approved):
All the changes look good.

mdkinney:
Approved for stable tag

Issue #11578, opened 2025-10-03
[Bug]: Pre-OS graphics fail to load when deleting GPIO driver using FMMT tool


varunupa:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [ ] Firmware
- [x] Tool
- [ ] Unit Test

### What packages are impacted?

BaseTools

### Which targets are impacted by this bug?

RELEASE, DEBUG

### Current Behavior

when we delete GPIO driver from FMMT tool , seeing Pre-OS graphics failed to load as it can't find required PPI for graphics to run
Though removing GPIO driver before building image has no impact on Pre-OS graphics  
FMMT cmd:
FMMT.py -d driver-guid  <input-binaryImage.bin>  <output-binaryImage.bin>

### Expected Behavior

Deleting GPIO driver from FMMT tool should not have impact on Pre-OS graphics
FMMT cmd:
FMMT.py -d driver-guid  <input-binaryImage.bin> <output-binaryImage.bin>

### Steps To Reproduce

1. Build platform image 
2.  Removed GPIO driver with FMMT Tool - "FMMT.py -d driver-guid  <input-binaryImage.bin> <output-binaryImage.bin>"
3. Test Pre-OS graphics with generated image

### Build Environment

```markdown
- OS(s):Ubuntu 
- Tool Chain(s): GCC
```

### Version Information

```text
edk2/master
```

### Urgency

Medium

### Are you going to fix this?

Someone else needs to fix it

### Do you need maintainer feedback?

No maintainer feedback needed

lgao4:
@varunupa , which platform can reproduce this issue? You may find the platform owner to discuss this issue.

PR #12464, opened 2026-04-15
BaseTools/FMMT: Rebase PE/COFF images after DeleteFfs

YuweiChen1110:
When FMMT -d removes an FFS from a Firmware Volume, all subsequent FFS files shift to lower offsets. If those FFS files contain PE32/TE images with pre-built ImageBase values, the ImageBase no longer matches the actual ROM address. This causes PeiCore to detect a non-zero delta at load time, copy the image to heap, and apply relocations - which can corrupt absolute pointers (e.g. PPI GUID pointers) and break XIP execution.

ReplaceFfs() already handles this via _rebase_subsequent_ffs_in_current_fv(), but DeleteFfs() had no rebase logic at all.

Add _rebase_after_delete() which iterates all FFS children after the deleted one, applies address_delta = -deleted_size to their PE32/TE ImageBase and .reloc entries via the existing _rebase_ffs_pe_sections() helper, and recalculates FFS checksums.

The call is placed before the node is removed from the tree so Child.index() can still locate the deleted FFS position.

Linked issue: https://github.com/tianocore/edk2/issues/11578

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested with command "FMMT -d binary_file ffs_name output_file" for most of Intel's platforms for both build process and boot process.

## Integration Instructions

N/A


AshrafAliS (approved):

AshrafAliS:
@lgao4 @gapalomi

lgao4 (approved):

PR #12472, opened 2026-04-16
MdeModulePkg/Library: introduce ArmFfaConsoleDebugLib

LeviYeoReum:
# Description

This patch adds ArmFfaConsoleDebugLib. A debug library that utilizes FF-A Console Log API to print debug messages to the console.

This is useful in context of running the StandaloneMm payload as a S-EL0 SP on top of SPMC where StandaloneMm isn't allowed to access console device or no console device (e.x) Hafnium.

NOTE:
  FFA_CONSOLE_LOG* could be used by secure partition only.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Linking StandaloneMm with this library.
Check log printed correcly.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugStandaloneMmLib.c
+++ b/MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugStandaloneMmLib.c
@@ -1,0 +1,513 @@
+/** @file
+  Base Debug library instance based on Arm FF-A Console log.
+
+  Copyright (c) 2025, Arm Limited. All rights reserved.<BR>
+  Copyright (c) 2006 - 2019, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include <Uefi/UefiBaseType.h>
+
+#include <IndustryStandard/ArmFfaSvc.h>
+#include <IndustryStandard/ArmStdSmc.h>
+
+#include <Library/BaseLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/ArmFfaLib.h>
+#include <Library/ArmSmcLib.h>
+#include <Library/ArmSvcLib.h>
+#include <Library/DebugLib.h>
+#include <Library/PrintLib.h>
+#include <Library/PcdLib.h>
+#include <Library/DebugPrintErrorLevelLib.h>
+
+//
+// Define the maximum debug and assert message length that this library supports
+//
+#define MAX_DEBUG_MESSAGE_LENGTH  0x100
+
+//
+// Define Maxmium available number of registers per FF-A call.
+//
+#define MAX_REG64_COUNT  16
+
+//
+// VA_LIST can not initialize to NULL for all compiler, so we use this to
+// indicate a null VA_LIST
+//
+STATIC VA_LIST  mVaListNull;
+
+/**
+  Convert FFA return code to RETURN_STATUS.
+
+  @param [in] FfaStatus          Ffa return Status
+
+  @retval RETURN_STATUS          return value correspond RETURN_STATUS to FfaStatus
+
+**/
+STATIC
+RETURN_STATUS
+EFIAPI
+FfaStatusToRetStatus (
+  IN UINTN  FfaStatus
+  )
+{
+  switch ((UINT32)FfaStatus) {
+    case ARM_FFA_RET_SUCCESS:
+      return RETURN_SUCCESS;
+    case ARM_FFA_RET_INVALID_PARAMETERS:
+      return RETURN_INVALID_PARAMETER;
+    case ARM_FFA_RET_NO_MEMORY:
+      return RETURN_OUT_OF_RESOURCES;
+    case ARM_FFA_RET_BUSY:
+      return RETURN_NO_RESPONSE;
+    case ARM_FFA_RET_DENIED:
+      return RETURN_ACCESS_DENIED;
+    case ARM_FFA_RET_ABORTED:
+      return RETURN_ABORTED;
+    case ARM_FFA_RET_NODATA:
+      return RETURN_NOT_FOUND;
+    case ARM_FFA_RET_NOT_READY:
+      return RETURN_NOT_READY;
+    case ARM_FFA_RET_RETRY:
+      return RETURN_TIMEOUT;
+    default:
+      return RETURN_UNSUPPORTED;
+  }
+}
+
+/**
+  Convert FfArgs to RETURN_STATUS.
+
+  @param [in] FfaArgs            Ffa arguments
+
+  @retval RETURN_STATUS           return value correspond RETURN_STATUS to FfaStatus
+
+**/
+STATIC
+RETURN_STATUS
+EFIAPI
+FfaArgsToRetStatus (
+  IN ARM_FFA_ARGS  *FfaArgs
+  )
+{
+  UINT32  FfaStatus;
+
+  if (FfaArgs == NULL) {
+    FfaStatus = ARM_FFA_RET_INVALID_PARAMETERS;
+  } else if (IS_FID_FFA_ERROR (FfaArgs->Arg0)) {
+    /*
+     * In case of error, the Arg0 will be set to the fid FFA_ERROR.
+     * and Error code is set in Arg2.
+     */
+    FfaStatus = FfaArgs->Arg2;
+  } else if (FfaArgs->Arg0 == ARM_FFA_RET_NOT_SUPPORTED) {
+    /*
+     * If Some FF-A ABI doesn't support, it sets ARM_FFA_RET_NOT_SUPPORTED
+     * in Arg0 and other register has no meaning.
+     * In this case, set Arg2 as ARM_FFA_RET_NOT_SUPPORTED so that
+     * FfaStatusToEfiStatus (FfaARgs.Arg2) returns proper EFI_STATUS.
+     */
+    FfaStatus = ARM_FFA_RET_NOT_SUPPORTED;
+  } else {
+    FfaStatus = ARM_FFA_RET_SUCCESS;
+  }
+
+  return FfaStatusToRetStatus (FfaStatus);
+}
+
+/**
+  Trigger FF-A ABI call according to PcdFfaLibConduitSmc.
+
+  @param [in, out]  FfaArgs        Ffa arguments
+
+**/
+STATIC
+VOID
+EFIAPI
+ArmFfaConsoleCall (
+  IN OUT ARM_FFA_ARGS  *FfaArgs
+  )
+{
+  if (PcdGetBool (PcdFfaLibConduitSmc)) {
+    ArmCallSmc ((ARM_SMC_ARGS *)FfaArgs);
+  } else {
+    ArmCallSvc ((ARM_SVC_ARGS *)FfaArgs);
+  }
+}
+
+/**
+  Log through FFA_CONSOLE_LOG supervisor call.
+
+  @param  Buffer           Point of data buffer which need to be written.
+  @param  NumberOfBytes    Number of output bytes which are cached in Buffer.
+
+  @retval 0                Write data failed.
+  @retval !0               Actual number of bytes written.
+**/
+STATIC
+UINTN
+EFIAPI
+ArmFfaConsoleLog (
+  IN CHAR8  *Buffer,
+  IN UINTN  NumberOfBytes
+  )
+{
+  RETURN_STATUS  Status;
+  UINTN          MaxCharsPerArmFfaConsoleCall;
+  UINTN          RemainBytes;
+  UINTN          WriteBytes;
+  ARM_FFA_ARGS   FfaArgs;
+
+  MaxCharsPerArmFfaConsoleCall = sizeof (UINT64) * MAX_REG64_COUNT;
+  RemainBytes                  = NumberOfBytes;
+
+  while (RemainBytes > 0) {
+    ZeroMem (&FfaArgs, sizeof (ARM_FFA_ARGS));
+    WriteBytes = (RemainBytes > MaxCharsPerArmFfaConsoleCall) ? MaxCharsPerArmFfaConsoleCall : RemainBytes;
+
+    /*
+     * StandaloneMm in Arm is supported 64-bit only so
+     * SPMC must be in the 64-bit execution state and
+     * according to FF-A specification Chapter 11, bullet 6
+     * SPMC must implement both SMC32 and SMC64 conventions of the ABI
+     * if it runs in the 64-bit execution state.
+     *
+     * Therefore, use FFA_CONSOLE_LOG_AARH64 only.
+     */
+    FfaArgs.Arg0 = ARM_FID_FFA_CONSOLE_LOG_AARCH64;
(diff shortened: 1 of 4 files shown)


MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugLib.c
```diff
@@ -0,0 +1,557 @@
...
+    case ARM_FFA_RET_DENIED:
+      return RETURN_ACCESS_DENIED;
+    case ARM_FFA_RET_ABORTED:
+      return RETURN_ABORTED;
+    case ARM_FFA_RET_NODATA:
+      return RETURN_NOT_FOUND;
+    case ARM_FFA_RET_NOT_READY:
+      return RETURN_NOT_READY;
+    case ARM_FFA_RET_RETRY:
+      return RETURN_TIMEOUT;
```
kuqin12, line 80: Is this the only difference between this new function and the `FfaStatusToEfiStatus` from ffa lib? Can we add this case handling to ffa lib instead? It seems inappropriate to eat this return code by returning `EFI_UNSUPPORTED` in the existing function.

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugLib.c
```diff
@@ -0,0 +1,557 @@
...
+
+  @param [in] FfaArgs            Ffa arguments
+
+  @retval RETURN_STATUS           return value correspond RETURN_STATUS to FfaStatus
+
+**/
+STATIC
+RETURN_STATUS
+EFIAPI
+FfaArgsToRetStatus (
```
kuqin12, line 97: This is also very similar to `FfaArgsToEfiStatus` just not handling interrupted case. But do we really need to carry multiple copies of this logic?

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugLib.c
```diff
@@ -0,0 +1,557 @@
...
+**/
+STATIC
+VOID
+EFIAPI
+ArmFfaConsoleCall (
+  IN OUT ARM_FFA_ARGS  *FfaArgs
+  )
+{
+  if (PcdGetBool (PcdFfaLibConduitSmc)) {
+    ArmCallSmc ((ARM_SMC_ARGS *)FfaArgs);
```
kuqin12, line 140: Should this be `ArmCallFfa`?

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugLib.c
```diff
@@ -0,0 +1,557 @@
...
+  )
+{
+  RETURN_STATUS  Status;
+  UINTN          Fid;
+  UINTN          MaxCharsPerArmFfaConsoleCall;
+  UINTN          Idx;
+  UINTN          RegIdx;
+  UINTN          RemainBytes;
+  UINTN          WriteBytes;
+  UINTN          Regs[MAX_REG64_COUNT];
```
kuqin12: Shall we remove the usage of this buffer? It seems to be a transient buffer that will be thrown away immediately. Otherwise, this is piling up more stack usage (FFA_ARGS, Regs, Buffer[MAX_DEBUG_MESSAGE_LENGTH]).

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugLib.c
LeviYeoReum: Yes. I'll add via another patch. At that time. I've missed.

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugLib.c
LeviYeoReum: The tragedy of this situation is for circular dependency.
If we use "ArmFfaLib"'s `FfaArgToEfiStatus()`, we need to link the ArmFfaLib for "ArmFfaConsoleDebugLib"
But "ArmFfaLib's" function also calls "ArmFfaConsoleDebugLig via DEBUG()"

I think the best way for this to make some helpers as "static inline" in "ArmFfaLib.h"
But, AFAIK, edk2 doesn't allow this...

So, For preventing circular dependency, I think we need to this..

Any thought?

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugLib.c
LeviYeoReum: As I said, If we use `ArmCallFfa()`, that means we need to link "ArmFfaLib" which makes a circular dependency with ArmFfaConsoleLib... unless we makes "ArmCallFfa()" as static inline.

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugLib.c
LeviYeoReum: I believe we couldn't remove this buffer TBH via some global variable or dynamic allocated buffer.

First, If it's changed to global variable, there is the time where DEBUG macro is called but DATA section isn't yet changed with RW (In this case we couldn't use) in StnadlaoneMm Entry point.

Second, If it's changed to dynamic allocation, there is also "DEBUG()" usage which makes a recursive.

And As you know, we couldn't just copy a Buffer in case of FFA_CONSOLE_LOG32 where
register size should be 4 bytes -- for example, 
5th char should be located at w1 but, in case of 64, it should be located at x0.

Am I missing something?

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugLib.c
kuqin12: How about we move the primitives of `ARM_FID_FFA_CONSOLE_LOG_AARCH64` and FFA_ARGS into ffa lib? And this debug lib just calls the new function?

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugLib.c
kuqin12: I meant to directly operate on `FfaArgs`. The `Regs` as of now is for preparation purpose, while it eventually will end up being copied to `FfaArgs`. Can we directly write to the final destination?

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugLib.c
kuqin12: My only concern is `mUse64Fid` being a new global, if we taking this approach. Do we really need to support both FIDs? Is there any case where the 64bit FID is not supported but 32bit FID is?

LeviYeoReum: Good idea. in that case though ArmFfaLib call's the "DEBUG()", it wouldn't trigger recursive call.
Let me try again with this way!  Thanks!

LeviYeoReum: > How about we move the primitives of ARM_FID_FFA_CONSOLE_LOG_AARCH64 and FFA_ARGS into ffa lib? And this debug lib just calls the new function?

Sorry. Although in the function level, There wouldn't be a recursive.
However, althogh we do this in ArmFfaLib. I still seems to reamin "circular library dependency" 
  - DebugLib needs ArmFfaLib
  - ArmFfaLib needs DebugLib.

Is it okay or Am I missing something? (I think the best way to change ArmFfaCall() and Status function into STATIC INLINE and locate it in header file, But IIRC, edk2 doesn't allow this..?)

LeviYeoReum: > My only concern is `mUse64Fid` being a new global, if we taking this approach. Do we really need to support both FIDs? Is there any case where the 64bit FID is not supported but 32bit FID is?

The maximum number of characters for CONSOLE_LOG32 is 24
but, maximum number of characters for CONSOLE_LOG64 is 128.
IOW, we could reduce quite amount of SMC by using 64 bit FID if we can use this.

Though this is debug library, If we can reduce the number of call of SMC, I think it would be good.

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugLib.c
LeviYeoReum: > I meant to directly operate on FfaArgs. The Regs as of now is for preparation purpose, while it eventually will end up being copied to FfaArgs. Can we directly write to the final destination?

Oh. You're right. I'll change it. Thanks ;)

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugLib.c
kuqin12: > Is it okay or Am I missing something? (I think the best way to change ArmFfaCall() and Status function into STATIC INLINE and locate it in header file, But IIRC, edk2 doesn't allow this..?)

It is okay for now. But I am concerned that will cause more problems down the road. I think the issue is that the current FFA lib is not a primitive library. It contains too much UEFI specific implementation.

I am thinking about refactoring the library so that the primitives can live its true "BASE" type separately (I know I mentioned that a while back..). But that is way beyond the scope of this PR. Let's table it for now.

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugLib.c
kuqin12: As mentioned in the other comment, the root of the problem is beyond this PR. Will close this comment and create a github issue to track the refactoring discussion.

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugLib.c
LeviYeoReum: @kuqin12 
https://github.com/tianocore/edk2/pull/12476/changes/fbab7a2166a8239cd3067f4f8f56106a7ce7332d

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugLib.c
LeviYeoReum: > It is okay for now. But I am concerned that will cause more problems down the road. I think the issue is that the current FFA lib is not a primitive library. It contains too much UEFI specific implementation.

I am thinking about refactoring the library so that the primitives can live its true "BASE" type separately (I know I mentioned that a while back..). But that is way beyond the scope of this PR. Let's table it for now.

Agree. might we can make a primitive for like ArmFfaRawLib which is primitive and ArmFfaLib uses this primitive library.

Let's handle later for this. Thanks!

kuqin12: Correct. I captured that in this github issue: https://github.com/tianocore/edk2/issues/12477. Along with an extra idea :) We can take it from there.

kuqin12 (approved):

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugStandaloneMmLib.c
```diff
@@ -0,0 +1,557 @@
+/** @file
```
samimujawar, line 1: Considering 
> NOTE:
>  FFA_CONSOLE_LOG* could be used by secure partition only.

Should the ArmFfaConsoleDebugLib be renamed to ArmFfaConsoleDebugStMmLib?

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugStandaloneMmLib.c
```diff
@@ -0,0 +1,557 @@
...
+  }
+
+  Regs        = &FfaArgs.Arg2;
+  RemainBytes = NumberOfBytes;
+
+  while (RemainBytes > 0) {
+    ZeroMem (Regs, sizeof (UINTN) * MAX_REG64_COUNT);
+    WriteBytes = (RemainBytes > MaxCharsPerArmFfaConsoleCall) ? MaxCharsPerArmFfaConsoleCall : RemainBytes;
+
+    for (Idx = 0; Idx < WriteBytes; Idx++) {
```
samimujawar: Would this code be simpler if you can make the type of Regs as UINT8*, and do CopyMem()?

LeviYeoReum: We couldn't. When the FFA_CONSOLE_LOG32 where register size is 4 bytes -- 
for example, 5th char should be located at w1 but, in case of 64, it should be located at x0.
IOW, if some char sequence is:
    0x11 0x22 0x33 0x44 0x55

in LOG64, 0x55 should be located at 4th index of Buffer (inner x0)
However, LOG32, 0x55 should be on the 8th index (which is in the w1).

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugStandaloneMmLib.c
LeviYeoReum: Okay. I'll change the file name.
But StMm to StnadaloneMm.

vbpandya: Nit: The title of the PR also ?

MdeModulePkg/Library/ArmFfaConsoleDebugLib/ArmFfaConsoleDebugStandaloneMmLib.c
samimujawar: Considering that StMM is AArch64 only, we should only support the 64 bit version of the FID.

LeviYeoReum:
Gentle ping @lgao4

lgao4 (approved):

lgao4 (approved):

LeviYeoReum:
This is some error for CI not from the code:

```
...
Failures
 - opencppcoverage - opencppcoverage not installed. The package was not found with the source(s) listed.
```

Would you requeue again please?

Thanks!

LeviYeoReum:
@mergifyio queue

PR #12482, opened 2026-04-18
SecurityPkg/AuthVariableLib: harden signature list filtering

20000419:
# Description
 
This change hardens `FilterSignatureList()` in `SecurityPkg/Library/AuthVariableLib/AuthService.c` against malformed `EFI_SIGNATURE_LIST` input by introducing a `GetSignatureListEntryCount()` helper that validates structure bounds before the filter walks list entries.
 
The old inline arithmetic lacked guards for `SignatureSize == 0` (division by zero), integer overflow on `sizeof(EFI_SIGNATURE_LIST) + SignatureHeaderSize`, and payload-to-entry alignment. The new helper centralizes these checks and returns `EFI_INVALID_PARAMETER` on any structural inconsistency. `FilterSignatureList()` now propagates failures back to `AuthServiceInternalUpdateVariableWithTimeStamp()`, which aborts the append operation on error.
 
The currently reachable append-update paths already validate payload format through `CheckSignatureListFormat()` before `FilterSignatureList()` is reached, so this PR is defense-in-depth hardening rather than a fix for a currently end-to-end exploitable path on `master`. No behavior change is intended for well-formed signature lists; for malformed input the patched code fails explicitly instead of proceeding with unchecked arithmetic.
 
- [ ] Breaking change?
- [x] Impacts security?
- [ ] Includes tests?
## How This Was Tested
 
Built the patched module locally in WSL with:
 
```text
build -p SecurityPkg/SecurityPkg.dsc -m SecurityPkg/Library/AuthVariableLib/AuthVariableLib.inf -a X64 -b DEBUG -t GCC5
```
 
Manually verified that appending a valid signature to `db` via authenticated variable update succeeds without regression. No host-based unit test is included in this PR; adding `FilterSignatureList` unit tests under `SecurityPkg/Test/` is tracked as a follow-up.

### DIFF
--- a/SecurityPkg/Library/AuthVariableLib/AuthService.c
+++ b/SecurityPkg/Library/AuthVariableLib/AuthService.c
@@ -274,6 +274,7 @@ AuthServiceInternalUpdateVariableWithTimeStamp (
   IN EFI_TIME  *TimeStamp
   )
 {
+  EFI_STATUS          Status;
   EFI_STATUS          FindStatus;
   VOID                *OrgData;
   UINTN               OrgDataSize;
@@ -299,12 +300,15 @@ AuthServiceInternalUpdateVariableWithTimeStamp (
       // For variables with formatted as EFI_SIGNATURE_LIST, the driver shall not perform an append of
       // EFI_SIGNATURE_DATA values that are already part of the existing variable value.
       //
-      FilterSignatureList (
-        OrgData,
-        OrgDataSize,
-        Data,
-        &DataSize
-        );
+      Status = FilterSignatureList (
+                 OrgData,
+                 OrgDataSize,
+                 Data,
+                 &DataSize
+                 );
+      if (EFI_ERROR (Status)) {
+        return Status;
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


20000419:
Gentle ping @jyao1 @mxu9 — this PR has been open for a week with all 
160 checks passing (1 neutral, 159 successful). It's a small 
defense-in-depth hardening in SecurityPkg/AuthVariableLib scoped to 
FilterSignatureList(); no behavior change for well-formed input. 
Would appreciate your review when you get a chance. Happy to split 
or adjust scope if needed.

jyao1:
The change looks good to me.

> Manually verified that appending a valid signature to db via authenticated variable update succeeds without regression. No host-based unit test is included in this PR; adding FilterSignatureList unit tests under SecurityPkg/Test/ is tracked as a follow-up.

Thanks for doing that test.
Could you please confirm, if you have enroll a real DB from MSFT and enroll a real DBX from UEFI CA, to ensure they are still working as expected?

Host based UnitTest can be added later. I have no concern.
And I prefer to enroll real DB/DBX as part of the integration test.

20000419:
> The change looks good to me.
> 
> > Manually verified that appending a valid signature to db via authenticated variable update succeeds without regression. No host-based unit test is included in this PR; adding FilterSignatureList unit tests under SecurityPkg/Test/ is tracked as a follow-up.
> 
> Thanks for doing that test. Could you please confirm, if you have enroll a real DB from MSFT and enroll a real DBX from UEFI CA, to ensure they are still working as expected?
> 
> Host based UnitTest can be added later. I have no concern. And I prefer to enroll real DB/DBX as part of the integration test.

Thanks, confirmed with a real-object OVMF integration test.

I rebuilt the current branch with Secure Boot enabled:

build -p OvmfPkg/OvmfPkgX64.dsc -a X64 -b DEBUG -t GCC5 -D SECURE_BOOT_ENABLE=TRUE -D SMM_REQUIRE=TRUE
build -p SecurityPkg/SecurityPkg.dsc -m SecurityPkg/Library/AuthVariableLib/AuthVariableLib.inf -a X64 -b DEBUG -t GCC5

Then I booted the built OVMF image in QEMU with a fresh variable store. The test first enrolled a test PK and the real Microsoft KEK CA 2011 from Microsoft secureboot_objects v1.6.4-signed, moving firmware into User Mode. The debug log showed:

SecureBoot variable updated according to PK change. Remeasure the value!

Then I applied signed amd64 DB/DBX payloads from Microsoft secureboot_objects v1.6.4-signed:

- DB: DBUpdate2024.bin, DBUpdate3P2023.bin, DBUpdateOROM2023.bin
- DBX: DBXUpdate.bin, DBXUpdate2024.bin

All updates returned EFI_SUCCESS:

SetVariable(KEK, SETUP\KEK.auth): Success
SetVariable(PK, SETUP\PK.auth): Success
SetVariable(db, DB\amd64\DBUpdate2024.bin): Success
SetVariable(db, DB\amd64\DBUpdate3P2023.bin): Success
SetVariable(db, DB\amd64\DBUpdateOROM2023.bin): Success
SetVariable(dbx, DBX\amd64\DBXUpdate.bin): Success
SetVariable(dbx, DBX\amd64\DBXUpdate2024.bin): Success
AuthVarRealEnrollTest: final Success

Final variable sizes:
KEK: 1560 bytes
PK:  855 bytes
db:  4493 bytes
dbx: 22431 bytes

So real Microsoft DB enrollment and real DBX enrollment still work as expected with this change. I did not add a host-based unit test in this PR.

jyao1 (approved):

PR #12483, opened 2026-04-19
RedfishPkg: fix memory leaks

ghbaccount:
# Description

This PR provides fixes for some findings in RedfishPkg.
In general it refines and fixes trivial memory leaks on an error path.

## How This Was Tested

Smoke test.

## Integration Instructions

N/A

### DIFF
--- a/RedfishPkg/Library/HiiUtilityLib/HiiUtilityInternal.c
+++ b/RedfishPkg/Library/HiiUtilityLib/HiiUtilityInternal.c
@@ -621,6 +621,7 @@ StorageToConfigResp (
                       (VOID **)&HiiConfigRouting
                       );
       if (EFI_ERROR (Status)) {
+        FreePool (TempConfigRequest);
         return Status;
       }
 
@@ -662,6 +663,7 @@ StorageToConfigResp (
       break;
   }
 
+  FreePool (TempConfigRequest);
   return Status;
 }
 
@@ -1246,6 +1248,10 @@ LoadFormSetStorage (
                   (VOID **)&HiiConfigRouting
                   );
   if (EFI_ERROR (Status)) {
+    if (ConfigRequest != NULL) {
+      FreePool (ConfigRequest);
+    }
+
     return;
   }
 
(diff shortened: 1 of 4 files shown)


szsam (approved):

RedfishPkg/Library/HiiUtilityLib/HiiUtilityInternal.c
```diff
@@ -1269,18 +1275,9 @@ LoadFormSetStorage (
...
-
-    return;
-  }
-
-  if (Storage->Type != EFI_HII_VARSTORE_NAME_VALUE) {
-    if (ConfigRequest != NULL) {
-      FreePool (ConfigRequest);
-    }
+  ASSERT (Storage->ConfigRequest != NULL);
+  if (ConfigRequest != NULL) {
```
nicklela, line 1279: Comparing to original code path, we don't want to free `ConfigRequest` when `Storage->Type` equals `EFI_HII_VARSTORE_NAME_VALUE`. And this is missing in new code path.

ghbaccount: When Storage->Type equals EFI_HII_VARSTORE_NAME_VALUE we don't even allocate ConfigRequest (string 1205).

The same unconditional free was in original code also (string 1273). For me original code looks like there are some remnants from old code were not removed for some reason, may be during review or merge.
But now I have looked at git history again and saw these were introduced by a single commit...

Anyway I decided that simple `if (ConfigRequest != NULL)` condition is enough for both branches (error and success)

Please clarify that on error path we free  ConfigRequest unconditionally but in case of success we do not do that.

ghbaccount: Hi, @nicklela !
What do you think about my comment above? May be I wasn't clear?
Do you still want to keep the original behavior?

nicklela: Hi @ghbaccount , sorry for my late response.

When `Storage->Type` equals `EFI_HII_VARSTORE_NAME_VALUE`, it goes to else case at line 1224: https://github.com/tianocore/edk2/blob/6a6ec8a228b6dd99f9a52a78c2a4c82be9b73ec8/RedfishPkg/Library/HiiUtilityLib/HiiUtilityInternal.c#L1224

And `ConfigRequest` is allocated at line 1226: 
https://github.com/tianocore/edk2/blob/6a6ec8a228b6dd99f9a52a78c2a4c82be9b73ec8/RedfishPkg/Library/HiiUtilityLib/HiiUtilityInternal.c#L1226

But I agree that we need to release `ConfigRequest` before we leave this function, if it is not NULL. I approved this change.

nicklela (approved):

PR #12488, opened 2026-04-20
OvmfPkg/IntelTdx: add toggle to disable firmware UI

katexochen:
In many cases, an interactive firmware UI isn't needed, and smaller interface surface desireable for CoCo guests. Add an opt-int toggle to disable firmware UI.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Applied patch downstream and built with` -DBUILD_FIRMWARE_UI=FALSE`.

## Integration Instructions

N/A

### DIFF
--- a/OvmfPkg/IntelTdx/IntelTdxX64.fdf
+++ b/OvmfPkg/IntelTdx/IntelTdxX64.fdf
@@ -216,8 +216,6 @@ INF  MdeModulePkg/Universal/Console/ConPlatformDxe/ConPlatformDxe.inf
 INF  MdeModulePkg/Universal/Console/ConSplitterDxe/ConSplitterDxe.inf
 INF  MdeModulePkg/Universal/Console/TerminalDxe/TerminalDxe.inf
 INF  MdeModulePkg/Universal/BdsDxe/BdsDxe.inf
-INF  MdeModulePkg/Application/UiApp/UiApp.inf
-INF  MdeModulePkg/Application/BootManagerMenuApp/BootManagerMenuApp.inf
 INF  OvmfPkg/QemuKernelLoaderFsDxe/QemuKernelLoaderFsDxe.inf
 INF  MdeModulePkg/Universal/DevicePathDxe/DevicePathDxe.inf
 INF  MdeModulePkg/Universal/Disk/DiskIoDxe/DiskIoDxe.inf
@@ -303,7 +301,10 @@ INF  MdeModulePkg/Universal/Acpi/BootGraphicsResourceTableDxe/BootGraphicsResour
 INF  MdeModulePkg/Universal/Disk/UdfDxe/UdfDxe.inf
 INF  OvmfPkg/VirtioFsDxe/VirtioFsDxe.inf
 
-INF MdeModulePkg/Logo/LogoDxe.inf
+INF  MdeModulePkg/Logo/LogoDxe.inf
+
+INF  MdeModulePkg/Application/UiApp/UiApp.inf
+INF  MdeModulePkg/Application/BootManagerMenuApp/BootManagerMenuApp.inf
 
 #
 # Usb Support


kraxel:
You can just move things over to the `FV.NCCFV` firmware volume, these modules will not be loaded in TDX mode.  Which surely makes sense for `UiApp`.  Not convinced  wrt `BootManagerMenuApp`, starting existing BootNNNN entries is pretty much the only thing which the boot menu can do, I don't see much attack surface there.

katexochen:
> Not convinced wrt `BootManagerMenuApp`, starting existing BootNNNN entries is pretty much the only thing which the boot menu can do, I don't see much attack surface there.

Even if its a simple interface, it still is an interface. That's why measurements of the MenuApp and its options are required. Which is, again, an added complexity in the measurement story, both for the verifier as well as [for measurement pre-calculation](https://github.com/edgelesssys/contrast/pull/2179/changes#diff-c91b5d8304b88d1008fadc4df2b185f5403cdd62ced9704b84e90389986be479).

kraxel:
> > Not convinced wrt `BootManagerMenuApp`, starting existing BootNNNN entries is pretty much the only thing which the boot menu can do, I don't see much attack surface there.
> 
> Even if its a simple interface, it still is an interface. That's why measurements of the MenuApp and its options are required. Which is, again, an added complexity in the measurement story

Why do you measure the boot entries?  The firmware already measures the initial state of the uefi variable store, that should cover the boot entries too.

Also: If you consider measuring the boot entries important nevertheless, why does removing BootManagerMenuApp change the picture?  BDS will still try the boot entries one after the other.

katexochen:
> Why do you measure the boot entries?

That's what TDVF is doing. We are not measuring this in addition, but we need to pre-calculate the measurements upfront (the code I linked). Not including the MenuApp in the build will also exclude the measurement events made for the boot menu.

kraxel:
> > Why do you measure the boot entries?
> 
> That's what TDVF is doing.

Then maybe stop doing that?  I don't see how this improves security, but it makes pre-calculation more complex, not only because of the menu app but also because the autogenerated entries for virtio-blk devices are not fixed.  They change if the pci address of the virtio-blk device changes.

That said I have no hard objections against excluding the menu app in tdx mode, so just move both to the `FV.NCCFV` firmware volume.

kraxel (approved):

katexochen:
> Then maybe stop doing that? I don't see how this improves security, but it makes pre-calculation more complex, not only because of the menu app but also because the autogenerated entries for virtio-blk devices are not fixed. They change if the pci address of the virtio-blk device changes.

Not sure I understand what you are suggesting here. For my setup with direct kernel boot, this change removes all the Boot#### from the tdeventlog. Are you saying that if I would add virtio-blk disks, BdsDxe would auto-create new boot entries that would be measured? And that should be disabled, too?

ardbiesheuvel (approved):

kraxel:
> > Then maybe stop doing that? I don't see how this improves security, but it makes pre-calculation more complex, not only because of the menu app but also because the autogenerated entries for virtio-blk devices are not fixed. They change if the pci address of the virtio-blk device changes.
> 
> Not sure I understand what you are suggesting here. For my setup with direct kernel boot, this change removes all the Boot#### from the tdeventlog. Are you saying that if I would add virtio-blk disks, BdsDxe would auto-create new boot entries that would be measured? And that should be disabled, too?

Yes, OVMF will create BootNNNN entries for disks, otherwise virtual machines would simply not be able to boot from disk.  I think happens after checking for a kernel to boot, so when using direct kernel boot you probably don't see that.

mikebeaton:
This has failed to push.

katexochen:
> This has failed to push.

Sorry. I enabled `Allow edits by maintainers`, please try again.

ardbiesheuvel:
@mergifyio requeue

PR #12489, opened 2026-04-20
MdeModulePkg/Universal/ExStatusCodeHandler: Add ExStatusCodeHandlerPei

xunfengd:
# Description

Add `ExStatusCodeHandlerPei` as an extension of the existing `StatusCodeHandlerPei`. The new handler keeps full compatibility with the legacy PEIM but adds three capabilities that the current PEI status code handler cannot provide:

- Thread-safe serial output. Serialize status-code prints across multiple threads/CPUs so messages from concurrent producers no longer interleave on the UART.
- Shadow-safe serial callbacks. Work correctly with `PcdMigrateTemporaryRamFirmwareVolumes` by fixing up serial-port callback pointers after PEIM shadowing, so the handler keeps reporting once PEI is migrated to permanent memory.
- Runtime gating of serial output. Drive registration of the serial worker from `PcdStatusCodeUseSerial`, which itself is derived from `GetDebugPrintErrorLevel ()`, so platforms can disable serial status code output without rebuilding.

No existing module is modified. `StatusCodeHandlerPei` continues to work as-is; platforms opt in to the new PEIM in their DSC/FDF.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Built `MdeModulePkg` with the new PEIM and its supporting libraries/PPIs/GUIDs using the standard edk2 CI build matrix. Ran the edk2 CI package-level checks (build, EccCheck, package tests) on `MdeModulePkg` to confirm no regressions. Boot-tested the new `ExStatusCodeHandlerPei` on an internal platform DSC that opts in to the PEIM, and verified:
- Serial status code output still appears during PEI.
- Output remains correct after PEIM shadowing with `PcdMigrateTemporaryRamFirmwareVolumes` enabled.
- Concurrent prints from multiple CPUs are serialized and do not interleave on the UART.
- Setting debug print error level to 0 disables serial status code output at runtime without rebuilding.

## Integration Instructions

Platforms that want the new capabilities should add `MdeModulePkg/Universal/ExStatusCodeHandler/Pei/ExStatusCodeHandlerPei.inf` (and its required library instances for `SemaphoreLib` and `ModuleAddrCalcLib`) to their DSC/FDF in place of, or in addition to, `StatusCodeHandlerPei`. Existing platforms that do not consume the new PEIM require no changes.

### DIFF
--- a/MdeModulePkg/Include/Guid/SerialPortConfig.h
+++ b/MdeModulePkg/Include/Guid/SerialPortConfig.h
@@ -1,0 +1,125 @@
+/** @file
+  Provides serial port configuration.
+
+  Copyright (c) 2025, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef _SERIAL_PORT_CONFIG_H
+#define _SERIAL_PORT_CONFIG_H
+
+#include <Uefi/UefiBaseType.h>
+
+#define SERIAL_PORT_CONFIG_MAX       3
+#define SERIAL_PORT_CONFIG_SIZE_MAX  100
+
+typedef union {
+  struct {
+    /**
+      Enables the eSPI interface for serial logging.
+    **/
+    UINT8    EspiUartEnable  : 1;
+
+    /**
+      Enables the local UART controller for serial logging.
+    **/
+    UINT8    LocalUartEnable : 1;
+
+    /**
+      Following values are supported for COM selection:
+      000 3F8h - 3FFh (COM 1)
+      001 2F8h - 2FFh (COM 2)
+      010 3E8h - 3EFh (COM 3)
+      011 2E8h - 2EFh (COM 4)
+    **/
+    UINT8    EspiComSelect  : 3;
+    UINT8    LocalComSelect : 3;
+  } Fields;
+
+  UINT8    Raw;
+} PLATFORM_SERIAL_PORT_CONFIG;
+
+typedef UINT8 SERIAL_PORT_DEVICE_CONFIG[SERIAL_PORT_CONFIG_SIZE_MAX];
+
+typedef
+RETURN_STATUS
+(*SERIAL_PORT_INITIALIZE) (
+  IN SERIAL_PORT_DEVICE_CONFIG  SerialPortDeviceConfig
+  );
+
+typedef
+UINTN
+(*SERIAL_PORT_WRITE) (
+  IN SERIAL_PORT_DEVICE_CONFIG  SerialPortDeviceConfig,
+  IN UINT8                      *Buffer,
+  IN UINTN                      NumberOfBytes
+  );
+
+typedef
+UINTN
+(*SERIAL_PORT_READ) (
+  IN SERIAL_PORT_DEVICE_CONFIG  SerialPortDeviceConfig,
+  IN UINT8                      *Buffer,
+  IN UINTN                      NumberOfBytes
+  );
+
+typedef
+BOOLEAN
+(*SERIAL_PORT_POLL) (
+  IN SERIAL_PORT_DEVICE_CONFIG  SerialPortDeviceConfig
+  );
+
+typedef struct {
+  /*
+    UART device enabled by the platform.
+  */
+  BOOLEAN                      EnabledByPlatform;
+
+  /*
+    UART device enabled by the platform and available on the current platform.
+  */
+  BOOLEAN                      Enabled;
+
+  /*
+    UART device initialization hook.
+  */
+  SERIAL_PORT_INITIALIZE       Initialize;
+
+  /*
+    UART device write hook.
+  */
+  SERIAL_PORT_WRITE            WriteBytes;
+
+  /*
+    UART device read hook.
+  */
+  SERIAL_PORT_READ             ReadBytes;
+
+  /*
+    UART device poll hook.
+  */
+  SERIAL_PORT_POLL             Poll;
+
+  /*
+    UART device configuration data. Convert to a device-specific type before
+    use.
+  */
+  SERIAL_PORT_DEVICE_CONFIG    DeviceConfig;
+} SERIAL_PORT_CONFIG_SINGLE;
+
+typedef struct {
+  /*
+    Array of all UART device configurations.
+  */
+  SERIAL_PORT_CONFIG_SINGLE    Configurations[SERIAL_PORT_CONFIG_MAX];
+
+  /*
+    Number of UART devices present.
+  */
+  UINT8                        Count;
+
+  BOOLEAN                      InitializeDevices;
+} SERIAL_PORT_CONFIG;
+
+#endif // _SERIAL_PORT_CONFIG_H
--- a/MdeModulePkg/Include/Library/ModuleAddrCalcLib.h
+++ b/MdeModulePkg/Include/Library/ModuleAddrCalcLib.h
@@ -1,0 +1,36 @@
+/** @file
+  Calculate Module Offset
+
+  Copyright (c) 2025, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+/**
+  Calculate the module entry point offset after PEIM shadowing.
+
+  @param[in, out] ModuleOffset    Absolute offset between the original PEIM
+                                  entry point and the shadowed entry point.
+  @param[in, out] ModulePositive  TRUE if the PEIM moved to a higher address.
+                                  FALSE if it moved to a lower address.
+
+  This helper supports PEIMs that may run twice: once from temporary storage and
+  once after shadowing into permanent memory.
+
+  This helper supports both 32-bit and 64-bit PEIMs.
+
+  @retval EFI_SUCCESS          The PEIM is running before shadowing. Ignore the
+                               returned offset and direction values.
+  @retval EFI_ALREADY_STARTED  The PEIM is running after shadowing. ModuleOffset
+                               and ModulePositive contain valid results.
+
+  @return Status  The return value is expected to be either EFI_SUCCESS or
+                  EFI_ALREADY_STARTED because a PEIM is shadowed at most once.
+**/
+EFI_STATUS
+ModuleOffsetCalculator (
+  IN     EFI_PEI_FILE_HANDLE  FileHandle,
+  IN     EFI_GUID             *CallerGuid,
+  IN OUT UINTN                *ModuleOffset,
+  IN OUT BOOLEAN              *ModulePositive
+  );
--- a/MdeModulePkg/Include/Library/SemaphoreLib.h
+++ b/MdeModulePkg/Include/Library/SemaphoreLib.h
@@ -1,0 +1,373 @@
+/** @file
+  The Semaphore Library API provides the necessary functions to acquire/release
+  the global or a socket's semaphore.
+  This API is designed to allow a calling agent to acquire a global (the SBSP)
+  semaphore or a socket's semaphore.  It also provides functionality to release
+  the semaphore and check if ownership has been obtained.  If a semaphore is
+  desired, an agent should first attempt to acquire it, then check if it has
+  ownership.  If ownership has not been obtained, the agent must wait until
+  ownership has been obtained before proceeding.  Once the desired task is complete
+  the semaphore must be released.  Semaphores should be used when ensuring
+  exclusive access to resources among CPU sockets is necessary.
+
+  Copyright (c) 2025, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef _SEMAPHORE_LIB_H_
+#define _SEMAPHORE_LIB_H_
+
+#include <Uefi.h>
+
+///
+/// Used to identify which system semaphore is being accessed. There are
+/// currently two System Semaphores available for use (0 and 1).
+///
+typedef enum {
+  SystemSemaphore0,     ///< Semaphore 0 - Used for SPD/SMBus access
+  SystemSemaphore1,     ///< Semaphore 1 - Used for Debug print
+  SystemSemaphore2,
+  SystemSemaphore3,
+  SystemSemaphoreMax
+} SYSTEM_SEMAPHORE_NUMBER;
+
+///
+/// Used to identify which local semaphore is being accessed. There are
+/// currently two Local Semaphores available for use (0 and 1).
+///
+typedef enum {
+  LocalSemaphore0,
+  LocalSemaphore1,
+  LocalSemaphore2,
+  LocalSemaphore3,
+  LocalSemaphoreMax
+} LOCAL_SEMAPHORE_NUMBER;
+
+#define DO_NOT_ADD_TO_QUEUE  FALSE
+#define ADD_TO_QUEUE         TRUE
+
+/**
+  Acquire a global (BSP) semaphore for the calling socket.
+
+  Used for ensuring exclusive access to resources among CPU sockets.
+
+  IMPORTANT:
+    The functions must be called in the sequence below:
+      ......
+      Owned = AcquireGlobalSemaphore (SystemSemaphore0, ADD_TO_QUEUE, &QNum);
+      while (!Owned) {
+        Owned = IsGlobalSemaphoreOwned (SystemSemaphore0, QNum));
+      }
+      DoSomething ();
+      ReleaseGlobalSemaphore (SystemSemaphore0);
+      ......
+
+  @param[in]  SemaphoreNumber   - SYSTEMSEMAPHORE register number (0 or 1)
+  @param[in]  AddToQueue        - specifies whether to add the requesting
+                                  socket to the queue (TRUE or FALSE)
+  @param[out] QueueNumber       - assigned place in line of semaphore request,
+                                  if adding to queue
+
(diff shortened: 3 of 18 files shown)


MdeModulePkg/Include/Guid/SerialPortConfig.h
```diff
@@ -0,0 +1,125 @@
+/** @file
+  Provides serial port configuration.
```
apop5, line 2: Is this necessary for adding ExStatusCodeHandler?

ExStatusCodeHandler is consuming a SerialPortLib instance, and making call directly to it. Why is this introducing a serial port config as a HOB?

MdeModulePkg/Universal/ExStatusCodeHandler/Pei/ExSerialStatusCodeWorker.c
```diff
@@ -0,0 +1,202 @@
...
+                  Value,
+                  Instance
+                  );
+  }
+
+  //
+  // No EX info, just call SerialPort Lib function to do print and exit.
+  //
+  if (ExDebugInfo == NULL) {
+    SerialPortWrite ((UINT8 *)Buffer, CharCount);
```
apop5, line 172: SemaphoreLib is being introduced, but there is no use of it here. Can you explain the expected usage?

If the system is attempting to prevent concurrent access to the serial ports, then it seems like it would be consumed prior to calling SerialPortWrite

xunfengd:
> ## Pull Request Formatting Issues
> ⚠️ Remove the following template lines from your PR description: `<_Include a description of the change and why this change was made._>` `<_For each item, place an "x" in between `[`and`]`if true. Example:`[x]` (you can also check items in GitHub UI)_>` `<_Create the PR as a Draft PR if it is only created to run CI checks._>` `<_Delete lines in \<\> tags before creating the PR._>` `<_Describe the test(s) that were run to verify the changes._>` `<_Describe how these changes should be integrated. Use N/A if nothing is required._>`
> 
> Address these issues and the validation will automatically re-run when you update your pull request.

All  template lines have been removed. Thanks!

MdeModulePkg/Include/Library/ModuleAddrCalcLib.h
```diff
@@ -0,0 +1,41 @@
+/** @file
+  Calculate Module Offset
+
+  Copyright (c) 2025, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef _MODULE_ADDR_CALC_LIB_H_
```
apop5: There is exactly 1 consumer of this library class. Should this library be deleted, and the functionality rolled into the ExStatusCodeHandler directly?

MdeModulePkg/Library/SemaphoreLibNull/SemaphoreLibNull.inf
```diff
@@ -0,0 +1,29 @@
+## @file
+# Null instance of the SemaphoreLib.
```
apop5, line 2: The library class is named SemaphoreLib, the null instance should be named SemaphoreLibNull. The Hardware in the name isn't needed.

MdeModulePkg/MdeModulePkg.dsc
```diff
@@ -109,6 +109,7 @@
   VariableFlashInfoLib|MdeModulePkg/Library/BaseVariableFlashInfoLib/BaseVariableFlashInfoLib.inf
   IpmiCommandLib|MdeModulePkg/Library/BaseIpmiCommandLibNull/BaseIpmiCommandLibNull.inf
   SpiHcPlatformLib|MdeModulePkg/Library/BaseSpiHcPlatformLibNull/BaseSpiHcPlatformLibNull.inf
+  HardwareSemaphoreLib|MdeModulePkg/Library/HardwareSemaphoreLibNull/HardwareSemaphoreLibNull.inf
```
apop5: This is wrong, it should be SemaphoreLIb because that is the library class.

MdeModulePkg/Include/Library/SemaphoreLib.h
```diff
@@ -0,0 +1,373 @@
...
+  desired, an agent should first attempt to acquire it, then check if it has
+  ownership.  If ownership has not been obtained, the agent must wait until
+  ownership has been obtained before proceeding.  Once the desired task is complete
+  the semaphore must be released.  Semaphores should be used when ensuring
+  exclusive access to resources among CPU sockets is necessary.
+
+  Copyright (c) 2025, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
```
apop5, line 16: edk2 has switched to using `#pragma once` instead of include header guard.

MdeModulePkg/Include/Ppi/ExReportStatusCodeHandler.h
```diff
@@ -0,0 +1,39 @@
+/** @file
+  Define the EFI_PEI_EX_RSC_HANDLER_PPI used to register
+  ExSerialStatusCodeReportWorker for ReportStatusCode() notifications.
+
+  Copyright (c) 2025, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef __EX_REPORT_STATUS_CODE_HANDLER_PPI_H__
```
apop5, line 10: edk2 has switched to using `#pragma once` instead of include header guard.

MdeModulePkg/Universal/ExStatusCodeHandler/Pei/ExStatusCodeHandlerPei.h
```diff
@@ -0,0 +1,140 @@
+/** @file
+  Internal include file for Status Code Handler PEIM.
+
+  Copyright (c) 2006 - 2012, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef __EX_STATUS_CODE_HANDLER_PEI_H__
```
apop5: edk2 has switched to using `#pragma once` instead of include header guard.

MdeModulePkg/Universal/ExStatusCodeHandler/Pei/ExStatusCodeHandlerPei.inf
```diff
@@ -0,0 +1,71 @@
...
+  SerialPortLib
+  ReportStatusCodeLib
+  PrintLib
+  DebugLib
+  DebugPrintErrorLevelLib
+  BaseMemoryLib
+  ModuleAddrCalcLib
+
+[Guids]
+  ## SOMETIMES_PRODUCES   ## HOB
```
apop5: Drop these empty comment lines

MdeModulePkg/Library/ModuleAddrCalcLibNull/ModuleAddrCalcLibNull.inf
```diff
@@ -0,0 +1,32 @@
...
+#
+#  Copyright (c) 2025, Intel Corporation. All rights reserved.<BR>
+#
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+[Defines]
+  INF_VERSION     = 0x00010017
+  BASE_NAME       = ModuleAddrCalcLibNull
```
apop5: This isn't a NULL library. If this library can be removed, and the functionality directly into ExStatusCodeHanlder, then it becomes moot. Otherwise, the basename, files names, etc should have the NULL removed

MdeModulePkg/Universal/ExStatusCodeHandler/Pei/ExStatusCodeHandlerPei.c
```diff
@@ -0,0 +1,167 @@
...
+EFI_PEI_PPI_DESCRIPTOR  mExStatusCodeHandlerPpiList[] = {
+  {
+    EFI_PEI_PPI_DESCRIPTOR_PPI | EFI_PEI_PPI_DESCRIPTOR_TERMINATE_LIST,
+    &gEfiPeiExStatusCodeHandlerPpiGuid,
+    &mExStatusCodeHandlerPpi
+  }
+};
+
+// {560A1DA9-CADB-421D-ADAD-A583E2143601}
+EFI_GUID  gExStatusCodeHandlerModuleMigrationGuid = {
```
apop5: This is being used to identify the GuidHob....but it is not used anywhere else.

Why not use Autogen.c's gEfiCallerIdGuid, which is generated to be the FILE_GUID from the INF ?

MdeModulePkg/Universal/ExStatusCodeHandler/Pei/ExStatusCodeHandlerPei.c
```diff
@@ -0,0 +1,167 @@
...
+    Status = SerialPortInitialize ();
+    ASSERT_EFI_ERROR (Status);
+    Status = RscHandlerPpi->Register (ExSerialStatusCodeReportWorker);
+    ASSERT_EFI_ERROR (Status);
+  }
+
+  //
+  // Install Report Status Code Handler PPI
+  //
+  Status = PeiServicesInstallPpi (mExStatusCodeHandlerPpiList);
```
apop5, line 155: Can you explain why this is necessary? This is a PPI, that will need to be located and called from another PEIM to register an additional statuscode handler into the Rsc, but why? Could this just install the additional RSC handler without the need for another module to force its install?

MdeModulePkg/Include/Guid/SerialPortConfig.h
xunfengd: Hi @apop5,
On the server platform, support for various serial controllers (eSPI, BMC, etc.) is required. During the firmware boot process, serial port configurations are first retrieved from Setup/NVS and subsequently stored in the HOB. When the serial ports are initialized, the system prioritizes loading configurations from the HOB; this approach enhances the flexibility of the serial port initialization process.

MdeModulePkg/Universal/ExStatusCodeHandler/Pei/ExSerialStatusCodeWorker.c
xunfengd: Hi @apop5 ，
Yeah, nice catch! if "ExDebugInfo == NULL" , it means that it's a concurrent access so we could directly call the serial output without PrintAcquire. 
eg for the  concurrent access,
   if (ExDebugInfo->PrintSyncAcquire != NULL) {
     ExDebugInfo->PrintSyncAcquire ();
   }

MdeModulePkg/Include/Library/ModuleAddrCalcLib.h
xunfengd: Hi @apop5 ，
ModuleAddrCalcLib is a general-purpose foundational library designed for calculating module offsets when Module Migration. Currently, we have numerous drivers that load during the pre-memory phase and require migration to the post-memory phase; this library enables us to precisely locate the specific function calls involved in post-mem phase.

MdeModulePkg/Library/SemaphoreLibNull/SemaphoreLibNull.inf
xunfengd: Hi @apop5 ,
Thanks for your reminder.  will rename it later.

MdeModulePkg/MdeModulePkg.dsc
xunfengd: Hi @apop5 ,
Thanks for your reminder. will rename it later.

MdeModulePkg/Universal/ExStatusCodeHandler/Pei/ExStatusCodeHandlerPei.h
xunfengd: Hi @apop5 ,
Thanks for your reminder. will remove it later.

MdeModulePkg/Universal/ExStatusCodeHandler/Pei/ExStatusCodeHandlerPei.inf
xunfengd: Hi @apop5 ,
Thanks for your reminder. will remove it later.

MdeModulePkg/Include/Guid/SerialPortConfig.h
apop5: If I am understanding this correctly, you are attempting to have multiple serial output devices. One for BMC, one for eSPI, and others, and all of these are expecting to take every DEBUG message passed through ReportStatusCode, and be able to output them to their respective associated device? 

if that is the case, this seems like a very circular route to get that functionality. It would be easier for each associated output device to have their own standard RSC handler that gets registered, no additional Hobs taking up CAR, no ExStatusCodeHandler PEIM needed.

Am I misunderstanding the goal here?

MdeModulePkg/Library/ModuleAddrCalcLibNull/ModuleAddrCalcLibNull.inf
xunfengd: Hi @apop5 ,
Thanks for your reminder. will refactor it later.

MdeModulePkg/Universal/ExStatusCodeHandler/Pei/ExStatusCodeHandlerPei.c
xunfengd: @apop5 
Thanks for your reminder. will use gEfiCallerIdGuid later instead.

MdeModulePkg/Universal/ExStatusCodeHandler/Pei/ExStatusCodeHandlerPei.c
xunfengd: ExStatusCodeHandlerPei already self-registers its own serial worker on entry — the PPI is purely an extension point for downstream handlers. If you'd prefer, I can drop the PPI from this PR and add it later when a second consumer is upstreamed.

MdeModulePkg/Include/Guid/SerialPortConfig.h
xunfengd: Hi @apop5, 
That is not the case! During a single boot cycle, we do not intend to output UART data to both the BMC and eSPI simultaneously. The selection of the UART interface depends on NVS/setup configurations—or even board jumpers—as defined by this specific configuration option.

The SerialPortConfig HOB is consumed by the SerialPortLib instance, not by ExStatusCodeHandlerPei. Server platforms support multiple physical serial controllers (eSPI/SIO/BMC/MMIO UART) whose base address, stride, baud rate, etc. are read from Setup/NVS at early PEI and need to be passed to the serial library. Producing one HOB lets a single SerialPortLib instance retarget at runtime instead of building per-controller libraries.

That said, the HOB definition is independent of this PEIM. If you'd prefer, I can split the SerialPortConfig.h GUID into a separate PR or drop it and rebase ExStatusCodeHandlerPei to not reference it directly — please let me know which you prefer.

PS:
The motivation isn't routing one print to several devices; it's the three capabilities the existing StatusCodeHandlerPei lacks: (a) cross-CPU serialization on the UART, (b) callback fix-up after PEIM shadowing under PcdMigrateTemporaryRamFirmwareVolumes, and (c) runtime gating via PcdStatusCodeUseSerial. These are orthogonal to per-device handlers — any number of additional RSC handlers can still be registered. We deliberately did not modify StatusCodeHandlerPei so platforms that don't need these features are unaffected.

The StatusCodeHandler is highly useful in practical bootloader deployments, and it typically features implementations across the PEI, DXE, MM, and RT phases. By leveraging mechanisms such as PPIs and Protocols, a platform can easily register various platform-specific feature handlers—such as the TraceHubHandler, FatalErrorHandler, HECIHandler, and AMTHandler. Whenever a module invokes the `ReportStatusCode` function, these registered handlers are called in a sequential loop, thereby significantly enhancing efficiency (as there is no need to implement separate, standalone libraries to replicate these specific functionalities).

MdeModulePkg/Universal/ExStatusCodeHandler/Pei/ExStatusCodeHandlerPei.h
xunfengd: Done

MdeModulePkg/Universal/ExStatusCodeHandler/Pei/ExStatusCodeHandlerPei.inf
xunfengd: Done

MdeModulePkg/Library/ModuleAddrCalcLibNull/ModuleAddrCalcLibNull.inf
xunfengd: Remove the null done

MdeModulePkg/Universal/ExStatusCodeHandler/Pei/ExStatusCodeHandlerPei.c
xunfengd: remove the gExStatusCodeHandlerModuleMigrationGuid  done and try to use gEfiCallerIdGuid

panyongg:
The PR [12489](https://github.com/tianocore/edk2/pull/12489) is sued to replace https://github.com/tianocore/edk2/pull/11870, as 11870 will be closed.

PR #12491, opened 2026-04-20
MdeModulePkg: Don't Allow Guard Pages to Cross Bin Boundaries

os-d:
# Description

Currently, the DXE page allocator does not ensure that guard page allocations stay within the bin that it is attempting to allocate within. As a result, S4 resume is jeopardized by bins expanding due to guard pages, either into other bins or out of bins. This is caught by a new assert in CoreGetMemoryMap() to ensure the bins are correct.

This fixes this by changing the internal heap guard API to return the adjusted size and start address of a proposed allocation. The page allocator then can ensure that the adjusted allocation still fits within the bin it is attempting to allocate within; if not, it will search for another descriptor.

This is attempting to fix the same issue as https://github.com/tianocore/edk2/pull/12384. After offline discussion, it was determined that a different approach than that PR was desired to create a smaller change.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Repro'd a similar issue on OVMF as described here: https://github.com/tianocore/edk2/pull/12384. After this change, OVMF with heap guard enabled boots to shell instead of hitting the assert.

## Integration Instructions

N/A.

### DIFF
--- a/MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
+++ b/MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
@@ -814,21 +814,37 @@ UnsetGuardForMemory (
   memory blocks, and try to use it as the Guard page of the memory to be
   allocated.
 
-  @param[in]  Start           Start address of free memory block.
-  @param[in]  Size            Size of free memory block.
-  @param[in]  SizeRequested   Size of memory to allocate.
+  A guard page is not allowed to be placed at page 0 because it may be used for
+  NULL pointer detection or have special meaning if it is left mapped. This API
+  will reject the memory block if the only available place for the head guard is
+  page 0.
+
+  @param[in,out]  Start           Start address of the free memory block. On successful return, this will contain the
+                                  adjusted start address of the proposed allocation.
+  @param[in]      Size            Size of free memory block.
+  @param[in,out]  SizeRequested   Size of memory to allocate. On successful return, this will contain the adjusted size
+                                  of memory to allocate.
 
   @return The end address of memory block found.
   @return 0 if no enough space for the required size of memory and its Guard.
 **/
 UINT64
 AdjustMemoryS (
-  IN UINT64  Start,
-  IN UINT64  Size,
-  IN UINT64  SizeRequested
+  IN OUT UINT64  *Start,
+  IN UINT64      Size,
+  IN OUT UINT64  *SizeRequested
   )
 {
-  UINT64  Target;
+  UINT64   Target;
+  BOOLEAN  HasHeadGuard;
+  UINT64   OriginalSizeRequested;
+  UINT64   OriginalStart;
+
+  HasHeadGuard = TRUE;
+
+  if ((Start == NULL) || (SizeRequested == NULL) || (*SizeRequested == 0) || (Size < *SizeRequested)) {
+    return 0;
+  }
 
   //
   // UEFI spec requires that allocated pool must be 8-byte aligned. If it's
@@ -836,36 +852,53 @@ AdjustMemoryS (
   // make sure alignment of the returned pool address.
   //
   if ((PcdGet8 (PcdHeapGuardPropertyMask) & BIT7) == 0) {
-    SizeRequested = ALIGN_VALUE (SizeRequested, 8);
+    *SizeRequested = ALIGN_VALUE (*SizeRequested, 8);
   }
 
-  Target = Start + Size - SizeRequested;
-  ASSERT (Target >= Start);
-  if (Target == 0) {
+  // Take into account the alignment requirement
+  OriginalSizeRequested = *SizeRequested;
+  OriginalStart         = *Start;
+
+  Target = OriginalStart + Size - OriginalSizeRequested;
+  if ((Target <= EFI_PAGE_SIZE) || (Target < OriginalStart)) {
+    // If Target is at page 1 or below, we don't have a shared guard page and can't add a new one,
+    // so reject this memory block. Page 0 is reserved for NULL pointer detection.
+    // Our alignment requirement may have caused Target to be outside of the free memory block,
+    // so reject the block in that case as well.
     return 0;
   }
 
-  if (!IsGuardPage (Start + Size)) {
+  if (!IsGuardPage (OriginalStart + Size)) {
     // No Guard at tail to share. One more page is needed.
-    Target -= EFI_PAGES_TO_SIZE (1);
+    Target         -= EFI_PAGES_TO_SIZE (1);
+    *SizeRequested += EFI_PAGES_TO_SIZE (1);
   }
 
   // Out of range?
-  if (Target < Start) {
+  if ((Target < OriginalStart) || (Target <= EFI_PAGE_SIZE)) {
     return 0;
   }
 
+  // We are done adjusting Target at this point, so set the start of the proposed allocation
+  // to Target
+  *Start = Target;
+
+  // Check if we have a head guard to share. If not, we need one more page for the head Guard.
+  if (!IsGuardPage (Target - EFI_PAGES_TO_SIZE (1))) {
+    HasHeadGuard    = FALSE;
+    *Start         -= EFI_PAGES_TO_SIZE (1);
+    *SizeRequested += EFI_PAGES_TO_SIZE (1);
+  }
+
   // At the edge?
-  if (Target == Start) {
-    if (!IsGuardPage (Target - EFI_PAGES_TO_SIZE (1))) {
-      // No enough space for a new head Guard if no Guard at head to share.
-      return 0;
-    }
+  if ((Target == OriginalStart) && !HasHeadGuard) {
+    // Not enough space for a new head guard page and there isn't one to share
+    return 0;
   }
 
   // OK, we have enough pages for memory and its Guards. Return the End of the
   // free space.
-  return Target + SizeRequested - 1;
+  return Target + OriginalSizeRequested - 1;
 }
 
 /**
--- a/MdeModulePkg/Core/Dxe/Mem/HeapGuard.h
+++ b/MdeModulePkg/Core/Dxe/Mem/HeapGuard.h
@@ -275,18 +275,25 @@ AdjustMemoryF (
   memory blocks, and try to use it as the Guard page of the memory to be
   allocated.
 
-  @param[in]  Start           Start address of free memory block.
-  @param[in]  Size            Size of free memory block.
-  @param[in]  SizeRequested   Size of memory to allocate.
+  A guard page is not allowed to be placed at page 0 because it may be used for
+  NULL pointer detection or have special meaning if it is left mapped. This API
+  will reject the memory block if the only available place for the head guard is
+  page 0.
+
+  @param[in,out]  Start           Start address of the free memory block. On successful return, this will contain the
+                                  adjusted start address of the proposed allocation.
+  @param[in]      Size            Size of free memory block.
+  @param[in,out]  SizeRequested   Size of memory to allocate. On successful return, this will contain the adjusted size
+                                  of memory to allocate.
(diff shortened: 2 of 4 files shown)


os-d:
@AmyChan-Intel can you please confirm that this PR resolves the issue you reported?

os-d:
@mdkinney this change does not have any intersection with my PEI memory bins PR, all the logic changes are strictly in the DXE heap guard and page allocator code, not any of the bin specific code that will be shared between PEI and DXE.

AmyChan-Intel:
> @AmyChan-Intel can you please confirm that this PR resolves the issue you reported?

Yes, this pr does resolve the issue we meet.

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
```diff
@@ -814,58 +814,80 @@ UnsetGuardForMemory (
   memory blocks, and try to use it as the Guard page of the memory to be
   allocated.
 
-  @param[in]  Start           Start address of free memory block.
-  @param[in]  Size            Size of free memory block.
-  @param[in]  SizeRequested   Size of memory to allocate.
+  @param[in,out]  Start           Start address of the free memory block. On successful return, this will contain the
+                                  adjusted start address of the proposed allocation.
+  @param[in]      Size            Size of free memory block. On successful return, this will contain the adjusted size
```
mdkinney: Size is [in] only.  Comment seems wrong that the adjusted size is returned.  Formal parameter below is pass by value, not reference.

MdeModulePkg/Core/Dxe/Mem/HeapGuard.h
```diff
@@ -275,18 +275,21 @@ AdjustMemoryF (
   memory blocks, and try to use it as the Guard page of the memory to be
   allocated.
 
-  @param[in]  Start           Start address of free memory block.
-  @param[in]  Size            Size of free memory block.
-  @param[in]  SizeRequested   Size of memory to allocate.
+  @param[in,out]  Start           Start address of the free memory block. On successful return, this will contain the
+                                  adjusted start address of the proposed allocation.
+  @param[in]      Size            Size of free memory block. On successful return, this will contain the adjusted size
```
mdkinney: Same issue here.  [in] only, but description says it will return adjusted size.

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
os-d: Thanks, will fix here and in the header. Originally, I had Size as the param being adjusted, hence the comment there, but realized it needed to be SizeRequested and missed fixing the comment.

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
```diff
@@ -814,58 +814,80 @@ UnsetGuardForMemory (
...
+  if (!IsGuardPage (OriginalStart + Size)) {
     // No Guard at tail to share. One more page is needed.
-    Target -= EFI_PAGES_TO_SIZE (1);
+    Target         -= EFI_PAGES_TO_SIZE (1);
+    *SizeRequested += EFI_PAGES_TO_SIZE (1);
   }
 
   // Out of range?
-  if (Target < Start) {
+  if (Target < OriginalStart) {
```
mdkinney: ASSERT() is added above.  Should ASSERT() be here as well if the one page adjustment triggers this condition?

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
```diff
@@ -814,58 +814,80 @@ UnsetGuardForMemory (
...
     return 0;
   }
 
+  // Check if we have a head guard to share. If not, we need one more page for the head Guard.
+  if (!IsGuardPage (Target - EFI_PAGES_TO_SIZE (1))) {
```
mdkinney, line 887: Do you need to check if Target-EFI_PAGES_TO_SIZE(1) underflows before this check with an ASSERT() like above?

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
os-d: The ASSERT() above was already in the code, I just changed it to accommodate `OriginalStart` instead of `Start`. That assert is making sure the input parameters are valid (with the 8 byte alignment), i.e. that SizeRequested is <= Size (which I added a new check for, though doesn't take the 8 byte alignment into account, as this assert does). So, this assert could probably be dropped.

In this case (line 871), we can't assert because this is an expected case. The page allocator can give this function a free memory space that is exactly the same size as the size it is requesting. This is valid if there are shared guard pages on either side of the free space. So, we expect that this case may be rejected by heap guard and the page allocator will find the next free space that might be valid.

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
os-d: Do you mean underflows the integer (i.e. Target < 0x1000) or underflows the free memory range?

If the former, I think that's fair here and above where we decrement Target (as well as decrementing *Start). We may be in very low addresses (though if in page 0 whatever got this will crash).

For the latter, this does need a fix. *Start should be set to Target before this point so we correctly return the start of the proposed allocation area (including guards) and not adjust the start of the free memory space. That would incorrectly reject valid memory regions. I will fix that and give another test.

Target doesn't get decremented here (heap guard will magically later know it can take the page before the allocation when the actual guard pages are created) and so the check on line 882 catches the case where we would underflow the free memory range, because Target would == OriginalStart.

os-d: I resolved this by setting `*Start = Target` after we are done adjusting Target and before this point. I also amended the check above on line 858 to check if Target is <= EFI_PAGE_SIZE and reject if so, which would catch the edge case of integer underflow.

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
os-d: I ended up dropping the ASSERT() above, since I think that could hit on a legitimate case (the allocator offers up a block for a pool that is the right size but not with the 8 byte alignment requirement for the size and so heap guard rejects that region and the allocator tries another region). Before, the assert was just being used to attempt to reject that case without actually doing a check. I now have a check that will reject that case.

os-d:
@mdkinney I have addressed your feedback and re-tested the repro on OvmfPkg. This continues to fix the issue but now doesn't reject some valid memory blocks. Please re-review when you have a chance.

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
```diff
@@ -814,58 +814,86 @@ UnsetGuardForMemory (
...
 
-  Target = Start + Size - SizeRequested;
-  ASSERT (Target >= Start);
-  if (Target == 0) {
+  // Take into account the alignment requirement
+  OriginalSizeRequested = *SizeRequested;
+  OriginalStart         = *Start;
+
+  Target = OriginalStart + Size - OriginalSizeRequested;
+  if ((Target <= EFI_PAGE_SIZE) || (Target < OriginalStart)) {
```
mdkinney, line 863: Should this is < EFI_PAGE_SIZE to allow Target == 0x1000?

Also, this is related to the other feature for Null page protection - PcdNullPointerDetectionPropertyMask.  This used to check for 0, and this now checks for EFI_PAGE_SIZE.  Do we need to use PcdNullPointerDetectionPropertyMask for EFI_PAGE_SIZE check and 0 if PcdNullPointerDetectionPropertyMask is disabled?

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
mdkinney: Does Target need a 0 or EFI_PAGE_SIZE check here as well since there is a path where Target can be reduced by one page?

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
os-d: I thought about having this check depend on PcdNullPointerDetectionPropertyMask, but I decided it was simpler to always disallow guard pages at page 0 because either a platform has that PCD set and so we should avoid it or it has page 0 mapped and that can have special meaning (i.e. grub will attempt to read structures from page 0 without checking to see if it is mapped and crash the system if not).

If a system was so memory constrained that the only valid allocation was to include a guard page at page 0, then on the next allocation it would need to promote all the guard pages to regular memory anyway and allocate from them.

If you'd prefer I change this check to check against the PCD, but I thought it was safer to always disallow guard pages on page 0. Let me know what you think.

I have `Target <= 0x1000` here because we need a head guard. We know there won't be a shared head guard at page 0 (based on my rationale above to not allow that) and so we would need to allocate one. But, I am disallowing allocating a guard page at page 0, so we need to fail out.

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
os-d: Yeah, that is fair. I have added a check for <= EFI_PAGE_SIZE for the same reason, we need a head guard and we know there won't be a shared guard and we are disallowing allocating one at page 0.

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
mdkinney: I am ok with this logic.  Do we need to update description of Heap Guard to state that it does not allow a guard page to be placed at page 0?

os-d: I took a look but I didn't find any central description of heap guard, so I updated the function's header comment. Let me know if there is another spot you are thinking of.

mdkinney: I was thinking the MdeModulePkg.dec file where the PCD that controls enable/disable of Heap Guard:

```
  ## This mask is to control Heap Guard behavior.
  #
  # Note:
  #   a) Heap Guard is for debug purpose and should not be enabled in product
  #      BIOS.
  #   b) Due to the limit of pool memory implementation and the alignment
  #      requirement of UEFI spec, BIT7 is a try-best setting which cannot
  #      guarantee that the returned pool is exactly adjacent to head guard
  #      page or tail guard page.
  #   c) UEFI freed-memory guard and UEFI pool/page guard cannot be enabled
  #      at the same time.
  #
  #   BIT0 - Enable UEFI page guard.<BR>
  #   BIT1 - Enable UEFI pool guard.<BR>
  #   BIT2 - Enable SMM page guard.<BR>
  #   BIT3 - Enable SMM pool guard.<BR>
  #   BIT4 - Enable UEFI freed-memory guard (Use-After-Free memory detection).<BR>
  #   BIT6 - Enable non-stop mode.<BR>
  #   BIT7 - The direction of Guard Page for Pool Guard.
  #          0 - The returned pool is near the tail guard page.<BR>
  #          1 - The returned pool is near the head guard page.<BR>
  # @Prompt The Heap Guard feature mask
  gEfiMdeModulePkgTokenSpaceGuid.PcdHeapGuardPropertyMask|0x0|UINT8|0x30001054
```

os-d: Sure, I have added a note there

mdkinney (approved):

os-d:
@lgao4 this hasn't been up long, but fixes a latent bug in the heap guard system that platforms are hitting since a check was added that asserts in that case. Can you please review when you have a chance so we can unblock platforms?

lgao4 (approved):

PR #12493, opened 2026-04-22
Global: Fix some incorrect MODULE_TYPEs in INF files

EricGao2015:
# Description

The following drivers-PciPlatformDxe in UefiPayloadPkg, SemihostFs in ArmPkg, BootGraphicsResourceTableDxe, HiiResourcesSampleDxe, RegularExpressionDxe in MdeModulePkg, and AndroidFastbootTransportTcpDxe, AndroidFastbootTransportUsbDxe, GdbStub in EmbeddedPkg-do not follow UEFI driver model, so the MODULE_TYPE should be DXE_DRIVER instead of UEFI_DRIVER.

In EmbeddedPkg, FastbootTransportTcpDxe Driver is dependent on gEfiSimpleTextOutProtocolGuid, FastbootTransportUsbDxe Driver is dependent on gUsbDeviceProtocolGuid, GdbStub Driver is dependent on gEfiDebugSupportProtocolGuid. So the [Depex] should be updated too.

No functional changes.

## How This Was Tested

Pass CI check.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Filesystem/SemihostFs/SemihostFs.inf
+++ b/ArmPkg/Filesystem/SemihostFs/SemihostFs.inf
@@ -12,7 +12,7 @@
   INF_VERSION                    = 0x00010005
   BASE_NAME                      = SemihostFs
   FILE_GUID                      = C5B9C74A-6D72-4719-99AB-C59F199091EB
-  MODULE_TYPE                    = UEFI_DRIVER
+  MODULE_TYPE                    = DXE_DRIVER
   VERSION_STRING                 = 1.0
 
   ENTRY_POINT                    = SemihostFsEntryPoint
@@ -40,3 +40,5 @@
   gEfiSimpleFileSystemProtocolGuid
   gEfiDevicePathProtocolGuid
 
+[Depex]
+  TRUE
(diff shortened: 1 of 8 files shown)


mdkinney:
Some of the proposed changes from UEFI Driver to DXE Driver do not look correct.  For example NetworkPkg TLS that does Service Binding Protocol is part of UEFI Driver Model and should not be changed to a DXE Driver.

EricGao2015:
> Some of the proposed changes from UEFI Driver to DXE Driver do not look correct. For example NetworkPkg TLS that does Service Binding Protocol is part of UEFI Driver Model and should not be changed to a DXE Driver.

Thank you for your correction. I revereted the TlsDxe commit.
Must a driver that implements any one of the protocols required by the UEFI Driver Model be classified as a UEFI_DRIVER, even if it does not implement the Driver Binding Protocol?

mdkinney:
> > Some of the proposed changes from UEFI Driver to DXE Driver do not look correct. For example NetworkPkg TLS that does Service Binding Protocol is part of UEFI Driver Model and should not be changed to a DXE Driver.
> 
> Thank you for your correction. I revereted the TlsDxe commit. Must a driver that implements any one of the protocols required by the UEFI Driver Model be classified as a UEFI_DRIVER, even if it does not implement the Driver Binding Protocol?

Drivers that use the Driver Binding Protocol are a subset of all possible UEFI Drivers.  A good reference is here:

https://tianocore-docs.github.io/edk2-UefiDriverWritersGuide/draft/3_foundation/37_uefi_images/README.7.html#37-uefi-images

Image Types

PR #12495, opened 2026-04-23
Global: Remove duplicate items  in INF files

EricGao2015:
# Description

The INF specification said, each source file/protocol/guid/libraryclass/PCD must be listed only once per section. But many drivers don't follow this rule. This PR removed the duplicate items in INF files.

No functional changes.

## How This Was Tested

Pass CI check.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Bus/Pci/PciSioSerialDxe/PciSioSerialDxe.inf
+++ b/MdeModulePkg/Bus/Pci/PciSioSerialDxe/PciSioSerialDxe.inf
@@ -55,10 +55,9 @@
 
 [Protocols]
   gEfiSioProtocolGuid                           ## TO_START
-  gEfiDevicePathProtocolGuid                    ## TO_START
+  gEfiDevicePathProtocolGuid                    ## TO_START ## BY_START
   gEfiPciIoProtocolGuid                         ## TO_START
   gEfiSerialIoProtocolGuid                      ## BY_START
-  gEfiDevicePathProtocolGuid                    ## BY_START
 
 [FeaturePcd]
   gEfiMdeModulePkgTokenSpaceGuid.PcdSerialUseHalfHandshake|FALSE   ## CONSUMES
--- a/MdeModulePkg/Core/PiSmmCore/PiSmmIpl.inf
+++ b/MdeModulePkg/Core/PiSmmCore/PiSmmIpl.inf
@@ -75,9 +75,6 @@
   ## SOMETIMES_CONSUMES ## Event
   ## SOMETIMES_PRODUCES ## UNDEFINED # Used to do smm communication
   gEfiEventExitBootServicesGuid
-  ## SOMETIMES_CONSUMES ## Event
-  ## SOMETIMES_PRODUCES ## UNDEFINED # Used to do smm communication
-  gEfiEventReadyToBootGuid
   gEfiEventVirtualAddressChangeGuid             ## CONSUMES             ## Event
   gEfiEndOfDxeEventGroupGuid                    ## CONSUMES             ## Event
   gLoadFixedAddressConfigurationTableGuid       ## SOMETIMES_CONSUMES   ## SystemTable
--- a/ArmPkg/Drivers/ArmGicDxe/ArmGicDxe.inf
+++ b/ArmPkg/Drivers/ArmGicDxe/ArmGicDxe.inf
@@ -46,7 +46,6 @@
   IoLib
   PcdLib
   TimerLib
-  UefiLib
 
 [Protocols]
   gHardwareInterruptProtocolGuid  ## PRODUCES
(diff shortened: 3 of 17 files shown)


MdeModulePkg/Core/PiSmmCore/PiSmmIpl.inf
```diff
@@ -77,7 +77,6 @@
   gEfiEventExitBootServicesGuid
   ## SOMETIMES_CONSUMES ## Event
   ## SOMETIMES_PRODUCES ## UNDEFINED # Used to do smm communication
-  gEfiEventReadyToBootGuid
```
os-d, line 80: I think the two lines above this can be removed, too (although in general those comments are odd)

os-d (approved):

mdkinney (approved):

makubacki (approved):
Ideally, something would be done to prevent or reduce the likelihood of future duplicates being introduced, but fixing is fine for now.

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.inf
EricGao2015: The two comment lines have removed.

EricGao2015:
> Ideally, something would be done to prevent or reduce the likelihood of future duplicates being introduced, but fixing is fine for now.

Currently, BaseTools tolerates the duplicates，The duplicate ones can be gently merged. I'll consider adding some checks in BaseTools to detect the duplicates after the PR merged.

AshrafAliS (approved):

gguo11837463 (approved):
UefiPayloadPkg looks good from me.

MarsDoge (approved):

leiflindholm (approved):
Approved for ArmPkg, EmbeddedPkg, NetworkPkg, ShellPkg.

EricGao2015:
Could this PR be merged now?

MdeModulePkg/Bus/Pci/PciSioSerialDxe/PciSioSerialDxe.inf
```diff
@@ -58,7 +58,6 @@
   gEfiDevicePathProtocolGuid                    ## TO_START
   gEfiPciIoProtocolGuid                         ## TO_START
   gEfiSerialIoProtocolGuid                      ## BY_START
```
mikebeaton: Assuming the previous comments were correct, I think you should have `## TO_START ## BY_START` as the comment after the instance you are keeping. (Which I believe means consumes and produces.)

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.inf
mikebeaton: Assuming that these multi-line comments are meant to all apply to the item following them, then I am not sure just removing these is correct.

Seems like the best bet might be to move this item with its two comment lines up above, replacing the single item with its inline comments (which you are currently keeping).

MdeModulePkg/Bus/Pci/PciSioSerialDxe/PciSioSerialDxe.inf
EricGao2015: Applied.

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.inf
```diff
@@ -75,9 +75,6 @@
   ## SOMETIMES_CONSUMES ## Event
   ## SOMETIMES_PRODUCES ## UNDEFINED # Used to do smm communication
   gEfiEventExitBootServicesGuid
-  ## SOMETIMES_CONSUMES ## Event
-  ## SOMETIMES_PRODUCES ## UNDEFINED # Used to do smm communication
-  gEfiEventReadyToBootGuid
```
mikebeaton, line 80: Assuming that these multi-line comments are meant to all apply to the item following them, then I am not sure just removing these is correct.

Seems like the best bet might be to move this item with its two comment lines up above, replacing the single item with its inline comments (which you are currently keeping).

EricGao2015: > Assuming that these multi-line comments are meant to all apply to the item following them, then I am not sure just removing these is correct.

Seems the 2 comment lines don't apple to the item following them, you can refer to https://github.com/tianocore/edk2/commit/53ec4d7f98a6af8b00c891c3f13791967e52b6a3#diff-9fd9006043020cbccbc1c884003aede2a10664c8411de9ddde5396085bbff609R77-R81
So I prefer to remove the 3 lines together because they have added together before.

> 
> Seems like the best bet might be to move this item with its two comment lines up above, replacing the single item with its inline comments (which you are currently keeping).

mikebeaton: Well @os-d was also fine with it, so okay.

mikebeaton:
@leiflindholm - Polite ping - this failed to push yesterday.

PR #12497, opened 2026-04-23
StandaloneMmPkg: Fix possible infinite loop due invalid Hob

khaliid2040:
# Description

The second call to CreateMmFoundationHobList () expects the address of the beginning of the platform hob list in the third parameter. However it has being given the beginning of the HobList which is PHIT. This causes later wrong address calculations which in turns puts MM core in infinite loop, because the IPL passed hob without EFI_HOB_TYPE_END_OF_HOB_LIST.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

OVMF X64 with standaloneMm and SMM profiling enabled.

## Integration Instructions
N//A

### DIFF
--- a/StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
+++ b/StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
@@ -504,7 +504,7 @@ CreateMmHobList (
   Status = CreateMmFoundationHobList (
              (UINT8 *)HobList + PhitHobSize + PlatformHobSize,
              &FoundationHobSize,
-             HobList,
+             (UINT8 *)HobList + PhitHobSize,
              PlatformHobSize,
              MmFvBase,
              MmFvSize,


khaliid2040:
There seems a breakage with this patch. So converted to draft until it is resolved.

ardbiesheuvel:
I understand the problem, but I don't like the fix tbh. There are some weird interdependencies here that suggest that this entire code flow needs to be refactored to do things properly.

lzeng14:
Curious about why PcdCpuSmmRestrictedMemoryAccess = FALSE may cause two callings of CreateMmFoundationHobList() to return HobList with
different size and hobs ?

khaliid2040:
@ardbiesheuvel and @lzeng14 Between two calls the platform hob is copied to HobList. But, the offset of platform hob in the hob list ( the third parameter of CreateMmFoundationHobList) is wrong, it points to beginning of the hob while the beginning of the platform is in (HobList + PhitHobSize). I updated the patch which fixes the root cause.

niruiyu (approved):

PR #12502, opened 2026-04-25
BaseTools: Add ClangToolChain build plugin

apop5:
# Description

Add a new edk2-pytools build plugin that automatically configures CLANG_BIN for CLANGPDB and CLANGDWARF tool chains. The plugin resolves the clang binary path by checking, in order:

1. CLANG_BIN environment variable
2. System PATH
3. Default LLVM install path (Windows)
4. VS-installed clang (Windows)

The plugin also reports clang and linker (lld-link/lld) versions to the version aggregator for build logging.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Tested under windows with LLVM 
Tested under windows with MSVC LLVM installed 

Tested under linux with llvm
Tested under linux with CLANG_BIN set

## Integration Instructions
No integration necessary.

### DIFF
--- a/BaseTools/Plugin/ClangToolChain/ClangToolChain.py
+++ b/BaseTools/Plugin/ClangToolChain/ClangToolChain.py
@@ -1,0 +1,434 @@
+# @file ClangToolChain.py
+# Plugin to configures paths for the Clang tool chain
+##
+# This plugin works in conjunction with the tools_def
+#
+# Copyright (c) Microsoft Corporation
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+import logging
+import shutil
+from pathlib import Path
+from io import StringIO
+from edk2toolext.environment.plugintypes.uefi_build_plugin import (
+    IUefiBuildPlugin,
+)
+from edk2toolext.environment import shell_environment
+from edk2toolext.environment import version_aggregator
+from edk2toollib.utility_functions import GetHostInfo
+from edk2toollib.utility_functions import RunCmd
+import edk2toollib.windows.locate_tools as locate_tools
+from edk2toollib.windows.locate_tools import FindWithVsWhere
+
+SUPPORTED_TOOL_CHAINS = ("CLANGPDB", "CLANGDWARF")
+
+class ClangToolChain(IUefiBuildPlugin):
+
+    def do_post_build(self, thebuilder):
+        """No-op post-build hook required by IUefiBuildPlugin."""
+        return 0
+
+    def do_pre_build(self, thebuilder):
+        """Configure CLANG_BIN and report tool versions for CLANGPDB/CLANGDWARF.
+
+        Resolves the clang binary in this order:
+          1. CLANG_BIN environment variable
+          2. System PATH
+          3. Default LLVM install path (Windows)
+          4. VS-installed clang (Windows only)
+
+        Returns 0 on success, 1 if clang cannot be found.
+        """
+        self.Logger = logging.getLogger("ClangPdbToolChain")
+        ##
+        # CLANGPDB/CLANGDWARF
+        # - Need to find the clang path.
+        # - Report path and version for logging
+        #
+        # if CLANG_BIN already set the plugin will confirm it exists and
+        # get the version of clang
+        # If not set it will look for clang on the path.  If found it will
+        # configure for that.
+        # if still not found it will try the default install directory.
+        # finally an error will be reported if not found
+        ##
+        tool_chain_tag = thebuilder.env.GetValue("TOOL_CHAIN_TAG")
+        if tool_chain_tag.upper() not in SUPPORTED_TOOL_CHAINS:
+            return 0
+
+        HostInfo = GetHostInfo()
+        # Environment Variable
+        clang_bin_env = shell_environment.GetEnvironment().get_shell_var(
+            "CLANG_BIN"
+        )
+        ClangBin = Path(clang_bin_env, "clang.exe" if HostInfo.os == "Windows" else "clang") if clang_bin_env else None
+        # 1. Check CLANG_BIN from the environment
+        if ClangBin and not ClangBin.is_file():
+            self.Logger.error(
+                "CLANG_BIN is set to '%s' but '%s' was not found.",
+                clang_bin_env,
+                ClangBin,
+            )
+            return 1
+        # 2. CLANG_BIN was not set; check system path.
+        if not ClangBin:
+            ClangBin = Path(w) if (w := shutil.which("clang")) else None
+        # 3. Check Windows LLVM install path.
+        if not ClangBin and HostInfo.os == "Windows":
+            ClangBin = Path("C:/Program Files/LLVM/bin/clang.exe")
+        # 4. Finally try MSVC LLVM install.
+        if (not ClangBin or not ClangBin.is_file()) and HostInfo.os == "Windows":
+            vs_clang_path = self._configure_windows_vs_host_for_clang(
+                HostInfo
+            )
+            ClangBin = (
+                vs_clang_path
+                if vs_clang_path and vs_clang_path.is_file()
+                else None
+            )
+
+        if ClangBin and ClangBin.is_file():
+            clang_bin_parent = ClangBin.parent
+            if not clang_bin_env:
+                shell_environment.GetEnvironment().set_shell_var(
+                    "CLANG_BIN",
+                    clang_bin_parent.as_posix() + "/",
+                )
+            version_aggregator.GetVersionAggregator().ReportVersion(
+                "CLANG_BIN",
+                str(ClangBin),
+                version_aggregator.VersionTypes.INFO,
+            )
+            clang_version = self._get_clang_version(clang_bin_parent)
+            if clang_version:
+                version_aggregator.GetVersionAggregator().ReportVersion(
+                    "CLANG",
+                    clang_version,
+                    version_aggregator.VersionTypes.TOOL,
+                )
+            else:
+                self.Logger.warning("Could not determine clang version")
+
+            # Report the linker version based on toolchain
+            if tool_chain_tag == "CLANGPDB":
+                lld_version = self._get_linker_version(clang_bin_parent, "lld-link")
+                if lld_version:
+                    version_aggregator.GetVersionAggregator().ReportVersion(
+                        "lld-link",
+                        lld_version,
+                        version_aggregator.VersionTypes.TOOL,
+                    )
+                else:
+                    self.Logger.warning("Could not determine lld-link version")
+            elif tool_chain_tag == "CLANGDWARF":
+                lld_version = self._get_linker_version(clang_bin_parent, "lld")
+                if lld_version:
+                    version_aggregator.GetVersionAggregator().ReportVersion(
+                        "lld",
+                        lld_version,
+                        version_aggregator.VersionTypes.TOOL,
+                    )
+                else:
+                    self.Logger.warning("Could not determine lld version")
+        else:
+            self.Logger.error("Could not find clang executable.")
+            return 1
+
+        return 0
+
+    def _configure_windows_vs_host_for_clang(self, host_info):
+        """Configure the VS host environment and locate the VS-bundled clang.
+
+        Sets CLANG_HOST_BIN and relevant VS shell variables, then searches the
+        VS LLVM install tree for clang.exe.
+
+        Args:
+            host_info: HostInfo object from GetHostInfo().
+
+        Returns:
+            Path to clang.exe if found, otherwise None.
+        """
+
+        interesting_keys = [
+            "ExtensionSdkDir",
+            "INCLUDE",
+            "LIB",
+            "LIBPATH",
+            "UniversalCRTSdkDir",
+            "UCRTVersion",
+            "WindowsLibPath",
+            "WindowsSdkBinPath",
+            "WindowsSdkDir",
+            "WindowsSdkVerBinPath",
+            "WindowsSDKVersion",
+            "VCToolsInstallDir",
+            "Path",
+        ]
+
+        # check to see if host is configured
+        # HostType for VS tools should be (defined in tools_def):
+        # x86   == 32bit Intel
+        # x64   == 64bit Intel
+        # arm64 == 64bit Arm
+        host_type = shell_environment.GetEnvironment().get_shell_var(
+            "CLANG_VS_HOST"
+        )
+        if host_type is not None:
+            host_type = host_type.lower()
+            self.Logger.info(
+                f"CLANG_VS_HOST defined by environment.  Value is {host_type}"
+            )
+        else:
+            # figure it out based on host info
+            if host_info.arch == "x86":
+                if host_info.bit == "32":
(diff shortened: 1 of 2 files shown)


BaseTools/Plugin/ClangToolChain/ClangToolChain.py
```diff
@@ -0,0 +1,432 @@
...
+        #
+        # if CLANG_BIN already set the plugin will confirm it exists and
+        # get the version of clang
+        # If not set it will look for clang on the path.  If found it will
+        # configure for that.
+        # if still not found it will try the default install directory.
+        # finally an error will be reported if not found
+        ##
+        tool_chain_tag = thebuilder.env.GetValue("TOOL_CHAIN_TAG")
+        if tool_chain_tag in ("CLANGPDB", "CLANGDWARF"):
```
Javagedes: Please reverse this with an early abort so we don't have the core logic nested in an if statement.

Javagedes: please make a global variable at the top of this file that is a list of the supported toolchains. it does not look like we use these anywhere else in the file but still a simple enough change in case we eventually need to use these more than this single time.

BaseTools/Plugin/ClangToolChain/ClangToolChain.py
```diff
@@ -0,0 +1,432 @@
...
+            HostInfo = GetHostInfo()
+            clang_bin_env = shell_environment.GetEnvironment().get_shell_var(
+                "CLANG_BIN"
+            )
+            ClangBin = Path(clang_bin_env, "clang") if clang_bin_env else None
+
+            # 1. Check CLANG_BIN from the environment
+            if not ClangBin or not ClangBin.is_file():
```
Javagedes: Based on my above comment, if ClangBin exists but is not a file, then I think we need to throw a warning so the user knows we are not using what they are specifying (because it does not contain the clang binary) and are instead going to search for our own.

BaseTools/Plugin/ClangToolChain/ClangToolChain.py
```diff
@@ -0,0 +1,432 @@
...
+        # if still not found it will try the default install directory.
+        # finally an error will be reported if not found
+        ##
+        tool_chain_tag = thebuilder.env.GetValue("TOOL_CHAIN_TAG")
+        if tool_chain_tag in ("CLANGPDB", "CLANGDWARF"):
+            HostInfo = GetHostInfo()
+            clang_bin_env = shell_environment.GetEnvironment().get_shell_var(
+                "CLANG_BIN"
+            )
+            ClangBin = Path(clang_bin_env, "clang") if clang_bin_env else None
```
Javagedes: I think I would prefer you make this `clang` or `clang.exe` depending on the OS, that way if ClangBin exists but is not a file, then we've got an actual error. Right now you are treating it as "<clang_bin_env>/clang does not exist, we must be on windows, so just set that is clang bin and move on. 

But this will hide the issue of someone setting CLANG_BIN to a bad path where clang does not actually exist.

BaseTools/Plugin/ClangToolChain/ClangToolChain.py
```diff
@@ -0,0 +1,432 @@
...
+
+            # 1. Check CLANG_BIN from the environment
+            if not ClangBin or not ClangBin.is_file():
+                # 2. CLANG_BIN didn't point to a valid directory.
+                # Check system path.
+                ClangBin = shutil.which("clang")
```
Javagedes: you are reusing "ClangBin" with multiple types and it is getting confusing. It starts off as a `Path` object but can end up as a string type object. I'd simplify everything to just be a `Path` object, so you don't have to continue doing special `Path` wrapping down beliw.

BaseTools/Plugin/ClangToolChain/ClangToolChain.py
```diff
@@ -0,0 +1,432 @@
...
+                        vs_clang_path = (
+                            self._configure_windows_vs_host_for_clang(
+                                HostInfo
+                            )
+                        )
+                        if vs_clang_path and vs_clang_path.is_file():
+                            ClangBin = vs_clang_path
+            if ClangBin and Path(ClangBin).is_file():
+                clang_bin_parent = Path(ClangBin).parent
+                shell_environment.GetEnvironment().set_shell_var(
```
Javagedes, line 93: you are setting CLANG_BIN above if you find it, but then fall back to setting CLANG_BIN again here. I would delay the setting of CLANG_BIN to a single point (probably here) instead of sporadically setting it throughout this code.

BaseTools/Plugin/ClangToolChain/ClangToolChain.py
```diff
@@ -0,0 +1,432 @@
...
+                elif HostInfo.os == "Windows":
+                    # 3. Check Windows Llvm Install Path.
+                    ClangBin = Path("C:/Program Files/LLVM/bin/clang")
+                    if ClangBin.is_file():
+                        shell_environment.GetEnvironment().set_shell_var(
+                            "CLANG_BIN",
+                            ClangBin.parent.as_posix() + "/",
+                        )
+                    else:
+                        # 4.Finally try MSVC LLVM install
```
Javagedes: nit: you have `3.` outside of the first if statement but `4.` inside of the else statement. I'd move one of them up or down a tab to make it a bit more clear.

BaseTools/Plugin/ClangToolChain/ClangToolChain.py
```diff
@@ -0,0 +1,432 @@
...
+        Sets CLANG_HOST_BIN and relevant VS shell variables, then searches the
+        VS LLVM install tree for clang.exe.
+
+        Args:
+            host_info: HostInfo object from GetHostInfo().
+
+        Returns:
+            Path to clang.exe if found, otherwise None.
+        """
+        import edk2toollib.windows.locate_tools as locate_tools
```
Javagedes: We should probably just import this at the top, unless you have a reason it is here.

BaseTools/Plugin/ClangToolChain/ClangToolChain.py
```diff
@@ -0,0 +1,432 @@
...
+        Checks the shell variable *varname* first; falls back to vswhere.
+
+        Args:
+            vs_version: VS version string passed to FindWithVsWhere (e.g. "vs2022").
+            varname: Shell variable name to check first, or None.
+
+        Returns:
+            Install path string, or None if not found.
+        """
+        from edk2toollib.windows.locate_tools import FindWithVsWhere
```
Javagedes: same here.

BaseTools/Plugin/ClangToolChain/ClangToolChain.py
apop5: Updated in latest

apop5: done, and added the missing .upper() during the comparison

BaseTools/Plugin/ClangToolChain/ClangToolChain.py
apop5: done.

BaseTools/Plugin/ClangToolChain/ClangToolChain.py
apop5: should be addressed

BaseTools/Plugin/ClangToolChain/ClangToolChain.py
apop5: unified to only set at the bottom, instead of setting in multiple places.

BaseTools/Plugin/ClangToolChain/ClangToolChain.py
```diff
@@ -0,0 +1,435 @@
...
+            return 0
+
+        HostInfo = GetHostInfo()
+        # Environment Variable
+        clang_bin_env = shell_environment.GetEnvironment().get_shell_var(
+            "CLANG_BIN"
+        )
+        ClangBin = Path(clang_bin_env, "clang.exe" if HostInfo.os == "Windows" else "clang") if clang_bin_env else None
+
+        # 1. Check CLANG_BIN from the environment
```
Javagedes, line 65: All of your nested if statements basically try and set clangbin. If it fails, do another inner if statement. so on and so on. Due to that, I think you can collapse all the nesting into top level if statements:

```python
# 1. Check CLANG_BIN from the environment.
if ClangBin and not ClangBin.is_file():
    self.Logger.error(
        "CLANG_BIN is set to '%s' but '%s' was not found.",
        clang_bin_env,
        ClangBin,
    )
    return 1

# 2. CLANG_BIN was not set; check the system PATH.
if not ClangBin:
    ClangBin = Path(w) if (w := shutil.which("clang")) else None

# 3. On Windows, check the default LLVM install.
if not ClangBin and HostInfo.os == "Windows":
    candidate = Path("C:/Program Files/LLVM/bin/clang.exe")
    if candidate.is_file():
        ClangBin = candidate

# 4. Finally, try the MSVC LLVM install.
if not ClangBin and HostInfo.os == "Windows":
    candidate = self._configure_windows_vs_host_for_clang(HostInfo)
    if candidate and candidate.is_file():
        ClangBin = candidate
```

makubacki:
@Javagedes & @apop5, at least at first glance it looks like there is duplication in [BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py](https://github.com/tianocore/edk2/blob/master/BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py).

https://github.com/tianocore/edk2/blob/d7f133a16428a5d4d1d1021a48d4192cc88a6a13/BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py#L321-L338

https://github.com/tianocore/edk2/blob/d7f133a16428a5d4d1d1021a48d4192cc88a6a13/BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py#L447-L453

...

Can you clarify if this is expected?

apop5:
> @Javagedes & @apop5, at least at first glance it looks like there is duplication in [BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py](https://github.com/tianocore/edk2/blob/master/BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py).
> 
> https://github.com/tianocore/edk2/blob/d7f133a16428a5d4d1d1021a48d4192cc88a6a13/BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py#L321-L338
> 
> https://github.com/tianocore/edk2/blob/d7f133a16428a5d4d1d1021a48d4192cc88a6a13/BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py#L447-L453
> 
> ...
> 
> Can you clarify if this is expected?

Yes, this is expected. 

The WindowsVsToolChain is about getting access to make when running under windows. LLVM tool chain does not provide a make application. 
https://github.com/tianocore/edk2/blob/master/BaseTools/Conf/tools_def.template#L984

This plugin is about CLANG_BIN. CLANG_BIN is a prefix which supplies where the LLVM tool chain exists. 
https://github.com/tianocore/edk2/blob/master/BaseTools/Conf/tools_def.template#L996

BaseTools/Plugin/ClangToolChain/ClangToolChain.py
apop5: applied in latest.

makubacki:
> Yes, this is expected.

I don’t have time to review it in detail at the moment, but at a high level the separation between the two does not appear particularly clean. That said, if you and the rest of the reviewers are comfortable with it, I’m fine proceeding.

makubacki:
> I don’t have time to review it in detail at the moment, but at a high level the separation between the two does not appear particularly clean. That said, if you and the rest of the reviewers are comfortable with it, I’m fine proceeding.

I wanted to follow up with a more detailed reply but conceptually I think it'd be easier to reason about plugin support for Clang if all related logic could be consolidated into the new Clang plugin.

BaseTools/Plugin/ClangToolChain/ClangToolChain.py
```diff
@@ -0,0 +1,434 @@
...
+from io import StringIO
+from edk2toolext.environment.plugintypes.uefi_build_plugin import (
+    IUefiBuildPlugin,
+)
+from edk2toolext.environment import shell_environment
+from edk2toolext.environment import version_aggregator
+from edk2toollib.utility_functions import GetHostInfo
+from edk2toollib.utility_functions import RunCmd
+import edk2toollib.windows.locate_tools as locate_tools
+from edk2toollib.windows.locate_tools import FindWithVsWhere
```
antklein, line 21: Recommended re-organization.

```suggestion
import logging
import shutil
from io import StringIO
from pathlib import Path

import edk2toollib.windows.locate_tools as locate_tools
from edk2toolext.environment import shell_environment, version_aggregator
from edk2toolext.environment.plugintypes.uefi_build_plugin import (
    IUefiBuildPlugin,
)
from edk2toollib.utility_functions import GetHostInfo, RunCmd
from edk2toollib.windows.locate_tools import FindWithVsWhere
```

BaseTools/Plugin/ClangToolChain/ClangToolChain.py
```diff
@@ -0,0 +1,434 @@
...
+        if ClangBin and not ClangBin.is_file():
+            self.Logger.error(
+                "CLANG_BIN is set to '%s' but '%s' was not found.",
+                clang_bin_env,
+                ClangBin,
+            )
+            return 1
+        # 2. CLANG_BIN was not set; check system path.
+        if not ClangBin:
+            ClangBin = Path(w) if (w := shutil.which("clang")) else None
```
antklein, line 75: Does `shutil.which("clang")` work for both Windows and Unix? Does it need to differentiate between "clang" and "clang.exe"?

antklein, line 75: Should `.resolve()` be added to handle symlinks?
```suggestion
            ClangBin = Path(w).resolve() if (w := shutil.which("clang")) else None
```

apop5: It worked in testing on Windows 11 and Docker container based on ubuntu 24

antklein (approved):

BaseTools/Plugin/WindowsVsToolChain/WindowsVsToolChain.py
antklein: Minor change to fix this comparison syntax, since you are making edits to the file.
```suggestion
            if shell_environment.GetEnvironment().get_shell_var("VS2019_PREFIX") is not None:
```

antklein:
Hi @Javagedes, @gapalomi, @lgao4, @mdkinney, @mikebeaton, @ponchofigueroa, @spbrogan, @YuweiChen1110, @AshrafAliS,

I'm requesting your support in reviewing this PR. It has been active for many months, and all current reviews have been addressed.

Javagedes (approved):

lgao4 (approved):

PR #12505, opened 2026-04-27
3945_arm_cca_v5_s5: Introduce an Arm CCA helper library (ArmCcaLib)

samimujawar:
# Introduce an Arm CCA helper library (ArmCcaLib)

This series introduces ArmCcaLib to ArmVirtPkg, providing helper APIs for detecting whether the current execution state is a Realm and to query the Realm IPA width. To minimise costly Realm exits from repeated RSI calls, GUID HOBs are added to cache key Realm information such as IPA width and execution state. The library transparently uses cached HOB data when available and falls back to RSI calls during early boot. 

A Null instance of ArmCcaLib is included to preserve compatibility for non-CCA builds. In addition, a helper is added to identify protected (Realm) MMIO regions using RSI_IPA_STATE_GET.

Note: This is a PR series that includes patches from PR https://github.com/tianocore/edk2/pull/12224

- [ ] Breaking change?
  - No
- [ ] Impacts security?
  - No
- [ ] Includes tests?

## How This Was Tested

Only build tested. Actual functionality can only be tested when the remaining patches from PR https://github.com/tianocore/edk2/pull/12224 are integrated in the future.

## Integration Instructions

N/A


ardbiesheuvel (approved):

leiflindholm:
Following onto a comment I made on a previous PR, and a calibration with Ard: let's have a discussion once this whole thing is merged as to whether anything should move to MdePkg/UefiCpuPkg (because it's architectural) or whether it makes more sense for it to stay in ArmVirtPkg (as here is the only place it'll be used).

samimujawar:
> Following onto a comment I made on a previous PR, and a calibration with Ard: let's have a discussion once this whole thing is merged as to whether anything should move to MdePkg/UefiCpuPkg (because it's architectural) or whether it makes more sense for it to stay in ArmVirtPkg (as here is the only place it'll be used).

Ack. I will add this to the list of modules/libraries to be moved to MdePkg/UefiCpuPkg.

leiflindholm:
> > Following onto a comment I made on a previous PR, and a calibration with Ard: let's have a discussion once this whole thing is merged as to whether anything should move to MdePkg/UefiCpuPkg (because it's architectural) or whether it makes more sense for it to stay in ArmVirtPkg (as here is the only place it'll be used).
> 
> Ack. I will add this to the list of modules/libraries to be moved to MdePkg/UefiCpuPkg.

To clarify: I'm no longer certain that is the right thing to do, but I think the topic merits discussion (once the whole feature is merged).

PR #12509, opened 2026-04-27
ArmPkg: Flush data cache when mapping region as WC

cfernald:
# Description

When mapping a region as WC (Normal Memory Non-Cacheable), dirty cache lines may persistent, and depending on the caching behavior of the platform, may be written back corrupting data in the future. To prevent this, the data cache should be flushed and invalidated for the range after the mapping has been changed, but before accessing the memory. Currently, the EFI_CPU_ARCH_PROTOCOL implementation makes the caller responsible for this, but this is prone to obscure and difficult to diagnose memory issues if this is missed. It is safer to always perform the flush when mapping as WC, even if this may be redundant in some cases.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested booting to operating system and with one-off shell test application.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/CpuDxe/CpuMmuCommon.c
+++ b/ArmPkg/Drivers/CpuDxe/CpuMmuCommon.c
@@ -204,6 +204,23 @@ SetGcdMemorySpaceAttributes (
 }
 
 /**
+  Checks if the specified ARM translation table attributes are cacheable memory.
+
+  @param[in]  Attributes  ARM memory attributes.
+
+  @retval TRUE if the attributes are cacheable, FALSE otherwise.
+**/
+STATIC
+BOOLEAN
+IsArmAttributeCacheable (
+  IN UINT64  Attributes
+  )
+{
+  return ((Attributes & TT_ATTR_INDX_MASK) == TT_ATTR_INDX_MEMORY_WRITE_THROUGH) ||
+         ((Attributes & TT_ATTR_INDX_MASK) == TT_ATTR_INDX_MEMORY_WRITE_BACK);
+}
+
+/**
   This function modifies the attributes for the memory region specified by BaseAddress and
   Length from their current attributes to the attributes specified by Attributes.
 
@@ -238,6 +255,7 @@ CpuSetMemoryAttributes (
   UINTN       RegionBaseAddress;
   UINTN       RegionLength;
   UINTN       RegionArmAttributes;
+  BOOLEAN     FlushCache;
 
   if (mIsFlushingGCD) {
     return EFI_SUCCESS;
@@ -261,8 +279,34 @@ CpuSetMemoryAttributes (
   if (EFI_ERROR (Status) || (RegionArmAttributes != ArmAttributes) ||
       ((BaseAddress + Length) > (RegionBaseAddress + RegionLength)))
   {
-    return ArmSetMemoryAttributes (BaseAddress, Length, EfiAttributes, 0);
-  } else {
-    return EFI_SUCCESS;
+    //
+    // If the region was previously mapped as a cacheable normal memory,
+    // and is changed to device or non-cacheable memory, ensure any
+    // stale cache lines are written back and invalidated. If this is not done,
+    // depending on the caching behavior of the platform, dirty cache lines may
+    // be written back corrupting data in the future, or stale cache lines may
+    // persist if caching is enabled later.
+    //
+    FlushCache = FALSE;
+    if (!EFI_ERROR (Status) && IsArmAttributeCacheable (RegionArmAttributes) && !IsArmAttributeCacheable (ArmAttributes)) {
+      FlushCache = TRUE;
+    }
+
+    Status = ArmSetMemoryAttributes (BaseAddress, Length, EfiAttributes, 0);
+    if (EFI_ERROR (Status)) {
+      return Status;
+    }
+
+    if (FlushCache) {
+      //
+      // A data barrier is required to ensure that all page updates are complete and the TLB is flushed prior to the
+      // cache invalidation to avoid any unexpected cache lines being created.
+      //
+
+      ArmDataSynchronizationBarrier ();
+      WriteBackInvalidateDataCacheRange ((VOID *)(UINTN)BaseAddress, (UINTN)Length);
+    }
   }
+
+  return EFI_SUCCESS;
 }


vishalo:
@cfernald does this change have any impact on overall boot time on the platform this was tested?

ArmPkg/Drivers/CpuDxe/CpuMmuCommon.c
```diff
@@ -261,8 +261,23 @@ CpuSetMemoryAttributes (
...
+    }
+
+    //
+    // When mapping a region as WC (Normal Memory Non-Cacheable), ensure any
+    // stale cache lines are written back and invalidated. If this is not done,
+    // depending on the caching behavior of the platform, dirty cache lines may
+    // be written back corrupting data in the future, or stale cache lines may
+    // persist if caching is enabled later.
+    //
+    if ((EfiAttributes & EFI_MEMORY_CACHETYPE_MASK) == EFI_MEMORY_WC) {
```
leiflindholm: To the point of the comment you're adding, the problem is with transition between WC and !WC. So should we be checking
`(EfiAttributes ^ RegionArmAttributes) & EFI_MEMORY_WC`
?

ardbiesheuvel: Doesn't this apply equally to `EFI_MEMORY_UC`?

leiflindholm: Good point. I guess it does.
So
`(EfiAttributes ^ RegionArmAttributes) & (EFI_MEMORY_UC | EFI_MEMORY_WC)`
?

os-d: When @cfernald and I discussed this, the rationale was that device memory shouldn't be mapped as normal cacheable memory at any point. Perhaps there are scenarios we are unaware of here. But if you have a device there and flush something that exists in the cache to it, that seems unpredictable. If you were to do anything for `EFI_MEMORY_UC` an invalidate (no write back) would seem safer? Though then it is also questionable about deciding for the platform that the cache may not be written back.

ardbiesheuvel: > When @cfernald and I discussed this, the rationale was that device memory shouldn't be mapped as normal cacheable memory at any point.

Agreed, but that is not what I am referring to here. A shareable, cached mapping of a physical region that does not take part in the cache coherency protocol will cause lockups and other problems.

But we are specifically dealing with a cached mapping that is being converted into a WC mapping, and so we should be able to assume that the underlying physical region tolerates being mapped as cacheable.

> Perhaps there are scenarios we are unaware of here. But if you have a device there and flush something that exists in the cache to it, that seems unpredictable. If you were to do anything for `EFI_MEMORY_UC` an invalidate (no write back) would seem safer? Though then it is also questionable about deciding for the platform that the cache may not be written back.

There is no difference whatsoever between WC and UC mappings of DRAM in this regard. WC permits unaligned accesses and DC ZVA instructions, and so it behaves more as ordinary memory when writing, but in terms of cache behavior, we should treat them the same.

os-d: Do we expect UC mappings of DRAM? If we do (and thinking on it, I believe we do since core code assume x86 caching semantics and so tends to set UC on host memory shared with devices) then I agree, we should also have this same behavior for UC.

We won't be able to distinguish between what is actually device memory vs DRAM, but I think we can take the stance that when device memory is being mapped, there should not have been a cacheable mapping for it and so invalidating and writing the data cache is safe.

ardbiesheuvel: > Do we expect UC mappings of DRAM? If we do (and thinking on it, I believe we do since core code assume x86 caching semantics and so tends to set UC on host memory shared with devices) then I agree, we should also have this same behavior for UC.
> 

`UC` is generally not the best choice for mapping DRAM, but both the UEFI spec and the ARM ARM permit it, and it is widespread in existing code.

> We won't be able to distinguish between what is actually device memory vs DRAM, but I think we can take the stance that when device memory is being mapped, there should not have been a cacheable mapping for it and so invalidating and writing the data cache is safe.

That was precisely my point. Once a cacheable+shareable mapping exists, the cat is out of the bag and any bad behavior that we might be able to trigger by explicit cache maintenance is likely to occur spontaneously as well, due to the fact that MMIO regions are not set up to take part in the interconnect's cache coherency protocols.

cfernald: @leiflindholm , to your original point, yes I can optimize this by only doing this when the type is changing.

@ardbiesheuvel , I'm happy to add UC as a condition here as well. As Oliver mentioned, it was really an opinionated optimization to only focus on WC but for the sake of safety it is probably best to just do both.

I will add both of those changes.

cfernald: @leiflindholm & @ardbiesheuvel , apologies for the delay getting back to this. I have updated the logic to strictly check for a remap from a cacheable type to a non-cacheable type. I'll resolve this thread. Please let me know if you have outstanding concerns.

cfernald:
> @cfernald does this change have any impact on overall boot time on the platform this was tested?

No meaningful impact. On the systems I've tested, this almost never occurs naturally so this may not even occur on a given boot. Most of the testing is done against a test that forces the scenario.

ArmPkg/Drivers/CpuDxe/AArch64/Mmu.c
```diff
@@ -573,3 +573,19 @@ GetMemoryRegion (
...
+  @retval TRUE if the attributes are cacheable, FALSE otherwise.
+**/
+BOOLEAN
+IsArmAttributeCacheable (
+  IN UINT64  Attributes
+  )
+{
+  return ((Attributes & TT_ATTR_INDX_MASK) == TT_ATTR_INDX_MEMORY_WRITE_THROUGH) ||
+         ((Attributes & TT_ATTR_INDX_MASK) == TT_ATTR_INDX_MEMORY_WRITE_BACK);
+}
```
ardbiesheuvel: Please make this a `STATIC` local function in `CpuMmuCommon.c`

ArmPkg/Drivers/CpuDxe/CpuMmuCommon.c
```diff
@@ -261,8 +262,29 @@ CpuSetMemoryAttributes (
...
+      FlushCache = TRUE;
+    }
+
+    Status = ArmSetMemoryAttributes (BaseAddress, Length, EfiAttributes, 0);
+    if (EFI_ERROR (Status)) {
+      return Status;
+    }
+
+    if (FlushCache) {
+      ArmDataSynchronizationBarrier ();
```
ardbiesheuvel, line 306: Please add a comment that the `DSB` is needed to defer the cache maintenance until after TLB invalidation has completed (assuming that is why you added it)

cfernald: That is the intention, to ensure all page updates and TLB invalidation have completed prior to the write-back invalidate to remove any possibility for new cache lines being pulled in after this barrier. 

It may be a slightly redundant as there are already data barriers in the page table swap entry. Let me know if you have an opinion here.

ArmPkg/Drivers/CpuDxe/AArch64/Mmu.c
cfernald: Done.

ArmPkg/Drivers/CpuDxe/CpuMmuCommon.c
cfernald: Added the comment though, let me know if you believe any change is needed to the barrier.

ardbiesheuvel (approved):

rcran-ampere:
This change is causing a crash on my Ampere Altra systems (an ADLINK Ampere Altra Dev Kit with 96GB RAM and an ASRock Rack ALTRAD8UD-1L2T with 1.5TB). I verified that the commit before this one boots successfully. I added some DEBUG lines, and see the following:
```
add-symbol-file /home/bcran/src/uefi/Build/Altra1L2T/DEBUG_GCC/AARCH64/ArmPkg/Drivers/CpuDxe/CpuDxe/DEBUG/ArmCpuDxe.dll 0xFA876000
Loading driver at 0x000FA875000 EntryPoint=0x000FA8793F0 ArmCpuDxe.efi
InstallProtocolInterface: BC62157E-3E33-4FEC-9920-2D3B36D750DF FA99BD18
ProtectUefiImageCommon - 0xFA99B040
  - 0x00000000FA875000 - 0x000000000000C000
PROGRESS CODE: V03040002 I0
ReplaceTableEntry: splitting block entry with MMU disabled
InstallProtocolInterface: 26BACCB1-6F42-11D4-BCE7-0080C73C8881 FA87D0A0
InstallProtocolInterface: F4560CF6-40EC-4B4A-A192-BF1D57D0B189 FA87D068
MemoryProtectionCpuArchProtocolNotify:
InitializeDxeNxMemoryProtectionPolicy: StackBase = 0x00000000BF9C0000  StackSize = 0x0000000000020000
InitializeDxeNxMemoryProtectionPolicy: applying strict permissions to active memory regions
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: Status: Success, IsArmAttributeCacheable(RegionArmAttributes): 1, IsArmAttributeCacheable(ArmAttributes): 0
CpuSetMemoryAttributes: calling ArmSetMemoryAttributes
CpuSetMemoryAttributes: FlushCache is TRUE.
CpuSetMemoryAttributes: calling ArmDataSynchronizationBarrier
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x0000000090000000, 798883840)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: Status: Success, IsArmAttributeCacheable(RegionArmAttributes): 1, IsArmAttributeCacheable(ArmAttributes): 0
CpuSetMemoryAttributes: calling ArmSetMemoryAttributes
CpuSetMemoryAttributes: FlushCache is TRUE.
CpuSetMemoryAttributes: calling ArmDataSynchronizationBarrier
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x00000000BF9C0000, 4096)


Synchronous Exception at 0x00000000FA879B7C
PC 0x0000FA879B7C (0x0000FA875000+0x00004B7C) [ 0] ArmCpuDxe.dll
PC 0x0000FA8764D4 (0x0000FA875000+0x000014D4) [ 0] ArmCpuDxe.dll
PC 0x0000FA878754 (0x0000FA875000+0x00003754) [ 0] ArmCpuDxe.dll
PC 0x0000BF9F057C (0x0000BF9E0000+0x0001057C) [ 1] DxeCore.dll
PC 0x0000BF9E175C (0x0000BF9E0000+0x0000175C) [ 1] DxeCore.dll
PC 0x0000BF9E511C (0x0000BF9E0000+0x0000511C) [ 1] DxeCore.dll
PC 0x0000BF9F4900 (0x0000BF9E0000+0x00014900) [ 1] DxeCore.dll
PC 0x0000BF9E7508 (0x0000BF9E0000+0x00007508) [ 1] DxeCore.dll
PC 0x0000FA879700 (0x0000FA875000+0x00004700) [ 2] ArmCpuDxe.dll
PC 0x0000BF9EAE90 (0x0000BF9E0000+0x0000AE90) [ 3] DxeCore.dll
PC 0x0000BF9F91B8 (0x0000BF9E0000+0x000191B8) [ 3] DxeCore.dll
PC 0x0000BF9E8E44 (0x0000BF9E0000+0x00008E44) [ 3] DxeCore.dll

[ 0] /home/bcran/src/uefi/Build/Altra1L2T/DEBUG_GCC/AARCH64/ArmPkg/Drivers/CpuDxe/CpuDxe/DEBUG/ArmCpuDxe.dll
[ 1] /home/bcran/src/uefi/Build/Altra1L2T/DEBUG_GCC/AARCH64/MdeModulePkg/Core/Dxe/DxeMain/DEBUG/DxeCore.dll
[ 2] /home/bcran/src/uefi/Build/Altra1L2T/DEBUG_GCC/AARCH64/ArmPkg/Drivers/CpuDxe/CpuDxe/DEBUG/ArmCpuDxe.dll
[ 3] /home/bcran/src/uefi/Build/Altra1L2T/DEBUG_GCC/AARCH64/MdeModulePkg/Core/Dxe/DxeMain/DEBUG/DxeCore.dll

  X0 0x00000000BF9C0000   X1 0x0000000000001000   X2 0x00000000FA879B7C   X3 0x0000000000000040
  X4 0x0000000000000001   X5 0x0000000000000001   X6 0x0000000000000000   X7 0x00000000FA87AB23
  X8 0x0000000000000002   X9 0x0000000000000001  X10 0x00000000000000FF  X11 0x0000000000000002
 X12 0x0000000000000002  X13 0x0000000000000000  X14 0x0000000000000000  X15 0x0000000000000000
 X16 0x00000000FA876000  X17 0x0000000000000000  X18 0x0000000000000000  X19 0x00000000BF9C0040
 X20 0x00000000BF9C1000  X21 0x00000000FA879B7C  X22 0x0000000000000040  X23 0x0000000000000001
 X24 0x0000000000006000  X25 0x0000000000006000  X26 0x00000000FA996C18  X27 0x00000000FFBFF420
 X28 0x00000000FFBFF420   FP 0x00000000BF9DF800   LR 0x00000000FA8764D4  

  V0 0x0000000000000000 0000000000000000   V1 0x0000000000000000 41732F4579C98000
  V2 0x0000000000000000 4089000000000000   V3 0x0000000000000000 0000000000000000
  V4 0x0000000000000000 0000000000000000   V5 0x0000000000000000 0000000000000000
  V6 0x0000000000000000 0000000000000000   V7 0x0000000000000000 0000000000000000
  V8 0x0000000000000000 0000000000000000   V9 0x0000000000000000 0000000000000000
 V10 0x0000000000000000 0000000000000000  V11 0x0000000000000000 0000000000000000
 V12 0x0000000000000000 0000000000000000  V13 0x0000000000000000 0000000000000000
 V14 0x0000000000000000 0000000000000000  V15 0x0000000000000000 0000000000000000
 V16 0x0000000000000000 0000000000000000  V17 0x0000000000000000 0000000000000000
 V18 0x0000000000000000 0000000000000000  V19 0x0000000000000000 0000000000000000
 V20 0x0000000000000000 0000000000000000  V21 0x0000000000000000 0000000000000000
 V22 0x0000000000000000 0000000000000000  V23 0x0000000000000000 0000000000000000
 V24 0x0000000000000000 0000000000000000  V25 0x0000000000000000 0000000000000000
 V26 0x0000000000000000 0000000000000000  V27 0x0000000000000000 0000000000000000
 V28 0x0000000000000000 0000000000000000  V29 0x0000000000000000 0000000000000000
 V30 0x0000000000000000 0000000000000000  V31 0x0000000000000000 0000000000000000

  SP 0x00000000BF9DF800  ELR 0x00000000FA879B7C  SPSR 0x800002C9  FPSR 0x00000010
 ESR 0x9600014B          FAR 0x00000000BF9C0000

 ESR : EC 0x25  IL 0x1  ISS 0x0000014B

Data abort: Access flag fault, third level

Stack dump:
  00000BF9DF700: 654D746553757043 7274744179726F6D 203A736574756269 20676E696C6C6163
  00000BF9DF720: 6361426574697257 64696C61766E496B 4361746144657461 676E615265686361
  00000BF9DF740: 3030303078302865 4339464230303030 3034202C30303030 FF9F000A0D293639
  00000BF9DF760: 00000000BF9C1000 0000000000000001 0000001E00000001 00000000FA87B027
  00000BF9DF780: 00000000BF9DF830 00000000FA8786F8 0000000000000018 00000000FA87C741
  00000BF9DF7A0: 00000000BF9C0000 0000000000001000 0000000000000001 0000000000006000
  00000BF9DF7C0: 0000000000006000 00000000FA996C18 00000000FFBFF420 00000000FFBFF420
  00000BF9DF7E0: FF9F000000000B3F 00000000BFFFE000 00000000BF9DF800 00000000FA876514
> 00000BF9DF800: 00000000BF9DF830 00000000FA878754 0000000000000018 00000000FA87C741
  00000BF9DF820: 00000000BF9C0000 0000000000001000 00000000BF9DF8A0 00000000BF9F057C
  00000BF9DF840: 0000000000000000 0000000000000024 0000000000006000 0000000000000000
  00000BF9DF860: 0000000000001000 00000000BF9C0000 00000000BF9DF8A0 0000000000000000
  00000BF9DF880: 00000000BF9FC950 004000000000070C 0000000000020000 00000000BF9C0000
  00000BF9DF8A0: 00000000BF9DF970 00000000BF9E175C 00000000BF9C0000 0000000000006000
  00000BF9DF8C0: 0000000000001000 00000000BF9C1000 00000000BF9C1000 0000000000000001
  00000BF9DF8E0: 800000000000400E 0000000000000000 000000000002701F FFFFFFFFFFFD9FFF
ASSERT [ArmCpuDxe] CpuExceptionHandlerLib.c(438): ((BOOLEAN)(0==1))
```

cfernald:
> This change is causing a crash on my Ampere Altra systems (an ADLINK Ampere Altra Dev Kit with 96GB RAM and an ASRock Rack ALTRAD8UD-1L2T with 1.5TB). I verified that the commit before this one boots successfully. I added some DEBUG lines, and see the following:
> ...

I think I may have missed an obvious issue in the change in the event memory types were changed on unmapped memory. I'll test this. It's not clear to me yet _why_ that would be done as part of the NX protections though. Might be worth a look at well.

Will try to reproduce this to confirm the theory.

os-d:
> I think I may have missed an obvious issue in the change in the event memory types were changed on unmapped memory. I'll test this. It's not clear to me yet _why_ that would be done as part of the NX protections though. Might be worth a look at well.

The DXE Core tends to have a bad view of memory attributes and very blatantly drops things left and right (which we know as that was a key design point in fixing in Patina :). I'm unfortunately not terribly surprised that this occurred.

I suspect the problem is that the GCD has the wrong caching attribute here (quite possibly 0, which is invalid) and so the NX protection code, which gets the cacheability attribute from the GCD to set, sets the wrong thing, changing the cacheability on unmapped memory (perhaps a guard page or something). Let me know how your testing goes, I can help look into the memory protections changing the caching attribute, if that's what it ends up being.

cfernald:
@rcran-ampere, would it be possible for you to confirm that the commit in this PR resolves your issue?

https://github.com/tianocore/edk2/pull/12698

rcran-ampere:
> @rcran-ampere, would it be possible for you to confirm that the commit in this PR resolves your issue?
> 
> #12698

With that change, it gets further but then hangs.
```
add-symbol-file /home/bcran/src/uefi/Build/Altra1L2T/DEBUG_GCC/AARCH64/ArmPkg/Drivers/CpuDxe/CpuDxe/DEBUG/ArmCpuDxe.dll 0xFA876000
Loading driver at 0x000FA875000 EntryPoint=0x000FA8793FC ArmCpuDxe.efi
InstallProtocolInterface: BC62157E-3E33-4FEC-9920-2D3B36D750DF FA99BD18
ProtectUefiImageCommon - 0xFA99B040
  - 0x00000000FA875000 - 0x000000000000C000
PROGRESS CODE: V03040002 I0
ReplaceTableEntry: splitting block entry with MMU disabled
InstallProtocolInterface: 26BACCB1-6F42-11D4-BCE7-0080C73C8881 FA87D0A0
InstallProtocolInterface: F4560CF6-40EC-4B4A-A192-BF1D57D0B189 FA87D068
MemoryProtectionCpuArchProtocolNotify:
InitializeDxeNxMemoryProtectionPolicy: StackBase = 0x00000000BF9C0000  StackSize = 0x0000000000020000
InitializeDxeNxMemoryProtectionPolicy: applying strict permissions to active memory regions
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: Status: Success, IsArmAttributeCacheable(RegionArmAttributes): 1, IsArmAttributeCacheable(ArmAttributes): 0
CpuSetMemoryAttributes: calling ArmSetMemoryAttributes
CpuSetMemoryAttributes: FlushCache is TRUE.
CpuSetMemoryAttributes: calling ArmDataSynchronizationBarrier
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x0000000090000000, 798883840)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: Status: Success, IsArmAttributeCacheable(RegionArmAttributes): 1, IsArmAttributeCacheable(ArmAttributes): 0
CpuSetMemoryAttributes: calling ArmSetMemoryAttributes
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: Status: Success, IsArmAttributeCacheable(RegionArmAttributes): 1, IsArmAttributeCacheable(ArmAttributes): 0
CpuSetMemoryAttributes: calling ArmSetMemoryAttributes
CpuSetMemoryAttributes: FlushCache is TRUE.
CpuSetMemoryAttributes: calling ArmDataSynchronizationBarrier
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x00000000BFA08000, 192512)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: Status: Success, IsArmAttributeCacheable(RegionArmAttributes): 1, IsArmAttributeCacheable(ArmAttributes): 0
CpuSetMemoryAttributes: calling ArmSetMemoryAttributes
CpuSetMemoryAttributes: FlushCache is TRUE.
CpuSetMemoryAttributes: calling ArmDataSynchronizationBarrier
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x00000000BFA45000, 4853760)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: Status: Success, IsArmAttributeCacheable(RegionArmAttributes): 1, IsArmAttributeCacheable(ArmAttributes): 0
CpuSetMemoryAttributes: calling ArmSetMemoryAttributes
CpuSetMemoryAttributes: FlushCache is TRUE.
CpuSetMemoryAttributes: calling ArmDataSynchronizationBarrier
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x00000000BFF10000, 614400)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: Status: Success, IsArmAttributeCacheable(RegionArmAttributes): 1, IsArmAttributeCacheable(ArmAttributes): 0
CpuSetMemoryAttributes: calling ArmSetMemoryAttributes
CpuSetMemoryAttributes: FlushCache is TRUE.
CpuSetMemoryAttributes: calling ArmDataSynchronizationBarrier
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x00000000BFFB3000, 982261760)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: Status: Success, IsArmAttributeCacheable(RegionArmAttributes): 1, IsArmAttributeCacheable(ArmAttributes): 0
CpuSetMemoryAttributes: calling ArmSetMemoryAttributes
CpuSetMemoryAttributes: FlushCache is TRUE.
CpuSetMemoryAttributes: calling ArmDataSynchronizationBarrier
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x00000000FA881000, 1093632)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: Status: Success, IsArmAttributeCacheable(RegionArmAttributes): 1, IsArmAttributeCacheable(ArmAttributes): 0
CpuSetMemoryAttributes: calling ArmSetMemoryAttributes
CpuSetMemoryAttributes: FlushCache is TRUE.
CpuSetMemoryAttributes: calling ArmDataSynchronizationBarrier
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x00000000FA995000, 86421504)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: Status: Success, IsArmAttributeCacheable(RegionArmAttributes): 1, IsArmAttributeCacheable(ArmAttributes): 0
CpuSetMemoryAttributes: calling ArmSetMemoryAttributes
CpuSetMemoryAttributes: FlushCache is TRUE.
CpuSetMemoryAttributes: calling ArmDataSynchronizationBarrier
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x00000000FFD90000, 2555904)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: Status: Success, IsArmAttributeCacheable(RegionArmAttributes): 1, IsArmAttributeCacheable(ArmAttributes): 0
CpuSetMemoryAttributes: calling ArmSetMemoryAttributes
CpuSetMemoryAttributes: FlushCache is TRUE.
CpuSetMemoryAttributes: calling ArmDataSynchronizationBarrier
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x0000080000000000, 2147483648)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: Status: Success, IsArmAttributeCacheable(RegionArmAttributes): 1, IsArmAttributeCacheable(ArmAttributes): 0
CpuSetMemoryAttributes: calling ArmSetMemoryAttributes
CpuSetMemoryAttributes: FlushCache is TRUE.
CpuSetMemoryAttributes: calling ArmDataSynchronizationBarrier
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x0000080100000000, 1644972474368)
```

cfernald:
@rcran-ampere , Thanks for your help on this. Could you have it print the new and old arm attributes when FlushCache is set to true? That is `RegionArmAttributes`, `ArmAttributes`, and ideally `EfiAttributes`?

rcran-ampere:
@cfernald Here's the updated output. I removed some debug prints since they were no longer needed. 
```
ReplaceTableEntry: splitting block entry with MMU disabled
InstallProtocolInterface: 26BACCB1-6F42-11D4-BCE7-0080C73C8881 FA87D0A0
InstallProtocolInterface: F4560CF6-40EC-4B4A-A192-BF1D57D0B189 FA87D068
MemoryProtectionCpuArchProtocolNotify:
InitializeDxeNxMemoryProtectionPolicy: StackBase = 0x00000000BF9C0000  StackSize = 0x0000000000020000
InitializeDxeNxMemoryProtectionPolicy: applying strict permissions to active memory regions
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: FlushCache=TRUE. RegionArmAttributes: 0x004000000000070C, ArmAttributes: 002000000000041C, EfiAttributes: 0000000000004000
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x0000000090000000, 798883840)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: FlushCache=TRUE. RegionArmAttributes: 0x000000000000070C, ArmAttributes: 002000000000041C, EfiAttributes: 0000000000004000
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x00000000BFA08000, 192512)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: FlushCache=TRUE. RegionArmAttributes: 0x000000000000070C, ArmAttributes: 002000000000041C, EfiAttributes: 0000000000004000
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x00000000BFA45000, 4853760)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: FlushCache=TRUE. RegionArmAttributes: 0x000000000000070C, ArmAttributes: 002000000000041C, EfiAttributes: 0000000000004000
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x00000000BFF10000, 614400)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: FlushCache=TRUE. RegionArmAttributes: 0x000000000000070C, ArmAttributes: 002000000000041C, EfiAttributes: 0000000000004000
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x00000000BFFB3000, 982261760)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: FlushCache=TRUE. RegionArmAttributes: 0x004000000000070C, ArmAttributes: 002000000000041C, EfiAttributes: 0000000000004000
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x00000000FA881000, 1093632)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: FlushCache=TRUE. RegionArmAttributes: 0x004000000000070C, ArmAttributes: 002000000000041C, EfiAttributes: 0000000000004000
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x00000000FA995000, 86421504)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: FlushCache=TRUE. RegionArmAttributes: 0x000000000000070C, ArmAttributes: 002000000000041C, EfiAttributes: 0000000000004000
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x00000000FFD90000, 2555904)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: FlushCache=TRUE. RegionArmAttributes: 0x004000000000070C, ArmAttributes: 002000000000041C, EfiAttributes: 0000000000004000
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x0000080000000000, 2147483648)
CpuSetMemoryAttributes: done flushing cache
CpuSetMemoryAttributes: done
CpuSetMemoryAttributes: enter
CpuSetMemoryAttributes: FlushCache=TRUE. RegionArmAttributes: 0x004000000000070C, ArmAttributes: 002000000000041C, EfiAttributes: 0000000000004000
CpuSetMemoryAttributes: calling WriteBackInvalidateDataCacheRange(0x0000080100000000, 1644972474368)
```

cfernald:
@rcran-ampere , it seems like a 0x17F00000000 byte range is being converted from 0x3 (WRITE_BACK) to index 0x7 (Unused, but should then be device memory I think). This seems like a bug in itself, do you know if the memory at 0x80100000000 is DRAM or MMIO, and if the initial WB is expected?

The 0x7 seems to come from the fact that the caller is not providing any caching attributes, so it defaults to this unused value.

os-d:
> @rcran-ampere , it seems like a 0x17F00000000 byte range is being converted from 0x3 (WRITE_BACK) to index 0x7 (Unused, but should then be device memory I think). This seems like a bug in itself, do you know if the memory at 0x80100000000 is DRAM or MMIO, and if the initial WB is expected?
> 
> The 0x7 seems to come from the fact that the caller is not providing any caching attributes, so it defaults to this unused value.

Presumably this is the case I was mentioning, that has occurred in a number of cases, where the GCD doesn’t have cache attributes set, so when the memory protection code fetches cache attrs from the GCD, there are none. It may be that there is a callsite not ferching the attrs. This is where the memory attribute protocol is safer.

The way it is intended to work is that InitMmu() is called with the correct attrs via the platform lib. Then CpuDxe queries all attrs and updates the GCD with the actual attrs, then memory protections are applied. It appears something broke down in that process, which exposed the bug here. The process would have been broken before this change, of course.

cfernald:
> > @rcran-ampere , it seems like a 0x17F00000000 byte range is being converted from 0x3 (WRITE_BACK) to index 0x7 (Unused, but should then be device memory I think). This seems like a bug in itself, do you know if the memory at 0x80100000000 is DRAM or MMIO, and if the initial WB is expected?
> > The 0x7 seems to come from the fact that the caller is not providing any caching attributes, so it defaults to this unused value.
> 
> Presumably this is the case I was mentioning, that has occurred in a number of cases, where the GCD doesn’t have cache attributes set, so when the memory protection code fetches cache attrs from the GCD, there are none. It may be that there is a callsite not ferching the attrs. This is where the memory attribute protocol is safer.
> 
> The way it is intended to work is that InitMmu() is called with the correct attrs via the platform lib. Then CpuDxe queries all attrs and updates the GCD with the actual attrs, then memory protections are applied. It appears something broke down in that process, which exposed the bug here. The process would have been broken before this change, of course.

I agree. This code could definitely be more resilient. I'm debating if `CpuSetMemoryAttributes` should be made more resilient to this, such that if no caching attribute is provided, then the previous attribute will be maintained. @os-d, what do you think?

os-d:
> I agree. This code could definitely be more resilient. I'm debating if `CpuSetMemoryAttributes` should be made more resilient to this, such that if no caching attribute is provided, then the previous attribute will be maintained. @os-d, what do you think?

I'm hesitant to change the semantics of the CpuArch protocol. The house of cards that is the edk2 memory protections is clearly very delicate and I'm worried that will break more than it fixes.

I am able to repro the memory protections sending 0 as the cache attr for NX protections on top of tree ArmVirtPkg. Let me take a look at untangling that. Obviously Project Mu has a large set of changes for memory protections that are on my docket to upstream. Depending on how far this thread gets pulled we may need more of those to make the system work. And again, even with the Project Mu changes, the overall architecture does not work of splitting page table management out from the core (which leads again into the Patina design decision to own page table management in the core :).

I hope we can get a change that will fix the sending 0 as the cache attr and then put an assert in the code to catch further cases.

os-d:
Well, checking, but this is almost certainly the issue (true on X64 as well):

image

The CpuArch protocol gets installed before the GCD is updated with the page table attributes and the memory protections code runs when the CpuArch protocol is installed. No attributes in the GCD at that point.

cfernald:
@rcran-ampere , I have pushed an updated commit to my new PR that will avoid flushing in incorrect scenarios like this. Could you please confirm this resolves your issue?

rcran-ampere:
> @rcran-ampere , I have pushed an updated commit to my new PR that will avoid flushing in incorrect scenarios like this. Could you please confirm this resolves your issue?

Thanks - that fixed it.

PR #12513, opened 2026-04-28
ShellPkg/UefiShellLevel3: Code cleanup/refactor

pierregondois:
# Description

Small cleanup/refactor of the UefiShellLevel3 commands.
No functional change should be induced by these patches.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

### DIFF
--- a/ShellPkg/Library/UefiShellLevel3CommandsLib/Echo.c
+++ b/ShellPkg/Library/UefiShellLevel3CommandsLib/Echo.c
@@ -17,53 +17,24 @@ STATIC CONST SHELL_PARAM_ITEM  ParamList[] = {
   { NULL,    TypeMax  }
 };
 
-/**
-  Function for 'echo' command.
+/** Main function of the 'REcho' command.
 
-  @param[in] ImageHandle  Handle to the Image (NULL if Internal).
-  @param[in] SystemTable  Pointer to the System Table (NULL if Internal).
+  @param[in] Package    List of input parameter for the command.
 **/
+STATIC
 SHELL_STATUS
-EFIAPI
-ShellCommandRunEcho (
-  IN EFI_HANDLE        ImageHandle,
-  IN EFI_SYSTEM_TABLE  *SystemTable
+MainCmdREcho (
+  LIST_ENTRY  *Package
   )
 {
-  EFI_STATUS    Status;
-  LIST_ENTRY    *Package;
   SHELL_STATUS  ShellStatus;
   UINTN         ParamCount;
-  CHAR16        *ProblemParam;
   UINTN         Size;
   CHAR16        *PrintString;
 
-  Size         = 0;
-  ProblemParam = NULL;
-  PrintString  = NULL;
-  ShellStatus  = SHELL_SUCCESS;
-
-  //
-  // initialize the shell lib (we must be in non-auto-init...)
-  //
-  Status = ShellInitialize ();
-  ASSERT_EFI_ERROR (Status);
-
-  //
-  // parse the command line
-  //
-  Status = ShellCommandLineParseEx (ParamList, &Package, &ProblemParam, TRUE, TRUE);
-  if (EFI_ERROR (Status)) {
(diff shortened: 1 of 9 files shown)


ShellPkg/Library/UefiShellLevel3CommandsLib/Echo.c
leiflindholm: This looks like another one of those default init blocks that have been added in the past to get compilers to stop complaining about fragile code. I bet we could drop all three of these, and get rid of the ShellStatus variable as well.

Actually, I stand corrected, Size and PrintString need initializing - but maybe we can move it down to the init condition for the for loop where this matters (new line 74)?

ShellPkg/Library/UefiShellLevel3CommandsLib/GetMtc.c
```diff
@@ -32,28 +32,29 @@ MainCmdMtc (
...
-    //
-    if (ShellStatus == SHELL_SUCCESS) {
-      ShellPrintHiiDefaultEx (STRING_TOKEN (STR_GET_MTC_OUTPUT), gShellLevel3HiiHandle, Mtc);
-    }
+  //
+  // Get the monotonic counter count
+  //
+  Status = gBS->GetNextMonotonicCount (&Mtc);
+  if (Status == EFI_DEVICE_ERROR) {
+    ShellStatus = SHELL_DEVICE_ERROR;
```
leiflindholm: The ShellStatus variable is pretty pointless in this function too.
These could all return directly. In fact if we had an EfiStatusToShellStatus helper, we could tidy this up further.

ShellPkg/Library/UefiShellLevel3CommandsLib/Help.c
```diff
@@ -453,14 +454,13 @@ ShellCommandRunHelp (
...
       }
-
-      //
-      // free the command line package
-      //
-      ShellCommandLineFreeVarList (Package);
     }
   }
 
+  if (Package != NULL) {
```
leiflindholm: This is clearly a useful bugfix, but:
could we move this right to the end, with a label attached to it, and use a few well placed gotos to nuke some levels of indentation while getting rid of success handling?

ShellPkg/Library/UefiShellLevel3CommandsLib/Echo.c
pierregondois: Yes sure, should be done now

ShellPkg/Library/UefiShellLevel3CommandsLib/GetMtc.c
pierregondois: Yes sure, `ShellStatus` can only be `SHELL_SUCCESS` when reaching the print statement, so this also simplifies things a bit.

ShellPkg/Library/UefiShellLevel3CommandsLib/Help.c
pierregondois: I forgot to handle the `Help` command.
I removed the patch which was specific to the `Help` command and did the same operations.

leiflindholm: And that made the previous bug fall out in the wash?

pierregondois: yes and no, this is is now done in:
```ShellPkg/UefiShellLevel3: Extract MainCmdXXX() function```::```ShellPkg/Library/UefiShellLevel3CommandsLib/Help.c```.

Effectively the patch states there is no functional change. Should I extract it in a separate patch ?

pierregondois: I added back the patch `ShellPkg/Help: Free package on all paths`

leiflindholm: I don't think that's necessary, but it would be worth changing the commit message to point out that while not intending functional changes, the refactoring as a side effect ends up removing a memory leak in said function.

leiflindholm (approved):

PR #12519, opened 2026-04-29
CryptoPkg: Add Pkcs7Encrypt and Pkcs7Decrypt

Flickdm:
# Description

This is adding PKCS7Encrypt(..) which is used by OfflineCrashDump to encrypt the information being passed back to the Operating System.

Real implementation of Pkcs7Encrypt was done by @idigdoug and I'm simply upstreaming.

- [ ] Breaking change?
  - No
- [ ] Impacts security?
  - No
- [x] Includes tests?
-  Unit tests

## How This Was Tested

Unit Tests and Used in Production

## Integration Instructions

N/A


Flickdm:
It may be worth having a conversation about implementing Pkcs7Decrypt if anything for testing / validation. Openssl would be trivial effectively wrapping PKCS7_verify. If we wanted to support mbedtls that would be more problematic. 

@idigdoug do you have thoughts?

idigdoug:
> It may be worth having a conversation about implementing Pkcs7Decrypt if anything for testing / validation. Openssl would be trivial effectively wrapping PKCS7_verify. If we wanted to support mbedtls that would be more problematic.
> 
> @idigdoug do you have thoughts?

I would be happy to provide an implementation of Pkcs7Decrypt if it seems useful. For OfflineCrashDump, Pkcs7Decrypt would only be used for testing, not in production, so it would be an open question for whether it is worth any binary size increase. Or maybe it's disabled by default?

liyi77:
How about to use CMS?

idigdoug:
> How about to use CMS?

I could switch the OpenSSL implementation to use the CMS API, but I think right now the CMS support isn't enabled by EDK2's default build of OpenSSL. Once EDK2 starts transitioning to enable CMS support, this should definitely be updated to use the CMS API instead -- that would be great.

The MbedTLS implementation is logically CMS-compatible, though it uses the older hashing option. For consistency of behavior between MbedTLS and OpenSSL, the newer hashing in the MbedTLS implemetation should probably be enabled at the same time as we switch the OpenSSL implementation to use CMS.

liyi77:
Yes I just realize that we are currently using the no-cms config, then I am OK to use pkcs7.

jyao1 (approved):

jyao1:
> It may be worth having a conversation about implementing Pkcs7Decrypt if anything for testing / validation. Openssl would be trivial effectively wrapping PKCS7_verify. If we wanted to support mbedtls that would be more problematic.
> 
> @idigdoug do you have thoughts?

Usually we always add Encryption together with Decryption. I prefer to add both Pkcs7Encrypt / Pkcs7Decrypt.
If there is no usage, the platform can disable it via PCD.

I do not think mbedtls would be a concern. It is legal that one API is only supported by one crypto_lib. It is OK to return unsupported in mbedtls.

idigdoug:
> Usually we always add Encryption together with Decryption. I prefer to add both Pkcs7Encrypt / Pkcs7Decrypt. If there is no usage, the platform can disable it via PCD.

I'll work with Doug Flick to implement a Pkcs7Decrypt function. Should that go in a separate PR?

idigdoug:
@liyi77 would you be able to take a look at this? Thanks!

liyi77:
> @liyi77 would you be able to take a look at this? Thanks!

Hi could you implement  Pkcs7Encrypt / Pkcs7Decrypt in same PR?

idigdoug:
> > @liyi77 would you be able to take a look at this? Thanks!
> 
> Hi could you implement Pkcs7Encrypt / Pkcs7Decrypt in same PR?

Done.

liyi77 (approved):
LGTM

idigdoug:
> > It may be worth having a conversation about implementing Pkcs7Decrypt if anything for testing / validation. Openssl would be trivial effectively wrapping PKCS7_verify. If we wanted to support mbedtls that would be more problematic.
> > @idigdoug do you have thoughts?
> 
> Usually we always add Encryption together with Decryption. I prefer to add both Pkcs7Encrypt / Pkcs7Decrypt. If there is no usage, the platform can disable it via PCD.
> 
> I do not think mbedtls would be a concern. It is legal that one API is only supported by one crypto_lib. It is OK to return unsupported in mbedtls.

@jyao1 I've added Decryption support. Let me know if it looks ok to you. Thanks!

liyi77:
@mergify rebase

liyi77:
CI build is still failing; please try manually rebasing or creating a new PR.

idigdoug:
We're hitting https://github.com/tianocore/edk2/issues/13009

liyi77:
@mergify rebase

Issue #12517, opened 2026-04-29
[Bug]: Base.h IA32 GCC _AlignOf


apop5:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

MdePkg

### Which targets are impacted by this bug?

RELEASE, DEBUG

### Current Behavior

Base.h defines static asserts for checking that alignment of variables matches size (i.e. UINT64 is aligned on a 8 byte boundary)
[edk2/MdePkg/Include/Base.h at master · tianocore/edk2](https://github.com/tianocore/edk2/blob/master/MdePkg/Include/Base.h#L836)
 
When compiling for i386 for SysV ABI, the _Alignof(UINT64) is returning 4 instead of the static asserts requirement of 8. This cases the STATIC_ASSERTS for UINT64/INT64.
 
This was 

_Alignof is a C11 operator 

### Expected Behavior

the static asserts should not trigger for ia32 builds.

### Steps To Reproduce

This was reproduced building a platform with IA32 PEI phase. 

This can be reproduced using gdbolt compiler explorer and the gcc flags -m32, examine the output of various gcc compiler versions

```c
#include <stdio.h>

int main()
{
    printf("Alifnof : %d\n", _Alignof(long long));
    printf("__alignof__ : %d\n", __alignof__(long long));
    printf("align : %d\n", sizeof(long long));
    return 0;
}
```

4
8
8

### Build Environment

```markdown
- OS(s): Ubuntu
- Tool Chain(s): GCC (version 12)
```

### Version Information

```text
commit: edk2-stable202511
```

### Urgency

Low

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

No maintainer feedback needed

### Anything else?

__alignof__ is a gcc extension that predates C11. 

Using __alignof__ passes the static asserts.

apop5:
Further investigation:

.aslc source file being compiled under IA32 using GCC would fail the static asserts

Tracked back to IA32_ASLCC_FLAGS no longer containing the `-malign-double` CC flag. This was lost during the large changes sets to tools_def.template.

Adding the -malign-double results in expected behavior in compiler explorer, and in building physical platforms

apop5:
Fixed in #12520

PR #12520, opened 2026-04-29
BaseTools/Conf: Add -malign-double to IA32 ASLCC flags

apop5:
# Description

When compiling an IA32 .aslc source file that includes Base.h, the GCC
static asserts for fundamental type sizes fail because 64-bit types such
as UINT64 are only 4-byte aligned by default in 32-bit mode. Adding
-malign-double causes the compiler to align 64-bit values on 8-byte
boundaries, matching the alignment assumed by the static asserts.

This fix is applied to the GCC, GCCNOLTO, CLANGPDB, and CLANGDWARF
toolchain IA32 ASLCC_FLAGS entries.

REF: https://github.com/tianocore/edk2/issues/12517

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Older platform with IA32 PEI stage started failing static asserts in GCC 12.x
Changes pass compilation on platform.

## Integration Instructions
N/A


MdePkg/Include/Base.h
```diff
@@ -769,10 +769,14 @@ typedef UINTN *BASE_LIST;
 // Standard C++ operator.
 //
 #define ALIGNOF(TYPE)  alignof (TYPE)
-#elif defined (__GNUC__) || defined (__clang__) || (defined (_MSC_VER) && _MSC_VER >= 1900)
+#elif defined (__GNUC__) || defined (__clang__)
```
mdkinney: _AlignOf is define in C11.  WOuld prefer _AlignOf to be used for GCC and Clang if compiler is C11 or higher.  Please update to only use __alignof__ for GCC and Clang if compiler is configured below C11.

apop5: @mdkinney 
I researched with the compiler explorer more and `-malign-double` is important.
The build failure is coming from a .aslc source file which included base.h

Between all the changes to tools_def.template, `-malign-double` has been dropped for IA32_ASLCC_FLAGS, which is resulting in the static_assert being tripped.

I'll change the PR to explicitly, for IA32_ASLCC_FLAGS have `-malign-double` appended to it.

mdkinney (approved):

makubacki:
`push` label added with approval present and the stable tag freeze completed.

PR #12529, opened 2026-04-30
BaseTools/Capsule: Prevent to Read the STDOUT Content as Signature

linjaso1:
# Description

- Within the capsule generate script, it is using the STDOUT result as signature while signing the hash digest via OpenSSL tool.

- There would have incorrect result when the user terminal have the output when executing the startup script.

- Incorrect the content of signature would make the verification failed.

- Use the ```-output``` flag to export the signature then read it back as the resolution.

- [ ] Breaking change?
  - No, this is bug-fixes without any feature impact.
- [ ] Impacts security?
  - No, this change do not include any security fix.
- [X] Includes tests?
  - Test on the platform which enabled the capsule update feature.

## How This Was Tested

- Tried to generate the capsule and tested on the correspond platform

## Integration Instructions

- N/A

### DIFF
--- a/BaseTools/Source/Python/Capsule/GenerateCapsule.py
+++ b/BaseTools/Source/Python/Capsule/GenerateCapsule.py
@@ -10,7 +10,7 @@
 # keep the tool as simple as possible, it has the following limitations:
 #   * Do not support vendor code bytes in a capsule.
 #
-# Copyright (c) 2018 - 2024, Intel Corporation. All rights reserved.<BR>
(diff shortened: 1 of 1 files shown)


linjaso1:
Hi @gapalomi / @lgao4 / @mikebeaton / @ponchofigueroa / @YuweiChen1110 ,

Could you help to review this PR?

Thanks.

mikebeaton (approved):

lgao4 (approved):

linjaso1:
Hi @lgao4 ,

Could you help to add ```push``` tag to merge this PR?

Thanks.

PR #12530, opened 2026-04-30
NetworkPkg: Fix Coverity static analysis issues

AbuthahirM11:
This PR fixes Coverity static analysis findings in NetworkPkg modules.

**NetworkPkg/Dhcp6Dxe/Dhcp6Utility.c**
Replace the ASSERT (Instance->Config != NULL) in Dhcp6UpdateIaInfo with
an explicit NULL check that returns EFI_DEVICE_ERROR. Also add a
complementary guard in Dhcp6GenerateIaCb alongside the existing check for
Instance->IaCb.Ia. These are defensive changes to prevent potential NULL
pointer dereferences.

**NetworkPkg/Ip6Dxe/Ip6Driver.c**
In Ip6CleanService, Ip6FreeNeighborEntry may attempt to send packets via
MNP. Defensively move the neighbor table cleanup to occur before MNP teardown
so that the MNP child handle and its resources remain valid during neighbor
entry cleanup.

**NetworkPkg/Ip6Dxe/Ip6Nd.c**
In `Ip6ProcessRouterAdvertise`, `IpSb->InterfaceId` is dereferenced in
`CopyMem` to form a stateless address. Add a guard to ensure
`IpSb->InterfaceId` is not NULL before entering the block.

**NetworkPkg/UefiPxeBcDxe/PxeBcDhcp6.c**
When caching the DHCPv6 discover packet to `Mode->DhcpDiscover` in
`PxeBcDhcp6CallBack`, the destination was incorrectly specified as
`Mode->DhcpDiscover.Dhcpv4` (the DHCPv4 union member). Change it to
`Mode->DhcpDiscover` to correctly reference the union.

Cc: Mike Beaton <mjsbeaton@gmail.com>

Signed-off-by: Abuthahir M <abuthahirm@ami.com>

### DIFF
--- a/NetworkPkg/Dhcp6Dxe/Dhcp6Utility.c
+++ b/NetworkPkg/Dhcp6Dxe/Dhcp6Utility.c
@@ -1335,7 +1335,7 @@ Dhcp6GenerateIaCb (
   UINT32        IaSize;
   EFI_DHCP6_IA  *Ia;
 
-  if (Instance->IaCb.Ia == NULL) {
+  if ((Instance->IaCb.Ia == NULL) || (Instance->Config == NULL)) {
     return EFI_DEVICE_ERROR;
   }
 
--- a/NetworkPkg/Ip6Dxe/Ip6Driver.c
+++ b/NetworkPkg/Ip6Dxe/Ip6Driver.c
@@ -190,6 +190,14 @@ Ip6CleanService (
   Ip6CleanPrefixListTable (IpSb, &IpSb->OnlinkPrefix);
   Ip6CleanPrefixListTable (IpSb, &IpSb->AutonomousPrefix);
 
+  //
+  // Free the Neighbor Discovery resources before MNP teardown.
+  //
+  while (!IsListEmpty (&IpSb->NeighborTable)) {
+    NeighborCache = NET_LIST_HEAD (&IpSb->NeighborTable, IP6_NEIGHBOR_ENTRY, Link);
+    Ip6FreeNeighborEntry (IpSb, NeighborCache, FALSE, TRUE, EFI_SUCCESS, NULL, NULL);
+  }
+
   if (IpSb->RouteTable != NULL) {
     Ip6CleanRouteTable (IpSb->RouteTable);
     IpSb->RouteTable = NULL;
@@ -229,14 +237,6 @@ Ip6CleanService (
 
   if (IpSb->RecvRequest.MnpToken.Event != NULL) {
     gBS->CloseEvent (IpSb->RecvRequest.MnpToken.Event);
-  }
-
-  //
-  // Free the Neighbor Discovery resources
-  //
-  while (!IsListEmpty (&IpSb->NeighborTable)) {
-    NeighborCache = NET_LIST_HEAD (&IpSb->NeighborTable, IP6_NEIGHBOR_ENTRY, Link);
-    Ip6FreeNeighborEntry (IpSb, NeighborCache, FALSE, TRUE, EFI_SUCCESS, NULL, NULL);
   }
 
   return EFI_SUCCESS;
--- a/NetworkPkg/UefiPxeBcDxe/PxeBcDhcp6.c
+++ b/NetworkPkg/UefiPxeBcDxe/PxeBcDhcp6.c
@@ -2053,7 +2053,7 @@ PxeBcDhcp6CallBack (
       //
       // Cache the dhcp discover packet to mode data directly.
       //
-      CopyMem (&Mode->DhcpDiscover.Dhcpv4, &Packet->Dhcp6, Packet->Length);
+      CopyMem (&Mode->DhcpDiscover.Dhcpv6, &Packet->Dhcp6, Packet->Length);
       break;
 
     case Dhcp6RcvdAdvertise:
--- a/NetworkPkg/Dhcp6Dxe/Dhcp6Io.c
+++ b/NetworkPkg/Dhcp6Dxe/Dhcp6Io.c
@@ -538,7 +538,9 @@ Dhcp6UpdateIaInfo (
   T1 = 0;
   T2 = 0;
 
-  ASSERT (Instance->Config != NULL);
+  if (Instance->Config == NULL) {
+    return EFI_DEVICE_ERROR;
+  }
 
   // OptionLen is the length of the Options excluding the DHCP header.
   // Length of the EFI_DHCP6_PACKET from the first byte of the Header field to the last
--- a/NetworkPkg/Ip6Dxe/Ip6Nd.c
+++ b/NetworkPkg/Ip6Dxe/Ip6Nd.c
@@ -2277,6 +2277,7 @@ Ip6ProcessRouterAdvertise (
           //
           if ((PrefixList == NULL) &&
               (PrefixOption.ValidLifetime != 0) &&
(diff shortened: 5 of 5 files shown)


NetworkPkg/UefiPxeBcDxe/PxeBcDhcp6.c
```diff
@@ -2054,7 +2054,7 @@ PxeBcDhcp6CallBack (
       //
       // Cache the dhcp discover packet to mode data directly.
       //
-      CopyMem (&Mode->DhcpDiscover.Dhcpv4, &Packet->Dhcp6, Packet->Length);
+      CopyMem (&Mode->DhcpDiscover, &Packet->Dhcp6, Packet->Length);
```
mikebeaton: Wouldn't this more logically be:
```
CopyMem (&Mode->DhcpDiscover.Dhcpv6, &Packet->Dhcp6, Packet->Length);
```
?

NetworkPkg/Ip6Dxe/Ip6Driver.c
```diff
@@ -190,6 +190,14 @@ Ip6CleanService (
...
   Ip6CleanPrefixListTable (IpSb, &IpSb->AutonomousPrefix);
 
+  //
+  // Free the Neighbor Discovery resources before MNP teardown.
+  //
+  while (!IsListEmpty (&IpSb->NeighborTable)) {
+    NeighborCache = NET_LIST_HEAD (&IpSb->NeighborTable, IP6_NEIGHBOR_ENTRY, Link);
+    Ip6FreeNeighborEntry (IpSb, NeighborCache, FALSE, TRUE, EFI_SUCCESS, NULL, NULL);
+  }
+
```
mikebeaton, line 200: What was the actual Coverity issue here? I can't spot it.

AbuthahirM11: The Coverity defect is FORWARD_NULL. In the original ordering, IpSb->Mnp was explicitly set to NULL during MNP teardown, but Ip6FreeNeighborEntry was called afterward. That function can internally attempt to send NDP packets via IpSb->Mnp, which at that point is NULL. Coverity tracked the explicit IpSb->Mnp = NULL assignment and flagged the subsequent dereference through Ip6FreeNeighborEntry as a forward null. The fix moves the neighbor table cleanup before MNP teardown so IpSb->Mnp is still valid when Ip6FreeNeighborEntry executes.

NetworkPkg/UefiPxeBcDxe/PxeBcDhcp6.c
AbuthahirM11: You're right. Using &Mode->DhcpDiscover.Dhcpv6 is more precise and consistent with the rest of the file (e.g., Mode->DhcpAck.Dhcpv6 and Mode->ProxyOffer.Dhcpv6). Updated in the latest commit.

NetworkPkg/Ip6Dxe/Ip6Driver.c
mikebeaton: Thank you - resolved. I cannot mark resolved (as no write access), but I believe you can (as PR author).

NetworkPkg/UefiPxeBcDxe/PxeBcDhcp6.c
mikebeaton: Resolved

mikebeaton (approved):
I believe these changes all look good.

AbuthahirM11:
@leiflindholm @ajfish @ardbiesheuvel @makubacki @mdkinney @Zclarkwilliams — Could you please take a look and approve this PR when you get a chance?

NetworkPkg/Dhcp6Dxe/Dhcp6Utility.c
makubacki, line 1: The current commit for this change is: [NetworkPkg/Dhcp6Dxe: Fix NULL pointer dereference in Dhcp6GenerateIaCb](https://github.com/tianocore/edk2/pull/12530/changes/700e6e802cdbc3ebe2110b1498e87e98b1a5f5af)

This is not "fixing a NULL pointer dereference". It is **preventing** a potential NULL pointer dereference.

Also, this function (`Dhcp6GenerateIaCb()`) only has one caller - `Dhcp6UpdateIaInfo()` and that function references `Instance->Config` shortly before calling `Dhcp6GenerateIaCb()`:

https://github.com/tianocore/edk2/blob/b03a21a63e3bd001f52c527e5a57feddb53a690b/NetworkPkg/Dhcp6Dxe/Dhcp6Io.c#L635-L646

Furthermore, that function has an `ASSERT()`:

https://github.com/tianocore/edk2/blob/b03a21a63e3bd001f52c527e5a57feddb53a690b/NetworkPkg/Dhcp6Dxe/Dhcp6Io.c#L523-L541

So, afaict, this doesn't have an effective impact. If you would like to keep a change, I suggest the above `ASSERT (Instance->Config != NULL);` return an error instead and update the commit message to accurately reflect this is a defensive change.

NetworkPkg/Ip6Dxe/Ip6Driver.c
makubacki, line 1: The commit message for [NetworkPkg/Ip6Dxe: Move neighbor table cleanup before MNP teardown](https://github.com/tianocore/edk2/pull/12530/changes/3301c5e232334be981f5ad691b8cc4a60ee57b87) says:

> In Ip6CleanService, Ip6FreeNeighborEntry may send packets via MNP.
If the MNP child handle is destroyed before the neighbor table is
cleared, this results in a use-after-free. Move the neighbor table
cleanup to occur before MNP teardown to fix this ordering issue.

For the claim "this results in use-after-free" to be made, the commit message should describe viable call paths in which a use-after-free is actually possible. I'm not saying that is not possible, but it needs to be explained. Otherwise, this is a defensive measure by simply moving resource cleanup before tearing down the subsystem it depends on.

NetworkPkg/Dhcp6Dxe/Dhcp6Utility.c
AbuthahirM11: @makubacki You are correct that the original description was inaccurate. I have updated the change to replace the ASSERT (Instance->Config != NULL) in Dhcp6UpdateIaInfo with an explicit if (Instance->Config == NULL) { return EFI_DEVICE_ERROR; }, and added a complementary guard in Dhcp6GenerateIaCb as well. The commit message now accurately describes these as defensive changes to prevent potential NULL pointer dereferences.

NetworkPkg/Ip6Dxe/Ip6Driver.c
AbuthahirM11: @makubacki I have updated the commit message to reflect this as a defensive measure — moving neighbor table cleanup before MNP teardown to ensure MNP resources remain valid — rather than claiming a confirmed use-after-free.

makubacki: Thanks, lgtm now.

AbuthahirM11:
@leiflindholm @ajfish @ardbiesheuvel @makubacki @mdkinney @Zclarkwilliams — Could you please take a look and approve this PR when you get a chance?

makubacki:
@mergifyio refresh

makubacki (approved):

PR #12536, opened 2026-05-01
ArmPkg, StandaloneMmPkg: optimise memory permission change

LeviYeoReum:
# Description

In StandaloneMm, the allowed memory permission combinations are ROX, RO,
 and RW.

At the initial stage, all regions are mapped as ROX. When drivers are loaded for relocation, the image is first mapped as RW.

However, the current mechanism introduces significant overhead when changing memory permissions. For example,
when changing permissions from ROX to RW, the process is performed in two steps:

  1. Clear the execute (X) permission (ROX → RO)
  2. Add the write (W) permission (RO → RW)

To execute step (1), the system first retrieves the permissions of the target region. This may trigger multiple SMC calls, depending on the permissions of the pages within that region.

This permission retrieval operation is redundant, as there is no need to maintain an intermediate state. Nevertheless,
it can cause additional SMC calls and coherence operations from the SPMC (e.g., TLB flushes).

Therefore, this redundant operation should be removed, and in most cases, the memory permission should be updated with a single “set memory permission” operation.

- [ ] Breaking change?
  - N/A
- [ ] Impacts security?
  - N/A
- [ ] Includes tests?
  - N/A

## How This Was Tested

Booting test on FVP_RevC with StandaloneMm.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Include/Library/StandaloneMmMmuLib.h
+++ b/ArmPkg/Include/Library/StandaloneMmMmuLib.h
@@ -8,24 +8,96 @@
 
 #pragma once
 
+/**
+  Set XN bit preserving other permission.
+
+  @param [in]  BaseAddress     Base address for the memory region.
+  @param [in]  Length          Length of the memory region.
+
+  @retval EFI_SUCCESS             Request successfull.
+  @retval EFI_INVALID_PARAMETER   A parameter is invalid.
+  @retval EFI_NOT_READY           Callee is busy or not in a state to handle
+                                  this request.
+  @retval EFI_UNSUPPORTED         This function is not implemented by the
+                                  callee.
+  @retval EFI_ABORTED             Message target ran into an unexpected error
+                                  and has aborted.
+  @retval EFI_ACCESS_DENIED       Access denied.
+  @retval EFI_OUT_OF_RESOURCES    Out of memory to perform operation.
+
+**/
 EFI_STATUS
 ArmSetMemoryRegionNoExec (
   IN  EFI_PHYSICAL_ADDRESS  BaseAddress,
   IN  UINT64                Length
   );
 
+/**
+  Clear XN bit preserving other permission.
+
+  @param [in]  BaseAddress     Base address for the memory region.
+  @param [in]  Length          Length of the memory region.
+
+  @retval EFI_SUCCESS             Request successfull.
+  @retval EFI_INVALID_PARAMETER   A parameter is invalid.
+  @retval EFI_NOT_READY           Callee is busy or not in a state to handle
+                                  this request.
+  @retval EFI_UNSUPPORTED         This function is not implemented by the
+                                  callee.
+  @retval EFI_ABORTED             Message target ran into an unexpected error
+                                  and has aborted.
+  @retval EFI_ACCESS_DENIED       Access denied.
+  @retval EFI_OUT_OF_RESOURCES    Out of memory to perform operation.
+
+**/
 EFI_STATUS
 ArmClearMemoryRegionNoExec (
   IN  EFI_PHYSICAL_ADDRESS  BaseAddress,
   IN  UINT64                Length
   );
 
+/**
+  Set the memory to read-only while preserving execute permission.
+
+  @param [in]  BaseAddress     Base address for the memory region.
+  @param [in]  Length          Length of the memory region.
+
+  @retval EFI_SUCCESS             Request successfull.
+  @retval EFI_INVALID_PARAMETER   A parameter is invalid.
+  @retval EFI_NOT_READY           Callee is busy or not in a state to handle
+                                  this request.
+  @retval EFI_UNSUPPORTED         This function is not implemented by the
+                                  callee.
+  @retval EFI_ABORTED             Message target ran into an unexpected error
+                                  and has aborted.
+  @retval EFI_ACCESS_DENIED       Access denied.
+  @retval EFI_OUT_OF_RESOURCES    Out of memory to perform operation.
+
+**/
 EFI_STATUS
 ArmSetMemoryRegionReadOnly (
   IN  EFI_PHYSICAL_ADDRESS  BaseAddress,
   IN  UINT64                Length
   );
 
+/**
+  Set the memory to read-write while preserving execute permission.
+
+  @param [in]  BaseAddress     Base address for the memory region.
+  @param [in]  Length          Length of the memory region.
+
+  @retval EFI_SUCCESS             Request successfull.
+  @retval EFI_INVALID_PARAMETER   A parameter is invalid.
+  @retval EFI_NOT_READY           Callee is busy or not in a state to handle
+                                  this request.
+  @retval EFI_UNSUPPORTED         This function is not implemented by the
+                                  callee.
+  @retval EFI_ABORTED             Message target ran into an unexpected error
+                                  and has aborted.
+  @retval EFI_ACCESS_DENIED       Access denied.
+  @retval EFI_OUT_OF_RESOURCES    Out of memory to perform operation.
+
+**/
 EFI_STATUS
 ArmClearMemoryRegionReadOnly (
   IN  EFI_PHYSICAL_ADDRESS  BaseAddress,
--- a/ArmPkg/Include/Library/StandaloneMmMmuLib.h
+++ b/ArmPkg/Include/Library/StandaloneMmMmuLib.h
@@ -57,7 +57,7 @@ ArmClearMemoryRegionNoExec (
   );
 
 /**
-  Set the memory to read-only while preserving execute permission.
+  Change memory permission as RO ignoring former permission.
 
   @param [in]  BaseAddress     Base address for the memory region.
   @param [in]  Length          Length of the memory region.
@@ -75,13 +75,13 @@ ArmClearMemoryRegionNoExec (
 
 **/
 EFI_STATUS
-ArmSetMemoryRegionReadOnly (
+ArmSetMemoryRegionReadOnlyPerm (
   IN  EFI_PHYSICAL_ADDRESS  BaseAddress,
   IN  UINT64                Length
   );
 
 /**
-  Set the memory to read-write while preserving execute permission.
+  Change memory permission as RW ignoring former permission.
 
   @param [in]  BaseAddress     Base address for the memory region.
   @param [in]  Length          Length of the memory region.
@@ -99,7 +99,31 @@ ArmSetMemoryRegionReadOnly (
 
 **/
 EFI_STATUS
-ArmClearMemoryRegionReadOnly (
+ArmSetMemoryRegionReadWritePerm (
+  IN  EFI_PHYSICAL_ADDRESS  BaseAddress,
+  IN  UINT64                Length
+  );
+
+/**
+  Change memory permission as ROX ignoring former permission.
+
+  @param [in]  BaseAddress     Base address for the memory region.
+  @param [in]  Length          Length of the memory region.
+
+  @retval EFI_SUCCESS             Request successfull.
+  @retval EFI_INVALID_PARAMETER   A parameter is invalid.
+  @retval EFI_NOT_READY           Callee is busy or not in a state to handle
+                                  this request.
+  @retval EFI_UNSUPPORTED         This function is not implemented by the
+                                  callee.
+  @retval EFI_ABORTED             Message target ran into an unexpected error
+                                  and has aborted.
+  @retval EFI_ACCESS_DENIED       Access denied.
+  @retval EFI_OUT_OF_RESOURCES    Out of memory to perform operation.
+
+**/
+EFI_STATUS
+ArmSetMemoryRegionReadOnlyExecPerm (
   IN  EFI_PHYSICAL_ADDRESS  BaseAddress,
   IN  UINT64                Length
   );
--- a/ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
+++ b/ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
@@ -45,19 +45,24 @@ STATIC
 BOOLEAN
 EFIAPI
 IsFfaMemoryAbiSupported (
-  OUT UINT32  *Version
+  OUT UINT32  *Version OPTIONAL
   )
 {
   EFI_STATUS  Status;
+  UINT32      CurrentVersion;
 
-  *Version = 0;
+  CurrentVersion = 0;
 
   Status = ArmFfaLibGetVersion (
              ARM_FFA_CREATE_VERSION (ARM_FFA_MAJOR_VERSION, ARM_FFA_MINOR_VERSION),
-             Version
+             &CurrentVersion
              );
   if (EFI_ERROR (Status)) {
     return FALSE;
+  }
+
+  if (Version != NULL) {
+    *Version = CurrentVersion;
   }
 
   return TRUE;
--- a/ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
+++ b/ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
@@ -277,6 +277,24 @@ RequestMemoryPermissionChange (
   return SendMemoryPermissionRequest (UseFfaAbis, &SvcArgs, &Ret);
 }
 
+/**
+  Set XN bit preserving other permission.
+
+  @param [in]  BaseAddress     Base address for the memory region.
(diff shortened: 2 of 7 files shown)


ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
```diff
@@ -28,6 +28,9 @@
 #include <Library/DebugLib.h>
 #include <Library/PcdLib.h>
 
+#define EFI_PAGE_SIZE_ALIGN_UP(Size) \
```
leiflindholm: With an EFI prefix, this would probably be better added to UefiBaseType.h in a preceding commit. And it would be a useful addition there.
Alternatively, drop the prefix.

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
```diff
@@ -50,7 +53,9 @@ IsFfaMemoryAbiSupported (
 {
   EFI_STATUS  Status;
 
-  *Version = 0;
+  if (Version != NULL) {
```
leiflindholm: Unrelated to rest of change, please break out separately.

Also, *Version is not `OPTIONAL`, either for this function or for `ArmFfaLibGetVersion`, so if you add the test, `Version` being `NULL` should be treated as an error.

ArmPkg/Include/Library/StandaloneMmMmuLib.h
```diff
@@ -8,26 +8,122 @@
 
 #pragma once
 
+/**
```
leiflindholm, line 11: The additions of the comment headers for existing functions is good, but unrelated to the functional change. Please break out into separate commit.

leiflindholm (changes requested):

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
LeviYeoReum: Okay. I'll add this into ` UefiBaseType.h`

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
LeviYeoReum: Thanks to point out this. I'll add this with separated patch.

ArmPkg/Include/Library/StandaloneMmMmuLib.h
LeviYeoReum: Okay. I'll separate it. Thanks!

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
LeviYeoReum: I found we can use just `ALIGN_VALUE`. So I'll remove this. Thanks!

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
```diff
@@ -400,130 +440,167 @@ ArmClearMemoryRegionNoExec (
...
-               Length,
-               &MemoryAttributes,
-               &PageCount
-               );
-    if (EFI_ERROR (Status)) {
-      break;
-    }
+  if (Length == 0) {
+    /// Nothing to do.
+    return EFI_SUCCESS;
```
kuqin12, line 492: Is this true? Or would it better be invalid_parameter?

ArmPkg/Include/Library/StandaloneMmMmuLib.h
```diff
@@ -8,26 +8,122 @@
...
+  @retval EFI_UNSUPPORTED         This function is not implemented by the
+                                  callee.
+  @retval EFI_ABORTED             Message target ran into an unexpected error
+                                  and has aborted.
+  @retval EFI_ACCESS_DENIED       Access denied.
+  @retval EFI_OUT_OF_RESOURCES    Out of memory to perform operation.
+
+**/
 EFI_STATUS
 ArmClearMemoryRegionReadOnly (
```
kuqin12: The other function changes are good, but the only concern is this one: it also removes the other code permissions. That makes the functionality not suitable for this function name anymore. Would it be better to name it as "ArmSetMemoryRegionReadWrite"?

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
LeviYeoReum: Since `SectionHeader.Misc.VirtualSize` could be 0 Legnth -- like .reloc section: 

```
[8001 0] UpdateMmFoundationPeCoffPermissions: Mapping section 1 of image at 0x7004000 with RW-XN permissions
[8001 0] UpdateMmFoundationPeCoffPermissions: Section 2 of image at 0x7004000 has 0x42000040 permissions
[8001 0] UpdateMmFoundationPeCoffPermissions: Section 2 of image at 0x7004000 has .reloc name
[8001 0] UpdateMmFoundationPeCoffPermissions: Section 2 of image at 0x7004000 has 0x701A000 address
[8001 0] UpdateMmFoundationPeCoffPermissions: Section 2 of image at 0x7004000 has 0x16000 data
api_ffa_mem_perm_set page_count 0
ERROR: FFA_MEM_PERM_SET: page_count was zero
```

I think for this case, just make skip for this.

ArmPkg/Include/Library/StandaloneMmMmuLib.h
LeviYeoReum: This is what I did at first with the same reason But, @samimujawar wants to keep the function name for this.
@samimujawar, Should we keep the function name?

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -184,7 +184,7 @@ EFI_STATUS
 EFIAPI
 ArmFfaLibGetVersion (
   IN  UINT32  RequestVersion,
-  OUT UINT32  *CurrentVersion
+  OUT UINT32  *CurrentVersion OPTIONAL
```
leiflindholm: This change turns a function called "GetVersion" into something that may or may not return a version.
So either the function should be renamed, a new function only checking compatibility should be added, or the argument should remain mandatory.

Regardless, the prototype in MdePkg/Include/Library/ArmFfaLib.h should be updated to match anything that happens here.

leiflindholm (changes requested):

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: TBH, I think I just want to keep ArmFfaLibGetVersion() without optional.
But, the code was written the "CurrentVersion" as if this is optional parameter.
I think it would be better to remove this check but return error in case of this parameter is NULL
and it's enough to change SetPermission's IsFfaMemoryAbiSupport.

Is it okay for you?

leiflindholm: That works for me.

ArmPkg/Include/Library/StandaloneMmMmuLib.h
LeviYeoReum: I've talked with @samimujawar, and I'll change to the function name.
Thanks!

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -190,6 +190,12 @@ ArmFfaLibGetVersion (
   EFI_STATUS    Status;
   ARM_FFA_ARGS  FfaArgs;
 
+  if (CurrentVersion == NULL) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  *CurrentVersion = 0;
```
leiflindholm: If this zero-init is needed, I suggest it should be broken out into a separate commit.

It is not necessary for the stated change, and is an (in my opinion unnecessary) change of behaviour.

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
```diff
@@ -45,17 +45,23 @@ STATIC
...
-  *Version = 0;
+  CurrentVersion = 0;
 
   Status = ArmFfaLibGetVersion (
              ARM_FFA_CREATE_VERSION (ARM_FFA_MAJOR_VERSION, ARM_FFA_MINOR_VERSION),
-             Version
+             &CurrentVersion
              );
+
+  if (Version != NULL) {
```
leiflindholm, line 64: Feels wrong to update *Version before checking Status. Invert order?

We also seem to be doing a lot of setting CurrentVersion to 0 (here, and then unconditionally in ArmFfaLibGetVersion in the preceding commit) when both functions actually return a status.

leiflindholm (changes requested):

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: Okay. I'll remove it anyway callee will check the return value first.

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
LeviYeoReum: Okay. I'll do it later. Thanks!

leiflindholm (approved):
I'm happy with the code.

Would like to see comments from @samimujawar or @ardbiesheuvel on the mappings API change itself.

ArmPkg/Include/Library/StandaloneMmMmuLib.h
```diff
@@ -27,7 +27,7 @@
 
 **/
 EFI_STATUS
-ArmSetMemoryRegionNoExec (
+ArmSetMemoryRegionNoExecBit (
   IN  EFI_PHYSICAL_ADDRESS  BaseAddress,
```
ardbiesheuvel, line 31: Why is it better to put `Bit` at the end?

ardbiesheuvel:
> I'm happy with the code.
> 
> Would like to see comments from @samimujawar or @ardbiesheuvel on the mappings API change itself.

Looks fine to me, although I don't see the point of the change that puts `Bit` at the end of a bunch of function names.

ArmPkg/Include/Library/StandaloneMmMmuLib.h
LeviYeoReum: Well to distingush with "ArmSet/ClearMemoryRegionReadOnly()" But since the following patch changes with Perm suffix, I think it doesn't need to add Bit suffix. I'll remove it. 

Thanks!

LeviYeoReum:
> > I'm happy with the code.
> > Would like to see comments from @samimujawar or @ardbiesheuvel on the mappings API change itself.
> 
> Looks fine to me, although I don't see the point of the change that puts `Bit` at the end of a bunch of function names.

Thanks. I'll remove `Bit` suffix.

ardbiesheuvel (approved):

kuqin12 (approved):

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
samimujawar, line 183: Document error codes?

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
```diff
@@ -309,6 +309,7 @@ ArmSetMemoryRegionNoExec (
   UINTN       Size;
   UINT32      PageCount;
 
+  Status     = EFI_SUCCESS;
```
samimujawar: Please add a comment why the Status is set here to success.

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
```diff
@@ -391,6 +392,7 @@ ArmClearMemoryRegionNoExec (
...
```
samimujawar: same comment as above.

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
```diff
@@ -460,71 +462,44 @@ ArmClearMemoryRegionNoExec (
...
-                            );
-    }
+  UINT32   PermissionRequest;
+  BOOLEAN  UseFfaAbis;
 
-    if (Length < Size) {
-      Length = Size;
-    }
+  if (Length == 0) {
+    /// Nothing to do.
```
samimujawar: Add explanation why length can be 0

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
```diff
@@ -542,65 +517,93 @@ ArmSetMemoryRegionReadOnly (
...
-  BOOLEAN     UseFfaAbis;
-  UINT32      Version;
-  UINTN       Size;
-  UINT32      PageCount;
+  UINT32   PermissionRequest;
+  BOOLEAN  UseFfaAbis;
 
-  UseFfaAbis = IsFfaMemoryAbiSupported (&Version);
+  if (Length == 0) {
+    /// Nothing to do.
```
samimujawar: same comment as above.

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
```diff
@@ -542,65 +517,93 @@ ArmSetMemoryRegionReadOnly (
...
+ArmSetMemoryRegionReadOnlyExecPerm (
+  IN  EFI_PHYSICAL_ADDRESS  BaseAddress,
+  IN  UINT64                Length
+  )
+{
+  UINT32   PermissionRequest;
+  BOOLEAN  UseFfaAbis;
+
+  if (Length == 0) {
+    /// Nothing to do.
```
samimujawar: same comment as previous one.

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
LeviYeoReum: JFYI: see the https://github.com/tianocore/edk2/pull/12536#discussion_r3186248202

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
LeviYeoReum: Okay. But JFYI: see the https://github.com/tianocore/edk2/pull/12536#discussion_r3186248202

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
LeviYeoReum: Done

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
LeviYeoReum: Done

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
LeviYeoReum: Done

ArmPkg/Library/StandaloneMmMmuLib/ArmMmuStandaloneMmLib.c
samimujawar: I understand, but the explanation needs to be there in the code. The comment you referenced above will be lost when the PR is merged.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
LeviYeoReum: Thanks. I'll add

samimujawar (approved):

PR #12151, opened 2026-02-10
CodeQL: scan all core packages and update CLI to v2.24.1

szsam:
# Description

  This PR includes two commits:

  1. .github/workflows/codeql.yml: expand the CodeQL workflow matrix so all core
     packages are scanned.
      - Adds package/arch entries for:
          - EmbeddedPkg (IA32, X64)
          - NetworkPkg (IA32,X64)
          - RedfishPkg (IA32,X64)
          - SignedCapsulePkg (IA32,X64)
          - UefiPayloadPkg (IA32,X64)
  2. BaseTools/Plugin/CodeQL: update CodeQL CLI external dependencies from
     v2.18.1 to v2.24.1.
      - Updates source URLs, versions, and SHA256 values in:
          - codeqlcli_ext_dep.yaml
          - codeqlcli_linux_ext_dep.yaml
          - codeqlcli_windows_ext_dep.yaml
      - Pins CodeQlQueries.qls to codeql/cpp-queries@1.5.10 (matching CodeQL CLI
        v2.24.1).

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

The updated `.github/workflows/codeql.yml`.

## Integration Instructions

N/A

PR #12540, opened 2026-05-04
.github/workflows/codeql.yml: Scan all core packages

szsam:
# Description

Follow-up to stale PR #12151.

Expand the CodeQL package matrix to include EmbeddedPkg, NetworkPkg, RedfishPkg, and UefiPayloadPkg with the appropriate IA32/X64 arch lists. With this update, CodeQL now scans all core packages, improving static analysis coverage.

- [ ] Breaking change
- [ ] Impacts security
- [ ] Includes tests

## How This Was Tested

Github workflow `.github/workflows/codeql.yml`

## Integration Instructions

N/A


szsam:
@makubacki @mdkinney @lgao4 @os-d @Javagedes @spbrogan
 Please review this PR if you have time.

makubacki (approved):
As I mentioned in https://github.com/tianocore/edk2/pull/12151, an increase in resources used by GitHub runners is a reason these packages were not added in the past. However, if there is interest in addressing CodeQL issues in these packages, I'm not opposed to running it there as well.

os-d:
> As I mentioned in #12151, an increase in resources used by GitHub runners is a reason these packages were not added in the past. However, if there is interest in addressing CodeQL issues in these packages, I'm not opposed to running it there as well.

At a minimum, I’d recommend not running it for IA32. That could be done as a separate PR to remove building it for IA32 in all packages.

I don’t have much of an opinion here, other than I don’t want to increase CI times too much. I think the argument for running these would be stronger if fixes for the CodeQL errors came in and then the CI was enabled for it; e.g. if this is just going to add more errors but isn’t actually going to be used to fix things I don’t think it is that useful. Requiring new CodeQL errors be fixed before a PR is merged can be useful, but it isn’t clear to me what this would add beyond the existing GH CodeQL comments that get posted.

lgao4 (approved):

makubacki:
@szsam, would you be interested in dropping `IA32` in this commit? I agree there is not much value there and it can always be tweaked later.

szsam:
> @szsam, would you be interested in dropping `IA32` in this commit? I agree there is not much value there and it can always be tweaked later.

Dropped IA32 for newly added packages.

Issue #12390, opened 2026-04-02
[Bug]:PXE crash in NetworkPkg when link disconnects during NBP download (NULL/invalid Private access)


kajal-agarwal-in:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

NetworkPkg

### Which targets are impacted by this bug?

DEBUG, RELEASE

### Current Behavior

- When the PXE network cable is disconnected during an active NBP download, the PXE stack does not abort cleanly.
- Instead, it repeatedly encounters device errors from the USB/network stack.
- Multiple failures are observed from Snp->Receive() and USB transfers returning EFI_DEVICE_ERROR.
- Despite these errors, execution continues inside UefiPxeBcDxe.
- This leads to an invalid memory access and eventual crash.
- The failure manifests as a synchronous exception with the program counter set to a poisoned value (0xAFAFAFAFAFAFAFAF).
- This strongly indicates a use‑after‑free or invalid access to the PXE Private structure.

>>Start PXE over IPv4.
Station IP address is 192.168.0.123
Server IP address is 192.168.0.120
NBP filename is Boot\HelloTest.efi
NBP filesize is 50000000 Bytes
Downloading NBP file...
XhcCheckUrbResult: TRANSACTION_ERROR! Completecode = 4
XhcBulkTransfer: error - Device Error, transfer - 200
MnpReceivePacket: Snp->Receive() = Device Error.
XhcCheckUrbResult: TRANSACTION_ERROR! Completecode = 4
XhcBulkTransfer: error - Device Error, transfer - 200
MnpReceivePacket: Snp->Receive() = Device Error.
XhcCheckUrbResult: TRANSACTION_ERROR! Completecode = 4
XhcBulkTransfer: error - Device Error, transfer - 200
MnpReceivePacket: Snp->Receive() = Device Error.
XhcCheckUrbResult: TRANSACTION_ERROR! Completecode = 4
XhcBulkTransfer: error - Device Error, transfer - 200
MnpReceivePacket: Snp->Receive() = Device Error.
XhcControlTransfer: error - Device Error, transfer - 100
XhcControlTransfer: error - Device Error, transfer - 100
XhcControlTransfer: error - Device Error, transfer - 100
XhcControlTransfer: error - Device Error, transfer - 100
XhcControlTransfer: error - Device Error, transfer - 100
[...]

PR #12542, opened 2026-05-05
NetworkPkg/UefiPxeBcDxe: Exit gracefully when PXE is disconnected

kajal-agarwal-in:
This PR resolves [[Issue #12390](https://github.com/tianocore/edk2/issues/12390)], where PXE boot crashes occur in NetworkPkg when the network link disconnects during NBP (Network Boot Program) download. The root cause is invalid or NULL private data access triggered by unexpected PXE disconnection.

Description
- Added validation checks to ensure private data structures are valid before use.
- Improved error handling in DHCP and MTFTP flows to return meaningful status codes instead of crashing.
- Updated driver logic to exit gracefully when PXE is disconnected, preventing hangs and resource leaks.
- These changes strengthen PXE boot reliability by ensuring the system handles disconnect scenarios cleanly.

- [ ] Breaking change?
No. This update does not introduce new library classes or move modules; it only enhances existing driver behavior.
- [ ] Impacts security?
No. This change does not affect cryptographic algorithms or security-sensitive code paths.
- [ ] Includes tests?
No explicit tests included. Validation was performed through functional boot testing.
## How This Was Tested

Functional boot testing was performed by reproducing the issue as below
- Start a PXE boot installation.
- Disconnect the PXE cable mid‑installation.
- Verified that Mtftp4Dxe and UefiPxeBcDxe exit gracefully without hangs or resource leaks.
- Confirmed normal PXE boot continues to work when PXE remains connected.

## Integration Instructions
- Consumers of NetworkPkg can integrate these changes directly.
- No build or configuration changes are needed.

### DIFF
--- a/NetworkPkg/UefiPxeBcDxe/PxeBcDhcp4.c
+++ b/NetworkPkg/UefiPxeBcDxe/PxeBcDhcp4.c
@@ -1717,6 +1717,12 @@ PxeBcDhcp4Dora (
     goto ON_EXIT;
   }
 
+  if ((Private == NULL) ||
+      (Private->Signature != PXEBC_PRIVATE_DATA_SIGNATURE))
+  {
+    return Status;
+  }
+
   //
   // Get the acquired IPv4 address and store them.
   //
@@ -1749,6 +1755,13 @@ PxeBcDhcp4Dora (
   AsciiPrint ("\n");
 
 ON_EXIT:
+
+  if ((Private == NULL) ||
+      (Private->Signature != PXEBC_PRIVATE_DATA_SIGNATURE))
+  {
+    return Status;
+  }
+
   if (EFI_ERROR (Status)) {
     Dhcp4->Stop (Dhcp4);
     Dhcp4->Configure (Dhcp4, NULL);
--- a/NetworkPkg/UefiPxeBcDxe/PxeBcImpl.c
+++ b/NetworkPkg/UefiPxeBcDxe/PxeBcImpl.c
@@ -470,6 +470,10 @@ EfiPxeBcDhcp (
     Status = PxeBcDhcp4Dora (Private, Private->Dhcp4);
   }
 
+  if (Private->Signature != PXEBC_PRIVATE_DATA_SIGNATURE) {
+    return Status;
+  }
+
   //
   // Reconfigure the UDP instance with the default configuration.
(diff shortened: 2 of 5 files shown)


NetworkPkg/UefiPxeBcDxe/PxeBcDhcp4.c
```diff
@@ -1717,6 +1717,12 @@ PxeBcDhcp4Dora (
     goto ON_EXIT;
   }
 
+  if ((Private == NULL) ||
+      (Private->Signature != PXEBC_PRIVATE_DATA_SIGNATURE))
+  {
+    return Status;
+  }
+
```
mikebeaton, line 1725: We've already used Private several times above here in this method. If it was NULL or pointing to memory with the wrong signature, we would already have accessed incorrect memory multiple times.

Also, Status can only be EFI_SUCCESS at this point, given the conditional immediately above.

NetworkPkg/UefiPxeBcDxe/PxeBcDhcp4.c
```diff
@@ -1749,6 +1755,13 @@ PxeBcDhcp4Dora (
   AsciiPrint ("\n");
 
 ON_EXIT:
+
+  if ((Private == NULL) ||
+      (Private->Signature != PXEBC_PRIVATE_DATA_SIGNATURE))
+  {
+    return Status;
+  }
```
mikebeaton, line 1763: Similarly to the above, Private has already been used, for good or bad, before the first jump to ON_EXIT.

NetworkPkg/UefiPxeBcDxe/PxeBcImpl.c
```diff
@@ -470,6 +470,10 @@ EfiPxeBcDhcp (
     Status = PxeBcDhcp4Dora (Private, Private->Dhcp4);
   }
 
+  if (Private->Signature != PXEBC_PRIVATE_DATA_SIGNATURE) {
+    return Status;
+  }
```
mikebeaton, line 475: Similarly, we've used whatever Private points to before we get to this test. (Unless the idea that the memory can be zeroed out from under us? If so the fix doesn't seem sufficient, as surely that can happen at any time, so these changes would only be working round a flawed design...)

Issue #11962, opened 2026-01-05
[Bug]: LPA2 support regresses emulated boot w/ -cpu max


dannf:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

ArmVirtPkg

### Which targets are impacted by this bug?

RELEASE

### Current Behavior

After updating Debian to edk2-stable202511, booting emulated AARCH64 guests w/ `-cpu max` hangs:
```
qemu-system-arm64 -cpu max -machine virt -bios /usr/share/qemu-efi-aarch64/QEMU_EFI.fd -m 1G -chardev stdio,mux=on,id=char0 -nographic -monitor none -serial chardev:char0
UEFI firmware (version 2025.11-2 built at 10:42:49 on Jan  3 2026)
```

Bisection identified commit 90771630bfe123de8e4bacc96f98ffe4d8308153. It is still reproducible with latest git (commit 81c95449749469480f0124a1416f22b4ecc67584). As well as with https://github.com/tianocore/edk2/pull/11961. I also found that the `virt-6.2` machine type boots fine, while `virt-7.0` does not. The difference between the 2 is that `virt-6.2` has `lpa2` disabled.

### Expected Behavior

No hang.

### Steps To Reproduce

Debian unstable, build command:
```
. edksetup.sh; export MAKEFLAGS=; build -a AARCH64 -t GCC5 -p ArmVirtPkg/ArmVirtQemu.dsc -DCC_MEASUREMENT_ENABLE=TRUE -DNETWORK_HTTP_BOOT_ENABLE=TRUE -DNETWORK_IP6_ENABLE=TRUE -DNETWORK_TLS_ENABLE --pcd PcdFirmwareVendor=L"Debian distribution of EDK II\\0" --pcd PcdFirmwareVersionString=L"2025.11-2\\0" --pcd PcdFirmwareReleaseDateString=L"01/03/2026\\0" --pcd PcdUninstallMemAttrProtocol=TRUE -b RELEASE
```

Test command - QEMU 10.2.0:

`qemu-system-aarch64 -cpu max -machine virt-7.0 -bios /usr/share/qemu-efi-aarch64/QEMU_EFI.fd -m 1G -chardev stdio,mux=on,id=char0 -nographic -monitor none -serial chardev:char0`

### Build Environment

```markdown
- OS(s): Debian unstable
[...]

PR #12545, opened 2026-05-05
ArmVirtPkg/ArmPlatformLibQemu: Support early ID map on LPA2 capable CPUs

ardbiesheuvel:
ArmVirtQemu uses an initial early ID map in emulated NOR flash that is activated right out of reset, in order to avoid running a good chunk of the SEC and PEI stages with the MMU and caches disabled.

This early ID map is currently not compatible with LPA2, and the associated startup code consciously omits 52-bit physical addressing, as it was not supported for 4k page sizes when this code was written.

This means that by the time ArmConfigureMmu() runs, the TCR control register is already programmed with a value that reflects the non-LPA2 nature of the active page tables. Given that the MMU is on at that point, TCR cannot simply be updated to a value that enables LPA2. Yet this is what happens currently, and so the boot hangs really early when running ArmVirtQemu on LPA2 capable CPUs.

Disabling the MMU again at this point is problematic, because it would require cache maintenance on the stack and on other live RAM regions.

Given that the early page tables are in read-only NOR flash, updating them on the fly to make the LPA2 compatible is impossible. So instead, provide a second set of early page tables, but using 5 levels of paging for LPA2, and omitting the shareability attributes in the page table descriptors, as LPA2 repurposes these bits as physical address bits.

Also update the startup code to program TCR.IPS and TCR.DS appropriately, but only if LPA2 is supported and the PArange is 52 bits: this reflects the logic in ArmConfigureMmu(), ensuring that the TCR value it calculates is identical to the one chosen by the early code.

### DIFF
--- a/ArmVirtPkg/Library/ArmPlatformLibQemu/ArmPlatformHelper.S
+++ b/ArmVirtPkg/Library/ArmPlatformLibQemu/ArmPlatformHelper.S
@@ -43,6 +43,8 @@
  .set    ID_AA64MMFR1_VH_MASK,  0xf << 8
  .set    HCR_EL2_E2H,           0x1 << 34
 
+ .set    ID_AA64MMFR0_TG4_52,   0xf << 28
+ .set    TCR_DS,                0x1 << 59
 
 ASM_FUNC(ArmPlatformPeiBootAction)
 #ifdef CAVIUM_ERRATUM_27456
@@ -75,15 +77,21 @@ ASM_FUNC(ArmPlatformPeiBootAction)
   mov_i  x0, mairval
   mov_i  x1, tcrval
   adrp   x2, idmap
-  orr    x2, x2, #0xff << 48     // set non-zero ASID
   mov_i  x3, sctlrval
 
   mrs    x6, id_aa64mmfr0_el1    // get the supported PA range
+  and    x4, x6, #ID_AA64MMFR0_TG4_52 // check for LPA2 support
   and    x6, x6, #0xf            // isolate PArange bits
   cmp    x6, #6                  // 0b0110 == 52 bits
+  csel   x4, x4, xzr, eq         // ignore LPA2 capability unless PArange == 52
+  ccmp   x4, xzr, #0, eq         // permit PS == 52 bits iff LPA2 is supported
   sub    x6, x6, #1              // subtract 1
   cinc   x6, x6, ne              // add back 1 unless PArange == 52 bits
   bfi    x1, x6, #32, #3         // copy updated PArange into TCR_EL1.IPS
+  cbz    x4, 1f
+  orr    x1, x1, #TCR_DS
+  adrp   x2, idmap_lpa2          // switch to LPA2 idmap
+1:orr    x2, x2, #0xff << 48     // set non-zero ASID
 
   cmp    x6, #3                  // 0b0011 == 42 bits
   sub    x6, x6, #1              // subtract 1
--- a/ArmVirtPkg/Library/ArmPlatformLibQemu/IdMap.S
+++ b/ArmVirtPkg/Library/ArmPlatformLibQemu/IdMap.S
@@ -13,49 +13,57 @@
 
   .set      TT_MT_DEV, 0x0 << 2                 // MAIR #0
   .set      TT_MT_MEM, (0x3 << 2) | (0x3 << 8)  // MAIR #3
+  .set      TT_MT_MEM_LPA2, (0x3 << 2)          // MAIR #3
 
-  .set      PAGE_XIP,  TT_TYPE_PAGE  | TT_MT_MEM | TT_AF | TT_RO | TT_NG
-  .set      BLOCK_XIP, TT_TYPE_BLOCK | TT_MT_MEM | TT_AF | TT_RO | TT_NG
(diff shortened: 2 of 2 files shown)


ArmVirtPkg/Library/ArmPlatformLibQemu/IdMap.S
```diff
@@ -13,17 +13,30 @@
 
   .set      TT_MT_DEV, 0x0 << 2                 // MAIR #0
   .set      TT_MT_MEM, (0x3 << 2) | (0x3 << 8)  // MAIR #3
+  .set      TT_MT_MEM_LPA2, (0x3 << 2)          // MAIR #3
+
+  .set      BLOCK_DEV, TT_TYPE_BLOCK | TT_MT_DEV | TT_AF | TT_XN | TT_NG
```
leiflindholm: Why does this line move?

ArmVirtPkg/Library/ArmPlatformLibQemu/IdMap.S
```diff
@@ -13,17 +13,30 @@
...
+  .set      BLOCK_MEM_LPA2, TT_TYPE_BLOCK | TT_MT_MEM_LPA2 | TT_AF | TT_XN | TT_NG
+
   .section  ".rodata.idmap", "a"
   .align    12
 
-idmap:      /* level 0 */
+   .macro   make_idmap, name, lpa=
+  .globl    \name
+\name:
+  .ifnb     \lpa
```
leiflindholm: I don't know how much it matters to keep this code future-proof, and I don't exactly expect churn, but it feels a little bit awkward that there's a parameter called lpa that if specified means "generate map for FEAT_LPA2" and if not means "generate map for pre-FEAT_LPA2".

Could we have a comment?

ArmVirtPkg/Library/ArmPlatformLibQemu/IdMap.S
ardbiesheuvel: because it is the only line that does not incorporate `TT_MT_MEM`. The others all do, and are grouped together.

ArmVirtPkg/Library/ArmPlatformLibQemu/IdMap.S
ardbiesheuvel: Sure.

ArmVirtPkg/Library/ArmPlatformLibQemu/IdMap.S
leiflindholm: Fair enough.

ArmVirtPkg/Library/ArmPlatformLibQemu/ArmPlatformHelper.S
```diff
@@ -75,15 +77,21 @@ ASM_FUNC(ArmPlatformPeiBootAction)
...
+  and    x4, x6, #ID_AA64MMFR0_TG4_52 // check for LPA2 support
   and    x6, x6, #0xf            // isolate PArange bits
   cmp    x6, #6                  // 0b0110 == 52 bits
+  csel   x4, x4, xzr, eq         // ignore LPA2 capability unless PArange == 52
+  ccmp   x4, xzr, #0, eq         // permit PS == 52 bits iff LPA2 is supported
   sub    x6, x6, #1              // subtract 1
   cinc   x6, x6, ne              // add back 1 unless PArange == 52 bits
   bfi    x1, x6, #32, #3         // copy updated PArange into TCR_EL1.IPS
+  cbz    x4, 1f
+  orr    x1, x1, #TCR_DS
```
LeviYeoReum, line 92: For a E2H it is 59 bit. but none E2H case, Is it 32 bit IIRC?:
  - https://github.com/tianocore/edk2/blob/master/MdePkg/Include/AArch64/AArch64Mmu.h#L132

ardbiesheuvel: This code can only run in E2H mode at EL2.

ArmVirtPkg/Library/ArmPlatformLibQemu/IdMap.S
ardbiesheuvel: I'll make it self-documenting instead.

ArmVirtPkg/Library/ArmPlatformLibQemu/ArmPlatformHelper.S
LeviYeoReum: Ah I miss line 67 Thanks.

LeviYeoReum (approved):

leiflindholm (approved):

ardbiesheuvel:
@lgao4 @mdkinney 

Please consider this for the stable tag as a bugfix.

lgao4:
> @lgao4 @mdkinney
> 
> Please consider this for the stable tag as a bugfix.

Seemly, this is a clear bug fix. I am OK to merge it.

oshoval:
Hi
Seems this fix this issue
https://github.com/tianocore/edk2/issues/11962#issuecomment-4418779208
(at least in the case i mentioned there)

when can we expect please a new official F43 release for it ?

cc @dannf fyi please

Thanks !

PR #12550, opened 2026-05-06
Close the Redfish interface when the connection is no longer needed

changab:
# Description
Close the Redfish interface when the connection is no longer needed. This is the follow-up PR of https://github.com/tianocore/edk2/pull/11437.

The use case is on edk2-redfish-clinet, https://github.com/tianocore/edk2-redfish-client/pull/129.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Tested on AMD platform.

## Integration Instructions

N/A

### DIFF
--- a/RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.c
+++ b/RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.c
@@ -41,6 +41,43 @@ EFI_DRIVER_BINDING_PROTOCOL  gRedfishConfigDriverBinding = {
 };
 
 /**
+  Disconnect the interface communication with REST EX protocol.
+
+  @param[in]   Event    Event whose notification function is being invoked.
+  @param[out]  Context  Pointer to the Context buffer
+
+**/
+VOID
+EFIAPI
+RedfishConfigDisconnection (
+  IN  EFI_EVENT  Event,
+  OUT VOID       *Context
+  )
+{
+  EFI_STATUS                   Status;
+  EFI_REDFISH_DISCOVERED_LIST  *DiscoveredList;
+
+  if (gRedfishServiceDiscovered) {
+    DiscoveredList = (EFI_REDFISH_DISCOVERED_LIST *)Context;
+    if (DiscoveredList == NULL) {
+      DEBUG ((DEBUG_ERROR, "%a: DiscoveredList == NULL, failed to release Redfish service.\n", __func__));
+      return;
+    } else {
+      Status = gEfiRedfishDiscoverProtocol->ReleaseRedfishService (gEfiRedfishDiscoverProtocol, DiscoveredList);
+      FreePool (DiscoveredList);
+      if (!EFI_ERROR (Status)) {
+        gBS->CloseEvent (Event);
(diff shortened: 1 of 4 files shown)


RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.c
```diff
@@ -331,6 +371,33 @@ RedfishServiceDiscoveredCallback (
...
+                    EVT_NOTIFY_SIGNAL,
+                    TPL_CALLBACK,
+                    RedfishConfigDisconnection,
+                    (VOID *)ThisDiscoveredList,
+                    &gEdkIIRedfisEventRedfishInterfaceDisconnectionGuid,
+                    &ThisEvent
+                    );
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "%a: Fail to register Redfish transport interface disconnection event.\n", __func__));
+      FreePool (ThisDiscoveredList);
```
igorkulchytskyy, line 397: Should we free the allocated memory of ThisDiscoveredList at the end of the function?
Then, if later we should add some code in between with a goto to exit, we do not need to remember to free the memory

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.c
```diff
@@ -320,8 +357,11 @@ RedfishServiceDiscoveredCallback (
   OUT VOID       *Context
   )
 {
+  EFI_STATUS                       Status;
+  EFI_EVENT                        ThisEvent;
   EFI_REDFISH_DISCOVERED_TOKEN     *RedfishDiscoveredToken;
   EFI_REDFISH_DISCOVERED_INSTANCE  *RedfishInstance;
+  EFI_REDFISH_DISCOVERED_LIST      *ThisDiscoveredList;
 
   RedfishDiscoveredToken = (EFI_REDFISH_DISCOVERED_TOKEN *)Context;
```
igorkulchytskyy, line 366: Maybe it would be good to add a check if Context is not NULL?

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.c
changab: @igorkulchytskyy,  ThisDiscoveredList is passed to the callback function as the context. So, we free ThisDiscoveredList in the callback function.

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.c
changab: Good catch, I think we need another PR for this, as this is a separate fix.

RedfishPkg/RedfishConfigHandler/RedfishConfigHandlerDriver.c
igorkulchytskyy: Yes. Got it

igorkulchytskyy (approved):

nicklela (approved):
LGTM

PR #12553, opened 2026-05-07
DynamicTablesPkg/AmlLib: Add AmlMethodParamTypeBuffer support

abdattar:
Extend AML_METHOD_PARAM_TYPE enum with AmlMethodParamTypeBuffer (= 4) to allow callers to pass a raw byte Buffer as an AML method argument.

Add a Buffer case to the switch in AmlCodeGenInvokeMethod() that delegates to the existing AmlCodeGenBuffer() helper, with NULL-pointer and zero-length guards matching the pattern of the other cases. AmlCodeGenReturnInvokeMethod() already delegates to AmlCodeGenInvokeMethod() so it picks up the new case automatically.

Update AML_METHOD_PARAM documentation: clarify that DataSize carries the byte count when Type is AmlMethodParamTypeBuffer, and update the union/struct field comments accordingly.

Cc: Pierre Gondois <pierre.gondois@arm.com>
Cc: Sami Mujawar <sami.mujawar@arm.com>

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Tested on AMD platform

## Integration Instructions
N/A

### DIFF
--- a/DynamicTablesPkg/Library/Common/AmlLib/CodeGen/AmlCodeGen.c
+++ b/DynamicTablesPkg/Library/Common/AmlLib/CodeGen/AmlCodeGen.c
@@ -4138,6 +4138,26 @@ AmlCodeGenInvokeMethod (
           }
 
           break;
+        case AmlMethodParamTypeBuffer:
+          if ((Parameters[Index].Data.Buffer == NULL) !=
+              (Parameters[Index].DataSize == 0))
+          {
+            ASSERT_EFI_ERROR (EFI_INVALID_PARAMETER);
+            Status = EFI_INVALID_PARAMETER;
+            goto exit_handler;
+          }
+
+          Status = AmlCodeGenBuffer (
+                     Parameters[Index].Data.Buffer,
+                     (UINT32)Parameters[Index].DataSize,
+                     &ObjectNode
+                     );
+          if (EFI_ERROR (Status)) {
+            ASSERT_EFI_ERROR (Status);
+            goto exit_handler;
+          }
+
+          break;
         default:
           ASSERT_EFI_ERROR (EFI_INVALID_PARAMETER);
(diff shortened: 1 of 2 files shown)


DynamicTablesPkg/Library/Common/AmlLib/CodeGen/AmlCodeGen.c
```diff
@@ -4137,6 +4137,26 @@ AmlCodeGenInvokeMethod (
             goto exit_handler;
           }
 
+          break;
+        case AmlMethodParamTypeBuffer:
+          if ((Parameters[Index].Data.Buffer == NULL) ||
```
pierregondois: Just one comment, `AmlCodeGenBuffer()` allows to create empty buffers if `Buffer == NULL and BufferSize == 0`. To also allow it here, maybe this should be:
```
          // Allow empty Buffer creation
          if ((Parameters[Index].Data.Buffer == NULL) &&
              (Parameters[Index].DataSize == 0))
```
?

abdattar: Done.

pierregondois: The condition should actually be:
```
          if ((Parameters[Index].Data.Buffer == NULL) !=
              (Parameters[Index].DataSize == 0))
```
in order to allow empty buffers while rejecting invalid parameters.

Sorry for the bad suggestion. Would it be possible to update it again ?

abdattar: Done. replaced with "!=" operator.

pierregondois (approved):

PR #12556, opened 2026-05-07
MdePkg: StackCheckLib: Update __stack_chk_guard for GCC 17

### LINKED ISSUE #12547: [Bug]: __stack_chk_guard definition is wrong

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

Other

RELEASE, NOOPT, NO-TARGET, DEBUG

### Current Behavior

https://github.com/tianocore/edk2/blob/f3d49c80246dc2923bdbb49ba3cd46c747660f5c/MdePkg/Include/Library/StackCheckLib.h#L20
https://github.com/tianocore/edk2/blob/f3d49c80246dc2923bdbb49ba3cd46c747660f5c/MdePkg/Library/StackCheckLib/StackCheckLibCommonGcc.c#L24
https://github.com/tianocore/edk2/blob/f3d49c80246dc2923bdbb49ba3cd46c747660f5c/MdePkg/Library/StackCheckLibNull/StackCheckLibNullGcc.c#L11

Currently the symbol is defined as `void*`.

### Expected Behavior

GCC trunk (as of May 5, 2026) now expects it to be defined as `uintptr_t`.

### Steps To Reproduce

GCC 17+

### Build Environment

```markdown
GCC 17+
```

### Version Information

```text
hash f3d49c80246dc2923bdbb49ba3cd46c747660f5c .
```

### Urgency

High

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

_No response_

os-d:
# Description

GCC 17 (upcoming release) is updating the type of
__stack_chk_guard from a void * to a uintptr_t.

This is only a type change, not a size change. However, it does cause the GCC 17 build to fail because of the type mismatch.

uintptr_t is defined as unsigned long, which is not a type edk2 defines. It has the same size as UINTN on a given system, but on 64 bit systems we have the same type issue because UINTN is defined as unsigned long long, not unsigned long.

To work around this and avoid defining a new type in edk2, use the compiler built in __UINTPTR_TYPE__ which is the underlying definition in GCC.

This is a backwards compatible change with previous versions of GCC and CLANGDWARF but also fixes the upcoming definition change for GCC 17.

Closes #12547 .

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Built GCC 17 at the commit referenced in the GCC bugzilla for this: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=121911.
Built OvmfPkgIa32X64 and ArmVirtQemu with this GCC and confirmed I saw the build failure and the new definition works and boots.

Then used my system GCC (GCC 15) and confirmed the same scenarios worked.

Then enabled stack cookies on CLANGDWARF as an extra test and confirmed the same scenarios work.

## Integration Instructions

N/A.

### DIFF
--- a/MdePkg/Include/Library/StackCheckLib.h
+++ b/MdePkg/Include/Library/StackCheckLib.h
@@ -17,7 +17,10 @@
 // and the return address so that continuously writing past the stack variables will cause
 // the stack cookie to be overwritten. Before the function returns, the stack cookie value
 // will be checked and if there is a mismatch then StackCheckLib handles the failure.
-extern VOID  *__stack_chk_guard;
+// __UINTPTR_TYPE__ will be the same bitwidth as UINTN, but is a different type that edk2 does
+// not define (e.g. on X64 this is unsigned long) and so we cannot use UINTN here. Newer versions
+// of GCC define this as uintptr_t, which is unsigned long on X64.
+extern __UINTPTR_TYPE__  __stack_chk_guard;
 
 /**
   Called when a stack cookie check fails. The return address is the failing address.
--- a/MdePkg/Library/StackCheckLib/StackCheckLibCommonGcc.c
+++ b/MdePkg/Library/StackCheckLib/StackCheckLibCommonGcc.c
@@ -21,7 +21,7 @@ TriggerStackCookieInterrupt (
   VOID
   );
 
-VOID  *__stack_chk_guard = (VOID *)(UINTN)STACK_COOKIE_VALUE;
+__UINTPTR_TYPE__  __stack_chk_guard = (__UINTPTR_TYPE__)STACK_COOKIE_VALUE;
 
 /**
   This function gets called when a gcc/clang generated stack cookie fails. This implementation calls into a platform
--- a/MdePkg/Library/StackCheckLibNull/StackCheckLibNullGcc.c
+++ b/MdePkg/Library/StackCheckLibNull/StackCheckLibNullGcc.c
@@ -8,7 +8,7 @@
 #include <Uefi.h>
 #include <Library/StackCheckLib.h>
 
-VOID  *__stack_chk_guard = (VOID *)(UINTN)0x0;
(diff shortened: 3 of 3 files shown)


trofi (approved):
Fixes the `gcc-17` build for me as well.

os-d:
Friendly ping to @mdkinney and @lgao4 on this one.

@ardbiesheuvel perhaps you have thoughts here?

MdePkg/Include/Library/StackCheckLib.h
```diff
@@ -17,7 +17,10 @@
 // and the return address so that continuously writing past the stack variables will cause
 // the stack cookie to be overwritten. Before the function returns, the stack cookie value
 // will be checked and if there is a mismatch then StackCheckLib handles the failure.
-extern VOID  *__stack_chk_guard;
+// __UINTPTR_TYPE__ will be the same bitwidth as UINTN, but is a different type that edk2 does
+// not define (e.g. on X64 this is unsigned long) and so we cannot use UINTN here. Newer versions
+// of GCC define this as uintptr_t, which is unsigned long on X64.
+extern __UINTPTR_TYPE__  __stack_chk_guard;
```
lgao4, line 23: older version gcc compiler also defines this type __UINTPTR_TYPE__?

os-d: @lgao4 yes, I tested with GCC 15 and it appears this has been in GCC (and so CLANGDWARF also) from much earlier than edk2’s minimum supported version. I expect this to be fully backwards compatible with supported GCC versions (and so CLANGDWARF as well, if stack check lib support was added for it).

lgao4: If this change is compatible, I agree it.

lgao4 (approved):

ardbiesheuvel (approved):

PR #12563, opened 2026-05-08
MdeModulePkg/ArmFfaLib: Fix VM ID handling for RX release and RXTX unmap

shenghuliu:
### **Description**

This is follow-up of https://github.com/tianocore/edk2/pull/12235

According to the FF‑A specification, when FFA_RX_RELEASE or FFA_RXTX_UNMAP is invoked from UEFI as a non-secure virtual instance, the w1 register must be zero (MBZ). Supplying a non‑zero value in this context results in the SPMC returning FFA_INVALID_PARAMETER.

When these functions are invoked from UEFI as a non-secure physical instance, FFA_ID_GET always returns a VM ID of zero. As a result, passing a non‑zero VM ID is unnecessary and may be invalid.

Therefore, in edk2, w1 (VM ID or partition ID) should always be set to zero when invoking FFA_RX_RELEASE and FFA_RXTX_UNMAP.

 [ ] Breaking change?

 [ ] Impacts security?

 [ ] Includes tests?

### **How This Was Tested**
This change was tested on an internal device that complies with the Hafnium implementation.

### **Integration Instructions**
There is no special requirements.

### DIFF
--- a/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
+++ b/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
@@ -2,6 +2,8 @@
   Arm Ffa library common code.
 
   Copyright (c) 2024, Arm Limited. All rights reserved.<BR>
+  Copyright (c) Qualcomm Technologies, Inc. All rights reserved.<BR>
+
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
    @par Glossary:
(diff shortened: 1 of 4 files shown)


shenghuliu:
Sorry for long delay due to personal leave. Please review and let me know your comments. thanks.

LeviYeoReum (commented):
Would you describe commit message more detail?
Just likeL

    According to the FF-A specification, when
    FFA_RXTX_UNMAP is invoked from any instance other than non-secure
    physical, the w1 register must be zero (MBZ). If a non-zero value is
    supplied in this context, the SPMC returns FFA_INVALID_PARAMETER.

    The UEFI operates as a guest or non-secure
    physical instance where the partition ID is always zero and is not
    invoked from a hypervisor context where w1 carries a VM ID and 
    StandlaoneMm operates as a secure physical instance. 

    Therefore, passing a VM ID is unnecessary and potentially invalid in
    edk2 for FFA_RXTX_UNMAP. 
    Set w1 (partition id) always zero for FFA_RXTX_UNMAP.  

and uncrustify please for CI complain.

shenghuliu:
Adding @leiflindholm and @vishalo for review.

shenghuliu:
@LeviYeoReum  Incorrect commits are fixed - this is ready for review now

LeviYeoReum (approved):

kuqin12 (approved):

PR #12567, opened 2026-05-11
Tpl control protocol

ardbiesheuvel:


os-d:
I did a quick smoke test in the two problem areas I had tested with before, both looked good.

The first was OvmfPkg with this diff

```c
diff --git a/OvmfPkg/LocalApicTimerDxe/LocalApicTimerDxe.c b/OvmfPkg/LocalApicTimerDxe/LocalApicTimerDxe.c
index c8f08b983e..fce81b8ef0 100644
--- a/OvmfPkg/LocalApicTimerDxe/LocalApicTimerDxe.c
+++ b/OvmfPkg/LocalApicTimerDxe/LocalApicTimerDxe.c
@@ -79,6 +79,7 @@ TimerInterruptHandler (
   }

   mTplControl->RestoreTpl (OriginalTPL);
+  gBS->Stall (100000);
 }

 /**
diff --git a/OvmfPkg/OvmfPkgX64.dsc b/OvmfPkg/OvmfPkgX64.dsc
index 29027fa63f..28b8179dac 100644
--- a/OvmfPkg/OvmfPkgX64.dsc
+++ b/OvmfPkg/OvmfPkgX64.dsc
@@ -579,6 +579,7 @@
 !endif

 [PcdsFixedAtBuild]
+  gEfiMdeModulePkgTokenSpaceGuid.PcdCpuStackGuard|TRUE
   gEfiMdeModulePkgTokenSpaceGuid.PcdStatusCodeMemorySize|1
 !if $(SMM_REQUIRE) == FALSE
   gEfiMdeModulePkgTokenSpaceGuid.PcdResetOnMemoryTypeInformationChange|FALSE
 ```  
 
 With NestedInterruptTplLib reverted (and without your change) this hits a stack overflow. Now it just hangs (as expected).
 
 The other case was the UsbBusDxe case I mentioned in the other PR, it is an example of an event that relies on interrupts being enabled and that case boots to the shell (with broken versions of the interrupt handling it would hang waiting for an interrupt).

So immediate thoughts are this looks good, I will review and do some more testing later this week, few other things queued up before I can get back to it.

mdkinney:
Adding a new protocol dependency to the DXE Core is adding a new Arch Protocol which would require PI Specification changes and changing the set of protocols required before UEFI Drivers can be dispatched.

ardbiesheuvel:
> Adding a new protocol dependency to the DXE Core is adding a new Arch Protocol which would require PI Specification changes and changing the set of protocols required before UEFI Drivers can be dispatched.

IMO there are two ways to go about this:
- keep this as an EDK2 internal 'optimization' with no impact on the PI specification, limiting its use to in-tree timer drivers that are affected by the issue,
- contribute the protocol to the PI specification, making it available to any PI compliant component, including out-of-tree DXE drivers that implement interrupt handlers that require this functionality.

I have no strong preference either way, but if we go with a PI spec change, I'd prefer for someone else to take ownership of this.

Note that there is no protocol dependency being introduced in the core code here: only the OVMF specific timer driver gains a dependency on the TPL protocol exposed by DXE core, but everything else works happily as before, including UEFI drivers having implicit dependencies on the set of PI architectural protocols.

os-d:
> > Adding a new protocol dependency to the DXE Core is adding a new Arch Protocol which would require PI Specification changes and changing the set of protocols required before UEFI Drivers can be dispatched.
> 
> IMO there are two ways to go about this:
> 
> * keep this as an EDK2 internal 'optimization' with no impact on the PI specification, limiting its use to in-tree timer drivers that are affected by the issue,
> * contribute the protocol to the PI specification, making it available to any PI compliant component, including out-of-tree DXE drivers that implement interrupt handlers that require this functionality.
> 
> I have no strong preference either way, but if we go with a PI spec change, I'd prefer for someone else to take ownership of this.
> 

As I mentioned in the other PR, I think this makes sense to go the PI spec because this is across a core <-> platform boundary. The platforms in edk2 are only in edk2 out of convenience, they could just as easily be in edk2-platforms or private. If we agree on approach, I can take it there (I would do code first in edk2, so all review would still occur here; if you wanted to convert this PR into a code first PR and file an edk2 issue, I could also help file the mantis and point the PIWG here. Whatever works).

> Note that there is no protocol dependency being introduced in the core code here: only the OVMF specific timer driver gains a dependency on the TPL protocol exposed by DXE core, but everything else works happily as before, including UEFI drivers having implicit dependencies on the set of PI architectural protocols.

Agreed, I don't believe this is an architectural protocol because DXE core doesn't consume it; it produces it. If we take this to the PI spec, it could go as the current standalone protocol or one could imagine adding it as part of DXE services.

PR #12570, opened 2026-05-12
OvmfPkg: MmControlPei: Fix QEMU does not send MMI to all cores

kuqin12:
# Description

Because there was no negotiation, the SMM controller from QEMU will not broadcast the SMI to all cores, causing only the BSP getting interrupted into MMI.

This change fixed the issue by invoking the negotiation routine and program the register to enable broadcasting.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

This was tested locally and verified that all cores will be coalesced in SMM.

## Integration Instructions

N/A


OvmfPkg/SmmControl2Dxe/MmControlPei.c
```diff
@@ -244,6 +245,13 @@ MmControlPeiEntryPoint (
     goto FatalError;
   }
 
+  //
+  // QEMU can inject SMIs in different ways, negotiate our preferences.
+  //
+  NegotiationSuccessful = NegotiateSmiFeatures ();
+
+  ASSERT (NegotiationSuccessful);
+
```
ardbiesheuvel, line 254: Why does unsuccessful negotiation trigger an ASSERT() failure but isn't considered a fatal error?

kuqin12: Invoking a single-core SMI may or may not lead to system malfunction, depending on how it is used. In the x64 STMM case (where this module is used), the system is still able to boot into the UEFI shell.

In addition, the existing DXE counterpart does not treat this as a fatal error:
https://github.com/tianocore/edk2/blob/b03a21a63e3bd001f52c527e5a57feddb53a690b/OvmfPkg/SmmControl2Dxe/SmmControl2Dxe.c#L247

So I’ve left this as an assertion to flag a potential integration issue, rather than enforcing a hard failure. That said, I’m open to changing this behavior if there is preference to a stricter handling.

kuqin12: Marking this as resolved. But please reopen it if the response is not adequate enough.

kuqin12:
@kraxel @mikebeaton @ardbiesheuvel @jyao1 would you mind reviewing the change? This will allow the QEMU to not only send SIPI from BSP. Any feedback is welcome. Thanks~

ardbiesheuvel:
Looks ok to me. @kraxel @niruiyu?

kraxel (approved):

cfernald (approved):

kuqin12:
@ardbiesheuvel @jyao1 do you have further feedback on this change? Otherwise, would you mind helping to merge the change? Thanks in advance!

PR #12571, opened 2026-05-12
Global: Replace manual alignment checks with helper macros

szsam:
# Description

Replace manual alignment checks with IS_ALIGNED() and
ADDRESS_IS_ALIGNED().

Convert the following bitmask and modulo forms:

- ((E & ((PowOf2Expr) - ONE)) == ZERO)
- ((E & ((PowOf2Expr) - ONE)) != ZERO)
- ((E % (PowOf2Expr)) == ZERO)
- ((E % (PowOf2Expr)) != ZERO)

to the corresponding helper macro forms:

+ IS_ALIGNED (E, PowOf2Expr)
+ !IS_ALIGNED (E, PowOf2Expr)

PowOf2Expr is limited to known power-of-two expressions, including
SIZE_* and BASE_* macros, EFI_PAGE_SIZE, CPU_STACK_ALIGNMENT,
RUNTIME_PAGE_ALLOCATION_GRANULARITY, sizeof() of UEFI integer types
(e.g. BOOLEAN, CHAR16, UINT32, UINTN) and pointer types, and 1 << E1
expressions.

~~For the bitmask form, replace `((E & (SZ - ONE)) == ZERO)` and
`!= ZERO` with `IS_ALIGNED (E, SZ)` for any expression `E` and `SZ`.
This assumes existing code already ensures that `SZ` is a power of 2.~~

~~For the modulo form, replace `((E % SZ) == ZERO)` and `!= ZERO` with
`IS_ALIGNED (E, SZ)` only for known edk2 constant macros that are
power of 2: `EFI_PAGE_SIZE` and `SIZE_[0-9]+[KMGTPE]B` defined in
`Base.h`.~~

Address checks that cast the checked value to UINTN are written with
ADDRESS_IS_ALIGNED().

This standardizes alignment checks, removes duplicated hand-written
bitmask logic, and makes pointer alignment checks explicit with the
address-specific helper.

Related commit: 94a50ef5

The change was generated with the Coccinelle semantic patch below.

Generated by coccinelle script:
```smpl
@power_of_2_expr@
expression PowOf2Expr;
expression E1;
typedef BOOLEAN, CHAR8, CHAR16, INT8, UINT8, INT16, UINT16, INT32, UINT32, INT64, UINT64, INTN, UINTN;
type ScalarType = { BOOLEAN, CHAR8, CHAR16, INT8, UINT8, INT16, UINT16, INT32, UINT32, INT64, UINT64, INTN, UINTN };
type AnyType;
type PointerType = AnyType *;
idexpression ScalarType ScalarValue;
idexpression PointerType PointerValue;
constant SizeBase =~ "^(SIZE|BASE)_(1|2|4|8|16|32|64|128|256|512)[KMGTPE]B$";
constant NamedPowerOf2 =~ "^(EFI_PAGE_SIZE|CPU_STACK_ALIGNMENT|RUNTIME_PAGE_ALLOCATION_GRANULARITY)$";
constant ONE = {1, 1U, 1u};
@@
(
(
  SizeBase
|
  NamedPowerOf2
|
  ONE << E1
|
  sizeof (ScalarType)
|
  sizeof (PointerType)
|
  sizeof (ScalarValue)
|
  sizeof (PointerValue)
)
&
PowOf2Expr
)

@depends on power_of_2_expr disable is_zero,isnt_zero@
expression E;
expression power_of_2_expr.PowOf2Expr;
constant ONE = {1, 1U, 1u};
constant ZERO = {0, 0U, 0u};
@@
(
  ((E & (E - ONE)) == ZERO)
|
- ((E & ((PowOf2Expr) - ONE)) == ZERO)
+ IS_ALIGNED (E, PowOf2Expr)
|
  ((E & (E - ONE)) != ZERO)
|
- ((E & ((PowOf2Expr) - ONE)) != ZERO)
+ !IS_ALIGNED (E, PowOf2Expr)
|
- ((E % (PowOf2Expr)) == ZERO)
+ IS_ALIGNED (E, PowOf2Expr)
|
- ((E % (PowOf2Expr)) != ZERO)
+ !IS_ALIGNED (E, PowOf2Expr)
)

@address_is_aligned@
typedef UINTN;
expression *Address;
expression Alignment;
@@
- IS_ALIGNED ((UINTN) Address, Alignment)
+ ADDRESS_IS_ALIGNED (Address, Alignment)

@normalize_aligned disable paren expression@
expression E, SZ;
@@
(
- (IS_ALIGNED (E, SZ))
+ IS_ALIGNED (E, SZ)
|
- (!IS_ALIGNED (E, SZ))
+ !IS_ALIGNED (E, SZ)
)

@normalize_macro_args disable paren expression@
expression E, SZ;
@@
(
- IS_ALIGNED ((E), SZ)
+ IS_ALIGNED (E, SZ)
|
- IS_ALIGNED (E, (SZ))
+ IS_ALIGNED (E, SZ)
)
```

- [ ] Breaking change
- [ ] Impacts security
- [ ] Includes tests

## How This Was Tested

CI

## Integration Instructions

N/A


os-d:
Please split this up into one commit per package. No reason to have a single commit here.

DynamicTablesPkg/Library/Acpi/Common/AcpiSsdtPcieLib/SsdtPcieGenerator.c
```diff
@@ -797,7 +797,7 @@ GeneratePciCrs (
 
       case PCI_SS_M32:
         ASSERT ((AddrMapInfo->PciAddress & ~MAX_UINT32) == 0);
-        ASSERT (((AddrMapInfo->PciAddress + AddrMapInfo->AddressSize - 1) & ~MAX_UINT32) == 0);
+        ASSERT (IS_ALIGNED (~MAX_UINT32, AddrMapInfo->PciAddress + AddrMapInfo->AddressSize));
```
ardbiesheuvel: Did you review these changes before posting them? Does this change make sense  to you?

Bitwise manipulation involving `constant - 1` only makes sense if `constant` is a power of two, and I don't think your script takes that into account, right?

szsam: Thank you for the feedback.

Yes, I reviewed the change before posting it.

I kept this change because, after expansion of the `IS_ALIGNED` macro,

```c
IS_ALIGNED (~MAX_UINT32, AddrMapInfo->PciAddress + AddrMapInfo->AddressSize)
```

becomes equivalent to:

```c
((AddrMapInfo->PciAddress + AddrMapInfo->AddressSize - 1) & ~MAX_UINT32) == 0
```

I do realize now that this change makes the code harder to understand.

For replacements involving modulo expressions, such as replacing:

```c
(E % SZ) == 0
```

with:

```c
IS_ALIGNED (E, SZ)
```

the script *only replaces `SZ` when it is one of the known power-of-two size constants*, such as `EFI_PAGE_SIZE` or `SIZE_[0-9]+[KMGTPE]B`.

However, for replacements of the bitmask form:

```c
(E & (SZ - 1)) == 0
```

with:

```c
IS_ALIGNED (E, SZ)
```

the script currently matches *arbitrary expressions for `SZ`*.

My reasoning was that the original bitmask form is only valid when `SZ` is a power of two. If `SZ` is not a power of two, then the existing code is already likely buggy, because `(SZ - 1)` would not be a valid alignment mask.

That said, I agree this assumption may still be too broad for an automated cleanup. *How about restricting the bitmask-form replacements to known EDK2 constants that are guaranteed to be powers of two, similar to what the script already does for modulo-form replacements?* This would also avoid replacing cases like:

```c
((AddrMapInfo->PciAddress + AddrMapInfo->AddressSize - 1) & ~MAX_UINT32) == 0
```

Looking forward to your thoughts.

ardbiesheuvel: > Thank you for the feedback.
> 
> Yes, I reviewed the change before posting it.
> 
> I kept this change because, after expansion of the `IS_ALIGNED` macro,
> 
> ```c
> IS_ALIGNED (~MAX_UINT32, AddrMapInfo->PciAddress + AddrMapInfo->AddressSize)
> ```
> 
> becomes equivalent to:
> 
> ```c
> ((AddrMapInfo->PciAddress + AddrMapInfo->AddressSize - 1) & ~MAX_UINT32) == 0
> ```
> 

The fact that `IS_ALIGNED (a, b)` performs `(a & (b - 1)) == 0` under the hood does not mean that we should replace every occurrence that matches that pattern, even if they have nothing to do with alignment.

This check is a convoluted way of ensuring that the address range in question is addressable with 32 bits. It has nothing to do with alignment, so it should not use this macro.
...
> However, for replacements of the bitmask form:
> 
> ```c
> (E & (SZ - 1)) == 0
> ```
> 
> with:
> 
> ```c
> IS_ALIGNED (E, SZ)
> ```
> 
> the script currently matches _arbitrary expressions for `SZ`_.
> 
> My reasoning was that the original bitmask form is only valid when `SZ` is a power of two. If `SZ` is not a power of two, then the existing code is already likely buggy, because `(SZ - 1)` would not be a valid alignment mask.
> 

The code is only buggy if your assumption is that the code is performing an alignment check, but this is not always the case when the pattern `(a & (b - 1)) == 0` is used.

> That said, I agree this assumption may still be too broad for an automated cleanup. _How about restricting the bitmask-form replacements to known EDK2 constants that are guaranteed to be powers of two, similar to what the script already does for modulo-form replacements?_ This would also avoid replacing cases like:
> 
> ```c
> ((AddrMapInfo->PciAddress + AddrMapInfo->AddressSize - 1) & ~MAX_UINT32) == 0
> ```
> 
> Looking forward to your thoughts.

Yes that makes more sense.

And please split this up in per-package patches, as @os-d suggested. And do not close this PR and reopen another one for that: use Git to update your underlying branch, and the changes will be reflected in the PR immediately.

szsam:
I made the replacement more restrictive: `(E & ((PowOf2Expr) - ONE))` or `(E % PowOf2Expr)` is replaced with `IS_ALIGNED` only when `PowOf2Expr` is guaranteed to be a power of 2. This includes `SIZE_*` and `BASE_*` macros; `EFI_PAGE_SIZE`, `CPU_STACK_ALIGNMENT`, and `RUNTIME_PAGE_ALLOCATION_GRANULARITY`; `sizeof()` expressions for UEFI integer types, such as `BOOLEAN`, `CHAR16`, `UINT32`, and `UINTN`, as well as pointer types; and `1 << E1` expressions.

Also, I have split the PR into one commit per package.

@os-d @ardbiesheuvel Please review.

os-d (approved):
Approved for EmulatorPkg.

lgao4 (approved):

lgao4:
The changes in MdePkg and MdeModulePkg are good to me.

kilaterlee (approved):
Approved for LoongArch64*

ardbiesheuvel (approved):

PR #12572, opened 2026-05-13
MdeModulePkg/PciBusDxe: Scan funcs when func 0 is absent

MarsDoge:
# Description        
## Summary                                                                                                                                          
- Add `PcdPciScanFuncIfFunc0Absent`, default `FALSE`.
- Let PciBusDxe continue collecting device information for functions 1..7 when function 0 is absent and the PCD is enabled.
- Preserve existing platform behavior by default.

## Background
Some virtualized PCI topologies may expose selected non-zero functions to a guest while function 0 is hidden. Platforms that require this behavior can opt in through the new Feature PCD.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Tested on a physical platform that exposes a multi-function PCI device
where function 0 is intentionally unavailable after its clock is disabled,
while function 1 is still required by the platform.
    
With the new PCD enabled, PciBusDxe continued scanning functions 1..7
after function 0 was not detected, and the required function 1 device was
discovered during PCI enumeration.

    Built OVMF with the new PCD disabled and enabled:
    
    bash
    build -p OvmfPkg/OvmfPkgX64.dsc -a X64 -b DEBUG -t GCC \
      --pcd PcdPciScanFuncIfFunc0Absent=FALSE
    
    build -p OvmfPkg/OvmfPkgX64.dsc -a X64 -b DEBUG -t GCC \
      --pcd PcdPciScanFuncIfFunc0Absent=TRUE
    
    
    Booted OVMF with QEMU using an e1000e device placed at function 1 while
    function 0 was absent:
    
    bash
    qemu-system-x86_64 \
      -machine q35,accel=tcg \
      -m 512M \
      -nodefaults \
      -display none \
      -monitor none \
      -no-reboot \
      -bios Build/OvmfX64/DEBUG_GCC/FV/OVMF.fd \
      -debugcon file:/tmp/qemu-pci-func-test/ovmf-debug.log \
      -global isa-debugcon.iobase=0x402 \
      -device e1000e,bus=pcie.0,addr=0x5.1,multifunction=on
    
    
    QEMU reported:
    
    text
    Bus  0, device   5, function 1:
      Ethernet controller: PCI device 8086:10d3
    
    
    With the PCD set to FALSE, OVMF did not discover [00|05|01].
    With the PCD set to TRUE, OVMF discovered it:
    
    text
    PciBus: Discovered PCI @ [00|05|01]  [VID = 0x8086, DID = 0x10D3]

## Integration Instructions
Platforms that need to discover valid non-zero PCI functions when function 0 is absent can enable the new Feature PCD in their DSC.

### DIFF
--- a/MdeModulePkg/Bus/Pci/PciBusDxe/PciBusDxe.inf
+++ b/MdeModulePkg/Bus/Pci/PciBusDxe/PciBusDxe.inf
@@ -90,6 +90,7 @@
   gEfiMdeModulePkgTokenSpaceGuid.PcdPciBridgeIoAlignmentProbe       ## CONSUMES
   gEfiMdeModulePkgTokenSpaceGuid.PcdUnalignedPciIoEnable            ## CONSUMES
   gEfiMdeModulePkgTokenSpaceGuid.PcdPciDegradeResourceForOptionRom  ## CONSUMES
+  gEfiMdeModulePkgTokenSpaceGuid.PcdPciScanFuncIfFunc0Absent       ## CONSUMES
 
 [Pcd]
   gEfiMdeModulePkgTokenSpaceGuid.PcdSrIovSystemPageSize         ## SOMETIMES_CONSUMES
--- a/MdeModulePkg/Bus/Pci/PciBusDxe/PciEnumeratorSupport.c
+++ b/MdeModulePkg/Bus/Pci/PciBusDxe/PciEnumeratorSupport.c
@@ -266,10 +266,21 @@ PciPciDeviceInfoCollector (
                  );
 
       if (EFI_ERROR (Status) && (Func == 0)) {
+        if (!FeaturePcdGet (PcdPciScanFuncIfFunc0Absent)) {
+          //
+          // Preserve default behavior for physical platforms: go to next
+          // device if there is no Function 0.
+          //
+          break;
+        }
+
         //
-        // go to next device if there is no Function 0
+        // Some virtualized PCI topologies may let a hypervisor expose
+        // selected non-zero functions to a guest while Function 0 is absent.
+        // Some platforms may require probing such functions.
+        // Keep scanning only when enabled by platform policy.
         //
-        break;
+        continue;
       }
 
       if (!EFI_ERROR (Status)) {
--- a/MdeModulePkg/MdeModulePkg.dec
+++ b/MdeModulePkg/MdeModulePkg.dec
@@ -882,6 +882,12 @@
   # @Prompt Enable PCI bridge IO alignment probe.
   gEfiMdeModulePkgTokenSpaceGuid.PcdPciBridgeIoAlignmentProbe|FALSE|BOOLEAN|0x0001004e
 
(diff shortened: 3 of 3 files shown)


apop5:
I don't agree with this change. I understand you are attempting to gate it through a PCD, but the change is against the PCI specification:

PCI Local Bus Specification 3.0 (I know its old, but it still applies)

3.2.2.3.4. Selection of a Device’s Configuration Space

`They are also required to always implement function 0 in the device.`

MarsDoge:
Thanks for the review. @apop5
    
I agree that for standard PCI enumeration, function 0 is required. The PCI
specification is clear about that, and this patch is not intended to change
the default PCI-compliant behavior.
    
The use case here is a non-standard/platform-specific topology where
firmware knows that function 0 may be intentionally unavailable, while
another function is still valid and required by the platform.
    
Linux has a similar compatibility path for this class of situation. In
drivers/pci/probe.c::pci_scan_slot(), Linux normally stops scanning a slot
when function 0 is absent, but it explicitly does not break when
hypervisor_isolated_pci_functions() is true:
  
```c  
      /*
       * Function 0 is required unless we are running on a hypervisor
       * that passes through individual PCI functions.
       */
      if (!dev && fn == 0 && !hypervisor_isolated_pci_functions())
              break;
```

Reference:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/pci/probe.c?id=70eda68668d1476b459b64e69b8f36659fa9dfa8#n2871
        
So I am not arguing that a missing function 0 is PCI-spec compliant. The
point is that there are real environments where a platform/hypervisor may
intentionally expose individual PCI functions, and a strict scan would miss
a valid non-zero function.
    
This patch keeps the default behavior unchanged by making the new PCD
default to FALSE. Only platforms that explicitly opt in get the relaxed
scanning behavior.

MarsDoge:
I understand @apop5's PCI spec concern, and I am not trying to change the default enumeration behavior. I am only trying to understand whether a default-disabled platform opt-in is acceptable for this kind of non-standard topology.
    
Linux has a similar exception path for unusual PCI function exposure: hypervisor_isolated_pci_functions() allows pci_scan_slot() to continue after Function 0 is missing. There is also pci=pcie_scan_all / PCI_SCAN_ALL_PCIE_DEVS for broader PCIe device scanning. I mention it only as another example of the default-strict / explicit-opt-in model.
    
For edk2, my intent is similar: keep the default behavior strict, and only allow a platform that explicitly opts in to use the relaxed scan behavior.
    
If this policy does not belong in generic PciBusDxe, I would appreciate guidance on where this platform-specific scan policy should live.
    
@jkolakow @lgao4 Could you also share your thoughts on this direction?

lgao4 (approved):

lgao4:
The change is good to me.

MarsDoge:
Hi, @lgao4  I just rebased to the latest baseline. Is this PR ready to be merged ?

PR #12573, opened 2026-05-13
 OvmfPkg,SecurityPkg: Document explicitly the referenced TCG Spec. revisions

quite:
# Description

Where sections in TCG Specification documents are referenced, be explicit about which revision of the document that is referenced.

This is a pure change to documentation, to the comments in the C code.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

### DIFF
--- a/OvmfPkg/Tcg/TdTcg2Dxe/TdTcg2Dxe.c
+++ b/OvmfPkg/Tcg/TdTcg2Dxe/TdTcg2Dxe.c
@@ -316,7 +316,7 @@ InternalDumpHex (
 
   This function initialize TD_EVENT_HDR for EV_NO_ACTION
   Event Type other than EFI Specification ID event. The behavior is defined
-  by TCG PC Client PFP Spec. Section 9.3.4 EV_NO_ACTION Event Types
+  by TCG PC Client PFP Spec. 00.21 (2016) Section 9.3.4 EV_NO_ACTION Event Types
 
   @param[in, out]   NoActionEvent  Event Header of EV_NO_ACTION Event
   @param[in]        EventSize      Event Size of the EV_NO_ACTION Event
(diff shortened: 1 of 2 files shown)


quite:
the failing check indicates the following is an issue? not sure how to overcome um
```
 * OvmfPkg
 * SecurityPkg
```

jyao1 (approved):

jyao1 (changes requested):
I am OK to add spec version/revision.

But I have some opens:
1) About version indication.
It should be full version/revision.
See https://trustedcomputinggroup.org/resource/tcg-efi-protocol-specification/, and https://trustedcomputinggroup.org/resource/pc-client-specific-platform-firmware-profile-specification/.
Only adding revision is weird.

2) About version selection.
I am not sure why you choose "TCG PC Client PFP Spec. 00.21 (2016)".
If you look at PcdTcgPfpMeasurementRevision, it indicates more.

3) About update completeness.
We already have PFP version description in Tcg2Dxe.inf.
I prefer we update that as well, and ensure both are synced.

quite:
I originally wanted to read where and when measurements were done, and I came across the line `// TCG PC Client PFP spec Section 2.4.4.5 Step 4` in Tcg2Dxe.c. The only spec pdf where I could find a section `2.4.4.5` (which also made sense in the context) was "00.21 (2016)". Also in the same file I saw `according to TCG PC Client PFP spec 0021 Section 2.4.4.2`. So I thought I could just built upon that, and improve the references slightly for the person who comes next and wants to follow them.

(It also looks to me like back in 2016, the spec was not written to say explicitly "Version 1.0", but instead "Level 00 Revision 21" -- maybe "Level 00" implies "version 1.0". But I now saw that in the errata for this spec, it does say "Version 1.0; Revision 0.21", so I guess the spec was "Version 1.0".)

But it seems to be a fair bit more complicated than I first thought. It's not just that the section references have become outdated (and are missing which spec document aversion/revision they point into), but these would also need to be maintained when edk2 is updated to for a new spec version. I'm sorry, I think this I'm not able to get on top of a good way to solve this at this time, and sorry for wasting your time mhm.

PR #12577, opened 2026-05-14
ManageabilityPkg/KcsLib: Fix KCS Read Phase missing dummy byte and incorrect size check

panjinbao-void:
## Summary

Two bug fixes in ManageabilityTransportKcsLib that cause KCS interface to get stuck in READ_STATE, resulting in BMC NOT_READY failures during continuous IPMI command sequences.

### Bug 1: Missing dummy byte read leaves KCS in READ_STATE

When `ResponseBufferSize == actual BMC response size`, the read loop exits via `ReadLength >= *Length` without completing the IPMI Spec Figure 9-7 handshake. The host never reads the final dummy byte at IDLE_STATE, so the KCS interface stays stuck in READ_STATE. The next IPMI command then sees a non-idle interface and fails with `EFI_NOT_READY`.

This is extremely obvious when sending multiple IPMI commands back to back.

### Bug 2: Incorrect response size equality check

`KcsTransportSendCommand()` saves `*ResponseDataSize` (buffer capacity) before calling `KcsTransportRead()`, then returns `EFI_DEVICE_ERROR` if actual bytes read != buffer capacity. Since IPMI responses are variable-length, most commands return fewer bytes than the buffer can hold — this check incorrectly rejects valid responses.

When this check triggers, it also causes the host to skip the read-phase drain (Bug 1), compounding the NOT_READY failure.

## Test Plan

- Tested on SpaceMit V100 EVB platform with BMC over KCS
- Verified IPMI commands succeed when response is shorter than buffer
- Verified KCS interface returns to IDLE_STATE after buffer-full reads
- Confirmed no regression in normal IPMI command flow (Get Device ID, Get SEL Info, etc.)
- Continuous IPMI command sequences no longer trigger NOT_READY

Signed-off-by: Pan Jinbao <jinbao.pan@spacemit.com>

### DIFF
--- a/ManageabilityPkg/Library/ManageabilityTransportKcsLib/KcsCommon.c
+++ b/ManageabilityPkg/Library/ManageabilityTransportKcsLib/KcsCommon.c
@@ -685,7 +685,6 @@ KcsTransportSendCommand (
 {
   EFI_STATUS  Status;
   UINT8       *RspHeader;
-  UINT32      ExpectedResponseDataSize;
 
   if ((RequestData != NULL) && (RequestDataSize == 0)) {
     DEBUG ((DEBUG_ERROR, "%a: Mismatched values of RequestData and RequestDataSize\n", __func__));
@@ -759,24 +758,13 @@ KcsTransportSendCommand (
 
     FreePool (RspHeader);
 
-    ExpectedResponseDataSize = *ResponseDataSize;
-    Status                   = KcsTransportRead (ResponseData, ResponseDataSize);
+    Status = KcsTransportRead (ResponseData, ResponseDataSize);
     if (EFI_ERROR (Status)) {
       DEBUG ((DEBUG_ERROR, "KCS response read Failed with Status(%r)\n", Status));
     }
 
     // Print out the response payloads.
     if (*ResponseDataSize != 0) {
-      if (ExpectedResponseDataSize != *ResponseDataSize) {
-        DEBUG ((
-          DEBUG_ERROR,
-          "Expected KCS response size : %d is not matched to returned size : %d.\n",
-          ExpectedResponseDataSize,
-          *ResponseDataSize
-          ));
-        return EFI_DEVICE_ERROR;
-      }
-
       HelperManageabilityDebugPrint ((VOID *)ResponseData, (UINT32)*ResponseDataSize, "KCS Response Data:\n");
       Status = KcsCheckResponseData (ResponseData, *ResponseDataSize, AdditionalStatus);
     } else {
--- a/ManageabilityPkg/Library/ManageabilityTransportKcsLib/KcsCommon.c
+++ b/ManageabilityPkg/Library/ManageabilityTransportKcsLib/KcsCommon.c
@@ -383,6 +383,47 @@ KcsTransportRead (
     }
   }
 
+  //
+  // Per IPMI 2.0 Spec Figure 9-7, drain remaining bytes until
+  // BMC transitions to IDLE_STATE.
+  //
+  while (TRUE) {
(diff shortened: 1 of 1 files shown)


ManageabilityPkg/Library/ManageabilityTransportKcsLib/KcsCommon.c
```diff
@@ -383,6 +383,47 @@ KcsTransportRead (
     }
   }
 
+  //
```
changab, line 386: Hi @panjinbao-void,
I dont see the figure 9-7 you mentioned in IPMI spec 2.0 to drain the remaining bytes.  Could you please point me to the right figure? Or screenshot the flow and paste it here to help with the review. Thanks.

Also, @abdattar is verifying this change on our platform too.

panjinbao-void: Hi @changab, thanks for the review.
Apologies for the incorrect figure number – the figures in Section 9.15 of the IPMI v2.0 rev 1.1 spec are unnumbered (all printed as "Figure 9-, " with no digit after the dash). I was counting them sequentially.
The "drain remaining bytes to IDLE" logic refers to the flowchart titled "KCS Interface BMC to SMS Read Transfer Flow Chart" (Section 9.15, 2nd diagram). When the state becomes IDLE_STATE, the host must wait for BF-E/- Read to finish draining data bytes from the data buffer. This is the step where the interface buffer was already full, leaving the interface stuck in READ_STATE.

ManageabilityPkg/Library/ManageabilityTransportKcsLib/KcsCommon.c
```diff
@@ -383,6 +383,47 @@ KcsTransportRead (
     }
   }
 
+  //
+  // Per IPMI 2.0 Spec Figure 9-7, drain remaining bytes until
+  // BMC transitions to IDLE_STATE.
+  //
+  while (TRUE) {
```
changab, line 390: But is this already handled in line 368?

panjinbao-void: The handling at line 368 only triggers when the BMC transitions to IDLE_STATE while ReadLength < *Length (i.e., actual response is shorter than the buffer).
The new drain loop handles the distinct case where ReadLength == *Length — the while loop exits by its boundary condition before seeing IDLE_STATE. At that point the BMC is still in READ_STATE, and without draining, the KCS interface is left stuck.
So the two paths are:
- Actual response shorter than buffer → existing idle check at ~line 368 handles it
- Actual response equal to buffer capacity → loop exits early, existing code never reaches idle check, new drain loop handles it

changab: Ok, I see. Thanks for the explanation. Please check my comment below.

ManageabilityPkg/Library/ManageabilityTransportKcsLib/KcsCommon.c
```diff
@@ -383,6 +383,47 @@ KcsTransportRead (
...
+      if (EFI_ERROR (Status)) {
+        *Length = ReadLength;
+        return Status;
+      }
+
+      KcsRegisterWrite8 (KCS_REG_DATA_OUT, IPMI_KCS_CONTROL_CODE_READ);
+    } else {
+      *Length = ReadLength;
+      return EFI_DEVICE_ERROR;
+    }
```
changab, line 424: According to the diagram, we should consider both IpmiKcsReadState and the "else values" as the error condition. I think we should follow the Error Exit process, right? If so, could you please also provide a function to process the error exit? Thanks!

PR #12582, opened 2026-05-14
MdePkg: Coerce MSVC/CLANGPDB AARCH64 to Use AAPCS64 ABI for VA_LIST

os-d:
# Description

MSVC AARCH64 and CLANGPDB AARCH64 currently use the MS ABI for variadics in edk2. However, this breaks the AAPCS64 ABI and interopability with GCC/CLANGDWARF built binaries.

Because the architecture defines the single ABI, MSVC AARCH64 has little usage, and CLANGPDB AARCH64 is new to edk2, coerce these toolchains to use the AAPCS64 ABI instead of the MS ABI. This allows interopability between all supported AARCH64 toolchains in edk2 for variadics.

Note that variadic functions themselves are unaffected by this change, e.g. functions that take ... as the last parameter. This affects the VA_LIST structure and the associated helpers to parse it.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested in two ways:
- With the original failing case in CryptoPkg that passes a VA_LIST across a protocol. When one side is built with GCC and one with CLANGPDB, this was broken. After this fix it works.
- With two test drivers that communicate via protocol, calling variadic functions, calling functions that take VA_LIST as a parameter, and functions that take VA_LIST* as a parameter. Inputs/outputs were validated they were correct. All four cases of producer/consumer built with GCC/CLANGPDB were tested and all passed. Prior to the change all had failed.

## Integration Instructions

N/A.

### DIFF
--- a/.pytool/Plugin/SpellCheck/cspell.base.yaml
+++ b/.pytool/Plugin/SpellCheck/cspell.base.yaml
@@ -27,6 +27,7 @@
     "ignoreWords": [],
     "words": [
         "AABBCCDD",
+        "AAPCS",
         "ABPID",
         "ACPIBAR",
         "ACPICA",
--- a/MdePkg/Include/Base.h
+++ b/MdePkg/Include/Base.h
@@ -579,14 +579,27 @@ struct _LIST_ENTRY {
 
 #if defined (_M_ARM64)
 //
-// MSFT ARM variable argument list support.
+// MSVC AARCH64 and CLANGPDB AARCH64 hit this path. By default, they use the MS ABI for VA_LIST and
+// friends. However, this is against the AAPCS64 ABI and cause interop issues with GCC and CLANGDWARF.
+// These definitions coerce MSVC/CLANGPDB to use the AAPCS64 ABI.
 //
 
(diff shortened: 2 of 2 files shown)


os-d:
@ardbiesheuvel I can't add you as a reviewer because my write privileges are revoked during the hard freeze. This can wait until after that, btw, I don't think it is urgent to fix because of the low usage of passing VA_LIST or VA_LIST* over an ABI boundary.

ardbiesheuvel (approved):
Note that VA_LIST is not defined in the PI or UEFI specifications, and so any use of VA_LIST on an API boundary is EDK2 specific, but it seems that cat is already out of the bag.

Javagedes (approved):

ardbiesheuvel:
Let's go ahead with this, although I'd strongly suggest to avoid `VA_LIST` parameters on `EFIAPI` boundaries at all cost.

os-d:
> Let's go ahead with this, although I'd strongly suggest to avoid `VA_LIST` parameters on `EFIAPI` boundaries at all cost.

Agreed.

PR #12589, opened 2026-05-18
ArmVirtPkg: Factorize common modules in ArmVirtPkg.dsc.inc

pierregondois:
# Description

Factorize common modules in ArmVirtPkg.dsc.inc.

No functional change except for Kvmtool which changes the Metronome module implementation.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Ran Kvmtool and CloudHv

### DIFF
--- a/ArmVirtPkg/ArmVirt.dsc.inc
+++ b/ArmVirtPkg/ArmVirt.dsc.inc
@@ -85,6 +85,7 @@ DEFINE FD_SIZE_IN_MB    = 3
   # Ramdisk Requirements
   #
   FileExplorerLib|MdeModulePkg/Library/FileExplorerLib/FileExplorerLib.inf
+  FrameBufferBltLib|MdeModulePkg/Library/FrameBufferBltLib/FrameBufferBltLib.inf
 
   # Allow dynamic PCDs
   #
@@ -141,6 +142,9 @@ DEFINE FD_SIZE_IN_MB    = 3
 
   # PCI Libraries
   PciLib|MdePkg/Library/BasePciLibPciExpress/BasePciLibPciExpress.inf
+  PciSegmentLib|MdePkg/Library/BasePciSegmentLibPci/BasePciSegmentLibPci.inf
+  PciHostBridgeLib|OvmfPkg/Fdt/FdtPciHostBridgeLib/FdtPciHostBridgeLib.inf
+  PciPcdProducerLib|OvmfPkg/Fdt/FdtPciPcdProducerLib/FdtPciPcdProducerLib.inf
   PciExpressLib|OvmfPkg/Library/BaseCachingPciExpressLib/BaseCachingPciExpressLib.inf
   PciCapLib|OvmfPkg/Library/BasePciCapLib/BasePciCapLib.inf
   PciCapPciSegmentLib|OvmfPkg/Library/BasePciCapPciSegmentLib/BasePciCapPciSegmentLib.inf
--- a/ArmVirtPkg/ArmVirt.dsc.inc
+++ b/ArmVirtPkg/ArmVirt.dsc.inc
@@ -235,6 +235,9 @@ DEFINE FD_SIZE_IN_MB    = 3
 [LibraryClasses.common.UEFI_DRIVER]
   # resolve RngLib via RngDxe's EFI_RNG_PROTOCOL for UEFI_DRIVER type drivers such as TlsDxe
   RngLib|MdePkg/Library/DxeRngLib/DxeRngLib.inf
+  ExtractGuidedSectionLib|MdePkg/Library/DxeExtractGuidedSectionLib/DxeExtractGuidedSectionLib.inf
+  PerformanceLib|MdeModulePkg/Library/DxePerformanceLib/DxePerformanceLib.inf
+  MemoryAllocationLib|MdePkg/Library/UefiMemoryAllocationLib/UefiMemoryAllocationLib.inf
 
 [LibraryClasses.common.DXE_CORE]
   HobLib|MdePkg/Library/DxeCoreHobLib/DxeCoreHobLib.inf
@@ -251,11 +254,6 @@ DEFINE FD_SIZE_IN_MB    = 3
   PerformanceLib|MdeModulePkg/Library/DxePerformanceLib/DxePerformanceLib.inf
   MemoryAllocationLib|MdePkg/Library/UefiMemoryAllocationLib/UefiMemoryAllocationLib.inf
   HiiLib|MdeModulePkg/Library/UefiHiiLib/UefiHiiLib.inf
-
-[LibraryClasses.common.UEFI_DRIVER]
-  ExtractGuidedSectionLib|MdePkg/Library/DxeExtractGuidedSectionLib/DxeExtractGuidedSectionLib.inf
-  PerformanceLib|MdeModulePkg/Library/DxePerformanceLib/DxePerformanceLib.inf
-  MemoryAllocationLib|MdePkg/Library/UefiMemoryAllocationLib/UefiMemoryAllocationLib.inf
 
 [LibraryClasses.common.DXE_RUNTIME_DRIVER]
   MemoryAllocationLib|MdePkg/Library/UefiMemoryAllocationLib/UefiMemoryAllocationLib.inf
--- a/ArmVirtPkg/ArmVirt.dsc.inc
+++ b/ArmVirtPkg/ArmVirt.dsc.inc
@@ -433,6 +433,17 @@ DEFINE FD_SIZE_IN_MB    = 3
   }
 
   #
+  # Multiple Console IO support
+  #
+  MdeModulePkg/Universal/Console/ConPlatformDxe/ConPlatformDxe.inf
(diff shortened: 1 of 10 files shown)


ArmVirtPkg/ArmVirtKvmTool.dsc
```diff
@@ -272,7 +272,6 @@
   }
   MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciHostBridgeDxe.inf {
     <LibraryClasses>
-      NULL|OvmfPkg/Fdt/FdtPciPcdProducerLib/FdtPciPcdProducerLib.inf
```
ardbiesheuvel, line 376: What about the DEPEX ?

ArmVirtPkg/ArmVirt.dsc.inc
```diff
@@ -464,6 +464,19 @@ DEFINE FD_SIZE_IN_MB    = 3
   OvmfPkg/VirtioFsDxe/VirtioFsDxe.inf
 !endif
 
+  #
+  # Platform Driver
+  #
+  EmbeddedPkg/Drivers/FdtClientDxe/FdtClientDxe.inf
+!if $(PLAT_XEN) == FALSE
```
leiflindholm: This may be logically true, but would it not be more clear to enumerate the == TRUE cases?

Alternatively, stacking variables like (in ArmVirt.dsc.inc):
```
!if $(PLAT_XEN) == FALSE
  DEFINE ENABLE_VIRTIO = TRUE
!else
  DEFINE ENABLE_VIRTIO = FALSE
!endif
```
and then filter virtio support on that?

ArmVirtPkg/ArmVirtKvmTool.dsc
```diff
@@ -241,7 +241,7 @@
       NULL|ArmVirtPkg/Library/NorFlashKvmtoolLib/NorFlashKvmtoolLib.inf
   }
 
-  MdeModulePkg/Universal/Metronome/Metronome.inf
+  EmbeddedPkg/MetronomeDxe/MetronomeDxe.inf
```
leiflindholm: Why this change rather than the opposite?

ArmVirtPkg/ArmVirtKvmTool.dsc
pierregondois: `FdtPciPcdProducerLib.inf` waits for `gFdtClientProtocolGuid` to be published, but it doesn't seem there is any other chain of dependency regarding `PciHostBridgeDxe.inf`.
When testing Kvmtool, everything was fine.

ArmVirtPkg/ArmVirt.dsc.inc
pierregondois: Yes right indeed.
Your comment doesn't apply to the following pattern where a subset of components is not loaded right ?

```
#
# Multiple Console IO support
#
MdeModulePkg/Universal/Console/ConPlatformDxe/ConPlatformDxe.inf
MdeModulePkg/Universal/Console/TerminalDxe/TerminalDxe.inf
MdeModulePkg/Universal/SerialDxe/SerialDxe.inf
!if $(PLAT_XEN) == FALSE
MdeModulePkg/Universal/Console/ConSplitterDxe/ConSplitterDxe.inf
MdeModulePkg/Universal/Console/GraphicsConsoleDxe/GraphicsConsoleDxe.inf
!endif
```

ArmVirtPkg/ArmVirtKvmTool.dsc
pierregondois: It was easier to only test KvmTool.
`MdeModulePkg/Universal/Metronome/Metronome.inf` has a thiner granularity, but I'm not sure it is actually relevant for virtual platforms.

ArmVirtPkg/ArmVirtKvmTool.dsc
ardbiesheuvel: Yes, so by removing `FdtPciPcdProducerLib.inf` here, `PciHostBridgeDxe.inf` no longer inherits the DEPEX on `gFdtClientProtocolGuid`, and so it may be dispatched before `FdtClientDxe`

ArmVirtPkg/ArmVirt.dsc.inc
leiflindholm: Not that that couldn't be useful also, but no - those could probably be considered in future if it grows in complexity or affected platforms.

The virtio case kind of stood out in the patch.

ArmVirtPkg/ArmVirtKvmTool.dsc
leiflindholm: Well, the reason I'm asking is I don't really understand why we still have the EmbeddedPkg one at all. Did we at one point not have a NanoSecondDelay on Arm and decided this was the way around it?

So if we're picking one over the other, I would prefer not picking the "but on Arm" one, such that it will be easier to drop that one in the future.

ArmVirtPkg/ArmVirtKvmTool.dsc
pierregondois: I'm not sure sure I understand why `PciHostBridgeDxe.inf` itself needs to wait for `gFdtClientProtocolGuid`.

There is however an implicit dependency chain over it as:
- `MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciHostBridgeDxe.inf` needs `gEfiCpuIo2ProtocolGuid` (DEPEX)
- `gEfiCpuIo2ProtocolGuid` is produced by `UefiCpuPkg/CpuMmio2Dxe/CpuMmio2Dxe.inf`
- `UefiCpuPkg/CpuMmio2Dxe/CpuMmio2Dxe.inf` relies on `gEfiMdePkgTokenSpaceGuid.PcdPciIoTranslation`
- `gEfiMdePkgTokenSpaceGuid.PcdPciIoTranslation` is populated by `OvmfPkg/Fdt/FdtPciPcdProducerLib/FdtPciPcdProducerLib.inf` (per-component lib)

ArmVirtPkg/ArmVirtKvmTool.dsc
pierregondois: I m not sure whether this was ironic or not, but indeed the ArmPkg TimerLib doesn't seem to implement NanoSecondDelay() at a ns precision:
https://github.com/tianocore/edk2/blob/b03a21a63e3bd001f52c527e5a57feddb53a690b/ArmPkg/Library/ArmArchTimerLib/ArmArchTimerLib.c#L104

KvmTool/CloudHv have a system counter freq of 1GHz, so this perfectly fits the 1ns resolution.

pierregondois: I can add a patch to change `NanoSecondDelay()` and move all ArmVirtPkg platforms to `MdeModulePkg/Universal/Metronome/Metronome.inf` if it suits everybody

ArmVirtPkg/ArmVirtKvmTool.dsc
ardbiesheuvel: > I'm not sure sure I understand why `PciHostBridgeDxe.inf` itself needs to wait for `gFdtClientProtocolGuid`.
> 

Because the DTB tells us where the host bridge is, and how it needs to be configured.

ArmVirtPkg/ArmVirtKvmTool.dsc
leiflindholm: > I m not sure whether this was ironic or not, but indeed the ArmPkg TimerLib doesn't seem to implement NanoSecondDelay() at a ns precision:

Not in the least ironic.
Who cares what granularity it actually implements? The function exists, and using the MdeModulePkg Metronome with ArmPkg TimerLib doesn't end up with any worse resolution than using the EmbeddedPkg one with the same TimerLib. And other TimerLib implementations (or future improvements to ArmPkg TimerLib) could improve resolution.

The 1GHz resolution is also an SBSA requirement. (which does not imply an ability to actually wait for a single ns)

ArmVirtPkg/ArmVirtKvmTool.dsc
pierregondois: Yes right. Just to be sure, the dependency over `gFdtClientProtocolGuid` is:
- respected (cf. above)
- somehow indirect: the module doesn't actually wait for `gFdtClientProtocolGuid` itself, but rather for a protocol that depends on `gFdtClientProtocolGuid`

Also, other ArmVirtPkg platforms don't describe this dependency over `gFdtClientProtocolGuid`.
Is the patch ok, or are there some concerns I didn't get ?

ArmVirtPkg/ArmVirtKvmTool.dsc
ardbiesheuvel: > I can add a patch to change `NanoSecondDelay()` and move all ArmVirtPkg platforms to `MdeModulePkg/Universal/Metronome/Metronome.inf` if it suits everybody

Yes, please.

pierregondois:
Gentle ping if there is any new review

leiflindholm (approved):

PR #12590, opened 2026-05-18
MdeModulePkg: GCD: Use Alignment Requirement In Bin Size Calc

os-d:
# Description

This PR consists of two commits:

MdeModulePkg: GCD: Use Alignment Requirement in Bin Size Calc
--
Currently CalculateTotalMemoryBinSizeNeeded() does not take runtime alignment granularity considerations into account. This
means that the GCD initialization code can choose resource desc HOBs to use for the bin region that are actually too small and
fail to initialize the bins.

This fixes CalculateTotalMemoryBinSizeNeeded() to take the alignment requirements into consideration, both for size and for alignment of the bin address range.

MdeModulePkg: Dxe Core: Use CalculateTotalMemoryBinSizeNeeded()
--
CoreSetMemoryTypeInformationRange() currently calculates the bin size needed independently from the CalculateTotalMemoryBinSizeNeeded() function. This commit updates to use that function and remove the duplication.

Note: This PR was split off from #12086 and is required to merge before it and the second PR #12591 split off from it.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested with the memory bins PR #12086 and ensured bins were calculated correctly and passed from PEI to DXE.

## Integration Instructions

N/A.

### DIFF
--- a/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
+++ b/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
@@ -135,25 +135,6 @@ GLOBAL_REMOVE_IF_UNREFERENCED CONST CHAR8  *mGcdAllocationTypeNames[] = {
 };
 
 /**
-  Calculate total memory bin size needed.
-
-  @param BinTop                The top address of the memory bins. This is an optional parameter.
-                               When NULL, the returned size meets the alignment requirements as long as
-                               the base address selected also meets the alignment requirements. When
-                               non-NULL, then the returned BinTop value and the returned size both meet
-                               the alignment requirements. When non-NULL, this will be updated on
-                               output to the new top address of the memory bins that must be used to
-                               satisfy alignment requirements.
-
-  @return The total memory bin size needed.
-
-**/
-UINT64
-CalculateTotalMemoryBinSizeNeeded (
-  IN OUT OPTIONAL EFI_PHYSICAL_ADDRESS  *BinTop
-  );
-
-/**
   Dump the entire contents if the GCD Memory Space Map using DEBUG() macros when
   PcdDebugPrintErrorLevel has the DEBUG_GCD bit set.
 
--- a/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
+++ b/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
@@ -135,6 +135,25 @@ GLOBAL_REMOVE_IF_UNREFERENCED CONST CHAR8  *mGcdAllocationTypeNames[] = {
 };
 
 /**
+  Calculate total memory bin size needed.
+
+  @param BinTop                The top address of the memory bins. This is an optional parameter.
+                               When NULL, the returned size meets the alignment requirements as long as
+                               the base address selected also meets the alignment requirements. When
+                               non-NULL, then the returned BinTop value and the returned size both meet
+                               the alignment requirements. When non-NULL, this will be updated on
+                               output to the new top address of the memory bins that must be used to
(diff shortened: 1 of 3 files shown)


os-d:
@mdkinney this is the first PR split off from the memory bins and the one that will need to be merged first. I cannot add you as a reviewer because my write access is revoked during the hard freeze.

MdeModulePkg/Core/Dxe/Gcd/Gcd.c
```diff
@@ -2165,25 +2165,63 @@ CoreConvertResourceDescriptorHobAttributesToCapabilities (
...
 CalculateTotalMemoryBinSizeNeeded (
-  VOID
+  IN OUT OPTIONAL EFI_PHYSICAL_ADDRESS  *BinTop
   )
 {
   UINTN   Index;
   UINT64  TotalSize;
+  UINT64  Granularity;
+
+  if (gMemoryTypeInformation == NULL) {
```
mdkinney: How would this ever be NULL since it is a global variable?  And it is not a parameter

MdeModulePkg/Core/Dxe/Gcd/Gcd.c
```diff
@@ -2165,25 +2165,63 @@ CoreConvertResourceDescriptorHobAttributesToCapabilities (
 }
 
 /**
-  Calculate total memory bin size neeeded.
+  Calculate total memory bin size needed.
 
-  @return The total memory bin size neeeded.
+  @param BinTop                The top address of the memory bins. This is an optional parameter.
```
mdkinney, line 2170: Looks like alignment requirements are considered of BinTop is NULL or not.

MdeModulePkg/Core/Dxe/Gcd/Gcd.c
os-d: Correct, I'll drop this check. This is because originally this was part of the larger PR where MemoryTypeInformation is a pointer passed in, so the check makes sense there, but when moving it to before the changes that moved the globals to parameters, I didn't fix this.

MdeModulePkg/Core/Dxe/Gcd/Gcd.c
os-d: Correct, address alignment considerations come into consideration when BinTop is non-null. Size alignment is always done because gMemoryTypeInformation already contains the aligned page counts.

@mdkinney are you asking for a change in the comment? I have already noted this here, but happy to update.

os-d:
@mdkinney I have addressed one comment and have a question on the change you are looking for in the other, please re-review.

MdeModulePkg/Core/Dxe/Gcd/Gcd.c
mdkinney: I think a minor update to the comment would help.  When NULL, the returned size meets the alignment requirements as long as the base address selected also meets the alignment requirements.  When non-NULL, then returned TopBin value and the returned size both meet the alignment requirements.  Is that correct?

os-d: That is correct. I have updated the comment to your suggestion.

os-d:
@mdkinney I have updated the comment, please re-review

mdkinney (approved):

os-d:
@lgao4 any comments on this PR, if not, can we merge it? It was split off from a PR that has been open for 4 months that has other approvals, so at this point I don’t think other review is going to come.

lgao4 (approved):

PR #12591, opened 2026-05-18
MdeModulePkg: Dxe Core: Allocate Memory Bins Contiguously

os-d:
# Description

Currently, there is no guarantee that the memory bins will be allocated contiguously. However, there are assumptions in the code that bins are allocated contiguously, such as the GCD init code requiring a free memory region be large enough for a contiguous bin range on DXE Core init.

Ensuring contiguous bins also makes a cleaner model for the bins and keeps the memory map in a more standard configuration between different configurations, as platforms today can pass a resource descriptor HOB to DXE core to describe the bin range and this only supports a contiguous range.

This also sets up reusing the memory bin logic for PEI memory bin support which will use the aforementioned resource descriptor HOB to pass the bin range to DXE.

This commit updates CoreAddMemoryDescriptor() to allocate a contiguous range.

Note: This was split off from #12086 and is required to go in before that PR does.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested with the PEI memory bin changes.

## Integration Instructions

N/A.

### DIFF
--- a/MdeModulePkg/Core/Dxe/Mem/Page.c
+++ b/MdeModulePkg/Core/Dxe/Mem/Page.c
@@ -650,6 +650,13 @@ CoreSetMemoryTypeInformationRange (
     }
   }
 
+  DEBUG ((
+    DEBUG_INFO,
+    "%a: Inherited range 0x%llx - 0x%llx for memory bins\n",
+    __func__,
+    Start,
+    Start + Length -1
+    ));
   mMemoryTypeInformationInitialized = TRUE;
 }
 
@@ -677,11 +684,10 @@ CoreAddMemoryDescriptor (
   )
 {
   EFI_PHYSICAL_ADDRESS  End;
-  EFI_STATUS            Status;
   UINTN                 Index;
-  UINTN                 FreeIndex;
-  UINT64                Alignment;
-  UINT64                BinSize;
+  EFI_PHYSICAL_ADDRESS  BaseAddress;
+  EFI_PHYSICAL_ADDRESS  LastBinAddress;
+  UINT64                RequiredSize;
 
   if ((Start & EFI_PAGE_MASK) != 0) {
     return;
@@ -718,6 +724,42 @@ CoreAddMemoryDescriptor (
     return;
   }
 
+  BaseAddress  = 0;
+  RequiredSize = CalculateTotalMemoryBinSizeNeeded (NULL);
+  if (RequiredSize == 0) {
+    mMemoryTypeInformationInitialized = TRUE;
+    return;
+  }
+
+  // To ensure we get a contiguous range of memory for our bins, we will attempt to allocate
+  // all of the memory needed in one go. If that works, we can then carve it up into the individual bins.
+  // Our size is already aligned to the correct granularity, allocate aligned pages to ensure the base address is
+  // aligned.
+  BaseAddress = (EFI_PHYSICAL_ADDRESS)(UINTN)AllocateAlignedPages (
+                                               EFI_SIZE_TO_PAGES ((UINTN)RequiredSize),
+                                               RUNTIME_PAGE_ALLOCATION_GRANULARITY
+                                               );
+
+  if (BaseAddress == 0) {
+    DEBUG ((
+      DEBUG_INFO,
+      "%a: Could not allocate contiguous pages for all memory bins. It will be attempted again when more memory is added.\n",
+      __func__
+      ));
+    return;
+  }
+
+  DEBUG ((
+    DEBUG_INFO,
+    "%a: Allocated 0x%llx - 0x%llx for memory bins\n",
+    __func__,
+    BaseAddress,
+    BaseAddress + RequiredSize - 1
+    ));
+
+  LastBinAddress         = BaseAddress + RequiredSize;
+  mDefaultMaximumAddress = BaseAddress - 1;
+
   //
   // Loop through each memory type in the order specified by the gMemoryTypeInformation[] array
   //
@@ -731,71 +773,9 @@ CoreAddMemoryDescriptor (
     }
 
     if (gMemoryTypeInformation[Index].NumberOfPages != 0) {
-      Alignment = DEFAULT_PAGE_ALLOCATION_GRANULARITY;
-      if ((gMemoryTypeInformation[Index].Type == EfiReservedMemoryType) ||
-          (gMemoryTypeInformation[Index].Type == EfiACPIMemoryNVS) ||
-          (gMemoryTypeInformation[Index].Type == EfiRuntimeServicesCode) ||
-          (gMemoryTypeInformation[Index].Type == EfiRuntimeServicesData))
-      {
-        Alignment = RUNTIME_PAGE_ALLOCATION_GRANULARITY;
-      }
-
-      BinSize = EFI_PAGES_TO_SIZE ((UINTN)gMemoryTypeInformation[Index].NumberOfPages);
-      BinSize = ALIGN_VALUE (BinSize, Alignment);
-
-      gMemoryTypeInformation[Index].NumberOfPages = (UINT32)EFI_SIZE_TO_PAGES ((UINTN)BinSize);
-
-      //
-      // Allocate pages for the current memory type from the top of available memory
-      //
-      Status = CoreAllocatePages (
-                 AllocateAnyPages,
-                 Type,
-                 gMemoryTypeInformation[Index].NumberOfPages,
-                 &mMemoryTypeStatistics[Type].BaseAddress
-                 );
-      if (EFI_ERROR (Status)) {
-        //
-        // If an error occurs allocating the pages for the current memory type, then
-        // free all the pages allocates for the previous memory types and return.  This
-        // operation with be retied when/if more memory is added to the system
-        //
-        for (FreeIndex = 0; FreeIndex < Index; FreeIndex++) {
-          //
-          // Make sure the memory type in the gMemoryTypeInformation[] array is valid
-          //
-          Type = (EFI_MEMORY_TYPE)(gMemoryTypeInformation[FreeIndex].Type);
-          if ((UINT32)Type > EfiMaxMemoryType) {
-            continue;
-          }
-
-          if (gMemoryTypeInformation[FreeIndex].NumberOfPages != 0) {
-            CoreFreePages (
-              mMemoryTypeStatistics[Type].BaseAddress,
-              gMemoryTypeInformation[FreeIndex].NumberOfPages
-              );
-            mMemoryTypeStatistics[Type].BaseAddress    = 0;
-            mMemoryTypeStatistics[Type].MaximumAddress = MAX_ALLOC_ADDRESS;
-          }
-        }
-
-        return;
-      }
-
-      //
-      // Compute the address at the top of the current statistics
-      //
-      mMemoryTypeStatistics[Type].MaximumAddress =
-        mMemoryTypeStatistics[Type].BaseAddress +
-        LShiftU64 (gMemoryTypeInformation[Index].NumberOfPages, EFI_PAGE_SHIFT) - 1;
-
-      //
-      // If the current base address is the lowest address so far, then update the default
-      // maximum address
-      //
-      if (mMemoryTypeStatistics[Type].BaseAddress < mDefaultMaximumAddress) {
-        mDefaultMaximumAddress = mMemoryTypeStatistics[Type].BaseAddress - 1;
-      }
+      mMemoryTypeStatistics[Type].BaseAddress    = LastBinAddress - EFI_PAGES_TO_SIZE (gMemoryTypeInformation[Index].NumberOfPages);
+      mMemoryTypeStatistics[Type].MaximumAddress = LastBinAddress - 1;
+      LastBinAddress                             = mMemoryTypeStatistics[Type].BaseAddress;
     }
   }
 
@@ -804,6 +784,10 @@ CoreAddMemoryDescriptor (
   // those memory areas can be freed for future allocations, and all future memory
   // allocations can occur within their respective bins
   //
+  FreeAlignedPages (
+    (VOID *)(UINTN)BaseAddress,
+    EFI_SIZE_TO_PAGES ((UINTN)RequiredSize)
+    );
   for (Index = 0; gMemoryTypeInformation[Index].Type != EfiMaxMemoryType; Index++) {
     //
     // Make sure the memory type in the gMemoryTypeInformation[] array is valid
@@ -814,10 +798,6 @@ CoreAddMemoryDescriptor (
     }
 
     if (gMemoryTypeInformation[Index].NumberOfPages != 0) {
-      CoreFreePages (
-        mMemoryTypeStatistics[Type].BaseAddress,
-        gMemoryTypeInformation[Index].NumberOfPages
-        );
       mMemoryTypeStatistics[Type].NumberOfPages   = gMemoryTypeInformation[Index].NumberOfPages;
       gMemoryTypeInformation[Index].NumberOfPages = 0;
     }


os-d:
@mdkinney this is the second PR split off from the PEI memory bins PR, similarly I can't add you as a reviewer because my write access is revoked during the hard freeze.

As mentioned in the PR description, this currently contains the two commits from #12590 because it is intended to go in after it. Only the third commit should be reviewed in this PR; the other two will be dropped when the first PR is merged.

mdkinney:
@os-d can you please update since #12590 has been merged?

os-d:
> @os-d can you please update since #12590 has been merged?

Yes, I am out of town through Tues, will rebase on Weds and let you know.

os-d:
@mdkinney I have updated this PR to rebase on master, dropping the two commits from #12590 . No changes to the only remaining commit in this PR. Please review when you have a chance.

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -731,71 +768,9 @@ CoreAddMemoryDescriptor (
...
-        LShiftU64 (gMemoryTypeInformation[Index].NumberOfPages, EFI_PAGE_SHIFT) - 1;
-
-      //
-      // If the current base address is the lowest address so far, then update the default
-      // maximum address
-      //
-      if (mMemoryTypeStatistics[Type].BaseAddress < mDefaultMaximumAddress) {
-        mDefaultMaximumAddress = mMemoryTypeStatistics[Type].BaseAddress - 1;
-      }
+      mMemoryTypeStatistics[Type].BaseAddress    = LastBinAddress - LShiftU64 (gMemoryTypeInformation[Index].NumberOfPages, EFI_PAGE_SHIFT);
```
mdkinney: The code being replaced here looks at the alignment requirements for each memory type.  That part of the logic seems to have been dropped and this computation of BaseAddress and adjustment to number of pages.  Am I missing where that is being done?

mdkinney:
The commit message does not seem to be correct.  This change is to DXE Core only and is implementing a stricter requirement to allocate all bins in contiguous memory.  I think this is in preparation of a future PEI change so this logic works when DXE creates the bins or DXE inherits the bins from PEI in a future PR.  This PR and commit message should focus on adding the stricter requirement for all bins being allocated in a contiguous buffer.

This seems like it may be a non-backwards compatible change because there could be memory HOBs from PEI where non of them are large enough for the  contiguous allocation.  Do we need to add "Breaking Change" flag and describe what a platform needs to do if the allocation of bins that worked before this change fails after this change?

MdeModulePkg/Core/Dxe/Mem/Page.c
os-d: @mdkinney that is being done here: https://github.com/tianocore/edk2/pull/12591/changes#diff-5112fd915996f95133e8e0e56eb9032d77120a527d30f1c276011f4032ee06deR734-R739.

By calling `AllocateReservedPages()` for the calculated size from `CalculateTotalMemoryBinSizeNeeded()` (which per the last PR is updated to correctly calculate the bin sizes with alignment considered) we guarantee the base address is aligned (the allocator guarantees that for us) and that the size is correct.

os-d:
> The commit message does not seem to be correct. This change is to DXE Core only and is implementing a stricter requirement to allocate all bins in contiguous memory. I think this is in preparation of a future PEI change so this logic works when DXE creates the bins or DXE inherits the bins from PEI in a future PR. This PR and commit message should focus on adding the stricter requirement for all bins being allocated in a contiguous buffer.
> 

I'll update the commit message.

> This seems like it may be a non-backwards compatible change because there could be memory HOBs from PEI where non of them are large enough for the contiguous allocation. Do we need to add "Breaking Change" flag and describe what a platform needs to do if the allocation of bins that worked before this change fails after this change?

I don't think this actually could be a breaking change here, because the GCD initialization code has always assumed that the bin range will be contiguous, it just never actually allocated it as such. But when it initializes, it calls `CalculateTotalMemoryBinSizeNeeded()` to get the size of the bins + a static minimum size for the DXE core. It then searches through all the HOBs to find a region that will work, doing its best to merge some together if it has to. If it doesn't find a range, it will assert. That's a long stretch of logic, but it is it here: https://github.com/tianocore/edk2/blob/f17ed80f61487a19271c218f7b569163e83d26eb/MdeModulePkg/Core/Dxe/Gcd/Gcd.c#L2417C3-L2614.

So, even without this change, the scenario you are describing would have hit that assert because the GCD logic is assuming contiguous bins. And in release builds, it would have failed allocating when [`CoreAddMemoryDescriptor()` is called](https://github.com/tianocore/edk2/blob/f17ed80f61487a19271c218f7b569163e83d26eb/MdeModulePkg/Core/Dxe/Gcd/Gcd.c#L2640) on that very first region, because it is the only region DXE has available at that point and `CoreAddMemoryDescriptor()` then expects to initialize the all of the bins from the first region (it will [try again later if it couldn't](https://github.com/tianocore/edk2/blob/f17ed80f61487a19271c218f7b569163e83d26eb/MdeModulePkg/Core/Dxe/Mem/Page.c#L757-L782)).

So, behavior wise in the scenario you describe it should be the same:
- Debug builds assert
- Release builds fail to allocate bins the first time, when enough memory is added later, it gets allocated

Let me know if I'm missing something.

MdeModulePkg/Core/Dxe/Mem/Page.c
mdkinney: I agree the single large allocation is correct.  It is the mMemoryStatistics BaseAddress and NumberOfPages assignments and updates that do not appear to match previous logic.

os-d:
@mdkinney I have updated the commit message and PR description, let me know if that seems reasonable.

MdeModulePkg/Core/Dxe/Mem/Page.c
os-d: Ah, gotcha, thanks for clarifying.

gMemoryTypeInformation[Index].NumberOfPages is already guaranteed to have the correctly aligned number of pages: https://github.com/tianocore/edk2/blob/f17ed80f61487a19271c218f7b569163e83d26eb/MdeModulePkg/Core/Dxe/Mem/Page.c#L734-L746. This was part of your recent fix: https://github.com/tianocore/edk2/commit/6834d43991741737acf9e53350cb9293e1c9c2f5.

Because the single large allocation has the correctly aligned base and end address and the sizes are correctly aligned in gMemoryTypeInformation, each individual base address (calculated by `End - EFI_PAGES_TO_SIZE(gMemoryTypeInformation[Index].NumberOfPages)`) is therefore correctly aligned.

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -650,6 +650,13 @@ CoreSetMemoryTypeInformationRange (
     }
   }
 
+  DEBUG ((
```
mdkinney, line 653: What is the purpose of the new debug message in this PR?  Why is term 'inherited' used?

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -718,6 +724,37 @@ CoreAddMemoryDescriptor (
...
+  RequiredSize = CalculateTotalMemoryBinSizeNeeded (NULL);
+  if (RequiredSize == 0) {
+    mMemoryTypeInformationInitialized = TRUE;
+    return;
+  }
+
+  // To ensure we get a contiguous range of memory for our bins, we will attempt to allocate
+  // all of the memory needed in one go. If that works, we can then carve it up into the individual bins. We allocate
+  // reserved pages to ensure runtime page allocation granularity is taken into account.
+  BaseAddress = (EFI_PHYSICAL_ADDRESS)(UINTN)AllocateReservedPages (
```
mdkinney: This uses a memory allocation lib that may or may not honor the alignment requirements.  Would it make more sense to directly call the service inside the DXE Core to do this allocation instead of depending on the memory allocation lib?

For example, the code being replaced that did this piecemeal used CoreAllocagePages().

MdeModulePkg/Core/Dxe/Mem/Page.c
os-d: I wanted print statements for where the bin locations are. There are two ways DXE core can get bins, either via allocation (which is changed in this PR and so has the print added) and inheriting a bin range from pre-DXE. For now, that's only inherited from the platform, but in the future could be inherited from PEI.

So, because I was adding the print for the allocation case, I also added a print for the inherited case.

I can change the language of it if you want and I can make it a separate commit if you prefer.

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -804,6 +779,10 @@ CoreAddMemoryDescriptor (
   // those memory areas can be freed for future allocations, and all future memory
   // allocations can occur within their respective bins
   //
+  FreePages (
```
mdkinney: If code above is updated to use CoreAllocatePages(), then this should use CoreFreePages()

MdeModulePkg/Core/Dxe/Mem/Page.c
mdkinney: I agree.  Thanks!

MdeModulePkg/Core/Dxe/Mem/Page.c
os-d: The reason that I used MemoryAllocationLib here is to set up sharing this code between PEI core and DXE core. Obviously those two different environments have different internal allocation semantics, so MemoryAllocationLib is the abstraction point between them.

However, I agree that a given lib could not have the alignment requirement correct. In fact, EmulatorPkg's DXE_CORE instance of MemoryAllocationLib does not (though EmulatorPkg is only supported for IA32 and X64 so in practice that doesn't matter).

For the standard MemoryAllocationLib for DXE_CORE, it does just use CoreAllocatePages().

The size will be correct, because that is calculated independently accounting for the alignment, but the base address could be misaligned.

In order to set up sharing this code between PEI and DXE, but still handle the rare case where a non-standard MemoryAllocationLib for the cores are used, I can change this to do something like:

```c
 BaseAddress = (EFI_PHYSICAL_ADDRESS)(UINTN)AllocateReservedPages ((UINTN)RShiftU64 (RequiredSize, EFI_PAGE_SHIFT));
 
  if (BaseAddress == 0) {
    DEBUG ((DEBUG_ERROR, "%a: Could not allocate contiguous pages for all memory bins\n", __func__));
    ASSERT (FALSE);
    return;
  }
  
  // new code
  AddressToFree = BaseAddress;

  if (BaseAddress & (RUNTIME_PAGE_ALLOCATION_GRANULARITY - 1) != 0) {
    // non-standard MemoryAllocationLib used, ensure alignment ourselves
   FreePages(BaseAddress, (UINTN)RShiftU64 (RequiredSize, EFI_PAGE_SHIFT));

   RequiredSize += RUNTIME_PAGE_ALLOCATION_GRANULARITY;
   BaseAddress = (EFI_PHYSICAL_ADDRESS)(UINTN)AllocateReservedPages ((UINTN)RShiftU64 (RequiredSize, EFI_PAGE_SHIFT));
 
  if (BaseAddress == 0) {
    DEBUG ((DEBUG_ERROR, "%a: Could not allocate contiguous pages for all memory bins\n", __func__));
    ASSERT (FALSE);
    return;
   }

    AddressToFree = BaseAddress;
    BaseAddress = ALIGN_VALUE (BaseAddress, RUNTIME_PAGE_ALLOCATION_GRANULARITY);
  }
 
  ...
  FreePages(AddressToFree, (UINTN)RShiftU64 (RequiredSize, EFI_PAGE_SHIFT));

```

I think it would be unlikely we'd ever hit that case, but then it would be covered. Thoughts?

mdkinney: There are aligned allocation APIs in the MemoryAllocationLib

```
/**
  Allocates one or more 4KB pages of type EfiReservedMemoryType at a specified alignment.

  Allocates the number of 4KB pages specified by Pages of type EfiReservedMemoryType with an
  alignment specified by Alignment.  The allocated buffer is returned.  If Pages is 0, then NULL is
  returned.  If there is not enough memory at the specified alignment remaining to satisfy the
  request, then NULL is returned.

  If Alignment is not a power of two and Alignment is not zero, then ASSERT().
  If Pages plus EFI_SIZE_TO_PAGES (Alignment) overflows, then ASSERT().

  @param  Pages                 The number of 4 KB pages to allocate.
  @param  Alignment             The requested alignment of the allocation.  Must be a power of two.
                                If Alignment is zero, then byte alignment is used.

  @return A pointer to the allocated buffer or NULL if allocation fails.

**/VOID *
EFIAPI
AllocateAlignedReservedPages (
  IN UINTN  Pages,
  IN UINTN  Alignment
  );

/**
  Frees one or more 4KB pages that were previously allocated with one of the aligned page
  allocation functions in the Memory Allocation Library.

  Frees the number of 4KB pages specified by Pages from the buffer specified by Buffer.  Buffer
  must have been allocated on a previous call to the aligned page allocation services of the Memory
  Allocation Library.  If it is not possible to free allocated pages, then this function will
  perform no actions.

  If Buffer was not allocated with an aligned page allocation function in the Memory Allocation
  Library, then ASSERT().
  If Pages is zero, then ASSERT().

  @param  Buffer                The pointer to the buffer of pages to free.
  @param  Pages                 The number of 4 KB pages to free.

**/
VOID
EFIAPI
FreeAlignedPages (
  IN VOID   *Buffer,
  IN UINTN  Pages
  );
```

os-d: A much easier solution I forgot about :). Thanks for the reminder, I’ll switch to that.

os-d:
@mdkinney I switched to `AllocateAlignedPages()`. Please re-review.

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -804,6 +781,10 @@ CoreAddMemoryDescriptor (
...
```
mdkinney: Needs to be FreeAlignedPages() to match allocation.

os-d: Oops, thanks for catching that, fixed.

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -718,6 +724,39 @@ CoreAddMemoryDescriptor (
...
+  // aligned.
+  BaseAddress = (EFI_PHYSICAL_ADDRESS)(UINTN)AllocateAlignedPages (
+                                               (UINTN)RShiftU64 (RequiredSize, EFI_PAGE_SHIFT),
+                                               RUNTIME_PAGE_ALLOCATION_GRANULARITY
+                                               );
+
+  if (BaseAddress == 0) {
+    DEBUG ((DEBUG_ERROR, "%a: Could not allocate contiguous pages for all memory bins\n", __func__));
+    ASSERT (FALSE);
+    return;
```
mdkinney, line 749: Since CoreAddMemoryDescriptor() can be called many times with different sizes, and the one large enough for the contiguous allocation may not be the first descriptor added, I do not think this should be a DEBUG_ERROR or an ASSERT().  It is simply that the specific descriptor added is not big enough.

os-d: Good point. I'll change this to an info and drop the assert.

os-d: Fixed.

mdkinney:
All the updates LGTM.

One last comment  I see several RShift64 and LShiftU64 for EFI_PAGE_SHIFT.  In general I prefer to use EFI_SIZE_TO_PAGES() and EFI_PAGES_TO_SIZE() to make the code more readable.  Is there a reason those are not being used?

os-d:
> All the updates LGTM.
> 
> One last comment I see several RShift64 and LShiftU64 for EFI_PAGE_SHIFT. In general I prefer to use EFI_SIZE_TO_PAGES() and EFI_PAGES_TO_SIZE() to make the code more readable. Is there a reason those are not being used?

@mdkinney I agree and I prefer the macros as well. However, the macros directly use the shift operators: https://github.com/tianocore/edk2/blob/8812785128a3574abf5da87cd3388bd435073a7f/MdePkg/Include/Uefi/UefiBaseType.h#L202, which means they can’t be used for shifting 64 bit numbers (such as addresses) for IA32 because it will generate compiler intrinsics. Because IA32 PEI is still supported (and IA32 DXE is still buildable in the codebase) and this code is being shared with PEI, I had to use [R|L]ShiftU64.

The macros are intended to be used across different bitwidth types, so I don’t think we can easily change them to not directly use the shift operators, but I’d need to do some more thinking there. If you want to explore that route, I can take a look, but would prefer to do so after this PR goes in and then fix up the whole codebase in areas that couldn’t use the macros because of IA32 compilation.

mdkinney:
> > All the updates LGTM.
> > One last comment I see several RShift64 and LShiftU64 for EFI_PAGE_SHIFT. In general I prefer to use EFI_SIZE_TO_PAGES() and EFI_PAGES_TO_SIZE() to make the code more readable. Is there a reason those are not being used?
> 
> @mdkinney I agree and I prefer the macros as well. However, the macros directly use the shift operators:
> 
> https://github.com/tianocore/edk2/blob/8812785128a3574abf5da87cd3388bd435073a7f/MdePkg/Include/Uefi/UefiBaseType.h#L202
> 
> , which means they can’t be used for shifting 64 bit numbers (such as addresses) for IA32 because it will generate compiler intrinsics. Because IA32 PEI is still supported (and IA32 DXE is still buildable in the codebase) and this code is being shared with PEI, I had to use [R|L]ShiftU64.
> The macros are intended to be used across different bitwidth types, so I don’t think we can easily change them to not directly use the shift operators, but I’d need to do some more thinking there. If you want to explore that route, I can take a look, but would prefer to do so after this PR goes in and then fix up the whole codebase in areas that couldn’t use the macros because of IA32 compilation.

Do we ever expect the size of these allocations to be > 4GB on 32-bit platforms?  Use of direct shift work for type UINTN is always safe, so typecast to UINTN to use that macro is how this has been addressed elsewhere.

I believe the reason that LShuftU64() and RShiftU64() are not used in these macros is because that would indirectly add a dependency on BaseLib from use of the include file with these macros.

os-d:
> Do we ever expect the size of these allocations to be > 4GB on 32-bit platforms? Use of direct shift work for type UINTN is always safe, so typecast to UINTN to use that macro is how this has been addressed elsewhere.
> 
> I believe the reason that LShuftU64() and RShiftU64() are not used in these macros is because that would indirectly add a dependency on BaseLib from use of the include file with these macros.

@mdkinney Yeah, that's a fair assumption to make. I have updated to use the macros. Please re-review.

mdkinney (approved):

mdkinney:
> > Do we ever expect the size of these allocations to be > 4GB on 32-bit platforms? Use of direct shift work for type UINTN is always safe, so typecast to UINTN to use that macro is how this has been addressed elsewhere.
> > I believe the reason that LShuftU64() and RShiftU64() are not used in these macros is because that would indirectly add a dependency on BaseLib from use of the include file with these macros.
> 
> @mdkinney Yeah, that's a fair assumption to make. I have updated to use the macros. Please re-review.

Updates look good.

os-d:
@lgao4 any comments here? Otherwise, can we merge the PR? This is another one split out from the PEI memory bins PR that has been open for >4 months, so I don't anticipate any other review.

makubacki:
> @lgao4 any comments here? Otherwise, can we merge the PR? This is another one split out from the PEI memory bins PR that has been open for >4 months, so I don't anticipate any other review.

@lgao4, can you please reply if you plan to review the latest changes? This PR has been open for a month so it's difficult to tell if messages are being missed or there is intent to review it. If we don't hear back in a day or so, I think we can move on with Mike's detailed review.

lgao4 (approved):

lgao4:
I have no other comments.

PR #12592, opened 2026-05-19
DynamicTablesPkg : Add HMAT generator 

gonzoua:
# Description

Add HMAT generator to DynamicTablesPkg. This is a reworked version of the original submission (PR https://github.com/tianocore/edk2/pull/10915). 

- [ ] Breaking change?
- [ ] Impacts security?
- [x] Includes tests?

### Build

```
host$ docker run --rm -ti -v "${HOME}":"${HOME}" \
    -e EDK2_DOCKER_USER_HOME="${HOME}" \
    -v $(pwd):/host -w /host \
    --name edk2-ubuntu-dev \
    ghcr.io/tianocore/containers/ubuntu-22-dev:latest /bin/bash
docker$ export WORKSPACE=$(pwd)
docker$ export EDK_TOOLS_PATH=/host/edk2/BaseTools
docker$ source edk2/edksetup.sh
docker$ make -j$(nproc) -C edk2/BaseTools
docker$ git clone -b test-hmat https://github.com/gonzoua/edk2-dut.git /tmp/edk2-dut
docker$ export PACKAGES_PATH=$(pwd):/tmp/edk2-dut
docker$ build -t GCC -a AARCH64 -p /tmp/edk2-dut/Arm64Dut.dsc
docker$ build -t GCC -a X64 -b NOOPT -p edk2/DynamicTablesPkg/Test/DynamicTablesPkgHostTest.dsc
```

#### QEMU

```
export CODE=Build/Arm64Dut/$B/FV/QEMU_EFI.fd
export VARS=Build/Arm64Dut/$B/FV/QEMU_VARS.fd

dd of="QEMU_EFI-pflash.raw" if="/dev/zero" bs=1M count=64
dd of="QEMU_EFI-pflash.raw" if="${CODE}" conv=notrunc
dd of="QEMU_VARS-pflash.raw" if="/dev/zero" bs=1M count=64
dd of="QEMU_VARS-pflash.raw" if="${VARS}" conv=notrunc

qemu-system-aarch64 -m 4G -cpu cortex-a57 -M virt -nographic \
       -drive file=QEMU_EFI-pflash.raw,format=raw,if=pflash \
       -drive file=QEMU_VARS-pflash.raw,format=raw,if=pflash 
```

In EFI shell run `acpiview -s HMAT`, the output does not contain errors:
```
...
Table Statistics:                                                                                                                                                                                             
        0 Error(s)                                                                                                                                                                                            
        0 Warning(s)                                                                                                                                                                                          
Shell> 
```

#### Unit test

```
docker$ Build/DynamicTablesPkg/HostTest/NOOPT_GCC/X64/HmatGeneratorGoogleTest
...
[  PASSED  ] 5 tests.
```


DynamicTablesPkg/Library/Common/MetadataObjLib/MetadataObj.c
pierregondois: NIT: `RootHandle`, same update for `DynamicTablesPkg/Include/Library/MetadataObjLib.h`

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,705 @@
...
+
+  MemProximityDomainAttrOffset = TableSize;
+  TableSize                   +=
+    (sizeof (EFI_ACPI_6_4_HMAT_STRUCTURE_MEMORY_PROXIMITY_DOMAIN_ATTRIBUTES) * MemProximityDomainAttrCount);
+
+  MemLatBwOffset = TableSize;
+
+  // The largest SSLBI entry is full matrix where all proximity
+  // domains act as initiators and targets. Go for max case, adjust size later.
+  MaxLatBwSize = GetSllbiTableSize (ProximityDomainInfoCount, ProximityDomainInfoCount);
```
pierregondois: Hello Oleksandr,
I think we had a similar issue in the SlitGenerator.

The array has the size of  `Initiator Proximity Domains` * `Target Proximity Domains`.
Unfortunately, we cannot describe the list of Initiator/Target proximity domains.
So for instance, if the proximity domains are (unluckily):
- Initiator: 10,11
- Target: 12,13

then there is no way to know that the initiator starts at 10 and the target starts at 12. We only know the maximal value of the proximity domains, and we need to create an array of 11*13.

This also means that the number of proximity domains we get from `GetEArchCommonObjProximityDomainInfo()` won't help to estimate the size of the array.
In the above example, we have 4 different values (10,11,12,13), but the array is not at max 4*4 (as computed in the code), but 11*13.

Another example, with the following, the HMAT tests fail:
```
diff --git a/DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/GoogleTest/HmatGeneratorGoogleTest.cpp b/DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/GoogleTest/HmatGeneratorGoogleTest.cpp
index 126820c4a41b..ccf65b661b4c 100644
--- a/DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/GoogleTest/HmatGeneratorGoogleTest.cpp
+++ b/DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/GoogleTest/HmatGeneratorGoogleTest.cpp
@@ -440,8 +440,8 @@ protected:
     },
     {
       .Flags                    = 1,
-      .InitiatorProximityDomain = 3,
-      .MemoryProximityDomain    = 4
+      .InitiatorProximityDomain = 9,
+      .MemoryProximityDomain    = 10
     }
   };
 
```

If you think the code from the SlitGenerator can be re-used, maybe the common code could be placed in `DynamicTablesPkg/Include/Library/CmObjHelperLib.h` ?

gonzoua: I understand the issue with SLIT and non-sequential domain IDs (we actually ran into it independently some time ago) but I don't think this applies to HMAT. The difference is that HMAT SSLBI matrix includes lists of initiators and targets, so in your example above the actual SSLBI record content would be:
```
[10 11]  [12 13] [4 UINT16 relation values].
```

So if provided data is valid we can infer the max size of this data structure. It does not depend on the doimain ID values, only on a number of them. I have validation checks to ensure that data is sane and it's not a sparse matrix.

gonzoua:
Force-push to  [3204b5a](https://github.com/tianocore/edk2/commit/3204b5adc7c49762135c4948630466b1f15f572e):

* Update documentation for `MetadataFreeHandle`

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
pierregondois: Ah yes, right indeed, sorry I completely missed it.
I will review the generator next week if that's ok

gonzoua: Yes, no rush with the review.

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,705 @@
...
+  @retval EFI_SUCCESS   If DomainId was found
+  @retval EFI_NOT_FOUND If DomainId wasn't found
+**/
+STATIC
+EFI_STATUS
+FindDomainIndex (
+  IN  UINT32  DomainId,
+  IN  UINT32  *DomainIds,
+  IN  UINTN   DomainIdCount,
+  OUT UINTN   *DomainIdIndex
```
pierregondois: NIT: Should be marked as `OPTIONAL`

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,705 @@
...
+**/
+STATIC
+EFI_STATUS
+AddMemLatBwInfo (
+  IN      CONST EDKII_CONFIGURATION_MANAGER_PROTOCOL   *CONST                           CfgMgrProtocol,
+  IN      CONST CM_ARCH_COMMON_MEMORY_LAT_BW_INFO                                       *CmMemLatBwInfo,
+  IN      CONST CM_ARCH_COMMON_PROXIMITY_DOMAIN_RELATION_INFO                           *CmMemLatBwRelations,
+  IN            UINTN                                                                   CmMemLatBwRelationCount,
+  IN            UINT32                                                                  ProximityDomainCount,
+  IN  OUT       EFI_ACPI_6_4_HMAT_STRUCTURE_SYSTEM_LOCALITY_LATENCY_AND_BANDWIDTH_INFO  *MemLatBwInfo,
```
pierregondois, line 144: `IN` only parameter no ?

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,705 @@
...
+  UINT32      DomainId;
+  UINTN       InitiatorIdx;
+  UINTN       TargetIdx;
+  UINT16      *Relations;
+
+  ASSERT (CfgMgrProtocol != NULL);
+  ASSERT (MemLatBwInfo != NULL);
+  ASSERT (CmMemLatBwInfo != NULL);
+  ASSERT (CmMemLatBwRelations != NULL);
+  ASSERT (CmMemLatBwRelationCount > 0);
```
pierregondois, line 164: Similarly to `AddMemProximityDomainAttrInfo()`, I think the function should start with that in case no subtable is present.

```
if (CmMemLatBwRelationCount == 0) {
  return EFI_SUCCESS;
}
```

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,705 @@
...
+  ASSERT (RecordLength != NULL);
+
+  MemLatBwInfo->Type = EFI_ACPI_6_4_HMAT_TYPE_SYSTEM_LOCALITY_LATENCY_AND_BANDWIDTH_INFO;
+
+  CopyMem (&MemLatBwInfo->Flags, &CmMemLatBwInfo->Flags, sizeof (UINT8));
+  MemLatBwInfo->DataType        = CmMemLatBwInfo->DataType;
+  MemLatBwInfo->MinTransferSize = CmMemLatBwInfo->MinTransferSize;
+  MemLatBwInfo->EntryBaseUnit   = CmMemLatBwInfo->EntryBaseUnit;
+
+  UINT8  *TablePtr = (UINT8 *)MemLatBwInfo;
```
pierregondois: Ideally the declaration should be at the beginning of the function and the initialization on another line:
```
UINT8  *TablePtr;

TablePtr = (UINT8 *)MemLatBwInfo;
```

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,705 @@
...
+
+  TablePtr            += sizeof (EFI_ACPI_6_4_HMAT_STRUCTURE_SYSTEM_LOCALITY_LATENCY_AND_BANDWIDTH_INFO);
+  InitiatorDomainCount = 0;
+  InitiatorDomains     = (UINT32 *)TablePtr;
+
+  // Append Initiator ProximityDomains.
+  for (Index = 0; Index < CmMemLatBwRelationCount; Index++) {
+    Status = GetProximityDomainId (
```
pierregondois, line 193: Would it be possible to create a wrapper calling `GetProximityDomainId() + FindDomainIndex()` and re-use it to populate the target domains and the relation values aswell ?

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,705 @@
...
+      TargetDomainCount,
+      CmMemLatBwRelationCount,
+      InitiatorDomainCount * TargetDomainCount
+      ));
+    return EFI_INVALID_PARAMETER;
+  }
+
+  Relations = (UINT16 *)(TargetDomains + TargetDomainCount);
+  for (Index = 0; Index < CmMemLatBwRelationCount; Index++) {
+    if (InitiatorDomainCount * TargetDomainCount != CmMemLatBwRelationCount) {
```
pierregondois: Should be `(CmMemLatBwRelations[Index].Relation > MAX_UINT16)` + Maybe move it closer to the Relations assignment statement ?

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,705 @@
...
+    return EFI_SUCCESS;
+  }
+
+  ASSERT (MemProximityDomainAttrInfo != NULL);
+  ASSERT (CmMemProximityDomainAttrInfo != NULL);
+
+  while (MemProximityDomainAttrCount-- != 0) {
+    MemProximityDomainAttrInfo->Type   = EFI_ACPI_6_4_HMAT_TYPE_MEMORY_PROXIMITY_DOMAIN_ATTRIBUTES;
+    MemProximityDomainAttrInfo->Length = sizeof (EFI_ACPI_6_4_HMAT_STRUCTURE_MEMORY_PROXIMITY_DOMAIN_ATTRIBUTES);
+    CopyMem (&MemProximityDomainAttrInfo->Flags, &CmMemProximityDomainAttrInfo->Flags, sizeof (UINT16));
```
pierregondois, line 343: NIT: is it possible to add some validation for the flags ?

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,705 @@
...
+  ASSERT (MemLatBwInfo != NULL);
+  ASSERT (CmMemLatBwInfo != NULL);
+  ASSERT (CmMemLatBwRelations != NULL);
+  ASSERT (CmMemLatBwRelationCount > 0);
+  ASSERT (ProximityDomainCount > 0);
+  ASSERT (RecordLength != NULL);
+
+  MemLatBwInfo->Type = EFI_ACPI_6_4_HMAT_TYPE_SYSTEM_LOCALITY_LATENCY_AND_BANDWIDTH_INFO;
+
+  CopyMem (&MemLatBwInfo->Flags, &CmMemLatBwInfo->Flags, sizeof (UINT8));
```
pierregondois, line 174: NIT: is it possible to add some validation for the flags ?

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,705 @@
...
+
+  while (MemProximityDomainAttrCount-- != 0) {
+    MemProximityDomainAttrInfo->Type   = EFI_ACPI_6_4_HMAT_TYPE_MEMORY_PROXIMITY_DOMAIN_ATTRIBUTES;
+    MemProximityDomainAttrInfo->Length = sizeof (EFI_ACPI_6_4_HMAT_STRUCTURE_MEMORY_PROXIMITY_DOMAIN_ATTRIBUTES);
+    CopyMem (&MemProximityDomainAttrInfo->Flags, &CmMemProximityDomainAttrInfo->Flags, sizeof (UINT16));
+
+    Status = GetProximityDomainId (
+               CfgMgrProtocol,
+               0,
+               CmMemProximityDomainAttrInfo->InitiatorProximityDomain,
```
pierregondois, line 357: NIT: We could directly give &MemProximityDomainAttrInfo->InitiatorProximityDomain to GetProximityDomainId()

We could also validate that (InitiatorProximityDomain != MemoryProximityDomain)

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,705 @@
...
+  Status = GetEArchCommonObjProximityDomainInfo (
+             CfgMgrProtocol,
+             CM_NULL_TOKEN,
+             &ProximityDomainInfo,
+             &ProximityDomainInfoCount
+             );
+
+  if (EFI_ERROR (Status)) {
+    DEBUG (
+      (
```
pierregondois: NIT: did uncrustify do that ?

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,705 @@
...
+  }
+
+  Status = GetEArchCommonObjMemoryProximityDomainAttrInfo (
+             CfgMgrProtocol,
+             CM_NULL_TOKEN,
+             &MemProximityDomainAttrInfo,
+             &MemProximityDomainAttrCount
+             );
+
+  if (EFI_ERROR (Status)) {
```
pierregondois: I think: `if (EFI_ERROR (Status) && (Status != EFI_NOT_FOUND))`
Same comment for `GetEArchCommonObjMemoryLatBwInfo()`

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,705 @@
...
+       TableSize,
+       Status
+      )
+      );
+    goto error_handler;
+  }
+
+  Hmat = (EFI_ACPI_6_4_HETEROGENEOUS_MEMORY_ATTRIBUTE_TABLE_HEADER *)*Table;
+
+  if (TableSize > MAX_UINT32) {
```
pierregondois: Maybe better to move it before the allocation

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,705 @@
...
+               MemLatBwInfos[Index].RelativeDistanceArray,
+               &MemLatBwRelations,
+               &MemLatBwRelationsCount
+               );
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "%a: Failed to get relative distances array: %r\n", __func__, Status));
+      goto error_handler;
+    }
+
+    Status = AddMemLatBwInfo (
```
pierregondois, line 624: Need `LatBwEntrySize = 0;` if we allow to return early when `MemLatBwRelationsCount == 0`

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,705 @@
...
+
+  MemLatBwOffset = TableSize;
+
+  // The largest SSLBI entry is full matrix where all proximity
+  // domains act as initiators and targets. Go for max case, adjust size later.
+  MaxLatBwSize = GetSllbiTableSize (ProximityDomainInfoCount, ProximityDomainInfoCount);
+  TableSize   += MemLatBwCount * MaxLatBwSize;
+
+  // Allocate the Buffer for HMAT table
+  *Table = (EFI_ACPI_DESCRIPTION_HEADER *)AllocateZeroPool (TableSize);
```
pierregondois, line 567: Should we call `AllocateCopyPool()` once we know the actual size of the HMAT table ?

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: As a pointer - yes, but it is modified down in the function, for instance lines 167, 170

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: Fixed

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: This is different from AttrInfo. The presence of the table is signalled by a number of `EArchCommonObjMemoryLatBwInfo` objects. This function handles number of relations referenced from the LatBwInfo. If it's empty - there is no reason to have the table in the first place so it's internal inconsistency and should be treated as an error.

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: Done. The only possible invalid value is memory hierarchy enum.

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: Fixed

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: Done

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: I think I did it once and then copy-pasted it and Uncrustify was OK with it. Fixed.

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: Fixed

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: Fixed the former.

As for the latter - I didn't find any restriction on it. Spec states that the initiator field provides additional information but never states that the domain should be different from memory's one.

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: Fixed

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: Fixed

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: Done

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
pierregondois: FWIU the sub-table describes the link/associativity that exists between proximity domains.
A proximity domain should always be associated with itself, so describing such relation would not bring any new information.

IMO it would be nice to at least have an `ASSERT()` to warn. This can be changed if a real use case is reported.

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
pierregondois: FWIU it should be theoretically possible to only have `Memory Proximity Domain Attributes Structure`.
On the other side (and as you said) it would be a bit meaningless do do so.

It should be ok to be restrictive and update the code later if someone ever comes with such case.

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
pierregondois: Is it possible to also add a check for `Access attributes` ? Only 2 values are possible

pierregondois:
Hello Oleksandr,
Just a few comments a left I think.
Also would it be possible to add some `ASSERT_EFI_ERROR (Status);` or `ASSERT (FALSE);` in the error paths ?

pierregondois:
@mdkinney and @lgao4, would it be possible to have a look at the MdePkg patch ?

lgao4:
> @mdkinney and @lgao4, would it be possible to have a look at the MdePkg patch ?

The change in Mdepkg looks good.

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: Ugh, somehow I didn't add my comment here when uploading previous version. I experimented with this idea but it doesn't improve readability for the code. There are four instances when this combination is used but two slightly distinct use cases. The wrapper that can be used for both would have an overloaded semantics. I'd prefer to keep it this way and when/if more instances occure I'll create a wrapper. But for now it's just shifts complexitiy to another place.

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: It's a flags field so my understanding it may have none of them, or both present. So any combination of bits for Access field is a valid one.

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: Agree, I added proper validation, ASSERT would be a half-measure. This is unlikely and can be changed later if required.

gonzoua:
> Hello Oleksandr, Just a few comments a left I think. Also would it be possible to add some `ASSERT_EFI_ERROR (Status);` or `ASSERT (FALSE);` in the error paths ?

Added some , if you think there are more places they should be added to - let me know.

gonzoua:
Force-push to  [42d75f6](https://github.com/tianocore/edk2/commit/42d75f65917401233e193576cc43b71e9ccbe6cd)  added:

* Check for InitiatorDomain not to be equal to Target domain for Memory Proximity Domain Attributes record.
*  ASSERT checks for error status for some error paths.

gonzoua:
Gentle reminder: I believe I addressed all feedback items.

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
pierregondois: ah yes, right indeed

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
pierregondois: ok understood

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,738 @@
...
+  );
+
+/** Retrieve the SSLBI Info */
+GET_OBJECT_LIST (
+  EObjNameSpaceArchCommon,
+  EArchCommonObjMemoryLatBwInfo,
+  CM_ARCH_COMMON_MEMORY_LAT_BW_INFO
+  );
+
+/** Find an index of the specified doamin ID in an array.
```
pierregondois: `domain`

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,738 @@
...
+  EFI_ACPI_6_4_HETEROGENEOUS_MEMORY_ATTRIBUTE_TABLE_HEADER  *Hmat;
+  VOID                                                      *FinalTable;
+
+  ASSERT (This != NULL);
+  ASSERT (AcpiTableInfo != NULL);
+  ASSERT (CfgMgrProtocol != NULL);
+  ASSERT (Table != NULL);
+  ASSERT (AcpiTableInfo->TableGeneratorId == This->GeneratorID);
+  ASSERT (AcpiTableInfo->AcpiTableSignature == This->AcpiTableSignature);
+
```
pierregondois, line 494: Is it possible to add a revision check:
```
if ((AcpiTableInfo->AcpiTableRevision < This->MinAcpiTableRevision) ||
    (AcpiTableInfo->AcpiTableRevision > This->AcpiTableRevision))
```

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,738 @@
...
+  return EFI_SUCCESS;
+}
+
+/** Add the MemProxDomainAttr Information record to the HMAT Table.
+
+  @param [in]  CfgMgrProtocol                   Pointer to the Configuration Manager Protocol Interface.
+  @param [in]  MemProximityDomainAttrInfo       Pointer to a Proximity Domain Attribute array in an HMAT table.
+                                                The caller is responsible for allocating enough memory to fit
+                                                the data.
+  @param [in]  CmMemProximityDomainAttrInfo     Pointer ConfigurationManager objects array wuth Domain Attribute data.
```
pierregondois: `with`

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,738 @@
...
+             CfgMgrProtocol,
+             CM_NULL_TOKEN,
+             &ProximityDomainInfo,
+             &ProximityDomainInfoCount
+             );
+
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "ERROR: HMAT: Failed to get MemProximityDomain Info. Status = %r\n",
```
pierregondois: `MemProximityDomain` -> `ProximityDomainInfo`

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
```diff
@@ -0,0 +1,738 @@
...
+/** Computes how many bytes are required to hold an ACPI HMAT SSLBI table.
+
+  @param [in] InitiatorProximityDomainCount Number of initiator domains in an SSLBI record.
+  @param [in] TargetProximityDomainCount    Number of target domains in an SSLBI record.
+
+  @retval Bytes required to store SSLBI record.
+**/
+STATIC
+UINT32
+GetSllbiTableSize (
```
pierregondois: `GetSllbiTableSize` -> `GetSslbiTableSize` ?

pierregondois (commented):
Hello Oleksandr,
I just come back from a 2 weeks break, sorry for the delay.
Just a few typos I think, otherwise all good for me

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: Fixed

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: Fixed

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: Fixed

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: Fixed

DynamicTablesPkg/Library/Acpi/Common/AcpiHmatLib/HmatGenerator.c
gonzoua: Done.

pierregondois (approved):

Issue #10495, opened 2024-03-11
EFI_SIMPLE_NETWORK_PROTOCOL.Reset() returns incorrect status code (Bugzilla Bug 4729)


tianocore-issues:
This issue was created automatically with [bugzilla2github](https://github.com/makubacki/bugzilla2github/tree/bz2gh_tianocore)
# Bugzilla Bug 4729
Date: 2024-03-11T08:55:49+00:00
From: @nicklela
To:   @nicklela
CC:   @lgao4

Last updated: 2024-07-11T03:46:22+00:00

tianocore-issues:
## Comment 22776
Date: 2024-03-11 08:55:49 +0000
From: @nicklela

- Industry Specification: ---
- Release Observed: EDK II Master
- Releases to Fix: EDK II Master
- Target OS: ---
- Bugzilla Assignee(s): @nicklela

During SCT test, SCT reports below failure while testing EFI_SIMPLE_NETWORK_PROTOCOL.Reset() function:

------------------------------------------------------------
UEFI 2.6
Test Configuration #0
------------------------------------------------------------
Function Test for Reset
------------------------------------------------------------
Logfile: &quot;\_EFI\BOOT\bbr\SCT\Log\NetworkSupportTest\SimpleNetworkProtocolTest0\R
eset_Func_0_0_BA628A58-E318-4E4A-BAC2-A7833ADF42D9.log&quot;
Test Started: 02/15/24  04:56a
------------------------------------------------------------
ExtendedVerification = 0 is not implemented!
EFI_SIMPLE_NETWORK_PROTOCOL.Reset - Invoke Reset() with ExtendedVerification is 
FALSE and verify interface correctness within test case -- FAILURE
30314E89-DB26-4B01-90F3-04D31B19A601
/home/runner/work/arm-systemready/arm-systemready/SR/scripts/edk2-test/uefi-sct/
SctPkg/TestCase/UEFI/EFI/Protocol/SimpleNetwork/BlackBoxTest/SimpleNetworkBBTest
Function.c:525:Status - Success

ExtendedVerification = 0 is not implemented!
EFI_SIMPLE_NETWORK_PROTOCOL.Reset - Invoke Reset() with ExtendedVerification is 
TRUE and verify interface correctness within test case -- PASS
A3135B96-F9C6-45B6-AE87-15CAAE317EFB
/home/runner/work/arm-systemready/arm-systemready/SR/scripts/edk2-test/uefi-sct/
[...]

PR #5901, opened 2024-07-11
NetworkPkg/SnpDxe: return error for unsupported parameter

nicklela:
# Description

REF:https://bugzilla.tianocore.org/show_bug.cgi?id=4729

From SCT testing report, Reset() does not support the case when ExtendedVerification is set to FALSE. So, we should return EFI_INVALID_PARAMETER in this case. For details, please refer to Bug 4729.

- [ ] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [ ] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

## How This Was Tested

N/A

## Integration Instructions

N/A

PR #12593, opened 2026-05-19
NetworkPkg/SnpDxe: Fix UEFI SCT regression in SnpUndi32Reset

lewischendell:
# Description

Fix UEFI SCT regression in SnpUndi32Reset caused by upstream EDK2 spec alignment

Upstream EDK2 changed SnpUndi32Reset to return EFI_INVALID_PARAMETER when ExtendedVerification is FALSE per spec alignment (edk2 #10495, PR #5901).
This caused UEFI SCT failures in test cases 5.11.1.4.1 and 5.11.1.4.2 which expected EFI_SUCCESS.

Remove early EFI_INVALID_PARAMETER return and retain only warning message to allow normal SNP reset flow via PxeReset(), returning EFI_SUCCESS on successful reset operation.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

  1. Running UEFI SCT (System Configuration Test) - Specifically the Network Protocols SNP/PXE/BIS  test suite
  2. Targeted test cases - Test cases 5.11.1.4.1 and 5.11.1.4.2 in14_Network_Protocols_SNP_PXE_BIS_Test.md
  3. Verification - Confirming that after removing the early EFI_INVALID_PARAMETER return, the SnpUndi32Reset() function now returns EFI_SUCCESS when ExtendedVerification is FALSE, allowing the SCT tests to pass

## Integration Instructions

N/A

### DIFF
--- a/NetworkPkg/SnpDxe/Reset.c
+++ b/NetworkPkg/SnpDxe/Reset.c
@@ -103,7 +103,6 @@ SnpUndi32Reset (
   //
   if (!ExtendedVerification) {
     DEBUG ((DEBUG_WARN, "ExtendedVerification = %d is not implemented!\n", ExtendedVerification));
-    return EFI_INVALID_PARAMETER;
   }
 
   if (This == NULL) {


leiflindholm:
cc: @nicklela 

@lewischendell I think this change is correct, but I think we need to untangle why the previous change was referred to as required to resolve another SCT error.

lewischendell:
Hi @leiflindholm , @nicklela 
This PR addresses the same SCT test case, but I’m unclear why the previous code change was needed—it appears to break the SCT test. Does ARM use a different SCT implementation for validation?
The previous change causes the SCT test to fail immediately. Returning EFI_INVALID_PARAMETER from SNP.Reset() leads the test to raise EFI_TEST_ASSERTION_FAILED, as seen in the SCT error-handling logic below:

https://github.com/tianocore/edk2-test/blob/17bf62e0302770f27ecdd0267d566d4d386b6137/uefi-sct/SctPkg/TestCase/UEFI/EFI/Protocol/SimpleNetwork/BlackBoxTest/SimpleNetworkBBTestFunction.c#L510
```c
  // Assertion Point 4.4.2.1
  // Call Reset() function with ExtendedVerification is FALSE.
 Status = SnpInterface->Reset (SnpInterface, FALSE);
  if (EFI_ERROR(Status)) {
    if (EFI_UNSUPPORTED == Status) {
      AssertionType = EFI_TEST_ASSERTION_PASSED;
    } else {
      AssertionType = EFI_TEST_ASSERTION_FAILED;
    }
  }
``` 
Before applying this PR, SCT log reported an Invalid Parameter error, which causes the test to fail. After this PR, the return value is EFI_SUCCESS, and the test passes.

**Before PR - Test Failed:** [Reset_Func_0_0_BA628A58-E318-4E4A-BAC2-A7833ADF42D9_BeforePR_Failed.log](https://github.com/user-attachments/files/29299797/Reset_Func_0_0_BA628A58-E318-4E4A-BAC2-A7833ADF42D9_BeforePR_Failed.log)
```
Device Path: Acpi(PNP0A03,0)/Pci(1F|6)/Mac(888888888788)

EFI_SIMPLE_NETWORK_PROTOCOL.Reset - Invoke Reset() with ExtendedVerification is FALSE and verify interface correctness within test case -- FAILURE
30314E89-DB26-4B01-90F3-04D31B19A601
/home/lewisc/tianocore/edk2/SctPkg/TestCase/UEFI/EFI/Protocol/SimpleNetwork/BlackBoxTest/SimpleNetworkBBTestFunction.c:566:Status - Invalid Parameter
```
**After PR - Test Pass:** [Reset_Func_0_0_BA628A58-E318-4E4A-BAC2-A7833ADF42D9_AfterPR_Pass.log](https://github.com/user-attachments/files/29299930/Reset_Func_0_0_BA628A58-E318-4E4A-BAC2-A7833ADF42D9_AfterPR_Pass.log)
```
Device Path: Acpi(PNP0A03,0)/Pci(1F|6)/Mac(888888888788)

EFI_SIMPLE_NETWORK_PROTOCOL.Reset - Invoke Reset() with ExtendedVerification is FALSE and verify interface correctness within test case -- PASS
30314E89-DB26-4B01-90F3-04D31B19A601
/home/lewisc/tianocore/edk2/SctPkg/TestCase/UEFI/EFI/Protocol/SimpleNetwork/BlackBoxTest/SimpleNetworkBBTestFunction.c:566:Status - Success
```

**Side note:**
The failure in SimpleNetworkBBTestConformance for SNP.Reset() appears to have the same root cause—any unexpected return value (e.g., EFI_INVALID_PARAMETER) is treated as a failure unless it is explicitly allowed (such as EFI_UNSUPPORTED or EFI_NOT_STARTED).
https://github.com/tianocore/edk2-test/blob/17bf62e0302770f27ecdd0267d566d4d386b6137/uefi-sct/SctPkg/TestCase/UEFI/EFI/Protocol/SimpleNetwork/BlackBoxTest/SimpleNetworkBBTestConformance.c#L454
```c
  //
  // Assertion Point 5.4.2.1
  // Call Reset() function when network interface not start.
  //
  Status = SnpInterface->Reset (SnpInterface, FALSE);
  if ((Status == EFI_NOT_STARTED) && (SnpInterface->Mode->State == EfiSimpleNetworkStopped)) {
    AssertionType = EFI_TEST_ASSERTION_PASSED;
  } else {
    if (EFI_UNSUPPORTED == Status) {
      AssertionType = EFI_TEST_ASSERTION_PASSED;
    } else {
      AssertionType = EFI_TEST_ASSERTION_FAILED;
    }
  }
```

NetworkPkg/SnpDxe/Reset.c
```diff
@@ -103,7 +103,6 @@ SnpUndi32Reset (
   //
   if (!ExtendedVerification) {
     DEBUG ((DEBUG_WARN, "ExtendedVerification = %d is not implemented!\n", ExtendedVerification));
```
nicklela, line 105:
```
@retval EFI_INVALID_PARAMETER One or more of the parameters has an unsupported value.
```
At line 85, it says that this function returns invalid parameter when one or more of the parameters has an unsupported value. 

At line 105, when `ExtendedVerification = FALSE`, it says that this parameter value is not implemetned (not supported), we should return EFI_INVALID_PARAMETR to follow UEFI specification.

PR #12595, opened 2026-05-19
OvmfPkg: Seperate OvmfPkg reset vectors from UefiCpuPkg

### LINKED ISSUE #12333: [Bug]: Remove -I from OvmfPkg reset vectors

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

OvmfPkg

_No response_

### Current Behavior

OvmfPkg reset vector modules include files from UefiCpuPkg as follows:

https://github.com/tianocore/edk2/blob/183b578ec6b476108d75ac0d22c2016ab8111fc0/OvmfPkg/ResetVector/ResetVector.inf#L32-L34

https://github.com/tianocore/edk2/blob/183b578ec6b476108d75ac0d22c2016ab8111fc0/OvmfPkg/Bhyve/ResetVector/ResetVector.inf#L32-L34

https://github.com/tianocore/edk2/blob/183b578ec6b476108d75ac0d22c2016ab8111fc0/OvmfPkg/XenResetVector/XenResetVector.inf#L32-L34

This was brought up as an issue during a recent change: https://github.com/tianocore/edk2/pull/12312#discussion_r2966590447

### Expected Behavior

The required files need to either be copied into OvmfPkg or OvmfPkg needs to use the ResetVector component from UefiCpuPkg.

### Steps To Reproduce

See code links above.

### Build Environment

```markdown
N/A
```

### Version Information

```text
Commit: 183b578
Stable Tag: edk2-stable202602
```

### Urgency

Low

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

_No response_

khaliid2040:
# Description

The current OvmfPkg reset vectors  directly include and reference UefiCpuPkg reset vector. This has been brought up on #12312 as violation to edk2 packaging standards. A package shouldn't reference the private code of another package because it is a layering violation. So, this PR resolves that violation and makes all OvmfPkg reset vectors to be independent of UefiCpuPkg.

Also, according to DEC specification 1.27:

>All paths must be relative to the directory that contains the DEC file. Use of absolute paths, or WORKSPACE relative paths is prohibited.

### Details of the PR

This PR achieves the seperation by first copying referenced files from UefiCpuPkg (which are about 12 files). In order to minimize the duplication all OvmfPkg reset vectors will share those files internally independent of UefiCpuPkg. To miantain the organization of the include directory two directories have been introduced Ia16 which houses common 16-bit code that all reset vectors may reference and and Ia32 which houses 32-bit and common code.

Thus, all OvmfPkg reset vectors have been forwarded to reference that local include directory relative to package directory instead using `WORKSPACE` or absolute paths. No runtime changes have been introduced.

All of these has been done as per discussions in the linked issue.

Resolves: #12333 
- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Built locally and project CI.

## Integration Instructions

N/A

### DIFF
--- a/OvmfPkg/Bhyve/ResetVector/ResetVector.inf
+++ b/OvmfPkg/Bhyve/ResetVector/ResetVector.inf
@@ -29,10 +29,6 @@
   MdePkg/MdePkg.dec
   UefiCpuPkg/UefiCpuPkg.dec
 
-[BuildOptions]
-   *_*_IA32_NASMB_FLAGS = -I$(WORKSPACE)/UefiCpuPkg/ResetVector/Vtf0/
-   *_*_X64_NASMB_FLAGS = -I$(WORKSPACE)/UefiCpuPkg/ResetVector/Vtf0/
-
 [Pcd]
   gUefiOvmfPkgTokenSpaceGuid.PcdOvmfSecPageTablesBase
   gUefiOvmfPkgTokenSpaceGuid.PcdOvmfSecPageTablesSize
--- a/OvmfPkg/Bhyve/ResetVector/ResetVector.nasmb
+++ b/OvmfPkg/Bhyve/ResetVector/ResetVector.nasmb
@@ -32,16 +32,16 @@
   %error "Either ARCH_IA32 or ARCH_X64 must be defined."
 %endif
 
-%include "CommonMacros.inc"
+%include "Ia32/CommonMacros.inc"
 
-%include "PostCodes.inc"
+%include "Ia32/PostCodes.inc"
 
 %ifdef DEBUG_PORT80
-  %include "Port80Debug.nasm.inc"
+  %include "Ia32/Port80Debug.nasm.inc"
 %elifdef DEBUG_SERIAL
-  %include "SerialDebug.nasm.inc"
+  %include "Ia32/SerialDebug.nasm.inc"
 %else
-  %include "DebugDisabled.nasm.inc"
+  %include "Ia32/DebugDisabled.nasm.inc"
 %endif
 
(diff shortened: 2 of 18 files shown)


tlendacky:
There's already a version of ResetVectorVtf0.nasm.inc in the OvmfPkg/ResetVector/Ia16 directory. Should you really be creating another one? Should you just put the files in the appropriate arch directories in the OvmfPkg/ResetVector/ directory (without replacing the ones that are already present)?

I really hate all this code duplication. Is there a better way to do this rather than copying all those files?

khaliid2040:
> There's already a version of ResetVectorVtf0.nasm.inc in the OvmfPkg/ResetVector/Ia16 directory. Should you really be creating another one? Should you just put the files in the appropriate arch directories in the OvmfPkg/ResetVector/ directory (without replacing the ones that are already present)?

There is no replacement. The files that OvmfPkg reset vectors were referencing from UefiCpuPkgare only copied to OvmfPkg, the other files have been kept as they are.
> 
> I really hate all this code duplication. Is there a better way to do this rather than copying all those files?

Unfortunately, this is the best way we could make them share the code. But we could move all these files to OvmfPkg/ResetVector, however in that case we should use `WORKSPACE` or absolute path which DEC spec prohibits. Even if we do it the way you suggested then still we have to copy each file to each module sub-directories.

The approach this PR took was to seperate OvmfPkg reset vectors from UefiCpuPkg while minimizing the duplication by making reset vectors share those copied files in central place like `OvmfPkg/Include/Ia32` and `OvmfPkg/Include/Ia16`.

PR #12596, opened 2026-05-19
UnitTestFrameworkPkg: Add CMocka MockUefiRuntimeServicesTableLib

vasudevansambandan:
Unit tests exercising code that calls UEFI Runtime Services have no way to control or verify gRT calls without a mock implementation. No such mock existed in UnitTestFrameworkPkg for CMocka-based tests.

This adds a CMocka mock library for UefiRuntimeServicesTableLib covering all Runtime Services: GetTime, SetTime, GetWakeupTime, SetWakeupTime, SetVirtualAddressMap, ConvertPointer, GetVariable, GetNextVariableName, SetVariable, GetNextHighMonotonicCount, ResetSystem, UpdateCapsule, QueryCapsuleCapabilities, and QueryVariableInfo.

Each mock validates input parameters per the UEFI specification before delegating to CMocka mock() and will_return(). A static gMockRuntime table is assigned to gRT at link time. Helper macros in the header reduce will_return() and expect_*() boilerplate for success, EFI_BUFFER_TOO_SMALL, and error cases.

### DIFF
--- a/UnitTestFrameworkPkg/Include/Library/MockUefiRuntimeServicesTableLib.h
+++ b/UnitTestFrameworkPkg/Include/Library/MockUefiRuntimeServicesTableLib.h
@@ -1,0 +1,878 @@
+/** @file MockUefiRuntimeServicesTableLib.h
+  Mock Unit Test UEFI Runtime Services Table Library Implementation for Unit Testing
+
+  This library provides comprehensive mocking of UEFI Runtime Services
+  functions using CMocka framework. Each function can return data based
+  on mock() implementation.
+
+  Copyright (c) 2026, American Megatrends International LLC. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#ifndef __MOCK_UEFI_RUNTIME_SERVICES_TABLE_LIB_H__
+#define __MOCK_UEFI_RUNTIME_SERVICES_TABLE_LIB_H__
+
+#include <Uefi.h>
+#include <Library/BaseLib.h>
+
+/**
+  Returns the current time and date information, and the time-keeping capabilities
+  of the hardware platform.
+
+  @param[out]  Time         A pointer to storage to receive a snapshot of the current time.
+  @param[out]  Capabilities An optional pointer to a buffer to receive the real time clock
+                            device's capabilities.
+
+  @retval EFI_SUCCESS           The operation completed successfully.
+  @retval EFI_INVALID_PARAMETER Time is NULL.
+  @retval EFI_DEVICE_ERROR      The time could not be retrieved due to hardware error.
+**/
+EFI_STATUS
+EFIAPI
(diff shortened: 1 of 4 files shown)


vasudevansambandan:
@spbrogan @mdkinney @makubacki Kindly review the pull request and share your feedback.

UnitTestFrameworkPkg/Include/Library/MockUefiRuntimeServicesTableLib.h
makubacki: Have you seen the existing `MockUefiRuntimeServicesTableLib.h` file?

https://github.com/tianocore/edk2/blob/master/MdePkg/Test/Mock/Include/GoogleTest/Library/MockUefiRuntimeServicesTableLib.h

spbrogan (changes requested):
In Edk2 we don't have super clear guidance or examples about mocking (and how to do it) but I think this library implementation introduces a lot of unnecessary complexity and goes against the idea of mocking.  Although you are following the UEFI spec for return codes the idea of unit testing and using mocking is that return codes and call execution of those dependencies are entirely managed within your test.  For example, if you want the API to return efi invalid parameter when a null is passed for a certain parameter then the solution is to do that using the mock rather than expecting the functional code in the library to do that.

spbrogan:
this is how it is done with the google test framework. 

https://github.com/tianocore/edk2/blob/master/MdePkg/Test/Mock/Library/GoogleTest/MockUefiRuntimeServicesTableLib/MockUefiRuntimeServicesTableLib.cpp

vasudevansambandan:
@spbrogan  Thank you for your feedback. Sorry for the delayed response. Here is our update:

Unit test mock implementations align with edk2 practices. Key points:
 
1. **Parameter validation in mocks** – We return error statuses (e.g., `EFI_INVALID_PARAMETER`) rather than using `assert_non_null()` because it provides better visibility into how production code handles and propagates errors, potentially exposing multiple issues in a single test run instead of stopping at the first failure.
 
2. **Alignment with edk2** – We compared our `MockGetVariable()`/`MockSetVariable()` against edk2's SecureBootVariableLib tests. Both perform similar validation (required parameter checks, `check_expected_ptr()`, buffer validation, error path handling like `EFI_NOT_FOUND`/`EFI_BUFFER_TOO_SMALL`). Our approach follows the same philosophy.
 
3. **Common vs. local mock libraries** – We recommend continuing to use common mock libraries (e.g., `MockUefiRuntimeServicesTableLib`) over local per-test mocks to avoid duplication, ensure consistency, and simplify maintenance.
 
4. **Helper macros** – Providing helper macros for common success/error scenarios improves test readability, encourages comprehensive error-path coverage, and ensures consistent testing across suites.

vasudevansambandan:
@spbrogan @dubek @elmarco @fishman Please review my latest comment and share your feedback.

PR #12598, opened 2026-05-20
ArmPkg: MmCommunicationDxe: Fix Comm Buffer Init

os-d:
# Description

Currently, MmCommunicationDxe expects the MM comm buffer to either be not added by the platform or added and allocated by the platform. However, not all platforms follow this pattern.

This commit makes the handling more generic by checking to see if a GCD descriptor exists that covers this range. If it does, the capabilities are updated as needed. If only a partial desc exists for this range, the driver will fail as that is a platform misconfiguration.

If the descriptor does not exist, the driver will add the buffer.

Then, for all successful cases, the attributes are updated on the buffer to what the driver expects them to be.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on a physical ARM64 platform that was failing to boot due to the MM comm buffer being described in a resource descriptor HOB but not allocated. After this change, it successfully booted. Also did a regression test on another ARM64 physical platform that was succeeding with the existing logic and saw it boot.

## Integration Instructions

Platforms should either not produce a resource descriptor HOB for the MM comm buffer region or should produce a reserved resource descriptor HOB that covers the entire MM comm buffer region.

### DIFF
--- a/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
+++ b/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
@@ -5,6 +5,9 @@
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
+
+#include <PiDxe.h>
+
 #include <Library/ArmLib.h>
 #include <Library/ArmFfaLib.h>
 #include <Library/ArmSmcLib.h>
@@ -23,6 +26,8 @@
 #include <IndustryStandard/ArmStdSmc.h>
 #include <IndustryStandard/ArmFfaSvc.h>
 #include <IndustryStandard/MmCommunicate.h>
+
+#define COMM_BUFFER_ATTRS  (EFI_MEMORY_WB | EFI_MEMORY_XP | EFI_MEMORY_RUNTIME)
 
 //
 // Partition ID if FF-A support is enabled
@@ -690,8 +695,9 @@ MmCommunication2Initialize (
   IN EFI_SYSTEM_TABLE  *SystemTable
   )
 {
-  EFI_STATUS  Status;
-  UINTN       Index;
+  EFI_STATUS                       Status;
+  UINTN                            Index;
+  EFI_GCD_MEMORY_SPACE_DESCRIPTOR  GcdDescriptor;
 
   // Initialize to make mm communication
   Status = InitializeCommunication ();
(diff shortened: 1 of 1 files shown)


LeviYeoReum (approved):

ardbiesheuvel (approved):

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
```diff
@@ -708,83 +714,74 @@ MmCommunication2Initialize (
...
+      ));
+    ASSERT (FALSE);
+    goto ReturnErrorStatus;
+  } else if (GcdDescriptor.GcdMemoryType != EfiGcdMemoryTypeReserved) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a: Existing GCD memory space is not EfiGcdMemoryTypeReserved for the MM-NS Buffer Memory Space\n",
+      __func__
+      ));
+    ASSERT_EFI_ERROR (Status);
```
samimujawar: Would Status be EFI_SUCCESS here in this case?

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
```diff
@@ -708,83 +714,74 @@ MmCommunication2Initialize (
...
+               ((GcdDescriptor.BaseAddress + GcdDescriptor.Length) >=
+                (mNsCommBuffMemRegion.PhysicalBase + mNsCommBuffMemRegion.Length))))
+  {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a: Existing GCD memory space does not completely cover MM-NS Buffer Memory Space\n",
+      __func__
+      ));
+    ASSERT (FALSE);
+    goto ReturnErrorStatus;
```
samimujawar, line 751: Should Status be updated here to reflect the error?
Also, the purpose of 'goto ReturnErrorStatus' in the original code was to remove the memory space. Since, this has now been removed, would it be better to return from the function with the error code?

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
```diff
@@ -708,83 +714,74 @@ MmCommunication2Initialize (
...
+             (COMM_BUFFER_ATTRS))
+  {
+    // set the capabilities if we need to. If we fail, still try setting the attributes
+    // the worst case is it will error there.
+    Status = gDS->SetMemorySpaceCapabilities (
+                    mNsCommBuffMemRegion.PhysicalBase,
+                    mNsCommBuffMemRegion.Length,
+                    COMM_BUFFER_ATTRS
+                    );
+    ASSERT_EFI_ERROR (Status);
```
samimujawar: Should we remove the assert and replace it with a warning message? Otherwise, the debug builds will stop the execution and the remaining code path would never be tested. Which also means the debug and the release build behaviour would be different.

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
os-d: Yes, good catch, thanks, I'll update this to ASSERT(FALSE)

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
os-d: Yeah, here and below I'll set the Status so we return a failure

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
os-d: My rationale here was that we want to catch in debug builds if SetMemorySpaceCapabilities() fails, so we can fix the code. However, calling SetMemorySpaceAttributes() without the correct capabilities is safe, you'll just get an error say unsupported. So then the release code would return an error.

So, same semantics, just gives an opportunity to catch it in the release code. But, I could also add a goto ReturnErrorStatus if SetMemorySpaceCapabilities() fails and leave the assert. Then the same behavior (more or less) and we can still catch it in debug builds.

TBH, the fall through doesn't really matter, I've done that in other places not because I've ever hit it, just because in production I don't like to fail unless I really need to; since SetMemorySpaceCapabilities() isn't necessarily required (who knows why it failed, maybe the capability did get set and then it failed), might as well go to SetMemorySpaceAttributes(), which is the case that actually needs to succeed or else we do have to fail.

So, I'll check for failure and goto ReturnErrorStatus here, in practice that is what is going to happen anyway.

os-d:
@samimujawar I have addressed your comments, please re-review

samimujawar (approved):

makubacki:
@mergifyio queue

makubacki:
@mergifyio refresh

PR #12599, opened 2026-05-21
Tpl fix v3

ardbiesheuvel:
# Description

Alternative proposal for preventing unbounded recursion in the timer interrupt handler.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

OVMF boot on KVM

## Integration Instructions

Downstream implementations of timer interrupt handlers should be updated to avoid manipulating the TPL before invoking the callback provided by the DXE core, if the unbounded recursion guarantee is needed.

Without such changes, everything will work exactly as before, given that `CoreTimerTick()` will perform a no-op TPL raise/restore of TPL_HIGH_LEVEL to TPL_HIGH_LEVEL and vice versa when called with the TPL already at TPL_HIGH_LEVEL.

### DIFF
--- a/ArmPkg/Drivers/TimerDxe/TimerDxe.c
+++ b/ArmPkg/Drivers/TimerDxe/TimerDxe.c
@@ -289,16 +289,8 @@ TimerInterruptHandler (
   IN  EFI_SYSTEM_CONTEXT         SystemContext
   )
 {
-  EFI_TPL  OriginalTPL;
-  UINT64   CurrentValue;
-  UINT64   CompareValue;
-
-  //
-  // DXE core uses this callback for the EFI timer tick. The DXE core uses locks
-  // that raise to TPL_HIGH and then restore back to current level. Thus we need
-  // to make sure TPL level is set to TPL_HIGH while we are handling the timer tick.
-  //
-  OriginalTPL = gBS->RaiseTPL (TPL_HIGH_LEVEL);
+  UINT64  CurrentValue;
+  UINT64  CompareValue;
 
   // Signal end of interrupt early to help avoid losing subsequent ticks
   // from long duration handlers
@@ -333,8 +325,6 @@ TimerInterruptHandler (
     ArmGenericTimerReenableTimer ();
     ArmInstructionSynchronizationBarrier ();
   }
-
-  gBS->RestoreTPL (OriginalTPL);
 }
 
 /**
--- a/EmulatorPkg/CpuRuntimeDxe/Cpu.c
+++ b/EmulatorPkg/CpuRuntimeDxe/Cpu.c
@@ -36,7 +36,6 @@ CPU_ARCH_PROTOCOL_PRIVATE  mCpuTemplate = {
       CpuIoServiceWrite
     }
   },
-  TRUE
 };
 
 #define EFI_CPU_DATA_MAXIMUM_LENGTH  0x100
@@ -277,10 +276,6 @@ EmuEnableInterrupt (
   IN EFI_CPU_ARCH_PROTOCOL  *This
   )
 {
-  CPU_ARCH_PROTOCOL_PRIVATE  *Private;
-
-  Private                 = CPU_ARCH_PROTOCOL_PRIVATE_DATA_FROM_THIS (This);
-  Private->InterruptState = TRUE;
   gEmuThunk->EnableInterrupt ();
   return EFI_SUCCESS;
 }
@@ -291,10 +286,6 @@ EmuDisableInterrupt (
   IN EFI_CPU_ARCH_PROTOCOL  *This
   )
 {
-  CPU_ARCH_PROTOCOL_PRIVATE  *Private;
-
-  Private                 = CPU_ARCH_PROTOCOL_PRIVATE_DATA_FROM_THIS (This);
-  Private->InterruptState = FALSE;
   gEmuThunk->DisableInterrupt ();
   return EFI_SUCCESS;
 }
@@ -306,14 +297,11 @@ EmuGetInterruptState (
   OUT BOOLEAN               *State
   )
 {
-  CPU_ARCH_PROTOCOL_PRIVATE  *Private;
-
   if (State == NULL) {
     return EFI_INVALID_PARAMETER;
   }
 
-  Private = CPU_ARCH_PROTOCOL_PRIVATE_DATA_FROM_THIS (This);
-  *State  = Private->InterruptState;
+  *State = gEmuThunk->GetInterruptState ();
   return EFI_SUCCESS;
 }
 
--- a/EmulatorPkg/CpuRuntimeDxe/CpuDriver.h
+++ b/EmulatorPkg/CpuRuntimeDxe/CpuDriver.h
@@ -45,7 +45,6 @@ typedef struct {
   //
   // Local Data for CPU interface goes here
   //
-  BOOLEAN                  InterruptState;
 } CPU_ARCH_PROTOCOL_PRIVATE;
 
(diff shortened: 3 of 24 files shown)


ardbiesheuvel:
@mdkinney @os-d updated version of my minimal counterproposal.

Note that there are no functional changes as a result *unless* the timer interrupt handler invokes `CoreTimerTick()` with a TPL below `TPL_HIGH_LEVEL`, which is what is being proposed for `LocalApicTimerDxe` already. In that case, re-enabling interrupts is deferred to `ERET`, producing the behavior that we need in order to avoid unbounded recursion.

In all other cases, things work exactly as before, without the need to infer from the interrupt state whether or not the caller may be an interrupt handler, and without breaking CPU arch protocol implementations that [erroneously] cache the interrupt state.

os-d:
I had an offline conversation with @mdkinney last week on this topic, summarizing my thoughts here that I shared then, as well as from the other PR:
- In general, I favor this approach more than the other proposed approach. I like that it is opt in (mostly; I am asking a teammate to test on a physical Intel system, I don't have the ability to do so for edk2 changes, I want to make sure that this works with HpetDxe; when I was prototyping some changes on Patina that mirrored the original proposed edk2 changes, I saw a physical Intel system hang when HpetDxe was launched; however, on non-Intel platforms, both virtual and physical, the changes worked and in fact late last week I was accidentally able to reproduce the interrupt burst case in an emulator and my Patina changes did fix it and allow forward progress. HpetDxe is an example of a timer interrupt handler that doesn't raise/restore TPL itself, so it would be affected by the behavior change here). I think this scopes the risk down. Also, frankly I think this change is much less complex than the original change and with the latest updates to it, I do think it covers the various scenarios (again, need to do some more testing here).
- That being said, Mike had raised the concern that CoreRestoreTPL() does have the inherent issue that it can cause a stack overflow, which is not ideal to leave in. However, because the issue is currently only on virtual platforms (currently ignoring the scenarios where a physical platforms takes too long in an  interrupt handler and so we take the timer interrupt before returning because nothing makes that platform boot, we've just turned a stack overflow into a hang, which is a net positive, but again, we still don't boot, the platform needs fixes) and a subset of them at that, I'm more okay with leaving the potential problem in CoreRestoreTPL(). But, I can certainly accept the argument that CoreRestoreTPL() itself should be made more resilient in case of interrupt controller HW changes in the future, etc., though this is of course hard to predict.
- TL;DR, from a personal standpoint I lean towards this implementation for simplicity and less risk, though I accept the original proposal as well, trading a broader fix for more complexity/risk. Mostly, I'd love to see more test cases here, hopefully the ones Mike has can be iterated upon to better test these cases. I think there is certainly a path there; I haven't had cycles to help craft those tests. Once we've decided on a path, I can help do some additional testing.

ardbiesheuvel:
> * In general, I favor this approach more than the other proposed approach. I like that it is opt in (mostly; I am asking a teammate to test on a physical Intel system, I don't have the ability to do so for edk2 changes, I want to make sure that this works with HpetDxe; when I was prototyping some changes on Patina that mirrored the original proposed edk2 changes, I saw a physical Intel system hang when HpetDxe was launched; however, on non-Intel platforms, both virtual and physical, the changes worked and in fact late last week I was accidentally able to reproduce the interrupt burst case in an emulator and my Patina changes did fix it and allow forward progress. HpetDxe is an example of a timer interrupt handler that doesn't raise/restore TPL itself, so it would be affected by the behavior change here). I think this scopes the risk down. Also, frankly I think this change is much less complex than the original change and with the latest updates to it, I do think it covers the various scenarios (again, need to do some more testing here).

Exactly.

> * That being said, Mike had raised the concern that CoreRestoreTPL() does have the inherent issue that it can cause a stack overflow, which is not ideal to leave in.

It does not. `CoreRestoreTpl()` never calls itself recursively with the same TPL, only with a strictly higher one, and so its stack utilization is bounded. The only way it ever gets invoked recursively with the same TPL is over the back of a timer interrupt. So that is what we should fix.

> However, because the issue is currently only on virtual platforms (currently ignoring the scenarios where a physical platforms takes too long in an  interrupt handler and so we take the timer interrupt before returning because nothing makes that platform boot, we've just turned a stack overflow into a hang, which is a net positive, but again, we still don't boot, the platform needs fixes) and a subset of them at that, I'm more okay with leaving the potential problem in CoreRestoreTPL(). But, I can certainly accept the argument that CoreRestoreTPL() itself should be made more resilient in case of interrupt controller HW changes in the future, etc., though this is of course hard to predict.

VIrtual platforms are important, and specifically confidential compute, where the entity that is in charge of queuing interrupts is untrusted (the VMM). Note that this condition may be ephemeral, and resolve itself within a period of time that might hardly be noticeable by a person, in which case not crashing is a much better course of action.

> * TL;DR, from a personal standpoint I lean towards this implementation for simplicity and less risk, though I accept the original proposal as well, trading a broader fix for more complexity/risk. Mostly, I'd love to see more test cases here, hopefully the ones Mike has can be iterated upon to better test these cases. I think there is certainly a path there; I haven't had cycles to help craft those tests. Once we've decided on a path, I can help do some additional testing.

Thanks. Agree that the testing is tricky here. But I

os-d:
> > * That being said, Mike had raised the concern that CoreRestoreTPL() does have the inherent issue that it can cause a stack overflow, which is not ideal to leave in.
> 
> It does not. `CoreRestoreTpl()` never calls itself recursively with the same TPL, only with a strictly higher one, and so its stack utilization is bounded. The only way it ever gets invoked recursively with the same TPL is over the back of a timer interrupt. So that is what we should fix.
> 

Fair point.

> > However, because the issue is currently only on virtual platforms (currently ignoring the scenarios where a physical platforms takes too long in an  interrupt handler and so we take the timer interrupt before returning because nothing makes that platform boot, we've just turned a stack overflow into a hang, which is a net positive, but again, we still don't boot, the platform needs fixes) and a subset of them at that, I'm more okay with leaving the potential problem in CoreRestoreTPL(). But, I can certainly accept the argument that CoreRestoreTPL() itself should be made more resilient in case of interrupt controller HW changes in the future, etc., though this is of course hard to predict.
> 
> VIrtual platforms are important, and specifically confidential compute, where the entity that is in charge of queuing interrupts is untrusted (the VMM). Note that this condition may be ephemeral, and resolve itself within a period of time that might hardly be noticeable by a person, in which case not crashing is a much better course of action.
> 

Completely agree virtual platforms are important and we should fix this problem for them. My point just being the problem is primarily for a subset of platforms which informs my assessment of the approach that should be taken to fix it. The confidential compute angle is a good one, I didn't think about that angle.

> > * TL;DR, from a personal standpoint I lean towards this implementation for simplicity and less risk, though I accept the original proposal as well, trading a broader fix for more complexity/risk. Mostly, I'd love to see more test cases here, hopefully the ones Mike has can be iterated upon to better test these cases. I think there is certainly a path there; I haven't had cycles to help craft those tests. Once we've decided on a path, I can help do some additional testing.
> 
> Thanks. Agree that the testing is tricky here. But I

Looks like we lost part of your message.

ardbiesheuvel:
> > > * TL;DR, from a personal standpoint I lean towards this implementation for simplicity and less risk, though I accept the original proposal as well, trading a broader fix for more complexity/risk. Mostly, I'd love to see more test cases here, hopefully the ones Mike has can be iterated upon to better test these cases. I think there is certainly a path there; I haven't had cycles to help craft those tests. Once we've decided on a path, I can help do some additional testing.
> > 
> > 
> > Thanks. Agree that the testing is tricky here. But I
> 
> Looks like we lost part of your message.

It fell out of my brain too :-(

Let's park this until Mike gets back next week.

PR #12605, opened 2026-05-22
BaseTools: Add support for preserving build ID

cfernald:
# Description

Adds an optional flag that copies the GNU build-id note from the input ELF file into the output PE/COFF firmware image as a dedicated ".bldid" section. The build ID is emitted by the linker as a unique fingerprint of the binary and allows custom post-build and debugging tools to reliably match a firmware image against its corresponding unstripped ELF and debug symbols, without relying on file names, timestamps, or build paths.

This notable opts to use a non-standard section name ".bldid" to store the build ID. This approach was chosen to keep genfw and the parsers simple since the full "build-id" name would require redirecting the section name. While this breaks from standard conventions, this is not impactful since GenFW is already creating a non-standard artifact for the PE image with the associated ELF symbol file.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested on QEMU builds with GCC and GenFW with the flag to maintain build ID. This was used to match binaries to symbols automatically. This change has been carried for many months in MU without regression.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Source/C/GenFw/Elf64Convert.c
+++ b/BaseTools/Source/C/GenFw/Elf64Convert.c
@@ -130,6 +130,8 @@ STATIC UINT32 mHiiRsrcOffset;
 STATIC UINT32 mRelocOffset;
 STATIC UINT32 mDebugOffset;
 STATIC UINT32 mExportOffset;
+STATIC UINT32 mBuildIdOffset;
+STATIC BOOLEAN mBuildIdFound;
 //
 // Used for RISC-V relocations.
 //
@@ -227,6 +229,10 @@ InitializeElf64 (
     ElfFunctions->WriteExport = WriteExport64;
   }
 
+  if (mBuildIdFlag) {
+    mCoffNbrSections++;
+  }
+
   return TRUE;
 }
 
@@ -288,6 +294,22 @@ IsHiiRsrcShdr (
   Elf_Shdr *Namedr = GetShdrByIndex(mEhdr->e_shstrndx);
 
   return (BOOLEAN) (strcmp((CHAR8*)mEhdr + Namedr->sh_offset + Shdr->sh_name, ELF_HII_SECTION_NAME) == 0);
+}
+
+STATIC
+BOOLEAN
+IsBuildIdShdr (
+  Elf_Shdr *Shdr
+  )
+{
+  Elf_Shdr *Namedr = GetShdrByIndex(mEhdr->e_shstrndx);
+
+  if (Namedr->sh_offset + Shdr->sh_name >= mFileBufferSize) {
+    Error (NULL, 0, 3000, "Invalid", "IsBuildIdShdr: Name offset %lu is larger then file size %lu", mEhdr->e_shstrndx, mFileBufferSize);
+    exit(EXIT_FAILURE);
+  }
+
+  return (BOOLEAN) (strcmp((CHAR8*)mEhdr + Namedr->sh_offset + Shdr->sh_name, ELF_BUILD_ID_SECTION_NAME) == 0);
 }
 
 STATIC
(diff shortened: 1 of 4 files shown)


BaseTools/Source/C/GenFw/Elf64Convert.c
```diff
@@ -227,6 +229,10 @@ InitializeElf64 (
     ElfFunctions->WriteExport = WriteExport64;
   }
 
+  if (mBuildIdFlag) {
+    mCoffNbrSections ++;
```
os-d:
```suggestion
    mCoffNbrSections++;
```
small nit but the space between scares me :)

BaseTools/Source/C/GenFw/Elf64Convert.c
```diff
@@ -290,6 +296,22 @@ IsHiiRsrcShdr (
...
+STATIC
+BOOLEAN
+IsBuildIdShdr (
+  Elf_Shdr *Shdr
+  )
+{
+  Elf_Shdr *Namedr = GetShdrByIndex(mEhdr->e_shstrndx);
+
+  if (Namedr->sh_offset + Shdr->sh_name >= mFileBufferSize) {
+    assert(FALSE);
```
os-d: I didn't look too far in the file, but my knowledge of BaseTools is that this pattern is more common than assert:

```c
Error (NULL, 0, 3000, "Invalid", "GetShdrByIndex: Index %u is too high.", Num);
exit(EXIT_FAILURE);
```

From line 250 (with additions) of this file

os-d (approved):
Approving as we have been using this for a year or so now and seeing it work well for identifying DWARF symbols in the pseudo-PE/COFF image.

As we've discussed offline, I don't love that we are renaming the section, but it is already a frankenimage and the only reasonable option.

BaseTools/Source/C/GenFw/Elf64Convert.c
```diff
@@ -290,6 +296,22 @@ IsHiiRsrcShdr (
...
+IsBuildIdShdr (
+  Elf_Shdr *Shdr
+  )
+{
+  Elf_Shdr *Namedr = GetShdrByIndex(mEhdr->e_shstrndx);
+
+  if (Namedr->sh_offset + Shdr->sh_name >= mFileBufferSize) {
+  Error (NULL, 0, 3000, "Invalid", "IsBuildIdShdr: Name offset %lu is larger then file size %lu", mEhdr->e_shstrndx, mFileBufferSize);
+  exit(EXIT_FAILURE);
+  }
```
makubacki, line 310: I assume it was not intentional to skip indentation in the block.


```suggestion
  if (Namedr->sh_offset + Shdr->sh_name >= mFileBufferSize) {
    Error (NULL, 0, 3000, "Invalid", "IsBuildIdShdr: Name offset %lu is larger then file size %lu", mEhdr->e_shstrndx, mFileBufferSize);
    exit(EXIT_FAILURE);
  }
```

BaseTools/Source/C/GenFw/GenFw.c
```diff
@@ -1576,6 +1581,13 @@ Routine Description:
       continue;
     }
 
+    if (stricmp (argv[0], "--build-id") == 0) {
+      mBuildIdFlag = TRUE;
+      argc --;
+      argv ++;
```
makubacki: For consistency with the other comment that called out the space.

```suggestion
      argc--;
      argv++;
```

makubacki (approved):

BaseTools/Source/C/GenFw/Elf64Convert.c
cfernald: Correct, my format on save messes up this file, and clearly i've become lazy without it.

ardbiesheuvel (approved):

PR #12606, opened 2026-05-22
BaseTools/Python: Remove error message with email address

vishalo:
# Description

The unhandled exception handlers in build.py, GenFds/GenFds.py, Trim/Trim.py, and the UPT tools (InstallPkg, InventoryWs, MkPkg, ReplacePkg, RmPkg, TestInstall) prints a message asking users to send email to `devel@edk2.groups.io` when an unexpected Python exception occurred:

`
Please send email to devel@edk2.groups.io for help,
attaching following call stack trace!`

Remove this message so that downstream builds do not direct users to the upstream edk2 mailing list for tool failures that may be unrelated to upstream edk2. Most if these errors are due to downstream modifications.

Also remove the now-unused MSG_EDKII_MAIL_ADDR constant from Common/DataType.py and the MSG_EDKII_MAIL_ADDR and MSG_SEARCH_FOR_HELP constants from UPT/Logger/StringTable.py.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
Build ArmVirPkg with incorrectly formatted dsc file and check error log.

## Integration Instructions
N/A


leiflindholm:
This PR is very welcome, but it's kind of missing the why:
When build.py encounters an error in a config file and prints a fairly unhelpful message with accompanying stack trace, the cause is nearly always downstream modifications, at which point there is not much point in doing what the old message says and email the mailing list.

vishalo:
> This PR is very welcome, but it's kind of missing the why: When build.py encounters an error in a config file and prints a fairly unhelpful message with accompanying stack trace, the cause is nearly always downstream modifications, at which point there is not much point in doing what the old message says and email the mailing list.

Thank you @leiflindholm  - updated the PR description.

leiflindholm (approved):
Yeah, the existing message hinders more than helps.

Ideally we should improve the error output, but this at least stops confusing downstream consumers (who really should know better than leaking their internal build config details, but...).

vishalo:
@gapalomi, @lgao4, @mikebeaton, @ponchofigueroa, @YuweiChen1110 - could you please review ?

lgao4 (approved):

vishalo:
@mergifyio queue

leiflindholm (approved):

vishalo:
Unable to set "Editable by maintainers".

szsam:
> Unable to set "Editable by maintainers".

You cannot enable this option for a fork owned by an organization. See https://docs.github.com/en/pull-requests/reference/forks

PR #12608, opened 2026-05-25
MdeModulePkg: Add DepexDumpLib for driver dependency dump support

LinusAtDell:
# Description

This library provides diagnostic capabilities for DXE driver dependency expressions (Depex). It
helps developers understand why drivers are not being dispatched by converting Depex RPN (Reverse
Polish Notation) streams into human-readable format and providing detailed dependency analysis.

Key features:
* DepexToInfixString(): Converts Depex RPN to parenthesized infix strings
* PrintDriverDepexSummary(): Provides comprehensive driver dependency analysis including missing
  protocols, dependency chains, and circular dependency detection
* PrintDriverDepexSummaryUniversal(): Universal version for both DXE and SMM drivers
* Pure formatting library with no runtime state modifications
* Debug output integration for driver dispatch troubleshooting

The library is integrated into the DXE dispatcher to automatically print dependency information when
drivers are discovered but not loaded.

## Checklist

* Breaking change?
  * [ ] **Breaking change** - Does this PR cause a break in build or boot behavior?
  * Examples: Does it add a new library class or move a module to a different repo.
* Impacts security?
  * [ ] **Security** - Does this PR have a direct security impact?
  * Examples: Crypto algorithm change or buffer overflow fix.
* Includes tests?
  * [ ] **Tests** - Does this PR include any explicit test code?
  * Examples: Unit tests or integration tests.

## How This Was Tested

The library was tested by:
1. Building successfully for X64 architecture in DEBUG and RELEASE configurations
2. Integrating into DXE core dispatcher to verify dependency dump output
3. Testing DepexToInfixString() with various Depex opcodes (AND, OR, NOT, PUSH, BEFORE, AFTER, TRUE, FALSE)
4. Verifying PrintDriverDepexSummary() output format and content
5. Testing with drivers that have missing dependencies to verify diagnostic output
6. Validating CHAR8/CHAR16 string handling consistency across all functions

## Integration Instructions

N/A - This is a new library that can be optionally used by platforms for driver dependency debugging.

### DIFF
--- a/MdeModulePkg/Core/Dxe/Dispatcher/DepexDebug.c
+++ b/MdeModulePkg/Core/Dxe/Dispatcher/DepexDebug.c
@@ -1,0 +1,771 @@
+/** @file
+  Depex debug helper for DXE Core.
+
+  This file provides diagnostic capabilities for DXE driver dependency
+  expressions (Depex). It helps developers understand why drivers are not
+  being dispatched by converting Depex RPN (Reverse Polish Notation) streams
+  into human-readable format and providing detailed dependency analysis.
+
+  Copyright (c) 2026, Dell Inc. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include "DxeMain.h"
+#include <Library/MemoryAllocationLib.h>
+#include <Library/PrintLib.h>
+#include <Library/DebugLib.h>
+#include <Library/UefiBootServicesTableLib.h>
+#include <Library/BaseLib.h>
+#include <Protocol/LoadedImage.h>
+
+//
+// Type definitions (local to this file)
+//
+typedef struct {
+  EFI_GUID    DriverGuid;
+  EFI_GUID    ProtocolGuid;
+  BOOLEAN     IsCycle;
+} DEPENDENCY_LINK;
+
+typedef struct {
+  DEPENDENCY_LINK    *Links;
+  UINTN              LinkCount;
+  UINTN              Capacity;
+  BOOLEAN            CycleDetected;
+  UINTN              CycleStartIndex;
+} DEPENDENCY_CHAIN;
+
+typedef struct {
+  EFI_GUID    FileNameGuid;
+  BOOLEAN     Discovered;
+  BOOLEAN     Scheduled;
+  BOOLEAN     Loaded;
+  BOOLEAN     Requested;
+  BOOLEAN     Unrequested;
+  BOOLEAN     Untrusted;
+  BOOLEAN     Initialized;
+  BOOLEAN     OpenProtoLocked;
+  BOOLEAN     DepexProtocolError;
+  BOOLEAN     LastEvalTrue;
+  UINT32      Round;
+} DEPEX_DIAG_DRIVER_STATE;
+
+typedef struct {
+  CHAR8    *Buffer;
+  UINTN    Capacity;
+  UINTN    Length;
+} STRING_BUILDER;
+
+//
+// Constants for string buffer sizes
+//
+#define DEPEX_STRING_BUFFER_SIZE       64
+#define DEPEX_INITIAL_CHAIN_CAPACITY   16
+#define DEPEX_CHAIN_GROWTH_MULTIPLIER  2
+#define DEPEX_STACK_SIZE               32
+#define DEPEX_MAX_GUID_STRING_LENGTH   64
+#define DEPEX_OUTPUT_BUFFER_SIZE       4096
+#define DEPEX_STRING_EXTRA_SIZE        4
+
+//
+// Constants for debug output formatting
+//
+#define DEPEX_DEBUG_INDENT_SPACING  2
+#define DEPEX_DEBUG_OTHER_FLAGS     "  Other flags:\n"
+
+// Standard Depex opcodes per PI spec
+#define EFI_DEP_BEFORE        0x00
+#define EFI_DEP_AFTER         0x01
+#define EFI_DEP_PUSH          0x02
+#define EFI_DEP_AND           0x03
+#define EFI_DEP_OR            0x04
+#define EFI_DEP_NOT           0x05
+#define EFI_DEP_TRUE          0x06
+#define EFI_DEP_FALSE         0x07
+#define EFI_DEP_END           0x08
+#define EFI_DEP_SOR           0x09
+#define EFI_DEP_REPLACE_TRUE  0xff
+
+#define MAX_TRACE_DEPTH  6
+
+// External variables from Dispatcher.c
+extern LIST_ENTRY  mDiscoveredList;
+
+// Internal cache variables
+STATIC DEPEX_DIAG_DRIVER_STATE  *mDriverStateCache = NULL;
+STATIC UINTN                    mDriverStateCount  = 0;
+
+/**
+  Convert a GUID to a string representation for debug output.
+
+  @param Guid       Pointer to the GUID to convert.
+  @param Buffer     Buffer to store the GUID string.
+  @param BufferSize Size of the buffer in bytes.
+
+  @retval Pointer to the GUID string, or NULL if conversion fails.
+**/
+STATIC
+CHAR8 *
+GuidToString (
+  IN EFI_GUID  *Guid,
+  IN CHAR8     *Buffer,
+  IN UINTN     BufferSize
+  )
+{
+  if ((Guid == NULL) || (Buffer == NULL)) {
+    return NULL;
+  }
+
+  AsciiSPrint (Buffer, BufferSize, "%g", Guid);
+  return Buffer;
+}
+
+/**
+  Build a cache of driver state information from the DXE core's discovered list.
+
+  @param Cache  Pointer to receive the allocated driver state array.
+  @param Count  Pointer to receive the number of drivers in the cache.
+
+  @retval EFI_SUCCESS            Cache successfully built.
+  @retval EFI_INVALID_PARAMETER  Cache or Count is NULL.
+  @retval EFI_OUT_OF_RESOURCES   Memory allocation failed.
+**/
+STATIC
+EFI_STATUS
+DiagBuildDriverStateCache (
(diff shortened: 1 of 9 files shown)


LinusAtDell:
@khaliid2040 @lgao4 @mikebeaton

Currently when a driver is not dispatched, the only output is:

  "Driver %g was discovered but not loaded!!"

There is no indication of *which* protocol is missing or *why* the Depex
failed. During bring-up this forces developers to manually parse Depex
bytecode or add temporary debug patches.

This PR adds `DepexDumpLib` to automatically print the missing protocol
and the Depex expression in human-readable form at the point of dispatch
failure, with a NULL instance for platforms that don't need it.

Please take a look. Thanks!

lgao4:
> @khaliid2040 @lgao4 @mikebeaton
> 
> Currently when a driver is not dispatched, the only output is:
> 
> "Driver %g was discovered but not loaded!!"
> 
> There is no indication of _which_ protocol is missing or _why_ the Depex failed. During bring-up this forces developers to manually parse Depex bytecode or add temporary debug patches.
> 
> This PR adds `DepexDumpLib` to automatically print the missing protocol and the Depex expression in human-readable form at the point of dispatch failure, with a NULL instance for platforms that don't need it.
> 
> Please take a look. Thanks!

I don't agree to add new library class for DxeCore module, because it needs to update every platform DSC to add this library instance. I suggest to directly add dump depex debug message logic in DxeCore.

LinusAtDell:
> > @khaliid2040 @lgao4 @mikebeaton
> > Currently when a driver is not dispatched, the only output is:
> > "Driver %g was discovered but not loaded!!"
> > There is no indication of _which_ protocol is missing or _why_ the Depex failed. During bring-up this forces developers to manually parse Depex bytecode or add temporary debug patches.
> > This PR adds `DepexDumpLib` to automatically print the missing protocol and the Depex expression in human-readable form at the point of dispatch failure, with a NULL instance for platforms that don't need it.
> > Please take a look. Thanks!
> 
> I don't agree to add new library class for DxeCore module, because it needs to update every platform DSC to add this library instance. I suggest to directly add dump depex debug message logic in DxeCore.

OK, noted. I will modify the patch to implement this in DxeCore and update it later.

LinusAtDell:
After applying this patch, 
Driver BE602BDD-5A74-406B-A9E4-BA8F9646F944 was discovered but not loaded!!
Driver BE602BDD-5A74-406B-A9E4-BA8F9646F944 not dispatched
[DEPEX] missing Protocol(BC8324C2-0C86-4B5B-97BB-76BC0A9234EE)

LinusAtDell:
@khaliid2040 @lgao4 @mikebeaton @jiaxinwu  @niruiyu 


Based on previous feedback, this version removes the DepexDumpLib approach. The functionality is implemented directly in both DxeCore and SmmCore as helper source files, ensuring consistent behavior across DXE and SMM without introducing a new library class or requiring DSC changes.

You can refer to the following log:

Driver BE602BDD-5A74-406B-A9E4-BA8F9646F944 was discovered but not loaded!!
Driver BE602BDD-5A74-406B-A9E4-BA8F9646F944 not dispatched
[DEPEX] missing Protocol(BC8324C2-0C86-4B5B-97BB-76BC0A9234EE)

Could you please take a look and share your feedback? Thanks!

MdeModulePkg/Core/Dxe/Dispatcher/DepexDebug.c
```diff
@@ -0,0 +1,851 @@
+/** @file
+  Depex debug helper for DXE Core.
+
```
khaliid2040, line 3: Please move this file to `Dispatcher` directory. It will keep things organized.

MdeModulePkg/Core/Dxe/DxeMain.h
```diff
@@ -2802,3 +2802,22 @@ EFI_STATUS
...
+  @param[in] DebugLevel   The debug level to print the summary at.
+
+  @retval EFI_SUCCESS             Summary printed successfully.
+  @retval EFI_INVALID_PARAMETER   DriverEntry is NULL.
+  @retval EFI_OUT_OF_RESOURCES    Memory allocation failed.
+**/
+EFI_STATUS
+CorePrintDriverDepexSummary (
+  IN EFI_CORE_DRIVER_ENTRY  *DriverEntry,
+  IN UINT32                 DebugLevel
```
khaliid2040: Please use boolean PCDs to dynamically disable and enable the feature. This makes it flexible without requiring to mask some log levels. And you don't need to pass DebugLevel.

MdeModulePkg/Core/PiSmmCore/PiSmmCore.h
```diff
@@ -1350,3 +1350,22 @@ VOID
...
+  @param[in] DebugLevel   The debug level to print the summary at.
+
+  @retval EFI_SUCCESS             Summary printed successfully.
+  @retval EFI_INVALID_PARAMETER   DriverEntry is NULL.
+  @retval EFI_OUT_OF_RESOURCES    Memory allocation failed.
+**/
+EFI_STATUS
+SmmPrintDriverDepexSummary (
+  IN EFI_SMM_DRIVER_ENTRY  *DriverEntry,
+  IN UINT32                DebugLevel
```
khaliid2040: Same, remove DebugLevel and use boolean PCD you add in `MdeModulePkg.dec` .

MdeModulePkg/Core/PiSmmCore/Dependency.c
```diff
@@ -349,10 +349,15 @@ SmmIsSchedulable (
         Status = PopBool (&Operator);
         if (EFI_ERROR (Status)) {
           DEBUG ((DEBUG_DISPATCH, "  RESULT = FALSE (Unexpected error)\n"));
+          SmmPrintDriverDepexSummary (DriverEntry, DEBUG_VERBOSE);
```
khaliid2040: This is not optimal because it requires you to call in every switch case statement. I would suggest you just like DXE version to move to main dispatcher routine and call SmmPrintDriverDepexSummary() if SmmIsSchedulable() returns FALSE.

khaliid2040:
Also, please split the DxeCore and PiSmmCore changes into two different commits.

MdeModulePkg/Core/Dxe/Dispatcher/DepexDebug.c
LinusAtDell: Thank you for the suggestion.
I moved DepexDebug.c under the Dispatcher/ directory to keep the structure organized and consistent

MdeModulePkg/Core/Dxe/DxeMain.h
LinusAtDell: Thank you, this makes sense.
I removed the DebugLevel parameter and switched to boolean PCD-based control so the feature can be enabled/disabled dynamically.

MdeModulePkg/Core/PiSmmCore/PiSmmCore.h
LinusAtDell: Thank you for pointing this out.
I applied the same approach in PiSmmCore: removed DebugLevel and used the boolean PCDs added in MdeModulePkg.dec.

MdeModulePkg/Core/PiSmmCore/Dependency.c
LinusAtDell: Great suggestion, thank you.
I updated the flow to match DXE: the summary call is now made from the main SMM dispatcher path when
SmmIsSchedulable() returns FALSE, instead of calling it in multiple switch/error paths.

LinusAtDell:
> Also, please split the DxeCore and PiSmmCore changes into two different commits.

Thank you for the reminder.
I split the changes into two separate commits (DXE and PiSmmCore) to keep review scope cleaner.

LinusAtDell:
@khaliid2040 @lgao4 @mikebeaton @jiaxinwu @niruiyu
Thank you again for all the detailed feedback.
I have now addressed all requested items in the latest update, including:
   • file placement cleanup
   • switching to boolean PCD-based control (and removing DebugLevel parameter plumbing)
   • moving the PiSmmCore summary call to the dispatcher flow
   • splitting DXE and PiSmmCore into separate commits
Could you please take another look when you have time?
I’d really appreciate your re-review

lgao4:
> @khaliid2040 @lgao4 @mikebeaton @jiaxinwu @niruiyu Thank you again for all the detailed feedback. I have now addressed all requested items in the latest update, including: • file placement cleanup • switching to boolean PCD-based control (and removing DebugLevel parameter plumbing) • moving the PiSmmCore summary call to the dispatcher flow • splitting DXE and PiSmmCore into separate commits Could you please take another look when you have time? I’d really appreciate your re-review

DxeCore  CoreIsSchedulable () function can print the depex evaluation result with DEBUG_DISPATCH print error level. If user wants to know the detail information, user can configure PCD PcdDebugPrintErrorLevel to enable DEBUG_DISPATCH level. So, this patch may not be necessary.

LinusAtDell:
> > @khaliid2040 @lgao4 @mikebeaton @jiaxinwu @niruiyu Thank you again for all the detailed feedback. I have now addressed all requested items in the latest update, including: • file placement cleanup • switching to boolean PCD-based control (and removing DebugLevel parameter plumbing) • moving the PiSmmCore summary call to the dispatcher flow • splitting DXE and PiSmmCore into separate commits Could you please take another look when you have time? I’d really appreciate your re-review
> 
> DxeCore CoreIsSchedulable () function can print the depex evaluation result with DEBUG_DISPATCH print error level. If user wants to know the detail information, user can configure PCD PcdDebugPrintErrorLevel to enable DEBUG_DISPATCH level. So, this patch may not be necessary.

Thank you @gao4 for pointing this out. After reviewing CoreIsSchedulable(), I confirmed that the existing DEBUG_DISPATCH output already includes "PUSH GUID(%g) = FALSE" for each missing protocol, which provides the same information. I'll close this PR and rely on the existing mechanism

PR #12609, opened 2026-05-25
MdeModulePkg/SdMmcPciHcDxe: Support override for SD 1.8v signaling switch

NumberOneGit:
Some platforms (e.g. Raspberry Pi) have SDHCI implementations where the internal signaling voltage control isn't wired up and therefore needs to be set externally, in a non-standard way.

Power cycling the card to get it out of 1.8V is also kind of necessary, but for now it can be done in the Reset/InitHost phase override.

# Description
This is an old patch Mario Bălănică <mariobalanica02@gmail.com> made in development of the Raspberry Pi 5 UEFI


NumberOneGit:
This commit was created a few years ago to allow use of storage media on Raspberry Pi 5.  Let me know what needs to be done to bring it to the state required for approval.

PR #12611, opened 2026-05-26
NetworkPkg/DpcDxe: Stop adding entries on allocation failure

MarsDoge:
# Description

DpcQueueDpc() expands the DPC entry free list when it is empty. If an allocation fails after at least one entry has already been added to the free list, the current code continues and inserts the failed allocation result into the list.

Stop expanding the free list when an allocation fails and the free list already contains entries. The caller can then continue with one of the entries that was successfully allocated.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested
PXE/HTTP/iSCSI

## Integration Instructions

N/A

### DIFF
--- a/NetworkPkg/DpcDxe/Dpc.c
+++ b/NetworkPkg/DpcDxe/Dpc.c
@@ -145,6 +145,8 @@ DpcQueueDpc (
           ReturnStatus = EFI_OUT_OF_RESOURCES;
           goto Done;
         }
+
+        break;
       }
 
       //


makubacki (approved):
> If an allocation fails after at least one entry has already been added to the free list, the current code continues and inserts the failed allocation result into the list.

Even worse than inserting the failed allocation result into the list, the current code potentially dereferences a null pointer in the process. LGTM

SaloniKasbekar (approved):

makubacki:
@SaloniKasbekar, since you've also approved, I assume you're okay with me adding the push label?

SaloniKasbekar:
Yes, ok with the push label.

PR #12612, opened 2026-05-26
MdeModulePkg/GptLib: Bound GPT entry array size to MAX_UINT32

jmestwa-coder:
Noticed the overflow guard in PartitionValidGptTable caps NumberOfPartitionEntries against MAX_UINTN, but the value it guards, NumberOfPartitionEntries * SizeOfPartitionEntry, is a UINT32 product that truncates. On a 64-bit build a header declaring 16 entries of 0x10000000 bytes clears the check while the product wraps to 0, so the entry array is allocated empty and the per-entry loop reads far past it. Bound against MAX_UINT32 to match the width of the multiplication.

### DIFF
--- a/MdeModulePkg/Library/GptLib/Gpt.c
+++ b/MdeModulePkg/Library/GptLib/Gpt.c
@@ -198,10 +198,13 @@ PartitionValidGptTable (
   //
   // Ensure PartitionEntryLBA * BlockSize and
   // NumberOfPartitionEntries * SizeOfPartitionEntry don't overflow when they
-  // are later used to read and size the partition entry array.
+  // are later used to read and size the partition entry array. Both entry
+  // fields are UINT32, so the product is evaluated in 32-bit arithmetic and
+  // must be bounded by MAX_UINT32 to avoid truncation before it is widened
+  // to UINTN.
   //
   if ((PartHdr->PartitionEntryLBA > DivU64x32 (MAX_UINT64, BlockSize)) ||
-      (PartHdr->NumberOfPartitionEntries > DivU64x32 (MAX_UINTN, PartHdr->SizeOfPartitionEntry)))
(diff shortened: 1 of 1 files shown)


jmestwa-coder:
gentle ping

szsam:
This patch is valid.

The type of `NumberOfPartitionEntries * SizeOfPartitionEntry` is `UINT32`, strictly smaller than `MAX_UINTN` on 64-bit systems.
https://github.com/tianocore/edk2/blob/a70c8729668f30de067f4b9db2e69baba283856e/MdeModulePkg/Universal/Disk/PartitionDxe/Gpt.c#L322


An alternative fix is to cast `NumberOfPartitionEntries` to `UINTN` before multiplication.
``` diff
- PartEntry = AllocatePool (PrimaryHeader->NumberOfPartitionEntries * PrimaryHeader->SizeOfPartitionEntry); 
+ PartEntry = AllocatePool ((UINTN)PrimaryHeader->NumberOfPartitionEntries * PrimaryHeader->SizeOfPartitionEntry); 

```

szsam (approved):

jmestwa-coder:
the cast fixes the allocation, but the same 32-bit product is used at seven other spots in this file:

- the ReadDisk size at Gpt.c:332, which has to stay in agreement with the alloc size at 322
- the entry array CRC path at 560/570/578
- PartitionRestoreGptTable at 653/664/675

bounding once in PartitionValidGptTable covers all of them, since every one of those paths only runs on a header that already passed validation. it also rejects the header outright instead of accepting one that asks for a ~4GB entry array, which is what the existing check was already trying to do.

jmestwa-coder:
rebased on master. the GPT parser was extracted into MdeModulePkg/Library/GptLib in 00a865d and the MAX_UINTN check moved with it, so the same one-line bound is now applied there. updated the title to match the new location.

jmestwa-coder:
any update?

PR #12616, opened 2026-05-26
NetworkPkg/DnsDxe: Bound query name length in ParseDnsResponse

jmestwa-coder:
ParseDnsResponse() runs AsciiStrLen over the query name taken straight from the received packet, before any length check. A response whose name field carries no terminating zero makes the scan walk past the packet buffer returned by NetbufGetByte. Bounding the scan to the remaining packet length keeps it inside the datagram, and the existing check then rejects a name that never terminates.

### DIFF
--- a/NetworkPkg/DnsDxe/DnsImpl.c
+++ b/NetworkPkg/DnsDxe/DnsImpl.c
@@ -1218,7 +1218,7 @@ ParseDnsResponse (
   //
   QueryName = (CHAR8 *)(RxString + sizeof (*DnsHeader));
 
-  QueryNameLen = (UINT32)AsciiStrLen (QueryName) + 1;
+  QueryNameLen = (UINT32)AsciiStrnLenS (QueryName, RemainingLength) + 1;
 
   //
   // Check whether the remaining packet length is available or not.


jmestwa-coder:
pushed a fix for the PatchCheck failure:
- added the missing Signed-off-by
- filled in the commit message body

makubacki (approved):

jmestwa-coder:
any update?

makubacki:
@jmestwa-coder, the latest commit formatting looks good to me. Please rebase locally and push (automatic rebase is prevented due to github workflow files being touched), then I'll add the push label back.

jmestwa-coder:
rebased onto latest master and pushed, commit itself is unchanged. should be good for the push label now.

mikebeaton (approved):

mikebeaton:
Apologies, I jumped too soon. Added comment and requested changes.

NetworkPkg/DnsDxe/DnsImpl.c
```diff
@@ -1218,7 +1218,7 @@ ParseDnsResponse (
   //
   QueryName = (CHAR8 *)(RxString + sizeof (*DnsHeader));
 
-  QueryNameLen = (UINT32)AsciiStrLen (QueryName) + 1;
+  QueryNameLen = (UINT32)AsciiStrnLenS (QueryName, RemainingLength) + 1;
```
mikebeaton: What happens if the length limit is hit? Don't we get a QueryNameLen which implies that the valid string (which should have a NUL) ends just outside the packet. So basically, if QueryNameLen == RemainingLength after this call, we know we've got an invalid packet, and should abort.

mikebeaton (changes requested):

NetworkPkg/DnsDxe/DnsImpl.c
jmestwa-coder: Right, when the limit is hit AsciiStrnLenS returns RemainingLength, so the old version relied on the following `RemainingLength <= QueryNameLen + sizeof (DNS_QUERY_SECTION)` check to abort. Made it explicit instead:
- compare the AsciiStrnLenS result against RemainingLength and return EFI_ABORTED before adding the terminator
- rebased onto current master, commit message updated to match

mikebeaton: @jmestwa-coder -

Sorry for wasting your time, you're right, additional fix is not wrong, but not required.

If you could revert to your previous version (sorry!), I can approve.

PR #12624, opened 2026-05-28
RedfishPkg/BmcUsbNicLib: Fix MAC address underflow check

Opadc:
## Description

HostInterfaceIpmiCheckMacAddress compares the host-end USB NIC MAC
address against the BMC-end MAC address minus one in the last octet.

When the BMC-end last octet is 0x00, C integer promotion evaluates the
subtraction as -1 instead of the expected UINT8 wraparound value 0xff.
This prevents the documented BMC ...:00 to host ...:ff case from
matching correctly.

Cast the subtraction result back to UINT8 before comparing it with the
host-end MAC address octet.

## Verification

- `python3 BaseTools/Scripts/PatchCheck.py -1`

### DIFF
--- a/RedfishPkg/Library/PlatformHostInterfaceBmcUsbNicLib/PlatformHostInterfaceBmcUsbNicLib.c
+++ b/RedfishPkg/Library/PlatformHostInterfaceBmcUsbNicLib/PlatformHostInterfaceBmcUsbNicLib.c
@@ -751,7 +751,7 @@ HostInterfaceIpmiCheckMacAddress (
              (VOID *)&IpmiLanChannelMacAddress.Addr,
              IpmiLanMacAddressSize - 1
              ) != 0) ||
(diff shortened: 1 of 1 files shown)


nicklela (changes requested):
The change looks good to me but please address the patch check failure:

```
 * Missing Signed-off-by! (Note: this must be added by the code contributor!)
https://www.tianocore.org/tianocore-wiki.github.io/development/contribution-guides/commit_message_format.html
```

Opadc:
Thanks Nickle. The PatchCheck failure was caused by an unsigned merge commit. I removed it and force-pushed the PR branch.

The PR now contains a single signed-off commit:
fac39f93ed RedfishPkg/BmcUsbNicLib: Fix MAC address underflow check

I also reran PatchCheck locally and it passed:
python3 BaseTools/Scripts/PatchCheck.py -1

nicklela (approved):

nicklela:
@changab  please check this when you have chance, thanks

abdattar (approved):

PR #12628, opened 2026-05-28
BaseTools/Build: Fix arch macro expansion scope in DSC parser

kirkchoutw:
In the EDK2 DSC parser, macro expansion for the arch field in `[Components.$(MACRO)]` sections was only performed on the Component record itself inside `__ProcessComponent()`. Sub-items under that component (e.g. `<LibraryClasses>`, `<BuildOptions>`) retained the unexpanded `$(MACRO)` string in their scope, causing arch matching to fail. As a result, per-module library class overrides and build options under macro-based arch sections were silently ignored during the build.

**Changes:**
- Moved the `ReplaceMacro()` call for `self._Scope[0][0]` from `__ProcessComponent()` to the main processing loop in `__ProcessAll()`, immediately after `self._Scope` is assigned. This ensures macro expansion of the arch field applies to all record types (Component entries, LibraryClasses sub-items, BuildOptions sub-items, etc.), not just Component records.
- Removed the now-redundant `ReplaceMacro()` call from `__ProcessComponent()`.
- Added a warning when a macro in the arch field cannot be resolved, since the EDK2 DSC spec does not allow unresolved macros in final section headers (e.g. `[Components.]` or `[Components.$(UndefString)]`).

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verified with a DSC file using `[Components.$(PEI_ARCH)]` sections containing per-module `<LibraryClasses>` and `<BuildOptions>` sub-items. Before the fix, the sub-items were silently ignored (arch field remained as the unexpanded macro string). After the fix, all sub-items correctly resolve and apply to the expected architecture target.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Source/Python/Workspace/MetaFileParser.py
+++ b/BaseTools/Source/Python/Workspace/MetaFileParser.py
@@ -1400,6 +1400,17 @@ class DscParser(MetaFileParser):
 
             self._Scope = [[S1, S2, S3]]
             #
+            # Expand macros in arch field for all records so that per-module
+            # sub-items (e.g. <LibraryClasses> under [Components.$(PEI_ARCH)])
+            # resolve correctly, not just the Component record itself.
+            #
+            self._Scope[0][0] = ReplaceMacro(self._Scope[0][0], self._Macros)
+            if '$(' in self._Scope[0][0] and S1 != TAB_ARCH_COMMON:
+                EdkLogger.warn("Parser",
+                               "Macro in arch field was not resolved. "
+                               "'%s' used in section header is not defined." % S1,
+                               File=self._FileWithError, Line=LineStart)
+            #
             # For !include directive, handle it specially,
             # merge arch and module type in case of duplicate items
             #
@@ -1712,7 +1723,6 @@ class DscParser(MetaFileParser):
 
     def __ProcessComponent(self):
         self._ValueList[0] = ReplaceMacro(self._ValueList[0], self._Macros)
-        self._Scope[0][0] = ReplaceMacro(self._Scope[0][0], self._Macros)
 
     def __ProcessBuildOption(self):
         self._ValueList = [ReplaceMacro(Value, self._Macros, RaiseError=False)


kirkchoutw:
Friendly ping @YuweiChen1110 @gapalomi @lgao4 @mikebeaton ,whenever you have time, I'd appreciate a review. Thanks!

lgao4:
could you give the test case? I may want to see your case in DSC.

kirkchoutw:
Hi @lgao4 ,
Below is test case ,step and pass criteria.
[TestArchMacroPkg.zip](https://github.com/user-attachments/files/29059728/TestArchMacroPkg.zip)

Please validate the behavior using the following steps on a clean edk2 tree:

1.Copy the test package directory into the workspace root:
TestArchMacroPkg/
2.Set up the build environment:
3.Build the package:
build -p TestArchMacroPkg/TestArchMacroPkg.dsc -a IA32 -t GCC -b DEBUG
Verify the generated artifacts:

Check the module GNUmakefile:

Build/TestArchMacroPkg/DEBUG_GCC/IA32/TestArchMacroPkg/Module/TestPeiDriver/TestPeiDriver/GNUmakefile

Pass criteria:

TestBaseMemoryLib appears in the link dependencies for TestPeiDriver.
-DPEI_CUSTOM_FLAG appears in the CC_FLAGS for TestPeiDriver.

Interpretation:

If both are present, architecture macro expansion within sub-items is working correctly (fixed behavior).
If they are missing and the default BaseMemoryLib is used instead, sub-item scope expansion is not being applied (bug behavior).


I would recommend preserving the build folders generated before and after applying this PR, and then using Beyond Compare to compare them.

This can help clearly identify any differences in the generated build artifacts, including GNUmakefiles, library dependency resolution, build options, and other outputs affected by the change.
image

kirkchoutw:
Hi @lgao4 ,

The requested test cases have been added to the comment. Could you please review it when you have time?

Thanks!

lgao4 (approved):

PR #12622, opened 2026-05-28
MdePkg: Introduce DT_FIXUP protocol

dimorinny:
# Description

Previous related discussion: https://github.com/tianocore/edk2/pull/11667

Adds the DT_FIXUP_PROTOCOL header. This protocol allows firmware to apply boot-time fixups to the device tree. It is expected to be invoked by the boot manager before passing the updated device tree to the OS.

This protocol is [used](https://github.com/U-Boot-EFI/EFI_DT_FIXUP_PROTOCOL/blob/master/references.rst) by:

1. Android Generic Bootloader (GBL)
2. Systemd-boot, Systemd-stub
3. Ubuntu's grub2

Original proposal:
https://github.com/U-Boot-EFI/EFI_DT_FIXUP_PROTOCOL

- [ ] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [ ] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

## How This Was Tested

Used by GBL on top of QCOM EDK2-based FW.

## Integration Instructions

N/A

PR #12630, opened 2026-05-29
.pytool,.azurepipelines: Add Cspell Ignore Regex List

os-d:
# Description

This adds two regexes to ignore in cspell: copyright notices (which may have companies not in the companies dictionary or individuals listed) and hex values (which may have things like 0xABEABCAB...).

While we are here, edk2 is using an old version of cspell that doesn't do as good of a job of ignoring some things it should. Update to the latest. This only brought one additional failure, which is added to the global ignore list as it comes from the gcc/clang -fstack-protector flag.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Ran the spell check plugin against the failing copyright notice from https://github.com/tianocore/edk2/pull/12622 and it now passes. Ran the spell check plugin against the whole repo with the new version.

## Integration Instructions

N/A.

### DIFF
--- a/.azurepipelines/templates/spell-check-prereq-steps.yml
+++ b/.azurepipelines/templates/spell-check-prereq-steps.yml
@@ -17,6 +17,6 @@ steps:
     #checkLatest: false # Optional
   condition: and(gt(variables.pkg_count, 0), succeeded())
 
-- script: npm install -g cspell@5.20.0
+- script: npm install -g cspell@10.0.0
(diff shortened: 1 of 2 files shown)


Javagedes (approved):

makubacki (approved):

leiflindholm (approved):
Thank you!

PR #12632, opened 2026-05-30
NetworkPkg/Dhcp6Dxe: bound IA inner option length to buffer

jmestwa-coder:
Dhcp6SeekInnerOptionSafe trusts the IA option's declared length field but never checks it against OptionLen, so a reply claiming more inner data than the datagram holds makes Dhcp6SeekOption walk past the packet buffer. Bound the declared inner length against the remaining buffer for both IA_NA and IA_TA. Adds host tests for the over-declared case.

### DIFF
--- a/NetworkPkg/Dhcp6Dxe/Dhcp6Io.c
+++ b/NetworkPkg/Dhcp6Dxe/Dhcp6Io.c
@@ -750,6 +750,13 @@ Dhcp6SeekInnerOptionSafe (
     }
 
     IaInnerLenTmp -= DHCP6_SIZE_OF_COMBINED_IAID_T1_T2;
+
+    //
+    // Verify the declared inner length does not run past the option buffer.
+    //
+    if (IaInnerLenTmp > OptionLen - DHCP6_MIN_SIZE_OF_IA_NA) {
+      return EFI_DEVICE_ERROR;
+    }
   } else if (IaType == Dhcp6OptIata) {
     //
     // Verify the OptionLen is valid.
@@ -769,6 +776,13 @@ Dhcp6SeekInnerOptionSafe (
     }
 
     IaInnerLenTmp -= DHCP6_SIZE_OF_IAID;
+
+    //
+    // Verify the declared inner length does not run past the option buffer.
+    //
+    if (IaInnerLenTmp > OptionLen - DHCP6_MIN_SIZE_OF_IA_TA) {
+      return EFI_DEVICE_ERROR;
+    }
   } else {
     return EFI_DEVICE_ERROR;
   }
--- a/NetworkPkg/Dhcp6Dxe/GoogleTest/Dhcp6IoGoogleTest.cpp
+++ b/NetworkPkg/Dhcp6Dxe/GoogleTest/Dhcp6IoGoogleTest.cpp
@@ -699,6 +699,182 @@ TEST_F (Dhcp6SeekInnerOptionSafeTest, InvalidOption) {
   ASSERT_EQ (Result, EFI_DEVICE_ERROR);
 }
 
+// Test Description:
+// This test verifies that Dhcp6SeekInnerOptionSafe rejects an IANA option whose
+// declared length runs past the end of the option buffer.
+TEST_F (Dhcp6SeekInnerOptionSafeTest, IANAInnerLenExceedsBufferExpectFail) {
+  EFI_STATUS           Status;
+  UINT8                Option[sizeof (DHCPv6_OPTION_IA_NA) + SEARCH_PATTERN_LEN] = { 0 };
+  UINT32               OptionLength                                              = sizeof (Option);
+  DHCPv6_OPTION_IA_NA  *OptionPtr                                                = (DHCPv6_OPTION_IA_NA *)Option;
+
+  UINT8   *InnerOptionPtr   = NULL;
+  UINT16  InnerOptionLength = 0;
+
+  OptionPtr->Header.Code = Dhcp6OptIana;
+  OptionPtr->Header.Len  = HTONS (0xFFFF); // Declare far more inner data than the buffer holds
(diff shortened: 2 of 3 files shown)


makubacki:
@jmestwa-coder, commits in this repo are merged as-is and the commit in the PR is currently missing a commit description and a signed-off-by line (checked by `tianocore.PatchCheck`). Please add those.

jmestwa-coder:
Done:
- added a commit description explaining the over-declared inner length and the out-of-bounds read it causes
- added my signed-off-by

PatchCheck should be happy now.

NetworkPkg/Dhcp6Dxe/Dhcp6Io.c
```diff
@@ -750,6 +750,13 @@ Dhcp6SeekInnerOptionSafe (
     }
 
     IaInnerLenTmp -= DHCP6_SIZE_OF_COMBINED_IAID_T1_T2;
+
+    //
+    // Verify the declared inner length does not run past the option buffer.
+    //
+    if (IaInnerLenTmp > OptionLen - DHCP6_MIN_SIZE_OF_IA_NA) {
```
makubacki, line 757: Can you please add parentheses around `DHCP6_MIN_SIZE_OF_IA_NA` in Dchp6Impl.h? `DHCP6_MIN_SIZE_OF_IA_TA` already has them.

https://github.com/tianocore/edk2/blob/95d38da6c63b151a2cf75f51fa01625749c155d0/NetworkPkg/Dhcp6Dxe/Dhcp6Impl.h#L150-L152

NetworkPkg/Dhcp6Dxe/GoogleTest/Dhcp6IoGoogleTest.cpp
```diff
@@ -699,6 +699,60 @@ TEST_F (Dhcp6SeekInnerOptionSafeTest, InvalidOption) {
...
+  EFI_STATUS           Status;
+  UINT8                Option[sizeof (DHCPv6_OPTION_IA_NA) + SEARCH_PATTERN_LEN] = { 0 };
+  UINT32               OptionLength                                              = sizeof (Option);
+  DHCPv6_OPTION_IA_NA  *OptionPtr                                                = (DHCPv6_OPTION_IA_NA *)Option;
+
+  UINT8   *InnerOptionPtr   = NULL;
+  UINT16  InnerOptionLength = 0;
+
+  OptionPtr->Header.Code = Dhcp6OptIana;
+  OptionPtr->Header.Len  = HTONS (0xFFFF); // Declare far more inner data than the buffer holds
```
makubacki, line 715: Thanks for adding the tests. Could you try including a test with a smaller incorrect size (like an inner length that is off by one)?

makubacki: A test that verifies an inner length that exactly matches `OptionLen - DHCP6_MIN_SIZE_OF_IA_NA` is successful would also be nice for coverage.

makubacki (approved):

NetworkPkg/Dhcp6Dxe/Dhcp6Io.c
jmestwa-coder: good catch. added the parens in Dhcp6Impl.h to match DHCP6_MIN_SIZE_OF_IA_TA. without them the macro expanded as OptionLen - 4 + 12, so the IA_NA bound was effectively OptionLen + 8 and too loose. it now binds as OptionLen - 16. folded into the same commit.

NetworkPkg/Dhcp6Dxe/GoogleTest/Dhcp6IoGoogleTest.cpp
jmestwa-coder: added IANAInnerLenOffByOneExpectFail (plus the IATA variant), declaring an inner length one byte past OptionLen - DHCP6_MIN_SIZE_OF_IA_*. both reject with EFI_DEVICE_ERROR.

jmestwa-coder: also added IANAInnerLenExactMatchExpectSuccess and the IATA variant, where the inner length exactly equals OptionLen - DHCP6_MIN_SIZE_OF_IA_*. both pass and return the expected inner length.

jmestwa-coder:
Pushed an update:
- parenthesized DHCP6_MIN_SIZE_OF_IA_NA in Dhcp6Impl.h so the IA_NA bound binds as OptionLen - 16
- added off-by-one over-declared tests for IA_NA and IA_TA (reject)
- added exact-boundary tests for IA_NA and IA_TA (accept, return expected inner length)

All folded into the single commit.

makubacki (approved):

makubacki:
In case you didn't see it, there are a couple warnings in the tests that need to be accounted for:

```
WARNING - Compiler #4267 from D:\a\1\s\NetworkPkg\Dhcp6Dxe\GoogleTest\Dhcp6IoGoogleTest.cpp(799): 'argument': conversion from 'size_t' to 'UINT16', possible loss of data
WARNING - Compiler #4267 from D:\a\1\s\NetworkPkg\Dhcp6Dxe\GoogleTest\Dhcp6IoGoogleTest.cpp(862): 'argument': conversion from 'size_t' to 'UINT16', possible loss of data
```

jmestwa-coder:
Thanks, missed those. Both come from DHCP6_SIZE_OF_IAID being a sizeof, so the IA_TA length expressions widen to size_t before HTONS. Cast the two HTONS arguments to UINT16 (lines 799 and 862). Folded into the same commit and pushed.

makubacki:
If NetworkPkg reviewers would like to review this, please do so soon. Otherwise, I'll merge it within a couple of days.

PR #12633, opened 2026-05-30
Fix/qemukernelloader fwcfg size overflow

AristoChen:
# Description

 QemuKernelFetchBlob() in OvmfPkg/QemuKernelLoaderFsDxe reads the kernel blob's component sizes (kernel setup size + kernel image size) from QEMU fw_cfg, which is host-controlled, and sums them into a UINT32 with no overflow check. The blob is then served from a single UINT32-sized AllocatePool(), while the copy loop streams each component's full UINT32 size into that buffer. A host that reports, e.g., setup size 0xFFFFFF00 and image size 0x200 makes the UINT32 sum wrap to 0x100, so a 256-byte allocation is subsequently written with ~4 GiB of attacker-controlled bytes: a controllable heap buffer overflow. The fetch runs before VerifyBlob(), so the SEV/TDX measured-boot hash gate does not cover it; on confidential-compute platforms (host outside the guest TCB) this is a guest integrity break.

  This PR fixes the overflow by accumulating the component sizes in a UINT64 and rejecting any total that does not fit in UINT32 (EFI_INVALID_PARAMETER) before allocating, so the allocation can no longer be undersized by a wrapping sum.

  It also adds a host-based unit test that drives QemuKernelFetchBlob() through a programmable fw_cfg stub. OvmfPkg had no host-based unit test platform, so the PR introduces OvmfPkg/Test/OvmfPkgHostTest.dsc and wires it into CI via the HostUnitTestCompilerPlugin and HostUnitTestDscCompleteCheck plugins.

- [ ] Breaking change?
- [x] Impacts security?
- [x] Includes tests?

## How This Was Tested

Built and ran the new host-based unit test on Linux/GCC X64 with the UnitTestFrameworkPkg host DSC (AddressSanitizer enabled by default):

     build -p OvmfPkg/Test/OvmfPkgHostTest.dsc -a X64 -t GCC -b NOOPT
     Build/OvmfPkg/HostTest/NOOPT_GCC/X64/QemuKernelLoaderFsDxeUnitTest

Two cases run: a well-formed kernel blob is fetched correctly, and a size sum that overflows UINT32 is rejected. With the fix applied both cases pass. To confirm the test is meaningful, the fix was reverted while keeping the test: the overflow case then reproduces the heap buffer overflow under AddressSanitizer (heap-buffer-overflow WRITE in InternalMemSetMem, allocation undersized to 256 bytes), while the well-formed case still passes.
## Integration Instructions

N/A

### DIFF
--- a/OvmfPkg/QemuKernelLoaderFsDxe/UnitTest/QemuKernelLoaderFsDxeUnitTest.c
+++ b/OvmfPkg/QemuKernelLoaderFsDxe/UnitTest/QemuKernelLoaderFsDxeUnitTest.c
@@ -1,0 +1,294 @@
+/** @file
+  Host-based unit test for QemuKernelLoaderFsDxe blob fetching.
+
+  Exercises QemuKernelFetchBlob() against a programmable fw_cfg stub to
+  reproduce the integer-overflow heap buffer overflow tracked as
+  review-notes D1 (CWE-190 -> CWE-787): the per-blob sizes returned by the
+  (host-controlled) fw_cfg interface are summed into a UINT32 with no overflow
+  check, so a wrapping sum makes the allocation far smaller than the number of
+  bytes the copy loop subsequently writes.
+
+  The test #includes the module under test as a single translation unit so the
+  STATIC QemuKernelFetchBlob() and its STATIC fw_cfg item table are reachable.
+  The fw_cfg, HOB, device-path and blob-verifier dependencies have no
+  host-library instance, so minimal stubs are provided locally; everything else
+  links against the standard host libraries.
+
+  Copyright (c) 2026, Canonical Ltd. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#include <stdio.h>
+#include <string.h>
+#include <stdarg.h>
+#include <stddef.h>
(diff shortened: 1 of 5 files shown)


OvmfPkg/QemuKernelLoaderFsDxe/UnitTest/QemuKernelLoaderFsDxeUnitTest.c
```diff
@@ -0,0 +1,294 @@
...
+  IN  EFI_STATUS    FetchStatus
+  )
+{
+  return EFI_SUCCESS;
+}
+
+//
+// Pull in the module under test (STATIC QemuKernelFetchBlob + mKernelBlobItems).
+//
+#include "../QemuKernelLoaderFsDxe.c"
```
EricGao2015, line 161: I'am not sure if including C file is permitted in edk2.

kraxel: This is a test case, I think it is fine there.

kraxel:
edk2 has `MdePkg/Include/Library/SafeIntLib.h` for that, with functions like `SafeUint32Add`.

PR #12638, opened 2026-06-01
BaseTools/Source/Python: Add -noattr to capsule signing openssl commands

mkopec:
# Description

This makes capsule PKCS#7 signatures reproducible.

openssl smime -sign embeds a signingTime signed attribute (current UTCTime) by default. Because the attribute is inside the signed digest, the resulting RSA signature differs on every invocation even when the key and payload are identical, breaking reproducible builds of signed FMP capsules.

Pass -noattr to drop all signed attributes from the SignerInfo. The signature is then computed directly over the content, producing byte-identical output across runs.

The FMP update path is unaffected: FmpAuthenticationLibPkcs7 calls Pkcs7Verify() without extracting signed attributes, and rollback protection uses FwVersion / LowestSupportedVersion in FMP_PAYLOAD_HEADER plus the auth header MonotonicCount, not signingTime. signingTime is only consumed by DxeImageVerificationLib for Secure Boot dbt checks on PE/COFF images, which is a separate code path.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

A coreboot image with edk2 as payload was built, and then a capsule was generated multiple times with multiple-second pauses between tool calls. SHA256 sums of the generated capsules were compared to confirm all capsules were identical.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Source/Python/Pkcs7Sign/Pkcs7Sign.py
+++ b/BaseTools/Source/Python/Pkcs7Sign/Pkcs7Sign.py
@@ -201,7 +201,7 @@ if __name__ == '__main__':
     #
     # Sign the input file using the specified private key and capture signature from STDOUT
     #
-    Process = subprocess.Popen('%s smime -sign -binary -signer "%s" -outform DER -md sha256 -certfile "%s"' % (OpenSslCommand, args.SignerPrivateCertFileName, args.OtherPublicCertFileName), stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
+    Process = subprocess.Popen('%s smime -sign -binary -noattr -signer "%s" -outform DER -md sha256 -certfile "%s"' % (OpenSslCommand, args.SignerPrivateCertFileName, args.OtherPublicCertFileName), stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
     Signature = Process.communicate(input=FullInputFileBuffer)[0]
     if Process.returncode != 0:
       sys.exit(Process.returncode)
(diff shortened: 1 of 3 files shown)


BaseTools/Source/Python/Pkcs7Sign/Pkcs7Sign.py
```diff
@@ -201,7 +201,7 @@
     #
     # Sign the input file using the specified private key and capture signature from STDOUT
     #
-    Process = subprocess.Popen('%s smime -sign -binary -signer "%s" -outform DER -md sha256 -certfile "%s"' % (OpenSslCommand, args.SignerPrivateCertFileName, args.OtherPublicCertFileName), stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
+    Process = subprocess.Popen('%s smime -sign -binary -noattr -signer "%s" -outform DER -md sha256 -certfile "%s"' % (OpenSslCommand, args.SignerPrivateCertFileName, args.OtherPublicCertFileName), stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
```
SergiiDmytruk, line 204: Maybe `-noattr` should be added also to the `Readme.md` in this directory which seems to match this call:
https://github.com/tianocore/edk2/blob/808ee1b53d94feaf80e832b5e817fa10ce2f28be/BaseTools/Source/Python/Pkcs7Sign/Readme.md?plain=1#L147

mkopec: Good catch, done in https://github.com/tianocore/edk2/pull/12638/commits/0dc6b11ce238c8b3286e3f04ffb0c787523f2bfd

SergiiDmytruk (approved):
Can't resolve the thread I've started, yet can approve.  GitHub is weird.

filipleple (approved):

SergiiDmytruk:
> This PR has been automatically marked as stale because it has not had activity in 60 days. It will be closed if no further activity occurs within 7 days. Thank you for your contributions.

Keep it open.

PR #12641, opened 2026-06-02
NetworkPkg/IScsiDxe: bound value length in IScsiBuildKeyValueList

jmestwa-coder:
Noticed IScsiBuildKeyValueList runs AsciiStrLen on the value pointer before checking it against the remaining data-segment bytes. A login or CHAP response whose last key=value isn't NUL-terminated makes the scan read past the received PDU buffer, since the segment copied off the wire (AllocatePool(Len) + NetbufQueCopy) carries no terminator. Bound it with AsciiStrnLenS(Value, Len) so the scan stays inside the segment and the existing SafeUint32Sub rejects an unterminated value.

### DIFF
--- a/NetworkPkg/IScsiDxe/IScsiProto.c
+++ b/NetworkPkg/IScsiDxe/IScsiProto.c
@@ -1922,7 +1922,7 @@ IScsiBuildKeyValueList (
 
     KeyValuePair->Value = Data;
 
-    Status = SafeUint32Add ((UINT32)AsciiStrLen (KeyValuePair->Value), 1, &Result);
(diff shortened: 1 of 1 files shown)


jmestwa-coder:
Added the missing Signed-off-by and pushed, PatchCheck should be green now. The CLANGPDB failure was a Windows agent git-fetch crash, unrelated to the change.

makubacki (approved):

makubacki:
LGTM. Will merge by Wednesday if no other concerns. Attn NetworkPkg reviewers: @mikebeaton, @SaloniKasbekar, @Zclarkwilliams

SaloniKasbekar (approved):

PR #12642, opened 2026-06-03
NetworkPkg/Ip6Dxe: Enforce RFC 4861 8-octet minimum in ND redirect parse

MarkWuDell:
NetworkPkg/Ip6Dxe: Enforce RFC 4861 8-octet minimum in ND redirect parse

RFC 4861 Section 4.6 defines every Neighbor Discovery option
length in units of 8 octets. Therefore, valid ND option
regions must be consumed as an exact multiple of 8 bytes, and
trailing 1-7 byte fragments are malformed.

This change hardens ND option parsing in two place:
01. Ip6ProcessRedirect() in Ip6Nd.c
- Change loop guard from:
   while (Length > 0)
  to
   while (Length >= 8)
02. Ip6IsNDOptionValid() in Ip6Option.c
- Change loop guard from:
   while (Offset + sizeof (IP6_OPTION_HEADER) -1 < OptionLen)
  to
   while (Offset + 8 <= (UINT32)OptionLen)
- Change final return from:
   return TRUE;
  to
   return (Offset == (UINT32)OptionLen);

Together these updates reject malformed trailing fragments
and preserve behavior for well-formed ND option regions.

Signed-off-by: Mark Wu <mark.wu1@dell.com>

### DIFF
--- a/NetworkPkg/Ip6Dxe/Ip6Nd.c
+++ b/NetworkPkg/Ip6Dxe/Ip6Nd.c
@@ -2511,10 +2511,15 @@ Ip6ProcessRedirect (
   // Check the options. The only interested option here is the target-link layer
   // address option.
   //
+  // Per RFC 4861 Section 4.6, every ND option length is in units of 8 octets,
+  // so the smallest valid option is 8 bytes (Length field = 1). Guard with
+  // ">= 8" instead of "> 0" to avoid parsing malformed trailing fragments
+  // that cannot form a complete option.
+  //
   Length          = Packet->TotalSize - 40;
   Option          = (UINT8 *)(IcmpDest + 1);
   LinkLayerOption = NULL;
-  while (Length > 0) {
+  while (Length >= 8) {
     switch (*Option) {
       case Ip6OptionEtherTarget:
 
--- a/NetworkPkg/Ip6Dxe/Ip6Option.c
+++ b/NetworkPkg/Ip6Dxe/Ip6Option.c
@@ -173,10 +173,12 @@ Ip6IsNDOptionValid (
 
   //
   // RFC 4861 states that Neighbor Discovery packet can contain zero or more
-  // options. Start processing the options if at least Type + Length fields
-  // fit within the input buffer.
+  // options. RFC 4861 Section 4.6 states that every ND option length is in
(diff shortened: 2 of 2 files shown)


makubacki (approved):

NetworkPkg/Ip6Dxe/Ip6Nd.c
```diff
@@ -2510,10 +2511,15 @@ Ip6ProcessRedirect (
...
+  // Per RFC 4861 Section 4.6, every ND option length is in units of 8 octets,
+  // so the smallest valid option is 8 bytes (Length field = 1). Guard with
+  // ">= 8" instead of "> 0" to avoid parsing malformed trailing fragments
+  // that cannot form a complete option.
+  //
   Length          = Packet->TotalSize - 40;
   Option          = (UINT8 *)(IcmpDest + 1);
   LinkLayerOption = NULL;
-  while (Length > 0) {
+  while (Length >= 8) {
```
SaloniKasbekar, line 2522: Your change looks good. However, Ip6IsNDOptionValid() also currently accepts a buffer with a trailing 1-byte fragment after a valid option sequence - which goes against the RFC. 

while (Offset + sizeof (IP6_OPTION_HEADER) - 1 < OptionLen)

So if parsing stops with leftover bytes smaller than 2, the function exits the loop and returns TRUE, even though the option area was not fully consumed.

Could you help fix that failure as well in this PR?

MarkWuDell: Hello @SaloniKasbekar,

Thanks for your review.
I have updated Ip6Option.c accordingly.

makubacki: @SaloniKasbekar, if you can please resolve the conversation, I'm happy to add the push label.

MarkWuDell: Hello @SaloniKasbekar,

When you have time, could you please review the latest updates and help resolve the conversations if everything looks good?
Thanks for your time!

mdkinney: @Zclarkwilliams Can you please review this conversation and the final changes and resolve if you agree with latest version?

PR #12643, opened 2026-06-03
MdeModulePkg/UdfDxe: bound file identifier descriptor to directory data

jmestwa-coder:
ReadDirectoryEntry only bounds each File Identifier Descriptor's start offset, so oversized identifier lengths on a crafted UDF volume make GetFidDescriptorLength and the DuplicateFid copy read past the directory buffer.


jmestwa-coder:
added the missing sign-off, that was the PatchCheck failure. the CLANGPDB job died in git fetch on the windows agent, not a build error, so the re-push should clear it.

PR #12647, opened 2026-06-05
MdeModulePkg/UsbBusPei: Use dynamic buffer for USB configuration data

DC-Damien:
USB devices whose configuration descriptor TotalLength exceeds 1024 bytes (e.g. IR cameras with large descriptor tables) previously hit an EFI_DEVICE_ERROR hard-limit and failed to enumerate in PEI.

Replace the fixed array with a UINT8 * pointer and dynamically allocate the exact amount of memory required via
PeiServicesAllocatePool() after the TotalLength is learned from the initial 4-byte descriptor probe. The minimum-size sanity check (TotalLength must cover its own header) is preserved; the 1024-byte upper-bound check is removed.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Test in real hardware.

## Integration Instructions

N/A.

### DIFF
--- a/MdeModulePkg/Bus/Usb/UsbBusPei/UsbPeim.c
+++ b/MdeModulePkg/Bus/Usb/UsbBusPei/UsbPeim.c
@@ -3,6 +3,7 @@ The module to produce Usb Bus PPI.
 
 Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.<BR>
 Copyright (c) 2006 - 2018, Intel Corporation. All rights reserved.<BR>
+Copyright (c) 2026 Dell Inc. All rights reserved.<BR>
 
 SPDX-License-Identifier: BSD-2-Clause-Patent
 
@@ -722,7 +723,7 @@ PeiUsbGetAllConfiguration (
   )
 {
   EFI_STATUS                 Status;
-  EFI_USB_CONFIG_DESCRIPTOR  *ConfigDesc;
+  EFI_USB_CONFIG_DESCRIPTOR  ConfigDescHeader;
   PEI_USB_IO_PPI             *UsbIoPpi;
   UINT16                     ConfigDescLength;
   UINT8                      *Ptr;
@@ -735,7 +736,7 @@ PeiUsbGetAllConfiguration (
   UsbIoPpi = &PeiUsbDevice->UsbIoPpi;
 
   //
-  // First get its 4-byte configuration descriptor
+  // First get its 4-byte configuration descriptor to learn TotalLength.
   //
   Status = PeiUsbGetDescriptor (
              PeiServices,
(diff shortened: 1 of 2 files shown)


DC-Damien:
cc @lgao4 .

MdeModulePkg/Bus/Usb/UsbBusPei/UsbPeim.c
```diff
@@ -737,6 +738,11 @@ PeiUsbGetAllConfiguration (
   //
   // First get its 4-byte configuration descriptor
   //
+  Status = PeiServicesAllocatePool (4, (VOID **)&PeiUsbDevice->ConfigurationData);
```
lgao4: The fixed length is used here. So, I suggest to define one local variable for this usage instead of allocate pool.

DC-Damien: So we need to use one local variable, and the code base would be something like this so that we don`t have to allocate memory in PEI phase, is that right ? 
```
   EFI_STATUS                 Status;
+  EFI_USB_CONFIG_DESCRIPTOR  ConfigDescHeader;^M
   EFI_USB_CONFIG_DESCRIPTOR  *ConfigDesc;
   PEI_USB_IO_PPI             *UsbIoPpi;
   UINT16                     ConfigDescLength;
@@ -736,20 +737,15 @@ PeiUsbGetAllConfiguration (
   UsbIoPpi = &PeiUsbDevice->UsbIoPpi;

   //
-  // First get its 4-byte configuration descriptor
+  // First get its 4-byte configuration descriptor to learn TotalLength.^M
   //
-  Status = PeiServicesAllocatePool (4, (VOID **)&PeiUsbDevice->ConfigurationData);
-  if (EFI_ERROR (Status)) {
-    return Status;
-  }
-
   Status = PeiUsbGetDescriptor (
              PeiServices,
              UsbIoPpi,
              (USB_DT_CONFIG << 8), // Value
              0,                    // Index
              4,                    // Length
-             PeiUsbDevice->ConfigurationData
+             (UINT8 *)&ConfigDescHeader^M
              );

   if (EFI_ERROR (Status)) {
@@ -759,21 +755,20 @@ PeiUsbGetAllConfiguration (

   MicroSecondDelay (USB_GET_CONFIG_DESCRIPTOR_STALL);

-  ConfigDesc       = (EFI_USB_CONFIG_DESCRIPTOR *)PeiUsbDevice->ConfigurationData;
-  ConfigDescLength = ConfigDesc->TotalLength;
-
-  Status = PeiServicesAllocatePool (ConfigDescLength, (VOID **)&PeiUsbDevice->ConfigurationData);
-  if (EFI_ERROR (Status)) {
-    return Status;
```

DC-Damien:
Hi @lgao4, Please let me know if you have any feedback about this. Thanks!.

DC-Damien:
Hi @lgao4, May I check with you if you have any feedback on this ? Thanks.

lgao4 (approved):

lgao4:
Sorry for the late response.

PR #12654, opened 2026-06-08
MdePkg: Remove contructor from Mpxy Library

hschauhan:
# Description
Cases when MPXY extension is not present or it fails, the system boot fails. This is easily reproducible with Qemu master branch.

Remove the constructor and let RAS agent client initialize the MPXY library and fail graceully in case of errors but the boot process continue.

## How This Was Tested

Tested on Qemu master branch. System boots and HEST table is not added


vlsunil (approved):

vlsunil:
@mergifyio queue

PR #12662, opened 2026-06-11
OvmfPkg: Add SVSM support to VirtMmCommunicationDxe

kraxel:
- **MdePkg/Svsm: add SVSM_UEFI_MM_PROTOCOL**
- **UefiCpuPkg/AmdSvsmLib: add AmdSvsmUefiMmRequest**
- **OvmfPkg/AmdSvsmLib: add AmdSvsmUefiMmRequest**
- **OvmfPkg/VirtMmCommunicationDxe: add SVSM support**

# Description

The SVSM_UEFI_MM_PROTOCOL protocol allows to send MM communication buffers to
the SVSM (instead of edk2 MM code) and let SVSM manage EFI variables that way.

This PR adds support for this to VirtMmCommunicationDxe and also adds a helper
to AmdSvsmLib.

### DIFF
--- a/OvmfPkg/Library/AmdSvsmLib/AmdSvsmLib.c
+++ b/OvmfPkg/Library/AmdSvsmLib/AmdSvsmLib.c
@@ -12,6 +12,7 @@
 #include <Library/AmdSvsmLib.h>
 #include <Register/Amd/Msr.h>
 #include <Register/Amd/Svsm.h>
+#include <Register/Amd/SvsmMsr.h>
 
 #define PAGES_PER_2MB_ENTRY  512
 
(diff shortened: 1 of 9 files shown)


kraxel:
@tlendacky please have a look.

OvmfPkg/Library/AmdSvsmLib/AmdSvsmLib.c
```diff
@@ -640,3 +641,30 @@ AmdSvsmQueryProtocol (
...
+{
+  SVSM_CALL_DATA  SvsmCallData;
+  SVSM_FUNCTION   Function;
+  UINT64          Caa;
+  UINTN           Ret;
+
+  Function.Id.Protocol = SVSM_UEFI_MM_PROTOCOL;
+  Function.Id.CallId   = SVSM_UEFI_MM_REQUEST;
+
+  // this works at runtime too
```
tlendacky: This comment should be expanded a bit to indicate why you are reading the MSR vs calling AmdSvsmSnpGetCaa(), i.e. the fact that this can be an EFI Runtime Service call and the PCD value is likely not correct because the OS will have set its own value, so read the MSR to have the whichever #VC handler is active return the proper value.

OvmfPkg/VirtMmCommunicationDxe/VirtMmCommunication.c
```diff
@@ -20,7 +20,8 @@
 
 VOID                  *mCommunicateBuffer;
 EFI_PHYSICAL_ADDRESS  mCommunicateBufferPhys;
-BOOLEAN               mUsePioTransfer = FALSE;
+BOOLEAN               mHaveSvsmProtocol = FALSE;
```
tlendacky: Just a nit, should this be consistent and be named mUseSvsmProtocol ?

tlendacky:
> @tlendacky please have a look.

Just some minor comments, otherwise it all looks reasonable and correct.

tlendacky (approved):

OvmfPkg/Library/AmdSvsmLib/AmdSvsmLib.c
kraxel: Updated.

OvmfPkg/VirtMmCommunicationDxe/VirtMmCommunication.c
kraxel: Done.

ardbiesheuvel:
For the record, this approach implies that yet another implementation of the EDK2 specific MM Get/Setvariable API is going to be deployed, this time inside the guest at a higher SVSM level?

I think this is absolutely fine, and it is probably a better place for it than QEMU. I just wanted to call it out.

So no objections from me, but please fix the uncrustify error spotted by the CI.

kraxel:
> For the record, this approach implies that yet another implementation of the EDK2 specific MM Get/Setvariable API is going to be deployed, this time inside the guest at a higher SVSM level?

Higher VMPL level, but yes, this is correct.  SVSM running at VMPL #0 manages the EFI vars, and edk2/OS running at VMPL #2 goes access them via VirtMmCommunicationDxe.  Implemented in rust, `virtfw-varstore` at crates.io.

> I think this is absolutely fine, and it is probably a better place for it than QEMU. I just wanted to call it out.

For confidential VMs you clearly don't want this in qemu.  For non-confidential VMs it continues to be useful as service provided by qemu.  And long-term I want both SVSM and QEMU use the same rust implementation. 

> but please fix the uncrustify error spotted by the CI.

Done.

ardbiesheuvel (approved):

PR #12670, opened 2026-06-12
Improving error handling in Variable Smm

PaddyDeng-v:
# Description

Gracefully fail in `MmVariableServiceInitialize`: Some platform may disable dead loop on assert. In this case, we should also abort the function in assert cases.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

After apply both commits, the platform can still boot to OS. No error is happing.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Universal/Variable/RuntimeDxe/Variable.c
+++ b/MdeModulePkg/Universal/Variable/RuntimeDxe/Variable.c
@@ -3697,6 +3697,9 @@ GetHobVariableStore (
 /**
   Initializes variable store area for non-volatile and volatile variable.
 
+  On any failure, the caller must invoke VariableCommonUninitialize() to
+  release resources that may have been partially allocated.
+
   @retval EFI_SUCCESS           Function successfully executed.
   @retval EFI_OUT_OF_RESOURCES  Fail to allocate enough memory resource.
 
@@ -3726,7 +3729,6 @@ VariableCommonInitialize (
   //
   Status = InitNonVolatileVariableStore ();
   if (EFI_ERROR (Status)) {
-    FreePool (mVariableModuleGlobal);
     return Status;
   }
 
@@ -3752,11 +3754,6 @@ VariableCommonInitialize (
   //
   Status = GetHobVariableStore (VariableGuid);
   if (EFI_ERROR (Status)) {
-    if (mNvFvHeaderCache != NULL) {
-      FreePool (mNvFvHeaderCache);
-    }
-
-    FreePool (mVariableModuleGlobal);
     return Status;
   }
 
@@ -3771,15 +3768,6 @@ VariableCommonInitialize (
   mVariableModuleGlobal->ScratchBufferSize = ScratchSize;
   VolatileVariableStore                    = AllocateRuntimePool (PcdGet32 (PcdVariableStoreSize) + ScratchSize);
   if (VolatileVariableStore == NULL) {
-    if (mVariableModuleGlobal->VariableGlobal.HobVariableBase != 0) {
-      FreePool ((VOID *)(UINTN)mVariableModuleGlobal->VariableGlobal.HobVariableBase);
-    }
-
-    if (mNvFvHeaderCache != NULL) {
-      FreePool (mNvFvHeaderCache);
-    }
-
-    FreePool (mVariableModuleGlobal);
     return EFI_OUT_OF_RESOURCES;
   }
 
@@ -3799,6 +3787,47 @@ VariableCommonInitialize (
   VolatileVariableStore->Reserved1 = 0;
 
   return EFI_SUCCESS;
+}
+
+/**
+  Uninitialize variable store area, freeing all resources allocated by
+  VariableCommonInitialize().
+
+**/
+VOID
+VariableCommonUninitialize (
+  VOID
+  )
+{
+  if (mVariableModuleGlobal != NULL) {
+    if (mVariableModuleGlobal->VariableGlobal.HobVariableBase != 0) {
+      FreePool ((VOID *)(UINTN)mVariableModuleGlobal->VariableGlobal.HobVariableBase);
+      mVariableModuleGlobal->VariableGlobal.HobVariableBase = 0;
+    }
+
+    if (mVariableModuleGlobal->VariableGlobal.VolatileVariableBase != 0) {
+      FreePool ((VOID *)(UINTN)mVariableModuleGlobal->VariableGlobal.VolatileVariableBase);
+    }
+
+    FreePool (mVariableModuleGlobal);
+    mVariableModuleGlobal = NULL;
+  }
+
+  if (mNvFvHeaderCache != NULL) {
+    //
+    // In real NV mode, mNvVariableCache points into the same allocation as mNvFvHeaderCache.
+    //
+    ASSERT ((UINTN)mNvVariableCache == (UINTN)mNvFvHeaderCache + mNvFvHeaderCache->HeaderLength);
+    FreePool (mNvFvHeaderCache);
+    mNvFvHeaderCache = NULL;
+    mNvVariableCache = NULL;
+  } else if ((mNvVariableCache != NULL) && (PcdGet64 (PcdEmuVariableNvStoreReserved) == 0)) {
+    //
+    // In emulated NV mode without a pre-reserved store, mNvVariableCache is a dynamic allocation.
+    //
+    FreePool (mNvVariableCache);
+    mNvVariableCache = NULL;
+  }
 }
 
 /**
--- a/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableDxe.c
+++ b/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableDxe.c
@@ -568,7 +568,12 @@ ProtocolIsVariablePolicyEnabled (
   @param[in] ImageHandle    The firmware allocated handle for the EFI image.
   @param[in] SystemTable    A pointer to the EFI System Table.
 
-  @retval EFI_SUCCESS       Variable service successfully initialized.
+  @retval EFI_SUCCESS           Variable service successfully initialized.
+  @retval EFI_OUT_OF_RESOURCES  Insufficient memory to allocate variable
+                                storage or event/protocol resources.
+  @retval EFI_VOLUME_CORRUPTED  The non-volatile variable store is corrupted.
+  @retval Others                An error from protocol installation, event
+                                creation, or variable policy initialization.
 
 **/
 EFI_STATUS
@@ -581,9 +586,40 @@ VariableServiceInitialize (
   EFI_STATUS  Status;
   EFI_EVENT   ReadyToBootEvent;
   EFI_EVENT   EndOfDxeEvent;
+  EFI_EVENT   FtwNotifyEvent;
+  BOOLEAN     VariableLockProtocolInstalled;
+  BOOLEAN     VarCheckProtocolInstalled;
+  BOOLEAN     RuntimeServicesUpdated;
+  BOOLEAN     FtwNotifyRegistered;
+  BOOLEAN     VirtualAddressChangeRegistered;
+  BOOLEAN     ReadyToBootEventCreated;
+  BOOLEAN     EndOfDxeEventCreated;
+  BOOLEAN     VariablePolicyLibInitialized;
+  BOOLEAN     VariablePolicyProtocolInstalled;
+
+  ReadyToBootEvent                = NULL;
+  EndOfDxeEvent                   = NULL;
+  FtwNotifyEvent                  = NULL;
+  VariableLockProtocolInstalled   = FALSE;
+  VarCheckProtocolInstalled       = FALSE;
+  RuntimeServicesUpdated          = FALSE;
+  FtwNotifyRegistered             = FALSE;
+  VirtualAddressChangeRegistered  = FALSE;
+  ReadyToBootEventCreated         = FALSE;
+  EndOfDxeEventCreated            = FALSE;
+  VariablePolicyLibInitialized    = FALSE;
+  VariablePolicyProtocolInstalled = FALSE;
 
   Status = VariableCommonInitialize ();
-  ASSERT_EFI_ERROR (Status);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a: VariableCommonInitialize failed - %r. Aborting variable init.\n",
+      __func__,
+      Status
+      ));
+    goto ErrorExit;
+  }
 
   Status = gBS->InstallMultipleProtocolInterfaces (
                   &mHandle,
@@ -591,7 +627,17 @@ VariableServiceInitialize (
                   &mVariableLock,
                   NULL
                   );
-  ASSERT_EFI_ERROR (Status);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a: Installing gEdkiiVariableLockProtocolGuid failed - %r. Aborting variable init.\n",
+      __func__,
+      Status
+      ));
+    goto ErrorExit;
+  }
+
+  VariableLockProtocolInstalled = TRUE;
 
   Status = gBS->InstallMultipleProtocolInterfaces (
                   &mHandle,
@@ -599,35 +645,46 @@ VariableServiceInitialize (
                   &mVarCheck,
                   NULL
                   );
-  ASSERT_EFI_ERROR (Status);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a: Installing gEdkiiVarCheckProtocolGuid failed - %r. Aborting variable init.\n",
+      __func__,
+      Status
+      ));
+    goto ErrorExit;
+  }
+
+  VarCheckProtocolInstalled = TRUE;
 
   SystemTable->RuntimeServices->GetVariable         = VariableServiceGetVariable;
   SystemTable->RuntimeServices->GetNextVariableName = VariableServiceGetNextVariableName;
   SystemTable->RuntimeServices->SetVariable         = VariableServiceSetVariable;
   SystemTable->RuntimeServices->QueryVariableInfo   = VariableServiceQueryVariableInfo;
-
-  //
-  // Now install the Variable Runtime Architectural protocol on a new handle.
-  //
-  Status = gBS->InstallProtocolInterface (
-                  &mHandle,
-                  &gEfiVariableArchProtocolGuid,
-                  EFI_NATIVE_INTERFACE,
-                  NULL
-                  );
-  ASSERT_EFI_ERROR (Status);
+  RuntimeServicesUpdated                            = TRUE;
 
   if (!PcdGetBool (PcdEmuVariableNvModeEnable)) {
     //
     // Register FtwNotificationEvent () notify function.
     //
-    EfiCreateProtocolNotifyEvent (
-      &gEfiFaultTolerantWriteProtocolGuid,
-      TPL_CALLBACK,
-      FtwNotificationEvent,
-      (VOID *)SystemTable,
-      &mFtwRegistration
-      );
+    FtwNotifyEvent = EfiCreateProtocolNotifyEvent (
+                       &gEfiFaultTolerantWriteProtocolGuid,
+                       TPL_CALLBACK,
+                       FtwNotificationEvent,
+                       (VOID *)SystemTable,
+                       &mFtwRegistration
+                       );
+    if (FtwNotifyEvent == NULL) {
+      DEBUG ((
+        DEBUG_ERROR,
+        "%a: EfiCreateProtocolNotifyEvent for FTW failed. Aborting variable init.\n",
+        __func__
+        ));
+      Status = EFI_OUT_OF_RESOURCES;
+      goto ErrorExit;
+    }
+
+    FtwNotifyRegistered = TRUE;
   } else {
     //
     // Emulated non-volatile variable mode does not depend on FVB and FTW.
@@ -643,7 +700,17 @@ VariableServiceInitialize (
                   &gEfiEventVirtualAddressChangeGuid,
                   &mVirtualAddressChangeEvent
                   );
-  ASSERT_EFI_ERROR (Status);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a: CreateEventEx for VirtualAddressChange failed - %r. Aborting variable init.\n",
+      __func__,
+      Status
+      ));
+    goto ErrorExit;
+  }
+
+  VirtualAddressChangeRegistered = TRUE;
 
   //
   // Register the event handling function to reclaim variable for OS usage.
@@ -654,7 +721,17 @@ VariableServiceInitialize (
              NULL,
              &ReadyToBootEvent
              );
-  ASSERT_EFI_ERROR (Status);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a: EfiCreateEventReadyToBootEx failed - %r. Aborting variable init.\n",
+      __func__,
+      Status
+      ));
+    goto ErrorExit;
+  }
+
+  ReadyToBootEventCreated = TRUE;
 
   //
   // Register the event handling function to set the End Of DXE flag.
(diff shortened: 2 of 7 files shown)


MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
```diff
@@ -1173,15 +1182,33 @@ MmVariableServiceInitialize (
...
+  }
 
   Status = gMmst->MmInstallProtocolInterface (
                     &VariableHandle,
                     &gEdkiiSmmVarCheckProtocolGuid,
                     EFI_NATIVE_INTERFACE,
                     &mSmmVarCheck
                     );
-  ASSERT_EFI_ERROR (Status);
+  if (EFI_ERROR (Status)) {
```
makubacki, line 1248: @PaddyDeng-v, I think you need to consider system-state cleanup since this is a behavior change. Previously, an assertion call either:

- Did nothing
- Triggered a breakpoint
- Entered a deadloop

This change adds a new possibility to return partially through the initialization flow at various points to firmware execution.

If subsequent initialization fails, `gEfiSmmVariableProtocolGuid` should likely be uninstalled as it exposes MM variable services within the MM environment that would be in a "partially initialized" state.

For example, if `gEfiSmmVariableProtocolGuid` is installed successfully (Line 1179), but `gEdkiiSmmVarCheckProtocolGuid` installation fails (Line 1196), that means the call to register the variable SMI handler (Line 1237) is not executed. MM drivers (using `gEfiSmmVariableProtocolGuid` directly) might participate in a flow where they expect variables written from outside the MM environment (which relies on the SMI handler) and that is not available.

These types of dependencies can be established and broken down throughout initialization, but I think the simplest answer is that "if initialization fails, variable services are not available". So, interfaces in global databases (like the protocol database) are removed.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
```diff
@@ -505,6 +505,12 @@ SmmVariableHandler (
   SmmVariableFunctionHeader = (SMM_VARIABLE_COMMUNICATE_HEADER *)CommBuffer;
   switch (SmmVariableFunctionHeader->Function) {
     case SMM_VARIABLE_FUNCTION_GET_VARIABLE:
+      if (mNvVariableCache == NULL) {
```
MatthiasHuang: [`mNvVariableCache`](https://github.com/PaddyDeng-v/edk2/blob/5b559c9724e7e8a11da9ea3bb9122420bb3e8017/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableNonVolatile.c#L322) is  assigned in [`InitNonVolatileVariableStore()`](https://github.com/PaddyDeng-v/edk2/blob/5b559c9724e7e8a11da9ea3bb9122420bb3e8017/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableNonVolatile.c#L294) via  [`VariableCommonInitialize()`](https://github.com/PaddyDeng-v/edk2/blob/5b559c9724e7e8a11da9ea3bb9122420bb3e8017/MdeModulePkg/Universal/Variable/RuntimeDxe/Variable.c#L3705),  which runs at the very top of [`MmVariableServiceInitialize`](https://github.com/PaddyDeng-v/edk2/blob/5b559c9724e7e8a11da9ea3bb9122420bb3e8017/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c#L1199). Since commit 1 now aborts before [`MmiHandlerRegister`](https://github.com/PaddyDeng-v/edk2/blob/5b559c9724e7e8a11da9ea3bb9122420bb3e8017/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c#L1285) when that fails, the SMI handler should only ever be registered when `mNvVariableCache` is non-NULL, so these per-case NULL checks look unreachable in the normal flow. Are they guarding a specific scenario (e.g. a custom/downstream init ordering)? If so, a brief comment would help future readers; if not, they could be dropped, since commit 1 is the actual safeguard

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
```diff
@@ -505,6 +505,12 @@ SmmVariableHandler (
...
     case SMM_VARIABLE_FUNCTION_GET_VARIABLE:
+      if (mNvVariableCache == NULL) {
+        DEBUG ((DEBUG_ERROR, "GetVariable: NV variable cache not initialized!\n"));
+        Status = EFI_NOT_READY;
+        goto EXIT;
+      }
+
       if (CommBufferPayloadSize < OFFSET_OF (SMM_VARIABLE_COMMUNICATE_ACCESS_VARIABLE, Name)) {
         DEBUG ((DEBUG_ERROR, "GetVariable: SMM communication buffer size invalid!\n"));
         return EFI_SUCCESS;
```
MatthiasHuang, line 510: Small consistency note: the new guards correctly set `Status` and `goto EXIT` (so  [`ReturnStatus`](https://github.com/PaddyDeng-v/edk2/blob/5b559c9724e7e8a11da9ea3bb9122420bb3e8017/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c#L1043) is written  back), but the sibling buffer-size checks in these cases still `return EFI_SUCCESS` without setting `ReturnStatus`. Because the DXE caller reads  `SmmVariableFunctionHeader->ReturnStatus` directly, that path can surface a stale value from the previous operation. Converting these to `Status = EFI_INVALID_PARAMETER; goto  EXIT;` would make the handler consistent. (Pre-existing; optional, but you're already in these cases.)

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
PaddyDeng-v: Hi @makubacki ,

I added clean up path in the latest commit. Thanks for the commenting.
Please kindly help to review it.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
PaddyDeng-v: Hi @MatthiasHuang ,
Your opinion sounds reasonable. I just removed the second commit from the PR. Thanks.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
PaddyDeng-v: Code change regarding smi handler has been removed. Close this conversation.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
makubacki, line 1283: In Traditional MM, this will install `gEfiSmmVariableProtocolGuid` in the DXE protocol data (`gBS`). Because of the coupled nature of Traditional MM, the protocol is often used a dependency in Traditional MM drivers that need to use variable services where they rely on the protocol in the DXE database (not the MM/`gMmst` database). For example: [Tcg2Smm.inf](https://github.com/tianocore/edk2/blob/master/SecurityPkg/Tcg/Tcg2Smm/Tcg2Smm.inf)

You currently have code to uninstall the protocol from the MM protocol database based on `SmmVariableProtocolInstalled`, but not the DXE protocol database. It will, at least, need to be uninstalled in the DXE database as well.

Also, due to the protocols (in both databases) being a common dependency for module dispatch, I think it's also worth considering a rearrangement of actions in the function so that fallible pre-conditions (like allocating buffers) occur before mutating overall system state (like installing and then quickly uninstalling a protocol).

makubacki: I suggest moving this `gMmst->MmAllocatePool()` call earlier in the flow. Right now, if the allocation fails, `gEfiSmmVariableProtocolGuid` and `gEdkiiSmmVarCheckProtocolGuid` have to be uninstalled if it fails. That disruption to the protocol database can be avoided by verifying the allocation is successful earlier.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
```diff
@@ -1192,14 +1226,35 @@ MmVariableServiceInitialize (
                     mVariableBufferPayloadSize,
                     (VOID **)&mVariableBufferPayload
                     );
-  ASSERT_EFI_ERROR (Status);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a: VariableCommonInitialize failed - %r. Aborting SMM variable init.\n",
```
makubacki: Stale copy/paste message.

```suggestion
      "%a: gMmst->MmAllocatePool failed - %r. Aborting SMM variable init.\n",
```

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
```diff
@@ -1153,15 +1153,29 @@ MmVariableServiceInitialize (
...
+  BOOLEAN     SmmVariableProtocolInstalled = FALSE;
+  BOOLEAN     SmmVarCheckProtocolInstalled = FALSE;
+  BOOLEAN     SmmVariableHandlerRegistered = FALSE;
+  BOOLEAN     SmmEndOfDxeNotifyRegistered  = FALSE;
+  BOOLEAN     SmmFtwNotifyRegistered       = FALSE;
 
   //
   // Variable initialize.
   //
   Status = VariableCommonInitialize ();
```
makubacki, line 1185: `VariableCommonInitialize()` allocates a very large buffer assigned to `mVariableModuleGlobal->VariableGlobal.VolatileVariableBase` and nothing in that function is unwound like here.

For symmetry/cleanliness, if all of this is being undone on error paths, consider adding a `VariableCommonUninitialize()` function that knows how to undo its counterpart's initialization.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
```diff
@@ -1192,14 +1226,35 @@ MmVariableServiceInitialize (
...
-  ASSERT_EFI_ERROR (Status);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a: VariableCommonInitialize failed - %r. Aborting SMM variable init.\n",
+      __func__,
+      Status
+      ));
+    goto ErrorExit;
+  }
```
makubacki: Right now, `mVariableBufferPayloadSize` is reset to `0` if `mVariableBufferPayload != NULL`, but if this allocation fails `mVariableBufferPayload` is still `NULL`. I think you also reset it here to keep global state clean.


```suggestion
  if (EFI_ERROR (Status)) {
    mVariableBufferPayloadSize = 0;
    DEBUG ((
      DEBUG_ERROR,
      "%a: VariableCommonInitialize failed - %r. Aborting SMM variable init.\n",
      __func__,
      Status
      ));
    goto ErrorExit;
  }
```

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
```diff
@@ -1225,7 +1290,17 @@ MmVariableServiceInitialize (
...
+        "%a: MmRegisterProtocolNotify failed - %r. Aborting SMM variable init.\n",
+        __func__,
+        Status
+        ));
+      goto ErrorExit;
+    }
+
+    SmmFtwNotifyRegistered = TRUE;
 
     SmmFtwNotificationEvent (NULL, NULL, NULL);
```
makubacki: Since you're focused on fixing error handling in this function, the return value from `SmmFtwNotificationEvent()` is entirely ignored right now. It would be nice if it was also handled.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
makubacki: Please describe the potential error codes now potentially returned from the function.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
```diff
@@ -1153,15 +1153,29 @@ MmVariableServiceInitialize (
   )
 {
   EFI_STATUS  Status;
-  EFI_HANDLE  VariableHandle;
-  VOID        *SmmFtwRegistration;
-  VOID        *SmmEndOfDxeRegistration;
+  EFI_HANDLE  VariableHandle               = NULL;
```
makubacki: I know the code was already written this way, but `VariableHandle` is set to `NULL` again before it is used:

```c
  VariableHandle = NULL;
  Status         = gMmst->MmInstallProtocolInterface (
                            &VariableHandle,
                            &gEfiSmmVariableProtocolGuid,
                            EFI_NATIVE_INTERFACE,
                            &gSmmVariable
                            );
```

I don't see where it is modified until then, so you can drop this redundant assignment.

```suggestion
```

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
```diff
@@ -1225,7 +1290,17 @@ MmVariableServiceInitialize (
                       SmmFtwNotificationEvent,
                       &SmmFtwRegistration
                       );
-    ASSERT_EFI_ERROR (Status);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((
+        DEBUG_ERROR,
+        "%a: MmRegisterProtocolNotify failed - %r. Aborting SMM variable init.\n",
```
makubacki: This is the exact same message as the one used above for `gEfiMmEndOfDxeProtocolGuid`. Can you please make it unique, so it is easier to map to the corresponding call?

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
PaddyDeng-v: Added `VariableClearNotifySmmReady ()` interface to clean up the traditional SMM / standalone MM side effects.
Since `VariableNotifySmmReady ()` in standalone MM is a no-op, `VariableClearNotifySmmReady ()` is also a no-op there.

PaddyDeng-v: Relocated MmAllocatePool.
It is currently only after `VariableCommonInitialize()`

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
PaddyDeng-v: Fixed in the latest commit

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
PaddyDeng-v: Created `VariableCommonUninitialize ()` to free up space. Since `VariableCommonInitialize ()` is also used in `VariableDxe.c`. A similar clean up pattern is applied to `VariableServiceInitialize()` in `VariableDxe.c`

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
PaddyDeng-v: Since `SmmFtwNotificationEvent` is also registered as a notification, I think `SmmFtwNotificationEvent` should clean its side effect in its own error exit path. 
At the direct call here, the error could be expected because the protocol it requires is not installed yet. I think we can keep the current design, or use `if (Status != EFI_SUCCESS && Status != EFI_NOT_FOUND) { goto ErrorExit; }`.

Please let me know your thoughts, and if you prefer adding `SmmFtwNotificationEvent` own error exit path in this PR or another. Thanks.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
PaddyDeng-v: Updated debug message to describe the notify we are registering.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
PaddyDeng-v: Updated

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
PaddyDeng-v: Added comment describe error status return from this function. Other errors return from deeper function code are summarized as `Others`

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
PaddyDeng-v: Updated

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
makubacki: @PaddyDeng-v, you're correct that it's possible that `gEfiSmmFaultTolerantWriteProtocolGuid` might not be installed yet and an error return code is expected. This call handles the case where the protocol is already installed.

I was just asking for the return value to be handled here at all. Your suggestion of `if (Status != EFI_SUCCESS && Status != EFI_NOT_FOUND) { goto ErrorExit; }` seems reasonable.

makubacki:
@lgao4 & @mikebeaton, I think that this PR is at a point where you should give a review as well.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmm.c
PaddyDeng-v: Added `if (Status != EFI_SUCCESS && Status != EFI_NOT_FOUND) { goto ErrorExit; }` as discussed.
Tested no issue on booting to OS with the current code change.

MdeModulePkg/Universal/Variable/RuntimeDxe/Variable.c
```diff
@@ -3801,6 +3801,49 @@ VariableCommonInitialize (
   return EFI_SUCCESS;
 }
 
+/**
+  Uninitialize variable store area.
+
+  @retval EFI_SUCCESS           Function successfully executed.
+
+**/
+EFI_STATUS
```
makubacki: nit: This function does not need to return a value if it only is going to return `EFI_SUCCESS` and the return value will be ignored anyway.

MdeModulePkg/Universal/Variable/RuntimeDxe/Variable.c
```diff
@@ -3801,6 +3801,49 @@ VariableCommonInitialize (
...
+    }
+
+    FreePool (mVariableModuleGlobal);
+    mVariableModuleGlobal = NULL;
+  }
+
+  if (mNvFvHeaderCache != NULL) {
+    //
+    // In real NV mode, mNvVariableCache points into the same allocation as mNvFvHeaderCache.
+    //
```
makubacki, line 3819: This would be more effectively confirmed with an assertion that they are the same.

makubacki:
@lgao4, can you please review as well?

MdeModulePkg/Universal/Variable/RuntimeDxe/Variable.c
PaddyDeng-v: Updated as suggested

MdeModulePkg/Universal/Variable/RuntimeDxe/Variable.c
PaddyDeng-v: Update as suggested

MdeModulePkg/Universal/Variable/RuntimeDxe/Variable.c
```diff
@@ -3801,6 +3801,47 @@ VariableCommonInitialize (
...
+      mVariableModuleGlobal->VariableGlobal.HobVariableBase = 0;
+    }
+
+    if (mVariableModuleGlobal->VariableGlobal.VolatileVariableBase != 0) {
+      FreePool ((VOID *)(UINTN)mVariableModuleGlobal->VariableGlobal.VolatileVariableBase);
+    }
+
+    FreePool (mVariableModuleGlobal);
+    mVariableModuleGlobal = NULL;
+  }
```
makubacki, line 3814: The original `VariableCommonInitialize()` implementation was hidden by the GitHub UI for most of the review (since it wasn't changed), but looking at it now, `VariableCommonInitialize()` already frees some buffers that are freed in `VariableCommonUninitialize()` but it does not set the global variable to `NULL`, so there could be a double free.

---

For example, in `VariableCommonInitialize()`:

```c
  //
  // Init non-volatile variable store.
  //
  Status = InitNonVolatileVariableStore ();
  if (EFI_ERROR (Status)) {
    FreePool (mVariableModuleGlobal);
    return Status;
  }
```

Also in `VariableCommonInitialize()`:

```c
  //
  // Get HOB variable store.
  //
  Status = GetHobVariableStore (VariableGuid);
  if (EFI_ERROR (Status)) {
    if (mNvFvHeaderCache != NULL) {
      FreePool (mNvFvHeaderCache);
    }

    FreePool (mVariableModuleGlobal);
    return Status;
  }
```

If those fail, the buffer is already freed, the variable is not set to `NULL`, and on return `VariableServiceInitialize()` execution goes to `ErrorExit`:

```c
  Status = VariableCommonInitialize ();
  if (EFI_ERROR (Status)) {
    DEBUG ((
      DEBUG_ERROR,
      "%a: VariableCommonInitialize failed - %r. Aborting variable init.\n",
      __func__,
      Status
      ));
    goto ErrorExit;
  }
```

`ErrorExit` calls `VariableCommonUninitialize()`:

```c
  VariableCommonUninitialize ();

  ASSERT_EFI_ERROR (Status);
  return Status;
```

Which will then try to use and free the freed buffer again:

```c
/**
  Uninitialize variable store area, freeing all resources allocated by
  VariableCommonInitialize().

**/
VOID
VariableCommonUninitialize (
  VOID
  )
{
  if (mVariableModuleGlobal != NULL) {
    if (mVariableModuleGlobal->VariableGlobal.HobVariableBase != 0) {
      FreePool ((VOID *)(UINTN)mVariableModuleGlobal->VariableGlobal.HobVariableBase);
      mVariableModuleGlobal->VariableGlobal.HobVariableBase = 0;
    }

    if (mVariableModuleGlobal->VariableGlobal.VolatileVariableBase != 0) {
      FreePool ((VOID *)(UINTN)mVariableModuleGlobal->VariableGlobal.VolatileVariableBase);
    }

    FreePool (mVariableModuleGlobal);
    mVariableModuleGlobal = NULL;
  }
```

---

In the new design with `VariableCommonUninitialize()` and these being internal-only functions, I think it is clearest to just route all `VariableCommonInitialize()` related cleanup to `VariableCommonUninitialize()`.

Update `VariableCommonInitialize()`'s function description to state `VariableCommonUninitialize()` should be called on any failure, drop all of the redundant cleanup code in `VariableCommonInitialize()`, and then verify all code interacting with the cleanup code in `VariableCommonUninitialize()` is coordinating with it correctly.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableDxe.c
```diff
@@ -615,19 +672,40 @@ VariableServiceInitialize (
                   EFI_NATIVE_INTERFACE,
                   NULL
                   );
```
makubacki: Consider moving protocol installation for `gEfiVariableArchProtocolGuid` to the end of the function. Uninstalling it doesn't account for notifications that may have already happened on installation and logic in those drivers probably also don't account for removal. It's best to not install it until confidence is high the variable service will be present.

makubacki (commented):
I know PR comments are stretching on a bit, but this is an impactful change with a lot of ramifications.

MdeModulePkg/Universal/Variable/RuntimeDxe/Variable.c
PaddyDeng-v: Updated and did a simple boot test.
Please help to review it. Thanks!

PaddyDeng-v:
Hi @makubacki and @lgao4 ,

I think every existing discussion in this PR is resolved.
Could you please help to provide your input on the current state, or we could start preparing to merge it?

Regards,
Paddy

MarcChen46:
Hi @makubacki and @lgao4 
Please kindly review this PR and see if you have any comment, if no concern, please approve it.
Understand the code is locked for stable tag release now, we can wait the new stable tag get released then merge the PR after that. :)

lgao4:
> Hi @makubacki and @lgao4 Please kindly review this PR and see if you have any comment, if no concern, please approve it. Understand the code is locked for stable tag release now, we can wait the new stable tag get released then merge the PR after that. :)

I have no other comments. The change is good to me.

lgao4 (approved):

MdeModulePkg/Universal/Variable/RuntimeDxe/Variable.c
makubacki: Lgtm, thanks

makubacki (approved):

MarcChen46:
@lgao4 , please help merging this PR since the `edk2-stable202608` has been released.

PaddyDeng-v:
Updated the PR to base on the latest origin/main

PR #12671, opened 2026-06-12
MdeModulePkg: Dxe: Skip FV Extraction When FV3 HOB Found

os-d:
# Description

Currently, the DXE dispatcher will skip extracting an FV file when an FV2 HOB is found for it, as that indicates pre-DXE extracted it.

However, the dispatcher does not check for FV3 HOBs, which also can describe extracted FVs. That can result in extracting the same FV in DXE that is already extracted, which can be a large performance hit.

This updates the DXE dispatcher to check for the existence of either an FV2 or FV3 HOB for this FV and skip extracting if either is found.

- [ ] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [ ] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

## How This Was Tested

Tested on ArmVirtPkg/OvmfPkg by not producing FV2 HOBs for extracted regions, only FV3, and seeing DXE correctly skip re-extracting them.

## Integration Instructions

N/A.

PR #12672, opened 2026-06-12
EmbeddedPkg, OvmfPkg: Improve FV2/3 HOB Handling

### LINKED ISSUE #9780: PrePiLib: incorrect population of FV2 HOB FvName field (Bugzilla Bug 2094)
This issue was created automatically with [bugzilla2github](https://github.com/makubacki/bugzilla2github/tree/bz2gh_tianocore)
# Bugzilla Bug 2094
Date: 2019-08-19T17:34:52+00:00
From: Dmitry Antipov &lt;<dmanti>&gt;
To:   Matthew Carlson &lt;<macarl>&gt;
CC:   @lgao4, macarl

Last updated: 2020-10-13T21:47:07+00:00

os-d:
# Description

This PR improves handling of FV3 HOBs in EmbeddedPkg and OvmfPkg, which have not been updated to follow what PEI Core does.

There are three commits in this PR:

EmbeddedPkg: PrePi.h: Drop Duplicate HOB Defintions
--
PrePi.h has some duplicate HOB definitions from HobLib.h. Instead of redefining them, use HobLib.h as this already defined for the PrePi environment.

EmbeddedPkg: PrePiLib: Skip Extracting if FV3 HOB Found
--
PrePiLib only searches for the existence of an FV2 HOB to determine if an FV has already been extracted. However, FV2 HOBs have been superceded by FV3 HOBs for describing extracted FVs.

This commit updates PrePiLib to check for the existence of either an FV2 or FV3 HOB for this FV and skips extracting if either is found.

This follows the same proposed behavior in https://github.com/tianocore/edk2/pull/12671.

EmbeddedPkg: PrePiLib: Zero Volume Info Buffer Before Writing
--
FfsGetVolumeInfo() in PrePiLib does not zero the VolumeInfo buffer before writing to it. However, it has some fields that are expected to be zero if certain conditions are not met.

For example, the FvName is expected to be all 0's if the extended header is not present in the FV. Currently, the FvName field will point to garbage if the caller does not zero the buffer first.

Flip the expectation and have FfsGetVolumeInfo() zero the buffer before it writes to it.

EmbeddedPkg, OvmfPkg: PrePiLib: Use Correct FvName in FV2 HOB Production
--
PrePiLib is incorrectly producing FV2 HOBs. The FvName field of the FV2 HOB is intended to be the parent FV's name. This way the HOB consumer phase can identify when an FV has already been extracted: if the file name of the extracted FV and the parent FV name match in the FV2 HOB.

With the incorrect FvName, DXE will re-extract the already extracted FV, which can be a large performance hit (some platforms we have seen ~1 second).

This requires an update to FfsProcessFvFile to pass the parent FV handle through so the FV2 HOB can be correctly produced. OvmfPkg consumers are also updated.

Closes #9780 .

OvmfPkg: PeilessStartupLib: Fix FV2 HOB Production
--
PeilessStartupLib is incorrectly producing FV2 HOBs. The FvName field of the FV2 HOB is intended to be the parent FV's name. This way the HOB consumer phase can identify when an FV has already been extracted: if the file name of the extracted FV and the parent FV name match in the FV2 HOB.

With the incorrect FvName, DXE will re-extract the already extracted FV, which can be a large performance hit (some platforms we have seen ~1 second).

EmbeddedPkg, OvmfPkg: Produce FV3 HOBs for Extracted FVs
--
EmbeddedPkg's PrePiLib and OvmfPkg's PeilessStartupLib (which uses PrePiLib) only produce FV2 HOBs for extracted FVs. However, FV3 HOBs have superceded FV2 HOBs for describing extracted FVs.

This commit updates both of these libs to also produce FV3 HOBs, following PEI's pattern of producing an FV, FV2, and FV3 HOB for each extracted FV.

Eventually, only FV3 HOBs will be produced, so this paves the way for that.

In order to produce the FV3 HOB, the authentication status is required to be piped out from PrePiLib when the GUIDed extraction occurs. As such, a breaking change is introduced to add an optional parameter to `FfsFindSectionDataWithHook()` that allows the authentication status to be retrieved.

- [x] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Tested booting OvmfPkg/IntelTdx to shell with this change and with only producing FV3 HOBs. Confirmed that extracted FVs were being re-extracted with current code, after the changes, not being re-extracted.

## Integration Instructions

Callers of `FfsFindSectionDataWithHook()` must update the call to add the `AuthenticationStatus` parameter. For existing use cases, this can always be `NULL`. If the authentication status is desired to be returned, then a UINT32 pointer may be passed here.

Callers of `FfsProcessFvFile()` must update the call to add the `ParentVolumeHandle`.

### DIFF
--- a/EmbeddedPkg/Include/Library/PrePiLib.h
+++ b/EmbeddedPkg/Include/Library/PrePiLib.h
@@ -180,7 +180,8 @@ FfsGetVolumeInfo (
 /**
   Get Fv image from the FV type file, then add FV & FV2 Hob.
 
-  @param FileHandle      File handle of a Fv type file.
+  @param FvFileHandle        File handle of a Fv type file.
+  @param ParentVolumeHandle  Parent volume handle, for filling out FvName field in FV2 Hob
 
   @retval EFI_NOT_FOUND  FV image can't be found.
   @retval EFI_SUCCESS    Successfully to process it.
@@ -189,7 +190,8 @@ FfsGetVolumeInfo (
 EFI_STATUS
 EFIAPI
 FfsProcessFvFile (
-  IN  EFI_PEI_FILE_HANDLE  FvFileHandle
+  IN  EFI_PEI_FILE_HANDLE  FvFileHandle,
+  IN EFI_PEI_FV_HANDLE     ParentVolumeHandle
   );
 
 /**
@@ -209,22 +211,6 @@ FfsAnyFvFindFirstFile (
   IN  EFI_FV_FILETYPE      FileType,
   OUT EFI_PEI_FV_HANDLE    *VolumeHandle,
   OUT EFI_PEI_FILE_HANDLE  *FileHandle
-  );
-
-/**
-  Get Fv image from the FV type file, then add FV & FV2 Hob.
-
-  @param FileHandle  File handle of a Fv type file.
-
-
-  @retval EFI_NOT_FOUND  FV image can't be found.
-  @retval EFI_SUCCESS    Successfully to process it.
-
-**/
-EFI_STATUS
-EFIAPI
-FfsProcessFvFile (
-  IN  EFI_PEI_FILE_HANDLE  FvFileHandle
   );
 
 /**
--- a/EmbeddedPkg/Include/Library/PrePiLib.h
+++ b/EmbeddedPkg/Include/Library/PrePiLib.h
@@ -68,9 +68,11 @@ EFI_STATUS
 
   @param  SectionType           The value of the section type to find.
   @param  SectionCheckHook      A hook which can check if the section is the target one.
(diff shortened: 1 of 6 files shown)


os-d:
I need to add one change here, I realized that a previous Project Mu commit wasn't upstreamed that fixed the FV2 production here, the current code is incorrectly using the extracted FV name in the FV2 HOB instead of the parent FV.

os-d:
> I need to add one change here, I realized that a previous Project Mu commit wasn't upstreamed that fixed the FV2 production here, the current code is incorrectly using the extracted FV name in the FV2 HOB instead of the parent FV.

Updated and linked the GH issue.

ardbiesheuvel (approved):

PR #12182, opened 2026-02-21
MdePkg: Add Stack Cookie Support for CLANGPDB IA32, X64, and AArch64

os-d:
# Description

This PR provides a series of commits to enable stack cookies on CLANGPDB for IA32, X64, and AArch64. CLANGPDB was failing to link StackCheckLib because it had already performed LTO on its dependencies (BaseLib, DebugLib) and StackCheckLib cannot be LTO'd. As a result, this PR also drops all dependencies from StackCheckLib, which both simplifies the implementation and fixes the issue for adopting CLANGPDB support.

See the commit messages for detailed descriptions of the changes.

Closes #12002 

- [x] Breaking change?
  - **Breaking change** - Does this PR cause a break in build or boot behavior?
  - Examples: Does it add a new library class or move a module to a different repo.
- [x] Impacts security?
  - **Security** - Does this PR have a direct security impact?
  - Examples: Crypto algorithm change or buffer overflow fix.
- [ ] Includes tests?
  - **Tests** - Does this PR include any explicit test code?
  - Examples: Unit tests or integration tests.

## How This Was Tested

Tested on OVMFIA32, OVMFX64, and ArmVirtQemu with stack cookie violations in PEI, DXE, traditional SMM, and standalone MM (the latter two for OVMF only, of course). These were tested with VS2022, GCC, and CLANGPDB for all architectures where supported and on both Windows and Linux for CLANGPDB.

In all cases, when combined with https://github.com/tianocore/edk2/pull/12181, the stack cookie exception was printed clearly with the faulting address.

## Integration Instructions

This is marked a breaking change because it removes PcdStackCookieExceptionVector and StackCheckFailureHookLib. However, for most platforms (which were taking the default PCD value and the null lib) no change is required, but if platforms were overriding these, they would need to update to register an interrupt handler for exception ID 0x42.

CLANGPDB users will now have stack cookies enabled; by default StackCheckLibNull will be linked and stack cookie corruptions will be ignored. View the README for enablement instructions.

This also changes StackCheckLib to be able to be compiled under LTO. It uses MdeLibs.dsc.inc to forcibly include the relevant symbols. If a platform does not use this file, they should include the following in their DSC:

```inf
[BuildOptions]
  #
  # Linker flags to ensure that stack cookie symbols are not optimized out when StackCheckLib is included.
  #
  MSFT:*_*_IA32_DLINK_FLAGS = /INCLUDE:___security_cookie
[...]

PR #12677, opened 2026-06-15
Add the stack protector for LoongArch64

kilaterlee:
# Description

Patch1: Added LoongArch64 stack cookie interrupt instance, calling `CpuBreakpoint` to stop the CPU.
Patch2: Add LoongArch64 support for dynamic stack cookie entry point library
Patch3: Enable stack protector for LoongArch64

## How This Was Tested

Build a  module using the `-stack-protector-all`, and try to tamper with the "guard value" after it has been loaded the `guard` but before it is compared, it should `break` the CPU and tell us the stack has been changed.

## Integration Instructions

N/A

### DIFF
--- a/MdePkg/Library/DynamicStackCookieEntryPointLib/DxeCoreEntryPoint.inf
+++ b/MdePkg/Library/DynamicStackCookieEntryPointLib/DxeCoreEntryPoint.inf
@@ -19,7 +19,7 @@
 
 
 #
-#  VALID_ARCHITECTURES           = IA32 X64 AARCH64
+#  VALID_ARCHITECTURES           = IA32 X64 AARCH64 LOONGARCH64
 #
 
 [Sources]
@@ -36,6 +36,9 @@
 [Sources.AARCH64]
   AArch64/DynamicCookieGcc.S | GCC
 
+[Sources.LOONGARCH64]
+  LoongArch64/DynamicCookieGcc.S | GCC
+
 [Packages]
   MdePkg/MdePkg.dec
 
@@ -43,3 +46,6 @@
   BaseLib
   DebugLib
   StackCheckLib
+
+[LibraryClasses.LOONGARCH64]
+  RngLib
--- a/MdePkg/Library/DynamicStackCookieEntryPointLib/LoongArch64/DynamicCookieGcc.S
+++ b/MdePkg/Library/DynamicStackCookieEntryPointLib/LoongArch64/DynamicCookieGcc.S
@@ -1,0 +1,71 @@
+#------------------------------------------------------------------------------
+#
+# Copyright (c) Microsoft Corporation.
+# Copyright (c) 2026, Loongson Technology Corporation Limited. All rights reserved.<BR>
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+# Module Name:
+#
+#   DynamicCookieGcc.S
+#
+# Abstract:
+#
+#   Generates random number through the RdTime instruction on a 64-bit LOONGARCH64 platform
+#   to store a random value in the GCC __stack_check_guard stack cookie.
+#   The first byte is 0'd to prevent string copy functions from clobbering
+#   the stack cookie.
+#
+# Notes:
+#
+# Use the RngLib library to get random numbers. LoongArch64 currently does not support like
+# `RdRand` or `RNDR` instructions, it can use the `RdTime` and a xorshift64 algorithms to
+# generate pseudo-random numbers.
+#
+#------------------------------------------------------------------------------
+
+#include <Library/RngLib.h>
+
+ASM_GLOBAL ASM_PFX (_ModuleEntryPoint)
+
+#------------------------------------------------------------------------------
+# VOID
+# EFIAPI
+# _ModuleEntryPoint (
+#   The parameters are saved to the stack when it calls other functions.
+# )
+#------------------------------------------------------------------------------
+ASM_PFX(_ModuleEntryPoint):
+  addi.d    $sp, $sp, -0x20
+  st.d      $a0, $sp, 0x0
+  st.d      $a1, $sp, 0x8
+  st.d      $ra, $sp, 0x10
+
+  //
+  // Use the stack to pass parameters.
+  //
+  addi.d    $a0, $sp, 0x18
+  bl        GetRandomNumber64
+
+  //
+  // Load the random number from the stack.
+  //
+  ld.d      $t0, $sp, 0x18
+
+  //
+  // Zero the first byte of the random value
+  //
+  bstrins.d $t0, $zero, 7, 0
+
+  la.local  $t1, __stack_chk_guard
+  st.d      $t0, $t1, 0
+  dbar      0
+
+  ld.d      $a0, $sp, 0x0
+  ld.d      $a1, $sp, 0x8
+  ld.d      $ra, $sp, 0x10
+  addi.d    $sp, $sp, 0x20
+
+c_entry:
+  b         _CModuleEntryPoint     // Jump to the C module entry point
+
+.end
--- a/MdePkg/Library/StackCheckLib/LoongArch64/StackCookieInterrupt.S
+++ b/MdePkg/Library/StackCheckLib/LoongArch64/StackCookieInterrupt.S
@@ -1,0 +1,21 @@
+#------------------------------------------------------------------------------
+#
+# LoongArch64/StackCookieInterrupt.S
+#
+# Copyright (c) 2026, Loongson Technology Corporation Limited. All rights reserved.<BR>
+#
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+#------------------------------------------------------------------------------
+
+#include <Library/BaseLib.h>
+ASM_GLOBAL ASM_PFX(TriggerStackCookieInterrupt)
+
+#/**
+#  Calling `CpuBreakpoint` to create a breakpoint to the LoongArch64 exception handle and stop.
+#**/
+
+ASM_PFX(TriggerStackCookieInterrupt):
+  bl    CpuBreakpoint
+  jirl  $zero, $ra, 0
+.end
--- a/BaseTools/Conf/tools_def.template
+++ b/BaseTools/Conf/tools_def.template
@@ -586,10 +586,10 @@ NOOPT_*_*_OBJCOPY_ADDDEBUGFLAG     = --add-gnu-debuglink="$(DEBUG_DIR)/$(MODULE_
 *_*_*_DTCPP_PATH                   = DEF(DTCPP_BIN)
 *_*_*_DTC_PATH                     = DEF(DTC_BIN)
 
-# All supported GCC archs except LOONGARCH64 support -mstack-protector-guard=global, so set that on everything except LOONGARCH64
+# Because LOONGARCH64 supports global stack protector guard by default, it does not support the parameter -mstack-protector-guard=global, and this parameter will be supported in GCC17. So set that on every thing except LOONGARCH64.
 DEFINE GCC_ALL_CC_COMMON               = -g -Os -fshort-wchar -fno-builtin -fno-strict-aliasing -Wall -Werror -Wno-array-bounds -include AutoGen.h -fno-common -fstack-protector
 DEFINE GCC_IA32_X64_CC_FLAGS           = -mstack-protector-guard=global
-DEFINE GCC_LOONGARCH64_CC_FLAGS        = DEF(GCC_ALL_CC_COMMON) -mabi=lp64d -fno-asynchronous-unwind-tables -Wno-address -fno-short-enums -fsigned-char -ffunction-sections -fdata-sections -march=loongarch64 -mno-memcpy -Werror -Wno-maybe-uninitialized -Wno-stringop-overflow -Wno-pointer-to-int-cast -no-pie -fno-stack-protector -mno-explicit-relocs -mno-relax
+DEFINE GCC_LOONGARCH64_CC_FLAGS        = DEF(GCC_ALL_CC_COMMON) -mabi=lp64d -fno-asynchronous-unwind-tables -Wno-address -fno-short-enums -fsigned-char -ffunction-sections -fdata-sections -march=loongarch64 -mno-memcpy -Werror -Wno-maybe-uninitialized -Wno-stringop-overflow -Wno-pointer-to-int-cast -no-pie -mno-explicit-relocs -mno-relax
 DEFINE GCC_AARCH64_CC_COMMON           = DEF(GCC_ALL_CC_COMMON) -mlittle-endian -fno-short-enums -fverbose-asm -funsigned-char  -ffunction-sections -fdata-sections -Wno-address -fno-asynchronous-unwind-tables -fno-unwind-tables -fno-pic -fno-pie -ffixed-x18 -mstack-protector-guard=global
 DEFINE GCC_AARCH64_CC_XIPCOMMON        = -mstrict-align -mgeneral-regs-only
 DEFINE GCC_RISCV64_CC_XIPFLAGS         = -mstrict-align -mgeneral-regs-only
--- a/MdePkg/Library/DynamicStackCookieEntryPointLib/StandaloneMmDriverEntryPoint.inf
+++ b/MdePkg/Library/DynamicStackCookieEntryPointLib/StandaloneMmDriverEntryPoint.inf
@@ -23,7 +23,7 @@
 #
 # The following information is for reference only and not required by the build tools.
 #
-#  VALID_ARCHITECTURES           = X64 AARCH64
+#  VALID_ARCHITECTURES           = X64 AARCH64 LOONGARCH64
 #
 
 [Sources]
@@ -40,6 +40,9 @@
 [Sources.AARCH64]
   AArch64/DynamicCookieGcc.S | GCC
 
+[Sources.LOONGARCH64]
+  LoongArch64/DynamicCookieGcc.S | GCC
+
 [Packages]
   MdePkg/MdePkg.dec
 
@@ -48,6 +51,9 @@
   DebugLib
   MmServicesTableLib
   StackCheckLib
+
+[LibraryClasses.LOONGARCH64]
+  RngLib
 
 [Protocols]
   gEfiLoadedImageProtocolGuid      ## SOMETIMES_CONSUMES
--- a/MdePkg/Library/DynamicStackCookieEntryPointLib/UefiApplicationEntryPoint.inf
+++ b/MdePkg/Library/DynamicStackCookieEntryPointLib/UefiApplicationEntryPoint.inf
@@ -18,7 +18,7 @@
   LIBRARY_CLASS                  = UefiApplicationEntryPoint|UEFI_APPLICATION
 
 #
-#  VALID_ARCHITECTURES           = IA32 X64 AARCH64
+#  VALID_ARCHITECTURES           = IA32 X64 AARCH64 LOONGARCH64
 #
 
 [Sources]
@@ -35,6 +35,9 @@
 [Sources.AARCH64]
   AArch64/DynamicCookieGcc.S | GCC
 
+[Sources.LOONGARCH64]
+  LoongArch64/DynamicCookieGcc.S | GCC
+
 [Packages]
   MdePkg/MdePkg.dec
 
@@ -43,3 +46,6 @@
   DebugLib
   BaseLib
   StackCheckLib
+
+[LibraryClasses.LOONGARCH64]
+  RngLib
--- a/MdePkg/Library/DynamicStackCookieEntryPointLib/UefiDriverEntryPoint.inf
+++ b/MdePkg/Library/DynamicStackCookieEntryPointLib/UefiDriverEntryPoint.inf
@@ -20,7 +20,7 @@
 #
 # The following information is for reference only and not required by the build tools.
 #
-#  VALID_ARCHITECTURES           = IA32 X64 AARCH64
+#  VALID_ARCHITECTURES           = IA32 X64 AARCH64 LOONGARCH64
 #
 
 [Sources]
@@ -40,11 +40,17 @@
 [Sources.AARCH64]
   AArch64/DynamicCookieGcc.S | GCC
 
+[Sources.LOONGARCH64]
+  LoongArch64/DynamicCookieGcc.S | GCC
+
 [LibraryClasses]
   UefiBootServicesTableLib
   DebugLib
   BaseLib
   StackCheckLib
+
+[LibraryClasses.LOONGARCH64]
+  RngLib
 
 [Protocols]
   gEfiLoadedImageProtocolGuid                   ## SOMETIMES_CONSUMES
--- a/MdePkg/Library/StackCheckLib/StackCheckLib.inf
+++ b/MdePkg/Library/StackCheckLib/StackCheckLib.inf
@@ -30,6 +30,9 @@
   AArch64/StackCookieInterrupt.S   |GCC
   AArch64/StackCookieInterrupt.asm |MSFT
 
+[Sources.LOONGARCH64]
+  LoongArch64/StackCookieInterrupt.S | GCC
+
 [Packages]
   MdePkg/MdePkg.dec
 


MdePkg/Library/DynamicStackCookieEntryPointLib/DxeCoreEntryPoint.inf
```diff
@@ -36,10 +36,16 @@
...
 [Packages]
   MdePkg/MdePkg.dec
 
 [LibraryClasses]
   BaseLib
   DebugLib
   StackCheckLib
+
+[LibraryClasses.LOONGARCH64]
+  RngLib
```
os-d, line 51: Is it safe to link RngLib to every module? It ha s no other dependencies?

StackCheckLib is moving to a model where there are no dependencies, I'll go further in another comment.

os-d (commented):
StackCheckLib is moving to a different model that will have no dependencies in this PR: https://github.com/tianocore/edk2/pull/12182. The review has stalled and I have not kicked it because I have other higher priority PRs that I'm asking that set of reviewers to prioritize, but as soon as those are in (I'm hoping this week), I'll return to this.

Can this change wait for that PR? Or can you test on top of that PR (it also needs this BaseTools change: https://github.com/tianocore/edk2/pull/12373). I don't have a way to test LoongArch64, so ideally we can refactor StackCheckLib and then have you come in after that, testing with the new design.

At the very least, can we remove RngLib as a dependency? That does not work with the new model of having no dependencies and dropping the LTO disabling.

kilaterlee:
> StackCheckLib is moving to a different model that will have no dependencies in this PR: #12182. The review has stalled and I have not kicked it because I have other higher priority PRs that I'm asking that set of reviewers to prioritize, but as soon as those are in (I'm hoping this week), I'll return to this.
> 
> Can this change wait for that PR? Or can you test on top of that PR (it also needs this BaseTools change: #12373). I don't have a way to test LoongArch64, so ideally we can refactor StackCheckLib and then have you come in after that, testing with the new design.
> 
> At the very least, can we remove RngLib as a dependency? That does not work with the new model of having no dependencies and dropping the LTO disabling.

Hmm... let's me clarify: do you mean that the StackCheckLib shouldn't depend anything? In other words, it should like be a `leaf module` that doesn't call any other modules or functions?

os-d:
> > StackCheckLib is moving to a different model that will have no dependencies in this PR: #12182. The review has stalled and I have not kicked it because I have other higher priority PRs that I'm asking that set of reviewers to prioritize, but as soon as those are in (I'm hoping this week), I'll return to this.
> > Can this change wait for that PR? Or can you test on top of that PR (it also needs this BaseTools change: #12373). I don't have a way to test LoongArch64, so ideally we can refactor StackCheckLib and then have you come in after that, testing with the new design.
> > At the very least, can we remove RngLib as a dependency? That does not work with the new model of having no dependencies and dropping the LTO disabling.
> 
> Hmm... let's me clarify: do you mean that the StackCheckLib shouldn't depend anything? In other words, it should like be a `leaf module` that doesn't call any other modules or functions?

Correct. But I just looked back at the code and I see you did not add RngLib for StackCheckLib itself, only the dynamic entry points, so that does meet the goal. The dynamic entry points can have dependencies, though again I’ll ask: is it safe to have RngLib as a dependency for every module? I.e. you have it on the DxeCore Entry Point lib, does RngLib have any library constructors (at least in planned instances, it can’t be guaranteed)? Does it call out to any other functionality? If either of those is true or could be true, it can’t be used in all entry points. Is it possible to reproduce the RngLib functionality in the shared assembly? Or is it too platform specific? It may also be that LoongArch64 doesn’t use dynamic stack cookies if it can’t generate random numbers easily.

kilaterlee:
> > > StackCheckLib is moving to a different model that will have no dependencies in this PR: #12182. The review has stalled and I have not kicked it because I have other higher priority PRs that I'm asking that set of reviewers to prioritize, but as soon as those are in (I'm hoping this week), I'll return to this.
> > > Can this change wait for that PR? Or can you test on top of that PR (it also needs this BaseTools change: #12373). I don't have a way to test LoongArch64, so ideally we can refactor StackCheckLib and then have you come in after that, testing with the new design.
> > > At the very least, can we remove RngLib as a dependency? That does not work with the new model of having no dependencies and dropping the LTO disabling.
> > 
> > 
> > Hmm... let's me clarify: do you mean that the StackCheckLib shouldn't depend anything? In other words, it should like be a `leaf module` that doesn't call any other modules or functions?
> 
> Correct. But I just looked back at the code and I see you did not add RngLib for StackCheckLib itself, only the dynamic entry points, so that does meet the goal. The dynamic entry points can have dependencies, though again I’ll ask: is it safe to have RngLib as a dependency for every module? I.e. you have it on the DxeCore Entry Point lib, does RngLib have any library constructors (at least in planned instances, it can’t be guaranteed)? Does it call out to any other functionality? If either of those is true or could be true, it can’t be used in all entry points. Is it possible to reproduce the RngLib functionality in the shared assembly? Or is it too platform specific? It may also be that LoongArch64 doesn’t use dynamic stack cookies if it can’t generate random numbers easily.

Yes, I simply called the RngLib function once at the EntryPoint, and let's me talk about that why me imported the `RngLib`:
Since LoongArch SPEC doesn't `TRNG` instructions, but it has a SE(Security Engine,  commonly found in Loongson CPUs) can  generate the true random numbers, and this process is not implemented via an instruction. Currently, we consider the SE is not a part of the ISA, so I just called the RngLib to created the `PRN` at this moment.
A other reason is that I plan to add the `SecurityEngineLib` in the `UefiCpuPkg/Library` in the next step, the RngLib can  consumed it, thereby avoiding code duplication, I believe a unified entry point for random number generation is more appropriate. The code hierarchy is roughly as follows:
```
Caller ->
  RngLib
   if (SE)
      ->SELib
   else
     PRNG
```

As mentioned above,`RngLib` may call out to other libraries. In my mind, all of the libraries within these modules are statically linked, so they are not truly call out. Do you consider this way is not safe?

os-d:
> Yes, I simply called the RngLib function once at the EntryPoint, and let's me talk about that why me imported the `RngLib`: Since LoongArch SPEC doesn't `TRNG` instructions, but it has a SE(Security Engine, commonly found in Loongson CPUs) can generate the true random numbers, and this process is not implemented via an instruction. Currently, we consider the SE is not a part of the ISA, so I just called the RngLib to created the `PRN` at this moment. A other reason is that I plan to add the `SecurityEngineLib` in the `UefiCpuPkg/Library` in the next step, the RngLib can consumed it, thereby avoiding code duplication, I believe a unified entry point for random number generation is more appropriate. The code hierarchy is roughly as follows:
> 
> ```
> Caller ->
>   RngLib
>    if (SE)
>       ->SELib
>    else
>      PRNG
> ```
> 

Sure and sharing the code is reasonable. My concern is only around dependencies and particular for the core entry points.

> As mentioned above,`RngLib` may call out to other libraries. In my mind, all of the libraries within these modules are statically linked, so they are not truly call out. Do you consider this way is not safe?

The concern is if RngLib has a constructor, it won't be called on DXE Core entry. If RngLib used a protocol, it would be broken on DXE Core entry. If RngLib has any assumptions it is only being used after core services are available, it would be broken.

I took a look at BaseRngLib, which I assume is what you assume will be used? That looks okay to use for LoongArch64, there is a constructor, which won't get called until later, but it just returns EFI_SUCCESS in the LoongArch64 case.

However, a platform can use any instance and from the core side, we can't control that. But I'm okay to chalk that up to platform misconfiguration. But this is a dangerous area, a lot of assumptions don't hold here, so I wanted to follow up on it.

All that being said, I don't have any intersection with LoongArch64 platforms, so as long as you deem this appropriate and the library stays scoped to LoongArch64 (as it already is), I'm okay with this going in.

I'm also fine for this to go in first before the CLANGPDB support/refactor PR goes in, but I will reach out for you to test LoongArch64 at that time.

os-d (approved):

kilaterlee:
> > Yes, I simply called the RngLib function once at the EntryPoint, and let's me talk about that why me imported the `RngLib`: Since LoongArch SPEC doesn't `TRNG` instructions, but it has a SE(Security Engine, commonly found in Loongson CPUs) can generate the true random numbers, and this process is not implemented via an instruction. Currently, we consider the SE is not a part of the ISA, so I just called the RngLib to created the `PRN` at this moment. A other reason is that I plan to add the `SecurityEngineLib` in the `UefiCpuPkg/Library` in the next step, the RngLib can consumed it, thereby avoiding code duplication, I believe a unified entry point for random number generation is more appropriate. The code hierarchy is roughly as follows:
> > ```
> > Caller ->
> >   RngLib
> >    if (SE)
> >       ->SELib
> >    else
> >      PRNG
> > ```
> 
> Sure and sharing the code is reasonable. My concern is only around dependencies and particular for the core entry points.
> 
> > As mentioned above,`RngLib` may call out to other libraries. In my mind, all of the libraries within these modules are statically linked, so they are not truly call out. Do you consider this way is not safe?
> 
> The concern is if RngLib has a constructor, it won't be called on DXE Core entry. If RngLib used a protocol, it would be broken on DXE Core entry. If RngLib has any assumptions it is only being used after core services are available, it would be broken.
> 
> I took a look at BaseRngLib, which I assume is what you assume will be used? That looks okay to use for LoongArch64, there is a constructor, which won't get called until later, but it just returns EFI_SUCCESS in the LoongArch64 case.
> 
> However, a platform can use any instance and from the core side, we can't control that. But I'm okay to chalk that up to platform misconfiguration. But this is a dangerous area, a lot of assumptions don't hold here, so I wanted to follow up on it.
> 
> All that being said, I don't have any intersection with LoongArch64 platforms, so as long as you deem this appropriate and the library stays scoped to LoongArch64 (as it already is), I'm okay with this going in.
> 
> I'm also fine for this to go in first before the CLANGPDB support/refactor PR goes in, but I will reach out for you to test LoongArch64 at that time.

I seemingly see your point: you are concerned that if the constructor in `BaseRngLib` executes other operations—such as making decisions based on certain features—those operations might not execute before the entry point. You are right, I hadn't think about that. If the RngLib relies on constructor to verify whether the SE is alive, it can not be used at this stage.

os-d:
> I seemingly see your point: you are concerned that if the constructor in `BaseRngLib` executes other operations—such as making decisions based on certain features—those operations might not execute before the entry point. You are right, I hadn't think about that. If the RngLib relies on constructor to verify whether the SE is alive, it can not be used at this stage.

Exactly. Or that another instance of RngLib is chosen that calls out to another library that expects a constructor to run or attempts to use a dynamic PCD or protocol, etc.

So, again, at this stage BaseRngLib is safe to use for LoongArch64 and I will defer to you on whether the risk that a platform switches to using a different RngLib can just be chalked up to platform misconfiguration and if your future plans for RNG align with using RngLib.

kilaterlee:
> > I seemingly see your point: you are concerned that if the constructor in `BaseRngLib` executes other operations—such as making decisions based on certain features—those operations might not execute before the entry point. You are right, I hadn't think about that. If the RngLib relies on constructor to verify whether the SE is alive, it can not be used at this stage.
> 
> Exactly. Or that another instance of RngLib is chosen that calls out to another library that expects a constructor to run or attempts to use a dynamic PCD or protocol, etc.
> 
> So, again, at this stage BaseRngLib is safe to use for LoongArch64 and I will defer to you on whether the risk that a platform switches to using a different RngLib can just be chalked up to platform misconfiguration and if your future plans for RNG align with using RngLib.

So, this is a hard choise for me:
a. Completely give up the constructor in BaseRngLib.
b. Copy a same `PRNG` code from BaseRngLib in EntryPoint library.
c. Maybe another RngLib instance(extremely low probability).
I don't know which way is the best, maybe option b? But I don't like it...

os-d:
> > > I seemingly see your point: you are concerned that if the constructor in `BaseRngLib` executes other operations—such as making decisions based on certain features—those operations might not execute before the entry point. You are right, I hadn't think about that. If the RngLib relies on constructor to verify whether the SE is alive, it can not be used at this stage.
> > 
> > 
> > Exactly. Or that another instance of RngLib is chosen that calls out to another library that expects a constructor to run or attempts to use a dynamic PCD or protocol, etc.
> > So, again, at this stage BaseRngLib is safe to use for LoongArch64 and I will defer to you on whether the risk that a platform switches to using a different RngLib can just be chalked up to platform misconfiguration and if your future plans for RNG align with using RngLib.
> 
> So, this is a hard choise for me: a. Completely give up the constructor in BaseRngLib. b. Copy a same `PRNG` code from BaseRngLib in EntryPoint library. c. Maybe another RngLib instance(extremely low probability). I don't know which way is the best, maybe option b? But I don't like it...

AARCH64 has shared assembly macros in MdePkg, depending on the length and complexity of the PRNG code, that could be an option: https://github.com/tianocore/edk2/blob/3a1cfddd9a12bcdbbbcd0ff98c87a8747d08e433/MdePkg/Include/AArch64/AsmMacroLib.h#L17.

I.e. the asm goes in a C macro in MdePkg, the RngLib constructor and dynamic stack cookie entry point both use it.

That has its own ugliness, but could be workable.

Otherwise the code duplication is probably the best alternative.

Another option is not having a dynamic stack cookie for the core entry points. For other components, it should be fine, the constructor will get called first. The presence of a static stack cookie is still better than none.

There is another option that would change how the dynamic stack cookie is updated for LoongArch64 only, but that one I would suggest we don't do because it seems brittle and puts it different than other archs, which is only likely to lead to breakage.

kilaterlee:
> AARCH64 has shared assembly macros in MdePkg, depending on the length and complexity of the PRNG code, that could be an option:
> 
> https://github.com/tianocore/edk2/blob/3a1cfddd9a12bcdbbbcd0ff98c87a8747d08e433/MdePkg/Include/AArch64/AsmMacroLib.h#L17
> 
> .
> I.e. the asm goes in a C macro in MdePkg, the RngLib constructor and dynamic stack cookie entry point both use it.
> 
> That has its own ugliness, but could be workable.
> 
> Otherwise the code duplication is probably the best alternative.
> 
> Another option is not having a dynamic stack cookie for the core entry points. For other components, it should be fine, the constructor will get called first. The presence of a static stack cookie is still better than none.
> 
> There is another option that would change how the dynamic stack cookie is updated for LoongArch64 only, but that one I would suggest we don't do because it seems brittle and puts it different than other archs, which is only likely to lead to breakage.

I examined the AARCH64 code, it looks like we need prepare some ASM or other code, and the RngLib constructor or dyamic stack cookie should call it first--such as making decisions logic. However, this still fails to avoid code duplication, as the `ArchGetRandomNumber` family API does not take the extra parameters to declare which `RNG` should be used, that is the constructor's job.

os-d:
> I examined the AARCH64 code, it looks like we need prepare some ASM or other code, and the RngLib constructor or dyamic stack cookie should call it first--such as making decisions logic. However, this still fails to avoid code duplication, as the `ArchGetRandomNumber` family API does not take the extra parameters to declare which `RNG` should be used, that is the constructor's job.

I'm not quite following why code duplication would still occur.

What I would imagine is:

MdePkg/**/LoongArch64.h -> Two macros, one that determines RNG source and the other that gets a random number based on source.

Dynamic stack cookie entry point -> Use macros to determine RNG source and get random number.

RngLib -> Constructor uses macro to determine RNG source and saves it in a global variable. The constructor won't be used for the dynamic stack cookie entry point, but for all other uses and modules it will. The RngLib functions then use the global to determine source and use the macros to get a random number.

That should avoid the duplication, right? I don't like having the asm in macros, but it does solve the code duplication problem.

kilaterlee:
> > I examined the AARCH64 code, it looks like we need prepare some ASM or other code, and the RngLib constructor or dyamic stack cookie should call it first--such as making decisions logic. However, this still fails to avoid code duplication, as the `ArchGetRandomNumber` family API does not take the extra parameters to declare which `RNG` should be used, that is the constructor's job.
> 
> I'm not quite following why code duplication would still occur.
> 
> What I would imagine is:
> 
> MdePkg/**/LoongArch64.h -> Two macros, one that determines RNG source and the other that gets a random number based on source.
> 
> Dynamic stack cookie entry point -> Use macros to determine RNG source and get random number.
> 
> RngLib -> Constructor uses macro to determine RNG source and saves it in a global variable. The constructor won't be used for the dynamic stack cookie entry point, but for all other uses and modules it will. The RngLib functions then use the global to determine source and use the macros to get a random number.
> 
> That should avoid the duplication, right? I don't like having the asm in macros, but it does solve the code duplication problem.

I see, you mean that the dynamic stack cookie and RngLib both call a same API which located some .h file. If this approach is taken, it seems some or all RNG logic would be moved into the .h file, and all of them should completed via ASM macros.
This method certainly solves the code duplication problem, but... it looks like not very comfortable...
And I got another way that all of the options are in the RngLib:
Both the constructor and the `GetRandom` API call an external library to determine the RNG source. This utilizes a global variable with three states: `Not sure`, `TRNG is ready`, and `TRNG is not ready`. The `GetRandom` API reads this global variable: if the state is `Not sure`, it makes its own determination; otherwise, it directly uses the value established by the constructor.

os-d:
> I see, you mean that the dynamic stack cookie and RngLib both call a same API which located some .h file. If this approach is taken, it seems some or all RNG logic would be moved into the .h file, and all of them should completed via ASM macros. This method certainly solves the code duplication problem, but... it looks like not very comfortable...

Yep, I don’t love it, but there is prior “art” here. It’s really the trade off being code duplication or nicer interfaces. Again, you may also decide that dynamic stack cookies for DXE core don’t work well for LoongArch64 until an RNG instruction is available (I think you said that was on the roadmap).

> And I got another way that all of the options are in the RngLib: Both the constructor and the `GetRandom` API call an external library to determine the RNG source. This utilizes a global variable with three states: `Not sure`, `TRNG is ready`, and `TRNG is not ready`. The `GetRandom` API reads this global variable: if the state is `Not sure`, it makes its own determination; otherwise, it directly uses the value established by the constructor.

You could work that out too. The decision there is whether a new library class is worthwhile for this use case.

kilaterlee:
> You could work that out too. The decision there is whether a new library class is worthwhile for this use case.

Okay, as for whether to use the new library or PCD, we can decide later; perhaps we could use the new SE library(Because we need to know whether the SE or TRNG is alive).
And I think dynamic stack cookie is always a consumer, so the review can proceed, as far the RngLib update, that should be discussed in another PR. :)

kilaterlee:
@MarsDoge @AydenMeng @lgao4 Please review this patch set, THX!:)

lgao4:
I have same comments on RngLib dependency. Have you the plan to resolve it in future?

kilaterlee:
> I have same comments on RngLib dependency. Have you the plan to resolve it in future?

Are you also worried that the constructor might fail to execute in certain situations?

MdePkg/Library/DynamicStackCookieEntryPointLib/LoongArch64/DynamicCookieGcc.S
```diff
@@ -0,0 +1,69 @@
...
+  bl       GetRandomNumber64
+
+  la.pcrel $a0, random_num
+  ld.d     $t0, $a0, 0x0
+
+  //
+  // Zero the first byte of the random value
+  //
+  srli.d   $t0, $t0, 8
+  slli.d   $t0, $t0, 8
```
AydenMeng: I suggest using `bstrins.d $t0, $zero, 8, 0`

MdePkg/Library/DynamicStackCookieEntryPointLib/LoongArch64/DynamicCookieGcc.S
```diff
@@ -0,0 +1,69 @@
...
+
+#------------------------------------------------------------------------------
+# VOID
+# EFIAPI
+# _ModuleEntryPoint (
+#   The parameters are saved to the stack when it calls other functions.
+# )
+#------------------------------------------------------------------------------
+ASM_PFX(_ModuleEntryPoint):
+  addi.d   $sp, $sp, -0x20
```
AydenMeng: To be precise, we can only use 0x18 bytes below sp, right?

MdePkg/Library/StackCheckLib/LoongArch64/StackCookieInterrupt.S
```diff
@@ -0,0 +1,21 @@
...
+ASM_GLOBAL ASM_PFX(TriggerStackCookieInterrupt)
+
+#/**
+#  Calling `CpuBreakpoint` to create a breakpoint to the LoongArch64 exception handle and stop.
+#**/
+
+ASM_PFX(TriggerStackCookieInterrupt):
+  bl    CpuBreakpoint
+  jirl  $zero, $ra, 0
+  .end
```
AydenMeng: There might be a wrong tab here.

MdePkg/Library/DynamicStackCookieEntryPointLib/LoongArch64/DynamicCookieGcc.S
kilaterlee: We stack should be 16-byte aligned.

MdePkg/Library/DynamicStackCookieEntryPointLib/LoongArch64/DynamicCookieGcc.S
kilaterlee: Okay.

MdePkg/Library/StackCheckLib/LoongArch64/StackCookieInterrupt.S
kilaterlee: Yes.

kilaterlee:
> > I have same comments on RngLib dependency. Have you the plan to resolve it in future?
> 
> Are you also worried that the constructor might fail to execute in certain situations?

@lgao4 Hi Liming, ping again. Do you worried that the constructor might fail?

MarsDoge:
I suggest using a stack scratch slot here instead of the `random_num` label.

random_num is currently placed in the text section, so
GetRandomNumber64(&random_num) writes into the code section. This may break
when image/code section protection marks .text as read-only/executable.

The random value is only temporary, so using the current stack frame should be
cleaner than adding a .data/.bss object.

For example:
```asm
  addi.d   $a0, $sp, 0x18
  bl       GetRandomNumber64
  beqz     $a0, restore_and_c_entry
  ld.d     $t0, $sp, 0x18
...
```
Remove the following operations
```asm

  la.pcrel $a0, random_num
  ...
random_num:
  .dword 0x0
```

kilaterlee:
> I suggest using a stack scratch slot here instead of the `random_num` label.
> 
> random_num is currently placed in the text section, so GetRandomNumber64(&random_num) writes into the code section. This may break when image/code section protection marks .text as read-only/executable.
> 
> The random value is only temporary, so using the current stack frame should be cleaner than adding a .data/.bss object.
> 
> For example:
> 
> ```assembly
>   addi.d   $a0, $sp, 0x18
>   bl       GetRandomNumber64
>   beqz     $a0, restore_and_c_entry
>   ld.d     $t0, $sp, 0x18
> ...
> ```
> 
> Remove the following operations
> 
> ```assembly
>   la.pcrel $a0, random_num
>   ...
> random_num:
>   .dword 0x0
> ```

That's a good idea, lets me try.

lgao4:
> > I suggest using a stack scratch slot here instead of the `random_num` label.
> > random_num is currently placed in the text section, so GetRandomNumber64(&random_num) writes into the code section. This may break when image/code section protection marks .text as read-only/executable.
> > The random value is only temporary, so using the current stack frame should be cleaner than adding a .data/.bss object.
> > For example:
> > ```assembly
> >   addi.d   $a0, $sp, 0x18
> >   bl       GetRandomNumber64
> >   beqz     $a0, restore_and_c_entry
> >   ld.d     $t0, $sp, 0x18
> > ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Remove the following operations
> > ```assembly
> >   la.pcrel $a0, random_num
> >   ...
> > random_num:
> >   .dword 0x0
> > ```
> 
> That's a good idea, lets me try.

This solution doesn't need to depend on RngLib. Right?

lgao4:
> > I suggest using a stack scratch slot here instead of the `random_num` label.
> > random_num is currently placed in the text section, so GetRandomNumber64(&random_num) writes into the code section. This may break when image/code section protection marks .text as read-only/executable.
> > The random value is only temporary, so using the current stack frame should be cleaner than adding a .data/.bss object.
> > For example:
> > ```assembly
> >   addi.d   $a0, $sp, 0x18
> >   bl       GetRandomNumber64
> >   beqz     $a0, restore_and_c_entry
> >   ld.d     $t0, $sp, 0x18
> > ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Remove the following operations
> > ```assembly
> >   la.pcrel $a0, random_num
> >   ...
> > random_num:
> >   .dword 0x0
> > ```
> 
> That's a good idea, lets me try.

This solution doesn't need to depend on RngLib. Right?

kilaterlee:
> > > I suggest using a stack scratch slot here instead of the `random_num` label.
> > > random_num is currently placed in the text section, so GetRandomNumber64(&random_num) writes into the code section. This may break when image/code section protection marks .text as read-only/executable.
> > > The random value is only temporary, so using the current stack frame should be cleaner than adding a .data/.bss object.
> > > For example:
> > > ```assembly
> > >   addi.d   $a0, $sp, 0x18
> > >   bl       GetRandomNumber64
> > >   beqz     $a0, restore_and_c_entry
> > >   ld.d     $t0, $sp, 0x18
> > > ...
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Remove the following operations
> > > ```assembly
> > >   la.pcrel $a0, random_num
> > >   ...
> > > random_num:
> > >   .dword 0x0
> > > ```
> > 
> > 
> > That's a good idea, lets me try.
> 
> This solution doesn't need to depend on RngLib. Right?

No, it simply removes `random_num` and uses a stack instead.
So, regarding the dependency on RngLib, are you worried about security issues or the inability to execute the constructor?

MarsDoge (commented):
Hi, @lgao4 sorry for the confusion. Let me clarify: 
The point of my comment  [#issuecomment-4851580637](https://github.com/tianocore/edk2/pull/12677/#issuecomment-4851580637) was about where the temporary random value is stored, the GetRandomNumber64(&random_num) writes into the image text section. That may break once code/image section protection makes .text read-only/executable.

The RngLib dependency / constructor-safety discussion is a separate topic from this storage-location fix.

lgao4:
> Hi, @lgao4 sorry for the confusion. Let me clarify: The point of my comment [#issuecomment-4851580637](https://github.com/tianocore/edk2/pull/12677/#issuecomment-4851580637) was about where the temporary random value is stored, the GetRandomNumber64(&random_num) writes into the image text section. That may break once code/image section protection makes .text read-only/executable.
> 
> The RngLib dependency / constructor-safety discussion is a separate topic from this storage-location fix.

OK. I see.

lgao4:
I have no other comment.

MdePkg/Library/DynamicStackCookieEntryPointLib/LoongArch64/DynamicCookieGcc.S
```diff
@@ -0,0 +1,71 @@
...
+#
+# Notes:
+#
+# Use the RngLib library to get random numbers. LoongArch64 currently does not support like
+# `RdRand` or `RNDR` instructions, it can use the `RdTime` and a xorshift64 algorithms to
+# generate pseudo-random numbers.
+#
+#------------------------------------------------------------------------------
+
+#inlcude <Library/RngLib.h>
```
MarsDoge: typo issue

MdePkg/Library/DynamicStackCookieEntryPointLib/LoongArch64/DynamicCookieGcc.S
```diff
@@ -0,0 +1,71 @@
...
+
+  //
+  // Load the random number from the stack.
+  //
+  ld.d      $t0, $sp, 0x18
+
+  //
+  // Zero the first byte of the random value
+  //
+  bstrins.d $t0, $zero, 8, 0
```
MarsDoge: ` bstrins.d $t0, $zero, 8, 0` clears bits 0..8. 
If the intent is to zero the first byte, this should be ` bstrins.d $t0, $zero, 7, 0 `.

MdePkg/Library/DynamicStackCookieEntryPointLib/LoongArch64/DynamicCookieGcc.S
```diff
@@ -0,0 +1,71 @@
...
+  addi.d    $sp, $sp, -0x20
+  st.d      $a0, $sp, 0x0
+  st.d      $a1, $sp, 0x8
+  st.d      $ra, $sp, 0x10
+
+  //
+  // Use the stack to pass parameters.
+  //
+  addi.d    $a0, $sp, 0x18
+  bl        GetRandomNumber64
```
MarsDoge, line 47: One remaining issue: the return value from GetRandomNumber64() should be checked.
Follow the existing IA32/X64/AARCH64 behavior: if RNG generation fails, keep the build-time/static stack cookie and continue to _CModuleEntryPoint.

suggestion:
```asm
bl        GetRandomNumber64
beqz      $a0, restore_and_c_entry
...
restore_and_c_entry:
  ld.d      $a0, $sp, 0x0
  ld.d      $a1, $sp, 0x8
  ld.d      $ra, $sp, 0x10
  addi.d    $sp, $sp, 0x20
...

```

MdePkg/Library/DynamicStackCookieEntryPointLib/LoongArch64/DynamicCookieGcc.S
kilaterlee: done

MdePkg/Library/DynamicStackCookieEntryPointLib/LoongArch64/DynamicCookieGcc.S
kilaterlee: done

MdePkg/Library/DynamicStackCookieEntryPointLib/LoongArch64/DynamicCookieGcc.S
kilaterlee: Since both RngLibTimer and the RngLib will return `false` if the parameter is `NULL`, and the parameter here is guaranteed not to be `NULL`—I believe there is no need to check the return value.

MarsDoge (approved):
LGTM overall!

PR #12681, opened 2026-06-16
CryptoPkg: Match OpenSSL's default security level

yadutaf:
# Description

`TlsNew()` explicitly sets the default security level to 3. The current default in OpenSSL is security level 2 which is inherited by Linux distributions like Ubuntu 26.04. This is also the security level that was announced in https://edk2.groups.io/g/devel/topic/115039926.

- [ ] Breaking change?
- [x] Impacts security? - Lower the default SSL security level to match OpenSSL's / Ubuntu 26.04
- [ ] Includes tests?

## How This Was Tested

This was tested with Qemu on Ubuntu 26.04 by HTTPS-booting https://boot.netboot.xyz/ipxe/netboot.xyz-snponly.efi. This specific URL works with Curl and Firefox Nightly but fails in EDK II without this patch with:

```
TlsDoHandshake X509_verify_result: 66 (EE certificate key too weak)
```

With this patch, the boot succeeds.

## Integration Instructions

N/A

### DIFF
--- a/CryptoPkg/Library/TlsLib/TlsInit.c
+++ b/CryptoPkg/Library/TlsLib/TlsInit.c
@@ -186,11 +186,6 @@ TlsNew (
   }
 
   //
-  // This retains compatibility with previous version of OpenSSL.
-  //
-  SSL_set_security_level (TlsConn->Ssl, 3);
-
-  //
   // Initialize the created SSL Object
   //
   SSL_set_info_callback (TlsConn->Ssl, NULL);


liyi77:
some discussion about security level 3: https://github.com/tianocore/edk2/pull/11331

Recommend to use TlsSetSecurityLevel () if need.

yadutaf:
Hi @liyi77. Thanks for sharing the discussion. I understand a key point was being future-proof with respect to security while providing a dedicated API for integrators.

A limitation, however, is that, by raising the minimum above the current industry-standard, this is raising the bar for the users, indeed users can not or hardly patch their BIOSes, including the OVMF build for Qemu and provided by the distributions. This might discourage even more the use of HTTPS and push back users to the low friction clear text HTTP.

Another approach could be to remove this line of code entirely so that the default would transparently align with upstream's OpenSSL as the embedded version is updated.

liyi77:
I understand your point.
I prefer this approach to align with OpenSSL upstream.
> Another approach could be to remove this line of code entirely so that the default would transparently align with upstream's OpenSSL as the embedded version is updated.

yadutaf:
@liyi77 Thanks for your feedback. I have rebased and updated the PR accordingly.

liyi77 (approved):

yadutaf:
@liyi77 It looks like the merge failed. Should I change anything on the PR ?

liyi77:
> @liyi77 It looks like the merge failed. Should I change anything on the PR ?

seems not a real code issue, please try to rebase pr to master

yadutaf:
> seems not a real code issue, please try to rebase pr to master

Sure. Done.

liyi77:
@mergify queue

PR #12682, opened 2026-06-16
ArmVirtPkg: allow runtime control of serial debug log level via fw_cfg

luigix25:
ArmVirtPkg: introduce compile and runtime control of serial debug log level

Add a new PCD `PcdSerialDebugPrintErrorLevel` that overrides the verbosity
set in `DebugPrintErrorLevel` for the serial port debug output and does
not affect memory debug logging.

Accepted values are:
- "silent": DEBUG_ERROR only
- "verbose": use the `DebugPrintErrorLevel` value.
- a hex bitmask (e.g. "0x80000040").

This PCD value can be overridden at runtime using the fw_cfg entry
"opt/org.tianocore/DebugLevel" without rebuilding the firmware.

The runtime override takes priority over the compile-time one.  When
neither is set, the serial output uses the compiled-in
PcdDebugPrintErrorLevel.

SEC and PEI_CORE phases do not support this override because they run
from flash, where global variables are not available.  Supporting it
would require parsing the device tree on every debug write to locate
the fw_cfg device and read the DebugLevel value — unnecessary overhead
given the low volume of logs in those phases.

ParseSerialDebugLevel() is defined as a STATIC inline function in
DebugLevelParser.h rather than a separate library class, because it
is a small helper shared by only two compilation units (Flash.c and
PlatformPeiLib.c) and adding a full library class would be excessive.

Example QEMU command line:

  -fw_cfg name=opt/org.tianocore/DebugLevel,string=silent
  -fw_cfg name=opt/org.tianocore/DebugLevel,string=verbose
  -fw_cfg name=opt/org.tianocore/DebugLevel,string=0x80000000

Suggested-by: Gerd Hoffmann <kraxel@redhat.com>

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

launch qemu with `-fw_cfg name=opt/org.tianocore/DebugLevel,string=0xdebuglevel`

## Integration Instructions
N/A


luigix25:
> ## Pull Request Formatting Issues
> 
> ⚠️ Remove the following template lines from your PR description: `<_Describe how these changes should be integrated. Use N/A if nothing is required._>`
> 
> Address these issues and the validation will automatically re-run when you update your pull request.

Fixed.

kraxel:
The doc update needs to go into a separate commit to avoid updating both `ArmVirtPkg` and `OvmfPkg` with one patch.

kraxel:
Looks good to me overall.  I'm wondering whenever we should have a separate default value for the serial debug level (`PcdSerialDebugPrintErrorLevel` ?) though.  I'd like to be able to have a firmware build which ...

* logs everything to memory debug log
* logs nothing to the serial line by default
* allows to enable logging to the serial line.

... and I think this is not possible.

luigix25:
> Looks good to me overall. I'm wondering whenever we should have a separate default value for the serial debug level (`PcdSerialDebugPrintErrorLevel` ?) though. I'd like to be able to have a firmware build which ...
> 
>     * logs everything to memory debug log
> 
>     * logs nothing to the serial line by default
> 
>     * allows to enable logging to the serial line.
> 
> 
> ... and I think this is not possible.

Correct. Currently the only option is to set debug log to maximum and then use this to decrease serial log output.

luigix25:
- Documentation moved to a different commit

luigix25:
[@kraxel]
- Added new PCD `PcdSerialDebugPrintErrorLevel`. This sets the debug output level for serial port.
- It accepts 3 possible values:
  - `verbose`: default value. it falls back to `PcdDebugPrintErrorLevel`
  - `silent`: `DEBUG_ERROR`
  - `0xVALUE`
- This PCD value can be overridden using  `opt/org.tianocore/DebugLevel`

Note that the override is not available for SEC and PEI.

luigix25:
CI is failing because I created an header file that contains a function (`ParseSerialDebugLevel`). This is because I need to consume that function from `PeiPlatformLib` and from Flash.c (within `DebugLib`)

I _really_ don't want to create a new library for a 10 line function, and I didn't find a suitable library to reuse. So putting everything in an header seemed a good idea to me but apparently this is not allowed in edk2.

Very open to suggestions

kraxel:
> CI is failing because I created an header file that contains a function (`ParseSerialDebugLevel`). This is because I need to consume that function from `PeiPlatformLib` and from Flash.c (within `DebugLib`)
> 
> I _really_ don't want to create a new library for a 10 line function, and I didn't find a suitable library to reuse. So putting everything in an header seemed a good idea to me but apparently this is not allowed in edk2.
> 
> Very open to suggestions

Hmm, tricky.  Maybe make `PcdSerialDebugPrintErrorLevel` an UINT32 PCD so we do not have to parse much in the first place in the early boot case?  Support the convenience strings `silent` and `verbose` for the fw_cfg override only?

luigix25:
> > CI is failing because I created an header file that contains a function (`ParseSerialDebugLevel`). This is because I need to consume that function from `PeiPlatformLib` and from Flash.c (within `DebugLib`)
> > I _really_ don't want to create a new library for a 10 line function, and I didn't find a suitable library to reuse. So putting everything in an header seemed a good idea to me but apparently this is not allowed in edk2.
> > Very open to suggestions
> 
> Hmm, tricky. Maybe make `PcdSerialDebugPrintErrorLevel` an UINT32 PCD so we do not have to parse much in the first place in the early boot case? Support the convenience strings `silent` and `verbose` for the fw_cfg override only?

mmh then fw_cfg would not be a 1:1 override for the PCD. I don't really like it :/ 
So, do you prefer code duplication or a new library?

kraxel:
> mmh then fw_cfg would not be a 1:1 override for the PCD. I don't really like it :/ So, do you prefer code duplication or a new library?

IMHO code duplication is fine in this case, a new library for 10-20 lines of code is overkill indeed.  But please add a comment in both places pointing to the other copy to reduce the risk that they run out of sync.

luigix25:
[@kraxel]
- Duplicated (sigh) `ParseSerialDebugLevel` function
- Rebased to upstream

kraxel (approved):

ardbiesheuvel:
Does this affect `RELEASE` builds too?

luigix25:
> Does this affect `RELEASE` builds too?

@ardbiesheuvel I don't think so. DebugLib is not used in `RELEASE` builds. 
IOW RELEASE builds don't produce any logs.

My code does actually nothing.

Do you have any concerns?

luigix25:
Ping :)

ardbiesheuvel (approved):

ArmVirtPkg/ArmVirtCloudHv.dsc
```diff
@@ -46,6 +46,8 @@
 
   TpmPlatformHierarchyLib|SecurityPkg/Library/PeiDxeTpmPlatformHierarchyLibNull/PeiDxeTpmPlatformHierarchyLib.inf
   ArmTransferListLib|ArmPkg/Library/ArmTransferListLib/ArmTransferListLib.inf
+  QemuFwCfgLib|OvmfPkg/Library/QemuFwCfgLib/QemuFwCfgLibNull.inf
```
pierregondois, line 49: @luigix25 

I think this should also be done for ArmVirtXen.dsc.
Unfortunately, the `ArmVirtPkg/PlatformCI/.azurepipelines/Ubuntu-XXX.yml` files don't support Xen, so it was not caught by the CI, but I get:

```
build -a AARCH64 -t GCC -p ArmVirtPkg/ArmVirtXen.dsc

[...]

build.py...
edk2/ArmVirtPkg/ArmVirtXen.dsc(...): error 4000: Instance of library class [QemuFwCfgSimpleParserLib] is not found for module [edk2/ArmVirtPkg/PrePi/ArmVirtPrePiUniCoreRelocatable.inf], [QemuFwCfgSimpleParserLib] is:
        consumed by edk2/ArmVirtPkg/Library/PlatformPeiLib/PlatformPeiLib.inf
```

luigix25: oops! I will a PR!

luigix25: Done #12872

PR #12686, opened 2026-06-17
OvmfPkg/LoongArchVirt: Set default terminal type to TTYTERM

EricGao2015:
# Description

Replace the legacy VT100 default with TTYTERM, which is more suitable for modern terminal emulators and command-line utilities.

## How This Was Tested

Add debug message in TerminalDxe driver and check if the change has taken effect.

## Integration Instructions

N/A


kraxel (approved):

kilaterlee (approved):
LGTM

PR #12701, opened 2026-06-19
SecurityPkg: Correct sanity check for File and FileBuffer inputs

### LINKED ISSUE #12678: [Bug]: 057a611ae6 ("SecurityPkg: Added basic DxeImageVerificationLib tests", 2025-06-22) DxeImageVerificationHandler Sanity Check

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

SecurityPkg

NO-TARGET

### Current Behavior

-> In this   [commit](https://github.com/tianocore/edk2/commit/1854eee1b464ccd9cd3aec55141178e10a6dc502)  changes, the File parameter of DxeImageVerificationHandler() was explicitly marked as OPTIONAL, with the commit description stating:

"According to the File description, the File is optional and can be NULL."

→ However, a sanity check was added to DxeImageVerificationHandler() to satisfy the unit test ImageTypeVerifySanity in GoogleTest/DxeImageVerificationLibGoogleTest.cpp, which asserts EXPECT_EQ(Status, EFI_INVALID_PARAMETER) when File = NULL.

Refer: [SecurityPkg: Added basic DxeImageVerificationLib tests · tianocore/edk2@057a611](https://github.com/tianocore/edk2/commit/057a611ae6e3c4a1b2811220576a6b48be349cb8#diff-f2776f21c91fbe2262821aab370f649ed41d8ff5a7451caa1faf24ddf2a3fc12) 

→ The sanity check in DxeImageVerificationHandler() should be updated to align with the logic already used in ExecuteSecurity2Handlers(), which correctly returns EFI_INVALID_PARAMETER only when both File and FileBuffer are NULL.

ExecuteSecurity2Handlers {
           .
		   .
		   .
		   .
		   .
		   //
  // Invalid case if File and FileBuffer are both NULL.
  //
  if ((File == NULL) && (FileBuffer == NULL)) {
    return EFI_INVALID_PARAMETER;
  }
            .
			.
			.
			.
}

### Expected Behavior

DxeImageVerificationHandler should allow once FileBuffer valid, it shouldn't restrict the flow on DxeImageVerificationHandler as File is an Option Parameter

### Steps To Reproduce

Connect the RAID Controller.
Boot into Setup.
Observe that the Storage Option ROMs are not visible.

Due to the sanity check, the image verification process is being restricted even though the FileBuffer is valid. As a result, the Storage OpROMs fail to load during setup.

### Build Environment

```markdown
- OS(s): Windows 11
- Tool Chain(s): VS2019
```

### Version Information

```text
https://github.com/tianocore/edk2/commit/057a611ae6e3c4a1b2811220576a6b48be349cb8#diff-f2776f21c91fbe2262821aab370f649ed41d8ff5a7451caa1faf24ddf2a3fc12
```

### Urgency

High

I will fix it

Maintainer feedback requested

### Anything else?

  if ((File == NULL) && (FileBuffer == NULL)) {
    return EFI_INVALID_PARAMETER;
  }

Sanity check should be like above two parameter check same as caller function.

vignesh032820-Ami:
# Description

Add a sanity check to ensure that either File or FileBuffer is provided before proceeding with security verification.
If both are NULL, return EFI_INVALID_PARAMETER. This change prevents verification from running without a valid input buffer and improves robustness of the check.

- [ ] Breaking change?
  - No – This does not alter build or boot behavior.
- [ ] Impacts security?
  - Yes – Prevents potential misuse of verification routines with invalid inputs.
- [ ] Includes tests?
  - No – No explicit test code included, but verified through functional build and runtime checks.

## How This Was Tested

Built the affected module successfully in the EDK2 environment.

Verified runtime behavior by passing valid and invalid File/FileBuffer combinations.

Confirmed that invalid inputs correctly return EFI_INVALID_PARAMETER without proceeding to verification.

## Integration Instructions

This patch can be merged directly into master.

Fixes #12678

Signed-off-by: Vignesh G <vigneshg@ami.com>

### DIFF
--- a/SecurityPkg/Library/DxeImageVerificationLib/DxeImageVerificationLib.c
+++ b/SecurityPkg/Library/DxeImageVerificationLib/DxeImageVerificationLib.c
@@ -1705,9 +1705,13 @@ DxeImageVerificationHandler (
   IsFoundInDatabase = FALSE;
 
   //
-  // Sanity check
+  // Sanity check:
+  // Ensure that either File or FileBuffer is provided.
+  // Return EFI_INVALID_PARAMETER if both are NULL.
+  // This prevents security verification from proceeding
+  // when no valid input buffer is available.
   //
-  if (File == NULL) {
+  if ((File == NULL) && (FileBuffer == NULL)) {
     return EFI_INVALID_PARAMETER;
   }
 


SecurityPkg/Library/DxeImageVerificationLib/DxeImageVerificationLib.c
```diff
@@ -1705,9 +1705,13 @@ DxeImageVerificationHandler (
...
   //
-  // Sanity check
+  // Sanity check:
+  // Ensure that either File or FileBuffer is provided.
+  // Return EFI_INVALID_PARAMETER if both are NULL.
+  // This prevents security verification from proceeding
+  // when no valid input buffer is available.
   //
-  if (File == NULL) {
+  if ((File == NULL) && (FileBuffer == NULL)) {
```
cfernald, line 1714: Did you mean for this to be a `||`? It seems as written this would allow a caller who provides `File == NULL` but `FileBuffer != NULL` to progress further than before.

vignesh032820-Ami: - The File parameter is optional in DxeImageVerificationHandler.
- Callers often provide only FileBuffer (with File == NULL).
- If we used ||, those valid cases would be rejected.
- The correct logic is to fail only when both are NULL, which is why && is required.
- This preserves expected caller behavior while ensuring security verification does not proceed with invalid inputs.

```
**//
// Sanity check:
// Ensure that either File or FileBuffer is provided.
// Return EFI_INVALID_PARAMETER only if both are NULL.
// File is optional, so a valid FileBuffer alone must be accepted.
//
if (File == NULL && FileBuffer == NULL) {
    return EFI_INVALID_PARAMETER;
}**
```

- Previously, the check could restrict flow even when a valid FileBuffer was provided.
- The updated condition aligns with the intended design: FileBuffer is sufficient for verification, File is optional.
- This change improves robustness without breaking existing caller expectations.

cfernald: I see your point, thanks for clarifying.

cfernald:
@vignesh032820-Ami The change looks good. Please fix your commit message to pass the patch check. Namely, start with `SecurityPkg:` and add a commit message describing the change and how it fixes the behavior as you mentioned in your comment.

cfernald (approved):

PR #12704, opened 2026-06-19
FatPkg: Clean any volume caches during exit boot services

cfernald:
# Description

The current implementation assumes the caller will perform the necessary cleanup before exiting boot services. This has been observed to drop some cached file writes that occur even before the application calling exit boot services is launched.

This commit adds a per-volume pre-ExitBootServices event to flush any dirty caches and perform other cleanup the volume to ensure all write data are persisted and consistent.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- [x] One off shell test to check for dropped writes
- [x] Boot to OS on Q35 QEMU platform, confirmed no repro
- [x] Boot to OS on Patina QEMU platform, confirmed no repro

## Integration Instructions

N/A

### DIFF
--- a/FatPkg/EnhancedFatDxe/Fat.inf
+++ b/FatPkg/EnhancedFatDxe/Fat.inf
@@ -72,6 +72,7 @@
   gEfiFileInfoGuid                      ## SOMETIMES_CONSUMES   ## UNDEFINED
   gEfiFileSystemInfoGuid                ## SOMETIMES_CONSUMES   ## UNDEFINED
   gEfiFileSystemVolumeLabelInfoIdGuid   ## SOMETIMES_CONSUMES   ## UNDEFINED
+  gEfiEventBeforeExitBootServicesGuid   ## CONSUMES             ## Event
 
 [Protocols]
   gEfiDiskIoProtocolGuid                ## TO_START
--- a/FatPkg/EnhancedFatDxe/Misc.c
+++ b/FatPkg/EnhancedFatDxe/Misc.c
@@ -331,6 +331,14 @@ FatDiskIo (
   //
   Status = EFI_VOLUME_CORRUPTED;
   if (Offset + BufferSize <= Volume->VolumeSize) {
+    if (Volume->CachingDisabled) {
+      //
+      // Caching has been turned off for this volume. Convert
+      // IO mode to raw disk access equivalent.
+      //
+      IoMode = (IO_MODE)RAW_ACCESS (IoMode);
+    }
+
     if (CACHE_ENABLED (IoMode)) {
       //
       // Access cache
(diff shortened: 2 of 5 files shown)


FatPkg/EnhancedFatDxe/Fat.inf
```diff
@@ -72,6 +72,7 @@
   gEfiFileInfoGuid                      ## SOMETIMES_CONSUMES   ## UNDEFINED
   gEfiFileSystemInfoGuid                ## SOMETIMES_CONSUMES   ## UNDEFINED
   gEfiFileSystemVolumeLabelInfoIdGuid   ## SOMETIMES_CONSUMES   ## UNDEFINED
+  gEfiEventBeforeExitBootServicesGuid   ## CONSUMES             ## Event
```
os-d, line 75: I assume this is done at Before EBS because boot services are required during the cleanup? If that's the case, can you add a comment, perhaps in the event creation on why Before EBS is chosen?

FatPkg/EnhancedFatDxe/Misc.c
```diff
@@ -449,6 +449,54 @@ FatFreeDirEnt (
...
+    return;
+  }
+
+  Status = FatAcquireLockOrFail ();
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_WARN, "%a: FAT lock busy, skipping flush of %p\n", __func__, Volume->Handle));
+    return;
+  }
+
+  Status = FatCleanupVolume (Volume, NULL, EFI_SUCCESS, NULL);
```
os-d: Instead of calling FatCleanupVolume, should we be calling FatCloseFile? That would be doing it at the file level instead of the volume level, but that would ensure any non blocking writes are completed before flushing and clean up the internal state just so that some other driver couldn't do a final write after we have flushed.

However, that would remove the file out from under a potential writer. So perhaps just doing the `FatWaitNonblockingTask()` before calling `FatCleanupVolume()`? At least anything actively pending is ensured to be written, though I'd have to trace through the code to see if this matters for the cache.

FatPkg/EnhancedFatDxe/Fat.inf
cfernald: Yeah, we are still going to rely on the block IO and it's dependencies beyond that. The thinking is boot services proper, those may not be functional. Will add a comment.

FatPkg/EnhancedFatDxe/Misc.c
cfernald: I'm wondering if we just need to be less invasive thinking about it. I think the best approach is to ensure that all caches are flushed without making assumptions on who may have those. It seems valid that a caller may intend on flushing their own caches using the file system during this event so its best if we dont destroy they possibility. I will rework this to do best effort to flush any caches without destroying any open handles.

FatPkg/EnhancedFatDxe/Fat.inf
cfernald: Added a comment about this on the register

FatPkg/EnhancedFatDxe/Misc.c
cfernald: Made this so it only flushes without destroying any external handles.

os-d:
We've chatted offline on this, but the inherent race conditions make me concerned that we'll still miss cases. I'm wondering if we can disable the cache on this event so that ours is the final flush, if there are any writes past this point they'll go directly to the disk. Reads will be slower as a result, but I think that's safer and going to be the most coherent state.

cfernald:
@os-d, I added a flag that will skip the cache on further IO access. Let me know what you think

os-d (approved):

PR #12713, opened 2026-06-23
MdeModulePkg/UsbBusDxe: Fix UsbPortReset might run into recursive loop

JaredPan:
UsbSelectConfig will introduce the UsbConnectDriver call. If this UsbPortReset has happened in the Usb device driver Start() routine and the device FW cannot be recovered by PortReset, the UsbSelectConfig will introduce the recursive loop.

[Suggested solution]
Since UsbPortReset should not change the Bus Topology, the Reset flow should only SetAddress and reconfigure the device.

Signed-off-by: Marlboro Chuang <marlboro.chuang@dell.com>

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

### DIFF
--- a/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbBus.c
+++ b/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbBus.c
@@ -816,13 +816,14 @@ UsbIoPortReset (
   IN EFI_USB_IO_PROTOCOL  *This
   )
 {
-  USB_INTERFACE  *UsbIf;
-  USB_INTERFACE  *HubIf;
-  USB_DEVICE     *Dev;
-  EFI_TPL        OldTpl;
-  EFI_STATUS     Status;
-  UINT8          DevAddress;
-  UINT8          Config;
+  USB_INTERFACE    *UsbIf;
+  USB_INTERFACE    *HubIf;
+  USB_DEVICE       *Dev;
+  EFI_TPL          OldTpl;
+  EFI_STATUS       Status;
+  UINT8            DevAddress;
+  UINT8            Config;
+  USB_DEVICE_DESC  *DevDesc;
 
   OldTpl = gBS->RaiseTPL (USB_BUS_TPL);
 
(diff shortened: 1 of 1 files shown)


JaredPan:
Hi @lgao4 and @jkolakow,

This PR fixes a potential recursive loop in the USB enumeration flow related to UsbSelectConfig and UsbPortReset.

Could you please help review and share your feedback?

Thanks!

lgao4:
So, Reset doesn't need to ReConfig?

JaredPan:
> So, Reset doesn't need to ReConfig?

Hi @lgao4, 

Port Reset still does reconfigure the hardware — UsbSetAddress and UsbSetConfig are retained and restore the device to CONFIGURED state. What's removed is UsbSelectConfig, which creates new USB_INTERFACE objects and calls UsbConnectDriver / ConnectController. Per UEFI Spec 2.11 §17.2.18, UsbPortReset does not change the bus topology, so the existing software interfaces and driver bindings remain valid. Re-running UsbSelectConfig is what causes the recursive loop when the calling driver's Start() triggers another port reset.

lgao4:
> > So, Reset doesn't need to ReConfig?
> 
> Hi @lgao4,
> 
> Port Reset still does reconfigure the hardware — UsbSetAddress and UsbSetConfig are retained and restore the device to CONFIGURED state. What's removed is UsbSelectConfig, which creates new USB_INTERFACE objects and calls UsbConnectDriver / ConnectController. Per UEFI Spec 2.11 §17.2.18, UsbPortReset does not change the bus topology, so the existing software interfaces and driver bindings remain valid. Re-running UsbSelectConfig is what causes the recursive loop when the calling driver's Start() triggers another port reset.

I see. What functionality has been verified?

JaredPan:
> I see. What functionality has been verified?

Hi @lgao4 ,
This code has been deployed on the production platform for many years and has successfully passed manufacturing tests as well as product shipment qualification tests.

lgao4 (approved):

PR #12720, opened 2026-06-24
NetworkPkg/SnpDxe: Fix MAC address passthrough by reading station address after init

JaredPan:
During SnpUndi32Initialize(), CurrentAddress is unconditionally
overwritten with PermanentAddress before UNDI initialization. This
causes MAC address passthrough (MacPassthru) to fail, as the NIC's
actual current address which may differ from its permanent address
is lost.

After UNDI initialization completes, call PxeGetStnAddr() to read the
NIC's station address via the UNDI interface and update CurrentAddress,
PermanentAddress, and BroadcastAddress in the mode structure with the
values reported by the hardware.

The call is added to both initialization paths: the cable-detect
success path and the fallback no-cable-detect path.

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

### DIFF
--- a/NetworkPkg/SnpDxe/Initialize.c
+++ b/NetworkPkg/SnpDxe/Initialize.c
@@ -188,6 +188,7 @@ SnpUndi32Initialize (
   )
 {
   EFI_STATUS  EfiStatus;
+  EFI_STATUS  StnAddrStatus;
   SNP_DRIVER  *Snp;
   EFI_TPL     OldTpl;
 
@@ -253,6 +254,11 @@ SnpUndi32Initialize (
   //
   if (Snp->CableDetectSupported) {
     if (PxeInit (Snp, PXE_OPFLAGS_INITIALIZE_DETECT_CABLE) == EFI_SUCCESS) {
+      StnAddrStatus = PxeGetStnAddr (Snp);
+      if (EFI_ERROR (StnAddrStatus)) {
+        DEBUG ((DEBUG_WARN, "%a: failed to refresh station address (%r)\n", __func__, StnAddrStatus));
+      }
+
       goto ON_EXIT;
     }
   }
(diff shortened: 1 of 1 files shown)


JaredPan:
Hi @ajfish @ardbiesheuvel @leiflindholm @makubacki and @mdkinney,

This PR fixes an issue where CurrentAddress gets overwritten during SnpUndi32Initialize(), causing MAC passthrough (MacPassthru) to fail. After UNDI initialization, it now retrieves the NIC station address via PxeGetStnAddr() and updates the mode structure accordingly.
The change is applied to both initialization paths (with and without cable detect).
Could you please help take a look when you have time? Thanks!

NetworkPkg/SnpDxe/Initialize.c
```diff
@@ -253,6 +253,7 @@ SnpUndi32Initialize (
   //
   if (Snp->CableDetectSupported) {
     if (PxeInit (Snp, PXE_OPFLAGS_INITIALIZE_DETECT_CABLE) == EFI_SUCCESS) {
+      PxeGetStnAddr (Snp);
```
makubacki: This looks reasonable to me overall, but it would be nice for a NetworkPkg reviewer to confirm.

For these two calls added, it would be useful to at least print the return value here in case of a failure for debug.

Note: Code below assumes a new `StnAddrStatus` variable is used.

```suggestion
      StnAddrStatus = PxeGetStnAddr (Snp);
      if (EFI_ERROR (StnAddrStatus)) {
        DEBUG ((DEBUG_WARN, "%a: failed to refresh station address (%r)\n", __func__, StnAddrStatus));
      }
```

JaredPan: Hi @makubacki,
Thanks for the review and suggestion. I've made the updates.

makubacki (approved):
Approving but it would be great for a NetworkPkg reviewer to review as well. Attn: @mikebeaton, @SaloniKasbekar, @Zclarkwilliams

SaloniKasbekar (approved):

PR #12721, opened 2026-06-24
UefiPayloadPkg/DxeHobLib: Add lazy initialization for gHobList

JaredPan:
This change adds dynamic initialization of gHobList in GetHobList() to handle cases where library constructors are executed before DxeHobListLibConstructor.

Issue:
On ARM platforms using ArmMmuBaseLib, the ArmMmuBaseLibConstructor calls GetFirstGuidHob() before xeHobListLibConstructor has initialized gHobList, causing an ASSERT.

Root Cause:
ArmMmuBaseLib's constructor uses HobLib functions, but UefiPayloadPkg's DxeHobLib assumes gHobList is already initialized by DxeHobListLib's constructor.

Solution:
Add the same lazy initialization pattern already used in MdePkg/Library/DxeHobLib, which dynamically retrieves the HOB list from the System Configuration Table if gHobList is NULL.

This change:
- Aligns behavior with MdePkg/Library/DxeHobLib
- Has no impact on existing functionality
- Improves robustness for ARM platforms

# Description

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?


UefiPayloadPkg/Library/DxeHobLib/HobLib.c
```diff
@@ -35,7 +36,14 @@ GetHobList (
   VOID
   )
 {
-  ASSERT (gHobList != NULL);
+  EFI_STATUS  Status;
+
+  if (gHobList == NULL) {
+    Status = EfiGetSystemConfigurationTable (&gEfiHobListGuid, &gHobList);
+    ASSERT_EFI_ERROR (Status);
+    ASSERT (gHobList != NULL);
```
benjamindoron: I don't like seeing two ASSERTs one after another that mean the same thing. It's convention to use return values, like EFI_STATUS, so I'm not going to recommend dropping the `Status` assignment. How about dropping this line (44)?

JaredPan: Hi @benjamindoron,
Thanks for the review. I've made the updates.

JaredPan:
Hi @benjamindoron @gdong1 @gguo11837463 @jameslu8 @Sean-StarLabs @shuoliu0 ,

This PR fixes a constructor ordering issue in UefiPayloadPkg's DxeHobLib where HobLib APIs may be called before gHobList is initialized, leading to an ASSERT on some ARM platforms.

The change adds lazy initialization in GetHobList(), similar to MdePkg's DxeHobLib, to dynamically retrieve the HOB list when needed.

Could you please take a look? Thanks!

gguo11837463 (approved):

UefiPayloadPkg/Library/DxeHobLib/HobLib.c
JaredPan: Hi @benjamindoron,

I've made the updates. Could you please take a look? Thanks!

gdong1 (approved):
looks good to me.

PR #12727, opened 2026-06-25
MdePkg/UefiDevicePathLib: Fix the potential memory leak issue

EricGao2015:
# Description

The function DevPathToTextUsbWWID() allocates NewStr when the input SerialNumber lacks a null terminator. However, this allocated memory is never freed after use, resulting in a memory leak.

This patch adds the missing FreePool() for NewStr before the function returns, ensuring that the allocated buffer is properly freed and eliminating the potential leak.

By the way, add check for the allocated buffer NewStr to avoid null pointer dereference.

## How This Was Tested

N/A

## Integration Instructions

N/A


MdePkg/Library/UefiDevicePathLib/DevicePathToText.c
AshrafAliS, line 1008: do we need fallback when AllocagePool fails on RELEASE mode.

EricGao2015: Strictly speaking, I think adding `if (NewStr != NULL)` is necessary.

EricGao2015:
@lgao4 @mdkinney Hi reviewers, could you review this PR if you have time? Thank you.

lgao4 (approved):

EricGao2015:
@lgao4 Hi Liming, can this PR be merged now?

lgao4 (approved):

PR #12730, opened 2026-06-25
MdeModulePkg: Print HOB memory type names

Damien-Chen:
…ug log

When iterating EFI_HOB_TYPE_MEMORY_ALLOCATION entries, the existing
DEBUG output only prints the numeric MemoryType value in hexadecimal.
This makes it necessary to manually cross-reference the EFI_MEMORY_TYPE
enum to understand which memory region belongs to which memory type.

Add a new GetMemoryTypeName() API to HobPrintLib that returns the
human-readable string for a given EFI_MEMORY_TYPE value. Use this
function in the DEBUG_CODE_BEGIN() blocks of both DxeMain and PeiCore
dispatcher to append the type name to the existing format string:

Also fix the existing mMemoryTypeStr array in HobPrintLib.c which
incorrectly mapped index 15 to "EfiMaxMemoryType" instead of
"EfiUnacceptedMemoryType".

Before:
```
Memory Allocation 0x00000006 0x812000 - 0x812FFF
Memory Allocation 0x0000000A 0x7EF78000 - 0x7EFFFFFF
Memory Allocation 0x0000000A 0x814000 - 0x81FFFF
```

After:
```
Memory Allocation 0x00000006 0x812000 - 0x812FFF (EfiRuntimeServicesData)
Memory Allocation 0x0000000A 0x7FF78000 - 0x7FFFFFFF (EfiACPIMemoryNVS)
Memory Allocation 0x0000000A 0x814000 - 0x81FFFF (EfiACPIMemoryNVS)
```

# Description

- [x] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Build OvmfX64.dsc with GCC and test in QEMU x86-64 system

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
+++ b/MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
@@ -404,10 +404,11 @@ DxeMain (
     if (GET_HOB_TYPE (Hob) == EFI_HOB_TYPE_MEMORY_ALLOCATION) {
       DEBUG ((
         DEBUG_INFO | DEBUG_LOAD,
-        "Memory Allocation 0x%08x 0x%0lx - 0x%0lx\n", \
-        Hob.MemoryAllocation->AllocDescriptor.MemoryType,                      \
-        Hob.MemoryAllocation->AllocDescriptor.MemoryBaseAddress,               \
-        Hob.MemoryAllocation->AllocDescriptor.MemoryBaseAddress + Hob.MemoryAllocation->AllocDescriptor.MemoryLength - 1
+        "Memory Allocation 0x%08x 0x%0lx - 0x%0lx (%a)\n",
+        Hob.MemoryAllocation->AllocDescriptor.MemoryType,
+        Hob.MemoryAllocation->AllocDescriptor.MemoryBaseAddress,
+        Hob.MemoryAllocation->AllocDescriptor.MemoryBaseAddress + Hob.MemoryAllocation->AllocDescriptor.MemoryLength - 1,
+        GetMemoryTypeName (Hob.MemoryAllocation->AllocDescriptor.MemoryType)
         ));
     }
   }
--- a/MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
+++ b/MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
@@ -397,65 +397,9 @@ DxeMain (
 
   DEBUG ((DEBUG_INFO | DEBUG_LOAD, "HOBLIST address in DXE = 0x%p\n", HobStart));
 
-  DEBUG_CODE_BEGIN ();
-  EFI_PEI_HOB_POINTERS  Hob;
-
-  for (Hob.Raw = HobStart; !END_OF_HOB_LIST (Hob); Hob.Raw = GET_NEXT_HOB (Hob)) {
-    if (GET_HOB_TYPE (Hob) == EFI_HOB_TYPE_MEMORY_ALLOCATION) {
-      DEBUG ((
-        DEBUG_INFO | DEBUG_LOAD,
-        "Memory Allocation 0x%08x 0x%0lx - 0x%0lx (%a)\n",
-        Hob.MemoryAllocation->AllocDescriptor.MemoryType,
-        Hob.MemoryAllocation->AllocDescriptor.MemoryBaseAddress,
-        Hob.MemoryAllocation->AllocDescriptor.MemoryBaseAddress + Hob.MemoryAllocation->AllocDescriptor.MemoryLength - 1,
-        GetMemoryTypeName (Hob.MemoryAllocation->AllocDescriptor.MemoryType)
-        ));
-    }
-  }
-
-  for (Hob.Raw = HobStart; !END_OF_HOB_LIST (Hob); Hob.Raw = GET_NEXT_HOB (Hob)) {
-    if (GET_HOB_TYPE (Hob) == EFI_HOB_TYPE_FV) {
-      DEBUG ((
-        DEBUG_INFO | DEBUG_LOAD,
-        "FV Hob            0x%0lx - 0x%0lx\n",
-        Hob.FirmwareVolume->BaseAddress,
-        Hob.FirmwareVolume->BaseAddress + Hob.FirmwareVolume->Length - 1
-        ));
-    } else if (GET_HOB_TYPE (Hob) == EFI_HOB_TYPE_FV2) {
-      DEBUG ((
-        DEBUG_INFO | DEBUG_LOAD,
-        "FV2 Hob           0x%0lx - 0x%0lx\n",
-        Hob.FirmwareVolume2->BaseAddress,
-        Hob.FirmwareVolume2->BaseAddress + Hob.FirmwareVolume2->Length - 1
-        ));
-      DEBUG ((
-        DEBUG_INFO | DEBUG_LOAD,
-        "                  %g - %g\n",
-        &Hob.FirmwareVolume2->FvName,
-        &Hob.FirmwareVolume2->FileName
-        ));
-    } else if (GET_HOB_TYPE (Hob) == EFI_HOB_TYPE_FV3) {
-      DEBUG ((
-        DEBUG_INFO | DEBUG_LOAD,
-        "FV3 Hob           0x%0lx - 0x%0lx - 0x%x - 0x%x\n",
-        Hob.FirmwareVolume3->BaseAddress,
-        Hob.FirmwareVolume3->BaseAddress + Hob.FirmwareVolume3->Length - 1,
-        Hob.FirmwareVolume3->AuthenticationStatus,
-        Hob.FirmwareVolume3->ExtractedFv
-        ));
-      if (Hob.FirmwareVolume3->ExtractedFv) {
-        DEBUG ((
-          DEBUG_INFO | DEBUG_LOAD,
-          "                  %g - %g\n",
-          &Hob.FirmwareVolume3->FvName,
-          &Hob.FirmwareVolume3->FileName
-          ));
-      }
-    }
-  }
-
-  DEBUG_CODE_END ();
-
+  DEBUG_CODE ( 
+    PrintHobList (HobStart, NULL); 
+  ); 
   //
   // Initialize the Event Services
   //
--- a/MdeModulePkg/Core/Pei/Dispatcher/Dispatcher.c
+++ b/MdeModulePkg/Core/Pei/Dispatcher/Dispatcher.c
@@ -1094,7 +1094,6 @@ PeiCheckAndSwitchStack (
     //
     DEBUG_CODE_BEGIN ();
     UINT32                *StackPointer;
-    EFI_PEI_HOB_POINTERS  Hob;
 
     for (  StackPointer = (UINT32 *)SecCoreData->StackBase;
            (StackPointer < (UINT32 *)((UINTN)SecCoreData->StackBase + SecCoreData->StackSize)) \
@@ -1121,18 +1120,7 @@ PeiCheckAndSwitchStack (
       "  temporary memory heap occupied by memory pages: %d bytes.\n",
       (UINT32)(UINTN)(Private->HobList.HandoffInformationTable->EfiMemoryTop - Private->HobList.HandoffInformationTable->EfiFreeMemoryTop)
       ));
-    for (Hob.Raw = Private->HobList.Raw; !END_OF_HOB_LIST (Hob); Hob.Raw = GET_NEXT_HOB (Hob)) {
-      if (GET_HOB_TYPE (Hob) == EFI_HOB_TYPE_MEMORY_ALLOCATION) {
-        DEBUG ((
-          DEBUG_INFO,
-          "Memory Allocation 0x%08x 0x%0lx - 0x%0lx (%a)\n",
-          Hob.MemoryAllocation->AllocDescriptor.MemoryType,
-          Hob.MemoryAllocation->AllocDescriptor.MemoryBaseAddress,
-          Hob.MemoryAllocation->AllocDescriptor.MemoryBaseAddress + Hob.MemoryAllocation->AllocDescriptor.MemoryLength - 1,
-          GetMemoryTypeName (Hob.MemoryAllocation->AllocDescriptor.MemoryType)
-          ));
-      }
-    }
+    PrintHobList (Private->HobList.Raw, NULL);
 
     DEBUG_CODE_END ();
 
--- a/MdeModulePkg/Core/Pei/Dispatcher/Dispatcher.c
+++ b/MdeModulePkg/Core/Pei/Dispatcher/Dispatcher.c
@@ -1125,10 +1125,11 @@ PeiCheckAndSwitchStack (
       if (GET_HOB_TYPE (Hob) == EFI_HOB_TYPE_MEMORY_ALLOCATION) {
         DEBUG ((
           DEBUG_INFO,
-          "Memory Allocation 0x%08x 0x%0lx - 0x%0lx\n", \
-          Hob.MemoryAllocation->AllocDescriptor.MemoryType,               \
-          Hob.MemoryAllocation->AllocDescriptor.MemoryBaseAddress,        \
-          Hob.MemoryAllocation->AllocDescriptor.MemoryBaseAddress + Hob.MemoryAllocation->AllocDescriptor.MemoryLength - 1
+          "Memory Allocation 0x%08x 0x%0lx - 0x%0lx (%a)\n",
+          Hob.MemoryAllocation->AllocDescriptor.MemoryType,
+          Hob.MemoryAllocation->AllocDescriptor.MemoryBaseAddress,
+          Hob.MemoryAllocation->AllocDescriptor.MemoryBaseAddress + Hob.MemoryAllocation->AllocDescriptor.MemoryLength - 1,
+          GetMemoryTypeName (Hob.MemoryAllocation->AllocDescriptor.MemoryType)
           ));
       }
     }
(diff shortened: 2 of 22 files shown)


khaliid2040:
We already have `MdeModulePkg/Library/HobPrintLib` which prints the Hobs in detailed including memory allocation hobs. I think it is better if PeiCore and DxeCore both use that library instance to print hobs including memory allocation hobs. But this requires platform Integration though.

os-d:
> We already have `MdeModulePkg/Library/HobPrintLib` which prints the Hobs in detailed including memory allocation hobs. I think it is better if PeiCore and DxeCore both use that library instance to print hobs including memory allocation hobs. But this requires platform Integration though.

I agree with using HobPrintLib. It is a BASE lib, so it should be applicable to the cores, though it needs confirmation. The platform integration is very minimal here (i.e. a single line in a DSC). I think the breaking change RFC (which is not yet merged, but appears to be ready to) can be followed for this.

Damien-Chen:
After investigation, `HobPrintLib` is a better choice so that we can avoid adding redundant code in PeiCore and DxeCore.

MdeModulePkg/Core/Pei/Dispatcher/Dispatcher.c
khaliid2040: Replace this hob loop and call PrintHobList().

MdeModulePkg/Library/HobPrintLib/HobPrintLib.c
```diff
@@ -39,7 +39,7 @@ CHAR8  *mMemoryTypeStr[] = {
   "EfiMemoryMappedIOPortSpace",
   "EfiPalCode",
   "EfiPersistentMemory",
-  "EfiMaxMemoryType"
```
khaliid2040, line 42: Don't change this string it is based on UEFI EFI_MEMORY_TYPE

```C
typedef enum {
   EfiReservedMemoryType,
   EfiLoaderCode,
   EfiLoaderData,
   EfiBootServicesCode,
   EfiBootServicesData,
   EfiRuntimeServicesCode,
   EfiRuntimeServicesData,
   EfiConventionalMemory,
   EfiUnusableMemory,
   EfiACPIReclaimMemory,
   EfiACPIMemoryNVS,
   EfiMemoryMappedIO,
   EfiMemoryMappedIOPortSpace,
   EfiPalCode,
   EfiPersistentMemory,
   EfiUnacceptedMemoryType,
   EfiMaxMemoryType
} EFI_MEMORY_TYPE;
```

Make sure it matches that enum.

MdeModulePkg/Library/HobPrintLib/HobPrintLib.c
```diff
@@ -382,6 +382,28 @@ PrintFv3Hob (
   return EFI_SUCCESS;
 }
 
+/**
```
khaliid2040: This function is unneeded. You shouldn't add new code to library, as i said before the library will handle printing the hobs.

MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
khaliid2040: Please use PrintHobList() it will print all hobs. You should replace the entire code block in DEBUG_CODE_BEGIN upto DEBUG_CODE_END and call that function consider StandaloneMmPkg core as example:

https://github.com/tianocore/edk2/blob/0809aa4dceb1fafc1aa11de07cb6c8735d77bf9e/StandaloneMmPkg/Core/StandaloneMmCore.c#L896-L898


Also it maybe better to just copy from it and replace DEBUG_CODE_BEGIN and DEBUG_CODE_END with DEBUG_CODE.

khaliid2040 (changes requested):

khaliid2040:
And please split the commit into per package, each package should have a commit.

MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
Damien-Chen: The function PrintHobList would print very detail information of each memory allocation HOB.

However, the original objective is to have memory type for each memory allocation like the comment in the PR 

```
Memory Allocation 0x00000006 0x812000 - 0x812FFF (EfiRuntimeServicesData)
Memory Allocation 0x0000000A 0x7FF78000 - 0x7FFFFFFF (EfiACPIMemoryNVS)
Memory Allocation 0x0000000A 0x814000 - 0x81FFFF (EfiACPIMemoryNVS)
```

If we want to replace PrintHobList information, the debug log would be a litte bit of complex.

os-d: Can you share what the output with HobPrintLib is?

MdeModulePkg/Library/HobPrintLib/HobPrintLib.c
os-d: Yes, this should be a separate commit in the PR to fix the issue that HobPrintLib wasn't updated when EfiUnacceptedMemoryType was added.

os-d:
> I agree with using HobPrintLib. It is a BASE lib, so it should be applicable to the cores, though it needs confirmation. The platform integration is very minimal here (i.e. a single line in a DSC). I think the breaking change RFC (which is not yet merged, but appears to be ready to) can be followed for this.

https://github.com/tianocore/tianocore-wiki.github.io/blob/main/rfc/text/0003-edk2-breaking-change-and-release-process.md has been checked in, so if this PR continues with HobPrintLib, the breaking change process should be followed there.

MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
Damien-Chen: After using PrintHobList, the debug boot log will looks like

```
HOBLIST address in DXE = 0x7FB37018
Print all Hob information from Hob 0x7FB37018
HOB[0]: Type = EFI_HOB_TYPE_HANDOFF, Offset = 0x0, Length = 0x38
   BootMode            = 0x0
   EfiMemoryTop        = 0x7FF78000
   EfiMemoryBottom     = 0x7BB6C000
   EfiFreeMemoryTop    = 0x7F400000
   EfiFreeMemoryBottom = 0x7BB95838
   EfiEndOfHobList     = 0x7BB95830
HOB[1]: Type = EFI_HOB_TYPE_MEMORY_POOL, Offset = 0x38, Length = 0x48
   AllocationSize    = 0x40
HOB[2]: Type = EFI_HOB_TYPE_MEMORY_POOL, Offset = 0x80, Length = 0x208
   AllocationSize    = 0x200
HOB[3]: Type = EFI_HOB_TYPE_MEMORY_POOL, Offset = 0x288, Length = 0x1C8
   AllocationSize    = 0x1C0
HOB[4]: Type = EFI_HOB_TYPE_MEMORY_POOL, Offset = 0x450, Length = 0x108
   AllocationSize    = 0x100
....
```
Compare to original version, PrintHobList will generate more line of debug message. For me I prefer the debug log should be clean and tidy. However, if you think it is good to use PrintHobList in PeiCore and DxeCore, than I`m also good at this. Let me know your feedback.

Damien-Chen:
> > I agree with using HobPrintLib. It is a BASE lib, so it should be applicable to the cores, though it needs confirmation. The platform integration is very minimal here (i.e. a single line in a DSC). I think the breaking change RFC (which is not yet merged, but appears to be ready to) can be followed for this.
> 
> https://github.com/tianocore/tianocore-wiki.github.io/blob/main/rfc/text/0003-edk2-breaking-change-and-release-process.md has been checked in, so if this PR continues with HobPrintLib, the breaking change process should be followed there.

Yes, I will follow this.

makubacki:
> > > I think the breaking change RFC (which is not yet merged, but appears to be ready to) can be followed for this.
> > 
> > https://github.com/tianocore/tianocore-wiki.github.io/blob/main/rfc/text/0003-edk2-breaking-change-and-release-process.md has been checked in, so if this PR continues with HobPrintLib, the breaking change process should be followed there.
> 
> Yes, I will follow this.

@Damien-Chen, the files described in that process are being added to the repo in https://github.com/tianocore/edk2/pull/12772. Please wait for that PR to complete and then you can add your change to BREAKING-CHANGES.md. Since you're one of the first to go through the process, I'm happy to help you prepare your entry for that file.

MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
os-d: In general, I think it is useful to have HOB information printed upon entrance to DXE. However, I do agree HobPrintLib is fairly verbose and HOBs like the pool allocations aren’t very valuable to see.

But, I do think it makes sense to centralize on that library if you want to add additional information to the HOB prints. That is after all its purpose. The main objection to the original implementation was a duplicate table of strings.

You could explore an update to HobPrintLib. Perhaps a new API that prints tighter information. Or updating the existing one. I think the verbose information is okay, too.

Overall, I would weigh whether having the memory type pretty printed is worthwhile to have a new lib dependency on the cores.

Damien-Chen: Do you prefer adding a single one API in HobPrintLib so that we can easily get memory type name ? like one of commit in this PR, Since this is a helper function Lib so I think it is acceptable to have this kind of modification.

os-d: I don’t have a major preference here, but if you are already going to the step of having the core rely on HobPrintLib, I would suggest having the lib print the HOBs. If you think the standard API is too verbose, I would probably make it less verbose rather than introducing a new less verbose API. The only other consumer is StandaloneMmCore and it has the same considerations as these cores for what is valuable to printz

Damien-Chen:
Hi @khaliid2040, @os-d. I made some modification on PrintHobList and use this single API to replace previous change. Let me know if you think if it needs more memory HOB information to be printed or not in DxeCore and PeiCore.

os-d:
> Hi @khaliid2040, @os-d. I made some modification on PrintHobList and use this single API to replace previous change. Let me know if you think if it needs more memory HOB information to be printed or not in DxeCore and PeiCore.

Thanks for continuing to work on different paths for this PR.

You still have the old commits in the PR, please remove those and only have the commits which are intended to be checked in.

I think dropping the generic HOB print info makes sense, it is not that useful.

I would suggest printing `EFI_HOB_TYPE_MEMORY_POOL` and `EFI_HOB_TYPE_GUID_EXTENSION` at VERBOSE since those are generally not helpful to see from the core side, but printing all the other HOBs. I do think that is valuable from the core side to see FVs reported, memory allocations, and resources described. You may need to check if some of those HOBs are also printed elsewhere in the early core and those prints can be dropped, too.

lgao4:
I don't think it is necessary to add new HobPrintLib. This Library is only used in PeiCore and DxeCore. And, it will break every the existing platform DSC. I suggest to keep the same code logic in these two modules.

os-d:
> I don't think it is necessary to add new HobPrintLib. This Library is only used in PeiCore and DxeCore. And, it will break every the existing platform DSC. I suggest to keep the same code logic in these two modules.

HobPrintLib already exists and is used in StMM core. I personally do not believe adding a single line to a DSC is a major breaking change, very easy to fix, and we have a breaking change process to follow.

That being said, I do think this change is minor and so I have also been on the fence about whether it is worthwhile to add a new library dependency for. Because it was centralizing on an existing library, I was okay with it. But, the change desired is very small, we could say just duplicate the list of memory types to strings or say no the change. That’s why I was suggesting adding more functionality than just the memory allocation HOBs.

@lgao4 because this is a trivial printing change, I’m fine if you want to avoid HobPrintLib and add duplicate strings. I do think it would be nice to print out all the relevant HOBs on entering DXE, which HobPrintLib does, but it’s not critical to do.

Damien-Chen:
> I don't think it is necessary to add new HobPrintLib. This Library is only used in PeiCore and DxeCore. And, it will break every the existing platform DSC. I suggest to keep the same code logic in these two modules.

The original modification was add struct definition that contains all EFI memory type and embedded in DxeCore and PeiCore, and after investigating reviewer suggestion, it`s more adjustable to use HobPrintLib.

If we don`t want to use HobPrintLib or use embedded string to print more memory information, I suggest we can remove the debug memory allocation message from DxeCore and PeiCore.

```
Memory Allocation 0x00000006 0x812000 - 0x812FFF
Memory Allocation 0x0000000A 0x7EF78000 - 0x7EFFFFFF
Memory Allocation 0x0000000A 0x814000 - 0x81FFFF
```
Since I can`t find more valuable information from it as a developer, that is the reason why this PR is created and want to resolve.

os-d:
> > I don't think it is necessary to add new HobPrintLib. This Library is only used in PeiCore and DxeCore. And, it will break every the existing platform DSC. I suggest to keep the same code logic in these two modules.
> 
> The original modification was add struct definition that contains all EFI memory type and embedded in DxeCore and PeiCore, and after investigating reviewer suggestion, it`s more adjustable to use HobPrintLib.
> 
> If we don`t want to use HobPrintLib or use embedded string to print more memory information, I suggest we can remove the debug memory allocation message from DxeCore and PeiCore.
> 
> ```
> Memory Allocation 0x00000006 0x812000 - 0x812FFF
> Memory Allocation 0x0000000A 0x7EF78000 - 0x7EFFFFFF
> Memory Allocation 0x0000000A 0x814000 - 0x81FFFF
> ```
> 
> Since I can`t find more valuable information from it as a developer, that is the reason why this PR is created and want to resolve.

I don't think we should drop the prints. It is fairly easy to find the print, see what it is printing, and look up the memory type. If you wanted to add: `Memory Type:` to that print, it would eliminate one of those steps.

As a core developer, I find it very useful to have the HOB information printed, which is why I was suggesting to expand what is printed. But, if we scope to your specific concern here, I think it is fine to add the simple information to the string and call it good. I'm less inclined to add duplications of the table of strings, but it is not the end of the world to do.

Damien-Chen:
> > > I don't think it is necessary to add new HobPrintLib. This Library is only used in PeiCore and DxeCore. And, it will break every the existing platform DSC. I suggest to keep the same code logic in these two modules.
> > 
> > 
> > The original modification was add struct definition that contains all EFI memory type and embedded in DxeCore and PeiCore, and after investigating reviewer suggestion, it`s more adjustable to use HobPrintLib.
> > If we don`t want to use HobPrintLib or use embedded string to print more memory information, I suggest we can remove the debug memory allocation message from DxeCore and PeiCore.
> > ```
> > Memory Allocation 0x00000006 0x812000 - 0x812FFF
> > Memory Allocation 0x0000000A 0x7EF78000 - 0x7EFFFFFF
> > Memory Allocation 0x0000000A 0x814000 - 0x81FFFF
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Since I can`t find more valuable information from it as a developer, that is the reason why this PR is created and want to resolve.
> 
> I don't think we should drop the prints. It is fairly easy to find the print, see what it is printing, and look up the memory type. If you wanted to add: `Memory Type:` to that print, it would eliminate one of those steps.
> 
> As a core developer, I find it very useful to have the HOB information printed, which is why I was suggesting to expand what is printed. But, if we scope to your specific concern here, I think it is fine to add the simple information to the string and call it good. I'm less inclined to add duplications of the table of strings, but it is not the end of the world to do.

I think my previous suggestion may have been a little over the top. The goal here is simply to make the debug messages clearer and easier to follow, which is one of their primary purposes. This PR removes some steps to help developers better understand the memory allocation details.

khaliid2040:
> I don't think it is necessary to add new HobPrintLib. This Library is only used in PeiCore and DxeCore. And, it will break every the existing platform DSC. I suggest to keep the same code logic in these two modules.

The reason i suggested this library was we could achieve multiple things in one modular way. There is alot of information the code in PeiCore and DxeCore don't print. When it comes to printing hob information people want different kinda information in different ways. 

Anyways, i was okay with the original intent of this PR except that small duplication i thought we could achieve in more modular way.

@Damien-Chen I would suggest you to keep the original logic of the library but make the unnecessary information to be printed at `DEBUG_VERBOSE` level. If you still don't like it then maybe we should stick with your original code.

Damien-Chen:
> > I don't think it is necessary to add new HobPrintLib. This Library is only used in PeiCore and DxeCore. And, it will break every the existing platform DSC. I suggest to keep the same code logic in these two modules.
> 
> The reason i suggested this library was we could achieve multiple things in one modular way. There is alot of information the code in PeiCore and DxeCore don't print. When it comes to printing hob information people want different kinda information in different ways.
> 
> Anyways, i was okay with the original intent of this PR except that small duplication i thought we could achieve in more modular way.
> 
> @Damien-Chen I would suggest you to keep the original logic of the library but make the unnecessary information to be printed at `DEBUG_VERBOSE` level. If you still don't like it then maybe we should stick with your original code.

I think it should more reasonable, but I would prefer this enhancement should be a separate PR.

Damien-Chen:
> > > > I think the breaking change RFC (which is not yet merged, but appears to be ready to) can be followed for this.
> > > 
> > > 
> > > https://github.com/tianocore/tianocore-wiki.github.io/blob/main/rfc/text/0003-edk2-breaking-change-and-release-process.md has been checked in, so if this PR continues with HobPrintLib, the breaking change process should be followed there.
> > 
> > 
> > Yes, I will follow this.
> 
> @Damien-Chen, the files described in that process are being added to the repo in #12772. Please wait for that PR to complete and then you can add your change to BREAKING-CHANGES.md. Since you're one of the first to go through the process, I'm happy to help you prepare your entry for that file.

Let me know what should I do for this, thanks.

makubacki:
> > > > > I think the breaking change RFC (which is not yet merged, but appears to be ready to) can be followed for this.
> > > > 
> > > > 
> > > > https://github.com/tianocore/tianocore-wiki.github.io/blob/main/rfc/text/0003-edk2-breaking-change-and-release-process.md has been checked in, so if this PR continues with HobPrintLib, the breaking change process should be followed there.
> > > 
> > > 
> > > Yes, I will follow this.
> > 
> > 
> > @Damien-Chen, the files described in that process are being added to the repo in #12772. Please wait for that PR to complete and then you can add your change to BREAKING-CHANGES.md. Since you're one of the first to go through the process, I'm happy to help you prepare your entry for that file.
> 
> Let me know what should I do for this, thanks.

The files are in the repo, and the process is summarized here: https://www.tianocore.org/tianocore-wiki.github.io/development/contribution-guides/breaking_change_process.html

Please let me know if you have any specific questions.

Issue #12587, opened 2026-05-15
[Bug]: TPM table offset to LAML and LASL are incorrect


kuqin12:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

SecurityPkg

### Which targets are impacted by this bug?

DEBUG

### Current Behavior

There are 3 TPM2 table implementations:

- https://github.com/tianocore/edk2/blob/b03a21a63e3bd001f52c527e5a57feddb53a690b/SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.c#L79
- https://github.com/tianocore/edk2/blob/b03a21a63e3bd001f52c527e5a57feddb53a690b/SecurityPkg/Tcg/Tcg2Acpi/Tcg2Acpi.c#L87
- https://github.com/tianocore/edk2/blob/b03a21a63e3bd001f52c527e5a57feddb53a690b/DynamicTablesPkg/Library/Acpi/Common/AcpiTpm2Lib/Tpm2Generator.c#L276

The Laml and Lasa fields offsets were incorrect. According to the spec, they should be 68 byte from the beginning of the table, regardless of the size of the `Start Method Specific Parameters`: https://trustedcomputinggroup.org/wp-content/uploads/TCG-ACPI-General-Specification_v1.50_pub.pdf (Table 7). But the offset were all at 12 (other than the dynamic package v5 case)

### Expected Behavior

The Laml and Lasa fields offsets were incorrect. According to the spec, they should be 68 byte from the beginning of the table, regardless of the size of the `Start Method Specific Parameters`: https://trustedcomputinggroup.org/wp-content/uploads/TCG-ACPI-General-Specification_v1.50_pub.pdf (Table 7).

### Steps To Reproduce

Check code implementation above.

### Build Environment

```markdown
- OS(s): Linux + Windows
- Tool Chain(s): GCC, VS2022, CLANGPDB
```

### Version Information

```text
Top of master.
```

### Urgency

Medium

### Are you going to fix this?

Someone else needs to fix it

### Do you need maintainer feedback?

Maintainer feedback requested

### Anything else?

Given these are optional fields carried over from legacy TPM 1.2 spec and the Tcg2 protocol already provided the tcg event log buffer address, we should consider not populating the fields completely.

Raymond-MS:
This should be able to be closed now that #12857 has been merged.

[...]

PR #12732, opened 2026-06-25
SecurityPkg: Introduce Dynamic TCG log scaling

Raymond-MS:
# Description

Implemented dynamic TCG log scaling in Tcg2Dxe. When the log would become truncated it instead now dynamically scales doubling the size each time. An ERROR log is reported that an increase to your base log size should occur such that scaling is not necessary. This is a precaution against platforms that log a lot and the addition of new hashing algorithms for PQC. The log is allocated in BootServices memory. Tests were added via TcgLogTest which includes a DXE driver and a UEFI shell UnitTest app. The DXE driver handles pre-ReadyToBoot tests while the TestApp handles post-ReadyToBoot tests as well as gathering the test results from the DXE driver. Markdown documents were created to detail the changes.

Added the Truncation event marker to the end of the FinalEventLog when it becomes truncated. Added a event signal for when scaling occurs on the normal event log. Consumers can trigger callbacks on this event; the test app uses this to know when scaling occurs.

Added an ACPI log region for the ACPI table LAML/LASA. This region does not scale and can become truncated indicated by the Truncation event marker.

- [ ] Breaking change?
- [ ] Impacts security?
- [x] Includes tests?

## How This Was Tested

Ran CI locally. Tested on the QemuArmVirt platform. 

## Integration Instructions

Include the TcgLogTest .inf's to your platform .dsc and .fdf files. You will need to include both the TcgLogTestDxe and TcgLogTestApp for full functionality.

### DIFF
--- a/SecurityPkg/Include/Guid/Tcg2EventLogScaled.h
+++ b/SecurityPkg/Include/Guid/Tcg2EventLogScaled.h
@@ -1,0 +1,17 @@
+/** @file
+  Defines the GUID used to signal that the TCG event log has been dynamically
+  scaled. Consumers may register a notification callback on this event group
+  to react to the scaling event.
+
+  Copyright (c), Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#pragma once
+
+#define TCG2_EVENT_LOG_SCALED_GUID \
+  { \
+    0x9b4f7c2a, 0x1d3e, 0x4a8b, { 0x9c, 0x6f, 0x5e, 0x8d, 0x2a, 0x1b, 0x4c, 0x7f } \
+  }
+
+extern EFI_GUID  gTcg2EventLogScaledGuid;
--- a/SecurityPkg/Tcg/Tcg2Dxe/README.md
+++ b/SecurityPkg/Tcg/Tcg2Dxe/README.md
@@ -1,0 +1,148 @@
+# Tcg2Dxe
+
+Tcg2Dxe is a DXE-phase UEFI driver that publishes the TCG2 protocol defined
+by the [TCG EFI Protocol Specification](https://trustedcomputinggroup.org/resource/tcg-efi-protocol-specification/).
+Its main responsibilites are to expose a standard interface to a TPM device,
+measure components and events into PCRs, support measured boot, and enable
+secure boot attestation.
+
+## Dynamic Event Log Scaling
+
+The TCG event log is initially allocated with a fixed size defined by a
+PCD: PcdTcgLogAreaMinLen. As firmware components log measured boot
+events the log fills up. Traditionally, when the log is full, subsequent events
+are dropped and the log is marked as truncated.
+
+Tcg2Dxe extends this behavior with **dynamic scaling**: when the log is about
+to overflow, the driver doubles its allocation, copies the existing log into
+the new buffer, and frees the old one. This allows the log to grow as needed
+and avoids losing events.
+
+### How It Works
+
+1. **Scaling check** — Before logging a TCG 2.0 event,
+   `TcgLogDynamicScalingNeeded` calculates whether the new event would exceed
+   the current allocation (`EventLogAreaStruct->Laml`).
+
+2. **Reallocation** — When scaling is needed, `TcgScaleEventLog` allocates a
+   new `EfiBootServicesData` region at twice the current size, copies the
+   existing log, updates the `Lasa`/`Laml` fields in the event log area
+   struct, and frees the old region.
+
+3. **Logging** — After scaling, the new event is logged into the resized buffer
+   via `TcgDxeLogEvent` inside a TPL-raised critical section.
+
+### NormalEventLog vs. FinalEventLog
+
+Tcg2Dxe maintains two distinct event log regions:
+
+| Log | Memory Type | Lifetime | Can Scale |
+| --- | ----------- | -------- | --------- |
+| **Normal log** | `EfiBootServicesData` | Available until `ExitBootServices` | Yes |
+| **Final Events log** | `EfiACPIMemoryNVS` | Persistent | No |
+| **ACPI event log** | `EfiACPIMemoryNVS` | Persistent | No |
+
+- The **Normal log** is the main log copy which is returned via `GetEventLog`.
+  It can grow dynamically via scaling. Note that previous calls to `GetEventLog`
+  could contain stale data if the log was scaled after. It is recommended to
+  call `GetEventLog` each time access is required.
+- The **Final Events log** (`EFI_TCG2_FINAL_EVENTS_TABLE`) records events
+  logged after `GetEventLog` has been called. It is installed as a UEFI
+  configuration table so the OS can discover events that occurred between its
+  call to `GetEventLog` and `ExitBootServices`. Because the **Final Events log**
+  does not scale, it can become truncated.
+- The **ACPI event log** is a fixed-size mirror of the normal log, allocated
+  in `EfiACPIMemoryNVS` and sized from `PcdTcgLogAreaMinLen`. Because this region
+  does not scale, it can become truncated. Its address and length are published
+  to `PcdTpm2AcpiTableLasa` and `PcdTpm2AcpiTableLaml` so `Tcg2Acpi` picks them
+  up when populating the TPM2 ACPI table.
+
+### Scale Limit
+
+The number of times the normal event log region may be dynamically scaled is
+capped by `TCG_EVENT_LOG_MAX_SCALE_COUNT`. Each successful scale doubles the
+allocation, so this caps total growth at `PcdTcgLogAreaMinLen << TCG_EVENT_LOG_MAX_SCALE_COUNT`.
+Once the limit is reached:
+
+1. `TcgScaleEventLog` refuses to scale further and returns
+   `EFI_OUT_OF_RESOURCES`.
+2. `EventLogAreaStruct->EventLogTruncated` is set to `TRUE`, so subsequent
+   `GetEventLog` callers see `EventLogTruncated == TRUE`.
+3. `HashLogExtendEvent` returns `EFI_VOLUME_FULL` for events that would have
+   triggered the refused scale.
+
+### Scaling Notification (`gTcg2EventLogScaledGuid`)
+
+Each time the normal log is successfully resized, `TcgScaleEventLog` calls
+`EfiEventGroupSignal (&gTcg2EventLogScaledGuid)` to notify interested parties
+that the log moved in memory.
+
+Consumers that cache the log base address returned by `GetEventLog` (for
+example, parsers walking the log incrementally) must invalidate their cache
+on this signal and call `GetEventLog` again to get the current `Lasa`/last
+entry. A typical consumer:
+
+```c
+gBS->CreateEventEx (
+       EVT_NOTIFY_SIGNAL,
+       TPL_CALLBACK,
+       OnTcgEventLogScaled,
+       Context,
+       &gTcg2EventLogScaledGuid,
+       &Event
+       );
+```
+
+The event is declared in `gTcg2EventLogScaledGuid` (see
+`SecurityPkg/Include/Guid/Tcg2EventLogScaled.h`) and listed in `Tcg2Dxe.inf`.
+
+## FinalEventLog Truncation Marker
+
+Because the **FinalEventLog** is fixed-size and cannot scale, it can fill
+up before `ExitBootServices`. When the next event would overflow the log,
+`TcgDxeLogEvent` calls `AppendTruncationMarker` which writes a final
+`EV_NO_ACTION` event whose payload is the ASCII string
+`TCG_LOG_TRUNCATION_EVENT_STRING`. The `NumberOfEvents` counter in
+`EFI_TCG2_FINAL_EVENTS_TABLE` is incremented to include the marker, and
+`EventLogTruncated` is set so subsequent attempts return `EFI_VOLUME_FULL`
+without re-appending the marker.
+
+To guarantee the marker always fits, FinalEventLog initialization in
+`SetupEventLog` subtracts `GetTruncationEventSize()` from the usable `Laml`:
+
+```c
+mTcgDxeData.FinalEventLogAreaStruct[Index].Laml =
+    PcdGet32 (PcdTcg2FinalLogAreaLen)
+  - sizeof (EFI_TCG2_FINAL_EVENTS_TABLE)
+  - GetTruncationEventSize ();
+```
+
+`AppendTruncationMarker` temporarily restores this reserved space so
+`TcgCommLogEvent` will accept the marker write.
+
+OS-side and pre-OS consumers can detect FinalEventLog truncation by:
+
+- walking `EFI_TCG2_FINAL_EVENTS_TABLE` and inspecting the last entry for an
+  `EV_NO_ACTION` event whose payload begins with `"TCG Event Log Truncated"`.
+
+## ACPI Event Log Truncation Marker
+
+The **ACPI event log** uses the same truncation-marker mechanism as
+FinalEventLog. Because it is fixed-size and does not scale, `TcgDxeLogEvent`
+appends an `EV_NO_ACTION` truncation event whose payload is
+`TCG_LOG_TRUNCATION_EVENT_STRING` when the next event would overflow the
+region. `EventLogTruncated` is then set so subsequent writes silently skip the
+ACPI region without re-appending the marker. Unlike FinalEventLog, the
+truncated ACPI region does **not** cause `HashLogExtendEvent` to return
+`EFI_VOLUME_FULL`.
+
+To guarantee the marker always fits, the ACPI event log initialization in
+`SetupEventLog` subtracts `GetTruncationEventSize()` from the usable `Laml`:
+
+```c
+mTcgDxeData.AcpiEventLogAreaStruct[Index].Laml =
+    PcdGet32 (PcdTcgLogAreaMinLen) - GetTruncationEventSize ();
+```
+
+Truncation can be detected by walking the region and inspecting the last
+entry for the `"TCG Event Log Truncated"` payload.
--- a/SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
+++ b/SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
@@ -15,6 +15,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Guid/GlobalVariable.h>
 #include <Guid/HobList.h>
 #include <Guid/TcgEventHob.h>
+#include <Guid/Tcg2EventLogScaled.h>
 #include <Guid/EventGroup.h>
 #include <Guid/EventExitBootServiceFailed.h>
 #include <Guid/ImageAuthentication.h>
@@ -67,6 +68,13 @@ TCG2_EVENT_INFO_STRUCT  mTcg2EventInfo[] = {
 
 #define TCG_EVENT_LOG_AREA_COUNT_MAX  2
 
+// Maximum number of times the TCG event log can be dynamically scaled
+// before the log is considered truncated.
+#define TCG_EVENT_LOG_MAX_SCALE_COUNT  4
+
+// Payload written as the last event of the FinalEventLog when it becomes truncated.
+#define TCG_LOG_TRUNCATION_EVENT_STRING  "TCG Event Log Truncated"
+
 typedef struct {
   EFI_TCG2_EVENT_LOG_FORMAT    EventLogFormat;
   EFI_PHYSICAL_ADDRESS         Lasa;
@@ -76,6 +84,7 @@ typedef struct {
   BOOLEAN                      EventLogStarted;
   BOOLEAN                      EventLogTruncated;
   UINTN                        Next800155EventOffset;
+  UINTN                        ScaleCount;
 } TCG_EVENT_LOG_AREA_STRUCT;
 
 // Mapping of TPM return status to BIOS/OS TPM support and related flags (TPMPresentFlag, TpmUpdateFlag)
@@ -93,6 +102,7 @@ typedef struct _TCG_DXE_DATA {
   BOOLEAN                             GetEventLogCalled[TCG_EVENT_LOG_AREA_COUNT_MAX];
   TCG_EVENT_LOG_AREA_STRUCT           FinalEventLogAreaStruct[TCG_EVENT_LOG_AREA_COUNT_MAX];
   EFI_TCG2_FINAL_EVENTS_TABLE         *FinalEventsTable[TCG_EVENT_LOG_AREA_COUNT_MAX];
+  TCG_EVENT_LOG_AREA_STRUCT           AcpiEventLogAreaStruct[TCG_EVENT_LOG_AREA_COUNT_MAX];
 } TCG_DXE_DATA;
 
 TCG_DXE_DATA  mTcgDxeData = {
@@ -124,6 +134,7 @@ VARIABLE_TYPE  mVariableType[] = {
 };
 
 EFI_HANDLE  mImageHandle;
+BOOLEAN     mReadyToBoot = FALSE;
 
 /**
   Measure PE image into TPM log based on the authenticode image hashing in
@@ -931,6 +942,128 @@ TcgCommLogEvent (
 }
 
 /**
+  Get TPML_DIGEST_VALUES compact binary buffer size.
+
+  @param[in]     DigestListBin    TPML_DIGEST_VALUES compact binary buffer.
+
+  @return TPML_DIGEST_VALUES compact binary buffer size.
+**/
+UINT32
+GetDigestListBinSize (
+  IN VOID  *DigestListBin
+  )
+{
+  UINTN          Index;
+  UINT16         DigestSize;
+  UINT32         TotalSize;
+  UINT32         Count;
+  TPMI_ALG_HASH  HashAlg;
+  UINT8          *Current;
+
+  Current   = (UINT8 *)DigestListBin;
+  Count     = ReadUnaligned32 ((UINT32 *)Current);
+  TotalSize = sizeof (Count);
+  Current  += sizeof (Count);
+  for (Index = 0; Index < Count; Index++) {
+    HashAlg    = ReadUnaligned16 ((UINT16 *)Current);
+    TotalSize += sizeof (HashAlg);
+    Current   += sizeof (HashAlg);
+
+    DigestSize = GetHashSizeFromAlgo (HashAlg);
+    TotalSize += DigestSize;
+    Current   += DigestSize;
+  }
+
+  return TotalSize;
+}
+
+/**
+  Build the EV_NO_ACTION truncation marker event.
+
+  @param[out]  EventHdr      TCG_PCR_EVENT2 header initialized for the
+                             truncation marker.
+  @param[out]  EventHdrSize  Size in bytes of the serialized header
+                             (excluding the payload string).
+**/
+STATIC
+VOID
+BuildTruncationEvent (
+  OUT TCG_PCR_EVENT2_HDR  *EventHdr,
+  OUT UINT32              *EventHdrSize
+  )
+{
+  InitNoActionEvent (EventHdr, sizeof (TCG_LOG_TRUNCATION_EVENT_STRING));
+  *EventHdrSize = (UINT32)(sizeof (EventHdr->PCRIndex) +
+                           sizeof (EventHdr->EventType) +
+                           GetDigestListBinSize (&EventHdr->Digests) +
+                           sizeof (EventHdr->EventSize));
+}
+
+/**
+  Compute the total size in bytes of the EV_NO_ACTION truncation marker event
+  (header for the active PCR banks + truncation payload).
+
+  Used to reserve headroom in the Final Events log so the marker is always
+  guaranteed to fit when truncation occurs.
+
+  @return  Size in bytes of the truncation marker event.
+**/
+STATIC
+UINTN
+GetTruncationEventSize (
+  VOID
+  )
+{
+  TCG_PCR_EVENT2_HDR  NoActionEvent;
+  UINT32              EventHdrSize;
+
+  BuildTruncationEvent (&NoActionEvent, &EventHdrSize);
+
+  return EventHdrSize + sizeof (TCG_LOG_TRUNCATION_EVENT_STRING);
+}
+
+/**
+  Append an EV_NO_ACTION truncation marker as the final entry of the Final
+  Events log. The marker is written into the headroom reserved at log
+  initialization (see Final log Laml setup in SetupEventLog) so it is always
+  guaranteed to fit even when the log is otherwise full.
+
+  @param[in,out] EventLogAreaStruct  Final Events log area.
+
+  @retval EFI_SUCCESS  The truncation marker was logged.
+  @retval Other        TcgCommLogEvent failed; nothing was logged.
+**/
+STATIC
+EFI_STATUS
+AppendTruncationMarker (
+  IN OUT TCG_EVENT_LOG_AREA_STRUCT  *EventLogAreaStruct
+  )
+{
+  EFI_STATUS          Status;
+  TCG_PCR_EVENT2_HDR  TruncationHdr;
+  UINT32              TruncationHdrSize;
+
+  BuildTruncationEvent (&TruncationHdr, &TruncationHdrSize);
+
+  // Restore the reserved space so TcgCommLogEvent accepts the marker.
+  EventLogAreaStruct->Laml += TruncationHdrSize + sizeof (TCG_LOG_TRUNCATION_EVENT_STRING);
+
+  Status = TcgCommLogEvent (
+             EventLogAreaStruct,
+             &TruncationHdr,
+             TruncationHdrSize,
+             (UINT8 *)TCG_LOG_TRUNCATION_EVENT_STRING,
+             sizeof (TCG_LOG_TRUNCATION_EVENT_STRING)
+             );
+
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "%a: Failed to log truncation marker - %r\n", __func__, Status));
+  }
+
+  return Status;
+}
+
+/**
   Add a new entry to the Event Log.
 
   @param[in] EventLogFormat  The type of the event log for which the information is requested.
@@ -1015,6 +1148,11 @@ TcgDxeLogEvent (
                NewEventSize
                );
     if (Status == EFI_OUT_OF_RESOURCES) {
+      Status = AppendTruncationMarker (EventLogAreaStruct);
+      if (!EFI_ERROR (Status)) {
+        (mTcgDxeData.FinalEventsTable[Index])->NumberOfEvents++;
+      }
+
       EventLogAreaStruct->EventLogTruncated = TRUE;
       return EFI_VOLUME_FULL;
     } else if (Status == EFI_SUCCESS) {
@@ -1028,41 +1166,34 @@ TcgDxeLogEvent (
     }
   }
 
-  return Status;
-}
+  //
+  // Also record to the fixed-size ACPI event log region.
+  //
+  EventLogAreaStruct = &mTcgDxeData.AcpiEventLogAreaStruct[Index];
 
-/**
-  Get TPML_DIGEST_VALUES compact binary buffer size.
-
-  @param[in]     DigestListBin    TPML_DIGEST_VALUES compact binary buffer.
-
-  @return TPML_DIGEST_VALUES compact binary buffer size.
-**/
-UINT32
-GetDigestListBinSize (
-  IN VOID  *DigestListBin
-  )
-{
-  UINTN          Index;
-  UINT16         DigestSize;
-  UINT32         TotalSize;
-  UINT32         Count;
-  TPMI_ALG_HASH  HashAlg;
-
-  Count         = ReadUnaligned32 (DigestListBin);
-  TotalSize     = sizeof (Count);
-  DigestListBin = (UINT8 *)DigestListBin + sizeof (Count);
-  for (Index = 0; Index < Count; Index++) {
-    HashAlg       = ReadUnaligned16 (DigestListBin);
-    TotalSize    += sizeof (HashAlg);
-    DigestListBin = (UINT8 *)DigestListBin + sizeof (HashAlg);
-
-    DigestSize    = Tpm2GetHashSizeFromAlgo (HashAlg);
-    TotalSize    += DigestSize;
-    DigestListBin = (UINT8 *)DigestListBin + DigestSize;
+  if (EventLogAreaStruct->Lasa == 0) {
+    // No need to handle ACPI event log region.
+    return EFI_SUCCESS;
   }
 
-  return TotalSize;
+  if (!EventLogAreaStruct->EventLogTruncated) {
+    Status = TcgCommLogEvent (
+               EventLogAreaStruct,
+               NewEventHdr,
+               NewEventHdrSize,
+               NewEventData,
+               NewEventSize
+               );
+    if (Status == EFI_OUT_OF_RESOURCES) {
+      AppendTruncationMarker (EventLogAreaStruct);
+      EventLogAreaStruct->EventLogTruncated = TRUE;
+      Status                                = EFI_SUCCESS;
+    } else if (Status == EFI_SUCCESS) {
+      EventLogAreaStruct->EventLogStarted = TRUE;
+    }
+  }
+
+  return Status;
 }
 
 /**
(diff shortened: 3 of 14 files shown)


SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1627,21 +1898,13 @@ SetupEventLog (
...
-                        &Lasa
-                        );
-      }
+
+      Status = gBS->AllocatePages (
+                      AllocateAnyPages,
+                      EfiBootServicesData,
+                      EFI_SIZE_TO_PAGES (PcdGet32 (PcdTcgLogAreaMinLen)),
+                      &Lasa
+                      );
```
jyao1, line 1989: This is a big behavior change.

The original Lasa is in ACPINVS, but now it is in BootServiceData.

Where do you put the LASA finally in the ACPI table?

SecurityPkg/Tcg/Tcg2Dxe/README.md
```diff
@@ -0,0 +1,120 @@
...
+   via `TcgDxeLogEvent` inside a TPL-raised critical section.
+
+### NormalEventLog vs. FinalEventLog
+
+Tcg2Dxe maintains two distinct event log regions:
+
+| Log | Memory Type | Lifetime | Can Scale |
+| --- | ----------- | -------- | --------- |
+| **Normal log** | `EfiBootServicesData` | Available until `ExitBootServices` | Yes |
+| **Final Events log** | `EfiACPIMemoryNVS` | Persistent | No |
```
jyao1, line 42: I am confused on the description.

The original Lasa is in ACPINVS, but now it is in BootServiceData.

Where do you put the LASA finally in the ACPI table?

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: You are correct. The TCG spec shows that the LAML/LASA in the TPM2 ACPI table are optional fields. The idea is not to populate them at all as doing so would mean scaling the NVS region and re-installing the table if it had already been published which isn't ideal. As far as we can tell in our investigations, nobody is using LAML/LASA to acquire the log region, instead, most are using the Tcg2Protocol GetEventLog call to acquire the log. This + the FinalEventLog generates the full log after ExitBootServices. In the test framework I have added, I put in a test that mimics what most do to acquire the log:

1. Call GetEventLog through the Tcg2Protocol. Note that this would normally happen on ReadyToBoot or sometime before ExitBootServices. This activates the FinalEventLog which will contain all logs between the call to GetEventLog and ExitBootServices where the protocol is no longer available.
2. Make a local copy of the log.
3. Acquire the FinalEventLog. Note this would normally happen after ExitBootServices.
4. Append the FinalEventLog onto the local copy of the log generating the full log.

I validate that another call to GetEventLog matches the local copy + FinalEventLog. From a Windows and Linux perspective, this is how we have seen it done.

cfernald: @jyao1, as Raymond mentioned, this is an intentional decision to drop the optional TPM2 TCG log in favor of the TCG protocol which is actually used. The thinking being that the ACPI table memory is wasted as it is not used and will quickly become incomplete for many PCRs anyways. Dropping this allows the TCG log to scale to avoid truncation without impacting runtime memory usage or hibernate scenarios that expect runtime memory stability.

Please let us know if you are aware of scenarios or reasons which the ACPI table log would need to be kept.

MdePkg/Include/IndustryStandard/Tpm2Acpi.h
```diff
@@ -79,6 +77,57 @@ typedef struct {
   UINT8     Reserved[8];
 } EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_ARM_FFA;
 
+typedef union {
+  UINT8                                                     PlatformSpecificParameters[EFI_TPM2_ACPI_TABLE_START_METHOD_SPECIFIC_PARAMETERS_MAX_SIZE_REVISION_4];
+  EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_ARM_SMC    SmcParameters;
+} EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_V4;
+
```
cfernald, line 88: Consider adding this to avoid this being broken in the future. E.g. version N adds something to SMC parameters and someone edits the existing struct instead of creating a new V_ one. 
```c
STATIC_ASSERT(sizeof(EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_V4) == EFI_TPM2_ACPI_TABLE_START_METHOD_SPECIFIC_PARAMETERS_MAX_SIZE_REVISION_4)
```

MdePkg/Include/IndustryStandard/Tpm2Acpi.h
```diff
@@ -79,6 +77,57 @@ typedef struct {
...
+typedef struct {
+  EFI_ACPI_DESCRIPTION_HEADER    Header;
+  UINT32                         Flags;
+  UINT64                         AddressOfControlArea;
+  UINT32                         StartMethod;
+  UINT8                          PlatformSpecificParameters[EFI_TPM2_ACPI_TABLE_START_METHOD_SPECIFIC_PARAMETERS_MAX_SIZE];
+  UINT32                         Laml; // Optional
+  UINT64                         Lasa; // Optional
+} EFI_TPM2_ACPI_TABLE_TEMPLATE;
```
cfernald, line 137: Might as well just define this to the latest version

```c
// Most recent (and largest) definition of the ACPI table
typedef EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_V5 EFI_TPM2_ACPI_TABLE_TEMPLATE
```

SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.c
```diff
@@ -356,54 +325,57 @@ PublishTpm2 (
...
 
-  mTpm2AcpiTemplate.Header.Revision = PcdGet8 (PcdTpm2AcpiTableRev);
+  mTpm2AcpiTemplate.Header.Signature = EFI_ACPI_5_0_TRUSTED_COMPUTING_PLATFORM_2_TABLE_SIGNATURE;
+  mTpm2AcpiTemplate.Header.Length    = sizeof (EFI_TPM2_ACPI_TABLE_V5);
+  mTpm2AcpiTemplate.Header.Revision  = PcdGet8 (PcdTpm2AcpiTableRev);
   DEBUG ((DEBUG_INFO, "Tpm2 ACPI table revision is %d\n", mTpm2AcpiTemplate.Header.Revision));
 
+  // FF-A is only supported in revisions 5 and up.
   if (mTpm2AcpiTemplate.Header.Revision < EFI_TPM2_ACPI_TABLE_REVISION_5) {
-    DEBUG ((DEBUG_ERROR, "%a The minimum revision supported for TPM over FFA table is 5, not %d.\n", __func__, mTpm2AcpiTemplate.Header.Revision));
```
cfernald: I think it's helpful to keep the expected vs found version.

SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.c
```diff
@@ -356,54 +325,57 @@ PublishTpm2 (
...
-    ControlArea->Command                        = PcdGet64 (PcdTpmBaseAddress) + 0x80;
-    ControlArea->Response                       = PcdGet64 (PcdTpmBaseAddress) + 0x80;
-  } else {
-    DEBUG ((DEBUG_ERROR, "TPM2 InterfaceType get error! %d\n", InterfaceType));
+  if (InterfaceType != Tpm2PtpInterfaceCrb) {
+    DEBUG ((DEBUG_ERROR, "TPM over FF-A only supports CRB interface\n"));
     return EFI_UNSUPPORTED;
   }
 
+  mTpm2AcpiTemplate.Flags = (mTpm2AcpiTemplate.Flags & 0xFFFF0000) | PcdGet8 (PcdTpmPlatformClass);
```
cfernald: Please define with mask (and the one below) and put them in the Tpm2Acpi.h

`0xFFFF0000`
`0x0000FFFF`

SecurityPkg/Tcg/Tcg2Acpi/Tcg2Acpi.c
```diff
@@ -815,17 +787,46 @@ PublishTpm2 (
...
+        // Remove LAML/LASA from the length if either is 0.
+        mTpm2AcpiTemplate.Header.Length -= (sizeof (UINT32) + sizeof (UINT64));
+      }
+
+      break;
+
+    default:
+      mTpm2AcpiTemplate.Header.Length = sizeof (EFI_TPM2_ACPI_TABLE_TEMPLATE);
+      DEBUG ((DEBUG_ERROR, "TPM2 revision get error! %d\n", mTpm2AcpiTemplate.Header.Revision));
+      ASSERT (FALSE);
```
cfernald, line 826: Shouldn't we fail? without asserts this will publish a potentially invalid table.

SecurityPkg/Tcg/Tcg2Acpi/Tcg2Acpi.c
```diff
@@ -839,9 +840,11 @@ PublishTpm2 (
       break;
     case Tpm2PtpInterfaceFifo:
     case Tpm2PtpInterfaceTis:
+      mTpm2AcpiTemplate.StartMethod = EFI_TPM2_ACPI_TABLE_START_METHOD_TIS;
       break;
     default:
       DEBUG ((DEBUG_ERROR, "TPM2 InterfaceType get error! %d\n", InterfaceType));
+      ASSERT (FALSE);
```
cfernald, line 849: I think we should fail here too.

SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.c
```diff
@@ -356,54 +325,57 @@ PublishTpm2 (
...
+  DEBUG ((DEBUG_INFO, "Tpm2 ACPI table PlatformClass is %d\n", (mTpm2AcpiTemplate.Flags & 0x0000FFFF)));
+
+  Tpm2AcpiTableV5       = (EFI_TPM2_ACPI_TABLE_V5 *)&mTpm2AcpiTemplate;
+  Tpm2AcpiTableV5->Laml = PcdGet32 (PcdTpm2AcpiTableLaml);
+  Tpm2AcpiTableV5->Lasa = PcdGet64 (PcdTpm2AcpiTableLasa);
+  if ((Tpm2AcpiTableV5->Laml == 0) || (Tpm2AcpiTableV5->Lasa == 0)) {
+    // Remove LAML/LASA from the length if either is 0.
+    mTpm2AcpiTemplate.Header.Length -= (sizeof (UINT32) + sizeof (UINT64));
```
cfernald: Consider using `= EFI_FIELD_OFFSET(EFI_TPM2_ACPI_TABLE_V5, Laml)` instead

SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.c
```diff
@@ -356,54 +325,57 @@ PublishTpm2 (
...
     );
 
-  mTpm2AcpiTemplate.Header.Revision = PcdGet8 (PcdTpm2AcpiTableRev);
+  mTpm2AcpiTemplate.Header.Signature = EFI_ACPI_5_0_TRUSTED_COMPUTING_PLATFORM_2_TABLE_SIGNATURE;
+  mTpm2AcpiTemplate.Header.Length    = sizeof (EFI_TPM2_ACPI_TABLE_V5);
+  mTpm2AcpiTemplate.Header.Revision  = PcdGet8 (PcdTpm2AcpiTableRev);
   DEBUG ((DEBUG_INFO, "Tpm2 ACPI table revision is %d\n", mTpm2AcpiTemplate.Header.Revision));
 
+  // FF-A is only supported in revisions 5 and up.
   if (mTpm2AcpiTemplate.Header.Revision < EFI_TPM2_ACPI_TABLE_REVISION_5) {
```
cfernald: Should this be `==`? After this check, this code continues and initializing the table with the PCD provided revision, but assuming it's version 5. If future version append mandatory fields, this could would break that contract.

SecurityPkg/Include/Guid/Tcg2EventLogScaled.h
```diff
@@ -0,0 +1,20 @@
+/** @file
+  Defines the GUID used to signal that the TCG event log has been dynamically
+  scaled. Consumers may register a notification callback on this event group
+  to react to the scaling event.
+
+  Copyright (c), Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#ifndef TCG2_EVENT_LOG_SCALED_H_
```
cfernald: We have switched to `#pragma once`

e.g. https://github.com/tianocore/edk2/commit/9326c0eb0a9b5b82fef335e46fbd85868d318a7d

SecurityPkg/Tcg/Tcg2Dxe/README.md
cfernald: As mentioned in the other comment thread, this change drops the optional ACPI fields.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -930,6 +940,126 @@ TcgCommLogEvent (
...
+{
+  UINTN          Index;
+  UINT16         DigestSize;
+  UINT32         TotalSize;
+  UINT32         Count;
+  TPMI_ALG_HASH  HashAlg;
+
+  Count         = ReadUnaligned32 (DigestListBin);
+  TotalSize     = sizeof (Count);
+  DigestListBin = (UINT8 *)DigestListBin + sizeof (Count);
```
cfernald: nit: would be cleaner to just define a new `UINT8 *Current` instead of using and constantly having to cast `DigestListBin`.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1183,6 +1434,26 @@ TcgDxeLogHashEvent (
           EventSizePtr           = Tpm2CopyDigestListToBuffer (DigestBuffer, DigestList, mTcgDxeData.BsCap.ActivePcrBanks);
           CopyMem (EventSizePtr, &NewEventHdr->EventSize, sizeof (NewEventHdr->EventSize));
 
+          // Continually scale until we have enough space to log. We need to dynamically
+          // scale the TCG log before we enter a critical region.
+          while (TRUE) {
```
cfernald, line 1467: is this to account for the event being too large for the scaling?

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
jyao1: I do not think this is a good idea.

As far as I know, some Linux systems are still rely on the event log in ACPI table.
(See https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/plain/drivers/char/tpm/eventlog/acpi.c and https://github.com/torvalds/linux/blob/master/include/acpi/actbl3.h)

Also this is a TCG defined feature, existing for a long time.
(See https://trustedcomputinggroup.org/wp-content/uploads/TCG_ACPIGeneralSpec_v1p3_r8_pub.pdf)

I have big concern to drop this feature, before Linux drop it.

Raymond-MS: I just want to mention that the code you posted for Linux is using V4 of the table. We are publishing V5 of the table with the StartMethod Specific Parameters length increased to 16 from 14. The code does no validation of the table revision and in this situation would break because they offset into the table wrong. Their LAML/LASA would be completely wrong. Regardless of whether or not we keep the ACPI region, this platform is going to break. 

I understand that this has been a long-standing feature of the TCG spec for TPM2 ACPI table ever since it was added in V4, however, from its inception, the LAML/LASA have always been marked as optional. Wouldn't it be a platform's responsibility to gracefully handle the LAML/LASA not being populated?

A temporary solution while we allow platforms to move away from the ACPI table LAML/LASA could be having a fixed ACPI log region for the TPM2 table that can become truncated in the event it fills. The idea would be that this would only be available to give platforms time to swap away from LAML/LASA in lieu of the protocol + FinalEventTable. Would this work for you?

jyao1: At least the current V4 is working now. But this patch will break it. That is not a good idea.

The code does not support V5/V6 support yet. If the intention is to support V5/V6, please keep the current design. It is totally irrelevant to dynamic TCG logging.

I prefer we split the requirement (support V5/V6) and (support dynamic TCG logging).
The first part is TCG spec requirement. I will fully support your work if you would like to add V5/V6 support.

But the second part is a pure EDK2 feature, which we have not agreed yet.
I am still confused on the requirement. Why we need the log to be dynamic???
Given a platform, we validated at manufacture time, then we will know what the event log looks like, and we will publish RIM accordingly.
If the event log is truncated, it should be discovered immediately, and the firmware dev is expected to enlarge the event log buffer. Without that, how you publish the final RIM?

I treat the truncation is a defensive action and move the system forward with the partial log. Then it will give you chance to fix. I would not expect the truncation happening at the end user, because that will cause TCG RIM based attestation failure.

Raymond-MS: I disagree. We posted this bug not long ago: https://github.com/tianocore/edk2/issues/12587. If you look at Tcg2Acpi.h, the latest table revision is 5. I am fixing the issue where the platform specific parameters are incorrect. The TCG spec for V5 of the TPM2 table should have the platform specific parameters set to 16. This should have already broken that platform as mentioned earlier.

As for having the log be dynamic, we have noticed there are platforms that have the capability to log an excessive amount. This causes the log to become truncated, in those instances it can be hard to detect until much later. With dynamic scaling you can prevent the log from becoming truncated and pass validation with a valid log while still getting an indicator that your log size must be increased. With PQC coming soon and the addition of multiple hashing algorithms it is entirely possible the log will truncate, we want to avoid that if at all possible so that we can still end up with a valid log.

I also want to mention that from the ACPI table perspective, you would have no indication that your log ever became truncated. This was another reason why we removed the LAML/LASA. Without adding the truncation event marker as I did for the FinalEventTable, a user of the ACPI table would not be able to know they have a truncated log.

cfernald: IMO, part of this stems from the choice to effectively move existing fields from version 4 to version 5. Or to put fields after a variable size field without a field specifying the size of that field in the first place. Either way, the TCG spec made a breaking change for TPM2 version 5. 

Now we are stuck between a rock and a hard place. We need to support version 5 because there are new start methods, FF-A, that *requires* us to use version 5. So just not implementing version 5 is not really an option.

So even if we separate these two problems, Raymond's point still holds that any existing software assuming that LAML/LASA are at offsets 64/68 (instead of 68/72) will break.

On the point of the scaling log, there are a few reasons that we believe makes this important.

1. With the changes for PQC algorithms with larger certificates, larger hash sizes, and more active banks, the size of log events will grow. 

2. The measured events are not constant and are likely to grow in the lifetime of the device. The DBx is a great example of an event whose data gets added to the log will grow through the lifetime of the system as software/certificates are revoked.

3. Besides that, UEFI is not the only one to measure to the log, and some scenarios like PXE may have retries and limits that may cause variable iterations of certain log events to be measured that may not be fully predictable.

I will support splitting this into 2 PRs if that helps to clarify the conversation though.

jyao1: > So even if we separate these two problems, Raymond's point still holds that any existing software assuming that LAML/LASA are at offsets 64/68 (instead of 68/72) will break.

I agree with this statement. That is bug of "existing software", and other existing software need to fix the issue and move to a different offset if they want to support V5/V6.

But that is NOT a reason that you can remove the feature for V4. (Just because of one bug in other existing software, which does not make any sense to me.)

We should follow the TCG spec and provide maximum compatibility as needed.
A feature can be dropped only if 1) spec has dropped it, 2) no existing software is using that.
When both conditions are satisfied, we can start discussing when/how to drop.

Raymond-MS: I understand. I proposed a solution in my earlier comments. Would this work for you?

To prevent the need to go back and read the old message the proposal was to have a fixed sized NVS log region for the ACPI table that can become truncated.

cfernald: To be clear, I'm not opposed to keeping the ACPI LAML/LASA, though I think there is a decision tree that we should work though. The first being: do we believe we need dynamic scaling for the TCG log. Our conclusion has been yes, but I'd like to get your thoughts on the matter, independent of the ACPI table.

If you agree, there are some impacts that become problematic for the ACPI table. namely,

1. The impacts of scaling the log on permanent memory fragmentation, memory buckets, and memory map stability for OS hibernate compatibility.
2. If you diverge the tables on expansion (to avoid 1) then you need to have complexity to diverge and signal that the ACPI table is truncated.
3. Not strictly related, but with the ACPI table we are burning (effectively) permanent memory for a structure that is almost immediately out-of-date for the full PCR context.

Our original assessment, which appears to be wrong, was that the ACPI fields are unused and so it wasn't worth the complexity to support.

jyao1: >  The first being: do we believe we need dynamic scaling for the TCG log.

To be honest, I am not sure. That is implementation choice.
If you think there is real value, I am not opposed to add dynamic scaling as a new feature.

Let's discuss from both technical perspective, and business perspective.

My technical points:
1) If we decide to add dynamic scaling, this new feature SHOULD NOT impact the existing feature, such as ACPI table based event log, and Memory BIN design for S4 resume, etc.
2) If we want to unconditionally support dynamic scaling, then we should remove the use of `PcdTcgLogAreaMinLen` and `PcdTcg2FinalLogAreaLen`. They are no longer valid, if LogArea can be scaled.

If the proposal can satisfy above points, then I do not have technical concern.

And we also need to think about the consequence of scaling. E.g. what if the LogArea is scaled too much? OK? or not-OK?

> As for having the log be dynamic, we have noticed there are platforms that have the capability to log an excessive amount.

From business part, as I mentioned before, I assume platform owner should check the PCR and Event Log, then provide a RIM, before the production is shipped.
I am surprised that you have observed a platform having truncated log. It seems a bug to me.

May I know the root-cause? Which causes the platform having excessive log?
Is that normal in production (e.g. an expected use case)? or abnormal (e.g. under attack, or something)? or test only (e.g. in manufacture time)?
Only if we understand the cause, then we can think about the proper solution.

Raymond-MS: I understand, this is why I proposed having an ACPI log region that gets mapped to the LAML/LASA fields in the table. This way users of the ACPI table can access a version of the log. The concern is that if the normal log is scaling the ACPI log is going to become truncated. As for your second point, I am not sure we would want to remove the PCDs. The idea of dynamic scaling is to provide a mechanism for platforms with unexpected logging to still get a valid log instead of a truncated log. The goal is still for the platform to update the PcdTcgLogAreaMinLen value to something that will prevent scaling from occurring. PcdTcg2FinalLogAreaLen should not be removed either as the FinalEventLog does not dynamically scale; neither will the ACPI log.

I put in a limit for the number of times we allow dynamic scaling to occur (4). After this point we give up and mark the log as truncated.

MdePkg/Include/IndustryStandard/Tpm2Acpi.h
Raymond-MS: Added.

MdePkg/Include/IndustryStandard/Tpm2Acpi.h
Raymond-MS: I don't think this is necessary. This should always be the latest because MAX_SIZE should always follow the latest revision. This prevents the need to make this change every time a new revision gets added. Only the MAX_SIZE needs to change to the latest revision size.

SecurityPkg/Include/Guid/Tcg2EventLogScaled.h
Raymond-MS: Fixed.

SecurityPkg/Tcg/Tcg2Acpi/Tcg2Acpi.c
Raymond-MS: This function doesn't really follow that flow of exiting early but I will update it as I agree there is no reason to continue with publishing the table if it's invalid.

SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.c
Raymond-MS: The comment is correct that V5 and up should support the FF-A parameters which is what Tcg2AcpiFfa is meant to do, however, we should be checking for != to V5 because we are specifically using V5. If we update in the future to use a version greater than V5 we will need to make sure to publish that version correctly.

SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.c
Raymond-MS: Sounds good.

SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.c
Raymond-MS: Sounds good.

SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.c
Raymond-MS: Added.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: Updated.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: Yes, this was something Michael had brought up where if the log was set so small that one single scale wouldn't allow for the log to fit we should scale until we either exhaust the attempts or it fits. This should normally never happen but it is better for us to gracefully handle it.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
jyao1: > The goal is still for the platform to update the PcdTcgLogAreaMinLen value to something that will prevent scaling from occurring. 

OK. If the goal is still let platform to update the LogAreaLen then there is no need to add dynamic  scalling feature at all.

To me, the real requirement seems to be **a platform owner need to have a mechanism to know if the log is truncated or not.**
(Currently, there is no such way. because only OS will call the GetEventLog API to know if it is truncated. Also, ACPI table does not provide such info.)

Do you agree? If so, that could be a much lightweight solution.

cfernald: @jyao1. to answer your question about platforms where this was previously observed, it was server platforms that would boot primarily using PXE, which could lead to many measurements due to retry logic. That is one problem, and presumably a platform owner could do somtehing meaningful here.

However, for client devices I'm not sure what the platform owner would do exactly. Most user don't even know what the TPM is, much less the TCG log. The problem is that the TCG log length cannot or may not be fully predicted by the platform manufacturer because of the increased cost of scenarios, certificates, revocations, etc. This is quite impactful for an OS that has to enable security features on these devices that require a complete and accurate TCG log. So even if 10% of manufacturers inaccurately predict the required space, that 10% of devices that may become unusable for some security features.

It's a bit of hindsight, but this all really seems far more complicated than it needs to be. The TCG log simply need to just contain the measurements made by UEFI (and before). It is a snapshot of the state at exit boot services and will quickly become incomplete as the OS take control and extends further measurements. In my opinion, while i'm okay keeping it in the ACPI table, its value there is pretty significant diminished by its immediate staleness. The most important thing is that the OS and user can retrieve an accurate and complete log no matter what. To that end I would like to make sure that even if the ACPI tables version is truncated because of limitations with memory bins, for example, that the protocol + final event log are always complete.

Are you suggesting an alternative where instead of dynamically scaling the platform would just update some persistent information and reboot, similar to memory bins?

Raymond-MS: @jyao1, I understand where you are coming from and with the current implementation we get what you are asking for; a way for the platform to know if the log is truncated (i.e. via the Truncation flag in GetEventLog and now the Truncation event marker in both the ACPI log and the FinalEventLog). However, I don't see why there should be any reluctance to adopt a dynamically scaling log. We won't always be able to predict scenarios in which the log may become truncated. As mentioned earlier, with PQC and the addition of more hashing algorithms the log may bloat significantly. With dynamic scaling, it's possible to still receive a valid log while still getting some indication that the default log size is invalid and needs to be increased. With dynamic scaling, it is entirely possible to never see a truncated log which I think is the ideal situation, right?

jyao1: @cfernald 
> it was server platforms that would boot primarily using PXE, which could lead to many measurements due to retry logic. That is one problem, and presumably a platform owner could do somtehing meaningful here.

Thanks to let me know.
Yes, I assume platform owner should finally publish the RIM.
If RIM does not include event log, then they do not care.
If RIM does include event log, then they should care about event log, and check event log, then publish.

> Most user don't even know what the TPM is, much less the TCG log. 

Exactly. Event log is NOT designed for end user.
Event log is meaningful in attestation scenario.
If the manufacturer does not publish a good event log in RIM, a standalone event log will not be very useful.

> The problem is that the TCG log length cannot or may not be fully predicted by the platform manufacturer because of the increased cost of scenarios, certificates, revocations, etc.

First, TCG does not require platform manufacturer to predict.
TCG just asks platform manufacture to publish final RIM. If they cannot predict, that is fine. They can boot and collect the first time event log, then evaluate and publish.

Second, if there is worried that event log will be changed in some scenarios, and people do not know what they are, then it is NOT a trusted scenarios. People just lose the trust statement, because they do not know.
For example, an attacker can keep launching different images in hundreds, even thousand times, and keep adding new items to the event log. 
Of course, we can keep scaling and keep logging, but at one time, the memory will be almost exhausted.
Then we still cannot see a full log, but a truncated log.

That is why TCG only require a reasonable size of log, to cover all **legal conditions** which the platform is known and trusted.
I do not think there is requirement to always have a full log for any illegal conditions. To me that is a mission impossible.

> while i'm okay keeping it in the ACPI table, its value there is pretty significant diminished by its immediate staleness. 

Allow me to share the history,
Event log in ACPI table was designed for Linux since TPM 1.2, back to 20 years ago.
Event log in TCG2_Protocol was introduced by MSFT since TPM 2.0, back to 10 years ago. The reason is that we want to support multiple TCG event log formats for compatibility consideration. A single event log in ACPI cannot server the compatibility.
But later, we believe TPM 1.2 can be dropped, then a single TPM 2.0 log format in ACPI is possible, and Linux people like to use ACPI table, instead of EFI protocol to get information.

If we skip the compatibility support phase for event log, then ACPI table is the only choice. (Not weird TCG2 protocol + final log)

> Are you suggesting an alternative where instead of dynamically scaling the platform would just update some persistent information and reboot, similar to memory bins?

That is an interesting idea.
To be honest, I was not suggesting it before. But it is worth considering.

One problem we encountered before is that if we only support scale up, then it might become too large in some illegal condition. We might need a way to scale down.
Another problem is time of refresh. Memory BIN refresh is done at ready to boot event, because we assume no special memory will be allocated later. But is that statement also true for event log? Even worse, if we need to enlarge event log area after that, then we might break the memory BIN feature.

I believe we need to think through the corner cases from technical perspective.

@Raymond-MS 
> However, I don't see why there should be any reluctance to adopt a dynamically scaling log. 

Please allow me to explain my logic:
1) **understand the problem statement first**
2) **figure out the solution to resolve the problem, but not over-design**

If the problem statement is: Platform has no way to know log is truncated.
Then the solution should be: Design such mechanism to notify platform.

To me, scaling log means the platform owner should not know the size of event log, it should just work. That is why I think we should remove PcdTcgLogAreaMinLen and PcdTcg2FinalLogAreaLen.
However, you still want to keep them, which means, you do want the platform owner has the knowledge of event log size, which devalue the scaling log proposal. That also confused me on what your goal is, what problem you are trying to resolve.

To generalize, if there is a simpler way to resolve the problem, I always prefer simpler way, instead of introducing a complicated solution and break the compatibility.

I also mentioned before, "If you think there is real value, I am not opposed to add dynamic scaling as a new feature."
**To be clear, what I am looking for is:**
1) **a clear problem statement**, that dynamic scaling log is the only solution or best solution. (So far I am not convinced that it is a must-have based on your description.)
2) **a compatible technical proposal**, to keep current existing feature (such as ACPI event log, and memory BIN, etc).

Please enlighten me.

spbrogan:
@jyao1  - Here is my attempt at a problem statement. 

Pre-boot TCG logging contains both statically configured platform things[^1] + user (and boot target) configured things[^2] and thus it is not practical for a platform vendor to determine a static size of the log.  In-Market platforms with existing firmware have seen some errors and it is expected to get significantly worse in the coming years.  Errors are catastrophic for platform/boot security and as designed today have no path for user-based recovery.   

[^1]: These are events like PCR0, PCR1, and a reasonable assumption of PCR2,3,4,5,7
[^2]: These are PCR2,3,4,5 & 7.  Reasonable assumptions could be made in the past, but this is not holding up with changes occurring in the industry.  Thinks like the KEK/DB/DBX content size changes have caused overflows in PCR7.  Multiple boot attempts because of user defined boot order have caused overflows in PCR 4,5.

jyao1:
Thanks Sean.
Here is summary of the discussion:

1) One PR for one feature only. Please split ACPI V5/V6 change from the Dynamic Log Scaling feature. This PR should just deal with scaling. The ACPI change should be in a different PR.
2) No breaking change for existing feature. The ACPI log in NVS can be as is. The new dynamic scaling is only for DXE phase TCG2 protocol GetEventLog API. (Scaling ACPI NVS is too risky and that might break other features.)
3) The final Event Log is NOT scaled in this PR. (It is kept as is.) It might be considered in other PR.
4) No PCD change. The original 2 PCD are still valid. PcdTcgLogAreaMinLen is used to control ACPI based log. PcdTcg2FinalLogAreaLen is used to control final event log.

Raymond-MS:
> Thanks Sean. Here is summary of the discussion:
> 
> 1. One PR for one feature only. Please split ACPI V5/V6 change from the Dynamic Log Scaling feature. This PR should just deal with scaling. The ACPI change should be in a different PR.
> 2. No breaking change for existing feature. The ACPI log in NVS can be as is. The new dynamic scaling is only for DXE phase TCG2 protocol GetEventLog API. (Scaling ACPI NVS is too risky and that might break other features.)
> 3. The final Event Log is NOT scaled in this PR. (It is kept as is.) It might be considered in other PR.
> 4. No PCD change. The original 2 PCD are still valid. PcdTcgLogAreaMinLen is used to control ACPI based log. PcdTcg2FinalLogAreaLen is used to control final event log.

Should have addressed all feedback from our discussion. Removed the ACPI table changes from this PR and pushed up: https://github.com/tianocore/edk2/pull/12857. ACPI log is in NVS memory and only the protocol log scales (up to a max of 4 times). FinalEventLog does not scale and can become truncated same as the ACPI log with the truncation event marker. No PCD changes made.

Raymond-MS:
@jyao1 Would it be possible to request for https://github.com/tianocore/edk2/pull/12857 to be reviewed and merged before this? The TcgLogTestApp uses the updates to the ACPI table in order to validate. I would need to overhaul the tests to only use the current ACPI changes and then restore them once the ACPI table updates go in which seems like overkill. If the ACPI changes could go in first, I should be able to rebase and the errors I am seeing should go away.

jyao1:
> @jyao1 Would it be possible to request for #12857 to be reviewed and merged before this? The TcgLogTestApp uses the updates to the ACPI table in order to validate. I would need to overhaul the tests to only use the current ACPI changes and then restore them once the ACPI table updates go in which seems like overkill. If the ACPI changes could go in first, I should be able to rebase and the errors I am seeing should go away.

Yes. I agree.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1121,6 +1247,155 @@ CopyDigestListBinToBuffer (
...
+
+  Status = gBS->AllocatePages (
+                  AllocateAnyPages,
+                  EfiBootServicesData,
+                  EFI_SIZE_TO_PAGES ((UINTN)NewLaml),
+                  &NewLasa
+                  );
+
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "Failed to allocate new TCG event log\n"));
```
cfernald, line 1304: Should you set `EventLogAreaStruct->EventLogTruncated = TRUE;` here too?

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1183,6 +1459,26 @@ TcgDxeLogHashEvent (
...
+                                     NewEventHdr->EventSize
+                                     );
+
+            if (!DynamicScalingNeeded) {
+              break;
+            }
+
+            Status = TcgScaleEventLog (&mTcgDxeData.EventLogAreaStruct[Index]);
+            if (EFI_ERROR (Status)) {
+              DEBUG ((DEBUG_ERROR, "Unable to scale the TCG event log!\n"));
```
cfernald, line 1481: You could also just set `EventLogAreaStruct->EventLogTruncated = TRUE;` here instead to be sure.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1832,7 +2173,7 @@ SetupEventLog (
 
         mTcgDxeData.FinalEventLogAreaStruct[Index].EventLogFormat        = mTcg2EventInfo[Index].LogFormat;
         mTcgDxeData.FinalEventLogAreaStruct[Index].Lasa                  = Lasa + sizeof (EFI_TCG2_FINAL_EVENTS_TABLE);
-        mTcgDxeData.FinalEventLogAreaStruct[Index].Laml                  = PcdGet32 (PcdTcg2FinalLogAreaLen) - sizeof (EFI_TCG2_FINAL_EVENTS_TABLE);
+        mTcgDxeData.FinalEventLogAreaStruct[Index].Laml                  = PcdGet32 (PcdTcg2FinalLogAreaLen) - sizeof (EFI_TCG2_FINAL_EVENTS_TABLE) - GetTruncationEventSize ();
```
cfernald, line 2180: Why is the truncation event size remove from Laml here? I understand the `EFI_TCG2_FINAL_EVENTS_TABLE` since it is not part of the log, but the truncation event would be, right?

Raymond-MS: This makes sure that we have enough room for the truncation event. Without it we need to keep track of the size such that if the current event + sizeof (truncation event) would exceed the size of the log. I understand where you are coming from though. I will investigate.

Raymond-MS: Okay, I did a little bit of digging. This LAML field is only local to this file and is used to guarantee we have enough room for the truncation marker event. For FinalEventLog, the header indicates how many entries exist within the log so there is no issue there. As for the ACPI log, the LAML that is set in the ACPI table is the true LAML not this value used in Tcg2Dxe. Both instances should be safe.

To answer your question, we remove the size of the truncation marker event to reserve enough space for it in the log. The truncation marker event is indeed part of the LAML and should be reported in both instances.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: I think setting EventLogTruncated here when we attempt to scale is the best option, it handles all cases and consolidates setting this flag into a single location when scaling. Updated.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: Moved setting the EventLogTruncated flag to a single location if scaling fails.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
cfernald: I see. That is reasonable. Thanks!

cfernald (approved):

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1622,26 +1917,71 @@ SetupEventLog (
   DEBUG ((DEBUG_INFO, "SetupEventLog\n"));
 
   //
-  // 1. Create Log Area
+  // 0. Create ACPI Log Area.
   //
   for (Index = 0; Index < sizeof (mTcg2EventInfo)/sizeof (mTcg2EventInfo[0]); Index++) {
     if ((mTcgDxeData.BsCap.SupportedEventLogs & mTcg2EventInfo[Index].LogFormat) != 0) {
-      mTcgDxeData.EventLogAreaStruct[Index].EventLogFormat = mTcg2EventInfo[Index].LogFormat;
-      if (PcdGet8 (PcdTpm2AcpiTableRev) >= 4) {
```
jyao1, line 1630: May I know why this check is removed for ACPI logging?

If the ACPI table revision is smaller, then we there is no need to create LASA/LAML.

Raymond-MS: The check before was if the table revision was >= 4 || LogFormat == TCG2, below this I check for LogFormat == TCG2 and only continue on that specific format. At that point we do not care what the table revision is, right? My understanding is that TCG1.2 has plans to be phased out in the near future which is why this code doesn't really support it.

jyao1: > My understanding is that TCG1.2 has plans to be phased out in the near future which is why this code doesn't really support it.

Please note: this is Tcg2Dxe, and it only supports TPM2. Not TPM 1.2 support at all.
My comment for the table revision is irrelevant to TCG 1.2.

Second, according to TCG ACPI table spec, if revision is smaller than 4, then there is NO LASA/LAML.
There is no need to allocate ACPI_NVS. That is why we put original logic there.
For example, if a platform configure the ACPI table revision 3, why the UEFI need allocate NVS but no one uses it.

I am not convinced that we should remove table revision check and always allocate LASA/LAML even it is NOT needed.

Raymond-MS: But that isn't how it worked before. As mentioned in my previous comment, the check used to be:

```
      if ((PcdGet8 (PcdTpm2AcpiTableRev) >= 4) ||
          (mTcg2EventInfo[Index].LogFormat == EFI_TCG2_EVENT_LOG_FORMAT_TCG_2))
      {
        //
        // Report TCG2 event log address and length, so that they can be reported in TPM2 ACPI table.
        // Ignore the return status, because those fields are optional.
        //
        PcdSet32S (PcdTpm2AcpiTableLaml, (UINT32)mTcgDxeData.EventLogAreaStruct[Index].Laml);
        PcdSet64S (PcdTpm2AcpiTableLasa, mTcgDxeData.EventLogAreaStruct[Index].Lasa);
      }
```

LAML/LASA get set regardless of PcdTpm2AcpiTableRev if the LogFormat is TCG2. Is this a bug? Was this meant to be an &&?

I understand where you are coming from in terms of why allocate the NVS memory region if we are not going to set LAML/LASA but I already handle that via the LogFormat check. If the LogFormat is TCG2, we allocate the region and set the LAML/LASA as indicated in the above code segment.

As for TPM1.2 support, I am confused as to why you are saying Tcg2Dxe does not support it at all. If that is the case why does this still exist?

```
TCG2_EVENT_INFO_STRUCT  mTcg2EventInfo[] = {
  { &gTcgEventEntryHobGuid,  EFI_TCG2_EVENT_LOG_FORMAT_TCG_1_2 },
  { &gTcgEvent2EntryHobGuid, EFI_TCG2_EVENT_LOG_FORMAT_TCG_2   },
};

#define TCG_EVENT_LOG_AREA_COUNT_MAX  2
```

We are looping through the EventLogAreaStruct and EventLogFormat (TCG2/TCG1.2) throughout the file. Even in Tcg2DxeLogHashEvent there is a case statement to handle TCG1.2 logging.

What I meant before was there are plans to remove all of this code from Tcg2Dxe in the future.

jyao1: 1) About PcdTpm2AcpiTableRev

The original code logic (https://github.com/tianocore/edk2/blob/master/SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c#L1630-L1644) is:
 * If PcdTpm2AcpiTableRev >= 4, the event log is in ACPINvs.
 * Else, the event log is in BootServiceData. (**No ACPINvs allocated**)
 * Both ACPI event log and normal event log are same.

Since you want to add dynamic scaling feature, we need to split.

In your code, the logic is:
 * the ACPI event log is in ACPINvs (no scaling) (**ACPINvs allocated even PcdTpm2AcpiTableRev < 4**)
 * normal event log is in BootServiceData (support scaling)

That is NOT aligned with original logic, because there is no need to allocate ACPINvs if PcdTpm2AcpiTableRev < 4.

My expectation is to align with original logic.
 * If PcdTpm2AcpiTableRev >= 4, the ACPI event log is in ACPINvs (no scaling)
 * Else ACPI event log is absent. (**No ACPINvs allocated**)
 * normal event log is in BootServiceData (support scaling) - regardless PcdTpm2AcpiTableRev

Please let me know if that is clear.

2) About TCG 1.2

I think there is misunderstanding. I mixed TPM 1.2 and TCG 1.2 log format.
Yes, TCG2 DXE still support TCG 1.2 log format, but it does not support TPM 1.2.

> If the LogFormat is TCG2, we allocate the region and set the LAML/LASA as indicated in the above code segment.

You are right. And I think you fixed an issue that ACPINvs is allocated even in TCG1.2 style before.

My comment is to do more: If (the LogFormat is TCG2) **and (PcdTpm2AcpiTableRev >= 4)**, we allocate the region and set the LAML/LASA

Raymond-MS: Sounds good, I updated the check to include checking the table revision. I also fixed an issue related to this to make sure we only log to the ACPI log region if it was successfully allocated.

jyao1 (approved):

PR #12734, opened 2026-06-26
OvmfPkg/VirtioInputDxe: Add mouse and tablet support in virtio-input driver

jiaqingz-dev:
# Description

This series generalizes the existing `OvmfPkg/VirtioKeyboardDxe` driver into a single `VirtioInputDxe` driver that supports virtio keyboard, mouse and tablet devices on a shared transport/dispatch core.

`VirtioKeyboardDxe` added in 8bc9f5a2bc42 ("OvmfPkg: Virtio based keyboard driver implementation") only consumes virtio-input keyboard events and produces `EFI_SIMPLE_TEXT_INPUT(_EX)_PROTOCOL`. However, virtio-input is a generic device class that also carries relative (mouse) and absolute (tablet) pointing devices. This series reuses the existing virtqueue/event plumbing and adds the two pointer protocols, so a single driver can drive `virtio-keyboard-pci`, `virtio-mouse-pci` and `virtio-tablet-pci`.

The work is split into three phases, ordered so that every commit builds and is bisectable:

1. **Fix and clean up the existing keyboard driver** (still named
   `VirtioKeyboardDxe`), so the bug fixes can be backported independently of the
   new feature:
   - `Fix protocol uninstall in BindingStop` - pass the handle by value to `UninstallMultipleProtocolInterfaces()` (it takes `EFI_HANDLE`, not `EFI_HANDLE *`), so `Stop()` no longer fails with `EFI_INVALID_PARAMETER`.
   - `Fix RingMap check in UninitRing` - correct the unmap guard.
   - `Close WaitForKey(Ex) events in Uninit()` - avoid leaking the `WaitForKey`/`WaitForKeyEx` events on teardown.
   - `Drop unused functions and variables`, `Fix typo in IsKeyRegistered()`, `Simplify WaitForKey event handling`, `Declare internal functions as STATIC` - cleanup, no functional change.

2. **Restructure for multi-device support:**
   - `Rename VirtioKeyboardDxe to VirtioInputDxe` (OvmfPkg + ArmVirtPkg DSC/FDF and the `PrepareVirtioInputDevicePath` reference in `BdsPlatform.c`).
   - `OvmfPkg/IndustryStandard: Add type definitions for virtio input device` - new `Include/IndustryStandard/VirtioInput.h` with the virtio 1.x input config/event structures.
   - `Fix Reset() function to only flush input state` - per the UEFI spec, `Reset()` should drop pending input rather than re-initialize the device.
   - `Add keyboard capability detection` - probe the `EV_KEY`/`EV_REL`/`EV_ABS` bitmaps so the driver installs only the protocols a device actually reports, and fails binding with `EFI_UNSUPPORTED` when none are present.
   - `Split keyboard logic into VirtioKeyboard.c` - separate keyboard handling from the shared core, no functional change.

3. **Add the new pointer devices:**
   - `Add virtio mouse support` - `VirtioMouse.c` produces `EFI_SIMPLE_POINTER_PROTOCOL`, translating `EV_REL` events into relative movement and handling mouse button button states.
   - `Add virtio tablet support` - `VirtioTablet.c` produces `EFI_ABSOLUTE_POINTER_PROTOCOL`, reading the per-axis `abs_info` min/max ranges and translating `EV_ABS` events into absolute coordinates.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

- Built `OvmfPkg/OvmfPkgX64.dsc` (X64, GCC) in `DEBUG` and `RELEASE`
- Booted OVMF under QEMU with `-device virtio-keyboard-pci`, `-device virtio-mouse-pci` and `-device virtio-tablet-pci` and verified inputs with a custom input testing EFI application.
- Booted Xen PVH guest (QEMU GPEX with no PS2/USB keyboard/mouse/tablet by default, virtio-input-pci only) and verified with the same input testing application.

## Integration Instructions

Any out of tree platform referencing `VirtioKeyboardDxe` should use the new `VirtioInputDxe` instead.


OvmfPkg/Include/IndustryStandard/VirtioInput.h
```diff
@@ -0,0 +1,76 @@
...
+typedef struct {
+  UINT32    Min;
+  UINT32    Max;
+  UINT32    Fuzz;
+  UINT32    Flat;
+  UINT32    Res;
+} VIRTIO_INPUT_ABS_INFO;
+#pragma pack ()
+
+#pragma pack(1)
```
AshrafAliS: pragma pack can be added to full file instead of repeating for every struct

OvmfPkg/VirtioInputDxe/VirtioInput.h
```diff
@@ -0,0 +1,200 @@
...
+
+typedef struct {
+  UINTN                      Signature;
+  EFI_KEY_DATA               KeyData;
+  EFI_KEY_NOTIFY_FUNCTION    KeyNotificationFn;
+  LIST_ENTRY                 NotifyEntry;
+} VIRTIO_INPUT_IN_EX_NOTIFY;
+
+// Data structure representing ring buffer
+typedef struct {
```
AshrafAliS, line 46: pragma pack not needed?

jiaqingz-dev: It is not needed here, it's a struct holding device status used by the driver. The comment here seems to be not precise. It was moved directly from original `VirtioKeyboardDxe`...

kraxel:
Nice.

I think the logic for key press vs. button press is a bit too simple:
* Anything below 256 is a keypress, that is correct.
* The 0x100 -> 0x15f range is for all kinds of button events.
* Anything above 0x160 is keypress again.

See `/usr/include/linux/input-event-codes.h`.

jiaqingz-dev:
> Nice.
> 
> I think the logic for key press vs. button press is a bit too simple:
> 
> * Anything below 256 is a keypress, that is correct.
> * The 0x100 -> 0x15f range is for all kinds of button events.
> * Anything above 0x160 is keypress again.
> 
> See `/usr/include/linux/input-event-codes.h`.

Fixed

kraxel (approved):

jiaqingz-dev:
Hi, @ardbiesheuvel @jyao1 Could you please help review this PR? Thank you very much.

jiaqingz-dev:
Just bring some attention to this PR

Hi, @elmarco @leiflindholm @ardbiesheuvel @samimujawar @mikebeaton @jyao1 @tperard @bexcran
Would you please help reviewing this change? Thanks a lot.

ardbiesheuvel (approved):

PR #12740, opened 2026-06-29
OvmfPkg: Add PcdEfiShellAutoboot to optionally restore shell autoboot

luigix25:
# Description

Commits 896907b53b and ca278f14c2 set LOAD_OPTION_CATEGORY_APP on the EFI Shell boot option, preventing it from being used as a fallback boot target. This is desirable for interactive use but breaks automated testing frameworks like edk2-sct that rely on the shell starting automatically after each VM reboot.

Add a new FeaturePcd PcdEfiShellAutoboot (default FALSE) that enables or disables autoboot on the shell.
Users who need autoboot can enable it at build time via:

  build --pcd PcdEfiShellAutoboot=TRUE

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Build edk2 with `--pcd PcdEfiShellAutoboot=TRUE`
Run VM without any disk attached and check if shell autoboots 

## Integration Instructions

N/A

### DIFF
--- a/OvmfPkg/Library/PlatformBootManagerLib/BdsPlatform.c
+++ b/OvmfPkg/Library/PlatformBootManagerLib/BdsPlatform.c
@@ -1693,7 +1693,9 @@ PlatformBootManagerAfterConsole (
   PlatformRegisterFvBootOption (
     &gUefiShellFileGuid,
     L"EFI Internal Shell",
-    LOAD_OPTION_ACTIVE | LOAD_OPTION_CATEGORY_APP,
+    FeaturePcdGet (PcdEfiShellAutoboot)
+      ? LOAD_OPTION_ACTIVE
+      : LOAD_OPTION_ACTIVE | LOAD_OPTION_CATEGORY_APP,
     ShellEnabled
     );
 
--- a/OvmfPkg/Library/PlatformBootManagerLib/PlatformBootManagerLib.inf
+++ b/OvmfPkg/Library/PlatformBootManagerLib/PlatformBootManagerLib.inf
@@ -59,6 +59,9 @@
   QemuFwCfgSimpleParserLib
   PlatformBootManagerCommonLib
 
+[FeaturePcd]
+  gUefiOvmfPkgTokenSpaceGuid.PcdEfiShellAutoboot
+
 [Pcd]
   gUefiOvmfPkgTokenSpaceGuid.PcdEmuVariableEvent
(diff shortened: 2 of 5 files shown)


luigix25:
cc @kilaterlee

kilaterlee:
> cc @kilaterlee

@luigix25 Hi Luigix, sorry for misspelling your name in the email. I think this PR looks good, thanks a lot!

kilaterlee (approved):
I have tested it and it works.

kraxel:
I think if a test case depends on EFI shell it should explicitly run EFI shell instead of depending on OVMF implementation details.  Also note that any OVMF build with secure boot support will not include EFI Shell in the firmware image for security reasons.

One option to do so is to copy EFI shell to a boot media and pass that to the VM.  CI does that, see `FlashRomImage` in `OvmfPkg/PlatformCI/PlatformBuildLib.py`.

Another option is to use qemu direct kernel boot.  When using OVMF any EFI binary can be passed there, so using `qemu -kernel Shell.efi` works.

kilaterlee:
> I think if a test case depends on EFI shell it should explicitly run EFI shell instead of depending on OVMF implementation details. Also note that any OVMF build with secure boot support will not include EFI Shell in the firmware image for security reasons.
> 
> One option to do so is to copy EFI shell to a boot media and pass that to the VM. CI does that, see `FlashRomImage` in `OvmfPkg/PlatformCI/PlatformBuildLib.py`.
> 
> Another option is to use qemu direct kernel boot. When using OVMF any EFI binary can be passed there, so using `qemu -kernel Shell.efi` works.

@edhay Hi Edhaya, what do you think about this issue?

edhay:
> LOAD_OPTION_CATEGORY_APP

Thank you @kilaterlee 

Hi @luigix25 
Will this change prevent executables like UEFI-SCT from continuing on restart if ekd2 built by default (PcdEfiShellAutoboot=FALSE)?
If yes, why dont we make the default value TRUE?

luigix25:
> > LOAD_OPTION_CATEGORY_APP
> 
> Thank you @kilaterlee
> 
> Hi @luigix25 Will this change prevent executables like UEFI-SCT from continuing on restart if ekd2 built by default (PcdEfiShellAutoboot=FALSE)? If yes, why dont we make the default value TRUE?

This PR does not change current behavior.

Anyway, given @kraxel response I don't think this PR it's necessary after all.

PR #12745, opened 2026-07-01
MdeModulePkg, SecurityPkg: Fix GPT measurement mismatch (CVE-2024-13745)

r1chard-lyu:
# Description

As reported in CVE-2024-13745 via oss-sec
(https://seclists.org/oss-sec/2026/q2/727), `DxeTpm2MeasureBootLib` can
measure a GPT partition table that differs from the one parsed and used by
the `PartitionDxe` driver. The two components carry independent GPT parsing
logic, and the measurement code runs after PartitionDxe has already parsed
the table, allowing an attacker to present inconsistent partition-table
instances to different consumers and desynchronize what is recorded in
PCR[5] from what the system actually uses.

This series addresses the issue by:
- Introducing a shared `GptLib` to consolidate GPT parsing
  and validation logic.
- Updating `DxeTpm2MeasureBootLib` to use `GptLib`.
- Adding `GptLib` library mappings for affected platforms.
- Strengthening GPT header validation.
- Preventing `PartitionDxe` from continuing with an invalid
  primary GPT after recovery failure.

Ref: https://seclists.org/oss-sec/2026/q2/727
CVE: CVE-2024-13745

- [X] Breaking change?
- [X] Impacts security?
- [X] Includes tests?

## How This Was Tested
**Build regression**
- Built with GCC across 8 platforms (X64, IA32,
AARCH64, RISCV64): OvmfPkgX64, IntelTdxX64, MicrovmX64, ArmVirtQemu,
RiscVVirtQemu, etc. All compiled successfully.

**Integration test (QEMU + OVMF + swtpm, -D TPM2_ENABLE=TRUE)**
- Attack disk - a GPT image whose primary partition-entry-array CRC is
corrupted (one byte flipped at offset 1152) while the backup stays valid,
attached as a read-only virtio-blk drive so the primary table cannot be
restored. Confirmed fail-closed: fw debug log shows repeated
"Restore primary partition table error", GPT parsing aborts at the new
`goto Done`, and no partition child handles are created. Pre-fix code
fell through and parsed the invalid primary header.
- Normal disk - a valid, writable GPT image. Boots normally:
`map -r` in the UEFI Shell maps the GPT partition (HD(1,GPT,...)),
the debug log reports "Valid primary and Valid backup partition table"
with no restore errors. Confirms no regression.

**Host Based Unit Test**
- Coverage Validated GptLib parsing logic against valid tables, corrupted CRCs, and malformed GPT scenarios.
    - build -p MdeModulePkg/Test/MdeModulePkgHostTest.dsc -a X64 -t GCC -b NOOPT -m MdeModulePkg/Library/GptLib/UnitTest/GptLibUnitTestHost.inf
    - ./Build/MdeModulePkg/HostTest/NOOPT_GCC/X64/GptLibUnitTestHost

## Integration Instructions
This series adds a new library class, GptLib, consumed by PartitionDxe
and by DxeTpm2MeasureBootLib / DxeTpmMeasureBootLib. Any platform that
builds any of these modules must resolve the GptLib library class,
otherwise the build will fail:

```
GptLib|MdeModulePkg/Library/GptLib/GptLib.inf
```

In-tree platforms (the ArmVirt family, EmulatorPkg, UefiPayloadPkg, and
the OVMF platforms that build PartitionDxe or link the TPM measure-boot
libraries) are already wired up by this series. Out-of-tree platforms
need the same one-line mapping in their [LibraryClasses] sections.

Tracking Issue: tianocore/edk2#12808 (breaking change record: see the
[BREAKING-CHANGES.md entry](https://github.com/r1chard-lyu/edk2/blob/0034d845f9c7cfe7d1ee5cadf2c5add683dd45a3/BREAKING-CHANGES.md#breaking-change-new-gptlib-library-class-dependency)
added by this series).


cfernald (approved):
Approving for the SecurityPkg change.

jyao1:
@r1chard-lyu, I like the idea to create GptLib to centralize the check. Thanks for the work.

> **Integration test (QEMU + OVMF + swtpm, -D TPM2_ENABLE=TRUE)**
>
> Attack disk - a GPT image whose primary partition-entry-array CRC is
corrupted (one byte flipped at offset 1152) while the backup stays valid,
attached as a read-only virtio-blk drive so the primary table cannot be
restored. Confirmed fail-closed: fw debug log shows repeated
"Restore primary partition table error", GPT parsing aborts at the new
goto Done, and no partition child handles are created. Pre-fix code
fell through and parsed the invalid primary header.
Normal disk - a valid, writable GPT image. Boots normally:
map -r in the UEFI Shell maps the GPT partition (HD(1,GPT,...)),
the debug log reports "Valid primary and Valid backup partition table"
with no restore errors. Confirms no regression.

**Question for test:**
Do you think if it is possible to create some host based Unit Test for GptLib?
As such, we can cover more corner cases on mal GPT format.

r1chard-lyu:
> **Question for test:** Do you think if it is possible to create some host based Unit Test for GptLib? As such, we can cover more corner cases on mal GPT format.

@jyao1 Yes, I agree. If the current changes are acceptable, I can create a new PR to implement the host-based unit test and cover more corner cases for malformed GPT formats to ensure we cover a more complete range of test scenarios. Thanks for the suggestion.

jyao1:
@r1chard-lyu , yes, current patch looks good to me.

But I would prefer to have more unit test for it. That will increase the confidence.
You can create another commit to add unit test.

MarsDoge:
@r1chard-lyu Thanks for working on this. The GptLib direction looks good to me because it
lets PartitionDxe and measured boot use the same validated GPT view.
    
One scope question: this PR updates DxeTpm2MeasureBootLib, but
DxeTpmMeasureBootLib still seems to keep the old GPT measurement path, reading
the primary header directly from LBA 1 and validating it without the new
GptLib / partition-entry-array CRC check.
    
Since #12636 mentioned both DxeTpmMeasureBootLib and
DxeTpm2MeasureBootLib, should the legacy TCG/TPM1 path also be switched to
GptLib, or is it intentionally out of scope for CVE-2024-13745?

r1chard-lyu:
> @r1chard-lyu Thanks for working on this. The GptLib direction looks good to me because it lets PartitionDxe and measured boot use the same validated GPT view.
> 
> One scope question: this PR updates DxeTpm2MeasureBootLib, but DxeTpmMeasureBootLib still seems to keep the old GPT measurement path, reading the primary header directly from LBA 1 and validating it without the new GptLib / partition-entry-array CRC check.
> 
> Since #12636 mentioned both DxeTpmMeasureBootLib and DxeTpm2MeasureBootLib, should the legacy TCG/TPM1 path also be switched to GptLib, or is it intentionally out of scope for CVE-2024-13745?

Thank you for looking into this. I think there's no reason not to apply the same update to the legacy TCG/TPM1 path. Since we are already aware of the issue, it should also be switched to GptLib. I will add the corresponding changes to this PR. Thanks!

r1chard-lyu:
> @r1chard-lyu , yes, current patch looks good to me.
> 
> But I would prefer to have more unit test for it. That will increase the confidence. You can create another commit to add unit test.

@jyao1 Thanks for the feedback! That makes complete sense. I've added additional commits with host-based unit tests to cover both normal and malformed GPT formats. Please take a look and let me know if you have any further suggestions.

jyao1 (approved):

MarsDoge (approved):
I re-checked the latest version, and the scope now looks consistent with the
discussion in #12636: PartitionDxe, DxeTpm2MeasureBootLib and
DxeTpmMeasureBootLib all select the GPT data through the shared GptLib
validation path, so the measured GPT view should stay aligned with the parser
used by the firmware.

This looks good to me.

lgao4:
So, this fix will require to add new GptLib into the platform DSC.

r1chard-lyu:
> So, this fix will require to add new GptLib into the platform DSC.

@lgao4, yes. Therefore, I have already updated the other platform DSCs accordingly.

lgao4:
> > So, this fix will require to add new GptLib into the platform DSC.
> 
> @lgao4, yes. Therefore, I have already updated the other platform DSCs accordingly.

If so, I would suggest to add generic MdeModuleLibs.dsc.inc file in MdeModulePkg. This file adds this new library instance. Then, update every platform DSC to include MdeModuleLibs.dsc.inc. This way can avoid the future incompatible change.

os-d:
> > > So, this fix will require to add new GptLib into the platform DSC.
> > 
> > 
> > @lgao4, yes. Therefore, I have already updated the other platform DSCs accordingly.
> 
> If so, I would suggest to add generic MdeModuleLibs.dsc.inc file in MdeModulePkg. This file adds this new library instance. Then, update every platform DSC to include MdeModuleLibs.dsc.inc. This way can avoid the future incompatible change.

@mdkinney has brought this up before as well. Another, more involved change, but to me better than dsc includes which can be convoluted, is the already edk2 build spec defined (but not implemented in BaseTools) idea of a recommended library class. If one isn’t specified in a dsc, then the default is chosen. Of course, that is magic in another way :) and requires much more work.

I think that an MdeModuleLibs.dsc.inc probably would bridge the gap between people who want/need to add new library dependencies and those who feel that is a breaking change to be avoided, as much as I personally try to avoid DSC include files.

os-d:
> > > > So, this fix will require to add new GptLib into the platform DSC.
> > > 
> > > 
> > > @lgao4, yes. Therefore, I have already updated the other platform DSCs accordingly.
> > 
> > 
> > If so, I would suggest to add generic MdeModuleLibs.dsc.inc file in MdeModulePkg. This file adds this new library instance. Then, update every platform DSC to include MdeModuleLibs.dsc.inc. This way can avoid the future incompatible change.
> 
> @mdkinney has brought this up before as well. Another, more involved change, but to me better than dsc includes which can be convoluted, is the already edk2 build spec defined (but not implemented in BaseTools) idea of a recommended library class. If one isn’t specified in a dsc, then the default is chosen. Of course, that is magic in another way :) and requires much more work.
> 
> I think that an MdeModuleLibs.dsc.inc probably would bridge the gap between people who want/need to add new library dependencies and those who feel that is a breaking change to be avoided, as much as I personally try to avoid DSC include files.

@lgao4 I resurrected a branch I started some time ago to add the recommended library instance functionality: https://github.com/tianocore/edk2/pull/12788

r1chard-lyu:
> > > So, this fix will require to add new GptLib into the platform DSC.
> > 
> > 
> > @lgao4, yes. Therefore, I have already updated the other platform DSCs accordingly.
> 
> If so, I would suggest to add generic MdeModuleLibs.dsc.inc file in MdeModulePkg. This file adds this new library instance. Then, update every platform DSC to include MdeModuleLibs.dsc.inc. This way can avoid the future incompatible change.

@lgao4 Thanks for the suggestion! I've added MdeModulePkg/MdeModuleLibs.dsc.inc (following the MdePkg/MdeLibs.dsc.inc pattern), and the platform DSCs now resolve GptLib via `!include MdeModulePkg/MdeModuleLibs.dsc.inc`.

Please let me know if you have any further feedback!

r1chard-lyu:
> > > > > So, this fix will require to add new GptLib into the platform DSC.
> > > > 
> > > > 
> > > > @lgao4, yes. Therefore, I have already updated the other platform DSCs accordingly.
> > > 
> > > 
> > > If so, I would suggest to add generic MdeModuleLibs.dsc.inc file in MdeModulePkg. This file adds this new library instance. Then, update every platform DSC to include MdeModuleLibs.dsc.inc. This way can avoid the future incompatible change.
> > 
> > 
> > @mdkinney has brought this up before as well. Another, more involved change, but to me better than dsc includes which can be convoluted, is the already edk2 build spec defined (but not implemented in BaseTools) idea of a recommended library class. If one isn’t specified in a dsc, then the default is chosen. Of course, that is magic in another way :) and requires much more work.
> > I think that an MdeModuleLibs.dsc.inc probably would bridge the gap between people who want/need to add new library dependencies and those who feel that is a breaking change to be avoided, as much as I personally try to avoid DSC include files.
> 
> @lgao4 I resurrected a branch I started some time ago to add the recommended library instance functionality: #12788 

@os-d Yes, I agree MdeModuleLibs.dsc.inc seems like a good middle ground to bridge the gap between the different groups for now. Thanks for reviving that in #12788! Would the two mechanisms be complementary, given that the fallback only applies when a DSC has no mapping? If so, I'm happy to add @RecommendedInstance to the GptLib consumer INFs as a follow-up once it lands.

os-d:
> > > > > > So, this fix will require to add new GptLib into the platform DSC.
> > > > > 
> > > > > 
> > > > > @lgao4, yes. Therefore, I have already updated the other platform DSCs accordingly.
> > > > 
> > > > 
> > > > If so, I would suggest to add generic MdeModuleLibs.dsc.inc file in MdeModulePkg. This file adds this new library instance. Then, update every platform DSC to include MdeModuleLibs.dsc.inc. This way can avoid the future incompatible change.
> > > 
> > > 
> > > @mdkinney has brought this up before as well. Another, more involved change, but to me better than dsc includes which can be convoluted, is the already edk2 build spec defined (but not implemented in BaseTools) idea of a recommended library class. If one isn’t specified in a dsc, then the default is chosen. Of course, that is magic in another way :) and requires much more work.
> > > I think that an MdeModuleLibs.dsc.inc probably would bridge the gap between people who want/need to add new library dependencies and those who feel that is a breaking change to be avoided, as much as I personally try to avoid DSC include files.
> > 
> > 
> > @lgao4 I resurrected a branch I started some time ago to add the recommended library instance functionality: #12788
> 
> @os-d Yes, I agree MdeModuleLibs.dsc.inc seems like a good middle ground to bridge the gap between the different groups for now. Thanks for reviving that in #12788! Would the two mechanisms be complementary, given that the fallback only applies when a DSC has no mapping? If so, I'm happy to add @RecommendedInstance to the GptLib consumer INFs as a follow-up once it lands.

I’d like to get feedback from @mdkinney on direction before this merges. Adding MdeModuleLibs.dsc.inc will still be a breaking change now (and so please follow the breaking change RFC if that is added) but can help prevent breaking changes in the future. In general, I don’t think the burden of an extra line in the DSC is very high, but others disagree and in general I’d like this problem to have a better solution. With dsc.inc, every package needs one. They also get complex and easily overridden on accident. Ideally, I’d like to avoid adding more dsc.inc files.

The recommended instance (or a version of it) makes it not a breaking change, but is a new paradigm.

r1chard-lyu:
> Adding MdeModuleLibs.dsc.inc will still be a breaking change now (and so please follow the breaking change RFC if that is added) but can help prevent breaking changes in the future.

Since this is a breaking change, an RFC has been submitted to comply with the RFC process:
https://github.com/tianocore/tianocore-wiki.github.io/pull/39

os-d:
> > Adding MdeModuleLibs.dsc.inc will still be a breaking change now (and so please follow the breaking change RFC if that is added) but can help prevent breaking changes in the future.
> 
> Since this is a breaking change, an RFC has been submitted to comply with the RFC process: [tianocore/tianocore-wiki.github.io#39](https://github.com/tianocore/tianocore-wiki.github.io/pull/39)

@r1chard-lyu sorry for the confusion, you do not need to submit an RFC for a new library dependency. You would fall into this category: https://github.com/tianocore/tianocore-wiki.github.io/blob/main/rfc/text/0003-edk2-breaking-change-and-release-process.md#non-removal-breaking-changes. Concise instructions: https://github.com/tianocore/tianocore-wiki.github.io/blob/main/rfc/text/0003-edk2-breaking-change-and-release-process.md#guide-level-explanation.

You only need to submit an RFC where you want or need broader community discussion and alignment. Typically this would be for large sweeping changes or controversial changes.

r1chard-lyu:
> @r1chard-lyu sorry for the confusion, you do not need to submit an RFC for a new library dependency. You would fall into this category: https://github.com/tianocore/tianocore-wiki.github.io/blob/main/rfc/text/0003-edk2-breaking-change-and-release-process.md#non-removal-breaking-changes. Concise instructions: https://github.com/tianocore/tianocore-wiki.github.io/blob/main/rfc/text/0003-edk2-breaking-change-and-release-process.md#guide-level-explanation.
> 
> You only need to submit an RFC where you want or need broader community discussion and alignment. Typically this would be for large sweeping changes or controversial changes.

Thanks for pointing out the exact spot. I'll add the other required items later, per the documentation guidelines for non-breaking changes.

r1chard-lyu:
@lgao4, after some discussion with the os-d, I restored the original approach in this PR, which does not use `MdeModuleLibs.dsc.inc`.

Introducing `MdeModuleLibs.dsc.inc` at this point would cause several concerns, as discussed in https://github.com/tianocore/edk2/pull/12745#issuecomment-4931418430 and https://github.com/tianocore/edk2/pull/12788. This is a significant change and may require an RFC to allow for broader discussion. Therefore, it is outside the scope of this submission and would also block further progress on this PR.

For these reasons, I prefer the original approach. What are your thoughts on this?

As an additional note, I added a separate commit to follow the process outlined in https://github.com/tianocore/edk2/pull/12745#issuecomment-4970335351. This additional commit is intended to comply with the breaking change process; everything else remains unchanged.

lgao4:
> @lgao4, after some discussion with the os-d, I restored the original approach in this PR, which does not use `MdeModuleLibs.dsc.inc`.
> 
> Introducing `MdeModuleLibs.dsc.inc` at this point would cause several concerns, as discussed in [#12745 (comment)](https://github.com/tianocore/edk2/pull/12745#issuecomment-4931418430) and #12788. This is a significant change and may require an RFC to allow for broader discussion. Therefore, it is outside the scope of this submission and would also block further progress on this PR.
> 
> For these reasons, I prefer the original approach. What are your thoughts on this?
> 
> As an additional note, I added a separate commit to follow the process outlined in [#12745 (comment)](https://github.com/tianocore/edk2/pull/12745#issuecomment-4970335351). This additional commit is intended to comply with the breaking change process; everything else remains unchanged.

I am OK for your agreement. I have no other comment.

lgao4 (approved):

r1chard-lyu:
@cfernald @jyao1 @lgao4 
Is there anyone else who needs to review this, or is it ready to be merged?

vbpandya:
This PR introduced a build failure in edk2-platforms
https://github.com/tianocore/edk2-platforms/pull/1017 fixes Platforms/ARM/*

r1chard-lyu:
> This PR introduced a build failure in edk2-platforms [tianocore/edk2-platforms#1017](https://github.com/tianocore/edk2-platforms/pull/1017) fixes Platforms/ARM/*

Yes, this is an expected breaking change. Thanks for fixing it in edk2-platforms!

makubacki:
> > This PR introduced a build failure in edk2-platforms [tianocore/edk2-platforms#1017](https://github.com/tianocore/edk2-platforms/pull/1017) fixes Platforms/ARM/*
> 
> Yes, this is an expected breaking change. Thanks for fixing it in edk2-platforms!

For others in the future, the [Breaking Change process](https://www.tianocore.org/tianocore-wiki.github.io/development/contribution-guides/breaking_change_process.html) has an [edk2-platforms step](https://www.tianocore.org/tianocore-wiki.github.io/rfc/text/0003-edk2-breaking-change-and-release-process.html#cross-repository-coordination-edk2-platforms) for this reason.

PR #12749, opened 2026-07-01
ShellPkg: Fix smbiosview Type 26 location/status decode

vbpandya:
## Description

Fix the SMBIOS shell viewer decode for Type 26 Voltage Probe
LocationAndStatus and add missing Type 28 Temperature Probe location
decode values.

smbiosview extracts Type 26 location from bits 4:0 and status from
bits 7:5, matching the SMBIOS structure definition. However, the lookup
tables used for those decoded values were swapped: VPLocationTable
contained status strings and VPStatusTable contained location strings.

As a result, a valid Type 26 record could be displayed as:

  Voltage Probe - Location: OK
  Voltage Probe - Status: Processor

Swap the table contents so the decoded output correctly reports:

  Voltage Probe - Location: Processor
  Voltage Probe - Status: OK

Also extend the Type 28 Temperature Probe location table with the SMBIOS
defined values 0x0C through 0x0F:

  0x0C Front Panel Board
  0x0D Back Panel Board
  0x0E Power System Board
  0x0F Drive Back Plane

### DIFF
--- a/ShellPkg/Library/UefiShellDebug1CommandsLib/SmbiosView/QueryTable.c
+++ b/ShellPkg/Library/UefiShellDebug1CommandsLib/SmbiosView/QueryTable.c
@@ -3246,33 +3246,6 @@ TABLE_ITEM  VPLocationTable[] = {
   },
   {
     0x03,
-    L" OK "
-  },
-  {
-    0x04,
-    L" Non-critical "
-  },
-  {
-    0x05,
-    L" Critical "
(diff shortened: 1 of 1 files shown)


vbpandya:
CI failure looks unrelated to the PR, I'll wait for the initial review and then if there are comment I'll address those and hopefully the CI will be back by then.

pierregondois (approved):

PR #12750, opened 2026-07-01
FmpDevicePkg/FmpDxe: Reduce NV variable reads for FMP state

shubham-msft:
## Description

In PopulateDescriptor(), each FMP device previously retrieved Version, LowestSupportedVersion, LastAttemptVersion, and LastAttemptStatus through multiple accesses to the same FmpControllerState NV variable.

Cache the FmpControllerState data obtained from a single GetVariable() call and reuse it to populate all four fields. This eliminates redundant NV variable accesses during FMP descriptor population while preserving existing behavior.

- [ ] Breaking change?
  - No.
  - No API changes, interface changes, structure layout changes, or behavioural changes were introduced.

- [ ] Impacts security?
  - No.
  - The change only caches data that has already been read from existing firmware state storage. No authentication, cryptography, privilege, memory ownership, or buffer handling logic was modified.

- [ ] Includes tests?
  - No additional test code added.

## How This Was Tested

Tested on multiple ARM64 and x86-64 platforms.

Validated the FMP controller-state caching change against a baseline build during normal boot and firmware update scenarios.

- Verified that Version, LowestSupportedVersion, LastAttemptVersion, and LastAttemptStatus matched the baseline implementation, confirming functional equivalence.
- Confirmed the cached controller-state path was exercised during runtime.
- Successfully completed firmware update validation and verified correct update-state information after reboot.
- All FMP devices complete descriptor population successfully.
- Observed reduced descriptor population time due to elimination of redundant non-volatile state reads.
- No functional regressions were observed across repeated boot and update cycles.
- Boot time improvements in the range of 15 to 40ms across the platforms.

The change improves existing functionality by eliminating redundant accesses to firmware state data, resulting in reduced descriptor population time and measurable boot-time improvements without any observed regressions.

## Integration Instructions

N/A

### DIFF
--- a/FmpDevicePkg/FmpDxe/FmpDxe.c
+++ b/FmpDevicePkg/FmpDxe/FmpDxe.c
@@ -214,15 +214,15 @@ GetImageTypeNameString (
   2. Check if we have a variable for lowest supported version (this will be updated with each capsule applied)
   3. Check Fixed at build PCD
 
-  @param[in] Private  Pointer to the private context structure for the
-                      Firmware Management Protocol instance.
+  @param[in] FmpControllerState  The cached FMP Controller State, or NULL if the
+                                 state could not be retrieved.
 
   @retval  The largest value
 
 **/
 UINT32
 GetLowestSupportedVersion (
-  FIRMWARE_MANAGEMENT_PRIVATE_DATA  *Private
+  FMP_CONTROLLER_STATE  *FmpControllerState
   )
 {
   EFI_STATUS  Status;
@@ -259,7 +259,7 @@ GetLowestSupportedVersion (
   //
   // Check the lowest supported version UEFI variable for this device
   //
-  VariableLowestSupportedVersion = GetLowestSupportedVersionFromVariable (Private);
+  VariableLowestSupportedVersion = GetLowestSupportedVersionFromFmpControllerState (FmpControllerState);
   if (VariableLowestSupportedVersion > ReturnLsv) {
     ReturnLsv = VariableLowestSupportedVersion;
   }
@@ -283,8 +283,9 @@ PopulateDescriptor (
   FIRMWARE_MANAGEMENT_PRIVATE_DATA  *Private
   )
 {
-  EFI_STATUS  Status;
-  UINT32      DependenciesSize;
+  EFI_STATUS            Status;
+  UINT32                DependenciesSize;
+  FMP_CONTROLLER_STATE  *FmpControllerState;
 
   if (Private == NULL) {
     DEBUG ((DEBUG_ERROR, "FmpDxe(%s): PopulateDescriptor() - Private is NULL.\n", mImageIdName));
@@ -314,6 +315,8 @@ PopulateDescriptor (
   //
   GenerateFmpVariableNames (Private);
 
+  FmpControllerState = GetFmpControllerState (Private);
+
   //
   // Get the version.  Some devices don't support getting the firmware version
   // at runtime.  If FmpDeviceLib does not support returning a version, then
@@ -322,7 +325,7 @@ PopulateDescriptor (
   Status = FmpDeviceGetVersion (&Private->Descriptor.Version);
   if (Status == EFI_UNSUPPORTED) {
     Private->RuntimeVersionSupported = FALSE;
-    Private->Descriptor.Version      = GetVersionFromVariable (Private);
+    Private->Descriptor.Version      = GetVersionFromFmpControllerState (FmpControllerState);
   } else if (EFI_ERROR (Status)) {
     //
     // Unexpected error.   Use default version.
@@ -358,7 +361,7 @@ PopulateDescriptor (
                                         );
   }
 
-  Private->Descriptor.LowestSupportedImageVersion = GetLowestSupportedVersion (Private);
+  Private->Descriptor.LowestSupportedImageVersion = GetLowestSupportedVersion (FmpControllerState);
 
   //
   // Get attributes from the FmpDeviceLib
@@ -390,8 +393,12 @@ PopulateDescriptor (
     Private->Descriptor.Size = 0;
   }
 
-  Private->Descriptor.LastAttemptVersion = GetLastAttemptVersionFromVariable (Private);
-  Private->Descriptor.LastAttemptStatus  = GetLastAttemptStatusFromVariable (Private);
+  Private->Descriptor.LastAttemptVersion = GetLastAttemptVersionFromFmpControllerState (FmpControllerState);
+  Private->Descriptor.LastAttemptStatus  = GetLastAttemptStatusFromFmpControllerState (FmpControllerState);
+
+  if (FmpControllerState != NULL) {
+    FreePool (FmpControllerState);
+  }
 
   //
   // Get the dependency from the FmpDependencyDeviceLib.
--- a/FmpDevicePkg/FmpDxe/VariableSupport.c
+++ b/FmpDevicePkg/FmpDxe/VariableSupport.c
@@ -87,14 +87,14 @@ DeleteFmpVariable (
   Retrieve the FMP Controller State UEFI Variable value.  Return NULL if
   the variable does not exist or if the size of the UEFI Variable is not the
   size of FMP_CONTROLLER_STATE.  The buffer for the UEFI Variable value
-  if allocated using the UEFI Boot Service AllocatePool().
+  is allocated using the UEFI Boot Service AllocatePool().  Caller must free
+  the returned buffer with FreePool().
 
   @param[in] Private  Private context structure for the managed controller.
 
   @return  Pointer to the allocated FMP Controller State.  Returns NULL
            if the variable does not exist or is a different size than expected.
 **/
-static
 FMP_CONTROLLER_STATE *
 GetFmpControllerState (
   IN FIRMWARE_MANAGEMENT_PRIVATE_DATA  *Private
@@ -323,40 +323,31 @@ GenerateFmpVariableNames (
 /**
   Returns the value used to fill in the Version field of the
   EFI_FIRMWARE_IMAGE_DESCRIPTOR structure that is returned by the GetImageInfo()
-  service of the Firmware Management Protocol.  The value is read from a UEFI
-  variable.  If the UEFI variables does not exist, then a default version value
-  is returned.
+  service of the Firmware Management Protocol.  The value is extracted from the
+  provided FmpControllerState.  If FmpControllerState is NULL or the Version
+  field is not valid, then a default version value is returned.
 
-  UEFI Variable accessed: GUID = gEfiCallerIdGuid, Name = L"FmpState"
-
-  @param[in] Private  Private context structure for the managed controller.
+  @param[in] FmpControllerState  The cached FMP Controller State, or NULL if the
+                                 state could not be retrieved.
 
   @return  The version of the firmware image in the firmware device.
 **/
 UINT32
-GetVersionFromVariable (
-  IN FIRMWARE_MANAGEMENT_PRIVATE_DATA  *Private
+GetVersionFromFmpControllerState (
+  IN FMP_CONTROLLER_STATE  *FmpControllerState
   )
 {
-  FMP_CONTROLLER_STATE  *FmpControllerState;
-  UINT32                Value;
+  UINT32  Value;
 
-  Value              = DEFAULT_VERSION;
-  FmpControllerState = GetFmpControllerState (Private);
-  if (FmpControllerState != NULL) {
-    if (FmpControllerState->VersionValid) {
-      Value = FmpControllerState->Version;
-      DEBUG ((
-        DEBUG_INFO,
-        "FmpDxe(%s): Get variable %g %s Version %08x\n",
-        mImageIdName,
-        &gEfiCallerIdGuid,
-        Private->FmpStateVariableName,
-        Value
-        ));
-    }
-
-    FreePool (FmpControllerState);
+  Value = DEFAULT_VERSION;
+  if ((FmpControllerState != NULL) && FmpControllerState->VersionValid) {
+    Value = FmpControllerState->Version;
+    DEBUG ((
+      DEBUG_INFO,
+      "FmpDxe(%s): FMP Controller State Version %08x\n",
+      mImageIdName,
+      Value
+      ));
   }
 
   return Value;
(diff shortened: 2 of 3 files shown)


shubham-msft:
This change caches the FMP controller state obtained from a single retrieval operation and reuses it during descriptor population, eliminating redundant accesses while preserving existing behaviour. Validation details are included in the PR description.

makubacki:
@shubham-msft, was the 10-40ms improvement from platforms using the edk2 UEFI variable driver with [gEfiMdeModulePkgTokenSpaceGuid.PcdEnableVariableRuntimeCache](https://github.com/tianocore/edk2/blob/7779b1a0424aa96877b600f227c55039b2a4e10d/MdeModulePkg/MdeModulePkg.dec#L799) set to `TRUE`?

There are multiple layers to reading a UEFI variable. With that set to `TRUE`, the read does not trigger an SMI (or actually fetch from SPI flash) when cached.

FmpDevicePkg/FmpDxe/FmpDxe.c
antklein: Should the option to attempt reading from the cached variable be added to the GetLowestSupportedVersion function?

shubham-msft:
> @shubham-msft, was the 10-40ms improvement from platforms using the edk2 UEFI variable driver with [gEfiMdeModulePkgTokenSpaceGuid.PcdEnableVariableRuntimeCache](https://github.com/tianocore/edk2/blob/7779b1a0424aa96877b600f227c55039b2a4e10d/MdeModulePkg/MdeModulePkg.dec#L799) set to `TRUE`?
> 
> There are multiple layers to reading a UEFI variable. With that set to `TRUE`, the read does not trigger an SMI (or actually fetch from SPI flash) when cached.

Correcting my previous response here -
Although  `PcdEnableVariableRuntimeCache` is set to `TRUE `by default, it seems to be only getting consumed by the standard EDK2 SMM variable driver (`MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe`) likely applicable for x86 platforms. Even on x86 platforms, some improvement has been observed.
However, ARM64 platforms backed by TrustZone-based NVRAM access do not make use of the PCD `PcdEnableVariableRuntimeCache`. Hence, every `GetVariable()`call is routed through TrustZone. The optimization performed in this PR helps avoid the TrustZone accesses hence saving time in the range of 10-40ms.

FmpDevicePkg/FmpDxe/FmpDxe.c
shubham-msft: `GetLowestSupportedVersion()` which internally calls `GetLowestSupportedVersionFromVariable()` has been left untouched intentionally - same approach as the 4  *`FromVariable()`  getter functions which are also still present but no longer called from  `PopulateDescriptor()` . The design keeps all original functions unmodified and adds cached variants alongside them. This minimizes the diff against upstream and preserves the original functions for any future callers that still want to call them.

`GetLowestSupportedVersionCached()` is the newly modified function, now called from `PopulateDescriptor()`, which internally calls `GetLowestSupportedVersionFromCachedState()`.

FmpDevicePkg/FmpDxe/FmpDxe.c
```diff
@@ -314,6 +315,13 @@ PopulateDescriptor (
   //
   GenerateFmpVariableNames (Private);
 
+  //
+  // Fetch FMP Controller State once from NV storage and use it for all
+  // descriptor fields below.  This avoids redundant GetVariable2() calls
+  // hence saving few ms per call.
+  //
+  FmpControllerState = GetFmpControllerState (Private);
```
makubacki: Imo, the code pattern here is obvious, and a comment is not needed. `GetFmpControllerState()` clearly says it returns `FMP_CONTROLLER_STATE *` to a pool allocated buffer and this function references that data throughout its scope, freeing it at the end.

As already discussed in a separate conversation thread, performance is dependent on UEFI variable driver implementation, and I don't think this generic FMP code should make any assumption about platform variable performance metrics.


```suggestion
  FmpControllerState = GetFmpControllerState (Private);
```

FmpDevicePkg/FmpDxe/VariableSupport.c
```diff
@@ -320,6 +319,154 @@ GenerateFmpVariableNames (
   }
 }
 
+//
+// Cached-state variants of the getter functions.
+// These extract values from a pre-fetched FMP_CONTROLLER_STATE without
+// performing an additional NV variable read.  Used by PopulateDescriptor()
+// to avoid redundant GetVariable2() calls.
+//
+
```
makubacki: Again, I think this comment is overly verbose the function's responsibilities are/should be obvious from their documentation & signature.


```suggestion
```

FmpDevicePkg/FmpDxe/VariableSupport.c
```diff
@@ -320,6 +319,154 @@ GenerateFmpVariableNames (
...
+/**
+  Returns the Version from a pre-fetched FMP controller state.
+
+  @param[in] Private              Private context structure for the managed controller.
+  @param[in] FmpControllerState   Cached state from GetFmpControllerState().
+
+  @return  The firmware version, or DEFAULT_VERSION if field is not valid.
+**/
+UINT32
+GetVersionFromCachedState (
```
makubacki: This generic helper is named according to caller flow assumptions - that the caller has cached `FMP_CONTROLLER_STATE`. Really, that's irrelevant. It's job is to simply return the `Version` value from a given `FMP_CONTROLLER_STATE`.

This exact name doesn't have to be taken, but I suggest staying away from over assuming how the helper is used.

```suggestion
GetVersionFromFmpControllerState (
```

FmpDevicePkg/FmpDxe/VariableSupport.c
```diff
@@ -320,6 +319,154 @@ GenerateFmpVariableNames (
...
+/**
+  Returns the LowestSupportedVersion from a pre-fetched FMP controller state.
+
+  @param[in] Private              Private context structure for the managed controller.
+  @param[in] FmpControllerState   Cached state from GetFmpControllerState().
+
+  @return  The LSV, or DEFAULT_LOWESTSUPPORTEDVERSION if field is not valid.
+**/
+UINT32
+GetLowestSupportedVersionFromCachedState (
```
makubacki: Similar to the [comment on Line 338](https://github.com/tianocore/edk2/pull/12750/changes#r3537364605):

```suggestion
GetLowestSupportedVersionFromFmpControllerState (
```

FmpDevicePkg/FmpDxe/VariableSupport.c
```diff
@@ -320,6 +319,154 @@ GenerateFmpVariableNames (
...
+/**
+  Returns the LastAttemptStatus from a pre-fetched FMP controller state.
+
+  @param[in] Private              Private context structure for the managed controller.
+  @param[in] FmpControllerState   Cached state from GetFmpControllerState().
+
+  @return  The last attempt status, or DEFAULT_LASTATTEMPTSTATUS if field is not valid.
+**/
+UINT32
+GetLastAttemptStatusFromCachedState (
```
makubacki: Similar to the [comment on Line 338](https://github.com/tianocore/edk2/pull/12750/changes#r3537364605):

```suggestion
GetLastAttemptStatusFromFmpControllerState (
```

FmpDevicePkg/FmpDxe/VariableSupport.c
```diff
@@ -320,6 +319,154 @@ GenerateFmpVariableNames (
...
+/**
+  Returns the LastAttemptVersion from a pre-fetched FMP controller state.
+
+  @param[in] Private              Private context structure for the managed controller.
+  @param[in] FmpControllerState   Cached state from GetFmpControllerState().
+
+  @return  The last attempt version, or DEFAULT_LASTATTEMPTVERSION if field is not valid.
+**/
+UINT32
+GetLastAttemptVersionFromCachedState (
```
makubacki: Similar to the [comment on Line 338](https://github.com/tianocore/edk2/pull/12750/changes#r3537364605):

```suggestion
GetLastAttemptVersionFromFmpControllerState (
```

FmpDevicePkg/FmpDxe/VariableSupport.c
makubacki: These functions are fairly similar to their existing counterparts to get the information from the variable.

In this case `GetVersionFromVariable()`:

https://github.com/tianocore/edk2/blob/c0c8ca0ea0b1b6e185bfd1430640da944d3dbcd2/FmpDevicePkg/FmpDxe/VariableSupport.c#L336-L363

These APIs are internal to FmpDxe so changing the signature does not affect external code.

Did you consider updating the existing functions to take an `OPTIONAL` `FMP_CONTROLLER_STATE`? Then the function could simply read from the given `FMP_CONTROLLER_STATE` or call `GetFmpControllerState (Private)` based on whether the argument is provided.

The function could then be more generically named like `GetVersionFromFmpControllerState()`.

It just seems a little redundant right now to have core logic duplicated between two functions whose sole difference lies in how the incoming data is provided.

FmpDevicePkg/FmpDxe/FmpDxe.c
antklein: @shubham-msft, from what I can see `PopulateDescriptor()` still uses this function `GetLowestSupportedVersion()`.

https://github.com/tianocore/edk2/pull/12750/changes#diff-b7d25c5db4ca8c3c746f2865cc98140c2fc9e65b765e772ae0300f7cf7234565L361

```c
Private->Descriptor.LowestSupportedImageVersion = GetLowestSupportedVersion (Private);
```

shubham-msft: Updated the code now. Sorry that was a miss from my side after having tested the changes locally. Incorporated the suggestions from @makubacki as well.

FmpDevicePkg/FmpDxe/FmpDxe.c
shubham-msft: Made the suggested code changes

FmpDevicePkg/FmpDxe/VariableSupport.c
shubham-msft: Made the suggested code changes in all the places.

FmpDevicePkg/FmpDxe/FmpDxe.c
```diff
@@ -314,6 +318,8 @@ PopulateDescriptor (
   //
   GenerateFmpVariableNames (Private);
 
+  FmpControllerState = GetFmpControllerState (Private);
```
antklein, line 318: Is this needed as part of FmpDxe.c? Why does the caller of the Get<X>FromVariable need to manage their own cached FmpControllerState?

By moving the handling of the cached information to be explicitly internal to VariapleSupport.c it would reduce API overhead when FmpDxe interacts with VariableSupport.c.

Personally, since this optimization is not optional, I would prefer to have all of the logic handled internal to VariableSupport.c.

shubham-msft: The proposed approach keeps the cached state as an explicit, caller-owned OPTIONAL parameter rather than a module-level static inside `VariableSupport.c`. This avoids hidden global mutable state with implicit lifetime - the data flow is explicit with allocation, usage, and cleanup all happening within a single well-defined scope in `PopulateDescriptor()`. More importantly, `FmpDxe `is instantiated per FMP device, each with its own `FIRMWARE_MANAGEMENT_PRIVATE_DATA`. A static cache could silently serve stale data if one instance's cached state is accessed by another.` FmpDxe.c` treats the state as an opaque allocation - it calls `GetFmpControllerState()` to obtain it, passes it through to the getters, and calls `FreePool()` to release it, without ever inspecting the internal fields. Each getter independently handles the NULL case by falling back to its own `GetFmpControllerState()` call, so they remain fully self-contained and callable outside of `PopulateDescriptor()` without any dependency on the caching mechanism.

FmpDevicePkg/FmpDxe/FmpDxe.c
makubacki, line 1: I think this is getting close to a clean, efficient solution but there is one more thing that *looks* like it can be improved (and simplified).

---

Afacit, everything funnels through `PopulateDescriptor()` and early on that calls `GetFmpControllerState()`:

```c
  //
  // Generate UEFI Variable names used to store status information for this
  // FMP instance.
  //
  GenerateFmpVariableNames (Private);

  FmpControllerState = GetFmpControllerState (Private);
```

After that, `PopulateDescriptor()` calls:

- `GetVersionFromFmpControllerState()` (currently Line 331)
  
  ```c
      Private->Descriptor.Version      = GetVersionFromFmpControllerState (Private, FmpControllerState);
  ```

- `GetLowestSupportedVersion()` (currently Line 367)
  
  ```c
    Private->Descriptor.LowestSupportedImageVersion = GetLowestSupportedVersion (Private, FmpControllerState);
  ```
  - Which can call `GetLowestSupportedVersionFromFmpControllerState()` (currently Line 265)
  
    ```c
      //
      // Check the lowest supported version UEFI variable for this device
      //
      VariableLowestSupportedVersion = GetLowestSupportedVersionFromFmpControllerState (Private, FmpControllerState);
    ```

- `GetLastAttemptVersionFromFmpControllerState()` (currently Line 399)
  
  ```c
    Private->Descriptor.LastAttemptVersion = GetLastAttemptVersionFromFmpControllerState (Private, FmpControllerState);
  ```

- `GetLastAttemptStatusFromFmpControllerState()` (currently Line 400)
  
  ```c
    Private->Descriptor.LastAttemptStatus  = GetLastAttemptStatusFromFmpControllerState (Private, FmpControllerState);
  ```

Private->Descriptor.LastAttemptStatus  = GetLastAttemptStatusFromFmpControllerState (Private, FmpControllerState);

---

Each of the functions called out above either directly or indirectly call `GetFmpControllerState()`.

Because the `...FromFmpControllerState()` functions in VariableSupport.c all have this block at the beginning:

```c
  if (FmpControllerState == NULL) {
    FmpControllerState = GetFmpControllerState (Private);
    AllocatedState     = TRUE;
  }
```

If the variable was not present, `FmpControllerState` is `NULL` (because that is what `GetFmpControllerState()` returns when the variable is not found) so each of these keep trying to get the variable and they're all going to keep failing because it simply does not exist. Also, in the process this error message will keep getting printed:

`DEBUG ((DEBUG_ERROR, "FmpDxe(%s): Failed to get the controller state.  Status = %r\n", mImageIdName, Status));`.

So, it works to hold the value and avoid repeated reads when the variable is present, but could you just drop the optional parameter (e.g. `IN FMP_CONTROLLER_STATE              *FmpControllerState  OPTIONAL`) and that block in each function and just use the initial read to determine if it is present or not?

shubham-msft: Thanks for the suggestion. This is actually what my original implementation looked like - pure field extractors that take a pre-read  `FmpControllerState`  and return the value or a `default`, with no internal NV reads. Originally I had these as separate  `*FromCachedState ` functions alongside the existing  `*FromVariable`  functions. The  `OPTIONAL  `parameter and internal  `GetFmpControllerState()` fallback were added in a later iteration when merging both variants into a single function per the earlier review feedback. I've now reverted to the simpler original design - the key difference being that the  `*FromVariable ` functions are removed entirely rather than kept alongside. The single `GetFmpControllerState() ` call in  `PopulateDescriptor()`  determines presence once, and all getters just extract from that result. This eliminates the repeated failed reads and error messages you identified and removes the `AllocatedState `tracking, hence optimizing each getter with no change in behavior.

FmpDevicePkg/FmpDxe/VariableSupport.c
```diff
@@ -323,40 +322,35 @@ GenerateFmpVariableNames (
...
+  if ((FmpControllerState != NULL) && FmpControllerState->VersionValid) {
+    Value = FmpControllerState->Version;
+    DEBUG ((
+      DEBUG_INFO,
+      "FmpDxe(%s): Get variable %g %s Version %08x\n",
+      mImageIdName,
+      &gEfiCallerIdGuid,
+      Private->FmpStateVariableName,
+      Value
+      ));
```
makubacki, line 350: nit: Since a UEFI variable is not being read here anymore.

Also, update similar messages in other functions.

```suggestion
    DEBUG ((
      DEBUG_INFO,
      "FmpDxe(%s): FMP Controller State Version %08x\n",
      mImageIdName,
      Value
      ));
```

makubacki (approved):

FmpDevicePkg/FmpDxe/FmpDxe.c
makubacki: Thanks. The update looks good.

FmpDevicePkg/FmpDxe/VariableSupport.c
makubacki: To match the description in VariableSupport.h:

```
  Retrieve the FMP Controller State UEFI Variable value.  Return NULL if
  the variable does not exist or if the size of the UEFI Variable is not the
  size of FMP_CONTROLLER_STATE.  The buffer for the UEFI Variable value
  is allocated using the UEFI Boot Service AllocatePool().  Caller must free
  the returned buffer with FreePool().
```

FmpDevicePkg/FmpDxe/VariableSupport.c
shubham-msft: With the updated DEBUG messages no longer referencing ` Private->FmpStateVariableName`  or ` &gEfiCallerIdGuid`, the  `Private  `parameter is now unused in all four getter functions ( `GetVersionFromFmpControllerState `,  `GetLowestSupportedVersionFromFmpControllerState `,  `GetLastAttemptStatusFromFmpControllerState `,  `GetLastAttemptVersionFromFmpControllerState `). do you recommend removing the  `Private ` parameter from these functions, or keep it for potential future use?

makubacki: I suggest dropping it.

FmpDevicePkg/FmpDxe/VariableSupport.c
shubham-msft: Made the suggested changes.

FmpDevicePkg/FmpDxe/VariableSupport.c
shubham-msft: Made the suggested changes.

antklein (approved):

shubham-msft:
Hi @makubacki , @lgao4 , @mdkinney , looking for workflow approval to merge the PR.

makubacki:
> Hi @makubacki , @lgao4 , @mdkinney , looking for workflow approval to merge the PR.

You ideally need a PR approval from @lgao4 or @mdkinney. I was happy to review the PR but they are FmpDevicePkg maintainers and I am not, so it would be nice from them to approve modifications to the package. @lgao4 & @mdkinney, if you'd like to just proceed with my review, let me know.

lgao4:
> > Hi @makubacki , @lgao4 , @mdkinney , looking for workflow approval to merge the PR.
> 
> You ideally need a PR approval from @lgao4 or @mdkinney. I was happy to review the PR but they are FmpDevicePkg maintainers and I am not, so it would be nice from them to approve modifications to the package. @lgao4 & @mdkinney, if you'd like to just proceed with my review, let me know.

Michael: You are familiar with this module. Can you work as this package maintainer?

lgao4 (approved):

makubacki:
> Michael: You are familiar with this module. Can you work as this package maintainer?

Thanks for the suggestion @lgao4. I'll consider if I have enough time for it.

shubham-msft:
@mergifyio queue

PR #12755, opened 2026-07-02
ShellPkg/SmbiosView: Improve SMBIOS Type 42 decoding

xunfengd:
Update smbiosview to decode Management Controller Host Interface (Type 42) records more clearly and robustly.

The MC host interface type table only reported MCTP host interfaces as a generic range. Add the MCTP host interface subtype values defined in MdePkg, including PCIe VDM and MMBI, so Type 42 records can be identified from smbiosview output.

Also improve Type 42 formatted-area parsing by using unaligned reads for InterfaceTypeSpecificData, validating the formatted structure length before reading protocol records, and advancing each protocol record by its full header and data length.

This prevents protocol record mis-parsing when ProtocolTypeDataLen is zero, such as for IPMI-over-MMBI Type 42 records.

Change-Id: I10c974560ff2a5385d159494a5b285569003a061

# Description

Update `smbiosview` to decode Management Controller Host Interface (Type 42) records more clearly and robustly.

The MC host interface type table previously reported MCTP host interfaces as a generic range. This change adds the MCTP host interface subtype values defined in `MdePkg`, including PCIe VDM and MMBI, so Type 42 records can be identified directly from `smbiosview` output.

This also improves Type 42 formatted-area parsing by using unaligned reads for `InterfaceTypeSpecificData`, validating the formatted structure length before reading protocol records, and advancing each protocol record by its full header and data length. This prevents protocol record mis-parsing when `ProtocolTypeDataLen` is zero, such as for IPMI-over-MMBI Type 42 records.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Verified that the modified `SmbiosView` source files report no editor diagnostics.

Reviewed `smbiosview` output from an SMBIOS Type 42 record containing an IPMI protocol record with `ProtocolTypeDataLen` set to zero. Confirmed the parser can identify the Management Controller Host Interface record and walk the protocol record without mis-parsing subsequent data.

## Integration Instructions

N/A

### DIFF
--- a/ShellPkg/Library/UefiShellDebug1CommandsLib/SmbiosView/PrintInfo.c
+++ b/ShellPkg/Library/UefiShellDebug1CommandsLib/SmbiosView/PrintInfo.c
@@ -1285,15 +1285,24 @@ SmbiosPrintStructure (
     {
       MC_HOST_INTERFACE_PROTOCOL_RECORD  *MCHostInterfaceProtocolRecord;
       UINT8                              *RecordsPointer;
+      UINT8                              *RecordsEnd;
       UINT8                              MCHostInterfaceProtocolNumber;
       UINT8                              ProtocolTypeDataLen;
+      UINT8                              InterfaceTypeSpecificDataLength;
+      UINTN                              RecordLength;
 
       DisplayMCHostInterfaceType (Struct->Type42->InterfaceType, Option);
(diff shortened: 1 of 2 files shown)


pierregondois:
Hello Liangqi,
Is this a draft PR, or is the intent to get the PR merged ?

PR #12756, opened 2026-07-03
MdePkg/BaseIoLibIntrinsic: Add MMIO-backed I/O port fallback for LoongArch

MarsDoge:
# Description

Add a LoongArch64 IoLib implementation that maps legacy I/O port accesses
to the platform-provided MMIO translation window described by
`PcdPciIoTranslation`.

On LoongArch64, `BaseIoLibIntrinsic` currently uses `IoLibNoIo.c`, so legacy
I/O port accesses assert. This is problematic for UEFI drivers that still use
`IoRead*()` / `IoWrite*()` for legacy I/O port devices such as PS/2
controllers.

This change selects a new `IoLibPortMmio.c` source for LoongArch64. If
`PcdPciIoTranslation` is non-zero, legacy I/O port accesses are translated
to MMIO and forwarded through the existing MMIO helpers. If the PCD is zero,
the code keeps the `IoLibNoIo.c`-compatible behavior: assert and return a
harmless value without touching MMIO address 0.

Dynamic PCD access is used because LoongArchVirt discovers the I/O translation
window from FDT and updates `PcdPciIoTranslation` dynamically. The `PcdLib`
dependency is scoped to `LOONGARCH64` so other `BaseIoLibIntrinsic`
architecture variants do not gain an unused dependency.

This keeps the translation inside the IoLib implementation instead of:
- making UEFI drivers depend on PI `EFI_CPU_IO2_PROTOCOL`, or
- duplicating the translation logic in each legacy I/O consumer.

Related previous discussions:
- https://github.com/tianocore/edk2/pull/11414
- https://github.com/tianocore/edk2/pull/11550

# How This Was Tested

- `BaseTools/Scripts/PatchCheck.py <commit>`: pass
- `build -p OvmfPkg/LoongArchVirt/LoongArchVirtQemu.dsc -a LOONGARCH64 -t GCC -b DEBUG`: pass
- Verified on LoongArch hardware with an `IoLib`-based legacy I/O driver;
  the driver can access the target I/O device normally through the
  MMIO-backed IoLib path.

### DIFF
--- a/MdePkg/Library/BaseIoLibIntrinsic/BaseIoLibIntrinsic.inf
+++ b/MdePkg/Library/BaseIoLibIntrinsic/BaseIoLibIntrinsic.inf
@@ -55,7 +55,7 @@
   IoLibNoIo.c
 
 [Sources.LOONGARCH64]
-  IoLibNoIo.c
+  IoLibPortMmio.c
 
 [Packages]
   MdePkg/MdePkg.dec
@@ -64,4 +64,11 @@
   DebugLib
   BaseLib
   RegisterFilterLib
+
+[LibraryClasses.LOONGARCH64]
+  PcdLib
+
+
+[Pcd.LOONGARCH64]
+  gEfiMdePkgTokenSpaceGuid.PcdPciIoTranslation
 
--- a/MdePkg/Library/BaseIoLibIntrinsic/IoLibPortMmio.c
+++ b/MdePkg/Library/BaseIoLibIntrinsic/IoLibPortMmio.c
@@ -1,0 +1,806 @@
+/** @file
+  I/O library for platforms that map legacy I/O port accesses to MMIO.
+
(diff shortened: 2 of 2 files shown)


MarsDoge:
Hi @mdkinney,

Could you please help review this PR when you have time?

This patch follows the direction discussed in #11414 and #11550: avoid making
UEFI drivers depend on the PI CPU I/O protocol, avoid duplicating
`PcdPciIoTranslation` handling in individual drivers, and keep the translation
inside an IoLib implementation.

The updated version adds a LoongArch64 path in `BaseIoLibIntrinsic`, uses
dynamic PCD access for `PcdPciIoTranslation`, and scopes the `PcdLib`
dependency to `LOONGARCH64` only.

PatchCheck, LoongArchVirt full build, and LoongArch hardware validation with an
`IoLib`-based legacy I/O driver have passed.

Please let me know whether this IoLib-based direction looks acceptable for
MdePkg, following the previous discussions in #11414 and #11550.

PR #12760, opened 2026-07-03
MdeModulePkg/ScsiDiskDxe: Don't fail media detection on FUA query error

MrChromebox:
# Description

Commit ee58614bbb ("MdeModulePkg/ScsiDiskDxe: Check Write Caching and FUA support") added a best-effort FUA-support query to ScsiDiskDetectMedia(). It issues a MODE SENSE(10) for the Caching mode page via ScsiDiskFuaMode() and assigns the result to the function's shared Status variable.

When the device rejects that command, Status is left as EFI_DEVICE_ERROR and falls through to the function's return, even though the media was already detected successfully (capacity read, media present). Callers treat this as fatal:
ScsiDiskDriverBindingStart() skips installing Block I/O, and ScsiDiskReadBlocks()/WriteBlocks() return EFI_DEVICE_ERROR. For non-fixed devices DetectMedia() re-runs on every access, so the disk is never usable.

This is observed on UFS: the LUN answers the MODE SENSE with a Target Failure ("UfsExecScsiCmds() fails with Target Failure"), which previously worked (edk2-stable202602) now breaks UFS drive detection (edk2-stable202605).

The FUA query is only used to refine ScsiDiskDevice->FuaMode, which already defaults to TRUE (the prior always-set-FUA behavior). Its failure must not affect media detection, so stop assigning its result to Status and hoist the device-type check out of the retry loop.

Fixes: ee58614bbb ("MdeModulePkg/ScsiDiskDxe: Check Write Caching and FUA support")

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

TEST=build/boot google/yaviks Chromebook with UFS, drive is detected and bootable.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Bus/Scsi/ScsiDiskDxe/ScsiDisk.c
+++ b/MdeModulePkg/Bus/Scsi/ScsiDiskDxe/ScsiDisk.c
@@ -2541,10 +2541,9 @@ ScsiDiskDetectMedia (
   //
   // Get FUA Mode
   //
-  for (Retry = 0; Retry < MaxRetry; Retry++) {
-    if (ScsiDiskDevice->DeviceType == EFI_SCSI_TYPE_DISK) {
-      Status = ScsiDiskFuaMode (ScsiDiskDevice);
-      if (!EFI_ERROR (Status)) {
+  if (ScsiDiskDevice->DeviceType == EFI_SCSI_TYPE_DISK) {
+    for (Retry = 0; Retry < MaxRetry; Retry++) {
+      if (!EFI_ERROR (ScsiDiskFuaMode (ScsiDiskDevice))) {
         break;
       }
     }


os-d:
@MrChromebox my question on this one is should `FuaMode` default to `FALSE`? All SCSI devices support FUA off, IIUC, so that would seem a safer default if the device doesn't respond as expected to the mode sense cmd.

In your case is the device giving back a check condition with illegal request because it doesn't support that mode sense 10 page? Or is the device actually choking in some way? If the device just doesn't support the page, then I think this approach is okay. If the device is choking on it, it may be questionable to use the device at all.

@annie-li do you have thoughts on this? You made the original change here. @jkolakow @lgao4 thoughts?

MrChromebox:
@os-d AIUI, the issue is (many? most? all?) UFS devices, which use the SCSI subsystem, not supporting the mode 10 page

os-d:
> @os-d AIUI, the issue is (many? most? all?) UFS devices, which use the SCSI subsystem, not supporting the mode 10 page

Then supporting that makes sense to me, again my question is just whether we should default FUA mode to false, since we can't tell if the device supports it or not.

lgao4:
> > @os-d AIUI, the issue is (many? most? all?) UFS devices, which use the SCSI subsystem, not supporting the mode 10 page
> 
> Then supporting that makes sense to me, again my question is just whether we should default FUA mode to false, since we can't tell if the device supports it or not.

Seemly, this is a separate issue about the default policy.

MrChromebox:
is it preferable that I drop this and submit a new PR which changes the default policy?

PR #12761, opened 2026-07-03
FatPkg: Fix FAT32 mount on small 4K-sector ESP partitions

MrChromebox:
# Description

Trust the FAT32 BPB instead of requiring MaxCluster >= 65525, which rejected valid UFS EFS volumes and blocked internal boot.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?
 
## How This Was Tested

TEST=build/boot google/joxer with 200MB EFS (51024 clusters)

## Integration Instructions

N/A

### DIFF
--- a/FatPkg/EnhancedFatDxe/Init.c
+++ b/FatPkg/EnhancedFatDxe/Init.c
@@ -390,7 +390,12 @@ FatOpenDevice (
     Volume->FatEntrySize = sizeof (UINT16);
     DirtyMask            = FAT16_DIRTY_MASK;
   } else {
-    if (Volume->MaxCluster < FAT_MAX_FAT16_CLUSTER) {
+    //
+    // FAT32 layout comes from the BPB (32-bit FAT).  Small volumes such as a
+    // ~200 MiB ESP with 4 KB sectors can be valid FAT32 while MaxCluster is
+    // still below FAT_MAX_FAT16_CLUSTER; mkfs.fat and Linux use this layout.
+    //
+    if (Volume->MaxCluster < FAT_MAX_FAT12_CLUSTER) {
       return EFI_VOLUME_CORRUPTED;
     }
 


FatPkg/EnhancedFatDxe/Init.c
```diff
@@ -390,7 +390,12 @@ FatOpenDevice (
     Volume->FatEntrySize = sizeof (UINT16);
     DirtyMask            = FAT16_DIRTY_MASK;
   } else {
-    if (Volume->MaxCluster < FAT_MAX_FAT16_CLUSTER) {
+    //
+    // FAT32 layout comes from the BPB (32-bit FAT).  Small volumes such as a
+    // ~200 MiB ESP with 4 KB sectors can be valid FAT32 while MaxCluster is
+    // still below FAT_MAX_FAT16_CLUSTER; mkfs.fat and Linux use this layout.
+    //
+    if (Volume->MaxCluster < FAT_MAX_FAT12_CLUSTER) {
```
os-d, line 398: The FAT spec is an odd document because it is ancient, there are a couple different versions that circulate and Microsoft doesn't host an official spec anymore. But, this is listed:

Image

https://www.cs.fsu.edu/~cop4610t/assignments/project3/spec/fatspec.pdf.

Other versions that are slightly newer (2005) still have the same calculation but have dropped the one and only one statement. However, I think this is an expectation that is pervasive in the spec. If the volume size shrinks for a FAT32 volume, then the sectors per cluster is expected to also shrink to keep the total cluster size higher than 65,525.

So, from my reading of this, the mkfs.fat utility has the wrong behavior and either parameters need to be adjusted or the tool needs to be updated. Do you agree?

MrChromebox: @os-d yes, I agree with your reading. I'm not sure if this was a user-created partition, or distro installer created one. But either way, mkfs.fat shouldn't be creating out-of-spec partitions

